PR #286, opened 2021-12-03
Make the fonts compiled into flutter accessible to svelte

bitdivine:
# Motivation
The svelte app should use the same fonts as flutter.  However the flutter fonts don't appear to be available from download from the canister, they are probably compiled into the flutter app, so we need a separate copy for svelte.  Eventually the flutter app will be gone anyway so this copy will be the only copy remaining.

# Changes
- Copied the flutter fonts into the svelte codebase.
- Provided the urls to the fonts in the global.css, so that they can be used.


ielashi:
I am wondering if we shouldn't use these fonts at all. They're non-standard, take up a lot of space and take up a lot of space. Can we not get away with standard fonts to keep the app snappy? Sure, it'll look slightly different, but I think that's acceptable. WDYT?

bitdivine:
@ielashi  Dom is keen on these fonts.  You would need to clear any change with him.  Also, note that when we have different paths served by flutter/svelte, when moving between tabs the top part of the page should look the same in both and the difference in font, although slight, is quite noticeable.

ielashi:
@bitdivine A few things to keep in mind:

* Some of these fonts (e.g. Roboto) are very rarely used, if it all. Maybe worth only pulling fonts in if we know for certain that they need to be used?
* Having ~500kb of fonts doesn't sound like a good practice. Do we really need all these variations of the same font? If having that font is really a strict requirement, I'd advise we choose one of these variations, use it for the title and tabs (if that's what Dom cares about), and use standard fonts elsewhere.
* We don't have enough space to deploy all these fonts within the canister. There's a workaround that we already have for this, but that will be additional work in the way of getting v0 out the door.

We should be avoiding repeating the same mistakes that took us down to where we are now.

ielashi (approved):

bitdivine:
@ielashi Good points.  Thank you.  Knowing the background helps.  To reduce size for a given font set, some options are:
- Make a separate assets canister
- Reuse the files already used by flutter (I decided not to in an effort to keep the two codebases separate but that may not be the right decision)
- Compress the fonts; I have not tried serving compressed data from a canister but it should be possible.

Note: Personally website load speed is near the top of my list of priorities when it comes to aesthetics but I don't believe that we can avoid circular.  Roboto we can probably skip (I have not heard any strong argument in favour of keeping it whereas Dom and Jan have both mentioned circular).  I'll experiment with using just one circular family member and varying font-weight.

ielashi:
We definitely don't need an assets canister. We already have what's called "stable assets" - assets that we can upload separately that don't included in the canister's binary. We used for a [few things](https://github.com/dfinity/nns-dapp/blob/main/rs/src/main.rs#L307), including fonts.

PR #289, opened 2021-12-07
The votes cast on a proposal are always displayed now

anishh2003:
For each proposal , the user can now always see the votes cast

when user has neurons with voting power: 
Screenshot 2021-12-07 at 15 07 08

Screenshot 2021-12-07 at 15 07 26

when user has no neurons /neurons with no voting power: 
Screenshot 2021-12-09 at 13 17 49


hpeebles:
Can you add a screenshot of how it looks when you have no neurons?

anishh2003:
> Can you add a screenshot of how it looks when you have no neurons?

hpeebles (approved):

PR #292, opened 2021-12-09
Sync svelte login and logout

bitdivine:
# Motivation
If a user logs in on one tab, they should be logged in on all tabs, likewise if a user logs out on one tab, they should be logged out on all.  There is a security ticket regarding logging out:  https://dfinity.atlassian.net/browse/FOLLOW-229

# Changes
- Create a broadcast mechanism - this is a channel shared between tabs on the same origin that can be used for them to communicate.
- If one tab logs out, it broadcasts the fact, others listen to the channel and log out as well.
- If one tab logs in, the other tabs initialize their login status; assuming the login is real they will also log in.

# Testing
- Start a server with `frontend/svelte$ nns run dev`
- Open three tabs
- Log in to any one.
- Verify that all three tabs log in
- Log out of any one
- Verify that all three tabs log out

Tested on Chrome and Firefox.  Simulated Safari and IE by setting the check for BroadcastChannel to false.


frontend/svelte/src/Auth.svelte
```diff
@@ -10,16 +11,20 @@
   let authClient;
   let identityProvider = process.env.INTERNET_IDENTITY_URL;
```
bitdivine, line 13: Change:  `initAuth()` checked for transitions from not logged in, on ititial page load, to logged in, if the user was already authenticated.  The code is now extended to handle the case where the user was logged in but is now logged out.  This can happen if the user logged out in a different tab.

frontend/svelte/src/Auth.svelte
```diff
@@ -10,16 +11,26 @@
...
+      JSON.stringify({ checkAuth: { wasSignedIn, isAuthenticated } })
+    );
+    if (wasSignedIn !== isAuthenticated) {
+      if (isAuthenticated) {
+        onSignIn();
+      } else {
+        signOut();
+      }
     }
   };
```
bitdivine, line 27: AuthSync triggers checkAuth if the user logged in or out in another tab.

frontend/svelte/src/Auth.svelte
```diff
@@ -31,11 +42,12 @@
         onError: reject,
       });
     });
-    onSignedIn();
+    onSignIn();
+    authSync.onSignIn();
   };
```
bitdivine, line 45: Consistent naming.

frontend/svelte/src/Auth.svelte
```diff
@@ -10,16 +11,23 @@
   let authClient;
   let identityProvider = process.env.INTERNET_IDENTITY_URL;
 
-  // Sets initial login status
-  const initAuth = async () => {
+  // Check for any change in authentication status and act upon it.
+  const checkAuth = async () => {
+    const wasSignedIn = signedIn;
```
nmattia, line 16: I'm a bit confused, what is the difference between being "signed in" and "authenticated"?

frontend/svelte/src/AuthSync.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Creates a login sync.
+   *
+   * @param {Function} callback - invoked whenever login status may have changed in another tab.
+   */
+  constructor(callback: Function) {
+    if (typeof BroadcastChannel === "undefined") {
+      // Safari, IE.  (Safari is expected to have BroadcastChannels soon.)
+      setInterval(callback, AuthSync.INTERVAL);
```
nmattia, line 18: If I understand correctly, if there is no broadcast channel, the callback is called every 10s?

Might be a silly question, but is there anything clearing the interval when/if the AuthSync is destroyed?

frontend/svelte/src/AuthSync.ts
```diff
@@ -0,0 +1,45 @@
...
+    } else {
+      // Chrome, Firefox, Edge.
+      this.broadcastChannel = new BroadcastChannel("nns-auth");
+      this.broadcastChannel.onmessage = function (event) {
```
nmattia, line 22: `event` is not used here. Why do we differentiate between `SIGN_IN` and `SIGN_OUT`?

nmattia (commented):
A few questions!

frontend/svelte/src/AuthSync.ts
bitdivine: Not a silly question, the auth sync is never destroyed, it checks forever.  However a check doesn't involve an https request, it checks local shared data.  It is precisely the poll we were referring to earlier.  You can verify that there are no http requests by looking in the network tab.

Note: I have not dug in to find out why, but a delegation lasts only 30 minutes by default, but after 30 minutes the `authClient.isAuthenticated()` still returns true.  This sounds like a bug in authClient to me, however I have not dived into the code to find out exactly what is happening.

frontend/svelte/src/Auth.svelte
bitdivine: None, at a high conceptual level.  At a low level, `signedIn` is the variable used to control whether svelte shows the login page whereas `.isAuthenticated()` is an AuthClient API method used to set `signedIn`.

frontend/svelte/src/AuthSync.ts
bitdivine: It helps debugging.  If you listen to the channel in a console while you have five tabs open and are logging in and out and you are wondering "Was the last logout broadcast?" it helps to see a history of "login, logout, login" rather than "change,change,change".

frontend/svelte/src/AuthSync.ts
bitdivine: I have a separate PR with logout if the user browses to another tab and stays on it, based on the visibility API, however I would like to get to the bottom of this "is authenticated even after the delegation has expired" behaviour.

frontend/svelte/src/AuthSync.ts
nmattia: Ok, I think this is really worth clarifying in the code. I've spent a few minutes trying to figure out where those values were being read!

nmattia (approved):
LGTM!

PR #293, opened 2021-12-10
Add svelte to canister

bitdivine:
# Motivation
The svelte app is to be run alongside the main app.  We have decided to host it in the same canister as the existing app, but under a different path.

# Changes
- Update the build script to include the svelte app.
  - The svelte app is under the path /v2, e.g. https://txssk-maaaa-aaaaa-aaanq-cai.nnsdapp.dfinity.network/v2/index.html
  - The svelte app is not used or referenced by the flutter app, so it can be visited by someone who knows it is there, but it will not be encountered as part of normal dapp usage.
- Update the svelte index.html to use relative paths, as it being served from /v2 rather than the root.
  - An alternative would be to change the html root, however we will want to refer to the flutter app so that is not desirable.

# Testing
- This code has been deployed to: https://txssk-maaaa-aaaaa-aaanq-cai.nnsdapp.dfinity.network/#/loading
- The svelte app is not used but can be accessed under /v2: https://txssk-maaaa-aaaaa-aaanq-cai.nnsdapp.dfinity.network/v2/index.html

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,26 +13,40 @@ set -x
 (cd frontend/ts && ./build.sh)
 
 # build the flutter app
-cd frontend/dart || exit
-if [[ $DEPLOY_ENV = "mainnet" ]]; then
-  flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=FLUTTER_WEB_CANVASKIT_URL=/assets/canvaskit/
-else
-  # For all networks that are not main net, build with the staging config
-  flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=DEPLOY_ENV=staging
-fi
-sed -i -e 's/flutter_service_worker.js?v=[0-9]*/flutter_service_worker.js/' build/web/index.html
+pushd frontend/dart || exit
+  if [[ $DEPLOY_ENV = "mainnet" ]]; then
+    flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=FLUTTER_WEB_CANVASKIT_URL=/assets/canvaskit/
+  else
+    # For all networks that are not main net, build with the staging config
+    flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=DEPLOY_ENV=staging
+  fi
+  sed -i -e 's/flutter_service_worker.js?v=[0-9]*/flutter_service_worker.js/' build/web/index.html
+popd || exit
+
+rm -f assets.tar.xz
+rm -fr web-assets
+cp -R frontend/dart/build/web/ web-assets
+
+# Build the svelte app
+pushd frontend/svelte || exit
+  npm ci
+  npm run build
+popd || exit
+
+rm -fr web-assets/v2
+cp -R frontend/svelte/public web-assets/v2
 
 # Bundle into a tight tarball
 # On macOS you need to install gtar + xz
 # brew install gnu-tar
 # brew install xz
+pushd web-assets || exit
+  # Remove the assets/NOTICES file, as it's large in size and not used.
+  rm assets/NOTICES
+  tar cJv --mtime='2021-05-07 17:00+00' --sort=name --exclude .last_build_id -f ../assets.tar.xz . || \
+  gtar cJv --mtime='2021-05-07 17:00+00' --sort=name --exclude .last_build_id -f ../assets.tar.xz .
+popd || exit
 
-cd build/web/ || exit
-# Remove the assets/NOTICES file, as it's large in size and not used.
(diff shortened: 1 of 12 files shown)


build.sh
```diff
@@ -13,26 +13,40 @@ set -x
 (cd frontend/ts && ./build.sh)
 
 # build the flutter app
```
bitdivine, line 15: This first part - building is the same as before, just with pushd/popd.

build.sh
```diff
@@ -13,26 +13,40 @@ set -x
...
-fi
-sed -i -e 's/flutter_service_worker.js?v=[0-9]*/flutter_service_worker.js/' build/web/index.html
+pushd frontend/dart || exit
+  if [[ $DEPLOY_ENV = "mainnet" ]]; then
+    flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=FLUTTER_WEB_CANVASKIT_URL=/assets/canvaskit/
+  else
+    # For all networks that are not main net, build with the staging config
+    flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=DEPLOY_ENV=staging
+  fi
+  sed -i -e 's/flutter_service_worker.js?v=[0-9]*/flutter_service_worker.js/' build/web/index.html
```
bitdivine: Now, instead of bundling (we have to wait for that until the svelte has been built) we pop out and copy the build artefacts to a common build dir.

build.sh
```diff
@@ -13,26 +13,40 @@ set -x
...
+popd || exit
+
+rm -f assets.tar.xz
+rm -fr web-assets
+cp -R frontend/dart/build/web/ web-assets
+
```
bitdivine, line 29: Now we build the svelte and add it to the common build dir

build.sh
```diff
@@ -13,26 +13,40 @@ set -x
...
+# Build the svelte app
+pushd frontend/svelte || exit
+  npm ci
+  npm run build
+popd || exit
+
+rm -fr web-assets/v2
+cp -R frontend/svelte/public web-assets/v2
```
bitdivine, line 38: Finally we can bundle, but we bundle the common dir rather than the dir inside frontend/dart.

frontend/svelte/public/index.html
```diff
@@ -11,10 +11,10 @@
...
     />
```
bitdivine, line 13: As the svelte app is being served from /v2 these paths need to be relative.

nmattia (approved):
Same comment as before about `pushd`/`popd` (I'd rather have `cd $ABSOLUTE_PATH`, because right now one lone `cd foo` will break everything) but other than that LGTM

frontend/svelte/public/index.html
```diff
@@ -11,10 +11,10 @@
...
 
-    <link rel="stylesheet" href="/global.css" />
-    <link rel="stylesheet" href="/build/bundle.css" />
+    <link rel="stylesheet" href="./global.css" />
+    <link rel="stylesheet" href="./build/bundle.css" />
```
nmattia, line 15: out of curiosity, where does this `build/` come from?

ielashi:
Questions:

1. Did we try deploying this to a testnet to ensure that we're still below the canister size limit?
2. We should probably add something like [this](https://github.com/dfinity/nns-dapp/blob/main/rs/src/assets.rs#L197) for the `/v2` endpoint, so that visit `/v2` will load the `index.html` and not return a 404.

frontend/svelte/public/index.html
bitdivine: From svelte.  In `frontend/svelte/public/` the `index.html` and `global.css` are [tracked in git](https://github.com/dfinity/nns-dapp/tree/main/frontend/svelte/public) and can be adjusted manually, whereas `build` is compiled from the .svelte files.  See the [.gitignore in `frontend/svelte`](https://github.com/dfinity/nns-dapp/blob/main/frontend/svelte/.gitignore).

bitdivine:
@ielashi  - thank you for the review.
- Yes, the code has been deployed to a testnet, see the links in the "Testing" section of the description.
- The index handling is a good idea.  I was not sure that we would need it but it is a tad cleaner.

ielashi (approved):
@bitdivine Ah, sorry I am used to people not writing good PR descriptions, so I am used to not looking at it. Thanks for the clarity :)

bitdivine:
@ielashi   Done.  I support `/v2/` but not `/v2` as if I serve `/v2` , resources such as `global.css` which are under `/v2/global.css` are instead sought under `/global.css`.  I.e. the browser does not know that `/v2` is a directory and uses the parent as the directory in which to look for relative resources.  We could add a redirect, but I don't expect any normal user to visit either of these URLs manually.  They are just for aesthetics.

PR #297, opened 2021-12-14
Restyle svelte login page

bitdivine:
# Motivation
The NNS login page has a new background image, text and styling.

# Changes
- Add new background image to svelte
- Set new text
- Adjust styling slightly towards the new blueprint
  - Note: I am awaiting exact colour codes from the artist.  The current implementation demos the different styles that appear to be present in the design - sudden colour changes, colour gradients and a very gentle washing effect, from grey to grey-yellow back to shades of grey, almost invisible but also obviously not constant.

# Test
The svelte app is deployed here: https://wxns6-qiaaa-aaaaa-aaaqa-cai.nnsdapp.dfinity.network/v2/


frontend/svelte/src/Auth.svelte
```diff
@@ -117,18 +119,46 @@
     width: 100vw;
     height: 100vh;
     z-index: 100;
-    background-color: var(--background-grey);
+    background-color: black;
+    background-image: url("/assets/assets/nns_background.jpeg");
```
ielashi, line 123: I think you also need:

```
 background-size: cover;
 background-repeat: no-repeat;
```

For the image to stretch and not repeat itself.

ielashi (approved):

nmattia (approved):
Looking nice! The button now looks relatively big but I can live with that

PR #300, opened 2021-12-14
feat: update agent-js from `0.10.1` to `0.10.2`

Daniel-Bloom-dfinity:
`agent-js` (as of `0.10.2`) avoids `ic0.app` subdomains, and thus 307 redirects.


ielashi:
This is only updating the svelte code. Should we updating flutter as well?

Daniel-Bloom-dfinity:
> Should we updating flutter as well?

Yes, but that's a more complicated upgrade since it's going from `0.9.1` to `0.10.2`, so I'm putting together a separate PR for that.

ielashi (approved):

PR #301, opened 2021-12-15
Get the ICP to cycles conversion rate from the CMC

hpeebles:
This PR makes the NNS Dapp frontend get the ICP to cycles conversion rate from the cycles minting canister rather than from the registry.


nzoghb (approved):

hpeebles:
Not sure why the clippy check wouldn't run, but I've run `cargo clippy` locally and everything passes.

PR #305, opened 2021-12-17
Deduplicate build.sh

bitdivine:
# Motivation
The dockerfile and build.sh duplicated code, fell out of sync and eventually caused a deploy to fail.  Therefore we wish to:

- call build.sh in the dockerfile rather than duplicating the code
- be able to deploy docker builds to the testnet, so that we can verify that docker builds work

# Changes
- Call build.sh in the dockerfile instead of replicating the steps.
- Take the target network as an env var.
- Cleanup and fix until testnet deployments work.

# Testing

## Differences observed

- A difference:  Before assets/NOTICES was included in the docker build.  In build .shit is removed with:
```
# Remove the assets/NOTICES file, as it's large in size and not used.
rm assets/NOTICES
```
That might explain why test deployments have more space than prod....

If the NOTICES file is left in, new docker builds are identical to the old.  But I will remove the NOTICES file as that seems to be the intent.  The hash was compared by checking out:

```
375bd498c56953cf4d98472956b1056e52bf306a = last commit before svelte.
```
building, then cherry picking this PR:
```
541e3f1803e288d09982aa4a95a509b2589db07f = build commit
```
and building again.  The build differed but commenting out the NOTICES line in build.sh gave the same hash.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -27,3 +27,7 @@ frontend/dart/ios/Runner/GeneratedPluginRegistrant.m
 frontend/dart/lib/generated_plugin_registrant.dart
 frontend/dart/pubspec.lock
 frontend/dart/build/web
+frontend/svelte/node_modules/
+frontend/svelte/public/build/
+frontend/svelte/public/assets/
+**/*.wasm
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+**/*.wasm
 assets.tar.xz
 out
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 # docker build -t nns-dapp .
 # docker run --rm --entrypoint cat nns-dapp /nns-dapp.wasm > nns-dapp.wasm
 
-FROM ubuntu:20.10
+FROM ubuntu:20.10 as builder
 SHELL ["bash", "-c"]
 
 ARG rust_version=1.54.0
@@ -15,7 +15,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
     apt -yq update && \
     apt -yqq install --no-install-recommends curl ca-certificates \
         build-essential pkg-config libssl-dev llvm-dev liblmdb-dev clang cmake \
-	git
+	git jq
 
 # Install node
 RUN curl --fail -sSf https://raw.githubusercontent.com/creationix/nvm/v0.34.0/install.sh | bash
@@ -46,20 +46,20 @@ ENV PATH=/flutter/bin:$PATH
 # Install IC CDK optimizer
 RUN cargo install --version 0.3.1 ic-cdk-optimizer
 
+# Install dfx
+COPY dfx.json dfx.json
+RUN DFX_VERSION="$(jq -cr .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
+
+# Start the second container
+FROM builder
+SHELL ["bash", "-c"]
+ARG DEPLOY_ENV=mainnet
+RUN echo $DEPLOY_ENV
+
+# Build
 COPY . .
+RUN ./build.sh
 
-ENV DEPLOY_ENV=mainnet
-RUN cd frontend/ts && ./build.sh
-
-RUN cd frontend/dart && flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=FLUTTER_WEB_CANVASKIT_URL=/assets/canvaskit/
-# Remove random hash from flutter output
-RUN sed -i -e 's/flutter_service_worker.js?v=[0-9]*/flutter_service_worker.js/' frontend/dart/build/web/index.html
-
-# Bundle into a tight tarball
-RUN cd frontend/dart/build/web/ && tar cJv --mtime='2021-05-07 17:00+00' --sort=name --exclude .last_build_id -f ../../../../assets.tar.xz .
-RUN ls -sh assets.tar.xz; sha256sum assets.tar.xz
-
-RUN cargo build --locked --target wasm32-unknown-unknown --release --package nns-dapp
-RUN cp target/wasm32-unknown-unknown/release/nns-dapp.wasm .
-RUN ic-cdk-optimizer nns-dapp.wasm -o nns-dapp.wasm
+# Copy the wasm to the traditional location.
+RUN cp "$(jq -rc '.canisters["nns-dapp"].wasm' dfx.json)" nns-dapp.wasm
 RUN ls -sh nns-dapp.wasm; sha256sum nns-dapp.wasm
--- a/build.sh
+++ b/build.sh
@@ -2,6 +2,11 @@
 
 set -e
 
+if [[ $DEPLOY_ENV = "nobuild" ]]; then
+  echo "Skipping build as requested"
+  exit 0
+fi
+
 if ! [[ $DEPLOY_ENV = "testnet" ]] && ! [[ $DEPLOY_ENV = "mainnet" ]] && ! [[ $DEPLOY_ENV = "local" ]]; then
   echo "Which deployment environment? Set DEPLOY_ENV to 'testnet' or 'mainnet' or 'local'"
   exit 1
@@ -58,6 +63,5 @@ else
 fi
 
 echo Optimising wasm
-wasm-opt target/wasm32-unknown-unknown/release/nns-dapp.wasm --strip-debug -Oz -o target/wasm32-unknown-unknown/release/nns-dapp-opt.wasm
-
+ic-cdk-optimizer target/wasm32-unknown-unknown/release/nns-dapp.wasm -o target/wasm32-unknown-unknown/release/nns-dapp-opt.wasm
 sha256sum target/wasm32-unknown-unknown/release/nns-dapp-opt.wasm
--- a/scripts/deploy-docker-build-to-testnet
+++ b/scripts/deploy-docker-build-to-testnet
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -euo pipefail
+cd "$(dirname "$(realpath "$0")")/.." || exit
+
+WASM_PATH="$(jq -rc '.canisters["nns-dapp"].wasm' dfx.json)"
+
+docker build --build-arg DEPLOY_ENV=testnet -t nns-dapp .
+mkdir -p target/wasm32-unknown-unknown/release/
+# shellcheck disable=SC2094 # The source and destination files are not the same, the first is in docker.
+docker run --rm --entrypoint cat nns-dapp "$WASM_PATH" >"$WASM_PATH"
+docker run --rm --entrypoint cat nns-dapp /assets.tar.xz >assets.tar.xz
+DEPLOY_ENV=nobuild dfx deploy --network testnet
+
+set +x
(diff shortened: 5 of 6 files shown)


.dockerignore
```diff
@@ -27,3 +27,6 @@ frontend/dart/ios/Runner/GeneratedPluginRegistrant.m
 frontend/dart/lib/generated_plugin_registrant.dart
 frontend/dart/pubspec.lock
 frontend/dart/build/web
```
bitdivine, line 29: I am not sure why `.dockerignore` !== `.gitignore`.  Maybe we should make them the same?  If not, at least the dockerignore should ignore svelte build artefacts when copying files into the container.

.gitignore
```diff
@@ -1,3 +1,4 @@
+nns-dapp.wasm
```
bitdivine, line 1: This file is created by following the instructions in the readme.  There are no wasm files in the repo, maybe we could ignore all of them.

Dockerfile
```diff
@@ -3,7 +3,7 @@
 # docker build -t nns-dapp .
 # docker run --rm --entrypoint cat nns-dapp /nns-dapp.wasm > nns-dapp.wasm
 
-FROM ubuntu:20.10
+FROM ubuntu:20.10 as builder
```
bitdivine, line 6: The first half of the dockerfile installs dependencies.  By splitting it off explicitly we create a relatively strong guarantee that it won't have to be rebuilt often.  Given that docker builds are very slow and small things can cause a rebuild, this seems worthwhile.

Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
     apt -yq update && \
     apt -yqq install --no-install-recommends curl ca-certificates \
         build-essential pkg-config libssl-dev llvm-dev liblmdb-dev clang cmake \
```
bitdivine, line 17: We need jq and dfx (installed later).  Previously the build script was not working correctly in docker because of the absence of dfx.

Dockerfile
```diff
@@ -46,20 +46,17 @@ ENV PATH=/flutter/bin:$PATH
 # Install IC CDK optimizer
 RUN cargo install --version 0.3.1 ic-cdk-optimizer
 
-COPY . .
```
bitdivine, line 49: This is where we start the second docker container, based on the first that contains just build tools.

Dockerfile
```diff
@@ -46,20 +46,17 @@ ENV PATH=/flutter/bin:$PATH
...
+SHELL ["bash", "-c"]
+ARG DEPLOY_ENV=mainnet
+RUN echo $DEPLOY_ENV
+COPY . .
 
-RUN cd frontend/dart && flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=FLUTTER_WEB_CANVASKIT_URL=/assets/canvaskit/
-# Remove random hash from flutter output
-RUN sed -i -e 's/flutter_service_worker.js?v=[0-9]*/flutter_service_worker.js/' frontend/dart/build/web/index.html
+# Install dfx
+RUN DFX_VERSION="$(jq -cr .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
```
bitdivine, line 57: Dfx is installed here rather than in the tools section because the dfx version is in dfx.json which is only available in the second docker container.

Dockerfile
```diff
@@ -46,20 +46,17 @@ ENV PATH=/flutter/bin:$PATH
...
 
-# Bundle into a tight tarball
-RUN cd frontend/dart/build/web/ && tar cJv --mtime='2021-05-07 17:00+00' --sort=name --exclude .last_build_id -f ../../../../assets.tar.xz .
-RUN ls -sh assets.tar.xz; sha256sum assets.tar.xz
+RUN ./build.sh
```
bitdivine, line 59: Finally we get to replace docker script with build.sh.

Dockerfile
```diff
@@ -46,20 +46,17 @@ ENV PATH=/flutter/bin:$PATH
...
 
-RUN cargo build --locked --target wasm32-unknown-unknown --release --package nns-dapp
-RUN cp target/wasm32-unknown-unknown/release/nns-dapp.wasm .
-RUN ic-cdk-optimizer nns-dapp.wasm -o nns-dapp.wasm
+RUN cp "$(jq -rc '.canisters["nns-dapp"].wasm' dfx.json)" nns-dapp.wasm
```
bitdivine, line 61: build.sh now uses the same optimiser as docker so we can just copy out the docker build artefact.

build.sh
```diff
@@ -2,6 +2,11 @@
 
 set -e
 
+if [[ $DEPLOY_ENV = "nobuild" ]]; then
+  echo "Skipping build as requested"
+  exit 0
+fi
```
bitdivine, line 8: Is it possible to run `dfx deploy` without dfx running the build script?  We need to be able to deploy a docker-built wasm file.  The bypass here is not very elegant but works.

build.sh
```diff
@@ -44,6 +49,5 @@ else
 fi
 
 echo Optimising wasm
-wasm-opt target/wasm32-unknown-unknown/release/nns-dapp.wasm --strip-debug -Oz -o target/wasm32-unknown-unknown/release/nns-dapp-opt.wasm
-
+ic-cdk-optimizer target/wasm32-unknown-unknown/release/nns-dapp.wasm -o target/wasm32-unknown-unknown/release/nns-dapp-opt.wasm
```
bitdivine, line 52: This is the optimizer used in production.

scripts/deploy-docker-image-to-testnet
```diff
@@ -0,0 +1,12 @@
+set -eux
```
bitdivine, line 1: Running this successfully deployed a docker build to the testnet.

scripts/deploy-local-build-to-testnet
```diff
@@ -0,0 +1,2 @@
+DEPLOY_ENV=testnet dfx deploy --network testnet
```
bitdivine, line 1: Running this successfully deployed a local build to the testnet.

.dockerignore
nmattia: there are things like the README that you probably want in git but not in the docker builds

Dockerfile
nmattia: how about pinning it as well? (unrelated to this PR)

.gitignore
nmattia: what about ignoring this in the dockerfile? you probably don't want to invalidate the layer just because this is around.

Dockerfile
nmattia: note: you could also `COPY dfx.json dfx.json` in the first stage

build.sh
```diff
@@ -2,6 +2,11 @@
 
 set -e
 
+if [[ $DEPLOY_ENV = "nobuild" ]]; then
```
nmattia, line 5: Where does `nobuild` come from?

---

EDIT: found it. Would be nice to add a line here explaining what this is.

scripts/deploy-docker-build-to-testnet
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
nmattia, line 1: this looks like one of those scripts that will go out of date, but happy to go with it for now!

scripts/deploy-docker-build-to-testnet
```diff
@@ -0,0 +1,15 @@
...
+cd "$(dirname "$(realpath "$0")")/.." || exit
+
+WASM_PATH="$(jq -rc '.canisters["nns-dapp"].wasm' dfx.json)"
+
+docker build --build-arg DEPLOY_ENV=testnet -t nns-dapp .
+mkdir -p target/wasm32-unknown-unknown/release/
+# shellcheck disable=SC2094 # The source and destination files are not the same, the first is in docker.
+docker run --rm --entrypoint cat nns-dapp "$WASM_PATH" >"$WASM_PATH"
+docker run --rm --entrypoint cat nns-dapp /assets.tar.xz >assets.tar.xz
+DEPLOY_ENV=nobuild dfx deploy --network testnet
```
nmattia, line 12: It's really odd having this `nobuild` env just for the sake of `dfx deploying` that we're not `dfx build`ing. Would it be possible to bypass `dfx deploy` altogether? This looks like a recipe for disaster.

scripts/deploy-docker-build-to-testnet
ielashi: @bitdivine FYI we already have a `deploy.sh` script. Maybe we can move these two deploy scripts into `deploy.sh`? I think this can also address nmattia's concern below.

.dockerignore
bitdivine: so construct .dockerignore from .gitignore and a few additional files?

nmattia: I'm simply giving an example of why `.dockerignore` !== `.gitignore` !

Dockerfile
nmattia: I meant adding a sha256 digest, but we can do that separately!

build.sh
nmattia: All good!

nmattia (approved):
LGTM, we can always improve further in subsequent PRs!

bitdivine:
Full transcript of the test notes:

```
375bd498c56953cf4d98472956b1056e52bf306a = last commit before svelte.

541e3f1803e288d09982aa4a95a509b2589db07f = build commit
```

# Check build before the commit:

## Create a branch at the last commit before svelte:

```
max@sinkpad:~/dfn/nns-dapp/branches/dockerfile-for-testnet (1:29)$ git checkout -b test-build
Switched to a new branch 'test-build'
max@sinkpad:~/dfn/nns-dapp/branches/dockerfile-for-testnet (1:39)$ git reset --hard 375bd498c56953cf4d98472956b1056e52bf306a
```

## Build and copy out the wasm and assets
```
    docker build -t nns-dapp .
...
    622f1ac7820dadd8d683166d8c6c1eb0573ebc87dbdff54762e8f2a506281c7b  nns-dapp.wasm

    docker run --rm --entrypoint cat nns-dapp /nns-dapp.wasm > nns-dapp.wasm
    docker run --rm --entrypoint cat nns-dapp /assets.tar.xz > assets.tar.xz
    mv assets.tar.xz assets-before.tar.xz
    mv nns-dapp.wasm nns-dapp-before.wasm
```

## Repeat to verify that the buil dis actually reproducible
```
    docker build -t nns-dapp .
    622f1ac7820dadd8d683166d8c6c1eb0573ebc87dbdff54762e8f2a506281c7b  nns-dapp.wasm
```
## With the new docker build command
```
    git cherry-pick 541e3f1803e288d09982aa4a95a509b2589db07f

    MERGE CONFLICT.

    fixing in a way that I hope will not affect the outcome...

    New commit: 9c531908cd6544af7a7c2c9259c1c5b78c614b6d

    docker build -t nns-dapp .
    c6f64f6e72b6b9ac55f7f49b864eea16b67e1c208f1149dc8d2947f6c3927fd6  nns-dapp.wasm

    Check:  The old docker image included the notices file.  In the new it doesn't.  The build script says:

    `# Remove the assets/NOTICES file, as it's large in size and not used.`

    This may explain why production builds are short of space.

    If I comment out the line in build.sh that removes the notices file:

    docker build -t nns-dapp .
    622f1ac7820dadd8d683166d8c6c1eb0573ebc87dbdff54762e8f2a506281c7b  target/wasm32-unknown-unknown/release/nns-dapp-opt.wasm
```

PR #313, opened 2021-12-22
on changing tabs , the correct url is now shown

anishh2003:
This commit solves the problem of having only accounts being shown in the url on tab navigation . The url now changes dynamically on tab navigation.


frontend/dart/lib/ui/home/home_tabs_widget.dart
```diff
@@ -109,61 +100,100 @@ class _HomePageState extends State<HomePage> {
...
-          ),
+      ),
+      body: SizedBox.expand(
+        child: TabBarView(
+          controller: _tabController,
+          children: [
+            AccountsTabWidget(),
+            NeuronsPage(),
+            GovernanceTabWidget(),
+            CansitersPage()
```
hpeebles, line 195: Typo

anishh2003: this has been changed to CanistersPage() now

hpeebles (approved):

PR #317, opened 2021-12-23
Update styling

bitdivine:
# Motivation
We have styling updates from Jan and Dom.

Here is the new look (screenshot with just the text and colour changes, not the whole screen):

Screenshot from 2021-12-23 13-22-15

# Changes
- Update text and styling of the login page.

### DIFF
--- a/frontend/svelte/src/Auth.svelte
+++ b/frontend/svelte/src/Auth.svelte
@@ -78,7 +78,7 @@
   {#if !signedIn && authClient}
     <div class="auth-overlay">
       <div />
-      <h1>the INTERNET COMPUTER</h1>
+      <h1>INTERNET COMPUTER</h1>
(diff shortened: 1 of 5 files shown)


frontend/svelte/src/Auth.svelte
```diff
@@ -65,8 +65,8 @@
   };
 
   // Sets login status on first load.
-  onMount(() => {
-    checkAuth();
+  onMount(async () => {
```
nmattia, line 68: curious why this changed? guess it was overlooked previously? did the linter not complain?

anishh2003 (approved):

frontend/svelte/src/Auth.svelte
bitdivine: Oops, that isn't meant to be in this PR.  That is the login loop fix.  I'll put it in a separate PR.  Thank you!

PR #318, opened 2021-12-23
Deployment does not need a wallet

bitdivine:
# Motivation
Deploying a canister to a testnet does not need a wallet, and if there is no wallet there is a rather confusing error message.

`deploy.sh` already uses `--no-wallet`.  The same was omitted from the docker testnet deploy script but should be included.

# Changes

- Add `--no-wallet` to the docker testnet deploy script
- Add `-x` to the same script, as it is helpful to see which step it is at.


scripts/deploy-docker-build-to-testnet
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euo pipefail
+set -euxo pipefail
```
nmattia, line 2: is this for debugging or meant to be merged? I saw that many scripts in this repo do have `set -x`, which makes the output _very_ verbose I find; I'd rather have meaningful `echo`s where necessary

bitdivine: This script is short and I really do want every line it prints.  See the +x below when this stops being true.

nmattia: I reluctantly approve then :grumpyface:

nmattia (approved):

PR #319, opened 2021-12-23
Fix login loop

bitdivine:
# Motivation
The following bug was observeable in the previous version of this code:

- Log in to the nns, by clicking a link to the NNS ending in /v2 and then logging in as usual
- Keeping that logged in tab open, again click the link to .../v2
- The browser was likely to enter a login/logout loop.

The code from this PR has been deployed here and should work correctly:

https://qhbym-qaaaa-aaaaa-aaafq-cai.peopleparty01.dfinity.network/v2/

# Testing
Do not deploy until all of these have been tested
- [x] chrome on Linux
- [x] Firefox on Linux
- [x] Chrome on Android
- [ ] Chrome on Mac
- [ ] FF on Mac
- [ ] Safari on Mac
- [ ] iphone
- [ ] IE on Windows
- [ ] Chrome on Windows
- [ ] FF on windows


frontend/svelte/src/Auth.svelte
```diff
@@ -59,14 +59,15 @@
     signedIn = false;
     authSync.onSignOut();
     // Ensure that all data is wiped
-    // ... if we have unencrypted data in local storage, delete it here.
+    // ... if we have data in local storage, (delegations, account data, ...), delete it here.
+    window.localStorage.clear();
```
nmattia, line 63: The comment suggests there was an intent to do so... just to clarify, did you just forget to actually clear the storage or was this supposed to be handled in some other way?

frontend/svelte/src/Auth.svelte
```diff
@@ -59,14 +59,15 @@
...
     // ... wipe data in ephemeral state, but in the next tick allow repaint to finish first.
     setTimeout(() => location.reload(), 100);
   };
 
   // Sets login status on first load.
-  onMount(() => {
-    checkAuth();
+  onMount(async () => {
+    await checkAuth();
```
nmattia, line 70: Shouldn't the linter catch promises that haven't been awaited?

nmattia (commented):
Instead of clearing localStorage, shouldn't we file a bug with `agent-js` and have `authClient.logout();` do a proper logout?

bitdivine:
Note: This PR has fallen behind main and has a merge conflict, so I have merged in the latest main.  Re-testing....

bitdivine:
> Instead of clearing localStorage, shouldn't we file a bug with `agent-js` and have `authClient.logout();` do a proper logout?

There are probably cases where it is useful to retain local storage even when logged out.  Ditto local databases, cookies and any other local state one can think of.

frontend/svelte/src/Auth.svelte
nmattia: I think we can enforce this: https://eslint.org/docs/rules/require-await, wanna have a go (in a different PR)?

frontend/svelte/src/AuthSync.ts
```diff
@@ -38,18 +38,31 @@ export class AuthSync {
     this.#onChange(AuthSync.SIGN_OUT);
   }
 
+  /**
+   * Gets the currently stored action.
+   */
+  #getStoredAction() {
```
nmattia, line 44: Looks like the timestamp is never read; why write it?

frontend/svelte/src/Auth.svelte
```diff
@@ -40,11 +40,14 @@
       });
     });
     onSignIn();
-    authSync.onSignIn();
   };
 
   // Gets a local copy of user data.
   const onSignIn = async () => {
+    authSync.onSignIn();
+    if (process.env.REDIRECT_TO_LEGACY) {
```
nmattia, line 48: What sets this?

frontend/svelte/src/Auth.svelte
```diff
@@ -56,13 +59,20 @@
...
     setTimeout(() => location.reload(), 100);
   };
 
   // Sets login status on first load.
-  onMount(checkAuth);
+  onMount(async () => {
+    await checkAuth();
+    // If logged out by flutter, we still need to broadcast the logout status.
+    if (!signedIn) {
+      authSync.onSignOut();
```
nmattia, line 73: doesn't `authCheck` already call `signOut` if we're not signed in, which in turn calls `authSync.onSignOut`?

nmattia: in order to make things a bit more symmetrical (at the expense of a few CPU cycles) why not call `signIn` in `authCheck` (if it turns out we're signed in), as opposed to only calling `onSignIn`?

bitdivine:
The svelte routing mechanism has changed and it makes more sense to deploy the updated mechanism at once.

PR #322, opened 2021-12-23
refactor: Update agent-js from `0.9.1` to `0.10.2`

Daniel-Bloom-dfinity:
agent-js (as of `0.10.2`) avoids `ic0.app` subdomains, and thus 307 redirects.

The leap from `0.9` to `0.10` includes a breaking change which eliminated the use of Blobs and changed the way buffers were used. This change accounts for that.


frontend/ts/src/ledger/secp256k1.ts
```diff
@@ -57,10 +51,16 @@ export class Secp256k1PublicKey implements PublicKey {
       );
     }
 
-    const rawKey = blobFromUint8Array(
-      key.subarray(Secp256k1PublicKey.DER_PREFIX.length)
-    );
-    if (!this.derEncode(rawKey).equals(key)) {
+    ArrayBuffer;
```
ielashi, line 54: We don't need this line, do we?

ielashi: FYI this whole file was copied from agent-js. Until we use the agent-js version, maybe instead we should copy https://github.com/dfinity/agent-js/blob/2fe3dd99cddfcf45c6d9d5b7a199a86285ce9740/packages/identity-ledgerhq/src/identity/secp256k1.ts into this file to keep them in sync?

ielashi (commented):
LGTM - left one minor comment. I'll do some testing of this on the testnet (once it's working again) before approval.

ielashi (approved):
Tested on a testnet and LGTM!

PR #325, opened 2022-01-03
build: prettier configuration

peterpeterparker:
# Motivation

Defines a `prettier` configuration for the svelte app.

# Changes

- I took the default from sveltekit (see below `.prettierrc`), removed the `useTabs` and added a couple of personal preferences but not against any other options.

```
{
        "useTabs": true,
        "singleQuote": true,
        "trailingComma": "none",
        "printWidth": 100
}
```


nmattia:
@peterpeterparker you'll probably need to re-push to trigger the CI. Right now frontend formatting changes are pushed by the GitHub Actions, but due to GH Actions limitations (and because we haven't set a custom token) this push will not trigger a CI run.

frontend/svelte/.prettierrc
```diff
@@ -0,0 +1,16 @@
+{
```
nmattia, line 1: Some options here (like no trailing comma, using single quotes, no bracket spacing) impact quite a few files and (ideally) would need to be ported to other repos for consistency. Would be great if you could clarify a bit why you prefer these. Thanks!

peterpeterparker: - [no trailing comma](https://prettier.io/docs/en/options.html#trailing-commas): if you do not set trailing comma to `none` you get formatting such as following which I personally find harder to read. It is also the default prettier settings provided by sveltekit (see [template](https://github.com/sveltejs/kit/blob/1d76d66038647c8a67dbb011cf9c0e231fd086a6/packages/create-svelte/shared/%2Bprettier/.prettierrc)).

```
// Trailing comma
const hello = {
  a: '1',
  b: 2,
}

// Trailing comma: none
const hello = {
  a: '1',
  b: 2
}
```

- [single quotes](https://prettier.io/docs/en/options.html#quotes): I kind of feel single quotes usage in JavaScript dominate the scene. It is also the default prettier settings provided by sveltekit (see [template](https://github.com/sveltejs/kit/blob/1d76d66038647c8a67dbb011cf9c0e231fd086a6/packages/create-svelte/shared/%2Bprettier/.prettierrc)).

On the contrary, for scss and css, I think double quotes is the common use, that's why I override the rules for these.

- [bracket spacing](https://prettier.io/docs/en/options.html#bracket-spacing): no particular reason, I am fine with or without, we can get rid of it in my PR if you rather like

- [arrowParens](https://prettier.io/docs/en/options.html#arrow-function-parentheses): I mostly use arrow parents because I try to always specify types but here again but we can get rid of it if you rather like

- [bracketSameLine](https://prettier.io/docs/en/options.html#bracket-line): it makes HTML easier to read and is a bit less ugly in my opinion

nmattia: Thanks for the details answer!

I'd suggest keeping the trailing commas and not having the brackets on the same line, because this makes things simpler when you edit the code and makes the diffs more consistent. I'd also keep the space in the brackets if there's no other reason. Other than that, LGTM!

What do you think?

peterpeterparker: done. I've remove both `"bracketSpacing": false` and `"bracketSameLine": true` from this PR prettier config

bitdivine:
Note: There is already a (blank) .prettierrc in frontend/ts.  We don't HAVE to be consistent with the existing code but consistency is generally a good thing.  If we are going to have a meaningful (non-trivial) .prettierrc we should probably put it in /frontend/ or even in the root of the repo so that it applied equally to all.  Prettier looks in parent directories (by memory and confirmed with a test just now) so this should work: (it does, tested)

```
# Copy the proposed config into the root of the repo:
cp frontend/svelte/.prettierrc .prettierrc.json
# There are merge conflicts, so get the latest main:
git fetch ; git reset --hard origin/main
# Add the new config, remove the existing ones
git add .prettierrc.json
git rm frontend/svelte/.prettierrc.json frontend/ts/.prettierrc.json
git commit -a -m "Make a common .prettierrc"
# Run prettier everywhere:
pushd frontend/ts ; npm ci ; npm run format ; popd
pushd frontend/svelte ; npm ci ; npm run format ; popd
git commit -a -m "Format"
# Then merge fast before there are more merge conflicts!
```

Note: Given that this affects code that @ielashi  is working on, we should check that he is OK with this course of action and these formatting rules.  If so, let's get it in without a long review.  What do you think, @peterpeterparker  and @nmattia ?

bitdivine:
Regarding choice of rules, this is something I generally do not care about, as long as there is a set of rules per repo so that it never takes human time.  Here are a few policies in different dfn or afiliated projects:
```
max@sinkpad:~ (12)$ cat /home/max/projects/open-chat/v2/frontend/.prettierrc
{
    "svelteBracketNewLine": false,
    "tabWidth": 4,
    "useTabs": false,
    "printWidth": 100
}
max@sinkpad:~ (25)$ cat /home/max/dfn/icfront/agent-js/.prettierrc
{
  "trailingComma": "all",
  "tabWidth": 2,
  "printWidth": 100,
  "semi": true,
  "bracketSpacing": true,
  "useTabs": false,
  "singleQuote": true,
  "quoteProps": "consistent",
  "arrowParens": "avoid"
}
max@sinkpad:~ (41)$ cat /home/max/002/boundary_node/typescript/service-worker/.prettierrc
{
  "trailingComma": "all",
  "tabWidth": 2,
  "printWidth": 100,
  "semi": true,
  "bracketSpacing": true,
  "useTabs": false,
  "singleQuote": true,
  "quoteProps": "consistent",
  "arrowParens": "avoid"
}
```

peterpeterparker:
Agree, it would be useful to have the same configuration across projects.

Regarding choice of rules, I am flexible only "single quotes in TS/JS"  would make me happy 😉.

ielashi:
I have no preference personally as well and I don't mind extending the prettier config to `frontend/ts`

bitdivine:
Done.  If this is really what you want, please agree, approve and merge.  Given that Islam and I don't mind either way, it's up to @peterpeterparker  and @nmattia.  I would only ask that we make such changes rarely. :-)

frontend/svelte/.prettierrc
nmattia: Can you also re-add the trailing commas, for the reasons explained above? thanks!

>  this makes things simpler when you edit the code and makes the diffs more consistent

nmattia:
@bitdivine LGTM, modulo the trailing commas! Also thanks for reformatting and keeping the commits separate. Can you re-add the trailing commas? Then I reckon we're ready to merge

peterpeterparker:
It's been a while, I guess we will go on with the default format.

PR #326, opened 2022-01-03
build: eslint

peterpeterparker:
# Motivation

Detects JS errors ahead of compilation time.

# Changes

- adds default eslint configuration

# Notes

- to ignore folders in eslint, an `.eslintignore` can be created but needs to finds place at the root. since we have a mono-repo, to avoid conflicts, I just configured the linter to lint the svelte `src` folder

- this PR does not adds yet the linter to the ci or build chain

### DIFF
--- a/frontend/svelte/.eslintrc.cjs
+++ b/frontend/svelte/.eslintrc.cjs
@@ -1,0 +1,24 @@
+module.exports = {
+  root: true,
+  parser: "@typescript-eslint/parser",
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "prettier",
+  ],
+  plugins: ["svelte3", "@typescript-eslint"],
+  ignorePatterns: ["*.cjs"],
+  overrides: [{ files: ["*.svelte"], processor: "svelte3/svelte3" }],
+  settings: {
+    "svelte3/typescript": () => require("typescript"),
+  },
+  parserOptions: {
+    sourceType: "module",
+    ecmaVersion: 2020,
+  },
+  env: {
+    browser: true,
+    es2017: true,
+    node: true,
+  },
+};
--- a/frontend/svelte/.eslintrc.cjs
+++ b/frontend/svelte/.eslintrc.cjs
@@ -15,10 +15,17 @@ module.exports = {
   parserOptions: {
     sourceType: "module",
     ecmaVersion: 2020,
+    project: ["./tsconfig.json"]
   },
   env: {
     browser: true,
     es2017: true,
     node: true,
+  },
+  rules: {
+    "@typescript-eslint/strict-boolean-expressions": [
+      2,
+      { allowString: false, allowNumber: false },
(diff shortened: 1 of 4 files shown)


frontend/svelte/.eslintrc.cjs
```diff
@@ -0,0 +1,24 @@
...
+  plugins: ["svelte3", "@typescript-eslint"],
+  ignorePatterns: ["*.cjs"],
+  overrides: [{ files: ["*.svelte"], processor: "svelte3/svelte3" }],
+  settings: {
+    "svelte3/typescript": () => require("typescript"),
+  },
+  parserOptions: {
+    sourceType: "module",
+    ecmaVersion: 2020,
+  },
```
nmattia, line 18:
```suggestion
  },
      "rules": {
        "@typescript-eslint/strict-boolean-expressions": [
            2,
            { "allowString" : false,
                "allowNumber" : false
            }
        ]
    },
```

Would be great to have the same rules as in https://github.com/dfinity/internet-identity/blob/291650809945d761631d1babe163f70d89584534/.eslintrc.json !

peterpeterparker: Done

nmattia (approved):
Thanks, LGTM!

nmattia:
@peterpeterparker the GitHub action pushed formatting changes and you'll need to re-trigger CI

peterpeterparker:
CI lint [checks](https://github.com/dfinity/nns-dapp/runs/4704196021?check_suite_focus=true) are failing but most of these are either solved or deprecated by [#3289](https://github.com/dfinity/nns-dapp/pull/3289), so we might wait on my other PR to go ahead here I would say.

bitdivine:
My comments are similar to the comments about the prettierrc.  We already have .eslintrc config files so we should probably just elevate one to the root of the repo.  There was actually a svelte PR that included that but it was an early sketch and didn't get merged.

frontend/svelte/package.json
```diff
@@ -8,6 +8,7 @@
     "dev": "npm run import-assets && rollup -c -w",
     "start": "sirv public",
     "check": "svelte-check --tsconfig ./tsconfig.json",
+    "lint": "prettier --check --plugin-search-dir=. . && eslint src",
```
nmattia, line 11: nitpick: what about only running the linter in `lint`? Is this meant as a "let's catch everything" as opposed to just linting?

nmattia (commented):
Looking great, thanks! Before we merge, let's make sure CI is checking this. You can add a step similar to the [formatting check](https://github.com/dfinity/nns-dapp/blob/4a6767868fd9eec338d0155a193b7a6295127d70/.github/workflows/checks.yml#L39-L41).

peterpeterparker:
Let's keep the existing eslint setup as in root.

In the future, if we notice we might need more default svelte eslint settings, we can have a look at this in a cleaner / more global way as the PR I provided.

PR #332, opened 2022-01-06
Fix deploy.sh

bitdivine:
# Motivation
`./deploy.sh` had a bug in the cd line: Unlike most scripts, `deploy.sh` is not in `scripts/` but in the root of the repo, so moving up one directory is inappropriate.

# Changes
Removed the `/..` from the cd line.

# Testing
* Created a clean new workspace
* Ran `./deploy.sh testnet`
* The code ran without error and the site is here: https://qsgjb-riaaa-aaaaa-aaaga-cai.nnsdapp.dfinity.network/


nmattia (commented):
How about using `git rev-parse --show-toplevel`? It's not likely this script will be called without the `.git` around

bitdivine:
@nmattia Because this already works, makes fewer assumptions about the environment and is the standard mechanism for moving to the appropriate directory.

bitdivine:
By the way, with git show-toplevel, that introduces an ambiguity in the context of git workspaces.  show-toplevel is not all that simple.  Why add complexity?

nmattia (approved):

PR #333, opened 2022-01-06
Use svelte login page

bitdivine:
# Motivation
The svelte router has changed, so the mechanism for switching between svelte and flutter needs to be updated.

# Changes
* Added flutter code to redirect to svelte for login.
* No new svelte code is needed as the redirect code is already present.
* Changed the default of REDIRECT_TO_LEGACY to redirect by default.

# Testing
This code has been deployed to:
- https://q3fc5-haaaa-aaaaa-aaahq-cai.nnsdapp.dfinity.network
- https://nns.winning.gold <-- shortcut to the above
- QR code for nns.winning.gold

Please run tests USER_001 and USER_003 from the [manual testing runbook](https://www.notion.so/NNS-Dapp-Manual-Testing-Runbook-b8a881b4066f4dfeb81538649aa63239#24f8bdacee684b46809002590088372e) on these platforms:

[There is a discussion thread for the results.](https://app.slack.com/client/T43F9UHS5/C01S03NBM7S/thread/C01S03NBM7S-1641552855.017800)

- [x] chrome on Linux - Done, Max
- [x] Firefox on Linux - Done, Max
- [x] Chrome on Android - Done, Max
- [x] Chrome on Mac
 - [x] Maria encountered a concurrency bug, tracked here: https://app.slack.com/client/T43F9UHS5/C01S03NBM7S/thread/C01S03NBM7S-1641562611.029900
   - I believe that this is "how mac works" rather than a bug.  Considering this as resolved. 
- [x] FF on Mac
  - Fails on FF72.  Works of FF95.  Need to define which versions we support but 72 is old enough that we probably don't care.
- [x] Safari on Mac
- [x] Safari on iphone
- [x] Edge on Windows
- [x] Chrome on Windows
- [x] FF on windows
 
Automated testing:
- I have written a selenium script to verify that login works.  It does.  Need to align with @nmattia on the selenium testing - whether to merge this or whether his is already ready for merging.


frontend/dart/web/logout.js
```diff
@@ -0,0 +1,10 @@
+/* If ever there is no delegation, go to the svelte login page. */
+(() => {
+     function checkLoginStatus() {
+       if (null === window.localStorage.getItem("ic-delegation")) {
```
nmattia, line 4: this makes some assumptions about auth-client's internals, correct?

frontend/svelte/src/routes/Accounts.svelte
```diff
@@ -9,8 +9,10 @@
   });
 </script>
 
-<Layout>
-  <section>
-    <h1>ICP page</h1>
-  </section>
-</Layout>
+{#if !process.env.REDIRECT_TO_LEGACY}
```
nmattia, line 12: why disable this? I thought we could never land here anyway

nmattia (commented):
Few comments; I know this is still a draft, sorry if I jumped the gun!

frontend/svelte/src/routes/Accounts.svelte
bitdivine: We can and do, but when we do, the page is redirected to flutter.  During the redirect this page is visible.

frontend/dart/web/logout.js
bitdivine: I don't believe that this is internal; it is a stable, expected behaviour.  It's not something that is going to change over the next month or so anyway.  It can only break by us updating the ic-agent we use in the NNS and if we update it in a breaking way, that will show up in testing.  How is your automated test suite coming along?

bitdivine:
@nmattia - NP - it is scheduled to go into production today.

nmattia (approved):
Let's do this!

PR #334, opened 2022-01-07
EXC-805: Add console command for `addHotKey`.

ielashi:
This PR adds a console command that allows users to add hotkeys to
their neurons. Example:

```
await nns.governance.addHotKey("<neuron-id>", "<principal>");
```

The motivation for this is to resolve https://forum.dfinity.org/t/8-year-locked-neuron-missing-from-neurons-tab

### DIFF
--- a/frontend/ts/src/ServiceApi.ts
+++ b/frontend/ts/src/ServiceApi.ts
@@ -126,6 +126,7 @@ export default class ServiceApi {
     this.cyclesMintingService = cyclesMintingServiceBuilder(agent);
     this.icManagementService = icManagementBuilder(agent);
     this.identity = identity;
+    setupExports(this.governanceService);
   }
 
   /* ACCOUNTS */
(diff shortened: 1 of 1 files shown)


frontend/ts/src/ServiceApi.ts
```diff
@@ -643,3 +644,23 @@ async function ledgerService(identity: Identity): Promise<LedgerService> {
...
+function setupExports(governance: GovernanceService) {
+  // @ts-ignore
+  window["nns"] = {
+    governance: {
+      // Modifying the method signature to be a bit more ergonomic for the console.
+      addHotKey: function (neuronId: NeuronId, principal: PrincipalString) {
+        return governance.addHotKey({
+          neuronId: neuronId,
+          principal: principal,
+        });
```
bitdivine, line 662: Presumably this can be made more compact:
```suggestion
        return governance.addHotKey({ neuronId, principal });
```
But this is not important.  If ci is green, please press merge!

bitdivine (approved):

frontend/ts/src/ServiceApi.ts
ielashi: Prettier doesn't seem to like that unfortunately.

PR #342, opened 2022-01-11
feat: spa routing

peterpeterparker:
# Motivation

Replace navigation based on `location.replace()` with a more conventional SPA / html5 approach i.e. a routing based on the [history api] and `location.hash` as fallback for old browser.

The goal is also to minimize full page refresh and by extension the amount of requests on the assets canister for the JS resources.

Finally, this change will allow us to implement "404" pages.

# Changes

- introduces a new global store that contains the current route path
- replaces  `location.replace()` with history API and fallback in route utils
- adds a new function `navigate()` to navigate programmatically
- comment navigation actions to give more context

# Note

- the new function `navigate()` is not yet used . I used it for test purpose and I can image we might need it in the future.
- navigation actions (accounts, neurons, etc.) remain normal link `<a/>` for the moment. let's see in the future if we rather transform these to `<button/>`, or hijack these, to use the `navigate()` function


bitdivine:
Thank you.  Can you deploy this to the testnet so that we can test it?  If deployment fails, please let me know and I can either deploy it or (better) help you fix the deployment.

frontend/svelte/src/lib/utils/route.utils.ts
```diff
@@ -1,11 +1,69 @@
...
+    return;
+  }
+
+  history.pushState({}, "", `${baseUrl()}${path}`);
+};
+
+const concatPathQuery = ({ path, query }: { path: string; query?: string }) =>
+  `${path}${query ? `?${query}` : ""}`;
+
+const supportHistory = (): boolean => {
```
nmattia, line 49:
```suggestion
const supportsHistory = (): boolean => {
```

frontend/svelte/src/lib/utils/route.utils.ts
```diff
@@ -1,11 +1,69 @@
...
+  const ua: string = window.navigator.userAgent;
+  if (
+    (ua.indexOf("Android 2.") !== -1 || ua.indexOf("Android 4.0") !== -1) &&
+    ua.indexOf("Mobile Safari") !== -1 &&
+    ua.indexOf("Chrome") === -1 &&
+    ua.indexOf("Windows Phone") === -1
```
nmattia, line 55: so these devices have a `window.history.pushState` but it doesn't work?

frontend/svelte/src/lib/utils/route.utils.ts
```diff
@@ -1,11 +1,69 @@
...
+  ) {
+    return false;
+  }
+  return window.history && "pushState" in window.history;
+};
+
+const baseUrl = (): string => {
```
nmattia, line 62: I'm a bit confused. Doesn't this return the path, instead of the base URL? I get the following running it on this page:

```js
path = new URL(document.baseURI).pathname;
path.replace(/\/$/, "");
> "/dfinity/nns-dapp/pull/342/files" = $1
```

frontend/svelte/src/lib/stores/route.store.ts
```diff
@@ -0,0 +1,39 @@
+import { writable } from "svelte/store";
+import { pushHistory, replaceHistory, routePath } from "../utils/route.utils";
+
+export interface RouteStore {
+  path: string;
+}
+
+/**
```
nmattia, line 8: 🙏

frontend/svelte/src/lib/components/PrivateRoute.svelte
```diff
@@ -17,7 +18,12 @@
     }
 
     // Redirect to root, user needs to sign in
-    window.location.replace(`${appPath()}/?redirect=${routeContext()}`);
+    // We replace the url to get the redirect query params after successful sign in, to redirect user to current page
```
nmattia, line 21:
```suggestion
    // We replace the url to get the redirect query params after successful sign in, to redirect user to the current page
```

frontend/svelte/src/lib/utils/route.utils.ts
```diff
@@ -1,11 +1,69 @@
...
+
+  if (!supportHistory()) {
+    window.location.replace(path);
+    return;
+  }
+
+  history.replaceState({}, "", `${baseUrl()}${path}`);
+};
+
+export const pushHistory = (params: { path: string; query?: string }) => {
```
nmattia, line 35: nitpick: why split the `path` and `query` params if we're just gonna reconstruct the full path anyway? Same applies to `replaceHistory`. I'd rather do the reconstruction in the function that calls this one.

frontend/svelte/src/lib/stores/route.store.ts
```diff
@@ -0,0 +1,39 @@
...
+    subscribe,
+
+    update: ({ path }: { path: string }) =>
+      update((state: RouteStore) => ({ ...state, path })),
+
+    navigate: ({ path, query }: { path: string; query?: string }) => {
+      update((state: RouteStore) => ({ ...state, path }));
+
+      pushHistory({ path, query });
+    },
```
nmattia, line 29: Since we don't use `navigate()` I'd suggest removing it; we can always re-add it later. It's dead code and might create confusion.

nmattia (commented):
Thanks, this is great! Few comments. Also, can you give a brief example of a typical redirect (in the history case)? i.e. the paths involved and what the history looks like; I'm still wrapping my head around this push history!

frontend/svelte/src/lib/utils/route.utils.ts
peterpeterparker: I read in a book that even though we can think testing only `window.history.pushState` would be enough testing also the agent for really old phones can be useful.

frontend/svelte/src/lib/stores/route.store.ts
peterpeterparker: I would rather like to keep it for the time being. In a couple of weeks, if we don't use it, then we will delete it.
It's going to be more efficient than deleting it, PR wasn't that easy to develop...

frontend/svelte/src/lib/utils/route.utils.ts
nmattia: Ah, I see! Can you add a note about this? Would be great if you could find the reference too!

frontend/svelte/src/lib/stores/route.store.ts
nmattia: I fear that by the time you'll want to use it, code around it might have changed and it won't do exactly what you intended it to do in the first place. In the meantime it's code that still needs to be maintained (for instance if the signature of `pushHistory` changes) and will confuse people.

peterpeterparker: I fear that by the time I want to use it, I'll have to invest again x hours to rebuild it.
Let's keep it for a while, if it becomes a pain, then I'll remove it any day.

frontend/svelte/src/lib/utils/route.utils.ts
peterpeterparker: I removed the function. We do not use the `<base/>` in this app and, as you noticed, might not fully worked out.

No `baseUrl()`, no confusion.

frontend/svelte/src/lib/utils/route.utils.ts
peterpeterparker: Done.

frontend/svelte/src/lib/utils/route.utils.ts
nmattia: Thanks for the clarification!

frontend/svelte/src/lib/utils/route.utils.ts
peterpeterparker: yes in the utils it would be possible to receive only a "fullPath" that contain both instead of two params

on the contrary, on the store, we do not want the "query", only the "path" is meaningful

> A store to handle the current route ("which page to display").

frontend/svelte/src/lib/utils/route.utils.ts
nmattia: Ok, I was thinking about a link to an article; just the title to the book won't really be helpful if someone just comes to check the code. 

According to stackoverflow™ it's enough to check that `pushHistory` exists and that its type isn't `undefined` ([link](https://stackoverflow.com/questions/6824991/how-to-find-if-a-browser-supports-history-pushstate-or-not#comment93694452_6825002)). Unless we have a good reason not to, can we keep the code as simple as possible?

frontend/svelte/src/lib/utils/route.utils.ts
nmattia: I think this is misleading because it suggests that `pushHistory` will actually do something special with the query params; moreover the abstraction would break if you ever need to call with a path that's not already split (i.e. `/foo?bar=baz` would need to be split on `?` first, just to fit `pushHistory`'s interface, then reconstructed on the other side). But you're call :)

frontend/svelte/src/lib/utils/route.utils.ts
peterpeterparker: Code and comment updated. Done.

frontend/svelte/src/lib/utils/route.utils.ts
peterpeterparker: I see the point. If we defer the logic to the store, then the store will have to be refactored for kind of the same reason.

Therefore, I would suggest to let it as it is in this PR. Subsequent PRs can be provided for refactoring purpose.

frontend/svelte/src/lib/components/Guard.svelte
```diff
@@ -1,8 +1,16 @@
...
+  import { routePath } from "../utils/route.utils";
+
+  // Browser back button has been clicked, we reflect the new browser url to the route
+  const updateRoute = () => routeStore.update({ path: routePath() });
 </script>
 
-<svelte:window on:storage={async () => await authStore.init()} />
+<svelte:window
+  on:storage={async () => await authStore.init()}
+  on:popstate={updateRoute}
```
nmattia, line 12:
```suggestion
  // Browser back button has been clicked, we reflect the new browser url to the route
  on:popstate={() => routeStore.update({ path: routePath() })}
```

frontend/svelte/src/lib/components/Guard.svelte
```diff
@@ -1,8 +1,16 @@
 <script lang="ts">
   import { authStore } from "../stores/auth.store";
+  import { routeStore } from "../stores/route.store";
+  import { routePath } from "../utils/route.utils";
+
+  // Browser back button has been clicked, we reflect the new browser url to the route
+  const updateRoute = () => routeStore.update({ path: routePath() });
```
nmattia, line 7:
```suggestion
```

frontend/svelte/src/lib/utils/route.utils.ts
```diff
@@ -1,11 +1,61 @@
...
+  window.history &&
+  "pushState" in window.history &&
+  typeof window.history.pushState != "undefined";
+
+/**
+ * Find base href:
+ * // - https://something.com/ -> / (local development)
+ * // - https://something.com/v2/ => /v2/ (if the app is deployed on a server that serves from root ("/"), our case)
+ */
+export const baseHref = (): string => {
```
nmattia, line 58: why "href"? I've only ever seen it used in anchor tag attributes, don't even know what it stands for

frontend/svelte/src/lib/components/Guard.svelte
peterpeterparker: Does not work i think because you would add a comment within a component

An  Html `<!-- something -->` before the tag might work better?

frontend/svelte/rollup.config.js
```diff
@@ -88,6 +88,7 @@ export default {
             ? "https://qjdve-lqaaa-aaaaa-aaaeq-cai.nnsdapp.dfinity.network/"
             : "https://identity.ic0.app/")
       ),
+      "process.env.BASE_HREF": JSON.stringify(process.env.BASE_HREF || "/v2/"),
```
nmattia, line 91:
```suggestion
      // Specifies where the svelte app is loaded (typically "/" in local development and "/v2" in production)
      "process.env.BASE_HREF": JSON.stringify(process.env.BASE_HREF || "/v2/"),
```

frontend/svelte/src/lib/components/Nav.svelte
```diff
@@ -1,20 +1,24 @@
...
 </script>
 
 <nav>
-  {#each routes as { path, label }}
-    <a href={path} class:selected={currentPath === path} aria-label={label}
-      ><span>{label}</span></a
+  {#each routes as { context, label }}
+    <a
+      href={`${baseUrl}#/${context}`}
+      class:selected={currentContext === context}
```
nmattia, line 20: I think it'd really be a good idea to start typing the "Context" as an enum, comparing strings here is quite dangerous. Not in this PR though, just a thought.

frontend/svelte/src/lib/components/PrivateRoute.svelte
```diff
@@ -17,7 +18,12 @@
     }
 
     // Redirect to root, user needs to sign in
-    window.location.replace(`${appPath()}/?redirect=${routeContext()}`);
+    // We replace the url to get the redirect query params after successful sign in, to redirect user to the current page
+    // We do not navigate, i.e. push to browser history, because we do not want to stack this redirect in the back navigation
+    routeStore.replace({
+      path: "/",
+      query: `redirect=${routeContext()}`,
```
nmattia, line 25: might be a silly question: why don't we just go to e.g. `/#/accounts` directly instead of using `redirect=accounts`?

frontend/svelte/src/lib/utils/route.utils.ts
```diff
@@ -1,11 +1,61 @@
-// Get the current route from the location hash to lower case prefixed with a slash. Example: #/ICP => /#/icp
-export const routePath = (): string => `/${window.location.hash.toLowerCase()}`;
-
-export const appPath = (): string => {
-  const { pathname } = window.location;
-  return `${pathname.substring(0, pathname.lastIndexOf("/"))}`;
+/**
+ * The pathname and the hash regardless of the base href
```
nmattia, line 2:
```suggestion
 * The pathname and the hash without base href
```

frontend/svelte/src/lib/utils/route.utils.ts
```diff
@@ -1,11 +1,61 @@
...
+export const routePath = (): string => {
+  const base: string = baseHref();
+  const { pathname, hash } = window.location;
+  return `${pathname.replace(base, "/")}${hash.toLowerCase()}`;
 };
 
 // e.g. #/accounts => accounts
-export const routeContext = (): string =>
-  routePath().split("/").slice(2)[0].toLowerCase();
+export const routeContext = (): string => {
```
nmattia, line 11: I guess this module could have some unit tests, lots of assumptions here

frontend/svelte/src/routes/Accounts.svelte
```diff
@@ -1,12 +1,19 @@
...
   import Layout from "../lib/components/Layout.svelte";
-  import { onMount } from "svelte";
+  import type { Unsubscriber } from "svelte/types/runtime/store";
+  import { AuthStore, authStore } from "../lib/stores/auth.store";
+  import { onDestroy } from "svelte";
 
-  onMount(() => {
-    if (process.env.REDIRECT_TO_LEGACY) {
-      window.location.replace("/#/accounts");
+  // TODO: To be removed once this page has been implemented
```
nmattia, line 7: Reminder: unclear why `onMount` doesn't work here

frontend/svelte/src/lib/utils/route.utils.ts
peterpeterparker: Yep generally speaking this utils needs test but jest was added after the pr was created.
Later on in separate PRs as for the existing components not corvered.

frontend/svelte/src/lib/components/PrivateRoute.svelte
peterpeterparker: In an url like `http://localhost:8080/?redirect=/#/accounts` the `/#/accounts` is still interpreted as the `hash` of the url by the browser. We do not want that, we want the `redirect=xxxx` to be a search param.

Note: above code `routeStore.replace` is new but the feature itself,`redirect=${routeContext()}` etc., exists already in `main` branch

frontend/svelte/src/lib/components/Nav.svelte
peterpeterparker: Yes, well noted. We have some more pages to add in next weeks ("account details" for example). We can do that at the same time.

frontend/svelte/src/lib/components/Guard.svelte
peterpeterparker: Resolve today together in a call with @nmattia

frontend/svelte/src/lib/components/PrivateRoute.svelte
nmattia: Gotcha. My question was not "why don't we redirect to `/?redirect=/#/accounts`" but "why don't we redirect to `/#/accounts`" ?

peterpeterparker: That was my first idea too but with all the redirects scenario we have to handle, if we keep `/#/accounts` in case of redirect to sign in, there was one scenario in which I got stuck on a blank page. With query params, no issue.

nmattia: Ok, gotcha. Would be good to add a note about that as well, but I'm guessing it's gonna go away anyway when we get rid of flutter!

peterpeterparker: It will ease for sure. Again not new to this PR, if a must, can / should be treated in a separate new PR I would say.

nmattia: Ah yes, definitely! I just got curious here, didn't think of asking the first time around. Thanks for the clarification!

frontend/svelte/src/routes/Accounts.svelte
peterpeterparker: You are right, with last improvements we do not need any changes there anymore. Reverted as in main branch.

nmattia (approved):
LGTM! Would also be great to have a comment explaining what's happening with the base href

peterpeterparker:
Comment about script base href added

frontend/svelte/scripts/copy.index.mjs
```diff
@@ -2,6 +2,11 @@
 
 import { readFileSync, writeFileSync } from "fs";
 
+/**
+ * Rollup takes care of the JS and CSS bundles. Here we copy the index.html from the source to the output folder.
+ * At the same time, we also update the <base /> reference (see comment below).
+ * By pre-rendering this information, we make it static and guarantee it will be set when the app - the JS code - access it at runtime.
```
nmattia, line 8:
```suggestion
 * By pre-rendering this information, we make it static and guarantee it will be set when the app - the JS code - accesses it at runtime.
```

PR #343, opened 2022-01-11
test: jest

peterpeterparker:
# Motivation

In addition to e2e tests, we also need unit and dom tests for the components and functions.

# Changes

- add [jest](https://jestjs.io)
- add [svelte-testing-library](https://testing-library.com/docs/svelte-testing-library/intro/) to setup env best practices and render components for testing purpose
- add a first test for the `Header` component

# Note

I created the first test in a sub-folder `__tests__` but got no preferences about the file structure. If you rather like to have the tests elsewhere, like next to the components or functions, let me know. (=> we decided to use `src/tests`).


frontend/svelte/package.json
```diff
@@ -8,7 +8,10 @@
     "dev": "npm run import-assets && rollup -c -w",
     "start": "sirv public",
     "check": "svelte-check --tsconfig ./tsconfig.json",
-    "format": "prettier --write --plugin-search-dir=. ."
+    "format": "prettier --write --plugin-search-dir=. .",
+    "test": "jest",
+    "test:watch": "npm run test -- --watchAll",
+    "test:coverage": "jest --coverage"
```
nmattia, line 14: I've never really found this to be useful; what's your experience?

frontend/svelte/src/__tests__/lib/components/Header.spec.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * @jest-environment jsdom
```
nmattia, line 2: what is this?

frontend/svelte/svelte.config.test.cjs
```diff
@@ -0,0 +1,3 @@
+const preprocess = require("svelte-preprocess");
```
nmattia, line 1: What is this?

nmattia (commented):
This looks great! 

Yeah, let's move `__tests__` to `frontend/svelte/tests`, so that people know where to look. WDYT?

frontend/svelte/svelte.config.test.cjs
peterpeterparker: Compiles tests from ts to js. Same as in `rollup.config.js` to bundle the app.

Since it's already here, it's easier to use it than using and setting babel only for the tests purpose.

frontend/svelte/src/__tests__/lib/components/Header.spec.ts
peterpeterparker: it tells the test to use jsdom i.e. to emulate the dom. without it the test is run as a node test.

frontend/svelte/svelte.config.test.cjs
nmattia: Thanks! Can you add a note about this?

frontend/svelte/src/__tests__/lib/components/Header.spec.ts
nmattia: Gotcha, thanks!

frontend/svelte/package.json
peterpeterparker: in previous gig the ci was set to fail if the test coverage didn't reach 80%
does not necessary means much but could be used to force writing tests

frontend/svelte/svelte.config.test.cjs
peterpeterparker: Done.

frontend/svelte/package.json
nmattia: then I suggest we create a ticket, we can then discuss if we want a CI check for test coverage or not; in the meantime we can remove this. What do you think?

peterpeterparker: Done.

frontend/svelte/jest-setup.ts
```diff
@@ -0,0 +1,5 @@
+import "@testing-library/jest-dom";
```
nmattia, line 1: noob question: what is this file and where does `util` come from?

nmattia (commented):
Looking great! One more thing; can you add a line to [run the tests on CI](https://github.com/dfinity/nns-dapp/blob/6b04d6241d735ee0bd23bbf3bf932f8e83217801/.github/workflows/checks.yml#L173)?

frontend/svelte/jest-setup.ts
peterpeterparker: Added a comment about `util` and tests added to CI.

nmattia (approved):
LGTM, thanks!

PR #345, opened 2022-01-11
EXC-822: Add principal to hardware wallet accounts page.

ielashi:
This is necessary if, for example, you want to fill out a KYC form. Only hardware wallet accounts are effected by this change.

Example:

Screen Shot 2022-01-11 at 16 59 46


anishh2003:
> @anishh2003 Maybe you can help with making the UI a bit more palatable? :)



> more

I have added in the UI changes

anishh2003 (approved):

PR #352, opened 2022-01-12
EXC-828: Upgrade agent-js to 0.10.2

ielashi:
This PR is the 2nd iteration of 26989ab. It fixes two issues related to type conversions:

1. **Converting Buffer to ArrayBuffer.** Instead of (incorrectly) using `.buffer`, we now use `bufferToArrayBuffer`.
2. **Converting Uint8Array to ArrayBuffer.** Instead of (incorrectly) using `.buffer`, we now use `uint8ArraytoArrayBuffer`.


ielashi:
> I'm not aware superficially of any reason that you cannot access an ArrayBuffer from a Uint8Array via .buffer

See https://stackoverflow.com/questions/37228285/uint8array-to-arraybuffer for a counter-example (second answer).

hpeebles (approved):

PR #354, opened 2022-01-13
Include followee suggestions from governance canister

anishh2003:


frontend/dart/lib/ui/neurons/following/topic_followeees_widget.dart
```diff
@@ -77,48 +78,57 @@ class TopicFolloweesWidget extends StatelessWidget {
...
-                          child: Text(
-                              FolloweeSuggestion.followerSuggestions
-                                      .firstWhereOrNull(
-                                          (element) => element.id == e)
-                                      ?.name ??
-                                  e,
-                              style: context.textTheme.bodyText2),
+
+    return FutureBuilder(
+        future: context.icApi.followeeSuggestions(),
```
hpeebles, line 83: I don't think you need this here. You only need to get the suggestions if they click 'Add followee'.

frontend/dart/lib/ui/neurons/following/followee_suggestions.dart
```diff
@@ -114,8 +120,15 @@ class FolloweeSuggestion {
 
   FolloweeSuggestion(this.name, this.text, this.id);
 
-  static List<FolloweeSuggestion> followerSuggestions = [
-    FolloweeSuggestion("DFINITY Foundation", "", "27"),
-    FolloweeSuggestion("Internet Computer Association", "", "28"),
-  ];
+  // static List<FolloweeSuggestion> followerSuggestions = [
```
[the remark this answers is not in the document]
anishh2003: this has been removed now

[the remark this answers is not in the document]
hpeebles: I guess you mean you've made the change locally... because this PR hasn't been updated...

frontend/dart/lib/ui/neurons/following/topic_followeees_widget.dart
hpeebles: We had a call... this is needed to populate the names of the neurons you are following.

frontend/ts/src/canisters/governance/Service.ts
```diff
@@ -146,6 +147,34 @@ export default class Service implements ServiceInterface {
       : null;
   };
 
+  public listKnownNeurons = async (
+    certified: boolean
+  ): Promise<Array<KnownNeuron>> => {
```
ielashi, line 152: New endpoints should be added to `nns-js` IMO, not here. That way, the Svelte app can reuse it when the time comes.

frontend/dart/lib/ic_api/web/web_ic_api.dart
```diff
@@ -561,6 +566,19 @@ class PlatformICApi extends AbstractPlatformICApi {
         (_) => {print("[${DateTime.now()}] Syncing canisters complete.")});
   }
 
+  Future<List<FolloweeSuggestion>> followeeSuggestions() {
+    return this.suggestedFolloweesCache!.get();
+  }
+
+  Future<List<FolloweeSuggestion>> _followeeSuggestions(
+      [bool certified = true]) async {
```
ielashi, line 574: Are we using this flag anywhere, or are we always calling this with `certified` set to true?

rs/src/main.rs
```diff
@@ -312,3 +325,16 @@ pub enum GetAccountResponse {
...
+    id: NeuronId,
+    name: String,
+    description: Option<String>,
+}
+
+impl KnownNeuron {
+    pub fn new(id: NeuronId, name: String, description: Option<String>) -> KnownNeuron {
+        KnownNeuron { id, name, description }
+    }
+}
```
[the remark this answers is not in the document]
hpeebles: These backend changes are now all gone.

frontend/dart/lib/ic_api/web/web_ic_api.dart
```diff
@@ -28,6 +30,7 @@ class PlatformICApi extends AbstractPlatformICApi {
   AccountsSyncService? accountsSyncService;
   NeuronSyncService? neuronSyncService;
   ProposalSyncService? proposalSyncService;
+  SuggestedFolloweesCache? suggestedFolloweesCache;
```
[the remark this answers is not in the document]
hpeebles: It's because the fields are initialized within the `buildServices` method rather than within a ctor. So the field is undefined when the type is first constructed. This is why all of the fields are optional.

We should probably instead have a static method which builds the services and then calls into a private ctor. That way the fields wouldn't need to be optional. But with the Svelte version looming I don't think this matters.

frontend/ts/src/canisters/governance/Service.ts
hpeebles: This is now being done in parallel.

bitdivine (approved):
Verified that the GUI works as advertised.  Was unable to test adding new followees due to limited time but it appears that the "worst case" scenario is the status quo.

PR #357, opened 2022-01-14
Cache docker build

nmattia:
This adds caching to our docker build through [`buildx` Github Actions support](https://github.com/moby/buildkit#github-actions-cache-experimental). This saves 75% of the build time (22mn -> 5mn). The `.dockerignore` is also updated to fix some issues when local files ending up polluting the build.

Big, big thanks @crazy-max for your help!

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -24,12 +24,15 @@ jobs:
       # This is fine as long as this layer is rarely rebuilt. Moreover we make
       # sure to have exact cache hits (i.e. no restore-keys matching) to ensure
       # the cache is not uploaded (unless there is a cache miss).
-      # NOTE: about files involved
+      #
+      # The `hashFiles` should include every single file involved in the
+      # "builder" build, otherwise the action will fetch the cache but start
+      # from scratch nonetheless, never uploading the build output.
       - name: Cache docker layers
         uses: actions/cache@v2
         with:
           path: /tmp/.buildx-cache
-          key: ${{ runner.os }}-buildx-${{ hashFiles( 'Dockerfile', 'Cargo.lock', 'Cargo.toml', 'rs/Cargo.toml', 'rs/nns_functions_candid_gen/**', 'dfx.json' ) }}
+          key: ${{ runner.os }}-buildx-${{ hashFiles( '.dockerignore', 'Dockerfile', 'Cargo.lock', 'Cargo.toml', 'rs/Cargo.toml', 'rs/nns_functions_candid_gen/**', 'dfx.json' ) }}
 
       - name: Build base Docker image
         uses: docker/build-push-action@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -11,7 +11,75 @@ jobs:
...
+          tags: nns-dapp-builder
+          file: Dockerfile
+          cache-from: type=gha,scope=builder-stage
+          # TODO: directly feed image from first build
+          # NOTE: Here we ensure we only export the cache if there was no cache
+          # hit.  Otherwise, we export many GBs (a few minutes) to never
+          # actually use the exported data.
+          # NOTE: the empty string must be at the end, otherwise `''` is
+          # considered falsy and is never used
+          cache-to: ${{ (! steps.docker-cache.outputs.cache-hit) && 'type=gha,scope=cached-stage,mode=max' || '' }}
```
crazy-max, line 55: Not needed if using `type=gha`

nmattia: thanks!

.github/workflows/docker-build.yaml
```diff
@@ -11,7 +11,75 @@ jobs:
...
+          context: .
+          push: false # otherwise docker tries to push to dockerhub
+          tags: nns-dapp
+          load: true # effectively makes the image usable in 'docker run'
+          file: Dockerfile
+          cache-from: type=gha,scope=cached-stage
+
       - run: mkdir out/
       - run: docker run --rm --entrypoint cat nns-dapp /nns-dapp.wasm > out/nns-dapp.wasm
       - run: docker run --rm --entrypoint cat nns-dapp /assets.tar.xz > out/assets.tar.xz
```
crazy-max, line 85: IIUC you want to output some artifacts created in build nns-app? If that's the case you can remove the `load` and simply output the desired files from a `scratch` stage using `outputs: ./out` (yes you can output to a local folder not just an image).

nmattia: That's AWESOME! thank you so much!

.github/workflows/docker-build.yaml
```diff
@@ -11,10 +11,32 @@ jobs:
...
+          cache-from: type=gha,scope=cached-stage
+          cache-to: type=gha,scope=cached-stage,mode=max
+          outputs: type=cacheonly
+          target: builder
+
+      - name: Build nns-dapp Docker image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: false # otherwise docker tries to push to dockerhub
```
crazy-max, line 34: you can remove this `push` is a shorthand for `--output=type=registry` and as you specify `outputs: ./out` it will not been set anyway: https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#push

nmattia: Thanks. They default to `false` anyway, just wanted to make it clear. But you're right, doesn't add much. Removed!

Dockerfile
```diff
@@ -46,12 +48,23 @@ ENV PATH=/flutter/bin:$PATH
...
 
+# Pre-build all cargo dependencies. Because cargo doesn't have a build option
+# to build only the dependecies, we pretend that our project is a simple, empty
+# `lib.rs`. Then we remove the dummy source files to make sure cargo rebuild
+# everything once the actual source code is COPYed (and e.g. doesn't trip on
+# timestamps being older)
+COPY Cargo.lock .
+COPY Cargo.toml .
+COPY rs/Cargo.toml rs/Cargo.toml
+COPY rs/nns_functions_candid_gen ./rs/nns_functions_candid_gen
```
bitdivine, line 59: Is the idea that the API will change only very rarely?  Fair enough.

bitdivine (approved):

Dockerfile
nmattia: Yes!

PR #358, opened 2022-01-14
Fix delegation ttl to 30 min

frederikrothenberger:
Fixes https://dfinity.atlassian.net/browse/FOLLOW-449


nmattia (approved):
Rubber stamp, no context

robin-kunzler:
Thanks for fixing this @frederikrothenberger ! As mentioned, if it would be easy to test this somehow, that would be great. But I leave it up to you to decide.

PR #359, opened 2022-01-14
Add non-Docker build to CI

nmattia:
This adds a non-Docker, full build which is more easily cachable than
the Docker build. This is also much faster than a Docker build (~5mn instead of ~22mn).

This will be used by the end-to-end tests
(embryo here: https://github.com/dfinity/nns-dapp/pull/341)

This will be used by or reuse builds artefacts from checks.yml as soon as I have time to link the two, see https://dfinity.atlassian.net/browse/L2-191.


nmattia:
@ielashi probably not, unless we can have a really robust, reliable and reproducible setup without it. But we'll see! Thanks for the review!

PR #360, opened 2022-01-17
Create build script for dart code

nmattia:
This makes the top-level `build.sh` more readable and is more in line
with e.g. `frontend/ts`'s `build.sh`.

### DIFF
--- a/frontend/dart/build.sh
+++ b/frontend/dart/build.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+#
+# Build the flutter app
+
+set -euo pipefail
+
+MYPATH="$(
+  cd -- "$(dirname "$0")" >/dev/null 2>&1
+  pwd -P
+)"
+
+echo "Assuming dart codebase path is $MYPATH"
+cd "$MYPATH"
+
+if [[ $DEPLOY_ENV = "mainnet" ]]; then
+  flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=FLUTTER_WEB_CANVASKIT_URL=/assets/canvaskit/
+else
+  # For all networks that are not mainnet, build with the staging config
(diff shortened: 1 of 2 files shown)


frontend/dart/build.sh
```diff
@@ -0,0 +1,23 @@
...
+cd "$MYPATH"
+
+if [[ $DEPLOY_ENV = "mainnet" ]]; then
+  flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=FLUTTER_WEB_CANVASKIT_URL=/assets/canvaskit/
+else
+  # For all networks that are not mainnet, build with the staging config
+  flutter build web --web-renderer html --release --no-sound-null-safety --pwa-strategy=none --dart-define=DEPLOY_ENV=staging
+fi
+
+# Remove the random hash from flutter output
```
bitdivine, line 22: This is a cache buster. If you redeploy and the flutter_service_worker.js is still in an end user's browser cache, or any intermediate http cache, they might have an unhappy time.  Why do you want to remove this?  For consistency?  Maybe use the git commit instead?

nmattia: I didn't do that, I just moved the code around

nmattia: as to why: pretty sure it's in order to have a fixed sha256

bitdivine: I see.  It seems to be new in this PR but was actually introduced by @nomeata in:
```
commit fc379b34ad2667443d2e24e231e4bd8c76bf97f6
Author: Joachim Breitner <mail@joachim-breitner.de>
Date:   Sat May 8 17:00:11 2021 +0200

    Remove random hash from flutter output
```
Let's keep it for now but revisit it as needed.  The IC does not have much of a story around caching yet.  Once it has decent caching this might well become an issue.

bitdivine (approved):

PR #362, opened 2022-01-18
refactor: auth store

peterpeterparker:
# Motivation

Make @nmattia happy 😉.

Joke beside, this PR removes the `signedIn` flag in the state management to handle the auth state with the `indentity` only. 
Using a single flag is less prone to issue for such an important information.

In addition, it refactors a couple of TODO we still had to implement.

# Changes

- rename function `init` to `sync`
- only and only if user is authenticated (`await authClient.isAuthenticated()`) set principal in store i.e. ignore anonymous principal created by auth client
- create new function to check if signed in based on principal
- remove `export` in `export const initAuthStore = () => {`
- add a comment in store to explain “how it works”
- add tests


nmattia (approved):
LGTM, thanks! Looks great

PR #363, opened 2022-01-18
feat: i18n

peterpeterparker:
# Motivation

Simple solution to localize texts of the app by using a store (without 3rd party dependencies).

# Changes

- add a new `readable` store that track the translations
- add a json file `en.json` that contains the texts / translations
- create a script that generates the TypeScript definitions (interfaces) from the English json file (type support available when using the translations)
- update components to use translations


frontend/svelte/scripts/i18n.types.mjs
```diff
@@ -0,0 +1,37 @@
...
+  const lang = `lang: Languages;`;
+
+  const main = `\n\ninterface I18n {${lang}${data
+    .map((i) => `${i.key}: ${i.name};`)
+    .join("")}}`;
+  const interfaces = data
+    .map((i) => `\n\ninterface ${i.name} {${i.properties.join("")}}`)
+    .join("");
+
+  const output = prettier.format(`${interfaces}${main}`, { parser: "babel" });
```
nmattia, line 32: very sweet! Can you also add a comment saying that the file was auto-generated by the script?

nmattia (commented):
Very nice! How would this work once we have more than one language?

peterpeterparker:
@nmattia for multiple languages, we would have to:

1. modify the store from a `readable` store to a `writable` store
2. add a function that lazy load the selected language and apply the values to the store

nmattia:
Thanks! So we'd have to wait a bit before displaying any text? or would we first display in english and then reload the store once it's been loaded? Either way I like the approach, just wondering how much async overhead there would be for multiple languages

frontend/svelte/scripts/i18n.types.mjs
nmattia: Ah, I meant in `i18n.types.mjs`. Isn't the other file wiped every time the command is run?

peterpeterparker:
We would use the function to:

1. load the default language at startup
2. change the language on the fly if user select another language

Yes the store will be per default 'en' but it will be fast enough to not notice the language load at startup.

For example in Stencil:

- store: https://github.com/deckgo/deckdeckgo/blob/main/studio/src/app/stores/i18n.store.ts
- startup loader: https://github.com/deckgo/deckdeckgo/blob/805c59b8e6c6f6821d7ee0ad9eadc4e10f24e0ef/studio/src/app/services/lang/lang.service.ts#L19

nmattia (approved):
LGTM!

PR #364, opened 2022-01-18
Script to deploy from CI artefact

bitdivine:
# Motivation
People wish to deploy to a testnet for reasons including, but not limited to:
- Not being able to build locally, due to Mac M1 chips
- Not wanting to waste time building when CI has already built a wasm file.

# Changes
- Added a script to deploy a wasm file downloaded from CI.

Note: github assets cannot be downloaded by anonymous users.  In the envisaged use case, a human user will be choosing a build artifact so they might as well click "Download" while they are about it.

Note: It is assumed that the download directory is ~/Downloads.  This is the default on Mac and Linux.  We can support other use cases as the need arises.

# Testing
I  have used the script.  It deployed a canister, as it should.

Also run by @peterpeterparker on an M1 mac.  Worked in a clean repo, but needed the wasm build directory to be created.  I have added a line to create this to the script.

Please run the script as well.  Here is a sample CI build to download: https://github.com/dfinity/nns-dapp/actions/runs/1712880217


scripts/deploy-ci-to-testnet
```diff
@@ -0,0 +1,23 @@
...
+# Get the asset
+DOWNLOADS_DIR="$HOME/Downloads"
+echo
+read -n 1 -p "Please download the 'Backend wasm module.zip' docker build artifact from github to $DOWNLOADS_DIR and press enter...  "
+
+DOWNLOAD_FILENAME="$(ls -t "$DOWNLOADS_DIR" | grep -E '^Backend wasm module( [(][0-9]+[)])*.zip$' | head -n1)"
+DOWNLOAD_PATH="$DOWNLOADS_DIR/$DOWNLOAD_FILENAME"
+WASM_PATH="$(jq -rc '.canisters["nns-dapp"].wasm' dfx.json)"
+WASM_FILENAME="nns-dapp.wasm"
+TMPDIR="$(mktemp -d)"
```
nmattia, line 14: Use [`TMPDIR`](https://en.wikipedia.org/wiki/TMPDIR) as a var name is probably not a good idea

nmattia (commented):
The script isn't very robust (relies on specific paths and things like the unfortunate `DEPLOY_ENV`) and requires manual interaction anyway. I think a list of steps to run would be much better

scripts/deploy-ci-to-testnet
bitdivine: I can namespace it, although that dir is the tmpdir for everything in this script.  Do you have a specific example of what might go wrong?

nmattia: Better thanks! `TMPDIR` may be `export`ed by default and is the filesystem's default temporary directory. Tools that you use here (like `unzip`, etc) may read that and use the directory for their own, private temporary files

bitdivine: Exactly, default is the key and it is standard practice, indeed required, to override TMPDIR when one wants a program to put its tmp files in a different location.  So when that is the intention, that is correct.  Here it does't really matter either way but I fail to see what would be wrong if we really wanted all the tmpfiles to be in this directory, not just the wasm artifact.

bitdivine:
Regarding DEPLOY_ENV - that is the canonical way, and indeed the only way, of deploying this repo without building.  That can be changed, perhaps, but that is not the subject of this PR so it would be inappropriate to address it here.

.github/workflows/docker-build.yaml
```diff
@@ -28,23 +28,36 @@ jobs:
...
           target: builder
 
-      - name: Build nns-dapp Docker image
+  assets:
+    needs: builder
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+       include:
+         - DEPLOY_ENV: "testnet"
```
nmattia, line 37: I doubt this will work since `dfx` will try to look up the canister ID (which hasn't been created in the docker image)

scripts/deploy-ci-to-testnet
```diff
@@ -0,0 +1,39 @@
...
+WASM_PATH="$(jq -rc '.canisters["nns-dapp"].wasm' dfx.json)"
+WASM_FILENAME="nns-dapp.wasm"
+UNZIP_TMPDIR="$(mktemp -d)"
+mkdir -p "$(dirname "$WASM_PATH")"
+cp "$DOWNLOAD_PATH" "$UNZIP_TMPDIR"
+pushd "$UNZIP_TMPDIR"
+unzip "$DOWNLOAD_FILENAME"
+popd
+mv "$UNZIP_TMPDIR/$WASM_FILENAME" "$WASM_PATH"
+rm -fr "${UNZIP_TMPDIR:-this-is-not-the-dir-you-are-looking-for}"
```
nmattia, line 35: How could `UNZIP_TMPDIR` not be set? i.e. when would we `rf -fr this-is-not...`?

scripts/deploy-ci-to-testnet
```diff
@@ -0,0 +1,39 @@
...
+DEPLOY_ENV=nobuild dfx deploy --no-wallet --network testnet
+
+set +x
```
nmattia, line 38: I don't see a `set -x`, what does this do?

nmattia (approved):
I think there are a few things that are not ideal:
* The dockerfile changes look good (module the potential issue with `dfx` not finding the canister) but I don't think this is the right way about. We should probably invest some time (not much I reckon) getting rid of this `DEPLOY_ENV` and make it a post-build config parameter.
* I am not sure the `deploy-ci-to-testnet` script is all that useful. It requires manual interaction, relies on artefacts that in general are GCed after a week, and basically just unzips a downloaded file. If we wanted something like this I'd rather push actual releases to GitHub and have no manual step.

All that being said, still gonna rubber stamp this and will probably fix it later on

.github/workflows/docker-build.yaml
bitdivine: It won't - this needs the canister_ids.json mentioned.  There is no canister-ds.json commited in the repo so that developers don't clobber each other's work, and canister_ids.json is in the gitignore.  There used to be a canister_ids.json committe din git but that was removed, in July if I recall correctly.

What we can do is to have a `canister_ids.default.json` and use that e.g. here.  It is also useful if other testnets that just want to use nns, not develop it, use the default canister ID.  I can add that here or wait for your change.

Added here but can revert.

scripts/deploy-ci-to-testnet
bitdivine: When a developer tinkers with this script.  Much better to tell them that they have a bug than potentially wipe their system.  That said, I use that by habit and in this case, as there is no stem so this would reduce to `rm -fr ""` which is safe, so the additional safety mechanism isn't needed.  I'll drop it.

scripts/deploy-ci-to-testnet
bitdivine: Good point.  I had a set -x but dropped it.

.github/workflows/docker-build.yaml
```diff
@@ -38,6 +38,8 @@ jobs:
          - DEPLOY_ENV: "mainnet"
     steps:
       - uses: actions/checkout@v2
+      - name: Create canister_ids.json
+        run: cp canister_ids.default canister_ids.json
```
nmattia: This is too much of a hack; let's merge https://github.com/dfinity/nns-dapp/pull/365 and then you can set `OWN_CANISTER_ID` from the Dockerfile

nmattia (changes requested):

.github/workflows/docker-build.yaml
bitdivine: Switched to OWN_CANISTER_ID - please review.

nmattia: Thanks!

scripts/deploy-ci-to-testnet
```diff
@@ -0,0 +1,38 @@
...
+  echo
+  echo "Please download a testnet backend wasm module from:  https://github.com/dfinity/nns-dapp/actions/workflows/docker-build.yaml"
+  read -rn 1 -p "Press enter when done .... "
+  echo
+  read -rn 1 -p "What is your download dir? (default '$DOWNLOADS_DIR'): " ALT_DOWNLOADS_DIR
+  DOWNLOADS_DIR="${ALT_DOWNLOADS_DIR:-$DOWNLOADS_DIR}"
+
+  # Gets the most recent 'Backend wasm module' download:
+  # shellcheck disable=SC2010 # we really do want this, not globbing.
+  DOWNLOAD_FILENAME="$(ls -t "$DOWNLOADS_DIR" | grep -E '^Testnet backend wasm module( [(][0-9]+[)])?.zip$' | head -n1)"
```
nmattia, line 21:
```suggestion
  DOWNLOAD_FILENAME="$(ls --sort=time "$DOWNLOADS_DIR" | grep -E '^Testnet backend wasm module( [(][0-9]+[)])?.zip$' | head -n1)"
```

note: not sure how posix `-t` (and `--sort`) is

scripts/deploy-ci-to-testnet
```diff
@@ -0,0 +1,38 @@
...
+  DOWNLOAD_PATH="$DOWNLOADS_DIR/$DOWNLOAD_FILENAME"
+  echo "Sourcing WASM from: $DOWNLOAD_PATH"
```
nmattia, line 23:
```suggestion
```

scripts/deploy-ci-to-testnet
```diff
@@ -0,0 +1,38 @@
...
+fi
+
```
nmattia, line 25:
```suggestion
  echo "Sourcing WASM from: $DOWNLOAD_FILENAME"
```

scripts/deploy-ci-to-testnet
```diff
@@ -0,0 +1,38 @@
...
+WASM_PATH="$(jq -rc '.canisters["nns-dapp"].wasm' dfx.json)"
+WASM_FILENAME="nns-dapp.wasm"
+UNZIP_TMPDIR="$(mktemp -d)"
+mkdir -p "$(dirname "$WASM_PATH")"
+cp "$DOWNLOAD_PATH" "$UNZIP_TMPDIR"
+pushd "$UNZIP_TMPDIR"
+unzip "$DOWNLOAD_FILENAME"
+popd
+mv "$UNZIP_TMPDIR/$WASM_FILENAME" "$WASM_PATH"
```
nmattia, line 34: note: not sure about this but there's a chance this will mess with cargo (e.g. it may refuse to rebuild if the mtime of the downloaded wasm is more recent than that of the local files)

nmattia (approved):

PR #366, opened 2022-01-18
Add proxy

nmattia:
This adds a proxy that gives each local canister its own port and funnels the requests to the actual running replica. This is necessary, otherwise redirects between the svelte app and the flutter code fail when served from `dfx` and a local replica, since the local replica expects the `canisterId` query param to be set.

I'll sync with @krpeacock to figure out if this should be `dfx`'s default behavior, in the meantime this works just fine.

---

Used by https://github.com/dfinity/nns-dapp/pull/367

### DIFF
--- a/proxy/index.ts
+++ b/proxy/index.ts
@@ -67,7 +67,7 @@ const parsePortMappings = (args: string[]): Record<CanisterId, Port> | string =>
 //      5XX)
 // * listens on the specified port and proxies to
 //      '<replicaHost>/?canisterId=<canisterId>'
-const mkApp = (replicaHost: string, port: number, canisterId: string) => {
+const mkApp = ({replicaHost, port, canisterId} : { replicaHost: string, port: number, canisterId: string}) => {
   const app = express();
 
   // could use morgan's .token() thingy but really not worth it here
@@ -144,5 +144,5 @@ if (Object.keys(parsed.canistersToPorts).length == 0) {
 for (let canisterId in parsed.canistersToPorts) {
   let port = parsed.canistersToPorts[canisterId];
   console.log(`Forwarding ${port} to ${parsed.replicaHost}/?canisterId=${canisterId}`);
-  mkApp(parsed.replicaHost, port, canisterId);
+  mkApp({ replicaHost: parsed.replicaHost, port, canisterId });
 }
--- a/proxy/index.ts
+++ b/proxy/index.ts
@@ -120,7 +120,7 @@ const mkApp = (replicaHost: string, port: number, canisterId: string) => {
   });
 };
 
-const usage = 'USAGE: proxy --repilca-host http://... [<canister-id>:<port>]';
+const usage = 'USAGE: proxy --replica-host http://... [<canister-id>:<port>]';
 
 const args = process.argv.slice(2);
 
--- a/proxy/index.ts
+++ b/proxy/index.ts
@@ -27,7 +27,6 @@ const parseArgs = (args: string[]): CLIArguments | string => {
     }
     let [_, replicaHost] = replicaArgs;
 
-
     // Parse the rest of the args as canister/port mappings
     const canistersToPorts = parsePortMappings(args);
 
@@ -62,7 +61,6 @@ const parsePortMappings = (args: string[]): Record<CanisterId, Port> | string =>
 
     return canistersToPorts;
 }
-
 
 // An app that:
 // * logs the requests (green for 2XX, blue for 3XX, yellow for 4XX and red for
--- a/proxy/index.ts
+++ b/proxy/index.ts
@@ -96,15 +96,15 @@ const mkApp = ({replicaHost, port, canisterId} : { replicaHost: string, port: nu
     createProxyMiddleware({
       target: replicaHost,
       pathRewrite: (pathAndParams, req) => {
-        let queryParamsString = "?";
+        let queryParamsString = `?`;
 
         let [path, params] = pathAndParams.split("?");
 
-        if (params === undefined) {
-          queryParamsString = `?canisterId=${canisterId}`;
-        } else {
-          queryParamsString = `?${params}&canisterId=${canisterId}`;
+        if (params) {
+          queryParamsString += `${params}&`;
         }
+
+        queryParamsString += `canisterId=${canisterId}`;
 
         return (path += queryParamsString);
       },
--- a/proxy/index.ts
+++ b/proxy/index.ts
@@ -2,16 +2,74 @@ import express from "express";
 import morgan from "morgan";
 import { createProxyMiddleware } from "http-proxy-middleware";
 
-// Use value defined in dfx.json (.networks.local.bind)
-const REPLICA_HOST = "http://localhost:8080";
+type CanisterId = string;
 
-// Use value defined in .dfx/local/canister_ids.json (.nns-dapp.local) as
-// canister ID
-const canistersToPorts: any = {
-  "rwlgt-iiaaa-aaaaa-aaaaa-cai": 8086,
-};
+type Port = number;
 
-const mkApp = (port: number, canisterId: string) => {
+interface CLIArguments {
+    replicaHost: string,
+    canistersToPorts: Record<CanisterId, Port>,
+}
+
+// Parse all the arguments, returning an error string if parsing failed.
+const parseArgs = (args: string[]): CLIArguments | string => {
+
+    // Ensure '--replica-host http://...' was specified
+    const replicaArgIndex = args.indexOf("--replica-host");
+    if(replicaArgIndex == -1) {
+        return 'Please specify --replica-host';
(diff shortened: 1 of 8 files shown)


proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+      }\x1b[0m ${req.method} ${req.originalUrl} -> ${req.url} `;
+    })
+  );
+
+  app.all(
+    "*",
+    createProxyMiddleware({
+      target: replicaHost,
+      pathRewrite: (pathAndParams, req) => {
+        let queryParamsString = "?";
```
nmattia, line 99: not necessary, it's being reset below

proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    replicaHost: string,
+    canistersToPorts: Record<CanisterId, Port>,
+}
+
+// Parse all the arguments, returning an error string if parsing failed.
+const parseArgs = (args: string[]): CLIArguments | string => {
+
+    // Ensure '--replica-host http://...' was specified
+    const replicaArgIndex = args.indexOf("--replica-host");
+    if(replicaArgIndex == -1) {
```
peterpeterparker, line 19: Code formatting. prettier should probably be applied.

proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+type CanisterId = string;
+
+type Port = number;
+
+interface CLIArguments {
+    replicaHost: string,
+    canistersToPorts: Record<CanisterId, Port>,
+}
+
+// Parse all the arguments, returning an error string if parsing failed.
```
peterpeterparker, line 14: Any particular reason why you return string for errors and not throw errors?

Because it's a standalone script, I think it would be easier to throw errors and wrap the main call with a single try/catch.

I can also make the code cleaner because such piece of code as `if(typeof ... === "string") {` can then be dropped.

proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    if(replicaArgIndex == -1) {
+        return 'Please specify --replica-host';
+    }
+
+    // Remove the '--replica-host http://...' from the args and store the value
+    const replicaArgs = args.splice(replicaArgIndex, 2);
+    if(replicaArgs.length != 2) {
+        return 'No value for --replica-host';
+    }
+    let [_, replicaHost] = replicaArgs;
```
peterpeterparker, line 28: A few `let` can actually be `const`. This particular one and others in the script.

proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    // Ensure '--replica-host http://...' was specified
+    const replicaArgIndex = args.indexOf("--replica-host");
+    if(replicaArgIndex == -1) {
+        return 'Please specify --replica-host';
+    }
+
+    // Remove the '--replica-host http://...' from the args and store the value
+    const replicaArgs = args.splice(replicaArgIndex, 2);
+    if(replicaArgs.length != 2) {
```
peterpeterparker, line 25: should we test the `replicaHost` against `http://...` or another pattern?

if the value is omitted in the args, does it fails at a later stage?

correct cmd: `----replica-host https://....com rwajt-...:8086`
incorrect cmd: ----replica-host rwajt-...:8086

proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    if(typeof canistersToPorts === "string") {
+        return canistersToPorts;
+    } else {
+        return { replicaHost, canistersToPorts };
+    }
+}
+
+// Parse the canister ID to port mappings (rwajt-...:8086 rdm6x-...:443 ...)
+const parsePortMappings = (args: string[]): Record<CanisterId, Port> | string => {
+    let canistersToPorts: Record<CanisterId,Port> = {};
```
peterpeterparker, line 42: if you switch to "throw error" then `.reduce` can be use here

```
return args.reduce((acc: Record<CanisterId, Port>, arg: string) => {
        const tokens = arg.split(":");

        if (tokens.length != 2) {
            throw Error(`Could not parse '${arg}'`);
        }

        const [canisterId, portStr] = tokens;

        const port: number = parseInt(portStr);

        if(isNaN(port)) {
            throw Error(`Could not parse port '${portStr}' as number`);
        }

        return {
            ...acc,
            [canisterId]: port
        }
    }, {});
```

proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    return canistersToPorts;
+}
+
+// An app that:
+// * logs the requests (green for 2XX, blue for 3XX, yellow for 4XX and red for
+//      5XX)
+// * listens on the specified port and proxies to
+//      '<replicaHost>/?canisterId=<canisterId>'
+const mkApp = (replicaHost: string, port: number, canisterId: string) => {
```
peterpeterparker, line 70: When there are more than one param, it's handy to use object for the definition. Just a note of course.

```
const mkApp = ({replicaHost, port, canisterId}: {replicaHost: string, port: number, canisterId: string}) => {
```

proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+  app.all(
+    "*",
+    createProxyMiddleware({
+      target: replicaHost,
+      pathRewrite: (pathAndParams, req) => {
+        let queryParamsString = "?";
+
+        let [path, params] = pathAndParams.split("?");
+
+        if (params === undefined) {
```
peterpeterparker, line 103: it can be inlinied

```
queryParamsString = `?canisterId=${canisterId}${params ? `&${params}` : ''}`;
```

or simplified

```
queryParamsString = `?canisterId=${canisterId}`;

if (params) {
   queryParamsString += `&${params}`;
}
```

or a fun one (to be tested though)

```
queryParamsString = `?${[`canisterId=${canisterId}`, ...(params ? [params] : [])].join('&')}`;
```

proxy/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    process.exit(1);
+}
+
+if (Object.keys(parsed.canistersToPorts).length == 0) {
+    console.log("No canisters to proxy");
+    console.log(usage);
+    process.exit(1)
+}
+
+for (let canisterId in parsed.canistersToPorts) {
```
peterpeterparker, line 144: Not sure if we have the `eslint` rule [Array Foreach](https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md) active but if yes, you might have to replace the `for` with a `forEach`.

```
canistersToPorts.parsed.forEach((canisterId) => {
  ...
});
```

peterpeterparker (commented):
Coolio 👍

proxy/index.ts
nmattia: The reason is that errors don't appear in the type signature, so yes you could wrap everything in a `try`/`catch` (that was the first implementation) but then it's really not clear what can go wrong and how. I'd rather be explicit!

proxy/index.ts
nmattia: Definitely, overlooked. Thanks!

proxy/index.ts
nmattia: I thought about this then realised people might actually _need_ to pass in weird values; for now let's be flexible in what we allow there, but agreed this could be useful if we know for a fact that all hosts will start with `http://` (as opposed to simply e.g. `localhost:8080`)

proxy/index.ts
nmattia: Sticking to explicit errors for now, but thanks for sharing `.reduce`!

proxy/index.ts
nmattia: Thanks, done! I wish there was a way to not duplicate the parameter names (and still avoid e.g. `args: { ... }`).

proxy/index.ts
nmattia: Thanks! I've left out the clever solutions, and adapted to always have `canisterId` at the end (I think it's a bit clearer)

proxy/index.ts
nmattia: But according to `eslint`, we should instead use `for ... of`, right? So either way we're good?

proxy/index.ts
nmattia: Done, thanks! Thought ts would actually warn me; guess I should set up eslint!

proxy/index.ts
peterpeterparker: I had something different in mind then, all good then.

peterpeterparker (approved):

proxy/index.ts
bitdivine: Could you not have type `CLIArguments | Error` ?  The error doesn't have to be thrown.

bitdivine:
Please run "npm audit fix" in ./proxy.  Two of the dependencies have high severity vulnerabilities.

bitdivine:
Thanks! :-)

PR #369, opened 2022-01-20
Add daily GitHub task to security audit rs and js

bitdivine:
# Motivation
The NNS code is meant to be secure.  We rely on third party packages.  We would like to be aware of vulnerabilities.

# Changes
- Added npm and cargo audits.

Not included:
- Any mechanism for notifying users of failures.
  - Any user may subscribe to this job and so be notified of failures.  This provides push notifications to any interested parties.
  - A status badge can be added:
  -   `Audit`

### DIFF
--- a/.github/workflows/security-audit.yaml
+++ b/.github/workflows/security-audit.yaml
@@ -10,14 +10,14 @@ on:
 jobs:
   rust-audit:
     name: Rust Package Audit
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - name: Cargo Package Audit
         run: cargo audit
   js-audit:
     name: Javascript Package Audit
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
       - name: Audit packages
--- a/.github/workflows/security-audit.yaml
+++ b/.github/workflows/security-audit.yaml
@@ -7,6 +7,8 @@ on:
     # Default: Run on every PR.
   workflow_dispatch:
     # Default: Enables a button in the github UI to trigger the workflow manually.
+  push:
+    # For testing only - this clause should be deleted befor emerging
 jobs:
   rust-audit:
     name: Rust Package Audit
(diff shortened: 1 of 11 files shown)


.github/workflows/security-audit.yaml
```diff
@@ -0,0 +1,22 @@
...
+      - uses: actions-rs/audit-check@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+  js-audit:
+    name: Javascript Package Audit
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Audit packages
+      run: npm audit
```
nmattia, line 22: Shouldn't this run in a particular directory?

.github/workflows/security-audit.yaml
```diff
@@ -0,0 +1,22 @@
...
+    # Run daily
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+jobs:
+  rust-audit:
+    name: Rust Package Audit
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/audit-check@v1
```
nmattia, line 13: This doesn't look official. Couldn't we simply install cargo audit and run it ourselves?

nmattia (commented):
Nice! How are the failures reported?

.github/workflows/security-audit.yaml
bitdivine: npm audit will look, there is no need to specify, although it would be cool to list them explicitly in a top-level workspaces file, just as a top level Cargo.toml file can act as an idex.

.github/workflows/security-audit.yaml
bitdivine: Can do.  I read through the code before I started using it ~1 year ago but the overhead of running our own is small and we REALLY care about security, so...

.github/workflows/security-audit.yaml
bitdivine: Uh, actually it isn't as I remember.  I have added a manual search and a TODO to define workspaces.

PR #370, opened 2022-01-20
Add the powered by IC banner

bitdivine:
# Motivation
We are proud of running the NNS on the IC.  And there is this ticket: https://dfinity.atlassian.net/browse/L2-197?jql=project%20%3D%20L2%20AND%20text%20~%20powered

# Changes
- Added the powered by IC logo, as provided in the ic-components repo here: https://github.com/dfinity/dfn-components/blob/main/img/ic-badge-light.svg
- Updated the styling to:
  - https://www.figma.com/file/r9Ghb2EGC6xEyVvDGrT39g/Internet-Computer-Brand?node-id=1509%3A376
  - https://www.figma.com/file/6b3ucwM3ATiJ1ssJhpwQex/NNS-dapp-%2F-Edits?node-id=79%3A30

Note: As a programmer it is instinctive to pull in dfn-components, presumably as  a git submodule, rather than copy the svg.  However that is apparently against the convention of frontend devs where branding is concerned.  Example: Google provides svgs to copy, not a library.  Ionic actually provides both svgs to download and a library but the library is massive, so unless one is developing an ionic app, download is the common path.  Hence, the copy here.  When in Rome...

Note:  The spacing is identical to the historic spacing - The bottom row was 170 px high, it is now 130 plus 40 for the banner, so there should be no effects on the layout beyond that part of the screen.

Note: The flutter files show further style changes, but they are not covered in the ticket so I'd rather get clarification on them and make any changes in a separate PR.

Screenshot from 2022-01-20 16-29-58


frontend/svelte/src/routes/Auth.svelte
```diff
@@ -67,6 +67,13 @@
       <span class="green">ICP</span> and <span class="blue">governance</span>
     </p>
     <button on:click={signIn}>LOGIN</button>
+    <img
+      src="/assets/assets/ic-badge-light.svg"
+      role="presentation"
+      alt="Powered by IC Banner"
+      loading="lazy"
+      class="powered-by-ic-banner"
```
peterpeterparker, line 75: `class` can be removed

peterpeterparker (changes requested):

peterpeterparker:
@bitdivine indeed the provided design introduces changes in the overall "login - auth" page UI. it is the outcome of some discussions and I did not updated verbally the ticket but only with the related screenshots.

I am fine if we proceed with these UI changes in a separate PR but in my opinion and understanding -we have to develop these too because the "Login" button and the banner without any further changes does not play well together in terms of UX, better and cooler in the new design.

Regarding the [dfn-component](https://github.com/dfinity/dfn-components), let's have a look together soon.

frontend/svelte/src/routes/Auth.svelte
bitdivine: Done

bitdivine:
@peterpeterparker - we can try to get the style changes in.  The page would presumably look something like this:  https://sgymv-uiaaa-aaaaa-aaaia-cai.nnsdapp.dfinity.network

That is a draft; sizes are taken from figma which is in px not rem and the colours don't match the ones we have already defined.

peterpeterparker:
@bitdivine sure if we can do it in the same PR that would cool

The draft looks good  👍

A couple of things I noticed spontaneously in comparison to Figma file, if it can help:

- the media breakpoints (login button in center or bottom for mobile) should be based on the width of the screen (not height)
- section "login" and title on mobile need a bit of spacing
- "ICP and governance" needs a bit of spacing
- "login" button :hover and :active feedback can be styled
- "login" text in the button is not centered

bitdivine:
Re breakpoints: Agreed; there was already a height breakpoint in the code so I preserved that.  Flutter has breakpoints at: width 768 and 1200px and for this we need only the latter breakpoint as tablet and mobile use the same font sizes, if I recall correctly.

peterpeterparker (approved):
Thx!

PR #371, opened 2022-01-20
Update build script

nmattia:
This scratches an old itch I had and fixes some documentation for running the docker build. In particular:

* Adds `scripts/docker-build` which outputs `nns-dapp.wasm` in the top-level (that's the final, optimized wasm).
* Removes `source` from dfx.json. `source` (i.e. `assets.tar.xz`) isn't used in `custom` builds AFAICT. 
* Specifies `nns-dapp.wasm` as the wasm input for `dfx`, there's no reason to actually modify cargo's generated files in `target` with `ic-cdk-optimizer`; instead we only read those files, and place the (optimized) output in the top-level.
* Performs some checks on the tar version used. Not only does this not obfuscate tar errors anymore (which was done with `||`) but also clarifies the intent. It also gives the user directions on how to install GNU tar.
* Clarifies `build.sh` a ton by:
  * adding a diagram of what happens during the build
  * making most steps resemble each other (svelte, dart and ts builds)
  * use absolute paths with `cd` as opposed to `pushd`ing (greatly reduces the cognitive load since the person reading the scripts doesn't have to keep track of `$PWD`) (also applied to other build scripts)
  * actually use `set -euo pipefail` for sanity (also applied to other build scripts)
  * remove `web-assets`, which really was a temp directory in disguise (as are assets.tar.xz and ic_agent.js, but right now we can't get rid of those because they are referenced in resp the rust and dart builds)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Use this with
 #
 # docker build . -t nns-dapp
-# container_id=$(docker create nns-dapp)
+# container_id=$(docker create nns-dapp no-op)
 # docker cp $container_id:nns-dapp.wasm nns-dapp.wasm
 # docker rm --volumes $container_id
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,9 @@
 # Use this with
 #
-# docker build -t nns-dapp .
-# docker run --rm --entrypoint cat nns-dapp /nns-dapp.wasm > nns-dapp.wasm
+# docker build . -t nns-dapp
+# container_id=$(docker create nns-dapp)
+# docker cp $container_id:nns-dapp.wasm nns-dapp.wasm
+# docker rm --volumes $container_id
 
 # This is the "builder", i.e. the base image used later to build the final
 # code.
@@ -76,10 +78,8 @@ RUN echo "OWN_CANISTER_ID: '$OWN_CANISTER_ID'"
 COPY . .
 RUN ./build.sh
 
-# Copy the wasm to the traditional location.
-RUN cp "$(jq -rc '.canisters["nns-dapp"].wasm' dfx.json)" nns-dapp.wasm
 RUN ls -sh nns-dapp.wasm; sha256sum nns-dapp.wasm
 
-FROM scratch
+FROM scratch AS scratch
 COPY --from=build /nns-dapp.wasm /
 COPY --from=build /assets.tar.xz /
--- a/build.sh
+++ b/build.sh
@@ -57,6 +57,29 @@ set -x
 #################
 # assets.tar.xz #
 #################
+
+# we need GNU tar (see below) so we check early
+if tar --help | grep GNU >/dev/null
+then
+    echo "found GNU tar as tar"
+    tar="tar"
+elif command -v gtar >/dev/null
+then
+    echo "found GNU tar as gtar"
+    tar="gtar"
+else
+    echo "did not find GNU tar, please install"
+    echo "  brew install gnu-tar"
+    exit 1
+fi
+
+if ! command -v xz >/dev/null
+then
+    echo "did not find xz, please install"
+    echo "  brew install xz"
+    exit 1
+fi
+
 tarball_dir=$(mktemp -d)
 echo "using $tarball_dir for tarball directory"
 cp -R "$TOPLEVEL/frontend/dart/build/web/". "$tarball_dir"/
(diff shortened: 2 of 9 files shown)


Dockerfile
```diff
@@ -1,7 +1,9 @@
 # Use this with
 #
-# docker build -t nns-dapp .
-# docker run --rm --entrypoint cat nns-dapp /nns-dapp.wasm > nns-dapp.wasm
+# docker build . -t nns-dapp
+# container_id=$(docker create nns-dapp)
```
bitdivine, line 4: This line does not work:
```
max@sinkpad:~/dfn/nns-dapp/branches/nm-fix-dockerbuild (5:25)$ docker build . -t nns-dapp
...[snip]
max@sinkpad:~/dfn/nns-dapp/branches/nm-fix-dockerbuild (3:51)$ docker create nns-dapp
Error response from daemon: No command specified  <-- this happens if the image is "scratch"
```

nmattia: ah, forgot to add something there!

nmattia, line 4:
```suggestion
# container_id=$(docker create nns-dapp no-op)
```

nmattia: @bitdivine try now?

bitdivine: Aha, this works:
```
# docker build . -t nns-dapp
# container_id=$(docker create nns-dapp whatever)
# docker cp $container_id:nns-dapp.wasm nns-dapp.wasm
# docker rm --volumes $container_id
```

nmattia: see comment above, it was a copy/paste error. The actual script running this has `docker create nns-dapp no-op`.

build.sh
```diff
@@ -14,48 +36,84 @@ fi
...
+  exit 1
+fi
+
+if ! command -v xz >/dev/null; then
+  echo "did not find xz, please install"
+  echo "  brew install xz"
+  exit 1
+fi
+
+tarball_dir=$(mktemp -d)
```
bitdivine, line 80: A fixed directory is more useful for local development.  At the moment you can do things such as `cd web-assets ; python -m http.server` and look at the full built front end very cheaply, and update the contents of the directory without restarting the web server.  Please keep using a fixed directory rather than a tmpdir.

build.sh
```diff
@@ -14,48 +36,84 @@ fi
...
 rm assets/NOTICES
-tar cJv --mtime='2021-05-07 17:00+00' --sort=name --exclude .last_build_id -f ../assets.tar.xz . ||
-  gtar cJv --mtime='2021-05-07 17:00+00' --sort=name --exclude .last_build_id -f ../assets.tar.xz .
-popd || exit
+"$tar" cJv --mtime='2021-05-07 17:00+00' --sort=name --exclude .last_build_id -f "$TOPLEVEL/assets.tar.xz" .
 
-ls -sh assets.tar.xz
-sha256sum assets.tar.xz
+cd "$TOPLEVEL"
+rm -rf "$tarball_dir"
```
bitdivine, line 96: Please use a fixed path and don't delete the assets dir.

build.sh
nmattia: done!

build.sh
nmattia: done!

build.sh
```diff
@@ -14,48 +36,84 @@ fi
...
-npm run build
-popd || exit
+#################
+# assets.tar.xz #
+#################
 
-rm -fr web-assets/v2
-cp -R frontend/svelte/public web-assets/v2
+# we need GNU tar (see below) so we check early
+if tar --help | grep GNU >/dev/null; then
```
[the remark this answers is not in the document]
nmattia: unfortunately that's not POSIX!

nmattia:
By the way, [shasum of `mainnet` assets on main](https://github.com/dfinity/nns-dapp/runs/4881792515?check_suite_focus=true#step:4:1758):
```
#31 [build 6/6] RUN ls -sh nns-dapp.wasm; sha256sum nns-dapp.wasm
1759
#31 0.074 2.7M nns-dapp.wasm
1760
#31 0.092 3dac3ae445042a6889d0ad7a874248da7588add4bd06c67f2a2912e322832547  nns-dapp.wasm
1761
#31 DONE 0.1s
```

[On this branch](https://github.com/dfinity/nns-dapp/runs/4895644338?check_suite_focus=true#step:4:2601):
```
#30 [build 5/5] RUN ls -sh nns-dapp.wasm; sha256sum nns-dapp.wasm
2602
#30 0.535 2.7M nns-dapp.wasm
2603
#30 0.555 3dac3ae445042a6889d0ad7a874248da7588add4bd06c67f2a2912e322832547  nns-dapp.wasm
2604
#30 DONE 0.6s
```

bitdivine (approved):
Thanks!

PR #372, opened 2022-01-20
Get rid of DEPLOY_ENV=nobuild

nmattia:


scripts/deploy-ci-to-testnet
```diff
@@ -33,6 +33,6 @@ unzip "$DOWNLOAD_FILENAME"
 popd
 mv "$UNZIP_TMPDIR/$WASM_FILENAME" "$WASM_PATH"
 rm -fr "${UNZIP_TMPDIR}"
-DEPLOY_ENV=nobuild dfx deploy --no-wallet --network testnet
+dfx canister --no-wallet --network testnet install --all
```
bitdivine, line 36: This fails of the canister already exists.  Perhaps there is a command that deploys without building - I believe there should be - but could not find one.  The closest I could find was:
```
$ dfx canister --no-wallet --network testnet install -m reinstall nns-dapp
Warning!
You are about to reinstall the nns-dapp canister.
This will OVERWRITE all the data and code in the canister.

YOU WILL LOSE ALL DATA IN THE CANISTER.

Do you want to proceed? yes/No

Refusing to reinstall canister without approval
```
if I recall correctly.

nmattia: I know -- that's why this is a draft PR! but thanks for contributing

PR #374, opened 2022-01-25
feat: voting filters

peterpeterparker:
# Motivation

Add static filters to the "Voting" page.

# Changes

- new `<Checkbox />`
- new `<Chip />`
- new `<Modal />`
- new `<Filter />` (card)
- add static filters on voting page
- add static filter "hide my proposal"
- add a generic modal for the "voting filters"

# Note

Filters are provided as static and are not functional. In subsequent PRs, the effective queries of the proposals is going to be implemented with the help of these filters. Therefore it is also worth to note that handling the data, the filters (passing props vs store) is subject to changes.

# Screenshots

a
b
c
d
e
f
g
h


frontend/svelte/src/lib/components/Checkbox.svelte
```diff
@@ -0,0 +1,59 @@
...
+  export let inputId: string;
+  export let checked: boolean;
+
+  export let color: "dark" | "light" = "light";
+  export let text: "block" | "inline" = "inline";
+
+  export let selector: string | undefined = undefined;
+
+  const dispatch = createEventDispatcher();
+  const select = () => dispatch("select");
```
nmattia, line 13: nitpick: looks like the `select` event is generally used for selecting text: 

> select events can be dispatched only on form <input type="text"> and <textarea> elements.
> - [source](https://developer.mozilla.org/en-US/docs/Web/API/Element/select_event)

could also be nice to add some doc here so that the user knows how to interact with the component without having to read the code!

frontend/svelte/src/lib/components/Filter.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
```
nmattia, line 1: I think the name of this component is a bit odd; why is it a "Filter" if it (itself) contains filters?

frontend/svelte/src/lib/components/VotingFilters.svelte
```diff
@@ -0,0 +1,73 @@
...
+    >Proposal Status</Filter
+  >
+</div>
+
+<Checkbox
+  inputId="hide-my-proposals"
+  checked={false}
+  color="dark"
+  text="block"
+  selector="hide-my-proposals"
```
nmattia, line 47:
```suggestion
  selector="hide-unavailable-proposals"
```

frontend/svelte/src/lib/modals/VotingFilterModal.svelte
```diff
@@ -0,0 +1,43 @@
...
+  import Checkbox from "../components/Checkbox.svelte";
+
+  export let props: VotingFilterModalProps | undefined;
+
+  let visible: boolean;
+  let title: string;
+  let filters: VotingFilters | undefined;
+
+  // TODO(#L2-206): do we want a store or pass props?
+  let activeTopics: VotingFilters[] = [];
```
nmattia, line 14: These are the "selected" topics, right? Might be worth renaming or adding a comment, I thought this was the list of all topics

frontend/svelte/src/lib/modals/VotingFilterModal.svelte
```diff
@@ -0,0 +1,43 @@
...
+
+  $: visible = props !== undefined;
+  $: title = props?.title || "";
+  $: filters = props?.filters;
+
+  const dispatch = createEventDispatcher();
+  const close = () => dispatch("close");
+
+  const select = (topic: VotingFilters) =>
```
nmattia, line 23: nit: I think this can be simplified (and performance improved, not that it matters though) by using a `dict` of topic to bool

nmattia, line 23: `select` is misleading; I thought this would only "select" a topic whereas it really selects and deselects (i.e. toggles)

nmattia (commented):
Few comments, thanks!

frontend/svelte/src/lib/components/Checkbox.svelte
peterpeterparker: I have prefixed all custom events with `nns` - e.g. `nnsChange`, `nnsFilter` or `nnsClose`

Comment added

frontend/svelte/src/lib/modals/VotingFilterModal.svelte
peterpeterparker: rename to `nnsChange`

peterpeterparker: seems more verbose, because we would have first to initialize a local state for the props (enum -> dict), with no big performance improvement since there are max 15 string in the list. I keep in mind in next PRs though, my be easier to use once functionally integrated.

frontend/svelte/src/lib/components/Filter.svelte
peterpeterparker: Not against a better naming, any idea? `FilterCard`?

Capture d’écran 2022-01-25 à 11 25 30

nmattia: Yeah I like `FilterCard`!

peterpeterparker: `FiltersCard` done.

nmattia (approved):
LGTM!

PR #376, opened 2022-01-25
Fix docker build

bitdivine:
# Motivation
The prod release process broke due to changes introduced for ci testing.  This PR is an attempt to find a clean solution that works for both.

Design goals:
- Simplify code.
- Have a documented algorithm for setting OWN_CANISTER_ID.
- Centralize logic, so that developers don't need to search through the codebase to try to figure out how variables are set.
  - Note: We can and probably should also set the default own_canister_id in this code block.  That requires more changes and more testing though, and right now deployments break so it would be better to add that later.

# Changes
The logic for determining OWN_CANISTER_ID implements this design:
```
# Here the OWN_CANISTER_ID is set.  The value is taken from the following, in order of priority:
# - The env var OWN_CANISTER_ID
# - canister_ids.json
# - try to deduce the value.
```

### DIFF
--- a/scripts/docker-build
+++ b/scripts/docker-build
@@ -11,6 +11,23 @@ cd "$SCRIPTS_DIR/.."
 
 DEPLOY_ENV=${DEPLOY_ENV:-mainnet}
 
+# Here the OWN_CANISTER_ID is set.  The value is taken from the following, in order of priority:
+# - The env var OWN_CANISTER_ID
+# - canister_ids.json
+# - try to deduce the value.
+if test -n "${OWN_CANISTER_ID:-}"
+then echo "Using env var OWN_CANISTER_ID: $OWN_CANISTER_ID"
+fi
+
+if test -z "${OWN_CANISTER_ID:-}" && test -e canister_ids.json
+then OWN_CANISTER_ID="$(jq -r '.["nns-dapp"]["'"$DEPLOY_ENV"'"]' canister_ids.json)"
+     if test -n "${OWN_CANISTER_ID:-}"
+     then echo "Using caniser ID from canister_ids.json: $OWN_CANISTER_ID"
+     fi
+fi
+
+if test -z "${OWN_CANISTER_ID:-}"
+then
 case "$DEPLOY_ENV" in
     mainnet)
         OWN_CANISTER_ID=""
@@ -24,6 +41,7 @@ case "$DEPLOY_ENV" in
         exit 1
         ;;
 esac
+fi
 
 echo "DEPLOY_ENV: $DEPLOY_ENV"
 echo "OWN_CANISTER_ID: $OWN_CANISTER_ID"
--- a/scripts/deploy-docker-build-to-testnet
+++ b/scripts/deploy-docker-build-to-testnet
@@ -2,7 +2,7 @@
 set -euxo pipefail
 cd "$(dirname "$(realpath "$0")")/.." || exit
 
-./scripts/docker-build
+DEPLOY_ENV=testnet ./scripts/docker-build
 
 DEPLOY_ENV=nobuild dfx deploy --no-wallet --network testnet
 


scripts/docker-build
```diff
@@ -11,6 +11,23 @@ cd "$SCRIPTS_DIR/.."
 
 DEPLOY_ENV=${DEPLOY_ENV:-mainnet}
 
+# Here the OWN_CANISTER_ID is set.  The value is taken from the following, in order of priority:
+# - The env var OWN_CANISTER_ID
+# - canister_ids.json
+# - try to deduce the value.
+if test -n "${OWN_CANISTER_ID:-}"
```
nmattia, line 18:
```suggestion
if [ -n "${OWN_CANISTER_ID:-}" ]
```

here and elsewhere

scripts/docker-build
```diff
@@ -11,6 +11,23 @@ cd "$SCRIPTS_DIR/.."
...
+then echo "Using env var OWN_CANISTER_ID: $OWN_CANISTER_ID"
+fi
+
+if test -z "${OWN_CANISTER_ID:-}" && test -e canister_ids.json
+then OWN_CANISTER_ID="$(jq -r '.["nns-dapp"]["'"$DEPLOY_ENV"'"]' canister_ids.json)"
```
nmattia, line 23: I'd rather not depend on `jq` if we can avoid it, and more importantly I'd rather not depend on temporary, `dfx` created file. Why not simply read `OWN_CANISTER_ID` on line `22`? Then feel free to add as much logic as you want in `deploy-docker-build-to-testnet`.

Also it's now odd to have different heuristics for `testnet` and `local` (whereas AFAICT they both depend on whatever canister ID was created)

scripts/docker-build
```diff
@@ -11,6 +11,23 @@ cd "$SCRIPTS_DIR/.."
...
 
+# Here the OWN_CANISTER_ID is set.  The value is taken from the following, in order of priority:
+# - The env var OWN_CANISTER_ID
+# - canister_ids.json
+# - try to deduce the value.
+if test -n "${OWN_CANISTER_ID:-}"
+then echo "Using env var OWN_CANISTER_ID: $OWN_CANISTER_ID"
+fi
+
+if test -z "${OWN_CANISTER_ID:-}" && test -e canister_ids.json
```
nmattia, line 22: `test -z "${OWN_CANISTER_ID:-}"`: You have the exact opposite test above (`-n`), why not add the test for `canister_ids` etc inside an else branch?

scripts/docker-build
```diff
@@ -11,6 +11,23 @@ cd "$SCRIPTS_DIR/.."
...
+then OWN_CANISTER_ID="$(jq -r '.["nns-dapp"]["'"$DEPLOY_ENV"'"]' canister_ids.json)"
+     if test -n "${OWN_CANISTER_ID:-}"
+     then echo "Using caniser ID from canister_ids.json: $OWN_CANISTER_ID"
+     fi
+fi
+
+if test -z "${OWN_CANISTER_ID:-}"
```
nmattia, line 29: This now doesn't really make sense; why not invert the logic and add a `testnet` branch?

nmattia (changes requested):

nmattia:
Thought: why not remove all the `DEPLOY_ENV` and `OWN_CANISTER_ID` logic from docker-build and lift this to the callers? I think it makes most sense (and then docker-build just reads the values and sets them as build-args)

bitdivine:
Getting `OWN_CANISTER_ID` from canister_ids.json requires jq.  Are you OK with requiring developers to have jq installed?

nmattia:
This isn't ideal but sure if that's the only way! Also didn't you wanna reduce the docker build to a single line `docker build` with buildx and `-o .`?

bitdivine:
Test: f15670f623d7c7f679cb408117984edd508375c3

nmattia (approved):
Unblocking, clarified offline!

PR #377, opened 2022-01-26
feat: trim query string in route path and tests

peterpeterparker:
# Motivation

At the time we developed the routing, the test framework was not set. This PR adds the tests to fulfill the gap.

In addition, it provides a small improvement to the route path function to harmonize its behavior if used with or without hash routing.

# Changes

- trim query string from location hash in function `routePath`
- add tests

### DIFF
--- a/frontend/svelte/jest.config.cjs
+++ b/frontend/svelte/jest.config.cjs
@@ -11,4 +11,5 @@ module.exports = {
   setupFilesAfterEnv: ["<rootDir>/jest-setup.ts"],
   collectCoverageFrom: ["src/**/*.{ts,tsx,svelte,js,jsx}"],
   testPathIgnorePatterns: ["nns-js"],
+  testURL: 'https://nns.ic0.app/'
 };
--- a/frontend/svelte/src/lib/utils/route.utils.ts
+++ b/frontend/svelte/src/lib/utils/route.utils.ts
@@ -1,10 +1,10 @@
 /**
- * The pathname and the hash without base href
+ * The pathname and the hash without base href and without the query string
  */
 export const routePath = (): string => {
   const base: string = baseHref();
   const { pathname, hash } = window.location;
-  return `${pathname.replace(base, "/")}${hash.toLowerCase()}`;
+  return `${pathname.replace(base, "/")}${hash.replace(/\?.*/,'').toLowerCase()}`;
 };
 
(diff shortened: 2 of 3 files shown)


frontend/svelte/jest.config.cjs
```diff
@@ -11,4 +11,5 @@ module.exports = {
   setupFilesAfterEnv: ["<rootDir>/jest-setup.ts"],
   collectCoverageFrom: ["src/**/*.{ts,tsx,svelte,js,jsx}"],
   testPathIgnorePatterns: ["nns-js"],
+  testURL: "https://nns.ic0.app/",
```
nmattia, line 14: what exactly does this do? The documentation is a bit sparse:

```
testURL [string]​

Default: http://localhost

This option sets the URL for the jsdom environment. It is reflected in properties such as location.href.
```

peterpeterparker: In the tests, without specific configuration, following is defined:

```
expect(window.location.href).toEqual('http://localhost') //  => true
```

With the configuration:

```
expect(window.location.href).toEqual('http://localhost') //  => false
expect(window.location.href).toEqual('https://nns.ic0.app/') //  => true
```

nmattia: But this isn't correct, right? Or does it not matter what the actual href is?

peterpeterparker: I see you mean the default for the tests should be `https://nns.ic0.app/v2/` ?

`/v2/` is what we use now but my understanding is that asap. it should integrally replace the existing. Therefore will become `/`. That's why I thought it makes sense to have `/` as default already now in the tests.

It does not matter much except for the tests particular tests of this PR. If you would like to change the default configuration then I should adapt the test accordingly.

The app itself can of course work on any base href.

nmattia: Sorry, no, I mean that when the tests are run they are not running on `nns.ic0.app`. I guess this doesn't matter in the end, and you're doing this to have URLs similar to production, is that it? I guess I just don't understand the intent

peterpeterparker: > this doesn't matter in the end, and you're doing this to have URLs similar to production, is that it?

Yes

nmattia (approved):
LGTM!

PR #378, opened 2022-01-26
feat: spinner

peterpeterparker:
# Motivation

Add a spinner that can be used for actions that take a certain time till completion.

# Changes

- new component `<Spinner />`
- use spinner in `<Guard />`(until agent-js is loaded) and to display main account (until ledget balance is fetched)

# GIF

ezgif com-gif-maker(2)

### DIFF
--- a/frontend/svelte/src/lib/components/Guard.svelte
+++ b/frontend/svelte/src/lib/components/Guard.svelte
@@ -2,6 +2,7 @@
   import { authStore } from "../stores/auth.store";
   import { routeStore } from "../stores/route.store";
   import { routePath } from "../utils/route.utils";
+  import Spinner from "./Spinner.svelte";
 </script>
 
 <!-- storage: on every change in local storage we sync the auth state -->
@@ -12,7 +13,7 @@
 />
 
 {#await authStore.sync()}
-  <!-- TODO(L2-175): display a spinner or other animation while loading the auth -->
+  <Spinner />
 {:then value}
   <slot />
 {:catch error}
--- a/frontend/svelte/src/lib/components/Spinner.svelte
+++ b/frontend/svelte/src/lib/components/Spinner.svelte
@@ -1,0 +1,135 @@
+<!-- adapted source: https://github.com/angular/components/tree/master/src/material/progress-spinner -->
+<svg
+  preserveAspectRatio="xMidYMid meet"
+  focusable="false"
+  aria-hidden="true"
+  viewBox="0 0 100 100"
+  ><circle
+    cx="50%"
+    cy="50%"
+    r="45"
+  /></svg
+>
+
+<style lang="scss">
+  @use "sass:math";
+
+  svg {
+    width: 30px;
+    height: 30px;
+
+    animation: spinner-linear-rotate 2000ms linear infinite;
+
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+
+    // same as <circle r=""/> attribute
+    --radius: 45px;
+    --circumference: calc(#{math.$pi} * var(--radius) * 2);
+
+    // start the animation at 5%
+    --start: calc((1 - 0.05) * var(--circumference));
+    --end: calc((1 - 0.8) * var(--circumference));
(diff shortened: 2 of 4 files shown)


frontend/svelte/src/lib/components/Guard.svelte
```diff
@@ -12,7 +13,7 @@
 />
 
 {#await authStore.sync()}
-  <!-- TODO(L2-175): display a spinner or other animation while loading the auth -->
+  <Spinner />
```
nmattia, line 16: This is a full page spinner, right? Can you share a short screen recording?

frontend/svelte/src/lib/components/Spinner.svelte
```diff
@@ -0,0 +1,130 @@
+<!-- adapted source: https://github.com/angular/components/tree/master/src/material/progress-spinner -->
+<svg
+  preserveAspectRatio="xMidYMid meet"
+  focusable="false"
+  aria-hidden="true"
+  viewBox="0 0 100 100"><circle cx="50%" cy="50%" r="45" /></svg
```
nmattia, line 6: This is a circle that's drawn with the dashed line that moves around that you describe below, correct?

frontend/svelte/src/lib/components/Spinner.svelte
```diff
@@ -0,0 +1,130 @@
...
+      transform: rotate(0deg);
+    }
+    100% {
+      transform: rotate(360deg);
+    }
+  }
+
+  /* -global- */
+  @keyframes -global-spinner-stroke-rotate-100 {
+    0% {
```
nmattia, line 62: can't we generate the frames with sass? might be cleaner and less error prone

frontend/svelte/src/routes/Accounts.svelte
```diff
@@ -37,9 +38,9 @@
 
       {#if main}
         <AccountCard account={main}>{$i18n.accounts.main}</AccountCard>
+      {:else}
```
nmattia, line 41: noob svelte question: svelte will re-evaluate the `if/else` whenever `main` changes, right?

nmattia (approved):
LGTM, few questions though!

frontend/svelte/src/routes/Accounts.svelte
peterpeterparker: yes

frontend/svelte/src/lib/components/Spinner.svelte
peterpeterparker: it seems you can loop (https://stackoverflow.com/a/20491731/5404186) but every frames are different so not sure what it would add

also it needs to be compatible with `/* -global- */` otherwise svelte won't apply the keyframe

peterpeterparker: not against improvement if you have an idea?

frontend/svelte/src/lib/components/Guard.svelte
peterpeterparker: `position: abolute` i.e. related to its parent container

gif added to detail

frontend/svelte/src/lib/components/Spinner.svelte
nmattia: What's `-global-`?

> every frames are different 

Right, but they follow the same structure don't they? Or maybe not. Either way, this is also completely fine! (though as always I think it'd be nice to add a comment explaining how they work to make a nice spinner)

frontend/svelte/src/lib/components/Spinner.svelte
peterpeterparker: Yes

CSS is automatically scoped to the component except if you use `:global`

frontend/svelte/src/lib/components/Spinner.svelte
peterpeterparker: `-global-` is related to `global`. It's a trick to make svelte renders the keyframes (at least last time I checked)

I would say let's live with it then and invest time elsewhere (yolo).

PR #379, opened 2022-01-26
feat: improve modal a11y

peterpeterparker:
# Motivation

The `<Modal />` component is a dialog, therefore is role should be set as such ([MDN doc](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role))

# Changes

- set `role=dialog` and aria information


nmattia (approved):
TIL, nice!

PR #383, opened 2022-01-31
build: nns-js from npm

peterpeterparker:
# Motivation

[nns-js](https://github.com/dfinity/nns-js/) is public and has been published to [npm](https://www.npmjs.com/package/@dfinity/nns). Use it as a dependency instead of a local copy.

# Changes

- drops local nns-js copy
- add dependency


nmattia (approved):
LGTM!

PR #384, opened 2022-01-31
feat: toast

peterpeterparker:
# Motivation

We need to inform user about unexpected errors, therefore I created a `<Toast />` component.

# Changes

- add a new `<Toast />` component
- introduce a store to display `info` and `error` messages

# Screenshots

a
b
c
d
">

### DIFF
--- a/frontend/svelte/src/lib/components/Toast.svelte
+++ b/frontend/svelte/src/lib/components/Toast.svelte
@@ -1,12 +1,34 @@
-<script>
+<script lang="ts">
   import IconClose from "../icons/IconClose.svelte";
+  import { msg } from "../stores/msg.store";
+  import { fade, fly } from "svelte/transition";
+  import { translate } from "../utils/i18n.utils";
+  import { i18n } from "../stores/i18n";
+
+  let visible: boolean;
+
+  const close = () => msg.set(undefined);
+
+  $: visible = $msg !== undefined;
+  $: ({ labelKey, type: msgType } = $msg || {});
 </script>
 
-<div role="dialog" class="toast error">
-  <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse vel turpis tempor, bibendum nunc vel, condimentum ligula.</p>
+{#if visible}
+  <div
+    role="dialog"
+    class="toast"
+    class:error={msgType === "error"}
+    in:fly={{ y: 100, duration: 200 }}
+    out:fade={{ delay: 100 }}
+  >
+    <p>
+      {translate({ labelKey })}
+    </p>
 
-  <button><IconClose /></button>
-</div>
+    <button on:click={close} aria-label={$i18n.core.close}><IconClose /></button
+    >
+  </div>
+{/if}
 
 <style lang="scss">
   @use "../themes/mixins/text";
--- a/frontend/svelte/src/lib/components/Toast.svelte
+++ b/frontend/svelte/src/lib/components/Toast.svelte
@@ -10,14 +10,14 @@
   const close = () => msg.set(undefined);
 
   $: visible = $msg !== undefined;
-  $: ({ labelKey, type: msgType } = $msg || {});
+  $: ({ labelKey, level } = $msg || {});
 </script>
 
 {#if visible}
   <div
     role="dialog"
     class="toast"
-    class:error={msgType === "error"}
+    class:error={level === "error"}
     in:fly={{ y: 100, duration: 200 }}
     out:fade={{ delay: 100 }}
   >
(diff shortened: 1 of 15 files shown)


nmattia (commented):
Is there a way to make the toast bigger on phone? Right now it doesn't look particularly useful. I (for one) hate it when I'm only shown the first few words of a message!

peterpeterparker:
@nmattia we can add more lines on mobile for example if you want. currently two, four?

```
p {
    @include text.clamp(2);
    ...
```

nmattia:
Let's either show everything, or add a way of opening the message fullscreen, what do you think?

peterpeterparker:
as discussed in chat, we do 4 lines on mobile (done) and if in the future we realize we need a better solution to display unexpected errors with more details, we will create a new component (like a scrollable bottom modal or something)

Capture d’écran 2022-01-31 à 11 16 03

frontend/svelte/src/lib/components/Toast.svelte
```diff
@@ -0,0 +1,83 @@
+<script lang="ts">
```
nmattia, line 1: Few lines at the top of this file explaining how this works (i.e. toast is always present, but only set to visible when there's an update in the store) would be great

frontend/svelte/src/lib/stores/msg.store.ts
```diff
@@ -0,0 +1,8 @@
+import { writable } from "svelte/store";
+
+export interface Msg {
+  labelKey: string;
+  level: "info" | "error";
+}
+
+export const msg = writable<Msg | undefined>(undefined);
```
nmattia, line 8: Shouldn't this be a list of `Msg`? What happens if the user receives another toast before the first one was closed? (I mean I know what happens, just saying it may not be ideal)

frontend/svelte/src/lib/components/Toast.svelte
```diff
@@ -0,0 +1,83 @@
...
+  import IconClose from "../icons/IconClose.svelte";
+  import { msg } from "../stores/msg.store";
+  import { fade, fly } from "svelte/transition";
+  import { translate } from "../utils/i18n.utils";
+  import { i18n } from "../stores/i18n";
+
+  let visible: boolean;
+
+  const close = () => msg.set(undefined);
```
nmattia, line 10: I don't feel comfortable with having two "owners" of the stores, i.e. two (or more) components that have write access to it. I'd rather have a queue interface where everything can enqueue new messages, and the toast only dequeues them. This would also solve the race condition mentioned in another comment. WDYT?

frontend/svelte/src/lib/components/Toast.svelte
peterpeterparker: done

frontend/svelte/src/lib/components/Toast.svelte
peterpeterparker: Source: Svelte tutorial - documentation

https://svelte.dev/tutorial/writable-stores

frontend/svelte/src/lib/stores/msg.store.ts
peterpeterparker: I don't think we will have the use case where multiple msgs are going to be added to the toast. Therefore a single msg.

To go futher we can manage the messages with list and add also timeout to hide these automatically.

Before implementing x functions we might or might now use, let's add the basis, begin to use the component and increment from there.

frontend/svelte/src/lib/components/Toast.svelte
peterpeterparker: (in progress)

peterpeterparker: done

frontend/svelte/src/lib/stores/toasts.store.ts
```diff
@@ -0,0 +1,32 @@
...
+  const { subscribe, update } = writable<ToastMsg[]>([]);
+
+  return {
+    subscribe,
+
+    show(msg: ToastMsg) {
+      update((messages: ToastMsg[]) => [...messages, msg]);
+    },
+
+    hide(index: number) {
```
nmattia, line 24: won't the index change after you've called `hide` on an index? i.e. won't `close()` close the wrong toast?

peterpeterparker: as discussed in a chat, we will display one toast at a time and the button close will pops that particular msg

frontend/svelte/src/lib/components/Toasts.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+  import { toastsStore } from "../stores/toasts.store";
+  import Toast from "./Toast.svelte";
+</script>
+
+<!-- Display one toast message at a time -->
+{#if $toastsStore.length > 0}
+  <Toast msg={$toastsStore[0]} />
```
nmattia, line 8: I think this is a bit error prone

nmattia (approved):
Still a thing or two I find a bit error prone but in the end this is just a toast

PR #385, opened 2022-01-31
build: ts dev dependencies cleanup

peterpeterparker:
# Motivation

Clean up de dependencies

# Changes

- remove unused source-map-explorer
- moves browserify related dependencies to dev dependencies

# Note

I don't know well `ts` but couldn't even more dependencies be removed?
Not sure about it, just feels like there are maybe some unused dependencies in the project.


nmattia (approved):
LGTM!

> feels like there are maybe some unused dependencies in the project

You can always remove them and see if CI still passes...

peterpeterparker:
# Motivation

Clean up de dependencies

# Changes

- remove unused source-map-explorer
- moves browserify related dependencies to dev dependencies

bitdivine:
Note that the build may pass but fail at runtime, depending on what the dependency is.  But I believe that swapping to nns-js for svelte is safe, at least for the methods we use at the moment! :-)

PR #386, opened 2022-01-31
feat: display unexpected auth sync and sign-in errors

peterpeterparker:
# Motivation

Display unexpected auth sync and sign-in errors to users.

# Changes

- add `<Toasts />` to login page (it does not use the main layout)
- stack a toast message for unexpected auth sync errors
- stack a toast message for sign-in errors
- add `detail` to toast store to display an optional string error in addition to an internationalized label
- display 5 lines in toast instead of 4 on mobile devices to display more details about the error

### DIFF
--- a/frontend/svelte/src/lib/components/Guard.svelte
+++ b/frontend/svelte/src/lib/components/Guard.svelte
@@ -3,6 +3,16 @@
   import { routeStore } from "../stores/route.store";
   import { routePath } from "../utils/route.utils";
   import Spinner from "./Spinner.svelte";
+  import { toastsStore } from "../stores/toasts.store";
+
+  const syncAuthStore = async () => {
+    try {
+      await authStore.sync();
+    } catch (err) {
+      toastsStore.show({ labelKey: "error.auth_sync", level: "error" });
+      console.error(err);
+    }
+  };
 </script>
 
 <!-- storage: on every change in local storage we sync the auth state -->
@@ -12,10 +22,8 @@
   on:popstate={() => routeStore.update({ path: routePath() })}
 />
 
-{#await authStore.sync()}
+{#await syncAuthStore()}
   <Spinner />
 {:then value}
   <slot />
-{:catch error}
-  <!-- TODO(L2-176): display the errors -->
 {/await}
--- a/frontend/svelte/src/lib/components/Toast.svelte
+++ b/frontend/svelte/src/lib/components/Toast.svelte
@@ -52,6 +52,8 @@
     padding: var(--padding) calc(var(--padding) * 2);
     box-sizing: border-box;
 
+    z-index: calc(var(--z-index) + 999);
+
     @media (min-width: 880px) {
       max-width: var(--section-max-width);
     }
--- a/frontend/svelte/src/lib/components/Toast.svelte
+++ b/frontend/svelte/src/lib/components/Toast.svelte
@@ -12,8 +12,11 @@
 
   const close = () => toastsStore.hide();
 
-  $: ({ labelKey, level, detail } =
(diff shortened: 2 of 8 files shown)


frontend/svelte/src/lib/components/Guard.svelte
```diff
@@ -3,6 +3,16 @@
   import { routeStore } from "../stores/route.store";
   import { routePath } from "../utils/route.utils";
   import Spinner from "./Spinner.svelte";
+  import { toastsStore } from "../stores/toasts.store";
+
+  const syncAuthStore = async () => {
+    try {
+      await authStore.sync();
```
nmattia, line 10: Out of curiosity, what could go wrong here?

frontend/svelte/src/lib/components/Toast.svelte
```diff
@@ -12,7 +12,8 @@
 
   const close = () => toastsStore.hide();
 
-  $: ({ labelKey, level } = msg || { labelKey: "", level: "info" });
+  $: ({ labelKey, level, detail } =
+    msg || ({ labelKey: "", level: "info" } as ToastMsg));
```
nmattia, line 16: ts noob: why do you need to explicitly cast your object to a `ToastMsg`?

frontend/svelte/src/lib/components/Toast.svelte
```diff
@@ -23,7 +24,7 @@
   out:fade={{ delay: 100 }}
 >
   <p>
-    {translate({ labelKey })}
+    {translate({ labelKey })}{detail ? ` ${detail}` : ""}
```
nmattia, line 27: How about a [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element here?

frontend/svelte/src/lib/components/Guard.svelte
peterpeterparker: not much normally as long as agent-js has multiple try catch in its [create](https://github.com/dfinity/agent-js/blob/118d8145b25b1b9d8308ca5447d0090ef1af6a65/packages/auth-client/src/index.ts#L1559) function

frontend/svelte/src/lib/components/Toast.svelte
peterpeterparker: > $ npm run check
>
> /Users/davi.....mponents/Toast.svelte:15:26 
> Error: Property 'detail' does not exist on type 'ToastMsg | { labelKey: string; level: string; }'. (ts)
>
> $: ({ labelKey, level, detail } =
>    msg || ({ labelKey: "", level: "info" }));

frontend/svelte/src/lib/components/Toast.svelte
peterpeterparker: It would be an un-common behavior for such a component (see for example Material design [guidelines](https://material.io/archive/guidelines/components/snackbars-toasts.html#))

frontend/svelte/src/lib/components/Toast.svelte
nmattia: right, but casting explicitly will break at runtime if we add another attribute to `ToastMsg`. Shouldn't this be something like this?

``` ts
msg || ({ labelKey: "", level: "info", detail: undefined })
```

peterpeterparker: nope, at runtime types disappear therefore nothing will break. in addition if another attribute is added, it's like adding something that will be ignored / not used as we don't use it in the template aka no effect. therefore the explicit cast is safe.

but I don't mind specifying `detail: undefined`, done.

nmattia (approved):

PR #387, opened 2022-02-01
ci: download filename matcher for prefixed filenames

peterpeterparker:
# Motivation

Artifacts filenames contain prefix "NNS ..." - e.g "NNS testnet backend wasm module.zip"
The `deploy-ci-to-testnet` script except filename beginning with "Testnet ...".

This PR makes the script support prefix.

# Changes

- `deploy-ci-to-testnet` script grep regex for download filenames


nmattia (approved):
LGTM, though I think @bitdivine wrote this, he might want to confirm

peterpeterparker:
No feedback from @bitdivine but since I am most probably the only one that use the script and it has been accepted, I merge it.

PR #390, opened 2022-02-01
feat: "Get ICPs" UI and <Input /> component

peterpeterparker:
# Motivation

Init the testnet only feature "Get ICPs" and create an `<Input />` component that can be use across the app.

# Changes

- new `<Input />` component 
- "Get ICPs" component that is only rendered if `testnet`
- implement a `<form />` for the first time
- modify rollup config to use and replace `process.env.DEPLOY_ENV` in code
- implement a cheap ripple styling effect for button
- style globally buttons to spare reused CSS code
- new style definition for any "blue" buttons
- add a type definition for input event

# Note

No test are provided for the `<GetICPs />` component because it has a testnet only purpose. On the contrary a test was added to the `<Header />` to check that with default build config the component is not rendered.

"Get ICPs" feature itself will be implemented in separate PR(s).


frontend/svelte/src/global.d.ts
```diff
@@ -1 +1,5 @@
 /// <reference types="svelte" />
+
+type InputEventHandler = Event & {
```
nmattia, line 3: (noob question) what is this?

frontend/svelte/src/lib/components/GetICPs.svelte
```diff
@@ -0,0 +1,56 @@
+<script lang="ts">
```
nmattia, line 1: A line or two giving an overview of what the component does and how it does it would be greatly appreciated

frontend/svelte/src/lib/components/GetICPs.svelte
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  const onClose = () => {
+    visible = false;
+    inputValue = undefined;
+  };
+
+  let validForm: boolean;
+
+  $: validForm = inputValue === undefined || inputValue <= 0;
```
nmattia, line 24: Should this be the opposite?

frontend/svelte/src/lib/components/GetICPs.svelte
```diff
@@ -0,0 +1,56 @@
...
+</script>
+
+<button on:click={() => (visible = true)} class="open text">Get ICPs</button>
+
+<Modal {visible} on:nnsClose={onClose}>
```
nmattia, line 29: where does the `nnsClose` event come from?

frontend/svelte/src/lib/components/GetICPs.svelte
```diff
@@ -0,0 +1,56 @@
...
+  <span slot="title">Get ICPs</span>
+
+  <form on:submit|preventDefault={onSubmit}>
+    <span class="how-much">How much?</span>
+
+    <Input placeholderLabelKey="core.icp" name="icp" bind:value={inputValue} />
```
nmattia, line 35: ~~I can't find any documentation on `placeholderLabelKey`. What is that?~~ ok, just saw this is a custom component

frontend/svelte/src/lib/components/GetICPs.svelte
```diff
@@ -0,0 +1,56 @@
...
+
+    <button type="submit" class="submit" disabled={validForm}>Get</button>
```
nmattia, line 37: shouldn't this have a `form="..."`? Or is having it inside the form enough?

frontend/svelte/src/lib/components/Input.svelte
```diff
@@ -0,0 +1,82 @@
+<script lang="ts">
```
nmattia, line 1: A line or two giving an overview of what the component does and how it does it would be greatly appreciated

nmattia: Also a word on two on why we don't use the stock component would be nice

frontend/svelte/src/lib/components/Input.svelte
```diff
@@ -0,0 +1,82 @@
...
+  export let inputType: "number" | "text" = "number";
+  export let required: boolean = true;
+  export let spellcheck: boolean | undefined = undefined;
+  export let step: number | "any" | undefined = undefined;
+
+  export let value: string | number | undefined = undefined;
+
+  export let placeholderLabelKey: string;
+
+  const handleInput = ({ currentTarget }: InputEventHandler) =>
```
nmattia, line 14: What happens if `inputType === "text"` and I input `2`? What is `typeof value` then?

frontend/svelte/src/lib/components/Input.svelte
```diff
@@ -0,0 +1,82 @@
...
+      ? +currentTarget.value
+      : currentTarget.value);
+
+  $: step = inputType === "number" ? step || "any" : undefined;
+</script>
+
+<div class="input-block">
+  <input
+    type={inputType}
+    {required}
```
nmattia, line 25: is there a way we could just forward the attributes here as opposed to listing them all?

frontend/svelte/src/global.d.ts
peterpeterparker: Comment added:

// The type declaration of the input event is neither defined in node types nor in svelte.
// This extends the event with the currentTarget that is provided by the browser and that can be used to retrieve the input value.

frontend/svelte/src/lib/components/GetICPs.svelte
peterpeterparker: validForm renamed into `invalidForm`

frontend/svelte/src/lib/components/GetICPs.svelte
peterpeterparker: Forgot the todo, my plan was to add the check when I'll implement the effective submit function. TODO added.

// TODO: L2-188 - check if form is valid i.e. !invalidForm

frontend/svelte/src/lib/components/Input.svelte
peterpeterparker: The component will be soon moved to a dedicated `ui` folder that will contains all the UI - design system - components of the app. De facto it will be clear that the component is a custom UI component.

Therefore not sure what would be the added value of the comment?

frontend/svelte/src/lib/components/Input.svelte
peterpeterparker: if `inputType === "text"` and user type 2 in the inputfield then output `value` is `"2"`.
if `inputType === "number"` and user type 2 in the inputfield then output `value` is `2`.

frontend/svelte/src/lib/components/Input.svelte
peterpeterparker: I thought about it too but we would need to list the attributes anyway (I am not aware of an  existing out of the box enum of attributes) and in addition there is some logic, like for the `step`. that's why, I went with a selected list of properties (which are all tested). ok?

frontend/svelte/src/lib/components/GetICPs.svelte
peterpeterparker: done

frontend/svelte/src/lib/components/GetICPs.svelte
peterpeterparker: From the `<Modal />`. Close button or click on the backdrop.

frontend/svelte/src/lib/components/Input.svelte
nmattia: I'm confused then, I thought that if I typed `2`, the value would match the regex `/^(number)$/` and thus would be set to `+currentTarget.value` where `+` would effectively convert it to a `number`. I'm guessing I'm wrong about one of those steps?

nmattia: It's `inputType` that's being matched 🤦 . Why not `=== "number"`?

peterpeterparker: We might add some more types in the future (file, range, date, etc).

I am confident it works out because I tested it and the provided tests cover quite some cases including simulating the user typing characters in the field if set to number or text.

nmattia: > We might add some more types in the future (file, range, date, etc).

Then still, why regex-match instead of checking the value? Also AFAICT you're matching the string `"number"` exactly aren't you?

nmattia:
Note: you might want to cherry-pick https://github.com/dfinity/nns-dapp/pull/397/commits/c0f76cf6061e7e1af26e0b133ac7714f025b668d to fix the docker build

frontend/svelte/src/lib/components/Input.svelte
peterpeterparker: `inputType === "number"` done

nmattia: Thanks!

nmattia (approved):
LGTM, thanks!

PR #391, opened 2022-02-01
feat: Svelte neuron card

lmuntanerDfinity:
# Motivation

Create the card summary of a Neuron. Just the UI with hardcoded details.

# Changes

- Changed `Card` can be clickable.
- Create `NeuronCard` component. Hard-coded information for now.
- `NeuronCard` is in `components/neurons`, in advance of [L2-217](https://dfinity.atlassian.net/browse/L2-217).
- Added two new icons: Lock with time and closed lock.


lmuntanerDfinity:
@peterpeterparker @bitdivine just my first PR to see how many things I got wrong: coding style, PR comment, missing information, ...

frontend/svelte/src/lib/components/Card.svelte
```diff
@@ -1,14 +1,33 @@
...
-    <slot name="end" />
-  </div>
-
-  <slot />
-</article>
+<script lang="ts">
+  export let to: string = undefined;
+</script>
+
+{#if to}
```
peterpeterparker, line 5: 1. actually, after thinking about it and now that I see the code, I think we can / should stick to the `article` with a role

```
<article role="button" ...>
```

and apply the mixin `tappable`.

2. The card can be something else than a link aka a button. Therefor it should `dispatch` an event.

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -0,0 +1,70 @@
+<script lang="ts">
+  import IconLockClock from "../../icons/IconLockClock.svelte";
+  import Card from "../Card.svelte";
+</script>
+
+<Card to="/#/neurons/one">
```
peterpeterparker, line 6: add a `<!-- TODO. ... -->` so that we do not forget ;)

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -0,0 +1,70 @@
...
+  <p class="duration">3 Years, 64 Days Dissolve Delay</p>
+</Card>
+
+<style lang="scss">
+  h4,
+  h3 {
+    line-height: 1.3;
+  }
+
+  p {
```
peterpeterparker, line 30: you can drop the `p {margin: 0}` it will be fixed with my open PR that set globally such margin for paragraphs

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -0,0 +1,70 @@
...
+    margin: 0;
+  }
+
+  .lock {
```
peterpeterparker, line 34: feels like you can spare some `class` probably. style is automatically scoped so you can target for example `h3` directly. likewise with scss you can get children.

frontend/svelte/src/lib/icons/IconLockClock.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
+  type Size = "m" | "s";
+  export let size: Size = "m";
+</script>
+
+<!-- source: https://material.io/icons/ -->
+<!-- <svg
```
peterpeterparker, line 7: should we remove the commented svg?

frontend/svelte/src/routes/Neurons.svelte
```diff
@@ -21,16 +22,16 @@
   onDestroy(unsubscribe);
 </script>
 
-{#if !process.env.REDIRECT_TO_LEGACY}
-  <Layout>
-    <section>
-      <h1>{$i18n.neurons.title}</h1>
+<Layout>
```
peterpeterparker, line 25: the `{#if !process.env.REDIRECT_TO_LEGACY}` most probably should not be dropped yet?

peterpeterparker (changes requested):
Thanks! I'll go for a run, if we should discuss some stuffs we can have a look together in a call tomorrow :)

peterpeterparker:
Forgot to mention: can you add some spec tests? like for the new card action and to check that a card is rendered in the "neurons" page

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
nmattia: and create a ticket, and link to it :)

frontend/svelte/src/lib/icons/IconLockClock.svelte
lmuntanerDfinity: oh yes!

frontend/svelte/src/lib/icons/IconLockClock.svelte
```diff
@@ -0,0 +1,39 @@
...
+    d="M18 8h-1V6c0-2.76-2.24-5-5-5S7 3.24 7 6v2H6c-1.1 0-2 .9-2 2v10c0 1.1.9 2 2 2h12c1.1 0 2-.9 2-2V10c0-1.1-.9-2-2-2zm-6 9c-1.1 0-2-.9-2-2s.9-2 2-2 2 .9 2 2-.9 2-2 2zm3.1-9H8.9V6c0-1.71 1.39-3.1 3.1-3.1 1.71 0 3.1 1.39 3.1 3.1v2z"
+  />
+</svg> -->
+<svg
+  width="32"
+  height="33"
+  viewBox="0 0 32 33"
+  fill="currentColor"
+  xmlns="http://www.w3.org/2000/svg"
+  class={size}
```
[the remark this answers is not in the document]
lmuntanerDfinity: sure

frontend/svelte/src/routes/Neurons.svelte
lmuntanerDfinity: ah, ok

lmuntanerDfinity:
@peterpeterparker Did the changes, except for the Card with `role="button"`. We can talk about this one if you want.

frontend/svelte/src/lib/components/Card.svelte
peterpeterparker: We need `click` events when the card is use in modal (just had a look, like when you select an account in a modal) so even though `<a/>` would be semantically more correct in case of navigation, I rather like  `<article role="button" ...>` at the end because it makes the code simpler to maintain. 

Following:

```
if link
 <a... 
else if 
 <button...
else
  <article...
```

would be ok if the content of each a, button and article would not be copy/pasted (as currently in your pr) but extracted to a component (don't think you can create tag from string as you can do in jsx). also then we would have to check that `slot` can passed through two children (I never tried out yet).

what do you think?

note: events are alright,  we have functions to navigate programmatically

lmuntanerDfinity: ok, let's go with the `<article role="button"`.

frontend/svelte/src/lib/components/Card.svelte
```diff
@@ -1,4 +1,10 @@
-<article>
+<script lang="ts">
+  export let role: string = undefined;
+
+  const clickableRoles = ["button", "link"];
+</script>
+
+<article {role} on:click class={clickableRoles.includes(role) && "clickable"}>
```
lmuntanerDfinity, line 7: @peterpeterparker what do you think?

From the [docs](https://svelte.dev/docs#template-syntax-element-directives-on-eventname):
```
If the on: directive is used without a value, the component will forward the event, meaning that a consumer of the component can listen for it.
```

I guessed it made more sense to avoid code and in the end even make it more readable when being used. `<Card on:click` looks also nice. 😄

lmuntanerDfinity:

frontend/svelte/src/lib/components/Card.svelte
peterpeterparker: looks cool

peterpeterparker (approved):
LGTM, thanks!

frontend/svelte/src/lib/components/Card.svelte
```diff
@@ -1,4 +1,14 @@
-<article>
+<script lang="ts">
+  // check testing for firing events
```
nmattia, line 2: Is this comment related to the code below or is this a TODO? What is "check testing"?

frontend/svelte/src/lib/components/Card.svelte
```diff
@@ -1,4 +1,14 @@
...
+  export let role: string | undefined = undefined;
```
nmattia, line 3: Is this ever not a `link`, except in the tests? Otherwise I'd rather keep things simple until we need complexity

nmattia: Also, I see `AccountCard` also uses the `Card`. What's its `role`? Is it OK for the `Card` to have a `role` that's `undefined`?

nmattia: Finally, should `role` really be a `string | undefined`? What values can it have?

frontend/svelte/src/lib/components/Card.svelte
```diff
@@ -8,7 +18,12 @@
 </article>
 
 <style lang="scss">
+  @use "../themes/mixins/interaction";
+
   article {
+    display: block;
```
nmattia, line 24: isn't this the default for `article` elements?

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -0,0 +1,69 @@
+<script lang="ts">
+  import IconLockClock from "../../icons/IconLockClock.svelte";
+  import Card from "../Card.svelte";
+
+  const goToDetail = () => {
+    // Go to the detail page of Neuron
```
nmattia, line 6: Is this a todo?

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -0,0 +1,69 @@
...
+  };
+</script>
+
+<!-- TODO L2-218 -->
+<Card role="link" on:click={goToDetail} ariaLabel="go to neuron details">
+  <div slot="start" class="lock">
+    <!-- TODO data from props: L2-218 -->
+    <h4>14225061177570733325</h4>
```
nmattia, line 14: nitpick: I do wonder if semantically this should be `h4`/`h5` etc

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -0,0 +1,69 @@
...
+      gap: 0.3rem;
+    }
+  }
+
+  .white {
+    color: var(--background-contrast);
+  }
+
+  .currency {
+    display: flex;
```
nmattia, line 56: nit: I wonder if flex isn't too big a hammer when you could probably `text-align`

frontend/svelte/src/lib/icons/IconLockClock.svelte
```diff
@@ -0,0 +1,33 @@
...
+  const widths = {
+    s: 18,
+    m: 24,
+    l: 32,
+  };
+
+  const heights = {
+    s: 18,
+    m: 24,
+    l: 33,
```
nmattia, line 16: is this meant to be `33`? Looks a bit odd, I would have guessed `32`. Also, could we get away with just preserving the aspect ratio?

frontend/svelte/src/lib/icons/IconLockClock.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
+  import type { Size } from "../types/shared";
+
+  export let size: Size = "m";
```
nmattia, line 4: what about actually writing out the words? It took me a few second to realize those meant "small", "medium" and "large". Also, I thought `l` was a 1!

frontend/svelte/src/lib/types/shared.ts
```diff
@@ -0,0 +1 @@
+export type Size = "m" | "s" | "l";
```
nmattia, line 1: what about actually writing out the words? It took me a few second to realize those meant "small", "medium" and "large". Also, I thought `l` was a 1!

frontend/svelte/src/lib/icons/IconLockClock.svelte
nmattia, line 4: Is this ever not a `small` icon? Again, I'd prefer to avoid complexity until it's needed

frontend/svelte/src/lib/types/i18n.d.ts
```diff
@@ -8,6 +8,7 @@ interface I18nCore {
 
 interface I18nError {
   auth_sync: string;
+  sign_in: string;
```
nmattia, line 11: Where is this used?

nmattia (commented):
Few comments!

frontend/svelte/src/lib/components/Card.svelte
lmuntanerDfinity: It can be removed, it was a WIP todo for me, sorry

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: Yes, to be done in L2-218, same as all the others

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -0,0 +1,69 @@
...
+  import IconLockClock from "../../icons/IconLockClock.svelte";
+  import Card from "../Card.svelte";
+
+  const goToDetail = () => {
+    // Go to the detail page of Neuron
+  };
+</script>
+
+<!-- TODO L2-218 -->
+<Card role="link" on:click={goToDetail} ariaLabel="go to neuron details">
```
lmuntanerDfinity, line 11: @nmattia you were asking when do we use the role. Here.

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: Good question. I am also not sure. I chose h4 here because the one afterwards, Locked, could also me considered a heading, but it seems to me lower than this ID.

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: This might change with L2-218 when I use the ICP component.

frontend/svelte/src/lib/icons/IconLockClock.svelte
lmuntanerDfinity: makes sense to me

lmuntanerDfinity: That I know of, for now it's only small, yes. I can keep it always with 18.

I also prefer avoiding flexibility, but the svg I got from the previous project was this size, I assumed it was also used with it.

frontend/svelte/src/lib/icons/IconLockClock.svelte
lmuntanerDfinity: This is to preserve the `viewBox="0 0 32 33"`. I got this svg from the previous project.

frontend/svelte/src/lib/types/i18n.d.ts
lmuntanerDfinity: This was added by prettier

frontend/svelte/src/lib/types/shared.ts
lmuntanerDfinity: We might remove this, since we only use one size

frontend/svelte/src/lib/components/Card.svelte
lmuntanerDfinity: yes, true. I think I had an anchor tag before, instead of the `role` and needed this. But not anymore.

frontend/svelte/src/lib/components/Card.svelte
lmuntanerDfinity: I mentioned you where "role" is used.

Good point, I will change the type to allowed values

lmuntanerDfinity:
@nmattia ready for another review 😄

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
nmattia: Right, I was asking when this is _not_ a `link`. Is it _ever_ a `button`?

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
nmattia: Then I'd suggest we stick to something else, I rarely see anything beyond h3. But really this one is up to @peterpeterparker

nmattia: (really feel free to leave as is)

nmattia (approved):
LGTM, sorry for the delay!

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
peterpeterparker: yeah these information looks like `<p/>` to me too - i.e. not titles

something we can update when the implementation of the dynamic data will find place (L2-218)

lmuntanerDfinity: I will make a note on the issue to not forget about it

PR #392, opened 2022-02-02
Update payload definition for UpdateNodeOperatorConfigPayload

hpeebles:
This PR adds some recently added fields to the payload definition.

This is required so that the NNS Dapp can deserialize the candid serialized payload and subsequently render it as human readable JSON.


frontend/ts/src/canisters/governance/nnsFunctions/payloads.did
```diff
@@ -92,6 +92,8 @@ type UpdateIcpXdrConversionRatePayload = record {
 type UpdateNodeOperatorConfigPayload = record {
   node_operator_id : opt principal;
   node_allowance : opt nat64;
+  rewardable_nodes : vec record { text; nat32 };
+  dc_id : opt text;
```
bitdivine, line 96: The [corresponding protobuf](protobuf/def/registry/node_operator/v1/node_operator.proto) has one more field: `node_provider_principal_id`.  Can we include it or a comment stating why it has been excluded, please?

frontend/ts/src/canisters/governance/nnsFunctions/payloads.js
```diff
@@ -103,6 +103,8 @@ export const UpdateIcpXdrConversionRatePayload = IDL.Record({
 export const UpdateNodeOperatorConfigPayload = IDL.Record({
   node_operator_id: IDL.Opt(IDL.Principal),
   node_allowance: IDL.Opt(IDL.Nat64),
+  rewardable_nodes: IDL.Vec(IDL.Tuple(IDL.Text, IDL.Nat32)),
+  dc_id: IDL.Opt(IDL.Text),
```
bitdivine, line 107: I guess this is autogenerated from the above so might not have a comment if the field is intentionally omitted.

rs/nns_functions_candid_gen/src/payloads.rs
```diff
@@ -235,6 +237,15 @@ pub struct UpdateNodeOperatorConfigPayload {
...
     pub node_allowance: Option<u64>,
+
+    /// The ID of the data center where this Node Operator hosts nodes.
+    #[prost(message, optional, tag = "3")]
+    pub dc_id: Option<String>,
+
+    /// A map from node type to the number of nodes for which the associated
+    /// Node Provider should be rewarded.
+    #[prost(btree_map = "string, uint32", tag = "4")]
+    pub rewardable_nodes: BTreeMap<String, u32>,
```
bitdivine, line 248: Ditto with `node_provider_principal_id`.

frontend/ts/src/canisters/governance/nnsFunctions/payloads.did
```diff
@@ -92,6 +92,8 @@ type UpdateIcpXdrConversionRatePayload = record {
...
 };
 type UpdateNodeRewardsTableProposalPayload = record {
   new_entries : vec record { text; NodeRewardRates };
```
bitdivine, line 99: Out of curiosity, where does this file come from?

I see only one did file in the ic repo under rs/nns/governance and that seems to be quite different from this file here.

Does the did file here correspond to an API between the frontend and the governance canister directly or does the nns-dapp backend act as a proxy, getting all the related data?  I'm expecting or hoping that whichever canister is being called will have a matching did file.

bitdivine (approved):

bitdivine:
We discussed deploying this to mainnet, as a method of testing.

This doesn't work because the canister is too big:
```
Installing code for canister nns-dapp, with canister_id fcmxc-hiaaa-aaaai-qfnga-cai
The replica returned an HTTP Error: Http Error: status 400 Bad Request, content type "", content: Request 0x61b23c0de7c815d8de09abbe9ed20456131d39a3ffb6f6d51ded072c40795152 is too large. Message bytes 3039222 is bigger than the max allowed 2097152.
```

On a side note, creating the canister was considerably harder than on previous occasions, deploying other canisters.  The transition to 0.9.0 seems quite rough and error messages are not super helpful.

PR #395, opened 2022-02-03
Footer and toolbar  l2 213

mstrasinskis:
# Motivation
Create the bottom toolbar element.
image
[L2-213](https://dfinity.atlassian.net/jira/software/c/projects/L2/boards/182?modal=detail&selectedIssue=L2-213)

# Changes
- presented SectionWithToolbar (with the `toolbar` slot)
- <section> was replaced with <SectionWithToolbar> to add toolbar support and bottom gradient

# Note

- @peterpeterparker button future is not clear. Especially the custom width.
- This is a **SECOND** version. The **FIRST** one `6d18e4274f2e8d53d4b4ca54058716c5efa6ce7e` adds only the custom `<Footer>` w/o more DOM changes:

**FIRST version advantages**
- cleaner DOM
- less changes

**SECOND version advantages**
- no weird place-holder workaround
- no `position: fixed` used. So the section DOM is ready to be animated

# On Mobile

image


frontend/svelte/src/routes/Accounts.svelte
```diff
@@ -22,12 +23,17 @@
     async (accounts: AccountsStore) => (main = accounts?.main)
   );
 
+  // TODO: TBD
```
nmattia, line 26: thanks! please create a ticket and link from here (applies to other TBDs as well)

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
```diff
@@ -0,0 +1,99 @@
...
+<style lang="scss">
+  section {
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+
+    // wrapper for scrollable content
+    & > .wrapper {
```
nmattia, line 23: aren't svelte styles scoped to the component anyway? could this be any other wrapper?

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
```diff
@@ -0,0 +1,99 @@
+<section>
```
nmattia, line 1: Just wondering, wouldn't it be simpler (and more flexible) to only have a toolbar component?

frontend/svelte/src/lib/i18n/en.json
```diff
@@ -26,12 +26,15 @@
...
+    "copy_identifier": "Copy identifier to clipboard",
+    "new_transaction": "New Transaction",
+    "add_account": "Add Account"
   },
   "neurons": {
     "title": "Neurons",
     "text": "Earn rewards by staking your ICP in neurons. Neurons allow you to participate in governance on the Internet Computer by voting on Network Nervous System (NNS) proposals.",
-    "principal_is": "Your principal id is"
+    "principal_is": "Your principal id is",
+    "stack_neurons": "Stack Neurons"
```
nmattia, line 37: did you mean "stake"? (here and elsewhere)

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
mstrasinskis: It's done this way only because of performance 
image
But agree, in this particular case it doesn't give any advantage.

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
mstrasinskis: It would be simpler but then the toolbar should be outside the `<section>` (not sure because of semantics)

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
nmattia: Oh, TIL! thanks

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
peterpeterparker: I would rather like a toolbar component too.

Instead of

```
<Layout>
     <SectionWithToolbar>
          ...
```

I find following more readable

```
<Layout>
    ...
    
    <Toolbar />
```

It also follows other design system approach I know e.g. Ionic.

If outside the `<section>`then we can use a `<footer>`  to encapsulate the toolbar, maybe not that bad for semantic?

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
```diff
@@ -0,0 +1,99 @@
...
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+
+    // wrapper for scrollable content
+    .wrapper {
+      position: absolute;
+      top: 0;
```
peterpeterparker, line 25: assuming we create a `<Toolbar />`, this will moves but if not, `inset`  can be used to shorten the style

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
```diff
@@ -0,0 +1,99 @@
...
+    // bottom gradient
+    &::after {
+      content: "";
+      position: absolute;
+      bottom: 0;
+      left: 0;
+      right: 0;
+      height: var(--section-toolbar-height);
+
+      z-index: 1;
```
peterpeterparker, line 50: there is a global `--z-index` (I think) props. can be useful to use it everywhere when we set z-index so that we can find these easily if we ever have to push things up or down.

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
```diff
@@ -0,0 +1,99 @@
...
+
+  .toolbar {
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    right: 0;
+
+    height: var(--section-toolbar-height);
+
+    z-index: 2;
```
peterpeterparker, line 70: likewise `calc(--z-index....`

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
```diff
@@ -0,0 +1,99 @@
...
+
+    display: flex;
+    justify-content: center;
+    align-items: end;
+
+    // buttons
+    & > :global(button) {
+      pointer-events: all;
+
+      // TODO: remove <tmp_styles>
```
peterpeterparker, line 83: you are right, global style for button will come with https://github.com/dfinity/nns-dapp/pull/390

once 390, we can probably defined a `primary` class or something so that no just `submit` buttons are blue but does we want

frontend/svelte/src/lib/themes/theme.scss
```diff
@@ -49,13 +49,6 @@ main {
   color: var(--gray-200);
 }
 
-section {
-  padding: calc(2 * var(--padding)) calc(4 * var(--padding));
```
peterpeterparker, line 53: I guess this will be reverted if we create `<Toolbar />`  cmp right?

peterpeterparker:
@mstrasinskis cool pr 👍

can you provide a screenshot of mobile - smaller - devices ? how does it looks like if two buttons are added to the toolbar?
still enough space or should we display them in a column?

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
```diff
@@ -0,0 +1,99 @@
+<section>
+  <div class="wrapper">
+    <div class="content">
+      <slot />
+    </div>
+  </div>
+  {#if $$slots.toolbar}
```
lmuntanerDfinity, line 7: Cool, I didn't know about this!

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
lmuntanerDfinity: Are the semantics worth losing the flexibility?

frontend/svelte/src/routes/Accounts.svelte
mstrasinskis: Done. Good point 👍

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
mstrasinskis: Updated

frontend/svelte/src/lib/i18n/en.json
mstrasinskis: Done 🤦‍♂️

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
mstrasinskis: Thank you very much for your inputs. If no animation needed then the `<Toolbar>` would be definitely the most flexible solution. Will be refactored with the next commit 👍

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
mstrasinskis: Good point!

frontend/svelte/src/lib/components/SectionWithToolbar.svelte
mstrasinskis: Thanks, haven't noticed it.

peterpeterparker (approved):
As discussed with @mstrasinskis in a chat the last improvements and introduction of the `<Toolbar />` component looks good to me.

In addition, it is worth to note we replicate current production design with a `<footer />` (background gradient) that is displayed on every pages of the app even if there is no buttons. 

It can be challenged but if we do so, then in another PR 😉.

Finally, we agreed the "global button styling" should be improved, specially the spacing on mobile devices. Some improvements are provided in another PR (#390) that is not yet reviewed. Therefore, to go ahead, we opened a new JIRA Task (L2-229). As soon as this other is merged, @mstrasinskis will takeover the task and develop the necessary adjustements.

frontend/svelte/src/lib/components/ui/Toolbar.svelte
```diff
@@ -0,0 +1,44 @@
...
+    justify-content: center;
+    align-items: end;
+
+    @media (min-width: 300px) {
+      min-width: 280px;
+      width: 70%;
+    }
+
+    @media (min-width: 768px) {
+      margin-bottom: calc(3 * var(--padding));
```
peterpeterparker, line 25:
```
// same as section
      padding: 0 calc(3 * var(--padding));
      // same as section
      padding: 0 calc(3 * var(--padding));
      ```
      
      shouldn't be `4` now?

peterpeterparker (approved):

PR #396, opened 2022-02-03
L2-214: Dark Modal

lmuntanerDfinity:
# Motivation

Add the dark version of the modal.

# Changes

- Add `dark` prop to `Modal` component.
- Changed `Modal` title to h3. Previous Modal had different sizes for dark and light modals. Some dark modals had long titles. I changed to a midway solution. Old dark modal title was too small.

### DIFF
--- a/frontend/svelte/src/lib/i18n/en.json
+++ b/frontend/svelte/src/lib/i18n/en.json
@@ -65,5 +65,8 @@
     "accepted": "Adopted",
     "executed": "Executed",
     "failed": "Failed"
+  },
+  "modals": {
+    "back": "Back"
   }
 }
--- a/frontend/svelte/src/lib/modals/Modal.svelte
+++ b/frontend/svelte/src/lib/modals/Modal.svelte
@@ -10,6 +10,7 @@
   export let theme: "dark" | "light" = "light";
   // There is no way to know to know whether a parent is listening to the "nnsBack" event
   // https://github.com/sveltejs/svelte/issues/4249#issuecomment-573312191
+  // Please do not use `showBackButton` without listening on `nnsBack`
   export let showBackButton: boolean = false;
 
   let dark: boolean;
--- a/frontend/svelte/src/lib/modals/Modal.svelte
+++ b/frontend/svelte/src/lib/modals/Modal.svelte
@@ -3,16 +3,27 @@
   import { quintOut } from "svelte/easing";
   import IconClose from "../icons/IconClose.svelte";
   import { createEventDispatcher } from "svelte";
+  import IconBack from "../icons/IconBack.svelte";
+  import { i18n } from "../stores/i18n";
 
   export let visible: boolean = false;
+  export let theme: "dark" | "light" = "light";
+  // There is no way to know to know whether a parent is listening to the "nnsBack" event
+  // https://github.com/sveltejs/svelte/issues/4249#issuecomment-573312191
+  export let showBackButton: boolean = false;
+
+  let dark: boolean;
+  $: dark = theme === "dark";
 
   const dispatch = createEventDispatcher();
   const close = () => dispatch("nnsClose");
+  const back = () => dispatch("nnsBack");
 </script>
 
 {#if visible}
   <div
     class="modal"
+    class:dark
     transition:fade
     role="dialog"
     aria-labelledby="modalTitle"
@@ -24,9 +35,15 @@
       class="wrapper"
     >
       <div class="toolbar">
-        <h2 id="modalTitle"><slot name="title" /></h2>
-
-        <button on:click|stopPropagation={close} aria-label="Close"
+        {#if showBackButton}
+          <button
+            class="back"
+            on:click|stopPropagation={back}
+            aria-label={$i18n.modals.back}><IconBack /></button
+          >
+        {/if}
+        <h3 id="modalTitle"><slot name="title" /></h3>
+        <button on:click|stopPropagation={close} aria-label={$i18n.core.close}
           ><IconClose /></button
         >
       </div>
@@ -46,6 +63,28 @@
     inset: 0;
 
     z-index: calc(var(--z-index) + 998);
+
+    &.dark {
+      color: var(--background-contrast);
+
+      .wrapper {
+        background: none;
+      }
+
+      .toolbar {
+        background: var(--gray-50-background);
+        box-shadow: 0 2px 8px var(--background);
+
+        h3,
+        button {
+          color: var(--background-contrast);
+        }
+      }
+
+      .content {
+        background: var(--gray-50-background);
+      }
+    }
   }
 
   .backdrop {
@@ -85,20 +124,23 @@
     background: var(--gray-100);
     color: var(--gray-800);
 
-    display: flex;
-    justify-content: space-between;
+    display: grid;
+    grid-template-columns: var(--icon-width) 1fr var(--icon-width);
 
-    h2 {
+    h3 {
       color: inherit;
       font-weight: 400;
       margin-bottom: 0;
       line-height: 1.5;
+      text-align: center;
+      grid-column-start: 2;
     }
 
     button {
       display: flex;
       justify-content: center;
       align-items: center;
+      padding: 0;
 
       &:active,
       &:focus,
--- a/frontend/svelte/src/lib/components/Checkbox.svelte
+++ b/frontend/svelte/src/lib/components/Checkbox.svelte
@@ -4,7 +4,7 @@
   export let inputId: string;
   export let checked: boolean;
 
-  export let color: "dark" | "light" = "light";
+  export let theme: "dark" | "light" = "light";
   export let text: "block" | "inline" = "inline";
 
   export let selector: string | undefined = undefined;
@@ -19,7 +19,7 @@
 
 <div
   on:click|preventDefault={onClick}
-  class={`checkbox ${color} ${selector || ""}`}
+  class={`checkbox ${theme} ${selector || ""}`}
 >
   <label for={inputId} class={text}><slot /></label>
   <input
--- a/frontend/svelte/src/lib/components/VotingFilters.svelte
+++ b/frontend/svelte/src/lib/components/VotingFilters.svelte
@@ -46,7 +46,7 @@
 <Checkbox
   inputId="hide-unavailable-proposals"
   checked={false}
-  color="dark"
+  theme="dark"
   text="block"
   selector="hide-unavailable-proposals"
   >{$i18n.voting.hide_unavailable_proposals}</Checkbox
--- a/frontend/svelte/src/lib/icons/IconBack.svelte
+++ b/frontend/svelte/src/lib/icons/IconBack.svelte
@@ -1,0 +1,11 @@
+<!-- source: https://material.io/icons/ -->
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  width="24px"
+  height="24px"
+  viewBox="0 0 24 24"
+  fill="currentColor"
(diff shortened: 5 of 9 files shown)


lmuntanerDfinity:
@nmattia @mstrasinskis please review

Here are the light and dark versions:

Screenshot 2022-02-03 at 13 50 00

Screenshot 2022-02-03 at 13 50 24

frontend/svelte/src/lib/modals/Modal.svelte
```diff
@@ -24,11 +26,13 @@
       class="wrapper"
     >
       <div class="toolbar">
-        <h2 id="modalTitle"><slot name="title" /></h2>
+        <h3 id="modalTitle"><slot name="title" /></h3>
 
-        <button on:click|stopPropagation={close} aria-label="Close"
-          ><IconClose /></button
-        >
+        <div class="button-container">
```
peterpeterparker, line 31: don't get why you wrap the close button in a `div` with an `absolute` positioning?
what cannot be achieve with the current `flex` display?

peterpeterparker (changes requested):

frontend/svelte/src/lib/modals/Modal.svelte
lmuntanerDfinity: I tried, but then the button was not aligned with the title. I used this wrapper to set it the the height of the parent and then align the button in the center.

lmuntanerDfinity: Any idea of how to improve this?

peterpeterparker:
For accessibility reason shouldn't we define a base `color` for the content of the "dark" modal? 

In your first screenshot the content is not contrasting enough.

lmuntanerDfinity:
Good point, I designed in the way that the color is defined in the content, but good point, I can change the color of the content as well.

frontend/svelte/src/lib/modals/Modal.svelte
peterpeterparker: wasn't it previously?

we can have a look next week together.

generally I like to spare dom elements specially those for styling purpose, worth a shot.

lmuntanerDfinity: yes, but now I wanted the title in the middle, like in the dark mode.

But the middle of the whole width, without considering of the button. Therefore, I used display: flex, and to remove the button from the normal flow, I used `position: absolute`. But `position: absolute;` pushed the button to the top.

One option was to hardcode how far from the top the button lives with just `top: XXX`. But I thought this solution is not very responsive. Therefore, I added the extra `div`, that starts at `top: 0` and height of the parent and places the contents (button) in the middle.

peterpeterparker: a display `grid` might solve this but in the nns rewrite, we follow the design of the current app.

however if we see small improvements, we shall proceed. in this case, i see an improvement in constancy if both types of modal present the title with the same alignment. 

finally:

ios = modal title center
material design = dialog & modal title left

since we target both mobile and desktop, kind of feel we follow a style closer to material design.

that's why I suggest then to revert and keep the `space-between` aka align the title to the left in the dark modal as well.

peterpeterparker: speaking of, I just realize the "dark" modal might contains a back button, so we might need a grid anyway 🤪

do you want to foreseen such slot in this PR or should we do this afterwards?

lmuntanerDfinity: And what do we do with the "back arrow", see example below:

Screenshot 2022-02-03 at 15 32 13

That is not in the Material dialogs...

lmuntanerDfinity: I can use a grid, makes sense to me. I can do it here.

peterpeterparker: Yeah would look this this in material-ich

152362902-636fb786-f542-445a-909b-e440289f4c40
ch

let's try grid 🤘

frontend/svelte/src/lib/modals/Modal.svelte
```diff
@@ -85,20 +118,32 @@
     background: var(--gray-100);
     color: var(--gray-800);
 
-    display: flex;
-    justify-content: space-between;
+    display: grid;
```
lmuntanerDfinity, line 121: @peterpeterparker here is the solution with grid.

I avoided using an extra DOM element by using a `margin-left` on the title to make up for the missing back button.

Not super happy, but it's the best idea I got.

frontend/svelte/src/lib/modals/Modal.svelte
```diff
@@ -85,20 +118,32 @@
...
+    &.showBackButton {
+      grid-template-columns: var(--icon-width) 1fr var(--icon-width);
+
+      h3 {
+        margin-left: 0;
+      }
+    }
 
-    h2 {
+    h3 {
```
mstrasinskis, line 132: May I ask why the headline was degraded?

frontend/svelte/src/lib/modals/Modal.svelte
peterpeterparker: cool no extra dom element 👍

regarding spacing, `grid-column-start` ?

```
display: grid;
    grid-template-columns: var(--icon-width) 1fr var(--icon-width);

    h3 {
      color: inherit;
      font-weight: 400;
      margin-bottom: 0;
      line-height: 1.5;
      text-align: center;
      grid-column-start: 2;
    }
```

peterpeterparker: with above `.showBackButton` can be removed too

frontend/svelte/src/lib/modals/Modal.svelte
lmuntanerDfinity: The be consistent with Dark and Light modals. In the previous design, the light modal had a way bigger title than the light because in the dark some titles were longer.

To be consistent and still allow for longer titles, I think it made sense to reduce.

Do you think that semantically, a h2 is better fitted here?

frontend/svelte/src/lib/modals/Modal.svelte
lmuntanerDfinity: good point!

lmuntanerDfinity: @peterpeterparker change done!

frontend/svelte/src/lib/modals/Modal.svelte
mstrasinskis: Thank for description. Semantically I think h2 would be more preferable but it's totally not critical especially if we rebuild everything later :D

frontend/svelte/src/lib/modals/Modal.svelte
peterpeterparker: cool 👍

lmuntanerDfinity:
@peterpeterparker @nmattia @mstrasinskis ready for another review

frontend/svelte/src/lib/modals/Modal.svelte
```diff
@@ -24,8 +31,12 @@
       class="wrapper"
     >
       <div class="toolbar">
-        <h2 id="modalTitle"><slot name="title" /></h2>
-
+        {#if showBackButton}
+          <button class="back" on:click|stopPropagation={back} aria-label="Back"
```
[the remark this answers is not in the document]
lmuntanerDfinity: Done!

lmuntanerDfinity:
@peterpeterparker ready for review

peterpeterparker:
I did a try locally, looks good 👍.

However, doing so I noticed an inconsistency in the declaration of the dark property.

In this PR - in the `<Modal />`:

```
export let dark: boolean = false;
```

In the existing `<Checkbox />`:

```
export let color: "dark" | "light" = "light";
```

I think it would be good to have a consistent approach.

I kind of rather like the one of the checkbox because it feels maybe a bit more readable when used (`color="light"` vs `dark={false}`). What do you think @lmuntanerDfinity, any preferences?

frontend/svelte/src/lib/i18n/en.json
```diff
@@ -64,5 +64,9 @@
     "accepted": "Adopted",
     "executed": "Executed",
     "failed": "Failed"
+  },
+  "modals": {
+    "close": "Close",
```
nmattia, line 69: I think we can stick to [a single close](https://github.com/dfinity/nns-dapp/blob/1d5aa2c60606a33ae29c8f39959505a8f96976e9/frontend/svelte/src/lib/i18n/en.json#L3) for now

frontend/svelte/src/lib/modals/Modal.svelte
```diff
@@ -3,16 +3,24 @@
...
   import IconClose from "../icons/IconClose.svelte";
   import { createEventDispatcher } from "svelte";
+  import IconBack from "../icons/IconBack.svelte";
+  import { i18n } from "../stores/i18n";
 
   export let visible: boolean = false;
+  export let dark: boolean = false;
+  // There is no way to know to know whether a parent is listening to the "nnsBack" event
+  // https://github.com/sveltejs/svelte/issues/4249#issuecomment-573312191
+  export let showBackButton: boolean = false;
```
nmattia, line 13: so someone using this modal should only ever `showBackButton=true` if they listen `on:nnsBack`? Can we add a note about this?

lmuntanerDfinity: Another option is to use a callback: "onBack" for example. I can check for the presence of that and show/hide the arrow.

But the callback is not the pattern you are using now.

Where else do you propose to add a note? That's why I added the comment here.

frontend/svelte/src/lib/i18n/en.json
lmuntanerDfinity: 👍

lmuntanerDfinity:
> I did a try locally, looks good 👍.
> 
> However, doing so I noticed an inconsistency in the declaration of the dark property.
> 
> In this PR - in the `<Modal />`:
> 
> ```
> export let dark: boolean = false;
> ```
> 
> In the existing `<Checkbox />`:
> 
> ```
> export let color: "dark" | "light" = "light";
> ```
> 
> I think it would be good to have a consistent approach.
> 
> I kind of rather like the one of the checkbox because it feels maybe a bit more readable when used (`color="light"` vs `dark={false}`). What do you think @lmuntanerDfinity, any preferences?

I don't have a strong opinion on the pattern.

But yes, for the name of the prop. I'd rather call it "theme" than "color". Is that ok?

I can rename Checkbox and change it in the modal.

peterpeterparker:
@lmuntanerDfinity `theme` works for me. let's go for it 👍 . sure please also refactor `<Checkbox />`, thx!

```
export let theme: "dark" | "light" = "light";
```

lmuntanerDfinity:
@peterpeterparker changes applied

peterpeterparker (approved):
@lmuntanerDfinity LGTM, thanks!

I already approve it so that it won't be blocked but please do first resolve @nmattia open questions before merging it.

lmuntanerDfinity:
@peterpeterparker ready for one (last?) review 😅

frontend/svelte/src/lib/modals/Modal.svelte
nmattia: The comment isn't particularly clear, in particular it doesn't explicitly tell you to use both or neither. Maybe a line about this before the `export` list?

lmuntanerDfinity: done!

nmattia (approved):
LGTM!

PR #397, opened 2022-02-04
Buildkit

bitdivine:
# Motivation
The build asset extraction is needlessly complicated.  Docker has an environment variable "buildkit" that if set, copies the final state to an output dir.  It is for ... building things. :smile: 

# Changes
- Replace the canister extraction code with DOCKER_BUILDKIT=1

# Testing
Run `./scripts/docker-build` on:
- [x] Linux (Ubuntu 20.04 LTS) Docker version 20.10.10, build b485636 - DONE OK
- [ ] Mac AMD64 - Running now
  - Breaks on this PR but also on main, on `RUN cargo install --version 0.3.1 ic-cdk-optimizer`.
  - Given that the breakage is unrelated to this PR it is not a blocker, however Mac builds should be fixed.
- [ ] Windows - not sure this is used but maybe...


.dockerignore
```diff
@@ -27,7 +27,6 @@ frontend/dart/ios/Flutter/flutter_export_environment.sh
 frontend/dart/ios/Runner/GeneratedPluginRegistrant.h
 frontend/dart/ios/Runner/GeneratedPluginRegistrant.m
 frontend/dart/lib/generated_plugin_registrant.dart
-frontend/dart/pubspec.lock
```
bitdivine, line 30: Unrelated.  Main is broken and builds will fail without this.

scripts/docker-build
```diff
@@ -33,41 +33,22 @@ image_name="nns-dapp-$DEPLOY_ENV"
 
 echo "The following image name will be used: $image_name"
 
+OUTDIR=out
```
nmattia, line 36: Why copy to OUTDIR first? I'm guessing that docker just won't copy the image content to a non-empty dir; if that's the case, can you add a comment about it?

scripts/docker-build
```diff
@@ -33,41 +33,22 @@ image_name="nns-dapp-$DEPLOY_ENV"
...
+assets=(assets.tar.xz nns-dapp.wasm)
+rm -fr "${OUTDIR:-not-the-file-you-were-looking-for}"
```
nmattia, line 38:
```suggestion
rm -rf "${OUTDIR:-not-the-file-you-were-looking-for}"
```

For consistency with the rest of the world; everything I see `-fr` I need to re-read to make sure I know what this is doing

nmattia (commented):
Very nice!

scripts/docker-build
bitdivine: Because copying the output of a big complicated process straight into ./ is stupid, in the sense that it is begging for a disaster to happen.  It just needs the last stage to be omitted, e.g. by someone setting target, and bingo, a lot more gets copied out than was intended, clobbering any files in the local directory.  I will add a comment to that effect.

bitdivine: Done

scripts/docker-build
bitdivine: I thought it was only French who insisted on -fr rather than -rf? ;-)  But changed anyway, it makes zero difference.  I have worked in infra across many companies for many years and this is the first time that I have encountered anyone trying to enforce a canonical ordering for those flags.

scripts/docker-build
nmattia: Agreed. Thanks!

scripts/docker-build
```diff
@@ -33,41 +33,27 @@ image_name="nns-dapp-$DEPLOY_ENV"
...
+done
 
-if [[ $success = true ]]; then
-  echo "Removing container $container_id"
-  docker rm --volumes "$container_id"
-else
-  echo "container is left for inspection, once done run:"
-  echo "  docker rm -v $container_id"
-fi
+echo FIN
```
nmattia, line 59: ❤️

nmattia (approved):
LGTM, looks great!

scripts/docker-build
```diff
@@ -33,41 +33,27 @@ image_name="nns-dapp-$DEPLOY_ENV"
 
 echo "The following image name will be used: $image_name"
 
+# All the files in the final docker stage will be copied to this directory:
+OUTDIR=out
+rm -rf "${OUTDIR:-not-the-file-you-were-looking-for}"
+# These files will be copied out of OUTDIR into the repository root.
```
dskloetd, line 39: On line 3 it says "This outputs a single file,".
I guess that's no longer the case?

PR #398, opened 2022-02-04
Add NNS function payload definitions for functions 23 and 24

hpeebles:
This PR adds the payload definitions for 2 newly added NNS functions, RemoveNodeOperators (23) and RerouteCanisterRange (24).

The NNS Dapp needs to know these payload definitions so that it can deserialize the candid encoded payloads and then display them in human readable JSON format.


frontend/ts/src/canisters/governance/nnsFunctions/payloads.did
```diff
@@ -68,7 +68,15 @@ type RecoverSubnetPayload = record {
   state_hash : vec nat8;
   time_ns : nat64;
 };
+type RemoveNodeOperatorsPayload = record {
+  node_operators_to_remove : vec vec nat8;
```
bitdivine, line 72: Is the repetition of "vec" intentional?  Looking in the ic repo I see e.g.:

`node_operators_to_remove: vec![node_operator_id_1, node_operator_id_2]`

which looks like a 1-D vec.

hpeebles: Vec of Vec is correct.
It's a Vec of Principal.
And a principal is itself a Vec<u8>.
So combined you get a Vec<Vec<u8>>.

frontend/ts/src/canisters/governance/nnsFunctions/payloads.js
```diff
@@ -78,9 +78,17 @@ export const RecoverSubnetPayload = IDL.Record({
   state_hash: IDL.Vec(IDL.Nat8),
   time_ns: IDL.Nat64,
 });
+export const RemoveNodeOperatorsPayload = IDL.Record({
+  node_operators_to_remove: IDL.Vec(IDL.Vec(IDL.Nat8)),
```
bitdivine, line 82: Can `IDL.Vec(IDL.Nat8)` be written as `IDL.Principal`?

hpeebles: It probably can... but all of this is auto generated.
I simply copy. the rust type definitions into `payloads.rs` and everything else is auto generated.

bitdivine (approved):
Possible quibble around vec<u8> rather than principal but I guess either will render.

PR #399, opened 2022-02-04
Warn untracked

bitdivine:
# Motivation
We have had non-deterministic builds recently.  A possible cause is untracked files being included in the docker build.

# Changes
- List which files are copied in to the build stage
- Diff the above list with files tracked by git.

# Testing
- Running ./scripts/docker-build twice yielded a failure:  The second build was pulling in the asset files from the first run.
- Running on a completely squeaky clean repo passed.


scripts/docker-build
```diff
@@ -56,4 +56,15 @@ for file in "${assets[@]}"; do
   sha256sum "$file"
 done
 
+untracked_inputs() {
+  comm -1 -3 <(git ls-files | sort) <(sort out/build-inputs.txt)
+}
+if untracked_inputs | grep -q .; then
+  echo WARNING: BUILD IS NOT REPEATABLE
```
nmattia, line 63:
```suggestion
  echo WARNING: BUILD MAY NOT BE REPRODUCIBLE
```

I think "repeatable" means you can build the code again, though you might not get the same output; here I suspect you mean "reproducible"

scripts/docker-build
```diff
@@ -56,4 +56,15 @@ for file in "${assets[@]}"; do
...
+  echo The following files were included in the build but are not tracked by git:
```
nmattia, line 64: I'll admit that I find it very convenient, when developing, to include files in the build that are not necessarily tracked by git. Did this actually help with the reproducibility issue?

Dockerfile
```diff
@@ -75,11 +75,15 @@ ARG OWN_CANISTER_ID
 RUN echo "OWN_CANISTER_ID: '$OWN_CANISTER_ID'"
 
 # Build
-COPY . .
+# ... put only git-tracked files in the build directory
+COPY . /build
```
nmattia, line 79: What do you mean by "put only git-tracked files in the build directory"? Looks to me like all files (except the dockerignore-d ones) are added the the build dir

scripts/docker-build
bitdivine: The reason that I chose to report untracked files rather than checking whether git is clean and that .dockerfile does not include anything not in git is precisely to give developers a bit of freedom.  The former is nice in that it fails fast but is probably too restrictive for developers.

Dockerfile
bitdivine: The intention is that only git-tracked files will be in there.  In particular:
- Not files from the guest-os - this is implemented by putting the files in a directory rather than in root
- Not untracked files - this uses a "warn but do not fail" mechanism if untracked files are added here, to give developers space.

nmattia (approved):
Approving as this is useful; I do think it would be a good idea to only fail on `$CI` though

PR #405, opened 2022-02-07
Wait for all img tags to be complete

nmattia:
This makes sure that we wait for the background to be loaded before we
take the screenshots.

### DIFF
--- a/e2e-tests/specs/example.e2e.ts
+++ b/e2e-tests/specs/example.e2e.ts
@@ -4,19 +4,21 @@ describe("landing page", () => {
 
     await browser.$("h1").waitForExist();
 
+    // Wait for all images to be "complete", i.e. loaded
     browser.waitUntil(
       function () {
         return this.execute(function () {
-          let imgs = document.getElementsByTagName("img");
-          if (imgs.length > 0) {
-            return (
-              Array.prototype.every.call(imgs, (img) => {
-                return img.complete;
-              }) && document.readyState === "complete"
-            );
-          } else {
+          const imgs: HTMLCollectionOf<HTMLImageElement> =
+            document.getElementsByTagName("img");
+          if (imgs.length <= 0) {
             return true;
           }
+
+          return (
+            Array.prototype.every.call(imgs, (img) => {
+              return img.complete;
+            }) && document.readyState === "complete"
+          );
         });
       },
       { timeoutMsg: `image wasn't loaded` }
--- a/e2e-tests/specs/example.e2e.ts
+++ b/e2e-tests/specs/example.e2e.ts
@@ -4,6 +4,24 @@ describe("landing page", () => {
 
     await browser.$("h1").waitForExist();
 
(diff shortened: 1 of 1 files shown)


e2e-tests/specs/example.e2e.ts
```diff
@@ -4,6 +4,24 @@ describe("landing page", () => {
 
     await browser.$("h1").waitForExist();
 
+    browser.waitUntil(
+      function () {
+        return this.execute(function () {
+          let imgs = document.getElementsByTagName("img");
```
peterpeterparker, line 10: `const imgs`

if you want to go fancy you can even specify the type :

```
const imgs: HTMLCollectionOf<HTMLImageElement> ...
```

e2e-tests/specs/example.e2e.ts
```diff
@@ -4,6 +4,24 @@ describe("landing page", () => {
...
+          if (imgs.length > 0) {
```
peterpeterparker, line 11: I like to avoid if else, i find it more readable

```
if (imgs.length <= 0) {
    return true;
}

return (Array.prototype.every.call(imgs, (img) => ...
```

peterpeterparker (approved):
LGTM

e2e-tests/specs/example.e2e.ts
```diff
@@ -4,6 +4,26 @@ describe("landing page", () => {
 
     await browser.$("h1").waitForExist();
 
+    // Wait for all images to be "complete", i.e. loaded
+    browser.waitUntil(
+      function () {
+        return this.execute(function () {
+          const imgs: HTMLCollectionOf<HTMLImageElement> =
+            document.getElementsByTagName("img");
```
mstrasinskis, line 12: Theoretically querySelectorAll works faster because it returns a static list.

e2e-tests/specs/example.e2e.ts
```diff
@@ -4,6 +4,26 @@ describe("landing page", () => {
...
+          if (imgs.length <= 0) {
+            return true;
+          }
+
+          return (
+            Array.prototype.every.call(imgs, (img) => {
+              return img.complete;
+            }) && document.readyState === "complete"
```
mstrasinskis, line 20: If check document.readyState first then we could avoid a loop. But it's more FYI because of image quantity it's user wouldn't feel the difference :)

mstrasinskis (commented):
LGTM

e2e-tests/specs/example.e2e.ts
nmattia: ah, thanks! will change next time I update the code!

e2e-tests/specs/example.e2e.ts
nmattia: heh, that's true! to be completely honest I copy pasted most of this from here: https://github.com/webdriverio/webdriverio/issues/1880

PR #406, opened 2022-02-07
Refactor submit button

mstrasinskis:
# Motivation

Currently two button styles overlap each other (type=submit & .primary)
The idea was to refactor / rename the existing style [type="submit"] to .primary.
[L2-237](https://dfinity.atlassian.net/browse/L2-237)

# Changes

- Blue button colours were reversed to submit.
- Was refactored only `Get ICPs` button
image
## original:
image

- section padding was updated to avoid separate PR


peterpeterparker (approved):
Thanks!

PR #407, opened 2022-02-07
Use shasum instead of sha256sum

nmattia:
The `shasum` executable is installed on most systems, whereas
`sha256sum` is not.

Suggestion:
# Motivation
`sha256sum` is not installed by default on mac OsX.

Notes:
- scripts such as sha are usually provided by coreutils on *NIX systems.
- GNU coreutils [includes a family of SHA functions](https://en.wikipedia.org/wiki/List_of_GNU_Core_Utilities_commands) (sha1sum, sha256sum, ...) but NOT shasum
- Ubuntu 18.10 and 20.04 include shasum, but only because it is included in perl: 
```
  $ apt-file search bin/shasum
libdigest-sha-perl: /usr/bin/shasum       
perl: /usr/bin/shasum
```
- The OsX installation of shasum is also courtesy of its installation of perl.
- Mac used to install coreutils via BaseSystemBinaries, however that package is not installed on recent Macbooks.  It is not clear what provides the coreutils now.

Given that perl is, for better or worse, installed fairly universally it seems safer to rely on that than OsX adhering to coreutils.

# Changes
- Switch `sha256sum` to `shasum -a 256`

# Testing
- Has this been run on Mac?  Does ./scripts/docker-build work with this change? - Tested OK by @nmattia 
- Tested OK on Ubuntu 20.04


nmattia:
> Can you please clarify why this change is needed?

My Intel Mac doesn't have sha256sum; where does yours come from? Happy to use whichever is available (i.e. `if command -v ... ; then ...; else ...; fi`! WDYT?

nmattia:
My bad! I've mostly been using darwin systems recently and was under the impression that `shasum` was always available; on Darwin and on Linux. Should have said "shasum is more widespread than sha256sum".

Anyway, thought about the suggestion above?


> Happy to use whichever is available (i.e. if command -v ... ; then ...; else ...; fi! WDYT?

Shouldn't matter which exact tool we use anyway since we just print out the result!

bitdivine:
You could use either or.  I get the impression that the switch here is fine, as every Linux system I have checked (and I have been checking en-passant on other machines since, such as zh22) has both but Mac has just the one.  I did try to find an official list of what is is installed on Mac.  Apparently such utils used to be in BaseSystemBinaries and some were in XCode but BSB is no more and a scan of all packages showed that shasum is provided by none of them.  What guarantees can we rely on when Mac doesn't adhere to standards and doesn't seem to have an official list of what is installed?

Just please document in the commit used in main.  Figuring out why this PR is needed required effort and people reading the git log should not have to struggle to get the same info.

nmattia:
Done! Here's how the commit message reads:

> The `shasum` executable is more widely available than `sha256sum`. In particular, it appears that `sha256sum` is not installed on macOS by default, whereas `shasum` is.

nmattia:
Confirm it works on mac! Good call on the clarification.

PR #408, opened 2022-02-07
feat: infinite scroll component

peterpeterparker:
# Motivation

Proposals and other list are not fetched integrally when displayed but through subsets that are displayed when the user scrolls a specified distance from the bottom or top of the page.

This PR introduces an "infinite scroll" component which implements such behavior for any lists ("for each").

# Changes

- new `<InfiniteScroll />` component

# Notes

The component relies on the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

Such API is [not supported](https://caniuse.com/?search=intersection%20Observer) on Safari version older than v12.1 on desktop and v12.2 on mobile. There browsers are out of the scope of the browser we are currently supporting ("latest 2"). That's why the PR does not include any polyfill or some sort of custom code for old browsers.

# Tests

Unit test a basic since there is no implementation of the Intersection Observer in jest. It will need e2e tests for a proper behavioral tests.

### DIFF
--- a/frontend/svelte/jest-setup.ts
+++ b/frontend/svelte/jest-setup.ts
@@ -1,7 +1,32 @@
 import "@testing-library/jest-dom";
-
 // jsdom does not implement TextEncoder
 // Polyfill the encoders with node
-import { TextEncoder, TextDecoder } from "util";
+import { TextDecoder, TextEncoder } from "util";
+
 global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
+
+// @ts-ignore
+global.IntersectionObserver = class IntersectionObserver {
+  constructor(
+    private callback: (entries: IntersectionObserverEntry[]) => void,
+    private options?: IntersectionObserverInit
+  ) {}
+
+  observe(element: HTMLElement) {
+    this.callback([
+      {
+        isIntersecting: true,
+        target: element,
+      } as unknown as IntersectionObserverEntry,
+    ]);
+  }
+
+  disconnect() {
+    return null;
+  }
+
+  unobserve() {
+    return null;
+  }
+};
--- a/frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
+++ b/frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
@@ -1,0 +1,55 @@
+<script lang="ts">
+  import { afterUpdate, createEventDispatcher, onDestroy } from "svelte";
+
+  /**
+   * The Infinite Scroll component calls an action to be performed when the user scrolls a specified distance from the bottom or top of the page.
+   *
+   * Usage: To be wrapped around loops `<InfiniteScroll>{#each ...}</InfiniteScroll>`
+   *
+   * The component observe the last HTML element of the list using the IntersectionObserver.
+   * It set the reference after each re-render of the list. Pay attention to not trigger unnecessary updates.
+   */
+
+  export let options: IntersectionObserverInit = {
+    rootMargin: "300px",
+    threshold: 0,
+  };
+
+  let container!: HTMLDivElement;
+
+  const dispatch = createEventDispatcher();
+
+  const onIntersection = (entries: IntersectionObserverEntry[]) => {
+    const intersecting: IntersectionObserverEntry | undefined = entries.find(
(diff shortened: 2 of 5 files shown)


frontend/svelte/jest-setup.ts
```diff
@@ -1,7 +1,32 @@
...
-
 // jsdom does not implement TextEncoder
 // Polyfill the encoders with node
-import { TextEncoder, TextDecoder } from "util";
+import { TextDecoder, TextEncoder } from "util";
+
 global.TextEncoder = TextEncoder;
 global.TextDecoder = TextDecoder;
+
+// @ts-ignore
```
nmattia, line 9:
```suggestion
// Mocks the Intersection Observer API for the tests. This just matches the interface
// but doesn't do anything.
// https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
// @ts-ignore
```

note: do we actually test the behavior?

frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
```diff
@@ -0,0 +1,55 @@
...
+   * The component observe the last HTML element of the list using the IntersectionObserver.
+   * It set the reference after each re-render of the list. Pay attention to not trigger unnecessary updates.
+   */
+
+  export let options: IntersectionObserverInit = {
+    rootMargin: "300px",
+    threshold: 0,
+  };
+
+  let container: HTMLDivElement;
```
nmattia, line 18: I think this is called the "root" in the intersection observer API doc

frontend/svelte/jest-setup.ts
peterpeterparker: yes. the component won't compile without it and it tests the the `nnsIntersect` event is dispatched by the component if the intersection observer calls the callback with an intersecting elements.

being said, as I described in the PR details, it will need e2e tests.

frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
peterpeterparker: in the current implementation I do not use the `root`, I can add it though. Will try.

frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
```diff
@@ -0,0 +1,55 @@
+<script lang="ts">
+  import { afterUpdate, createEventDispatcher, onDestroy } from "svelte";
+
+  /**
+   * The Infinite Scroll component calls an action to be performed when the user scrolls a specified distance from the bottom or top of the page.
```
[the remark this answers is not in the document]
peterpeterparker: exactly, that's what I understand an "infinite scroll" behavior is ("when bottom is reached, load next").

I understand what you describe with your comment as a "virtual scroll" behavior - i.e. another behavior.

nmattia (approved):
Thanks for the clarification!

frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
peterpeterparker: just had a try, it actually ends in infinite loop if we attach the `root`. therefore better stick to current observation on the viewport.

frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
```diff
@@ -0,0 +1,55 @@
...
+  const onIntersection = (entries: IntersectionObserverEntry[]) => {
+    const intersecting: IntersectionObserverEntry | undefined = entries.find(
+      ({ isIntersecting }: IntersectionObserverEntry) => isIntersecting
+    );
+
+    if (intersecting === undefined) {
+      return;
+    }
+
+    dispatch("nnsIntersect");
```
mstrasinskis, line 31: Probably `observer.disconnect();` could be called here to avoid multiple fetching in case of eg slow network connection.

peterpeterparker: Thanks.

👉 https://github.com/dfinity/nns-dapp/pull/410

PR #409, opened 2022-02-07
Sort css

bitdivine:
# Motivation
The order of the compiled CSS clauses is not stable.  This results in non-reproducible builds.

# Changes
- Compile the css with one clause per line.  (Curiosity: public/build/bundle.css has only 20 clauses, I expected much more.)
- Order the lines.

# Testing
TODO: Compare builds across multiple platforms and repeated runs.

### @ commit f38a3c4a70e4c787e45e9adf28148d16c6a684fa

sinkpad build 1 (Max's Dfinity Linux laptop, Ubuntu 20.04)
```
3e2e4d36df0f5d00b2f1a2284df19209141cc73a770a52d4a221f6ee4e697b7e  assets.tar.xz
330134a9f73176bc710d33c90a938d55cbc464aab2fc35aad2c5dfe3364def54  nns-dapp.wasm
```
### @ commit 582ed5940bbd856f8bf8a8340a3589775dc78cd0
sinkpad build 1
```
c64b80d50cf1bba535bb85a7e38d943bf9f083c62676ceebb64c700130564bf5  assets.tar.xz
b9ec39ea717c10ce9a29448fbf7e8509c6ab9a15b06eccf8b05b860634e920b7  nns-dapp.wasm
```

sinkpad build 2
```
c64b80d50cf1bba535bb85a7e38d943bf9f083c62676ceebb64c700130564bf5  assets.tar.xz
b9ec39ea717c10ce9a29448fbf7e8509c6ab9a15b06eccf8b05b860634e920b7  nns-dapp.wasm
```

maxbook pro build 1
- Docker build failed at step 12/19 - compiling ic-cdk-optimizer

CI build https://github.com/dfinity/nns-dapp/actions/runs/1805896170
```
max@sinkpad:~/Downloads/ci-build-1805896170 (41)$ unzip NNS\ mainnet\ backend\ wasm\ module\ \(1\).zip 
Archive:  NNS mainnet backend wasm module (1).zip
  inflating: nns-dapp.wasm           
max@sinkpad:~/Downloads/ci-build-1805896170 (44)$ sha256sum nns-dapp.wasm 
b9ec39ea717c10ce9a29448fbf7e8509c6ab9a15b06eccf8b05b860634e920b7  nns-dapp.wasm
```

## @ commit ea323520037eb51631a4a09a6175fd8bb3518fdf
sinkpad build 1
```
476b5756af3a924e53feba534bb2b98f1b5cf0bb071b095d87712a445885c4c3  assets.tar.xz
d6c560d998fc096d0242f9323e92a134816d1ba46a6b6f613f29fe23930a21b0  nns-dapp.wasm
```
CI
```
d6c560d998fc096d0242f9323e92a134816d1ba46a6b6f613f29fe23930a21b0  nns-dapp.wasm
```


frontend/svelte/package.json
```diff
@@ -32,6 +32,7 @@
     "prettier": "^2.5.0",
     "prettier-plugin-svelte": "^2.5.0",
     "rollup": "^2.3.4",
+    "rollup-plugin-css-bundle": "^1.0.4",
```
lmuntanerDfinity, line 35: A general question, don't we prefer to remove all the carets `^`? Specially to make sure every build is exactly the same.

lmuntanerDfinity (approved):

frontend/svelte/scripts/sort-css
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+FILE=public/build/bundle.css
```
nmattia, line 2: can you make this absolute?

frontend/svelte/rollup.config.js
```diff
@@ -60,7 +61,7 @@ export default {
     }),
     // we'll extract any component CSS out into
     // a separate file - better for performance
-    css({ output: "bundle.css" }),
+    cssbundle(),
```
nmattia, line 64: what's the difference here? also the repo looks a bit outdated and not official

frontend/svelte/package.json
bitdivine: The lockfile is there to make sure that the build is exactly the same.

frontend/svelte/scripts/sort-css
bitdivine: No, I'd really rather not.

frontend/svelte/rollup.config.js
bitdivine: cssbundle emits one line per clause.

frontend/svelte/package.json
lmuntanerDfinity: ah, true, with the lockfile removing it is not necessary anymore

frontend/svelte/rollup.config.js
```diff
@@ -60,7 +61,7 @@ export default {
     }),
     // we'll extract any component CSS out into
```
peterpeterparker, line 62: 1. what about global styles? extracted in a single files?

2. is the output still minified?

3. does it still support autoprefixer?

4. about performance on the ic - are all the files loaded at once or lazy loaded?

frontend/svelte/rollup.config.js
peterpeterparker: 3+ years old plugin not updated sounds risky in terms of maintenance

nmattia: can you add a note?

frontend/svelte/scripts/sort-css
nmattia: why not? AFAICT running the script from any other dir than the svelte root will just fail, correct?

peterpeterparker (changes requested):
Since PR has been marked as "accepted", I allow myself to mark it as "request changes" because I would like clarification on my questions before any potential merge.

frontend/svelte/rollup.config.js
bitdivine: @peterpeterparker  I agree that a clearly well maintained solution would be preferable.  There are working solutions and optimal solutions.  Let's get a working solution first, unless you know something better already?  I had a brief look but did not find an alternative.

bitdivine: @nmattia - yes, I'll add a comment

frontend/svelte/rollup.config.js
bitdivine: @peterpeterparker  I think the easiest answer is: Look at the build.css.  You can see this in assets.tar.xz in here: https://github.com/dfinity/nns-dapp/actions/runs/1805896170

I am concerned about CSS ordering.  I think it should not affect svelte's internal styles because those are nicely and robustly scoped.  Global css could be a problem in theory.  I did note in the README for cssbundle that it says that the ordering matches the order in which css is added - so it sounds a bit as if the ordering should be deterministic.  If so, we don't need the sort.  I have not tested this, I emulated the solution used by people-parties and tweaked it a bit.

frontend/svelte/rollup.config.js
nmattia: How about [this](https://github.com/egoist/rollup-plugin-postcss) repo, that seems to at least care about the issue (at first glance) https://github.com/egoist/rollup-plugin-postcss/pull/295/files? Also, can you look for/create a ticket upstream? It really feels like we shouldn't be the first people to run into this

frontend/svelte/rollup.config.js
bitdivine: Note: I was curious about the @charset utf8.  I checked and the stylesheet has only ascii chars.

bitdivine: We did previously have a `frontend/svelte/public/global.css` but that was deleted.  If we want such CSS again, we can re-add it, and it won't be affected by this build.  I doubt that we will need it, assuming we write pure CSS.  What do you think?

frontend/svelte/scripts/sort-css
bitdivine: Ah, if that is the reason it's a good point.  The script is missing the usual "cd to a fixed location" at the beginning.  CWD is a global variable so should be immutable in most scripts.

frontend/svelte/rollup.config.js
peterpeterparker: Using a not maintained plugin is a red flag to me in terms of maintainability. We might solve something today but end-up not being able to build the app tomorrow.

I would suggest to find another maintained plugin, write a custom plugin for our particular need or a script.

frontend/svelte/scripts/sort-css
bitdivine: It was also missing the usual `set -euo pipefail`

frontend/svelte/rollup.config.js
peterpeterparker: I'll check questions 1 to 3 once we have solve the question of plugins.

peterpeterparker (approved):
As discussed in today's chat, we go on with it as a temporary solution but will find a solution that offers more guarantee in terms of maintainability for the future.

nmattia (changes requested):
Can you create a ticket, assign to me and link from here? thanks

frontend/svelte/rollup.config.js
peterpeterparker: 1. global style is generated and and added to `bundle.css`
2. css in minified, can be improved since the output a file with one line of css per components though
3. autoprefixer seems to still work
4. performance will be identical since it loads on single css file

bitdivine:
@nmattia - please don't block this PR for the Jira issue.

nmattia:
Since this is a bit controversial, and since we're going for a temporary hack I really don't want this to fall through the cracks... sorry for being a tad uptight :/

nmattia (approved):
Unblocking

PR #410, opened 2022-02-07
feat: disconnect observer while fetching

peterpeterparker:
# Motivation

Add a level of security to prevent multiple fetching in case the Infinite Scroll is used on a slow network.

# Changes

- disconnect observer before dispatching event

### DIFF
--- a/frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
+++ b/frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
@@ -42,7 +42,7 @@
   afterUpdate(() => {
     // The DOM has been updated. We reset the observer to the current last HTML element of the infinite list.
 
-    // We disconnect previous observer first. We want to avoid to obsevere multiple elements.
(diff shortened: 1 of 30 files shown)


frontend/svelte/src/lib/components/ui/InfiniteScroll.svelte
```diff
@@ -38,6 +41,8 @@
 
   afterUpdate(() => {
     // The DOM has been updated. We reset the observer to the current last HTML element of the infinite list.
+
+    // We disconnect previous observer first. We want to avoid to obsevere multiple elements.
```
nmattia, line 45: Something fishy in that last sentence

peterpeterparker: Done

mstrasinskis:
Just one thing -  `const` would be preferable here `let observer: IntersectionObserver`

peterpeterparker:
@mstrasinskis const typo fixed.

mstrasinskis (approved):
Thank you!

PR #414, opened 2022-02-08
feat: badge component

peterpeterparker:
# Motivation

A new inline block element `<Badge />` to display notification or status.

# Changes

- new `<Badge />` element

# Screenshot

Capture d’écran 2022-02-08 à 09 13 31

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Badge.svelte
+++ b/frontend/svelte/src/lib/components/ui/Badge.svelte
@@ -1,0 +1,36 @@
+<script lang="ts">
+    export let color: 'warning' | 'success' | undefined = undefined;
+</script>
+
+<div class={color}>
+    <slot />
+</div>
+
+<style lang="scss">
+  div {
+    display: inline-block;
+
+    width: fit-content;
+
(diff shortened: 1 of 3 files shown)


frontend/svelte/src/lib/components/ui/Badge.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+  export let color: "warning" | "success" | undefined = undefined;
```
nmattia, line 2: Shouldn't this have more colors, e.g. "pending", "executed", "accepted", "rejected"? or maybe colors that can be defined by the caller?

nmattia (approved):
LGTM, just one comment about how the badges will be used

frontend/svelte/src/lib/components/ui/Badge.svelte
peterpeterparker: actually all these other proposals status' color are `currentColor` aka grey.

but yes in the future can for sure be extended, like "error". just thought I make the component colors' name generic and not linked with the proposals status.

PR #415, opened 2022-02-08
L2-215: Wizard Component

lmuntanerDfinity:
# Motivation

Create a new component that enables forms and functionality to be displayed by steps.

NOTE: It doesn't work with two Wizard rendered in the same page.

# Changes

* New components: Wizard, WizardPanel
* New components share a store: wizardStore.
* New folder to group the different files working together.

# Tests

* New tests at `Wizard.spec.ts`.


lmuntanerDfinity:
@nmattia @peterpeterparker @mstrasinskis please review

I tried to enable multiple instances of Wizard, but I needed a context. And since currently we don't need the functionality, I decided to leave it simple

frontend/svelte/src/lib/components/ui/Wizard/Wizard.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+  import { selectedTabStore } from "./wizardStore";
+
+  export let selectedTabIndex = 0;
+
+  $: {
+    selectedTabStore.update(() => selectedTabIndex);
```
peterpeterparker, line 7: - this can be a `set`

- do we need the property `selectedTabIndex`? consumer of the wizard can use the store to select the page

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
```diff
@@ -0,0 +1,20 @@
...
+  const X_SLIDE_OFFSET = 200;
+  let slideOffset = X_SLIDE_OFFSET;
+
+  selectedTabStore.subscribe((index) => {
+    slideOffset = index > currentIndex ? X_SLIDE_OFFSET : -X_SLIDE_OFFSET;
+    currentIndex = index;
+  });
+</script>
+
+{#if currentIndex === index}
```
peterpeterparker, line 16: the local state `currentIndex` can't be dropped aka use the store directly in the comparison (`$sele...`) ?

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
```diff
@@ -0,0 +1,20 @@
+<script lang="ts">
+  import { fly } from "svelte/transition";
+  import { selectedTabStore } from "./wizardStore";
+
+  export let index: number;
+  let currentIndex = null;
+  const X_SLIDE_OFFSET = 200;
+  let slideOffset = X_SLIDE_OFFSET;
+
+  selectedTabStore.subscribe((index) => {
```
peterpeterparker, line 10: `unsubscribe` on `onDestroy` to avoid memory leak

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
```diff
@@ -0,0 +1,20 @@
+<script lang="ts">
+  import { fly } from "svelte/transition";
+  import { selectedTabStore } from "./wizardStore";
+
+  export let index: number;
+  let currentIndex = null;
```
peterpeterparker, line 6: typescript definition `number : null` ?

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
```diff
@@ -0,0 +1,20 @@
...
+  const X_SLIDE_OFFSET = 200;
+  let slideOffset = X_SLIDE_OFFSET;
```
peterpeterparker, line 8: you might not need the initializion here since the subscribe kicks in when the component is loaded / mounted

frontend/svelte/src/lib/components/ui/Wizard/wizardStore.ts
```diff
@@ -0,0 +1,3 @@
+import { writable } from "svelte/store";
+
+export const selectedTabStore = writable(0);
```
peterpeterparker, line 3: `writable<number>(0)`?

peterpeterparker (changes requested):
Thanks, just a couple of small feedback from my side and one more question:

Do we need a clean up - i.e. shouldn't we reset the store index to 0 when the wizard is destroyed?

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
```diff
@@ -0,0 +1,20 @@
+<script lang="ts">
+  import { fly } from "svelte/transition";
+  import { selectedTabStore } from "./wizardStore";
+
+  export let index: number;
+  let currentIndex = null;
+  const X_SLIDE_OFFSET = 200;
```
mstrasinskis, line 7: On desktop it's a bit wider ;)

frontend/svelte/src/lib/components/ui/Wizard/Wizard.svelte
lmuntanerDfinity: My idea was that the parent of the Wizard sets the index.

```
<Wizard selectedTabIndex={ 0 }>
```

This way the logic of when to change leave outside the Wizard component.

Ok to use `set`.

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
lmuntanerDfinity: why didn't TS complain?

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
lmuntanerDfinity: I wanted a simpler solution without accessing the DOM.

We don't know how wide the parent will be. What do you propose?

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
lmuntanerDfinity: oh yest!

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
lmuntanerDfinity: The store value can only be used with the `subscribe`, do you mean to create a variable `isCurrentIndex`?

```
selectedTabStore.subscribe((index) => {
    slideOffset = index > currentIndex ? X_SLIDE_OFFSET : -X_SLIDE_OFFSET;
    currentIndex = index;
  });

$: isCurrentIndex = currentIndex === index;
```

I would like to keep the `currentIndex` to be able to set the offset negative or positive.

frontend/svelte/src/lib/components/ui/Wizard/wizardStore.ts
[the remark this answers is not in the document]
lmuntanerDfinity: Oh, my bad.

I like `Step`, I will rename all to that.

frontend/svelte/src/tests/lib/components/ui/Wizard.spec.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { render } from "@testing-library/svelte";
+import WizardTest from "./WizardTest.svelte";
+
+describe("Wizard", () => {
```
mstrasinskis, line 8: 👍

lmuntanerDfinity:
@peterpeterparker @mstrasinskis ready for another round of review

Still not sure how to best "calculate" the offset for the animation. Any ideas?

frontend/svelte/src/lib/components/ui/Wizard/Wizard.svelte
peterpeterparker: Since the store cannot be scoped (?) it does not prevent any one to still set the index without using the props. In addition, it means that the consumer has to define a local state for the index too therefore potentially we can end up with a local state and a global state that have two different values. If you find a way to prevent using the global store in other component I am fine with the props but otherwise I think it's better without.

Another thing: instead of having a setter that set an index, why not having two functions `next()` and `prev()` ?

frontend/svelte/src/lib/components/ui/Wizard/WizardPanel.svelte
peterpeterparker: gotcha, thanks for the feedback

frontend/svelte/src/lib/components/ui/Wizard/WizardStep.svelte
```diff
@@ -0,0 +1,27 @@
...
+  import { fly } from "svelte/transition";
+  import { selectedStepStore } from "./wizardStore";
+
+  export let index: number;
+  let currentIndex: number | null = null;
+  let isCurrentIndex: boolean = false;
+  const X_SLIDE_OFFSET = 200;
+  let slideOffset: number | undefined;
+
+  const unsubscribe = selectedStepStore.subscribe((index) => {
```
peterpeterparker, line 13: `unsubscribe` can be typed

```
import type { Unsubscriber } from "svelte/types/runtime/store";
const unsubscribe: Unsubscriber...
```

frontend/svelte/src/lib/components/ui/Wizard/wizardStore.ts
peterpeterparker: You are right, thanks

peterpeterparker:
@lmuntanerDfinity for the animation have you try to use - get the container (`<div in:fly....`) width?

lmuntanerDfinity:
@peterpeterparker about using `next` or a prop.

I chose the more declarative way. I agree that we are duplicating the state. I still feel that the parent would want to know whether it's in the first, second or third screen. Maybe there are different events happening there.

In case we go for `next` and the parent wanted to know about the screen, we could then dispatch an event. Or the parent could decide to duplicate the state as well.

I don't have a perfect solution, I have the feeling that we should have started by using the Wizard in some specific use cases. After we have implemented two use cases, we can review the API we offer and decide whether we want to change or not.

Not a strong opinion on going for one or the other at the moment. But I propose to merge this and move to the actual use cases and then review the API after we have implemented two Wizards. What do you think @peterpeterparker ?

lmuntanerDfinity:
> @lmuntanerDfinity for the animation have you try to use - get the container (`<div in:fly....`) width?

I wanted to keep it simple. But it was actually simple to do in Svelte:

https://github.com/dfinity/nns-dapp/pull/415/files#diff-77a5058faa0a8b429ea739b4ce137c873cfc924554d10c7187e29030ab161a57R23

peterpeterparker:
> ... for the animation have you try to use - get the container (<div in:fly....) width?
>
> I wanted to keep it simple. But it was actually simple to do in Svelte:
> https://github.com/dfinity/nns-dapp/pull/415/files#diff-77a5058faa0a8b429ea739b4ce137c873cfc924554d10c7187e29030ab161a57R23

This is what I meant 👍

frontend/svelte/src/lib/components/ui/Wizard/WizardStep.svelte
peterpeterparker: yes you are right, that's more my opinion on providing explicit return types since all other unsubscribed of the apps are explicitly typed. we do not have (i think) any eslint rules about it (yet).

lmuntanerDfinity: Nops, we don't have eslint, not a list of this conventions.

lmuntanerDfinity:
@peterpeterparker would you rather expose the store as an API, or use the component Wizard for it?

Another option is the following:

```
// to be used on:click somewhere
const next = () => {
  wizard.next();
}

// ...

<Wizard bind:this={wizard}>
</Wizard>
```

This way, if we decide to move from store to context, we can keep the same API. If we expose the store and we decide refactor, we will have to change many other places.

I'd rather expose the least of the implementation details. Does that work for you?

peterpeterparker:
@lmuntanerDfinity sure that works for me too.

what bugs me is "local state vs global state not in sync" and "next prev are more suitable". as long as both are solved, I am happy.

frontend/svelte/src/lib/components/ui/Wizard/WizardStep.svelte
peterpeterparker: as you wish, I can live with it.

can you open a task in jira backlog "clean-up un-useful explicit return types or set an eslint rules about it"?

so that we can decide what we do and have a consistent coding style.

lmuntanerDfinity: I don't mind one or the other that much. I think it's more important the consistency.

We can discuss about implementing eslint during our call today. I am not sure if it goes well with prettify, though.

peterpeterparker: Sounds like a plan.

lmuntanerDfinity:
@peterpeterparker updated API with `next` and `back` exported methods.

In the future, we could export a `getCurrentIndex` method if we want.

I also updated the tests.

Ready for another review 😄

frontend/svelte/src/lib/components/ui/Wizard/WizardStep.svelte
```diff
@@ -0,0 +1,35 @@
...
+  let slideOffset: number | undefined;
+
+  const unsubscribe: Unsubscriber = selectedStepStore.subscribe((index) => {
+    slideOffset = index > currentIndex ? absolutOffset : -absolutOffset;
+    currentIndex = index;
+  });
+
+  onDestroy(unsubscribe);
+
+  $: absolutOffset = container ? container.clientWidth : DEFAULT_OFFSET;
```
peterpeterparker, line 24: fine as it is but just fyi it can be inlinied as following:

```
$: absolutOffset = container?.clientWidth || DEFAULT_OFFSET;
```

lmuntanerDfinity: I still don't have this feature on the top of my head. Thanks!

lmuntanerDfinity: Done

frontend/svelte/src/tests/lib/components/ui/WizardTest.svelte
```diff
@@ -0,0 +1,26 @@
+<script lang="ts">
+  import Wizard from "../../../../lib/components/ui/Wizard/Wizard.svelte";
+  import WizardStep from "../../../../lib/components/ui/Wizard/WizardStep.svelte";
+
+  let wizard: Wizard | undefined;
+
+  export function next() {
+    wizard.next();
```
peterpeterparker, line 8: I know just a test but since we don't have effective implementation yet and as it might serve as example, might be worth to check for undefined

```
wizard?.next();
wizard?.prev();
```

also, fyi too, not a blocker, you can also write it in a single line

```
export const next = () => wizard?.next();
```

peterpeterparker (approved):
LGTM. Thanks for the improvements!

frontend/svelte/src/tests/lib/components/ui/WizardTest.svelte
lmuntanerDfinity: 👍

nmattia (commented):
I'm curious about the architecture here. Why have all the `WizardStep`s as different component? Because of this each step needs to check to see if it should be shown or not.

Why not have a single `WizardStep` at a time, always shown in `Wizard`? i.e.

``` diff
<section>
-   <slot />
+  <WizardStep ...>
 </section>
```

which means, AFAICT, that you could get rid of the `selectedStepStore` and just have a local variable in `Wizard.svelte`. But maybe I'm missing something obvious, don't have much experience with Svelte!

lmuntanerDfinity:
Small question, what is "AFAICT"?

The Wizard does not have access to its children. Therefore, we can't filter and select the child would like to render.

Instead, we render all the children with "slot" and then it's each `WizardStep` that reads the store and decides to render the content or not.

Does that answer your question?

nmattia:
> Small question, what is "AFAICT"?

As Far As I Can Tell

> The Wizard does not have access to its children. Therefore, we can't filter and select the child would like to render.
> Instead, we render all the children with "slot" and then it's each WizardStep that reads the store and decides to render the content or not.

I understand! And I'm suggesting we do it differently, i.e. although show exactly one "step" in `Wizard`:

``` js
<Wizard>
  <WizardStep ... more config here>
</Wizard>
```

Where the "more config" bit depends on the currently selected step. This way, you can get rid of the store and potentially have multiple wizards

lmuntanerDfinity:
We could have all the logic in the parent or Wizard. But then the `Wizard` component ends up being a UI component that slides children.

We would need the state everytime we needed a Wizard.

```
let currentIndex = 0;

const next = () => {
  currentIndex += 1;
}

const back = () => {
  currentIndex -= 1;
}
//...

<Wizard>
  <WizardStep show={currentIndex === 0}>
    <div>First</div>
  </WizardStep>
  <WizardStep show={currentIndex === 1}>
    <div>Second</div>
  </WizardStep>
</Wizard>
```

If we do this, we could also skip the `WizardStep` altogether and use a class.

```
<Wizard>
  <div class:show={currentIndex === 0}>First</div>
  <div class:show={currentIndex === 1}>Second</div>
</Wizard>
```

We might also need some extra config to know whether the transition is from the left or from the right.

We would be adding logic every time we want to use the `Wizard`. Do you think this is a better pattern?

mstrasinskis (approved):

nmattia:
@lmuntanerDfinity your call, you're much more familiar with the issue than me! This LGTM as it is

PR #417, opened 2022-02-09
L2-251: Development agent for nns

lmuntanerDfinity:
# Motivation

NNS-JS should make requests to testnet during development.

# Changes

* Create an agent that expects a host as optional parameter and use it with nns-js.
* Rename `IDENTITY_SERVICE_URL` to `SERVICE_URL`. It is more than just for identity.

# Tests

No new functionality added; therefore, no new tests added. Yet, a current test had to be edited to mock importing a module.

### DIFF
--- a/frontend/svelte/rollup.config.js
+++ b/frontend/svelte/rollup.config.js
@@ -84,8 +84,8 @@ export default {
     replace({
       preventAssignment: true,
       "process.env.ROLLUP_WATCH": !!process.env.ROLLUP_WATCH,
-      "process.env.IDENTITY_SERVICE_URL": JSON.stringify(
-        process.env.IDENTITY_SERVICE_URL ||
+      "process.env.SERVICE_URL": JSON.stringify(
+        process.env.SERVICE_URL ||
           (process.env.DEPLOY_ENV === "testnet"
             ? "https://qjdve-lqaaa-aaaaa-aaaeq-cai.nnsdapp.dfinity.network/"
             : "https://identity.ic0.app/")
--- a/frontend/svelte/rollup.config.js
+++ b/frontend/svelte/rollup.config.js
@@ -84,8 +84,8 @@ export default {
     replace({
       preventAssignment: true,
       "process.env.ROLLUP_WATCH": !!process.env.ROLLUP_WATCH,
-      "process.env.SERVICE_URL": JSON.stringify(
-        process.env.SERVICE_URL ||
+      "process.env.IDENTITY_SERVICE_URL": JSON.stringify(
+        process.env.IDENTITY_SERVICE_URL ||
           (process.env.DEPLOY_ENV === "testnet"
             ? "https://qjdve-lqaaa-aaaaa-aaaeq-cai.nnsdapp.dfinity.network/"
             : "https://identity.ic0.app/")
--- a/frontend/svelte/src/lib/constants/utils.constants.ts
+++ b/frontend/svelte/src/lib/constants/utils.constants.ts
@@ -1,0 +1,1 @@
+export const serviceURL: string = process.env.IDENTITY_SERVICE_URL;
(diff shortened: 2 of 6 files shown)


lmuntanerDfinity:
The host seems to be the same as used for the `auth-client`.

Adding a screenshot showing the 200 received from testnet.

Screenshot 2022-02-09 at 13 46 20

frontend/svelte/rollup.config.js
```diff
@@ -84,8 +84,8 @@ export default {
     replace({
       preventAssignment: true,
       "process.env.ROLLUP_WATCH": !!process.env.ROLLUP_WATCH,
-      "process.env.IDENTITY_SERVICE_URL": JSON.stringify(
-        process.env.IDENTITY_SERVICE_URL ||
+      "process.env.SERVICE_URL": JSON.stringify(
```
peterpeterparker, line 87: Are you sure `IDENTITY_SERVICE_URL` is not use in the CI?

(don't know but just in case...)

frontend/svelte/src/lib/utils/agent.utils.ts
```diff
@@ -0,0 +1,6 @@
+import { HttpAgent } from "@dfinity/agent";
+
+const host = process.env.SERVICE_URL;
```
peterpeterparker, line 3: Isn't  this the exact same value as the global constant `identityProvider`? If yes, we can reuse the constant.

peterpeterparker: I mean if these are the same value we can do following:

```
export const createAgent = () => new HttpAgent({ host: identityProvider });
```

lmuntanerDfinity: I would need to import `identityProvider`, right?

peterpeterparker: Right. Again only if same value. It can be convenient in the editor to search every occurences of the same constant.

lmuntanerDfinity: Done

frontend/svelte/src/lib/constants/utils.constants.ts
```diff
@@ -0,0 +1 @@
+export const serviceURL: string = process.env.IDENTITY_SERVICE_URL;
```
nmattia, line 1: this wasn't renamed in the end, was it? it's still `IDENTITY_SERVICE_URL`

frontend/svelte/src/lib/utils/agent.utils.ts
```diff
@@ -0,0 +1,5 @@
+import { HttpAgent } from "@dfinity/agent";
+import { serviceURL } from "../constants/utils.constants";
+
+// To avoid being executed in tests that only import it
+export const createAgent = () => new HttpAgent({ host: serviceURL });
```
nmattia, line 5: just a note but I wonder if it's worth having an extra module for effectively one line:

```js
new HttpAgent({ host: serviceURL });
```

lmuntanerDfinity: I will also `fetchRootKey` inside here. Maybe also memoize the agent instead of creating a new one.

I don't think it's going to stay like this.

frontend/svelte/src/lib/constants/utils.constants.ts
lmuntanerDfinity: No, it was not renamed. Do you think we should?

lmuntanerDfinity: I am not even sure the API is returning what it's supposed to. We get a 200, but I can't be sure the data is good because I have not yet been able to test it properly

nmattia: Don't know if you should, but I think you said you would :) 

> Rename IDENTITY_SERVICE_URL to SERVICE_URL. It is more than just for identity.

> I am not even sure the API is returning what it's supposed to. 

Ok, we could clarify offline if needed!

frontend/svelte/src/lib/utils/agent.utils.ts
nmattia: ok, very careful with `fetchRootKey` then! gotta make sure it doesn't land in production call, so hiding it inside a utils function might be a bit dangerous

PR #419, opened 2022-02-09
Add new NNS function types to the whitelist

hpeebles:
# Motivation

A few new NNS functions have been added and in order to show their payloads we need to add their Ids to the whitelist.
This whitelist exists because some proposals have payloads which are too large for the NNS Dapp to render (eg. proposals which contain canister wasms).

# Changes

The new Ids have been added to the whitelist

# Tests

I have added an NNS function 21 proposal to the list of dummy proposals and can now see its payload, whereas before this commit the payload displayed as 'null'.


hpeebles:
Screenshot 2022-02-09 at 13 53 47

frontend/ts/src/ServiceApi.ts
```diff
@@ -522,100 +523,121 @@ export default class ServiceApi {
...
+            summary:
+              "Initialize datacenter records. For more info about this proposal, read the forum announcement: https://forum.dfinity.org/t/improvements-to-node-provider-remuneration/10553",
+            nnsFunction: 21,
+            payload: addOrRemoveDataCentersPayload,
+          });
+        console.log(manageNeuronResponse);
+      }
+    } catch (e) {
+      console.log(e);
+      throw e;
```
bitdivine, line 640: Cool.  How is this invoked?  Do you run this on the command line, with nodejs, or via a browser?

bitdivine (approved):
LGTM.  Love the expanded testing facilities.  Thanks!

PR #420, opened 2022-02-09
feat: list proposals according to filters

peterpeterparker:
# Motivation

List proposals and filter the queries according filters.

# Changes

- new component `ProposalCard` to render a proposal as card
- `VotingsFilters` renamed in `ProposalsFilters`
- enums `Topic`, `ProposalRewardStatus` and `ProposalStatus` from nns-js replace custom enums (not without pain)
- refactor `key` to `labelKey` in `FilterCard` and style width if no filters are rendered
- new constant for the queries pagination ("10 elements at a time")
- new constant for the default filters that should be displayed per default
- translations modified according new enums keys
- `VotingFiltersModal` renamed in `ProposalsFiltersModal`
- new global store for the list of proposals
- new global store for the filters of the proposals
- new enum utils to convert the enums defined in nns-js to keys, values and to subset of keys
- new proposal utils to list proposals, fetch next proposals, tells if all proposals are fetched and what the id of last proposal in store is
- new utils to `debounce` action
- `proposals` store enhanced in order to search for the proposals while amending the global store (if empty search only on mount)
- `proposals` store enhanced to link filters with the search of the proposals
- `proposals` store enhanced to support infinite scrolling
- `proposals` store enhanced to display a spinner if work is happening
- list of `proposals` set to certified query

### Tests

Tested manually and unit tests provided. Being said here it will also needs e2e, particularly to play with the filters and infinite scroll

### Note

Some remaining `// TODO` including two filters (only `includeRewardStatus` and `includeStatus` have been implemented.

A particular TODO regarding enums. It is a bit a pain to make enum and generic cohabit, happy to get help about it.

### Side Note

The PR is quite big, sorry for that. It's notably a side effect of the fact that I had to develop first the UI weeks ago without knowing what will do with nns-js. Therefore here I also had to replace all custom enums with enums of nns-js, that leads to more changes than just using a function.

### Screenshots

ezgif com-gif-maker(4)


frontend/svelte/src/routes/Proposals.svelte
```diff
@@ -1,15 +1,76 @@
...
+
+    // If proposals are already displayed we reset the store first otherwise it might give the user the feeling than the new filters were already applied while the proposals are still being searched.
+    await listProposals({ clearBeforeQuery: !emptyProposals() });
+
+    loading = false;
+  };
+
+  let debounceFindProposals: () => void | undefined;
+
+  // We do not want to fetch the proposals twice when the component is mounting because the filter subscriber will emit a first value
```
lmuntanerDfinity, line 47: Can a proposal be edited?

If so, it could happen that when the user comes back, a proposal that was fetched and stored has been updated. We should then fetch it again.

I'd rather not rely on the order of fetches, and make everything work despite the order. Although I am not sure that I understand here why the subsequent fetches are a problem.

What do you mean by "the filter subscriber will emit a first value"?

frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  proposalsStore.pushProposals(proposals);
+};
+
+const queryProposals = async ({
+  beforeProposal,
+}: {
+  beforeProposal: ProposalId | undefined;
+}): Promise<ProposalInfo[]> => {
+  // TODO(L2-206): use createAgent
+  const governance: GovernanceCanister = GovernanceCanister.create();
```
lmuntanerDfinity, line 59: Shuold we create one every time we `queryProposals` or could we reuse one created beforehand?

frontend/svelte/src/lib/components/proposals/ProposalCard.svelte
```diff
@@ -0,0 +1,33 @@
...
+  export let proposalInfo: ProposalInfo;
+
+  let proposal: Proposal | undefined;
+  let status: ProposalStatus | undefined;
+
+  let color: "warning" | "success" | undefined;
+
+  $: ({ proposal, status } = proposalInfo);
+
+  $: color =
```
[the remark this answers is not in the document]
peterpeterparker: If you wish I can do it for sure. There is only two of six that receive a particular colors. That's why I thought it is compact enough with ifs. But again, can do it.

frontend/svelte/src/lib/components/proposals/ProposalCard.svelte
```diff
@@ -0,0 +1,33 @@
...
+  import { ProposalStatus } from "@dfinity/nns";
+  import Badge from "../ui/Badge.svelte";
+  import { i18n } from "../../stores/i18n";
+
+  export let proposalInfo: ProposalInfo;
+
+  let proposal: Proposal | undefined;
+  let status: ProposalStatus | undefined;
+
+  let color: "warning" | "success" | undefined;
```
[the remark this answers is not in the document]
peterpeterparker: "color" is good to me, that's inspired from the Ionic framework. I can imagine that in the future we might have other options that are not types like "grey" or "light".

frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  return last?.id;
+};
+
+export const listProposals = async ({
+  clearBeforeQuery = false,
+}: {
+  clearBeforeQuery?: boolean;
+}) => {
+  if (clearBeforeQuery) {
+    proposalsStore.setProposals([]);
```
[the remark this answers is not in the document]
peterpeterparker: Yes, I was thinking to refactor accounts and keep the pattern of this PR. Works for you - shall I add a TODO? Or rather like the accounts related pattern?

frontend/svelte/src/routes/Proposals.svelte
peterpeterparker: If a proposal is edited => store is edited => come back => re-render with updated data of state mgmt => all good.

There should be no issue. If there is, then let's fix it that when we implement the edition itself.

> "the filter subscriber will emit a first value"

When you `.subscribe` you instantly get the current value in store. In case of that page, this value should trigger a new query only if the value - i.e. the filter - change, not if the value exists. Therefore the listener should be attached only when everything is ready otherwise it triggers a query on mount what we do not want.

frontend/svelte/src/lib/utils/proposals.utils.ts
peterpeterparker: Good question. It does not cost much to create it every time and I kind of rather like it that way because doing so we avoid a global const. Also doing so it avoid questions such as "what happens if the identity has timed out, should we reset the canister references?". Works for you?

lmuntanerDfinity: if we can live with it, yes, it makes sense. Not sure how much it affects the performance.

Creating the agent not sure whether it takes time, but in that case we can just memoize the agent and that's it.

frontend/svelte/src/lib/components/proposals/ProposalCard.svelte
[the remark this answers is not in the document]
peterpeterparker: Done.

frontend/svelte/src/routes/Proposals.svelte
lmuntanerDfinity: I mean that another user edits the proposal.

frontend/svelte/src/lib/utils/proposals.utils.ts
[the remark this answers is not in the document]
lmuntanerDfinity: "canisterServices" is too obvious? or "storeServices"? or "canisterHelpers"?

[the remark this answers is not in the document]
peterpeterparker: I like short names. So "services", "helpers", "providers" or "canisters"?

frontend/svelte/src/routes/Proposals.svelte
peterpeterparker: Oh I see. In my opinion, fetching data is costly and takes time, specially when certified (currently prod does not, new requirement). Therefore I think its better to stack these in stores for performance reason as showcased in yesterday 1pm call.

If proposals or any lists of data re-edited by other users shall be automatically updated, in my opinion in my need another mechanism ("refresh button" or "push events -> refresh") rather than fetching all the time the data (because I suppose fetching happens more often than re-editing).

But again only my opinion. I plan to discuss the "certified" question once everybody will had be able to test. We can discuss this at the same time too for sure. I take a good note.

lmuntanerDfinity: You can show the details from the store, while you fetch in the background and then the store is updated. The UX is good and at the same time you make sure that the data is updated.

Anyway, this can be done in another issue.

peterpeterparker: Still costly in terms of requests on the IC and in terms of UI can also be surprising but yes agree that's also another option. Anyway as I said noted, once everybody as tested, we speak about it including UX. Thanks for the feedback and review!

frontend/svelte/src/lib/utils/proposals.utils.ts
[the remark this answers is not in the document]
lmuntanerDfinity: If they have to be connected to the store, maybe "providers" or "services" in my opinion.

nmattia (approved):
Rubber stamping to unblock, although I did not have time to review

PR #424, opened 2022-02-10
Page skeleton layout

mstrasinskis:
# Motivation

Create the components for layout - skeleton of the “sub-pages” which should work with the footer.
[L2-212](https://dfinity.atlassian.net/browse/L2-212)

image
image

# Changes

- ++ `<HeadlessLayout>`
- ++ contentless `<Wallet>` page (as PoC)
- ++ contentless `<ProposalDetails>` page (as PoC)
- ++ primitive `<Tooltip>` component (for the back button)
- ++ `<Footer>` as separate component (reusability)
- ++ `routeStore.back()`
- ++ `routeStore.previousPath`
- Routes were refactored (reusability)
- Refactored page-path-matching to support wallet and proposal_detail dynamic urls

# Notes

- Isn't Tooltip to primitive? Is it used somewhere else?
- Are there cases that could break this logic? 
```
if (process.env.REDIRECT_TO_LEGACY) {
      window.location.replace(`/${window.location.hash}`);
}
```


frontend/svelte/src/routes/ProposalDetails.svelte
```diff
@@ -0,0 +1,45 @@
+<script lang="ts">
+  import { onDestroy, onMount } from "svelte";
+  import IsolatedPageLayout from "../lib/components/common/IsolatedPageLayout.svelte";
+  import { routeStore } from "../lib/stores/route.store";
+  import { AppPath } from "./routes";
+
+  // TODO: To be removed once this page has been implemented
+  onMount(() => {
+    if (process.env.REDIRECT_TO_LEGACY) {
+      // TODO: TBD
```
mstrasinskis, line 10: 📌 Not sure about this place

frontend/svelte/src/App.svelte
```diff
@@ -11,6 +11,9 @@
   import { accountsStore } from "./lib/stores/accounts.store";
   import { onDestroy } from "svelte";
   import { AuthStore, authStore } from "./lib/stores/auth.store";
+  import Wallet from "./routes/Wallet.svelte";
+  import ProposalDetails from "./routes/ProposalDetails.svelte";
+  import { AppPath } from "./routes/routes";
```
lmuntanerDfinity, line 16: 👏

frontend/svelte/src/lib/components/common/PrivateRoute.svelte
```diff
@@ -7,14 +7,15 @@
...
 
-  export let path: string;
+  export let path: AppPath;
   export let component: typeof SvelteComponent;
 
   let signedIn: boolean = false;
 
   const redirectLogin = () => {
-    if (signedIn || routePath() !== path) {
+    if (signedIn || comparePathWithRoutePath(path, routePath()) === false) {
```
lmuntanerDfinity, line 18:
```
signedIn || !comparePathWithRoutePath(path, routePath())
```

frontend/svelte/src/lib/icons/IconBackIosNew.svelte
```diff
@@ -0,0 +1,12 @@
+<!-- source: https://fonts.google.com/icons?selected=Material+Icons&icon.query=arrow+back+ios+new -->
```
lmuntanerDfinity, line 1: We already have a back icon for the Modal, should we use the same one?

frontend/svelte/src/routes/ProposalDetails.svelte
```diff
@@ -0,0 +1,45 @@
...
+    (state) => (previousPath = state.previousPath)
+  );
+
+  onDestroy(unsubscribe);
+
+  const goBack = () => {
+    if (previousPath === AppPath.Proposals) {
+      routeStore.back();
+    } else {
+      routeStore.navigate({
```
lmuntanerDfinity, line 27: If we always go back to Proposals, why checking? to keep the navigation history?

frontend/svelte/src/lib/components/accounts/AccountCard.svelte
```diff
@@ -3,13 +3,18 @@
...
   import Identifier from "../ic/Identifier.svelte";
+  import { routeStore } from "../../stores/route.store";
+  import { AppPath } from "../../../routes/routes";
 
   export let account: Account;
 
   $: ({ identifier, balance } = account);
+
+  const cardClick = () =>
+    routeStore.navigate({ path: `${AppPath.Wallet}/${identifier}` });
```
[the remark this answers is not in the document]
mstrasinskis: Not sure, because most of them (all_paths - 2) are static string only.

frontend/svelte/src/routes/ProposalDetails.svelte
mstrasinskis: Tried to improve UX a bit 😄
@peterpeterparker what are your thoughts?

frontend/svelte/src/lib/stores/route.store.ts
```diff
@@ -13,23 +14,46 @@ export interface RouteStore {
...
     },
 
+    back: () => {
+      update((state: RouteStore) => ({
+        ...state,
+        previousPath: state.path,
+        path: state.previousPath,
+      }));
+
+      history.back();
```
peterpeterparker, line 48: For constistency reason, can you also implement `back` in `route.utils` with fallback if the history API is not supported ?

frontend/svelte/src/routes/routes.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const pathValidation = (path: AppPath): string =>
+  ({
+    // exceptions only
+    [AppPath.Wallet]: "/#/wallet/[a-zA-Z0-9]+",
+    [AppPath.ProposalDetails]: "/#/proposal/[0-9]+",
+  }[path] || path);
+
+export const comparePathWithRoutePath: (
```
peterpeterparker, line 20: we have no lint rules about it but I like for maintanability reason, as soon as more than 1 param is available, to use descriptive / interface for params.

`export const comparePathWithRoutePath: ({path, routePath}: {path: AppPath, routePath: string})`

works for you?

frontend/svelte/src/routes/ProposalDetails.svelte
peterpeterparker: we probably need the check or a check because user can access the page with alternative routes on desktop, like through bookmarks.

frontend/svelte/src/routes/ProposalDetails.svelte
```diff
@@ -0,0 +1,45 @@
...
+
+  let previousPath: string = null;
+
+  const unsubscribe = routeStore.subscribe(
+    (state) => (previousPath = state.previousPath)
+  );
+
+  onDestroy(unsubscribe);
+
+  const goBack = () => {
```
peterpeterparker, line 23: code is duplified, can you extract a utiliy function?

frontend/svelte/src/routes/ProposalDetails.svelte
peterpeterparker: As far as I remember we placed the redirect in `onMount` on purpose and/or after some tests. Therefore as it follow other routes, I think it should be good. Doesn't mean of course that you cannot challenge it, just FYI :)

peterpeterparker (changes requested):
Looks cool and works well 👍

Just a couple of comments.

frontend/svelte/src/lib/stores/route.store.ts
mstrasinskis: Just asking because of not so simple implementation, do we really need to support IE <= 9 ([history api](https://caniuse.com/history))

peterpeterparker: I see. Alright then only maybe refactor (move function, so we have all navigation related functions at the same placee) and add a comment please.

Of course if we still need .back ;)

frontend/svelte/src/routes/routes.ts
mstrasinskis: Sure, done!

peterpeterparker (approved):
LGTM, thx

lmuntanerDfinity (approved):

PR #425, opened 2022-02-11
Feat: Add Fetch Root Key in development

lmuntanerDfinity:
# Motivation

Calls in testnet need to fetch the root key.

# Changes

When creating the agent, if the environment is development, then it fetches the root key.

# Tests

No new functionality.

### DIFF
--- a/frontend/svelte/rollup.config.js
+++ b/frontend/svelte/rollup.config.js
@@ -85,7 +85,7 @@ export default {
       preventAssignment: true,
       "process.env.ROLLUP_WATCH": !!process.env.ROLLUP_WATCH,
       "process.env.FETCH_ROOT_KEY": JSON.stringify(
-        process.env.DEPLOY_ENV ? true : false
+        process.env.DEPLOY_ENV === "testnet" ? true : false
       ),
       "process.env.IDENTITY_SERVICE_URL": JSON.stringify(
         process.env.IDENTITY_SERVICE_URL ||
--- a/frontend/svelte/rollup.config.js
+++ b/frontend/svelte/rollup.config.js
@@ -84,6 +84,9 @@ export default {
     replace({
       preventAssignment: true,
       "process.env.ROLLUP_WATCH": !!process.env.ROLLUP_WATCH,
+      "process.env.FETCH_ROOT_KEY": JSON.stringify(
+        process.env.DEPLOY_ENV ? true : false
+      ),
       "process.env.IDENTITY_SERVICE_URL": JSON.stringify(
         process.env.IDENTITY_SERVICE_URL ||
           (process.env.DEPLOY_ENV === "testnet"
@@ -102,7 +105,9 @@ export default {
           ? true
           : ["false", "0"].includes(process.env.REDIRECT_TO_LEGACY)
           ? false
-          : true // default
+          : process.env.DEPLOY_ENV === "testnet"
+          ? false
+          : true
       ),
     }),
 
--- a/frontend/svelte/src/lib/utils/agent.utils.ts
+++ b/frontend/svelte/src/lib/utils/agent.utils.ts
@@ -2,4 +2,12 @@ import { HttpAgent } from "@dfinity/agent";
 import { serviceURL } from "../constants/utils.constants";
 
 // To avoid being executed in tests that only import it
-export const createAgent = () => new HttpAgent({ host: serviceURL });
+export const createAgent = async () => {
+  const agent = new HttpAgent({ host: serviceURL });
+
+  if (process.env.FETCH_ROOT_KEY === "true") {
+    await agent.fetchRootKey();
+  }
+
+  return agent;
+};
--- a/frontend/svelte/src/tests/lib/utils/accounts.utils.spec.ts
+++ b/frontend/svelte/src/tests/lib/utils/accounts.utils.spec.ts
@@ -1,3 +1,4 @@
+import type { HttpAgent } from "@dfinity/agent";
 import type { ICP } from "@dfinity/nns";
 import { AccountIdentifier, LedgerCanister } from "@dfinity/nns";
 import { loadAccounts } from "../../../lib/utils/accounts.utils";
@@ -26,7 +27,8 @@ class MockLedgerCanister extends LedgerCanister {
 describe("accounts-utils", () => {
   beforeAll(() => {
     // Needed to prevent importing Http from @dfinity/agent
-    const mockCreateAgent = () => undefined;
+    const mockCreateAgent = () =>
+      new Promise<HttpAgent>((resolve) => resolve(undefined));
     jest.spyOn(agent, "createAgent").mockImplementation(mockCreateAgent);
   });
   const mockLedgerCanister: MockLedgerCanister = new MockLedgerCanister();
--- a/frontend/svelte/src/tests/lib/utils/accounts.utils.spec.ts
+++ b/frontend/svelte/src/tests/lib/utils/accounts.utils.spec.ts
@@ -1,4 +1,3 @@
-import type { HttpAgent } from "@dfinity/agent";
 import type { ICP } from "@dfinity/nns";
 import { AccountIdentifier, LedgerCanister } from "@dfinity/nns";
 import { loadAccounts } from "../../../lib/utils/accounts.utils";
@@ -27,8 +26,7 @@ class MockLedgerCanister extends LedgerCanister {
 describe("accounts-utils", () => {
   beforeAll(() => {
     // Needed to prevent importing Http from @dfinity/agent
-    const mockCreateAgent = () =>
-      new Promise<HttpAgent>((resolve) => resolve(undefined));
+    const mockCreateAgent = () => Promise.resolve(undefined);
     jest.spyOn(agent, "createAgent").mockImplementation(mockCreateAgent);
   });
   const mockLedgerCanister: MockLedgerCanister = new MockLedgerCanister();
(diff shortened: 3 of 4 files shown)


lmuntanerDfinity:
@nmattia please review. I think that we discussed it the other day. I ended up not using it in my issue, but I had forgot about it and we will need it next week when we start interacting with testnet using nns-js.

lmuntanerDfinity:
@mstrasinskis @peterpeterparker fyi, this is needed for the calls to testnet to work properly.

frontend/svelte/src/tests/lib/utils/accounts.utils.spec.ts
```diff
@@ -26,7 +27,8 @@ class MockLedgerCanister extends LedgerCanister {
 describe("accounts-utils", () => {
   beforeAll(() => {
     // Needed to prevent importing Http from @dfinity/agent
-    const mockCreateAgent = () => undefined;
+    const mockCreateAgent = () =>
+      new Promise<HttpAgent>((resolve) => resolve(undefined));
```
nmattia, line 31:
```suggestion
    const mockCreateAgent = Promise.resolve(undefined)
```

Is this actually being used? I'm surprised not to see any changes at call sites

frontend/svelte/src/lib/utils/agent.utils.ts
```diff
@@ -2,4 +2,12 @@ import { HttpAgent } from "@dfinity/agent";
 import { serviceURL } from "../constants/utils.constants";
 
 // To avoid being executed in tests that only import it
-export const createAgent = () => new HttpAgent({ host: serviceURL });
+export const createAgent = async () => {
+  const agent = new HttpAgent({ host: serviceURL });
+
+  if (process.env.FETCH_ROOT_KEY === "true") {
```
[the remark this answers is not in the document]
lmuntanerDfinity: Oops, I forgot to add the changes in rolllup confit, I always skip those because I change it for local development

frontend/svelte/src/tests/lib/utils/accounts.utils.spec.ts
lmuntanerDfinity: We mock the `LedgerCanister.create`, therefore the `createAgent` is never used in the tests.

```
jest
      .spyOn(LedgerCanister, "create")
      .mockImplementation((): LedgerCanister => mockLedgerCanister);
```

Yet, to avoid a dependency on fetch when creating the agent from `@dfinity/agent`, we need to avoid importing it. Without a mockImplementation, it imports the file, also the library and the error appears.

```
    Fetch implementation was not available. You appear to be in a Node.js context, but global.fetch was not available.

      4 | // To avoid being executed in tests that only import it
      5 | export const createAgent = async () => {
    > 6 |   const agent = new HttpAgent({ host: serviceURL });
        |                 ^
      7 |
      8 |   if (process.env.FETCH_ROOT_KEY === "true") {
      9 |     await agent.fetchRootKey();

      at getDefaultFetch (node_modules/@dfinity/agent/src/agent/http/index.ts:100:13)
      at new HttpAgent (node_modules/@dfinity/agent/src/agent/http/index.ts:147:38)
      at createAgent (src/lib/utils/agent.utils.ts:6:17)
      at loadAccounts (src/lib/utils/accounts.utils.ts:12:29)
      at Object.<anonymous> (src/tests/lib/utils/accounts.utils.spec.ts:41:23)
```

frontend/svelte/rollup.config.js
```diff
@@ -102,7 +105,9 @@ export default {
           ? true
           : ["false", "0"].includes(process.env.REDIRECT_TO_LEGACY)
           ? false
-          : true // default
+          : process.env.DEPLOY_ENV === "testnet"
```
lmuntanerDfinity, line 108: I added this to avoid having the change this every time we develop locally.

We can assume this, is that correct?

frontend/svelte/src/lib/utils/agent.utils.ts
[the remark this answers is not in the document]
lmuntanerDfinity: Here it is: https://github.com/dfinity/nns-dapp/pull/425/files#diff-74abfab6043a5b76506f46cdfcf671a49f4c45cb86796255a237b2b66bdd685fR87

lmuntanerDfinity:
@nmattia ready for another one 😸

frontend/svelte/src/tests/lib/utils/accounts.utils.spec.ts
nmattia: I'm still a bit confused about why having `mockCreateAgent` makes an error about `global.fetch` not existing disappear, but I don't have much experience with typescript :)

frontend/svelte/rollup.config.js
nmattia: I _think_ so! but does that mean that you set `DEPLOY_ENV` to `testnet` when developing locally?

nmattia: Wait, this is `REDIRECT_TO_LEGACY`, I misread. I never quite understood what it did so better check with @bitdivine

nmattia: I actually have no idea, I haven't actively worked on the frontend here!

nmattia (commented):
Unblocking! Though not quite approving because I didn't grasp everything re. the mock config, leaving this up to @bitdivine and @peterpeterparker

nmattia (approved):
Unblocking

frontend/svelte/rollup.config.js
peterpeterparker: `REDIRECT_TO_LEGACY` = svelte redirect to flutter

Therefore I think it cannot be extended as suggested here. Current staging deployment happens my understanding with Redirect and Testnet at the same time.

But yes @bitdivine can confirm.

frontend/svelte/src/lib/utils/agent.utils.ts
peterpeterparker, line 8: the value is stringifie in the rollup config => `if (process.env.FETCH_ROOT_KEY)`

peterpeterparker, line 8: the value is stringified in the rollup config => `if (process.env.FETCH_ROOT_KEY) {`

peterpeterparker (changes requested):
We are already using the configuration `DEPLOY_ENV` in `<Header/>`  for a check `{#if deployEnv === "testnet"}`.

Therefore I think it is not a good idea to have this test and another different configuration `FETCH_ROOT_KEY` that both check for the exact same equality on "testnet".

Possible solutions:

- rename `FETCH_ROOT_KEY` to something more generic and update  - use it in`<Header/>`

- remove `FETCH_ROOT_KEY` and use `DEPLOY_ENV` in `createAgent`

frontend/svelte/src/lib/utils/agent.utils.ts
lmuntanerDfinity: Yes, but depends on how we agree, this might change.

lmuntanerDfinity:
> We are already using the configuration `DEPLOY_ENV` in `<Header/>` for a check `{#if deployEnv === "testnet"}`.
> 
> Therefore I think it is not a good idea to have this test and another different configuration `FETCH_ROOT_KEY` that both check for the exact same equality on "testnet".
> 
> Possible solutions:
> 
> * rename `FETCH_ROOT_KEY` to something more generic and update  - use it in`<Header/>`
> * remove `FETCH_ROOT_KEY` and use `DEPLOY_ENV` in `createAgent`

I agree on refactor and simplify the process. Also the `REDIRECT_TO_LEGACY` I'd say:

```
      "process.env.REDIRECT_TO_LEGACY": JSON.stringify(
        ["true", "1"].includes(process.env.REDIRECT_TO_LEGACY)
          ? true
          : ["false", "0"].includes(process.env.REDIRECT_TO_LEGACY)
          ? false
          : process.env.DEPLOY_ENV === "testnet"
          ? false
          : true
```

Let me set up a document to see if I understand the usage of each and we can agree on how to simplify it.

frontend/svelte/rollup.config.js
```diff
@@ -84,6 +84,9 @@ export default {
     replace({
       preventAssignment: true,
       "process.env.ROLLUP_WATCH": !!process.env.ROLLUP_WATCH,
+      "process.env.FETCH_ROOT_KEY": JSON.stringify(
+        process.env.DEPLOY_ENV === "testnet" ? true : false
```
mstrasinskis, line 88: I guess `? true : false` is redundant.

frontend/svelte/rollup.config.js
bitdivine: Correct, pre-release deployments happen with REDIRECT_TO_LEGACY unset, defaulting to true, however I can always set the env var `REDIRECT_TO_LECACY=1` as that will take priority over this change, so this is not a problem for me.  I would be fine with that.  If this does get merged, please ping @hpeebles as he mostly develops on the flutter side of the app and would be surprised if flutter did not show up.  I was using the rule of thumb:  "Make it like production except when the user says otherwise."

lmuntanerDfinity: thanks! I am continuing the conversation in the Notion page, we can then use the decisions there to change this PR, update the config and also write this in comments or in a markdown doc in the same repo.

lmuntanerDfinity:
Closing, changes are stale and not worth it.

PR #427, opened 2022-02-14
Input with button

mstrasinskis:
# Motivation

Expand Input Component to take inner buttons (eg Stake Neuron / Amount field)
https://dfinity.atlassian.net/browse/L2-252

# Changes

- `<button class="input-button">` style
- `<Input>` component now has a `<slot name="button">`
- Updated Input styles (desktop/mobile) to look a bit more like flutter dapp input.

# Tests
- No logic to test 

# Notes

## Mobile
image
image

## Desktop
image
image


frontend/svelte/src/lib/components/ui/Input.svelte
```diff
@@ -31,13 +31,25 @@
...
+
   .input-block {
     position: relative;
 
     margin: calc(2 * var(--padding)) 0;
+
+    display: flex;
+    align-items: center;
+
+    :global(button) {
```
lmuntanerDfinity, line 49: what does this mean?

mstrasinskis: It applies the style to the :global( `selector` ). In this case to any `<button>` tag inside this input component.

lmuntanerDfinity: why not just `button`, since we are already inside `.input-block`?

I can't find the docs for this `:global` feature. I am sorry for my lack of SCSS knowledge... 😅

mstrasinskis: The button is from another scope, not part of the component. So it will be ignored I guess 🤷‍♂️

lmuntanerDfinity: the slot `<slot name="button" />` is inside `<div class="input-block">` so it should not be ignore if you use `button`, shouldn't it?

frontend/svelte/src/lib/components/ui/Input.svelte
```diff
@@ -31,13 +31,25 @@
...
+      position: absolute;
```
lmuntanerDfinity, line 50: With this approach, if the user writes until the end of the input, it will go behind the button, is that right?

Not that it matters much, we use it for amounts only for now, but just checking that I understand the solution.

mstrasinskis: Yes, you are right. Same for the flutter-app, so not sure if we need to improve this now.

lmuntanerDfinity: Not for this rewrite in my opinion. The only moments I have seen the usage of this button is for amounts. Amounts won't be that long, right? 🤑

mstrasinskis: I don't see any simple solution (w/o JS). So let's implement this in the next redesigned version ;)

frontend/svelte/src/lib/components/ui/Input.svelte
mstrasinskis: Unfortunately just a button selector doesn't work. My assumption -- local-scope mechanism can't find the button in the Input component, so it doesn't even generate the css code for the button block.

# with :global
```scss
    :global(button) {
      position: absolute;
      right: calc(2 * var(--padding));
    }
```
image

# w/o :global
```scss
    button {
      position: absolute;
      right: calc(2 * var(--padding));
    }
```
image

lmuntanerDfinity (approved):

frontend/svelte/src/lib/components/ui/Input.svelte
lmuntanerDfinity: we might not want buttons inside an input in the next redesigned version

peterpeterparker (approved):
LGTM

Not related to your PR but made me think:

- we maybe need to cleanup globally box-shadow variables. we began to have quite some shadows everywhere...

- i'm looking forward to have e2e snapshots comparison in place to tests style changes

PR #428, opened 2022-02-14
Add media.min-width unknown-breakpoint catching

mstrasinskis:
# Motivation

Currently media-mixin allows any value in argument and swallows unknown.

# Changes

- Added @throw in case of unknown breakpoint.

# Notes

Now it throws an error:

image


lmuntanerDfinity (approved):

peterpeterparker (approved):
TIL you can throw error in a mixin. Thx!

PR #429, opened 2022-02-14
feat: Add max_ingress_messages_per_block to UpdateSubnetPayload.

bitdivine:
# Motivation
The `max_ingress_messages_per_block` field is being added to the upstream `UpdateSubnetPayload` in this PR: https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/3037/diffs  The nns-dapp will have to be able to display that field.

# Changes
Add `max_ingress_messages_per_block` in all the relevant places.

# Tests
- Deployed to a testnet with an up-to-date governance canister.
- Verified that new update proposals render correctly.
- Observed that, as expected, old update proposals do not render.


frontend/ts/src/canisters/governance/nnsFunctions/payloads.did
```diff
@@ -116,6 +116,7 @@ type UpdateSubnetPayload = record {
   pfn_evaluation_period_ms : opt nat32;
   subnet_id : principal;
   max_ingress_bytes_per_message : opt nat64;
+  max_ingress_messages_per_block : opt nat64;
```
bitdivine, line 119: Note: The upstream did file is now committed - hooray - and may be found here:  https://gitlab.com/dfinity-lab/public/ic/-/blob/master/rs/registry/canister/canister/registry.did

Note that there is quite a delta between the upstream did and the one in this repo; we should perhaps reduce that delta.

hpeebles (approved):

rs/nns_functions_candid_gen/src/payloads.rs
```diff
@@ -98,6 +98,7 @@ pub struct UpdateSubnetPayload {
 
     pub ingress_bytes_per_block_soft_cap: Option<u64>,
     pub max_ingress_bytes_per_message: Option<u64>,
+    pub max_ingress_messages_per_block: Option<u64>,
```
hpeebles, line 101: You should update the url in the comment preceding this struct.

PR #430, opened 2022-02-14
Sort CSS using rollup plugin

nmattia:
This reverts to the original plugin (pre https://github.com/dfinity/nns-dapp/pull/409) and actually performs the sorting in the rollup config as opposed to using an extra bash script. The plugin introduced in #409 seems unmaintained, see also https://github.com/dfinity/nns-dapp/pull/409#discussion_r800525016

I've also created an issue upstream and waiting for more info on the maintainer re. what the root cause may be, and potentially how to avoid the `mkdir` step here: https://github.com/thgh/rollup-plugin-css-only/issues/42

--- 

NOTE: forgot to mention, by simply reverting to `css({ output: "bundle.css" })` I was able to reproduce the issue on every build. After sorting, I cannot reproduce the issue anymore.


peterpeterparker (approved):
LGTM. Thanks a lot!

Note about test: I tested the branch locally. Did a `npm run build` and then ran the app with a local simulated server (`npx serve public`). App did still look as excepted.

PR #431, opened 2022-02-14
build: i18n script mjs to js

peterpeterparker:
# Motivation

As of Node v16.14.0, json import requires an addition assertion (as in the [tc39 proposal](https://github.com/tc39/proposal-import-assertions).

Unfortunately, until v16.13.x, such assertion is not supported nor ignored.

Since we are using misc Node version and because we cannot add the assertion while remaining backwards compatible, the current easiest way to solve the issue it to transform the module script to a standard js script aka use good old `require`. 

# Changes

- transform `.mjs` i18n script to `.js`
- use `require`

# Tests

Add some modification in auto-generated i18n files, run a build, changes were correctly overwritten by build.

### DIFF
--- a/frontend/svelte/package.json
+++ b/frontend/svelte/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "copy:index": "node ./scripts/copy.index.mjs",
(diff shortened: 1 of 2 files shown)


nmattia (approved):
Thanks!

PR #432, opened 2022-02-14
Use PrincipalId rather than Vec<u8> in payload definition

hpeebles:
# Motivation

We should display Principals in human readable form (base32 encoded), but for the new RemoveNodeOperators proposal type, the field in the payload is defined using byte arrays rather than Principals. This is because the Rust type was auto generated from the proto type definition.

A Principal _is_ just a byte array but candid detects Principals and deserializes them into the Principal class. This allows us to display them in human readable form. So in order to make the RemoveNodeOperators proposal payloads display nicely we simply need to change the NNS Dapp's copy of the payload definition to use Principals rather than byte arrays.

# Changes

The RemoveNodeOperatorsPayload `node_operators_to_remove` field was changed from a Vec of byte array to a Vec of PrincipalId.
This PR also removes all of the unused proto attributes from the payload definitions.

# Tests

Forthcoming... I'm trying to create a dummy proposal of type RemoveNodeOperators but am currently running into deployment issues.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1753,8 +1753,6 @@ dependencies = [
  "candid_derive",
  "ic-base-types",
  "ic-registry-subnet-type",
- "prost",
- "prost-build",
  "serde",
 ]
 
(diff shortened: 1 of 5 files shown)


hpeebles:
Turns out this fix attempt didn't work since the candid encoding of a Principal is different to the encoding of a byte array.
```
Error: type mismatch: type on the wire rec_690, expect type principal
```

PR #439, opened 2022-02-14
L2 256 routing fallback

mstrasinskis:
# Motivation

Currently the user sees a blank page if there is an error in the url. The goal is to redirect to the `/accounts` page on any unknown urls.
https://dfinity.atlassian.net/browse/L2-256

# Changes

- ++ urlToAppPath()
- ++ navigation fallback in the App.svelte (because the routing happening here)

# Tests

Extended `routes.spec.ts`


frontend/svelte/src/App.svelte
```diff
@@ -26,7 +27,16 @@
     }
   );
 
-  onDestroy(unsubscribe);
+  const unsubscribeRoute = routeStore.subscribe((route) => {
```
peterpeterparker, line 30: - for consistency reason, can you add the types? we still have to discuss if we want explicit or implicit but until then, good to have everything look the same

- you can probably deconstruct the route, it is never null if I am not wrong

- goodie: i think it is more readable when the `test` for inequality and return

```
const unsubscribeRoute: Unsubscriber = routeStore.subscribe(({path}: RouteStore) => {
    if (urlToAppPath(path) !== null) {
       return;
    }
    
    routeStore.replace....
```

frontend/svelte/src/routes/routes.ts
```diff
@@ -10,13 +10,18 @@ export enum AppPath {
...
 
-export const pathValidation = (path: AppPath): string =>
+const pathValidation = (path: AppPath): string =>
   ({
     // exceptions only
     [AppPath.Wallet]: `${AppPath.Wallet}/[a-zA-Z0-9]+`,
     [AppPath.ProposalDetails]: `${AppPath.ProposalDetails}/[0-9]+`,
   }[path] || path);
 
+export const urlToAppPath = (url: string): AppPath =>
```
peterpeterparker, line 20: if you rename `url` to `routePath` you spare the assignment `routePath: url`

frontend/svelte/src/routes/routes.ts
```diff
@@ -10,13 +10,18 @@ export enum AppPath {
   ProposalDetails = "/#/proposal",
 }
 
-export const pathValidation = (path: AppPath): string =>
+const pathValidation = (path: AppPath): string =>
```
peterpeterparker, line 13: we have these functions in `route.ts` but also others in `route.utls.ts`. Probably worth cleaning a bit and not have functions there and there. I would suggest to move functions to `route.utils` but open to any other suggestion? what do you think?

peterpeterparker (changes requested):
A couple of comment about code but the fallback looks good 👍.

frontend/svelte/src/routes/routes.ts
mstrasinskis: True 🤦‍♂️

frontend/svelte/src/routes/routes.ts
mstrasinskis: I also don't like the current structure. The biggest disadvantage of moving to route.utils is that the utils then should know about the AppPath which is a part of business logic. I'd rather move them to route.store or you find having AppPath in the utils isn't so bad, @peterpeterparker?

mstrasinskis: There is always an option to extend the store with meta props like `isKnownPath` or `appPath: AppPath`

peterpeterparker: Agree with you, it is not elegant to have these in utils. In `route.store` sounds fine or maybe moving them in `constants` folder?

peterpeterparker: `isKnownPath` can also be cool!

peterpeterparker (approved):
LGTM, thanks!

I added a last comment about the naming of a function and if you can add one or two tests to `route.utils.spec.ts` for `testRoutePath` with sub-routes (`/proposal/123`), since we do not have such tests yet, that would be cool.

frontend/svelte/src/App.svelte
```diff
@@ -26,7 +27,19 @@
     }
   );
 
-  onDestroy(unsubscribe);
+  const unsubscribeRoute: Unsubscriber = routeStore.subscribe(
+    ({ isKnownPath }) => {
```
lmuntanerDfinity, line 31: why not moving the `isKnownPath` helper here and get it from the path?

I feel that we are duplicating state with the path and whether it's known or not. One is the calculation of the other.

PR #441, opened 2022-02-15
fix: i18n type definition keys

peterpeterparker:
# Motivation

Some i18n keys were not generated and added to the description file

# Changes

- generate keys from `en.json` and update `i18n.d.ts` with the related keys


mstrasinskis (approved):
LGTM

PR #442, opened 2022-02-15
feat: transfer icps on testnet

peterpeterparker:
# Motivation

Transfer ICPs to a main account for testing purpose. Feature limited to testnet.

# Changes

- effective implementation to transfer ICPs for testing purpose
- display related potential errors
- add `disabled` attribute to input
- fix `<Spinner />` not perfectly absolutely centered
- set a `min-height` to primary buttons

# Tests

No tests were added to test the transfer feature itself because it is a restricted test feature. A test is provided to ensure the function can only be called if the environment is "testnet".

# Screenshots

ezgif com-gif-maker


frontend/svelte/src/lib/services/dev.services.ts
```diff
@@ -0,0 +1,23 @@
+import type { BlockHeight, TransferError } from "@dfinity/nns";
+import { get } from "svelte/store";
+import { E8S_PER_ICP } from "../constants/icp.constants";
+import { AccountsStore, accountsStore } from "../stores/accounts.store";
+import { AuthStore, authStore } from "../stores/auth.store";
+import { acquireICPTs } from "../utils/dev.utils";
+
+export const getIPCs = async (icps: number) => {
```
lmuntanerDfinity, line 8: typo, to `getICPs`?

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -0,0 +1,59 @@
...
+  const publicKey = "Uu8wv55BKmk9ZErr6OIt5XR1kpEGXcOSOC1OYzrAwuk=";
+  const privateKey =
+    "N3HB8Hh2PrWqhWH2Qqgr1vbU9T3gb1zgdBD8ZOdlQnVS7zC/nkEqaT1kSuvo4i3ldHWSkQZdw5I4LU5jOsDC6Q==";
+  const identity = Ed25519KeyIdentity.fromKeyPair(
+    base64ToUInt8Array(publicKey),
+    base64ToUInt8Array(privateKey)
+  );
+
+  const agent: HttpAgent = new HttpAgent({
+    host: "https://nnsdapp.dfinity.network/",
```
lmuntanerDfinity, line 37: Add a todo to move this to env var or config?

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -0,0 +1,59 @@
...
+    identity,
+  });
+  await agent.fetchRootKey();
+
+  const ledgerCanister: LedgerCanister = LedgerCanister.create({ agent });
```
lmuntanerDfinity, line 42: how does it know which canister id to make the call to?

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -0,0 +1,59 @@
...
+  const privateKey =
+    "N3HB8Hh2PrWqhWH2Qqgr1vbU9T3gb1zgdBD8ZOdlQnVS7zC/nkEqaT1kSuvo4i3ldHWSkQZdw5I4LU5jOsDC6Q==";
+  const identity = Ed25519KeyIdentity.fromKeyPair(
+    base64ToUInt8Array(publicKey),
+    base64ToUInt8Array(privateKey)
+  );
+
+  const agent: HttpAgent = new HttpAgent({
+    host: "https://nnsdapp.dfinity.network/",
+    identity,
```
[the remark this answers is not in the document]
peterpeterparker: I did not check the canister source code but that's generally how the IC works. You create an `agent` with the `identity` of the signed in user and in the backend code you use the caller to identify the user.

Like in Motoko:

```
public shared query({ caller }) func get(key: Text) : async (?Data) {
    if (Utils.isPrincipalNotEqual(caller, user)) {
        throw Error.reject("User does not have the permission to get the data.");
    };
```

caller = the one who's calling = the principal of the signed in user if function is called with the agent from the app as described above.

frontend/svelte/src/lib/utils/dev.utils.ts
peterpeterparker: I add a TODO at first but then had the feeling it was not used anywhere else, that's why I removed the todo.
Added a comment now.

frontend/svelte/src/lib/utils/dev.utils.ts
peterpeterparker: my guess is that the ledger has the same id (a guess) and the transfer on this testnet is authorized because the host `https://nnsdapp.dfinity.network` is used

code is copied 1 to 1 from `ts` project.

frontend/svelte/src/lib/utils/dev.utils.ts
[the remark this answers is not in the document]
lmuntanerDfinity: So we are going to need this also in our ledger utils?

How do we get the identity of the loged in user?

frontend/svelte/src/lib/utils/dev.utils.ts
lmuntanerDfinity: if we have the identity we don't need the canister id?

Who can help us understand this>

frontend/svelte/src/lib/utils/dev.utils.ts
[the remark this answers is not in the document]
peterpeterparker: `authStore.getIdentity().getPrincipal()`

have a look to `accountsStore.sync()`

frontend/svelte/src/lib/utils/dev.utils.ts
peterpeterparker: generally speaking or for that PR?

for that PR:

identity = signed in user
canister id = no (my understanding, `host` is what we need for that particular service)

generally speaking:

identity = signed in user
canister id = provider

mstrasinskis (approved):
LGTM
Many thanks to @lmuntanerDfinity for questions and to @peterpeterparker for answers 👍

lmuntanerDfinity (approved):

PR #443, opened 2022-02-15
Update IC dependency to latest version

hpeebles:
# Motivation

The NNS Dapp has a dependency on the IC repo (https://github.com/dfinity/ic) but prior to this commit it was referencing a very old version.
This PR updates this dependency to the latest version along with a few other dependencies.

# Changes

- Update IC dependency to commit 42a4bc4b7917a44c58c9cf907d84b9fd8e908ed9
- Update a few additional dependencies (candid, serde, serde_bytes, serde_cbor).
- Update Rust version from 1.54.0 to 1.58.1

# Tests

I deployed the current mainnet commit to a testnet (https://sp3hj-caaaa-aaaaa-aaajq-cai.nnsdapp.dfinity.network/v2/), added a load of accounts, transactions and neurons, then successfully upgraded the canister to this version and can see the data is still there and the Dapp still works.


rs/nns_functions_candid_gen/src/payloads.rs
```diff
@@ -145,34 +144,28 @@ pub struct RecoverSubnetPayload {
...
+    pub firewall_config: String,
     /// List of allowed IPv4 prefixes
-    #[prost(string, repeated, tag = "2")]
-    pub ipv4_prefixes: ::prost::alloc::vec::Vec<::prost::alloc::string::String>,
+    pub ipv4_prefixes: Vec<String>,
     /// List of allowed IPv6 prefixes
-    #[prost(string, repeated, tag = "3")]
-    pub ipv6_prefixes: ::prost::alloc::vec::Vec<::prost::alloc::string::String>,
+    pub ipv6_prefixes: Vec<String>,
 }
```
bitdivine, line 155: Just curious, why are the prost types not needed any more?

bitdivine: And the prost tags?  I thought the prost tags were there to make sure that if a tag was removed or added in the middle, the values of later tags would not change.

rs/src/accounts_store.rs
```diff
@@ -98,23 +96,23 @@ pub enum TransactionToBeProcessed {
 #[derive(Copy, Clone, CandidType, Deserialize)]
 pub struct CreateCanisterArgs {
     pub controller: PrincipalId,
-    pub amount: ICPTs,
+    pub amount: Tokens,
```
bitdivine, line 99: The nns-js still uses ICP - should that convert as well?

rs/src/accounts_store.rs
```diff
@@ -1021,42 +1019,12 @@ impl AccountsStore {
     }
 
     fn get_transaction_index(&self, block_height: BlockHeight) -> Option<TransactionIndex> {
-        // The binary search methods are taken from here (they will be in stable rust shortly) -
```
bitdivine, line 1024: Nice

rs/src/canisters/governance.rs
```diff
@@ -1,8 +1,6 @@
-use candid::CandidType;
 use dfn_candid::candid;
-use dfn_core::api::PrincipalId;
 use ic_nns_constants::GOVERNANCE_CANISTER_ID;
-use serde::Deserialize;
+use ic_nns_governance::pb::v1::{ClaimOrRefreshNeuronFromAccount, ClaimOrRefreshNeuronFromAccountResponse};
```
bitdivine, line 3: Are these identical to the types they replace?

bitdivine (approved):

rs/src/accounts_store.rs
hpeebles: Yes, that way it'll be compatible with each of the SNS ledgers once they start appearing.
But there's no rush on that since serialization will still work due to the types being the same internally.

rs/src/canisters/governance.rs
hpeebles: Yes, I originally copied them from the gov canister's code, but now I thought it made more sense to simply pull in that code as a dependency.
Going forward we should regularly update the dependency to ensure the type definitions remain in sync.

rs/nns_functions_candid_gen/src/payloads.rs
hpeebles: We are only using this executable to generate the candid files, those candid files don't care about prost tags, so the output remained the same after removing the dependency on prost.

PR #444, opened 2022-02-15
build: CSP and other adjustements to make deploy happens

peterpeterparker:
# Motivation

Define the Content Security Policies (CSP) that should be applied in production.

Make the svelte app deployment to canisters - i.g. testnet and mainnet - happens (without adjustements the app can indeed be deployed but is not functionable).

# Changes

- prerender the CSP meta rules statically within the `index.html`
- apply rules (following [Internet Identity](https://github.com/dfinity/internet-identity/blob/c5709518ce3daaf7fdd9c7994120b66bd613f01b/src/internet_identity/src/main.rs#L824) example)
- refactor `favicon.png` to use the same asset in both svelte and flutter app
- remove CSP rule for local development only
- fix "no sourcemap in production build"
- extract logic and default environment variables to a new `env.config.mjs` script 

# Note

This PR does **not** add new env variables to the rollup configuration and per extension to the JS code. In addition to CSP, it only cleans up and group variables within a same configuration that can be use for the svelte build.


frontend/svelte/env.config.mjs
```diff
@@ -0,0 +1,60 @@
...
+const REDIRECT_TO_LEGACY = ["true", "1"].includes(
+  process.env.REDIRECT_TO_LEGACY
+)
+  ? true
+  : ["false", "0"].includes(process.env.REDIRECT_TO_LEGACY)
+  ? false
+  : true; // default
+
+export const envConfig = {
+  PRODUCTION: !process.env.ROLLUP_WATCH,
```
lmuntanerDfinity, line 49: wouldn't you use the `FETCH_ROOT_KEY` env var?

frontend/svelte/env.config.mjs
```diff
@@ -0,0 +1,60 @@
...
+  DEPLOY_ENV: process.env.DEPLOY_ENV,
+  REDIRECT_TO_LEGACY,
+  MAINNET,
+  HOST,
+  OWN_CANISTER_ID: process.env.OWN_CANISTER_ID,
```
lmuntanerDfinity, line 54: What do you think of:  https://dfinity.atlassian.net/browse/L2-265

We can do this when @nmattia comes back from holidays thouth. No rush.

frontend/svelte/env.config.mjs
```diff
@@ -0,0 +1,60 @@
...
+const GOVERNANCE_CANISTER_URL = `https://${GOVERNANCE_CANISTER_ID}${domain}/`;
+const LEDGER_CANISTER_URL = `https://${LEDGER_CANISTER_ID}${domain}/`;
+
+// When developing with live reload in svelte, redirecting to flutter is
+// not desirable.  The default should match production:
+// - false while svelte is inactive
+// - true while flutter is being replaced by svelte
+// - false after flutter has been replaced, but before all scaffolding has been removed
+// - the flag may then be removed.
+const REDIRECT_TO_LEGACY = ["true", "1"].includes(
```
lmuntanerDfinity, line 40: Can't we simplify this?

And also set it to `false` if we are in local environment? Maybe checking the `ROLLUP_WATCH` env var?

frontend/svelte/src/index.html
```diff
@@ -6,16 +6,18 @@
...
-    />
+    <link rel="icon" href="/assets/assets/favicon.png" type="image/png" />
 
     <link rel="stylesheet" href="./build/bundle.css" />
 
-    <script type="module" src="./build/bundle.js"></script>
+    <script
+      type="module"
+      src="./build/bundle.js"
+      nonce="bundle-369ac6c9-8078-4625-82f7-f37a9ca8fb16"
```
lmuntanerDfinity, line 16: what is this `bundle-369ac6c9-8078-4625-82f7-f37a9ca8fb16`?

frontend/svelte/env.config.mjs
peterpeterparker: I'll add `FETCH_ROOT_KEY` in the config too.

Note: This PR does not aims to introduce changes and new feature. It group and set the default values as they already exists. We future PR we can improve and add new variables.

frontend/svelte/env.config.mjs
peterpeterparker: I'll have a look, same as previously, this is not the goal of this PR.

frontend/svelte/src/index.html
peterpeterparker: It's the `nonce` identifier. I use a random uid because I read it is a bad practice to only identify nonce with short dummy names.

frontend/svelte/env.config.mjs
lmuntanerDfinity: I know, it was just FYI

frontend/svelte/env.config.mjs
lmuntanerDfinity: thanks

frontend/svelte/env.config.mjs
lmuntanerDfinity: ok

frontend/svelte/env.config.mjs
peterpeterparker: done

frontend/svelte/env.config.mjs
```diff
@@ -0,0 +1,68 @@
...
+// agent-js connects per default to mainnet with the anonymous identity
+const MAINNET = `https://${domainMainnet}`;
+
+const IDENTITY_SERVICE_URL =
+  process.env.IDENTITY_SERVICE_URL ||
+  (process.env.DEPLOY_ENV === "testnet"
+    ? `https://qjdve-lqaaa-aaaaa-aaaeq-cai.${domainTestnet}/`
+    : "https://identity.ic0.app/");
+
+const HOST =
```
lmuntanerDfinity, line 31: If we don't set deploy_env, that means local, right?

frontend/svelte/env.config.mjs
```diff
@@ -0,0 +1,68 @@
...
+  process.env.REDIRECT_TO_LEGACY
+)
+  ? true
+  : ["false", "0"].includes(process.env.REDIRECT_TO_LEGACY)
+  ? false
+  : true; // default
+
+export const envConfig = {
+  ENVIRONMENT,
+  DEPLOY_ENV: process.env.DEPLOY_ENV,
```
mstrasinskis, line 57: Something for later improvement: `DEPLOY_ENV` is used only to indicate "testnet". Probably we could simplify the name to `USE_TESTNET=true`.

frontend/svelte/env.config.mjs
```diff
@@ -0,0 +1,68 @@
+/**
+ * As long as we use the Flutter app, we need to handle the configuration with miscellaneous environment variables.
+ * In addition, some of these variables have default fallback values according their environment - e.g. testnet or mainnet.
+ *
+ * That's why we group here the logic and default values to expose a single object - envConfig - that contains the effective configuration.
+ *
+ * The configuration is use in the rollup build but also in the parser of the static files - e.g. build.index.mjs to output the index.html with a CSP.
+ */
+
+const ENVIRONMENT = process.env.ROLLUP_WATCH
```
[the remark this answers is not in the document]
peterpeterparker: you still can, that's fine. we need extra information for example to do a difference between `DEPLOY_ENV=testnet npm run dev`  and `DEPLOY_ENV=testnet npm run build`. There is the environment you use (DEPLOY_ENV) and the environment where you are (local, testnet or mainet = ENVIRONMENT).

Not against better naming though.

frontend/svelte/env.config.mjs
peterpeterparker: Actually in the Flutter app, where this env. variable is used, there is also `DEPLOY_ENV=staging` in addition to testnet and mainnet 🤷‍♂️

Did not implemented here for the moment because as far as I know we don't use it.

But sure, if it is useless and assuming we update the CI and the Flutter app, we cann rename to `USE_TESTNET` that would be clearer.

frontend/svelte/env.config.mjs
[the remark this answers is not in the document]
lmuntanerDfinity: ok, so we can use "testnet" in different environments?

lmuntanerDfinity (approved):
Still not fully understand all the env vars, but looks good to me. I just need more time, hehe.

peterpeterparker:
@lmuntanerDfinity it can definitely be cleaned (in flutter and ci too), that would make things easier, agree!

sure you can use testnet (backend) locally and deploy on testnet (frontend and backend) too. as I said, maybe there is better naming than `ENVIRONMENT`. ping me with better naming idea anytime.

 thx for the review

PR #447, opened 2022-02-16
LM-259: Add Account

lmuntanerDfinity:
# Motivation

Create SubAccount

# Changes

* New Modal `AddAccountModal`
* New accounts store method `createSubAccount`
* New accounts utils function `createSubAccount`
* Using new nns-js function `createSubAccount`

# Tests

* `AddAccountModal.spec.ts`
* `accounts.services.spec.ts`


lmuntanerDfinity:
Screenshots below

Screenshot 2022-02-16 at 10 01 35
Screenshot 2022-02-16 at 10 01 42

lmuntanerDfinity:
BLOCKED by https://github.com/dfinity/nns-js/pull/36

frontend/svelte/src/routes/Accounts.svelte
```diff
@@ -55,12 +59,13 @@
...
             >{$i18n.accounts.new_transaction}</button
           >
-          <button class="primary" on:click={addAccount}
+          <button class="primary" on:click={openAddAccountModal}
             >{$i18n.accounts.add_account}</button
           >
         </Toolbar>
       {/if}
     </svelte:fragment>
+    <AddAcountModal visible={showAddAccountModal} on:nnsClose={closeModal} />
```
mstrasinskis, line 68: There is one `<main>` tag in `<Layout>` component. Probably the Layout needs to be extended w/ the `hidden` prop.

> A document must not have more than one [main](https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element) element that does not have the [hidden](https://html.spec.whatwg.org/multipage/interaction.html#attr-hidden) attribute specified.

peterpeterparker:
FYI As we discussed in one of my PR, I provided a PR that refactor utils in services and refactored the logic to load the accounts as we expected. I guess once merged it will conflicts with your PR.

👉 https://github.com/dfinity/nns-dapp/pull/452

frontend/svelte/src/lib/services/accounts.services.ts
```diff
@@ -47,5 +53,40 @@ const loadAccounts = async ({
...
+  // TODO: Remove and understand L2-301
+  while (tries < MAX_TRIES && newSubAccount === undefined) {
+    tries += 1;
+    try {
+      newSubAccount = await nnsDapp.createSubAccount({
+        subAccountName: name,
+      });
+    } catch (currentError) {
+      error = currentError;
+      if (currentError.kind !== AccountNotFoundError.kind) {
```
mstrasinskis, line 80: Why not `currentError  instanceof AccountNotFoundError`?

frontend/svelte/src/lib/services/accounts.services.ts
```diff
@@ -47,5 +53,40 @@ const loadAccounts = async ({
...
+
+  const MAX_TRIES = 2;
+  let tries: number = 0;
+  let newSubAccount: SubAccountDetails | undefined;
+  let error: Error | undefined;
+  // TODO: Remove and understand L2-301
+  while (tries < MAX_TRIES && newSubAccount === undefined) {
+    tries += 1;
+    try {
+      newSubAccount = await nnsDapp.createSubAccount({
```
mstrasinskis, line 75: `error = undefined;` here? To avoid `throw error`on line 88. Or should it throw anyway?

frontend/svelte/src/routes/Accounts.svelte
```diff
@@ -64,12 +67,15 @@
...
           >
-          <button class="primary" on:click={addAccount}
+          <button class="primary" on:click={openAddAccountModal}
             >{$i18n.accounts.add_account}</button
           >
         </Toolbar>
       {/if}
     </svelte:fragment>
+    {#if showAddAccountModal}
+      <AddAcountModal visible={showAddAccountModal} on:nnsClose={closeModal} />
```
mstrasinskis, line 77: Will it work if `modal.visible` is true by default?

mstrasinskis (commented):
👍 Just a couple of questions.

frontend/svelte/src/lib/services/accounts.services.ts
lmuntanerDfinity: oh yes!

frontend/svelte/src/routes/Accounts.svelte
lmuntanerDfinity: mhm, in that case I could avoid the `visible={...}`, chall we change that?

frontend/svelte/src/lib/services/accounts.services.ts
mstrasinskis: 🤞

frontend/svelte/src/routes/Accounts.svelte
mstrasinskis: Could you please test that this wouldn't break the animation?

frontend/svelte/src/lib/services/accounts.services.ts
lmuntanerDfinity: It would be "not instance" do you think it's nicer? `!(currentError instanceof AccountNotFound)`?

frontend/svelte/src/routes/Accounts.svelte
lmuntanerDfinity: animation to appear looks good, to go out not so much. Any suggestion?

lmuntanerDfinity:
@peterpeterparker ready for review

frontend/svelte/src/lib/services/accounts.services.ts
```diff
@@ -47,5 +53,40 @@ const loadAccounts = async ({
...
+    agent: await createAgent({ identity, host: identityServiceURL }),
+    canisterId: OWN_CANISTER_ID,
+  });
+
+  const MAX_TRIES = 2;
+  let tries: number = 0;
+  let newSubAccount: SubAccountDetails | undefined;
+  let error: Error | undefined;
+  // TODO: Remove and understand L2-301
+  while (tries < MAX_TRIES && newSubAccount === undefined) {
```
peterpeterparker, line 72: can we rather comment the all block until it is clear what and why and how?

I am not comfortable merging that `while` in main branch.

frontend/svelte/src/lib/stores/accounts.store.ts
```diff
@@ -3,11 +3,29 @@ import type { Account } from "../types/account";
...
  * A store that contains the account information.
  */
-export const accountsStore = writable<AccountsStore>({
-  main: undefined,
-});
+
+export const initAccountsStore = () => {
+  const { subscribe, set, update } = writable<AccountsStore>({
+    main: undefined,
+    subAccounts: [],
```
peterpeterparker, line 16: Should be the all object of accounts | undefined? Something like

```
const { subscribe, set, update } = writable<Accounts | undefined>(undefined)
```

feels like an empty array of sub-accounts and a main account undefined does not match well together

frontend/svelte/src/lib/modals/AddAccountModal/SelectTypeAccount.svelte
```diff
@@ -0,0 +1,56 @@
+<script lang="ts">
```
peterpeterparker, line 1: no tests for SelectTypAccount?

frontend/svelte/src/lib/modals/AddAccountModal/AddNewAccount.svelte
```diff
@@ -0,0 +1,84 @@
+<script lang="ts">
```
peterpeterparker, line 1: no test for AddNewAccount?

frontend/svelte/src/lib/stores/accounts.store.ts
```diff
@@ -3,11 +3,29 @@ import type { Account } from "../types/account";
...
+    addSubAccount(newAccount) {
+      update((state) => ({
+        ...state,
+        subAccounts: [newAccount, ...state.subAccounts],
+      }));
+    },
+  };
+};
+
+export const accountsStore = initAccountsStore();
```
peterpeterparker, line 31: some tests?

peterpeterparker (changes requested):

frontend/svelte/src/lib/services/accounts.services.ts
mstrasinskis: If only this helps to get rid of `kind` property.

frontend/svelte/src/lib/stores/accounts.store.ts
mstrasinskis: LGTM So main is undefined and subAccounts[0..n] are undefined

frontend/svelte/env.config.mjs
```diff
@@ -31,8 +31,12 @@ const IDENTITY_SERVICE_URL =
 const HOST =
   process.env.HOST || (development ? `https://${domainTestnet}/` : "");
 
+// Testnet Canister Ids
```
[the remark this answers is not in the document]
lmuntanerDfinity: ah, I didn't know. I will update the comment.

frontend/svelte/src/lib/modals/AddAccountModal/AddAccountModal.svelte
```diff
@@ -0,0 +1,57 @@
...
+      <Transition {diff}>
+        <AddNewAccount on:nnsClose />
+      </Transition>
+    {/if}
+  </article>
+</Modal>
+
+<style lang="scss">
+  article {
+    height: 500px;
```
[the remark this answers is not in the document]
lmuntanerDfinity: the user won't be able to see it all.

I think this modal height could be handled in general as well.

Maybe in https://dfinity.atlassian.net/browse/L2-302 as well?

frontend/svelte/src/lib/modals/AddAccountModal/AddNewAccount.svelte
lmuntanerDfinity: I consider the components `AddNewAccount` and `SelectTypeAccount` as implementation details.

I test them in the `AddAccountModal.spec.ts`. In there I test both of them.

frontend/svelte/src/lib/services/accounts.services.ts
lmuntanerDfinity: sure

frontend/svelte/src/routes/Accounts.svelte
lmuntanerDfinity: it's not because of the `visible` but because of the `{#if `. Which I prefer to avoid having components rendered that do nothing.

frontend/svelte/src/lib/modals/Modal.svelte
```diff
@@ -6,7 +6,7 @@
   import IconBackIosNew from "../icons/IconBackIosNew.svelte";
   import { i18n } from "../stores/i18n";
 
-  export let visible: boolean = false;
+  export let visible: boolean = true;
```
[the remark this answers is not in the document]
peterpeterparker: alright. I'm surprised it pass without crashing a test or there weren't one. if none, can you add a small one for it?

frontend/svelte/src/lib/stores/accounts.store.ts
```diff
@@ -2,12 +2,30 @@ import { writable } from "svelte/store";
...
 
 /**
  * A store that contains the account information.
  */
-export const accountsStore = writable<AccountsStore>({
-  main: undefined,
-});
+
+export const initAccountsStore = () => {
+  const { subscribe, set, update } = writable<AccountsStore>({
```
peterpeterparker, line 14: I am not sure about the store definition. With this form we can potentially ends up in a state where `subAccounts` are loaded but the `mainAccount` isn't or the contrary. Means we potentially might display partial information and make functions available that should not, like creating a first sub account even though they would exists and stuffs like that

That's why I ask myself if it would not be better to have a store "all or nothing" - i.e. either the `mainAccount` and the `subAccounts` are loaded or nothing .

These accounts are linked money - user ICPs, therefore the precaution.

What do you think?

Something like `writable<Accounts | undefined>(`

peterpeterparker (changes requested):

frontend/svelte/src/lib/stores/accounts.store.ts
lmuntanerDfinity: So you propose that we cannot just add a `subAccount`, either set all the store or nothing else?

peterpeterparker: Yes that's my suggestion / question

lmuntanerDfinity: I created an issue to remove the add_account call, the one that its commented, I will start when I merge this.

https://dfinity.atlassian.net/browse/L2-301

What I could do is to always fetch main and subaccounts and set the store.

Even after creating a subaccount, we could make the call to fetch all new. This way, it's not only the store that could set partial information, but that we ensure that when fetching, we always fetch it together.

Otherwise, even though we use set when creating a new subaccount, we are relying that the data is there, but we cannot be sure about it as you say. If we want to be safe, it's not just the store, but also the fetch that we should to together.

Does that make sense? I can add it to the issue.

peterpeterparker: > What I could do is to always fetch main and subaccounts and set the store.
> 
> Even after creating a subaccount, we could make the call to fetch all new. This way, it's not only the store that could set partial information, but that we ensure that when fetching, we always fetch it together.

This ain't a guarantee that both values are set at once or not. It would programmatically amend that both are set at the same time in your code but does not gives the guarantee that these are always set at once. Likewise, in x months or somewhere else in the code, someone might add a piece of code that desynchronize both values.

It's first a general question first, do we want:

a. set all the store (main + sub) in one single block or nothing else - i.e. a store that contains all accounts or nothing
b. set either main and subaccount - i.e. a store that has either a main account only, either subaccounts only or both or nothing

For the reason I listed previously I feel it's rather more secure to have a.

But, it's a question, what do you @lmuntanerDfinity and @mstrasinskis think? a. or b.?

lmuntanerDfinity: I don't have a strong opinion. "a" is safer but less performant. If we prefer safety than UX, then we should go for "a".

We should establish the priorities of the application as a whole to have consistent patterns. I don't think that we should be talking only about account and subaccounts here. There will be other places where we will need to choose between performance and security.

lmuntanerDfinity: added comments, this task will be done in https://dfinity.atlassian.net/browse/L2-301

peterpeterparker (approved):
LGTM

About `main` and `subAccounts`, we, @lmuntanerDfinity @mstrasinskis and I, have decided in a chat to go with solution `a.` for security reason ("all or nothing - i.e. `main` and `sub` accounts are fetched or nothing is added to store"). 

In addition to the store API itself that will expose only a setter (no update), we also decided to secure this choice that we will synchronize "in block" the accounts information not only at boot time but also after any operation on the account - e.g. creating a sub-account.

This will be done in upcoming PRs.

PR #423, opened 2022-02-10
L2 218: Create Neuron UI

lmuntanerDfinity:
# Motivation

Develop the UI components to create a neuron.

# Changes

* New components in `modals/CreateNeuronModal`.
* Expanded `Input` to support dark theme.
* Expanded `button` styles to support full-width.
* Expanded Identifier to hide button to copy to the clipboard.

# Tests

* New test in route Neurons to check opening the modal
* Test for `CreateNeuronModal`. The inner components are implementation details. Tested here.

PR #448, opened 2022-02-16
feat: replace input fullWidth property with CSS variable

peterpeterparker:
# Motivation

Prefer CSS variables over properties for styling purpose.

Exception: block of styles such as applying a all theme or color - i.g. options that affect multiples styles.

# Changes

- remove `fullWidth`
- add CSS variable

# Note

Follow-up of last week PR [Create Neuron UI](https://github.com/dfinity/nns-dapp/pull/423). Works for you @lmuntanerDfinity?

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Input.svelte
+++ b/frontend/svelte/src/lib/components/ui/Input.svelte
@@ -11,8 +11,6 @@
 
   export let placeholderLabelKey: string;
 
-  // TODO: Review LM and DDB
-  export let fullWidth: boolean = false;
   export let theme: "dark" | "light" = "light";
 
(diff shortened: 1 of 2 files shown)


lmuntanerDfinity (approved):
Should we prefer this for all style rule props?

peterpeterparker:
@lmuntanerDfinity thanks for the review. as described in the details in my opinion I think yes indeed it works best but there exception like a `theme` or a `color` palette that applies to lots of variables.

lmuntanerDfinity:
Yes, I understand. that's why I said style rule, not just style. Same for "primary" or "secondary" if we wanted.

PR #452, opened 2022-02-16
refactor: accounts & proposals services

peterpeterparker:
# Motivation

Refactor accounts and proposals `utils` to `services` and refactor the loading of the accounts to follow our current parent ("do the work in a service then set the store").

# Changes

- move and rename `.utils.ts` to `services/*.services.ts`
- refactor `loadAccounts` to not initiate the loading within the store but to defer the work to service that sets ultimately the values in store
- refactor remaining functions of `proposals.utils` to not use imperatively the store - i.e. pass proposals values as param instead

# Tests

`.spec` tests updated accordingly.


mstrasinskis (approved):
LGTM 👍

PR #453, opened 2022-02-17
Display payloads for all ExecuteNNSFunction proposals

hpeebles:
WIP


peterpeterparker:
would it make sense to add the new feature directly in `nns-js` now instead of `frontend/ts`?
because of the rewrite, how do we keep up if things are added to `frontend/ts` - i.e. do we need these new options in `nns-js`?

just thinking out loud

hpeebles:
> would it make sense to add the new feature directly in `nns-js` now instead of `frontend/ts`? because of the rewrite, how do we keep up if things are added to `frontend/ts` - i.e. do we need these new options in `nns-js`?
> 
> just thinking out loud

The changes are mostly in the backend canister.
The change on the frontend is that we now call into the NNS Dapp canister to get proposal info rather than into the governance canister, we then deserialize the payload as JSON, rather than candid.
But this behaviour is specific to the NNS Dapp so I don't think it should also belong in nns-js.

rs/src/proposals.rs
```diff
@@ -0,0 +1,586 @@
...
+                result
+            }
+            Ok(None) => Err("Proposal not found".to_string()), // We shouldn't cache this as the proposal may simply not exist yet
+            Err(error) => Err(error), // We shouldn't cache this as the error may just be transient
+        }
+    }
+}
+
+// Check if the proposal has a payload, if yes, deserialize it then convert it to JSON.
+fn process_proposal(mut proposal_info: ProposalInfo) -> Result<ProposalInfo, String> {
```
lmuntanerDfinity, line 44: Silly question, why isn't this done in the governance canister?

hpeebles: The full payloads are serialized in the governance canister using candid.
Here, I deserialize the payloads, then transform them if needed, then once they are transformed I opted to serialize using JSON rather than candid simply so that the frontend would be able to render them without having to know the schema.

lmuntanerDfinity: Is that something that all the frontend projects using the governance canister will need?

hpeebles: If they want to show the payloads then yes they'll need to do essentially the same thing.
I have a call with some of the NNS guys later, I'll raise this point with them.

lmuntanerDfinity: Thanks for the response!

hpeebles:
I'm going to split this PR into smaller chunks

PR #454, opened 2022-02-17
Feat: Refactor Wizard

lmuntanerDfinity:
# Motivation

Refactor Wizard component to something simpler and enable multiple in the same page.

# Changes

* Removed `Wizard` and `Wizard` Components in favor of a `StateSteps` Class.
* New UI Component for transitions: `Transition`.

# Tests

* Tests for `StateSteps` class.


lmuntanerDfinity:

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
```diff
@@ -57,19 +69,19 @@
 >
   <span slot="title">{$i18n.neurons?.[titleKey]}</span>
   <main>
-    <Wizard>
-      <WizardStep index={0}>
+    {#if currentIndex === Steps.SelectAccount}
```
mstrasinskis, line 72: currentIndex -> currentStep: Steps

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
```diff
@@ -1,18 +1,26 @@
...
+  import {
+    initWizardStore,
+    State as WizardState,
+  } from "../../stores/wizard.store";
+  import Transition from "../../components/ui/Transition.svelte";
 
   export let visible: boolean;
 
+  const wizardStore = initWizardStore();
+  enum Steps {
```
peterpeterparker, line 19: I recently a hard time with `enum`, still got PTSD 🤣. if it is possible with `type` I would not be against.

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
```diff
@@ -1,18 +1,26 @@
...
-  import { wizardStore } from "../../components/ui/Wizard/wizardStore";
+  import {
+    initWizardStore,
+    State as WizardState,
+  } from "../../stores/wizard.store";
+  import Transition from "../../components/ui/Transition.svelte";
 
   export let visible: boolean;
 
+  const wizardStore = initWizardStore();
```
peterpeterparker, line 18: - we use the modal also without the wizard (filters for example), therefore not sure it is a good idea to initialize it every time for every cases

- is it safe to init the wizard in component related script?

peterpeterparker (changes requested):

frontend/svelte/src/lib/components/ui/Transition.svelte
```diff
@@ -0,0 +1,20 @@
+<script lang="ts">
+  import { fly } from "svelte/transition";
+
+  export let diff: number;
+  const DEFAULT_OFFSET = 200;
+  const ANIMATION_DURATION = 200;
+  let container: HTMLDivElement | undefined;
```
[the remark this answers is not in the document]
lmuntanerDfinity: `bind:this` does not set `null` after inspecting it. But I can't find any docs that support this statement.

frontend/svelte/.eslintrc.js
```diff
@@ -0,0 +1,41 @@
+module.exports = {
```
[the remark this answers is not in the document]
lmuntanerDfinity: No, this PR does not add any additional linter rules. I wanted to do the setup and then decide with the team which rules we want to enforce. I preferred not to make the decision myself.

`svelte-check` does not seem to provide functionality to check for consistency in how we use types. For example, ensuring that all the variables have a type instead of relying in implicit types.

The PR #326  seems to have been closed because you were looking for a more global way of linting (based on the comments). Not because `svelte-check` was a substitute.

I don't mind not adding eslint, but enforcing style rules without being able to check the automatically just calls for inconsistency.

frontend/svelte/src/lib/components/ui/Transition.svelte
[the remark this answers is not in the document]
peterpeterparker: interesting, thanks for having a look and for the feedback 👍

frontend/svelte/.eslintrc.js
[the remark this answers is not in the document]
peterpeterparker: Not against at all! Can you just then as I said above provide a specific PR and adds it to the CI build? Thanks

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
lmuntanerDfinity: I use an `enum` because I want to use 0 and 1 actually. The `enum` is just a more readable way to have 0 and 1.

Do you propose the following?

```
const Steps = {
  SelectAccount: 0,
  StakeNeuron: 1,
};
```

The, I would also need a type; otherwise I would be relying on implicit types. More lines. Do you prefer this? With or without type definition?

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
lmuntanerDfinity: This is what could enable us to have multiple Wizards in the same page each in a different step. If we initialize it in the component where it's used. Otherwise we are stuck with using one wizard per page.

I don't see any problem nor any article about not initializing stores inside a component.

I see this more as a way to reuse functionality. A class would also work, but it would not rerender. (I will try this)

frontend/svelte/.eslintrc.js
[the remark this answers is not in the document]
lmuntanerDfinity: Do you mean a PR just for the linter?

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
peterpeterparker: does the script needs to be declared as `module` then?
does the store needs to be destroyed when the component is destroyed or does it happens automatically (memory overflow...)?

it's cool if we can scope the store sure.

frontend/svelte/.eslintrc.js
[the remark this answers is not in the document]
lmuntanerDfinity: done! #456

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
lmuntanerDfinity: Well, once the component is dismounted, the store is not referenced by any other variable, so I guess that the garbage collector should do its work.

Actually, if that is true, I don't even need to reset the store `onDestroy`.

peterpeterparker: > I guess that the garbage collector should do its work.

Please analyze and be sure that the store is destroyed.

In addition, as I said in my [first](https://github.com/dfinity/nns-dapp/pull/454#discussion_r809167505) comment we do not have a wizard in every modal. Can you implement a solution for this as well?

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
peterpeterparker: alright then

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
lmuntanerDfinity: A store is just an object, if the other variables are also destroyed, I don't see why this one wouldn't.

I haven't found anywhere how to destroy stores. Yet, I found the "[Store Contract](https://svelte.dev/docs#component-format-script-4-prefix-stores-with-$-to-access-their-values-store-contract)", where we could create our own stores as long as we keep the API.

Therefore, we cas assume, that the same way we can create objects inside a component, we can also create stores.

peterpeterparker: - Modal name: gotcha, yes, super confusing. Please rename or comment with a TODO if renamed in another PR.

- Store: I'll check briefly Monday

frontend/svelte/src/lib/components/ui/Transition.svelte
```diff
@@ -0,0 +1,20 @@
...
+  let absolutOffset = 200;
+  let slideOffset: number | undefined;
+
+  $: absolutOffset = container?.clientWidth || DEFAULT_OFFSET;
```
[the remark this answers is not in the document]
lmuntanerDfinity: That's cool, but why would I do that? The width is already there, I use this for the transition. Not to change the width

lmuntanerDfinity:
@peterpeterparker @mstrasinskis 

I joined the Svelte Discord channel that they [suggest](https://github.com/sveltejs/svelte/issues/new/choose) in their Github page and asked the following question:

> question - is it a bad practice to create a store inside a component? what happens when the component is destroyed, is the store garbage collected?

There are thousands of users and two moderators. One of the moderators answered to the question and we started talking. Long story short, this was his final answer:

> yeah sure, you can make a mycustomstore.js file and export a function that returns a store from there
> it’s a fairly common pattern :slightly_smiling_face:
> so you would do in your component:
> import { storeCreator } from ‘storeCreator.js’;
> const mystore = storeCreator()
> no mystore only exists within the context of that instance and will be safely removed when the component is

What do you think?

frontend/svelte/src/lib/components/ui/Transition.svelte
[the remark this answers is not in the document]
mstrasinskis: These bindings are readonly — changing the values of width won't have any effect.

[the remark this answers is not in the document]
lmuntanerDfinity: aaahhh, you mean instead of using the container! Good one!

peterpeterparker:
@lmuntanerDfinity sounds good. as said previously, I still gonna have a brief look on Monday though 😉

frontend/svelte/src/lib/stores/StepsState.ts
```diff
@@ -0,0 +1,35 @@
+function enumSize<EnumType>(enm: EnumType): number {
+  return Object.values(enm).filter(isNaN).length;
+}
+
+export class StepsState<EnumType> {
```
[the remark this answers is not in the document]
lmuntanerDfinity: Something that could be worth considering, is that the "class pattern" will only work if the "functionality + state" is only used in one component.

Whereas the "store pattern" could be used in many components at the same time, hence the pattern. That means that in case we wanted this functionality to apply to multiple components, we would need to find a way to do so in other ways, such as exposing methods from the component or passing props.

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
```diff
@@ -1,18 +1,24 @@
...
+  import Transition from "../../components/ui/Transition.svelte";
+  import { StepsState } from "../../services/stepsState.services";
 
   export let visible: boolean;
 
+  enum Steps {
+    SelectAccount,
+    StakeNeuron,
+  }
+  let stateStep = new StepsState(Steps);
```
peterpeterparker, line 20: When we will discuss eslint, I am not against explicit types for object `let stateStep: StepsState = ...`

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
```diff
@@ -21,55 +27,46 @@
...
-  });
+  onDestroy(unsubscribeAccounts);
 
   const titleMapper: Record<string, string> = {
     "0": "select_source",
     "1": "stake_neuron",
   };
   let titleKey: string = titleMapper[0];
-  $: titleKey = titleMapper[currentIndex];
+  $: titleKey = titleMapper[stateStep.currentStep];
```
peterpeterparker, line 44: not sure but you can maybe destruct `stateStep`

```
let currentStep: number;
let diff: number;
$: ({currentStep, diff}) = stateStep

$: titleKey = titleMapper[currentStep];

...

showBackButton={currentStep === Steps.StakeNeuron}

....

{#if currentStep === Steps.SelectAccount}

...

<Transition diff={stateStep.diff}>
```

frontend/svelte/src/lib/services/stepsState.services.ts
```diff
@@ -0,0 +1,33 @@
+function enumSize<EnumType>(enm: EnumType): number {
```
peterpeterparker, line 1: you can move the utils to `enum.utils`

frontend/svelte/src/lib/services/stepsState.services.ts
```diff
@@ -0,0 +1,33 @@
...
+  return Object.values(enm).filter(isNaN).length;
+}
+
+export class StepsState<EnumType> {
```
peterpeterparker, line 5: the class is not generic as far as I understand, so why `StepsState<EnumType>` instead of `StepsState`?

frontend/svelte/src/lib/services/stepsState.services.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  public next(): StepsState<EnumType> {
+    if (this.currentStep < enumSize(this.steps) - 1) {
+      this.previousStep = this.currentStep;
+      this.currentStep = this.currentStep + 1;
+    }
+    return this;
+  }
+
+  public get diff(): number {
```
peterpeterparker, line 22: I honestly did not understand what would returns `diff` and what was the goal of `diff`. can live with it - i.e. can be merged as it - but if you find something more meaningful, not against

frontend/svelte/src/lib/components/ui/Transition.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
```
peterpeterparker, line 1: we have test for the transition? if yes, should the related spec file be renamed too? if no, can we - should we add a test?

peterpeterparker (changes requested):
No biggie, just a couple of small questions

frontend/svelte/src/lib/services/stepsState.services.ts
lmuntanerDfinity: good point, I also didn't find a better name.

frontend/svelte/src/lib/services/stepsState.services.ts
lmuntanerDfinity: sure

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
lmuntanerDfinity: looks good

frontend/svelte/src/lib/modals/CreateNeuronModal/Modal.svelte
lmuntanerDfinity: ok

frontend/svelte/src/lib/components/ui/Transition.svelte
peterpeterparker: alright I guess we can rely on the e2e tests for this

lmuntanerDfinity: yes, also the modal tests check that the new component is rendered, not that there is a transition, but that the transition doesn't mess it up.

frontend/svelte/src/lib/services/stepsState.services.ts
peterpeterparker: it is confusing because you do not specify the enum type when you instantiate the class in your component `new StepsState(Steps);`. can you please add the type?

frontend/svelte/src/lib/utils/enum.utils.ts
```diff
@@ -15,3 +15,7 @@ export const enumsKeys = <T>({
 }): string[] => {
   return values.map((value: T) => obj[value as unknown as string]);
 };
+
+export const enumSize = <EnumType>(enm: EnumType): number => {
```
peterpeterparker, line 19: cool, now we can even add a small test 😉

frontend/svelte/src/lib/services/stepsState.services.ts
peterpeterparker: only think I can think of is something like `direction` that return `forward` or `backward` but still not optional. maybe someday we will find a better naming

peterpeterparker (changes requested):

frontend/svelte/src/lib/utils/enum.utils.ts
lmuntanerDfinity: done

mstrasinskis (approved):
LGTM. Thanks for the state implementation 👍

frontend/svelte/src/lib/services/stepsState.services.ts
lmuntanerDfinity: I have it in the assignment
```
let stateStep: StepsState<typeof Steps> = new StepsState(Steps);
```

lmuntanerDfinity: done

peterpeterparker (approved):
Thx

PR #457, opened 2022-02-18
Feat: Canister Home Messaging

lmuntanerDfinity:
# Motivation

Add content in Canisters Home

# Changes

* Add messages for Canisters Home.
* Increase margin in messages in Neurons Home


lmuntanerDfinity:
@mstrasinskis @peterpeterparker a super simple one.

Here are the screenshots:

Screenshot 2022-02-18 at 13 50 09

Screenshot 2022-02-18 at 13 50 16

frontend/svelte/src/routes/Canisters.svelte
```diff
@@ -19,6 +20,15 @@
...
     <section>
       <h1>{$i18n.canisters.title}</h1>
+      <p>{$i18n.canisters.text}</p>
+      <ul>
+        <li>{$i18n.canisters.step1}</li>
+        <li>{$i18n.canisters.step2}</li>
+        <li>{$i18n.canisters.step3}</li>
+      </ul>
+      <p>
+        {$i18n.canisters.principal_is} "{$authStore.principal}"
```
peterpeterparker, line 30: `"{$authStore.principal}"` is automatically casted to string?

in `Neurons.svelte` we explicitly use `principal?.toText()`

lmuntanerDfinity: Good point, I didn't see it. It was working but it's better to change it.

lmuntanerDfinity:

peterpeterparker (changes requested):
Cool, looks good.

Just one thing, can you add a test?

Probably worth testing that the `principal.toText` is rendered

lmuntanerDfinity:
@peterpeterparker @mstrasinskis ready for review with a test.

Interesting to note that even though I didn't explicitly use the `subscribe` method, instead used the `$` to access the store value, I still had to mock the subscribe method. Otherwise, the value in the store was `undefined`.

peterpeterparker (approved):
Thanks for the PR

PR #461, opened 2022-02-21
feat: add identity, host and fetch root key to create agent 

peterpeterparker:
# Motivation

Canisters calls shall be identified with user' identity.
Host and fetch root key are useful to develop locally and use testnet.

# Changes

- refactor `authStore` to contain user' `Identity` instead of `Principal` ("one level of information higher)
- rename constant `serviceURL` to `identityServiceURL`
- add `Identity`, optional `host` and `fetchRootKey` to agent utils
- `createAgent` is now promise based
- use `Identity` and `identityServiceURL` in `accounts.services` to fetch account balance
- use `Identity` and `host` in `proposals.services` to fetch the list of proposals (was only done in prod, now support testnet)
- globally mock `createAgent` for all tests
- extend rollup config to build app with `HOST` and `FETCH_ROOT_KEY` information

### DIFF
--- a/frontend/svelte/jest-spy.ts
+++ b/frontend/svelte/jest-spy.ts
@@ -1,4 +1,4 @@
 import * as agent from "./src/lib/utils/agent.utils";
 
-const mockCreateAgent = () => undefined;
+const mockCreateAgent = () => Promise.resolve(undefined);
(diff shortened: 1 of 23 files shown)


lmuntanerDfinity (approved):
LGTM

I don't understand the difference between Principal and Identity yet, nor why do we need to `fetchRootKey`. But apart from that, let's merge this.

PR #462, opened 2022-02-21
feat: list proposals exclude topics

peterpeterparker:
# Motivation

Apply "Topics" filter to list proposals

# Changes

- add `excludeTopics` utils
- use filter to list proposals


frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -9,3 +11,14 @@ export const lastProposalId = (
   const { length, [length - 1]: last } = proposalInfos;
   return last?.id;
 };
+
+export const excludeTopics = (topics: Topic[]): Topic[] => {
```
mstrasinskis, line 15: Hmm... is it possible to move this logic in enum.utils?

mstrasinskis (approved):
LGTM

frontend/svelte/src/lib/utils/proposals.utils.ts
peterpeterparker: Good idea. Utils extracted and as discussed, I have added some comments for the existing utils regarding the "number-based enums".

mstrasinskis (approved):
LGTM. Thanks for the extra comments.

PR #468, opened 2022-02-22
feach: filter - exclude voted proposals

peterpeterparker:
# Motivation

Implement filter "Hide "Open" proposals where all your neurons have voted or are ineligible to vote" to filter the list or proposals.

# Changes

- add `excludeVotedProposals` to filter store
- toggle state on checkbox change
- hide matching proposal on the client side (see *note below)

# Note

Such a filter option, `excludeVotedProposals`, does not exists in the governance canister (see [candid](https://github.com/dfinity/nns-js/blob/2f849e2227000e08eb7e0ba121e7b6da065ebd80/candid/governance.did#L141) options to list proposals).

In the Flutter app, the filtering is applied on the client side once all the data have been fetched (as far as I understand - see piece of code in [governance tab](https://github.com/dfinity/nns-dapp/blob/db533f2a0450c9c5f394b59d5d9eaa10b2f3b195/frontend/dart/lib/ui/proposals/governance_tab_widget.dart#L212)).

```
[Proposal.A, Proposal.B, Proposal.NOT_MATCHING, Proposal.D] => [A, B, C]
```

Because the list of proposals can / shall be browsed - ux-wise - with an infinite scrolling feature, the proper solution would be to extend the canister to apply the filter in advance - on the queries.

Since we do a rewrite and I want to push forward, I decided to follow the existing Flutter path and also implement a hack that replicate the filtering on the client side.

However, the PR does not filter the results and render a `div` container for any proposals but hide - does not render - the content of the not matching proposals in the DOM. Doing so, the infinite scroll feature that is based on the intersection observer is still able to detect when the next batch of proposals shall be fetched.

```
div
   card
div
   card
div
   card
div // <--- not matching **
div
   card
div 
   card
   
** not matching the filter -> no card render but still observable by the intersection observer -> still compatible with the infinite scrolling   
```


frontend/svelte/src/lib/components/proposals/ProposalCard.svelte
```diff
@@ -24,14 +26,28 @@
   };
 
   $: color = colors[status];
+
+  // HACK: the governance canister does not implement a filter to hide proposals where all neurons have voted or are ineligible.
+  // That's why we hide these proposals on the client side only.
+  // We do not filter these types of proposals from the list but "only" hide these because removing them from the list is not compatible with an infinite scroll feature.
+  let hide: boolean;
```
mstrasinskis, line 33: nice 👍

mstrasinskis (approved):
LGTM

PR #471, opened 2022-02-23
feat: list proposal card display

peterpeterparker:
# Motivation

Style the cards in the list of proposal as in the Flutter app and add information `proposer` and `id`.

# Changes

- display `proposer` and `id`
- clamp title style
- style `Card` to display both slots `start` and `end` `flex` even on mobile
- add spacing around spinner

# Screenshots

Capture d’écran 2022-02-23 à 09 01 01
Capture d’écran 2022-02-23 à 09 01 05
Capture d’écran 2022-02-23 à 09 01 08


frontend/svelte/src/lib/components/proposals/ProposalCard.svelte
```diff
@@ -37,17 +42,26 @@
...
 
-<!-- TODO(L2-206): display all proposal information as in production -->
-<!-- TODO(L2-206): implement missing css styles - design -->
-
 <!-- We hide the card but keep an element in DOM to preserve the infinite scroll feature -->
 <div>
   {#if !hide}
     <Card>
-      <p slot="start">{proposal?.title}</p>
+      <p slot="start" class="title">{proposal?.title}</p>
```
mstrasinskis, line 49: How about `title={proposal?.title}`?

frontend/svelte/src/lib/components/proposals/ProposalCard.svelte
```diff
@@ -37,17 +42,26 @@
...
+      <p><small>Id: {id || ""}</small></p>
     </Card>
   {/if}
 </div>
+
+<style lang="scss">
+  @use "../../themes/mixins/text";
+
+  .title {
+    @include text.clamp(3);
```
mstrasinskis, line 64: Good idea, will do the same for the proposal detail instead of `overflow-wrap: anywhere;`  👍

mstrasinskis (changes requested):
Looks great, only one suggestion

frontend/svelte/src/lib/components/proposals/ProposalCard.svelte
peterpeterparker: Good idea. Done. Also destructed the `title` since I was already there ;)

mstrasinskis (approved):
Thanks for update! LGTM.

PR #473, opened 2022-02-23
feat: list proposals catch and display errors

peterpeterparker:
# Motivation

Catch and display errors for list proposals

# Changes

- try catch toasts

# Screenshots

Capture d’écran 2022-02-23 à 10 52 47


lmuntanerDfinity (approved):
LGTML

As a side comment, do we want to use some kind of error monitoring tool like Rollbar?

PR #488, opened 2022-02-25
fix: meta lang and charset

peterpeterparker:
# Motivation

Lighthouse reports an error about the charset and the lang is missing.

> Properly defines charset Error!
> A character encoding declaration is required. It can be done with a `<meta>` tag in the first 1024 bytes of the HTML or in the Content-Type HTTP response header. [Learn more](https://web.dev/charset/?utm_source=lighthouse&utm_medium=devtools).

# Changes

- fix charset
- add missing lang

# Error

Capture d’écran 2022-02-25 à 08 41 13


peterpeterparker:
Not enough, we probably need to try to add the charset in the response

> Add a Content-Type HTTP response header [#](https://web.dev/charset/?utm_source=lighthouse&utm_medium=devtools#http)
> 
> Configure your server to add a [Content-Type](https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Type) HTTP response header that includes a charset directive.
> 
> Content-Type: text/html; charset=UTF-8

peterpeterparker:
It doesn't solve the issue but it doesn't arm to set correctly charset and lang

mstrasinskis (approved):
LGTM, very positive changes 👍

PR #492, opened 2022-02-25
Fix Rust <-> candid mismatch

hpeebles:
# Motivation

This fixes the display of 'Send' transactions, which are currently failing due to a mismatch between the Rust types and the candid.

# Changes

- Reverts back to using TransferResult::Send rather than Transfer (the transaction itself is of type Transfer, but relative to the user who sent it, it is a 'Send').
- Updates the candid with the new TransactionType::Transfer variant.

# Tests

I have deployed this version to the link below and have made transactions of type 'Send', 'Receive', 'Stake Neuron' and 'Top-up Neuron', and can see these transactions all rendering correctly - https://wckdt-raaaa-aaaaa-aaatq-cai.nnsdapp.dfinity.network/
You can view the transactions by using II anchor 10003.

Creating / topping up canisters has not been tested since the cycles conversion rate is not initialized on the testnet.


peterpeterparker:
Can you also update the .did file in nns-dapp svelte? 

We should discuss how we keep the did file in sync in both flutter and svelte apps.

lmuntanerDfinity:
@hpeebles is this why we were getting alerts in some actions even though they were working?

bitdivine (approved):

hpeebles:
> @hpeebles is this why we were getting alerts in some actions even though they were working?

We spoke about this offline.

PR #495, opened 2022-02-27
Screenshot the landing page not the service worker

bitdivine:
# Motivation
The landing page test actually waits for any page load, which can be the service worker.  This happens fairly consistently if run against the testnet rather than a local toy but it depends on how fast the service worker loads.  Here is how to run the test against a real network:

```
cd e2etests && npm ci
then
NNS_DAPP_URL="https://${canister_id}.nnsdapp.dfinity.network" npm run wdio
or
NNS_DAPP_URL="https://nns.ic0.app" npm run wdio
```

# Changes
Wait for "page ready" state where the page is not the service worker loading page.

# Tests
I have run this against the testnet.  The screenshot was of the service worker loading page and is now of the landing page whereas previously it was mostly the service worker loading page.


e2e-tests/specs/example.e2e.ts
```diff
@@ -1,7 +1,19 @@
+/**
+ * Waits for a page to load.  Normally this is simple, but
+ * we have a service worker that can load completely before
+ * the actual page loads.
+ */
+function waitForLoad(browser) {
+      return browser.waitUntil(() => browser.execute(() => (document.readyState === 'complete') && (null === document.title.match(/Content Validation Bootstrap/i))), {timeout: 60_000});
```
lmuntanerDfinity, line 7: Isn't the title `'Network Nervous System'`??

Is this to make sure the Service Worker is loaded?

bitdivine: This is to make sure that the page is NOT the service worker; it puts no constraint on the final page title.  This problem will apply to any page that uses the service worker so I made the function target the service worker title rather than the NNS title.  This way the function can be re-used in any similar situation.

lmuntanerDfinity: Then maybe rename to `waitForServiceWorker`?

nmattia (approved):

PR #496, opened 2022-02-28
feat: reset and find proposals on route mount

peterpeterparker:
# Motivation

In a first version we want to mimic the Flutter app behavior by triggering a new search of the proposals each time the page "Voting" is opened.

# Changes

- trigger find and reset proposals store on mount
- reset proposals filter on mount

# Note

We keep the proposal store as it allows a smooth transition between list and detail page.


mstrasinskis (approved):
LGTM

PR #499, opened 2022-02-28
chore: deprecate unused FiltersCard

peterpeterparker:
# Motivation

We don't use anymore, and as far as I can tell we will not use, the component `FiltersCard`.

# Changes

- remove `FiltersCard`


lmuntanerDfinity (approved):
Always in favor of deleting code!

PR #500, opened 2022-02-28
feat: display text if no proposals found

peterpeterparker:
# Motivation

Display a hint to the user if no matching proposals are found or should be displayed.

# Changes

- detect if no proposals where fetched or if none matching proposals have to be rendered
- display a msg "No proposals found for the filters." if nothing match

# Screenshot

Capture d’écran 2022-02-28 à 11 22 21


mstrasinskis (approved):
LGTM 👍

frontend/svelte/src/routes/Proposals.svelte
```diff
@@ -84,13 +89,24 @@
...
     debounceFindProposals?.()
   );
 
   onDestroy(unsubscribe);
+
+  let nothingFound: boolean;
+  $: nothingFound =
+    initialized &&
+    !loading &&
+    !hasMatchingProposals({
```
lmuntanerDfinity, line 106: How about having a derived store that combines `$proposalsStore` and `$proposalsFiltersStore`? It could return an empty array when no matching proposals. And `null` or `undefined` when still searching?

peterpeterparker: We need elements in the DOM. https://github.com/dfinity/nns-dapp/pull/468

lmuntanerDfinity: Ah, ok, not sure I understand why that is incompatible with the Infinite Scroll. But I will look into it.

PR #501, opened 2022-02-28
feat: list proposals - "filter" modal footer button

peterpeterparker:
# Motivation

To improve the UX, we are not going to fetch new proposals every time a checkbox is clicked but rather make it as a user choice. To do so, we add a "filter" button in the filter modals for the list of proposals.

# Changes

- add button "Filter"
- apply filtering only if button is clicked
- add `footer` slot to modal
- improve `content` slot scrolling in modal

# Screenshots

Capture d’écran 2022-02-28 à 14 00 31

Capture d’écran 2022-02-28 à 14 00 40

### DIFF
--- a/frontend/svelte/src/lib/modals/Modal.svelte
+++ b/frontend/svelte/src/lib/modals/Modal.svelte
@@ -49,6 +49,8 @@
       <div class="content" id="modalContent">
         <slot />
       </div>
+
+      <slot name="footer" />
     </div>
   </div>
 {/if}
@@ -159,7 +161,12 @@
   }
 
   .content {
+    display: flex;
+    flex-direction: column;
+
+    max-height: calc(100vh - 156px);
     overflow-y: scroll;
+
     color: var(--gray-800);
   }
 </style>
--- a/frontend/svelte/src/lib/modals/ProposalsFilterModal.svelte
+++ b/frontend/svelte/src/lib/modals/ProposalsFilterModal.svelte
@@ -61,10 +61,13 @@
     }
   };
 
-  const onChange = (filter: Topic | ProposalRewardStatus | ProposalStatus) => {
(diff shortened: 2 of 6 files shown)


frontend/svelte/src/lib/modals/ProposalsFilterModal.svelte
```diff
@@ -81,4 +84,16 @@
       >
     {/each}
   {/if}
+
+  <svelte:fragment slot="footer">
+    <button class="primary" type="button" on:click={filter}>
```
lmuntanerDfinity, line 89: can't you put `slot="footer"` to the button directly?

frontend/svelte/src/lib/modals/Modal.svelte
```diff
@@ -159,7 +161,12 @@
   }
 
   .content {
+    display: flex;
+    flex-direction: column;
+
+    max-height: calc(100vh - 156px);
```
lmuntanerDfinity, line 167: This might change depending on how to solve it here: https://dfinity.atlassian.net/browse/L2-302

Does it make sense to add a TODO and issue number?

peterpeterparker: Actually solve L2-203

lmuntanerDfinity: Does this also fix the AddAccount and CreateNeuron modals?

peterpeterparker: Not sure if it fix but it does not screw these for sure, I had a quick look. It might thought yes if the slot footer is used for the actions.

Also feels that the weird effect on back might be improved since I modified the content as flex instead of block.

peterpeterparker: Definitely

mstrasinskis (approved):
LGTM

PR #503, opened 2022-03-01
feat: testnet banner

peterpeterparker:
# Motivation

Some users find their way to the application deployed on testnet (source [forum](https://forum.dfinity.org/t/internet-identity-doubt-in-creating-a-new-ii-anchor/11179/8?u=peterparker)). To make them aware they are accessing a version that has only test purpose, we can display a banner about it.

# Changes

- test purpose banner displayed if environment is "testnet" and not "local development"
- css variable `--header-offset` for layout

# Screenshots

Capture d’écran 2022-03-01 à 08 01 15
Capture d’écran 2022-03-01 à 08 01 38
Capture d’écran 2022-03-01 à 08 01 43

### DIFF
--- a/frontend/svelte/src/lib/components/common/Banner.svelte
+++ b/frontend/svelte/src/lib/components/common/Banner.svelte
@@ -51,11 +51,11 @@
     left: 0;
     right: 0;
 
-    margin: 0;
-
     &.headless {
       position: relative;
     }
+
+    z-index: var(--z-index);
 
     height: var(--header-offset);
 
@@ -77,9 +77,13 @@
     font-weight: 400;
     @include text.clamp(2);
 
+    margin: 0;
+
     grid-column-start: 2;
 
     text-align: center;
+
+    line-height: inherit;
   }
 
   button {
--- a/frontend/svelte/src/lib/components/common/Banner.svelte
+++ b/frontend/svelte/src/lib/components/common/Banner.svelte
@@ -3,11 +3,14 @@
   import IconClose from "../../icons/IconClose.svelte";
 
   export let headless: boolean = false;
+
   let visible: boolean = JSON.parse(
     localStorage.getItem("nnsdapp-testnet-banner-display") || "true"
   );
 
-  const deployEnv: string = process.env.DEPLOY_ENV;
+  const testnet: boolean = process.env.DEPLOY_ENV === 'testnet';
+  const localEnv: boolean = JSON.parse(process.env.ROLLUP_WATCH);
+  const banner: boolean = testnet && !localEnv;
 
   let rootStyle: string | undefined;
 
@@ -29,12 +32,12 @@
 </script>
 
 <svelte:head>
-  {#if deployEnv === "testnet" && rootStyle}
+  {#if banner && rootStyle}
     {@html rootStyle}
   {/if}
 </svelte:head>
 
-{#if deployEnv === "testnet" && visible}
+{#if banner && visible}
   <div class:headless>
     <h4>For <strong>test</strong> purpose only.</h4>
     <button on:click={close} aria-label={$i18n.core.close}><IconClose /></button
--- a/frontend/svelte/src/lib/components/common/Banner.svelte
+++ b/frontend/svelte/src/lib/components/common/Banner.svelte
@@ -1,0 +1,64 @@
+<script lang="ts">
+  export let headless: boolean = false;
+
+  const deployEnv: string = process.env.DEPLOY_ENV;
+
+  const rootStyle: string = `
+    <style>
+      :root {
+        --header-offset: 50px;
+      }
+    </style>
+  `;
+</script>
+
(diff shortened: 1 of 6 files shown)


frontend/svelte/src/lib/components/common/Banner.svelte
```diff
@@ -0,0 +1,97 @@
+<script lang="ts">
+  import { i18n } from "../../stores/i18n";
+  import IconClose from "../../icons/IconClose.svelte";
+
+  export let headless: boolean = false;
+
+  let visible: boolean = JSON.parse(
+    localStorage.getItem("nnsdapp-testnet-banner-display") || "true"
```
lmuntanerDfinity, line 8: maybe move the name of the storage key to a variable

frontend/svelte/src/lib/components/common/Banner.svelte
```diff
@@ -0,0 +1,97 @@
...
+
+  const close = () => {
+    visible = false;
+
+    localStorage.setItem("nnsdapp-testnet-banner-display", "false");
+  };
+</script>
+
+<svelte:head>
+  {#if banner && rootStyle}
```
lmuntanerDfinity, line 35: why not `banner && visible` and make the `rootStyle` always a string?

frontend/svelte/src/lib/components/common/Banner.svelte
```diff
@@ -0,0 +1,97 @@
...
+
+  let visible: boolean = JSON.parse(
+    localStorage.getItem("nnsdapp-testnet-banner-display") || "true"
+  );
+
+  const testnet: boolean = process.env.DEPLOY_ENV === "testnet";
+  const localEnv: boolean = JSON.parse(process.env.ROLLUP_WATCH || "false");
+  const banner: boolean = testnet && !localEnv;
+
+  let rootStyle: string | undefined;
```
lmuntanerDfinity, line 15: why not setting it already at the beginning?

```
const rootStyle = `...`;
```

frontend/svelte/src/lib/components/common/HeadlessLayout.svelte
```diff
@@ -5,10 +5,13 @@
   import Toasts from "../ui/Toasts.svelte";
   import Tooltip from "../ui/Tooltip.svelte";
   import Footer from "./Footer.svelte";
+  import Banner from "./Banner.svelte";
 
   const dispatch = createEventDispatcher();
 </script>
 
+<Banner headless={true} />
```
lmuntanerDfinity, line 13: Does `<Banner headless />` work?

frontend/svelte/src/lib/components/common/Banner.svelte
```diff
@@ -0,0 +1,97 @@
...
+<style lang="scss">
+  @use "../../themes/mixins/text";
+
+  div {
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+
+    &.headless {
```
lmuntanerDfinity, line 57: Why is this needed? Could you change something in `HeadlessLayout` to avoid the extra config on the banner?

lmuntanerDfinity (commented):
Looks neat! A few comments.

frontend/svelte/src/lib/components/common/Banner.svelte
peterpeterparker: ok

frontend/svelte/src/lib/components/common/Banner.svelte
peterpeterparker: works too ok

frontend/svelte/src/lib/components/common/Banner.svelte
peterpeterparker: solve with previous change

frontend/svelte/src/lib/components/common/Banner.svelte
peterpeterparker: That's the most straight forward solution because a css variable cannot be set in the layout itself since the the banner is a child of body. It would need to inject the variable on the body tag which ultimately is kind of uglier as a props.

frontend/svelte/src/lib/components/common/HeadlessLayout.svelte
peterpeterparker: Yes that would work too.

frontend/svelte/src/lib/components/common/Banner.svelte
lmuntanerDfinity: Yeah, I meant more changing the HeadlessLayout css so that this banner could also use `position: absolute;` but I see in the `HeadlessLayout` that the position needs to be relative to make the shadow visible. If it doesnot have a straightforward solution, this is extra prop is good.

lmuntanerDfinity (approved):

PR #508, opened 2022-03-01
feat: do not reload list of proposals on back

peterpeterparker:
# Motivation

To make the transition between "Proposal detail" and "Proposals list - Voting" pages smooth in case of "back" action, the proposals shall not be fetched again if already exist.

Scenario example: 

- Neurons > Voting > Fetch proposals > select Proposal > Detail > Back > No fetch - display existing proposals from store
- Proposal details > Fetch proposal > Back > Voting > Fetch proposals
- Neurons > Voting -> Fetch proposals  > Neurons > Voting > Fetch proposals

# Changes

- add `referrerPath` to route store
- do not reset proposals store on mount in `Proposals.svelte` if referrer is "Proposal detail" page and store not empty
- fix `navigate` and `replace` called at the same time on back in `ProposalDetail.svelte`

### DIFF
--- a/frontend/svelte/src/lib/stores/route.store.ts
+++ b/frontend/svelte/src/lib/stores/route.store.ts
@@ -18,7 +18,9 @@ export interface RouteStore {
  * The store contains following information:
  * - path: the application path - i.e. the page (see AppPath) that is displayed
  * - referrerPath: the page - i.e. the path - that linked to the current path. For example, if navigating from "Accounts" to "Voting" page, the referrerPath is `/#/accounts`. If the referrer is not the app, then is value is undefined.
- * - isKnownPath: tells if the path is a path that is supported by the app - e.g. useful to display a 404
+ * - isKnownPath: tells if the path is a path that is supported by the app
+ *  - currently is used to imitate the flutter dapp navigation (redirects the user to the first page "/accounts/")
+ *  - later could be used to display a 404
  *
  */
 const initRouteStore = () => {
--- a/frontend/svelte/src/lib/stores/route.store.ts
+++ b/frontend/svelte/src/lib/stores/route.store.ts
@@ -5,6 +5,7 @@ import { pushHistory, replaceHistory, routePath } from "../utils/route.utils";
 
 export interface RouteStore {
   path: AppPath | string;
+  previousPath?: AppPath | string;
   isKnownPath: boolean;
 }
 
@@ -27,6 +28,7 @@ const initRouteStore = () => {
       update((state: RouteStore) => ({
         ...state,
         path,
+        previousPath: state.path,
         isKnownPath: isAppPath(path),
       })),
 
@@ -34,6 +36,7 @@ const initRouteStore = () => {
       update((state: RouteStore) => ({
         ...state,
         path,
+        previousPath: state.path,
         isKnownPath: isAppPath(path),
       }));
 
@@ -44,6 +47,7 @@ const initRouteStore = () => {
       update((state: RouteStore) => ({
         ...state,
         path,
+        previousPath: state.path,
         isKnownPath: isAppPath(path),
       }));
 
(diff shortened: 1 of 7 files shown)


frontend/svelte/src/lib/stores/route.store.ts
```diff
@@ -13,6 +14,12 @@ export interface RouteStore {
  * - update: update the state value
  * - navigate: update the state value and push the new route to the browser history i.e. update browser url with an entry in the navigation stack
  * - replace: update the state value and replace the route in the browser history i.e. update browser url without modifying the navigation stack
+ *
+ * The store contains following information:
+ * - path: the application path - i.e. the page (see AppPath) that is displayed
+ * - referrerPath: the page - i.e. the path - that linked to the current path. For example, if navigating from "Accounts" to "Voting" page, the referrerPath is `/#/accounts`. If the referrer is not the app, then is value is undefined.
+ * - isKnownPath: tells if the path is a path that is supported by the app - e.g. useful to display a 404
```
mstrasinskis, line 21: Just as reminder: to mock the flutter dapp we navigate to accounts/ on unknown paths.

frontend/svelte/src/lib/stores/route.store.ts
```diff
@@ -5,6 +5,7 @@ import { pushHistory, replaceHistory, routePath } from "../utils/route.utils";
 
 export interface RouteStore {
   path: AppPath | string;
+  referrerPath?: AppPath | string;
```
mstrasinskis, line 8: Does it makes sense to set the default value for referredPath to `document.referrer`?

frontend/svelte/src/routes/ProposalDetail.svelte
```diff
@@ -44,23 +38,28 @@
...
-
       console.error(error);
       toastsStore.show({
         labelKey: "error.proposal_not_found",
         level: "error",
         detail: `id: "${proposalId}"`,
       });
 
-      // Wait a bit before redirection so the user recognizes on which page the error occures
+      // Wait a bit before redirection so the user recognizes on which page the error occurs
```
mstrasinskis, line 48: 🤦‍♂️ thanks

mstrasinskis (commented):
The code is definitely cleaner, but I'm still not sure about single `unsubscribe`. For example if I'm on the `http://127.0.0.1:8080/#/proposal/50` and manually update the url to be http://127.0.0.1:8080/#/proposal/`63` there is no update of the data or any other feedback to the user. And this is not the flutter dapp behaviour.

frontend/svelte/src/lib/stores/route.store.ts
peterpeterparker: I was my original idea but `document.referrer` does not returns the path

Navigate from `https://nns.ic0.app/#/proposals` to `https://nns.ic0.app/#/neurons` -> referrer = `https://nns.ic0.app/v2/`

peterpeterparker: Navigate from `http://localhost:8080/#/proposal/76` to `http://localhost:8080/#/proposals` -> referrer = `''`

frontend/svelte/src/routes/ProposalDetail.svelte
peterpeterparker: haha I am the one that generally make such mistakes, happy to have spotted one 😉

peterpeterparker:
> The code is definitely cleaner, but I'm still not sure about single unsubscribe. For example if I'm on the .../#/proposal/50 and manually update the url to be ...#/proposal/63 there is no update of the data or any other feedback to the user. And this is not the flutter dapp behaviour.

You are right. I revert the code but I add a `unsubscribe` to `goBack` and that should do the tricks for both use cases.

mstrasinskis (approved):
LGTM. Thanks.

PR #509, opened 2022-03-01
fix: missing csp

peterpeterparker:
# Motivation

The CSP for the svelte app is not injected because there is no more `env.PRODUCTION` config.

# Changes

- fix `env.PRODUCTION` with `env.ENVIRONMENT`

### DIFF
--- a/frontend/svelte/scripts/build.index.mjs
+++ b/frontend/svelte/scripts/build.index.mjs
@@ -43,7 +43,7 @@ const updateBaseHref = (content) =>
  */
 const updateCSP = (content) => {
   // In local development mode, no CSP rule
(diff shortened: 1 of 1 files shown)


nmattia (approved):
Rubber stamp; one note though: the II repo has [tests](https://github.com/dfinity/internet-identity/blob/33950f4518b0acd2691d6e2a52d687b3b540d408/backend-tests/backend-tests.hs#L393) that check that the correct headers are set; maybe you'll want to do the same here?

peterpeterparker:
Thanks. Will ping you anyway to move the csp to the backend to apply the rules to the http headers.

PR #513, opened 2022-03-02
L2-313: Fetch and render neurons

lmuntanerDfinity:
# Motivation

Fetch and render neurons

# Changes

* NeuronCard uses NeuronInfo as prop.
* New neurons store.
* Setup NeuronDetails page.
* Neurons route fetches neurons and uses neurons store.
* New utils secondsToDuration

# Tests

* NeuronCard component test
* neurons store test
* secondsToDuration util
* getNeurons from neurons.services
* route NeuronDetail test setup


lmuntanerDfinity:
UI does not change much since we already had a Neuron card, but here it is anyway:

Screenshot 2022-03-02 at 09 48 36
Screenshot 2022-03-02 at 09 48 44

frontend/svelte/src/lib/utils/date.utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  amount: number;
+};
+const createLabel = (labelKey: LabelKey, amount: bigint): LabelInfo => ({
+  labelKey,
+  amount: Number(amount),
+});
+
+export const secondsToDuration = (seconds: bigint): string => {
+  const i18nObj = get(i18n);
+  const years = seconds / BigInt(ONE_YEAR_SECONDS);
```
mstrasinskis, line 21: Not sure about using BigInts for calculation because in the end all of them are converted to Numbers.

mstrasinskis (commented):
LGTM. Just one suggestion.

frontend/svelte/src/lib/utils/date.utils.ts
mstrasinskis: Probably it's already outdated info https://betterprogramming.pub/the-downsides-of-bigints-in-javascript-6350fd807d#4823

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -1,44 +1,105 @@
...
   import Card from "../ui/Card.svelte";
 
-  const goToDetail = () => {
-    // TODO L2-218: Go to the detail page of Neuron
+  export let neuron: NeuronInfo;
+  // Setting default value avoids warning missing props during testing
+  export let role: undefined | "link" | "button" = undefined;
+  export let ariaLabel: string | undefined = undefined;
+
+  type StateInfo = {
```
peterpeterparker, line 18: would it make sense to extract these "just" to make the component smaller for readability?

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -1,44 +1,105 @@
...
+    textKey: string;
+    Icon?: typeof SvelteComponent;
+    colorVar: "--background-contrast" | "--yellow-500" | "--gray-200";
+  };
+  type StateMapper = {
+    [key: number]: StateInfo;
```
peterpeterparker, line 24: isn't key typeof the enum or something instead of number? not sure that's possible

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -1,44 +1,105 @@
...
+    },
+    [NeuronState.DISSOLVING]: {
+      textKey: "dissolving",
+      Icon: IconHistoryToggleOff,
+      colorVar: "--yellow-500",
+    },
   };
+  const stateInfo: StateInfo = stateTextMapper[neuron.state];
+
+  const isCommunityFund = !!neuron.joinedCommunityFundTimestampSeconds;
```
peterpeterparker, line 49: shouldn't `isCommunityFund` and `isHotKeyControl` be re-actively assigned (`$: ....`) since they are depending on the property?

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -1,44 +1,105 @@
...
+    <h4 class:hasNeuronControl={isCommunityFund || isHotKeyControl}>
+      {neuron.neuronId}
+    </h4>
+    {#if isCommunityFund}
+      <p class="neuron-control">{$i18n.neurons.community_fund}</p>
+    {/if}
+    {#if isHotKeyControl}
+      <p class="neuron-control">{$i18n.neurons.hotkey_control}</p>
+    {/if}
+    <h5 style={`color: var(${stateInfo.colorVar});`}>
```
peterpeterparker, line 66: is that correct? shouldn't be `style={`color: ${stateInfo.colorVar};`}` to apply the value?

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -1,44 +1,105 @@
...
-<!-- TODO L2-218 -->
-<Card role="link" on:click={goToDetail} ariaLabel="go to neuron details">
+<Card {role} on:click {ariaLabel}>
   <div slot="start" class="lock">
-    <!-- TODO data from props: L2-218 -->
-    <h4>14225061177570733325</h4>
-    <h5 class="white">
-      Locked
-      <IconLockClock />
+    <h4 class:hasNeuronControl={isCommunityFund || isHotKeyControl}>
```
peterpeterparker, line 57: can it be `class:has-neuron-control`? does Svelte support this or that's why you picked `hasNeuronControl`?

frontend/svelte/src/lib/i18n/en.json
```diff
@@ -120,7 +129,20 @@
     "adopt": "Adopt",
     "reject": "Reject"
   },
+  "neuron_detail": {
```
peterpeterparker, line 132:
```
  "neuron_detail": {
    "title": "Neuron"
  },
  "neuron": {
    "title": "Neuron"
  },
  ```
  
  we can probably spare one

File: frontend/svelte/src/lib/stores/neurons.store.ts [hunk 8]
```diff
@@ -0,0 +1,26 @@
...
+  const { subscribe, update, set } = writable<NeuronInfo[]>([]);
+
+  return {
+    subscribe,
+
+    setNeurons(neurons: NeuronInfo[]) {
+      set([...neurons]);
+    },
+
+    pushNeurons(neurons: NeuronInfo[]) {
```
peterpeterparker, line 20: `pushNeurons` is used?

frontend/svelte/src/routes/Neurons.svelte
```diff
@@ -7,13 +7,33 @@
...
   // TODO: To be removed once this page has been implemented
-  onMount(() => {
+  onMount(async () => {
     if (process.env.REDIRECT_TO_LEGACY) {
       window.location.replace("/#/neurons");
     }
-    // TODO: Fetch and render neurons L2-313
+    try {
+      isLoading = true;
+      await getNeurons();
```
peterpeterparker, line 27: should it be called `listNeurons` to follows `listProposals` pattern?
or should we rename the proposals functions?

peterpeterparker (changes requested):

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: Yes, I tried. It wasn't working.

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: aren't they recalculated when a new prop is passed?

I can change it.

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: no, I can do it with `-`, no problem. I got used to it because normally we don't do the `=` and use a JS variable directly.

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: I haven't tried this one, the string as inline style works: I got it from here: https://svelte.dev/repl/5fab983a191d459da149ef55e7962cf8?version=3.2.1

I guess this is not React, hehe

frontend/svelte/src/lib/i18n/en.json
lmuntanerDfinity: Not sure the intention of "neuron", I imagined we wanted only one key for the neuron details, but I can reuse it. And since it's proposal_detail the other one...

frontend/svelte/src/lib/stores/neurons.store.ts
lmuntanerDfinity: not yet. I added a todo that staking a neuron should return a neuron and use this method.

frontend/svelte/src/lib/utils/date.utils.ts
lmuntanerDfinity: I was afraid of losing data by converting it to numbers. It comes with bigint from the backend.

frontend/svelte/src/routes/Neurons.svelte
lmuntanerDfinity: ok, I will rename.

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: to extract the state logic? sure

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
peterpeterparker: a pity. thanks for the feedback.

frontend/svelte/src/lib/utils/date.utils.ts
lmuntanerDfinity: I think this is a question for the SDK team, if they are using bigint, despite the lower performance, there must be a reason. If I convert it to Number and then we lose some range, it could lead to wrong outcome.

lmuntanerDfinity: It's not that we will are going to do this calculation many times either...

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: I think it's how you say, but I will look into it. To be on the safe side I refactored it.

lmuntanerDfinity:

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
peterpeterparker: read it again, yes that's work. I was disturbed because I would not have solved it like this. I generally don't inline the variable but only values - i.e. I would have done something like:

```
<h5 style={`--my-local-color: ${stateInfo.colorVar};`}>

<style>
   h5 {
       color: var(--my-local-color);
    }
```

In that way the styling find still place in the css. Notably makes the css more readable to me because I notice it is variable without looking at the html code.

I can live with it.

frontend/svelte/src/lib/i18n/en.json
peterpeterparker: You can get rid of "neuron" and replace it with "neuron_detail". I added it, totally fine if you replace it. Just no double code ;)

frontend/svelte/src/lib/stores/neurons.store.ts
peterpeterparker: ok then can  your update the related comment that still refers to "proposals"?

peterpeterparker:
commented

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
```diff
@@ -1,44 +1,75 @@
...
-    // TODO L2-218: Go to the detail page of Neuron
-  };
+  export let neuron: NeuronInfo;
+  // Setting default value avoids warning missing props during testing
+  export let role: undefined | "link" | "button" = undefined;
+  export let ariaLabel: string | undefined = undefined;
+
+  let stateInfo: StateInfo = getStateInfo(neuron.state);
+  $: stateInfo = getStateInfo(neuron.state);
+  let isCommunityFund: boolean = !!neuron.joinedCommunityFundTimestampSeconds;
```
peterpeterparker, line 17: You can spare the default value in the declaration if I am not wrong

```
let isCommunityFund: boolean;
$: isCommunityFund = !!neuron.joinedCommunityFundTimestampSeconds;
// etc.
```

frontend/svelte/src/lib/stores/neurons.store.ts
```diff
@@ -0,0 +1,26 @@
+import type { NeuronInfo } from "@dfinity/nns";
+import { writable } from "svelte/store";
+
+/**
+ * A store that contains the neurons
+ *
+ * - setNeurons: replace the current list of neurons with a new list
+ * - pushNeurons: append neurons to the current list of neurons. Notably useful when the neurons are fetched in a page that implements an infinite scrolling.
```
peterpeterparker, line 8: but there's no infinite scrolling right? might want to cut that from the comment too

peterpeterparker (approved):
Thanks!

Just please solve the new two comments before merge.

frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
lmuntanerDfinity: let me try

lmuntanerDfinity: no no, it works like a charm.

frontend/svelte/src/lib/utils/date.utils.ts
mstrasinskis: It's cool. Thanks for explanation.

PR #516, opened 2022-03-02
feat: load users' canisters

peterpeterparker:
# Motivation

Get the list of canisters of the users with the NNSDapp canister.

# Changes

- copy `getCanisters` from `ts` to svelte app `NNSDappCanister`
- add a store that contains the canisters
- load the canisters `onMount` in the "Canisters" page

# Side note

This pattern is really similar to the one we have in routes "Neurons" and "Proposals" (`loading` state, spinner, catching errors etc.). In some point in the future we might refactor a bit things to avoid double code.


frontend/svelte/src/lib/services/canisters.services.ts
```diff
@@ -0,0 +1,46 @@
...
+import { NNSDappCanister } from "../canisters/nns-dapp/nns-dapp.canister";
+import type { CanisterDetails } from "../canisters/nns-dapp/nns-dapp.types";
+import { OWN_CANISTER_ID } from "../constants/canister-ids.constants";
+import { identityServiceURL } from "../constants/identity.constants";
+import { canistersStore } from "../stores/canisters.store";
+import { i18n } from "../stores/i18n";
+import { createAgent } from "../utils/agent.utils";
+
+export const listCanisters = async ({
+  clearBeforeQuery,
```
lmuntanerDfinity, line 12: do we need to clear before query here? Isn't a better experience to show the canisters?

If we want to do it always, we might want to do it as well in neurons.

Yet, I don't see the point of using a store if we always fetch and never reuse what we have fetched. We could then just use a variable inside each route component and avoid all the store functionality.

lmuntanerDfinity: This way we would also be sure that we never show stale data if we never store it.

frontend/svelte/src/routes/Canisters.svelte
```diff
@@ -29,6 +58,19 @@
         {$i18n.canisters.principal_is}
         {$authStore.identity?.getPrincipal().toText()}
       </p>
+
+      <!-- TODO(L2-335): infinite scroll and cards -->
```
lmuntanerDfinity, line 62: I also thought about it, is the functionality in nns-js or even in the canister available? I didn't find it for the neurons.

lmuntanerDfinity, line 62: I also thought about it, is the functionality in nns-js or even in the canister available? I didn't find it for the neurons.

frontend/svelte/src/tests/mocks/canisters.store.mock.ts
```diff
@@ -0,0 +1,22 @@
+import { NNSDappCanister } from "../../lib/canisters/nns-dapp/nns-dapp.canister";
+import type { CanisterDetails } from "../../lib/canisters/nns-dapp/nns-dapp.types";
+import { mockCanisters } from "./canisters.mock";
+
+// @ts-ignore
+export class MockNNSDappCanister extends NNSDappCanister {
```
lmuntanerDfinity, line 6: did you try the mock from `jest-extended-mock`? I felt it easier to work with that than having to create a new mock everytime.

lmuntanerDfinity (approved):
A few questions worth discussing for the future. But already approving it since they are not blocking.

frontend/svelte/src/lib/services/canisters.services.ts
peterpeterparker: 1/1 the exact same pattern - both UX and technically wise - as "Voting" page and proposals.

frontend/svelte/src/routes/Canisters.svelte
peterpeterparker: You are right not at the moment. Removed the "infinite scroll" comment.

frontend/svelte/src/tests/mocks/canisters.store.mock.ts
peterpeterparker: No did not try out. This solution is easy for me because I already used it multiple times.
Not against though, maybe you can show me next week how it work with the lib?

frontend/svelte/src/lib/services/canisters.services.ts
peterpeterparker: Store has probably here the biggest advantage that we don't have to pass states through an infinite number of components. We can for sure discuss it. Good to go forward though.

lmuntanerDfinity: ok

lmuntanerDfinity: will refactor neurons in a small separate PR then/

frontend/svelte/src/tests/mocks/canisters.store.mock.ts
peterpeterparker: Oh cool. Then that's actually what I used to test the canister service itself as you did 😉

PR #517, opened 2022-03-02
Sharpen up the sha code and add utilities

bitdivine:
# Motivation
We now have a lookup mechanism from wasm sha to commit(s) however it is still a bit rough.

# Changes
* Add utilities for looking up and listing shas.
* Disambiguate the reference used when marking the commit with the sha.
* Make the name of the git note less specific, so that other artifact shas can be included as well.  At the moment there is only one that seems worth tracking but it seems unnecessary to bake that assumption in.

# Tests

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -73,9 +73,9 @@ jobs:
          git config user.name "GitHub Actions Bot"
          git config user.email "<>"
          : Make a note of the WASM shasum.
-         NOTES_REF="refs/notes/${{ matrix.DEPLOY_ENV }}/wasm-sha"
-         SHA="$(sha256sum < "${{ matrix.DEPLOY_ENV }}-out/nns-dapp.wasm")"
-         echo "$NOTES_REF: $SHA"
+         NOTES_REF="refs/notes/${{ matrix.DEPLOY_ENV }}/sha"
+         SHA="$(sha256sum "${{ matrix.DEPLOY_ENV }}-out/nns-dapp.wasm")"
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -73,11 +73,12 @@ jobs:
          git config user.name "GitHub Actions Bot"
          git config user.email "<>"
```
bitdivine, line 74: * "NOTE" is actually the ref, not the text in the note.  Confusing, so renamed.
* The sha line has been changed so that we can add more files if needed.  No other files need to be hashed at the moment but why engineer ourselves into a corner unnecessarily?  The ref of the git note has been renamed to match this more general meaning.
* The "if git notes" part can reuse the reference, so that has been factored out.

scripts/shasum-log
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+# Prints the history of shasums in the current branch.
+set -eu
+# Fetch if possible:
+git fetch origin refs/notes/*:refs/notes/* 2>/dev/null || true
+# DEPLOY_ENV - typical values are 'testnet' and 'mainnet'
+NOTES_REF=refs/notes/${DEPLOY_ENV:-testnet}/sha
+git log --pretty=format:"%H %N" --show-notes="$NOTES_REF" "${@}" | awk '(NF>1)'
```
bitdivine, line 8: Two scripts to make the sha notes easy to use.

lmuntanerDfinity: These are used by the developer, right? Not the CI.

peterpeterparker:
Do we still need this PR open a year ago?

PR #520, opened 2022-03-03
feat: add title to <Toast> message

mstrasinskis:
# Motivation

Some long error messages are not fully visible (see screenshot)
image

# Changes

- Added title attribute to the message (quick solution)

# Tests

- Added title test


lmuntanerDfinity (approved):
thanks!

PR #522, opened 2022-03-03
Let the svelte linter fail

bitdivine:
# Motivation
The svelte linter rules have been or will soon be updated.  This causes, as expected, a lot of errors.  According to our plan, we are going to allow these errors to appear for a while, to give people time to fix them.  However for that github must not block PR merges on account of lint failures.

# Changes
* Put the svelte lint job in a separate task.
* Make a new dummy task that controls whether we may merge, rather than listing requirements in the admin panel.

Screenshot from 2022-03-03 15-37-36

# TODO
* Update the merge rules, but this can be done a few fays after this has been merged to minimise disruption.

# Tests
Let's see whether this PR is mergeable with a failed lint result....

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -194,6 +194,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Compile typescript
+        run: npm run build
+
       - name: Run linter
         run: npm run check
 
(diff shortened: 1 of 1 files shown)


lmuntanerDfinity:
Ok, shall I add the recommended bundle of rules and that's it then? I can do that in another PR

PR #525, opened 2022-03-04
L2-312: Setup Eslint Bundle

lmuntanerDfinity:
# Motivation

Add eslint rules.

# Changes

* Add bundle of rules to eslint.
* Add eslint ignore file.
* Fix style and typing issues.


lmuntanerDfinity:
@mstrasinskis @peterpeterparker @bitdivine please review

lmuntanerDfinity:
We have:

```
263 problems (226 errors, 37 warnings)
6 errors and 0 warnings potentially fixable with the `--fix` option.
```

Let's take it easy.

frontend/svelte/.eslintrc.js
```diff
@@ -2,9 +2,9 @@ module.exports = {
   root: true,
   parser: "@typescript-eslint/parser",
   extends: [
-    // TODO: Team meeting to decide rules
-    // "plugin:@typescript-eslint/recommended",
-    // "plugin:@typescript-eslint/recommended-requiring-type-checking",
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:@typescript-eslint/recommended-requiring-type-checking",
```
bitdivine, line 7: One idea:  Can we introduce these one at a time?  I.e. add one rule, fix everything over the next few days, add the next and so on?

A second point:  Beware the type checking.  I am all in favour of requiring ts types, however when I first pitched in with svelte I tried adding types everywhere and in some cases the linter would require a type annotation but the svelte parser choked on the annotation.  There was an open svelte issue about this.  I don't have the link and perhaps this has been fixed.  We can hope! :-)

frontend/svelte/.eslintrc.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
   env: {
     es6: true,
     browser: true,
+    node: true,
```
bitdivine, line 20: What does that do in this context?

bitdivine (approved):
No objections but please wait until others have reviewed as well.

frontend/svelte/.eslintrc.js
lmuntanerDfinity: For example, it recognizes `process` instead of raising an error as a not defined variable.

frontend/svelte/.eslintrc.js
lmuntanerDfinity: I haven't got any weird type annotation needed that the svelte parse had problems with.

But I removed the types part for now. I like the idea of introducing it per parts.

With only the recommended eslint rules we have `67 problems (30 errors, 37 warnings)`.

We need to add `eslint:recommended` together with `plugin:@typescript-eslint/recommended`. This is the recommendation in the typescript-eslint plugin [docs](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin#usage).

frontend/svelte/.eslintrc.js
```diff
@@ -35,5 +37,11 @@ module.exports = {
   },
   plugins: ["svelte3", "@typescript-eslint"],
   ignorePatterns: ["node_modules", ".eslintrc.js"],
-  rules: {},
+  rules: {
```
nmattia, line 40: How about adding the ones used in II as well? https://github.com/dfinity/internet-identity/blob/33950f4518b0acd2691d6e2a52d687b3b540d408/.eslintrc.json#L12-L19

lmuntanerDfinity: looks good to me and it keeps consistency across projects, so perfect. Thanks!

nmattia (approved):
Rubber stamping; please check with @peterpeterparker for details!

frontend/svelte/src/tests/routes/Accounts.spec.ts
```diff
@@ -43,9 +43,10 @@ describe("Accounts", () => {
 
     const titleRow = container.querySelector("section > div");
 
-    expect(titleRow.textContent).toEqual(
-      `Accounts ${formatICP(mockMainAccount.balance.toE8s())} ICP`
-    );
+    titleRow &&
```
mstrasinskis, line 46: Generally speaking about this approach, doesn't this change the logic of tests that if the component isn't presented in the DOM the test wouldn't fail?

lmuntanerDfinity: done! thanks for the call!

mstrasinskis (approved):
LGTM. Thanks 👍

frontend/svelte/.eslintrc.js
```diff
@@ -2,8 +2,9 @@ module.exports = {
   root: true,
   parser: "@typescript-eslint/parser",
   extends: [
-    // TODO: Team meeting to decide rules
-    // "plugin:@typescript-eslint/recommended",
+    "eslint:recommended",
```
peterpeterparker, line 5: In sveltekit they also extends prettier per default

```
extends: ['eslint:recommended', 'plugin:@typescript-eslint/recommended', 'prettier'],
```

something we also want?

frontend/svelte/src/lib/services/dev.services.ts
```diff
@@ -10,6 +10,10 @@ import { syncAccounts } from "./accounts.services";
 export const getICPs = async (icps: number) => {
   const { main }: AccountsStore = get(accountsStore);
 
+  if (!main) {
```
peterpeterparker, line 13: that's a logic change not eslint. should it be safely ignored or should an error be throw because such use case should not be reached?

frontend/svelte/.eslintrc.js
lmuntanerDfinity: I don't mind.

Here I also didn't integrate yet the type-checking to do it in smaller PRs.

I can create a PR to add "prettier" as well.

frontend/svelte/src/lib/services/dev.services.ts
```diff
@@ -28,5 +32,8 @@ export const getICPs = async (icps: number) => {
   }
 
   const { identity }: AuthStore = get(authStore);
-  await syncAccounts({ identity });
+
+  if (identity) {
```
peterpeterparker, line 36: as previous comment

frontend/svelte/src/lib/services/proposals.services.ts
```diff
@@ -130,6 +130,9 @@ const queryProposalInfo = async ({
 
 export const getProposalId = (path: string): ProposalId | undefined => {
   const pathDetail = path.split("/").pop();
+  if (pathDetail === undefined) {
```
peterpeterparker, line 133: as previous comment

frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -19,17 +19,17 @@ export const lastProposalId = (
...
   proposal: Proposal
 ): [string, string][] => {
   const key = proposalFirstActionKey(proposal);
-  if (!key) {
+  if (key === undefined) {
     return [];
   }
 
-  return Object.entries(proposal.action[key])
+  return Object.entries(proposal.action?.[key])
```
peterpeterparker, line 32: that will throw an error no? `Object.entries(undefined)` throw an error

frontend/svelte/src/tests/mocks/canisters.store.mock.ts
```diff
@@ -1,22 +0,0 @@
-import { NNSDappCanister } from "../../lib/canisters/nns-dapp/nns-dapp.canister";
```
peterpeterparker, line 1: why removed in this PR?

frontend/svelte/src/tests/mocks/ledger.canister.mock.ts
```diff
@@ -1,20 +0,0 @@
-import { AccountIdentifier, ICP, LedgerCanister } from "@dfinity/nns";
```
peterpeterparker, line 1: why removed in this PR?

peterpeterparker (changes requested):
A couple of questions

frontend/svelte/src/lib/services/dev.services.ts
lmuntanerDfinity: it would throw an error if it doesn't despite eslint, so I can throw an error here.

Although, this is somethig we need to fix at a broad level as well. How do we handle session expiration at the app level?

frontend/svelte/src/lib/utils/proposals.utils.ts
lmuntanerDfinity: same as before

frontend/svelte/src/tests/mocks/canisters.store.mock.ts
lmuntanerDfinity: there were some `null` types in the mock and it was easier to do the test with `jest-extended-mock` than to fix the nul l type errors in the mock.

frontend/svelte/src/tests/mocks/ledger.canister.mock.ts
lmuntanerDfinity: Ditto, there were some `null` type errors that were needed to be solved. But it was easier to use `jest-extended-mock` than to fix the type issues in the mock.

peterpeterparker:
Regarding rules these are the one I would spontaneously to suggest:

- no-else-return
- no-continue
- prefer-const
- prefer-destructuring

We can of course discuss about it. We shall also discuss the thematic "implicit vs explicit typing" (object-shorthand I guess).

lmuntanerDfinity:
@peterpeterparker since we are going to start working, can we add these rules in another PR? This touches a lot of files and we don't want to start getting into conflicts.

If I had known your preference on the rules I would have added them to this PR.

frontend/svelte/src/lib/services/proposals.services.ts
lmuntanerDfinity: I tried, and if pathDetail is undefined, parseInt returns NaN, therefore the `isFinite` returns false and the whole function returns `undefined`. There is no change in logic.

peterpeterparker:
@lmuntanerDfinity sure definitely in another PR 👍

frontend/svelte/.eslintrc.js
peterpeterparker: another pr sounds good 👍

lmuntanerDfinity:

I create the new Jira issue to discuss and to add new rules once we fix the ones we have here: https://dfinity.atlassian.net/browse/L2-347

frontend/svelte/src/routes/NeuronDetail.svelte
```diff
@@ -1,9 +1,11 @@
 <script>
   import HeadlessLayout from "../lib/components/common/HeadlessLayout.svelte";
   import { i18n } from "../lib/stores/i18n";
+
+  const showThisRoute = process.env.REDIRECT_TO_LEGACY === "never";
 </script>
 
-{#if process.env.REDIRECT_TO_LEGACY === "never"}
+{#if showThisRoute}
```
peterpeterparker, line 8: this is not related to eslint right or is it?

peterpeterparker (changes requested):

frontend/svelte/src/routes/NeuronDetail.svelte
lmuntanerDfinity: no, it was broken though. The previous commit added a refactor on `REDIRECT_TO_LEGACY`

peterpeterparker: @lmuntanerDfinity can you then provide the fix in a separate PR?

lmuntanerDfinity: `main` is broken withuot this change, I couldn't check that everything was still working witoht this fix.

lmuntanerDfinity:
@peterpeterparker changes in REDIRECT_TO_LEGACY merged from `main`.

peterpeterparker (approved):

PR #526, opened 2022-03-06
fix: stringify all proposalActionFields

mstrasinskis:
# Motivation

Currently only the payload was stringified. It appears that other fields could also contain objects (screenshot)
image

# Changes

- apply stringifyJson to all fields
- update post-formatting to simulate the flutter dapp

# Tests

- `should stringify all objects`


frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -31,10 +31,19 @@ export const proposalActionFields = (
 
   return Object.entries(proposal.action[key])
     .filter(([key]) => key !== "payloadBytes")
-    .map(([key, value]: [string, object]) => [
-      key,
-      key === "payload" ? stringifyJson(value) : `${value}`,
-    ]);
+    .map(([key, value]: [string, object]) => {
+      let asText = stringifyJson(value, { indentation: 2 });
+      if (asText?.length) {
```
peterpeterparker, line 36: - the test might not needed I think, you can use the `?` on each `asText = asText?.replace`

- can we extract the "remove redundant wrappers" in a function?

frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -31,10 +31,19 @@ export const proposalActionFields = (
 
   return Object.entries(proposal.action[key])
     .filter(([key]) => key !== "payloadBytes")
-    .map(([key, value]: [string, object]) => [
-      key,
-      key === "payload" ? stringifyJson(value) : `${value}`,
-    ]);
+    .map(([key, value]: [string, object]) => {
+      let asText = stringifyJson(value, { indentation: 2 });
```
peterpeterparker, line 35: - the test might not needed I think, you can use the `?` on each `asText = asText?.replace`

- can we extract the "remove redundant wrappers" in a function?

peterpeterparker (changes requested):

frontend/svelte/src/lib/utils/proposals.utils.ts
lmuntanerDfinity: agree with extracting it to a function, then it will be easy to test.

mstrasinskis: Not sure about separate function because it will pollute the global scope w/ a temporary solution. A test for formatting was added 👍

peterpeterparker: Then a local function? 

When I read the code, I first notice many lines in a .map but don't get their context until I read the comments and code.

If I read `.map(removeRedundantWrappers)` then I get the context instantly.

mstrasinskis: Refactoring is done. Thanks, much cleaner 👍

peterpeterparker:
@mstrasinskis I have merged it so everything is in for the demo, hope you don't mind

PR #535, opened 2022-03-07
fix: backdrop dismiss click propagation

peterpeterparker:
# Motivation

Modal's backdrop dismiss should not be propagated. For example, in the list of proposals, dismissing the modal leads to navigating to detail page 🤪.

# Changes

- add `stopPropagation` to backdrop dismiss

# Use case

Capture d’écran 2022-03-07 à 13 00 43


mstrasinskis (approved):
LGTM 👍

lmuntanerDfinity (approved):

PR #536, opened 2022-03-07
Make staging builds in CI

bitdivine:
# Motivation
We now have three levels of `REDIRECT_TO_LEGACY`:
- never == false == pure svelte
- prod == as in production, which currently means that all tabs are served by flutter
- staging == as in production, but wit hone more tab served as svelte.

As M1 macs cannot make docker builds, it would be useful if CI could produce at least "never" and "staging" for development.

We keep the existing names for existing builds to avoid disruption.

# Changes
- Add a third build for staging
- Rename "true" to "prod" as "true" will be misleading as soon as we have 1 tab in svelte in prod.
 
# Tests
The docker builds in CI have succeeded.
- The mainnet build should be unaltered by this change.  Checking:
  - The CI mainnet build hash is: 
  - 1e3a424407eac5a8dc1eeb27109cd48b0005264624302eec86a6d735b111dc9d
  - The merge root is: 354af16031b7da1b7d2852909346a3f325ea459a 
  - Building the merge root locally yields: 1e3a424407eac5a8dc1eeb27109cd48b0005264624302eec86a6d735b111dc9d
  - Result: PASS
- The staging deployment SHOULD work with `./scripts/deploy-ci-to-testnet`
  - The canister gets deployed and redirects for the voting tab.
  - Result:  PASS

# Future work not addressed in this PR
There is a bunch of errors related to OWN_CANISTER_ID.  I have checked, they are unrelated to this PR and caused by the way OWN_CANISTER_ID is set.  This needs to be addressed, however that is orthogonal to this PR.

### DIFF
--- a/scripts/deploy-ci-to-testnet
+++ b/scripts/deploy-ci-to-testnet
@@ -18,7 +18,7 @@ else
 
   # Gets the most recent 'Backend wasm module' download:
   # shellcheck disable=SC2010 # we really do want this, not globbing.
-  DOWNLOAD_FILENAME="$(ls -t "$DOWNLOADS_DIR" | grep -E '.*(T|t)estnet backend wasm module( [(][0-9]+[)])?.zip$' | head -n1)"
+  DOWNLOAD_FILENAME="$(ls -t "$DOWNLOADS_DIR" | grep -E '.*backend wasm module( [(][0-9]+[)])?.zip$' | head -n1)"
   DOWNLOAD_PATH="$DOWNLOADS_DIR/$DOWNLOAD_FILENAME"
   echo "Sourcing WASM from: $DOWNLOAD_PATH"
 fi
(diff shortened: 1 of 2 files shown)


scripts/deploy-ci-to-testnet
```diff
@@ -18,7 +18,7 @@ else
 
   # Gets the most recent 'Backend wasm module' download:
   # shellcheck disable=SC2010 # we really do want this, not globbing.
-  DOWNLOAD_FILENAME="$(ls -t "$DOWNLOADS_DIR" | grep -E '.*(T|t)estnet backend wasm module( [(][0-9]+[)])?.zip$' | head -n1)"
+  DOWNLOAD_FILENAME="$(ls -t "$DOWNLOADS_DIR" | grep -E '.*backend wasm module( [(][0-9]+[)])?.zip$' | head -n1)"
```
peterpeterparker, line 21: I was so proud of my `(T|t)estnet` regex 😉

peterpeterparker (approved):
LGTM

PR #539, opened 2022-03-07
Merge neurons

anishh2003:


hpeebles:
When a user requests to merge 2 neurons, the only check we do in the browser is that the 2 neurons are controlled by the same principal.
There are additional restrictions (eg. neither neuron can be in the community fund and the 2 neurons must have matching followees on the 'ManageNeuron' topic), but these are only checked in the governance canister since it seems possible that these restrictions will be dropped in the future.
Any errors which the governance canister returns are shown to the user so if any of the checks fail the user will be told the reason.

hpeebles:
This code is deployed to here - https://qvhpv-4qaaa-aaaaa-aaagq-cai.nnsdapp.dfinity.network/v2/
I have tested the various scenarios and everything looks to be working.
The target and source neurons are always assigned correctly so as to ensure age bonus is always preserved.
The error messages bubble up to the end user whenever they are unable to merge neurons (eg. the controllers don't match).

PR #552, opened 2022-03-09
style: fix modal overflow

peterpeterparker:
# Motivation

When the wizard is used in the modal, the transition are glitchy because an horizontal scroll is briefly displayed on transitions.

When the content of the modal fits, a vertical scrollbar is displayed. That should not be the case. Such a scroll should be displayed only if the content can be scrolled.

# Changes

- style `overflow-x` and `overflow-y` of the modal's content

### DIFF
--- a/frontend/svelte/src/lib/modals/Modal.svelte
+++ b/frontend/svelte/src/lib/modals/Modal.svelte
@@ -182,7 +182,8 @@
 
     min-height: calc(var(--modal-min-height) - var(--modal-toolbar-height));
     max-height: calc(100vh - 156px);
-    overflow-y: scroll;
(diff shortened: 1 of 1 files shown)


lmuntanerDfinity (approved):

lmuntanerDfinity:
Thanks!

peterpeterparker:
# Motivation

When the wizard is used in the modal, the transition are glitchy because an horizontal scroll is briefly displayed on transitions.

When the content of the modal fits, a vertical scrollbar is displayed. That should not be the case. Such a scroll should be displayed only if the content can be scrolled.

# Changes

- style `overflow-x` and `overflow-y` of the modal's content

PR #553, opened 2022-03-09
fix: display stake and align state and stake

peterpeterparker:
# Motivation

If I am not wrong, "Stake" text can be displayed even if no ICP.

This PR also fix the alignment in case not hotkey are displayed and no community fund are displayed.

# Changes

- display "Stake" text also without ICP
- align state and stake for all cases

# Screenshots

Is:

Capture d’écran 2022-03-09 à 10 28 03

Will be:

Capture d’écran 2022-03-09 à 10 24 59

### DIFF
--- a/frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
+++ b/frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
@@ -48,18 +48,21 @@
     {#if isHotKeyControl}
       <span class="neuron-control">{$i18n.neurons.hotkey_control}</span>
     {/if}
-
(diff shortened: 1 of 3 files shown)


lmuntanerDfinity (approved):
Thanks, I feel these kind of alingments should be defined in the issue when we create the UI component.

Since now we are rewriting the Flutter version there is no need for this, but I would like if these were defined in the new design.

peterpeterparker:
Thanks for the review. I operate well with few description but like I said, don't have much opinion on JIRA related process. Happy to follow what you guyz decide.

PR #554, opened 2022-03-09
Fix query proposal certified

mstrasinskis:
# Motivation

Currently `queryProposal` uses query to fetch the data. But all request should use update instead.

# Changes

- getProposal function -> certified: true

# Tests

- updated getProposal.spec


peterpeterparker (approved):
Thanks!

lmuntanerDfinity (approved):

PR #557, opened 2022-03-09
L2-330: Display new neuron details

lmuntanerDfinity:
# Motivation

Show new neuron details when updating the delay after creating it.

# Changes

* SetDissolveDelayModal expects a neuron and uses the data to render the details.
* votingPower util.

# Tests

* Upgraded CreateNeuronModal tests to check that information used to create a neuron is there.
* votingPower util tests

### DIFF
--- a/frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
+++ b/frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
@@ -12,6 +12,8 @@
   import { StepsState } from "../../services/stepsState.services";
   import SetDissolveDelay from "./SetDissolveDelay.svelte";
   import type { NeuronId } from "@dfinity/nns";
+  import type { NeuronInfo } from "@dfinity/nns";
+  import { neuronsStore } from "../../stores/neurons.store";
 
   enum Steps {
     SelectAccount,
@@ -28,7 +30,7 @@
     }
   );
 
-  let newNeuronId: NeuronId | undefined;
+  let newNeuron: NeuronInfo | undefined;
 
   const chooseAccount = () => {
     stateStep = stateStep.next();
@@ -39,7 +41,9 @@
   const goToDissolveDelay = ({
     detail,
   }: CustomEvent<{ neuronId: NeuronId }>) => {
-    newNeuronId = detail.neuronId;
+    newNeuron = $neuronsStore.find(
+      (neuron) => neuron.neuronId === detail.neuronId
+    );
     stateStep = stateStep.next();
   };
 
@@ -92,10 +96,10 @@
       </Transition>
     {/if}
     <!-- TODO: Manage edge case: https://dfinity.atlassian.net/browse/L2-329 -->
-    {#if currentStep === Steps.SetDissolveDelay && newNeuronId}
+    {#if currentStep === Steps.SetDissolveDelay && newNeuron}
       <Transition {diff}>
         <!-- TODO: Edit Followees https://dfinity.atlassian.net/browse/L2-337 -->
-        <SetDissolveDelay neuronId={newNeuronId} on:nnsNext={close} />
+        <SetDissolveDelay neuron={newNeuron} on:nnsNext={close} />
       </Transition>
     {/if}
   </main>
--- a/frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
+++ b/frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
@@ -42,7 +42,7 @@
     detail,
   }: CustomEvent<{ neuronId: NeuronId }>) => {
     newNeuron = $neuronsStore.find(
-      (neuron) => neuron.neuronId === detail.neuronId
+      ({ neuronId }) => neuronId === detail.neuronId
     );
     stateStep = stateStep.next();
   };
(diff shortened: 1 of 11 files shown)


lmuntanerDfinity:

No UI changes.

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -21,6 +21,18 @@ import { AuthStore, authStore } from "../stores/auth.store";
 import { neuronsStore } from "../stores/neurons.store";
 import { createAgent } from "../utils/agent.utils";
 
+export const getNeuron = async (
```
peterpeterparker, line 24: out of curiosity, any reason why this moves?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -70,16 +80,13 @@ export const listNeurons = async (): Promise<void> => {
   neuronsStore.setNeurons(neurons);
 };
 
-export const getNeuron = async (
-  neuronId: NeuronId
-): Promise<NeuronInfo | undefined> => {
-  const { canister, identity } = await governanceCanister();
+export const listNeuron = async (neuronId: NeuronId): Promise<void> => {
```
peterpeterparker, line 83: won't you mind using something else than `list...`? we use "list" to, well, list lists.

`loadNeuron` or something  maybe?

btw. it's correct it does not fetch a list of neurons anymore but only a single neuron?

frontend/svelte/src/lib/utils/neuron.utils.ts
```diff
@@ -37,3 +42,12 @@ const stateTextMapper: StateMapper = {
 
 export const getStateInfo = (neuronState: NeuronState): StateInfo =>
   stateTextMapper[neuronState];
+
+export const votingPower = (
```
peterpeterparker, line 46: can you use object params to follow the pattern `{stake, dis...}: {stak...`?

peterpeterparker (changes requested):
Looks good. Beside suggestions - i.e. comments only, two observations.

frontend/svelte/src/lib/modals/neurons/SetDissolveDelay.svelte
```diff
@@ -50,17 +58,22 @@
...
     <h5>{$i18n.neurons.neuron_id}</h5>
-    <p>{neuronId}</p>
+    <p>{neuron.neuronId}</p>
   </div>
   <div>
     <h5>{$i18n.neurons.neuron_balance}</h5>
-    <!-- TODO: Get Neuron info https://dfinity.atlassian.net/browse/L2-330 -->
-    <p>1.10 ICP Stake</p>
+    <p data-tid="neuron-stake">
+      {`${formatICP(neuronICP)} ${$i18n.neurons.icp_stake}`}
```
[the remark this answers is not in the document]
lmuntanerDfinity: I prefer to be specific of what separates them. Otherwise it doesn't seem clear what or how much should be separating both texts.

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: I use it in the functions below. Therefore, I preferred to move it above.

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
```diff
@@ -39,7 +41,9 @@
   const goToDissolveDelay = ({
     detail,
   }: CustomEvent<{ neuronId: NeuronId }>) => {
-    newNeuronId = detail.neuronId;
+    newNeuron = $neuronsStore.find(
+      (neuron) => neuron.neuronId === detail.neuronId
```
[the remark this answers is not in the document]
lmuntanerDfinity: sure, it will be caught by eslint at some point, so better now than later.

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: kind of open on the naming, here 'listNeuron' returns 'void'  and all other 'listSomething' return arrays of data. that's why I think it would be best to not call it 'listNeuron' because it can be misinterpreted before effectivelly readoing code, you know what I mean?

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: Thanks for the answer.

frontend/svelte/src/lib/modals/neurons/SetDissolveDelay.svelte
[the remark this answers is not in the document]
peterpeterparker: That would be fancy 🤘

frontend/svelte/src/lib/utils/neuron.utils.ts
lmuntanerDfinity: noted!

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: `listNeurons` returns void. Instead it adds it to the store. Same for `listProposals`.

lmuntanerDfinity:
@peterpeterparker ready for next round of review!

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: 🙈

alright let's skip it here and see if we can put together a pattern later. Thanks for the feedbacks

peterpeterparker (approved):
LGTM, thanks

PR #558, opened 2022-03-09
L2-311: Registration e2e in CI pipeline

lmuntanerDfinity:
# Motivation

Registration e2e tests working in the CI pipeline.

# Changes

* Added "internet_identity" project to `dfx.json`.
* Proxy Internet Identity to localhost:8087.
* Added IDENTITY_SERVICE_URL when building NNS Dapp for e2e tests.
* Deploy of NNS Dapp for e2e tests all in Svelte.

# Tests

* Enable registration e2e test.


dfx.json
```diff
@@ -6,6 +6,12 @@
       "candid": "rs/nns-dapp.did",
       "wasm": "nns-dapp.wasm",
       "build": "./build.sh"
+    },
+    "internet_identity": {
```
lmuntanerDfinity, line 10: I am not sure whether this will mess up with the deployment process of the NNS Dapp.

dfx.json
```diff
@@ -6,6 +6,12 @@
...
+      "type": "custom",
+      "wasm": "internet_identity.wasm",
+      "candid": "internet_identity.did",
+      "build": "curl -sSL https://github.com/dfinity/internet-identity/releases/download/release-initial/internet_identity_dev.wasm -o internet_identity.wasm"
```
lmuntanerDfinity, line 14: thanks @nmattia !

e2e-tests/package.json
```diff
@@ -12,7 +13,7 @@
     "@wdio/local-runner": "^7.16.13",
     "@wdio/mocha-framework": "^7.16.13",
     "@wdio/spec-reporter": "^7.16.13",
-    "chromedriver": "^97.0.0",
+    "chromedriver": "^98.0.0",
```
lmuntanerDfinity, line 16: I needed it to update it because I have Chrome v98 installed locally. It wasn't working with v97.

e2e-tests/test.js
```diff
@@ -47,37 +50,64 @@ console.log(`Using replica host: ${replica_host}`);
...
 const wdio = child_process.spawn("npm", ["run", "wdio"], {
-  env: { ...process.env, NNS_DAPP_URL: NNS_DAPP_URL },
+  env: { ...process.env, NNS_DAPP_URL, II_URL },
 });
 
 wdio.stdout.on("data", (data) => {
   console.log(`wdio: ${data}`);
 });
 
+wdio.stderr.on("data", (data) => {
```
lmuntanerDfinity, line 105: This helped me debug the error with the chrome version. I was lost without it.

internet_identity.did
```diff
@@ -0,0 +1 @@
+service : () -> {}
```
lmuntanerDfinity, line 1: this is needed to deploy II wasm to the local replica

lmuntanerDfinity:
@nmattia please take a look. It's ready for your review. Not sure ready for merge because there are many changes that I am not sure how they can affect the pipeline.

lmuntanerDfinity:
@bitdivine I would also appreciate your feedback if you have time. I added some parts that might affect the deployment process.

e2e-tests/test.js
```diff
@@ -47,37 +50,64 @@ console.log(`Using replica host: ${replica_host}`);
...
-proxy.stdout.on("close", (code) => {
-  console.log(`proxy returned with ${code}`);
+nnsDappProxy.stdout.on("close", (code) => {
+  console.log(`nns dapp proxy returned with ${code}`);
+});
+
+const iiProxy = child_process.spawn("npm", [
+  "run",
+  "proxy",
+  "--",
```
nmattia, line 80: why is this needed? doesn't `npm run test` put `proxy` on the $PATH?

e2e-tests/test.js
```diff
@@ -47,37 +50,64 @@ console.log(`Using replica host: ${replica_host}`);
...
+  "--replica-host",
+  replica_host,
+  `${ii_canister_id}:${II_PORT}`,
```
nmattia, line 83: What about this, and removing the other `proxy` call?
```suggestion
  `${ii_canister_id}:${II_PORT}`, `${canister_id}:${NNS_DAPP_PORT}`,
```

nmattia (commented):
Nice! That's a lot of different `DEPLOY_ENV`s that we have now, which all have their explicit configs. What about reducing this a bit and simply using `IDENTITY_SERVICE_URL`? As far as I can tell (but I might be missing something) it's the only config difference with `local`.

e2e-tests/test.js
lmuntanerDfinity: oh, I was having other problems and was trying whether that was the problem. I will revert the changes.

e2e-tests/test.js
lmuntanerDfinity: I didn't know I could do that!

.github/workflows/build.yml
```diff
@@ -94,9 +94,10 @@ jobs:
           dfx start --background
           dfx canister --no-wallet create --all
 
+      # IDENTITY_SERVICE_URL needs to match `e2e-tests/test.js`
```
nmattia, line 97: ❤️

e2e-tests/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "wdio": "wdio run ./wdio.conf.ts",
+    "proxy": "proxy",
```
nmattia, line 7:
```suggestion
```

e2e-tests/specs/example.e2e.ts
```diff
@@ -5,13 +5,21 @@
...
 function waitForLoad(browser) {
   // Check that the page has completely loaded, and we are not in the intermediate service worker bootstrap page:
-  return browser.waitUntil(() => browser.execute(() => {
-    // We do not want the service worker bootstrap page, so:
-    const serviceWorkerTitlePattern = /Content Validation Bootstrap/i;
-    const isServiceWorkerBootstrapPage = !!document.title.match(serviceWorkerTitlePattern);
-    // Check that the page has loaded and that it isn't the intermediate bootstrap page.
-    return (document.readyState === 'complete') && !isServiceWorkerBootstrapPage;
-  }), {timeout: 60_000});
+  return browser.waitUntil(
```
nmattia, line 8: did this change or is this only formatting? if the latter, let's try and avoid in the future!

e2e-tests/package.json
lmuntanerDfinity: thanks!

e2e-tests/test.js
```diff
@@ -13,17 +13,20 @@ const fs = require("fs");
  */
 const CANISTER_IDS_PATH = `${__dirname}/../.dfx/local/canister_ids.json`;
 let canister_id;
+let ii_canister_id;
```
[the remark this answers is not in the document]
lmuntanerDfinity: makes sense

e2e-tests/specs/example.e2e.ts
lmuntanerDfinity: formatting. weird, this was probably the format commit.

lmuntanerDfinity:
@nmattia any idea why the pipeline is green, when the test isn't passing?

```
[chrome 98.0.4758.102 linux #0-0] 1 failing
[chrome 98.0.4758.102 linux #0-0]
[chrome 98.0.4758.102 linux #0-0] 1) landing page register and back to dashboard
[chrome 98.0.4758.102 linux #0-0] waitUntil condition failed with the following reason: Can't call getElementText on element with selector "h1" because element wasn't found
```

But still, we have a green check.

nmattia:
@lmuntanerDfinity could it be that the return code from the `wdio` process is not bubbled up? In particular, I see this:

```
wdio returned with false
proxy returned with false
```

would be good to double check the data passed on `close` and potentially throw an error there, depending on the return code

bitdivine:
A minor niggle:  The screenshot shows a spinner, so it will most likely change with every run.  It needs to wait until the page is fully loaded.

lmuntanerDfinity:
@bitdivine good point. We can't rely to be fully loaded yet, we would need to deploy the Ledger Canister to the local replica and proxy the calls there. Otherwise, we are relying on the Ledger Canister to be deployed and working on testnet.

We can't leave it like this though.

Option 1: Remove screenshot of second test.
Option 2: Rely on Ledger Canister on testnet and wait for it to be loaded. The screenshot might change from time to time.

Anyway, I will create a ticket to deploy Ledger Canister to local replica and use it in the test.

What do you prefer?

nmattia:
@lmuntanerDfinity you could also `execute()` on the browser and do something like `document.querySelelector("#spinner").remove()` (or whatever the correct selector is)

lmuntanerDfinity:
@nmattia very good idea, I will try that.

lmuntanerDfinity:
@bitdivine @nmattia removed the spinner before the screenshot. Tests passing. Let me know if you have more suggestions.

e2e-tests/specs/example.e2e.ts
```diff
@@ -102,25 +110,26 @@ describe("landing page", () => {
...
+
+    // At this point it's still the login page
+    // We wait for the header of the dashboard
+    await browser.$('header').waitForExist({ timeout: 20_000 });
     const title = await browser.$("h1");
+    const titleText = await title.getText();
+
+    expect(titleText).toBe("Accounts");
+
+    const spinner = await browser.$('main section svg');
```
nmattia, line 129:
```suggestion
    // remove spinner to make screenshots deterministic
    const spinner = await browser.$('main section svg');
```

e2e-tests/test.js
```diff
@@ -70,14 +76,29 @@ proxy.stdout.on("close", (code) => {
...
+
 wdio.stdout.on("close", (code) => {
   console.log(`wdio returned with ${code}`);
   proxy.kill();
 });
+
+wdio.on('exit', (code, signal) => {
+  if (code > 0) {
+    // bubble up error from wdio tests
+    throw new Error('End-to-end tests failed');
```
nmattia, line 98:
```suggestion
    throw new Error(`End-to-end tests returned with ${code}`);
```

nmattia (approved):
LGTM, thanks!

PR #560, opened 2022-03-10
Fix: Eslint errors 2

lmuntanerDfinity:
# Motivation

Fix eslint errors, part 2

# Changes

* Fix no-require.
* Fix no-empty-function.
* Allow `ts-ignore` with description.
* Fix no-unused-var

# Tests

No extra tests.

### DIFF
--- a/frontend/svelte/.eslintrc.js
+++ b/frontend/svelte/.eslintrc.js
@@ -50,5 +50,11 @@ module.exports = {
         allowNumber: false,
       },
     ],
+    "@typescript-eslint/ban-ts-comment": [
+      2,
+      {
+        "ts-ignore": "allow-with-description",
+      },
+    ],
   },
 };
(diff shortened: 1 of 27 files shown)


lmuntanerDfinity:

Only `✖ 19 problems (6 errors, 13 warnings)` more to go!

But they are from different rules and I thought better to do it tomorrow in another PR.

mstrasinskis (approved):
LGTM

peterpeterparker:
@lmuntanerDfinity `import * as en...` are reported as error in my editor, that's why they were imported with `require`. I guess it depends on the node version or configuration. Any idea how to solve this?

Capture d’écran 2022-03-14 à 07 55 32

peterpeterparker:
@lmuntanerDfinity actual no, if I just open "svelte" I don't have the error. Being said, unfortunately can do that all the time because Webstorm goes banana on the memory allocation when I do so.

not a blocker though. If you ever had an idea of solution let me know.

lmuntanerDfinity:
We have a similar issues with VSCode, some features work or not work depending on opening the whole project or an individual one.

This seems to be a general problem in text editors. Personally, I have open only the `svelte` folder to work on the frontend. When I work on 2e2 for example, or want to search something in flutter, I open a new text editor with the whole project.

PR #563, opened 2022-03-10
Fix: Change VotesCard props pattern

lmuntanerDfinity:
# Motivation

Adapt `VotesCard` to follow same pattern as the other components in ProposalDetail.

# Changes

* `VotesCard` expects array of neurons.

# Tests

* Tests on `VotesCard` have been updated.


lmuntanerDfinity:
@mstrasinskis @peterpeterparker quick PR to adapt to new props pattern.

PR #570, opened 2022-03-14
L2 367 refactor cast vote card

mstrasinskis:
# Motivation

To improve readability and testing of voting process the CastVoteCard component was refactored.
**No functional changes** except a couple of bugfixes in voting error handling

# Changes

- CastVoteCard.svelte was splited: 
  - VotingCard.svelte
  - VotingNeuronSelect.svelte 
  - VotingConfirmationToolbar.svelte 
  - proposals.services / registerVotes (logic)
  - votingNeuronSelectStore (neurons selection)
- proposals.utils / selectedNeuronsVotingPover

## extra
- stringifyJSON - devMode

# Tests
Added more tests for voting process

- VotingCard.spec.ts
- VotingConfirmationToolbar.spec.ts 
- VotingNeuronSelect.spec.ts
- proposals.services.spec.ts
- proposals.store.spec.ts
- proposals.utils.spec.ts
- utils.spec.ts


frontend/svelte/jest-setup.ts
```diff
@@ -18,3 +19,8 @@ process.env.GOVERNANCE_CANISTER_ID = "rrkah-fqaaa-aaaaa-aaaaq-cai";
 process.env.LEDGER_CANISTER_ID = "ryjl3-tyaaa-aaaaa-aaaba-cai";
 process.env.IDENTITY_SERVICE_URL =
   "https://qjdve-lqaaa-aaaaa-aaaeq-cai.nnsdapp.dfinity.network";
+
+// testing-library setup
+configure({
+  testIdAttribute: "data-tid",
```
lmuntanerDfinity, line 25: cool 💯

lmuntanerDfinity (commented):
I have a few generic questions about stores, when to use them and when not.

We already have a neuronsStore, but I assume this is only going to be used in the NeuronsTab? Aren't we coupling our stores to the pages? What if the pages in the new design have a mix of neurons and proposals? We won't be able to reuse our stores, there would be duplication between stores.

If we agree that the stores are tightly coupled with a route, could we store them in the same folder maybe? Along the components that are used there?

Or is it that "proposals.store.ts" already couples this with the Proposals sections?

Should we do the same for accounts? In the neurons tab I use the accounts, should I use the accountsStore or use another store? Or accounts are a different resource than neurons and therefore we consider them more generic?

lmuntanerDfinity:
@mstrasinskis a comment that I might bring up in the nns intro actually, no need to reply here. I realized when I was writing it.

lmuntanerDfinity (approved):

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingConfirmationToolbar.svelte
```diff
@@ -0,0 +1,68 @@
...
+
+  const showAdoptConfirmation = () => {
+    selectedVoteType = Vote.YES;
+    showConfirmationModal = true;
+  };
+  const showRejectConfirmation = () => {
+    selectedVoteType = Vote.NO;
+    showConfirmationModal = true;
+  };
+  const cancel = () => (showConfirmationModal = false);
```
peterpeterparker, line 28: `selectedVoteType` should be resetted here no?

peterpeterparker (changes requested):
Is it possible that `votingNeuronSelectStore` is not yet resetted on destroy?

If correct, I think it would be cleaner to also clean it on destroy to not keep the last information in memory.

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingConfirmationToolbar.svelte
mstrasinskis: onDestroy was added to clean up the memory, good point 👍
Not sure about reseting the voteType. Currently it feels like unnecessary complexity because there is no way to start the process w/o voteType selection.

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingCard.svelte
```diff
@@ -0,0 +1,50 @@
...
+      neurons,
+      proposal: proposalInfo,
+    })
+  );
+  $: visible =
+    $votingNeuronSelectStore.neurons.length > 0 &&
+    proposalInfo.status === ProposalStatus.PROPOSAL_STATUS_OPEN;
+
+  const vote = async ({ detail }: { detail: { voteType: Vote } }) =>
+    registerVotes({
```
peterpeterparker, line 34: you don't `await` on purpose?

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingConfirmationToolbar.svelte
peterpeterparker: Alright and thx for the `onDestroy`

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingConfirmationToolbar.svelte
```diff
@@ -0,0 +1,68 @@
...
+  const confirm = async () => {
```
peterpeterparker, line 29: is not `async`

peterpeterparker (changes requested):
Two more questions regarding `async` (sorry not spotted these the first time)

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingCard.svelte
mstrasinskis: Hmm... Don't see a reason for that. True, will remove `async`

peterpeterparker (approved):
Thx!

PR #575, opened 2022-03-15
build: bump dependencies and fix type import

peterpeterparker:
# Motivation

Bump all dependencies to stay up-to-date and get last patches.

Fix import that are `type`.

# Changes

- bump all dependencies
- set import used for type safety only as `type`


lmuntanerDfinity (approved):
Great thanks!!

We should do this monthly.

PR #577, opened 2022-03-15
L2 387 no neurons initially on proposal detail

mstrasinskis:
# Motivation

🪲🪲🪲
1. the proposal detail info was shown before the neurons were loaded. That could make wrong impression that for example there was no voting done
2. after successful voting an error was always shown
3. in case the proposal id is changed directly in URL there was no visible feedback that the new proposal is loading

# Changes

- for now the unknown store state is fixed on the component level (could be changed after update/query research).

# Tests

- added test to not show error messages on successful response


peterpeterparker (approved):
LGTM, thanks

PR #567, opened 2022-03-11
L2-346: Logout expired session

lmuntanerDfinity:
# Motivation

User should be logged out when session expires.

# Changes

* Set a timeout to log the user out when session expires.

# Tests

* Upgrade auth tests to check that a timeout is set.

PR #582, opened 2022-03-16
feat: web workers and sign-out on invalid auth delegation

peterpeterparker:
# Motivation

This PR introduces the support for web workers to the nns-dapp svelte app and a feature that checks periodically in the background if the authentication delegation is still valid. In case of invalidity, the cronjob initiate an automatic sign-out of the user. 

# Changes

Worker:

- rollup dependency [@surma/rollup-plugin-off-main-thread](https://www.npmjs.com/package/@surma/rollup-plugin-off-main-thread) to use the worker as module
- enhance `rollup.config.js` to bundle both an app `bundle.js` and a new `worker.js`

Feature:

- `worker.ts` as main entry of the worker which also listen for messages of the client side
- `auth.worker.ts` to check periodically the auth status
- sync auth and worker in `App.svelte`
- new utils to retrieve delegation and identity saved by agent-js in the local storage
- add `worker.services.ts` which initialize the worker on the client side including the message listener

# Context

Feature wise, this PR has the same goal as PR [#567 - Logout expired session](https://github.com/dfinity/nns-dapp/pull/567). We shall decide which approach we rather like.

# Pros and cons

Things to be discuss and that can be enhanced. The pros and cons of this PR in comparison to #567 and maybe generally speaking.

## Cons

- #567 follows same approach as Flutter - there is no cron
- agent-js is extremely heavy. as implemented in this PR, its code will be duplicated in both client `bundle.js` and worker `worker.js`. "Twice the size". However, this might be improved with a better configuration or bundler (I wish we would use sveltekit - i.e vitejs)
- unfortunately web workers have no access to local storage. no a big deal per se, from the client side we send to the worker the delegation but without such limitation, we could have spared some lines of code

## Pros

- delegating work to web worker removes stress from the render side
- #567 applies changes to the auth store - this pr introduces zero changes in such store nor in the way we deal with the auth
- when we are going to develop new application, we can reuse this approach - i.e code almost as it is (1 to 1)
- feature can be easily extended with new messages. for example, if we would like to display a modal to inform the user about the automatic sign-out, this would be relatively straight forward
- web workers might be an interesting options for the topics "certified queries". I can image a scenario when we defer call to workers to perform "query calls + certified/update calls + check if all data are the same"

# Demo

For demo purpose, in addition to the delegation check, I added a counter that leads to a sign-out if a maximal value is reached.

ezgif com-gif-maker

# Performance

The worker is downloaded when the app is accessed - i.e. also on the login screen but is loaded asynchronously once the page is rendered.

Capture d’écran 2022-03-16 à 12 41 05

### DIFF
--- a/frontend/svelte/src/lib/services/worker.services.ts
+++ b/frontend/svelte/src/lib/services/worker.services.ts
@@ -1,0 +1,34 @@
+import type { AuthStore } from "../stores/auth.store";
+import { authStore } from "../stores/auth.store";
+import type { PostMessageEventData } from "../types/post-messages";
+import { localStorageAuth } from "../utils/auth.utils";
+
+const initWorker = () => {
+  const worker = new Worker("./build/worker.js", { type: "module" });
+
+  worker.onmessage = async ({ data }: MessageEvent<PostMessageEventData>) => {
+    const { msg } = data;
+
+    switch (msg) {
+      case "nnsSignOut":
+        await authStore.signOut();
+        return;
+    }
+  };
+
+  return {
+    syncAuthIdle: async (auth: AuthStore) => {
+      if (!auth.identity) {
+        worker.postMessage({ msg: "nnsStopIdleTimer" });
+        return;
+      }
+
+      worker.postMessage({
+        msg: "nnsStartIdleTimer",
+        data: await localStorageAuth(),
+      });
+    }
+  };
+};
+
+export const worker = initWorker();
--- a/frontend/svelte/src/lib/services/worker.services.ts
+++ b/frontend/svelte/src/lib/services/worker.services.ts
@@ -27,7 +27,7 @@ const initWorker = () => {
         msg: "nnsStartIdleTimer",
         data: await localStorageAuth(),
       });
-    }
+    },
   };
 };
 
--- a/frontend/svelte/src/lib/workers/auth.worker.ts
+++ b/frontend/svelte/src/lib/workers/auth.worker.ts
@@ -5,7 +5,7 @@ import type { LocalStorageAuth } from "../types/auth";
 let timer: NodeJS.Timeout | undefined = undefined;
 
 export const startIdleTimer = (data?: LocalStorageAuth) =>
-  (timer = setInterval(async () => await onIdleSignOut(data), 1000));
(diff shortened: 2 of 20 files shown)


frontend/svelte/src/lib/services/worker.services.ts
```diff
@@ -0,0 +1,34 @@
...
+  return {
+    syncAuthIdle: async (auth: AuthStore) => {
+      if (!auth.identity) {
+        worker.postMessage({ msg: "nnsStopIdleTimer" });
+        return;
+      }
+
+      worker.postMessage({
+        msg: "nnsStartIdleTimer",
+        data: await localStorageAuth(),
```
lmuntanerDfinity, line 28: Should we also send a message every time the local storage changes?

Otherwise the worker might have expired data.

In the future, we might want to refresh the session and local storage will change when we do so.

lmuntanerDfinity (changes requested):
It looks good to me, maybe add a small improvement

frontend/svelte/src/lib/services/worker.services.ts
peterpeterparker: It's actually already the case. If localstorage change -> `svelte:window` listener on storage kicks in -> refresh auth -> send message to worker

peterpeterparker: As discussed and tested together, above works out as expected.

frontend/svelte/src/lib/workers/auth.worker.ts
```diff
@@ -0,0 +1,38 @@
...
+export const stopIdleTimer = () => {
+  if (!timer) {
+    return;
+  }
+
+  clearInterval(timer);
+  timer = undefined;
+};
+
+const onIdleSignOut = async (data?: LocalStorageAuth) => {
```
mstrasinskis, line 19: Is there any async call in the function?

peterpeterparker: Good catch. Not anymore. There was when I tried the local storage but had to move it to the client side.
Fixed. Thx.

mstrasinskis (approved):
LGTM, thanks 👍

lmuntanerDfinity (approved):

PR #585, opened 2022-03-16
L2-399: End to end testing in testnet.

lmuntanerDfinity:
# Motivation

Prepare the project to add end-to-end tests easily against testnet.

# Changes

* Update Readme on how to run against a testnet.
* Rely on document title to switch between tabs instead of URL.
* Move some shared code to `common`.

# Tests

The changes are tests.


lmuntanerDfinity:
We can run e2e test against testnet with:

```
NNS_DAPP_URL=.... npm run wdio
```

We can now add tests that only run agains testnets, while we work on deploying Ledger and Governance canister locally.

bitdivine:
Running against the nnsdapp testnet...

Using an alias like this does not work:
NNS_DAPP_URL=https://nns-svelte.winning.gold npm run wdio  

But this works fine:
NNS_DAPP_URL=https://sbzkb-zqaaa-aaaaa-aaaiq-cai.nnsdapp.dfinity.network npm run wdio

e2e-tests/common/register.ts
```diff
@@ -0,0 +1,57 @@
+export const register = async (browser: WebdriverIO.Browser) => {
+  await browser.$("main button").waitForExist();
+  await browser.$("main button").click();
+  
+  // REGISTRATION
+  
+  // Internet Identity
+  // First title is from the Service Worker.
+  await browser.pause(5000);
```
bitdivine, line 9: This can be made faster for the impatient, with retries up to a limit of 5000ms, however the impatient will invariably add this optimization.

bitdivine (approved):

e2e-tests/common/register.ts
lmuntanerDfinity: hahaha, yes, indeed. I wanted to be on the safe side.

PR #589, opened 2022-03-21
feat: display skeleton text while loading voting history

peterpeterparker:
# Motivation

Loading voting history can take time and displaying another or multiple additional spinners would look weird. That's why we can present skeleton text while we are loading all the proposals of the selected voting history.

# Changes

- add ui element `SkeletonText`
- display skeleton texts until voting history proposal is loaded

# Screenshots

ezgif com-gif-maker

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Markdown.svelte
+++ b/frontend/svelte/src/lib/components/ui/Markdown.svelte
@@ -31,5 +31,5 @@
   {@html parse(removeHTMLTags(text))}
 {:else}
   <!-- fallback text content -->
-  <p>{text}</p>
+  <p data-tid="markdown-text">{text}</p>
 {/if}
--- a/frontend/svelte/src/lib/components/ui/SkeletonText.svelte
+++ b/frontend/svelte/src/lib/components/ui/SkeletonText.svelte
@@ -3,7 +3,7 @@
   export let animated: boolean = true;
 </script>
 
-<div class:animated data-tid="skeleton-text">
+<div class:animated data-tid="skeleton-text" aria-busy={true}>
   <span>&nbsp;</span>
 </div>
 
--- a/frontend/svelte/src/lib/components/ui/SkeletonText.svelte
+++ b/frontend/svelte/src/lib/components/ui/SkeletonText.svelte
@@ -1,0 +1,60 @@
+<!-- adapted source: https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/skeleton-text -->
+<script lang="ts">
+  export let animated: boolean = true;
+</script>
+
+<div class:animated data-tid="skeleton-text">
+  <span>&nbsp;</span>
+</div>
+
+<style lang="scss">
+  div {
+    display: block;
+
+    width: 100%;
+    height: inherit;
+
+    margin: calc(var(--padding) / 2) 0;
+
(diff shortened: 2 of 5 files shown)


frontend/svelte/src/lib/components/ui/SkeletonText.svelte
```diff
@@ -0,0 +1,63 @@
...
+  .animated {
+    position: relative;
+
+    background: linear-gradient(
+      to right,
+      var(--skeleton-text-background) 8%,
+      var(--skeleton-text-background-animated) 18%,
+      var(--skeleton-text-background) 33%
+    );
+    background-size: 800px 104px;
```
lmuntanerDfinity, line 45: why `800` and `104`?

lmuntanerDfinity (commented):
LGTM, just one question

frontend/svelte/src/lib/components/ui/SkeletonText.svelte
peterpeterparker: because of the animation `-global-skeleton-text-shimmer` my guess.

as per comment, adapter from https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/skeleton-text

lmuntanerDfinity (approved):

frontend/svelte/src/lib/components/ui/SkeletonText.svelte
```diff
@@ -0,0 +1,63 @@
+<!-- adapted source: https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/skeleton-text -->
+<script lang="ts">
+  export let animated: boolean = true;
+</script>
+
+<div class:animated data-tid="skeleton-text">
```
mstrasinskis, line 6: It would be better to replace div w/ some inline-block. Because of:

The [P](https://www.w3.org/TR/html401/struct/text.html#edef-P) element represents a paragraph. It cannot contain [block-level elements](https://www.w3.org/TR/html401/struct/global.html#block-inline) (including [P](https://www.w3.org/TR/html401/struct/text.html#edef-P) itself).

Not sure it's important in case of skeleton.

frontend/svelte/src/lib/components/ui/SkeletonText.svelte
```diff
@@ -0,0 +1,63 @@
...
+  <span>&nbsp;</span>
```
mstrasinskis, line 7: `aria-busy="true"` ?

frontend/svelte/src/lib/components/ui/Markdown.svelte
```diff
@@ -31,5 +31,5 @@
   {@html parse(removeHTMLTags(text))}
 {:else}
   <!-- fallback text content -->
-  <p>{text}</p>
+  <p data-tid="markdown-text">{text}</p>
```
mstrasinskis, line 34: Should the component be renamed into MarkdownText to have same pattern as skeleton?

mstrasinskis (commented):
Looks good. Just a couple of comments to consider.

frontend/svelte/src/lib/components/ui/SkeletonText.svelte
peterpeterparker: why not, thanks!

mstrasinskis (approved):
Thank you! LGTM

PR #590, opened 2022-03-21
test: workaround editors display only error resolveJsonModule

peterpeterparker:
# Motivation

Workaround to not display the error "Consider using '--resolveJsonModule' to import module with '.json' extension." in Webstorm and VS Code when the project is opened from root.

Not the prettiest but at least it solves the display of the misleading error.

# Changes

- import i18n as a mock with a single `ts-ignore`
- use exported mock in tests


lmuntanerDfinity (approved):
@mstrasinskis do you open VS Code from root? I have other problems apart from this one. But definitely not worth addressing in this PR.

I also don't need to open from root, I am fine opening only the `svelte` project.

But if this helps WebStorm, let's go with it. We will need to remember to import from the mock.

peterpeterparker:
Yeah with Webstorm I cannot open only `svelte`. I have to open from root otherwise the editor goes mental in terms of memory and cpu consumption.

mstrasinskis (approved):
LGTM

PR #591, opened 2022-03-21
L2 406 add combined request support query and update

mstrasinskis:
# Motivation

To increase the request speed instead of making one `update` request two request are done:
- `query` it's fast so the app could be rendered sooner
- `update` to provide more reliable data

## For Discussion
- what to extend w/ queryAndUpdate?
- do we need deep-compare (if yes then do we need lodash?)
- because `query` is so fast, do we need the workaround for proposalDetails page to search the proposal first in the proposalListStore?
- proposal store / pushProposals replaces store entries w/ certified version (do we need this practice?)

# Changes

- api.utils / queryAndUpdate
- extended with queryAndUpdate requests:
  - listNeurons
  - ~~findProposals~~
  - getProposal
 - updated neurons error handling (reuse pattern from promises.service)

# Tests

- api.utils
- proposal.store - replacement with certified proposals


frontend/svelte/src/lib/utils/api.utils.ts
```diff
@@ -0,0 +1,44 @@
...
+        if (certifiedDone) return;
+        onError?.({ certified: false, error });
+      }),
+
+    // update
+    request({ certified: true })
+      .then((response) => onLoad({ certified: true, response }))
+      .catch((error) => onError?.({ certified: true, error }))
+      .finally(() => (certifiedDone = true)),
+  ]).then();
```
peterpeterparker, line 43: You can spare this `.then()`

frontend/svelte/src/lib/utils/api.utils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  return Promise.race([
+    // query
+    request({ certified: false })
+      .then((response) => {
+        if (certifiedDone) return;
+        onLoad({ certified: false, response });
+      })
+      .catch((error: E) => {
+        if (certifiedDone) return;
```
peterpeterparker, line 34: should we ignore the query error if the update calls finished?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -63,8 +65,24 @@ export const listNeurons = async ({
...
-  const neurons: NeuronInfo[] = await queryNeurons({ identity });
-  neuronsStore.setNeurons(neurons);
+  return queryAndUpdate<NeuronInfo[], unknown>({
+    request: ({ certified }) => queryNeurons({ identity, certified }),
+    onLoad: ({ response: neurons }) => neuronsStore.setNeurons(neurons),
+    onError: ({ error, certified }) => {
+      console.error(error);
+
+      // Explicitly handle only UPDATE errors
+      if (certified === true) {
```
peterpeterparker, line 75: nth - an eslint rule we still have to discuss - I find easier to read when if returns

```
if (certified !== true) {
   return;
}

neuronsStore.setNeurons([]);

toastsStore.show({
   labelKey: "error.get_neurons",
   ...
```

peterpeterparker:
Really cool 👍

frontend/svelte/src/lib/stores/proposals.store.ts
```diff
@@ -37,11 +38,33 @@ const initProposalsStore = () => {
...
-      update((proposalInfos: ProposalInfo[]) => [
-        ...proposalInfos,
-        ...proposals,
-      ]);
+    pushProposals({
+      proposals,
+      certified,
+    }: {
+      proposals: ProposalInfo[];
+      certified: boolean | undefined;
```
lmuntanerDfinity, line 46: maybe we want the `certified` in the `ProposalInfo`? This way we don't lose this information.

frontend/svelte/src/lib/utils/api.utils.ts
mstrasinskis: True 👍

frontend/svelte/src/lib/services/neurons.services.ts
mstrasinskis: Just trying to avoid negative comparison. But agree; if we stick to one style, let's support it everywhere.

frontend/svelte/src/lib/stores/proposals.store.ts
mstrasinskis: Good point, maybe later as discussed.

peterpeterparker (approved):
LGTM, thx

PR #593, opened 2022-03-21
Unblock builds

bitdivine:
# Motivation
Builds are failing because of the git notes.

A quick resolution is needed to stop breaking other PRs, a deeper solution can follow.

Diagnosis:
The code does a git fetch to get the notes, however it looks as if it should be doing something like a git reset --hard origin/notes/* of just the notes.  Doing that properly will require learning more about notes.

# Changes
- Ignore git note failures

# Tests


peterpeterparker (approved):
Thanks for the workaround

PR #598, opened 2022-03-22
L2-322: Stake neurons from subaccounts

lmuntanerDfinity:
# Motivation

User can stake a neuron from subaccounts.

# Changes

* Add subaccounts to SelectAccount screen in CreatenNeuronModal.
* Add `fromSubAccount` parameter to all the stake neuron flow.
* New helper to transform from SubAccountArray to number that is needed for nns-js.

# Tests

* New test to check that user can select a subaccount in CreatenNeuronModal.

### DIFF
--- a/frontend/svelte/src/lib/api/utils.api.ts
+++ b/frontend/svelte/src/lib/api/utils.api.ts
@@ -1,18 +1,5 @@
 import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
 
-// Source: as in ts/src/canisters/converters.ts
-export const arrayOfNumberToUint8Array = (
-  numbers: Array<number>
-): Uint8Array => {
-  return new Uint8Array(numbers);
-};
-
-export const arrayOfNumberToArrayBuffer = (
-  numbers: Array<number>
-): ArrayBuffer => {
-  return arrayOfNumberToUint8Array(numbers).buffer;
-};
-
 export const arrayBufferToNumber = (buffer: ArrayBuffer): number => {
   const view = new DataView(buffer);
   return view.getUint32(view.byteLength - 4);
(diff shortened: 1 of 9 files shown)


lmuntanerDfinity:

frontend/svelte/src/lib/api/utils.api.ts
```diff
@@ -0,0 +1,30 @@
+import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
+
+// Source: as in ts/src/canisters/converters.ts
+export const arrayOfNumberToUint8Array = (
```
peterpeterparker, line 4: all these functions are exposed by nns-js

peterpeterparker (changes requested):

frontend/svelte/src/lib/api/utils.api.ts
lmuntanerDfinity: They might, but they are related to nns dapp canister, not to the ledger or governance canisters. Therefore, I decided to move them here and pass as subaccountIt what the ledger canister needs, not the type SubAccountArray, which is a type from NNS Dapp canister.

Does that make sense?

Or we also want nns dapp specific code in nns-js?

peterpeterparker: There functions are used within nns-js (toNeuron -> arrayOfNumberToUint8Array). Therefore I would not duplicate code.

peterpeterparker: But if you find a way to not use the functions within nns-js, I am not against moving the code in nns-dapp. Just don't like duplicate code ;)

lmuntanerDfinity: Removed unused code in nns-js in https://github.com/dfinity/nns-js/pull/62

And also refactored here.

lmuntanerDfinity:

peterpeterparker (approved):
LGTM

Since you have move `arrayBufferToNumber` and `toSubAccountId` it's maybe a good opportunity to add some tests for these? Not necessary in this PR.

frontend/svelte/src/lib/api/utils.api.ts
peterpeterparker: All clear, thx

lmuntanerDfinity:
Yes, I thought about tests for those, but to be honest, I would be copying the implementation in the test for the expected value.

I am not sure how that could help. Or just hardcode the current expected value and put it in a test. At least if we change the implementation we would be aware, I will do that in another PR.

PR #599, opened 2022-03-22
feat: adapt error mgmt to nns-js that throw errors instead of variant returns

peterpeterparker:
# Motivation

`nns-js` does not return variant ok-error types anymore but throw errors if there is, well, errors ([PR #59](https://github.com/dfinity/nns-js/pull/59)). That's why `nns-dapp` have to be adapted for that breaking change.

# Changes

- let `nns-js` throw errors - do not check return types for error objects

# Status

~~On hold for next nightly build of the library~~


frontend/svelte/src/lib/services/proposals.services.ts
```diff
@@ -210,26 +205,41 @@ const requestRegisterVotes = async ({
...
+      }
+
+      // We throw anyway unexpected errors
+      throw error;
+    }
+  };
+
+  // TODO: switch to Promise.allSettled -- https://dfinity.atlassian.net/browse/L2-369
+  const responses: Array<GovernanceError | undefined> = await Promise.all(
+    neuronIds.map((neuronId: NeuronId) => register(neuronId))
```
lmuntanerDfinity, line 233: `neuronIds.map(register)`?

frontend/svelte/src/lib/services/proposals.services.ts
```diff
@@ -210,26 +205,41 @@ const requestRegisterVotes = async ({
...
   );
-  const errors = responses.filter(Boolean);
+  const errors: GovernanceError[] = responses.filter(
+    Boolean
```
lmuntanerDfinity, line 236: If we used a different function, could we avoid casting?

```
responses.filter((r) => r instanceof GovernanceError);
```

lmuntanerDfinity (approved):
Minor comments, already approving for merge tomorrow

frontend/svelte/src/lib/services/proposals.services.ts
peterpeterparker: we would have to extend it in any case since we expicitely `try catch instanceof` to filter what we are looking for.
anyhow removed that particular cast.

frontend/svelte/src/lib/services/proposals.services.ts
peterpeterparker: nice

PR #603, opened 2022-03-22
feat: marked rendered types

peterpeterparker:
# Introduction

This PR is a branch of a branch so that @mstrasinskis can have a look and use what he think is useful or not for his PR.

# Motivation

I extracted the function to a utils because it was easier to debug.

I think the tests did not work out mostly because the mock had to be extended with the new `Rendered()` class.

I also extracted the `tsconfig` for the jest spec tests to a separate file. Not sure it is needed but it does not arm in any case.


mstrasinskis (approved):
Awesome! Thanks.

PR #604, opened 2022-03-22
L2 417 apply query and update strategy to get neuron

mstrasinskis:
# Motivation

Switch to queryAndUpdate approach for loading a neuron

[L2-417](https://dfinity.atlassian.net/browse/L2-417)

# Changes

- `i18n.error.get_neuron` was removed because the error handling is done in neuron service

# Tests

- no extra tests are necessary

### DIFF
--- a/frontend/svelte/src/lib/modals/proposals/ProposerModal.svelte
+++ b/frontend/svelte/src/lib/modals/proposals/ProposerModal.svelte
@@ -20,6 +20,7 @@
       return;
     }
 
+    // The fetched neuron belongs to a proposer so it should not be added to the neuronsStore
     loadNeuron({
       neuronId: proposer,
       identity: $authStore.identity,
--- a/frontend/svelte/src/lib/modals/proposals/ProposerModal.svelte
+++ b/frontend/svelte/src/lib/modals/proposals/ProposerModal.svelte
@@ -10,6 +10,7 @@
   import { onMount } from "svelte";
   import VotingHistoryCard from "../../components/neurons/VotingHistoryCard.svelte";
   import { authStore } from "../../stores/auth.store";
+  import { loadNeuron } from "../../services/neurons.services";
 
   export let proposer: NeuronId;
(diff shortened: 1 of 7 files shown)


frontend/svelte/src/lib/modals/proposals/ProposerModal.svelte
```diff
@@ -20,16 +20,12 @@
...
-      neuron = await queryNeuron({
-        neuronId: proposer,
-        identity: $authStore.identity,
-      });
-    } catch (err) {
-      neuron = undefined;
-
-      toastsStore.error({ labelKey: "error.get_neuron", err });
-    }
+    loadNeuron({
```
peterpeterparker, line 23: Should we add somewhere in the code a comment that the result of this particular query should not be added to the neuron store?

Right now it only query `queryNeuron`, no use of the store. We are going to use `loadNeuron` that uses the store (`return neuron || queryNeuron({ neuronId, identity, certified });`).

Maybe some day one will have the idea to had the result of the loading function to the store if the neuron wasn't found in the store.

lmuntanerDfinity: That is a good point. I had a similar case now. I needed to fetch the neuron again and push it to the store. It will be in an upcoming PR, I didn't do it with loadNeuron, but I change the `neuronsStore.update` to not add duplicated neurons and remove old ones.

Fixing the `update` should fix the case when the neuron is added twice.

lmuntanerDfinity (approved):
LGTM

frontend/svelte/src/lib/modals/proposals/ProposerModal.svelte
mstrasinskis: Thanks, updated!

peterpeterparker: Thanks for the comment Maks

peterpeterparker (approved):
LGTM

PR #605, opened 2022-03-23
feat: clear storage and reload browser on idle auth

peterpeterparker:
# Motivation

As for a standard logout we should clear storage and reload browser to clean the stores on automatic logout if the session becomes invalid.

In addition this PR adds tests that check that clear storage and reload of the browser are effectively called on any logout.

# Changes

- refactor/move `logout` function to a new service `auth.service`
- pass msg (`ToastMsg`)  that needs to be displayed after automatic logout through query params
- display msg on auth screen
- clean up msg from query params once displayed

### DIFF
--- a/frontend/svelte/src/lib/services/auth.services.ts
+++ b/frontend/svelte/src/lib/services/auth.services.ts
@@ -1,0 +1,82 @@
+import { authStore } from "../stores/auth.store";
+import { toastsStore } from "../stores/toasts.store";
+import type { ToastLevel, ToastMsg } from "../types/toast";
+import { replaceHistory } from "../utils/route.utils";
+
+const msgParam: string = "msg";
+const levelParam: string = "level";
+
+export const logout = async ({
+  msg = undefined,
+}: {
+  msg?: Pick<ToastMsg, "labelKey" | "level">;
+}) => {
+  await authStore.signOut();
+
+  appendOptionalMsgToUrl(msg);
+
+  window.localStorage.clear();
+
+  // We reload the page to make sure all the states are cleared
+  window.location.reload();
+};
+
+/**
+ * If a message was provided to the logout process - e.g. a message informing the logout happened because the session timedout - append the information to the url as query params
+ */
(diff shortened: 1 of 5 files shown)


frontend/svelte/src/lib/services/auth.services.ts
```diff
@@ -0,0 +1,82 @@
...
+import { toastsStore } from "../stores/toasts.store";
+import type { ToastLevel, ToastMsg } from "../types/toast";
+import { replaceHistory } from "../utils/route.utils";
+
+const msgParam: string = "msg";
+const levelParam: string = "level";
+
+export const logout = async ({
+  msg = undefined,
+}: {
```
lmuntanerDfinity, line 11: why not make the options object also optional? This way we can call `logout()`.

frontend/svelte/src/lib/services/auth.services.ts
```diff
@@ -0,0 +1,82 @@
...
+  msg?: Pick<ToastMsg, "labelKey" | "level">;
+}) => {
+  await authStore.signOut();
+
+  appendOptionalMsgToUrl(msg);
```
lmuntanerDfinity, line 16: wouldn't it be nice to have the `if (msg)` here and make the message mandatory as param of `appendMsgToUrl`?

lmuntanerDfinity (approved):
Looks very good! Minor code style comments, but already approving.

frontend/svelte/src/lib/services/auth.services.ts
peterpeterparker: because someone might just bind the function to a button `<button on:click={logout}`. doing so the param would not be `undefined` but a `MouseEvent | TouchEvent|KeyboardEvent`. Therefore, instead of adding some logic to prevent that the param is effectively a msg param, it easier to prevent it by using an object and gives also type security.

frontend/svelte/src/lib/services/auth.services.ts
peterpeterparker: I am not a big fan of `if something` in the middle of the code, rather like to defer logic to function but not against. Done.

PR #606, opened 2022-03-23
L2-403: Follow and unfollow known neurons

lmuntanerDfinity:
# Motivation

User can follow and unfollow known neurons.

# Changes

* Hardcode two known neurons in the governance api.
* New Component `KnownNeuronFollowItem`.
* Upgrade NewFolloweeModal to use new component.
* Add features to service `getNeuron` to not get cached data.
* neuronsStore.pushNeurons should not add duplicated neurons to store.

# Tests

* New tests in NewFolloweeModal to test for the unfollow functionality.
* New test for api util `toSubAccountId`.
* New test for pushNeurons.

### DIFF
--- a/frontend/svelte/src/lib/api/governance.api.ts
+++ b/frontend/svelte/src/lib/api/governance.api.ts
@@ -118,7 +118,25 @@ export const queryKnownNeurons = async ({
 }): Promise<KnownNeuron[]> => {
   const { canister } = await governanceCanister({ identity });
 
-  return canister.listKnownNeurons(certified);
+  const knownNeurons = await canister.listKnownNeurons(certified);
+
+  if (!knownNeurons.find(({ id }) => id === BigInt(27))) {
+    knownNeurons.push({
+      id: BigInt(27),
+      name: "DFINITY Foundation",
+      description: undefined,
+    });
+  }
+
+  if (!knownNeurons.find(({ id }) => id === BigInt(28))) {
+    knownNeurons.push({
+      id: BigInt(28),
+      name: "Internet Computer Association",
+      description: undefined,
+    });
+  }
+
+  return knownNeurons;
 };
 
 // TODO: Apply pattern to other canister instantiation L2-371
--- a/frontend/svelte/src/lib/api/governance.api.ts
+++ b/frontend/svelte/src/lib/api/governance.api.ts
@@ -7,6 +7,7 @@ import {
   LEDGER_CANISTER_ID,
 } from "../constants/canister-ids.constants";
 import { createAgent } from "../utils/agent.utils";
+import { dfinityNeuron, icNeuron } from "./constants.api";
 import { toSubAccountId } from "./utils.api";
 
 export const queryNeuron = async ({
@@ -120,20 +121,12 @@ export const queryKnownNeurons = async ({
 
   const knownNeurons = await canister.listKnownNeurons(certified);
 
-  if (!knownNeurons.find(({ id }) => id === BigInt(27))) {
-    knownNeurons.push({
-      id: BigInt(27),
-      name: "DFINITY Foundation",
-      description: undefined,
-    });
+  if (!knownNeurons.find(({ id }) => id === dfinityNeuron.id)) {
+    knownNeurons.push(dfinityNeuron);
   }
 
-  if (!knownNeurons.find(({ id }) => id === BigInt(28))) {
-    knownNeurons.push({
-      id: BigInt(28),
-      name: "Internet Computer Association",
-      description: undefined,
-    });
+  if (!knownNeurons.find(({ id }) => id === icNeuron.id)) {
+    knownNeurons.push(icNeuron);
   }
 
   return knownNeurons;
--- a/frontend/svelte/src/lib/components/neurons/KnownNeuronFollowItem.svelte
+++ b/frontend/svelte/src/lib/components/neurons/KnownNeuronFollowItem.svelte
@@ -1,20 +1,23 @@
 <script lang="ts">
   import type { KnownNeuron, NeuronId, Topic } from "@dfinity/nns";
+  import { createEventDispatcher } from "svelte";
   import { addFollowee, removeFollowee } from "../../services/neurons.services";
   import { authStore } from "../../stores/auth.store";
   import { i18n } from "../../stores/i18n";
-  import { toastsStore } from "../../stores/toasts.store";
   import Spinner from "../ui/Spinner.svelte";
 
   export let knownNeuron: KnownNeuron;
   export let topic: Topic;
   export let neuronId: NeuronId;
   export let isFollowed: boolean = false;
+  export let disabled: boolean = false;
 
   let loading: boolean = false;
+  const dispatcher = createEventDispatcher();
 
   const addKnownNeuronFollowee = async () => {
     loading = true;
+    dispatcher("nnsLoading", { loading: true });
     const toggleFollowee = isFollowed ? removeFollowee : addFollowee;
     await toggleFollowee({
       identity: $authStore.identity,
@@ -23,17 +26,14 @@
       followee: knownNeuron.id,
     });
     loading = false;
-    toastsStore.show({
-      labelKey: "new_followee.success_add_followee",
-      level: "info",
-    });
+    dispatcher("nnsLoading", { loading: false });
   };
 </script>
 
 <div data-tid={`known-neuron-item-${knownNeuron.id}`}>
   <p>{knownNeuron.name}</p>
   <!-- TODO: Fix style while loading - https://dfinity.atlassian.net/browse/L2-404 -->
-  <button class="secondary small" on:click={addKnownNeuronFollowee}>
+  <button class="secondary small" {disabled} on:click={addKnownNeuronFollowee}>
     {#if loading}
       <Spinner />
     {:else if isFollowed}
--- a/frontend/svelte/src/lib/components/neurons/KnownNeuronFollowItem.svelte
+++ b/frontend/svelte/src/lib/components/neurons/KnownNeuronFollowItem.svelte
@@ -1,0 +1,53 @@
+<script lang="ts">
+  import type { KnownNeuron, NeuronId, Topic } from "@dfinity/nns";
+  import { addFollowee, removeFollowee } from "../../services/neurons.services";
+  import { authStore } from "../../stores/auth.store";
+  import { i18n } from "../../stores/i18n";
+  import { toastsStore } from "../../stores/toasts.store";
+  import Spinner from "../ui/Spinner.svelte";
+
+  export let knownNeuron: KnownNeuron;
+  export let topic: Topic;
+  export let neuronId: NeuronId;
+  export let isFollowed: boolean = false;
+
+  let loading: boolean = false;
+
+  const addKnownNeuronFollowee = async () => {
+    loading = true;
+    const toggleFollowee = isFollowed ? removeFollowee : addFollowee;
+    await toggleFollowee({
+      identity: $authStore.identity,
+      neuronId: neuronId,
+      topic,
+      followee: knownNeuron.id,
+    });
+    loading = false;
+    toastsStore.show({
+      labelKey: "new_followee.success_add_followee",
+      level: "info",
+    });
+  };
+</script>
+
+<div data-tid={`known-neuron-item-${knownNeuron.id}`}>
+  <p>{knownNeuron.name}</p>
+  <!-- TODO: Fix style while loading - https://dfinity.atlassian.net/browse/L2-404 -->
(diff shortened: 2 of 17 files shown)


lmuntanerDfinity:
A [video](https://www.loom.com/share/2ff4644bf9f44257815220cffbdf8ca0) of how it looks.

<a href="https://www.loom.com/share/2ff4644bf9f44257815220cffbdf8ca0">
    <p>Network Nervous System frontend dapp - 22 March 2022 - Watch Video</p>
    
  </a>

lmuntanerDfinity:

frontend/svelte/src/lib/components/neurons/KnownNeuronFollowItem.svelte
```diff
@@ -0,0 +1,53 @@
...
+  import Spinner from "../ui/Spinner.svelte";
+
+  export let knownNeuron: KnownNeuron;
+  export let topic: Topic;
+  export let neuronId: NeuronId;
+  export let isFollowed: boolean = false;
+
+  let loading: boolean = false;
+
+  const addKnownNeuronFollowee = async () => {
```
mstrasinskis, line 16: suggestion: (toggle|update)KnownNeuronFollowee

frontend/svelte/src/lib/components/neurons/KnownNeuronFollowItem.svelte
```diff
@@ -0,0 +1,53 @@
...
+    loading = true;
+    const toggleFollowee = isFollowed ? removeFollowee : addFollowee;
+    await toggleFollowee({
+      identity: $authStore.identity,
+      neuronId: neuronId,
```
mstrasinskis, line 21: neuronId,

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -237,16 +244,21 @@ const getNeuron = async ({
   neuronId,
   identity,
   certified,
+  nocache = false,
 }: {
   neuronId: NeuronId;
   identity: Identity | null | undefined;
   certified: boolean;
+  nocache?: boolean;
```
mstrasinskis, line 252: Suggestion: probably something more specific then nocache would be better e.g. `forceFetch`

frontend/svelte/src/lib/stores/neurons.store.ts
```diff
@@ -19,8 +19,16 @@ const initNeuronsStore = () => {
       set([...neurons]);
     },
 
-    pushNeurons(neurons: NeuronInfo[]) {
-      update((neuronInfos: NeuronInfo[]) => [...neuronInfos, ...neurons]);
+    pushNeurons(newNeurons: NeuronInfo[]) {
+      update((oldNeurons: NeuronInfo[]) => {
+        const filteredNeurons = oldNeurons.filter(
```
mstrasinskis, line 24: Suggestion:
const newIds = Set(oldNeurons.map(({neuronId}) => neuronId))
then newIds.has(...)

mstrasinskis (commented):
Looks Good. Just a couple of suggestions.

frontend/svelte/src/lib/api/governance.api.ts
```diff
@@ -118,7 +118,25 @@ export const queryKnownNeurons = async ({
 }): Promise<KnownNeuron[]> => {
   const { canister } = await governanceCanister({ identity });
 
-  return canister.listKnownNeurons(certified);
+  const knownNeurons = await canister.listKnownNeurons(certified);
+
+  if (!knownNeurons.find(({ id }) => id === BigInt(27))) {
```
peterpeterparker, line 123: maybe extract the constant neurons to the `neuron.contants` and these two add to a separate function to make everything obvious?

actually do we need a function or `Set` can shorten the code? if `knowNeurons` shall no contain duplicate it maybe would work out?

```
// constants
export const dfinityNeuron = {id: BigInt(27)...

// api
return Array.from(new Set([knowNeurons, dfinityNeuron, icNeuron]))
```

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
```diff
@@ -16,13 +17,28 @@
 
   let followeeAddress: number | undefined;
   let loading: boolean = false;
+  let topicFollowees: NeuronId[];
+  $: {
+    const topicInfo = neuron.fullNeuron?.followees.find(
```
peterpeterparker, line 22: nth: es6

```
const topicInfo = neuron.fullNeuron?.followees.find(
      ({topic: followeeTopic}) => followeeTopic === topic
    );
topicFollowees = topicInfo.followees ?? [];
```

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
```diff
@@ -16,13 +17,28 @@
...
     });
   });
 
+  const followsKnownNeuron = ({
+    followees,
+    knownNeuronId,
+  }: {
+    followees: NeuronId[];
+    knownNeuronId: NeuronId;
+  }) => followees.find((id) => id === knownNeuronId) !== undefined;
```
peterpeterparker, line 40: - I did not get it at first because the return type of the function is missing, can you add it?

- Is it safe performance wise in Svelte to call a function directly from the template - i.e. not use the function as callback to assign a state and use the state in the template? In Angular for example that's possible but I see it as a bad practice that can leads to performance issue.

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -137,11 +137,17 @@ const setFolloweesHelper = async ({
...
+      certified: true,
+      nocache: true,
     });
+
+    if (!neuron) {
+      throw new Error("Neuron not found");
+    }
+    neuronsStore.pushNeurons([neuron]);
   } catch (error) {
     toastsStore.show({
```
peterpeterparker, line 152: nth: `toastStore.error(...`

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -137,11 +137,17 @@ const setFolloweesHelper = async ({
       topic,
       followees,
     });
-    await loadNeuron({
+    const neuron: NeuronInfo | undefined = await getNeuron({
       neuronId,
       identity,
-      setNeuron: (neuron: NeuronInfo) => neuronsStore.pushNeurons([neuron]),
+      certified: true,
+      nocache: true,
```
peterpeterparker, line 144: nth: should it be camel case? `noCache`

frontend/svelte/src/lib/stores/neurons.store.ts
peterpeterparker, line 24: if we would use a `Set` (see previous comment), maybe we can also use one here?

should we extract it to a function to make the goal of this filter obvious? if no, comment?

peterpeterparker (changes requested):

frontend/svelte/src/lib/api/governance.api.ts
lmuntanerDfinity: I like the idea of the constants.

Set won't work, objects will be different objects with the same data:

```
Array.from(new Set([{}, {}, {}])) // still an array with 3 empty elements
```

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
lmuntanerDfinity: * I'll add the return type. I forgot about it.
* I guess it could lead to problems with thousands of KnownNeurons. By then we'll have the new design (I hope).

What are the options?
* Use the `followees.find` directly in the template? I wanted to make it more readable and that would still raise the same performance problems I guess.
* Pass the `topicFollowees` to the `KnownNeuronFollowItem` and have the component do the `find`? That would still have the same performance problem, right?
* Am I missing something else?

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: 👍

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: yep, good point

frontend/svelte/src/lib/api/governance.api.ts
peterpeterparker:
```
[...new Set([knowNeurons, dfinityNeuron, icNeuron].map({id} =>{id}))]
```

https://stackoverflow.com/a/35092559

frontend/svelte/src/lib/stores/neurons.store.ts
lmuntanerDfinity: The `Set` doesn't work, but I like the helper.

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
peterpeterparker: 😉

frontend/svelte/src/lib/api/governance.api.ts
lmuntanerDfinity: Is this clearer than pussing? don't I need to map again back to the objects?

```
[...new Set([knowNeurons, dfinityNeuron, icNeuron].map({id} =>{id}))].map((id) => neuronDict[id])
```

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
lmuntanerDfinity: It's a good point, I think it's something worth researching and adding to our Code Conventions.

frontend/svelte/src/lib/components/neurons/KnownNeuronFollowItem.svelte
lmuntanerDfinity: good point!

frontend/svelte/src/lib/components/neurons/KnownNeuronFollowItem.svelte
lmuntanerDfinity: yes

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: this is a success toast

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: ok

frontend/svelte/src/lib/stores/neurons.store.ts
lmuntanerDfinity: Sorry, I am not sure I follow, could you write it all? You call it newIds, and then use the `oldNeurons`. And do a Set of the old array, which should already be unique.

lmuntanerDfinity:

frontend/svelte/src/lib/api/governance.api.ts
peterpeterparker: To me yes because as soon as I read `[...new Set([` I already understood that this has for goal to remove duplicates

But you are right that's personal. I can live with both, as you rather like.

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
peterpeterparker: If you find anything please let me know!

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: Woot? within a ` } catch (error) {` block?

lmuntanerDfinity: Github is not showing the line properly. Check the changes directly.

frontend/svelte/src/lib/api/governance.api.ts
lmuntanerDfinity: something also for the code conventions

frontend/svelte/src/lib/stores/neurons.store.ts
mstrasinskis: Sorry, I mean something like:
```ts
update((oldNeurons: NeuronInfo[]) => {
  const oldIds = new Set(oldNeurons.map(({ neuronId }) => neuronId));
  return [
    ...oldNeurons,
    ...newNeurons.filter(({ neuronId }) => !oldIds.has(neuronId)),
  ];
});
```
To avoid kind of O(mn) complexity.

lmuntanerDfinity: done!

PR #607, opened 2022-03-23
feat: query and update calls to sync accounts

peterpeterparker:
# Motivation

We want to use query and update calls to sync the accounts as well in order to improve the performances.

# Changes

- use `queryAndUpdate` to `syncAccounts`

# Note

It will be faster but it will still takes a couple of seconds to display the query results. In my local test, instead of 12s the query calls took 4s.


mstrasinskis (approved):
LGTM

PR #609, opened 2022-03-23
feat: refactor get identity for services and logout if no identity is provided

peterpeterparker:
# Motivation

- Refactor the way services access the identity - i.e. instead of passing the identity all the way down from the component, access to the identity on a `services` level.

- Solve the many todo that were open to display properly the error message for the edge case of the identity becoming null or undefined

- With the same approach of an auth becoming idle, logout the user automatically if a function of a service is queried but for some reason no identity was provided

# Note

This is an edge case. The auth - i.e. the identity - is guarded both in terms of session timeout and storage changes. Therefore in a normal behavior mode the automatic signout should already happen even before any call to a service.

# Changes

- add new function `getIdentity` that either return a valid `Identity` or proceed with a `logout`
- remove all `// TODO !identity throw error`
- remove usage of `$authStore.identity` for the services in the component
- refactor and rename `api.utils.ts` to `utils.services.ts` (see next point)
- get the identity from the store directly within `queryAndUpdate` function 
- use `getIdentity` for the update calls (that are not read operations)

# Tests

This new `getIdentity` function is now mocked globally, so it might be possible that we would be able to spare some mocks for the components. There are already quite some changes in this PR, therefore did not dig deeper on that matter.

I did no provided particular test for `getIdentity` because it is already covered by existing test and `logout` itself is already covered by its corresponding tests too.

# Screenshots

Simulated edge case UX.

ezgif com-gif-maker(1)

### DIFF
--- a/frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
+++ b/frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
@@ -16,7 +16,7 @@
   let followeeAddress: number | undefined;
   let loading: boolean = false;
 
-  onMount(async () => await listKnownNeurons());
+  onMount(() => listKnownNeurons());
 
   const addFolloweeByAddress = async () => {
     loading = true;
--- a/frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
+++ b/frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
@@ -5,7 +5,6 @@
   import Spinner from "../../components/ui/Spinner.svelte";
   import { listKnownNeurons } from "../../services/knownNeurons.services";
   import { addFollowee } from "../../services/neurons.services";
-  import { authStore } from "../../stores/auth.store";
   import { i18n } from "../../stores/i18n";
   import { sortedknownNeuronsStore } from "../../stores/knownNeurons.store";
   import { toastsStore } from "../../stores/toasts.store";
@@ -17,11 +16,7 @@
   let followeeAddress: number | undefined;
   let loading: boolean = false;
 
-  onMount(() => {
-    listKnownNeurons({
-      identity: $authStore.identity,
-    });
-  });
+  onMount(async () => await listKnownNeurons());
 
   const addFolloweeByAddress = async () => {
     loading = true;
@@ -38,7 +33,6 @@
       return;
     }
     await addFollowee({
-      identity: $authStore.identity,
       neuronId: neuron.neuronId,
       topic,
       followee,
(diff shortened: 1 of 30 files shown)


frontend/svelte/src/lib/services/utils.services.ts
```diff
@@ -12,20 +15,22 @@ export type QueryAndUpdateOnError<E> = (options: {
...
  * The returned promise notify when first fetched data are available.
  */
-export const queryAndUpdate = <R, E>({
+export const queryAndUpdate = async <R, E>({
   request,
   onLoad,
   onError,
 }: {
-  request: (options: { certified: boolean }) => Promise<R>;
+  request: (options: { certified: boolean; identity: Identity }) => Promise<R>;
```
mstrasinskis, line 23: Just because `params` is used everywhere as an argument name probably it make sense to rename it here too.

mstrasinskis (approved):
LGTM. A couple of suggestions but looks much cleaner now. Thanks!

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
```diff
@@ -17,11 +16,7 @@
   let followeeAddress: number | undefined;
   let loading: boolean = false;
 
-  onMount(() => {
-    listKnownNeurons({
-      identity: $authStore.identity,
-    });
-  });
+  onMount(async () => await listKnownNeurons());
```
lmuntanerDfinity, line 19: no need for `async` nor `await`

lmuntanerDfinity (approved):
Agree, much nicer. Minor comment, already approving.

frontend/svelte/src/lib/services/auth.services.ts
```diff
@@ -23,6 +25,28 @@ export const logout = async ({
...
+  /* eslint-disable-next-line no-async-promise-executor */
+  return new Promise<Identity>(async (resolve) => {
+    const identity: Identity | undefined | null = get(authStore).identity;
+
+    if (!identity) {
+      await logout({
+        msg: { labelKey: "error.missing_identity", level: "error" },
+      });
+
+      // We do not resolve on purpose. logout() does reload the browser
```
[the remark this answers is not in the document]
peterpeterparker: maybe yes. I had a first version with throw error but it actually was glichty so I went with a "safe is safe" approach. Do you mind if we keep the `return`?

frontend/svelte/src/lib/services/utils.services.ts
peterpeterparker: My bad. I should have use `options` has defined. Done. Replaced all `params` I used for `options`. Thx!

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
peterpeterparker: I don't like these promises that are imperatively executed, feels wrong in my mind. As you wish. I removed these here. Done.

lmuntanerDfinity: something for the guidelines!

PR #611, opened 2022-03-24
L2 369 switch to promise all settled for the cast vote function

mstrasinskis:
# Motivation

To improve error handling we switch from Promise.all to Promise.allSettled. Display individual errors (one per failure neuron update) in detail of the error message.
[L2-369](https://dfinity.atlassian.net/browse/L2-369)

# Changes

- display individual neuron errors (in detail)
- listNeurons error handling (refresh)
- reorganisation of register vote tests (error blocks are combined in single describe block)

# Tests

- proposal service

# Notes

@peterpeterparker Probably it makes sense to change the detail rendering style. So it doesn't look like a part of the error message.

# Screenshots

image

**empty error message fallback**
image

### DIFF
--- a/frontend/svelte/src/lib/services/proposals.services.ts
+++ b/frontend/svelte/src/lib/services/proposals.services.ts
@@ -225,7 +225,7 @@ const requestRegisterVotes = async ({
   }> = responses
     // add neuronId
     .map((result, i) => ({ neuronId: neuronIds[i], result }))
-    // use only not empty errors
+    // handle only not-empty errors
     .filter(
       ({ result }) =>
         result.status === "rejected" && result.reason instanceof Error
@@ -242,7 +242,6 @@ const requestRegisterVotes = async ({
     const detail: string = errors
       .map(({ neuronId, error }) => {
         const reason = errorToString(error);
-        console.log(typeof reason, error?.["message"], get(i18n).error.fail);
         return replacePlaceholders(get(i18n).error.register_vote_neuron, {
           $neuronId: neuronId.toString(),
           $reason:
--- a/frontend/svelte/src/lib/services/proposals.services.ts
+++ b/frontend/svelte/src/lib/services/proposals.services.ts
@@ -180,11 +180,8 @@ export const registerVotes = async ({
       identity,
     });
   } catch (error) {
-    // TODO: replace w/ console.error mock
-    if (!isNode()) {
-      // preserve in unit-test
-      console.error("vote unknown:", error);
-    }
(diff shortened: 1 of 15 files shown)


frontend/svelte/src/lib/services/proposals.services.ts
```diff
@@ -182,51 +186,53 @@ const requestRegisterVotes = async ({
...
-    .map(({ detail }: GovernanceError) =>
-      stringifyJson(detail.error_message, { indentation: 2 })
+  const errors: Array<{
+    neuronId: bigint;
+    error: Error;
+  }> = responses
+    // add neuronId
+    .map((result, i) => ({ neuronId: neuronIds[i], result }))
+    // handle only not-empty errors
+    .filter(
```
peterpeterparker, line 207: not sure but wouldn't it be possible to `filter` first (if I get it right)? then you can save a `map` and therefore better performance?

frontend/svelte/src/lib/services/proposals.services.ts
```diff
@@ -182,51 +186,53 @@ const requestRegisterVotes = async ({
...
   if (errors.length > 0) {
-    // TODO: replace w/ console.error mock
-    if (!isNode()) {
-      // avoid in unit-test
-      console.error("vote:", errorDetails);
-    }
+    console.error("vote", errors);
+
+    // neuronId next to the error text
+    const detail: string = errors
```
peterpeterparker, line 220: a bit like in the same direction as above, maybe we can integrate this is previous chain - do everything in one single chain?

like

```
 responses.
     .filter( .... only those which are errors)
     .map(... => {
         const neuronId = ...
         const reason = ...
         return replacePlaceholder...
```

frontend/svelte/src/lib/utils/error.utils.ts
```diff
@@ -1,6 +1,10 @@
+import { GovernanceError } from "@dfinity/nns";
+
 export const errorToString = (err?: unknown): string | undefined =>
   typeof err === "string"
     ? (err as string)
+    : err instanceof GovernanceError
```
peterpeterparker, line 6: Not related to the PR: should we add a `NNSJsError` abstract type and all errors of the lib should extend it? something we can discuss, just an idea I had reading these lines.

frontend/svelte/src/lib/services/proposals.services.ts
peterpeterparker: Ah sh*t you're right. Thanks for the feedback

PR #613, opened 2022-03-25
Tests may use any version of chrome

bitdivine:
# Motivation
The e2e tests should work with the latest few versions of chrome.  The major version of chrome changes every 2 weeks.  The package.json should be stable because it encodes our expectations of what we want to test.  The lockfile is there to fix the specific choice that matches our expectations.  Therefore `package.json` should have something like `latest` or maybe `*`.  The lockfile should be bumped with a newer chromedriver every 2 weeks.

## Alternatives
chromedriver could be taken from the installed chrome version, rather than npm.

# Changes
- Changed the chromedriver dependency from 98 to latest.
  - Note: The latest version at the time of creating this PR is ~99~ now 100.
  - Note: I would also be 100% fine with `*` as we will want to test recent, non-latest versions of chrome.  These tests are likely to be local rather than in CI.
- Added updating chromedriver to the nightly job.

# Next steps - not on this PR
- Testing a matrix of different browsers in CI

# Tests
With the latest chrome installed (v99):
```
# Make sure we have a clean repo
git checkout .
# Install packages per the lockfile
npm ci
# Install the latest chromedriver
npm install chromedriver
# Check the chromedriver version
./node_modules/.bin/chromedriver --version
ChromeDriver 99.0.4844.51 (d537ec02474b5afe23684e7963d538896c63ac77-refs/branch-heads/4844@{#875})
# Test
NNS_DAPP_URL="$(curl -v "https://nns-svelte.winning.gold" 2>&1 | awk '/location:/{print $3}')" npm run wdio
```

## Test nightly
I pushed this code on the nightly branch, which is used to test the nightly job.  [Here is the github action it triggered.](https://github.com/dfinity/nns-dapp/actions/runs/2088719563)


bitdivine:
Update:  I have now also bumped the version of chromedriver in the package.lock as otherwise CI fails, as chrome v100 is installed on CI but chromedriver was at version 98.

lmuntanerDfinity (approved):

PR #614, opened 2022-03-25
L2-352: NeuronMetaInfoCard UI

lmuntanerDfinity:
# Motivation

UI for the NeuronMetaInfoCard in Neuron Details page.

# Changes

* NeuronMetaInfoCard html and css.
* Move `formatVotingPower` to neuron utils.
* Expand NeuronCard to manage child elements.
* New utils regarding voting power calculations.
* New type button: `warning`.

# Tests

* Test NeuronMetaInfoCard.
* Test utils.

### DIFF
--- a/frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
+++ b/frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
@@ -8,7 +8,7 @@
   import {
     ageMultiplier,
     dissolveDelayMultiplier,
-    votingPower,
+    formatVotingPower,
   } from "../../utils/neuron.utils";
   import NeuronCard from "../neurons/NeuronCard.svelte";
   import Tooltip from "../ui/Tooltip.svelte";
@@ -20,7 +20,8 @@
   <section>
     <div class="space-between">
       <p>
-        {secondsToDate(neuron.createdTimestampSeconds)} - {$i18n.neurons.staked}
+        {secondsToDate(Number(neuron.createdTimestampSeconds))} - {$i18n.neurons
+          .staked}
       </p>
       <button class="primary small"
         >{$i18n.neuron_detail.join_community_fund}</button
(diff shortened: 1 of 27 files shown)


lmuntanerDfinity:
Screenshot:

Screenshot 2022-03-25 at 09 02 53

Responsiveness:

Screenshot 2022-03-25 at 09 15 29

Alignment of bottom elements:
Screenshot 2022-03-25 at 09 03 04

lmuntanerDfinity:

lmuntanerDfinity:
We might want to have a UI check with Mischa. Although in another PR.

I followed the same alignment and styles as the NeuronCard for consistency sake. But it does not align 100% with current card.

frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
```diff
@@ -1,11 +1,126 @@
...
+              )}
+            >
+              <span>
+                <IconInfo />
+              </span>
+            </Tooltip>
+          {/if}
+        {/if}
+      </p>
+      <div class="buttons">
```
peterpeterparker, line 60: This component has quite a size, would it be cool to move the buttons to a dedicated components?

frontend/svelte/src/lib/utils/date.utils.ts
```diff
@@ -43,3 +43,14 @@ export const secondsToDuration = (seconds: bigint): string => {
...
+
+export const secondsToDate = (seconds: number): string => {
+  const options: Intl.DateTimeFormatOptions = {
+    month: "long",
+    day: "numeric",
+    year: "numeric",
+  };
+  const millisedoncs = seconds * 1000;
+  // We only support english for now.
+  return new Date(millisedoncs).toLocaleDateString("en", options);
```
peterpeterparker, line 55: any particular reason why not using the `new Intl.DateTimeFormat('en-US').format(date)` api?

peterpeterparker (approved):
A couple of comments just in case there would be useful.

Agree we should review the ui with Mischa!

frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
lmuntanerDfinity: Actually, I was thinking about having a folder `actions` inside this `neuron-detail` to add all the buttons. Specially because each button can come with a modal and functionality. I can already do this in this PR, makes sense.

frontend/svelte/src/lib/utils/date.utils.ts
lmuntanerDfinity: It does not follow the same format as in the UI. The month for example should be the name, not number.

peterpeterparker: Maybe something also worth to ask Mischa? Thanks for the feedback

PR #615, opened 2022-03-25
json viewer for proposal payload data

anishh2003:
# Motivation

Proposal payload data is in an unreadable format.

# Changes

JsonViewer widget is now being used to show the json data 
Widget code copied from : https://pub.dev/packages/flutter_json_viewer
The widget has not been put in directly as a dependancy , since the colour , text parameters are not passed as arguments. Hence it has been copied and changes have been made to the colour parameters for the NNSDapp

### DIFF
--- a/frontend/dart/lib/ui/widgets/flutter_json_viewer.dart
+++ b/frontend/dart/lib/ui/widgets/flutter_json_viewer.dart
@@ -1,0 +1,356 @@
+library flutter_json_widget;
+
+import 'package:flutter/material.dart';
+
+class JsonViewer extends StatefulWidget {
+  final dynamic jsonObj;
+  JsonViewer(this.jsonObj);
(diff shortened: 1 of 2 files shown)


frontend/dart/lib/ui/widgets/flutter_json_viewer.dart
```diff
@@ -0,0 +1,356 @@
+library flutter_json_widget;
```
hpeebles, line 1: Is this copied from somewhere or did you write it?
If you copied it from somewhere you should put a comment at the top of the file saying where it is taken from.

hpeebles: You haven't pushed that comment... its still not there

hpeebles (approved):

PR #626, opened 2022-03-29
feat: disable and busy state on create subaccounts

peterpeterparker:
# Motivation

When subaccounts are being created, it should not be possible to update the input field and to close and navigate in the modal.

# Changes

- set `disabled` to modal actions on busy state
- set `disabled` to add account input on busy state
- use busy store to display a spinner above everything when a sub account is created
- add "accounts" to type of busy store
- add a readonly derived store to access the information `busy`
- refactor busy store to only exposes functions `startBusy`  and `stopBusy`
- few styles improvements to buttons and inputs to reflect the disabled status
- remove `test` from busy store - we now have more than one value for our tests

# Note

I noticed we should enhance the busy screen with a catch on keyboard events - i.e. we don't want user to be able to tab and enter navigation in the background. Will try to solve this in another PR.

# Screenshots

Capture d’écran 2022-03-29 à 10 12 56


frontend/svelte/src/lib/stores/busy.store.ts
```diff
@@ -31,4 +32,11 @@ const initBusyStore = () => {
   };
 };
 
-export const busyStore = initBusyStore();
+const busyStore = initBusyStore();
+
+export const { startBusy, stopBusy } = busyStore;
+
+export const busy: Readable<boolean> = derived(
```
lmuntanerDfinity, line 39: derived stores are the future!

lmuntanerDfinity (approved):

PR #627, opened 2022-03-29
Flutter and Svelte in parallel

bitdivine:
# Motivation
When testing Svelte it is useful to be able to compare with flutter.  At the moment there are several deployments, including prod and staging.  However it is not enough.  Sometimes there is a strange state and it is useful to be able to examine that state on the same canister, so there need to be redirects from neither svelte nor flutter.

References:
- https://dfinity.atlassian.net/browse/L2-440

# Changes
- Defined a new value for `REDIRECT_TO_LEGACY`.
- Removed old values such as `1`
- Documented the values of `REDIRECT_TO_LEGACY`.

# Tests

Deployed with these values for `REDIRECT_TO_LEGACY`:
staging    -> https://q3fc5-haaaa-aaaaa-aaahq-cai.nnsdapp.dfinity.network == https://nns-staging.winning.gold
prod       -> https://wflfh-4yaaa-aaaaa-aaata-cai.nnsdapp.dfinity.network == https://nns-prod.winning.gold
svelte     -> https://xifbj-tqaaa-aaaaa-aaauq-cai.nnsdapp.dfinity.network == https://nns-svelte.winning.gold
flutter    -> https://xta5m-jiaaa-aaaaa-aaawa-cai.nnsdapp.dfinity.network == https://nns-flutter.winning.gold
both       -> https://x2dwq-7aaaa-aaaaa-aaaxq-cai.nnsdapp.dfinity.network == https://nns-both.winning.gold
Added memorable names for each.

Verified:

### Staging
- Logged in, landed on the flutter accounts page.
- Tabbed around and got all tabs in flutter barring the proposals page.
- Added/removed v2 for each tab.  Was redirected back to the "correct" version in each case.

### Prod
- Verified that all pages are in flutter and that if I add v2 I am redirected to flutter.

### Svelte
- Logged in.  Verified that I land on the svelte accounts page.
- Verified that clicking on tabs, I always stay on svelte.
- Verified, for each tab, that going to the flutter equivalent I am redirected to svelte.

### Flutter
- Logged in.  Verified that I landed on th eflutter accounts page.
- Tabbed around.  Pure flutter.
- Verified that adding v2 result sin being redirected back to the equivalent flutter page.

### Both
- Logged in.  Verified that I landed on the svelte page.
- Verified that clicking on tabs always yields another svelte page.
- Removed the v2 and so ended up on a flutter page.
- Verified that clicking on tabs I always stayed on flutter.


deploy.sh
```diff
@@ -4,5 +4,5 @@ cd "$(dirname "$(realpath "$0")")" || exit
 NETWORK=$1
 
 echo "Deploying to $NETWORK..."
-DEPLOY_ENV=$NETWORK dfx deploy --no-wallet --network "$NETWORK"
+DEPLOY_ENV=$NETWORK dfx deploy --no-wallet --network "$NETWORK" nns-dapp
```
bitdivine, line 7: The ii canister is not part of this repo, so should not be deployed by default.  It can always be deployed explicitly if wanted.

frontend/svelte/env.config.mjs
```diff
@@ -44,21 +44,9 @@ const GOVERNANCE_CANISTER_URL = `https://${GOVERNANCE_CANISTER_ID}${domain}/`;
 const LEDGER_CANISTER_URL = `https://${LEDGER_CANISTER_ID}${domain}/`;
 const OWN_CANISTER_URL = `https://${OWN_CANISTER_ID}${domain}/`;
```
bitdivine, line 46: I have removed values with confusing values, such as "1".  Once that is done, this rewriting code does nothing and can be deleted.

deploy.sh
lmuntanerDfinity: Maybe add a comment that we have other dapps in `dfx.json` and that's why we need to specify this.

deploy.sh
```diff
@@ -4,5 +4,5 @@ cd "$(dirname "$(realpath "$0")")" || exit
 NETWORK=$1
 
 echo "Deploying to $NETWORK..."
```
lmuntanerDfinity, line 6: `echo "Deploying NNS Dapp to $NETWORK..."`?

lmuntanerDfinity (changes requested):
Not sure where we want this, but we also need to take care of the `npm run dev` during local development.

Either we add `REDIRECT_TO_LEGACY=svelte` to the package.json `dev` script.

Or you add a ternary in `env.config.mjs`:

```
const REDIRECT_TO_LEGACY = ENVIRONMENT === "local"
  ? "svelte "
  : process.env.REDIRECT_TO_LEGACY || "prod";
```

frontend/svelte/src/routes/ProposalDetail.svelte
```diff
@@ -25,7 +25,7 @@
   let neuronsReady = false;
 
   // TODO: To be removed once this page has been implemented
-  const showThisRoute = ["never", "staging"].includes(
+  const showThisRoute = ["svelte", "both", "staging"].includes(
```
peterpeterparker, line 28: 1. why `["svelte", "both", "staging"]` and sometimes only `["svelte", "both"]`?

2. can you move the duplicate code declaration to the constants?

something like

```
// constant
export const SHOW_ROUTE = ["svelte", "both", "staging"].includes(    process.env.REDIRECT_TO_LEGACY as string
  );
  );

// Components - routes
onMount(() => {
    if (!SHOW_ROUTE) {
      window.location.replace(`/${window.location.hash}`);
    }
  });
```

peterpeterparker (changes requested):

deploy.sh
bitdivine: Regarding your comment about local deployment, a similar consideration may apply here.  We could check `DEV_ENV` and deploy all canisters for `local`.  What would make life easiest for those people who mostly work locally?

frontend/svelte/src/routes/ProposalDetail.svelte
peterpeterparker: Thanks for the feedback, works for me.

Maybe add a comment about the `staging` value in that particular piece of code, pretty sure I'll forget about it 😉

peterpeterparker (approved):

deploy.sh
bitdivine: Per discussion, such logic is not required.  A comment is enough.

deploy.sh
bitdivine: Done.

deploy.sh
bitdivine: Done: Comment added.

lmuntanerDfinity (approved):
💯

PR #628, opened 2022-03-29
Fix: change II curl URL

lmuntanerDfinity:
# Motivation

URL used to get II wasm for e2e test is returning 404.

# Changes

* Change URL where we get II wasm.

# Tests

No extra tests

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -11,7 +11,7 @@
       "type": "custom",
       "wasm": "internet_identity.wasm",
       "candid": "internet_identity.did",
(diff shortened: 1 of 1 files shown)


nmattia (approved):
LGTM!

peterpeterparker:
Shouldn't we download the latest to avoid to have to provide new PR to set the version again?

> https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity_dev.wasm

lmuntanerDfinity:
good question @peterpeterparker , @frederikrothenberger @nmattia what do you think?

nmattia:
No, because the latest is always changing. If Internet Identity pushes some new flow that breaks your tests, then your tests will start failing on all your PRs with no apparent reason.

PR #632, opened 2022-03-30
Nightly

bitdivine:
# Motivation
We wish to test against the latest published version of the Internet Identity, but to version control which version we are using so that test results are reproducible.

# Changes
- Add a script that updates the version of II in `dfx.json`
- Add a nightly job that gets the latest II and commits the updated version if there is a change.

# Tests
- I have run `./scripts/update-ii` locally and verified that it works as expected.
- I have made a whitespace change to dfx.json to trigger a commit from the nightly job.  Pushing the nightly branch triggers the job.
  - And it created this commit: https://github.com/dfinity/nns-dapp/pull/632/commits/ff568d48e656f1c4bf221b1b9ca65c4be2506385


dfx.json
```diff
@@ -11,7 +11,7 @@
       "type": "custom",
       "wasm": "internet_identity.wasm",
       "candid": "internet_identity.did",
-      "build": "curl -sSL https://github.com/dfinity/internet-identity/releases/download/release-initial/internet_identity_dev.wasm -o internet_identity.wasm"
+      "build": "curl   -sSL \"https://github.com/dfinity/internet-identity/releases/download/release-2022-03-30/internet_identity_dev.wasm\" -o internet_identity.wasm"
```
lmuntanerDfinity: rebase from `main`, I added also the `--fail` to help with debugging.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Update II
+        run: ./scripts/update-ii
```
bitdivine, line 19: The commit is made with the same tool as used by the formatter.  Arguably using git directly is enough but it hasn't been smooth sailing when committing git notes so I'd rather use the tool until I understand what is happening there.

dfx.json
```diff
@@ -11,7 +11,7 @@
       "type": "custom",
       "wasm": "internet_identity.wasm",
```
bitdivine, line 12: I am likely to force various changes in here for test purposes, any changes SHOULD be undone by the gh action if you wait....

scripts/update-ii
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: Script to update URL, with some extreme caution as it executes a script with a URL pulled from over the super-trustworthy interwebs.

lmuntanerDfinity (approved):
LGTM

dfx.json
bitdivine: Already done, but thank you for catching it!

bitdivine:
@nmattia As it's your day off I won't wait to merge but if you have any feedback I'm happy to discuss it when you are back.

.github/workflows/nightly.yaml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  update-ii:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Update II
+        run: ./scripts/update-ii
+      - name: Commit updated Internet Identity - will not trigger rebuild
```
nmattia, line 20: I'd suggest not doing this; rather I would run the tests, and only notify you if they fail. By always updating II you're creating a lot of noise for no added benefit.

scripts/update-ii
```diff
@@ -0,0 +1,40 @@
...
+set -euo pipefail
+
+: "Getting URL from latest redirect..."
+II_URL="$(
+  set +o pipefail # We truncate the curl output, this is fine.
+  curl -v https://github.com/dfinity/internet-identity/releases/latest/download/internet_identity_dev.wasm 2>&1 |
```
nmattia, line 7: it's usually a bit brittle to use both stderr and stdout as output, though I can't quite figure out what you're doing with it here... Maybe worth using a mix of `-o` and `-w`?

scripts/update-ii
```diff
@@ -0,0 +1,40 @@
...
+    awk '(($1 == "<") && ($2 == "location:")){print $3; exit 0}' |
+    tr -d '[:space:]'
+)"
+echo "II_URL: ${II_URL}"
+
+: "Sanity checking URL..."
```
nmattia, line 13: you're doing a lot of URL cleaning and parsing and checking; wouldn't you be better off just querying the release API? should give you what you want as long as you trust GitHub and SSL:

```
$ curl -H "Accept: application/vnd.github.v3+json" https://api.github.com/repos/dfinity/internet-identity/releases | jq -cMr 'sort_by(.published_at) | reverse | .[0] | .assets | .[] | .browser_download_url'
https://github.com/dfinity/internet-identity/releases/download/release-2022-03-30/internet_identity_dev.wasm
https://github.com/dfinity/internet-identity/releases/download/release-2022-03-30/internet_identity_production.wasm
https://github.com/dfinity/internet-identity/releases/download/release-2022-03-30/internet_identity_test.wasm
```

scripts/update-ii
```diff
@@ -0,0 +1,40 @@
...
+  printf "Refusing to use URL containing chars: '%q'\n" "${unexpected_chars}"
+  exit 1
+}
+[[ "$II_URL" = https://github.com/dfinity/internet-identity/releases/* ]] || {
+  echo "Refusing to use URL from unexpected source."
+  exit 1
+}
+
+: "Defining build script..."
+export II_BUILD="$(printf 'curl --fail -sSL "%s" -o internet_identity.wasm' "$II_URL")"
```
nmattia, line 25: why `export`?

nmattia (commented):
General notes: there's a lot of `printf`s and overloading `:` to act as comment which is a bit confusing

.github/workflows/nightly.yaml
bitdivine: What noise?  This will produce one commit per II update, so about one per month.

scripts/update-ii
bitdivine: `stdout` and `stderr` are not being mixed.  Stdout is being thrown away. `-o` and `-w` aren't useful in determining the stable URL.

scripts/update-ii
bitdivine: I'm alreday pulling from github over ssl, so no, I don't consider that to be enough to be safe if the result is being dropped into a shell script.

scripts/update-ii
bitdivine: `II_BUILD` is unavailable in `jq` unless exported.

bitdivine:
Re general notes:
- Why is `printf` confusing?  `printf` is better defined than `echo`, so where a precise output is expected it is preferable.
- `:` is not being overloaded.  It is serving one purpose only:  To print messages when run in debug mode:
  -  `echo something` will always print
  - `: something` will print only when we have `set -x`

PR #634, opened 2022-03-30
feat: wizard modal

peterpeterparker:
# Motivation

1. Make it easier to create modals that are wizard
2. Remove duplicate code
3. Clean up style - i.e. align modal spacing and height style across all existing modals that use the wizard
4. Fix transition glitch when the wizard modal is opened and components are loading

# Changes

- new component `WizardModal` (mainly a refactoring of existing usage of the wizard)
- adapt existing Neuron and Account modal for new `WizardModal` and remove duplicate code
- refactor duplicate usage of `Transition` component to a single usage
- use `introend` to trigger loading and fix transition glitch
- move `StepState` from services to store and `step.state`
- suppress usage of enums for the wizard steps and use object instead that can contain meta information (such as `displayBackButton`)
- fix possible slow transition by navigating the wizard always and only based on the index - defer logic that checks if this or that is loaded (like "only if neuron is defined") within component and display a `Spinner` until loaded
- clean up style and spacing of the existing Neuron and Accounts wizard modal for both content that present buttons and list
- style the display of the modal back button with a small animation
- fix style of the section that is use in modal
- improve semantic by replacing top `article` that contain `article` within the modal with a top `section`

# Screenshots

Capture d’écran 2022-03-30 à 15 20 37
Capture d’écran 2022-03-30 à 15 20 44
Capture d’écran 2022-03-30 à 15 20 51
ms
Capture d’écran 2022-03-30 à 15 21 08
Capture d’écran 2022-03-30 à 15 21 13
Capture d’écran 2022-03-30 à 15 21 22
Capture d’écran 2022-03-30 à 15 21 41
Capture d’écran 2022-03-30 à 15 21 44
Capture d’écran 2022-03-30 à 15 21 56
Capture d’écran 2022-03-30 à 15 22 00


frontend/svelte/src/lib/components/accounts/SelectAccount.svelte
```diff
@@ -1,36 +1,48 @@
...
 
   const dispatch = createEventDispatcher();
   const chooseAccount = (selectedAccount: Account) => {
     dispatch("nnsSelectAccount", { selectedAccount });
   };
-  let { main: mainAccount } = $accountsStore;
+
+  let mainAccount: Account | undefined;
+
+  const unsubscribeAccounts: Unsubscriber = accountsStore.subscribe(
```
lmuntanerDfinity, line 17: Why this change? What issue was there with the `$accountsStore`?

frontend/svelte/src/lib/components/accounts/SelectTypeAccount.svelte
```diff
@@ -8,7 +8,7 @@
   };
 </script>
 
-<div class="wrapper">
+<div class="wizard-wrapper">
```
lmuntanerDfinity, line 11: This class is not used, can you remove it?

frontend/svelte/src/lib/components/neurons/ConfirmDissolveDelay.svelte
```diff
@@ -34,59 +40,60 @@
...
-    >
-      {#if loading}
-        <Spinner />
-      {:else}
-        {$i18n.neurons.confirm_delay}
-      {/if}
-    </button>
-  </div>
-</section>
+<div class="wizard-wrapper" data-tid="confirm-dissolve-delay-container">
```
lmuntanerDfinity, line 43: ditto, unused class

frontend/svelte/src/lib/components/neurons/EditFollowNeurons.svelte
```diff
@@ -4,29 +4,28 @@
...
 
-<section data-tid="edit-followers-screen">
-  <p>{$i18n.follow_neurons.description}</p>
-  <div>
-    {#each topics as topic}
-      <FollowTopicSection {neuron} {topic} />
-    {/each}
-  </div>
-</section>
+<div class="wizard-list" data-tid="edit-followers-screen">
```
lmuntanerDfinity, line 14: class not used

frontend/svelte/src/lib/components/ui/Transition.svelte
```diff
@@ -1,25 +1,33 @@
...
+      : -absolutOffset;
 </script>
 
-<div
-  bind:clientWidth={absolutOffset}
-  in:fly={{ x: slideOffset, duration: ANIMATION_DURATION }}
->
-  <slot />
-</div>
+{#key transition}
```
lmuntanerDfinity, line 18: cool!

frontend/svelte/src/lib/components/ui/Transition.svelte
```diff
@@ -1,25 +1,33 @@
 <script lang="ts">
   import { fly } from "svelte/transition";
 
-  export let diff: number;
+  export let transition: { diff: number } = { diff: 0 };
```
lmuntanerDfinity, line 4: why use an object instead of a number? To trigger a new transition?

Maybe add a comment, otherwise is not very clear why we use an object.

frontend/svelte/src/lib/modals/Modal.svelte
```diff
@@ -46,6 +47,7 @@
         <div class="toolbar">
           {#if showBackButton}
             <button
+              transition:fade={{ duration: 150 }}
```
lmuntanerDfinity, line 50: can we use the constant `ANIMATION_DURATION`? or another constant?

frontend/svelte/src/lib/modals/WizardModal.svelte
```diff
@@ -0,0 +1,66 @@
...
+
+  let stepState: StepsState;
+  $: stepState = new StepsState(steps);
+
+  export let currentStepIndex: number;
+  let currentStep: Step | undefined;
+  $: ({ currentStepIndex, currentStep } = stepState);
+
+  let transition: { diff: number };
+  $: stepState, (transition = { diff: stepState.diff });
```
lmuntanerDfinity, line 17: do you need this `stepState, `? Doesn't Svelte understand it needs to recalculate everytime stepState changes?

frontend/svelte/src/lib/modals/accounts/AddAccountModal.svelte
```diff
@@ -1,57 +1,37 @@
...
-  <article>
-    {#if currentStep === Steps.SelectAccount}
-      <Transition {diff}>
-        <SelectTypeAccount on:nnsSelect={handleSelectType} />
-      </Transition>
+<WizardModal {steps} bind:currentStepIndex bind:this={modal} on:nnsClose>
+  <svelte:fragment slot="title">{$i18n.accounts.add_account}</svelte:fragment>
+
+  <svelte:fragment>
+    {#if currentStepIndex === stepIndex({ name: "SelectAccount", steps })}
```
lmuntanerDfinity, line 30: Is this more readable than `{#if currentStep === Steps.SelectAccount}`?

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
```diff
@@ -79,66 +62,46 @@
...
-</Modal>
+<WizardModal {steps} bind:currentStepIndex bind:this={modal} on:nnsClose>
+  <svelte:fragment slot="title">{$i18n.neurons?.[titleKey]}</svelte:fragment>
 
-<style lang="scss">
-  main {
-    padding: calc(3 * var(--padding));
-  }
-</style>
+  <!-- TODO: Manage edge case: https://dfinity.atlassian.net/browse/L2-329 -->
```
lmuntanerDfinity, line 71: Since you are handling the errors showing a spinner, we can remove this comment and resolve this issue.

Or at least remove it from here.

frontend/svelte/src/lib/modals/WizardModal.svelte
```diff
@@ -0,0 +1,66 @@
...
+
+<style lang="scss">
+  section {
+    height: min(500px, calc(100vh - 156px - (2 * var(--padding))));
+    margin: 0;
+    padding: calc(2 * var(--padding));
+    max-width: 100%;
+  }
+
+  :global(.wizard-wrapper) {
```
lmuntanerDfinity, line 52: isn't this a little bit dark magic, that a component changes the style of another element by a class?

How are we going to remember in a few months how this works?

We also have warnings when a class is not used.

lmuntanerDfinity, line 52: isn't this a little bit dark magic, that a component changes the style of another element by a class?

How are we going to remember in a few months how this works?

We also have warnings when a class is not used.

frontend/svelte/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { render } from "@testing-library/svelte";
+import SetDissolveDelay from "../../../../lib/components/neurons/SetDissolveDelay.svelte";
+
+describe("SetDissolveDelay", () => {
```
lmuntanerDfinity, line 8: All the tests for the steps are in the `CreateNeuronModal.spec.ts`.

lmuntanerDfinity, line 8: All the tests for the steps are in the `CreateNeuronModal.spec.ts`.

lmuntanerDfinity (commented):
Some comments.

I remember a discussion about the Wizard that we resolved not to use the `wizard.next()` and `wizard.back`. I like it better this way, but I can't remember why we didn't do it this way before. I remember that I developed it first like that, and then we refactored it.

Maybe @mstrasinskis remembers why we refactored it. But I prefer it like this as well.

frontend/svelte/src/lib/modals/WizardModal.svelte
```diff
@@ -0,0 +1,66 @@
+<script lang="ts">
+  import { StepsState } from "../stores/steps.state";
+  import type { Steps, Step } from "../stores/steps.state";
+  import Modal from "./Modal.svelte";
+  import Transition from "../components/ui/Transition.svelte";
+
+  export let steps: Steps;
+
+  let stepState: StepsState;
+  $: stepState = new StepsState(steps);
```
lmuntanerDfinity, line 10: Do we need to `StepsState` now? Since it's always used through the component, doesn't it make more sense to just use a component state here?

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
```diff
@@ -79,66 +62,46 @@
...
-    {/if}
-    <!-- TODO: Manage edge case: https://dfinity.atlassian.net/browse/L2-329 -->
-    {#if currentStep === Steps.EditFolloNeurons && newNeuron}
-      <Transition {diff}>
-        <EditFollowNeurons neuron={newNeuron} />
-      </Transition>
-    {/if}
-  </main>
-</Modal>
+<WizardModal {steps} bind:currentStepIndex bind:this={modal} on:nnsClose>
```
mstrasinskis, line 68: I like the idea to bind state from the WizardModal. But can we bind `currentStepName` directly and get rid of `stepIndex` function?

frontend/svelte/src/lib/modals/WizardModal.svelte
mstrasinskis: We could link it more explicitly with the exported class name:
```svelte
<script context="module">
  export const WIZARD_WRAPPER = "wizard-wrapper";
</script>
```

```svelte
import { WIZARD_WRAPPER } from "../../modals/WizardModal.svelte";
<div class={WIZARD_WRAPPER}>
```

frontend/svelte/src/lib/stores/steps.state.ts
```diff
@@ -0,0 +1,47 @@
+export interface Step {
+  readonly name: string;
+  readonly showBackButton: boolean;
+}
+
+export type Steps = [Step, ...Step[]];
```
mstrasinskis, line 6: Nice idea!

mstrasinskis:
> 

It was done (probably) because there are situation(s) when the next step depends on the user input. So `next()` wouldn't help. But I also prefer to have next and handle the edge cases separately 👍

frontend/svelte/src/lib/components/accounts/SelectTypeAccount.svelte
peterpeterparker: See next comment.

frontend/svelte/src/lib/components/neurons/ConfirmDissolveDelay.svelte
peterpeterparker: See next comment.

frontend/svelte/src/lib/components/neurons/EditFollowNeurons.svelte
peterpeterparker: See next comment.

frontend/svelte/src/lib/modals/WizardModal.svelte
peterpeterparker: In `Modal.svelte` you've got the style related to the modal, that's why I added the style related to the `WizardModal.svelte`.

I can also add a `modal.scss` to the global theming if you rather like?

frontend/svelte/src/lib/components/ui/Transition.svelte
peterpeterparker: Comment added. Indeed an object is used to make svelte picks it needs to rerender even if diff change from 1 to 1 for example.

frontend/svelte/src/lib/modals/Modal.svelte
peterpeterparker: It's quite a unique case at the moment therefore I think it's fine with the value here. We do not have constant for the transition duration of our css rules neither. Being said, maybe something we can refactor globally in the future.

frontend/svelte/src/lib/modals/accounts/AddAccountModal.svelte
peterpeterparker: No but we need objects for the step. I could have use enum within the object but its a complexity I would like to spare therefore I used `string` type for the names. I could have use `currentStepIndex === 0` but then it makes the index fi that's why I added this dumb function.

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
peterpeterparker: Can you have a look once merged?

frontend/svelte/src/lib/modals/WizardModal.svelte
peterpeterparker: See other comment / discussion

frontend/svelte/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
peterpeterparker: I am aware of it. The spinner test is scoped to the component.

peterpeterparker: Same as previous comment

frontend/svelte/src/lib/modals/WizardModal.svelte
peterpeterparker: Good question, I though about it too. To some extension we can remove it and add the logic within the modal, on the other side, it is good to have the logic scoped in a dedicated state / store. Therefore kind of think it's fine to have it that way.

But also open to remove it, as you rather like. Something we can do after this PR is merged too.

frontend/svelte/src/lib/modals/accounts/AddAccountModal.svelte
peterpeterparker: Being said, I'll try Maks suggestion 👉 https://github.com/dfinity/nns-dapp/pull/634#discussion_r839274970

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
peterpeterparker: Done, it's cleaner, thx. I ticked because in neuron the title comes from a map but of course I could also refactor the map and then use name.

frontend/svelte/src/lib/components/accounts/SelectAccount.svelte
lmuntanerDfinity: ok, weird

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
lmuntanerDfinity: Yes

frontend/svelte/src/lib/modals/WizardModal.svelte
lmuntanerDfinity: yep, maybe later, but also I don't have a strong opinion.

frontend/svelte/src/lib/modals/WizardModal.svelte
peterpeterparker: I want to avoid `mixin` because it can multiply the css within the tree which we do not need here because the style is unique.

For example we don't mind having `<button class="primary"...`, it's the same approach.

I'll extract the style to a global theming style.

peterpeterparker: I've created a global `modal.scss` for such styles.

lmuntanerDfinity: Not a fan of `<button class="primary..." either 😅

But a discussion for another time.

mstrasinskis (approved):
LGTM 👍

PR #641, opened 2022-03-31
L2-355: Make Dummy Proposal

lmuntanerDfinity:
# Motivation

Testnet user can create dummy proposals.

# Changes

* Add method to make dummy proposals which are hardcoded in governance api.
* New service enabled only in testnet to make dummy proposals.
* New card in Neuron Details enabled only in testnet to trigger the creation of dummy proposals.

# Tests

No new tests were added because all the functionality is for testnet, not for production purposes.


frontend/svelte/src/lib/components/neuron-detail/NeuronProposalsCard.svelte
```diff
@@ -1,11 +1,29 @@
 <script lang="ts">
   import type { NeuronInfo } from "@dfinity/nns";
+  import { makeDummyProposals } from "../../services/neurons.services";
   import Card from "../ui/Card.svelte";
 
   export let neuron: NeuronInfo;
+
+  const makeProposals = () => {
+    makeDummyProposals(neuron.neuronId);
```
mstrasinskis, line 9: I know that it wasn't in flutter, but I already saw a couple of comments that it would be great to have some feedback if the function fails. What do you think about showing the toast message e.g. "Make Dummy Proposals fails. See dev console for more info"?

mstrasinskis (approved):
LGTM. There is one suggestion.

frontend/svelte/src/lib/components/neuron-detail/NeuronProposalsCard.svelte
lmuntanerDfinity: This is just for text, but I can add that in the service, makes sense for me.

mstrasinskis: Thank you :)

lmuntanerDfinity: done!

peterpeterparker:
@lmuntanerDfinity since already merged can you please improve following in a new PR:

- add / use `assertTestnet();` to `makeDummyProposals()` to be absolutely sure it is only use on testnet

in addition, it feels like we add lots of bytes in the prod bundle with the new code of `dummyProposals.utils.ts` that is not needed in production - i.e. since we do not use chunking it will unnecessary increase prod bundle size (that is already to big in my opinion) and per extension make the download of the js slower for no valid productive reason. so do we need this on staging or only locally would be enough? if local enough, then can we exclude it from the build? if staging needed, shall we add it as an external script like we do to markdown?  cc @mstrasinskis

PR #643, opened 2022-03-31
L2 449 hide open proposals issue

mstrasinskis:
# Motivation

https://dfinity.atlassian.net/browse/L2-449
Fix a couple of bugs:
- excludeVotedProposals filter logic
- do not fetch on excludeVotedProposals (with mocking)
- fix no data-tid support of FiltersButton

# Changes

- added lastAppliedFilter to proposalsFilterStore

# Tests

- excludeVotedProposals tests were updated


frontend/svelte/src/lib/components/ui/FiltersButton.svelte
```diff
@@ -7,11 +7,12 @@
    */
   export let totalFilters: number;
   export let activeFilters: number;
+  export let testId: string | undefined;
   const dispatch = createEventDispatcher();
   const filter = () => dispatch("nnsFilter");
 </script>
 
-<button on:click={filter} class="dark small"
+<button data-tid={testId} on:click={filter} class="dark small"
```
peterpeterparker, line 15: is `testId` or `data-tid`?

frontend/svelte/src/lib/components/ui/FiltersButton.svelte
```diff
@@ -7,11 +7,12 @@
    */
   export let totalFilters: number;
   export let activeFilters: number;
+  export let testId: string | undefined;
```
peterpeterparker, line 10: to be honest with you, not a big fan of adding props for test purpose only. 

wouldn't it work out too if we add the `data-id` to the `<div class="filters">`  container and then target `data-id="" > button` for the test or you rather like not too?

frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -77,7 +77,7 @@ export const hideProposal = ({
   return (
     excludeVotedProposals &&
     status === ProposalStatus.PROPOSAL_STATUS_OPEN &&
-    ballots.find(({ vote }) => vote !== Vote.UNSPECIFIED) !== undefined
+    ballots.find(({ vote }) => vote === Vote.UNSPECIFIED) === undefined
```
peterpeterparker, line 80: when there never was any vote, isn't `ballots` empty?

frontend/svelte/src/routes/Proposals.svelte
```diff
@@ -126,9 +140,11 @@
       <ProposalsFilters />
 
       <InfiniteScroll on:nnsIntersect={findNextProposals}>
-        {#each $proposalsStore as proposalInfo (proposalInfo.id)}
-          <ProposalCard {proposalInfo} />
-        {/each}
+        <div class:hidden>
```
peterpeterparker, line 143: this is already a container `div` within `ProposalCard`, can we apply the style there instead of adding another additionital div in the dom?

frontend/svelte/src/routes/Proposals.svelte
```diff
@@ -145,6 +161,11 @@
 {/if}
 
 <style lang="scss">
+  div.hidden {
+    visibility: hidden;
+    position: absolute;
```
peterpeterparker, line 166: does position absolute comply with the intersection observer of the infinite scroll component?

peterpeterparker (changes requested):
A couple of questions

frontend/svelte/src/lib/components/ui/FiltersButton.svelte
mstrasinskis: I'm also not a fan of test props only. But from my perspective it's also a dev props. So as developer I can faster find the right place in the DOM without guessing by structure or styling or clicking inspect. So personally I would prefer to have `data-id` for all the components but I also understand that it's quite brutal (not a compromise 😄)

Regarding this particular case if I got you right the suggestion is to use `nth-child`. It will work but to be honest I would prefer to use same approach as often as possible to also not confuse people during e2e tests writing.

Another advantage is a reusability. So most e2e test could be applied to the redesigned version w/o rewriting the selectors.

frontend/svelte/src/lib/utils/proposals.utils.ts
mstrasinskis: Hmm... Looks like all of them are in ballots but in state `UNSPECIFIED`
image

frontend/svelte/src/routes/Proposals.svelte
mstrasinskis: I guess we can but then we can't reuse the Spinner 😄
What in your opinion is a better approach?
1. remove displaying the spinner (apply visibility style to ProposalCard entries)
2. remove the fetching mock (instantly apply the filter)
3. do real fetch on checkbox change

frontend/svelte/src/routes/Proposals.svelte
mstrasinskis: Currently it doesn't trigger the intersection event but agree It's smells a bit.

frontend/svelte/src/lib/components/ui/FiltersButton.svelte
lmuntanerDfinity: Not a big fan of test props, but I think that in some cases it might make sense. I would like to avoid relying in the DOM too much for e2e testing and having to redo all the e2e testing.

If we follow this approach, we might even be able to reuse a lot of the e2e testing for the new design.

frontend/svelte/src/routes/Proposals.svelte
```diff
@@ -98,13 +99,26 @@
...
-    // Show spinner right away avoiding debounce
-    loading = true;
-    proposalsStore.setProposals([]);
-
-    debounceFindProposals?.();
-  });
+  const unsubscribe: Unsubscriber = proposalsFiltersStore.subscribe(
+    ({ lastAppliedFilter }) => {
+      if (lastAppliedFilter === "excludeVotedProposals") {
+        // mock fetching
```
lmuntanerDfinity, line 105: wow, is this the best way? Isn't this filter applied locally? Why don't we just change the list immediately?

Sorry, I am not very familiar with Proposals.

lmuntanerDfinity (commented):
A question.

frontend/svelte/src/lib/utils/proposals.utils.ts
peterpeterparker: I had a look again at the Dart code and it was checking for `UNSPECIFIED` there well I guess we will be good indeed.

```
.filter((proposal) =>
                          !excludeVotedProposals ||
                          proposal.status != ProposalStatus.Open ||
                          proposal.ballots.values
                              .any((ballot) => ballot.vote == Vote.UNSPECIFIED))
                              ```

frontend/svelte/src/routes/Proposals.svelte
peterpeterparker: It has to trigger the intersection. The observer does not start when the end of the page is reached but after 20%. I fear we will have some weird side effect and issue if not exactly all children of the observer trigger the intersection observer.

frontend/svelte/src/routes/Proposals.svelte
peterpeterparker: Let's discuss this today. To be honest I don't get the motivation of the PR, missing the scope.

frontend/svelte/src/routes/Proposals.svelte
mstrasinskis: Most of the time after clicking on the "votable proposals" user will see the same proposals. This is UX improvement to give a visual feedback that the list was updated.

frontend/svelte/src/routes/Proposals.svelte
peterpeterparker: It has been improved by setting `hidden` within the card

frontend/svelte/src/routes/Proposals.svelte
peterpeterparker: `position: absolute` won't be used

peterpeterparker (approved):
LGTM, thanks!

PR #646, opened 2022-04-01
L2-365: Neuron Detail follow functionality

lmuntanerDfinity:
# Motivation

User can edit followees from neuron details page.

# Changes

* New modal FollowNeuronsModal used in Neuron Details.
* Read followees and use them in EditFollowNeurons screen.

# Tests

* Test FollowNeuronsModal.

### DIFF
--- a/frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard.svelte
+++ b/frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard.svelte
@@ -1,11 +1,31 @@
 <script lang="ts">
   import type { NeuronInfo } from "@dfinity/nns";
+  import { i18n } from "../../stores/i18n";
   import Card from "../ui/Card.svelte";
+  import FollowNeuronsButton from "./actions/FollowNeuronsButton.svelte";
 
   export let neuron: NeuronInfo;
+  console.log(neuron);
 </script>
 
 <Card>
+  <h3 slot="start">{$i18n.neuron_detail.following_title}</h3>
+  <p>{$i18n.neuron_detail.following_description}</p>
   <!-- TODO: https://dfinity.atlassian.net/browse/L2-354 -->
-  Following Card {neuron.neuronId}
+  <div class="actions">
+    <FollowNeuronsButton {neuron} />
+  </div>
 </Card>
+
+<style lang="scss">
+  h3 {
+    margin-bottom: 0;
+  }
+  p {
+    margin-top: 0;
+  }
+  .actions {
+    display: flex;
+    justify-content: flex-end;
+  }
+</style>
--- a/frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard.svelte
+++ b/frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard.svelte
@@ -5,7 +5,6 @@
   import FollowNeuronsButton from "./actions/FollowNeuronsButton.svelte";
 
   export let neuron: NeuronInfo;
-  console.log(neuron);
 </script>
 
 <Card>
--- a/frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
+++ b/frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
@@ -12,6 +12,7 @@
     formatVotingPower,
     hasJoinedCommunityFund,
     isCurrentUserController,
+    isNeuronControllable,
   } from "../../utils/neuron.utils";
   import NeuronCard from "../neurons/NeuronCard.svelte";
   import Tooltip from "../ui/Tooltip.svelte";
@@ -21,7 +22,6 @@
   import SplitNeuronButton from "./actions/SplitNeuronButton.svelte";
   import DissolveActionButton from "./actions/DissolveActionButton.svelte";
   import DisburseButton from "./actions/DisburseButton.svelte";
-  import { isNeuronControllable } from "../../services/neurons.services";
   import { accountsStore } from "../../stores/accounts.store";
 
   export let neuron: NeuronInfo;
@@ -31,9 +31,10 @@
   let userControlled: boolean;
   $: userControlled = isCurrentUserController(neuron);
   let isControllable: boolean;
-  // Hack to call isNeuronControllable every time accounts store changes
-  // TODO: Move logic to `accountsStore` directly
-  $: $accountsStore, (isControllable = isNeuronControllable(neuron));
+  $: isControllable = isNeuronControllable({
+    neuron,
+    accounts: $accountsStore,
+  });
 </script>
 
 <NeuronCard {neuron}>
--- a/frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
+++ b/frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
@@ -22,6 +22,7 @@
   import DissolveActionButton from "./actions/DissolveActionButton.svelte";
   import DisburseButton from "./actions/DisburseButton.svelte";
   import { isNeuronControllable } from "../../services/neurons.services";
+  import { accountsStore } from "../../stores/accounts.store";
 
   export let neuron: NeuronInfo;
 
@@ -30,7 +31,9 @@
   let userControlled: boolean;
   $: userControlled = isCurrentUserController(neuron);
   let isControllable: boolean;
-  $: isControllable = isNeuronControllable(neuron);
+  // Hack to call isNeuronControllable every time accounts store changes
+  // TODO: Move logic to `accountsStore` directly
+  $: $accountsStore, (isControllable = isNeuronControllable(neuron));
 </script>
 
 <NeuronCard {neuron}>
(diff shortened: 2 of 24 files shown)


lmuntanerDfinity:
UI to show followees is not hardcoded nor the badges number:

Screenshot 2022-04-01 at 10 52 28

lmuntanerDfinity:

frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard.svelte
```diff
@@ -1,11 +1,31 @@
 <script lang="ts">
   import type { NeuronInfo } from "@dfinity/nns";
+  import { i18n } from "../../stores/i18n";
   import Card from "../ui/Card.svelte";
+  import FollowNeuronsButton from "./actions/FollowNeuronsButton.svelte";
 
   export let neuron: NeuronInfo;
+  console.log(neuron);
```
peterpeterparker, line 8: to be removed

frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
```diff
@@ -30,7 +31,9 @@
   let userControlled: boolean;
   $: userControlled = isCurrentUserController(neuron);
   let isControllable: boolean;
-  $: isControllable = isNeuronControllable(neuron);
+  // Hack to call isNeuronControllable every time accounts store changes
+  // TODO: Move logic to `accountsStore` directly
+  $: $accountsStore, (isControllable = isNeuronControllable(neuron));
```
peterpeterparker, line 36: - since this sync mechanism happens on a global scale it would make sense to move next to where it is set rather than in a sub-sub-sub-...-component in which the neuron is a local state

- is there test(s) for this?

frontend/svelte/src/lib/components/neurons/FollowTopicSection.svelte
```diff
@@ -18,14 +20,42 @@
...
+  type FolloweeData = {
+    neuronId: NeuronId;
+    name?: string;
+  };
+  let followees: FolloweeData[] = [];
+  $: {
+    const followesPerTopic = neuron.fullNeuron?.followees.find(
+      ({ topic: currentTopic }) => topic === currentTopic
+    );
+    if (followesPerTopic !== undefined) {
```
peterpeterparker, line 32: - can you extract the logic to a function to make it more obvious what's happening

- nth: es6

frontend/svelte/src/lib/modals/neurons/FollowNeuronsModal.svelte
```diff
@@ -0,0 +1,24 @@
...
+<Modal theme="dark" size="medium" on:nnsClose>
+  <span slot="title">{$i18n.neurons.follow_neurons_screen}</span>
+  <section>
+    <EditFollowNeurons {neuron} />
+  </section>
+</Modal>
+
+<style lang="scss">
+  section {
+    height: min(500px, calc(100vh - 156px - (2 * var(--padding))));
```
peterpeterparker, line 19: mixin?

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
```diff
@@ -43,6 +44,7 @@
   const addFolloweeByAddress = async () => {
     loading = true;
     loadingAddress = true;
+    startBusy("add-followee");
     let followee: bigint;
     if (followeeAddress === undefined) {
       return;
```
peterpeterparker, line 50: `loading = true` stays for ever  if `followeeAddress === undefined`?

frontend/svelte/src/lib/components/neuron-detail/actions/FollowNeuronsButton.svelte
```diff
@@ -0,0 +1,19 @@
...
+  import FollowNeuronsModal from "../../../modals/neurons/FollowNeuronsModal.svelte";
+
+  export let neuron: NeuronInfo;
+
+  let showModal: boolean = false;
+  const openModal = () => (showModal = true);
+  const closeModal = () => (showModal = false);
+</script>
+
+<button class="primary small" on:click={openModal}
```
peterpeterparker, line 13: is there a test?

peterpeterparker (changes requested):

frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
lmuntanerDfinity: Do you mean to move it to the `accountStore` as I mention above? Initially I considered this as a neuron util, like `isCurrentUserController` or `hasValidStake` for example. Then I realized that it needed to be refreshed. This might happen other times. Therefore, we need a pattern that we can use for this "utils" that are not pure functions. What exactly do you propose?

Yes, there are tests for `isNeuronControllable` in `neurons.services.spec.ts`.

frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard.svelte
lmuntanerDfinity: yes, thanks!

frontend/svelte/src/lib/components/neurons/FollowTopicSection.svelte
lmuntanerDfinity: sure

frontend/svelte/src/lib/modals/neurons/FollowNeuronsModal.svelte
lmuntanerDfinity: sure

frontend/svelte/src/lib/modals/neurons/NewFolloweeModal.svelte
lmuntanerDfinity: oh, good catch! The button is disabled, so this should never happen. But better be safe!

frontend/svelte/src/lib/components/neuron-detail/actions/FollowNeuronsButton.svelte
lmuntanerDfinity: There is a test in the Neuron Details page that the button is there, and there is a test for the modal functionality.

Shall I add a test that this button opens the modal?

frontend/svelte/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
peterpeterparker: Sorry forget this comment, I misleaded the usage of the state.

frontend/svelte/src/lib/components/neuron-detail/actions/FollowNeuronsButton.svelte
peterpeterparker: That works, thanks

lmuntanerDfinity:

frontend/svelte/src/lib/components/neurons/FollowTopicSection.svelte
```diff
@@ -18,14 +20,40 @@
   $: id = Topic[topic];
 
   let showNewFolloweeModal: boolean = false;
-  let followees: NeuronId[] = [];
+  type FolloweeData = {
+    neuronId: NeuronId;
+    name?: string;
+  };
+  let followees: FolloweeData[] = [];
+  $: {
```
peterpeterparker, line 28: Should we go a step further and extract this to a `utils` and add a test.  What do you think?

frontend/svelte/src/lib/themes/mixins/_modal.scss
```diff
@@ -0,0 +1,6 @@
+@mixin wrapper {
```
peterpeterparker, line 1: I would use something else than "wrapper" because the section contains already an element that is a ".wrapper" if I am not wrong

peterpeterparker (changes requested):

frontend/svelte/src/lib/components/neurons/FollowTopicSection.svelte
lmuntanerDfinity: It accesses a store, where we need to subscribe in case there are changes. We could not just extract all of this to a util.

This pattern of utils that access stores and we want to have the updated data always is a common patter.

We might want to find a nice pattern to use everywhere. Similar to the topic of `$: $accountsStore, isControllable = isNeuronControllable(neuron)`.

I am not very happy with the hack. But maybe we can refactor all this together when we find a pattern we like.

frontend/svelte/src/lib/themes/mixins/_modal.scss
lmuntanerDfinity: not in the mixing. This is a new mixing. But I can rename it.

frontend/svelte/src/lib/components/neurons/FollowTopicSection.svelte
peterpeterparker:
```
$ something = function($knownNeuronsStore)
```

peterpeterparker (approved):
Cool, thanks!

PR #647, opened 2022-04-04
feat: await import dummy proposal script

peterpeterparker:
# Motivation

Ship within prod bundle what needs to be shipped only.

# Changes

- `await import` to load dummy proposal script
- assert feature can only be called on testnet


mstrasinskis (approved):
Nice! I tried so hard to find the rollup config changes before I noticed .js 👍

peterpeterparker:
@mstrasinskis ahah I went for the less time consuming solution, so I just dropped the .js in the asset 😉. thx for the review

PR #649, opened 2022-04-04
L2-329: Manage errors stake neuron

lmuntanerDfinity:
# Motivation

Stake Neuron flow: manage edge cases, show errors gracefully and give a better UX on errors.

# Changes

* Add "invalidStates" array in the CreateNeuronModal. Error should be shown and modal closed on unexpected edge cases.
* Show busy screen when updating delay and staking neuron.
* Manage errors in services instead of component during the flow.
* Components ConfirmDissolveDelay, SetDissolveDelay and EditFollowNeurons are not rendered without a neuron.

### DIFF
--- a/frontend/svelte/src/lib/components/neurons/ConfirmDissolveDelay.svelte
+++ b/frontend/svelte/src/lib/components/neurons/ConfirmDissolveDelay.svelte
@@ -7,7 +7,11 @@
   import { secondsToDuration } from "../../utils/date.utils";
   import { replacePlaceholders } from "../../utils/i18n.utils";
   import { formatICP } from "../../utils/icp.utils";
-  import { votingPower } from "../../utils/neuron.utils";
+  import {
+    formatVotingPower,
+    neuronStake,
+    votingPower,
+  } from "../../utils/neuron.utils";
   import { startBusy, stopBusy } from "../../stores/busy.store";
 
   export let delayInSeconds: number;
@@ -16,18 +20,20 @@
 
   const dispatcher = createEventDispatcher();
   let neuronICP: bigint;
-  $: neuronICP = neuron.fullNeuron?.cachedNeuronStake ?? BigInt(0);
+  $: neuronICP = neuronStake(neuron);
 
   const updateNeuron = async () => {
     startBusy("update-delay");
     loading = true;
-    await updateDelay({
+    const neuronId = await updateDelay({
       neuronId: neuron.neuronId,
       dissolveDelayInSeconds: delayInSeconds,
     });
     stopBusy("update-delay");
-    dispatcher("nnsNext");
     loading = false;
+    if (neuronId !== undefined) {
+      dispatcher("nnsNext");
+    }
   };
 </script>
 
@@ -50,10 +56,12 @@
   <div class="voting-power">
     <h5>{$i18n.neurons.voting_power}</h5>
     <p>
-      {votingPower({
-        stake: neuronICP,
-        dissolveDelayInSeconds: delayInSeconds,
-      }).toFixed(2)}
+      {formatVotingPower(
+        votingPower({
+          stake: neuronICP,
+          dissolveDelayInSeconds: delayInSeconds,
+        })
+      )}
     </p>
   </div>
   <div>
--- a/frontend/svelte/src/lib/components/neurons/ConfirmDissolveDelay.svelte
+++ b/frontend/svelte/src/lib/components/neurons/ConfirmDissolveDelay.svelte
@@ -8,81 +8,68 @@
   import { replacePlaceholders } from "../../utils/i18n.utils";
   import { formatICP } from "../../utils/icp.utils";
   import { votingPower } from "../../utils/neuron.utils";
+  import { startBusy, stopBusy } from "../../stores/busy.store";
 
   export let delayInSeconds: number;
-  export let neuron: NeuronInfo | undefined;
+  export let neuron: NeuronInfo;
   let loading: boolean = false;
 
   const dispatcher = createEventDispatcher();
   let neuronICP: bigint;
-  $: neuronICP = neuron?.fullNeuron?.cachedNeuronStake ?? BigInt(0);
+  $: neuronICP = neuron.fullNeuron?.cachedNeuronStake ?? BigInt(0);
 
   const updateNeuron = async () => {
-    if (neuron === undefined) {
-      // TODO: Manage errors https://dfinity.atlassian.net/browse/L2-329
-      console.error("Neuron is not defined");
-      return;
-    }
-
+    startBusy("update-delay");
     loading = true;
-    try {
-      await updateDelay({
-        neuronId: neuron.neuronId,
-        dissolveDelayInSeconds: delayInSeconds,
-      });
-      dispatcher("nnsNext");
-    } catch (error) {
-      // TODO: Manage errors https://dfinity.atlassian.net/browse/L2-329
-      console.error(error);
-    } finally {
-      loading = false;
-    }
+    await updateDelay({
+      neuronId: neuron.neuronId,
+      dissolveDelayInSeconds: delayInSeconds,
+    });
+    stopBusy("update-delay");
+    dispatcher("nnsNext");
+    loading = false;
   };
 </script>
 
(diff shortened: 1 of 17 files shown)


lmuntanerDfinity:

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
```diff
@@ -68,40 +97,32 @@
...
-  <!-- TODO: Manage edge case: https://dfinity.atlassian.net/browse/L2-329 -->
-  {#if currentStepName === "StakeNeuron"}
-    <!-- we spare a spinner for the selectedAccount within StakeNeuron because we reach this step once the selectedAccount has been selected -->
-    {#if selectedAccount !== undefined}
-      <StakeNeuron
-        account={selectedAccount}
-        on:nnsNeuronCreated={goToDissolveDelay}
-      />
-    {/if}
+  {#if currentStepName === "StakeNeuron" && selectedAccount !== undefined}
```
peterpeterparker, line 103: I made these changes on purpose - i.e. in my opinion the condition on the step of the wizard should only test the wizard name and not additional logic such as `&& selectedAccount !== undefined` or `newNeuron !== undefined`

In case of edge case, assuming the `currentStepName` is set and the additional clause is not set, the UI can become quirky and the user can be lost because no step and no feedback are displayed on screen.

In other words, for a smooth UI, a step should always be displayed.

That's why I deferred such logic as `selectedAccount !== undefined` or `newNeuron !== undefined` within their respective component and displayed for example spinner while data are loading.

How to you prevent these issues - i.e. how do you prevent no step to be rendered?

peterpeterparker (changes requested):

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
lmuntanerDfinity: I guess that it depends on what we want to do and the nature of the error.

At the moment, if we have the dissolve delay step and no neuron, it's not because we are fetching it, it's because something went wrong. Therefore, showing the spinner, will not be very helpful, because it will stay there forevever. We need to tell the user about it.

Another way is to also change the logic of when to move steps. Move steps earlier, before the previous step has finished. In this case, the spinner would make sense because in the end, the previous step would finish and we would have the neuron there. But we could show the spinner outside of the component, we show either the component or the spinner. But not the spinner inside the component.

I try to reduce the amount of complexity in components and adding the possiblity of `undefined` adds a case to be handled in the component that we could avoid. I guess this is a guideline we should agree on.

Also about when to move step. At the moment, we don't move until the step has finished totally. Therefore, the next step without neuron is an edge case where a Spinner is not information enough for the user.

Maybe a call with @mstrasinskis as well?

peterpeterparker: - Spinner was only a placeholder, a skeleton or any other information can be used. 
- In my opinion and this particular case the complexity of testing if data are undefined or not should not be deferred to the parent because the parent is the wizard binder - i.e. beside nav within the wizard it should be kind of dumb

lmuntanerDfinity: Discussed on Slack

lmuntanerDfinity:

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
```diff
@@ -49,7 +51,34 @@
...
+    { stepName: "SetDissolveDelay", neuron: null },
+    { stepName: "ConfirmDissolveDelay", neuron: null },
+    { stepName: "EditFollowNeurons", neuron: null },
+  ];
+  $: {
+    newNeuron = $neuronsStore.find(({ neuronId }) => newNeuronId === neuronId);
+    const invalidState = invalidStates.find(({ stepName, neuron, account }) => {
+      return (
+        stepName === currentStep?.name &&
+        ((neuron === null && newNeuron === undefined) ||
```
peterpeterparker, line 71: it's not "invalid" if neuron is null and newNeuron is not undefined and the contrary? likewise for account?

in short: it should not be`or`?

lmuntanerDfinity: Maybe I should not deconstruct here. This is when the `invalidState.neuron === null && newNeuron === undefined`.

This is an invalid state:
```
  type InvalidState = {
    stepName: string;
    neuron?: null;
    account?: null;
  };
```

I could set the `newNeuron` initial value to `null` as well to make it comparable to the InvalidState.

peterpeterparker: Not a problem with destructing, I probably just don't get it 🤪

```
const invalidStates: InvalidState[] = [
    { stepName: "StakeNeuron", account: null },
    ...
    ```
    
What does `account: null` here mean? Is that "the state is invalid if the stepName is StakeNeuron and if the selected account is null"?

If so, what about renaming `account` to something maybe more obvious like `invalidCondition`? Or maybe even a function?

```
type InvalidState = { stepName: string; isInvalid: <T>(obj: T) => boolean };
```

**Review by mstrasinskis — COMMENTED** [review]
Looks good! Just a couple of comments.

File: frontend/svelte/src/lib/components/neurons/SetDissolveDelay.svelte [hunk 3]
```diff
@@ -36,7 +36,7 @@
     dispatcher("nnsSkipDelay");
   };
   let neuronICP: bigint;
-  $: neuronICP = neuron?.fullNeuron?.cachedNeuronStake ?? BigInt(0);
+  $: neuronICP = neuron.fullNeuron?.cachedNeuronStake ?? BigInt(0);
```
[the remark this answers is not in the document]
lmuntanerDfinity: good point!

frontend/svelte/src/lib/components/neurons/ConfirmDissolveDelay.svelte
```diff
@@ -8,81 +8,68 @@
...
-      console.error(error);
-    } finally {
-      loading = false;
-    }
+    await updateDelay({
+      neuronId: neuron.neuronId,
+      dissolveDelayInSeconds: delayInSeconds,
+    });
+    stopBusy("update-delay");
+    dispatcher("nnsNext");
```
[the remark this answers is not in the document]
lmuntanerDfinity: good point, I thought the same, not sure the idea of throwing only in the services it matches always.

frontend/svelte/src/lib/components/neurons/ConfirmDissolveDelay.svelte
```diff
@@ -8,81 +8,68 @@
...
+    <p>
+      {replacePlaceholders($i18n.neurons.icp_stake, {
+        $amount: formatICP(neuronICP),
+      })}
+    </p>
+  </div>
+  <div class="voting-power">
+    <h5>{$i18n.neurons.voting_power}</h5>
+    <p>
+      {votingPower({
```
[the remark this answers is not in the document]
lmuntanerDfinity: yep!

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
```diff
@@ -94,26 +134,29 @@
...
-  <!-- TODO: Manage edge case: https://dfinity.atlassian.net/browse/L2-329 -->
   {#if currentStep?.name === "SetDissolveDelay"}
-    <SetDissolveDelay
-      neuron={newNeuron}
-      on:nnsSkipDelay={goEditFollowers}
-      on:nnsConfirmDelay={goNext}
-      bind:delayInSeconds
-    />
+    {#if newNeuron !== undefined}
+      <SetDissolveDelay
```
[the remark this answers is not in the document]
lmuntanerDfinity: yes, thanks, it got back in after rebasing and solving conflicts.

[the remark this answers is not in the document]
lmuntanerDfinity: done

mstrasinskis (approved):
Thanks 👍

PR #653, opened 2022-04-05
test: fix warnings "<FiltersButton> was created with unknown prop 'data-tid'"

peterpeterparker:
# Motivation

Clear "<FiltersButton> was created with unknown prop 'data-tid'" throw while testing

# Changes

- remove unused `data-tid` incorrectly set to a cmp

# Screenshot

Warning used to be:

Capture d’écran 2022-04-05 à 07 43 33


peterpeterparker:
This will be solved by existing PR #643

PR #655, opened 2022-04-05
feat: account address input rules

peterpeterparker:
# Motivation

Implement address input rules as in Dart.

# Changes

- input is of type `test` and should be min 40 characters
- modify component `Input` to support `minlength`


lmuntanerDfinity (approved):
Great thanks!

PR #657, opened 2022-04-05
feat: new transaction store, context and init amount step

peterpeterparker:
# Motivation

To create new transaction, user have to select and enter several data such as selecting the source account, selecting the destination address or account and the amount of ICP.

This PR add a store that contains these information and scope contextually its usage to the wizard (to avoid props going up and down and to scope logic withing components - i.e. steps).

In addition, to display the result of the selection, this PR also initialize the third step of the "New transaction" wizard. 

# Changes

- new `transaction.store`
- new `NewTransactionAmount` step - component
- extract common function that return the constant transaction fee as string
- insert `NewTransactionAmount` into wizard
- use new store in a wizard `context`
- use context in `NewTransactionDestination` (renamed) and `NewTransactionSource` (renamed) to select source and destination information
- bind input and selection of the account in `NewTransactionDestination`

# Screenshot

Capture d’écran 2022-04-05 à 16 01 14


frontend/svelte/src/lib/components/accounts/Address.svelte
```diff
@@ -3,11 +3,12 @@
...
   import { ACCOUNT_ADDRESS_MIN_LENGTH } from "../../constants/accounts.constants";
   import { invalidAddress } from "../../utils/accounts.utils";
+  import { createEventDispatcher } from "svelte";
 
   export let address: string = "";
 
-  // TODO(L2-430): to be implemented
-  const submitAddress = () => console.log("submitAddress");
+  const dispatcher = createEventDispatcher();
+  const submitAddress = () => dispatcher("nnsAddress", { address });
```
lmuntanerDfinity, line 11: Why do we need to pass the address if the parent already has the address?

I think that you can forward `on:submit`. If below you leave: `on:submit|preventDefault`. And when you use `<Address on:submit={...} bind:address />`.

frontend/svelte/src/lib/components/accounts/NewTransactionAmount.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+  import { getContext } from "svelte";
+  import { formattedTransactionFeeICP } from "../../utils/icp.utils";
+  import { i18n } from "../../stores/i18n";
+  import type { TransactionContext } from "../../stores/transaction.store";
+  import { NEW_TRANSACTION_CONTEXT_KEY } from "../../stores/transaction.store";
+
+  const context: TransactionContext = getContext<TransactionContext>(
```
lmuntanerDfinity, line 8: If we rely on the context here, doesn't that mean that we won't be able to reuse this component?

Isn't it use anywhere else?

frontend/svelte/src/lib/components/accounts/NewTransactionDestination.svelte
```diff
@@ -0,0 +1,60 @@
...
+    }));
+
+    next?.();
+  };
+</script>
+
+<div>
+  <p>{$i18n.accounts.enter_address_or_select}</p>
+
+  <Address bind:address on:nnsAddress={chooseAddress} />
```
lmuntanerDfinity, line 44: I think what I mentioned before: `<Address bind:address on:submit={chooseAddress}`.

Or, if we rely on the event, no need to `bind:address`. I also like the idea of not binding and getting it from the event.

lmuntanerDfinity (changes requested):
One proposal and questions

frontend/svelte/src/lib/components/accounts/NewTransactionAmount.svelte
peterpeterparker: > Isn't it use anywhere else?

As far as I now it's only there but it won't be difficult to extend, features are well scoped to the component.

> If we rely on the context here, doesn't that mean that we won't be able to reuse this component?

Yes and no. Yes in that form you can only use the component where the context has been set but no, you can also use the component if you defined the context, just need to extend it to support different key. We can also extend the context to be more generic, like having a generic context for wizard.

```
export interface WizardContext<T> {
   store: T;
   next: () => void;
}

export interface NewTransactionContext extends WizardContext...

export interface CreateNeuronContext extends WizardContext...
```

A spontaneous idea, not a thing we should or we shall do. Just want to highlight the fact that it can I think be enhanced quite easily.

frontend/svelte/src/lib/components/accounts/Address.svelte
peterpeterparker: Done

lmuntanerDfinity (approved):

frontend/svelte/src/lib/stores/transaction.store.ts
```diff
@@ -0,0 +1,27 @@
...
+  selectedAccount: Account | undefined;
+  destinationAddress: string | undefined;
+}
+
+export interface TransactionContext {
+  store: Writable<TransactionStore>;
+  next: () => void;
+}
+
+export const NEW_TRANSACTION_CONTEXT_KEY = "new-transaction";
```
mstrasinskis, line 15: 😎 
```suggestion
export const NEW_TRANSACTION_CONTEXT_KEY = Symbol("new-transaction");
```

frontend/svelte/src/lib/components/accounts/NewTransactionDestination.svelte
```diff
@@ -0,0 +1,59 @@
...
+  <p>{$i18n.accounts.enter_address_or_select}</p>
+
+  <Address bind:address on:submit={chooseAddress} />
+
+  <!-- Prevent the component to be presented with a scroll offset when navigating between wizard steps -->
+  {#if mounted}
+    <SelectAccount
+      on:nnsSelectAccount={chooseAccount}
+      displayTitle={true}
+      disableSelection={!emptyAddress(address)}
```
mstrasinskis, line 50: This **!** emptyAddress looks wrong. But please ignore if it suppose to be like this 👍

mstrasinskis (approved):
LGTM. Just a couple of questions.

frontend/svelte/src/lib/components/accounts/NewTransactionDestination.svelte
peterpeterparker: Yeah it's correct, added a comment in the code

>

PR #660, opened 2022-04-06
Feat: Remove isCurrentUserController dependency

lmuntanerDfinity:
# Motivation

We are going to remove the property isCurrentUserController from the Neuron type in nns-js. We first need to remove the dependency here.

# Changes

* Neuron util `isCurrentUserController` now checks that the controller is the main account's principal.

# Tests

* Update tests to new functionality.


lmuntanerDfinity:
@mstrasinskis @peterpeterparker after discussing the @ielashi we should remove the property added by nns-js `isCurrentUserController` and instead, each app should do they own logic for it, not nns-js.

This PR is the first step towards removing it from nns-js.

lmuntanerDfinity:

frontend/svelte/src/lib/utils/neuron.utils.ts
```diff
@@ -90,10 +90,13 @@ export const formatVotingPower = (value: bigint): string =>
 export const hasJoinedCommunityFund = (neuron: NeuronInfo): boolean =>
   neuron.joinedCommunityFundTimestampSeconds !== undefined;
 
-export const isCurrentUserController = (neuron: NeuronInfo): boolean =>
-  neuron.fullNeuron?.isCurrentUserController === undefined
+export const isCurrentUserController = (
+  neuron: NeuronInfo,
+  accounts: AccountsStore
```
peterpeterparker, line 95: - why passing the all store and not just the main account?

- nth: object params `{neuron, mainAccount}`

peterpeterparker:
> We first need to remove the dependency here.

Why? The nns-js dependency is pinned we can proceed with both in parallel no?

frontend/svelte/src/lib/utils/neuron.utils.ts
lmuntanerDfinity: sure, I wanted to keep the same signature of the other helper, but it also works passing the main only

lmuntanerDfinity:
Well, just in case I change the nns-js, then we need to update nns-js because someone else adds functionality there that we need.

But yes, regarding only this refactor, we can proceed in parallel.

peterpeterparker (approved):

PR #664, opened 2022-04-06
Add CSP for the web worker

bitdivine:
# Motivation
Current main does not load on iOS.  The console gives the error:

Screenshot 2022-04-06 at 14 14 45

With this PR the page loads in Safari and there are no console errors.

Note:  There is a "worker-src" directive but it is [not supported by Safari](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/worker-src).

# Changes
- Added `child-src: 'self'` to the CSP.

# Tests
With this PR the page loads in Safari and there are no console errors.

### DIFF
--- a/frontend/svelte/scripts/build.index.mjs
+++ b/frontend/svelte/scripts/build.index.mjs
@@ -52,6 +52,7 @@ const updateCSP = (content) => {
         content="default-src 'none';
         connect-src 'self' ${cspConnectSrc()};
         img-src 'self' https://nns.raw.ic0.app/;
+        child-src 'self';
         manifest-src 'self';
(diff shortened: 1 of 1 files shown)


bitdivine:
@peterpeterparker  - I have deployed this PR to https://nns-prod.winning.gold so that should work for you.

https://nns-flutter.winning.gold does not have this change so I would expect you to see a black screen there.

peterpeterparker (approved):
Thank you for the links @bitdivine. Indeed now it is clear and this does set the CSP rules for the worker correctly, thank you.

frederikrothenberger (approved):
There was no worker-source before, so why did this work on Firefox & Chrome before? 🤔

PR #674, opened 2022-04-07
Factor out e2e deployment code

bitdivine:
# Motivation
e2e tests are run locally by developers and in CI.  Both need the same incantation to deploy the canisters.  Let's reuse rather than duplicate.

# Changes
- Moved the e2e canister deployment commands from in the `.github/workflows` to a script.

# Tests


e2e-tests/scripts/deploy-local-canisters
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -euo pipefail
+cd "$(dirname "$0")/../.."
+dfx canister --no-wallet create --all || echo Canisters may have been created already.  Proceeding...
```
bitdivine: In CI, the --clean will guarantee that there are no conflicts.  In a developer environment I may change the code and redeploy without restarting dfx.  I suspect that I may not be alone in that.

lmuntanerDfinity (approved):
LGTM!

PR #675, opened 2022-04-08
Update dfx to 0.9.2

bitdivine:
# Motivation
The dfx version 0.8.4 is now well behind the default version (0.9.2).  It is generally sensible to stay reasonably up to date.  Also, looking at the changelog, there are changes that would help us clean up our code.  Examples:

- `DFX_NETWORK` is available as an environment variable to the build command.
- There is [a standardized location](https://github.com/dfinity/sdk/blob/master/CHANGELOG.adoc#feat-remote-canister-support) for external canister IDs, such as governance, ledger and II.

# Changes
- Update dfx from 0.8.4 to 0.9.2
- Removed `--no-wallet` everywhere as it is now the default and using it throws an error.

# Blast radius
These may be affected:

- Local development environments
- Deployment to testnets
- CI
- Production is not deployed with dfx but it is built with dfx.

# Tests
See CI

### TODO: Test Production upgrade.
- Deploy nns-dapp to a testnet with dfx 0.8.4
- Upgrade with dfx 0.9.2 (verify that this works)
- Verify that the upgraded canister works.

# Release
As updating dfx is something we do rarely, I associate quite a lot of risk with it.  As such, I would consider doing a release with just this change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-20.04
     timeout-minutes: 60
     env:
-      DFX_VERSION: 0.8.3
+      DFX_VERSION: 0.9.2
       RUSTC_VERSION: 1.58.1
       FLUTTER_VERSION: 2.2.3
       IC_CDK_OPTIMIZER_VERSION: 0.3.1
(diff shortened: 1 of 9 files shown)


lmuntanerDfinity (approved):
I guess this will help with deploying Ledger and Governance canisters for e2e testing?

PR #676, opened 2022-04-08
Make the multi-tab login test work again

bitdivine:
# Motivation

# Changes

# Tests


e2e-tests/common/login.ts
```diff
@@ -0,0 +1,48 @@
...
+) => {
+  await browser.pause(2_000);
+  const loginButton = await getLoginButton(browser);
+  await loginButton.waitForExist();
+
+  // Check which windows already exist, then wait for a new window to appear.
+  const originalTabId = await browser.getWindowHandle();
+  await browser["screenshot"]("nns_dapp_login_page");
+  await loginButton.click();
+  await browser.pause(10_000);
```
bitdivine, line 16: Note: I am preparing code for here (and for `registration.ts`) that uses `waitUntil(..)` rather than browserPause.  I would like to keep the tempo of at least one e2e PR per day though and it's almost the end of the day.

e2e-tests/wdio.conf.ts
```diff
@@ -61,7 +61,8 @@ export const config: WebdriverIO.Config = {
       acceptInsecureCerts: true,
     },
   ],
-  logLevel: "info",
+  logLevel:
+    process.env.LOG_LEVEL === undefined ? "info" : process.env.LOG_LEVEL,
```
bitdivine, line 65: I personally find `info` too noisy - it just fills my screen with details and I have to scroll up searching for problems.  I prefer `warn`.  This change above lets me override the log level from the command line without changing any code.  I drop back to info if I have to debug.

e2e-tests/common/register.ts
```diff
@@ -116,7 +121,13 @@ export const register = async (browser: WebdriverIO.Browser): string => {
```
bitdivine, line 121: Registration has only succeeded if we actually log in, so I moved the wait for login into the registration function.

lmuntanerDfinity (commented):
A final comment.

I think that if the screenshots are automatic and are committed, we should remove the Spinner and Notification before doing it. Otherwise we might bet screenshots commits in randome PRs.

bitdivine:
I see just this one comment (are there more?), but it is a good idea.  I will do so using the existing code but add it directly in the screenshot command.  I do wonder whether there should be a CSS class for unstable items such as spinners that are expected to be omitted from screenshots.

lmuntanerDfinity:
If we add this PR as is, we can have screenshots being committed in any PR. I'd rather remove those things before, does it make sense?

I like the idea of a single data-tid for unstable elements, but for now you can reuse what I did in `home.e2e.ts` and create a helper out of it.

lmuntanerDfinity:
The annoying part of having screenshots committed, is that the Pipeline checks get stuck for some weird Github actions bug. It's not just the commit.

bitdivine:
Regarding screenshots:  In main, screenshots are no longer committed, so the concern of all PRs being affected by screenshots changing no longer applies.

lmuntanerDfinity (approved):
Great!

PR #678, opened 2022-04-08
Prevent hardware wallets from being attached to accounts multiple times

hpeebles:
# Motivation

This PR fixes a bug where a hardware wallet could be attached to a single account multiple times. This would result in the user who own's the account seeing duplicate transactions for the hardware wallet.

# Changes

- Prevent a hardware wallet from being attached to an account multiple times.

# Tests

I attached my hardware wallet twice on 2 test versions of the NNS Dapp, one with the fix, and one without.
On the one without the fix my transactions were duplicated, on the one with the fix they were not.

Use II anchor 27 to view the test account.

No fix - https://s24we-diaaa-aaaaa-aaaka-cai.nnsdapp.dfinity.network/#/wallet/bf7999cc05364166ec0bab77c69fa6842ec03160095ef404db71adfdd986f099

Fix - https://su63m-yyaaa-aaaaa-aaala-cai.nnsdapp.dfinity.network/#/wallet/bf7999cc05364166ec0bab77c69fa6842ec03160095ef404db71adfdd986f099


ielashi (approved):

ielashi:
@hpeebles Late to the party. LGTM!

PR #679, opened 2022-04-10
e2e tests: Save html on error

bitdivine:
# Motivation
When an e2e test fails, it is often due to DOM selectors not being as expected.  It is time-consuming to recreate a similar situation manually.  Wouldn't it be quicker to save the DOM on error and examine that directly?

# Changes
- Save the DOM on error, with the same name as the screenshot but with html instead of jpg.
- Sanitise the screenshot filename.
- Use a separate counter for each suite.

# Tests
- I added `throw Error("Foo")` into the first line of the redirects test.
- I chose this test as it has non-alphanumeric characters in the test names.
- Running that test suite with: `LOG_LEVEL=warn ./scripts/test-winning` yielded these screenshots and html files:
  ```
  landing-page_00_landing-page.png
  landing-page_01_home-page.png
  redirects_00_test-fail_goes-to-accounts-page-after-registration.html
  redirects_00_test-fail_goes-to-accounts-page-after-registration.png
  redirects_01_test-fail_Redirects-accounts-v2-accounts.html
  redirects_01_test-fail_Redirects-accounts-v2-accounts.png
  redirects_02_test-fail_Redirects-neurons-v2-neurons.html
  redirects_02_test-fail_Redirects-neurons-v2-neurons.png
  redirects_03_test-fail_Redirects-proposals-v2-proposals.html
  redirects_03_test-fail_Redirects-proposals-v2-proposals.png
  redirects_04_test-fail_Redirects-canisters-v2-canisters.html
  redirects_04_test-fail_Redirects-canisters-v2-canisters.png
  ```


bitdivine:
Note: The test failure is unrelated; there is a timeout.  CI seems to struggle a bit with the e2e tests.

We could do with a test whether the e2e tests are flaky on CI.  Maybe a nightly job that runs the e2e tests 10 times.  If some timeouts are a bit tight that should show up there.

peterpeterparker:
Don't know if possible with Selenium but with Cypress you can record the video of tests that fail. From my experience, that would be helpful.

bitdivine:
@peterpeterparker - I agree, and quite generally, information on failed tests is likely to help.

Now, we have a screenshot when a test fails; not in this PR but already in main.  Upgrading that to a video would be cool.

A video would not help with query selectors.  The video would show what happened but not why.  Thus this PR would be useful even with a video.

peterpeterparker:
This HTML would only be accessible if the tests are run locally or would it be possible to download it from the CI when tests fails in the CI?

bitdivine:
The HTML should be available for failed CI jobs as well, assuming that we also merge: https://github.com/dfinity/nns-dapp/pull/681 

At the moment, without that PR, we don't even get the screenshots of failed CI jobs.

e2e-tests/wdio.conf.ts
```diff
@@ -50,7 +57,7 @@ export const config: WebdriverIO.Config = {
       project: "tsconfig.json",
     },
   },
-  specs: ["./specs/**/*.ts"],
+  specs: [["./specs/home.e2e.ts", "./specs/**/*.ts"]],
```
peterpeterparker, line 60: 1. an array of an array? looks like a typo (`[["./sp...` vs `["./sp...`)

2. not related to your PR but since we are already modifying this line maybe worth scoping the test pattern from `"./specs/**/*.ts"` to `"./specs/**/*.e2e.ts"`

e2e-tests/wdio.conf.ts
```diff
@@ -39,7 +39,14 @@ export const config: WebdriverIO.Config = {
     if (undefined !== error) {
       // Filenames containing spaces are painful to work with on the command line,
       // so we replace any spaces in the test name when we create the screenshot.
-      browser["screenshot"](`test-fail-${test.title.replace(/ /g, "-")}`);
+      const screenshotName = `test-fail-${test.title.replace(/ /g, "-")}`;
```
mstrasinskis, line 42: Does it make sense to add a date to the file-name to make it more or less unique?

bitdivine: I think this is not needed because each run generates just one set of screenshots; indeed the screenshot directory is deleted and recreated at the start of the e2e tests:  https://github.com/dfinity/nns-dapp/blob/main/.github/workflows/build.yml#L113

e2e-tests/wdio.conf.ts
bitdivine: This is an accidental commit.  Good catch!  Array-of-array tells wdio to run the inner tests as a group, so in sequence on a single runner.  I thought it might help but it didn't.

e2e-tests/wdio.conf.ts
bitdivine: Update:  I updated the screenshot code to
- (a) sanitize paths and
- (b) name screenshots `${suite_name}_${screenshot_number_in_suite}_${user_provided_name}` instead of `${screenshot_number_in_run}_${user_provided_name}`
I then reused this carefully constructed path when saving the dom.

e2e-tests/wdio.conf.ts
```diff
@@ -1,34 +1,56 @@
-const { existsSync, mkdirSync } = require("fs");
+const { existsSync, mkdirSync, writeFileSync } = require("fs");
 const { NNS_DAPP_URL } = require("./common/constants");
 
 export const config: WebdriverIO.Config = {
   baseUrl: NNS_DAPP_URL,
```
bitdivine, line 5: Screenshot function updated to:
- sanitize the filename
- restart the counter with every suite and prefix the filename with the suite name.
This carefully constructed name is then used when saving the DOM, just with an `.html` suffix instead of `.jpg`.

e2e-tests/wdio.conf.ts
```diff
@@ -39,7 +61,7 @@ export const config: WebdriverIO.Config = {
     if (undefined !== error) {
       // Filenames containing spaces are painful to work with on the command line,
       // so we replace any spaces in the test name when we create the screenshot.
```
mstrasinskis, line 63: is the comment obsolete? Probably it would be more reliable to always apply some filename transformation function before writing.

e2e-tests/wdio.conf.ts
```diff
@@ -1,34 +1,56 @@
...
-        .padStart(2, "0");
-      if (true === browser["screenshots-taken"].has(name)) {
-        throw Error(`A screenshot with this name was already taken: '${name}'`);
-      }
-      browser["screenshots-taken"].add(name);
+    browser.addCommand(
+      "screenshot",
+      async (name: string, options: { saveDom: boolean } = {}) => {
+        const countStr: string = (browser["screenshot-count"]++)
+          .toFixed()
```
mstrasinskis, line 16: Is it possible to have not integer here?

mstrasinskis (approved):
LGTM. With a couple of questions.

e2e-tests/wdio.conf.ts
mstrasinskis: Thank you

e2e-tests/wdio.conf.ts
bitdivine: Yes, the transformation is now in the screenshot function.  Good catch, thank you.

e2e-tests/wdio.conf.ts
bitdivine: Not with the code as it is.  If the code gets changed in future, yes, anything is possible.  I can code this defensively; usually a good idea.

PR #680, opened 2022-04-10
Selectors used by e2e tests

bitdivine:
# Motivation
In e2e tests, DOM selectors are currently provided as strings within the tests themselves.  It seems preferable to define once what selectors are expected to work and then use those.

Note: `npm run lint` does NOT fail if say `IIConfirmRedirectPage.NONEXISTENT_SELECTOR` is used.  This is, to me, unexpected.  I see no obvious reason for this omission.

# Changes
- Define DOM selectors for e2e strings
- Add more DOM selectors.
- Consider naming these classes distinctively.  These classes are not based on the implemented code structure but on the views, as as UX designer would see them.  If we have a class with DOM tags and wdio navigation functions for say the login page or neurons tab, we would not want that class to be confused with the similarly named svelte class that implements the functionality.

# Future
- replace DOM selectors everywhere with these predefined strings.

# Tests
I have used these seelctors in e2e tests while developing locally, however I would prefer to commit these separately to establish the naming scheme early.


e2e-tests/components/ii-confirm-redirect.ts
```diff
@@ -0,0 +1,3 @@
+export class IIConfirmRedirectPage {
```
lmuntanerDfinity, line 1: Why not using an enum?

```
enum IIConfirmRedirectPage {
  CONFIRM_REDIRECT_BUTTON_DELECTOR = '#confirmRedirect';
}
```

Same for the rest.

lmuntanerDfinity, line 1: Why not using an enum?

```
enum IIConfirmRedirectPage {
  CONFIRM_REDIRECT_BUTTON_DELECTOR = '#confirmRedirect';
}
```

Same for the rest.

lmuntanerDfinity (commented):
Not sure why `npm run lint` does not fail with a missing property, it should.

Did you try with `strict: true` in the `tsconfig.json`?

e2e-tests/components/ii-confirm-redirect.ts
bitdivine: @lmuntanerDfinity  I have methods as well.  See e..g `skipAddingRecoveryMechanism(...)` below.  It seems natural to me to have all the tools needed to use a component accessible from the same container.

lmuntanerDfinity: Ok, makes sense to me to use classes if we want to reuse some functionality as well.

lmuntanerDfinity (approved):
Approving already, but I added a question.

PR #682, opened 2022-04-10
Environment argument for controlling the log level

bitdivine:
# Motivation
The e2e log level is too noisy for most purposes, it floods the screen with hard-to-read data and by its sheer volume hides errors.  Personally, when developing I use `logLevel: warn` and switch back to `logLevel: info` if I really need the data for debugging.

However, when changing the log level in the code, one ends up with a diff which I have to remove every time I commit or make a PR.  Better to have a default in the code and an external override.

# Changes
- Override the log level with an environment variable.

# Tests
I use this control regularly.  It works.


lmuntanerDfinity (approved):
LGTM!

Aren't you using it anywhere for now?

PR #683, opened 2022-04-10
Fix flaky tests: Wait for canister to be ready

bitdivine:
# Motivation
Sometimes tests fail because the canister installation wasn't ready:
```
2022-04-10T05:44:35.8392589Z [0-0]   Code: 404 (Not Found)
2022-04-10T05:44:35.8393240Z [0-0]   Body: Requested canister does not exist
2022-04-10T05:44:35.8394445Z [0-0] " class="svelte-ksskaf">An error occurred while loading the accounts information. Server returned an error:
2022-04-10T05:44:35.8397503Z [0-0]   Code: 404 (Not Found)
2022-04-10T05:44:35.8397996Z [0-0]   Body: Requested canister does not exist
```

Note: I checked whether the installation had failed; the installation was fine, some later checks worked, it just took a while for the canister to be ready.

# Changes
- Support `nobuild` for faster testing.
- Wait for the local network to be ready before deploying canisters.
- Wait for the nns-dapp canister to be ready before running tests.

# Tests
Running tests repeatedly on CI:  Success: 2/2
Actually, I can just do this and see the results below:
```
seq 10 | while read line ; do git pull ; git commit -m bump-$line --allow-empty ; git push ; sleep 30m ; done
```

### DIFF
--- a/e2e-tests/scripts/deploy-local-canisters
+++ b/e2e-tests/scripts/deploy-local-canisters
@@ -8,7 +8,8 @@ while (($# > 0)); do
   arg="$1"
   shift 1
   case "$arg" in
-  --help) cat <<-EOF
+  --help)
+    cat <<-EOF
 
 	Deploys nns-dapp and required canisters to localhost.
 
@@ -19,7 +20,8 @@ while (($# > 0)); do
 	  --nobuild == set DEPLOY_ENV=nobuild so that the existing wasm is used.
 
 	EOF
-        exit 0 ;;
+    exit 0
+    ;;
   --nobuild) DEPLOY_ENV=nobuild ;;
   esac
 done
--- a/e2e-tests/scripts/deploy-local-canisters
+++ b/e2e-tests/scripts/deploy-local-canisters
@@ -8,6 +8,18 @@ while (($# > 0)); do
   arg="$1"
   shift 1
   case "$arg" in
+  --help) cat <<-EOF
+
+	Deploys nns-dapp and required canisters to localhost.
+
+	Usage: "$0 [--nobuild] [--help]"
+
+	Flags:
+	  --help == print this help
+	  --nobuild == set DEPLOY_ENV=nobuild so that the existing wasm is used.
+
+	EOF
+        exit 0 ;;
   --nobuild) DEPLOY_ENV=nobuild ;;
   esac
 done
--- a/e2e-tests/scripts/deploy-local-canisters
+++ b/e2e-tests/scripts/deploy-local-canisters
@@ -17,9 +17,10 @@ wait_for_url() {
   local seconds="${2:-10}"
   local options="${3:-}"
   for ((i = 0; i < seconds; i++)); do
-    if curl ${options:+${options}} -so /dev/null "$url"
-    then return 0
-    else sleep 1
+    if curl ${options:+${options}} -so /dev/null "$url"; then
+      return 0
+    else
+      sleep 1
     fi
   done
   echo "URL not ready after ${seconds}s: curl ${options} -so /dev/null $url" >&2
--- a/e2e-tests/scripts/deploy-local-canisters
+++ b/e2e-tests/scripts/deploy-local-canisters
@@ -1,5 +1,34 @@
 #!/usr/bin/env bash
 set -euo pipefail
 cd "$(dirname "$0")/../.."
+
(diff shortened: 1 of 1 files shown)


e2e-tests/scripts/deploy-local-canisters
```diff
@@ -1,5 +1,56 @@
...
+  for ((i = 0; i < seconds; i++)); do
+    if curl ${options:+${options}} -so /dev/null "$url"; then
+      return 0
+    else
+      sleep 1
+    fi
+  done
+  echo "URL not ready after ${seconds}s: curl ${options:+${options}} -so /dev/null $url" >&2
+  return 1
+}
```
bitdivine, line 43: The two calls to this function address the problem of the network or canister not being ready.

e2e-tests/scripts/deploy-local-canisters
```diff
@@ -1,5 +1,58 @@
 #!/usr/bin/env bash
 set -euo pipefail
 cd "$(dirname "$0")/../.."
+
+DEPLOY_ENV="${DEPLOY_ENV:-local}"
+NETWORK_URL="localhost:8080"
+while (($# > 0)); do
```
lmuntanerDfinity, line 7: Sorry, I don't understand this loop. Why do we need a loop to deploy the canisters?

e2e-tests/scripts/deploy-local-canisters
```diff
@@ -1,5 +1,58 @@
...
+
+: "Creating and deploying canisters..."
 dfx canister --no-wallet create --all || echo Canisters may have been created already. Proceeding...
-REDIRECT_TO_LEGACY=svelte DEPLOY_ENV=local IDENTITY_SERVICE_URL=http://localhost:8087 dfx deploy --no-wallet --network local --argument '(null)'
+REDIRECT_TO_LEGACY=svelte DEPLOY_ENV="$DEPLOY_ENV" IDENTITY_SERVICE_URL=http://localhost:8087 dfx deploy --no-wallet --network local --argument '(null)'
+
+: "Waiting for the canister to return with HTTP code 200..."
+NNS_DAPP_CANISTER_ID="$(dfx canister --no-wallet --network local id nns-dapp)"
+NNS_DAPP_CANISTER_URL="${NNS_DAPP_CANISTER_ID}.${NETWORK_URL}"
+wait_for_url "$NNS_DAPP_CANISTER_URL" 10 --fail
```
lmuntanerDfinity, line 56: I would have expected the command `dfx deploy` to finish once the canisters are ready. But it's good to know that's not the case.

e2e-tests/scripts/deploy-local-canisters
bitdivine: @lmuntanerDfinity 

What this does: This reads the command line arguments. `$#` is the number of arguments.  Below the first argument is read with `arg=$1` and removed with `shift 1`.  Were there a flag that consumed multiple arguments, say purely hypothetically, `--someflag somearg1 somearg2` the additional arguments would have to be removed as well, in this example with `shift 2`.

Why: It is not needed to resolve the CI flakiness, however I needed it to debug the CI flakiness quickly and I suspect that this won't be the last time that there is a need to redeploy repeatedly so I left it in, and made it pretty.

bitdivine: Example of similar code: (no, I didn't copy this from SO but just to give some context)  https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash

I would critique the SO code as follows:
- If any of the branches in the case clause forgets to shift out the consiumed argument, th euser will get an infinite loop.  By shifting out the arg at the beginning we avoid that.
- `[[ $# -gt 0 ]]` uses the generic logical structures where as `(( ... ))` is dedicated to arithmetic.  This maybe just stylistic, the code shown works and I would allow it into a PR, but using `((` reduces the number of ways things could go wrong from a careless edit.

e2e-tests/scripts/deploy-local-canisters
bitdivine: Me too!  Why it does not the case I do not know.  Maybe dfx exits as soon as the wasm has been uploaded successfully, but before the canister has "booted".

lmuntanerDfinity (approved):

PR #686, opened 2022-04-11
refactor: style padding calc

peterpeterparker:
# Motivation

We often got the exact same `calc(... * var(-padding))` values in the code. We can spare some style duplication.

# Changes

- refactor common padding style

### DIFF
--- a/frontend/svelte/src/lib/themes/variables.scss
+++ b/frontend/svelte/src/lib/themes/variables.scss
@@ -5,6 +5,11 @@
   --icon-width: 24px;
 
   --padding: 10px;
+
+  --padding-half: calc(var(--padding) / 2);
+  --padding-2x: calc(2 * var(--padding));
+  --padding-3x: calc(3 * var(--padding));
+
   --border-radius: 10px;
 
   --z-index: 1;
@@ -15,7 +20,7 @@
   --footer-height: 20vh;
 
   --modal-small-width: calc(100% - (8 * var(--padding)));
-  --modal-medium-width: calc(100% - (2 * var(--padding)));
+  --modal-medium-width: calc(100% - var(--padding-2x));
 
   --animation-time-normal: 0.2s;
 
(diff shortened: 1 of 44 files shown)


lmuntanerDfinity (approved):
I agree with the refactor. But I think that before doing any kind of refactor you should mention it, and we should agree to it.

This has happened a few times already, and I don't feel comfortable seeing a PR that changes 50 files or refactors a whole piece of functionality without discussing it first.

frontend/svelte/src/lib/themes/variables.scss
```diff
@@ -5,6 +5,11 @@
   --icon-width: 24px;
 
   --padding: 10px;
+
+  --padding-half: calc(var(--padding) / 2);
+  --padding-2x: calc(2 * var(--padding));
+  --padding-3x: calc(3 * var(--padding));
```
mstrasinskis, line 11: What about 3,4 and even 8?

mstrasinskis (approved):
LGTM, just one question

frontend/svelte/src/lib/themes/variables.scss
mstrasinskis: There is also 7 with 1.5

peterpeterparker: Sure, will do

peterpeterparker: How should we called 1.5?

peterpeterparker: @mstrasinskis I have added 1.5, 2.5, 4 and 8

There is still three different usage of `calc` and `padding` thought that there are local custom values so I did not extract these. We can always extract later too.

peterpeterparker:
> This has happened a few times already, and I don't feel comfortable seeing a PR that changes 50 files or refactors a whole piece of functionality without discussing it first.

@lmuntanerDfinity PR took 10 min tops of development therefore it made specially sense for efficiency reason to first create it and discuss afterwards, what we did.

PR #670, opened 2022-04-07
Feat: Neuron authorization improvements

lmuntanerDfinity:
# Motivation

Add extra authorization checkpoints for neuron functionalities.

# Changes

* Helper to assert that neuron is controlled by user.
* Helper to assert that neuron is controllable by accounts.
* User helpers in services.
* Hide button if user not authorized in Neuron Details.

# Tests

* Upgrade and add tests to neuron services.
* New tests for each assert.

PR #687, opened 2022-04-11
fix: neuron utils wildcard import

peterpeterparker:
# Motivation

Follow up PR #670 to fix wildcard import.

# Changes

- es6 naming import instead of wildcard to helps bundler chunking process (if we ever use chunking)


lmuntanerDfinity (changes requested):
Check also the `NeuronMaturityCard` since you are at it, please. I did the same there.

peterpeterparker:
@lmuntanerDfinity thanks, good idea! done

lmuntanerDfinity (approved):

PR #698, opened 2022-04-12
fix: reset proposals filter on back from detail page

peterpeterparker:
# Motivation

The filters and per extension the proposals should not be resetted when navigating from the proposal detail page to the voting tab.

In addition we also noticed that older proposal have sometimes undefined title.

# Changes

- reset proposal store when filter changes only once mounted
- if title is undefined display an empty title not undefined


mstrasinskis (approved):
LGTM. Thanks!

PR #702, opened 2022-04-12
feat: do not provide "Unspecified" as proposal filter option

peterpeterparker:
# Motivation

We shouldn't include Unknown or Unspecified as options for the filters, since those values should never appear on proposal fields.

# Changes

- do not provide `Unspecified` respectively `0` value as option for the proposal filters


mstrasinskis (approved):
LGTM, thanks 👍

PR #718, opened 2022-04-18
feat: chunks get balance calls to ledger

peterpeterparker:
# Motivation

A user is reporting an issue in production that leads to the UI not being usable. The or parts of the problem seems to be the fact that the ledger throws errors to multiple / many requests (see screenshot on the [forum](https://forum.dfinity.org/t/nns-unusable-in-chrome/12224)).

Flutter used to chunks to max 10 parallel requests the balance queries to the ledger (see [code](https://github.com/dfinity/nns-dapp/blob/e9dd304d8aaa96f82e91936f23d1bf0c781ae9c5/frontend/ts/src/canisters/ledger/Service.ts#L35)) what is not implemented in the Svelte.

This PR reimplements such chunking method.

# Note

Ultimately we might need to add this to any ledger queries including loading the balance of all subaccounts?

# Changes

- chunk ledger query `accountBalance` to get the neurons balance


mstrasinskis (approved):
LGTM. Do you see an easy way to prevent `listNeurons` to be invoked before the previous call is done?

PR #719, opened 2022-04-18
feat: init neurons store for voting only once if needed

peterpeterparker:
# Motivation

A user has reported the "Voting - Proposal detail page" UI being "hardly usable" (see [forum](https://forum.dfinity.org/t/nns-unusable-in-chrome/12224)) because lots of queries are performed. Root cause of the issue isn't absolutely clear but [#718](https://github.com/dfinity/nns-dapp/pull/718) should improve the performance.

While comparing Flutter and Svelte app, I noticed that the Svelte app list the neurons each time the "Voting - Proposal detail page" is accessed while Flutter relies on the store being loaded once. That's why in order to minimize the queries users with many neurons would have to execute, I replicated this behavior to the Svelte app.

This PR improves the "Voting - Proposal detail page" as following:

- if the neurons store has been initialized once, then it does not list the neurons with the backend but will use neurons store
- it also consider the store has being initialized when the certified data have been received
- finally, enhance the usage of the busy store when user vote to the reload of the data

Overall this should result in producing less queries on the backend and improving the back and forth navigation from "Voting tab" to "Proposal detail page".

However, first loading and voting will feel a bit slower because the app will be considered as stable when data are reloaded with certified data. Depends if you see the glass half full of half empty, in this particular case I rather like the feedback "the app is a bit slow" better than "the app is barely usable" 😉.

# Note

I provide the PR as a draft for discussion but if you like the solution, worth to note it is from my point of view implemented and can be reviewed if you wish to.

# Changes

- following the pattern of the `proposal.store`, add the `certified` information within the `neurons.store`
- change the initial value of the neurons in store from `[]` to `undefined` in order to be able to detect if the store has been initialized or not
- add a derived store that prevent `undefined` neurons to be used for feature that are implicitly called when the store is initialized
- use the `busy.store` when reloading the proposal and neurons after voting

# Demo

ezgif com-gif-maker


frontend/svelte/src/lib/stores/neurons.store.ts
```diff
@@ -5,46 +5,65 @@ import {
   sortNeuronsByCreatedTimestamp,
 } from "../utils/neuron.utils";
 
-export type NeuronsStore = NeuronInfo[];
+export interface NeuronsStore {
+  neurons?: NeuronInfo[];
+  certified: boolean | undefined;
```
mstrasinskis, line 10: Probably it makes sense to add a comment that the `certified` is an optimistic value because of its rewriting by `pushNeurons`.

mstrasinskis (approved):
LGTM 👍 Just one comment.

frontend/svelte/src/lib/stores/neurons.store.ts
peterpeterparker: Sure, done

PR #721, opened 2022-04-19
Feat/request and response logs

mstrasinskis:
# Motivation

To get more information from user side the log messages were added.

# Changes

- added logWithTimestamp
- requests log their status with some details (e.g. neuronId)
  - to not disturb people with raw data in dev.console all IDs are converted into hashes (see `hashCode`)
- to be sure that the app doesn't try to vote w/ duplications `registerVotes` does extra step to be sure neuronIds are unique with extra console.error if not.
- voting error handling was updated to rely on all rejected requests (not only with human readable message

# Tests

- queryAndUpdate should log

# Notes

- Probably it's better to replace `hashCode` with something random based...

# Screenshots

https://user-images.githubusercontent.com/98811342/164051948-1e830e94-8b46-4d31-a4e4-0bda830404e9.mov


peterpeterparker (approved):
LGTM, thanks

PR #722, opened 2022-04-19
feat: "Loading your neurons..." spinner

peterpeterparker:
# Motivation

We want to voting feature to be available only when the neurons store has been fully initialized but we also want to provide a smooth UI. That's why instead of hiding the all proposal detail page while we load the information we need we can display the static content - i.e. the proposal detail - as fast as we can and display another spinner dedicated to the loading of the neurons.

Worth to note that the store is now loaded only once, therefore user should only face this spinner the first time the store is loaded.

# Changes

- split `ProposalDetail` checks for `proposalInfoStore` and `neuronsReady` to display two distinct spinner - i.e. display the `ProposalDetailCard` as fast as possible

# Screenshot

ezgif com-gif-maker (2)

### DIFF
--- a/frontend/svelte/src/routes/ProposalDetail.svelte
+++ b/frontend/svelte/src/routes/ProposalDetail.svelte
@@ -126,7 +126,7 @@
               <Spinner />
             </div>
 
-            <span><small>{$i18n.proposal_detail__vote.loading_neurons}</small></span>
+            <span><small>{$i18n.proposal_detail.loading_neurons}</small></span>
           </div>
         {/if}
       {:else}
--- a/frontend/svelte/src/routes/ProposalDetail.svelte
+++ b/frontend/svelte/src/routes/ProposalDetail.svelte
@@ -110,17 +110,42 @@
     >
 
     <section>
-      {#if $proposalInfoStore && neuronsReady}
+      {#if $proposalInfoStore}
         <ProposalDetailCard proposalInfo={$proposalInfoStore} />
-        <VotesCard proposalInfo={$proposalInfoStore} />
-        <VotingCard proposalInfo={$proposalInfoStore} />
-        <IneligibleNeuronsCard
-          proposalInfo={$proposalInfoStore}
-          neurons={$definedNeuronsStore}
-        />
+
+        {#if neuronsReady}
+          <VotesCard proposalInfo={$proposalInfoStore} />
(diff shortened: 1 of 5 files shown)


frontend/svelte/src/routes/ProposalDetail.svelte
```diff
@@ -110,17 +110,42 @@
...
         <Spinner />
       {/if}
     </section>
   </HeadlessLayout>
 {/if}
+
+<style lang="scss">
+  .loader {
+    display: flex;
+    flex-direction: column;
```
mstrasinskis, line 142: Will it look w/o `flex-direction: column;`?

frontend/svelte/src/tests/routes/ProposalDetail.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import en from "../mocks/i18n.mock";
+import { MockLedgerCanister } from "../mocks/ledger.canister.mock";
+import { buildMockNeuronsStoreSubscribe } from "../mocks/neurons.mock";
+import {
+  mockEmptyProposalsStoreSubscribe,
+  mockProposals,
+} from "../mocks/proposals.store.mock";
+import { mockRouteStoreSubscibe } from "../mocks/route.store.mock";
+
+describe("NeuronDetail", () => {
```
mstrasinskis, line 23: There is a typo in the word "Proposal" ;)

mstrasinskis, line 23: There is a typo in the word "Proposal" ;)

mstrasinskis (approved):
LGTM, just a minor comments.
Thanks for the test 👍

frontend/svelte/src/tests/routes/ProposalDetail.spec.ts
peterpeterparker: These copy paste 😅

frontend/svelte/src/routes/ProposalDetail.svelte
peterpeterparker: Without `flex-direction: column;`

Capture d’écran 2022-04-20 à 06 52 10

Without `flex`

Capture d’écran 2022-04-20 à 06 51 28

PR #723, opened 2022-04-19
fix: proposal filters counter

peterpeterparker:
# Motivation

Because we removed the "Unspecified" options of the proposal filters, the counter total was not displaying the correct total value anymore - e.g. 5/11 instead of 5/10 filters.

In addition, found out that `Topic` was used in a generic utils that was fortunately only used so far for such filter.

# Changes

- fix counter without "Unspecified" value
- fix generic utils

# Screenshot

Counters are now correct:

Capture d’écran 2022-04-19 à 19 10 30


mstrasinskis (approved):
LGTM, thanks 👍

PR #724, opened 2022-04-19
refactor: move principal stringifying to nns-dapp

mstrasinskis:
# Motivation

To avoid changing the global class in `nns-js` (see snippet below) this logic was moved to `nns-dapp`
```ts
// Represent Principals as strings rather than as byte arrays when serializing to JSON strings
// eslint-disable-next-line @typescript-eslint/no-explicit-any
(Principal.prototype as any).toJSON = function () {
  return this.toString();
};
```

# Changes

- util `stringifyJson` was extended and refactored to look cleaner and be ready for the next challenges!
- temp UI improvement (2. screen) (before we switched to a real json-view renderer)

# Tests

- stringifyJson should render principal as text

# Screens

## principal rendering
Screenshot 2022-04-19 at 22 27 09

```console.log(value, value.toString());```

## preserve text formatting
image


bitdivine:
I actually rather liked your solution in https://github.com/dfinity/nns-js/pull/80/files .  Can not that solution be pushed upstream, to https://github.com/dfinity/agent-js/blob/main/packages/principal/src/index.ts#L98 ?

peterpeterparker (approved):
LGTM, thanks

bitdivine (approved):

PR #731, opened 2022-04-20
fix: filters have no effect after back from detail page

peterpeterparker:
# Motivation

PR [#698](https://github.com/dfinity/nns-dapp/pull/698) had a missing flow. It indeed did not reset the list of proposals but had for effect to make the changes of the filters ignored after navigating back from the proposal detail page.

This PR resolves the issue.

# Changes

- if we don't reset the filter and list explicitly, we can consider the component as `initialized`


mstrasinskis (approved):
Thanks!

PR #735, opened 2022-04-21
fix: hide proposal without ballots when filter-checkbox selected

mstrasinskis:
# Motivation

Proposals with `votingPower===0` are visible after enabling the filter (Hide open...) but users was not able to vote using `votingPower===0 neurons`. This PR enables voting with `votingPower===0` neurons (related nns-js [PR](https://github.com/dfinity/nns-js/pull/86))
Was tested on prod-nns and both-nns

# Changes

- updated validation to be based on selected count instead of total > 0

# Tests

- added tests for button disable state based on selection

# Screenshots
selected text in dev console
image

image

image


frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
```diff
@@ -54,6 +56,29 @@ describe("VotingCard", () => {
...
+    neuronsStore.setNeurons({ neurons, certified: true });
+    const { container } = render(VotingCard, {
+      props: {
+        proposalInfo,
+      },
+    });
+    votingNeuronSelectStore.reset();
+    expect(container.querySelector("button[disabled]")).toBeNull();
+  });
+
```
bitdivine, line 81: I don't get the difference between these two tests.  If you wait a tick, the result should be 2, if you don't wait, the result should be null.  Is that correct?  Why is this interesting information?

mstrasinskis: Good catch! The test was updated. The reset is not needed here and the test works only because reset needs some time.

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingConfirmationToolbar.svelte
```diff
@@ -37,13 +39,13 @@
 <div role="toolbar">
   <button
     data-tid="vote-yes"
-    disabled={total === 0n}
+    {disabled}
```
peterpeterparker, line 42: should this be discussed with Mischa?

does the user understand why it is disabled...

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
```diff
@@ -54,6 +56,30 @@ describe("VotingCard", () => {
...
+  it("should enable action buttons any neuron is selected", async () => {
+    // changing the neuronStore automatically updates votingNeuronSelectStore with initial pre-selection of all neurons
+    neuronsStore.setNeurons({ neurons, certified: true });
+    const { container } = render(VotingCard, {
+      props: {
+        proposalInfo,
+      },
+    });
+    await tick();
```
lmuntanerDfinity, line 79: Why do you need a tick here?

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
```diff
@@ -54,6 +56,30 @@ describe("VotingCard", () => {
...
+      },
+    });
+    votingNeuronSelectStore.reset();
+    await tick();
+    expect(container.querySelectorAll("button[disabled]").length).toBe(2);
+  });
+
+  it("should enable action buttons any neuron is selected", async () => {
+    // changing the neuronStore automatically updates votingNeuronSelectStore with initial pre-selection of all neurons
+    neuronsStore.setNeurons({ neurons, certified: true });
```
lmuntanerDfinity, line 73: do you reset the store after every test?

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
```diff
@@ -54,6 +56,30 @@ describe("VotingCard", () => {
...
 
+  it("should disable action buttons if no neurons selected", async () => {
+    neuronsStore.setNeurons({ neurons, certified: true });
+    const { container } = render(VotingCard, {
+      props: {
+        proposalInfo,
+      },
+    });
+    votingNeuronSelectStore.reset();
+    await tick();
```
lmuntanerDfinity, line 67: ditto

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
```diff
@@ -54,6 +56,30 @@ describe("VotingCard", () => {
...
   });
 
+  it("should disable action buttons if no neurons selected", async () => {
+    neuronsStore.setNeurons({ neurons, certified: true });
+    const { container } = render(VotingCard, {
+      props: {
+        proposalInfo,
+      },
+    });
+    votingNeuronSelectStore.reset();
```
lmuntanerDfinity, line 66: isn't this the default? To not have anything selected?

lmuntanerDfinity (commented):
A few questions

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
mstrasinskis: Vice versa all neurons should be preselected. Changing the neuronStore automatically updates votingNeuronSelectStore with initial pre-selection of all neurons (src/lib/components/proposal-detail/VotingCard/VotingCard.svelte#34)

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
mstrasinskis: The tick is needed because the auto pre-selection occurs in the store subscription of the VotingCard component (comment was added)

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
mstrasinskis: Every test relies on the neuronsStore and makes changes to it. So there is no explicit store reset. But good point about reseting -- `restoreAllMocks` was added.

lmuntanerDfinity: If every test relies on the neuronsStore, shouldn't we need to reset it after the test and make sure that at the beginning of the test the store is as it should be then?

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
lmuntanerDfinity: ok, thansk!

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
lmuntanerDfinity: ok, got it!

lmuntanerDfinity (approved):
Approving, just a final question about tests.

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
mstrasinskis: removed

frontend/svelte/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
mstrasinskis: Added store reset in beforeEach and afterAll

PR #745, opened 2022-04-25
refactor: rename path id functions

peterpeterparker:
# Motivation

Make the goal and context of the function that read the ids from the route path more obvious - less generic.

# Changes

- rename `getProposalId` to `routePathProposalId`
- rename `getNeuronId` to `routePathNeuronId`


mstrasinskis (approved):
LGTM 👍

PR #746, opened 2022-04-25
fix: 'await' has no effect on the type of this expression

peterpeterparker:
# Motivation

eslint warning "'await' has no effect on the type of this expression"

# Changes

- remove unused response object


mstrasinskis (approved):
LGTM. Thank you.

PR #747, opened 2022-04-25
docs: fix jsdoc params definition

peterpeterparker:
# Motivation

eslint warning "JSDoc '@param' tag has name 'callback', but there is no parameter with that name. "

# Changes

- declare first the param object and then the param variables in jsdoc


mstrasinskis (approved):
LGTM Thanks!

PR #748, opened 2022-04-25
chore: missing types for log function

peterpeterparker:
# Motivation

eslint warning "Rest parameter 'args' implicitly has an 'any[]' type, but a better type may be inferred from usage. "

# Changes

- define args as type `string[]`


mstrasinskis (approved):
LGTM

PR #749, opened 2022-04-25
feat: toast display as success instead of info

peterpeterparker:
# Motivation

The toast is actually used to display "success" messages and not information. Therefore this PR makes its color matching its purpose.

# Changes

- toast message `info` refactored to `success`
- colors from dark to "green" success color

# Screenshot

Was:

Capture d’écran 2022-04-25 à 11 47 46

Suggestion:

Capture d’écran 2022-04-25 à 11 46 56

### DIFF
--- a/frontend/svelte/src/lib/types/toast.ts
+++ b/frontend/svelte/src/lib/types/toast.ts
@@ -1,4 +1,4 @@
-export type ToastLevel = "info" | "warn" | "error";
+export type ToastLevel = "success" | "warn" | "error";
 
 export interface ToastMsg {
   labelKey: string;
--- a/frontend/svelte/src/lib/components/neuron-detail/actions/DissolveActionButton.svelte
+++ b/frontend/svelte/src/lib/components/neuron-detail/actions/DissolveActionButton.svelte
@@ -39,7 +39,7 @@
     if (id !== undefined) {
       toastsStore.show({
         labelKey: successKey,
-        level: "info",
+        level: "success",
       });
     }
     closeModal();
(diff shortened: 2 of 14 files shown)


frontend/svelte/src/lib/types/toast.ts
```diff
@@ -1,4 +1,4 @@
-export type ToastLevel = "info" | "warn" | "error";
+export type ToastLevel = "success" | "warn" | "error";
```
lmuntanerDfinity, line 1: Don't we want to keep both? Art the moment we don't have any, but it would make sense to have an "info" for some use cases.

lmuntanerDfinity: I created an issue for this: https://dfinity.atlassian.net/browse/L2-447

I spoke with Mischa, and he agreed that some notifications, on demand, should disappear after a few seconds.

Something like expanding the `ToastMsg` type with `autoClose: boolean` for example.

We can either implement it here or in another PR. Just FYI,

peterpeterparker: > I spoke with Mischa, and he agreed that some notifications, on demand, should disappear after a few seconds.
> ...
> We can either implement it here or in another PR 

Will do then

> Don't we want to keep both? 

we don't use "info" currently at all, therefore my suggestion would be to not keep it but if you rather like to, I can reintroduce it. your call.

lmuntanerDfinity: no, it's ok then, let's remove "info". No strong opinion.

peterpeterparker:
@lmuntanerDfinity here you go for a new review.

note: scope has evolved a bit so description has been also updated accordingly.

peterpeterparker:
> Did you check the "show" after rebasing?

@lmuntanerDfinity yes

frontend/svelte/src/lib/stores/toasts.store.ts
```diff
@@ -18,11 +20,12 @@ const initToastsStore = () => {
       update((messages: ToastMsg[]) => [...messages, msg]);
     },
 
+    success({ labelKey }: Pick<ToastMsg, "labelKey">) {
+      this.show({ labelKey, level: "success", duration: 4000 });
```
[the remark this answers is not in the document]
peterpeterparker: Sure, done

PR #750, opened 2022-04-25
feat: close accounts action modal on error too

peterpeterparker:
# Motivation

Follow defined UX rule "actions modals should be closed on action error"

# Changes

- close modal "add sub account" in case of submit error
- close modal "transfer icp" in case of submit error

### DIFF
--- a/frontend/svelte/src/lib/services/accounts.services.ts
+++ b/frontend/svelte/src/lib/services/accounts.services.ts
@@ -2,6 +2,10 @@ import type { Identity } from "@dfinity/agent";
 import { createSubAccount, loadAccounts } from "../api/accounts.api";
 import { sendICP } from "../api/ledger.api";
 import { toSubAccountId } from "../api/utils.api";
+import {
+  NameTooLongError,
+  SubAccountLimitExceededError,
+} from "../canisters/nns-dapp/nns-dapp.errors";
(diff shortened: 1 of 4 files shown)


frontend/svelte/src/lib/services/accounts.services.ts
```diff
@@ -42,11 +46,25 @@ export const addSubAccount = async ({
...
+  try {
+    const identity: Identity = await getIdentity();
 
-  await createSubAccount({ name, identity });
+    await createSubAccount({ name, identity });
 
-  await syncAccounts();
+    await syncAccounts();
+  } catch (err: unknown) {
+    const labelKey =
```
lmuntanerDfinity, line 56: Maybe you can use a mapper like `mapNeuronErrorToToastMessage` but for accounts.

This way you can reuse it in all the `catch`.

Maybe not in the PR, though.

lmuntanerDfinity (approved):
LGTM, just one comment

frontend/svelte/src/lib/services/accounts.services.ts
peterpeterparker: Sure feel free to refactor it in another PR (refactor the error does not match the PR's goal) 👍

PR #751, opened 2022-04-25
chore: remove redundant returns

mstrasinskis:
# Motivation

Remove redundancy - await and return response in `Promise<void>` functions

# Tests

- no test is affected


peterpeterparker (approved):
Thx!

PR #511, opened 2022-03-01
Display the ManageNeuron topic if a neuron has followees on it

hpeebles:
# Motivation

We currently hide the ManageNeuron topic from the followee options. But some neurons may be created with followees on the ManageNeuron topic. In these cases it is currently impossible for the neuron owner to remove the followee via the UI.

# Changes

- Show the ManageNeuron followee configuration options for neurons who already have followees on the ManageNeuron topic.

# Tests

I removed the topic filter, then followed a neuron on the ManageNeuron topic, then applied this change. I could then see the ManageNeuron topic on the followee options page. But when I removed the neuron from the ManageNeuron topic, the topic itself disappeared, as intended.

PR #753, opened 2022-04-25
Fix/filter out manage neuron topic

mstrasinskis:
# Motivation

According to [Flutter](https://github.com/dfinity/nns-dapp/pull/511)
- NeuronManagement proposals are not public so we hide this topic
- unless the neuron already has followees on this topic

# Changes

- refactoring to switch to followeesByTopic util function
- apply topicsToFollow to visible topics

# Tests

- followeesByTopic
- topicsToFollow

# Screens

## Before
image

## After
image

### with ManageNeuron followee

image

image


frontend/svelte/src/tests/lib/utils/neuron.utils.spec.ts
```diff
@@ -879,4 +882,108 @@ describe("neuron-utils", () => {
...
+    it("should return topics without ManageNeuron", () => {
+      expect(topicsToFollow(neuronWithoutManageNeuron)).toEqual(
+        enumValues(Topic).filter((topic) => topic !== Topic.ManageNeuron)
+      );
+      expect(topicsToFollow(neuronWithoutFollowees)).toEqual(
+        enumValues(Topic).filter((topic) => topic !== Topic.ManageNeuron)
+      );
+    });
+
+    it("should return topics with ManageNeuron if already voted for it", () => {
```
lmuntanerDfinity, line 985: typo I guess, hehe. If neuron follows some neuron on the ManageNeuron topic.

lmuntanerDfinity (approved):
Thanks, LGTM, only a test description change.

Björn just mentioned to add this feature for heavy users: https://dfinity.slack.com/archives/C01S03NBM7S/p1650656373629389

frontend/svelte/src/lib/utils/neuron.utils.ts
```diff
@@ -426,3 +428,28 @@ export const isIdentityController = ({
...
 };
+
+export const followeesByTopic = ({
+  neuron,
+  topic,
+}: {
+  neuron: NeuronInfo | undefined;
+  topic: Topic;
+}): Followees | undefined =>
+  neuron === undefined
```
peterpeterparker, line 439: nth: instead if the comparison why not `neuron?.fullNeuron...`

if neuron is undefined then it returns undefined too or am I missing something?

frontend/svelte/src/lib/utils/neuron.utils.ts
```diff
@@ -426,3 +428,28 @@ export const isIdentityController = ({
...
+/**
+ * NeuronManagement proposals are not public so we hide this topic
+ * (unless the neuron already has followees on this topic)
+ * https://github.com/dfinity/nns-dapp/pull/511
+ */
+export const topicsToFollow = (neuron: NeuronInfo): Topic[] =>
+  enumValues(Topic).filter(
+    (topic) =>
+      topic !== Topic.ManageNeuron ||
+      followeesByTopic({ neuron, topic: Topic.ManageNeuron }) !== undefined
```
peterpeterparker, line 454: nth: for performance reason would maybe be better to assign `followeesByTopic` to a `const` before `filter`

peterpeterparker (approved):
LGTM, thanks. Two small nth comments

frontend/svelte/src/tests/lib/utils/neuron.utils.spec.ts
mstrasinskis: Good catch, thanks!

frontend/svelte/src/lib/utils/neuron.utils.ts
mstrasinskis: Good point 👍 Updated .

frontend/svelte/src/lib/utils/neuron.utils.ts
mstrasinskis: Relic. Removed.

PR #757, opened 2022-04-26
Fix/remove security utils

mstrasinskis:
# Motivation

1. because of CSP the sanitising is not needed on FE side. 
2. the current solution is not reliable (doesn't prevent all types of XSS attacks)

# Changes

- removed security.utils

# Tests

- removed security.utils tests


peterpeterparker (approved):
LGTM

mstrasinskis:
The functionality was replaced in another [PR](https://github.com/dfinity/nns-dapp/pull/760)

PR #758, opened 2022-04-26
feat: init add hardware wallet wizard

peterpeterparker:
# Motivation

- Initialize "Add hardware wallet to account" wizard
- Add field to enter hardware wallet name

# Changes

- new functional `HardwareWallerName` and dummy `HardwareWalletConnect` components
- refactor `SelectTypeAccount` to `AddAccountType`
- refactor `AddNewAccount` to `AddNewSubAccount`
- update `AddAccountModal` to support dynamic choice of two different types of wizard tracks
- add `add-account.store` store scoped to wizard context

# Screenshots

ezgif com-gif-maker (3)


frontend/svelte/src/lib/modals/accounts/AddAccountModal.svelte
```diff
@@ -1,31 +1,83 @@
...
+  const selectType = async (type: AccountType) => {
+    // Set the type in store and reset other values only if the new type is not the one that was previously used - e.g. user first select hardware wallet, entered a name, clicked continue, went twice back and go to subaccount
+    addAccountStore.update(({ type: previousType, hardwareWalletName }) => ({
+      type,
+      hardwareWalletName:
+        previousType === type ? hardwareWalletName : undefined,
+    }));
+
+    // We preserve the first step in the array because we want the current first step to *not* be re-rendered. It would cause a flickering of the content of the modal.
+    steps.splice(1, steps.length);
```
lmuntanerDfinity, line 62: is this better than having all the steps from the beginning and jumping from one to another one?

lmuntanerDfinity: And instead of a Select type, it moves from step 0, to step 2 for example.

frontend/svelte/src/lib/components/accounts/AddAccountType.svelte
```diff
@@ -1,20 +1,29 @@
...
+  const context: AddAccountContext = getContext<AddAccountContext>(
+    ADD_ACCOUNT_CONTEXT_KEY
+  );
+
+  const select = async (type: AccountType) => await context.selectType(type);
 </script>
 
 <div class="wizard-wrapper">
-  <div class="card-item" role="button" on:click={selectNewAccount}>
+  <div class="card-item" role="button" on:click={() => select("subAccount")}>
```
lmuntanerDfinity, line 18: here could be `on:click={goToCreateSubAccount}`

frontend/svelte/src/lib/components/accounts/AddAccountType.svelte
```diff
@@ -1,20 +1,29 @@
...
     <h4>{$i18n.accounts.new_linked_title}</h4>
     <span>{$i18n.accounts.new_linked_subtitle}</span>
   </div>
-  <!-- TODO: Select New Hardware Wallet -->
-  <div class="card-item" role="button" on:click={selectNewAccount}>
+  <div
+    class="card-item"
+    role="button"
+    on:click={() => select("hardwareWallet")}
```
lmuntanerDfinity, line 25: And here: `on:click={goToCreateHardwareWallet}`

lmuntanerDfinity (commented):
A few comments on the pattern used to switch between branches of the wizard.

frontend/svelte/src/lib/modals/accounts/AddAccountModal.svelte
peterpeterparker: I think it's better because otherwise we have to prevent `.next()` to being used anywhere in this wizard. Who knows in x months from now someone adds a step, use `.next()` and bim wrong step is displayed.

By making it dynamic, there is two distinct tracks.

frontend/svelte/src/lib/components/accounts/AddAccountType.svelte
peterpeterparker: why not ✅

frontend/svelte/src/lib/modals/accounts/AddAccountModal.svelte
lmuntanerDfinity: We spoke about adding a `set` method, so that we should be able to do `modal.set("SelectAccount")`. It could be used for this use case.

peterpeterparker: Sure, would not solve the `.next()` potential issue though except is there is no `next` 🤪.

lmuntanerDfinity (approved):

PR #759, opened 2022-04-27
L2 468 add titles to my votes icons ux improvement

wackyleo459:
# Motivation
Improvement is required for accessibility purposes.
Example: A person with a disability might not understand the `My Votes` section. For the visually-impaired, the voice-over feature currently reads aloud fragmented information with no clear meaning.

# Changes
- Title attribute has been added to for each neuron vote. This shows up as a tooltip when hovered over.
- Similarly, aria-label has been added as standard practice for enabling voice-over accessibility. Detail audio will say: 'Neuron xxxxxx has voted yes'. 

# Tests
- Tests for `title`
- Tests for `aria-label`
Screen Shot 2022-04-29 at 10 42 14 AM
Screen Shot 2022-04-29 at 10 40 40 AM


frontend/svelte/src/tests/lib/components/proposal-detail/VotesCard.spec.ts
```diff
@@ -120,5 +121,30 @@ describe("VotesCard", () => {
...
     });
+
+    it("should have title attribute per voted neuron for YES or NO", () => {
+      neuronsStore.setNeurons({ neurons: votedNeurons, certified: true });
+      render(VotesCard, {
+        props: {
+          proposalInfo: mockProposalInfo,
+        },
+      });
+      expect(screen.getByTitle(/has voted NO/)).toBeInTheDocument();
```
mstrasinskis, line 132: Better use labels from get(i18n...)

mstrasinskis, line 132: Better use labels from get(i18n...)

frontend/svelte/src/tests/lib/components/proposal-detail/VotesCard.spec.ts
```diff
@@ -120,5 +121,30 @@ describe("VotesCard", () => {
...
+      expect(screen.getByTitle(/has voted YES/)).toBeInTheDocument();
```
mstrasinskis, line 133: Better to use `getByTitle` w/o screen

mstrasinskis, line 133: Better to use `getByTitle` w/o screen

mstrasinskis (commented):
LGTM, only a couple of minor comments

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -75,15 +92,32 @@
...
+              {neuron.id} has voted {neuron.vote === Vote.YES ? "yes" : "no"} with
+              {$i18n.proposal_detail__vote.voting_power} of {neuron.votingPower}
+            </div>
+          {/if}
+          <li data-tid="neuron-data">
+            <p>{neuron.id}</p>
+            <p class="vote-details">
+              <span>{neuron.votingPower}</span>
+              {#if voteIconMapper[neuron.vote]}
+                <svelte:component this={voteIconMapper[neuron.vote]} />
```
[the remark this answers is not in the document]
wackyleo459: Is it sufficient to wrap as span with title of 'YES' / 'NO' or should it be something like 'thumb-up' / 'thumb-down'?

frontend/svelte/src/tests/lib/components/proposal-detail/VotesCard.spec.ts
peterpeterparker: `i18n.mock` for tests 😉

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -31,6 +33,21 @@
     [Vote.UNSPECIFIED]: undefined,
   };
   let neuronsVotedForProposal: CompactNeuronInfo[];
+  function voteMapper(inputNeuronId: NeuronId, vote: Vote) {
```
peterpeterparker, line 36: we have a convention (not linter yet) that we use objects for params as soon as more than one params

we also rarely use `function` but arrow function (no linter here neither)

`const voteMapper = ({inputNeuronId, vote}: {inputNeuronId: NeuronId, vote: Vote}) =>`

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -31,6 +33,21 @@
...
+    if (vote === Vote.NO) {
```
peterpeterparker, line 37: you can probably avoid code duplication by matching the enum to the translation

something like:


```
replacePlaceholders($i18n.proposal_detail__vote[Vote.NO]....
```

or create a map of translation indexed with the vote

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -75,15 +92,32 @@
...
-              <svelte:component this={voteIconMapper[neuron.vote]} />
-            {/if}
-          </p>
-        </li>
+        <div
+          role="img"
+          aria-labelledby="voteStatus"
+          title={voteMapper(neuron.id, neuron.vote)}
+        >
+          {#if neuron.vote === Vote.UNSPECIFIED}
```
peterpeterparker, line 100: Because the `div` is identified with an `id` that has to be unique I would move this `if` within the `<div class="visually-hidden" id="voteStatus">` to avoid the duplucation

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -75,15 +92,32 @@
...
+            <div class="visually-hidden" id="voteStatus">
+              Neuron ID {neuron.id} has not voted
```
peterpeterparker, line 102: text should be extracted to `i18n`

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -75,15 +92,32 @@
...
+            </div>
+          {:else}
+            <div class="visually-hidden" id="voteStatus">
+              {$i18n.neuron_detail.title}
+              {neuron.id} has voted {neuron.vote === Vote.YES ? "yes" : "no"} with
```
peterpeterparker, line 107: `i18n` too

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -158,4 +192,27 @@
       gap: var(--padding);
     }
   }
+
+  .visually-hidden {
+    border: 0;
+    padding: 0;
+    margin: 0;
+    position: absolute !important;
```
peterpeterparker, line 200: please do not use `!important`

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -158,4 +192,27 @@
       gap: var(--padding);
     }
   }
+
+  .visually-hidden {
```
peterpeterparker, line 196: would'nt `opacity:0;` and `visibility: hidden` do the job?

peterpeterparker (commented):
Thanks, a couple of comments

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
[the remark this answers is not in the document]
mstrasinskis: For the title and label the full text could be used. So we don't need to implement what is in the previous comment `... aria-label and title to <li>`

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
wackyleo459: Thank you for your feedback! I've altogether removed the div to make it simpler.

wackyleo459:
I have refactored:
- Title, and aria-label (changed from aria-labelledby) has been added to existing parent li element.
     - makes it simplified: does not add an an extra element to the dom.
- Voting status has been simplified to one label inside i18: `vote_status`
- Used a local mapper to ensure vote corresponds to correct string.

frontend/svelte/src/lib/components/proposal-detail/VotesCard.svelte
```diff
@@ -30,8 +31,22 @@
...
+  const voteMapper = (inputNeuronId: NeuronId, vote: Vote) => {
+    const stringMapper = {
+      [Vote.NO]: "voted NO",
+      [Vote.YES]: "voted YES",
+      [Vote.UNSPECIFIED]: "not voted",
+    };
+
+    return replacePlaceholders($i18n.proposal_detail__vote.vote_status, {
+      $neuronId: inputNeuronId.toString(),
+      $vote: stringMapper[vote],
```
mstrasinskis, line 46: Actually you could ignore `Vote.UNSPECIFIED`. It shouldn't be in the list. But all the texts (e.g. "voted NO") should be in the i18n file. There is already labels that can be used for the `Vote` status - `i18n.core.yes / no`. 
```ts
$vote: vote === Vote.YES ? $i18n.core.yes : $i18n.core.no,
```

frontend/svelte/src/lib/i18n/en.json
```diff
@@ -272,7 +272,8 @@
     "confirm_adopt_headline": "Adopt Proposal",
     "confirm_adopt_text": "You are about to cast $votingPower votes for this proposal, are you sure you want to proceed?",
     "confirm_reject_headline": "Reject Proposal",
-    "confirm_reject_text": "You are about to cast $votingPower votes against this proposal, are you sure you want to proceed?"
+    "confirm_reject_text": "You are about to cast $votingPower votes against this proposal, are you sure you want to proceed?",
+    "vote_status": "$neuronId has $vote"
```
mstrasinskis, line 276: Just tested it with a screen reader, sounds already great. What do you think about this tiny improvement?
```suggestion
    "vote_status": "Neuron $neuronId has voted $vote"
```

mstrasinskis (commented):
Thank you, just a couple of suggestions.

frontend/svelte/src/tests/lib/components/proposal-detail/VotesCard.spec.ts
```diff
@@ -120,5 +121,48 @@ describe("VotesCard", () => {
...
+      const { container } = render(VotesCard, {
+        props: {
+          proposalInfo: mockProposalInfo,
+        },
+      });
+      const element = container.querySelector(`[data-tid="neuron-data"]`);
+
+      expect(element).toBeInTheDocument();
+
+      if (element) {
```
mstrasinskis, line 163: Could you please get rid of the `if`?

mstrasinskis (commented):
Thanks, just a tiny request.

mstrasinskis (approved):

peterpeterparker (approved):
LGTM, thanks

PR #760, opened 2022-04-27
Feat/replace sanitizing

mstrasinskis:
# Motivation

To improve security (against XSS Attacks) custom "sanitiser" was replaced with [DOMPurify](https://github.com/cure53/DOMPurify).

# Changes

- `removeHTMLTags` replaced with `sanitize`
- `ScriptLoader` replaced with dynamic imports (marked)
- added silentConsoleErrors function

# Tests

- updated all marked related tests

# Notes
Most of the tests expects `<Markdown>` to render raw text in `.fallback` block (also displays console.error), so `silentConsoleErrors` was added to all components that display markdown.

# Screenshots

https://user-images.githubusercontent.com/98811342/165469560-55c02259-e025-446e-9ea3-fb9ced95265e.mov


frontend/svelte/src/lib/services/utils.services.ts
```diff
@@ -60,3 +62,16 @@ export const queryAndUpdate = async <R, E>({
       .finally(() => (certifiedDone = true)),
   ]);
 };
+
+/**
+ * Sanitize markdown text and convert it to HTML
+ */
+export const markdownToSanitisedHTML = async (
```
lmuntanerDfinity, line 69: why is this a `service` not a pure util?

frontend/svelte/src/lib/utils/markdown.utils.ts
```diff
@@ -50,3 +50,16 @@ export const renderer = (marked: Marked): Renderer => {
...
   return renderer;
 };
+
+/**
+ * Uses markedjs
+ * @see {@link https://github.com/markedjs/marked}
+ */
+export const markdownToHTML = async (): Promise<(text: string) => string> => {
+  const url = "/assets/assets/libs/marked.min.js";
+  const { marked }: { marked: Marked } = await import(url);
```
lmuntanerDfinity, line 60: is this going to load every time that we use the util? And rely on the browser cache I guess?

frontend/svelte/src/lib/utils/security.utils.ts
```diff
@@ -1,14 +1,9 @@
...
-  text?: string | undefined
-): string | null | undefined => {
-  if (text === undefined) return text;
-
-  decoder = decoder || document.createElement("div");
-  decoder.innerHTML = text;
-  return decoder.textContent;
+export const sanitize = async (): Promise<(text: string) => string> => {
+  const url = "/assets/assets/libs/purify.min.js";
+  const { sanitize: purify } = (await import(url)).default;
```
lmuntanerDfinity, line 7: ditto on importing every time it's used?

frontend/svelte/src/lib/components/ui/Markdown.svelte
```diff
@@ -1,41 +1,26 @@
...
-    on:nnsLoad={onLoad}
-    on:nnsError={onError}
-  />
-{:else if globalMarked !== undefined && text !== undefined}
-  {@html globalMarked?.parse(removeHTMLTags(text) ?? "", {
-    renderer: renderer(globalMarked),
-  })}
 {:else}
-  <!-- fallback text content -->
-  <p data-tid="markdown-text">{text}</p>
```
lmuntanerDfinity, line 40: Why did you remove the `data-tid`? can't we keep it?

lmuntanerDfinity (commented):
A few questions

frontend/svelte/src/lib/components/ui/Markdown.svelte
```diff
@@ -1,41 +1,26 @@
 <script lang="ts">
-  import type { marked } from "marked";
-  import ScriptLoader from "../common/ScriptLoader.svelte";
   import Spinner from "./Spinner.svelte";
-  import { removeHTMLTags } from "../../utils/security.utils";
-  import { renderer } from "../../utils/markdown.utils";
-  type Marked = typeof marked;
+  import { markdownToSanitisedHTML } from "../../services/utils.services";
```
peterpeterparker, line 3: typo "sanitize" not "sanitise"

frontend/svelte/src/lib/components/ui/Markdown.svelte
```diff
@@ -1,41 +1,26 @@
...
+  let error: boolean = false;
+  const transform = async (text: string) => {
+    try {
+      html = await markdownToSanitisedHTML(text);
+    } catch (err) {
+      console.error(err);
+      error = true;
+    }
   };
+  $: if (text !== undefined) transform(text);
```
peterpeterparker, line 17: you don't need to handle the promise here?

frontend/svelte/src/lib/utils/security.utils.ts
mstrasinskis: Do you suggest to cache the function instead?

frontend/svelte/src/lib/utils/markdown.utils.ts
peterpeterparker: no, the browser handles esm modules and load these only when needed - i.e. it will load it the first time and then reuse what was previously downloaded

checkout @mstrasinskis above video

frontend/svelte/src/lib/components/ui/Markdown.svelte
mstrasinskis: The return parameter is not used here and the error handling is done inside the function.

mstrasinskis: Would you prefer using `.then` instead of async in the transform function?

frontend/svelte/src/lib/components/ui/Markdown.svelte
mstrasinskis: Good point! Done.

frontend/svelte/src/lib/components/ui/Markdown.svelte
mstrasinskis: Thanks, fixed

frontend/svelte/src/lib/utils/markdown.utils.ts
lmuntanerDfinity: Oh! Sorry @mstrasinskis the screen of the laptop was not large enough and I didn't go down the main comment to see the video.

It's ok, I guess we can rely on the browser.

frontend/svelte/src/lib/services/utils.services.ts
mstrasinskis: I tend to avoid having dependencies between utils due to possible circular dependency problems in the future. But I'm don't have strong opinion on this.

frontend/svelte/src/lib/components/ui/Markdown.svelte
peterpeterparker: I always subscribe to the promise but know @lmuntanerDfinity sometimes does not (for good reason), so just wanted to know if there was a particular reason here not to, just out of curiosity.

frontend/svelte/src/lib/services/utils.services.ts
mstrasinskis: refactored

PR #761, opened 2022-04-27
feat: code-splitting

peterpeterparker:
# Motivation

1. Split the single `bundle.js` into multiple chunks (`bundle.js`, `nns.js` and `agent.js`) that are loaded in parallel to speed up boot time

2. Add support for chunking - code splitting. This will allow us to drop the libraries we added in `assets` folder to replace these with "normal" dependencies which makes versioning easier

# Note

Regarding 1. the original idea was to split bundle and vendor chunks but such a rule is not entirely compatible with 2. as dependencies of the chunks would be included in vendor. PR solution split what's commonly used but also pack the node_modules dependencies that are used within the related chunks.

# Changes

- set up `rollup.config` for code splitting and output to dir
- update `index.html` and CSP for the new main js file called `main.js` instead of `bundle.js`

# Screenshots

Performance - measure - are not reliable (sometimes it is faster, sometimes not, regardless of the solution) but following screenshots display the parallel loading and the load of the test chunks after the app is booted.

Is:

Capture d’écran 2022-04-27 à 12 55 48

With PR:

Capture d’écran 2022-04-27 à 12 51 09


mstrasinskis (approved):
LGTM 👍

PR #770, opened 2022-04-29
L2-485: updated topic color

wackyleo459:
# Motivation
To have topics appear less like buttons

# Changes
changed border and text color of Topic labels
Screen Shot 2022-04-28 at 4 43 44 PM


mstrasinskis (approved):
LGTM, thanks.

PR #771, opened 2022-05-02
Nns canisters locally

bitdivine:
# Motivation
We wish to be able to test with local NNS canisters:
- This allows us to test in CI
- We can test with modified NNS canisters, e.g. to accelerate maturity.

# Changes
- Add script to download pre-built canisters
- Add script to install canisters locally
- Add script to build patched canisters locally, for e.g. faster maturity
- Add script to find the URL needed to deploy NNS canisters locally

# Future PRs
To get the nns-dapp working well with local governance canisters:
- The local proxy needs to be removed.
- Some hard-wired values in frontend/svelte need to be un-hard-wired.

# Tests
- WITH ADDITIONAL CODE:  I have used locally deployed canisters to get ICP and create neurons.  This was however a hacked up set of scripts that predate this PR; this PR is a cleaned up copy of some of that code.  There were also changes to the svelte front end.

Screenshot from 2022-05-02 04-41-56

- The scripts in this PR are used in CI and the installation succeeds:
Screenshot from 2022-05-02 04-38-30


e2e-tests/scripts/nns-canister-download
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+set -euo pipefail
+cd "$(dirname "$0")/../.."
+
+IC_COMMIT=6dc5e2dcb50569e20891af02b846a7d9a58e3489
+DOWNLOAD_DIR="$PWD/target/ic"
```
lmuntanerDfinity, line 6: Is this supposed to be the same as the previous one: `$(realpath ../../target/ic)`?

e2e-tests/scripts/nns-canister-download
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+set -euo pipefail
+cd "$(dirname "$0")/../.."
+
+IC_COMMIT=6dc5e2dcb50569e20891af02b846a7d9a58e3489
```
lmuntanerDfinity, line 5: Maybe add a comment of why this commit? We'll forget otherwise.

e2e-tests/scripts/nns-canister-build
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+set -euo pipefail
+cd "$(dirname "$0")"
+
+DOWNLOAD_DIR="$(realpath ../../target/ic)"
+IC_COMMIT=eba88796cf8dff32f5788c9167cdd8e292b6072a
```
lmuntanerDfinity, line 6: Maybe add a comment of why this commit?

e2e-tests/scripts/nns-canister.Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+RUN git clone https://github.com/dfinity/ic && \
+    cd ic && \
+    git reset --hard ${IC_COMMIT} && \
+    rm -rf .git && \
+    cd ..
+
+RUN git config --global url."https://github.com/".insteadOf git://github.com/
+
+COPY nns-canister.patch /tmp/
```
lmuntanerDfinity, line 21: Maybe add a comment of why we do this?

This adds the hack to increase maturity in every heartbeat, right?

.github/workflows/build.yml
```diff
@@ -95,6 +95,12 @@ jobs:
 
       # IDENTITY_SERVICE_URL needs to match `e2e-tests/test.js`
       # REDIRECT_TO_LEGACY=svelte builds and deploys only Svelte
+      - name: Deploy NNS canisters
```
lmuntanerDfinity, line 98: Is it worth repeating that this need to go right after `dfx start --clean`?

lmuntanerDfinity (approved):
This looks promising and not at all straightforward! Thanks 🙏 !

My comments are to help us (me) in the future better understand what's going on here.

One general question, why do we need to build the ledger while we can download the other nns canisters? Isn't there a built ledger? Is it to enable adding the hack to increase maturity faster?

e2e-tests/scripts/nns-canister-build
bitdivine: I have added a comment.

e2e-tests/scripts/nns-canister-download
bitdivine: It is a commit from master; unfortunately Roman's branch is not on github so I can't use that directly and patch instead.

I have added a comment to describe how to choose a commit.

e2e-tests/scripts/nns-canister-download
bitdivine: Yes.  The build script replaces some prebuilt, standard artifacts.  A catch is that to build the alternative wasm,  docker is needed so we don't have a strong guarantee that that will work on developer machines.  If it doesn't, you can skip the nns-canisters-build script and just use the standard prebuilt wasm files locally.  If that turns out to be the case, we may also get CI to build the wasm files so that you can download them.

e2e-tests/scripts/nns-canister.Dockerfile
bitdivine: Yes.  Done

bitdivine:
> Is it to enable adding the hack to increase maturity faster?
Yes.  I believe that we will also need other hacks, such as speeding up "wait for quiet", but those may be covered by deploy arguments.

e2e-tests/scripts/nns-canister-download
lmuntanerDfinity: thanks

e2e-tests/scripts/nns-canister-build
lmuntanerDfinity: thanks

bitdivine:
@lmuntanerDfinity - I am pressing ahead with the automerge because it can take a while.  If you would like further changes please say and I'll have a look, regardless of whether this PR has merged yet or not.

e2e-tests/scripts/nns-canister.Dockerfile
lmuntanerDfinity: thanks

.github/workflows/build.yml
lmuntanerDfinity: sounds good

PR #774, opened 2022-05-02
feat: display question "accept reject proposal"

peterpeterparker:
# Motivation

Some users find themselves lost on the proposal detail page after some scroll and might not be sure on what proposal they are about to vote. That's why we add a sentence next to the vote actions that display the proposal id and topic again.

# Changes

- refactor proposal info destruction that is shared in various component
- display sentence “Do you want to adopt or reject Proposal 56789 for Governance?” with related proposal id and topic

# Screenshot

Capture d’écran 2022-05-02 à 10 39 20


mstrasinskis (approved):
LGTM. Wouldn't it be more user friendly to display the proposal title?

peterpeterparker:
@mstrasinskis good point, I'll give a try and provide in another PR if successful

PR #779, opened 2022-05-02
feat: &ndash; for question separator

peterpeterparker:
# Motivation

> That’s good. A minor thing: can you use “&ndash;” for the dash between the id and the title. a dash is not the same character as a “-” (minus). Just a minor thing…

Mischa

# Screenshot

Capture d’écran 2022-05-02 à 16 07 52


mstrasinskis (approved):
LGTM 👍

PR #781, opened 2022-05-02
chore: declare title for eslint checker

peterpeterparker:
# Motivation

Resolve eslint "Hint: Variable 'title' implicitly has an 'any' type, but a better type may be inferred from usage. (ts)"

# Changes

- declare title


mstrasinskis (approved):
LGTM

PR #784, opened 2022-05-03
Fix: replace identityServiceURL with host to interact with accounts canister

ghost:
# Motivation

User sees different subaccounts in Svelte and Flutter.

# Changes

* Use the host instead of the identitityServiceUrl for accounts.

# Tests

No changes.

### DIFF
--- a/frontend/svelte/src/lib/api/proposals.api.ts
+++ b/frontend/svelte/src/lib/api/proposals.api.ts
@@ -7,7 +7,7 @@ import type {
   Vote,
 } from "@dfinity/nns";
 import { GovernanceCanister, Topic } from "@dfinity/nns";
-import { LIST_PAGINATION_LIMIT } from "../constants/constants";
+import { DEFAULT_LIST_PAGINATION_LIMIT } from "../constants/constants";
 import type { ProposalsFiltersStore } from "../stores/proposals.store";
 import { createAgent } from "../utils/agent.utils";
 import { hashCode, logWithTimestamp } from "../utils/dev.utils";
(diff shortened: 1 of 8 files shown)


lmuntanerDfinity:
@bitdivine @peterpeterparker @mstrasinskis please review

frontend/svelte/src/lib/constants/host.constants.ts
```diff
@@ -0,0 +1 @@
+export const host: string = process.env.HOST as string;
```
peterpeterparker, line 1: there are additional duplicates of `host: process.env.HOST`, do you plan to replace these too?

lmuntanerDfinity: Those are in the `api` folder. I wasn't sure about using something from outside the `api` folder there.

I understand the `api` folder as a standalone module. But I can also change it if that is not the case.

peterpeterparker: `proposal.api` and `governance.api` are in `api`

lmuntanerDfinity: that's what I mean. And this file is not in `api` folder. If we import from this file, then the `api` folder is not independent and could not be used as a standaone module.

peterpeterparker: but you already do it, look at the change at line 12 of the `account.api.ts` of your PR

lmuntanerDfinity: oh, yes, that was already importing from constants, oh, well, then let's do it.

lmuntanerDfinity: I move it inside `api`, it's only used there anyway.

peterpeterparker (approved):
LGTM

As discussed `identityServiceURL` was introduced in https://github.com/dfinity/nns-dapp/pull/461/files. It is mandatory for login purpose but for canister better use the proper host.

bitdivine (approved):

PR #785, opened 2022-05-04
feat: connect hardware wallet

peterpeterparker:
# Motivation

Add feature to connect hardware wallet to an account.

# Changes

- integrate `ledger.identity` and `secp256k1`
- add `connectToHardwareWallet` service
- chunk service
- implement component to connect hardware wallet
- translate and throw errors
- add dependencies `@ledgerhq/hw-transport-node-hid-noevents`, `@ledgerhq/hw-transport-webhid` and `@zondax/ledger-icp`

# Screenshots

ezgif com-gif-maker

ezgif com-gif-maker(1)

ezgif com-gif-maker(2)

### DIFF
--- a/frontend/svelte/src/lib/components/accounts/HardwareWalletConnect.svelte
+++ b/frontend/svelte/src/lib/components/accounts/HardwareWalletConnect.svelte
@@ -1,62 +1,23 @@
 <script lang="ts">
   import { i18n } from "../../stores/i18n";
   import { LedgerConnectionState } from "../../constants/ledger.constants";
-  import Spinner from "../ui/Spinner.svelte";
-  import { LedgerIdentity } from "../../identities/ledger.identity";
-  import HardwareWalletInfo from "./HardwareWalletInfo.svelte";
+  import HardwareWalletConnectAction from "./HardwareWalletConnectAction.svelte";
 
   let connectionState: LedgerConnectionState =
     LedgerConnectionState.NOT_CONNECTED;
-  let ledgerIdentity: LedgerIdentity | undefined = undefined;
-
-  const connect = async () => {
-    const { connectToHardwareWallet } = await import(
-      "../../services/ledger.services"
-    );
-
-    await connectToHardwareWallet(
-      ({ ledgerIdentity: identity, connectionState: state }) => {
-        connectionState = state;
-        ledgerIdentity = identity;
-      }
-    );
-  };
 
   const onSubmit = () => {
     // TODO(L2-433): Attach wallet
     alert("TODO(L2-433): Attach wallet");
   };
 
-  let disabled, connecting, connected: boolean;
+  let disabled: boolean;
   $: disabled = connectionState !== LedgerConnectionState.CONNECTED;
-  $: connecting = connectionState === LedgerConnectionState.CONNECTING;
-  $: connected =
-    connectionState === LedgerConnectionState.CONNECTED &&
-    ledgerIdentity !== undefined;
 </script>
 
 <form on:submit|preventDefault={onSubmit} class="wizard-wrapper">
   <div>
-    {#if connected}
-      <h4>Connected to Hardware Wallet</h4>
-      <HardwareWalletInfo {ledgerIdentity} />
-    {:else}
-      <p>{$i18n.accounts.connect_hardware_wallet_text}</p>
-      <div class="connect">
-        {#if connecting}
-          <!-- // TODO(L2-433): spinner processing accessibility -->
-          <Spinner />
-        {:else}
-          <button
-            class="primary"
-            type="button"
-            on:click|stopPropagation={connect}
-          >
-            {$i18n.accounts.connect_hardware_wallet}
-          </button>
-        {/if}
-      </div>
-    {/if}
+    <HardwareWalletConnectAction bind:connectionState />
   </div>
 
   <button class="primary full-width submit" type="submit" {disabled}>
@@ -69,15 +30,6 @@
 
   form {
     @include modal.wizard-single-input-form;
-  }
-
-  h4,
-  p {
-    margin-bottom: var(--padding-2x);
-  }
-
-  .connect {
-    min-height: var(--button-min-height);
   }
 
   .submit {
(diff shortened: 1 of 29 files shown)


frontend/svelte/src/lib/components/accounts/HardwareWalletConnectAction.svelte
```diff
@@ -0,0 +1,61 @@
...
+
+  let connecting: boolean;
+  let connected: boolean;
+  $: connecting = connectionState === LedgerConnectionState.CONNECTING;
+  $: connected =
+    connectionState === LedgerConnectionState.CONNECTED &&
+    ledgerIdentity !== undefined;
+</script>
+
+{#if connected && ledgerIdentity !== undefined}
```
mstrasinskis, line 30: Seems `ledgerIdentity !== undefined` is already a part of `connected` state.

frontend/svelte/src/lib/components/accounts/HardwareWalletInfo.svelte
```diff
@@ -0,0 +1,37 @@
...
+  import { principalToAccountIdentifier } from "@dfinity/nns";
+  import { i18n } from "../../stores/i18n";
+
+  export let ledgerIdentity: LedgerIdentity;
+
+  let principalText: string;
+  let accountIdentifier: string;
+  $: principalText = ledgerIdentity.getPrincipal().toString();
+  $: accountIdentifier = principalToAccountIdentifier(
+    ledgerIdentity.getPrincipal()
```
mstrasinskis, line 12: Does it work w/o ledgerIdentity.getPrincipal()`.toText()`?

frontend/svelte/src/lib/keys/secp256k1.ts
```diff
@@ -0,0 +1,98 @@
...
+    0x06, 0x05, 0x2b, 0x81, 0x04, 0x00, 0x0a, // OID secp256k1
+    0x03, 0x42, // BIT STRING
+    0x00, // no padding
+  ]);
+
+  private static derEncode(publicKey: ArrayBuffer): DerEncodedPublicKey {
+    if (publicKey.byteLength !== Secp256k1PublicKey.RAW_KEY_LENGTH) {
+      const bl = publicKey.byteLength;
+      throw new TypeError(
+        `secp256k1 public key must be ${Secp256k1PublicKey.RAW_KEY_LENGTH} bytes long (is ${bl})`
```
mstrasinskis, line 34: Do we want those messages in i18n? It's used in a couple of places.

frontend/svelte/src/lib/identities/ledger.identity.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    try {
+      // Verify that the public key of the device matches the public key of this identity.
+      const devicePublicKey: Secp256k1PublicKey =
+        await LedgerIdentity.fetchPublicKeyFromDevice({
+          app,
+          derivePath: this.derivePath,
+        });
+
+      if (JSON.stringify(devicePublicKey) !== JSON.stringify(this.publicKey)) {
```
mstrasinskis, line 162: Because of using stringify probably it makes sense to add a catch with console.errror. For example if there will be some bigint in the future.

mstrasinskis (commented):
Looks good. A few questions.

frontend/svelte/src/lib/components/accounts/HardwareWalletConnect.svelte
```diff
@@ -1,11 +1,50 @@
...
+<style lang="scss">
+  @use "../../themes/mixins/modal.scss";
+
+  form {
+    @include modal.wizard-single-input-form;
+  }
+
+  .submit {
+    opacity: 0;
+    visibility: hidden;
```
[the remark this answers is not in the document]
peterpeterparker: Robot mode, I always use both at the same time. Sure why not just opacity, done.

frontend/svelte/src/lib/components/accounts/HardwareWalletConnectAction.svelte
peterpeterparker: Yes indeed, it was my first implementation but then the type checker complained about it 😉. I guess it does not notice it therefore have to explicitely add the check.

frontend/svelte/src/lib/components/accounts/HardwareWalletInfo.svelte
peterpeterparker: No, it returns the object (`Assigned expression type Principal is not assignable to type string `).
In Java it would though 😉

mstrasinskis: We are going to remove `toJson()` from nns-js. Not sure is it related.

frontend/svelte/src/lib/keys/secp256k1.ts
peterpeterparker: Good point. I am unsure yet if we keep this implementation within nns-dapp or if we use the 100% identical code from agent-js. So I have added a `// TODO(L2-433): if we keep this implementation within nns-dapp, hardcoded text should be extracted to i18n`. If we ultimately remove the implementation, that's something spared.

frontend/svelte/src/lib/identities/ledger.identity.ts
peterpeterparker: I think it is already catched by our pattern of using a try catch toast error around the services call. Is that correct? Does that work for you?

```
try {
   ...
   const ledgerIdentity: LedgerIdentity = await LedgerIdentity.create();
   ...
   } catch (err: unknown) {
    const ledgerErrorKey: boolean = err instanceof LedgerErrorKey;

    toastsStore.error({
      labelKey: ledgerErrorKey
        ? (err as LedgerErrorKey).message
        : "error__ledger.unexpected",
      ...(!ledgerErrorKey && { err }),
    });
    ```

frontend/svelte/src/lib/components/accounts/HardwareWalletInfo.svelte
peterpeterparker: don't think so
Capture d’écran 2022-05-05 à 13 01 10

mstrasinskis (approved):
LGTM. Thank you.

PR #797, opened 2022-05-06
Correct agent URL

bitdivine:
# Motivation
In production, agent.js should call https://nns.ic0.app not https://ic0.app

In fact, it calls both but it should only be calling https://nns.ic0.app.  The CSP is lax in allowing https://ic0.app.  Tightening the CSP however has to be done with extreme caution so is deferred until we can observe no calls to domains that would be blocked.

There is a second open question: Why we need to provide the host to agent.js at all.  I suspect that once we have completely removed the proxy, we can also omit the host, both from the agent and from the CSP, using `self` in the latter.

# Changes
- Correct the HOST in svelte.

# Tests
Running `DEPLOY_ENV=mainnet ./build.sh` shows this configuration in svelte:
```
{
  "svelteEnvironmentVariables": {
    "ENVIRONMENT": "mainnet",
    "DEPLOY_ENV": "mainnet",
    "FETCH_ROOT_KEY": false,
    "REDIRECT_TO_LEGACY": "prod",
    "MAINNET": "https://ic0.app",  <------------ NOTE: This is used in the CSP.  This is unchanged.
    "HOST": "https://nns.ic0.app", <------------ NOTE: This is the only value that has changed in the config.
    "OWN_CANISTER_ID": "qoctq-giaaa-aaaaa-aaaea-cai",
    "OWN_CANISTER_URL": "https://qoctq-giaaa-aaaaa-aaaea-cai.ic0.app/",
    "IDENTITY_SERVICE_URL": "https://identity.ic0.app/",
    "GOVERNANCE_CANISTER_ID": "rrkah-fqaaa-aaaaa-aaaaq-cai",
    "LEDGER_CANISTER_ID": "ryjl3-tyaaa-aaaaa-aaaba-cai",
    "GOVERNANCE_CANISTER_URL": "https://rrkah-fqaaa-aaaaa-aaaaq-cai.ic0.app/",
    "LEDGER_CANISTER_URL": "https://ryjl3-tyaaa-aaaaa-aaaba-cai.ic0.app/",
    "ROLLUP_WATCH": false
  }
}
```

### DIFF
--- a/frontend/svelte/env.config.mjs
+++ b/frontend/svelte/env.config.mjs
@@ -32,7 +32,7 @@ const IDENTITY_SERVICE_URL =
 
 // The host that nns-js connects to
 const HOST =
-  process.env.HOST || (development ? `https://${domainTestnet}/` : MAINNET);
+  process.env.HOST || (development ? `https://${domainTestnet}/` : "https://nns.ic0.app");
 
 // Canister Ids for testnet and mainnet
 const GOVERNANCE_CANISTER_ID = "rrkah-fqaaa-aaaaa-aaaaq-cai";
--- a/frontend/svelte/env.config.mjs
+++ b/frontend/svelte/env.config.mjs
@@ -32,7 +32,8 @@ const IDENTITY_SERVICE_URL =
 
 // The host that nns-js connects to
 const HOST =
-  process.env.HOST || (development ? `https://${domainTestnet}/` : "https://nns.ic0.app");
+  process.env.HOST ||
+  (development ? `https://${domainTestnet}/` : "https://nns.ic0.app");
 
 // Canister Ids for testnet and mainnet
 const GOVERNANCE_CANISTER_ID = "rrkah-fqaaa-aaaaa-aaaaq-cai";


peterpeterparker:
> In production, agent.js should call https://nns.ic0.app/ not https://ic0.app/

What's the source of this statement?

lmuntanerDfinity (commented):
According to a test in agent-js, this changes nothing:

https://github.com/dfinity/agent-js/blob/65e17f988cf4f910fa20fc105174ad238594920c/packages/agent/src/agent/http/http.test.ts#L157

Does it make sense to talk to them? As @peterpeterparker asks, where is it specified that we need to use that url?

bitdivine:
Good idea.  It is my understanding that we should be using a tight CSP that allows only the origin host and named canisters, but an argument against that is query string redirects to other canisters.  I have started a thread on it and invited @robin-kunzler  from the security team, and @frederikrothenberger  as he has also worked on the CSP:  https://app.slack.com/client/T43F9UHS5/C031X8JBUNP/thread/C031X8JBUNP-1651823465.784359

frederikrothenberger:
I do not agree with this change for the following reasons:
* Connecting to https://nns.ic0.app will lead to redirects in the future. See https://dfinity.atlassian.net/browse/BOUN-65.
* Allowing `https://nns.ic0.app` and `https://ic0.app` does _not_ allow the canister to make canister calls it otherwise couldn't. In fact, the target canister id is also contained in the path and the cbor payload. AFAIK, the hostname is irrelevant for routing canister calls (e.g. `/api/v2/canister/rdmx6-jaaaa-aaaaa-aaadq-cai/query`).

TLDR: Do not do this.
*Note:* `https://ic0.app` should be listed in the CSP as `connect-src` only.

bitdivine:
Ok, thanks for the feedback.  One final concern:  Previously, the host was undefined for mainnet only but set for testnets.  I specified the host for mainnet to make tests and mainnet more similar.  If the host is not specified for ic-agent, it defaults, if I understand correctly, to the domain of the page, so https://nns.ic0.app for mainnet.  As such, the current main branch has a different host from previous mainnet deployments.  Does this matter?  Does it affect the login identity or which accounts a user has?

If you know for certain, great, if not, I'll test with different HOST values on the testnet.

frederikrothenberger:
I know for certain (from experience with II) that if you specify `https://nns.ic0.app` as the host it will strip away the subdomain and use `https://ic0.app` anyways. See https://github.com/dfinity/agent-js/blob/60773413d294c1c9de968ed80b52894deb37f385/packages/agent/src/agent/http/index.ts#L186

As such, it will always behave differently on a testnet than on mainnet unless you are using a domain that ends in `ic0.app` for testing.

PR #799, opened 2022-05-06
L2-446: ToastsStore manages multiple notifications

lmuntanerDfinity:
# Motivation

User sees multiple notifications.

# Changes

* Render all messages in toastsStore.
* Move wrapper styles to a wrapper of all the Toast Component.
* Add timestamp property to toast to identify them later.

# Tests

* Fix and add new test cases in Toasts Component


lmuntanerDfinity:
Video of the UX:

https://user-images.githubusercontent.com/98735375/167082142-c1102ee9-de0b-463b-860e-155df7210a14.mov

lmuntanerDfinity:

peterpeterparker:
Did you review the ux with Mischa?

Looking at the video I spontaneously "fear" that the screen can - particularly in edge case - become fully covered by toast messages what I personally would not like and why I rather like to have the msg stacked as in the current solution but of course if it is what was defined, I am fine with it.

lmuntanerDfinity:
I did that because I had some bad experiences when something happened as an error but I didn't see it because I still had a notification. I will discuss with Mischa, good point.

frontend/svelte/src/lib/stores/toasts.store.ts
```diff
@@ -9,16 +9,24 @@ import { errorToString } from "../utils/error.utils";
...
+  const { subscribe, update, set } = writable<ToastMsg[]>([]);
 
   return {
     subscribe,
 
     show(msg: ToastMsg) {
-      update((messages: ToastMsg[]) => [...messages, msg]);
+      update((messages: ToastMsg[]) => {
+        const now = Date.now();
+        // To cover the edge case where two messages are sent at exactly the same time
```
mstrasinskis, line 23: It looks like the `timestamp` is always used as id. So probably it makes sense to initially generate it with a random?
Suggestion:
```
Math.floor(Math.random() * 1e6) + performance.now()
```

mstrasinskis (commented):
The code looks good just a tiny suggestion.

frontend/svelte/src/lib/stores/toasts.store.ts
lmuntanerDfinity: Sure, I like it better, thanks!

mstrasinskis:
> I did that because I had some bad experiences when something happened as an error but I didn't see it because I still had a notification. I will discuss with Mischa, good point.

We could always use maximum space of 2 messages and in case of more overlap them within reserved space (of 2 messages). Should gives right impression with the correct z-index.

frontend/svelte/src/lib/stores/toasts.store.ts
lmuntanerDfinity: done

lmuntanerDfinity:
@peterpeterparker @mstrasinskis we reviewed it with Mischa. He accepted this solution.

He also accepted the risk of the edge case of having multiple notifications that can cover part of the screen.

mstrasinskis (approved):
LGTM

frontend/svelte/src/lib/components/ui/Toasts.svelte
```diff
@@ -3,7 +3,33 @@
   import Toast from "./Toast.svelte";
 </script>
 
-<!-- Display one toast message at a time -->
-{#if $toastsStore.length > 0}
-  <Toast msg={$toastsStore[0]} />
-{/if}
+<!-- We need the wrapper to avoid having an error in the tests -->
```
peterpeterparker, line 6: can't you use the `div` then just for the test - use a test component that contains a div that contains the toast?

frontend/svelte/src/lib/stores/toasts.store.ts
```diff
@@ -9,16 +9,19 @@ import { errorToString } from "../utils/error.utils";
...
-  const { subscribe, update } = writable<ToastMsg[]>([]);
+  const { subscribe, update, set } = writable<ToastMsg[]>([]);
 
   return {
     subscribe,
 
     show(msg: ToastMsg) {
-      update((messages: ToastMsg[]) => [...messages, msg]);
+      update((messages: ToastMsg[]) => {
+        const id = Math.floor(Math.random() * 1e6) + performance.now();
```
peterpeterparker, line 22: instead of a random number wouldn't a built-in unique `Symbol()` do the job?

frontend/svelte/src/lib/stores/toasts.store.ts
```diff
@@ -37,8 +40,19 @@ const initToastsStore = () => {
       }
     },
 
-    hide() {
-      update((messages: ToastMsg[]) => messages.slice(1));
+    hide(idToHide?: number) {
+      // If not specified, we remove the first one
+      if (idToHide === undefined) {
```
peterpeterparker, line 45: these is a use case were the it is undefined?

peterpeterparker (commented):
Thanks a lot for clearing this with Mischa! Added few comment @lmuntanerDfinity

frontend/svelte/src/lib/stores/toasts.store.ts
lmuntanerDfinity: It's more of an edge case for now, the Type needs to support it as optional, because we add it afterwards.

I could use a `Partial<ToastMsg>` on `show` for example, and then convert it to `ToastMsg` with `id` not defined. But I thought this solutions was more straighforward.

frontend/svelte/src/lib/stores/toasts.store.ts
lmuntanerDfinity: I will try.

frontend/svelte/src/lib/components/ui/Toasts.svelte
lmuntanerDfinity: I can try.

PR #803, opened 2022-05-06
Remove configuration values from svelte

bitdivine:
# Motivation
This continues the effort of using only one central configuration.  This particular thrust is aimed at being able to deploy to other testnets, e.g. to temporary testnets to test fast maturity without disrupting the nnsdapp testnet, and different configurations to the same testnet.

And it works:
```
dfx deploy --network small11 nns-dapp
```
produced: https://qaa6y-5yaaa-aaaaa-aaafa-cai.small11.dfinity.network/

# Changes
The main change:
- In svelte, get configuration variables from the central configuration rather than recomputing the values.
  - The values are read from the JSON config file, but values MAY be overridden by environment variables.
  - Question:  Do we need environment variables for svelte?  Are they actually used?  Would it be convenient to use just the JSON config file?
- Set `DEPLOY_ENV` from `DFX_NETWORK`.  The latter is set as an environment variable by `dfx`, so now we can deploy with:
  `dfx deploy --network testnet nns-dapp`
  rather than using a wrapper script that does:
  `DEPLOY_ENV=testnet dfx deploy --network testnet nns-dapp`
- Construct default canister URLs from `HOST` and the canister ID.  Canister URLs can be given in dfx.json but if they take the standard form, they just duplicate information.

# Future tasks
- Rename `DEPLOY_ENV` to `DFX_NETWORK` and delete the code that juggles between the two variables.
- Delete the `deploy.sh` wrapper script.
- Construct default canister URLs from `HOST` and the canister ID.  Canister URLs can be given in dfx.json but if they take the standard form, they just duplicate information.
- `config.sh` contains some code duplication.  Factor it out.

# Tests
This PR has been deployed to the nnsdapp testnet and to the small11 testnet.
- https://qaa6y-5yaaa-aaaaa-aaafa-cai.small11.dfinity.network/v2/#/accounts
- https://va76m-bqaaa-aaaaa-aaayq-cai.nnsdapp.dfinity.network/v2/#/accounts


config.sh
```diff
@@ -58,10 +58,16 @@ local_deployment_data="$(
 : "- next, look at the environment,"
 : "- last is the defaults section in dfx.json"
 : ""
-: "After assembling the configuration, replace OWN_CANISTER_ID."
+: "After assembling the configuration:"
+: "- replace OWN_CANISTER_ID"
+: "- construct ledger and governance canister URLs"
 jq -s '
```
bitdivine, line 64: Note:  As we now as a matter of policy always have HOST pointing at the network address, without the canister ID, we can construct the default canister URL if it is not specified explicitly.

TODO: The common code can be factored out.

config.sh
```diff
@@ -73,16 +79,30 @@ get_var() {
...
+LEDGER_CANISTER_URL="$(get_var LEDGER_CANISTER_URL)"
+export LEDGER_CANISTER_URL
+
 OWN_CANISTER_ID="$(get_var OWN_CANISTER_ID)"
 export OWN_CANISTER_ID
+OWN_CANISTER_URL="$(get_var OWN_CANISTER_URL)"
+export OWN_CANISTER_URL
 
-# Mainnet has an undefined HOST.  The HOST is an identifier for the calling page.
-HOST="$(get_var HOST 2>/dev/null || echo "")"
```
bitdivine, line 80: HOST is now always expected to be defined, to reduce the number of different code paths, so we don't need this exception any more.

dfx.json
```diff
@@ -22,7 +22,7 @@
         "REDIRECT_TO_LEGACY": "prod",
         "HOST": "https://ic0.app",
         "OWN_CANISTER_ID": "qoctq-giaaa-aaaaa-aaaea-cai",
-        "OWN_URL": "https://nns.ic0.app",
+        "OWN_CANISTER_URL": "https://nns.ic0.app",
```
bitdivine, line 25: The name `OWN_CANISTER_URL` already existed in the svelte codebase, `OWN_URL` was a new addition for the same variable.

dfx.json
```diff
@@ -36,21 +36,33 @@
         "FETCH_ROOT_KEY": true,
         "REDIRECT_TO_LEGACY": "svelte",
         "HOST": "https://nnsdapp.dfinity.network",
-        "OWN_URL": "https://OWN_CANISTER_ID.nnsdapp.dfinity.network",
```
bitdivine, line 39: Thi shas the standard form, so it doesn't need to be listed.  Neither does GOVERNANCE_CANISTER_URL or LEDGER_CANISTER_URL in any of these networks.

Arguably we shoudl do the same with IDENTITY_SERVICE_URL and just provide IDENTITY_CANISTER_ID, but we don't always have the latter and I don't like variables that are defined sometimes!

dfx.json
```diff
@@ -36,21 +36,33 @@
...
+        "FETCH_ROOT_KEY": true,
+        "REDIRECT_TO_LEGACY": "svelte",
+        "HOST": "https://small11.dfinity.network",
+        "IDENTITY_SERVICE_URL": "https://qjdve-lqaaa-aaaaa-aaaeq-cai.small11.dfinity.network"
+      },
+      "providers": [
+        "http://[2a00:fb01:400:42:5000:eeff:feae:ab37]:8080"
+      ],
+      "type": "persistent"
+    },
```
bitdivine, line 58: This works.  Try it. :-)

frontend/svelte/env.config.mjs
```diff
@@ -7,6 +7,19 @@
...
+/**
+ * Returns the given environment variable, if defined and non-empty, else throws an error.
+ */
+function getRequiredEnvVar(key) {
+  const value = process.env[key];
+  if (undefined === value || value === "") {
+    throw new Error(`Environment variable '${key}' is undefined.`);
+  }
+  return value;
+}
```
bitdivine, line 19: Almost all the configuration is now read directly from the environment.  If a required value is missing, we panic rather than trying to guess.

Note:  I am getting the configuration from the environment due to tradition.  The data is also available in a JSON file, it might be easier to use that.  I don't mind either way.  PRs are welcome from anyone who finds that getting config from JSON is easier.

frontend/svelte/env.config.mjs
```diff
@@ -7,6 +7,19 @@
...
+
+const IDENTITY_SERVICE_URL = getRequiredEnvVar("IDENTITY_SERVICE_URL");
+
 const ROLLUP_WATCH = process.env.ROLLUP_WATCH === "true";
 
 const ENVIRONMENT = ROLLUP_WATCH
```
bitdivine, line 25: I am not sure that we need ENVIRONMENT, but keeping changes to the config passed to svelte minimal makes it easier to check whether this PR changes the behaviour of mainnet.  Removing redundant variables can follow in future PRs.

frontend/svelte/env.config.mjs
lmuntanerDfinity: I don't have a strong opinion, but the possibility to override with env vars seems a good feature to me.

Maybe for easily test locally or whatever it's easier not not recreate the whole config json every time and use env vars.

Not sure, but I'd say we can leave it. If we stop using env vars, then I would go to remove the complexity.

lmuntanerDfinity (commented):
Why do the screenshots show the modal now and not before?

They also show the Account amount, and before they didn't because the Ledger Canister was not deployed properly in the local replica. Does this PR fix that also?

bitdivine:
Thank you for th ereview @lmuntanerDfinity .  Even here the screenshot is only partially loaded; the main account has not appeared yet.  When it does, there is the additional complication that the main account includes an account ID, which will be different every time CI runs.

I am not going to cover fixing screenshots, including identifying and blanking out/filling with zeros known unstable elements in this PR, so I will stop saving screenshots until this is addressed.  Is that OK?

lmuntanerDfinity (approved):
Ok that looks good to me.

As you say, the account and principals will be different, therefore we can't ensure the screenshots will be the same.

PR #804, opened 2022-05-06
L2-339: Add Hardware Wallet Accounts to Store

lmuntanerDfinity:
# Motivation

User can see the hardware wallet linked to account.

# Changes

* Add hardwareWallets to Accounts Store.
* Map Hardware Wallet Account details to Account and fetch balance.
* Add "Hardware Wallet" badge.
* Add property "type" to Account to distinguish better different accounts.

# Tests

* New tests for hardware wallets.
* New tests for route Accounts.
* New tests for loadAccounts api function.


lmuntanerDfinity:
UI of Hardware Wallet Account:

Screenshot 2022-05-06 at 17 50 53

lmuntanerDfinity:

frontend/svelte/src/lib/api/accounts.api.ts
```diff
@@ -34,37 +50,57 @@ export const loadAccounts = async ({
     canisterId: LEDGER_CANISTER_ID,
   });
 
-  const mapAccount = async (
-    account: AccountDetails | SubAccountDetails
-  ): Promise<Account> => {
-    const balance: ICP = await ledger.accountBalance({
-      accountIdentifier: AccountIdentifier.fromHex(account.account_identifier),
+  const mapMainAccount = async (account: AccountDetails): Promise<Account> => ({
```
peterpeterparker, line 53: feels like `mapMainAccount`, `mapHardwareAccount` and `mapSubAccount` are really similar. can you refactor this and spare duplicate codes?

frontend/svelte/src/lib/api/accounts.api.ts
```diff
@@ -34,37 +50,57 @@ export const loadAccounts = async ({
...
+    ...mainAccount.sub_accounts.map(mapSubAccount),
+    ...mainAccount.hardware_wallet_accounts.map(mapHardwareAccount),
   ]);
 
   logWithTimestamp(`Loading Accounts certified:${certified} complete.`);
 
   return {
-    main,
-    subAccounts,
+    main: accounts.find(({ type }) => type === "main"),
```
peterpeterparker, line 101: for performance reason, we can maybe spare the iterations. what do you think of following?

```
const [main, subAccounts, hardwareWallets] = await Promise.all([
    mapMainAccount(mainAccount),
    await Promise.all(mainAccount.sub_accounts.map(mapSubAccount)),
    await Promise.all(mainAccount.hardware_wallet_accounts.map(mapHardwareAccount)),
  ]);
  
  return {
    main,
    subAccounts,
    hardwareWallets,
  };
  ```

File: frontend/svelte/src/lib/components/accounts/AccountBadge.svelte [hunk 3]
```diff
@@ -7,6 +7,8 @@
 
 <!-- No badge for main account -->
 
-{#if account.subAccount !== undefined}
+{#if account.type === "subAccount"}
```
peterpeterparker, line 10: why not using a type that match the i18n label so that we can spare a if - so that we can spare ifs anywhere if we want to display the type

```
#if account.type !== undefined}
<small>{$i18n.accounts[account.type]}</small>
{/if}
```

peterpeterparker (commented):
Thanks for the PR, few comments

frontend/svelte/src/lib/api/accounts.api.ts
lmuntanerDfinity: They share only two properties, account_identifier and balance (which uses a helper to share code), the rest are different. I would need ternary operators for all of them. I think it's more straighforward to read and maintain to repeat a little bit.

Take a look at ths screenshot with differences and similarities:
Screenshot 2022-05-09 at 07 21 43

lmuntanerDfinity: I can move the `getAccountBalance` inside the function, to use `ledger` and `certified` from the same scope as well.

This would reduce some lines.

frontend/svelte/src/lib/api/accounts.api.ts
lmuntanerDfinity: sure, that should work as well

frontend/svelte/src/lib/components/accounts/AccountBadge.svelte
lmuntanerDfinity: we would be coupling on with the other, but yes, sounds ok.

lmuntanerDfinity:

frontend/svelte/src/lib/api/accounts.api.ts
peterpeterparker: > the rest are different....Take a look at ths screenshot with differences and similarities:

You can also use generic for the types (`AccountDetails`, `SubaccountDetails`...), so only `principal` and `name` are not present in all types

You can either have a function that build everything with generic and some logic or for example have a function that create what's common and then spread the rest of the values.

> I think it's more straighforward to read and maintain to repeat a little bit.

I actually had a hard time reviewing and understanding the code. When it comes to me it wasn't easy to read and therefore would not be to maintain.

lmuntanerDfinity: I'll try to make it more readable for you

frontend/svelte/src/lib/api/accounts.api.ts
```diff
@@ -34,37 +42,33 @@ export const loadAccounts = async ({
...
-  const mapAccount = async (
-    account: AccountDetails | SubAccountDetails
-  ): Promise<Account> => {
-    const balance: ICP = await ledger.accountBalance({
-      accountIdentifier: AccountIdentifier.fromHex(account.account_identifier),
-      certified,
-    });
-
-    return {
+  const mapAccount =
```
peterpeterparker, line 45: Looks good, thanks 👍.

Since the account variables are defined as optional (`name?: string` not `name: string | undefined`), what about following:

```
const mapAccount =
    (type: AccountType) =>
    async (
      account: AccountDetails | HardwareWalletAccountDetails | SubAccountDetails
    ): Promise<Account> => ({
      identifier: account.account_identifier,
      balance: await getAccountBalance(account.account_identifier),
      type,
      ...("sub_account" in account && {subAccount: account.sub_account}),
      ...("name" in account && {name: account.name}),
      ...("principal" in account && {principal: account.principal}),
    });
```

What do you think?

lmuntanerDfinity: sure

lmuntanerDfinity: done

lmuntanerDfinity:

peterpeterparker (approved):
LGTM. Thanks for the improvements

PR #807, opened 2022-05-09
Fix/recover subnet payload state hash

mstrasinskis:
# Motivation

It's expected that `state_hash` is rendered as a hex string (dashboard). In the `did` file it's a `vec nat8;` so there nothing points that it should be rendered this way.
Current approach is to transform all arrays that looks like a hash (`hexStringBytes`) to hash-string (`bytesToHexString`).

## [Dashboard](https://dashboard.internetcomputer.org/proposal/58333)
image

# Changes

- Json component extended with `hash` type
- attribute `title` added to the hash field with original value

# Tests

- Json
- utils

# Screens

## Before
image

## After
image


frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -102,3 +102,17 @@ export const createChunks = <T>(
   }
   return chunks;
 };
+
+// e.g. state_hash
+export const hexStringBytes = (bytes: number[]): boolean =>
```
lmuntanerDfinity, line 107: Maybe `isHexStringBytes` to make it clear it returns a boolean?

lmuntanerDfinity (approved):
LGTM,. just a minor comment

frontend/svelte/src/lib/utils/utils.ts
mstrasinskis: Updated. Thanks.

PR #813, opened 2022-05-09
spawn neuron will not give slider option for hardware wallet

anishh2003:
# Motivation

Spawning neuron for hardware wallet with slider option has not been implemented yet

# Changes

For hardware wallet , we will show Spawn neuron based on full maturity


bitdivine (approved):
Thank you for the demo.

PR #815, opened 2022-05-09
Fix: Local dev setup

lmuntanerDfinity:
# Motivation

Fix local development setup.

# Changes

* Throw error if `config.json` does not exist.
* Add npm script to build `config.json` for local development.

# Tests

No new tests


lmuntanerDfinity:
@bitdivine @peterpeterparker please review

When running the config script from nodejs with a child process, it doesn't find "jq":

```
Error: Command failed: ../../config.sh
../../config.sh: line 78: jq: command not found
```

Therefore, I think that a quick solution now is to throw an error if the file is not present and then create the file in another step.

Another options is to create `config.json` every time we run `npm run dev`. But I thought it would be too much tasks there.

config.sh
```diff
@@ -71,7 +71,7 @@ jq -s '
   .GOVERNANCE_CANISTER_URL=( if (.GOVERNANCE_CANISTER_URL == null) then (.HOST | sub("^(?<p>https?://)";"\(.p)\($config.GOVERNANCE_CANISTER_ID).")) else .GOVERNANCE_CANISTER_URL end ) |
   .LEDGER_CANISTER_URL=( if (.LEDGER_CANISTER_URL == null) then (.HOST | sub("^(?<p>https?://)";"\(.p)\($config.LEDGER_CANISTER_ID).")) else .LEDGER_CANISTER_URL end ) |
   .OWN_CANISTER_URL=( if (.OWN_CANISTER_URL == null) then (.HOST | sub("^(?<p>https?://)";"\(.p)\($config.OWN_CANISTER_ID).")) else .OWN_CANISTER_URL end ) |
-  .OWN_CANISTER_URL=(.OWN_CANISTER_URL | sub("OWN_CANISTER_ID"; $config.OWN_CANISTER_ID))
```
lmuntanerDfinity, line 74: @bitdivine not sure this change applies, but I found it strange. Is it ok the change?

frontend/svelte/package.json
```diff
@@ -7,8 +7,9 @@
     "i18n": "node scripts/i18n.types.js",
     "import-assets-for-dev": "cp -R ../dart/assets public/assets && cp ../dart/web/manifest.json public/ && cp ../dart/web/favicon.ico public/",
     "import-assets": "rm -fr public/assets/assets && mkdir -p public/assets && npm run import-assets-for-dev",
-    "build": "rm -fr public/assets/assets && npm run i18n && rollup -c && npm run build:index",
-    "dev": "npm run import-assets && npm run i18n && BASE_HREF=/ ROLLUP_WATCH=true npm run build:index && rollup -c -w",
+    "build": "rm -fr public/assets/assets && npm run i18n && npm run build:config && rollup -c && npm run build:index",
```
lmuntanerDfinity, line 10: We also need config script here. In the CI the config script might not exist.

bitdivine (approved):

peterpeterparker (approved):
Thanks

PR #820, opened 2022-05-09
L2 467 fix modal height

wackyleo459:
# Modal height
- In Set Dissolve Delay: fixed modal height property to disable awkward scrolling, where height was neither small enough or big enough.

- In filter (topics) on Voting: ensured modal is centered and not covered by bottom tab-bar on iOS (safari) devices.

## Changes
Screen Shot 2022-05-09 at 1 45 49 PM

Screen Shot 2022-05-09 at 1 44 00 PM


frontend/svelte/src/lib/themes/mixins/_modal.scss
```diff
@@ -1,5 +1,5 @@
 @mixin section {
-  height: min(500px, calc(100vh - 156px - var(--padding-2x)));
+  max-height: calc(100vh - 156px - var(--padding-2x));
```
peterpeterparker, line 2: Looks generally good 👍

Only issue is that by removing the height that has a fix value the animation that happens when the modal loads becomes a bit glitchy

ezgif com-gif-maker(2)

wackyleo459: Let me know if you want me to target the `dissolve delay modal` in particular to set as min height 500px.

peterpeterparker: Thanks it solves one of the ui issue but the otherwise remains, opening modal is a bit glitchy.

Therefore, I would like to suggest the following: let's fix the modal height. It can solve both the general issue and the glitch.

To do so, we can set the following values (not the padding changes as well to reflect the change in `Modal.svelte`):

```
  --height: calc(100vh - var(--padding-6x) - 156px);
  min-height: var(--height);
  max-height: var(--height);
```

Does that work for you?

If yes, can you please within this PR also remove the `overflow-y: scroll;` in `SelectNeuronsToMerge.svelte`? I noticed that following this improvement it will lead to display two scrollbars within the same modal.

wackyleo459: I've been able to locate the place that was making modal transition glitchy, even when we gave it a min-height. So now, transition will be smooth even without a fixed modal height. I did try your suggestion, but the fixed height makes some modals (like Add Account) look too big (to me). Min height is set to 300px, but if you prefer closer to 500 let me know! 

I had to enable custom height for Dissolve Delay modal, to match the exact height once everything is loaded. So now if you need to set a certain height for a particular modal, you can. 

Also, curious what is the significance of 156px?

frontend/svelte/src/lib/modals/neurons/IncreaseDissolveDelayModal.svelte
```diff
@@ -36,7 +36,13 @@
   };
 </script>
 
-<WizardModal {steps} bind:currentStep bind:this={modal} on:nnsClose>
+<WizardModal
+  {steps}
+  customHeight="542px"
```
peterpeterparker, line 41: As much as possible we avoid to use JavaScript properties for styling purpose. If such a value has really to be passed through a component, then CSS variables should be used.

frontend/svelte/src/lib/themes/mixins/_modal.scss
peterpeterparker: Unfortunately still found one that is glitchy.

I would suggest to rollback to the fixed height solution I suggested. It makes the code simpler, introduces less changes of code and solve all the glitch.

Afterwards (once merged) we can have a look if we introduce a parameter to have compact size modal for smaller content.

Works for you?

ezgif com-gif-maker(7)

frontend/svelte/src/lib/modals/neurons/IncreaseDissolveDelayModal.svelte
wackyleo459: Understood, I'd like to see an example of this kind of use-case for future reference

peterpeterparker:
@wackyleo459 I did a review of the branch and ultimately noticed some few improvement ideas. Therefore I created a new branch (quicker than adding comments to the PR).

Please have a look at https://github.com/dfinity/nns-dapp/pull/842. If it works out for you we merge 842 in this PR 820 and then we merge it to main.

peterpeterparker (approved):
LGTM, thanks!

As briefly discussed, let's iterate from there. If fixes the issue, we can now improve its display for "medium" and various size of modals.

PR #822, opened 2022-05-10
feat: new transaction from hardware wallet

peterpeterparker:
# Motivation

Transfer ICP with the hardware wallet as source.

# Changes

- add `transformRequest` to ledger identity to use cbor to encode requests
- add function `getAccountIdentity` which returns either the user identity or the ledger identity linked to an account
- use above function in existing `transferICP` function
- display a successfull toast if send ICP works out
- add new `getLedgerIdentity` to ledger service that return the ledger identity by establishing a connection with the ledger
- in transaction information display a hint that the source is a hardware wallet
- keep wizard opens in case of ledger errors as the user might need a couple of tries to connect the ledger correctly

# UX

Note that UX changes such as the display of the hint, not implementing an unuseful connection step or keep the wizard open on error have been defined and reviewed together with Mischa.

# Demo

Success:
ezgif com-gif-maker(6)

Error:
ezgif com-gif-maker(5)


frontend/svelte/src/lib/identities/ledger.identity.ts
```diff
@@ -169,4 +177,32 @@ export class LedgerIdentity extends SignIdentity {
       await transport.close();
     }
   }
+
+  /**
+   * Required implementation for agent-js transformRequest.
+   *
+   * Without following function, transaction processed by the ledger would end in error "27012 - Data is invalid : Unexpected data type"
```
lmuntanerDfinity, line 184: 👏

frontend/svelte/src/lib/services/ledger.services.ts
```diff
@@ -82,13 +82,14 @@ export const registerHardwareWallet = async ({
...
+    toastsStore.error(
+      toLedgerError({
+        err,
+        fallbackErrorLabelKey: "error__attach_wallet.unexpected",
+      })
+    );
   }
 };
+
+export const getLedgerIdentity = (): Promise<LedgerIdentity> =>
```
lmuntanerDfinity, line 94: There is also a `getLedgerIdentity` in `accounts.services.ts`. Is it worth renaming to avoid confusion?

lmuntanerDfinity (approved):
Minor comments, LGTM!

frontend/svelte/src/lib/services/ledger.services.ts
peterpeterparker: good point, I will actually refactor + move thing a bit - i.e. move the function from `accounts.services.ts` to proxy to `ledger.services.ts`

give me 5...

peterpeterparker: done @lmuntanerDfinity 

good thing moving it that way also allow us to add test for the get ledger identity function because the logic isn't obfuscated in a private function anymore

PR #825, opened 2022-05-10
L2-527: Disable Confirm Spawn if not enough

lmuntanerDfinity:
# Motivation

User should not be able to confirm spawning a neuron if amount is not enough for a new neuron.

# Changes

* Calculate if percentageToSpawn is enough to create new neuron.
* Add disabled prop in SelectPercentage.

# Tests

* Add test for new use case.

### DIFF
--- a/frontend/svelte/src/lib/components/neuron-detail/SelectPercentage.svelte
+++ b/frontend/svelte/src/lib/components/neuron-detail/SelectPercentage.svelte
@@ -12,6 +12,7 @@
   export let neuron: NeuronInfo;
   export let percentage: number;
   export let buttonText: string;
+  export let disabled: boolean = false;
 
   let neuronICP: bigint;
   $: neuronICP = neuronStake(neuron);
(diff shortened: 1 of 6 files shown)


frontend/svelte/src/lib/components/neuron-detail/SelectPercentage.svelte
```diff
@@ -60,7 +61,7 @@
     data-tid="select-maturity-percentage-button"
     class="primary full-width"
     on:click={selectPercentage}
-    disabled={percentage === 0}
+    disabled={percentage === 0 || disabled}
```
peterpeterparker, line 64: you still need the `percentage === 0` since you introduced the `disabled` flag?

lmuntanerDfinity: good point, worth extracting out this logic and use only `disabled`

frontend/svelte/src/lib/modals/neurons/SpawnNeuronModal.svelte
```diff
@@ -33,6 +34,21 @@
   let percentageToSpawn: number = 0;
   let loading: boolean;
 
+  let enoughMaturityToSpawn: boolean;
+  $: {
```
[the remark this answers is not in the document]
lmuntanerDfinity: ok

lmuntanerDfinity:

peterpeterparker (approved):
LGTM, thanks

PR #828, opened 2022-05-11
Simplify deployments

bitdivine:
# Motivation
Make deployment simpler, faster and less error prone.  The dream:

* ./deploy.sh local <-- Deploys a complete local environment, with NNS canisters, II and NNS Dapp.
* Now run e2e tests or the front end development environment
* Ditto when deploying to testnets; that needs to get simpler, quicker and require less knowledge.

# Changes
* Add all the deployment steps to deploy.sh

# Future
* Use the global config in the e2e tests
* Delete unneeded code in e2e-tests
* make the local environment work well for testing on mobile
* The e2e tests use NNS_DAPP_URL, the rest of the code uses OWN_CANISTER_ID.  Make one var less.

# Tests
* This code is run in CI; that covers `./deploy.sh local`
* I have run `./deploy.sh local` followed by `cd frontend/svelte && npm ci && npm run dev` and verified that I can log in.
  * I similarly ran `./deploy.sh testnet` This failed at first as the wasm was too large.  On investigation num run dev creates some large files.  I resolved this with:
    * `git clean -fXd`
    * `./deploy.sh testnet`
    * `cd frontend/svelte`
    * `npm run dev`
    * Verify that I can open the dev URL, log in, get ICP.

UPDATE: In the main branch DEPLOY_ENV is now hard-wired in frontend/svelte/package.json ; this may cause problems, we have to see.

UPDATE:  I have tested this on M1.  It works with one big caveat:  All the dependencies needed to run build.sh need to be installed.  I had none of them installed so I had some work to do.  I recorded the steps I took and made a script.  I imagine that team members who use M1 on a daily basis have all of this installed already but it will help new team members and people moving to M1.


e2e-tests/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "wdio": "wdio run ./wdio.conf.ts",
-    "test": "tsc test.ts && node test.js",
```
lmuntanerDfinity, line 7: Remove the file `scripts/test` if not used anymore.

Does this also mean that we need to run `./deploy.sh` from the root folder before running the e2e tests?

lmuntanerDfinity (commented):
A couple of questions

deploy.sh
```diff
@@ -1,14 +1,204 @@
...
+      exit 1
+    }
+}
+
+
+if [[ "$START_DFX" == "true" ]]; then
+  echo
+  echo "Please run these commands in a separate terminal:"
+  echo
+  echo "  pkill dfx"
```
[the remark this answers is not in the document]
bitdivine: The problem is that dfx stop only works for dfx running in the current directory.  If dfx start was run in another project, say, `dfx stop` will not stop it and the following commands will fail.

e2e-tests/package.json
bitdivine: Rather, the plan is to remove test.ts as part of removing the proxy.  https://dfinity.atlassian.net/browse/L2-531  test.ts starts the proxy and then runs wdio.

The e2e tests work against any testnet, so you could deploy the canisters to localhost beforehand if that is where you want to test but you don't have to.

By the way, now that deploy.sh works for you, you should also be able deploy to the nnsdapp testnet with, e.g.:
```
./deploy.sh --nns-dapp nnsdapp_both
```

I suppose there should be a simple and obvious way of choosing which testnet to run against, with a sensible default, and the current implicit use of the global config without telling the user is not obvious.

I wonder whether I should include just deploy.sh in this PR and update the testing in a separate PR.  What do you think?

bitdivine: I think I would rather split the PR, if that is OK with you.

lmuntanerDfinity: yes, makes sense

deploy.sh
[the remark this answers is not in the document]
lmuntanerDfinity: ahh, that is true

e2e-tests/scripts/nns-canister-build
```diff
@@ -2,7 +2,7 @@
 set -euo pipefail
 cd "$(dirname "$0")"
 
-DOWNLOAD_DIR="$(realpath ../../target/ic)"
+DOWNLOAD_DIR="$(realpath ../..)/target/ic"
```
bitdivine, line 5: Minor fix: The code on the left breaks if target doesn't exist yet.

e2e-tests/package.json
bitdivine: Done.  Can we merge the deploy script please?

deploy.sh
```diff
@@ -1,14 +1,203 @@
...
+    DEPLOY_NNS_DAPP="true"
+    ;;
+  --open)
+    OPEN_NNS_DAPP="true"
+    ;;
+  --dry-run)
+    DRY_RUN="true"
+    ;;
+  *)
+    DFX_NETWORK="$env"
```
lmuntanerDfinity, line 112: what does this do? set `DFX_NETWORK` to "--dry-run"??

deploy.sh
```diff
@@ -1,14 +1,203 @@
...
+  esac
+fi
+
+echo
+echo START_DFX=$START_DFX
+echo DEPLOY_NNS_BACKEND=$DEPLOY_NNS_BACKEND
+echo DEPLOY_II=$DEPLOY_II
+echo DEPLOY_NNS_DAPP=$DEPLOY_NNS_DAPP
+echo OPEN_NNS_DAPP=$OPEN_NNS_DAPP
+[[ "$DRY_RUN" != "true" ]] || exit 0
```
lmuntanerDfinity, line 146: not sure I understand this line. What is it doing?

lmuntanerDfinity (approved):
Minor comments. Already approving.

lmuntanerDfinity:
@bitdivine why do you add as a change "Use deploy.sh in CI". I don't see that part?

bitdivine:
@lmuntanerDfinity - good point, it is no longer there.  Removing from the changelog.

deploy.sh
bitdivine: It sets DFX_NETWORK to the argument, so e.g. `./deploy.sh testnet` and `./deploy.sh local` deploy to `testnet` and `local` respectively.

lmuntanerDfinity: 👍

lmuntanerDfinity: because it's not any of the previous cases?

PR #831, opened 2022-05-11
feat: show principal and address on hardware wallet

peterpeterparker:
# Motivation

On account detail page, user should be able to call a function that displays principal and address on the hardware wallet.

# Changes

- fix ledger identity functions' callback that need `this` at runtime
- add `showAddressAndPubKeyOnHardwareWallet` to service
- create a component that contains a `button` to execute above function


frontend/svelte/src/routes/Wallet.svelte
```diff
@@ -48,7 +49,11 @@
     <svelte:fragment slot="header">{$i18n.wallet.title}</svelte:fragment>
 
     {#if mainAccount}
-      <section>TBD - TODO(L2-429)</section>
+      <section>
+        <h1>TBD - TODO(L2-429)</h1>
+
+        <HardwareWalletShowAction />
```
mstrasinskis, line 55: Just curious, shouldn't there be something like {#if isHardwareWallet .... ?

mstrasinskis (approved):
LGTM, one question?

frontend/svelte/src/routes/Wallet.svelte
peterpeterparker: Yes, definitely. There will be another action for the hardware wallet "Show neurons" and one for "Rename subaccount" so my idea is to create later an "AccountActionToolbar" component or something in which we will implement the logic you suggested. Just thought we can do that in a separate PR afterwards.

peterpeterparker: i'll do it in my next pr for "Rename subaccount"

PR #833, opened 2022-05-11
Pre-build nns canister docker file for caching

bitdivine:
# Motivation
CI builds of several steps are slow.  The worst offender is the NNS canister build.

# Changes
- Put the NNS canister build into a separate job and cache the outputs

# Tests
See the output and time of the "Native build" CI test.  Here is the history of that workflow: https://github.com/dfinity/nns-dapp/actions/workflows/build.yml

The runtime of that workflow is reduced from ~28 minutes to ~12 minutes.


.github/workflows/build.yml
```diff
@@ -12,18 +12,63 @@ name: Native build
 on:
   push:
```
bitdivine, line 14: The env vars are pushed up here, from the job to the workflow scope, so that they are available to all jobs.

.github/workflows/build.yml
```diff
@@ -12,18 +12,63 @@ name: Native build
...
-    runs-on: ubuntu-20.04
-    timeout-minutes: 60
-    env:
+
+env:
       DFX_VERSION: 0.9.2
       RUSTC_VERSION: 1.58.1
       FLUTTER_VERSION: 2.2.3
       IC_CDK_OPTIMIZER_VERSION: 0.3.1
+      IC_COMMIT: 6dc5e2dcb50569e20891af02b846a7d9a58e3489
```
bitdivine, line 21: This commit is used in several places across the codebase and needs to be factored out.  But not in this PR.

.github/workflows/build.yml
```diff
@@ -12,18 +12,63 @@ name: Native build
...
+          cache-from: type=gha,scope=cached-stage
+          cache-to: type=gha,scope=cached-stage,mode=max
+          outputs: target/ic
+
+  build:
+    needs: [nns_canisters]
+    runs-on: ubuntu-20.04
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v2
```
bitdivine, line 63: This is the same step as above, except that this time we are just fetching the canister wasm files from cache.  As the other job runs first, this should always be a cache hit.

.github/workflows/build.yml
```diff
@@ -97,8 +142,6 @@ jobs:
       # REDIRECT_TO_LEGACY=svelte builds and deploys only Svelte
       - name: Deploy NNS canisters
         run: |
-          ./e2e-tests/scripts/nns-canister-download
-          ./e2e-tests/scripts/nns-canister-build
```
bitdivine, line 101: The above two steps used to get and build the wasm files; they are replaced by the caching code of this PR.

peterpeterparker (approved):
Rubber stamp, thanks

PR #834, opened 2022-05-12
L2-523: Screen to add user's principal to hotkeys

lmuntanerDfinity:
# Motivation

User can add the principal of the current user to a new neuron creaed with a Hardware Wallet.

# Changes

* Add new step in CreateNeuronModal if the chosen account is a Hardware Wallet.
* New Screen AddUserToHotkeys.
* New neuron service addHotkeyFromHW.
* Governance Canister is created with "hardwareWallet" flag if identity is LedgerIdentity.

# Tests

* Add test for new use case in CreateNeuronModal
* Add test for new neuron service.
* Fix and change old tests.


lmuntanerDfinity:
Video of the new screen and flow:

Sorry for the long time, this is how it takes to confirm all the steps with the hardware wallet.

I created an issue to improve that: https://dfinity.atlassian.net/browse/L2-533

https://user-images.githubusercontent.com/98735375/168027975-ace2034d-2ded-46fc-b5aa-9249553a1c8a.mov

lmuntanerDfinity:

frontend/svelte/src/lib/components/neurons/AddUserToHotkeys.svelte
```diff
@@ -0,0 +1,116 @@
...
+  import type { NeuronInfo } from "@dfinity/nns";
+
+  export let account: Account;
+  export let neuron: NeuronInfo;
+
+  let loading: boolean = false;
+
+  let principal: Principal | undefined;
+  onMount(async () => {
+    principal = (await getIdentity()).getPrincipal();
```
peterpeterparker, line 23: why not "just" use the store subscriber `$authStore.identity?.getPrincipal()`?

frontend/svelte/src/lib/components/neurons/AddUserToHotkeys.svelte
```diff
@@ -0,0 +1,116 @@
...
+
+  const dispatcher = createEventDispatcher();
+  const dispatchNext = () => {
+    dispatcher("nnsNext");
+  };
+
+  const addCurrentUserToHotkey = async () => {
+    loading = true;
+    startBusy("add-hotkey-neuron");
+    if (principal !== undefined) {
```
peterpeterparker, line 37: I would not test the principal here but rather re-request the `getIdentity` within the service to get the principal

If the user wait x minutes between displaying this component and calling the function, it might become undefined

Of course we have the cron that logout but just for the state of the art

frontend/svelte/src/lib/components/neurons/AddUserToHotkeys.svelte
```diff
@@ -0,0 +1,116 @@
...
+      })}
+    </p>
+  </div>
+  <div class="buttons">
+    <button
+      on:click={dispatchNext}
+      data-tid="skip-add-principal-to-hotkey-modal"
+      class="secondary full-width">{$i18n.neurons.skip}</button
+    >
+    <button
```
peterpeterparker, line 80: should the button be disabled until the principal is set?

peterpeterparker:
can you add a TODO to parse the ledger error correctly? Technically you can do it already but I renamed/refactored the function that parse the error to toast in #835

IS:

Capture d’écran 2022-05-12 à 12 15 02

Should be:

Capture d’écran 2022-05-12 à 12 14 52

peterpeterparker:
"Update delay"? Shouldn't it be "add hotkey"?

Capture d’écran 2022-05-12 à 12 19 44

peterpeterparker:
After I added the hotkey, I faced an error when updating the delay, is that normal - i.e. this will be done in another PR?
I guess so, just in case

Capture d’écran 2022-05-12 à 12 20 53

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
```diff
@@ -101,15 +111,17 @@
     detail,
   }: CustomEvent<{ selectedAccount: Account }>) => {
     selectedAccount = detail.selectedAccount;
+    if (isHardwareWallet(selectedAccount)) {
+      steps.push(extraStepHW);
+    }
+    steps.push(...lastSteps);
```
peterpeterparker, line 117: - what happens if the user goes back - i.e. can the user goes back in the wizard ad should the steps be reverted?

- when I dynamically change the steps I generally add  `await tick()` before calling the next `modal.next` to avoid possible UI glitch. would that be something interesting here?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -484,6 +500,31 @@ export const mergeNeurons = async ({
   }
 };
 
+// This service is used when creating a new neuron from a Hardware Wallet.
+// The new neuron is not yet in the neuronsStore.
+export const addHotkeyFromHW = async ({
```
peterpeterparker, line 505: why do you need two different functions to add the hotkey - i.e. why can't use `addHotkey` for hw too?

and if really mandatory, why this function does not call `getAndLoadNeuronHelper` after the hotkey has been added?

frontend/svelte/src/lib/components/neurons/AddUserToHotkeys.svelte
lmuntanerDfinity: I need the principal to pass it to `addHotkey`.

frontend/svelte/src/lib/components/neurons/AddUserToHotkeys.svelte
lmuntanerDfinity: sure

frontend/svelte/src/lib/components/neurons/AddUserToHotkeys.svelte
lmuntanerDfinity: sure

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
lmuntanerDfinity: The next step cannot go back. None of them. There is no back button. Creating a neuron is the step where you can't go back, no matter whether it's hardware wallet or not.

lmuntanerDfinity: I'll add a `tick`

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: I tried to explain it in the comment but I see that I failed.

The neuron is not in the neuronsStore, therefore we can't rely on the helpers that use it. I didn't want to trigger a "getNeuron" call, which would trigger another hardware wallet call and two more confirmation for the user.

I decided for this to have a better UX without changing the code.

I also created an issue to avoid getting the neuron after creation: https://dfinity.atlassian.net/browse/L2-533

But that requires a few more changes and I'd rather do it in another PR.

lmuntanerDfinity:
Oops, I changed the button text.

Yes, the dissolve delay and the rest wil be done in another PR and issue: https://dfinity.atlassian.net/browse/L2-524

Which is blocked by this PR.

I don't understand where do I need to add the TODO to parse the error? Can you add a comment where you mean?

lmuntanerDfinity:

lmuntanerDfinity:
Updated screenshot of the screen:

Screenshot 2022-05-12 at 13 11 18

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
peterpeterparker: Super

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: Discussed in a call

PR #835, opened 2022-05-12
feat: rename subaccount

peterpeterparker:
# Motivation

Being able to rename linked account.

# Note

In Flutter the function is access through a button/text in the navbar of the account's detail page.
I suggest to move the action to a menubar within the page as it would follow the pattern of the hardware wallet actions that are already displayed within the page.
Mischa is agree with this change.

# Changes

- new account api and service function `renameSubaccount`
- new components to implement the feature `RenameSubaccount*.svelte`
- new component for the menubar `WalletActions`
- style toast for break word if error contains a super long word
- refactor error mapper / utils to parse custom error messages to toast

# Demo

gif


frontend/svelte/src/lib/utils/error.utils.ts
```diff
@@ -52,15 +50,23 @@ export const mapNeuronErrorToToastMessage = (error: Error): ToastMsg => {
...
   err,
+  errorsWithMessage = [],
   fallbackErrorLabelKey,
 }: {
-  err: unknown;
+  err: unknown | undefined;
+  errorsWithMessage?: { new (message?: string): Error }[];
   fallbackErrorLabelKey: string;
 }): { labelKey: string; err?: unknown } => {
   const ledgerErrorKey: boolean =
```
mstrasinskis, line 67: Probably just errorKey?

frontend/svelte/src/lib/services/accounts.services.ts
```diff
@@ -170,3 +182,54 @@ export const getAccountIdentity = async (
...
+
+const renameError = ({
+  labelKey,
+  err,
+}: {
+  labelKey: string;
+  err?: unknown;
+}): { success: boolean; err?: string } => {
+  toastsStore.error(
+    toToastError({
```
mstrasinskis, line 228: Does `toToastError` is preferable here?

frontend/svelte/src/lib/utils/error.utils.ts
```diff
@@ -52,15 +50,23 @@ export const mapNeuronErrorToToastMessage = (error: Error): ToastMsg => {
...
+ * If the error does not contain a matching "message", it fallbacks to the "fallbackErrorLabelKey" and add the error as details of the toast.
+ */
+export const toToastError = ({
   err,
+  errorsWithMessage = [],
   fallbackErrorLabelKey,
 }: {
-  err: unknown;
+  err: unknown | undefined;
+  errorsWithMessage?: { new (message?: string): Error }[];
```
mstrasinskis, line 64: Out of curiosity: why not blindly map the message to the i18n errors without checking of `errorsWithMessage`?

mstrasinskis (approved):
LGTM, just a couple of questions.

frontend/svelte/src/lib/utils/error.utils.ts
peterpeterparker: because the err message might actually be the i18n key, it would duplicate it in the displayed error

Pseudo code

```
// i18n
{
  error: {
     ledger_is_not_connected: "Your ledger is not connected."
     
// Throw
throw new CustomError('error.ledger_is_not_connected')

// blindly map

// Error toast UI
Your ledger is not connected. error.ledger_is_not_connected

// With mumbo jumbo

// Error toast UI
Your ledger is not connected.
```

frontend/svelte/src/lib/services/accounts.services.ts
peterpeterparker: No, just convenient at the end of the day. Can remove it if you rather like?

mstrasinskis: No, it's cool. Thanks.

frontend/svelte/src/lib/utils/error.utils.ts
mstrasinskis: I mean instead of the relying on the object type we could check the existence of the label in i18n (we kind of already do it in the toast) and don't add the {err} if the label exists. Probably I didn't get the concept 😄

peterpeterparker: Aaaah I get it, sure that would work too. Updated  and pushed in same pr code as follow:

```
export const toToastError = ({
  err,
  fallbackErrorLabelKey,
}: {
  err: unknown | undefined;
  fallbackErrorLabelKey: string;
}): { labelKey: string; err?: unknown } => {
  let errorKey: boolean = false;
  const message: string | undefined = (err as Error)?.message;

  if (message !== undefined) {
    const label: string = translate({labelKey: message});
    errorKey = label !== message;
  }

  return {
    labelKey: errorKey
      ? (err as { message: string }).message
      : fallbackErrorLabelKey,
    ...(!errorKey && { err }),
  };
};
```

PR #836, opened 2022-05-12
L2-514: Hide slider spawning from hardware wallet

lmuntanerDfinity:
# Motivation

Neurons controlled with hardware wallets do not choose percentage when spawning.

# Changes

* Add a flag prop on SpawnNeuronModal to identify steps to show.
* New neuron util "isNeuronControlledByHardwareWallet"
* "percentageToSpawn" property is optional

# Tests

* New test case for SpawnNeuronModal
* New test for neuron util "isNeuronControlledByHardwareWallet"


lmuntanerDfinity:
UX of hardware wallet controlled neuron:

https://user-images.githubusercontent.com/98735375/168080365-db17b3a6-dfea-4e0e-8e2e-7648ed926d6b.mov

UX of nns dapp controlled neuron:

https://user-images.githubusercontent.com/98735375/168080469-afe17b8e-ac9a-47d2-97f1-a6518076ecc1.mov

lmuntanerDfinity:

frontend/svelte/src/lib/api/governance.api.ts
```diff
@@ -124,7 +124,7 @@ export const spawnNeuron = async ({
   identity,
 }: {
   neuronId: NeuronId;
-  percentageToSpawn: number;
+  percentageToSpawn?: number;
```
peterpeterparker, line 127: can you add a comment somewhere that says why the `percentageToSpawn` - that it is not yet supported by the current ledger? pretty sure I'll forget

frontend/svelte/src/lib/modals/neurons/SpawnNeuronModal.svelte
```diff
@@ -88,10 +101,12 @@
           {replacePlaceholders(
             $i18n.neuron_detail.spawn_maturity_confirmation_a,
             {
-              $percentage: formatPercentage(percentageToSpawn / 100, {
-                minFraction: 0,
-                maxFraction: 0,
-              }),
+              $percentage: controlledByHarwareWallet
```
peterpeterparker, line 104: that logic "if hw 100 else math" is maybe worth a function?

frontend/svelte/src/lib/utils/accounts.utils.ts
```diff
@@ -19,8 +19,9 @@ export const getAccountByPrincipal = ({
     return accounts.main;
   }
 
-  // TODO: Check also the hardware wallets L2-433
-  return undefined;
+  return accounts.hardwareWallets?.find(
+    (account) => account.principal?.toText() === principal
```
peterpeterparker, line 23: if you want to go fancy, you can deconstruct 

```
const {hardwareWallets, main} = accounts;

if (main?.principal?.toText() === principal) {
    return accounts.main;
  }
  
  return hardwareWallets?.find(
    ({principal: hwPrincipal}) => hwPrincipal?.toText() === principal
```

lmuntanerDfinity:

lmuntanerDfinity:

frontend/svelte/src/lib/utils/neuron.utils.ts
```diff
@@ -168,6 +168,23 @@ export const isNeuronControllable = ({
     getAccountByPrincipal({ principal: fullNeuron.controller, accounts }) !==
       undefined);
 
+export const isNeuronControlledByHardwareWallet = ({
```
peterpeterparker, line 171: what's the different between `neurons.service -> isNeuronControlledByHardwareWallet` and `neuron.util -> isNeuronControlledByHardwareWallet`?

to avoid confusion, can or should one of these two function be renamed?

frontend/svelte/src/lib/api/governance.api.ts
```diff
@@ -125,7 +125,8 @@ export const spawnNeuron = async ({
   identity,
 }: {
   neuronId: NeuronId;
-  percentageToSpawn: number;
+  // percentageToSpawn is not yet supported by the ledger IC app
+  percentageToSpawn?: number;
```
peterpeterparker, line 129: do we have a test that check that this variable is not set in case of hardware wallet (if possible)?

frontend/svelte/src/lib/modals/neurons/SpawnNeuronModal.svelte
```diff
@@ -40,13 +51,23 @@
     percentage: percentageToSpawn,
   });
 
+  let percentageMessage: string;
+  $: percentageMessage = controlledByHarwareWallet
```
peterpeterparker, line 55: do we have a test that check that the message displays 100% for harware wallet?

frontend/svelte/src/lib/utils/neuron.utils.ts
lmuntanerDfinity: Sorry, old stuff. I will remove the service

frontend/svelte/src/lib/api/governance.api.ts
lmuntanerDfinity: Yes, I added the test in the SpawnNeuronModal that if the neuron is controlled by a hardware wallet it goes directly to the screen to confirm.

This is the closes I can do it. All the other functions that we use do not know about the account.

frontend/svelte/src/lib/modals/neurons/SpawnNeuronModal.svelte
lmuntanerDfinity: No, I have a test that it goes directly to the confirmation screen without going to the select percentage.

I can add the `expect` to check that 100% is present.

frontend/svelte/src/lib/utils/neuron.utils.ts
lmuntanerDfinity: done

lmuntanerDfinity:

lmuntanerDfinity:
Video with the new UI discussed with Mischa.

https://user-images.githubusercontent.com/98735375/168229200-5fcf6c28-83ab-48f1-ae2c-5c1ae015bffb.mov

frontend/svelte/src/lib/api/governance.api.ts
peterpeterparker: good to know there is test.

it leads me to ask my self if we should not move the logic from the component to the service?

right now the component implements the logic

```
percentageToSpawn: controlledByHarwareWallet
        ? undefined
        : percentageToSpawn,
```        
        
maybe we could move this to the service - i.e. make the service check if hardware wallet and if so use undefined and not percentage.

doing so we can implement more accurate test.

what do you think? overengineered or make sense?

frontend/svelte/src/lib/modals/neurons/SpawnNeuronModal.svelte
peterpeterparker: Yes maybe. There was multiple user reports about the issue before we fixed in Flutter, that's why I am specially paying attention to the subject...

frontend/svelte/src/lib/modals/neurons/CreateNeuronModal.svelte
```diff
@@ -124,6 +124,12 @@
   };
   const addNeuron = ({ detail }: CustomEvent<{ neuron: NeuronInfo }>) => {
     newNeuron = detail.neuron;
+    if (isHardwareWallet(selectedAccount)) {
+      toastsStore.show({
```
peterpeterparker, line 128: should we add a test for this? or more an e2e thing?

lmuntanerDfinity: It's a very specific UI thing, not sure we should test that with our tests.

frontend/svelte/src/lib/modals/neurons/SpawnNeuronModal.svelte
lmuntanerDfinity: done

frontend/svelte/src/lib/api/governance.api.ts
lmuntanerDfinity: I think it's a little overengineered. Specially since this is temporary until the new Ledger App, as well as this is in a project that we should substitute in one year.

lmuntanerDfinity:

frontend/svelte/src/lib/api/governance.api.ts
peterpeterparker: One can argue that for the same substitution reason, it would be safer to have the logic within a service because it makes easier to reuse the code. Glass half-full vs half-empty 😉.

Thanks for the feedback, alright then.

PR #837, opened 2022-05-12
[FIX] http not https

bitdivine:
# Motivation
Local dfx deployments use http, not https

# Changes
* Remove the offending line; the default is correct and doesn't need to be overridden.

# Tests
I have deployed canisters to a local dfx and they work.


peterpeterparker (approved):
LGTM

PR #841, opened 2022-05-12
Fix typescript in e2e tests

bitdivine:
# Motivation
There are a lot of typescript errors in the e2e tests, strangely not caught by eslint.  This makes it easy to write typos and similar mistakes that SHOULD be caught at compile time but that wdio either swallows silently or causes wdio to die with very hard to debug error messages.

So let's catch the typescript errors.

Running tsc on the e2e test code does just that, and actually finds problems.

# Changes
- Run tsc as part of `npm run lint`
- Fix 36 errors.
- Add esModuleInterop due to tsc warnings in third party libraries.

# Tests
The e2e tests pass locally, and should also pass on CI.

### DIFF
--- a/e2e-tests/common/constants.ts
+++ b/e2e-tests/common/constants.ts
@@ -18,8 +18,6 @@ export enum RedirectToLegacy {
 
 export const REDIRECT_TO_LEGACY: RedirectToLegacy = ((
   env = RedirectToLegacy.prod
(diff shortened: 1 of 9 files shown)


lmuntanerDfinity (approved):
Thanks! Wow, some of the errors were like not having TS at all, so weird.

PR #842, opened 2022-05-13
style: modal height review

peterpeterparker:
# Motivation

Review PR #820 and provide suggestions.

It contains some changes for the modal height ("small and medium height modals") but also some few changes for the spacing on mobile devices as I tested the size on the modal on such devices too.

### DIFF
--- a/frontend/svelte/src/lib/components/accounts/AddAccountType.svelte
+++ b/frontend/svelte/src/lib/components/accounts/AddAccountType.svelte
@@ -29,14 +29,19 @@
 
 <style lang="scss">
   @use "../../themes/mixins/interaction";
+  @use "../../themes/mixins/media";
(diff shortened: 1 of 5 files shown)


wackyleo459:
I've left comments above.  The AddHotKey modal could definitely be using class of `small`. I'm not sure if you wanted to address the Add Account modal. Perhaps the height can remain the same, but display content at the top. (That could be for another issue!). Other than that, the idea is great thank you!

peterpeterparker:
> I've left comments above. 

@wackyleo459 mmmh I don't see any commments.

> The AddHotKey modal could definitely be using class of small

`small` is for modal of type yes/no. it might need a new type "medium" but let's move forward first I would say and solve this in another PR.

PR #845, opened 2022-05-16
build: fix canister id for local development

peterpeterparker:
# Motivation

Testnet has been redeployed, fix the canister id that has changed.

# Changes

- update nns-both canister id in `package.json`


lmuntanerDfinity (approved):
thanks!

PR #846, opened 2022-05-16
L2-486: Neurons functionality compatible with hardware wallet

lmuntanerDfinity:
# Motivation

User is able to manage a neuron from a Hardware Wallet.

# Changes

* Get the identity based on the neuron controller where suited.
* Prefer the user identity if the functionality can be managed by a hotkey.
* Split and Join Community Fund only available for neurons whose controller is the nns dapp user. Like in Svelte.
* New neuron util "getIdentityByPrincipal" and "isNeuronControllableByUser".
* Helper "getAndLoadNeuron" uses user principal.

# Tests

* Fix broken tests.
* Add tests for HW cases in Merge Neurons Modal and "mapMergeable" util.
* Tests for "getIdentity" helpers.

### DIFF
--- a/frontend/svelte/src/lib/services/neurons.services.ts
+++ b/frontend/svelte/src/lib/services/neurons.services.ts
@@ -55,7 +55,11 @@ import {
   isIdentityController,
 } from "../utils/neuron.utils";
 import { createChunks, isDefined } from "../utils/utils";
-import { getAccountIdentity, syncAccounts } from "./accounts.services";
+import {
+  getAccountIdentity,
+  getAccountIdentityByPrincipal,
+  syncAccounts,
+} from "./accounts.services";
 import { getIdentity } from "./auth.services";
 import { queryAndUpdate } from "./utils.services";
 
@@ -116,12 +120,15 @@ const getNeuronFromStore = (neuronId: NeuronId): NeuronInfo | undefined =>
 const getIdentityByNeuron = async (neuronId: NeuronId): Promise<Identity> => {
   const { neuron } = await getIdentityAndNeuronHelper(neuronId);
 
-  if (neuron.fullNeuron === undefined) {
+  if (
+    neuron.fullNeuron === undefined ||
+    neuron.fullNeuron.controller === undefined
+  ) {
     throw new NotAuthorizedError();
   }
 
-  const neuronIdentity = await getAccountIdentity(
-    neuron.fullNeuron.accountIdentifier
+  const neuronIdentity = await getAccountIdentityByPrincipal(
+    neuron.fullNeuron.controller
   );
   if (isIdentityController({ neuron, identity: neuronIdentity })) {
     return neuronIdentity;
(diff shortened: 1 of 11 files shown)


lmuntanerDfinity:

Same UX as currently, except interaction with the ledger, which doesn't show in the video.

Merging maturity is fixed when using Ledger IC app v2.

Merge neurons from HW still not working. Hamish is working on the fix, as soon as that is in, I will also fix it in Svelte.

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -113,15 +118,20 @@ const getNeuronHW = async ({
...
+  ) {
     throw new NotAuthorizedError();
   }
 
-  const neuronIdentity = await getAccountIdentity(
-    neuron.fullNeuron.accountIdentifier
+  const neuronIdentity = await getAccountIdentityByPrincipal(
+    neuron.fullNeuron.controller
   );
   if (isIdentityController({ neuron, identity: neuronIdentity })) {
```
peterpeterparker, line 136: you don't need this check anymore I think since you get the neuron with the controller

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -484,7 +495,9 @@ export const mergeNeurons = async ({
         translate({ labelKey: messageKey ?? "error.governance_error" })
       );
     }
-    const identity: Identity = await getIdentityByNeuron(targetNeuronId);
+    const identity: Identity = await getIdentityOfNeuronController(
```
peterpeterparker, line 498: function is called `getIdentityOfNeuronController` but you pass a `targetNeuronId` - i.e. not a controller - as parameter. Is that correct?

lmuntanerDfinity: yes, the method read the neuron from the store, check the controller and returns the identity of that controller.

Maybe `getIdentityOfControllerByNeuronId`?

peterpeterparker: `getIdentityOfControllerByNeuronId` 👍

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: Yes, good point, I can remove it. It's an edge case (I hope) impossible to have.

lmuntanerDfinity: Actually, I checked. And I think it's better to have it.

`getIdentity` returns the current user identity by default if the account is not a hardware wallet.

In case we enabled visiting neurons which are not ours, we would need this service to throw the not authorized error.

At the moment, we should never reach it. But I think that based on the extra checks we have, this is not a crazy one either.

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: done

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: Works for me. Can you add a comment?

lmuntanerDfinity:

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: yes, good point

lmuntanerDfinity: done!

lmuntanerDfinity:

peterpeterparker (approved):
LGTM, thx

PR #850, opened 2022-05-16
Call directly into CMC to create or top up a canister

hpeebles:
# Motivation

The cycles minting canister has been modified with 2 new methods, `notify_create_canister` and `notify_top_up`. These new methods mean that people no longer need to call `notify` on the ledger when creating or topping up a canister. This will allow us to deprecate the `notify` method on the ledger and eventually remove it, simplifying the ledger.

This change makes the NNS Dapp call directly into the new CMC methods rather than `notify` on the ledger.

# Changes

- The NNS Dapp canister calls directly into `notify_create_canister` when creating a canister
- The NNS Dapp canister calls directly into `notify_top_up` when topping up a canister

# Tests

I have deployed this branch to https://qvhpv-4qaaa-aaaaa-aaagq-cai.small07.dfinity.network/ and have tested creating and topping up canisters. Both flows work as expected.


bogwar (approved):
LGTM!

PR #851, opened 2022-05-16
feat: list harware wallet neurons

peterpeterparker:
# Motivation

List the hardware wallet neurons from the action on the wallet detail page.

# Changes

- new modal and related display components
- new function `listNeuronsHardwareWallet` added to `ledger.services` and proxy

# Demo

ezgif com-gif-maker


frontend/svelte/src/lib/components/accounts/HardwareWalletListNeurons.svelte
```diff
@@ -0,0 +1,46 @@
+<script lang="ts">
```
lmuntanerDfinity, line 1: Maybe rename this file to something `XXButton`, like `ShowHardwareWalletNeuronsButton`?

Otherwise it seems that it will render the list itself.

peterpeterparker: Thanks, done!

PR #854, opened 2022-05-17
style: busy text as paragraph

peterpeterparker:
# Motivation

Design is personal but I feel like the text of the busy screen was screaming at me 😂.

# Changes

- display text with `p`
- add spacing
- limit width in case passed text would be really long

# Screenshots

Is:

Capture d’écran 2022-05-17 à 07 23 44

Will be:

Capture d’écran 2022-05-17 à 07 24 47
Capture d’écran 2022-05-17 à 07 26 43

### DIFF
--- a/frontend/svelte/src/lib/components/ui/BusyScreen.svelte
+++ b/frontend/svelte/src/lib/components/ui/BusyScreen.svelte
@@ -10,7 +10,7 @@
   <div data-tid="busy" transition:fade>
     <div class="content">
       {#if $busyMessageKey !== undefined}
-        <h4>{translate({ labelKey: $busyMessageKey })}</h4>
+        <p>{translate({ labelKey: $busyMessageKey })}</p>
       {/if}
       <span>
         <Spinner inline />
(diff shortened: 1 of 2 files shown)


frontend/svelte/src/lib/components/ui/BusyScreen.svelte
```diff
@@ -35,4 +35,9 @@
     justify-content: center;
     align-items: center;
   }
+
+  p {
+    padding-bottom: var(--padding);
```
lmuntanerDfinity, line 40: maybe use a `gap` in the `content`?

and ditto on the max-width.

In case we ever want to add more elements in the content, this applies to all of them.

What do you think?

lmuntanerDfinity (approved):
Minor comment.

Already approving.

frontend/svelte/src/lib/components/ui/BusyScreen.svelte
peterpeterparker: Agree gap would work too but thinking about it, maybe neither this padding nor the gap will looks well when we will add more elements 🤷‍♂️.

maybe we would like such a design:

p: line-height and no padding -> 5px spacing
p: line-height and no padding -> 5px spacing
spinner: margin-top ->  10px

so would say let's keep as it and review it when and if we add more elements

PR #856, opened 2022-05-17
chore: remove unused global transaction store

peterpeterparker:
# Motivation

Remove dead code and avoid misusage of a store.

# Changes

- remove unused global `transactionStore`
- remove unused global `addAccountStore`
- move the `transaction.store.ts` from folder "store" to "types" and renamed it into `transaction.context.ts`
- move the `add-account.store.ts` from folder "store" to "types" and renamed it into `add-account.context.ts`


mstrasinskis (approved):
LGTM, much cleaner 👍

PR #858, opened 2022-05-17
Patch and command needed to set the xrate

bitdivine:
# Motivation
The exchange rate needs to be set for operations such as creating a canister.

# Changes
- Patch the CMC canister to allow setting the exchange rate directly.
- Provide a script to call the exchange rate setting mechanism.

# Tests
- I have used this to set the exchange rate on small06, small07 and on local dfx.  This did indeed allow creating canisters.  

Note: This patch is necessary but not sufficient for a complete setup.  It is also necessary to run some proposals but those are in a separate PR.


hpeebles (approved):

nmattia (approved):
gix stamp of approval

PR #860, opened 2022-05-17
L2-571: Redirect if current user removed hotkeys

lmuntanerDfinity:
# Motivation

User is redirected to neurons page if it removes itself from the hotkeys.

# Changes

* Check that the neuron has `fullNeuron` when querying a neuron.
* Throw NotFound if user is not authorized to see fullNeuron.
* Change "removeHotkey" service to return neuron id as success if a NotAuthorized is thrown and the hotkey as been removed.

# Tests

* Add a test to check that user is redirected after removing itself from hotkeys.


lmuntanerDfinity:
UX Video:

https://user-images.githubusercontent.com/98735375/168837415-378326e4-d2e5-4f8f-8546-01450cea35ce.mov

frontend/svelte/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -1,3 +1,6 @@
+/**
+ * @jest-environment jsdom
```
lmuntanerDfinity, line 2: Needed because now we import routeStore which imports route.utils, which needs the dom

frontend/svelte/src/tests/lib/services/busy.services.spec.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * @jest-environment jsdom
```
lmuntanerDfinity, line 2: busy services imports from neurons store, which needs the dom.

frontend/svelte/src/tests/lib/services/proposals.services.spec.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * @jest-environment jsdom
```
lmuntanerDfinity, line 2: proposals services imports from neurons services which needs the dom

lmuntanerDfinity:

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -380,7 +383,7 @@ const getAndLoadNeuron = async (neuronId: NeuronId) => {
     certified: true,
     forceFetch: true,
   });
-  if (!neuron) {
+  if (!neuron || !userAuthorizedNeuron(neuron)) {
```
peterpeterparker, line 386: shouldn't it throws another type of error - i.e. it isn't "not found" it is "not authorized" or am I missing something?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -542,18 +545,29 @@ export const removeHotkey = async ({
...
     await removeHotkeyApi({ neuronId, identity, principal });
+    hotkeyRemoved = true;
 
     await getAndLoadNeuron(neuronId);
 
     return neuronId;
   } catch (err) {
-    toastsStore.show(mapNeuronErrorToToastMessage(err));
+    // It happens when the current user is removed from hotkey
+    if (err instanceof NotFoundError && hotkeyRemoved) {
```
peterpeterparker, line 562: you are assuming that absolutely 100% of the errors that can happen in `getAndLoadNeuron` are related to the hotkey that has been removed. looks a bit optimistic to me.

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -542,18 +545,29 @@ export const removeHotkey = async ({
...
+      toastsStore.show({
+        level: "success",
+        labelKey: "neurons.remove_hotkey_success",
+      });
+      routeStore.replace({ path: AppPath.Neurons });
```
peterpeterparker, line 567: this makes the function tight with the location - component and navigation - it is used - e.g. if someone ever used the function on the detail accounts page, user might be redirect to neurons. should we use an optional callback to perform the redirection?

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: We could also do that, I went more the Github model, where if you don't have access to the repo you get a 404.

I guess it's also a security measure, this way you can't even know whether the neuron exists or not. If you get Unauthorized, then you know it exists.

I added a comment that both are handled the same way in this case.

Shall I do the Unauthorized?

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: Good point. I can also check that the hotkey removed is the current user before showing this message.

Otherwise, fallback to the other error management.

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: I like the idea, I can even remove all this logic from here and simplify it.

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: oh so if the error is throw the user is logged out?

anyway was just thinking, if we display the same error and user reach out the support, it might be tricky to know which of the two errors the user is facing.

can live with it though, was just a comment.

lmuntanerDfinity: it depends on what we do with the error, I consider not finding a neuron the same as that the user has no access to that neuron.

But actually, maybe the error can be different, and we handle it similar in the UI. This way it's up to the caller to define the same UX for both, not the one raising the error.

That looks better.

lmuntanerDfinity: I throw the Unauthorized and deal with it in the caller. Much better. Thanks!

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: changed

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: done

lmuntanerDfinity:

frontend/svelte/src/lib/components/neuron-detail/NeuronHotkeysCard.svelte
```diff
@@ -26,10 +30,20 @@
       initiator: "remove-hotkey-neuron",
       neuronId: neuron.neuronId,
     });
-    await removeHotkey({
+    const currentIdentityPrincipal = (await getIdentity())
```
peterpeterparker, line 33: maybe just move this after the `removeHotkey`. in case of bad luck, the couple of seconds of the function might just be the time it needs until the identity becomes invalid

frontend/svelte/src/lib/components/neuron-detail/NeuronHotkeysCard.svelte
```diff
@@ -26,10 +30,20 @@
...
+      .getPrincipal()
+      .toText();
+    const maybeNeuronId = await removeHotkey({
       neuronId: neuron.neuronId,
       principalString: hotkey,
     });
+    if (currentIdentityPrincipal === hotkey && maybeNeuronId !== undefined) {
```
peterpeterparker, line 40: worth adding a small comment?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -542,17 +548,31 @@ export const removeHotkey = async ({
...
 
     await removeHotkeyApi({ neuronId, identity, principal });
+    removed = true;
 
     await getAndLoadNeuron(neuronId);
 
     return neuronId;
   } catch (err) {
+    if (removed && err instanceof NotAuthorizedError) {
+      // Checking the identity can be one of the reasons we got the error.
```
peterpeterparker, line 565: mmmh not sure here. `getIdentity` never throw. If no identity, it logs out

frontend/svelte/src/lib/components/neuron-detail/NeuronHotkeysCard.svelte
lmuntanerDfinity: sure

frontend/svelte/src/lib/components/neuron-detail/NeuronHotkeysCard.svelte
lmuntanerDfinity: yep

frontend/svelte/src/lib/services/neurons.services.ts
lmuntanerDfinity: true, it's `getIdentityOfControllerByNeuronId` that can throw and we don't want to check identity then.

lmuntanerDfinity:

peterpeterparker (approved):
LGTM, thx

PR #861, opened 2022-05-17
Read e2e env vars in the same way as svelte

bitdivine:
# Motivation
The e2e tests should read configuration parameters in the same way as svelte; the inconsistency can be jarring.

# Changes
- Copy the svelte env parsing code for use in e2e tests, and use it.
  - Note: I would love to factor out the common code but was unable to do so.  The e2e tests use typescript, svelte uses mjs and I was unable to get both working and passing the linter at the same time.  If someone can, I would be most thankful.  Why do we need mjs rather than ts?
- Strengthen the case when an env var is an invalid enum variant.

# Tests
Verified that e2e tests work.


config.ts
```diff
@@ -0,0 +1,52 @@
...
+  var value = process.env[key];
+  if (undefined === value || value === "") {
+    value = configFromFile[key];
+  }
+  if (undefined === value || value === "") {
+    throw new Error(`Environment variable '${key}' is undefined.`);
+  }
+  return value;
+}
+
```
bitdivine, line 25: Below is a new function, to handle the case where the configuration item should be an enum variant.

I would like some help here.  What should the type signature of `enumType` be?  Or can I call `Object.keys()` directly on `Type` without passing `enumType` as an argument?

e2e-tests/common/constants.ts
```diff
@@ -16,9 +18,8 @@ export enum RedirectToLegacy {
   both = "both",
 }
```
bitdivine, line 19: The config variables below are now read in the same way as svelte config vars.  Additionally, if an illegal value is passed to `REDIRECT_TO_LEGACY` an error is thrown rather than silently reverting to the default.

config.ts
lmuntanerDfinity: I can't find a better solution. But it seems the best at the moment:

https://stackoverflow.com/questions/17380845/how-do-i-convert-a-string-to-enum-in-typescript/41548441#41548441

From [the link](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings) there:

> Keep in mind that string enum members do not get a reverse mapping generated at all.

config.ts
```diff
@@ -0,0 +1,36 @@
...
+ * The configuration is use in the rollup build but also in the parser of the static files - e.g. build.index.mjs to output the index.html with a CSP.
+ */
+
+import * as configFromFile from "./frontend/ts/src/config.json";
+
+/**
+ * Returns the given environment variable, if defined and non-empty, else throws an error.
+ */
+export function getRequiredEnvVar(key) {
+  var value = process.env[key];
```
peterpeterparker, line 16: please avoid `var` (i guess this file is not covered by the eslint rule)

bitdivine: The core part makes sense to  me, thanks!  I am gradually getting used to `??`.  A part I don't get is: Why use a closure rather than a function?  A closure captures the lexical environment so that it can reference values there, so that state or even if empty, that state container is locked up for as long as the closure exists.  A function doesn't have such a lock, so is lighter weight and faster.  I realize that I started with HPC problems and so maybe I stress about CPU cycles and RAM usage more than maybe I should but to me the choice between closures and functions is:  Do I need the surrounding state?  Yes-> closure, no-> function.

peterpeterparker: As you wish. just don't use `var`, that's a bad practice in JS.

config.ts
bitdivine: Thank you for looking.  Interesting link with the implementation details.

config.ts
bitdivine: I agree with the var! The only place I expect to find it is in css.

lmuntanerDfinity (approved):
LGTM

PR #864, opened 2022-05-18
feat: instead of spinner, make buttons disabled if busy

peterpeterparker:
# Motivation

Instead of spinner replacing buttons when work is in progress, make buttons disabled if `$busy`. This to avoid two display two spinners on screen.

# Changes

- remove `loading` and `spinner`
- make buttons disabled if `$busy`

# Tests

No particular tests were provided because such use case was not tested before anyway.


lmuntanerDfinity (approved):
Perfect, thanks!

PR #867, opened 2022-05-18
refactor: use selected account context

peterpeterparker:
# Motivation

A dedicated context and store has been introduced for the selected account of the "Wallet -Account" details page. It can be used to avoid passing properties through components and to replace existing store information that duplicate the information.

# Changes

- use selected account context and store instead of passing props
- remove `selectedAccount` from hardware wallet neurons store and use selected account store instead


mstrasinskis (approved):
Thanks, very nice 👍

PR #868, opened 2022-05-18
feat: avoid transaction list flickering after new transaction

peterpeterparker:
# Motivation

When a user transfer ICP to another account fro the "Wallet" details page, the list of transaction has to be updated with the new transaction. To do so, we update the list when the accounts are sync again and per extension the selected account is updated - all good.

However when we do so we display a skeleton card first what has for effect to make the list flickers briefly.

This PR solve that particular UI effect.

# Changes

- preserve transaction history if account does not change before reloading the list of transactions


mstrasinskis (approved):
Good idea, thank you 👍

PR #869, opened 2022-05-18
button disabled hover state

wackyleo459:
# Motivation
Button when disabled should not be changing colors, or cursor change when hovered over.

- Added [disabled]:hover attribute that behaves the same as disabled when button is hovered over.
- Note: The primary buttons in the footer of pages such as `Merge Neurons`, `Stake Neurons`, `Add Transaction`, and `Add Account` are not disabled by default, but a modal pop-up will show instead -- blocking interaction with the buttons. 
- It may not be necessary to change these footer buttons to disabled, unless you want to guard against when page is somehow lagging and paused.

https://dfinity.atlassian.net/jira/software/c/projects/L2/boards/182?modal=detail&selectedIssue=L2-534&assignee=617f8802e79ff6006ffac7cc

### DIFF
--- a/frontend/svelte/src/lib/themes/button.scss
+++ b/frontend/svelte/src/lib/themes/button.scss
@@ -2,7 +2,8 @@
 @use "mixins/media";
 
 button {
-  &[disabled] {
+  &[disabled],
+  &[disabled]:hover {
     pointer-events: none;
   }
 
@@ -65,7 +66,8 @@ button {
 
   &.primary,
   &.input-button {
-    &[disabled] {
+    &[disabled],
+    &[disabled]:hover {
       background: var(--gray-400);
       color: var(--gray-600);
     }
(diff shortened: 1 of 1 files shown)


mstrasinskis (changes requested):
Thank you very much, looks much better 👍

Somehow, by clicking in some places I managed to trigger the animation (see the video below. MacOS/Chrome 101.0.4951.64). Also it would be cool to use the default cursor (not pointer). If it's reproducible could you please try to disable the animation? 

We don't need to support some old browsers, so `pointer-event: none;` in theory could be helpful here.

https://user-images.githubusercontent.com/98811342/169152156-bbedfdd6-adcf-4b65-8d61-e5851a0028fd.mov

frontend/svelte/src/lib/themes/button.scss
```diff
@@ -2,7 +2,8 @@
 @use "mixins/media";
 
 button {
-  &[disabled] {
+  &[disabled],
+  &[disabled]:hover {
```
mstrasinskis, line 6: I wonder if with the current solution `:hover` is still needed 🤔

mstrasinskis (approved):
LGTM. Thanks a lot, works as expected 👍

PR #874, opened 2022-05-19
Use setup in CI

bitdivine:
# Motivation
We now have a setup script that installs all dependencies for a developer.  It would be nice to run it in CI to provide some assurance that it keeps working.

Using the setup script also reduces the risk of developer and CI environments diverging.

# Changes
- Use the setup script rather than installing rust and flutter independently in CI.

# Tests
See CI!

Note:
- CI passes
- The native build is no slower than before, if anything it is quicker at 8 minutes rather than the usual 11.  Not sure why.  It could be an outlier.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,9 +91,12 @@ jobs:
 
       - name: Install Software
         run: |
+          set -x
           bash -x ./scripts/setup --profile ~/.bashrc
           . ~/.bashrc
-          echo "$PATH" | tr : "\n" >> $GITHUB_PATH
+          tail ~/.bashrc
+          command -v flutter
+          echo "$PATH" | tr : "\n" | tee /dev/stderr >> $GITHUB_PATH
 
       # Helps with debugging
       - name: Versions
@@ -105,6 +108,8 @@ jobs:
           rustc --version
           cargo --version
           ic-cdk-optimizer --version
+          tail ~/.bashrc
+          command -v flutter
 
       - name: Start replica
         run: |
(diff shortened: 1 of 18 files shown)


scripts/setup
```diff
@@ -149,14 +150,22 @@ install_flutter() {
   Darwin)
     FLUTTER_ZIP="flutter_macos_${FLUTTER_VERSION}-stable.zip"
     FLUTTER_URL="https://storage.googleapis.com/flutter_infra_release/releases/stable/macos/${FLUTTER_ZIP}"
```
bitdivine, line 152: Fix for the flutter installation:

The Mac file is .zip, the Linux is .tar.xz.  unzip doesn't work on the latter.  The Mac code had been tested on a blank machine, the Linux code skipped this step because flutter was already installed.

.github/workflows/build.yml
```diff
@@ -82,46 +82,18 @@ jobs:
...
-          rustup target add wasm32-unknown-unknown
-
       # Cache the ic-cdk-optimizer
       - uses: actions/cache@v2
         with:
           path: |
-            ~/.local/bin/ic-cdk-optimizer
-            target
-          key: ${{ runner.os }}-ic-cdk-optimizer-${{ env.IC_CDK_OPTIMIZER_VERSION }}
+            ~/.cargo/bin/ic-cdk-optimizer
```
lmuntaner, line 89: this is not in the `setup` script?

lmuntaner (approved):
LGTM, just one question

.github/workflows/build.yml
bitdivine: Yes it is, but building ic-optimizer is slow so it is worth caching the result.  Setup will install tools not already installed, so we can cache any parts that are slow to install.

PR #875, opened 2022-05-19
Add cycles minting canister ID

bitdivine:
# Motivation
We need to reference the cycles minting canister sometimes, so we need to add the canister ID to the config.

# Changes
* Add the cmc canister ID, using the same name used in the main IC repo.

# Tests
This should cause no changes as it is as yet unused.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -87,6 +87,7 @@
   "defaults": {
     "network": {
       "config": {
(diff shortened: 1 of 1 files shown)


lmuntanerDfinity (approved):
Thanks!

PR #878, opened 2022-05-19
Install ic-admin

bitdivine:
# Motivation
`ic-admin` is not installed by the setup script but it should be, as it is used for deployments and creating proposals on testnets.

# Changes
- Add ic-admin

# Tests
- Tested installation on:
  - An M1 Mac
  - An Ubuntu 20.04 Linux machine.


peterpeterparker (approved):
Rubber stamp

PR #880, opened 2022-05-20
L2-479: Validate account id client

lmuntaner:
# Motivation

User sees an invalid account error on blur.

# Changes

* Use same nns-js validation on blur for the Address component.

# Tests

* Update broken tests.
* Add new test for new use case of error appearing on blur.

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Input.svelte
+++ b/frontend/svelte/src/lib/components/ui/Input.svelte
@@ -16,6 +16,7 @@
   export let placeholderLabelKey: string;
 
   export let theme: "dark" | "light" = "light";
+  export let withErrorMessage: boolean = false;
 
   const handleInput = ({ currentTarget }: InputEventHandler) =>
     (value =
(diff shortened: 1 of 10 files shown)


lmuntaner:
UX video:

https://user-images.githubusercontent.com/4550653/169457581-3c209a19-cdac-4073-986f-d7ba631d9619.mov

lmuntaner:

I know that @peterpeterparker was not a fan of the layout shift, but with Mischa we agreed that this was the best bad solution for now.

bitdivine (approved):
LGTM, thanks!

frontend/svelte/src/lib/components/ui/Input.svelte
```diff
@@ -28,7 +29,7 @@
   $: placeholder = translate({ labelKey: placeholderLabelKey });
 </script>
 
-<div class={`input-block ${theme}`} class:disabled>
+<div class={`input-block ${theme} `} class:disabled class:withErrorMessage>
```
peterpeterparker, line 32: @lmuntaner can you style it using `:global` selector in `InputWithError`?

if not possible, please do not use a property for styling purpose only but rather a style, like CSS variable

lmuntaner: ok

peterpeterparker: https://github.com/dfinity/nns-dapp/pull/891

PR #883, opened 2022-05-20
e2e: USER_A1 Verify that main account exists

bitdivine:
# Motivation
We need e2e tests for all user stories, as described in [the manual tests](https://www.notion.so/NNS-Dapp-Manual-Testing-Accounts-87b7b439dd6a4f94b1459916f6b6a09c#e39e3f789730417e9a68eda88d16c6ad).  In particular, test USER_A1 requires that new users automatically get a main account.

# Changes
- Added an e2e component for the accounts tab.

# Tests

### DIFF
--- a/e2e-tests/specs/user-A01-new-user-gets-main-account.e2e.ts
+++ b/e2e-tests/specs/user-A01-new-user-gets-main-account.e2e.ts
@@ -5,12 +5,10 @@ import { register } from "../common/register";
 import { MyNavigator } from "../common/navigator";
 import { AccountsTab } from "../components/accounts-tab";
 
-describe("Makes Standard Users", () => {
-  /**
-   * Creates a user with:
-   * - ICP in the main account
-   * - A neuron
-   */
+/**
+ * Veifies that new users get a main account.
+ */
+describe("Users get a main account", () => {
   it("Setup: Create a new user", async () => {
     await browser.url("/");
     const userId = await register(browser);
--- a/e2e-tests/specs/user-A01-new-user-gets-main-account.e2e.ts
+++ b/e2e-tests/specs/user-A01-new-user-gets-main-account.e2e.ts
@@ -25,6 +25,10 @@ describe("Users get a main account", () => {
   });
 
   it("User should have a main account", async () => {
+    // Note: getAccountByName() returns a promise.  If the main
+    // account cannot be found within a reasaonable amount
+    // of time, the promise will fail and the await will
+    // fail the test.
     await new AccountsTab(browser).getAccountByName(
       "Main",
       "Checking that the main account exists"
--- a/e2e-tests/specs/user-A01-new-user-gets-main-account.e2e.ts
+++ b/e2e-tests/specs/user-A01-new-user-gets-main-account.e2e.ts
@@ -27,6 +27,9 @@ describe("Makes Standard Users", () => {
   });
 
   it("User should have a main account", async () => {
-    await new AccountsTab(browser).getAccountByName("Main", "Checking that the main account exists");
+    await new AccountsTab(browser).getAccountByName(
+      "Main",
+      "Checking that the main account exists"
+    );
   });
 });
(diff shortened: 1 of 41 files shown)


frontend/svelte/src/lib/components/accounts/AccountCard.svelte
```diff
@@ -9,14 +9,15 @@
   export let account: Account;
   export let showCopy: boolean = false;
   export let role: "button" | "link" | undefined = undefined;
+  export let testId: string = "accountCard";
```
lmuntaner, line 12: Why do you add it as prop?

lmuntaner: And I think that the convention was to use kebab-case, right?

e2e-tests/specs/user-A01-new-user-gets-main-account.e2e.ts
```diff
@@ -0,0 +1,33 @@
...
+    // Verify that we are on the accounts tab.
+    const navigator = new MyNavigator(browser);
+    await navigator.getElement(
+      AccountsTab.SELECTOR,
+      "Wait for the accounts tab"
+    );
+  });
+
+  it("User should have a main account", async () => {
+    await new AccountsTab(browser).getAccountByName(
```
lmuntaner, line 28: It feels weird to have a test that doesn't `expect` anything.

Also in the previous one.

If I read this, it seems that `getAccountByName` could return `undefined` and the test would still pass.

lmuntaner: Maybe renaming or checking the return?

```
const element = await new AccountsTab(browser).getAccountByName(...)
expect(element).toBeDefined();
```

frontend/svelte/src/lib/components/accounts/AccountCard.svelte
bitdivine: `AccountCard` is a special case of `Card`.  `Card` has `testId` that has the default value `card` but that can be set for a more specific value.  For `AccountCard` a better default is `account-card`, but of course it might be useful to override that in a yet more specific case.

but happy to change it to kebab case!

frontend/svelte/src/lib/components/accounts/AccountCard.svelte
```diff
@@ -9,14 +9,15 @@
...
 
   let identifier: string;
   let balance: ICPType;
 
   $: ({ identifier, balance } = account);
 </script>
 
-<Card on:click {role}>
+<Card on:click {role} {testId}>
```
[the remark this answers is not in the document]
bitdivine: I guess we can generalize if needed when the time arises.

e2e-tests/specs/user-A01-new-user-gets-main-account.e2e.ts
bitdivine: `getAccountByName` will fail if it can't find the main account, so `expect(..).toBeDefined()` will always be true.  If it were false, the promise on the previous line woul dhave failed and so the await there would have thrown an error, so the expect would never run.

bitdivine: I'll add a comment to that effect.

frontend/svelte/src/lib/components/accounts/AccountCard.svelte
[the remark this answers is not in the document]
lmuntaner: If it was intended it's ok. It was just a typo.

frontend/svelte/src/lib/components/accounts/AccountCard.svelte
lmuntaner: ok, it makes sense. But let's stick to kebab case for the values, no?

lmuntaner (approved):
LGTM

bitdivine:
Thank you for the review, @lmuntaner !

PR #886, opened 2022-05-20
Fix: Propose can load neuron

lmuntaner:
# Motivation

User can see the proposer modal

# Changes

* `loadNeuron` does not check if the user is authorized or not to see the neuron.

# Tests

* Test for bug.


mstrasinskis (approved):
LGTM. Thank you 👍

PR #887, opened 2022-05-21
E2e tidy

bitdivine:
# Motivation
Keeping the e2e code in good shape.

# Changes
- Added the e2e linter to CI.  This did not cause any errors but it would catch errors if a contributor was careless or unaware of the linter.
- Added a rule to require function signatures, and added missing signatures.
- Minor rearrangement of code.

# Tests
I have carefully resisted making any changes to the the code, to avoid having lint and code changes in the same PR, so this PR should change nothing in the way the e2e tests are run.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -208,6 +208,32 @@ jobs:
       - name: Run linter
         run: npm run check
 
+  e2e-lint:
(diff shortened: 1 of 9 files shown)


lmuntaner (approved):
LGTM

PR #888, opened 2022-05-21
E2e Accounts: Can create linked account

bitdivine:
# Motivation
We need end to end tests for all the manual tests.  This PR addresses [User_A02 Create new linked account](https://www.notion.so/NNS-Dapp-Manual-Testing-Accounts-87b7b439dd6a4f94b1459916f6b6a09c#028295f8c51e4c8fac692371602c4bc4)

# Changes
- Fix the method for getting an account - wdio selectors don't work like that.
- Add a method to the accounts tab for creating a linked account
- Add a test to verify that the account appears

# Tests
This is a test - see CI for the results! :-)

### DIFF
--- a/e2e-tests/components/accounts-tab.ts
+++ b/e2e-tests/components/accounts-tab.ts
@@ -20,11 +20,14 @@ export class AccountsTab extends MyNavigator {
     description: string,
     options?: { timeout?: number }
   ): Promise<WebdriverIO.Element> {
-    return this.getElement(
-      `${AccountsTab.SELECTOR} [data-tid="account-card"] [data-account-name="${name}"]`,
-      `Getting account '${name}'`,
-      options
+    const accountNameWithEscapedSingleQuotes = name.replace(/'/g, "\\'");
+    const element = await this.browser.$(
+      `//*[@data-tid = 'account-card' and .//*[@data-tid="account-name" and text() = '${accountNameWithEscapedSingleQuotes}']]`
     );
+    const timeout = options?.timeout ?? 5_000;
+    const timeoutMsg = `Timeout after ${timeout.toLocaleString()}ms waiting for "${description}" with account "${name}".`;
+    await element.waitForExist({ timeout, timeoutMsg });
+    return element;
   }
 
   /**
--- a/e2e-tests/components/accounts-tab.ts
+++ b/e2e-tests/components/accounts-tab.ts
@@ -8,10 +8,6 @@ export class AccountsTab extends MyNavigator {
   static readonly LINKED_ACCOUNT_NAME_SELECTOR: string = `#modalContent [data-tid="input-ui-element"]`;
   static readonly CREATE_LINKED_ACCOUNT_BUTTON_SELECTOR: string = `#modalContent [type="submit"]`;
 
-  constructor(browser: WebdriverIO.Browser) {
-    super(browser);
-  }
-
   /**
    * Creates a browser controller for the accounts tab.
    */
--- a/e2e-tests/components/accounts-tab.ts
+++ b/e2e-tests/components/accounts-tab.ts
@@ -36,36 +36,29 @@ export class AccountsTab extends MyNavigator {
       AccountsTab.SELECTOR,
       "Prerequisite: Verify that we are on the acconts tab"
     );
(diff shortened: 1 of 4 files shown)


e2e-tests/components/accounts-tab.ts
```diff
@@ -2,25 +2,70 @@ import { MyNavigator } from "../common/navigator";
...
    */
+  constructor(browser: WebdriverIO.Browser) {
+    super(browser);
+  }
+
   async getAccountByName(
     name: string,
     description: string,
     options?: { timeout?: number }
-  ) {
```
bitdivine, line 13: It turns out that the first wdio selector returns only the first card.  The text selector cannot be added to the first selector, so as it is there is no clean way of getting wdio to select a specific account.

Adding more data attributes to the dom resolves this problem, and makes this function into a special case of the existing getElement.

frontend/svelte/src/lib/components/accounts/AccountCard.svelte
```diff
@@ -13,12 +13,17 @@
...
 </script>
 
 <Card on:click {role} testId="account-card">
-  <div slot="start" class="title">
-    <h3><slot /></h3>
+  <div
+    slot="start"
+    class="title"
+    data-account-identifier={identifier}
+    data-account-name={name ?? "Main"}
```
bitdivine, line 24: Without the name and identifier as attributes, there is no clean way for wdio to select these elements, hence the addition.

Note: as this is used for e2e tests, it is preferable not to use the internationalized version of "Main" here.

peterpeterparker: woot? we cannot perform CSS queries with  wdio?

```
document.querySelector("[data-tid='account-card'] [slot='start']");
```

bitdivine: That won't get you a particular card.

peterpeterparker: It's an example, there are other selector like getting first card with `nth-child(1)`. Alternatively you can `querySelectorAll` and filter if the ordering is not relevant. Just spontaneous ideas.

If you want we can give a try together next week?

But if you really have no other choice, then I would definitely stick to a single `data-tid` attributes as identifier for testing purpose and avoid introducing more variations of attributes naming (`data-*`) that have testing purposes. This to avoid confusion and because our all suite of tests is based on and can be configured with a single attribute identifier for tests.

bitdivine: Let's discuss next week.

wdio's equivalent to querySelector is `browser.$(someString)` where someString is either a css query or a text content query but not both.  It is possible to compose them, but the composition is implemented badly, to the point that I consider it useless; its bugs cost me 2 hours already.  It is possible to run arbitrary js inside the browser and get the result, so normal `querySelectorAll(..)` followed by `filter(..)` could be used there, but that doesn't return a wdio element, so is incompatible with the rest of the test suite.  I am familiar with n-th child, but if looking for an account by name, the code would be to try getting n-th child for reach successive N, which is slow and unclean.  But we can go over it in person.  That is probably quicker.

Alternatively you might want to have a shot at implementing `getAccountByName(...): Promise<WebdriverIO.Element>` with the DOM as it is.

bitdivine: Another idea -  if we restrict ourselves to data-tid, this could be done with `data-tid="account-card-${account_id}-${account_name}"` and for matching on account ID do `[data-tid^="account-card-${account_d}-"]` and for account name match with `[data-tid^="account-card-"][data-tid$="-${account_name}"]` however the latter is buggy as it could not distinguish between `account_name="xx-yyy"` and `account_name="yyy"`.  Unless there are some forbidden characters for account names that we could use as the separator this would only ever be buggy.  Still, it was a space to explore.

bitdivine: Apropos forbidden characters, " " == space is a legal account name, as are two or three spaces.  I think maybe we should trim names.

peterpeterparker (approved):
LGTM, thanks

PR #890, opened 2022-05-22
feat: input type icp

mstrasinskis:
# Motivation

To simplify ICP amount entering the input field type=number was replaced with type=icp. 
Mock Flutter behaviour:
- use dot as separator regardless of locale
- max 8 digits after dot
- no exponential format support

# Changes

inpuType="icp"

# Tests

- Input.spec.ts

# Screenshots

## Editing

https://user-images.githubusercontent.com/98811342/169716766-9db1e1a6-7722-4aba-b8bf-5476ce71e4b1.mov

## Copy Paste

https://user-images.githubusercontent.com/98811342/169716771-76abe3cd-f3fd-48bc-bdab-e4f92ffdd7d9.mov

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Input.svelte
+++ b/frontend/svelte/src/lib/components/ui/Input.svelte
@@ -52,7 +52,12 @@
     }
 
     internalValueChange = true;
-    value = lastValidICPValue;
+    value =
+      lastValidICPValue === undefined
+        ? undefined
+        : typeof lastValidICPValue === "number"
+        ? lastValidICPValue.toFixed(8)
+        : +lastValidICPValue;
     icpValue = fixUndefinedValue(lastValidICPValue);
 
     // force dom update (because no active triggers)
@@ -67,7 +72,7 @@
       const currentValue = currentTarget.value;
 
       // handle invalid input
-      if (!isValidICPFormat(currentValue)) {
+      if (isValidICPFormat(currentValue) === false) {
         // restore value (e.g. to fix invalid paste)
         restoreFromValidValue();
         return;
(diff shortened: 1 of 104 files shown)


frontend/svelte/src/lib/components/ui/Input.svelte
```diff
@@ -1,7 +1,14 @@
 <script lang="ts">
   import { translate } from "../../utils/i18n.utils";
+
+  const isValidICP = (text: string) => /^[\d]*(\.[\d]{0,8})?$/.test(text);
```
lmuntaner, line 4: Maybe move this to `icp.utils.ts`?

frontend/svelte/src/lib/components/ui/Input.svelte
```diff
@@ -32,19 +105,21 @@
...
-    type={inputType}
+    type={inputType === "icp" ? "text" : inputType}
+    bind:this={inputElement}
     {required}
     {spellcheck}
     {name}
     {step}
     {disabled}
-    {value}
+    value={inputType === "icp" ? icpValue : value}
```
lmuntaner, line 115: Does that mean that when it's `icp`, the value is not bound here?

Because when we use an `Input` we do: `bind:value={inputValue}`

lmuntaner (commented):
A question.

Did yuo try it in different use cases to check that it was working?

peterpeterparker:
Works great with my Swiss-french keyboard 👍

Can you maybe add some tests in `Input.spec`?

frontend/svelte/src/lib/components/ui/Input.svelte
mstrasinskis: The binding is handled manually. Tests for binding were added.

lmuntaner (approved):

PR #891, opened 2022-05-23
Fix: Style variable instead of prop for Input

lmuntaner:
# Motivation

Refactor `Input` and remove the prop `withErrorMessage`.

# Changes

* Use a css variable instead of a prop to change Input style.

# Tests

No test changes.

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Input.svelte
+++ b/frontend/svelte/src/lib/components/ui/Input.svelte
@@ -16,7 +16,6 @@
   export let placeholderLabelKey: string;
 
   export let theme: "dark" | "light" = "light";
-  export let withErrorMessage: boolean = false;
 
   const handleInput = ({ currentTarget }: InputEventHandler) =>
     (value =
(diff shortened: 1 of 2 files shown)


lmuntaner:
@peterpeterparker PR removing Input prop `withErrorMessage`

bitdivine (approved):

frontend/svelte/src/lib/components/ui/Input.svelte
```diff
@@ -60,7 +59,10 @@
   .input-block {
     position: relative;
 
-    margin: var(--padding-2x) 0;
+    margin-top: var(--padding-2x);
```
peterpeterparker, line 62: adding only the new property `margin: var(--padding-2x) 0 var(--input-margin-bottom, var(--padding-2x));` would work too right? just for my know-how, splitting is fine

peterpeterparker (approved):
Thanks

frontend/svelte/src/lib/components/ui/Input.svelte
lmuntaner: Yes, it would work, but I found that harder to read.

PR #893, opened 2022-05-23
e2e: Test proposal filters

bitdivine:
# Motivation
Every user flow, as described by a manual test, needs to be implemented as an e2e test.  In particular, it is required that the mutiple-choice filters work, [in USER_V02](https://www.notion.so/NNS-Dapp-Manual-Testing-Voting-Tab-cbdb93bf29264a56ac25b2bf6d6bb992#eb80c871060846b29e5e5f2d04a97a89). 

# Changes
- Add an e2e test for filtering with the multiple-choice modals.

Note: This filter does NOT cover the filter for eligible proposals.

# Tests
See CI! :-)


e2e-tests/components/proposals-tab.ts
```diff
@@ -50,6 +51,89 @@ export class ProposalsTab extends MyNavigator {
...
+    enable: boolean = true
+  ): Promise<void> {
+    // The default window is too small, so some checkboxes and buttons are not shown in screenshots.
+    const { width, height } = await browser.getWindowSize();
+    await browser.setWindowSize(800, 1000);
+
+    // Open filter modal:
+    await this.click(
+      `[data-tid="${filterTid}"]`,
+      `Open filter modal for ${filterTid}`
```
peterpeterparker, line 72: not related to your PR: should we "hardcode" expected texts or should we use the i18n texts of the app?

```
await this.click(
      `[data-tid="${filterTid}"]`,
      `Open filter modal for ${filterTid}`
      
vs somethine like

await this.click(
      `[data-tid="${filterTid}"]`,
      translate(i18n.key.whatever, values)
```

e2e-tests/components/proposals-tab.ts
```diff
@@ -50,6 +51,89 @@ export class ProposalsTab extends MyNavigator {
...
+        );
+      },
+      {
+        timeoutMsg: `Waiting for the number of filter elements to be correct.`,
+        timeout: 5_000,
+      }
+    );
+
+    // Restore the former window size.
+    await browser.setWindowSize(width, height);
```
peterpeterparker, line 134: out of curiosity: why do you have to set and reset the window size within the test?

additional question: unlike in jest, there are no lifecycles function such as `beforeEach` or `afterAll`?

e2e-tests/specs/user-V02-proposal-filters-work.e2e.ts
```diff
@@ -0,0 +1,87 @@
...
+        proposalMetadataSelector,
+        "Proposal should appear, if it was not already displayed"
+      );
+      const disappears = await proposalsTab
+        .waitForGone(proposalMetadataSelector, "Seeing if it disappears", {
+          timeout: 1_000,
+        })
+        .then(() => true)
+        .catch(() => false);
+      expect(disappears).toBe(false);
```
peterpeterparker, line 58: you can probably also test that it does not throw?

peterpeterparker (approved):
LGTM, thanks

e2e-tests/components/proposals-tab.ts
bitdivine: I thought about that.  Normally I would say that the test code should not be sharing any code with the thing it is testing, otherwise a bug in the implementation could break the test as well.  One could argue that i18n is so simple that depending on that is OK.

Another approach is not to check text anywhere in the e2e tests and rely only on tids.

In some of my projects, the equivalent of i18n returns HTML (typically a span, maybe always, I will have to check) so that e.g. in some languages where some text is especially long, the text can be smaller or break at specific points.  Where I do that, I also have attributes in the HTML that can help with automated testing.  E.g. there could be an attribute which is the i18n key, so the test can just check that there is a span with the expected key.

For now, my focus is on getting as many e2e tests out as possible, so I am doing the simple thing:  Assuming fixed English.  Happy to refine that when we have more time.

e2e-tests/components/proposals-tab.ts
bitdivine: This is a function that could be used in various e2e tests.  I don't want to assume that the calling test is OK with this new page size.  So far, all resizes have been to make the page long but maybe that won't always be the case.

e2e-tests/specs/user-V02-proposal-filters-work.e2e.ts
bitdivine: I want it to throw! :-)  If it doesn't this test should fail.

e2e-tests/components/proposals-tab.ts
peterpeterparker: Thanks for the explanation. Makes sense. We can maybe use the i18n translation from the app the day we tested everything with screenshots - i.e. screenshot = hardcoded values, test a bit more dynamic. That way we might have both. Just a spontaneous idea.

e2e-tests/specs/user-V02-proposal-filters-work.e2e.ts
peterpeterparker: I meant something like https://stackoverflow.com/a/71768359/5404186

```
await expect(
  foo(params),
).resolves.not.toThrowError();
```

just in case you want to spare the boolean conversion. not sure it works, again just a comment

e2e-tests/components/proposals-tab.ts
peterpeterparker: thanks for the feedback, alright let's see then

PR #895, opened 2022-05-24
L2-608: Fix neuron stake update after transaction

lmuntaner:
# Motivation

User sees always new neuron stake after transaction.

# Changes

* "loadNeuron" checks balance when fetching a single neuron, same as when fetching the list of neurons.
* Check in "loadNeuron" can be skipped with `skipCheck` flag.
* "loadNeuron" also accepts the query strategy as param.
* Add a callback in the Transaction modal to be called when transaction is finished.

# Tests

* Test that the new callback in Transaction modal is called.
* Test new flags in "loadNeuron"

### DIFF
--- a/frontend/svelte/src/lib/components/accounts/NewTransactionReview.svelte
+++ b/frontend/svelte/src/lib/components/accounts/NewTransactionReview.svelte
@@ -9,6 +9,10 @@
   import { busy, startBusy, stopBusy } from "../../stores/busy.store";
   import { transferICP } from "../../services/accounts.services";
   import { isAccountHardwareWallet } from "../../utils/accounts.utils";
+  import { toastsStore } from "../../stores/toasts.store";
+
+  export let onTransactionComplete: (() => Promise<void>) | undefined =
+    undefined;
 
   const context: TransactionContext = getContext<TransactionContext>(
     NEW_TRANSACTION_CONTEXT_KEY
@@ -23,6 +27,11 @@
     startBusy({ initiator: "accounts" });
 
     const { success } = await transferICP($store);
+
+    if (success) {
+      await onTransactionComplete?.();
+      toastsStore.success({ labelKey: "accounts.transaction_success" });
+    }
 
     stopBusy("accounts");
 
--- a/frontend/svelte/src/lib/components/accounts/NewTransactionReview.svelte
+++ b/frontend/svelte/src/lib/components/accounts/NewTransactionReview.svelte
@@ -11,13 +11,10 @@
   import { isAccountHardwareWallet } from "../../utils/accounts.utils";
   import { toastsStore } from "../../stores/toasts.store";
 
-  export let onTransactionComplete: (() => Promise<void>) | undefined =
-    undefined;
-
   const context: TransactionContext = getContext<TransactionContext>(
     NEW_TRANSACTION_CONTEXT_KEY
   );
-  const { store }: TransactionContext = context;
+  const { store, onTransactionComplete }: TransactionContext = context;
 
   let amount: ICPType = $store.amount ?? ICPType.fromE8s(BigInt(0));
 
--- a/frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
+++ b/frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
@@ -18,7 +18,6 @@
           resolve();
         },
         handleError: () => {
-          console.log("in da error");
           resolve();
         },
       });
--- a/frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
+++ b/frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
@@ -1,27 +1,11 @@
 <script lang="ts">
   import type { NeuronInfo } from "@dfinity/nns";
-  import { loadNeuron } from "../../services/neurons.services";
-  import { neuronsStore } from "../../stores/neurons.store";
+  import { reloadNeuron } from "../../services/neurons.services";
   import NewTransactionModal from "../accounts/NewTransactionModal.svelte";
 
   export let neuron: NeuronInfo;
 
-  // Not resolve until the neuron has been loaded
-  const fetchUpdatedNeuron = () =>
-    new Promise<void>((resolve) => {
-      loadNeuron({
-        neuronId: neuron.neuronId,
-        forceFetch: true,
-        strategy: "update",
-        setNeuron: ({ neuron, certified }) => {
-          neuronsStore.pushNeurons({ neurons: [neuron], certified });
-          resolve();
-        },
-        handleError: () => {
-          resolve();
-        },
-      });
-    });
+  const fetchUpdatedNeuron = () => reloadNeuron(neuron);
 </script>
 
 <NewTransactionModal
--- a/frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
+++ b/frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
@@ -5,7 +5,7 @@
 
   export let neuron: NeuronInfo;
 
-  const fetchUpdatedNeuron = () => reloadNeuron(neuron);
+  const fetchUpdatedNeuron = () => reloadNeuron(neuron.neuronId);
 </script>
 
 <NewTransactionModal
--- a/frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
+++ b/frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-  import { createEventDispatcher } from "svelte";
   import type { NeuronInfo } from "@dfinity/nns";
   import { loadNeuron } from "../../services/neurons.services";
   import { neuronsStore } from "../../stores/neurons.store";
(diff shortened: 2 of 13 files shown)


lmuntaner:

There was an issue sometimes where after a transaction, the balance of the neuron was not always up to date. It was not always an issue since there is some kind background update that can take longer or not.

I realized that we also need to check the balance when loading only one, the same as we do with the list. And we need to do that specially after increasing the stake of a neuron.

frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
```diff
@@ -1,26 +1,32 @@
...
-  const dispatcher = createEventDispatcher();
-  const fetchUpdatedNeuron = async () => {
-    await loadNeuron({
-      neuronId: neuron.neuronId,
-      forceFetch: true,
-      setNeuron: ({ neuron, certified }) => {
-        neuronsStore.pushNeurons({ neurons: [neuron], certified });
-      },
+  // Not resolve until the neuron has been loaded
+  const fetchUpdatedNeuron = () =>
```
peterpeterparker, line 10: shouldn't this finds place in a service? like `reloadNeuron` or something?

frontend/svelte/src/lib/components/accounts/NewTransactionReview.svelte
```diff
@@ -24,6 +25,11 @@
 
     const { success } = await transferICP($store);
 
+    if (success) {
```
peterpeterparker, line 28: any particular reason why you pass a callback instead of dispatching an event?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -248,7 +248,10 @@ export const listNeurons = async ({
       // Query the ledger for each neuron
       // refresh those whose stake does not match their ledger balance.
       try {
-        await checkNeuronBalances(neurons);
+        const refetch = await checkNeuronBalances(neurons);
+        if (refetch) {
```
peterpeterparker, line 252: why not implementing this logic - refetch -> listNeurons - in the service?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -862,13 +865,19 @@ export const removeFollowee = async ({
 export const loadNeuron = ({
   neuronId,
   forceFetch = false,
+  // Check also the Neuro stake when loading only one.
```
peterpeterparker, line 868: typo "Neuro"

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -887,19 +896,37 @@ export const loadNeuron = ({
...
-
-      setNeuron({ neuron, certified });
+      if (!certified || skipCheck) {
+        setNeuron({ neuron, certified });
+        return;
+      }
+      // If the check is required, we don't want to call `setNeuron` until it has finished.
+      // For example: to avoid closing the "IncreaseStakeNeuronModal" before we finish this check.
+      const refetch = await checkNeuronBalances([neuron]);
+      if (refetch) {
```
peterpeterparker, line 918: I don't get it here. So we loaded a neuron, then we do a check, then we load it again?

Isn't there a risk to end up in an endless loop?

frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
```diff
@@ -1,26 +1,32 @@
...
+      loadNeuron({
+        neuronId: neuron.neuronId,
+        forceFetch: true,
+        strategy: "update",
+        setNeuron: ({ neuron, certified }) => {
+          neuronsStore.pushNeurons({ neurons: [neuron], certified });
+          resolve();
+        },
+        handleError: () => {
+          console.log("in da error");
```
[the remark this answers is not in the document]
lmuntaner: haha, sorry, I'll remove it. I need to resolve here as well.

frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
lmuntaner: Which service? The transferICP?

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: to be able to reuse the `checkNeuronBalances` more easily.

In one service, the call is to `listNeurons`, in the other is `loadNeuron`. It was easier to return a boolean whether the checks were done or not.

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: thanks!

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: This is the same pattern as `listNeurons` and the same pattern as in flutter.

There is no risk of infinite loop, because we only get here in the "update" call, and in the next call it's a "query" call.

I can add a comment to make it clearer.

frontend/svelte/src/lib/services/neurons.services.ts

lmuntaner:

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: Comment added

frontend/svelte/src/lib/components/accounts/NewTransactionReview.svelte
lmuntaner: sure

peterpeterparker: Yes same conclusion as you, the busy spinner makes the use of callback mandatory in that use case

frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
peterpeterparker: `neurons.service`

lmuntaner: This is the problem, the service used is the `transferICPs` from accounts.service. Fetching a neuron from there looks ugly.

frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
```diff
@@ -1,26 +1,31 @@
...
-      },
+  // Not resolve until the neuron has been loaded
+  const fetchUpdatedNeuron = () =>
+    new Promise<void>((resolve) => {
+      loadNeuron({
+        neuronId: neuron.neuronId,
+        forceFetch: true,
+        strategy: "update",
+        setNeuron: ({ neuron, certified }) => {
+          neuronsStore.pushNeurons({ neurons: [neuron], certified });
```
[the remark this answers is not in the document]
lmuntaner: It pushes, but it removes the previous one.

We don't rely on the order of the array to render. We use the array with a specific sort.

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: like my other question, there isn't a risk of endless loop here too?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -347,15 +350,15 @@ const claimNeurons =
       neuronIds.map((neuronId) => claimOrRefreshNeuron({ identity, neuronId }))
     );
 
-const checkNeuronBalances = async (neurons: NeuronInfo[]): Promise<void> => {
+const checkNeuronBalances = async (neurons: NeuronInfo[]): Promise<boolean> => {
```
peterpeterparker, line 353: `checkNeuronBalances = true or false`...what means true, what means false?

can you add some doc or use a more descriptive answer?

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: Thanks for the answer, trust you but also, honestly, feels like it's the kind of stuffs that will break someday in production.

implementing another pattern that does not implies having `loadNeuron` within `loadNeuron` isn't possible?

frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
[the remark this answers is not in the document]
peterpeterparker: definitely not related to your PR but we should refactor the name and update the documentation in the store as well (it still says it appends elements) later on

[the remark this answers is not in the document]
lmuntaner: ok

frontend/svelte/src/lib/modals/neurons/IncreaseNeuronStakeModal.svelte
peterpeterparker: I mean this + test

Capture d’écran 2022-05-25 à 15 03 18

Capture d’écran 2022-05-25 à 15 03 15

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: we pass `skipCheck: true` if we refetch, so no.

frontend/svelte/src/lib/services/neurons.services.ts

lmuntaner:

peterpeterparker:
Thanks a lot for the improvements and explanation!

Code wise looks good but I don't feel personally confident to accept the PR on my own because of the two points discussed above (the two functions calling themselves). Pattern-wise and use-case-wise, it feels to me - personal feeling, can be fully wrong - like the kind of stuffs that will break someday or will have a bad impact on performance.

So I would be glad if @mstrasinskis review also the PR. If it works for him, then shall be it but would be happy if he also gives a feedback.

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -912,6 +941,23 @@ export const loadNeuron = ({
   });
 };
 
+// Not resolve until the neuron has been loaded
+export const reloadNeuron = (neuron) =>
```
mstrasinskis, line 945: nth: neuronId to not confuse with the neuron from setNeuron callback?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -912,6 +941,23 @@ export const loadNeuron = ({
...
+  new Promise<void>((resolve) => {
+    loadNeuron({
+      neuronId: neuron.neuronId,
+      forceFetch: true,
+      strategy: "update",
```
mstrasinskis, line 950: Probably `skipCheck: true,` to emphasise this.

mstrasinskis (approved):
LGTM, I hope after getting rid of the queryAndUpdate we wouldn't need those refetches.

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: good point, I'll change it.

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: Yep, makes sense.

lmuntaner: but you mean `skipCheck: false`, right? This is the point of this `reloadNeuron`, the update is not enough. The stake from neuron and ledger account might not yeet match.

frontend/svelte/src/lib/services/neurons.services.ts

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: done

PR #896, opened 2022-05-24
build: chunk nns-js

peterpeterparker:
# Motivation

We don't want all users to have to load the hardware wallet code because not all of them actually use ledger devices.

# Changes

- modify `rollup.config` manual chunking to preserve `nns-js` chunks

# Observation

Local `npm run build` results.

Before changes:

before

After changes:

after


peterpeterparker:
Unfortunately it doesn't improve much the performance therefore I close this PR. Solution is probably dynamically importing any nns-js usage.

PR #897, opened 2022-05-24
merge and spawn neuron maturity sliders have label which shows ICP 

anishh2003:
# Motivation

merge and spawn neuron maturity sliders will now show the maturity ICP equivalent as label. 
So that users can make out how much of the of their maturity ICP is being used,

# Changes

IMG_20220524_113447


hpeebles (approved):

lmuntaner:
Who suggested these changes? I thought that we were not supposed to add the amount of ICP related to maturity.

hpeebles:
> Who suggested these changes? I thought that we were not supposed to add the amount of ICP related to maturity.

That's a good point actually.

PR #898, opened 2022-05-24
feat: load neurons after sign-in

peterpeterparker:
# Motivation

Some users feel like voting is currently a bit slow because the neurons have to be loaded. This behavior happens because most user jump in the "Voting" tab in svelte from the other tabs that are still written with "Flutter".

e.g. user sign in -> flutter -> load neurons in the background -> go to "voting" tab -> select proposal -> neurons have to be loaded again

It will be improved once we will have launch the other tabs rewritten in Svelte. However, in current implementation, the neurons - unlike the accounts - are not loaded after each sign-in but "only" when user access the "Neurons" tab.

This PR change this behavior by loading the neurons after each sign-in. This behavior is identical to what Flutter does and has for end effect to make feel that the proposal detail page is faster because often the neurons would have been already loaded when accessed the page.

e.g. user sign in -> svelte -> load neurons -> go to "voting" tab -> select proposal -> neurons are most probably already loaded or needs few time to finish to be loaded

# Changes

- call `listNeurons()` from `App` after sign-in

### DIFF
--- a/frontend/svelte/src/App.svelte
+++ b/frontend/svelte/src/App.svelte
@@ -21,6 +21,7 @@
   import BusyScreen from "./lib/components/ui/BusyScreen.svelte";
   import { worker } from "./lib/services/worker.services";
   import { bindDebugGenerator } from "./lib/utils/dev.utils";
+  import { listNeurons } from "./lib/services/neurons.services";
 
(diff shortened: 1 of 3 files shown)


mstrasinskis (approved):
LGTM

PR #899, opened 2022-05-24
L2-335: List CanisterCards

lmuntaner:
# Motivation

User can see the attached canisters in a list of cards.

# Changes

* New component "CanisterCard".
* New route "CanisterDetails". Empty for now with TODOs inside.
* Add new route to the accepted routes.
* Change in the canisters store to also keep whether the data is "certified" or not. Following the other store patterns.

# Tests

* Test for the new components: CanisterCard and route CanisterDetails.
* Extend route Canisters testing.
* Fix canisters store tests.

### DIFF
--- a/frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
+++ b/frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
@@ -1,0 +1,18 @@
+<script lang="ts">
+  import type { CanisterDetails } from "../../canisters/nns-dapp/nns-dapp.types";
+  import Card from "../ui/Card.svelte";
+
+  export let canister: CanisterDetails;
+  export let role: undefined | "link" | "button" | "checkbox" = undefined;
+  export let ariaLabel: string | undefined = undefined;
+
+  let canisterId: string;
+  $: canisterId = canister.canister_id.toText();
+  let title: string;
+  $: title = canister.name.length > 0 ? canister.name : canisterId;
+</script>
+
+<Card {role} {ariaLabel} on:click testId="canister-card">
+  <h3 slot="start">{title}</h3>
+  <p>{canisterId}</p>
+</Card>
--- a/frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
+++ b/frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
@@ -8,11 +8,13 @@
 
   let canisterId: string;
   $: canisterId = canister.canister_id.toText();
-  let title: string;
-  $: title = canister.name.length > 0 ? canister.name : canisterId;
+  let validName: boolean;
+  $: validName = canister.name.length > 0;
 </script>
 
 <Card {role} {ariaLabel} on:click testId="canister-card">
-  <h3 slot="start">{title}</h3>
-  <p>{canisterId}</p>
+  <h3 slot="start">{validName ? canister.name : canisterId}</h3>
+  {#if validName}
+    <p>{canisterId}</p>
+  {/if}
 </Card>
--- a/frontend/svelte/src/lib/stores/canisters.store.ts
+++ b/frontend/svelte/src/lib/stores/canisters.store.ts
@@ -1,8 +1,8 @@
 import { writable } from "svelte/store";
-import type { CanisterDetails } from "../canisters/nns-dapp/nns-dapp.types";
+import type { CanisterDetails as CanisterInfo } from "../canisters/nns-dapp/nns-dapp.types";
 
 export interface CanistersStore {
-  canisters: CanisterDetails[] | undefined;
+  canisters: CanisterInfo[] | undefined;
   certified: boolean | undefined;
 }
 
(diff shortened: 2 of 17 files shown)


lmuntaner:
UX video:

https://user-images.githubusercontent.com/4550653/170087757-a8e98c79-ca71-428e-9f6f-9edaaea45c4b.mov

lmuntaner:
Screenshot of the tab when there are no canisters:

Screenshot 2022-05-24 at 16 50 14

frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
```diff
@@ -0,0 +1,18 @@
...
+
+  let canisterId: string;
+  $: canisterId = canister.canister_id.toText();
+  let title: string;
+  $: title = canister.name.length > 0 ? canister.name : canisterId;
+</script>
+
+<Card {role} {ariaLabel} on:click testId="canister-card">
+  <h3 slot="start">{title}</h3>
+  <p>{canisterId}</p>
```
peterpeterparker, line 17: should probably not be display if the title is the canister id but i guess this card is still under development right?

frontend/svelte/src/lib/stores/canisters.store.ts
```diff
@@ -1,19 +1,27 @@
 import { writable } from "svelte/store";
 import type { CanisterDetails } from "../canisters/nns-dapp/nns-dapp.types";
 
+export interface CanistersStore {
+  canisters: CanisterDetails[];
```
peterpeterparker, line 5: as neurons and accounts, should it be undefined per default?

doing so we can make the difference between is loaded and not loaded

peterpeterparker: I know it was an array before but since we are already there...

frontend/svelte/src/routes/Canisters.svelte
```diff
@@ -50,17 +67,23 @@
...
-      <p>
+      <p class="last-info">
         {$i18n.canisters.principal_is}
         {$authStore.identity?.getPrincipal().toText()}
       </p>
 
-      <!-- TODO(L2-335): display cards -->
-      {#each $canistersStore as canister}
-        <p>{canister.name ?? canister.canister_id}</p>
+      {#each $canistersStore.canisters as canister}
```
peterpeterparker, line 75: maybe add a key?

peterpeterparker: I know also it was like that before 😉

peterpeterparker (approved):
LGTM, just minor comments if you can have a look before merge

frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
lmuntaner: No, this is the development of the card.

This is the current implementation, it's show twice 😕 

I can implement what you say.

frontend/svelte/src/lib/stores/canisters.store.ts
lmuntaner: ok

frontend/svelte/src/routes/Canisters.svelte
lmuntaner: ok

frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
peterpeterparker: oh yeah so let's improve a bit the UI 😉

if not exists, can you also open a jira task to add a "copy" button next to the principal id as we do for the account identifier, I think it would be cool. i did not asked mischa though, if you think we should, let's ask him first

frontend/svelte/src/lib/stores/canisters.store.ts
lmuntaner: proposals is an array btw. But I'll change.

lmuntaner: done

frontend/svelte/src/routes/Canisters.svelte
lmuntaner: done

frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
lmuntaner: It does not exist, I can create it and also ask Mischa.

frontend/svelte/src/lib/stores/canisters.store.ts
peterpeterparker: good point, might improve it too there.

note: in flutter canister are loaded on app boot
currently we load canister on view only, I guess it's enough
but as yesterday with the neuron if we ever add to move and load these on app boot too, this will be useful

peterpeterparker: thanks!

frontend/svelte/src/lib/components/canisters/CanisterCard.svelte
peterpeterparker: Thanks!

PR #900, opened 2022-05-24
Debug store trigger

mstrasinskis:
# Motivation

To be able to get the debug log also on mobile devices the trigger was added to the headlines and the busy spinner (see [descriptions](https://www.notion.so/Debug-state-trigger-99c2bcdd972c48fbbcdcea98965ba76b))

# Changes

- `triggerDebugReport` directive
- confirm dialog before log

# Tests

only manual

### DIFF
--- a/frontend/svelte/src/lib/utils/dev.utils.ts
+++ b/frontend/svelte/src/lib/utils/dev.utils.ts
@@ -72,7 +72,6 @@ export function triggerDebugReport(node: HTMLElement) {
 
   return {
     destroy() {
-      stop();
       node.style.touchAction = originalTouchActionValue;
       node.removeEventListener("click", click, false);
     },
--- a/frontend/svelte/src/lib/utils/dev.utils.ts
+++ b/frontend/svelte/src/lib/utils/dev.utils.ts
@@ -1,4 +1,6 @@
+import { get } from "svelte/store";
 import { generateDebugLogProxy } from "../proxy/debug.services.proxy";
+import { i18n } from "../stores/i18n";
 
 export const isNode = (): boolean =>
   typeof process !== "undefined" &&
@@ -59,7 +61,7 @@ export function triggerDebugReport(node: HTMLElement) {
       count++;
 
       if (count === 5) {
-        generateDebugLogProxy(confirm("Save the app state to the file"));
+        generateDebugLogProxy(confirm(get(i18n).core.save_log_file));
       }
     } else {
       startTime = now;
--- a/frontend/svelte/src/lib/utils/dev.utils.ts
+++ b/frontend/svelte/src/lib/utils/dev.utils.ts
@@ -47,7 +47,7 @@ export const digestText = async (text: string): Promise<string> => {
  */
 export function triggerDebugReport(node: HTMLElement) {
   const TWO_SECONDS = 2 * 1000;
-  const originalUserSelectValue: string = node.style.userSelect;
+  const originalTouchActionValue: string = node.style.touchAction;
 
   let startTime: number = 0;
   let count = 0;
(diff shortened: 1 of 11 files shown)


frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -43,24 +43,41 @@ export const digestText = async (text: string): Promise<string> => {
 };
 
 /**
- * To not have "nns-state" in the code
- * @returns "nns-state"
+ * Bind debug logger tigger to the node (6 clicks in 2 seconds)
```
lmuntaner, line 46: 😂

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -43,24 +43,41 @@ export const digestText = async (text: string): Promise<string> => {
...
+  let count = 0;
+
+  const click = () => {
+    const now = Date.now();
+
+    if (now - startTime <= TWO_SECONDS) {
+      count++;
+
+      if (count === 5) {
+        generateDebugLogProxy(confirm("Save the app state to the file")).then();
```
lmuntaner, line 62: why the `.then()`?

lmuntaner (approved):
Looks very good! I didn't know about the `use` directive. Very promising!

Just one question if you can reply before merging.

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -43,24 +43,41 @@ export const digestText = async (text: string): Promise<string> => {
...
+      }
     } else {
-      originalLog.apply(this, args);
+      startTime = now;
+      count = 0;
     }
   };
-};
+
+  node.addEventListener("click", click, true);
```
peterpeterparker, line 70: listener shouldn't be `{passive: true}`?

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -43,24 +43,41 @@ export const digestText = async (text: string): Promise<string> => {
...
+
+  // disable text selection
+  node.style.userSelect = "none";
```
peterpeterparker, line 73: is it a must?

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -43,24 +43,41 @@ export const digestText = async (text: string): Promise<string> => {
...
-      originalLog.apply(this, args);
+      startTime = now;
+      count = 0;
     }
   };
-};
+
+  node.addEventListener("click", click, true);
+
+  // disable text selection
```
peterpeterparker, line 72: should we set `touch-action: manipulation` since users has to perform many clicks?

did not test it on touch screen though, might not needed just a spontaneous thought

frontend/svelte/src/lib/utils/dev.utils.ts
mstrasinskis: Good point, this was a must for the press and hold implementation. Let's try w/o 👍

frontend/svelte/src/lib/utils/dev.utils.ts
mstrasinskis: For the sake of reliability. A bit redundant here, is removed.

frontend/svelte/src/lib/utils/dev.utils.ts
mstrasinskis: True, should be safer with `manipulation` 👍

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -43,24 +43,38 @@ export const digestText = async (text: string): Promise<string> => {
...
+  let count = 0;
+
+  const click = () => {
+    const now = Date.now();
+
+    if (now - startTime <= TWO_SECONDS) {
+      count++;
+
+      if (count === 5) {
+        generateDebugLogProxy(confirm("Save the app state to the file"));
```
peterpeterparker, line 62: i18n?

peterpeterparker (approved):
LGTM, thanks

PR #901, opened 2022-05-25
Build patched CMC canister and deploy

bitdivine:
# Motivation
The CMC canister needs to be patched in order for cycle creation commands to work on a testnet.  The patch allows the exchange rate for buying the initial canister cycles to be set directly.

# Changes
- Commit 1: Build the patched canister in Docker, with the other patched NNS canisters.
- Commit 2: Rename `--users` in deploy.sh to `--populate`
- Commit 3: Set the exchange rate as part of `--populate`
- Commit 4: Make a proposal to allow CMC to create canisters; do this as part of `--populate`
- Cleanup:
  - Run only relevant parts of `--populate`:
    -  On local `dfx` networks, there are no subnets so don't try to authorize subnets.  It will fail.
    - If the network uses flutter, don't try to use e2e tests to create users and neurons.  It will fail.
  - Bust the NNS canister cache, as otherwise the old artifacts without the CMC patch are used.

# Tests
Running:
```
./deploy.sh nnsdapp_both
```
yielded this: https://tmxop-wyaaa-aaaaa-aaapa-cai.nnsdapp.dfinity.network/#/canister/wxns6-qiaaa-aaaaa-aaaqa-cai

I verified that I can create canisters there:
Screenshot from 2022-05-25 05-31-12

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,7 +123,7 @@ jobs:
 
       - name: Create initial state
         run: |
-          ./deploy.sh --users
+          ./deploy.sh --populate
 
       - name: Build proxy
         working-directory: proxy
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,7 @@ env:
       FLUTTER_VERSION: 2.2.3
       IC_CDK_OPTIMIZER_VERSION: 0.3.1
       IC_COMMIT: 6dc5e2dcb50569e20891af02b846a7d9a58e3489
+      NNS_CACHING_KEY: dv-0004
 
 jobs:
   nns_canisters:
@@ -32,7 +33,7 @@ jobs:
         with:
           path: |
             target/ic
-          key: "nns-canisters-${{ env.IC_COMMIT }}-${{hashFiles( 'e2e-tests/scripts/nns-canister.*' )}}-dv-0003"
+          key: "nns-canisters-${{ env.IC_COMMIT }}-${{hashFiles( 'e2e-tests/scripts/nns-canister.*' )}}-${{ env.NNS_CACHING_KEY }}"
 
       - name: Download prebuilt nns canisters
         if: steps.create_nns_canister_cache.outputs.cache-hit != 'true'
@@ -70,7 +71,9 @@ jobs:
         with:
           path: |
             target/ic
-          key: "nns-canisters-${{ env.IC_COMMIT }}-${{hashFiles( 'e2e-tests/scripts/nns-canister.*' )}}-dv-0003"
+          key: "nns-canisters-${{ env.IC_COMMIT }}-${{hashFiles( 'e2e-tests/scripts/nns-canister.*' )}}-${{ env.NNS_CACHING_KEY }}"
+      - name: Show canister versions
+        run: sha256sum target/ic/*.wasm
 
       # Cache based on the Cargo.lock
       # The cache key is always an exact match or no match (i.e. no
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -54,21 +55,16 @@ jobs:
           cache-to: type=gha,scope=cached-stage,mode=max
           outputs: target/ic
 
+      - name: Show canister versions
+        run: sha256sum target/ic/*.wasm
+
```
bitdivine, line 60: Shows what is about to be put into into the cache.

.github/workflows/build.yml
```diff
@@ -54,21 +55,16 @@ jobs:
...
       - uses: actions/checkout@v2
 
-      - name: Get cached nns canisters
-        id: create_nns_canister_cache
-        uses: actions/cache@v2
-        with:
-          path: |
-            target/ic
-          key: "nns-canisters-${{ env.IC_COMMIT }}-dv-0001"
-
```
bitdivine, line 71: This has been moved down to where the canisters are installed, to keep the NNS canister code together.

.github/workflows/build.yml
```diff
@@ -111,6 +107,15 @@ jobs:
         run: |
           dfx start --background --clean
```
bitdivine, line 109: get canisters from cache - moved down from above.

.github/workflows/build.yml
```diff
@@ -111,6 +107,15 @@ jobs:
...
 
+      - name: Get cached nns canisters
+        id: create_nns_canister_cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            target/ic
+          key: "nns-canisters-${{ env.IC_COMMIT }}-${{hashFiles( 'e2e-tests/scripts/nns-canister.*' )}}-${{ env.NNS_CACHING_KEY }}"
```
bitdivine, line 116: Show the canister versions that came out from the cache.

e2e-tests/scripts/nns-canister.Dockerfile
```diff
@@ -48,6 +48,12 @@ RUN binary="governance-canister" && \
     features="test" && \
     ic-cdk-optimizer -o "$CARGO_TARGET_DIR/${binary}_${features}.wasm" "$CARGO_TARGET_DIR/wasm32-unknown-unknown/release/${binary}.wasm"
```
bitdivine, line 50: Build the custom CMC canister.

.github/workflows/build.yml
```diff
@@ -123,7 +128,7 @@ jobs:
 
       - name: Create initial state
         run: |
-          ./deploy.sh --users
+          ./deploy.sh --populate
```
lmuntaner, line 131: thanks!

e2e-tests/scripts/nns-canister.Dockerfile
lmuntaner: How do we know it's custom?

lmuntaner (approved):
LGTM, just a couple of questions!

Thanks!

e2e-tests/scripts/nns-canister.Dockerfile
lmuntaner: aaahh, yes, that explains it.

I thought more custom as hacked somehow, just like you hacked the governance canister to speed up maturity.

lmuntaner: Ah!! ok, the key is in the `nns-canisters.patch`. Since it was not part of the PR, I didn't see the customization.

PR #903, opened 2022-05-25
L2-616: Disable HW Neurons when merging

lmuntaner:
# Motivation

User cannot select a neuron controlled via hardware wallet to merge.

# Changes

* Only neurons controlled by user are mergeable.
* Add new message for neurons controlled via hardware wallet.

# Tests

* Adapt tests.


lmuntaner:
Screenshot:

Screenshot 2022-05-25 at 14 07 30

lmuntaner:

frontend/svelte/src/lib/utils/neuron.utils.ts
```diff
@@ -152,13 +152,13 @@ export const sortNeuronsByCreatedTimestamp = (
  */
 export const isNeuronControllableByUser = ({
   neuron: { fullNeuron },
-  identity,
+  accounts,
 }: {
   neuron: NeuronInfo;
-  identity?: Identity | null;
+  accounts: AccountsStore;
```
mstrasinskis, line 158: Probably providing the mainAccount instead the AccountsStore is more cleaner (to not expose the Store structure)

mstrasinskis (approved):
LGTM. One minor suggestion.

frontend/svelte/src/lib/utils/neuron.utils.ts
lmuntaner: I also thought about it, but it's a very common signature to pass the accounts store...

But I can change it, I also prefer to pass as minimum as possible to a function.

PR #906, opened 2022-05-25
Upgrade nns-js

bitdivine:
# Motivation
We need nicely formatted bitcoin proposals

# Changes
Bump nns-js

# Tests

### DIFF
--- a/frontend/svelte/package.json
+++ b/frontend/svelte/package.json
@@ -16,7 +16,7 @@
     "format": "prettier --write --plugin-search-dir=. .",
     "test": "TZ=UTC jest",
     "test:watch": "TZ=UTC npm run test -- --watchAll",
-    "update:nns": "npm update @dfinity/nns@nightly"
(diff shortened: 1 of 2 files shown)


frontend/svelte/package.json
```diff
@@ -63,7 +63,7 @@
   "dependencies": {
     "@dfinity/agent": "^0.11.1",
     "@dfinity/auth-client": "^0.11.1",
-    "@dfinity/nns": "nightly",
```
lmuntaner, line 66: we revert this change manually

lmuntaner (changes requested):

frontend/svelte/package.json
bitdivine: Good spot.  I have changed the command to `npm update ...` so that it does the right thing without manual changes.

lmuntaner (approved):
Thanks!

PR #907, opened 2022-05-25
mobile button supports 2 lines of text

wackyleo459:
# Motivation
- Buttons on mobile should support 2 lines of text
 
Screen Shot 2022-05-25 at 12 39 30 PM

# Changes
- Enabled specific height of 70px, at breakpoint 450px
- Did not specify mixin media, since this breakpoint was specifically for this use case only

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Toolbar.svelte
+++ b/frontend/svelte/src/lib/components/ui/Toolbar.svelte
@@ -39,6 +39,10 @@
         margin: 0 var(--padding-2x);
         max-width: 406px;
       }
+
+      @media (max-width: 450px) {
+        height: 70px;
+      }
     }
   }
 </style>


frontend/svelte/src/lib/components/ui/Toolbar.svelte
```diff
@@ -39,6 +39,10 @@
         margin: 0 var(--padding-2x);
         max-width: 406px;
       }
+
+      @media (max-width: 450px) {
```
mstrasinskis, line 43: Currently we follow mobile-first approach, that's why it's better to use `min-width`. There is a helper `src/lib/themes/mixins/_media.scss` ( e.g. `@include media.min-width(medium) { ...`)

frontend/svelte/src/lib/components/ui/Toolbar.svelte
```diff
@@ -39,6 +39,10 @@
...
+        height: 70px;
```
mstrasinskis, line 44: Please use global variables as much as possible. There are a different `--padding` variations in `variables.scss`.

mstrasinskis (commented):
Not bad, the only concern is that with the `height` solution the small button is bigger then it was and if there are 3 lines of text then it's not in the button anymore (screenshot)
image
In theory `line-height: 1` help fitting 2 lines of text in the standard size button. Could you please check it out?

peterpeterparker:
I feel like this is now solved on `main` branch? should we close this PR or is it something else to do?

wackyleo459:
Yes, it's solved by Maksims. He told me to hold off, I will close this and just confirm with Maksims tomorrow. Thank you!

PR #908, opened 2022-05-25
prevent overflow x on section

wackyleo459:
# Motivation
Neuron Detail page had it's page overflow because of tooltip placements
Screen Shot 2022-06-08 at 5 57 02 PM
# Changes
- provided custom positioning for any tooltip that overflows viewport
- auto shift will happen whenever window gets resized
Screen Shot 2022-06-08 at 5 58 52 PM


peterpeterparker:
Can you try to debug why this happens first?

Sure we can set an `overflow-x: hidden`, generally speaking it's a good thing. However I think it would be nice first to fix the root cause of the issue too. In other tabs we don't have such an issue therefore I think there is something off in that particular "Neuron details" page.

wackyleo459:
There was the tooltip for some of the buttons, that would overflow to the right.
So I created a custom "rightEdge" property for the Tooltip that would enable the position to be adjusted.

Screen Shot 2022-05-26 at 4 49 15 PM

frontend/svelte/src/lib/components/neuron-detail/actions/SpawnNeuronButton.svelte
```diff
@@ -40,6 +40,7 @@
       $amount: formatICP(BigInt(MIN_NEURON_STAKE)),
     }
   )}
+  rightEdge={true}
```
peterpeterparker, line 43: cool to hear you find the root cause of the issue!

can you implement a solution that do no need to pass properties for styling purpose. I think it should be generic, the tooltip should never overflow the viewport.

if there is really really no solution, then prefer using css variables for styling purpose rather than properties.

peterpeterparker:
@wackyleo459 we have merged your workaround in #917 because we are preparing a release, thanks for the tips!
You can merge `main` in your PR and remove the workaround here.

frontend/svelte/src/lib/components/neuron-detail/actions/SpawnNeuronButton.svelte
wackyleo459: I've prevented any element  from passing properties.

wackyleo459:
@peterpeterparker  
- If the tooltip's right-side placement `getBoundingClientRect().right` is beyond the viewport : calculated as width of main tag `clientWidth`, a `rightEdge` class name is added.
- The styling for rightEdge will add `left: auto`
- This way, original placement of tooltip remains SAME, which is centered below the anchored button.
### Further change
- Similarly applied class name `leftEdge` for potential tooltips that bleed outside left edge.
Before and after
Screen Shot 2022-06-01 at 11 12 28 PM
Screen Shot 2022-06-02 at 12 40 49 AM

frontend/svelte/src/lib/components/ui/Tooltip.svelte
```diff
@@ -1,15 +1,42 @@
...
   export let id: string;
   export let text = "";
   export let noWrap: boolean = false;
+  let tooltipComponent, rightBoundary, leftBoundary, mainWidth;
+  let rightEdge = false;
+  let leftEdge = false;
+
+  // if tooltip goes beyond viewport, assign class name of 'rightEdge' or 'leftEdge'
+  afterUpdate(() => {
+    rightBoundary = tooltipComponent.getBoundingClientRect().right;
```
peterpeterparker, line 13: I think you don't need to declare component variable as you only use `rightBoundary` for example in the function

If you want fancy stuffs you can also deconstruct

maybe worth also preventing the ref on the component not being set?

All in all something like

```
const {right, left} = tooltipComponent?.getBoundingClientRect();
```

frontend/svelte/src/lib/components/ui/Tooltip.svelte
```diff
@@ -1,15 +1,42 @@
...
+    leftBoundary = tooltipComponent.getBoundingClientRect().left;
+    mainWidth = document.querySelector("main")?.clientWidth;
```
peterpeterparker, line 15: not sure it would works but instead of query you can maybe use the viewport and use some black magic of svelte

```
<script lang="ts">
   let clientWidth;
</script>

<svelte:window bind:clientWidth={clientWidth} />
```

Again not sure it works, let me know!

peterpeterparker:
Cool ideas 👍

frontend/svelte/src/lib/components/ui/Tooltip.svelte
wackyleo459: Thank you for the detailed feedback. This is helpful to making my code more simple.

frontend/svelte/src/lib/components/ui/Tooltip.svelte
peterpeterparker: Oh good point, ade me think, does your current code works out if the user resize the browser while the tooltip is open?

If not, we should use `clientWidth` to trigger the function each time it changes, something like following:

```
<script lang="ts">
   let clientWidth;
   
   // afterUpdate becomes a function
   const setBoundary = () => { ...
   
   // each time the width change, recalculate the boundary
   $: clientWidth, (() => setBoudary())()
</script>

<svelte:window bind:clientWidth={clientWidth} />
```

Of course, if it already works out when user resize the browser, just forget my comment 😉

wackyleo459: The binding to <svelte:window bind:innerWidth> helped trigger the afterUpdate() every time window gets resized. So thanks for that bit of tip!

However, I didn't need a reactive statement after all ( it was a lot of trial errors to figure that one out...)

wackyleo459:
- created CSS .rightEdge and .leftEdge class names to help prevent tooltips from overflowing
- provided further calculated custom styling for any tooltip that still overflows viewport
- auto shift will happen whenever window gets resized

mstrasinskis (approved):
Very Good 👍
The next step (definitely not in this PR) could be to take into consideration also horizontal boundaries.

PR #910, opened 2022-05-27
L2-334: Setup Canister Detail Page

lmuntaner:
# Motivation

User can see the canister title in the canister detail page.

# Changes

* New canister service "loadCanisterDetails" that fetches canister details and adds them to the store.
* New store "canisterDetailsStore" which stores the data of a canister. It's a different type than from nns dapp.
* New canister service "routePathCanisterId" to read the canister id from the url.
* Setup of "ControllersCard" and "CyclesCard" to help parallel development.
* CanisterDetails route component triggers fetching canisters and details and read them from the stores.

# Tests

* New test for new two canister services.
* Extend CanisterDetails route tests.

### DIFF
--- a/frontend/svelte/src/lib/components/ui/SkeletonParagraph.svelte
+++ b/frontend/svelte/src/lib/components/ui/SkeletonParagraph.svelte
@@ -23,7 +23,7 @@
     );
     background: var(--skeleton-text-background);
 
-    line-height: 0.8;
+    line-height: var(--skeleton-text-line-height, 0.8);
 
     user-select: none;
     pointer-events: none;
--- a/frontend/svelte/src/lib/components/ui/SkeletonTitle.svelte
+++ b/frontend/svelte/src/lib/components/ui/SkeletonTitle.svelte
@@ -1,0 +1,13 @@
+<script>
+  import SkeletonParagraph from "./SkeletonParagraph.svelte";
+</script>
+
+<span>
+  <SkeletonParagraph />
+</span>
+
+<style lang="scss">
+  span {
+    --skeleton-text-line-height: 2;
+  }
+</style>
--- a/frontend/svelte/src/lib/services/canisters.services.ts
+++ b/frontend/svelte/src/lib/services/canisters.services.ts
@@ -1,11 +1,15 @@
 import type { Principal } from "@dfinity/principal";
 import {
   attachCanister as attachCanisterApi,
+  queryCanisterDetails,
   queryCanisters,
 } from "../api/canisters.api";
-import type { CanisterDetails } from "../canisters/nns-dapp/nns-dapp.types";
+import type { CanisterDetails } from "../canisters/ic-management/ic-management.canister.types";
+import type { CanisterDetails as CanisterInfo } from "../canisters/nns-dapp/nns-dapp.types";
 import { canistersStore } from "../stores/canisters.store";
 import { toastsStore } from "../stores/toasts.store";
+import { getPrincipalFromString } from "../utils/accounts.utils";
+import { getLastPathDetail } from "../utils/app-path.utils";
 import { getIdentity } from "./auth.services";
 import { queryAndUpdate } from "./utils.services";
 
@@ -18,7 +22,7 @@ export const listCanisters = async ({
     canistersStore.setCanisters({ canisters: undefined, certified: true });
   }
 
-  return queryAndUpdate<CanisterDetails[], unknown>({
+  return queryAndUpdate<CanisterInfo[], unknown>({
     request: (options) => queryCanisters(options),
     onLoad: ({ response: canisters, certified }) =>
       canistersStore.setCanisters({ canisters, certified }),
@@ -55,5 +59,26 @@ export const attachCanister = async (
   } catch (error) {
     // TODO: Manage proper errors https://dfinity.atlassian.net/browse/L2-615
     return { success: false };
+  }
+};
+
+export const routePathCanisterId = (path: string): Principal | undefined => {
+  const maybeIdString = getLastPathDetail(path);
+  return maybeIdString !== undefined
+    ? getPrincipalFromString(maybeIdString)
+    : undefined;
+};
+
+export const getCanisterDetails = async (
+  canisterId: Principal
+): Promise<CanisterDetails | undefined> => {
+  const identity = await getIdentity();
+  try {
+    return await queryCanisterDetails({
+      canisterId,
+      identity,
+    });
+  } catch (error) {
+    // TODO: manage errors https://dfinity.atlassian.net/browse/L2-615
   }
 };
--- a/frontend/svelte/src/lib/utils/canisters.utils.ts
+++ b/frontend/svelte/src/lib/utils/canisters.utils.ts
@@ -1,0 +1,13 @@
+import type { Principal } from "@dfinity/principal";
+import type { CanistersStore } from "../stores/canisters.store";
+
+export const getCanisterInfoById = ({
+  canisterId,
+  canistersStore,
+}: {
+  canisterId: Principal;
+  canistersStore: CanistersStore;
+}) =>
+  canistersStore.canisters?.find(
+    ({ canister_id }) => canister_id.toText() === canisterId.toText()
+  );
--- a/frontend/svelte/src/lib/utils/canisters.utils.ts
+++ b/frontend/svelte/src/lib/utils/canisters.utils.ts
@@ -1,4 +1,5 @@
 import type { Principal } from "@dfinity/principal";
+import type { CanisterDetails as CanisterInfo } from "../canisters/nns-dapp/nns-dapp.types";
 import type { CanistersStore } from "../stores/canisters.store";
 
 export const getCanisterInfoById = ({
@@ -7,7 +8,7 @@ export const getCanisterInfoById = ({
 }: {
   canisterId: Principal;
   canistersStore: CanistersStore;
-}) =>
+}): CanisterInfo | undefined =>
   canistersStore.canisters?.find(
     ({ canister_id }) => canister_id.toText() === canisterId.toText()
   );
--- a/frontend/svelte/src/routes/CanisterDetail.svelte
+++ b/frontend/svelte/src/routes/CanisterDetail.svelte
@@ -36,6 +36,7 @@
     if (!SHOW_CANISTERS_ROUTE) {
       window.location.replace("/#/canisters");
     }
+    // TODO L2-628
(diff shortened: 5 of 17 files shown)


lmuntaner:
New UX video

https://user-images.githubusercontent.com/4550653/170703862-37ecf449-48ae-42cc-ada4-5dded621b9f3.mov

lmuntaner:

frontend/svelte/src/lib/components/ui/SkeletonParagraph.svelte
```diff
@@ -1,9 +1,15 @@
 <!-- adapted source: https://github.com/ionic-team/ionic-framework/tree/main/core/src/components/skeleton-text -->
 <script lang="ts">
   export let animated: boolean = true;
+  export let className: string | undefined = undefined;
```
peterpeterparker, line 4: can we avoid to javascript properties for styling purpose?

frontend/svelte/src/lib/stores/canisters.store.ts
```diff
@@ -26,4 +27,27 @@ const initCanistersStore = () => {
   };
 };
 
+/**
+ * A store that contains canister details. Used in the Canister Details tab.
+ *
+ * - setCanister: replace the current canister
+ * - reset: sets store to undefined
+ */
+const initCanisterDetailsStore = () => {
```
peterpeterparker, line 36: instead of this store can you use the pattern context+store as for example `selected-account.context`?

frontend/svelte/src/lib/services/canisters.services.ts
```diff
@@ -35,3 +42,32 @@ export const listCanisters = async ({
...
 };
+
+export const routePathCanisterId = (path: string): Principal | undefined => {
+  const maybeIdString = getLastPathDetail(path);
+  return maybeIdString !== undefined
+    ? getPrincipalFromString(maybeIdString)
+    : undefined;
+};
+
+export const loadCanisterDetails = async ({
```
peterpeterparker, line 53: isn't the canister loaded in the store? if so, shouldn't be loaded from the store as for `getAccountFromStore`?

frontend/svelte/src/lib/components/ui/SkeletonParagraph.svelte
lmuntaner: Yes, I can use css-variables.

frontend/svelte/src/lib/stores/canisters.store.ts
lmuntaner: Why not a store? We use it for proposals and neurons details page.

We might also need to manage it from services, not just components. For example, after adding cycles or editing the controllers, it will be easier if we can set the store to the new canister details and that will rerender the component.

frontend/svelte/src/lib/services/canisters.services.ts
lmuntaner: We have two different data from the canister, at the moment, both come from the stores: canistersStore and canisterDetailsStore.

The part from the canistersStore, I follow a similar patter as you say with:
```
getCanisterInfoById({ canisterId, canistersStore: $canistersStore })
```

Then the detail come directly for `$canisterDetailsStore`.

Does that make sense?

frontend/svelte/src/lib/stores/canisters.store.ts
peterpeterparker: As the store is declared globally it can lead to visual quirks on navigation if for example the user navigates from detail page A to list to detail page B.

The pattern context+store solve this issue.

Wallet page has been implemented with this pattern.
For proposal detail page there is an open issue to convert it to context+store (L2-460).
We can open an issue for the neurons page if you wish.

About "after adding cycles ... etc" check how it is done with "renaming accounts".

action rename -> sync all accounts -> auto subscribe in detail page -> get account from store -> update store in context

It isn't complicated and it rerenders the component.

frontend/svelte/src/lib/services/canisters.services.ts
peterpeterparker: Ah I see, thanks for the explanation

frontend/svelte/src/lib/stores/canisters.store.ts
lmuntaner: ok

lmuntaner: done

frontend/svelte/src/lib/components/ui/SkeletonParagraph.svelte
lmuntaner: done

lmuntaner:

frontend/svelte/src/lib/components/ui/SkeletonTitle.svelte
```diff
@@ -0,0 +1,13 @@
+<script>
+  import SkeletonParagraph from "./SkeletonParagraph.svelte";
+</script>
+
+<span>
```
peterpeterparker, line 5: is it semantically correct to have a `p` within a `span`?

frontend/svelte/src/lib/utils/canisters.utils.ts
```diff
@@ -0,0 +1,13 @@
+import type { Principal } from "@dfinity/principal";
+import type { CanistersStore } from "../stores/canisters.store";
+
+export const getCanisterInfoById = ({
+  canisterId,
+  canistersStore,
+}: {
+  canisterId: Principal;
+  canistersStore: CanistersStore;
+}) =>
```
peterpeterparker, line 10: can you define the return type?

frontend/svelte/src/routes/CanisterDetail.svelte
```diff
@@ -1,33 +1,163 @@
...
+    type SelectCanisterDetailsStore,
+  } from "../lib/types/canister-detail.context";
+  import { debugSelectedCanisterStore } from "../lib/stores/debug.store";
+  import type { CanisterDetails } from "../lib/canisters/ic-management/ic-management.canister.types";
 
   onMount(async () => {
     if (!SHOW_CANISTERS_ROUTE) {
       window.location.replace("/#/canisters");
     }
+    await listCanisters({ clearBeforeQuery: false });
```
peterpeterparker, line 39: is this correct? shouldn't it call the `listcanisters` only if not yet loaded?

if yes, should we implement some logic here in the route or should we load the canister as in Flutter or as the accounts and neurons, when the app boots?

frontend/svelte/src/lib/stores/canisters.store.ts
peterpeterparker: thanks a lot!

frontend/svelte/src/lib/components/ui/SkeletonTitle.svelte
lmuntaner: This is a Skeleton and temporary element. I don't know, but I thought it was not so important for this.

I needed a way to select that one.

frontend/svelte/src/lib/utils/canisters.utils.ts
lmuntaner: sure

frontend/svelte/src/routes/CanisterDetail.svelte
lmuntaner: If they are loaded it would still make sense to reload it in case there has been a change.

I am not a fan of booting so many things at once. I am surprised that we are considering that and at the same time we take performance so careful. Is the first pain not considered performance?

lmuntaner:
@peterpeterparker I forgot to ask for a re-review yesterday

frontend/svelte/src/routes/CanisterDetail.svelte
peterpeterparker: > If they are loaded it would still make sense to reload it in case there has been a change.

we don't do that in any other tabs (accounts, neurons and proposals) for a smooth transition between list and detail page or are we?

> Is the first pain not considered performance?

the user experience is degraded if neurons and accounts are not load at boot time

my question was: "should we implement some logic here in the route or should we load the canister when app boots"?

if you think canister should not be loaded when the app boots, then implement some logic here in the route to not reload the canisters if not needed

but again, unless it is not what we do in accounts, neurons and proposals tabs which I think we do, at least for accounts and proposals

lmuntaner: In the other details page we can load only that neuron or that proposal, therefore we don't need to load all of them.

The problem here is that the nns dapp canister does not offer a function to request only one single canister, therefore, I need to query all of them.

The nns dapp is where the name is (would be or will be when implemented) stored, therefore, I also need the canister info from the NNS Dapp, not just from IC Management.

Not the best setup. But we have:
- id + name in nns dapp canister
- other details in IC Mgt canister

Since the user can also land directly in the details page, I can't rely that the canister is already in the store. Even if it was coming from the list of canisters.

In the proposals details for example, we request one proposal, even though it could be in the `proposalsStore` duplicating the data also in the `proposalInfoStore`.

We always refetch the data (in the background) when we go to proposals, canisters or neurons. So it makes sense to also do it here.

But I agree that we don't have a clear pattern on when to fetch or not fetch, or how to do so across the app. We never agreed in any specific way and only discussed case by case in the PRs.

peterpeterparker: Thanks for the explanation, I see the problematic. Maybe we can split the loading and do only what's need?
Anyhow we can do that afterwards. Not sure there is a way nor I'll make it but happy to have a look too.
Can you add a `//TODO L2-628`?

lmuntaner:
@peterpeterparker added the comment? Did you forget to approve or didn't you finish the review?

frontend/svelte/src/routes/CanisterDetail.svelte
lmuntaner: done

frontend/svelte/src/lib/components/ui/SkeletonTitle.svelte
peterpeterparker: Yep semantically speaking `p` isn't a valid child of `span`

span excepts [phrasing content](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-span-element)

phrasing content [does not contain](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2) p

peterpeterparker (approved):
LGTM, thanks

PR #912, opened 2022-05-30
e2e test: USER_A4 User can transfer funds to self

bitdivine:
# Motivation
e2e tests are required for all user stories, in particular for [USER_A4: New transactions (same user)](https://www.notion.so/NNS-Dapp-Manual-Testing-Accounts-87b7b439dd6a4f94b1459916f6b6a09c#32db788158154da4869cf0210aeb3a19)

# Changes
- Add a test to verify that a user can transfer funds from the main account to a linked account

# Tests
This is a test.  See CI.


e2e-tests/specs/user-A04-send-icp-same-user.e2e.ts
```diff
@@ -0,0 +1,86 @@
...
+import { AccountsTab } from "../components/accounts-tab";
+import { Header } from "../components/header";
+
+/**
+ * Verifies that new users get a main account.
+ */
+describe("Users can send ICP between accounts", () => {
+  const mainAccountName = "Main";
+  const linkedAccountName = "Grizly";
+  let mainAccountValueBefore = NaN; // Value is filled in during the test.
```
peterpeterparker, line 15: it's more common to use `undefined` for such case than `Not a number`, can you update this?

e2e-tests/specs/user-A04-send-icp-same-user.e2e.ts
```diff
@@ -0,0 +1,86 @@
...
+      "Wait for the accounts tab"
+    );
+  });
+
+  it("Setup: Get ICP", async () => {
+    await new Header(browser).getIcp(10);
+  });
+
+  it(`Setup: Create a linked account '${linkedAccountName}'`, async () => {
+    await browser.setWindowSize(800, 1000);
```
peterpeterparker, line 39: you don't rollback the window size here? not sure I get when it's rollbacked or not and it doesn't block the PR at all, just a question in case it would need a rollback of the size

e2e-tests/specs/user-A04-send-icp-same-user.e2e.ts
```diff
@@ -0,0 +1,86 @@
...
+      await accountsTab.getAccountByName(
+        linkedAccountName,
+        "Ger linked account to check balance after"
+      )
+    );
+    expect(linkedAccountValueAfter).toBe(
+      linkedAccountValueBefore + transferIcp
+    );
+    const fee = mainAccountValueBefore - transferIcp - mainAccountValueAfter;
+    const expectedMaxFee = 0.0001001; // Small allowance for javascript imprecision.
```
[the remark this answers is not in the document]
peterpeterparker: or the in-accurency is in the test itself?

peterpeterparker (approved):
LGTM, if you can have a look to few comments

e2e-tests/specs/user-A04-send-icp-same-user.e2e.ts
bitdivine: Good catch.  This might have just been for me when looking at screenshots.  Let me check.

e2e-tests/specs/user-A04-send-icp-same-user.e2e.ts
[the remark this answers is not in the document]
bitdivine: The inaccuracy was in the test itself, when we were providing full precision and I should have been using bigint.  Now the point is moot as we show only very limited precision.

PR #917, opened 2022-05-30
style: prevent overflow-x on neurons detail page (workaround)

peterpeterparker:
# Motivation

@wackyleo459 is currently working on #908 to fix the `overflow-x` on the neurons detail page.

Meanwhile we also want to release both "accounts" and "neurons" tabs in svelte. So, it would be nice to release with a clean ui on mobile too.

That's why, this PR provides the workaround @wackyleo459 had created in #908 to solve the issue as temporary solution while Sue Ann continues to work on the other PR.

# Changes

- set no `overflow-x` on the `main` element


mstrasinskis (approved):
LGTM

wackyleo459:
I've managed to get a Fix on L2-610. Please take a look at #908 and leave review. Thank you!

PR #920, opened 2022-05-30
feat: shorten transaction fee display

peterpeterparker:
# Motivation

`0.0001 ICP` is better readable than `0.000100 ICP`. Because it is a static value, we can allow ourselves to shorten it to make it more readable. Request from Mischa.

# Changes

- display static transaction fee with 4 digits

# Note

As we use `8` "hardcoded" for ICP, I also hardcoded 4. Not against a better idea!

# Screenshot

Capture d’écran 2022-05-30 à 21 54 57


lmuntaner (approved):
LGTM

PR #921, opened 2022-05-31
feat: pre-fill subaccount input for rename

peterpeterparker:
# Motivation

Pre-fill input with current subaccount name when displaying renaming modal . Mischa's input.

# Changes

- pre-fill input with current name


lmuntaner (approved):
LGTM

PR #922, opened 2022-05-31
feat: translate nns-dapp errors and display max length in account error msg

peterpeterparker:
# Motivation

- translate nns-dapp canister errors
- display max length (24 characters) in error message related to account's name

# Changes

- transform error msgs from plain text to i18n keys
- use placeholders `substitutions` object from throw error to display
- add and replace placeholders `substitutions` in toast msg (option)
- edit error msg with "max 24 chars" length

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Toast.svelte
+++ b/frontend/svelte/src/lib/components/ui/Toast.svelte
@@ -4,7 +4,7 @@
    */
   import { toastsStore } from "../../stores/toasts.store";
   import { fade, fly } from "svelte/transition";
-  import { translate } from "../../utils/i18n.utils";
+  import {type I18nSubstitutions, replacePlaceholders, translate} from "../../utils/i18n.utils";
   import { i18n } from "../../stores/i18n";
   import type { ToastLevel, ToastMsg } from "../../types/toast";
   import { onDestroy, onMount } from "svelte";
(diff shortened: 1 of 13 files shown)


frontend/svelte/src/lib/components/ui/Toast.svelte
```diff
@@ -17,13 +21,13 @@
   let labelKey: string;
   let level: ToastLevel;
   let detail: string | undefined;
-  let substitutions: { [from: string]: string } | undefined;
+  let substitutions: I18nSubstitutions | undefined;
 
   $: ({ labelKey, level, detail, substitutions } = msg);
-  $: text = replacePlaceholders(
-    `${translate({ labelKey })}${detail !== undefined ? ` ${detail}` : ""}`,
+  $: text = `${replacePlaceholders(
```
peterpeterparker, line 27: @lmuntaner here in main you replaced the placeholders in the `detail` too while in my PR I only replaced in the translations because I thought the `detail` is often the error itself. If you rather like as in main can revert for sure.

lmuntaner: I don't have a strong opinion. I thought that just in case there is also something to replace in `detail` as well. For now they are also only errors, but it might evolve differently. We'll revert if we need it.

lmuntaner (approved):
LGTM

frontend/svelte/src/lib/components/ui/Toast.svelte
peterpeterparker: Thanks for the feedback, sure let's do that.

PR #923, opened 2022-05-31
L2-227: Create Canister part II

lmuntaner:
# Motivation

User can create a canister from the main account or a subaccount.

# Changes

* Implemented the "ConfirmCyclesCanister" to create the canister on click.
* New canister service: "createCanister".
* Added the step to select an account when creating a canister.
* Added a prop to SelectAccount: "hideHarwareWalletAccounts"
* Added the parameter to create a canister from a subaccount in the canister api "createCanister"

# Tests

* Extend testing case in CreateOrLinkCanisterModal until canister creation.
* New test case in CreateOrLinkCanisterModal to check that hardware wallet accounts are not shown.
* New test for new canister service.
* New test case in SelectAccount for the new prop.

### DIFF
--- a/frontend/svelte/src/lib/api/canisters.api.ts
+++ b/frontend/svelte/src/lib/api/canisters.api.ts
@@ -6,7 +6,10 @@ import { principalToSubAccount } from "../canisters/cmc/utils";
 import { ICManagementCanister } from "../canisters/ic-management/ic-management.canister";
 import type { CanisterDetails } from "../canisters/ic-management/ic-management.canister.types";
 import type { NNSDappCanister } from "../canisters/nns-dapp/nns-dapp.canister";
-import type { CanisterDetails as CanisterInfo } from "../canisters/nns-dapp/nns-dapp.types";
+import type {
+  CanisterDetails as CanisterInfo,
+  SubAccountArray,
+} from "../canisters/nns-dapp/nns-dapp.types";
 import { CYCLES_MINTING_CANISTER_ID } from "../constants/canister-ids.constants";
 import { HOST } from "../constants/environment.constants";
 import { createAgent } from "../utils/agent.utils";
@@ -14,6 +17,7 @@ import { logWithTimestamp } from "../utils/dev.utils";
 import { CREATE_CANISTER_MEMO, TOP_UP_CANISTER_MEMO } from "./constants.api";
 import { sendICP } from "./ledger.api";
 import { nnsDappCanister } from "./nns-dapp.api";
+import { toSubAccountId } from "./utils.api";
 
 export const queryCanisters = async ({
   identity,
@@ -86,10 +90,12 @@ export const createCanister = async ({
   identity,
   amount,
   name,
+  fromSubAccount,
 }: {
   identity: Identity;
   amount: ICP;
   name?: string;
+  fromSubAccount?: SubAccountArray;
 }): Promise<Principal> => {
   logWithTimestamp("Create canister call...");
 
@@ -102,6 +108,8 @@ export const createCanister = async ({
     principal: CYCLES_MINTING_CANISTER_ID,
     subAccount: SubAccount.fromBytes(toSubAccount) as SubAccount,
   });
+  const fromSubAccountId =
+    fromSubAccount !== undefined ? toSubAccountId(fromSubAccount) : undefined;
 
   // Transfer the funds
   const blockHeight = await sendICP({
@@ -109,6 +117,7 @@ export const createCanister = async ({
     identity,
     to: recipient.toHex(),
     amount,
+    fromSubAccountId,
   });
 
   // If this fails or the client loses connection
--- a/frontend/svelte/src/lib/components/accounts/SelectAccount.svelte
+++ b/frontend/svelte/src/lib/components/accounts/SelectAccount.svelte
@@ -9,6 +9,7 @@
   export let disableSelection: boolean = false;
   export let filterIdentifier: string | undefined = undefined;
   export let displayTitle: boolean = false;
+  export let hideHarwareWalletAccounts: boolean = false;
 
   const dispatch = createEventDispatcher();
   const chooseAccount = (selectedAccount: Account) => {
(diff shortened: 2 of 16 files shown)


lmuntaner:
UX Video:

https://user-images.githubusercontent.com/4550653/171135502-1960eb0c-22a8-47fc-926d-618a00d0b34f.mov

lmuntaner:

The experience after the creation is not the best. We can't find the canister and since we don't have when the canister was created, the order is the same as returned by nns-dapp. Not sure which one is that, but not by order of creation.

I will talk with Mischa to improve this by maybe just showing the id in the notification and keeping the notification for longer instead of disappearing in 4 seconds. But that will be done in another PR after this is discussed with Mischa.

lmuntaner:
@peterpeterparker  Why is that not possible? in nns-prod there is also this conversio rate or 1 to 1. It depends on the exchange rate.

I still have a pending talk with Hamish about that, but I implemented the same calculations as in Flutter.

peterpeterparker:
currently 3 ICP are worth 18.4827 T Cycles

frontend/svelte/src/lib/components/canisters/ConfirmCyclesCanister.svelte
```diff
@@ -1,4 +1,94 @@
...
+    });
+    const { success } = await createCanister({
+      amount,
+      fromSubAccount: account.subAccount,
+    });
+    stopBusy("create-canister");
+    if (success) {
+      // We don't wait for `syncAccounts` to finish to give a better UX to the user.
+      // `syncAccounts` might be slow since it loads all accounts and balances.
+      syncAccounts();
```
peterpeterparker, line 36: could we move sync accounts into the service or create a service that does both "create canister" and "sync accounts"?

kind of think that such logic should not finds place within the ui layer

frontend/svelte/src/lib/modals/canisters/CreateOrLinkCanisterModal.svelte
```diff
@@ -72,18 +91,30 @@
...
+        {icpToCyclesRatio}
         bind:amount
         on:nnsClose
         on:nnsSelectAmount={selectAmount}
       />
     {/if}
-    {#if currentStep?.name === "ConfirmCycles"}
-      <ConfirmCyclesCanister on:nnsClose />
+    {#if currentStep?.name === "ConfirmCycles" && amount !== undefined && account !== undefined}
+      <ConfirmCyclesCanister
```
peterpeterparker, line 112: lots of props, events, handler back and forth between this components and child components. Is it ok for you if afterwards I try to refactor with a context+store as I did in accounts?

frontend/svelte/src/lib/components/canisters/ConfirmCyclesCanister.svelte
```diff
@@ -1,4 +1,94 @@
...
+    }
+  };
+</script>
+
+<div class="wizard-wrapper wrapper" data-tid="confirm-create-canister-screen">
+  <div class="content">
+    <div class="conversion">
+      <h3>{formatNumber(amount, { minFraction: 2, maxFraction: 2 })}</h3>
+      <p>{$i18n.core.icp}</p>
+      {#if tCyclesFormatted !== undefined}
```
peterpeterparker, line 50: on a slow network the ui gonna be a bit flacky because source and button might be displayed first. is it worth adding a spinner and display everything is ready or "just" add more content within the if to that elements do not move on screen?

lmuntaner: This will be undefined only if we don't have an exchange rate, yet, the exchange rate is loaded at the beginning of the wizard.

I'd say that we need to discuss how to handle the edge case where we don't have an exchange rate. At the moment, the user will be able to continue, just without seeing the T Cycles.

Maybe we want to do that, but showing a message that the rate is not yet available, so it's up to him to accept and calculate the T Cycles.

Or we might want to stop moving on to even add cycles until the rate is loaded.

frontend/svelte/src/lib/modals/canisters/CreateOrLinkCanisterModal.svelte
peterpeterparker: Beside the pattern itself which I like, I particularly think it can be useful here for the future if we ever implement/add accounts name. Kind of feel in a couple of weeks, months, it can make easier to comeback to the code and extend it.

Let's see @mstrasinskis feedback.

frontend/svelte/src/lib/components/canisters/ConfirmCyclesCanister.svelte
peterpeterparker: Good point. Agree with you, probably worth to think about it. Might be an edge case, might happens often on slow network, hard to tell but worth thinking about it.

Current prod UI is glitchy, if we can make it smooth, users, begin with me, are going to be happy 😉

peterpeterparker:
> not in testnet, hehe.

ah, gotcha! thx

frontend/svelte/src/lib/modals/canisters/CreateOrLinkCanisterModal.svelte
lmuntaner: I prefer simple props, I think that makes component more reusable.

frontend/svelte/src/lib/api/canisters.api.ts
```diff
@@ -102,13 +108,16 @@ export const createCanister = async ({
     principal: CYCLES_MINTING_CANISTER_ID,
     subAccount: SubAccount.fromBytes(toSubAccount) as SubAccount,
   });
+  const fromSubAccountId =
```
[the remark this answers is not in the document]
lmuntaner: done

frontend/svelte/src/lib/components/canisters/ConfirmCyclesCanister.svelte
lmuntaner: done

frontend/svelte/src/lib/modals/canisters/CreateOrLinkCanisterModal.svelte
peterpeterparker: in that sense yeah but not sure we gonna use `ConfirmCyclesCanister` elsewhere 😉.

it doesn't block this PR anyway, happy to do it afterwards.

frontend/svelte/src/lib/components/accounts/SelectAccount.svelte
```diff
@@ -52,14 +53,16 @@
...
 
-    {#each hardwareWalletAccounts as hardwareWalletAccount}
-      <AccountCard
-        role="button"
-        on:click={() => chooseAccount(hardwareWalletAccount)}
-        account={hardwareWalletAccount}
-        >{hardwareWalletAccount.name}</AccountCard
-      >
-    {/each}
+    {#if !hideHarwareWalletAccounts}
```
peterpeterparker, line 56: FYI I'll let you merge first this PR and then sync and update PR #924 accordingly

lmuntaner: ah, yes, I saw your changes there as well.

Thanks!

PR #924, opened 2022-05-31
feat: display title "My accounts" on select accounts component if user has hardware wallets too

peterpeterparker:
# Motivation

On select accounts component, the title "My accounts" can be displayed if user has sub-accounts. It can also be displayed if user has hardware wallets.

# Changes

- if `displayTitle` prop is set and user has no subaccounts but has hardware wallet, display title as well


lmuntaner (approved):
LGTM

PR #925, opened 2022-05-31
Fix: Improve split and spawn neuron

lmuntaner:
# Motivation

User can see the splitted or spawned neurons immediately after the action.

The amount to split a neuron will be the amount of the new neuron. This is more user friendly.

# Changes

* Add transaction fee to the amount on `splitNeuron` service.
* Neuron services "splitNeuron" and "spawnNeuron" do not load only one neuron, but call "listNeurons".

# Tests

* Test to check that the transaction fee is added to the amount when splitting a neuron.

### DIFF
--- a/frontend/svelte/src/lib/services/neurons.services.ts
+++ b/frontend/svelte/src/lib/services/neurons.services.ts
@@ -30,7 +30,7 @@ import {
 import { getNeuronBalance } from "../api/ledger.api";
 import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
 import { IS_TESTNET } from "../constants/environment.constants";
-import { E8S_PER_ICP } from "../constants/icp.constants";
+import { E8S_PER_ICP, TRANSACTION_FEE_E8S } from "../constants/icp.constants";
 import { MAX_CONCURRENCY } from "../constants/neurons.constants";
 import type { LedgerIdentity } from "../identities/ledger.identity";
 import { getLedgerIdentityProxy } from "../proxy/ledger.services.proxy";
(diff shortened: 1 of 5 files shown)


lmuntaner:

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -622,18 +622,16 @@ export const splitNeuron = async ({
...
+    const transactionFeeAmount = TRANSACTION_FEE_E8S / E8S_PER_ICP;
+    const stake = convertNumberToICP(amount + transactionFeeAmount);
 
     if (!isEnoughToStakeNeuron({ stake, withTransactionFee: true })) {
       throw new InsufficientAmountError();
     }
 
     await splitNeuronApi({ neuronId, identity, amount: stake });
-    toastsStore.success({
-      labelKey: "neuron_detail.split_neuron_success",
```
mstrasinskis, line 633: Should the label be also removed from `en.json`?

mstrasinskis (approved):
LGTM, just one question.

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: yes, thanks!

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -622,18 +622,16 @@ export const splitNeuron = async ({
...
-    });
 
-    await getAndLoadNeuron(neuronId);
+    await listNeurons();
```
peterpeterparker, line 634: shouldn't you add a test that check that `listNeurons` is effectively called after the function?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -704,7 +702,7 @@ export const spawnNeuron = async ({
 
     await spawnNeuronApi({ neuronId, percentageToSpawn, identity });
 
-    await getAndLoadNeuron(neuronId);
+    await listNeurons();
```
peterpeterparker, line 705: likewise

PR #926, opened 2022-05-31
Fix/l2 449 open proposals filter

mstrasinskis:
# Motivation

https://dfinity.atlassian.net/browse/L2-449
Because of applied in the neuronStore filter (`hasValidStake`), only neurons with enough stake+maturity are available in the store. But in the proposal.ballots list all neuron statuses are delivered. The 'open' filter is based on proposal.ballots state. In the end this confuses users because proposals with filtered-out (and not voted) neuronIds in their ballots list are shown.

image

# Changes

`hideProposal` also checks if is neuronId available `belongsToValidNeuron`

# Tests

- related to Proposals page tests were updated


frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -94,19 +96,26 @@ const matchFilters = ({
...
 }): boolean => {
   const { excludeVotedProposals } = filters;
 
   const { status, ballots } = proposalInfo;
+  const belongsToValidNeuron = (id: NeuronId) =>
+    neurons.find(({ neuronId }) => neuronId === id) !== undefined;
   const containsUnspecifiedBallot = (): boolean =>
-    // Sometimes ballots contains all neurons with Vote.UNSPECIFIED
-    // something ballots is empty (inconsistent backend behaviour)
-    ballots?.length === 0
```
peterpeterparker, line 107: I get it right, we assume this "inconsistent backend behaviour" doesn't exist anymore therefore we don't check it anymore?

frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -94,19 +96,26 @@ const matchFilters = ({
...
+  neurons,
 }: {
   proposalInfo: ProposalInfo;
   filters: ProposalsFiltersStore;
+  neurons: NeuronInfo[];
 }): boolean => {
   const { excludeVotedProposals } = filters;
 
   const { status, ballots } = proposalInfo;
+  const belongsToValidNeuron = (id: NeuronId) =>
```
peterpeterparker, line 108: not a must, just something that went through my mind why reading the code:

not sure what would be the impact on performance but to speed up if would maybe be possible to map `$definedNeuronsStore` - or create a new derived store - that contains only the list of `NeuronId`. 

so that in `containsUnspecifiedBallot` we would do `list.includes(ballotNeuronId)` instead of `array.find(...`

again, more a note to my future self probably

frontend/svelte/src/lib/utils/proposals.utils.ts
mstrasinskis: @peterpeterparker  Long story short, ballots === [] is also the case when all the neurons are too young for voting. So the previous logic conflicts with this case => was removed.

frontend/svelte/src/lib/utils/proposals.utils.ts
mstrasinskis: I'm pretty sure this is not a performance issue. Let's postpone it till the neuron store is updated.

peterpeterparker: Yeah definitely, like I said, more a note for my future self

peterpeterparker (approved):
LGTM thanks

PR #935, opened 2022-06-01
style: redo a different background for card block on the proposal detail page

peterpeterparker:
# Motivation

Card blocks on proposal detail page used to be presented with a different background. This redo this ui colors.

# Note

it used to be `--gray100` and now will be `-gray50` because I deprecated 100.

# Changes

- set background and contrast to `-gray50` for card block.

# Screenshots

Capture d’écran 2022-06-01 à 16 11 31
Capture d’écran 2022-06-01 à 16 11 49


mstrasinskis (approved):
LGTM, thanks

PR #940, opened 2022-06-02
feat: do not reload canisters data on back

peterpeterparker:
# Motivation

To smoothness the ux, do not reload the canisters data on "back" from canister detail page.

This PR aligns the behavior to the "Voting" behavior.

# Changes

- test if data should be reloaded on back
- refactor existing test of "Voting" tab to be reused for "Canisters" tab

### DIFF
--- a/frontend/svelte/src/lib/utils/navigation.utils.ts
+++ b/frontend/svelte/src/lib/utils/navigation.utils.ts
@@ -10,13 +10,13 @@ export const reloadOnBack = <T>({
   currentData,
 }: {
   expectedPreviousPath: AppPath;
-  effectivePreviousPath: AppPath | string;
-  currentData: T[];
+  effectivePreviousPath: AppPath | string | undefined;
+  currentData: T[] | undefined;
 }): boolean => {
   const isReferrerDetail: boolean = isRoutePath({
     path: expectedPreviousPath,
     routePath: effectivePreviousPath,
   });
 
-  return isArrayEmpty(currentData) || !isReferrerDetail;
+  return isArrayEmpty(currentData ?? []) || !isReferrerDetail;
 };
--- a/frontend/svelte/src/lib/utils/navigation.utils.ts
+++ b/frontend/svelte/src/lib/utils/navigation.utils.ts
@@ -1,0 +1,22 @@
+import type { AppPath } from "../constants/routes.constants";
+import { isRoutePath } from "./app-path.utils";
+import { isArrayEmpty } from "./utils";
+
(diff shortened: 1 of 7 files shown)


frontend/svelte/src/lib/utils/navigation.utils.ts
```diff
@@ -0,0 +1,22 @@
+import type { AppPath } from "../constants/routes.constants";
+import { isRoutePath } from "./app-path.utils";
+import { isArrayEmpty } from "./utils";
+
+// If the previous page is a particular detail page and if we have data in store, we don't reset and query the data in store after the route is mounted.
+// We do this to smoothness the back and forth navigation between the page and the detail page that have store that are not loaded at boot time.
+export const reloadRouteData = <T>({
```
lmuntaner, line 7: I thought that we agreed to always refetch the data. When did we decide not to?

lmuntaner: Maybe we could fetch the data, just not clearing the current store.

lmuntaner (commented):
A question.

Also, why not in Neurons and Account? Who and how do we make these decisions?

peterpeterparker:
Not in the "neurons" and "accounts" tab because these are loaded on the app scale and not in the particular route.

frontend/svelte/src/lib/utils/navigation.utils.ts
peterpeterparker: Months ago, in March or begin of April, when we launched the "Voting" tab

peterpeterparker: > Maybe we could fetch the data, just not clearing the current store.

This PR aligns to an existing behavior. So let's do the same as what's already in production in a first step.

If you want to discuss the behavior we can do it afterwards sure.

lmuntaner (approved):

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -139,3 +139,5 @@ export const mapPromises = async <T, R>(
 
   return Promise.all(items.map(async (item) => await fun(item)));
 };
+
+export const isArrayEmpty = <T>({ length }: T[]): boolean => length <= 0;
```
mstrasinskis, line 143: Nitpick: according to specs array.length can not be negative.

mstrasinskis (approved):
LGTM

frontend/svelte/src/lib/utils/utils.ts
peterpeterparker: Java PTSD 😉

Modified to `length === 0`

PR #944, opened 2022-06-02
fix: improve mobile button long text support

mstrasinskis:
# Motivation

To have the same button height for the buttons with the long text the line-height was decreased.

# Changes

- Change A and Button default line-height to be default
- remove redundant styles

# Before

image

image

image

# After

image

image

image


lmuntaner (approved):
LGTM

PR #945, opened 2022-06-02
L2-665: Improve getting neuron identity

lmuntaner:
# Motivation

User can perform neuron actions before accounts are loaded.

# Changes

* Check first if principal in authStore is the controller of a neuron in getIdentityOfControllerByNeuronId
* Rename NotAuthorizedError to NotAuthorizedNeuronError to customize the error message to neurons.
* getAccountIdentityByPrincipal does not throw an error when no account, instead returns null.

# Tests

* Add tests for getIdentityOfControllerByNeuronId.
* Add test case for getAccountIdentityByPrincipal

### DIFF
--- a/frontend/svelte/src/lib/services/accounts.services.ts
+++ b/frontend/svelte/src/lib/services/accounts.services.ts
@@ -226,14 +226,14 @@ export const getAccountIdentity = async (
 
 export const getAccountIdentityByPrincipal = async (
   principalString: string
-): Promise<Identity | LedgerIdentity> => {
+): Promise<Identity | LedgerIdentity | null> => {
   const accounts = get(accountsStore);
   const account = getAccountByPrincipal({
     principal: principalString,
     accounts,
   });
   if (account === undefined) {
-    throw new Error(`Account with principal ${principalString} not found!`);
+    return null;
   }
   return getAccountIdentity(account.identifier);
 };
--- a/frontend/svelte/src/lib/services/accounts.services.ts
+++ b/frontend/svelte/src/lib/services/accounts.services.ts
@@ -226,14 +226,14 @@ export const getAccountIdentity = async (
 
 export const getAccountIdentityByPrincipal = async (
   principalString: string
-): Promise<Identity | LedgerIdentity | null> => {
+): Promise<Identity | LedgerIdentity | undefined> => {
   const accounts = get(accountsStore);
   const account = getAccountByPrincipal({
     principal: principalString,
     accounts,
   });
   if (account === undefined) {
-    return null;
+    return;
   }
   return getAccountIdentity(account.identifier);
 };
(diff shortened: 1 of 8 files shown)


lmuntaner:

frontend/svelte/src/lib/services/accounts.services.ts
```diff
@@ -226,14 +226,14 @@ export const getAccountIdentity = async (
 
 export const getAccountIdentityByPrincipal = async (
   principalString: string
-): Promise<Identity | LedgerIdentity> => {
+): Promise<Identity | LedgerIdentity | null> => {
```
peterpeterparker, line 229: `getAccountByPrincipal` returns `undefined` so I would also returns `undefined` and not `null` to stick to the same pattern

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -110,28 +110,38 @@ export const getNeuronFromStore = (
...
 ): Promise<Identity> => {
-  const { neuron } = await getIdentityAndNeuronHelper(neuronId);
+  const { neuron, identity } = await getIdentityAndNeuronHelper(neuronId);
 
   if (
     neuron.fullNeuron === undefined ||
     neuron.fullNeuron.controller === undefined
   ) {
-    throw new NotAuthorizedError();
+    throw new NotAuthorizedNeuronError("Neuron has no controller");
```
peterpeterparker, line 123: i18n or the error is mapped later on?

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -110,28 +110,38 @@ export const getNeuronFromStore = (
...
   }
 
+  // Check whether identity from authStore is controller
+  if (neuron.fullNeuron.controller === identity.getPrincipal().toText()) {
```
peterpeterparker, line 127: not necessary in this pr, just an idea: we have a `isNeuronControllableByUser` maybe we can have a function `isNeuronControllableByIdentity` that contains this check and the previous one

frontend/svelte/src/lib/services/accounts.services.ts
lmuntaner: sure, makes sense

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: There is no i18n error mapping in neurons. I can do that in another neuron.

The error messages are used as details for now. I can follow your pattern in neurons as well in another PR.

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: Yes, it could be the helper for this two checks.

frontend/svelte/src/lib/services/accounts.services.ts
lmuntaner: done

lmuntaner:

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: Works for me, maybe in another PR indeed

peterpeterparker (approved):
LGTM, thx

PR #948, opened 2022-06-02
fix: button height

mstrasinskis:
# Motivation

Restore button height. It needs more time to find the better solution.

# Changes

Undo line-heigh.

# Tests

image

https://user-images.githubusercontent.com/98811342/171684060-2e8434f0-a714-4690-a289-571b9979d522.mov


peterpeterparker (approved):
Thanks

PR #957, opened 2022-06-07
feat: improve canister details page routing and data handling

peterpeterparker:
# Motivation

Improve canister details page routing and data handling.

- load canisters store only if needed
- display error message if canister not found while navigating to list page
- disable cycles action if loading

# Note

Pattern similar to `Wallet.svelte` (routing and data handling) and `ProposalDetail.svelte` (reload store).
There are possibilities of refactor to make such function generic and reused in other routes. We can tackle this if we want in other PR.

# Changes

- rename `refetchDetails` to `reloadDetails` and modify declaration to accept canister id as param
- remove `getCanisterInfoById`
- add `getCanisterFromStore` to service
- modify `routePathCanisterId` to return a `string` (needed for error message)
- update `<CanisterDetails />` with above listed patterns to handle route and to load stores only when needed
- refactor `canisterIdString` to avoid typescript checks
- disable cycles button on load

# Screenshots

ezgif com-gif-maker


frontend/svelte/src/lib/i18n/en.json
```diff
@@ -44,6 +44,7 @@
     "accounts_not_found": "An error occurred while loading the accounts information.",
     "account_not_found": "Sorry, the account \"account_identifier\" was not found",
     "transactions_not_found": "An error occurred while loading the transactions.",
+    "canister_not_found": "Sorry, the canister \"canister_id\" was not found",
```
lmuntaner, line 47: Don't we use dollar signs for parameters to better differentiate them?

frontend/svelte/src/routes/CanisterDetail.svelte
```diff
@@ -34,107 +33,134 @@
...
-      if (!isRoutePath({ path: AppPath.CanisterDetail, routePath })) {
-        return;
-      }
-      const newCanisterId = routePathCanisterId(routePath);
+  let routeCanisterId: string | undefined;
+  $: routeCanisterId = routePathCanisterId($routeStore.path);
+
+  let selectedCanister: CanisterInfo | undefined;
+  $: $canistersStore,
+    (selectedCanister = getCanisterFromStore(routeCanisterId));
```
lmuntaner, line 97: Is it a pattern you want to use instead of passing the store as param?

```
selectedCanister = getCanisterFromStore({ canisterStore: $canisterStore, routeCanisterId });
```

To be honest I feel it's a little hacky and I have not been using it in neurons. But we can refactor to keep consistency.

lmuntaner (approved):
Minor comments, already approving.

frontend/svelte/src/lib/i18n/en.json
peterpeterparker: I followed what was done for "account_not_found" 🤷‍♂️

frontend/svelte/src/routes/CanisterDetail.svelte
peterpeterparker: Not against refactoring. I moved and refactored `getAccountFromStore` to path store to utils in this PR. Will have a look to the rest (accounts) afterwards.

frontend/svelte/src/lib/i18n/en.json
peterpeterparker: Updated both

lmuntaner: thanks!

PR #959, opened 2022-06-07
refactor: get account from store path store

peterpeterparker:
# Motivation

Follow-up of #957 - refactor `getAccountFromStore` to not access the store directly but gets it as parameter

# Changes

- `getAccountFromStore` function's declaration


lmuntaner (approved):
LGTM thanks!

PR #961, opened 2022-06-07
Add `get_proposal_payload` to NNS Dapp canister

hpeebles:
# Motivation

This is the first step towards being able to display all proposal payloads (even those containing canister wasms).
This PR adds `get_proposal_payload` to the NNS Dapp canister which internally holds a cache of proposal payloads which have been transformed into JSON. Where a payload contains a wasm, that wasm is replaced with the wasm hash, massively reducing the payload size.
When getting a proposal's payload, either the payload is already in the cache, in which case it is simply returned, or if not found in the cache, a call is made to the governance canister to get the proposal, its payload is then converted into JSON, stored in the cache, and returned to the caller.

# Changes

- Add `get_proposal_payload` method
- Store cache of proposal payloads inside the NNS Dapp canister

# Tests

I have tested this new endpoint locally for a range of proposals using dfx (eg. `dfx canister --network testnet call nns-dapp get_proposal_payload '(114)'`).

There are also unit tests which ensure proposals can successfully be converted into JSON.


rs/src/proposals.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    fn identity<Out>(payload_bytes: &[u8]) -> Result<String, String>
+    where
+        Out: CandidType + Serialize + DeserializeOwned,
+    {
+        transform::<Out, Out>(payload_bytes)
+    }
+
+    match nns_function {
+        1 => identity::<CreateSubnetPayload>(payload_bytes),
```
lmuntaner, line 73: Could you add where do you get this list from?

I guess that more cases might come up and we will want to update this statement.

rs/src/proposals/payloads/nns_function_23.rs
```diff
@@ -0,0 +1,10 @@
+pub fn payload_44892() -> Vec<u8> {
```
lmuntaner, line 1: Maybe move these files inside a test folder along with the test?

`rs/src/proposals/tests/payloads/*` and `rs/src/proposals/tests/deserialization_test.rs`?

lmuntaner (commented):
Just a few comments. I will leave it to @bitdivine to approve and check the Rust code.

rs/src/proposals/payloads/nns_function_23.rs
hpeebles: Given that there is only a single test I didn't think it was worth creating a `deserialization_test.rs`, we can split in the future if more are added.

lmuntaner: it was more for the payloads than the test file. But yes, it can be done later.

rs/src/proposals.rs
lmuntaner: btw, does this mean we need to keep this list up to date when they add a new proposal type?

hpeebles: Yes, I just added 3 more proposal types.
It doesn't happen very often though.

lmuntaner: noted, thanks!

bitdivine:
@hpeebles Thank you for this PR.  Regarding the proposed cache:
- Is the space that it will use bounded?  Is there a risk of many large proposals taking up too much memory and breaking the nns-dapp?
- How would you expire old values from the cache?
- Will the cache be retained when the dapp is redeployed?

hpeebles:
> @hpeebles Thank you for this PR. Regarding the proposed cache:
> 
> * Is the space that it will use bounded?  Is there a risk of many large proposals taking up too much memory and breaking the nns-dapp?
> * How would you expire old values from the cache?
> * Will the cache be retained when the dapp is redeployed?

The cache is purposely not persisted between upgrades.
This is the simplest option in terms of the code, this handles cache expiration, and it means that if a field is added to a proposal before the NNS Dapp is updated, after the NNS Dapp is updated the field will appear correctly.
If we were to persist the cache at all, then we would need to build in additional logic to handle cases where we've cached incomplete proposals.

hpeebles:
PR updated to include a size limit.

Latest version is deployed here - https://uywlp-pqaaa-aaaaa-aaa4q-cai.nnsdapp.dfinity.network/v2/

rs/src/proposals.rs
```diff
@@ -0,0 +1,385 @@
...
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use std::cell::RefCell;
+use std::collections::BTreeMap;
+use std::fmt::Debug;
+use std::ops::DerefMut;
+
+type Json = String;
+
+const CACHE_SIZE_LIMIT: usize = 500;
```
bitdivine, line 25: Thank you.  What is the limit on the size of a proposal?  2M?

bitdivine: For the sake of speed it might make sense to make the cache just say 50, which is enough for most users, and that drops the size down to a level where it's unlikely to cause a problem.  We then buy ourselves time to test how big we can actually make this.  Would that work for you, @hpeebles ?

Sorry to be super-cautious but this would be one way of giving us a really bad day with no warning.

rs/src/proposals.rs
```diff
@@ -0,0 +1,385 @@
...
+            Ok(None) => Err("Proposal not found".to_string()), // We shouldn't cache this as the proposal may simply not exist yet
+            Err(error) => Err(error), // We shouldn't cache this as the error may just be transient
+        }
+    }
+}
+
+fn insert_into_cache(cache: &mut BTreeMap<u64, Json>, proposal_id: u64, payload_json: String) {
+    cache.insert(proposal_id, payload_json);
+
+    if cache.len() > CACHE_SIZE_LIMIT {
```
bitdivine, line 52: If we are unlucky, won't this delete the entry we just added?  Maybe delete first, then add?

rs/src/proposals.rs
hpeebles: The proposal payloads stored in the cache are really small. The main point of this work is to replace WASMs with their hashes. So although a proposal payload may contain a 3MB WASM, what we store in the cache will only be a few KB.

rs/src/proposals.rs
hpeebles: Yeah good plan.

hpeebles: Although that old proposal will get booted the next time a proposal is loaded in... I still think it's marginally better though.

hpeebles: Done.

rs/src/proposals.rs
hpeebles: I've reduced the limit to 100.

rs/src/proposals.rs
```diff
@@ -0,0 +1,385 @@
...
+fn process_proposal_payload(proposal_info: ProposalInfo) -> Json {
+    if let Some(Action::ExecuteNnsFunction(f)) = proposal_info.proposal.as_ref().map(|p| p.action.as_ref()).flatten() {
+        transform_payload_to_json(f.nns_function, &f.payload)
+            .unwrap_or_else(|_| "Unable to deserialize payload".to_string())
+    } else {
+        "Proposal has no payload".to_string()
+    }
+}
+
+fn transform_payload_to_json(nns_function: i32, payload_bytes: &[u8]) -> Result<String, String> {
```
bitdivine, line 70: Please check whether the JSON is < 2Mb.  If it is too large, it won't fit in a response.

bitdivine (approved):
Approving, but please address the comment regarding JSON size before merging.

Note:  I am still fundamentally opposed to adding yet more workarounds for proposal rendering to the nns-dapp, when there is already an agreed-upon, clean strategy whereby the canister that accepts the proposals has to make sure that there is a useful format available for reviews.  The workarounds consume a lot of time and add new, exciting ways in which the dapp can break.  Why are we building additional traps for ourselves?

hpeebles:
Screenshot 2022-06-23 at 11 06 21

PR #966, opened 2022-06-08
fix: various i18n typo

peterpeterparker:
# Motivation

Fix typo.


mstrasinskis (commented):
A couple of google suggestions to take into consideration:
```
Sorry, non of your principals is the controller of the neuron..
       none

Two neurons with same id can't be merged.
Two neurons with the same id can't be merged.

You cannot merge neurons controlled by hotkey.
You cannot merge neurons controlled by a hotkey.

created after proposal
created after the proposal

Principal returned by device does not match public key
Principal returned by the device does not match the public key
```

peterpeterparker:
Nice @mstrasinskis 👍. Fixed your suggested typo too in this PR.

lmuntaner (approved):
LGTM

PR #973, opened 2022-06-09
Start enabling FireFox e2e tests

bitdivine:
# Motivation
We would like to test all e2e tests in FireFox.  Unfortunately many existing tests that work in Chrome break on FireFox.

# Changes
* Enable all browsers by default but limit each test to the browsers that it is known to be working on reliably.

# Tests
Skipping firefox tests:
```
WDIO_BROWSER=firefox npm test
```
skips all tests apart from `/specs/home.e2e.ts`.

making sure that existing chrome tests are not skipped:
```
WDIO_BROWSER=all npm test
```
is still running but I can see that the Chrome tests are executing.  We should see a complete log in CI.


lmuntaner (approved):
Looks good to me, minor comments to make it a little bit more readable.

And how about adding in `/specs/home.e2e.ts`.

```
const compatibleBrowsers = ["chrome", "firefox"];

// same code
```

So that all the tests have the same pattern?

e2e-tests/specs/multi-tab-auth.e2e.ts
```diff
@@ -12,7 +13,15 @@ describe("multi-tab-auth", () => {
   const nnsTabs: Array<string> = [];
   let navigator;
 
-  before(() => {
+  before(function () {
+    if (
+      !["chrome"].includes(
```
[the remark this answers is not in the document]
bitdivine: Yes.  Actually, I was thinking that these lines would be very temporary but it is probably worth making this a function just for readability.  I'll do that now.

e2e-tests/specs/home.e2e.ts
```diff
@@ -1,8 +1,13 @@
 import { register } from "../common/register";
 import { waitForImages } from "../common/waitForImages";
 import { waitForLoad } from "../common/waitForLoad";
+import { skipUnlessBrowserIs } from "../common/test";
 
 describe("landing page", () => {
+  before(function () {
```
lmuntaner, line 7: Just out of curiosity,

Did you try?

```
before(skipUnlessBrowserIs.bind(this, ["chrome", "firefox"]);
```

lmuntaner (approved):
It looks even better, thanks!

Just a minor curiosity comment.

e2e-tests/specs/home.e2e.ts
bitdivine: That should be apply, no?  Or an additional () as with: .bind(thing, args)().  If I recall correctly apply(thing, args) = bind(thing).call(args).

bitdivine: But I have tested this. :-)

bitdivine: Oh I see, cool, yes, that would work.

bitdivine: Will loop back to this.

bitdivine: Or merge and loop back when I have time.

lmuntaner: apply calls the function. Maybe you could use it instead of `.bind()()`, but a matter of flavours.

Not a strong opinion, I was just wondering.

Thanks for trying.

PR #974, opened 2022-06-09
L2-601: Change X for IconClose

lmuntaner:
# Motivation

Use IconClose instead of letter X for buttons that remove delete a resource.

# Changes

* Change the X in two buttons for IconClose
* Add size prop to IconClose.

# Tests

Only UI changes, no extra tests added.


lmuntaner:
New UI

Screenshot 2022-06-09 at 14 11 15
Screenshot 2022-06-09 at 14 15 17

Old UI

Screenshot 2022-06-09 at 14 31 34
Screenshot 2022-06-09 at 14 31 48

PR #975, opened 2022-06-10
L2-615: Manage Canister Errors

lmuntaner:
# Motivation

User has appropriate feedback when something goes wrong in the Canisters tab.

# Changes

* nns-dapp canister throws different error classes depending on response when attaching or detaching canisters.
* canister services check errors in the catch and show errors accordingly.

# Tests

* Test that nns-dapp canister throws errors according to response.


peterpeterparker (approved):
LGTM, thx

PR #976, opened 2022-06-11
L2 659 split neuron button placement

wackyleo459:
# Motivation for NeuronMetaInfoCard

When viewport gets smaller in width: the Split Neuron button is placed awkwardly at the end. Also other buttons (Increase Dissolve Delay, Start Dissolving) align left because of flexwrap.

`.only-buttons` has justify-content: space-between. But this seems unnecessary.

# Changes
- made buttons align to right no matter the viewport.

Screen Shot 2022-06-02 at 12 50 43 AM

https://user-images.githubusercontent.com/38238542/173206138-7977af7c-e13e-46af-a96d-0d56f6df92f5.mov


peterpeterparker (approved):
LGTM, thanks

PR #978, opened 2022-06-13
fix: 'await' has no effect on the type of this expression

peterpeterparker:
# Motivation

Fix lint warning:

> Hint: 'await' has no effect on the type of this expression. 
>    const identity = await getLedgerIdentityProxy(hardwareWallet?.identifier);
>    const governance = await GenesisTokenCanister.create({
>      agent: await createAgent({ identity, host: HOST }),
>    });


lmuntaner (approved):
Thanks!

PR #979, opened 2022-06-13
docs: fix JSDoc '@param' tag has name 'settings', but there is no parameter with that name

peterpeterparker:
# Motivation

Fix lint warning:

> Hint: JSDoc '@param' tag has name 'settings', but there is no parameter with that name. 
>    * @param {Principal} canisterId
>    * @param {CanisterSettings} settings
>    * @returns Promise<void>


lmuntaner (approved):
Thanks!

lmuntaner (approved):
Thanks!

PR #981, opened 2022-06-13
feat: display account icp detailed value in a tooltip

peterpeterparker:
# Motivation

On "Wallet" page only, display account ICP value in a detailed way and in a tooltip.

# Changes

- refactor wallet summary to dedicated `<WalletSummary />` component
- use `<Tooltip />` to display detailed ICP value
- style tootlip to `inline-block` to align tooltip correctly on column display

# Screenshots

Capture d’écran 2022-06-13 à 08 35 05
Capture d’écran 2022-06-13 à 08 35 16
Capture d’écran 2022-06-13 à 08 35 30

### DIFF
--- a/frontend/svelte/src/lib/components/accounts/WalletSummary.svelte
+++ b/frontend/svelte/src/lib/components/accounts/WalletSummary.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import { accountName as getAccountName } from "../../utils/transactions.utils";
   import { i18n } from "../../stores/i18n";
-  import ICP from "../ic/ICP.svelte";
+  import ICPComponent from "../ic/ICP.svelte";
   import Identifier from "../ui/Identifier.svelte";
   import { isAccountHardwareWallet } from "../../utils/accounts.utils";
   import { getContext } from "svelte";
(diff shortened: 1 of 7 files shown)


frontend/svelte/src/lib/components/accounts/WalletSummary.svelte
```diff
@@ -0,0 +1,94 @@
...
+  let detailedICP: string;
+  $: detailedICP = formatICP({
+    value: accountBalance.toE8s(),
+    detailed: true,
+  });
+</script>
+
+<div class="title">
+  <h1>{accountName}</h1>
+  <Tooltip
```
lmuntaner, line 39: Maybe not in this PR, but it might be better if this is a component that we can use to wrap any ICP easily:

```
<TooltipICP icp={accountBalance}>
  <ICP ... />
</TooltipICP>
```

The text could be generic then: "2.312314142 ICP"

lmuntaner (approved):
LGTM, just a comment for later.

I guess we could also add this in the Neurons Tab.

frontend/svelte/src/lib/components/accounts/WalletSummary.svelte
peterpeterparker: Yeah agree but since it was explicitly decided that we should add the tooltip just there, I thought I should be specific. If we ever change our mind and use it elsewhere, then sure generic

mstrasinskis: Or make it as feature of ICPComponent :D

mstrasinskis (approved):
LGTM

PR #982, opened 2022-06-13
feat: theming support

peterpeterparker:
# Motivation

Make NNS-dapp support theming.

# Note

Pattern and code copied / applied from [Papyrs](https://github.com/papyrs/papyrs).

# Changes

- new `theme.store`
- add `theme.utils` to set `theme` attribute on root HTML element and update `meta theme-color` tag
- add `script` in `index.html` to set above attribute as fast as #possible

# Tests

Basic testing. It will need "screenshots tests" if another theme will ever be implemented.

# Screenshots

ezgif com-gif-maker (3)


frontend/svelte/src/lib/stores/theme.store.ts
```diff
@@ -0,0 +1,23 @@
+import { writable } from "svelte/store";
+import { Theme } from "../types/theme";
+import { applyTheme } from "../utils/theme.utils";
+
+const { theme: storageTheme }: Storage = localStorage;
```
mstrasinskis, line 5: Nice!

frontend/svelte/src/lib/stores/theme.store.ts
```diff
@@ -0,0 +1,23 @@
...
+    subscribe,
+
+    select: (theme: Theme) => {
+      applyTheme(theme);
+      set(theme);
+    },
+  };
+};
+
+export const theme = initTheme();
```
mstrasinskis, line 23: Just to have the same pattern: `themeStore`?

frontend/svelte/src/lib/utils/theme.utils.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const applyTheme = (theme: Theme) => {
+  const { documentElement, head } = document;
+
+  documentElement.setAttribute("theme", theme);
+
+  const color: string = getComputedStyle(documentElement).getPropertyValue(
+    "--gray-50-background"
+  );
+  head?.children
```
mstrasinskis, line 11: nth: a short comment why this is done.

frontend/svelte/src/lib/stores/theme.store.ts
```diff
@@ -0,0 +1,23 @@
...
+applyTheme(initialTheme);
+
+export const initTheme = () => {
+  const { subscribe, set } = writable<Theme>(initialTheme);
+
+  return {
+    subscribe,
+
+    select: (theme: Theme) => {
+      applyTheme(theme);
```
mstrasinskis, line 17: Just because of this call `applyTheme` and the fact the store is not used (yet) it feels more like a service (or an util function). But not a strong opinion, I don't know the future plans.

mstrasinskis (approved):
LGTM. Nice feature 👍

frontend/svelte/src/lib/utils/theme.utils.ts
peterpeterparker: Ultimately `--gray-50-background` will be replaced with something like `--background` or something more generic. So with this code we can get the color to apply to the `theme-color` without the need to hardcode and duplicate the color values in JS. Does it answer your question?

mstrasinskis: Sorry, I mean something like "Update theme-color meta for mobile devices". But probably it's too obvious.

mstrasinskis: Thank you :)

mstrasinskis (approved):

PR #983, opened 2022-06-13
L2-679: Improve merge maturity error with HW

lmuntaner:
# Motivation

User sees proper error when merging maturity of a HW neuron with old version.

# Changes

* Add ledger service "assertLedgerVersion" that throws if Ledger version is less than minVersion.
* Use ledger service in "mergeMaturity" service.
* New util "smallerVersion" that returns boolean comparing two versions.
* Manage LedgerErrorMessage type errors separately in error mapping.

# Tests

* Test for new util.
* Test for new ledger service.

### DIFF
--- a/frontend/svelte/src/lib/services/ledger.services.ts
+++ b/frontend/svelte/src/lib/services/ledger.services.ts
@@ -185,8 +185,8 @@ export const assertLedgerVersion = async ({
     return;
   }
 
-  const response = await identity.getVersion();
-  const currentVersion = `${response.major}.${response.minor}.${response.patch}`;
+  const { major, minor, patch } = await identity.getVersion();
+  const currentVersion = `${major}.${minor}.${patch}`;
   if (smallerVersion({ currentVersion, minVersion })) {
     const labels = get(i18n);
     throw new LedgerErrorMessage(
--- a/frontend/svelte/src/lib/services/ledger.services.ts
+++ b/frontend/svelte/src/lib/services/ledger.services.ts
@@ -11,6 +11,7 @@ import { LedgerErrorKey, LedgerErrorMessage } from "../types/ledger.errors";
 import { hashCode, logWithTimestamp } from "../utils/dev.utils";
 import { toToastError } from "../utils/error.utils";
 import { replacePlaceholders } from "../utils/i18n.utils";
+import { isSmallerVersion } from "../utils/utils";
 import { syncAccounts } from "./accounts.services";
 import { getIdentity } from "./auth.services";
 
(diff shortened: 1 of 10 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/173322156-f9a532f0-e5fb-41af-be69-7ba3532d2e2c.mov

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -152,3 +152,52 @@ export const mapPromises = async <T, R>(
...
+ * @returns boolean
+ */
+export const smallerVersion = ({
+  minVersion,
+  currentVersion,
+}: {
+  minVersion: string;
+  currentVersion: string;
+}): boolean => {
+  const [minMajor, minMinor, minPatch] = addZeros(
```
peterpeterparker, line 177: did you try `localeCompare`(https://stackoverflow.com/a/65687141/5404186) ?

frontend/svelte/src/lib/services/ledger.services.ts
```diff
@@ -171,3 +172,28 @@ export const listNeuronsHardwareWallet = async (): Promise<{
...
+}: {
+  identity: Identity | LedgerIdentity;
+  minVersion: string;
+}): Promise<void> => {
+  // Ignore when identity not LedgerIdentity
+  if (!(identity instanceof LedgerIdentity)) {
+    return;
+  }
+
+  const response = await identity.getVersion();
```
peterpeterparker, line 188: nth: `const {major, minor, patch} = await identity.getVersion();`

frontend/svelte/src/lib/utils/utils.ts
lmuntaner: Looks cool! I need to standarize the number of parts, but then I can use it!

frontend/svelte/src/lib/services/ledger.services.ts
lmuntaner: yep

lmuntaner: done

frontend/svelte/src/lib/utils/utils.ts
lmuntaner: done

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -686,6 +687,8 @@ export const mergeMaturity = async ({
       neuronId
     );
 
+    await assertLedgerVersion({ identity, minVersion: "2.0.6" });
```
peterpeterparker, line 690: should "2.0.6" be a constant?

lmuntaner: Sure, but the min version might be different per functionality.

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -152,3 +152,43 @@ export const mapPromises = async <T, R>(
 };
 
 export const isArrayEmpty = <T>({ length }: T[]): boolean => length === 0;
+
+const AMOUNT_VERSION_PARTS = 3;
+const addZeros = (nums: number[], amountZeros: number): number[] => {
+  const newNumbers = [...nums];
+  while (newNumbers.length < amountZeros) {
```
peterpeterparker, line 159: maybe it is possible with es6 without while, something like

```
const addZeros = (nums: number[], amountZeros: number): number[] => [...nums, ...[...Array(amountZeros - nums.length).keys()].map(_ => 0)]
```

what you think?

did not try maybe there is even a more "es6 like" solution?

lmuntaner: sure, but I will need to do a check before so that we don't fall in `new Array(-1)`

lmuntaner: done

peterpeterparker: ah yeah good point

PR #984, opened 2022-06-14
feat: menu

peterpeterparker:
# Motivation

The current tab based navigation reaches its limit as we cannot add new navigation options because of the lack of spacing. In addition, UI wise, displaying constantly the tabs uses space. Therefore this PR replaces the tabbar with a new menu based navigation.

# Notes

- UX behavior (menu appearing over content - i.e. not pushing content etc.) has been defined with Mischa.
- A split pane that should make the menu sticky on desktop and a review of the navbar(s) will be implemented separately

# Changes

- new `<Menu />` and `<MenuItem />` ui components
- new `<Menu />` component for the implementation of the menu within NNS-dapp
- update `<Nav />` to use above commponents
- backdrop refactored to new component `<Backdrop />` so that modals and menu can share the same backdrop effect
- `<GetIcp />` move to menu
- some new icons
- tabbar visually remoed
- tab "ICP" renamed in "Accounts"

# Screenshots

gif

Capture d’écran 2022-06-18 à 10 32 10
Capture d’écran 2022-06-18 à 10 32 17
Capture d’écran 2022-06-18 à 10 32 21
Capture d’écran 2022-06-18 à 10 32 31


frontend/svelte/src/lib/i18n/en.json
```diff
@@ -92,12 +92,13 @@
     "auth_sign_out": "You have been logged out because your session has expired."
   },
   "navigation": {
-    "icp": "ICP",
+    "accounts": "Accounts",
```
lmuntaner, line 95: 👏

frontend/svelte/src/lib/components/ui/Backdrop.svelte
```diff
@@ -0,0 +1,35 @@
...
+  import { createEventDispatcher } from "svelte";
+
+  export let disabled = false;
+
+  const dispatch = createEventDispatcher();
+  const close = () => dispatch("nnsClose");
+</script>
+
+<div
+  in:fade={{ duration: 75 }}
```
lmuntaner, line 12: Maybe add it as a constant

frontend/svelte/src/lib/components/ui/Backdrop.svelte
```diff
@@ -0,0 +1,35 @@
...
+  out:fade={{ duration: 250 }}
```
lmuntaner, line 13: ditto

lmuntaner (commented):
Only a few comments, but I like it! Thanks!

frontend/svelte/src/lib/components/ui/Backdrop.svelte
peterpeterparker: sure, done

frontend/svelte/src/lib/components/ui/Backdrop.svelte
```diff
@@ -0,0 +1,38 @@
...
+  class="backdrop"
+  on:click|stopPropagation={close}
+  class:disabledActions={disabled}
+/>
+
+<style lang="scss">
+  @use "../../themes/mixins/interaction";
+
+  .backdrop {
+    position: absolute;
```
mstrasinskis, line 26: Wouldn't `position: fixed` be more reliable?

frontend/svelte/src/lib/components/ui/Backdrop.svelte
```diff
@@ -0,0 +1,38 @@
+<script lang="ts">
+  import { fade } from "svelte/transition";
+  import { createEventDispatcher } from "svelte";
+
+  export let disabled = false;
```
mstrasinskis, line 5: `disablePointerEvents` or something more self explanatory would be great.

mstrasinskis (commented):
Looks great! A couple of minor comments.

frontend/svelte/src/lib/components/ui/Backdrop.svelte
peterpeterparker: Since parent that uses the `<Backdrop />` has to be `fixed`, `absolute` is probably fine. Also not new, this code was extracted / moved from `<Modal />` therefore kind if feel it's ok to leave it as it is.

frontend/svelte/src/lib/components/ui/Backdrop.svelte
peterpeterparker: sure. done.

mstrasinskis (approved):
LGTM, thank you.

PR #985, opened 2022-06-14
Fix: Round to 8 decimals when Selecting Cycles

lmuntaner:
# Motivation

Round to eight decimal places while the user selects ICP or Cycles to avoid weird behavior with Input type="icp".

# Changes

* After converting ICP or Cycles to the other, round to only 8 decimals.

# Tests

UI Change.

### DIFF
--- a/frontend/svelte/src/lib/components/canisters/SelectCyclesCanister.svelte
+++ b/frontend/svelte/src/lib/components/canisters/SelectCyclesCanister.svelte
@@ -21,19 +21,25 @@
   const setCycles = () =>
     (amountCycles =
       amount !== undefined && icpToCyclesExchangeRate !== undefined
-        ? convertIcpToTCycles({
-            icpNumber: amount,
-            exchangeRate: icpToCyclesExchangeRate,
-          })
+        ? // ICP Input misbehaves with more than eight decimals
+          Number(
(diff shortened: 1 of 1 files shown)


frontend/svelte/src/lib/components/canisters/SelectCyclesCanister.svelte
```diff
@@ -21,19 +21,25 @@
   const setCycles = () =>
     (amountCycles =
       amount !== undefined && icpToCyclesExchangeRate !== undefined
-        ? convertIcpToTCycles({
-            icpNumber: amount,
-            exchangeRate: icpToCyclesExchangeRate,
-          })
+        ? // ICP Input misbehaves with more than eight decimals
+          Number(
```
peterpeterparker, line 25: shouldn't this find place within `convertIcpToTCycles`?

lmuntaner: I think that is specific to how the Input type="icp" works, not necessarily we want to restrict the conversion to eight decimals.

peterpeterparker: Thanks for the feedback

PR #987, opened 2022-06-15
Move the canisters tab to svelte

bitdivine:
# Motivation
We would like to release the svelte version of the canisters tab.

# Changes
* Update REDIRECT_TO_LEGACY so that prod and staging use the svelte version of the canisters tab.
* Update documentation.
* Enable testnet population for "prod" builds; this requires svelte.

# Tests
- See CI.
- Also, I will test this with the redirect e2e test against a deployment to the nns-dapp testnet.....  Testing...done. OK

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -220,8 +220,7 @@ if [[ "$POPULATE" == "true" ]]; then
   # Create users and neurons
   # Note: Cannot be used with flutter.
   REDIRECT_TO_LEGACY="$(jq -re .REDIRECT_TO_LEGACY frontend/ts/src/config.json)"
-  [[ "$REDIRECT_TO_LEGACY" == "prod" ]] ||
-    [[ "$REDIRECT_TO_LEGACY" == "flutter" ]] || {
+  [[ "$REDIRECT_TO_LEGACY" == "flutter" ]] || {
     echo Creating users and neurons...
     pushd e2e-tests
     npm ci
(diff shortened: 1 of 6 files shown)


e2e-tests/specs/redirect-to-legacy.e2e.ts
```diff
@@ -11,26 +11,26 @@ import { skipUnlessBrowserIs } from "../common/test";
 const REDIRECTS = {
```
bitdivine, line 11: Note: The e2e tests are by default run on "both" only.  I would like to run this against "prod" on the testnet before we release, so I have updated the test.

deploy.sh
```diff
@@ -220,8 +220,7 @@ if [[ "$POPULATE" == "true" ]]; then
   # Create users and neurons
   # Note: Cannot be used with flutter.
```
bitdivine, line 221: This `--populate` command works only in svelte, which all tabs now are with a prod build.

frontend/dart/lib/data/env.dart
```diff
@@ -2,4 +2,4 @@ const REDIRECT_TO_LEGACY = String.fromEnvironment('REDIRECT_TO_LEGACY');
 bool showAccountsRoute() { return ["flutter", "both"].contains(REDIRECT_TO_LEGACY); }
```
bitdivine, line 2: Remove the canisters tab from flutter....

frontend/svelte/src/lib/constants/routes.constants.ts
```diff
@@ -29,9 +29,12 @@ export const SHOW_PROPOSALS_ROUTE = [
   "prod",
```
bitdivine, line 29: ... and add it to svelte.

peterpeterparker (approved):
Rubber stamp, thanks

PR #989, opened 2022-06-15
style: redo overflow-x on headless layout

peterpeterparker:
# Motivation

"Neuron" detail page overflows again on mobile devices. Therefore we should redo the `overflow-x` on the headless layout even though we have improved the positioned of the tooltip that leads to this styling issue.


lmuntaner (approved):
thanks!

PR #990, opened 2022-06-15
Fix: Revert Tooltip and add Padding

lmuntaner:
# Motivation

User sees NeuronCard of the same size when merging neurons.

# Changes

* Revert `display: inline-block;` in the Tooltip.
* Add padding bottom to merge neurons first screen to keep consistency.

# Tests

No test, UI fixes.


peterpeterparker:
tooltip solved in https://github.com/dfinity/nns-dapp/pull/992

PR #991, opened 2022-06-15
style: adjust button line-height to support 2 lines of text in big bu…

mstrasinskis:
# Motivation

[Different Button Height on Safari iPhone 13 ](https://dfinity.atlassian.net/browse/L2-709)

# Changes

- line-height is adjusted to support 2 lines within normal height.

# Tests

## before
image

## after
image

### small buttons should look the same
image

### DIFF
--- a/frontend/svelte/src/lib/themes/button.scss
+++ b/frontend/svelte/src/lib/themes/button.scss
@@ -51,7 +51,7 @@ button {
     min-height: var(--button-min-height);
 
     font-size: var(--font-size-small);
-    line-height: 1.25;
+    line-height: var(--line-height-standard);
 
     @include media.min-width(medium) {
       font-size: var(--font-size-h3);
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
LGTM.

Also in Modals it appears well I asume, right?

frontend/svelte/src/lib/themes/button.scss
```diff
@@ -51,6 +51,7 @@ button {
     min-height: var(--button-min-height);
 
     font-size: var(--font-size-small);
+    line-height: 1.25;
```
peterpeterparker, line 54: `--line-height-standard` is 1.3, that wouldn't work? just would rather like not to have an "hardcoded" value

peterpeterparker (approved):
Thanks

mstrasinskis:
> Also in Modals it appears well I asume, right?
@lmuntaner I didn't find any issue.

PR #992, opened 2022-06-15
fix: tooltip breaking list of cards width

peterpeterparker:
# Motivation

Tooltip `display` property breaks list of cards ("Merge neurons") width if displayed

# Changes

- make `display` property optional

# Screenshots

screenshot_2022-06-15_at_12 06 05

Capture d’écran 2022-06-15 à 12 24 28
Capture d’écran 2022-06-15 à 12 24 31
Capture d’écran 2022-06-15 à 12 24 41


lmuntaner (approved):
Thanks

PR #996, opened 2022-06-15
docs: fix there is no parameter with that name

peterpeterparker:
# Motivation

Docs fix

> Hint: JSDoc '@param' tag has name 'currentVersion', but there is no parameter with that name. 
>  * @param minVersion Ex: "1.0.0"
>  * @param currentVersion Ex: "2.0.0"
>  * @returns boolean


lmuntaner (approved):
thanks

PR #999, opened 2022-06-16
L2-718: Do not check neuron balance always

lmuntaner:
# Motivation

Do not check neuron balance every time we load a neuron.

# Changes

* Remove logic to check the neuron balance and claim the neuron if it doesn't match the stake.
* Call claimOrRefresh only on "reloadNeuron".

# Tests

* Remove test cases.
* Add test case for reloadNeuron.


lmuntaner:

peterpeterparker (approved):
LGTM, thanks!

PR #1000, opened 2022-06-16
Handle new flows for creating and topping up canisters

hpeebles:
# Motivation

In the Svelte version of the NNS Dapp, creating and topping up canisters is optimistically handled on the frontend, this makes the end to end process happen faster. However, this is a 3 step process and parts of it may fail. So this PR changes the backend code to process transactions which the NNS Dapp canister detects are create / top up canister transactions, ensuring that even if the frontend fails, the canister will still be created or topped up and linked to the user's account.

# Changes

- Detect create / top up canister transactions of the new type (these are sent to an account controlled by the CMC rather than controlled by the NNS Dapp)
- Process these transactions to create or top up the canisters

# Tests

This has been manually tested by myself and @lmuntaner at this url - https://uywlp-pqaaa-aaaaa-aaa4q-cai.nnsdapp.dfinity.network/

PR #1003, opened 2022-06-16
Chore: Handle Multiple Request to CMC

lmuntaner:
# Motivation

CMC notify calls might fail because the nns dapp canister also makes the calls. Ignore the error if it fails because of that.

# Changes

* Poll notifyCreateCanister or notifyTopUpCanister from CMCanister in canister api createCanister and topUpCanister.
* Add new util "poll" to poll any function.

# Tests

* Add test cases in Canister api functions.
* New test for util "poll".

### DIFF
--- a/frontend/svelte/src/lib/api/canisters.api.ts
+++ b/frontend/svelte/src/lib/api/canisters.api.ts
@@ -145,7 +145,7 @@ const pollNotifyCreateCanister = async ({
           controller,
           block_index: blockHeight,
         }),
-      shouldRecall: (error: Error) => error instanceof ProcessingError,
+      shouldExit: (error: Error) => !(error instanceof ProcessingError),
     });
   } catch (error) {
     if (error instanceof PollingLimitExceededError) {
@@ -240,7 +240,7 @@ const pollNotifyTopUpCanister = async ({
           canister_id: canisterId,
           block_index: blockHeight,
         }),
-      shouldRecall: (error: Error) => error instanceof ProcessingError,
+      shouldExit: (error: Error) => !(error instanceof ProcessingError),
     });
   } catch (error) {
     if (error instanceof PollingLimitExceededError) {
--- a/frontend/svelte/src/lib/api/canisters.api.ts
+++ b/frontend/svelte/src/lib/api/canisters.api.ts
@@ -200,7 +200,7 @@ export const createCanister = async ({
   // If this fails or the client loses connection, nns dapp backend polls the transactions
   // and will also notify to CMC the transaction if it's pending.
   // If the backend is faster to notify, we might get a ProcessingError.
-  // We poll CMC until we stop getting ProcessingError or we request more than MAX_POLLING_TIMES
+  // We poll CMC until we stop getting ProcessingError or we request more than DEFAUL_MAX_POLLING_ATTEMPTS
   const canisterId = await pollNotifyCreateCanister({
     cmc,
     controller: principal,
--- a/frontend/svelte/src/lib/api/canisters.api.ts
+++ b/frontend/svelte/src/lib/api/canisters.api.ts
@@ -128,6 +128,11 @@ export const detachCanister = async ({
   logWithTimestamp("Detaching canister call complete.");
 };
 
+const pollingLimit = (error: Error): boolean =>
+  error instanceof PollingLimitExceededError;
+const notProcessingError = (error: Error): boolean =>
+  !(error instanceof ProcessingError);
+
 // Polls CMC waiting for a reponse that is not a ProcessingError.
 const pollNotifyCreateCanister = async ({
   cmc,
@@ -140,15 +145,16 @@ const pollNotifyCreateCanister = async ({
 }): Promise<Principal> => {
   try {
     return await poll({
-      fn: () =>
+      fn: (): Promise<Principal> =>
         cmc.notifyCreateCanister({
           controller,
           block_index: blockHeight,
         }),
-      shouldExit: (error: Error) => !(error instanceof ProcessingError),
+      shouldExit: notProcessingError,
     });
   } catch (error) {
-    if (error instanceof PollingLimitExceededError) {
+    if (pollingLimit(error)) {
+      // TODO: i18n errors https://dfinity.atlassian.net/browse/L2-721
       throw new CMCError(
         "Error creating canister. Canister might be created, refresh the page. Try again if not."
       );
@@ -235,15 +241,16 @@ const pollNotifyTopUpCanister = async ({
 }): Promise<Cycles> => {
   try {
     return await poll({
-      fn: () =>
+      fn: (): Promise<Cycles> =>
         cmc.notifyTopUp({
           canister_id: canisterId,
           block_index: blockHeight,
         }),
-      shouldExit: (error: Error) => !(error instanceof ProcessingError),
+      shouldExit: notProcessingError,
     });
   } catch (error) {
-    if (error instanceof PollingLimitExceededError) {
+    if (pollingLimit(error)) {
+      // TODO: i18n errors https://dfinity.atlassian.net/browse/L2-721
       throw new CMCError(
         "Error topping up canister. ICP might have been transferred, refresh the page. Try again if not."
       );
--- a/frontend/svelte/src/lib/api/canisters.api.ts
+++ b/frontend/svelte/src/lib/api/canisters.api.ts
@@ -2,6 +2,8 @@ import type { Identity } from "@dfinity/agent";
 import { AccountIdentifier, ICP, SubAccount } from "@dfinity/nns";
 import type { Principal } from "@dfinity/principal";
 import { CMCCanister } from "../canisters/cmc/cmc.canister";
+import { CMCError, ProcessingError } from "../canisters/cmc/cmc.errors";
+import type { Cycles } from "../canisters/cmc/cmc.types";
 import { principalToSubAccount } from "../canisters/cmc/utils";
 import { ICManagementCanister } from "../canisters/ic-management/ic-management.canister";
 import type {
@@ -9,6 +11,7 @@ import type {
   CanisterSettings,
 } from "../canisters/ic-management/ic-management.canister.types";
 import type { NNSDappCanister } from "../canisters/nns-dapp/nns-dapp.canister";
+import { CanisterAlreadyAttachedError } from "../canisters/nns-dapp/nns-dapp.errors";
 import type {
   CanisterDetails as CanisterInfo,
   SubAccountArray,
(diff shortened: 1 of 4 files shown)


lmuntaner:
This has been tested along https://github.com/dfinity/nns-dapp/pull/1000

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -189,3 +189,50 @@ export const smallerVersion = ({
...
+  maxTimes?: number;
+  counter?: number;
+}): Promise<T> => {
+  if (counter >= maxTimes) {
+    throw new PollingLimitExceededError();
+  }
+  try {
+    return await fn();
+  } catch (error) {
+    if (!shouldRecall(error)) {
```
hpeebles, line 227: I think `shouldRecall` isn't quite the right name.
Maybe something like `shouldRetry` or invert it and call the function `shouldBreak` / `shouldExit`.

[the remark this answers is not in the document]
lmuntaner: I like `shouldExit`

lmuntaner: done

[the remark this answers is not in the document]
lmuntaner: done

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -189,3 +189,50 @@ export const smallerVersion = ({
...
+  });
+
+export class PollingLimitExceededError extends Error {}
+const MAX_POLLING_TIMES = 10;
+/**
+ * Function that polls a specific function, checking error with passed argument to recall or not.
+ *
+ * @param {Object} params
+ * @param {fn} params.fn Function to call
+ * @param {shouldExit} params.shouldExit Function to check whether function shuold stop polling.
```
peterpeterparker, line 205: "shuold"

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -189,3 +189,50 @@ export const smallerVersion = ({
...
+ * @param {maxTimes} params.maxTimes Param to override the default number of times to poll.
```
peterpeterparker, line 206: counter is not documented

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -189,3 +189,50 @@ export const smallerVersion = ({
...
+ */
+export const poll = async <T>({
+  fn,
+  shouldExit,
+  maxTimes = MAX_POLLING_TIMES,
+  counter = 1,
+}: {
+  fn: () => Promise<T>;
+  shouldExit: (err: Error) => boolean;
+  maxTimes?: number;
```
peterpeterparker, line 218: `maxTimes` should not be mandatory to avoid infinite loop?

peterpeterparker, line 218: naming suggestion: instead of `maxTimes`, `maxCount` or `maxAttempts` ?

I understand "...Times" with a time meaning, then read the code and noticed that `counter` is a number - i.e. not related to time

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -189,3 +189,50 @@ export const smallerVersion = ({
...
+ * @param {shouldExit} params.shouldExit Function to check whether function shuold stop polling.
+ * @param {maxTimes} params.maxTimes Param to override the default number of times to poll.
+ *
+ * @returns
+ */
+export const poll = async <T>({
+  fn,
+  shouldExit,
+  maxTimes = MAX_POLLING_TIMES,
+  counter = 1,
```
peterpeterparker, line 214: should not start with zero?

frontend/svelte/src/lib/utils/utils.ts
peterpeterparker, line 205: `shouldExist` is only in case of error, correct? maybe add this info in the comment?

frontend/svelte/src/lib/api/canisters.api.ts
```diff
@@ -158,30 +191,67 @@ export const createCanister = async ({
...
+  blockHeight,
+  canisterId,
+}: {
+  cmc: CMCCanister;
+  canisterId: Principal;
+  blockHeight: bigint;
+  counter?: number;
+}): Promise<Cycles> => {
+  try {
+    return await poll({
```
peterpeterparker, line 237: do you need the `await` here?

frontend/svelte/src/lib/api/canisters.api.ts
```diff
@@ -158,30 +191,67 @@ export const createCanister = async ({
...
+      fn: () =>
```
peterpeterparker, line 238: can you add the return type`fn: (): Promise<Cycles> => `?

frontend/svelte/src/lib/api/canisters.api.ts
```diff
@@ -158,30 +191,67 @@ export const createCanister = async ({
...
+        cmc.notifyTopUp({
+          canister_id: canisterId,
+          block_index: blockHeight,
+        }),
+      shouldExit: (error: Error) => !(error instanceof ProcessingError),
+    });
+  } catch (error) {
+    if (error instanceof PollingLimitExceededError) {
+      throw new CMCError(
```
peterpeterparker, line 247: i18n later on?

frontend/svelte/src/lib/api/canisters.api.ts
```diff
@@ -124,6 +128,35 @@ export const detachCanister = async ({
   logWithTimestamp("Detaching canister call complete.");
 };
 
+// Polls CMC waiting for a reponse that is not a ProcessingError.
+const pollNotifyCreateCanister = async ({
```
peterpeterparker, line 132: `pollNotifyCreateCanister` code look really similar to `pollNotifyTopUpCanister`. should we / can we refactor to avoid double code?

lmuntaner: I refactored the similarities already in the `poll` util.

But I guess I can still reuse some other parts.

frontend/svelte/src/lib/api/canisters.api.ts
lmuntaner: yes, otherwise the error doesn't go into this catch.

frontend/svelte/src/lib/api/canisters.api.ts
lmuntaner: sure

frontend/svelte/src/lib/api/canisters.api.ts
lmuntaner: I don't have the i18n setup for error in canisters yet. I'll create an issue for it

frontend/svelte/src/lib/utils/utils.ts
lmuntaner: thanks!

lmuntaner: yes, I can add it.

frontend/svelte/src/lib/utils/utils.ts
lmuntaner: adding it

frontend/svelte/src/lib/utils/utils.ts
lmuntaner: This is the first thing I check, therefore it made more sense to me that the first time the counter was 1, like, it's the first call, what do I want to do?

But now thinking about it, if maxTImes is `1` it would not call even the first one. Therefore, it makes sense to start at 0.

frontend/svelte/src/lib/utils/utils.ts
lmuntaner: It has a default value. It's always present, but not mandatory in the api.

lmuntaner: I like `maxAttempts`.

lmuntaner: done

frontend/svelte/src/lib/utils/utils.ts
lmuntaner: done

frontend/svelte/src/lib/api/canisters.api.ts
lmuntaner: done

frontend/svelte/src/lib/api/canisters.api.ts
lmuntaner: done

frontend/svelte/src/lib/api/canisters.api.ts
lmuntaner: done some refactor

lmuntaner:

frontend/svelte/src/lib/utils/utils.ts
peterpeterparker: Missed it, good point, thx

frontend/svelte/src/lib/api/canisters.api.ts
peterpeterparker: early morning...thanks

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -189,3 +189,51 @@ export const smallerVersion = ({
...
   );
 };
+
+const waitForMilliseconds = (milliseconds: number): Promise<void> =>
+  new Promise((resolve) => {
+    setTimeout(resolve, milliseconds);
+  });
+
+export class PollingLimitExceededError extends Error {}
+const MAX_POLLING_TIMES = 10;
```
peterpeterparker, line 199: `MAX_POLLING_ATTEMPTS` to reflect refactored param

lmuntaner: yep

peterpeterparker (approved):
LGTM, thanks!

As it is a new pattern, before merge, I would be happy if @mstrasinskis have look too

lmuntaner:
Sounds good to me, I'll ask him directly.

frontend/svelte/src/lib/utils/utils.ts
```diff
@@ -189,3 +189,51 @@ export const smallerVersion = ({
...
+    throw new PollingLimitExceededError();
+  }
+  try {
+    return await fn();
+  } catch (error) {
+    if (shouldExit(error)) {
+      throw error;
+    }
+  }
+  await waitForMilliseconds(500);
```
mstrasinskis, line 232: nth: just to make it a bit more predictable we could store the callTimestamp and then waitForMilliseconds(DEFAULT_ATTEMPT_INTERVAL_MILLISECONDS - (Date.now() - callTimestamp))

mstrasinskis (approved):
LGTM, nice approach. Just one minor suggestion.

PR #1004, opened 2022-06-17
Proposal payloads fe

mstrasinskis:
# Motivation

### proposal-payloads-fe  ➡ get_proposal_payload

Added separate call to get payloads `getProposalPayload` to support all payload types (including WASM hashes).
[nns-js related PR](https://github.com/dfinity/nns-js/pull/115)

https://user-images.githubusercontent.com/98811342/174315685-22b0ba8c-f52d-40b7-949d-3fd289b6c188.mov

# Changes

- getProposal calls listProposals to avoid raw payload loading
- all nnsFunction names in i18n / `nns_function_names`

- ++ nns-dapp canister `getProposalPayload`
- ++ proposal.api / `queryProposalPayload`
- ++ proposals.services / `getProposalPayload`
- ++ `proposalPayloadStore` to cache payloads
- ++ NnsFunctionDetails.svelte

- Json component `data-tid="json"`

# Tests

- updated governance.canister.mock
- api / queryProposalPayload
- services / getProposalPayload
- proposalPayloadStore
- getNnsFunctionId


frontend/svelte/src/lib/services/proposals.services.ts
```diff
@@ -239,6 +241,30 @@ const getProposal = async ({
   });
 };
 
+/**
+ * Return proposal payload
```
lmuntaner, line 245: Probably better "Loads proposal payload in proposalPayloadsStore"

frontend/svelte/src/lib/services/proposals.services.ts
```diff
@@ -239,6 +241,30 @@ const getProposal = async ({
...
+ */
+export const getProposalPayload = async ({
```
lmuntaner, line 247: Rename to `loadProposalPayload`?

frontend/svelte/src/lib/stores/proposals.store.ts
```diff
@@ -245,8 +251,33 @@ const initNeuronSelectStore = () => {
   };
 };
 
+const initProposalPayloadsStore = () => {
+  const map = new Map<bigint, ProposalPayload>();
+  const { subscribe, update } = writable<ProposalPayloadsStore>({
+    getPayload: (proposalId) => map.get(proposalId),
```
lmuntaner, line 257: I like where this pattern is going.

But how about when you serialized this for the debug store? Shouldn't the store just be data?

frontend/svelte/src/lib/stores/proposals.store.ts
```diff
@@ -245,8 +251,33 @@ const initNeuronSelectStore = () => {
...
+    setPayload: ({
+      proposalId,
+      payload,
+    }: {
+      proposalId: ProposalId;
+      payload: ProposalPayload;
+    }) =>
+      update((state) => {
+        map.set(proposalId, payload);
+        return state;
```
lmuntaner, line 272: Isn't it required for the data to be immutable inside a store then?

How does it know when to rerender? This is interesting...

frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -51,6 +48,16 @@ export const proposalActionFields = (
   });
 };
 
+export const getNnsFunctionId = (proposal: Proposal): number | undefined => {
```
lmuntaner, line 51: Can't we use an enum for the nnsFunctions?

frontend/svelte/src/lib/stores/proposals.store.ts
```diff
@@ -245,8 +251,33 @@ const initNeuronSelectStore = () => {
   };
 };
 
+const initProposalPayloadsStore = () => {
```
lmuntaner, line 254: maybe worth mentioning that `null` is considered no payload and `undefined` still loading.

lmuntaner (commented):
A few comments and minor changes suggested.

frontend/svelte/src/lib/stores/proposals.store.ts
mstrasinskis: I like using the Map because of performance less logic but you are right just a Map is not a first class citizen for the store. I'll rewrite it to use an object.

frontend/svelte/src/lib/stores/proposals.store.ts
lmuntaner: I like the idea of an object instead of an array as well. And the keys are the id of the payload.

frontend/svelte/src/lib/api/proposals.api.ts
```diff
@@ -76,16 +80,49 @@ export const queryProposal = async ({
   logWithTimestamp(
     `Querying Proposal (${hashCode(proposalId)}) certified:${certified} call...`
   );
+
   const governance: GovernanceCanister = GovernanceCanister.create({
     agent: await createAgent({ identity, host: HOST }),
   });
 
-  const response = await governance.getProposal({ proposalId, certified });
+  const response = await governance.listProposals({
```
peterpeterparker, line 88: would it be worth tweaking a bit the api function `queryProposals` to use this function instead of using `governance.listProposals`?

peterpeterparker, line 88: dumb question probably but `listProposals` with `limit: 1` is faster than `getProposalWithPayload` that's idea right? or the idea to use is to spare bandwith?

frontend/svelte/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
```diff
@@ -302,4 +303,22 @@ export class NNSDappCanister {
       account_identifier: accountIdentifier,
     });
   }
+
+  public async getProposalPayload({
+    proposalId,
+  }: {
+    proposalId: ProposalId;
+  }): Promise<object> {
+    // Currently works only with certifiedService
```
peterpeterparker, line 312: means the display of each proposal will be slower? is query call in works?

frontend/svelte/src/lib/api/proposals.api.ts
peterpeterparker: as discussed in a call: yes because payload can be quite large

frontend/svelte/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
peterpeterparker: as discussed in a call: the overall experience will feels faster because the proposal details will be displayed more quickly as we defer the loading of the payload

frontend/svelte/src/lib/utils/proposals.utils.ts
mstrasinskis: Not sure, because that [this list ](https://github.com/dfinity/nns-dapp/pull/1004/files#diff-874563872d9de773b78a8342e7c59752d689e51f4b910f3417fa3d6725f18697R402) needs to be maintained. Currently we use it only for mapping (i18n). But let't rename the `getNnsFunctionId` to `getNnsFunctionIndex` for readability. Works for you?

frontend/svelte/src/lib/stores/proposals.store.ts
mstrasinskis: Thanks. The comment was added to the proposal.service.

frontend/svelte/src/lib/api/proposals.api.ts
mstrasinskis: Do you suggest to remove `queryProposal` afterwards to not confuse future us?

frontend/svelte/src/lib/utils/proposals.utils.ts
lmuntaner: yes, sounds good

frontend/svelte/src/lib/api/proposals.api.ts
mstrasinskis: Good idea 👍

peterpeterparker: Sure works for me.

frontend/svelte/src/lib/stores/proposals.store.ts
```diff
@@ -40,6 +40,8 @@ export interface ProposalsStore {
   certified: boolean | undefined;
 }
 
+type ProposalPayload = object | null | undefined;
+export type ProposalPayloadsStore = { [proposalId: string]: ProposalPayload };
```
peterpeterparker, line 44: have you thought about using a `Map` instead of a `Record`?

can it help making the access to the variable without case maybe 

Instead of `$proposalPayloadsStore[`${proposalId}`];` something like `$proposalPayloadsStore.get(proposalId)`

mstrasinskis: It was originally implemented - [with a Map](https://github.com/dfinity/nns-dapp/pull/1004/commits/eada542b0fc06d89d0b7cbc576ac9193881f17aa#diff-e29cbd593d1969e4efadcf1c9cfd2185226d561242bed499c02cb516d7ed8b91R255)

Then after discussion Map was changed to the object.

@lmuntaner what do you think if we undo to the Map version but recreate the Map (e.g. `return new Map(oldMap)`) on update to make the data immutable?

lmuntaner: sounds good

lmuntaner (approved):

frontend/svelte/src/lib/stores/proposals.store.ts
mstrasinskis: Updated

lmuntaner (approved):
A couple of comments. Already approving.

frontend/svelte/src/lib/stores/proposals.store.ts
```diff
@@ -245,8 +247,45 @@ const initNeuronSelectStore = () => {
   };
 };
 
+const initProposalPayloadsStore = () => {
+  const { subscribe, update } = writable<ProposalPayloadsStore>(
+    new Map<ProposalId, ProposalPayload>()
```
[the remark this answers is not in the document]
mstrasinskis: Good catch! Fixed and tests were extended 👍

frontend/svelte/src/lib/components/proposal-detail/ProposalDetailCard/NnsFunctionDetails.svelte
```diff
@@ -0,0 +1,34 @@
...
+      proposalId,
+    });
+  }
+</script>
+
+<dt>{$i18n.proposal_detail.nns_function_name}</dt>
+<dd>{nnsFunctionName}</dd>
+{#if payload === undefined}
+  <Spinner size="small" inline={true} />
+{:else if payload !== null}
```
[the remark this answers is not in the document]
mstrasinskis: Jira ticket was created https://dfinity.atlassian.net/browse/L2-741

PR #1005, opened 2022-06-18
L2-716: Load transaction fee from Ledger Canister

lmuntaner:
# Motivation

Get the transaction fee from the Ledger Canister instead of using a hardcoded one.

# Changes

* New store for transaction fees.
* New transactionFee services to get the fee from the Ledger Canister
* New ledger api function to get the transaction fee.
* Transform some neurons constants into neuron utils.
* Read the transaction fee from the neuron utils instead of constant in components.

# Tests

* Test for the new store.
* Test for the new service.
* Test for the new ledger api function.
* Update broken tests after changes.

### DIFF
--- a/frontend/svelte/src/lib/services/neurons.services.ts
+++ b/frontend/svelte/src/lib/services/neurons.services.ts
@@ -30,7 +30,7 @@ import { getLedgerIdentityProxy } from "../proxy/ledger.services.proxy";
 import { startBusy, stopBusy } from "../stores/busy.store";
 import { definedNeuronsStore, neuronsStore } from "../stores/neurons.store";
 import { toastsStore } from "../stores/toasts.store";
-import { mainTransactionFeeNumberStore } from "../stores/transaction-fees.store";
+import { mainTransactionFeeStore } from "../stores/transaction-fees.store";
 import type { Account } from "../types/account";
 import { InsufficientAmountError } from "../types/common.errors";
 import {
@@ -501,7 +501,7 @@ export const splitNeuron = async ({
       neuronId
     );
 
-    const transactionFee = get(mainTransactionFeeNumberStore);
+    const transactionFee = get(mainTransactionFeeStore);
     const transactionFeeAmount = transactionFee / E8S_PER_ICP;
     const stake = convertNumberToICP(amount + transactionFeeAmount);
 
--- a/frontend/svelte/src/lib/services/neurons.services.ts
+++ b/frontend/svelte/src/lib/services/neurons.services.ts
@@ -23,13 +23,14 @@ import {
 } from "../api/governance.api";
 import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
 import { IS_TESTNET } from "../constants/environment.constants";
-import { E8S_PER_ICP, TRANSACTION_FEE_E8S } from "../constants/icp.constants";
+import { E8S_PER_ICP } from "../constants/icp.constants";
 import { MIN_VERSION_MERGE_MATURITY } from "../constants/neurons.constants";
 import type { LedgerIdentity } from "../identities/ledger.identity";
 import { getLedgerIdentityProxy } from "../proxy/ledger.services.proxy";
 import { startBusy, stopBusy } from "../stores/busy.store";
 import { definedNeuronsStore, neuronsStore } from "../stores/neurons.store";
 import { toastsStore } from "../stores/toasts.store";
+import { mainTransactionFeeNumberStore } from "../stores/transaction-fees.store";
 import type { Account } from "../types/account";
 import { InsufficientAmountError } from "../types/common.errors";
 import {
@@ -500,10 +501,11 @@ export const splitNeuron = async ({
       neuronId
     );
 
-    const transactionFeeAmount = TRANSACTION_FEE_E8S / E8S_PER_ICP;
+    const transactionFee = get(mainTransactionFeeNumberStore);
+    const transactionFeeAmount = transactionFee / E8S_PER_ICP;
     const stake = convertNumberToICP(amount + transactionFeeAmount);
 
-    if (!isEnoughToStakeNeuron({ stake, withTransactionFee: true })) {
+    if (!isEnoughToStakeNeuron({ stake, fee: transactionFee })) {
       throw new InsufficientAmountError();
     }
 
--- a/frontend/svelte/src/lib/services/neurons.services.ts
+++ b/frontend/svelte/src/lib/services/neurons.services.ts
@@ -501,11 +501,11 @@ export const splitNeuron = async ({
       neuronId
     );
 
-    const transactionFee = get(mainTransactionFeeStore);
-    const transactionFeeAmount = transactionFee / E8S_PER_ICP;
+    const fee = get(mainTransactionFeeStore);
+    const transactionFeeAmount = fee / E8S_PER_ICP;
     const stake = convertNumberToICP(amount + transactionFeeAmount);
 
-    if (!isEnoughToStakeNeuron({ stake, fee: transactionFee })) {
+    if (!isEnoughToStakeNeuron({ stake, fee })) {
       throw new InsufficientAmountError();
     }
 
--- a/frontend/svelte/src/lib/stores/transaction-fees.store.ts
+++ b/frontend/svelte/src/lib/stores/transaction-fees.store.ts
@@ -16,7 +16,7 @@ const defaultTransactionFees = {
  * - setMain: replace the current fee in `main`.
  */
 const initTransactionFeesStore = () => {
-  const { subscribe, update, set } = writable<TransactionFeesStore>(
+  const { subscribe, update } = writable<TransactionFeesStore>(
     defaultTransactionFees
   );
 
@@ -28,11 +28,6 @@ const initTransactionFeesStore = () => {
         ...data,
         main: fee,
       }));
-    },
-
-    // Used in the tests
-    reset() {
-      set(defaultTransactionFees);
     },
   };
 };
--- a/frontend/svelte/src/lib/stores/transaction-fees.store.ts
+++ b/frontend/svelte/src/lib/stores/transaction-fees.store.ts
@@ -1,0 +1,45 @@
+import { derived, writable } from "svelte/store";
+import { DEFAULT_TRANSACTION_FEE_E8S } from "../constants/icp.constants";
+
+export type TransactionFeesStore = {
+  // Main Ledger of IC
+  main: bigint;
+};
+
+const defaultTransactionFees = {
+  main: BigInt(DEFAULT_TRANSACTION_FEE_E8S),
+};
+
+/**
+ * A store that contains the transaction fees of ledgers.
+ *
+ * - setMain: replace the current fee in `main`.
+ */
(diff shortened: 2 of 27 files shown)


lmuntaner:

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -847,3 +850,16 @@ export const makeDummyProposals = async (neuronId: NeuronId): Promise<void> => {
 
 export const routePathNeuronId = (path: string): NeuronId | undefined =>
   getLastPathDetailId(path);
+
+export const minNeuronSplittable = (): number =>
```
peterpeterparker, line 854: why not adding these to `utils` and passing the `fee` as param as the other utils modified in this PR (`icp.utils`)?

frontend/svelte/src/lib/services/transactionFees.services.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getIdentity } from "./auth.services";
+
+export const loadMainTransactionFee = async () => {
+  try {
+    const identity = await getIdentity();
+    const fee = await transactionFee({ identity });
+    transactionsFeeStore.setMain(fee);
+  } catch (error) {
+    // Swallow error and continue with the DEFAULT_TRANSACTION_FEE value
+    console.log("Error getting the transaction fee from the ledger");
```
peterpeterparker, line 12: `console.error`?

frontend/svelte/src/lib/utils/neuron.utils.ts
```diff
@@ -94,12 +95,14 @@ export const votingPower = ({
...
 
+// TODO: Do we need this? What does it mean to have a valid stake?
+// TODO: https://dfinity.atlassian.net/browse/L2-507
 export const hasValidStake = (neuron: NeuronInfo): boolean =>
   // Ignore if we can't validate the stake
   neuron.fullNeuron
     ? neuron.fullNeuron.cachedNeuronStake +
         neuron.fullNeuron.maturityE8sEquivalent >
-      BigInt(TRANSACTION_FEE_E8S)
+      BigInt(DEFAULT_TRANSACTION_FEE_E8S)
```
peterpeterparker, line 105: so here it is not the same fee? should we have two / multiple constants for the fees then?

frontend/svelte/src/lib/services/transactionFees.services.ts
```diff
@@ -0,0 +1,15 @@
+import { transactionFee } from "../api/ledger.api";
```
peterpeterparker, line 1: rename service `transaction-fees.services.ts` instead of `transactionFees.services.ts` to follow pattern

peterpeterparker: likewise for `tran...store.ts`

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: I felt that the reason why this needs the fee is an implementation detail. For me the minimum amount to split doesn't necessarily have anything to do with a fee (conceptually), therefore I'd rather not expose it.

Same for the minimum amount to merge maturity.

The current calculation takes that into account, but not necessarily.

The same way that `MIN_NEURON_STAKE` does not depend on the transaction fee.

frontend/svelte/src/lib/services/transactionFees.services.ts
lmuntaner: sure

lmuntaner: we will need ot refactor the `knownNeurons.services.ts` as well then

frontend/svelte/src/lib/services/transactionFees.services.ts
lmuntaner: ok

frontend/svelte/src/lib/utils/neuron.utils.ts
lmuntaner: I didn't want to refactor this part because I think this part deserves a review as already stated in the issue.

I can't even answer the question of "What is a valid stake".

Another problem with this one, is that this is used inside the neurons store, therefore, we might not have yet the transaction fee. To to this properly, we should do a derived store that uses both stores and returns the valid neurons.

Since this looks like a big refactor not only related to the transaction fee, I didn't want to do it in this PR. But I added the comment so that we don't forget about it.

frontend/svelte/src/lib/services/transactionFees.services.ts
lmuntaner: done

frontend/svelte/src/lib/services/transactionFees.services.ts
lmuntaner: done

frontend/svelte/src/lib/services/neurons.services.ts
peterpeterparker: There isn't an edge case issue by not passing the fee as parameter?

e.g.:

```
$: max = maxICP({
    icp: account.balance,
    fee: $mainTransactionFeeNumberStore,
  });
```

is calculated when `account` change and when `fee` change. so if it takes 1min to fetch the fee, value will be still updated what would not be the case of these particular `functions`.

but maybe there isn't such edge case here?

frontend/svelte/src/lib/utils/neuron.utils.ts
peterpeterparker: Makes sense, thanks for the feedback

frontend/svelte/src/lib/services/neurons.services.ts
lmuntaner: good point, yes, there is this edge case.

Is it too crazy to create it as a derivedStore? Even the `MIN_NEURON_STAKE` to keep it consistent.

We have derived stores that return and update it?

Exposing the fee as a dependency is also another option.

What do you think?

peterpeterparker: First, apologies for the late answer.

Does it need a derived store? I assumed passing the store value from the component with an auto subscribe shouldn't do it?

isn't `minMaturityMerge()` equals `get(mainTransactionFeeNumberStore);` equals `$mainTransactionFeeNumberStore`?

maybe i am missing something, you can quickly show me tomorrow

lmuntaner: Yes, it is, but then we are exposing the logic of `minMaturityMerge`.

But yes, for now, I will go for the idea of passing the fee as parameter, it seems the easiest even though we are leaking implementation details.

lmuntaner:

frontend/svelte/src/lib/stores/transaction-fees.store.ts
```diff
@@ -0,0 +1,45 @@
...
+    // Used in the tests
+    reset() {
+      set(defaultTransactionFees);
+    },
+  };
+};
+
+export const transactionsFeeStore = initTransactionFeesStore();
+
+export const mainTransactionFeeNumberStore = derived(
```
peterpeterparker, line 42: `mainTransactionFeeStore` would be a bit shorter, do we really need the info "Number" within the name?

frontend/svelte/src/lib/stores/transaction-fees.store.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+
+    // Used in the tests
+    reset() {
+      set(defaultTransactionFees);
+    },
+  };
+};
+
+export const transactionsFeeStore = initTransactionFeesStore();
```
peterpeterparker, line 40: you use plural on "fee" (`initTransactionFeesStore` and `TransactionFeesStore`) or plural on "transaction" (`transactionsFeeStore`)

do we need plural? and if yes, can we use one?

frontend/svelte/src/lib/stores/transaction-fees.store.ts
lmuntaner: True, not really. I wanted to differentiate it from the main one, but just getting the main transaction fee is enough.

Will do

frontend/svelte/src/lib/stores/transaction-fees.store.ts
lmuntaner: True about the plural.

Maybe the ledger of each SNS has a different transaction fee and we also need to support that. It didn't add that much complexity to be prepared for it.

frontend/svelte/src/lib/stores/transaction-fees.store.ts
lmuntaner: done

frontend/svelte/src/lib/stores/transaction-fees.store.ts
lmuntaner: done

lmuntaner:

frontend/svelte/src/lib/stores/transaction-fees.store.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    setMain(fee: bigint) {
+      update((data) => ({
+        ...data,
+        main: fee,
+      }));
+    },
+
+    // Used in the tests
+    reset() {
```
peterpeterparker, line 34: I guess you can use `setMain` too in test at the moment at least?

frontend/svelte/src/lib/utils/neuron.utils.ts
```diff
@@ -604,3 +598,26 @@ export const votedNeuronDetails = ({
     .filter(
       (compactNeuronInfoMaybe) => compactNeuronInfoMaybe.vote !== undefined
     ) as CompactNeuronInfo[];
+
+export const minMaturityMerge = (fee: number): number => fee;
+
+export const hasEnoughMaturityToMerge = ({
+  neuron,
```
peterpeterparker, line 605: if you want fancy deconstruct `neuron: {fullNeuron}`

frontend/svelte/src/lib/stores/transaction-fees.store.ts
lmuntaner: Sure.

peterpeterparker (approved):
LGTM, thanks!!!

frontend/svelte/src/lib/services/neurons.services.ts
```diff
@@ -500,10 +501,11 @@ export const splitNeuron = async ({
       neuronId
     );
 
-    const transactionFeeAmount = TRANSACTION_FEE_E8S / E8S_PER_ICP;
+    const transactionFee = get(mainTransactionFeeStore);
```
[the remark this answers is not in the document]
lmuntaner: sure

frontend/svelte/src/lib/stores/transaction-fees.store.ts
lmuntaner: done

frontend/svelte/src/lib/utils/neuron.utils.ts
lmuntaner: done

frontend/svelte/src/lib/services/neurons.services.ts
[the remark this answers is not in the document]

PR #1008, opened 2022-06-20
L2-721: i18n Canister api errors

lmuntaner:
# Motivation

Errors from polling service in canisters api are translatable.

# Changes

* New Error type `ApiErrorKey` thrown when polling is exceeded.

# Tests

No extra tests

### DIFF
--- a/frontend/svelte/src/lib/api/canisters.api.ts
+++ b/frontend/svelte/src/lib/api/canisters.api.ts
@@ -2,7 +2,7 @@ import type { Identity } from "@dfinity/agent";
 import { AccountIdentifier, ICP, SubAccount } from "@dfinity/nns";
 import type { Principal } from "@dfinity/principal";
 import { CMCCanister } from "../canisters/cmc/cmc.canister";
-import { CMCError, ProcessingError } from "../canisters/cmc/cmc.errors";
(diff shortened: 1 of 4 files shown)


peterpeterparker:
Not against. In canisters if I remember correctly there are two types of such error in API, one with the message and another with the message plus some details. So if we want to make `ApiErrorKey` generic, we might need to extend it with the optional second param. Makes sense?

lmuntaner:
Yes, I can do that in another PR with all the canister errors.

For this PR, I didn't want to change the pattern and just fix the errors that are not translated.

We have a mapper in the canister services that checks each error type and translates it:

```
// Check CMC and IC Mgt Canister Errors
const canisterMapper: Array<[Function, string]> = [
  [InsufficientAmountError, "error.insufficient_funds"],
  [RefundedError, "error.canister_refund"],
  [InvalidaTransactionError, "error.canister_invalid_transaction"],
  [UserNotTheControllerError, "error.not_canister_controller_to_update"],
];
```

If we like the pattern I will implement it slowly in multiple PRs.

lmuntaner:
@peterpeterparker can you review?

peterpeterparker (approved):
Lgtm, thx

PR #1011, opened 2022-06-20
L2-727: Remove redirections to Flutter

lmuntaner:
# Motivation

Remove redirections from svelte to flutter.

# Changes

* Remove usage of `REDIRECT_TO_LEGACY` in svelte app.
* Remove all `SHOW_XXX` constants.
* Remove all checks with `SHOW_XXX` onMount to redirect to flutter.

# Tests

Not applicable.

### DIFF
--- a/frontend/svelte/env.config.mjs
+++ b/frontend/svelte/env.config.mjs
@@ -69,8 +69,6 @@ const LEDGER_CANISTER_URL = getRequiredEnvVar("LEDGER_CANISTER_URL");
 const OWN_CANISTER_URL = getRequiredEnvVar("OWN_CANISTER_URL");
 
 // Configuration
-// ... Redirects between the svelte and now deleted frontends
-const REDIRECT_TO_LEGACY = getRequiredEnvVar("REDIRECT_TO_LEGACY");
 // ... The testnet name
 const DEPLOY_ENV = getRequiredEnvVar("DEPLOY_ENV");
 // ... Whether to get the root key
@@ -80,7 +78,6 @@ export const envConfig = {
   ENVIRONMENT,
   DEPLOY_ENV,
   FETCH_ROOT_KEY,
-  REDIRECT_TO_LEGACY,
   MAINNET,
   HOST,
   OWN_CANISTER_ID,
--- a/frontend/svelte/src/routes/Neurons.svelte
+++ b/frontend/svelte/src/routes/Neurons.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import Layout from "../lib/components/common/Layout.svelte";
-  import { onDestroy, onMount } from "svelte";
+  import { onDestroy } from "svelte";
   import type { Unsubscriber } from "svelte/types/runtime/store";
   import { authStore } from "../lib/stores/auth.store";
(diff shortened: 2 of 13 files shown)


lmuntaner:
@peterpeterparker @bitdivine please review

frontend/svelte/env.config.mjs
```diff
@@ -69,8 +69,6 @@ const LEDGER_CANISTER_URL = getRequiredEnvVar("LEDGER_CANISTER_URL");
 const OWN_CANISTER_URL = getRequiredEnvVar("OWN_CANISTER_URL");
 
 // Configuration
-// ... Redirects between the svelte and flutter frontends
-const REDIRECT_TO_LEGACY = getRequiredEnvVar("REDIRECT_TO_LEGACY");
```
peterpeterparker, line 73: should we fully remove it form the config or have we gonna use it to obfuscate sns features?

lmuntaner: We might want to use feature flags for SNS, I agree, but I would use the DEPLOY_ENV for that and use flag specifically for that.

This env var us used from the build, to flutter, across all the deployment and build process. I believe that it's easier to totally remove it from the pipeline.

lmuntaner: @bitdivine what do you think? Do you want to use this env var for SNS?

frontend/svelte/src/routes/Neurons.svelte
```diff
@@ -54,70 +46,65 @@
...
+          {neuron}
+        />
+      {/each}
+    {/if}
+  </section>
+  <svelte:fragment slot="footer">
+    <Toolbar>
+      <button
+        data-tid="stake-neuron-button"
+        class="primary"
```
bitdivine, line 77: I think this is unrelated to removing redirects?  I have no idea whether it's good or bad or intentional but it looks unrelated.  Same below.

bitdivine (approved):
There a couple of small changes that aren't clearly related to removing redirects.  Are these intentional?

frontend/svelte/src/routes/Neurons.svelte
lmuntaner: oh, it should be full-width, good point. This happened during a conflict fix, thanks!

frontend/svelte/env.config.mjs
bitdivine: I had been thinking of it but at minimum we'd want to rename it.  A dedicated feature flag `SHOW_SNS` would work, and yes, we could enable that for some deployment targets in dfx.json

frontend/svelte/src/routes/Neurons.svelte
lmuntaner: fixed, I will do a quick recheck of all the main pages before merging the PR.

frontend/svelte/env.config.mjs
bitdivine: Personally I'm leaning more towards a feature flag.  What do you think?

lmuntaner: Yes, I agree.

PR #1014, opened 2022-06-21
feat: split pane

peterpeterparker:
# Motivation

In the future we aim to offer a better user experience on desktop. Adding a split pane layout - making the menu sticky if screen width > 1300px - is one of the first step towards that goal.

# Changes

- new component `<SplitPane />` and usage
- trigger to open the menu extracted fro the `<Menu />` and integrated into header (new cmp `<MenuButton />`)
- various style (`position`, `width` etc.) update to fit new layout concept
- `<HeadlessLayout />` dropped to use a single `<Layout />` component for any view of the app
- new `components/header` package that contains header related component of the dapp
- `<Menu />` adapted to support both normal behavior ("animated over content") or "sticky" mode
- `--padding-6x` dropped because effectively it was equals to `--padding-4x`
- `accounts.services.routePathAccountIdentifier()` because navigating back from Wallet to Accounts view lead to displaying an error if split pane is used. I am guessing that components was not destroyed as fast as without the pane.
- `<MenuItems />` update to display the menu selected on main pages but also on detail pages

# Screenshots

Capture d’écran 2022-06-21 à 11 19 41
Capture d’écran 2022-06-21 à 11 19 45
Capture d’écran 2022-06-21 à 11 19 47
Capture d’écran 2022-06-21 à 11 19 55
gif


frontend/svelte/src/lib/components/ui/SplitPane.svelte
```diff
@@ -0,0 +1,49 @@
...
+    position: absolute;
+    top: calc(var(--header-offset, 0px) + var(--header-height));
+    left: 0;
+    right: 0;
+    bottom: 0;
+
+    display: flex;
+    flex-flow: row nowrap;
+
+    pointer-events: none;
```
mstrasinskis, line 31: Is it done to disable clicking on the Backdrop?

mstrasinskis (approved):
Awesome feature. Thank you 👍

frontend/svelte/src/lib/components/ui/SplitPane.svelte
peterpeterparker: Good catch. It was probably useful in one of the first iteration but does not seem to be useful anymore. I'll remove it. Thanks!

PR #1015, opened 2022-06-21
style: footer button width 50%

peterpeterparker:
# Motivation

Distribute buttons in the footer.

# Changes

- use `full-width` when two buttons and set 50% as default

# Screenshots

Capture d’écran 2022-06-21 à 14 00 25
Capture d’écran 2022-06-21 à 14 00 27
Capture d’écran 2022-06-21 à 14 00 29
Capture d’écran 2022-06-21 à 14 00 32
Capture d’écran 2022-06-21 à 14 00 34
Capture d’écran 2022-06-21 à 14 00 40
Capture d’écran 2022-06-21 à 14 00 42
Capture d’écran 2022-06-21 à 14 00 45
Capture d’écran 2022-06-21 à 14 00 52
Capture d’écran 2022-06-21 à 14 00 54
Capture d’écran 2022-06-21 à 14 00 57
Capture d’écran 2022-06-21 à 14 00 59
Capture d’écran 2022-06-21 à 14 01 03
Capture d’écran 2022-06-21 à 14 01 06
Capture d’écran 2022-06-21 à 14 01 09

### DIFF
--- a/frontend/svelte/src/lib/components/common/Footer.svelte
+++ b/frontend/svelte/src/lib/components/common/Footer.svelte
@@ -41,7 +41,7 @@
     }
 
     :global(button) {
-      flex: 1 1;
+      width: 50%;
     }
   }
 </style>
--- a/frontend/svelte/src/routes/Accounts.svelte
+++ b/frontend/svelte/src/routes/Accounts.svelte
@@ -97,13 +97,13 @@
       {#if accounts}
         <Toolbar>
           <button
-            class="primary"
+            class="primary full-width"
             on:click={openNewTransaction}
             data-tid="open-new-transaction"
             >{$i18n.accounts.new_transaction}</button
           >
           <button
-            class="primary"
+            class="primary full-width"
             on:click={openAddAccountModal}
             data-tid="open-add-account-modal"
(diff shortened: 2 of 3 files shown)


peterpeterparker:
Note: same settings look nicer with new UI

Capture d’écran 2022-06-21 à 14 07 10
Capture d’écran 2022-06-21 à 14 07 12

frontend/svelte/src/lib/components/common/Footer.svelte
```diff
@@ -41,7 +41,7 @@
     }
 
     :global(button) {
-      flex: 1 1;
```
lmuntaner, line 44: I would prefer to wrap every button with the width we want, instead of using this global. But it's a matter of taste.

lmuntaner (approved):
Minor comment.

LGTM, thanks!

frontend/svelte/src/lib/components/common/Footer.svelte
peterpeterparker: There are zillion changes in the light theme branch so I try to keep things slim until we go forward with it .

PR #1025, opened 2022-06-22
Fix: Navigation issue on Neuron Detail

lmuntaner:
# Motivation

Fix issue when user tries to navigate out of the neurons detail page with the menu.

# Changes

* Skip checks in the NeuronDetail page when route changes but it's not a neuron detail page.
* Add extra check when getting the neuron id from the path in "routePathNeuronId".

# Tests

* Add test cases in routePathNeuronId.


mstrasinskis (approved):
LGTM.

PR #1028, opened 2022-06-23
L2-738: Improve FF error message with HW

lmuntaner:
# Motivation

Users that try to connect a hardware wallet with Firefox should know why it's failing.

# Changes

* Check the error thrown when trying to connect to the ledger device and change the message key for the error.

# Tests

Not applicable.


lmuntaner:

peterpeterparker (approved):
LGTM, thanks

Being said, should we ask Mischa if we should not be pro-active? Display earlier the user it won't be supported rather than at the end of the process when an error occurs?

Also Safari does not yet support it as well.

```
if ("hid" in navigator) {
  // The WebHID API is supported.
}
```

lmuntaner:
That's a good point. Not sure what's better. For versions we wait until the end, and that can still happen, but for devices we show a tooltip, for example we disable trying to merge neurons from HW.

Worth having Mischa's feedback.

PR #1031, opened 2022-06-23
Introduce new types so that rollback will succeed

hpeebles:
We don't want to merge this into master.

This is branched from the current prod version and solely introduces the 2 new enum variants contained in the next release candidate.

Adding these types allows us to perform a rollback from that release candidate if needed.


bitdivine:
Thank you @hpeebles .  I think this PR should ideally be frozen in this state so I will close it.  Hopefully we won't ever need to use it! :-)

PR #1034, opened 2022-06-24
Rm DEPLOY_ENV

bitdivine:
# Motivation
`DEPLOY_ENV` duplicates `DFX_NETWORK`.  The latter is set by `dfx` so there is no need of the former.

The only complication is that `DEPLOY_ENV` has one value that is not a network: `nobuild`.  This uses the existing WASM.  That `nobuild` functionality would have to be replaced by something else.

# Changes
* Replace `DEPLOY_ENV` with `DFX_NETWORK`, simplifying code where previously both were used.
* Introduce the `NNS_DAPP_WASM` environment variable.  If set, it points to a wasm file to use instead of building.

# Tests
Need to test the following:
- [x] `./deploy.sh` works for local deployments?
- [x] `./deploy.sh` works for deployments to the testnet?
- [x] `e2e-tests/scripts/deploy-local-canisters` works? (it is used in CI, so if CI passes it should be fine)
- [x] Setting `NNS_DAPP_WASM` works?
```
  max@sinkpad:~/dfn/nns-dapp/branches/no-deploy-env (5:18)$ NNS_DAPP_WASM=./nns-dapp.wasm ./deploy.sh --nns-dapp

START_DFX=false
DEPLOY_NNS_BACKEND=false
DEPLOY_II=false
DEPLOY_NNS_DAPP=true
POPULATE=false
OPEN_NNS_DAPP=false
Creating canister "nns-dapp"...
"nns-dapp" canister was already created and has canister id: "qhbym-qaaaa-aaaaa-aaafq-cai"
Deploying: nns-dapp
All canisters have already been created.
Building canisters...
Executing './build.sh'
Skipping build as the WASM file has already been provided.
Installing canisters...
Upgrading code for canister nns-dapp, with canister_id qhbym-qaaaa-aaaaa-aaafq-cai
Module hash 2d546e3253f6fc1387c883d9114cb1f657f28d66127a4ccd6e869dde5efbc301 is already installed.
Deployed canisters.
Deployed to: http://qhbym-qaaaa-aaaaa-aaafq-cai.localhost:8080
max@sinkpad:~/dfn/nns-dapp/branches/no-deploy-env (9:17)$
```


build.sh
```diff
@@ -19,18 +19,22 @@ set -euo pipefail
...
-
-if [[ $DEPLOY_ENV = "nobuild" ]]; then
-  echo "Skipping build as requested"
+# If the WASM has been provided, there is no need to build it:
+[[ "${NNS_DAPP_WASM:-}" == "" ]] || {
+  test -f "${NNS_DAPP_WASM}" || {
+    echo "File not found, or node is not a file: NNS_DAPP_WASM='${NNS_DAPP_WASM}'"
+    exit 1
+  } >&2
+  cp "${NNS_DAPP_WASM}" nns-dapp.wasm
```
lmuntaner, line 28: Is it a proble if the NNS_DAPP_WASM has the same name?

lmuntaner (commented):
Just a couple of comments for now. Looking good!

build.sh
bitdivine: cp file to self works.  Ah, but it does return a non-zero exit code.  Good catch.  I thought about that but didn't realize about the exit code.  I was going to demo the "no problem":
```
max@sinkpad:~ (2:22)$ echo foo > bar
max@sinkpad:~ (2:25)$ cp bar bar ; echo $?
cp: 'bar' and 'bar' are the same file
1
max@sinkpad:~ (2:28)$ 
```

e2e-tests/README.md
```diff
@@ -22,7 +22,7 @@ dfx start
 
 # in another shell, deploy the canister
 # in the repo root
-DEPLOY_ENV=nobuild dfx deploy --network local
+NNS_DAPP_WASM=nns-dapp.wasm dfx deploy --network local
```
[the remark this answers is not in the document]
bitdivine: I'll check the README on merge.  Done.  I note the line:
```
./deploy.sh --network local
```
If no network is specified, it defaults to local.  Should I reduce the above to just `./deploy.sh` or do you find the explicit network flag helpful?

build.sh
bitdivine: Fix ready, waiting for clarification on the README before adding:
```
  INSTALL_LOCATION="$(jq -r '.canisters["nns-dapp"].wasm' dfx.json)"
  [[ "$(realpath "${NNS_DAPP_WASM}")" == "$(realpath "${INSTALL_LOCATION}")" ]] ||
    cp "${NNS_DAPP_WASM}" "${INSTALL_LOCATION}"
```
Note: Arguably I should go through all the locations where `nns-dapp.wasm` is mentioned directly and use the value from `dfx.json` instead but that is outside the scope of this PR.

e2e-tests/README.md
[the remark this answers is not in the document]
lmuntaner: I prefer being explicit.

build.sh
lmuntaner: more than enough for now

PR #1035, opened 2022-06-24
fix: glitchy signout

peterpeterparker:
# Motivation

Follow-up of #1020. Adding a fixed background and removing the lazy loading of the images (that is loaded anyway) seems to make the signout less glitchy for the eyes.

Still has to be tested on mobile devices.

# Changes

- remove `loading=lazy` for the image as it is always loaded
- add a scree background behind login screen
- remove timeout for reload


lmuntaner (approved):
LGTM, thanks!

PR #1050, opened 2022-06-28
fix: <CardInfo> was created with unknown prop ...

peterpeterparker:
# Motivation

`NeuronCard` path many properties that are not used to display static information therefore Svelte and jest throw a warning `<CardInfo> was created with unknown prop ...`.

# Note

Not against a better idea that do not need to rewrite the all neuron card architecture.

# Changes

- new `NeuronCardContainer` that either renders the `Card` or the `CardInfo` and path props accordingly


mstrasinskis (approved):
LGTM

PR #1051, opened 2022-06-28
fix: missing i18n keys

peterpeterparker:
# Motivation

Some keys are missing in the `i18n` typescript definition


lmuntaner (commented):
Thanks!

peterpeterparker:
We do not need this anymore.

PR #1059, opened 2022-06-29
fix: hide dev banner

peterpeterparker:
# Motivation

Following our change in the layout, if the dev banner gets hidden, the spacing isn't removed because Svelte does not remove the style in head.

# Changes

- instead of `svelte:head` use style property on the document

# Original issue that is solved with this PR

remove


mstrasinskis (approved):
LGTM

PR #1065, opened 2022-06-29
L2 779 launchpad menu entry

mstrasinskis:
# Motivation

Add "Launch Pad" menu entry (not visible on mainnet yet)

https://user-images.githubusercontent.com/98811342/176535474-0858c4eb-2e19-4234-956c-a78db10e6cfd.mov

image

# Changes

- New component
- New Menu entry

# Tests

- MenuItem / should render slotted statusIcon


frontend/svelte/src/lib/components/common/MenuItems.svelte
```diff
@@ -49,16 +53,31 @@
       icon: IconSettingsApplications,
     },
   ];
+
+  if (IS_TESTNET) {
```
peterpeterparker, line 57: nth: set the array only once - i.e. no push

```
  const routes: {
    context: string;
    selected: boolean;
    label: string;
    icon: typeof SvelteComponent;
    statusIcon?: typeof SvelteComponent;
  }[] = [  const routes: {
    context: string;
    selected: boolean;
    label: string;
    icon: typeof SvelteComponent;
    statusIcon?: typeof SvelteComponent;
  }[] = [
     // existing options,
     ...(IS_TESTNET && {context: "launchpad",
            selected: isSelectedPath([
              AppPath.SNSLaunchpad,
              AppPath.SNSProjectDetail,
            ]),
            label: "launchpad",
            icon: IconRocketLaunch,
            statusIcon: New,})
```

frontend/svelte/src/lib/components/ui/MenuItem.svelte
```diff
@@ -39,12 +40,17 @@
...
+    :global(:nth-child(3)) {
+      margin-right: var(--padding-2x);
+    }
   }
 
   span {
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
+    flex: auto;
```
peterpeterparker, line 53: did you check that it would still works out (with the trimed three dots) if the menu entry is too long? you never know maybe some day we will translate the app in german 😉

`Launch Pa...   new`

frontend/svelte/src/lib/components/ui/New.svelte
```diff
@@ -0,0 +1,21 @@
+<script lang="ts">
+  import IconFiberNew from "../../icons/IconFiberNew.svelte";
+</script>
+
+<span>
```
peterpeterparker, line 5: - do we need the `span` - i.e. the sole icon `svg` is not enough?

- do we need to add an `aria-label` or `title` for accessibility?

frontend/svelte/src/lib/components/ui/MenuItem.svelte
mstrasinskis: Good catch. Fixed
image

frontend/svelte/src/lib/types/i18n.d.ts
```diff
@@ -23,6 +23,7 @@ interface I18nCore {
   log: string;
   principal_id: string;
   copy: string;
+  new: string;
```
peterpeterparker, line 26: `new` is not a reserved keyword? maybe not but maybe safer anyway to use another keyword?

mstrasinskis: Agree 👍

PR #1066, opened 2022-06-30
Bump Internet Identity

bitdivine:
# Motivation
- The version of II deployed to testnets is out of date.  The production II has new, nice features such as one click login that we would like to use on testnets as well.

# Changes
* Bump II

# Tests


lmuntaner (approved):
thanks!

PR #1070, opened 2022-07-01
L2-735: Use Collapsible component in ProjectInfoSection

lmuntaner:
# Motivation

User can see the information on max and min commitment items when clicking on them.

# Changes

* Changes in Collapsible component: add "none" to the iconSize prop values. Which hides the button.
* Move the extra clickable space to the CardBlock.
* Use the Collapsible in the ProjectInfoSection.
* Remove info icon from KeyValuePair.
* New InfoContextKey component that adds info icon and makes content collapsible.

# Tests

* Add proper test case to Collapsible component.
* Text for new component InfoContextKey.
* Remove test case for KeyValuePair.

### DIFF
--- a/frontend/svelte/src/lib/components/ui/Collapsible.svelte
+++ b/frontend/svelte/src/lib/components/ui/Collapsible.svelte
@@ -8,7 +8,7 @@
   export let initiallyExpanded: boolean = false;
   export let maxContentHeight: number | undefined = undefined;
 
-  export let iconSize: "small" | "medium" = "small";
+  export let iconSize: "small" | "medium" | "none" = "small";
 
   // Minimum height when some part of the text-content is visible (empirical value)
   const CONTENT_MIN_HEIGHT = 40;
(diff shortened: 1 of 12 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/176846085-0f68bba1-2068-47c2-9a74-58c09413e15b.mov

lmuntaner:
@mstrasinskis please review

lmuntaner:
@mstrasinskis please review again. After talking with Mischa, only the "key" is clickable.

frontend/svelte/src/lib/components/ui/Collapsible.svelte
```diff
@@ -104,10 +106,6 @@
     @include interaction.tappable;
     user-select: none;
 
-    // increase click area
```
mstrasinskis, line 107: Isn't it still relevant for not `none` version?

frontend/svelte/src/lib/components/ui/InfoContextKey.svelte
```diff
@@ -0,0 +1,29 @@
...
+<style lang="scss">
+  .wrapper {
+    display: flex;
+    align-items: center;
+    gap: var(--padding);
+  }
+
+  .icon {
+    color: var(--primary);
+    display: flex;
```
mstrasinskis, line 25: Wouldn't flex params from wrapper be enough if you set flex: none; for the icon?

frontend/svelte/src/lib/components/ui/Collapsible.svelte
lmuntaner: not really, it was messing up with the spacing around the values.

lmuntaner: Also, I thuoght that it made sense that the responsibility to provide for enough clicking space was for the slot itself, not from the Collapsible element

frontend/svelte/src/lib/components/ui/InfoContextKey.svelte
lmuntaner: not really, the stupid svg from the icon was not well aligned with the text next to it without this.

mstrasinskis: Thanks for the explanation.

PR #1073, opened 2022-07-01
feat: polyfill inset

peterpeterparker:
# Motivation

Polyfill `inset: 0` to support older devices (even though according [spec](https://caniuse.com/?search=inset) it should be widely supported) 

# Changes

- replace `inset: 0` with mixin
- replace other `inset: ...` with top, right, bottom and left props

# Tests


mstrasinskis (approved):
LGTM.

PR #1075, opened 2022-07-01
L2-786: Project Status with Tag Component

lmuntaner:
# Motivation

Use a tag to show the status of a project.

# Changes

* Create a Tag component from the Followee component.
* Use Tag in ProjectStatusSection.
* Use Tag instead of Badge for Proposals.
* New Color enum type.
* Remove ProposalColor type.
* Add logic for the status text in ProjectStatusSection.

# Tests

* Add test cases to ProjectStatusSection.
* Add tests for Tag component.


lmuntaner:
UI:

Screenshot 2022-07-01 at 13 56 21

lmuntaner:

frontend/svelte/src/lib/components/project-detail/ProjectStatusSection.svelte
```diff
@@ -46,8 +57,7 @@
   <div class="wrapper" data-tid="sns-project-detail-status">
     <div class="title">
       <h2>{$i18n.sns_project_detail.status}</h2>
-      <!-- TODO: Create another Badge for SNS? -->
-      <Badge>{$i18n.sns_project_detail.accepting}</Badge>
+      <Tag>{statusTextMapper[projectStatus]}</Tag>
```
peterpeterparker, line 60: not sure we need a new component but maybe the `Badge` tab can be redesigned and used here. last week Mischa suggested to use outline style for the proposal status too (see [Slack](https://dfinity.slack.com/archives/C03HZERHBT3/p1656498788649699?thread_ts=1656481631.667719&cid=C03HZERHBT3)).

lmuntaner: The current Badge looked very different, whereas the UI of the "Followee" (refactored) here looked exactly the same. 

Therefore, I just extracted the styles into a component. We could use this `Tag` component instead of the Badge and remove the Badge. Not sure whether it makes sense to refactor it, rathern than use the new component and remove the old one.

The styles of this component are used anyawy, so having it makes sense even if it's only fr the Followee component.

lmuntaner: Screenshot 2022-07-04 at 08 15 31

peterpeterparker: > The current Badge looked very different, whereas the UI of the "Followee" (refactored) here looked exactly the same.

Yes that's the goal of Mischa my understanding - i.e. use a "very different" display for the proposal badge too

Quoting him:

> just one thing came across now. Tags on the proposals have a different style than tags for followee topics.
> ...
> I quickly tried with outline style. It looked OK. 
> If quick, than you can change…

lmuntaner: I'll talk to him to see which change he means.

lmuntaner:

New UI:

Screenshot 2022-07-04 at 14 09 17
Screenshot 2022-07-04 at 14 09 25
Screenshot 2022-07-04 at 14 10 57
Screenshot 2022-07-04 at 14 11 15

frontend/svelte/src/lib/components/ui/Tag.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
+  import type { Color } from "../../types/theme";
+
+  export let color: Color | undefined = undefined;
+</script>
+
+<span data-tid="tag" class={color}><slot /></span>
```
peterpeterparker, line 7: we can probably spare that `span` with a `svelte:element` so that it renders directly the dom element like `li` or `h2` without adding a child. what you think?

lmuntaner: mhm, I like the idea. It's very flexible, but if we don't abuse it, it can be a good way.

frontend/svelte/src/lib/components/ui/Tag.svelte
```diff
@@ -0,0 +1,34 @@
+<script lang="ts">
+  import type { Color } from "../../types/theme";
+
+  export let color: Color | undefined = undefined;
+  export let tagName: "span" | "h3" = "span";
+</script>
+
+<svelte:element this={tagName} data-tid="tag" class={`${color} tag`}
```
peterpeterparker, line 8: this can probably add `class="undefined tag"` to the do. if correct, maybe good to `colod ?? ''` handles it

frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard/Followee.svelte
```diff
@@ -25,7 +26,7 @@
 
 <ul aria-labelledby={id}>
   {#each followee.topics as topic}
-    <li>{topicTitle(topic)}</li>
+    <li><Tag>{topicTitle(topic)}</Tag></li>
```
peterpeterparker, line 29: do you need the `span`? `<Tag tagName="li" />`

frontend/svelte/src/lib/components/project-detail/ProjectStatusSection.svelte
```diff
@@ -4,16 +4,18 @@
...
   import InfoContextKey from "../ui/InfoContextKey.svelte";
   import KeyValuePair from "../ui/KeyValuePair.svelte";
   import ProgressBar from "../ui/ProgressBar.svelte";
   import Spinner from "../ui/Spinner.svelte";
+  import Tag from "../ui/Tag.svelte";
   import CommitmentProgressBar from "./CommitmentProgressBar.svelte";
 
   export let project: SnsFullProject;
 
+  const nowInSeconds: number = Math.round(Date.now() / 1000);
```
peterpeterparker, line 18: nth: there are already a couple of `Math.round(Date.now() / 1000)` in the code, maybe worth extracting a function

frontend/svelte/src/lib/components/project-detail/ProjectStatusSection.svelte
```diff
@@ -27,15 +29,24 @@
       : undefined;
   let currentDateTillStartSeconds: number;
   $: currentDateTillStartSeconds = Number(
-    BigInt(Math.round(Date.now() / 1000)) - project.summary.swapStart
+    BigInt(nowInSeconds) - project.summary.swapStart
```
peterpeterparker, line 32: can the `project` change while the component is mounted? if yes, `nowInSeconds` would not be accurate and reflect passed time

frontend/svelte/src/lib/components/project-detail/ProjectStatusSection.svelte
lmuntaner: sure

frontend/svelte/src/lib/components/project-detail/ProjectStatusSection.svelte
lmuntaner: Yes, for a long time, if the user stays longer.

For the final countdown, maybe I can add create a setInterval, but for this PR I think it was too much.

frontend/svelte/src/lib/components/ui/Tag.svelte
lmuntaner: sure

frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard/Followee.svelte
lmuntaner: no, good point!

frontend/svelte/src/lib/components/ui/Tag.svelte
lmuntaner: done

frontend/svelte/src/lib/components/neuron-detail/NeuronFollowingCard/Followee.svelte
lmuntaner: done

lmuntaner:

frontend/svelte/src/lib/components/project-detail/ProjectStatusSection.svelte
peterpeterparker: i guess you can also "just" call the new function and that should do the trick?

is:

```
$: currentDateTillStartSeconds = Number(
    BigInt(nowSeconds) - project.summary.swapStart
  );
```

I meant:

```
$: currentDateTillStartSeconds = Number(
    BigInt(nowInSeconds()) - project.summary.swapStart
  );
```

peterpeterparker (approved):
LGTM, thanks

frontend/svelte/src/lib/components/project-detail/ProjectStatusSection.svelte
lmuntaner: well, that will only change the value when the `summary` changes. It will not add any kind of countdown.

lmuntaner: I'd rather update the value of `nowSeconds` in a counter inside the componend. Then, all these variables will be updated with it.

PR #1077, opened 2022-07-01
Chore: Rename or remove SNS from name

lmuntaner:
# Motivation

Rename files with SNS either to Sns or removing SNS altogether.

# Changes

* Rename files and their imports.

# Tests

* Fix test files.


frontend/svelte/src/lib/components/sns-launchpad/SNSProposals.svelte
```diff
@@ -5,7 +5,7 @@
   import { i18n } from "../../stores/i18n";
```
mstrasinskis, line 5: How about `sns-launchpad` folder and `snsProjects.store`?

mstrasinskis (approved):
LGTM, only one question.

frontend/svelte/src/lib/components/sns-launchpad/SNSProposals.svelte
lmuntaner: sure, even sns-project-detail

PR #1080, opened 2022-07-04
Use the SNS version matching the canisters

bitdivine:
# Motivation
When deploying, the sns version needs to match the canisters or else there can be hard-to-diagnose failures.

The matching `sns` binary is already downloaded, it just needs to be called explicitly in `deploy.sh`, rather than picking up whichever `sns` is in `PATH`.

# Changes
* Use the `sns` from the ic artifacts directory in `deploy.sh`

# Tests
* I have run `./deploy.sh --sns testnet` and it worked.  It still deploys "old" canisters from a week ago but that is orthogonal.


peterpeterparker (approved):
LGTM

PR #1081, opened 2022-07-04
L2-795: Swap Participation Flow 1 UI

lmuntaner:
# Motivation

User can select an account and enter an amount to participate in a swap.

# Changes

* New modal ParticipateSwapModal.
* New screens Participate and ReviewParticipate (setup only) used in the modal.
* New button class "secondary".
* Setup of new component "SelectAccountDropdown.
* Add minParticipationCommitment and maxParticipationCommitment to SNS Project mock data.

# Tests

* Tests for new Modal.
* Add test case in ProjectStatusSection that opens modal.

### DIFF
--- a/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+  import { onDestroy } from "svelte";
+
   import { accountsStore } from "../../stores/accounts.store";
   import { i18n } from "../../stores/i18n";
   import type { Account } from "../../types/account";
@@ -8,7 +10,7 @@
   export let skipHardwareWallets: boolean = false;
 
   let selectableAccounts: Account[] | undefined;
-  accountsStore.subscribe((accounts) => {
+  const unsubscribe = accountsStore.subscribe((accounts) => {
     if (accounts.main !== undefined) {
       selectedAccount = accounts.main;
       selectableAccounts = [
@@ -18,6 +20,8 @@
       ];
     }
   });
+
+  onDestroy(unsubscribe);
 </script>
 
 <!-- TODO: Implement https://dfinity.atlassian.net/browse/L2-800 -->
--- a/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -12,7 +12,7 @@
   let selectableAccounts: Account[] | undefined;
   const unsubscribe = accountsStore.subscribe((accounts) => {
     if (accounts.main !== undefined) {
-      selectedAccount = accounts.main;
+      selectedAccount = selectedAccount ?? accounts.main;
       selectableAccounts = [
         accounts.main,
         ...(accounts.subAccounts ?? []),
--- a/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -1,0 +1,38 @@
+<script lang="ts">
+  import { accountsStore } from "../../stores/accounts.store";
+  import { i18n } from "../../stores/i18n";
+  import type { Account } from "../../types/account";
+  import Spinner from "../ui/Spinner.svelte";
+
+  export let selectedAccount: Account | undefined = undefined;
+  export let skipHardwareWallets: boolean = false;
+
+  let selectableAccounts: Account[] | undefined;
+  accountsStore.subscribe((accounts) => {
+    if (accounts.main !== undefined) {
+      selectedAccount = accounts.main;
+      selectableAccounts = [
+        accounts.main,
+        ...(accounts.subAccounts ?? []),
+        ...(skipHardwareWallets ? [] : accounts.hardwareWallets ?? []),
+      ];
+    }
+  });
+</script>
+
+<!-- TODO: Implement https://dfinity.atlassian.net/browse/L2-800 -->
+{#if selectableAccounts !== undefined}
+  <select
+    bind:value={selectedAccount}
+    name="account"
+    data-tid="select-account-dropdown"
+  >
+    {#each selectableAccounts as account}
+      <option value={account}>
+        {account.name ?? $i18n.accounts.main_account}</option
+      >
+    {/each}
+  </select>
+{:else}
+  <Spinner />
+{/if}
--- a/frontend/svelte/src/lib/components/project-detail/Participate.svelte
+++ b/frontend/svelte/src/lib/components/project-detail/Participate.svelte
@@ -1,0 +1,125 @@
+<script lang="ts">
+  import type { ICP } from "@dfinity/nns";
+  import { createEventDispatcher } from "svelte";
+  import { i18n } from "../../stores/i18n";
+  import {
+    mainTransactionFeeStoreAsIcp,
+    mainTransactionFeeStore,
+  } from "../../stores/transaction-fees.store";
+
+  import type { Account } from "../../types/account";
+  import { maxICP } from "../../utils/icp.utils";
+  import SelectAccountDropdown from "../accounts/SelectAccountDropdown.svelte";
+  import IcpComponent from "../ic/ICP.svelte";
+  import Input from "../ui/Input.svelte";
+  import KeyValuePair from "../ui/KeyValuePair.svelte";
+
+  export let selectedAccount: Account | undefined = undefined;
+  export let amount: number | undefined = undefined;
+  // TODO: Handle min and max validations: https://dfinity.atlassian.net/browse/L2-798
+  export let minAmount: ICP;
+  export let maxAmount: ICP;
+
+  let max: number = 0;
+  $: max = maxICP({
+    icp: selectedAccount?.balance,
+    fee: $mainTransactionFeeStore,
+  });
+  const addMax = () => (amount = max);
+
+  let disableButton: boolean;
+  $: disableButton =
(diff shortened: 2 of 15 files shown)


lmuntaner:

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
```diff
@@ -0,0 +1,38 @@
...
+  import { accountsStore } from "../../stores/accounts.store";
+  import { i18n } from "../../stores/i18n";
+  import type { Account } from "../../types/account";
+  import Spinner from "../ui/Spinner.svelte";
+
+  export let selectedAccount: Account | undefined = undefined;
+  export let skipHardwareWallets: boolean = false;
+
+  let selectableAccounts: Account[] | undefined;
+  accountsStore.subscribe((accounts) => {
```
peterpeterparker, line 11: if you `subscribe` to a store you should also `unsubscribe` on destroy

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
```diff
@@ -0,0 +1,38 @@
...
+    if (accounts.main !== undefined) {
+      selectedAccount = accounts.main;
+      selectableAccounts = [
```
peterpeterparker, line 14: worse a `derivedStore`? I though you like these 😉)

frontend/svelte/src/lib/components/project-detail/Participate.svelte
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  const goNext = () => {
+    dispatcher("nnsNext");
+  };
+</script>
+
+<div class="wrapper" data-tid="sns-swap-participate-step-1">
+  <div>
+    {#if selectedAccount !== undefined}
```
peterpeterparker, line 46: you need that check? isn't that step not displayed if no account are selected therefore it's acceptable to render it empty?

frontend/svelte/src/lib/themes/button.scss
```diff
@@ -83,6 +84,25 @@ button {
...
 
+  &.secondary {
+    &[disabled],
+    &[disabled]:hover {
+      background: var(--disable);
+      color: var(--disable-contrast);
+      cursor: default;
+    }
+
+    background: var(--primary-contrast);
```
peterpeterparker, line 95: is that correct? `secondary` button uses `primary` color?

frontend/svelte/src/lib/components/project-detail/Participate.svelte
peterpeterparker, line 46: maybe I missed it but no test for this?

peterpeterparker:
Good point the question regarding UI of the input. I would also treat that in a separate issue.

Regarding "shall we create a new component", I would say as for the other topics. If we are able easily to integrate the new design in the existing features, that's a win. If too complicated, as the input is quite a complicated component, maybe there is a way "just" by defining a new style and not a full new component?

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
lmuntaner: Why a derived store? not sure I follow, hehe.

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
```diff
@@ -0,0 +1,38 @@
...
+    data-tid="select-account-dropdown"
+  >
+    {#each selectableAccounts as account}
+      <option value={account}>
+        {account.name ?? $i18n.accounts.main_account}</option
+      >
+    {/each}
+  </select>
+{:else}
+  <Spinner />
```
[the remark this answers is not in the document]
lmuntaner: This part has not yet been agreed. I will add it as a comment here: https://dfinity.atlassian.net/browse/L2-800

frontend/svelte/src/lib/themes/button.scss
lmuntaner: correct, it uses the same colors but reversed, the color as backbrouned and background as color.

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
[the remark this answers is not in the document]
peterpeterparker: Thanks

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
lmuntaner: done

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
lmuntaner: It's a transformation from the store to the array. It could be a useful derived store.

Although here, it wouldn't prevent it to use the subscribe. I still want to main account to be the selected by default. For that it's easier if I subscribe to the accounts store. That reminded me that I should not set "selectedAccount" if it is already defined, since this is just the default value. I will change that.

frontend/svelte/src/lib/components/project-detail/Participate.svelte
peterpeterparker: Yeah sure that works but it feels like this particular check "if no selected account is provided then the dropdown to select account should be displayed otherwise the user should not be able to select an account" is not tested - i.e. not the button but the dropdown. but again might have missed it.

lmuntaner:

Btw, what do you think of the "button" as link, used as Max now?

Screenshot 2022-07-04 at 17 22 18

Should we create another button class for this?

frontend/svelte/src/lib/themes/button.scss
lmuntaner: sure

frontend/svelte/src/lib/components/project-detail/Participate.svelte
lmuntaner: Not the current check. This check is a UI render check. If no selected account, then we can't show the amount of ICP.

But I will add a test case that the button is disabled if no account is selected.

frontend/svelte/src/lib/themes/button.scss
lmuntaner: done

frontend/svelte/src/lib/components/project-detail/Participate.svelte
lmuntaner: done

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
peterpeterparker: Yes definitely, you would just spare a couple of lines in the component and defer the map from object to array to the derived store. All good.

peterpeterparker (approved):
LGTM, thanks

frontend/svelte/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      participateButton && fireEvent.click(participateButton);
+      await waitFor(() =>
+        expect(queryByTestId("sns-swap-participate-step-2")).toBeFalsy()
+      );
+    });
+  });
+
+  describe("accountsStore is empty", () => {
+    it("should have disabled button if no account is selected", async () => {
```
dskloetd, line 99: The only line where I see this test checking if the button is disabled is line 109 and there the button is not disabled.
Do you remember what the intent was of this test?
I know the PR is old but the confusing combination of test description and expectation carries all the way to today:
https://github.com/dfinity/nns-dapp/blob/a7b822b4231e324753cc38d153322208e1662f71/frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts#L234

lmuntaner: That the button in the first screen is disabled if no account is selected.

By default, it will select the main account. Yet, the accounts might not be loaded.

This test checks that when the accounts are not loaded, the button is disabled.

dskloetd: On which line does it check that the button is disabled?
On line 109 it checks that the button is **enabled** and I don't see any other line where it checks the disabled state of the button.

lmuntaner: oh, true. This seems wrong.

This doesn't seem correct. We should fix it.

dskloetd: Yes, that's why I'm asking if you remember the original intent of the test :-).

dskloetd: Oh, are you saying the behavior is broken? I was assuming the test just tests something different.

lmuntaner: It seems broken, yes.

If no account is selected, the button should be disabled.

But I remember seeing it disabled while testing. Therefore, I believe that the bug is in the test it self.

dskloetd: Actually, at some point the test changed from being about accounts to being about the swap commitment.
But it continued stating that the button should be disabled while checking that the button is enabled:
https://github.com/dfinity/nns-dapp/blob/78eb5a04cf02f0571b0e1a827455faad15ecd8dc/frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts#L234

PR #1082, opened 2022-07-04
L2-803: Bug, Merge maturity enabled w/out maturity

lmuntaner:
# Motivation

User should not see merge maturity enabled without maturity.

# Changes

* Update enable or disable with neuron props changes.

# Tests

Not applicable.


lmuntaner:

Not sure how we allowed this to happen. Could we add a lint rul that disables "const" if combined with "$"? Shouldn't Svelte have a rule for this?

peterpeterparker (approved):
LGTM, thanks

peterpeterparker:
Why not the eslint. Maybe there is none per default because sometimes one might want to imperatively access the value of a store at a given a time?

PR #1089, opened 2022-07-06
L2-800: Dropdown Select Account

lmuntaner:
# Motivation

User sees a nice Dropdown.

# Changes

* New Dropdown UI component.
* SelectAccountDropdown uses new Dropdown UI Component.

# Tests

* Test for Dropdown UI component
* Test for SelectAccountDropdown component

### DIFF
--- a/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -3,8 +3,9 @@
   import { accountsStore } from "../../stores/accounts.store";
   import { i18n } from "../../stores/i18n";
   import type { Account } from "../../types/account";
+  import type { SelectOption } from "../../types/common";
   import { getAccountFromStore } from "../../utils/accounts.utils";
-  import Spinner from "../ui/Spinner.svelte";
+  import Dropdown from "../ui/Dropdown.svelte";
 
   export let selectedAccount: Account | undefined = undefined;
   export let skipHardwareWallets: boolean = false;
@@ -15,16 +16,22 @@
     accountsStore: $accountsStore,
   });
 
-  let selectableAccounts: Account[] | undefined;
+  let selectableAccounts: SelectOption[] = [];
+  const convertAccountToSelectOption = (account: Account): SelectOption => ({
+    value: account.identifier,
+    label: account.name ?? $i18n.accounts.main,
+  });
   const unsubscribe = accountsStore.subscribe(
     ({ main, subAccounts, hardwareWallets }) => {
       if (main !== undefined) {
         selectedAccountIdentifier =
           selectedAccountIdentifier ?? main.identifier;
         selectableAccounts = [
-          main,
-          ...(subAccounts ?? []),
-          ...(skipHardwareWallets ? [] : hardwareWallets ?? []),
+          convertAccountToSelectOption(main),
+          ...(subAccounts?.map(convertAccountToSelectOption) ?? []),
+          ...(skipHardwareWallets
+            ? []
+            : hardwareWallets?.map(convertAccountToSelectOption) ?? []),
         ];
       }
     }
@@ -33,19 +40,9 @@
   onDestroy(unsubscribe);
 </script>
 
-<!-- TODO: Implement https://dfinity.atlassian.net/browse/L2-800 -->
-{#if selectableAccounts !== undefined}
-  <select
-    bind:value={selectedAccountIdentifier}
-    name="account"
-    data-tid="select-account-dropdown"
-  >
-    {#each selectableAccounts as account}
-      <option value={account.identifier}>
-        {account.name ?? $i18n.accounts.main_account}</option
-      >
-    {/each}
-  </select>
-{:else}
-  <Spinner />
-{/if}
+<Dropdown
+  name="account"
+  bind:value={selectedAccountIdentifier}
+  testId="select-account-dropdown"
+  options={selectableAccounts}
+/>
--- a/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -17,9 +17,12 @@
   });
 
   let selectableAccounts: SelectOption[] = [];
-  const convertAccountToSelectOption = (account: Account): SelectOption => ({
-    value: account.identifier,
-    label: account.name ?? $i18n.accounts.main,
+  const convertAccountToSelectOption = ({
+    identifier,
+    name,
+  }: Account): SelectOption => ({
+    value: identifier,
+    label: name ?? $i18n.accounts.main,
   });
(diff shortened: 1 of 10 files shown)


lmuntaner:

lmuntaner:
Dropdown is still a native dropdown.

New UX:

https://user-images.githubusercontent.com/4550653/177561396-79f9b402-8a33-4f31-8b34-945b5ec627ad.mov

Screenshots:

Screenshot 2022-07-06 at 14 30 49
Screenshot 2022-07-06 at 14 31 07

frontend/svelte/src/lib/components/ui/Dropdown.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+  import IconExpandMore from "../../icons/IconExpandMore.svelte";
+  import type { SelectOption } from "../../types/common";
+
+  // Do not allow to bind objects, since they are mutable.
```
peterpeterparker, line 5: what do you mean?

frontend/svelte/src/lib/components/ui/Dropdown.svelte
```diff
@@ -0,0 +1,52 @@
...
+  export let value: string;
+  export let name: string;
+  export let options: SelectOption[];
+  export let testId: string | undefined = undefined;
+</script>
+
+<div>
```
peterpeterparker, line 12: can't we spare this `div`?

frontend/svelte/src/lib/components/ui/Dropdown.svelte
```diff
@@ -0,0 +1,52 @@
...
+  <select bind:value {name} data-tid={testId}>
+    {#each options as option}
```
peterpeterparker, line 14: nth: deconstruct option

frontend/svelte/src/lib/components/ui/Dropdown.svelte
```diff
@@ -0,0 +1,52 @@
...
+<style lang="scss">
+  div {
+    width: 100%;
+    position: relative;
+
+    select {
+      padding: var(--padding-2x) var(--padding-3x);
+      border-radius: var(--element-border-radius);
+
+      appearance: none;
```
peterpeterparker, line 32: works in Safari and Firefox?

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
```diff
@@ -15,16 +16,22 @@
     accountsStore: $accountsStore,
   });
 
-  let selectableAccounts: Account[] | undefined;
+  let selectableAccounts: SelectOption[] = [];
+  const convertAccountToSelectOption = (account: Account): SelectOption => ({
```
peterpeterparker, line 20: nth: deconstruct 😉

frontend/svelte/src/lib/components/ui/Dropdown.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+  import IconExpandMore from "../../icons/IconExpandMore.svelte";
+  import type { SelectOption } from "../../types/common";
+
+  // Do not allow to bind objects, since they are mutable.
+  export let value: string;
+  export let name: string;
+  export let options: SelectOption[];
```
peterpeterparker, line 8: have you though about passing the options as `slot`? would it make the usage easier?

frontend/svelte/src/lib/components/ui/Dropdown.svelte
lmuntaner: I first set the value of the options to be the Account object, but then, when the update call finished and the store was substituted with the new values, the "selected" account disappeared. The reason was that the old value was the account object got during the query, and the object even though it had the same information it was not the same object.

That got me thinking that since we do this query/update in many places, it might be risky to use objects as values. Instead, it will be better if we use the identifier of the object.

I will improve the comment. No need to talk about "bind" actually.

frontend/svelte/src/lib/components/ui/Dropdown.svelte
lmuntaner: good point. I had to change the border because it was set in FF.

I also checked in dark mode, and it wasn't looking so good, so I added a background as well.

More on screenshots

frontend/svelte/src/lib/components/accounts/SelectAccountDropdown.svelte
lmuntaner: hehe, yes

frontend/svelte/src/lib/components/ui/Dropdown.svelte
lmuntaner: Do you mean this:

```
<Dropdown
  name="account"
  bind:value={selectedAccountIdentifier}
  testId="select-account-dropdown"
>
  {#each selectableAccounts as { identifier, name }}
    <option value={identifier}>{name ?? $i18n.accounts.main}</option>
  {/each}
</Dropdown>
```

I like it better now, but if we even want to change the Dropdown to something custom made, then we'll probably need the options as a string of objects. As long as we stay with the native dropdown this is ok.

What do you think?

lmuntaner:
I changed the screenshtos.

There is no border and it should have a shadow.

I also changed the background so that it appears different in dark mode.

Screenshot 2022-07-07 at 08 02 37
Screenshot 2022-07-07 at 08 02 46

lmuntaner:

frontend/svelte/src/lib/components/ui/Dropdown.svelte
peterpeterparker: Thanks for the feedback. I guess it's because the  HTML `option` value is a `string` (https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement9 - i.e. not support object or maybe I am misunderstanding something.

frontend/svelte/src/lib/components/ui/Dropdown.svelte
peterpeterparker: thanks

frontend/svelte/src/lib/components/ui/Dropdown.svelte
lmuntaner: Let's do a call on Monday otherwise.

frontend/svelte/src/lib/components/ui/Dropdown.svelte
lmuntaner: Ok, sounds good to me.

lmuntaner: > You like better with slot or without with an array property?

I meant I prefer using `slot`, but I wasn't sure using that directly would enable us to move to a custom dropdown. But if we use our own DropdownItem, it will be possible.

The only part that I don't like is that the DropdownItem is not an independent component and needs to go always inside a Dropdown. But it's worth the coupling.

peterpeterparker (approved):
LGTM, thx

PR #1090, opened 2022-07-06
feat: load snses

peterpeterparker:
# Motivation

List the deployed Snses with `nns-js` and load actors and summary data with `sns-js`.

# Changes

- use `nns-js` - Sns wasm canisters - to list all deployed sns projects
- use `sns-js` to init Sns wrappers - i.e. query root canisters, discover all canister ids of the Sns and load related canisters
- refactor mock code to split "list and query summaries" and "list and query swap state"
- add lots of TODO and related jira task
- integrate new api with mock data so that the UI can still be showcased and used
- query and update calls for newly integrated features
- bump agent-js v0.12.0 (it's been a while since we did not install nightly version)
- lazy load nns-js and sns-js features through a proxy (proxy is needed to mock tests easily)
- refactor error that override constructor when not needed (message is default)

# Tests

No extended tests provided as mock data are still used. Being said, `listSummaries` is already set including the code to mock sns/nns-js.

# Achitecture

Defined with @mstrasinskis we have decided to implement a query+update calls strategy, at least for the first version, that covers all the interaction. So there are two tracks, query on a sns project, use a root id that has been resolved with a query on sns wasm canister and performs only query calls within the sns project. likewise, update calls on sns project, use a root id that has been resolved with an update call on sns wasm canister and performs only update calls within the sns project.

Untitled-2022-05-18-1339

### DIFF
--- a/frontend/svelte/src/lib/api/sns.api.ts
+++ b/frontend/svelte/src/lib/api/sns.api.ts
@@ -18,15 +18,6 @@ type RootCanisterId = string;
 let snsQueryWrappers: Map<RootCanisterId, SnsWrapper> | undefined;
 let snsUpdateWrappers: Map<RootCanisterId, SnsWrapper> | undefined;
 
-// TODO(L2-829): to be deleted
-export const mockAbout5SecondsWaiting = <T>(generator: () => T): Promise<T> =>
-  new Promise((resolve) =>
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
LGTM, thank you.

PR #1094, opened 2022-07-07
chore: Update dfx to version 0.10.1

bitdivine:
# Motivation
Deploying an sns test environment involves uploading wasm files using dfx.  Large arguments such as this are not supported by dfx 0.9.2.  They will be supported in dfx `0.10.2`.  Updating to `0.10.1` now reduces the risk of delay updating to `0.10.2` after it has been released.

# Changes
* Update the dfx version to 0.10.1
* Specified the subnet type in `dfx.json`, as message sizes are now checked by `dfx run` and the default subnet type is `application` which is too small.

# Tests
- [x] CI works
- [x] Local deployments to dfx work
- [x] Deployments to a testnet work (deployed to small11)


dfx.json
```diff
@@ -117,6 +117,9 @@
     }
   },
   "defaults": {
+    "replica": {
+      "subnet_type": "system"
```
bitdivine, line 121: Message types are now enforced.  The default subnet type is application, for which nns canisters and our own nns-dapp are too large, so deployment fails unless we set the correct value.

deploy.sh
```diff
@@ -224,7 +224,7 @@ if [[ "$DEPLOY_NNS_DAPP" == "true" ]]; then
   #        includes other canisters for testing purposes.  If testing you MAY wish
   #        to deploy these other canisters as well, but you probbaly don't.
   dfx canister --network "$DFX_NETWORK" create nns-dapp --no-wallet || echo "canister may have been created already"
-  dfx deploy --network "$DFX_NETWORK" nns-dapp
+  dfx deploy --network "$DFX_NETWORK" nns-dapp --no-wallet
```
bitdivine, line 227: This now needs to be set explicitly.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+* Update dfx to version 10.1
```
bitdivine, line 4: If we fill in the changelog as we go, deployment should be quicker! :-)

dfx.json
```diff
@@ -1,5 +1,5 @@
 {
-  "dfx": "0.9.2",
+  "dfx": "0.10.1",
```
bitdivine, line 2: This is the key change.

peterpeterparker (approved):

dfx.json
lmuntaner: 🤯

lmuntaner (approved):

PR #1095, opened 2022-07-07
Check whether the installed docker is recent enough

bitdivine:
# Motivation
Updating docker can fix build problems.  So, check whether docker is old and if so ask the user to update.

# Changes
* Check whether the docker major version number is at least 10.

# Tests


mstrasinskis (approved):
LGTM

PR #1097, opened 2022-07-08
Fix: Error message when toggling community fund

lmuntaner:
# Motivation

Fix notification message when toggling community fund.

# Changes

* Set the message before making the call.

# Tests

* Only messaging changes.


dubovitskaya (approved):
thanks!

PR #1100, opened 2022-07-11
L2-814: Icon Review Participation

lmuntaner:
# Motivation

Add icon in the screen to confirm the participation.

# Changes

* New IconSouth component.
* Add icon in the ReviewParticipation screen.

# Tests

Only UI changes.


lmuntaner:
New UI:

Screenshot 2022-07-11 at 11 04 09
Screenshot 2022-07-11 at 11 04 22

lmuntaner:
@mstrasinskis please review

frontend/svelte/src/lib/components/project-detail/ReviewParticipate.svelte
```diff
@@ -44,8 +45,10 @@
       </p>
     </div>
     <div class="highlight">
+      <span class="icon">
```
mstrasinskis, line 48: Is it possible to apply styling to the svg to avoid extra wrapper?

frontend/svelte/src/lib/icons/IconSouth.svelte
```diff
@@ -0,0 +1,8 @@
+<!-- Source: https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Asouth%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4048 -->
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  height="48"
+  width="48"
+  fill="currentColor"
```
mstrasinskis, line 6: `viewBox="0 0 48 48"`?

mstrasinskis (approved):
LGTM. Just a couple of questions.

frontend/svelte/src/lib/components/project-detail/ReviewParticipate.svelte
lmuntaner: I prefer that the color comes from the wrapper, and in this case only the icon is color primary not the text.

If I add the color to the svg, it will be harder to reuse and not following the pattern we use so far.

Maybe we want to pass the color as prop, but maybe we could do that in another PR and do it for all icons.

mstrasinskis: Thanks for the explanation. It's cool as it is 👍

PR #1101, opened 2022-07-11
Install idl2json

bitdivine:
# Motivation
When sns canisters are created via the sns-wasms canister, not dfx, they are no longer added to the `canister_ids.json`.  It is possible to use `idl2json` and `jq` to use the `sns` cli response to populate `canister_ids.json`, and so create  
a better developer experience.  This requires `idl2json` to be available.

# Changes
* Add code to install idl2json to the setup script

# Tests
* Installation works on:
  - [x] Linux
  - [x] Mac M1
  - [x] Mac Intel

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -203,12 +203,15 @@ install_sns_darwin() {
 
 install_idl2json() {
   (
+    set -euo pipefail # These are not inherited in a subsell.
     build_dir="$(mktemp -d)"
     cd "$build_dir"
     git clone https://github.com/dfinity/idl2json.git
     cd idl2json
     cargo build --release
     mv target/release/idl2json "$USER_BIN"
+    cd /tmp
+    rm -fr "${build_dir:-this-is-not-the-dir-you-are-looking-for}" # This is defensive code; build_dir SHOULD be defined.
   )
 }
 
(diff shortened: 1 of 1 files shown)


scripts/setup
```diff
@@ -60,6 +60,9 @@ if test -z "${USER_PROFILE:-}"; then
   }
 fi
```
bitdivine, line 61: This pair of lines is factored out from several installers below.

scripts/setup
```diff
@@ -191,21 +190,28 @@ install_ic_admin_darwin() {
...
 }
 install_sns_darwin() {
-  local USER_BIN="$HOME/.local/bin"
-  mkdir -p "$USER_BIN"
   curl "https://download.dfinity.systems/ic/${IC_COMMIT}/nix-release/x86_64-darwin/sns.gz" | gunzip >"$USER_BIN/sns"
   chmod +x "$USER_BIN/sns"
   # shellcheck disable=SC2016
   command -v sns ||
     append_to_profile "export PATH=\"\$PATH:${USER_BIN}\""
 }
```
bitdivine, line 203: The installation itself.  `idl2json` is not available as a prebuilt binary.

Dfinity-skaestle (commented):
Thanks Max!

Dfinity-skaestle (approved):

PR #1103, opened 2022-07-12
Bump IC commit to support SNS

bitdivine:
# Motivation
Spawning an SMS via the wasm canister requires a newer version of the IC repo.

# Changes
- Bump IC_COMMIT to head of master.

# Tests
_ I have deployed an SNS with this commit.  There are still integration problems but this is still progress.


bitdivine:
Integrated in this PR: https://github.com/dfinity/nns-dapp/pull/1104

PR #1104, opened 2022-07-12
Update wasms and dids

bitdivine:
# Motivation & Changes
There are several small updates to the SNS wasm files:
* Bump IC commit
* The swap canister is now available for download, so we don't need to build it any more.
* The ic-js repo uses different names for the did files; the names are are made consistent with that.

# Tests
* I have verified that after running the nns download and build scripts, all the did and wasm files listed in dfx.json are present and correct.


e2e-tests/scripts/nns-canister.Dockerfile
```diff
@@ -56,17 +56,13 @@ RUN binary="cycles-minting-canister" && \
...
-COPY --from=builder /ic/rs/rosetta-api/ledger.did /ledger.private.did
-COPY --from=builder /ic/rs/rosetta-api/ledger_canister/ledger.did /ledger.public.did
 COPY --from=builder /ic/rs/nns/governance/canister/governance.did /governance.did
-COPY --from=builder /ic/rs/sns/sale/canister/sale.did /swap.did
-COPY --from=builder /ic/rs/sns/root/canister/root.did /root.did
+COPY --from=builder /ic/rs/rosetta-api/ledger.did /ledger.private.did
+COPY --from=builder /ic/rs/rosetta-api/ledger_canister/ledger.did /ledger.did
+COPY --from=builder /ic/rs/nns/gtc/canister/gtc.did /genesis_token.did
+COPY --from=builder /ic/rs/nns/sns-wasm/canister/sns-wasm.did /sns_wasm.did
+COPY --from=builder /ic/rs/sns/swap/canister/swap.did sns_swap.did
```
peterpeterparker, line 65: Just in case it would be an issue, notice a `/` is missing in comparison to other copy for swap and root

```
COPY --... /genesis_token.did
COPY --... /sns_wasm.did
COPY --... sns_swap.did
COPY --... sns_root.did
```

peterpeterparker (approved):

e2e-tests/scripts/nns-canister.Dockerfile
bitdivine: Good catch.  It probably doesn't matter but it i sbest to be consistent.

PR #1106, opened 2022-07-12
refactor: move and rename suffix api constants

peterpeterparker:
# Motivation

Rename api constants and errors to `something.api.ts` to `api.something.ts` and move in related packages.

Doing so goal of files is more obvious, adjust pattern (`*.api.ts` are api wrappers) while remaing focused to api only.

# Changes

- renae and move `constants.api.ts` and `errors.api.ts`


lmuntaner (approved):
LGTM, but this makes the `api` folder dependent on the project. We won't be able to "move" the folder easily to another place and reuse it or even create a library if we wanted to.

Not that we have any kind of plan in the short term, but it's worth mentioning.

We also use some "utils" anyway.

peterpeterparker:
> We won't be able to "move" the folder easily to another place and reuse it or even create a library if we wanted to.

Yep aware of that. Thanks for the review.

PR #1109, opened 2022-07-12
Read build config from JSON

bitdivine:
# Motivation
At the moment, the IC_COMMIT appears in mutiple places in the codebase and it is possible to accidentally update some but not all the commits.  The same is true of dfx version, rust version and so on.  It would be nice to have this build configuration concentrated in one place.

# Changes
- Create tooling for defining build configuration in the `dfx.json` `build` section.
- Use this configuration for `IC_COMMIT`.  Other values can follow later.

# Tests
- [x] CI passes
- [x] Local deployment works
- [x] Deployment to a testnet works


peterpeterparker (approved):
rubber stamp, thx

PR #1110, opened 2022-07-12
L2-824: Elements in Dark Mode

lmuntaner:
# Motivation

Adapt some elements to Dark mode for the new design.

# Changes

* Use css variables that are changed per theme for inputs.
* Change secondary button style to adapt to new dark model

# Tests

Only style changes.

### DIFF
--- a/frontend/svelte/src/lib/themes/button.scss
+++ b/frontend/svelte/src/lib/themes/button.scss
@@ -116,10 +116,10 @@ button {
       cursor: default;
     }
 
-    background: var(--card-background);
+    background: var(--secondary);
+    color: var(--secondary-contrast);
     // Use inset border to avoid messing up with size.
     border: 1px inset var(--input-error-color, var(--input-border-color));
-    color: var(--secondary-text-color);
 
     @include effect.ripple-effect(--primary-contrast);
 
--- a/frontend/svelte/src/lib/themes/button.scss
+++ b/frontend/svelte/src/lib/themes/button.scss
@@ -68,6 +68,8 @@ button {
   &.secondary,
   &.danger,
   &.warning {
+    padding: var(--padding) var(--padding-2x);
+
     border-radius: var(--element-border-radius);
     box-shadow: var(--interaction-box-shadow);
 
@@ -106,6 +108,10 @@ button {
       background: var(--primary-tint);
       @include effect.ripple-effect(--primary-tint);
     }
+
+    &:hover {
+      background: var(--primary-tint);
+    }
   }
 
   &.secondary {
@@ -121,19 +127,11 @@ button {
     // Use inset border to avoid messing up with size.
     border: 1px inset var(--input-error-color, var(--input-border-color));
 
-    @include effect.ripple-effect(--primary-contrast);
+    @include effect.ripple-effect(--secondary-tint);
 
     &:focus {
-      background: var(--card-background-tint);
-      @include effect.ripple-effect(--card-background-tint);
-    }
-  }
-
-  &.primary {
-    padding: var(--padding) var(--padding-2x);
-
-    &:hover {
-      background: var(--primary-tint);
+      background: var(--secondary-tint);
+      @include effect.ripple-effect(--secondary-tint);
     }
   }
 
--- a/frontend/svelte/src/lib/themes/button.scss
+++ b/frontend/svelte/src/lib/themes/button.scss
@@ -116,15 +116,16 @@ button {
       cursor: default;
     }
 
-    // Same colors as primary but reversed
-    background: var(--primary-contrast);
-    color: var(--primary);
+    background: var(--card-background);
+    // Use inset border to avoid messing up with size.
+    border: 1px inset var(--input-error-color, var(--input-border-color));
+    color: var(--secondary-text-color);
 
     @include effect.ripple-effect(--primary-contrast);
 
     &:focus {
-      background: var(--primary-contrast);
-      @include effect.ripple-effect(--primary-contrast);
+      background: var(--card-background-tint);
+      @include effect.ripple-effect(--card-background-tint);
     }
   }
 
--- a/frontend/svelte/src/lib/themes/themes/light.scss
+++ b/frontend/svelte/src/lib/themes/themes/light.scss
@@ -22,7 +22,6 @@
 
     // Inputs
     --input-border-color: transparent;
-    --input-checked-background: var(--primary);
 
     // Menu
     --menu-background: #fafbfc;
--- a/frontend/svelte/src/lib/themes/themes/light.scss
+++ b/frontend/svelte/src/lib/themes/themes/light.scss
@@ -32,8 +32,6 @@
     --disable-contrast: #858585;
 
     // Palette
-    --secondary-text-color: var(--primary);
-
     --background: #f4f5f6;
     --background-rgb: 244, 245, 246;
     --background-contrast: #333333;
@@ -47,5 +45,12 @@
     --card-background-contrast-rgb: 51, 51, 51;
     --card-background-shade: #e0e0e0;
     --card-background-tint: #ffffff;
+
+    --secondary: #ffffff;
+    --secondary-rgb: 255, 255, 255;
+    --secondary-contrast: var(--primary);
+    --secondary-contrast-rgb: 0, 165, 255;
+    --secondary-shade: #e0e0e0;
+    --secondary-tint: #ffffff;
   }
 }
--- a/frontend/svelte/src/lib/themes/themes/light.scss
+++ b/frontend/svelte/src/lib/themes/themes/light.scss
@@ -20,6 +20,10 @@
     --json-principal-color: var(--warning-emphasis);
     --json-hash-color: var(--warning-emphasis);
 
+    // Inputs
+    --input-border-color: transparent;
+    --input-checked-background: var(--primary);
+
     // Menu
     --menu-background: #fafbfc;
     --menu-background-contrast: #333333;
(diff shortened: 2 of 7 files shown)


lmuntaner:
New UI:

Screenshot 2022-07-12 at 11 41 00
Screenshot 2022-07-12 at 11 41 19
Screenshot 2022-07-12 at 11 41 32

lmuntaner:

peterpeterparker:
the border on the secondary button makes it quite primary for my early morning eyes but gonna be used to it :wink:.
can you provide a screenshot of the border and input when focused or hovered?

lmuntaner:
The different effects of buttons won't make it to this MVP, Mischa has no time.

I raised the question here: https://www.notion.so/New-and-SNS-UI-Questions-be9394e103ea4f49a241338bc502d692#9a25e24da7534d2b9645a43ea796d27a

There is no hover style and this is the focus style:
Screenshot 2022-07-13 at 07 26 28

Normal to compare:
Screenshot 2022-07-13 at 07 27 33

frontend/svelte/src/lib/themes/button.scss
```diff
@@ -116,15 +116,16 @@ button {
       cursor: default;
     }
 
-    // Same colors as primary but reversed
-    background: var(--primary-contrast);
-    color: var(--primary);
+    background: var(--card-background);
+    // Use inset border to avoid messing up with size.
+    border: 1px inset var(--input-error-color, var(--input-border-color));
+    color: var(--secondary-text-color);
```
peterpeterparker, line 122: why not `--card-background-contrast`?

frontend/svelte/src/lib/themes/themes/light.scss
```diff
@@ -29,6 +33,8 @@
     --disable-contrast: #858585;
 
     // Palette
+    --secondary-text-color: var(--primary);
```
peterpeterparker, line 36: if `--secondary-text-color` is kept, you don't need this here I think

frontend/svelte/src/lib/themes/themes/light.scss
```diff
@@ -20,6 +20,10 @@
     --json-principal-color: var(--warning-emphasis);
     --json-hash-color: var(--warning-emphasis);
 
+    // Inputs
+    --input-border-color: transparent;
+    --input-checked-background: var(--primary);
```
peterpeterparker, line 25: why primary and not contrasting color?

frontend/svelte/src/lib/themes/button.scss
```diff
@@ -116,15 +116,16 @@ button {
       cursor: default;
     }
 
-    // Same colors as primary but reversed
-    background: var(--primary-contrast);
-    color: var(--primary);
+    background: var(--card-background);
```
peterpeterparker, line 119: you can use the form mixin for the background and border I think?

lmuntaner: Yes, but I though it looked a little bit ugly. Even though the rules are the same, I didn't consider them to have the same nature.

I can use it though.

frontend/svelte/src/lib/themes/button.scss
lmuntaner: the color for light needs to be primary, not `333333`.

lmuntaner: I thought this could also be used for links. I think they behave the same way

frontend/svelte/src/lib/themes/themes/light.scss
lmuntaner: I thought I saw somewhere it was primary, but I just asked Mischa because I don'f find it. I'll let you know.

frontend/svelte/src/lib/themes/themes/light.scss
lmuntaner: What do you mean? Would you put it somewhere else? Where?

frontend/svelte/src/lib/themes/button.scss
lmuntaner: I realised I can't use it, the border here is `inset` to not mess up with widths.

Unless we want to add widths to all the buttons, but just transparent. Not sure something to be done in this PR though.

frontend/svelte/src/lib/themes/button.scss
peterpeterparker: light.scss -> --card-background-contrast: #333333;

peterpeterparker: then maybe it's the point where we should define a `--secondary` color rather than reusing existing `--card-background` for secondary button. it will make the css more readable and maintable.

frontend/svelte/src/lib/themes/themes/light.scss
peterpeterparker: you can ignore this comment.

frontend/svelte/src/lib/themes/button.scss
peterpeterparker: oki doki, thanks for the feedback

frontend/svelte/src/lib/themes/button.scss
lmuntaner: I don't understand, sorry. The `background: var(--card-background);` works. It matches the colors.

What doesn't match is the `color: var(--secondary-text-color);`. Because I need the primary in light mode.

What do you propose, to not use the `--card-background` for the `background`?

frontend/svelte/src/lib/themes/themes/light.scss
lmuntaner: I was wrong, this needs to be the same as card-background. I will remove the `--input-checked-background` variable.

frontend/svelte/src/lib/themes/button.scss
peterpeterparker: Because to compose the button secondary it would now introduce one particular new variable (`--secondary-text-color`)  but still use other styles (e.g. `--card-background`). Therefore kind of think it becomes quite a "complex" composition and for that reason, for readability and maintainability, kind of think that introducing a `--secondary` set of colors would be better.

in variable and themes:

```
  --secondary: #xxxxx;
  --secondary-rgb: 0, xxx, 255;
  --secondary-contrast: #xxxx;
  --secondary-contrast-rgb: 255, xxx, 255;
  --secondary-shade: #xxxx;
  --secondary-tint: #xxxx;
  ```
  
  then used in button
  
 ```
 .secondary {
    background: var(--secondary....
    ....
```

frontend/svelte/src/lib/themes/themes/light.scss
lmuntaner: done

lmuntaner:
New checkbox UI when checked for light theme:

Screenshot 2022-07-13 at 11 28 00

frontend/svelte/src/lib/themes/button.scss
lmuntaner: Do you mean something like:

```
    background: var(--secondary);
    color: var(--secondary-contrast);
```

??

lmuntaner: Not sure how to create the tint, shade for those one. And do we need the rgb as well?

lmuntaner: done!

lmuntaner:

frontend/svelte/src/lib/themes/button.scss
```diff
@@ -116,15 +116,16 @@ button {
...
 
-    // Same colors as primary but reversed
-    background: var(--primary-contrast);
-    color: var(--primary);
+    background: var(--secondary);
+    color: var(--secondary-contrast);
+    // Use inset border to avoid messing up with size.
+    border: 1px inset var(--input-error-color, var(--input-border-color));
 
     @include effect.ripple-effect(--primary-contrast);
```
peterpeterparker, line 124: `--secondary-contrast` instead of primary?

frontend/svelte/src/lib/themes/button.scss
```diff
@@ -116,15 +116,16 @@ button {
...
 
     &:focus {
-      background: var(--primary-contrast);
-      @include effect.ripple-effect(--primary-contrast);
+      background: var(--card-background-tint);
```
peterpeterparker, line 127: `--secondary-contrast` instead of `--card-background-tint`?

frontend/svelte/src/lib/themes/button.scss
lmuntaner: yes, good one, thanks!

frontend/svelte/src/lib/themes/button.scss
lmuntaner: maybe better "--secondary-tint"

I think that we should refactor a little bit the buttons css. I might do that but in another PR

frontend/svelte/src/lib/themes/button.scss
lmuntaner: Almost, better --secondary-tint

https://user-images.githubusercontent.com/4550653/178730918-6de37f0b-b24d-46a1-badd-b0f097244889.mov

I will use the --secondary-tint

frontend/svelte/src/lib/themes/button.scss
lmuntaner: done

frontend/svelte/src/lib/themes/button.scss
lmuntaner: done

lmuntaner:

PR #1111, opened 2022-07-12
ENABLE_NEW_SPAWN_FEATURE feature flag

bitdivine:
# Motivation

Add the functionality to use env vars as feature flags.

# Changes

* Add ENABLE_NEW_SPAWN_FEATURE feature flag.

# Tests

Only env var changes.


dfx.json
```diff
@@ -68,24 +69,27 @@
       "providers": ["https://ic0.app/"],
       "type": "persistent"
     },
-    "testnet": {
+    "staging": {
```
bitdivine, line 72: We don't have flutter any more so I have renamed nns_both to staging, so that we can have features enabled on "testnet" but disabled and prod-like on "staging".

dfx.json
```diff
@@ -68,24 +69,27 @@
...
+        "ENABLE_XXX_FEATURE": false,
         "HOST": "https://nnsdapp.dfinity.network",
         "OWN_URL": "https://OWN_CANISTER_ID.nnsdapp.dfinity.network",
         "IDENTITY_SERVICE_URL": "https://qjdve-lqaaa-aaaaa-aaaeq-cai.nnsdapp.dfinity.network"
       },
       "providers": ["http://[2a00:fb01:400:42:5000:d1ff:fefe:987e]:8080"],
       "type": "persistent"
     },
-    "nnsdapp_both": {
+    "testnet": {
```
bitdivine, line 85: Testnet is actually unchanged apart from adding `"ENABLE_XXX_FEATURE": true,`; moving the former "both" up so that staging is next to prod confuses the diff tool a bit.

frontend/svelte/src/main.ts
```diff
@@ -1,5 +1,8 @@
 import App from "./App.svelte";
```
bitdivine: Please change this to whatever you want.

lmuntaner (approved):
LGTM!

dfx.json
```diff
@@ -60,6 +60,7 @@
         "BUILD": "release",
         "FETCH_ROOT_KEY": false,
         "REDIRECT_TO_LEGACY": "prod",
+        "ENABLE_XXX_FEATURE": false,
```
peterpeterparker, line 63: instead of boolean, would it be nice to have an object or list or what ever?  maybe there is more than one feature that we would like to unleash, like having features a and b on testnet but c on another testnet for the community

lmuntaner: yeah, even a yml file of features in the end as well.

I hope we can remove this env var next week when we release it to prod and think of a sustainable way to do it if we have time.

frontend/svelte/src/lib/constants/environment.constants.ts
```diff
@@ -2,6 +2,8 @@ export const DFX_NETWORK: string = String(process.env.DFX_NETWORK);
 export const HOST: string = process.env.HOST as string;
 export const ROLLUP_WATCH: boolean = process.env.ROLLUP_WATCH === "true";
 export const FETCH_ROOT_KEY: boolean = process.env.FETCH_ROOT_KEY === "true";
+export const ENABLE_NEW_SPAWN_FEATURE: boolean =
+  process.env.ENABLE_NEW_SPAWN_FEATURE === "true";
```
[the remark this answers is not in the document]
peterpeterparker: I don't have time to make that change so good to know it would not need much effort to switch from boolean to object. Let's move forward then and thanks @bitdivine !

PR #1117, opened 2022-07-14
Fix: Handle case where canister_ids.json does not exist yet

bitdivine:
# Motivation
When deploying an sns, the case where there is no existing `canister_ids.json` is not handled yet.  It should be.

# Changes
* Back up the `canister_ids.json` only if it exists
* If there is no `canister_ids.json`, create one with an empty dictionary.

# Tests
I have used the code like this: `rm canister_ids.json ; ./deploy.sh --sns small11`.  This part of the deployment worked fine.


lmuntaner (approved):
LGTM, thanks!

PR #1120, opened 2022-07-14
Bump ic version

bitdivine:
# Motivation

# Changes

# Tests

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -146,7 +146,7 @@
       "config": {
         "RUSTC_VERSION": "1.58.1",
         "IC_CDK_OPTIMIZER_VERSION": "0.3.1",
-        "IC_COMMIT": "28ea02bad62dc777378a53602b8cd3a515ab23a3"
+        "IC_COMMIT": "3409f63139688e8d1da5beeeb2efa181f1ba1b1f"
        },
       "packtool": ""
     }


lmuntaner (approved):
I tried in my PR but didn't know which to show, so thanks!

e2e-tests/scripts/nns-canister.Dockerfile
```diff
@@ -50,17 +50,18 @@ RUN binary="governance-canister" && \
...
-# Note: This is available as a download however we patch the source code to make testing easier.
-RUN binary="cycles-minting-canister" && \
-    cargo build --target wasm32-unknown-unknown --release -p "$binary"
-RUN binary="cycles-minting-canister" && \
-    ic-cdk-optimizer -o "$CARGO_TARGET_DIR/${binary}.wasm" "$CARGO_TARGET_DIR/wasm32-unknown-unknown/release/${binary}.wasm"
+## Note: This is available as a download however we patch the source code to make testing easier.
+#RUN binary="cycles-minting-canister" && \
+#    cargo build --target wasm32-unknown-unknown --release -p "$binary"
+#RUN binary="cycles-minting-canister" && \
+#    ic-cdk-optimizer -o "$CARGO_TARGET_DIR/${binary}.wasm" "$CARGO_TARGET_DIR/wasm32-unknown-unknown/release/${binary}.wasm"
```
bitdivine, line 57: I kept this as we should be able to build the CMC canister.  We may not need a customized CMC any more, right now, but we probably will at some point in the future.  I'd like to keep this code in so that I can test whether a fix to the CMC dependencies lets us build again.

scripts/propose
```diff
@@ -69,18 +69,23 @@ while (($# > 0)); do
     exit 0
```
bitdivine, line 69: The exchange rate proposal command doesn't follow the convention used in ic-admin of starting with `propose-to-` so this needs to be adapted.

scripts/setup
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 cd "$(dirname "$0")/.."
-. <(../../build-config.sh)
+. <(./build-config.sh)
```
bitdivine, line 4: Fix: This script runs in the root of the repo, so the build config script is in the current directory.  It just so happens that workspace nesting is 2 directories deep so running this in a workspace (e.g. `~/nns-dapp/branches/foo`) called the copy in main (`~/nns-dapp/`) so it "worked" before.  I am surprised that CI didn't fail.  I'll have a look there.

e2e-tests/scripts/nns-canister.Dockerfile
lmuntaner: Does that mean the CMC is downloaded as is then? Where is the code that does that?

lmuntaner (approved):
Thanks for asking again, it is indeed a different task, hehe.

PR #1122, opened 2022-07-15
Feat/l2 840 context store for project detail page

mstrasinskis:
# Motivation

The Sns project detail page should not load and use the all store of summaries but only use data related to the particular project.
[L2-840](https://dfinity.atlassian.net/browse/L2-840)

Click on loaded project then refresh

https://user-images.githubusercontent.com/98811342/179159676-b08f6a29-9e83-4d82-8767-1d622afb3f49.mov

# Changes

- switch from launchpad stores to context store on project-details page
- use summaries and swaps certified entries if available

# Tests

- ProjectDetail loading tests

### DIFF
--- a/frontend/svelte/src/lib/api/sns.api.ts
+++ b/frontend/svelte/src/lib/api/sns.api.ts
@@ -2,7 +2,6 @@ import type { HttpAgent, Identity } from "@dfinity/agent";
 import type { DeployedSns, SnsWasmCanister } from "@dfinity/nns";
 import { Principal } from "@dfinity/principal";
 import type { InitSns, SnsWrapper } from "@dfinity/sns";
-import type { GetStateResponse } from "@dfinity/sns/dist/candid/sns_swap";
 import { mockSnsSummaryList } from "../../tests/mocks/sns-projects.mock";
 import { HOST } from "../constants/environment.constants";
 import {
@@ -350,7 +349,7 @@ export const querySnsSwapState = async ({
   // TODO: current deployed swap canister on testnet is empty
   // Error: "Message": "IC0304: Attempt to execute a message on canister mr56c-4qaaa-aaaaa-aacgq-cai which contains no Wasm module"
   // const { swap } = await swapState({});
-  let swap: GetStateResponse;
+  let swap: unknown;
   try {
     swap = await swapState({});
     console.log("swap", swap);
--- a/frontend/svelte/src/lib/api/sns.api.ts
+++ b/frontend/svelte/src/lib/api/sns.api.ts
@@ -347,6 +347,9 @@ export const querySnsSwapState = async ({
     certified,
   });
 
+  // TODO: current deployed swap canister on testnet is empty
+  // Error: "Message": "IC0304: Attempt to execute a message on canister mr56c-4qaaa-aaaaa-aacgq-cai which contains no Wasm module"
+  // const { swap } = await swapState({});
   let swap: GetStateResponse;
   try {
     swap = await swapState({});
@@ -354,10 +357,6 @@ export const querySnsSwapState = async ({
   } catch (err) {
     console.error("swap TBD");
   }
-
-  // TODO: current deployed swap canister on testnet is empty
-  // Error: "Message": "IC0304: Attempt to execute a message on canister mr56c-4qaaa-aaaaa-aacgq-cai which contains no Wasm module"
-  // const { swap } = await swapState({});
 
   logWithTimestamp(
     `Getting Sns ${rootCanisterId} swap state certified:${certified} done.`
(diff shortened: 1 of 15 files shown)


frontend/svelte/src/lib/api/sns.api.ts
```diff
@@ -348,6 +347,14 @@ export const querySnsSwapState = async ({
     certified,
   });
 
+  let swap: GetStateResponse;
```
peterpeterparker, line 350: maybe move this block after related TODO next under?

frontend/svelte/src/lib/components/project-detail/ProjectInfoSection.svelte
```diff
@@ -6,11 +6,22 @@
...
 
-  export let summary: SnsSummary;
+  const { store: projectDetailStore } = getContext<ProjectDetailContext>(
+    PROJECT_DETAIL_CONTEXT_KEY
+  );
 
-  const minCommitmentIcp = ICP.fromE8s(summary.minParticipationCommitment);
-  const maxCommitmentIcp = ICP.fromE8s(summary.maxParticipationCommitment);
+  let summary: SnsSummary;
+  $: summary = $projectDetailStore.summary as SnsSummary;
```
peterpeterparker, line 20: - the `as SnsSummary` is required?
- nth: deconstruct the `summary` variable

frontend/svelte/src/lib/utils/dev.utils.ts
```diff
@@ -94,3 +94,9 @@ export function triggerDebugReport(node: HTMLElement) {
     },
   };
 }
+
+export const shuffle = <T>(items: T[]): T[] =>
```
peterpeterparker, line 98: add a `TODO: to be deleted - use for mock data only`?

frontend/svelte/src/routes/ProjectDetail.svelte
```diff
@@ -32,29 +37,86 @@
...
+  });
+
+  // TODO: add projectDetailStore to debug store
+
+  setContext<ProjectDetailContext>(PROJECT_DETAIL_CONTEXT_KEY, {
+    store: projectDetailStore,
   });
 
+  const loadSummary = (rootCanisterId: string) => {
+    if ($snsSummariesStore.certified === true) {
```
peterpeterparker, line 53: any particular reason why to use only certified summary and swap?

in other detail pages do we use only certified?

frontend/svelte/src/routes/ProjectDetail.svelte
```diff
@@ -32,29 +37,86 @@
...
+    swapState: undefined,
+  });
+
+  // TODO: add projectDetailStore to debug store
+
+  setContext<ProjectDetailContext>(PROJECT_DETAIL_CONTEXT_KEY, {
+    store: projectDetailStore,
   });
 
+  const loadSummary = (rootCanisterId: string) => {
```
peterpeterparker, line 52: so if i get it right, if the summaries store already contains certified values, then we display the summary we have in store otherwise we query+update the summary from backend, is that correct?

do we have the same approach in other detail page and if no, any advantages/disadvantages?

looks good, just want to be sure i get it.

peterpeterparker:
nth: generally speaking in the components we could destruct `let summary` to `$: ({minCommintment....} = summary)`

frontend/svelte/src/routes/ProjectDetail.svelte
mstrasinskis: It's done to keep the code simple. I don't think we use this strategy anywhere. Before the separate payload endpoint, there were two separate loading functions. But agree, it's not optimal. Let's display the query data and call `loadSnsSummary` in that case.

frontend/svelte/src/lib/components/project-detail/ProjectInfoSection.svelte
mstrasinskis: Yep, because according to the type it can be undefined or null. The parent component prevents rendering w/o summary. Probably the best solution could be to move spinner etc. in this component.

peterpeterparker: Probably best yes but maybe in another PR then. Meanwhile can you add a comment that explains that the type safety is inherited because of a check in parent component?

frontend/svelte/src/routes/ProjectDetail.svelte
peterpeterparker: Yes it will make the ux faster.

peterpeterparker: That being said, checking out the canister detail page, it's a bit the same logic - i.e. considering the store as ready only when certified, therefore I get what you implemented. We can also go forward with this PR and if needed improve it afterwards.

mstrasinskis: @peterpeterparker the logic was updated

peterpeterparker (approved):
LGTM, thanks.

Before merge please just add a comment about the type casting `as Summary`, would be appreciated.

PR #1123, opened 2022-07-15
Re enable xdr

bitdivine:
# Motivation
When using stock wasm canister builds, the API to set the exchange rate directly is not available.  The exchange rate can however be set by proposal for both stock and custom canisters.  The `ic-admin` subcommand for this has been added recently.

# Changes
- Set the xdr rate by proposal rather than by a special API.
- Adjust the `propose-to` script as ic-admin proposal commands no longer always start with `-propose-to`.
- A fix.

# Tests
See CI


scripts/propose
```diff
@@ -69,18 +69,23 @@ while (($# > 0)); do
     exit 0
```
bitdivine, line 69: Previously all ic-admin proposal commands started with `propose-to-`.  That is no longer the case so this needs to be adapted slightly.

dfx.json
```diff
@@ -148,7 +148,7 @@
       "config": {
         "RUSTC_VERSION": "1.58.1",
         "IC_CDK_OPTIMIZER_VERSION": "0.3.1",
-        "IC_COMMIT": "3409f63139688e8d1da5beeeb2efa181f1ba1b1f"
+        "IC_COMMIT": "34f6a95c969daf54a73fde341196d5fb9c81ff2b"
```
bitdivine, line 151: The commit needs to be bumped to include this ic-admin subcommand: https://gitlab.com/dfinity-lab/public/ic/-/merge_requests/6224/diffs

scripts/setup
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 cd "$(dirname "$0")/.."
-. <(../../build-config.sh)
+. <(./build-config.sh)
```
bitdivine, line 4: Fix: This only "worked" before because git workspaces are 2 directories deep, so this was calling the script in the parent workspace.

lmuntaner (approved):
Don't we need a new version of ic-admin?

PR #1124, opened 2022-07-15
Feat/sns get state

mstrasinskis:
# Motivation

Added swapState loading (partly mocked). Will be updated as soon the endpoint is done.

# Changes

- swapState loading functions
- switched detail page data source from the launchpad stores to the context store (important for future changes)

# Tests

- context store related tests


peterpeterparker (approved):
LGTM, thanks

PR #1125, opened 2022-07-15
Fix: Do not fail spawning neurons

lmuntaner:
# Motivation

Governance canister will be deployed first with the new spawning functionality. Until the new nns-dapp is release, the spawning neurons break the UI and show a blank page.

Fix the error and show only 0 ICP and that's it. At this point, we can't use the new state nor the new property. We just need to avoid the blank page.

# Changes

* stateInfo can be undefined for unknown states.
* Do not show state information if stateInfo is undefined.

# Tests

* Type changes only

### DIFF
--- a/frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
+++ b/frontend/svelte/src/lib/components/neurons/NeuronCard.svelte
@@ -26,7 +26,7 @@
   export let disabled: boolean | undefined = undefined;
 
   // TODO: https://dfinity.atlassian.net/browse/L2-366
-  let stateInfo: StateInfo;
+  let stateInfo: StateInfo | undefined;
   $: stateInfo = getStateInfo(neuron.state);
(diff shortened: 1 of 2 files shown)


lmuntaner:
UI

Screenshot 2022-07-15 at 14 03 53

peterpeterparker:
No test? If I get it right a state info used to be displayed all the time but it ain't the case anymore following this PR.

lmuntaner:
It is displayed if there is such a state.

I couldn't do such a test because the edge case is not even in the enum. Therefore, a test would be with a value that is not even in the enum which could not transplie.

peterpeterparker:
Gotcha, so if I get it right, the `{#if...` will actually be removed later on?  I would have got it directly with a comment "" 😉.

lmuntaner:
Yes, correct. We won't need the `{# if`, but it might make sense to keep it in case there are other additions to the enum in the future.

PR #1129, opened 2022-07-18
feat: proposal open until duration expires

peterpeterparker:
# Motivation

A proposal should remain open for vote according its duration not its status because it could be accepted or declined when the majority is reached but still accept votes.

# Logic

To detect if a proposal is still open for voting, the field `deadline_timestamp_seconds` should be observed. Even though it is always set by the backend it is (for some unknown reason) set as optional in the definition. Therefore a fallback should be implemented. This PR implement such a fallback as the existing code of the dashboard.

# Changes

- new function `isProposalOpenForVote` that uses `deadlineTimestampSeconds` or fallback to some logic implemented as in the dashboard
- replace checks on status `OPEN` in list (filter "Hide Open proposal ....") and display of voting card
- bump nns-js


frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -318,3 +321,46 @@ export const mapProposalInfo = (
     status,
   };
 };
+
+/**
+ * A proposal can be accepted or declined if the majority votes before its duration expires but, it remains open for voting until then.
+ * That is why we should not consider the status "OPEN" to present a proposal as open for voting but consider the duration.
+ */
+export const isProposalOpenForVotes = (proposalInfo: ProposalInfo): boolean =>
+  votingPeriodEnd(proposalInfo).getTime() >= new Date().getTime();
```
mstrasinskis, line 330: Is the backend time available? Not critical at all, but if yes it would be great not to rely on the user local settings.

frontend/svelte/src/lib/utils/proposals.utils.ts
```diff
@@ -318,3 +321,46 @@ export const mapProposalInfo = (
...
+  if (deadlineTimestampSeconds === undefined) {
+    return votingPeriodEndFallback(proposalInfo);
+  }
+
+  return new Date(Number(deadlineTimestampSeconds) * 1000);
+};
+
+// Following fallback logic follows what's originally implemented by the Dashboard
+
+const SHORT_VOTING_PERIOD_SECONDS = 60 * 60 * 12; // 12 hours
```
mstrasinskis, line 349: constants.ts?

mstrasinskis (approved):
LGTM. A couple of minor suggestions/questions.

mstrasinskis:
> should this utility finds place as provided by this PR in NNS-dapp or should the logic - particularly because there is a fallback even if probably never used - to NNS-js?

Personally, I would leave it here for now 👍

frontend/svelte/src/lib/utils/proposals.utils.ts
peterpeterparker: I thought about it too but since they are used in a fallback that should never actually find place, I though they should somehow not exist on a global scale neither since they are to some extension useless. Does that make sense?

mstrasinskis: Thanks for explanation.

PR #1133, opened 2022-07-18
feat: scope proposal detail data

peterpeterparker:
# Motivation

Currently we use global stores to present the proposal data / information in the proposal detail page which can lead to some UI quirks if the user navigate quickly from one proposal to the other (list -> detail 1 -> go back list -> detail 2 etc.).

In addition, if the proposal detail page is loaded with an unknown proposal id, no error is thrown, the screen is displayed as loading for ever.

# Changes

- use governance `getProposal` instead of `listProposal` to fetch a particular proposal in api (otherwise we cannot detect invalid proposal id)
- introduce context store for proposal
- reload proposal of context store after user vote
- navigation and data loading update according new context store in `ProposalDetail` page


frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingCard.svelte
```diff
@@ -38,11 +42,23 @@
...
+
   const vote = async ({ detail }: { detail: { voteType: Vote } }) =>
     await registerVotes({
       neuronIds: $votingNeuronSelectStore.selectedIds,
       vote: detail.voteType,
       proposalId: proposalInfo.id as bigint,
+      reloadProposalCallback: (
+        proposalInfo: ProposalInfo // we update store only if proposal id are matching even though it would be an edge case that these would not match here
+      ) =>
+        store.update(({ proposalId, proposal }) => ({
```
mstrasinskis, line 58: Nice approach. Just thinking, does it make sense to do nothing when the `id` doesn't match 🤔

mstrasinskis (approved):
A couple of questions but a really good update, thanks

frontend/svelte/src/lib/api/proposals.api.ts
```diff
@@ -85,14 +85,9 @@ export const queryProposal = async ({
...
-  const response = await governance.listProposals({
-    request: {
-      limit: 1,
-      beforeProposal: proposalId + BigInt(1),
-      includeRewardStatus: [],
-      excludeTopic: [],
-      includeStatus: [],
-    },
+  // We use getProposal and not listProposals because if the request proposal id does not exist, listProposals would return a random proposal anyway
+  const response = await governance.getProposal({
```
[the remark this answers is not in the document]
peterpeterparker: Good point. Redo and done.

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingCard.svelte
peterpeterparker: I would say yes. Only scenario I can think of is user not waiting for spinner and updating manually the url with a new proposal id. If she/he does so, then another proposal id has been loaded therefore the context store should not be updated. Also not sure throwing an error would fit such scenario.

What do you think? Throwing an error or you had other ideas?

frontend/svelte/src/lib/api/proposals.api.ts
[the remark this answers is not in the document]
mstrasinskis: Thank you :)

frontend/svelte/src/lib/components/proposal-detail/VotingCard/VotingCard.svelte
mstrasinskis: It's a known and valid scenario, so I would just skip the response w/o any error.

peterpeterparker: Agree. So it is what's is implemented in this PR or should I implement it differently it is what you mean?

```
store.update(({ proposalId, proposal }) => ({
   proposalId,
          proposal: proposalId === proposalInfo.id ? proposalInfo : proposal,
        })),
    });
```

mstrasinskis: I thought about something like:
```ts
reloadProposalCallback: (
        proposalInfo: ProposalInfo // we update store only if proposal id are matching even though it would be an edge case that these would not match here
      ) => 
// trigger update only when actual data arrives
$store.proposalId === proposalInfo.id &&

        store.update(({ proposalId, proposal }) => ({
          proposalId,
          proposal: proposalInfo,
        })),
```
But the current implementation also works for me 👍

peterpeterparker: Let's go with current implementation then since I tested it but if I ever comeback to this, will update. Thanks!

PR #1135, opened 2022-07-19
Re-enable custom canister builds

bitdivine:
# Motivation
We would like to deploy testnets and local `dfx run` instances with custom NNS canisters.  We used to have this functionality but builds using the traditional methods are currently broken and being fixed.

So, with broken builds, how do end-users validate NNS wasm builds?  By running:
```
./gitlab-ci/tools/docker-build-ic --artifacts="canisters"
```
which runs
```
cargo build --target wasm32-unknown-unknown --profile canister-release --bin candid-test-canister --bin cycles-minting-canister --bin genesis-token-canister --bin governance-canister --bin governance-mem-test-canister --bin ic-icrc1-archive --bin ic-icrc1-ledger --bin ic-nervous-system-common-test-canister --bin identity-canister --bin inter_canister_error_handling --bin json --bin ledger-archive-node-canister --bin ledger-canister --bin mem-utils-test-canister --bin memory-test-canister --bin nan_canonicalized --bin nns-ui-canister --bin panics --bin pmap_canister --bin proxy_canister --bin registry-canister --bin response-payload-test-canister --bin root-canister --bin sns-governance-canister --bin sns-root-canister --bin sns-swap-canister --bin sns-test-dapp-canister --bin sns-governance-mem-test-canister --bin sns-wasm-canister --bin stable --bin statesync-test-canister --bin test-notified --bin time --bin upgrade-test-canister --bin wasm --bin xnet-test-canister
```
Note that this uses `--bin` instead of `-p`.

We cannot do exactly the same as we would like to pas in feature flags that differ by canister but that is a small change.

# Changes
* Fix custom cabister builds by using `--bin` rather than `-p`
* Re-enable the cmc canister build

# Tests
- [ ] Builds on CI
  - This should suffice as the build is identical regardless of deployment target.


e2e-tests/scripts/nns-canister.Dockerfile
```diff
@@ -36,25 +36,19 @@ WORKDIR /ic/rs
 #       Otherwise the built binary will simply not be deployed by ic-nns-init.
 RUN binary=ledger-canister && \
     features="notify-method" && \
-    cargo build --target wasm32-unknown-unknown --release -p "$binary" --features "$features"
-RUN binary=ledger-canister && \
```
bitdivine, line 40: Running the optimization in a separate step risks getting the feature flags out of sync.  Having a separate step made was useful when debugging `ic-cd-optimizer` builds but generally it is not useful.  If we wanted to keep separate steps we could put the feature flags in a docker env var.

lmuntaner (approved):
Rubber stamp

PR #1136, opened 2022-07-19
feat: use swap effective data and filter sns in launchpad according status

peterpeterparker:
# Motivation

Read, map and begin to display Sns swap state.

# Changes

- rename custom types and mock data `swapState` to `swapCommitment`
- install nightly sns-js and update types accordingly
- split commitment and swap state in `sns.api` and `sns.services`- i.e. commiment use mock data and implement effective swap state queries
- new derived stores `openProjectsStore` and `committedProjectsStore` (open for launchpad, committed for upcoming neuron drop down)
- new components `ProjectStatus` and `ProjectSwapDetails` that effectively display swap state data loaded from the backend
- concact and map loaded data with utilities in `sns.utils`. We load data through services and api with candid types but we map them in our `SnsSummary` and we also filter all optional values so that the UI can work without zillion of ifs.
- some utils copied from sns-js
- rollup config to group sns js in a dedicated chunk (otherwise rollup group it in `Logo.js` 🤣)


frontend/svelte/src/lib/utils/sns.utils.ts
```diff
@@ -1,18 +1,89 @@
...
+export const concatSnsSummary = ([summary, swap]: [
+  Omit<SnsSummary, "swap"> | undefined,
+  QuerySnsSwapState | undefined
+]): SnsSummary | undefined => {
+  if (!summary || !swap) {
+    return undefined;
+  }
+
+  // Not sure this should ever happen
+  const possibleSwap: Swap | undefined = fromNullable(swap?.swap);
```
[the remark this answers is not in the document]
peterpeterparker: Done. with still `fromNullable` though to convert `[] | [whatever]` to `?whatever`

frontend/svelte/src/lib/api/sns.api.ts
```diff
@@ -203,28 +207,29 @@ const wrapper = async ({
 };
 
 // TODO(L2-751): remove mock data
-let mockSnsSummaries: SnsSummary[] = [];
+let mockSnsSummaries: Omit<SnsSummary, "swap">[] = [];
```
mstrasinskis, line 210: It feels that `Omit<SnsSummary, "swap">` deserves it's own type.

frontend/svelte/src/lib/utils/did.utils.ts
```diff
@@ -0,0 +1,10 @@
+// These utils are duplicated in nns-js and sns-js, we might want to create a utility lib someday
+// We do not expose the utility in these libraries for that reason
+
+export const toNullable = <T>(value?: T): [] | [T] => {
```
mstrasinskis, line 4: nth: because there already is `isNullable` and `nonNullable` probably it make sense **later** to rename `toNullable`

frontend/svelte/src/lib/utils/sns.utils.ts
```diff
@@ -1,34 +1,79 @@
...
+      fromNullable(entry.swap.init) !== undefined &&
+      fromNullable(entry.swap.state) !== undefined
+  );
+
+  return validSwapSummaries.map(({ swap, ...rest }) => ({
+    ...rest,
+    swap: {
+      // We know for sure that init and state are defined because we check in previous filter that there are not undefined
+      // TODO: There might be a cleaner way than a type cast to make TypeScript checks these are defined
+      init: fromNullable(swap.init) as SnsSwapInit,
```
mstrasinskis, line 46: Later we could think about making `fromNullable` a generic type guard (not 100% sure it will work)

peterpeterparker: Not sure, it is not meant to be a guard but rather to convert the candid `[] | [something]` to an exploitable type in Javascript `?something`. There might be use cases where the undefined is useful.

frontend/svelte/src/lib/api/sns.api.ts
peterpeterparker: Done. `QuerySnsSummary`

frontend/svelte/src/lib/utils/did.utils.ts
peterpeterparker: Discussed, later on.

PR #1137, opened 2022-07-19
Use new ledger

bitdivine:
# Motivation
The SNS now uses a different ledger canister, and the swap canister can now be uploaded.

# Changes
* Download the new SNS wasm canister in addition to the existing ledger, which is still needed for the NNS.
* Get wasm paths from dfx.json when uploading wasm canisters
* Include the swap canister in the list that is uploaded

# Tests
- [x] Deploy an SNS to a testnet

### DIFF
--- a/e2e-tests/scripts/nns-canister.Dockerfile
+++ b/e2e-tests/scripts/nns-canister.Dockerfile
@@ -46,6 +46,8 @@ RUN binary="governance-canister" && \
 
 FROM scratch AS scratch
 COPY --from=builder /ic/rs/nns/governance/canister/governance.did /governance.did
+COPY --from=builder /ic/rs/rosetta-api/ledger.did /ledger.private.did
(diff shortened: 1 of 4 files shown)


e2e-tests/scripts/nns-canister.Dockerfile
```diff
@@ -30,18 +30,6 @@ RUN export CARGO_TARGET_DIR=/ic/rs/target && \
 ENV CARGO_TARGET_DIR=/ic/rs/target
 WORKDIR /ic/rs
```
bitdivine, line 32: We appear not to need a custom NNS ledger.  If we do need it again in future, it is easy enough to add, as long as it compileds.  In the meantime, we can save ourselves interruptions from the build breaking.

lmuntaner (approved):
Nice!

PR #1138, opened 2022-07-19
Make cycles fast

bitdivine:
# Motivation
Cycle minting rate limiting is now applied in the CMC canister.  Creating an SMS is expensive enough that we run into the rate limit quite quickly (although it seems to me that we hit the limit much too quickly - need to check).  If we hit the rate limiter, no more cycles can be minted for an hour, which significantly slows down test deployments, especially if there are bugs and the deployment needs to be repeated.

# Changes
* Reduce the rate limit time from 1 hour to 2 minutes.  Long enough that we can see that the limit exists but not so long that testnet deployment times are impacted seriously.
* Deleted a CMC patch that we don't need any more.

# Tests

### DIFF
--- a/e2e-tests/scripts/nns-canister.patch
+++ b/e2e-tests/scripts/nns-canister.patch
@@ -33,17 +33,15 @@ diff --git a/rs/nns/cmc/src/main.rs b/rs/nns/cmc/src/main.rs
 index 2c02d80dc..5a6072dc7 100644
 --- a/rs/nns/cmc/src/main.rs
 +++ b/rs/nns/cmc/src/main.rs
-@@ -106,8 +106,8 @@ impl State {
-             minting_account_id: None,
-             authorized_subnets: BTreeMap::new(),
-             default_subnets: vec![],
--            icp_xdr_conversion_rate: None,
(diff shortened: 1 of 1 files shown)


e2e-tests/scripts/nns-canister.patch
```diff
@@ -33,17 +33,15 @@ diff --git a/rs/nns/cmc/src/main.rs b/rs/nns/cmc/src/main.rs
 index 2c02d80dc..5a6072dc7 100644
 --- a/rs/nns/cmc/src/main.rs
 +++ b/rs/nns/cmc/src/main.rs
```
bitdivine, line 35: On th eleft is a patch we don't need any more - it is deleted.  It used to set xdr rate by default but now it is overridden so completely ineffective. The XDR rate needs to be set after deployment.

On the right is the new patch.  It reduces the waiting time when the cycle limit is hit from 1hour to 2 minutes.  Long enough to understand what is happening, not so long as to make testnet sns deployments super tedious.

mstrasinskis (approved):
Rubber stamp 👍

PR #1141, opened 2022-07-20
Sns wasm canister

bitdivine:
# Motivation
The SNS wasm canister ID is needed by the svelte front end.

# Changes
* Add `WASM_CANISTER_ID` as an environment variable generated from `dfx canister id wasm_canister`

# Tests
- [x] Works in a front end dev environment?


frontend/svelte/src/lib/api/sns.api.ts
```diff
@@ -68,7 +68,7 @@ const listSnses = async ({
   // TODO(L2-828): extract property for wasm canister id
```
peterpeterparker, line 68: you can remove that todo

peterpeterparker (approved):
LGTM, thx

PR #1144, opened 2022-07-20
feat: list sns projects workaround for demo

peterpeterparker:
# Motivation

Swap are open when actually the sales starts. It means that even if a proposal is executed the project remains pending for a while. For the sake of the demo this PR displays the pending projects as soon as the related proposal has been executed.

This workaround will be removed after the demo. 

# Changes

- if pending and time window is set then display projects


mstrasinskis (approved):
LGTM

PR #1145, opened 2022-07-20
style: apply different font opacity (label, description)

mstrasinskis:
# Motivation

Apply font colors to the launchpad blocks to match the design.

# Changes

- New classes for font colors:
  - `label`
  - `description`
  - `label-contrast-color`
  - `description-contrast-color`
- Updated launchpad cards and project description font colors
- Updated ICP component (label)

# Notes

- not possible to add label color directly to the `KeyValue` component because of 
  1. the description icon (to not make it transparent)
  2. doubling the opacity label + description (`InfoContextKey`)
  - thinking about refactoring the `InfoContextKey` component to not having its content inside the label

# Screenshots

Screenshot 2022-07-20 at 19 52 23

Screenshot 2022-07-20 at 19 52 31

Screenshot 2022-07-20 at 19 53 08

Screenshot 2022-07-20 at 19 52 56


frontend/svelte/src/lib/themes/fonts.scss
```diff
@@ -26,6 +26,20 @@ body {
   line-height: var(--line-height-standard);
 }
 
+.label {
```
peterpeterparker, line 29: - should we style with class or should we create two components `<Label />` and `<Description />`? because it seems to be applied to few elements

- if class, instead of `<span class="label" />`, should we use and style `<label />`?

mstrasinskis: Let's evaluate the final result and find the best solution based on it (classes vs Components)

frontend/svelte/src/lib/themes/fonts.scss
```diff
@@ -26,6 +26,20 @@ body {
...
+  opacity: var(--light-opacity);
+}
+.description {
+  opacity: var(--very-light-opacity);
+}
+// sometimes opacity is not an option (e.g. when there are icons inside)
+.label-contrast-color {
```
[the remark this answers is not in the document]
mstrasinskis: Good point! Done.

mstrasinskis:
@peterpeterparker How about to get rid of `--card-background-contrast`? Is there a plan to use not `--background-contrast` values for cards?

peterpeterparker:
@mstrasinskis not against, like renaming it to `--card-contrast` and `--card` or what's your idea?

mstrasinskis:
To be continued in another PR https://github.com/dfinity/nns-dapp/pull/1173/files

PR #1147, opened 2022-07-20
feat: debug tooltip ios

peterpeterparker:
# Motivation

Debug PR. Safari on iOS 😫


peterpeterparker:
Solved in https://github.com/dfinity/nns-dapp/pull/1148

PR #1148, opened 2022-07-20
style: fix tooltip not centered per default in Safari

peterpeterparker:
# Motivation

The tooltip is not centered per default on safari because it's parent is for some reason not an inherited block

# Changes

- set default css value for tooltip parent to block

# Screenshot

Capture d’écran 2022-07-20 à 20 26 08
Capture d’écran 2022-07-20 à 20 26 21


lmuntaner (approved):
Awesome!

PR #1149, opened 2022-07-21
refactor: *Nullable -> *Nullish

mstrasinskis:
# Motivation

To remove confusion, rename `nullable` to `nullish`.
Goal:
- `null` | `undefined` → `nullish` 
- `[]` → `nullable`

# Changes

Renaming only


peterpeterparker (approved):
Nice 👍

P.S.: Ultimately two functions we can moved in https://github.com/dfinity/ic-js/pull/156 too 🤪

PR #1150, opened 2022-07-21
build: redo sns and nns js nightly

peterpeterparker:
# Motivation

Redo sns and nns js nightly to easily get last nns and sns js changes


lmuntaner (approved):
thanks!

PR #1153, opened 2022-07-21
L2-856: Minimum amount to top up neuron

lmuntaner:
# Motivation

A user reported that topping up a neuron didn't work.

The problem was that to refresh a neuron, the neuron's balance needs to be greater than 1 ICP.

The user's neuron had less than 1 ICP, and with the transaction, it didn't go above 1 ICP either.

Yet, the transaction went through, but the neuron stake was not increased.

Do not allow performing such transactions.

# Changes

* add prop validateTransaction to NewTransaction component.
* add property validateTransaction to TransactionContext.
* Use validateTransaction function before performing any transaction in NewTransactionReview.
* Use new validateTransaction in IncreaseNeuronStakeModal.

# Tests

* Add the specific use case in the IncreaseNeuronStakeModal tests.


lmuntaner:

Since you were also involved in the issue reported by the user, it makes more sense if you review the proposed solution.

frontend/svelte/src/lib/types/transaction.context.ts
```diff
@@ -13,6 +13,7 @@ export interface TransactionContext {
   next: () => void;
   back: () => void;
   onTransactionComplete?: () => Promise<void>;
+  validateTransaction?: (store: TransactionStore) => boolean;
```
peterpeterparker, line 16: would it be possible to not pass the all store but "only" values?

Just thinking that by passing the store it also makes possible to subscribe to the store within the validation function. we don't do but if someone would ever do in the future it can lead to some weird effect.

frontend/svelte/src/lib/i18n/en.json
```diff
@@ -57,6 +57,7 @@
     "unknown": "Sorry, there was an unexpected error. Please try again later.",
     "amount_not_valid": "Sorry, the amount is not valid",
     "amount_not_enough_stake_neuron": "Sorry, the amount is not large enough. You need a minimum of 1 ICP to stake a neuron",
+    "amount_not_enough_top_up_neuron": "Sorry, the amount is not large enough. The amount plus the current stake need to be at least 1 ICP",
```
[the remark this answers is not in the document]
lmuntaner: I can rewrite to "is too small"

frontend/svelte/src/lib/types/transaction.context.ts
lmuntaner: I redid it to:
```
      validateTransaction({
        amount: $store.amount,
        selectedAccount: $store.selectedAccount,
        destinationAddress: $store.destinationAddress,
      })
```

I think this was what you meant, right?

frontend/svelte/src/lib/i18n/en.json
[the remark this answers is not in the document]
lmuntaner: done

lmuntaner:

frontend/svelte/src/lib/types/transaction.context.ts
peterpeterparker: Yes that does it, thanks

peterpeterparker (approved):
LGTM, thx

PR #1157, opened 2022-07-22
fix: make sure to render unique transitions to not break the UI

mstrasinskis:
# Motivation

Get rid of duplicated transactions to not break the UI (dev env. only) and improve the UX. E.g. could be the case when user send ICP from `main` to `main` account (by address). In that case, no transactions are displayed, and even the back button is broken.

## Before
Screenshot 2022-07-22 at 10 37 29
Screenshot 2022-07-22 at 10 47 26

## After
Screenshot 2022-07-25 at 11 13 59

# Changes

- ~~apply `uniqueObjects` to the fetched transactions~~
- display one of the duplicated transaction as `Received` to improve the UI

# Tests

- `Wallet` test updated


peterpeterparker:
Isn't it correct two display the two transactions?

1. One displayed as minus operation minus the transaction fee
2. The other as a plus movement

If this is correct, then I would say that the bug is maybe in the key used in cmp `TransactionList` for the svelte each (`transaction.timestamp.timestamp_nanos`)

peterpeterparker (changes requested):
Just blocking it because @lmuntaner has accepted it while I tend to think it should not be merged, see my question.

But I can be fully wrong, code is fine, more a feature related question.

mstrasinskis:
> Isn't it correct two display the two transactions?
> 
>     1. One displayed as minus operation minus the transaction fee ("Sent ICP -10.0001")
> 
>     2. The other as a plus movement ("Received ICP +10.0000")
> 
> 
> Or there are no transaction fee if same account?
> 
> If this is correct, then I would say that the bug is maybe in the key used in cmp `TransactionList` for the svelte each (`transaction.timestamp.timestamp_nanos`)

Updated according to your suggestion. From UX perspective, if find it better:

image

frontend/svelte/src/lib/utils/transactions.utils.ts
```diff
@@ -153,9 +156,12 @@ export const mapTransaction = ({
 
   const type = transactionType(transaction);
   const date = new Date(Number(timestamp.timestamp_nanos / BigInt(1e6)));
-  const isReceive = from !== account.identifier;
+  const isReceive = toSelfTransaction === true || from !== account.identifier;
   const isSend = to !== account.identifier;
-  const useFee = showTransactionFee({ type, isReceive });
+  const useFee =
```
peterpeterparker, line 161: there isn't a fee in case of same receiver and sender?

frontend/svelte/src/lib/utils/transactions.utils.ts
```diff
@@ -183,3 +189,24 @@ export const transactionName = ({
...
+  transactions: Transaction[]
+): { transaction: Transaction; toSelfTransaction: boolean }[] => {
+  const resultTransactions = transactions.map((transaction) => ({
+    transaction: { ...transaction },
+    toSelfTransaction: false,
+  }));
+
+  for (let i = 0; i < resultTransactions.length - 1; i++) {
+    const { transaction } = resultTransactions[i];
+    const { transaction: nextTransaction } = resultTransactions[i + 1];
```
peterpeterparker, line 204: isn't this a bit optimistic? isn't there a more accurate information / id to link both?

peterpeterparker: as discussed this is a bit optimitic but it's not like we have other choices since it's the data we get.
we will add some comments and after the pr is merge continue to inspect with other teams to check backend data

frontend/svelte/src/lib/utils/transactions.utils.ts
peterpeterparker: as discussed maybe a comment

peterpeterparker (approved):
As discussed, maybe a bit more comments and let's go with it, not like we have much choice with the data we have.

Thanks!

PR #1162, opened 2022-07-22
L2-797: Swap participation functionality

lmuntaner:
# Motivation

Use can participate in a swap

# Changes

* New sns service "participateISwap"
* New sns api function " participateInSnsSwap"
* ParticipateSwapModal uses new sns service to submit participation.

# Tests

* Finish test of ParticipateSwapModal
* Add test for new sns service
* Add test for new sns api function

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,7 +101,6 @@ proxy/build/
 proxy/node_modules
 canister_ids.json
 e2e-tests/node_modules
-frontend/node_modules
 /deployment-config.json
 release/
 
--- a/.gitignore
+++ b/.gitignore
@@ -101,6 +101,7 @@ proxy/build/
 proxy/node_modules
 canister_ids.json
 e2e-tests/node_modules
+frontend/node_modules
 /deployment-config.json
 release/
 
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -465,7 +465,7 @@ export const participateInSnsSwap = async ({
   rootCanisterId: Principal;
   fromSubAccount?: SubAccountArray;
 }): Promise<void> => {
-  logWithTimestamp(`Participating in swap ${rootCanisterId}: call...`);
+  logWithTimestamp("Participating in swap: call...");
 
   const { canister: nnsLedger } = await ledgerCanister({ identity });
   const snsWrapper = await wrapper({
@@ -489,5 +489,5 @@ export const participateInSnsSwap = async ({
   // Notify participation
   await snsWrapper.notifyParticipation({ buyer: controller.toText() });
 
-  logWithTimestamp(`Participating in swap ${rootCanisterId}: done`);
+  logWithTimestamp("Participating in swap: done");
 };
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -468,15 +468,15 @@ export const participateInSnsSwap = async ({
   logWithTimestamp(`Participating in swap ${rootCanisterId}: call...`);
 
   const { canister: nnsLedger } = await ledgerCanister({ identity });
-  const principalSubaccont = SubAccount.fromPrincipal(controller);
-  const accountIdentifier = AccountIdentifier.fromPrincipal({
-    principal: rootCanisterId,
-    subAccount: principalSubaccont,
-  });
   const snsWrapper = await wrapper({
     identity,
     rootCanisterId: rootCanisterId.toText(),
     certified: true,
+  });
+  const principalSubaccont = SubAccount.fromPrincipal(controller);
+  const accountIdentifier = AccountIdentifier.fromPrincipal({
+    principal: snsWrapper.canisterIds.swapCanisterId,
+    subAccount: principalSubaccont,
   });
 
   // Send amount to the ledger
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -1,8 +1,15 @@
 import type { HttpAgent, Identity } from "@dfinity/agent";
-import type { DeployedSns, SnsWasmCanister } from "@dfinity/nns";
+import {
+  AccountIdentifier,
+  ICP,
+  SnsWasmCanister,
+  SubAccount,
+  type DeployedSns,
+} from "@dfinity/nns";
 import { Principal } from "@dfinity/principal";
 import type { InitSnsWrapper, SnsWrapper } from "@dfinity/sns";
 import { mockSnsSummaryList } from "../../tests/mocks/sns-projects.mock";
+import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
 import { HOST, WASM_CANISTER_ID } from "../constants/environment.constants";
 import {
   importInitSnsWrapper,
@@ -19,6 +26,7 @@ import type {
 } from "../types/sns.query";
 import { createAgent } from "../utils/agent.utils";
 import { logWithTimestamp, shuffle } from "../utils/dev.utils";
+import { ledgerCanister } from "./ledger.api";
 
 let snsQueryWrappers: Promise<Map<QueryRootCanisterId, SnsWrapper>> | undefined;
 let snsUpdateWrappers:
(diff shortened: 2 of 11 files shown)


lmuntaner:
@peterpeterparker @mstrasinskis I'd say this is ready for review.

I get the following error:

```
Canister si2b5-pyaaa-aaaaa-aaaja-cai trapped explicitly: Panicked at 'The token amount can only be refreshed when the canister is in the 'open' state'
```

Therefore, it seems as if it's hitting the proper endpoint. But since we haven't yet tested it I wouldn't merge it.

lmuntaner:
@peterpeterparker @mstrasinskis it's ready for review, not yet to be merged though.

It has been tested and it works.

frontend/src/lib/api/sns.api.ts
```diff
@@ -443,3 +451,43 @@ export const querySnsSwapCommitment = async ({
...
+
+  // Send amount to the ledger
+  await nnsLedger.transfer({
+    amount,
+    fromSubAccount,
+    to: accountIdentifier,
+  });
+
+  // Notify participation
+  await snsWrapper.notifyParticipation({ buyer: controller.toText() });
```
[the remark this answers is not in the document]
lmuntaner: Nops, one needs the other. I believe that the swap canister checks the ledger when we notify it. I think that we would be able to call notify any time we want and it would do nothing unless there were ICPs in the account.

frontend/src/lib/api/sns.api.ts
```diff
@@ -443,3 +451,43 @@ export const querySnsSwapCommitment = async ({
...
+  rootCanisterId,
+  fromSubAccount,
+}: {
+  amount: ICP;
+  controller: Principal;
+  identity: Identity;
+  rootCanisterId: Principal;
+  fromSubAccount?: SubAccountArray;
+}): Promise<void> => {
+  logWithTimestamp(`Participating in swap ${rootCanisterId}: call...`);
```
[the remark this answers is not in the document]
lmuntaner: good point

.gitignore
```diff
@@ -101,6 +101,7 @@ proxy/build/
 proxy/node_modules
 canister_ids.json
 e2e-tests/node_modules
+frontend/node_modules
```
peterpeterparker, line 104: don't need this I think. it's already ignored in the `frontend/.gitignore` file.

frontend/src/lib/api/sns.api.ts
```diff
@@ -443,3 +451,43 @@ export const querySnsSwapCommitment = async ({
...
+}: {
+  amount: ICP;
+  controller: Principal;
+  identity: Identity;
+  rootCanisterId: Principal;
+  fromSubAccount?: SubAccountArray;
+}): Promise<void> => {
+  logWithTimestamp("Participating in swap: call...");
+
+  const { canister: nnsLedger } = await ledgerCanister({ identity });
```
peterpeterparker, line 470: from an architecture point of view, shouldn't the `sns.services`implement these two calls and not `sns.api`?

My understanding was that:

- `ledger.api` is a layer over NNS to communicate with the ledger
- `sns.api` is a layer over SNS to communicate with canisters of SNS

so if this is correct, `sns.api` should not call and use NNS canisters

.gitignore
lmuntaner: Weird, for some reason VS Code doesn't take into account the local `.gitignore`. But I'll remove it, it's a matter of VS Code, not of git not working.

frontend/src/lib/api/sns.api.ts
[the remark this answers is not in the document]

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
```diff
@@ -16,17 +26,48 @@
...
 
   let accepted: boolean = false;
   const toggelAccept = () => (accepted = !accepted);
 
-  const participate = () => {
-    // TODO: https://dfinity.atlassian.net/browse/L2-797
+  const dispatcher = createEventDispatcher();
+  const participate = async () => {
+    // TODO: Manage errors https://dfinity.atlassian.net/browse/L2-798
+    if ($store.summary === undefined || $store.summary === null) {
```
peterpeterparker, line 42: not for this PR but for the error, spontaneous idea: there is a `assertNonNullish` that checks for null and throw error, maybe we can create a `assertNonNullishToast` or something that does the same but call the toast on error

frontend/src/lib/api/sns.api.ts
lmuntaner: We do this in the api many times now. For example, when creating or topping up canisters, two calls are done also from the api function.

It's not a strong opinion, but if the api is such a thin layer, I wonder why we even bother using it. Is it just to create the agent and canister class?

One reason it was nice to have the two canister calls in the api was that they needed some polling a specific error management. Absttracting this in the api allowed the service to have a different kind of logic and manage the errors , validations and updating the stores after the actions. This is what we do mainly in the service:

* Validations
* Error management
* Store management

Maybe we haven't talked enough about service vs api. Is it worth a meeting?

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
```diff
@@ -16,17 +26,48 @@
...
+    startBusy({
+      initiator: "project-participate",
+    });
+    const { success } = await participateInSwap({
+      account,
+      amount: icpAmount,
+      rootCanisterId: $store.summary.rootCanisterId,
+    });
+    if (success) {
+      await loadSnsSwapCommitment({
```
peterpeterparker, line 54: e.g. in canisters we list the canisters and sync accounts on success within the service. shouldn't `loadSnsSwapCommitment` finds place in the service as well?

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
lmuntaner: the toast might be different depending on the use case, but I guess we could pass also the key as parameter as well.

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
lmuntaner: I wish, but in here the data lives in the context, not in the store. Therefore, I need to do it from a place where I have access to the context.

The other option would be to pass the function to "loadSwapCommitment" to the service "participateInSwap"

```
    const { success } = await participateInSwap({
      account,
      amount: icpAmount,
      rootCanisterId: $store.summary.rootCanisterId,
      onLoad: ({ response: swapCommitment }) =>
          ($store.swapCommitment = swapCommitment),
    });
```

This felt a little weird to me, but I could also do that.

frontend/src/lib/api/sns.api.ts
peterpeterparker: Just a bit "afraid" that interconnecting these api will lead to break chunking and lazy loading particularly if we ever apply fully such strategy. We should also pay attention to circular dependencies. But your answer makes sense and since it's already the case in canister then let's move forward. Thanks for the feedback.

.gitignore
peterpeterparker: maybe it's because in `frontend/.gitignore` it's written `/node_modules/`? if you replace with e.g. `node_modules` it's better?

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
peterpeterparker: yes. `assertNonNullish` as the error msg as optional key too. again not for this PR, just an idea

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
peterpeterparker: Yes maybe an `onSuccess` or something callback. Kind of feel we already have that pattern somewhere no?

.gitignore
lmuntaner: no, not better... weird

It is git ignored, so that is not the problem. The problem seems to be in the editor.

I'll play around more later, but it's something specific to VS Code.

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
lmuntaner: I will use nonNullish for now, then maybe we can use the assertNonNullish. Although, we might need it still returning a boolean also to set the proper type.

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
lmuntaner: sure

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
peterpeterparker: Like I said, my comment was not meant for this PR, just to share an idea 😉. I see there is the todo to handle error in another pr, all good

.gitignore
lmuntaner: done

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
lmuntaner: done

lmuntaner:

sns-js has been updated and now it's ready to be merged.

frontend/src/lib/services/sns.services.ts
```diff
@@ -230,3 +238,38 @@ export const routePathRootCanisterId = (path: string): string | undefined => {
...
+    const accountIdentity = await getAccountIdentity(account.identifier);
+    await participateInSnsSwap({
+      identity: accountIdentity,
+      rootCanisterId,
+      amount,
+      controller: accountIdentity.getPrincipal(),
+      fromSubAccount: "subAccount" in account ? account.subAccount : undefined,
+    });
+    await loadSnsSwapCommitment({
+      strategy: "update",
```
peterpeterparker, line 263: it has to be "update" only?

lmuntaner: I would say yes. There has been a change right before, I'd say we can't rely on the query.

peterpeterparker: Yes but can't we do "query + update"? Let's see when we can effectively manually test this multiple times...

peterpeterparker (approved):
LGTM, thx

frontend/src/lib/services/sns.services.ts
lmuntaner: yes, probably better.

I am thinking whether I should wait for the "load" to finish before I "return".

I need to be able to test the whole flow, and getting the real commitment, not a mock one.

I will come back to this.

PR #1163, opened 2022-07-22
Sns start command

bitdivine:
# Motivation
The sns start proposal has a few parameters that need to be set correctly.  This script should help.

# Changes
* Add a command for a start sns proposal.

# Tests
```
$ ./scripts/sns/swap/start_swap --dfx-network small11

Please create a neuron, e.g. in the nns-dapp front end.
Press enter when done... 

Please enter the ID of the neuron: 13156289036852124334

Warning: The version of DFX used (0.11.0-beta.1) is different than the version being run (0.11.0).
This might happen because your dfx.json specifies an older version, or DFX_VERSION is set in your environment.
We are forwarding the command line to the old version. To disable this warning, set the DFX_WARNING=-version_check environment variable.

Please add 37hh3-vuc33-3b46f-f6tqo-x7m3l-542wq-ifkdj-ahsfb-y245e-lfy3k-lqe as a hotkey to the neuron.
Press enter when done... 

(
  record {
    2_171_433_291 = opt variant {
      3_217_030_240 = record {
        2_744_746_248 = opt record { 23_515 = 8 : nat64 };
      }
    };
  },
)
```


deploy.sh
```diff
@@ -295,9 +295,16 @@ if [[ "$DEPLOY_SNS" == "true" ]]; then
         --wasm-file "$(CANISTER="sns_$canister" jq -r '.canisters[env.CANISTER].wasm' dfx.json)" "$canister"
     done
   fi
+  echo "Checking cycle balance"
+  while dfx wallet --network "$DFX_NETWORK" balance | awk '{exit $1 >= 50.00}'; do
+    WALLET_CANISTER="$(dfx identity --network small11 get-wallet)"
```
peterpeterparker, line 300: small06, small12, now small11....lots of env, do we need small11? can we simplify?

scripts/sns/swap/start_swap
```diff
@@ -0,0 +1,136 @@
...
+                };
+                command = opt variant {
+                        MakeProposal = record {
+                                url = "${URL:-http://...}";
+                                title = opt "${TITLE:-my title}";
+                                summary = "${SUMMARY:-summary}";
+                                action = opt variant{ SetSnsTokenSwapOpenTimeWindow = record {
+                                        swap_canister_id = opt principal "$SWAP_CANISTER_ID";
+                                        request = opt record { open_time_window = opt record {
+                                                start_timestamp_seconds=$(($(date +%s) + 60 * 5)): nat64;
```
peterpeterparker, line 128: not familiar with the format, the swap should start and end in how many minutes?

peterpeterparker (approved):
LGTM, thanks.

Two questions not blocking the PR.

deploy.sh
bitdivine: Good catch, small11 should be a var.  that's what happens when checking that the script works on one testnet only...

scripts/sns/swap/start_swap
bitdivine: In 60 * 5 seconds.  Ah, do you mean the `date +%s` ?  I can define that beforehand as TIME_NOW_SECONDS to make it clear what that means.

peterpeterparker: Ah in 5min, good. I wanted to know how much time we have between creating the proposal to accept it as the info are then transfered to the swap. 5 sounds good, thanks

PR #1165, opened 2022-07-25
build: fix usage of env var for wasm canister id

peterpeterparker:
# Motivation

Setting a wasm canister id through env var did not work anymore.

# Changes

- use wasm caniter id env var if provided else search in dfx to generate `deployment-config.sh`
- remove `console.log` in config
- move `config.ts` to related e2e  package


frontend/svelte/package.json
```diff
@@ -8,6 +8,7 @@
     "import-assets": "rm -fr public && cp -R static public",
     "build": "npm run import-assets && npm run i18n && npm run build:config && rollup -c && npm run build:index",
     "dev": "npm run import-assets && npm run i18n && CANISTER_ID=q3fc5-haaaa-aaaaa-aaahq-cai DFX_NETWORK=testnet npm run build:config && BASE_HREF=/ ROLLUP_WATCH=true npm run build:index && rollup -c -w",
+    "small06": "npm run import-assets && npm run i18n && CANISTER_ID=sbzkb-zqaaa-aaaaa-aaaiq-cai WASM_CANISTER_ID=qsgjb-riaaa-aaaaa-aaaga-cai DFX_NETWORK=small06 npm run build:config && BASE_HREF=/ ROLLUP_WATCH=true npm run build:index && rollup -c -w",
```
lmuntaner, line 11: the canister_id and the wasm_canister_id might change. Not sure that having a "small06" script helps much wit that.

I think we should use `canister_ids.json` for the configuration. It's more flexible and it's gitignored, therefore, changes there do not need to be pushed.

I changed the docs on how to develop locally: https://www.notion.so/Local-Development-Svelte-App-33f928e8f0234e0dbc552c4faa75cc95

lmuntaner (changes requested):
Thanks!

I like the changes, except adding one more script for a specific testnet. Not sure that's the way to go.

frontend/svelte/package.json
peterpeterparker: Good point. I removed this script from the PR.

lmuntaner (approved):
Thanks!

PR #1167, opened 2022-07-25
Filter sns proposals by open status

mstrasinskis:
# Motivation

Display only `state=open` SNS proposals on the launchpad page

# Changes

- update store logic
- rename store

# Tests

- updated according to the new logic


peterpeterparker (approved):
LTGM, thanks!

PR #1168, opened 2022-07-25
feat: display sns projects status pending with time window, open and complete + sort

peterpeterparker:
# Motivation

When a Sns proposal is accepted, the decentralized sale can effectively begin later on that's why we already want to display such projects in the "opportunities" section. Likewise, we want to display these projects for a while in the same section if they complete.

Finally, to ensure everything is sorted always in a same way, we want to sort the sns projects according swap start date.

# Changes

- rename `openProjectsStore` to `activeProjects` store and apply new filter "pending + time window or open or complete"
- move store filter to a new `projects.utils` file
- sort sns concatenated projects according swap start date

### DIFF
--- a/frontend/src/lib/components/launchpad/Projects.svelte
+++ b/frontend/src/lib/components/launchpad/Projects.svelte
@@ -5,7 +5,7 @@
   } from "../../services/sns.services";
   import { i18n } from "../../stores/i18n";
   import {
-    launchPadProjectsStore,
+    activePadProjectsStore,
     type SnsFullProject,
     snsesCountStore,
     snsSummariesStore,
@@ -21,7 +21,7 @@
   import { AppPath } from "../../constants/routes.constants";
 
   let projects: SnsFullProject[] | undefined;
-  $: projects = $launchPadProjectsStore;
+  $: projects = $activePadProjectsStore;
 
   let projectCount: number | undefined;
   $: projectCount = $snsesCountStore;
--- a/frontend/src/lib/components/launchpad/Projects.svelte
+++ b/frontend/src/lib/components/launchpad/Projects.svelte
@@ -5,7 +5,7 @@
   } from "../../services/sns.services";
   import { i18n } from "../../stores/i18n";
   import {
-    openProjectsStore,
+    launchPadProjectsStore,
     type SnsFullProject,
     snsesCountStore,
     snsSummariesStore,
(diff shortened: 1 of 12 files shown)


frontend/src/lib/components/launchpad/Projects.svelte
```diff
@@ -5,7 +5,7 @@
   } from "../../services/sns.services";
   import { i18n } from "../../stores/i18n";
   import {
-    openProjectsStore,
+    activePadProjectsStore,
```
mstrasinskis, line 8: Out of curiosity, why not `launchpadProjectsStore`?

mstrasinskis (approved):
LGTM, thank you!

frontend/src/lib/components/launchpad/Projects.svelte
peterpeterparker: That was the first name but then I thought it make it really tight to screen name particularly if we ever used it elsewhere, that's why ultimately I thought about a more generic name. Make sense or you rather like launchPad...?

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -38,15 +63,17 @@ export const concatSnsSummaries = ([summaries, swaps]: [
...
-      init: fromNullable(swap.init) as SnsSwapInit,
-      state: fromNullable(swap.state) as SnsSwapState,
-    },
-  }));
+  return sortSnsSummaries(
+    validSwapSummaries.map(({ swap, ...rest }) => ({
+      ...rest,
+      swap: {
+        // We know for sure that init and state are defined because we check in previous filter that there are not undefined
+        // TODO: There might be a cleaner way than a type cast to make TypeScript checks these are defined
```
mstrasinskis, line 71: How about having separate function for this? 
e.g. `export const fromDefinedNullable = <T>(value: [T]): T =>  value[0];`

frontend/src/lib/components/launchpad/Projects.svelte
mstrasinskis: I don't have any strong opinion. So let's leave it ;)

mstrasinskis: Thank you for the explanation.

frontend/src/lib/utils/sns.utils.ts
peterpeterparker: Like it. Can you add it to the new utils lib?

PR #1169, opened 2022-07-25
feat: disable auth idle manager

peterpeterparker:
# Motivation

As NNS-dapp observes the local storage and sync the delegation status if anything changes it leads to some UX issues / quirks with the new idle manager of agent-js. Until we got time to implement it properly we decided to disable it.

Session duration of 30min, sync of the auth on storage change cronjob that verifies the delegation validity and warning about automatic logout are not affected by this PR - remains the same as they are since months. 

# Changes

- disable idle manager of auth client
- extract session duration to a constant


lmuntaner (approved):
Thanks!

PR #1170, opened 2022-07-26
Deploy wasm canister only

bitdivine:
# Motivation
For demos, and also for providing stable sns-wasm canister IDs, it is desirable to deploy the sns-wasm canister by itself without creating an SNS.

# Changes
* Separate wasm canister deployment out into a separate step, with a separate flag.
  * Note: If an SNS is requested and there is no wasm canister, one will be deployed even without the new explicit `--sns-wasm`, so existing scripts will not break on account of the new flag.

# Tests
* Used to deploy small12 - with a small hiccup with the II canister IDs.  Added a fix (deleting unnecessary configuration) to this PR.
* Deployed to small06 repeatedly until I observed no usability or functionality issues.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -350,7 +350,7 @@ if [[ "$POPULATE" == "true" ]]; then
   }
 fi
 
-if [[ "$DEPLOY_SNS_WASM_CANISTER" == "true" ]]; then
+if [[ "$DEPLOY_SNS" == "true" ]]; then
   echo "Checking cycle balance"
   while dfx wallet --network "$DFX_NETWORK" balance | awk '{exit $1 >= 51.00}'; do
     WALLET_CANISTER="$(dfx identity --network "$DFX_NETWORK" get-wallet)"
--- a/deploy.sh
+++ b/deploy.sh
@@ -202,15 +202,20 @@ if [[ "$GUESS" == "true" ]]; then
   esac
 fi
 
-echo
-echo DELETE_CANISTER_IDS=$DELETE_CANISTER_IDS
-echo DELETE_WALLET=$DELETE_WALLET
-echo START_DFX=$START_DFX
-echo DEPLOY_NNS_BACKEND=$DEPLOY_NNS_BACKEND
-echo DEPLOY_II=$DEPLOY_II
-echo DEPLOY_NNS_DAPP=$DEPLOY_NNS_DAPP
-echo POPULATE=$POPULATE
-echo OPEN_NNS_DAPP=$OPEN_NNS_DAPP
+if [[ "$DRY_RUN" == "true" ]] || [[ "$GUESS" == "true" ]]; then
+  echo
+  echo DELETE_CANISTER_IDS=$DELETE_CANISTER_IDS
+  echo DELETE_WALLET=$DELETE_WALLET
+  echo START_DFX=$START_DFX
+  echo DEPLOY_NNS_BACKEND=$DEPLOY_NNS_BACKEND
+  echo DEPLOY_SNS_WASM_CANISTER="${DEPLOY_SNS_WASM_CANISTER:-}"
+  echo DEPLOY_II=$DEPLOY_II
+  echo DEPLOY_NNS_DAPP=$DEPLOY_NNS_DAPP
+  echo POPULATE=$POPULATE
+  echo DEPLOY_SNS="$DEPLOY_SNS"
+  echo OPEN_NNS_DAPP=$OPEN_NNS_DAPP
+fi
+
 [[ "$DRY_RUN" != "true" ]] || exit 0
 [[ "$GUESS" != "true" ]] || {
   echo
@@ -259,7 +264,8 @@ if [[ "$START_DFX" == "true" ]]; then
   read -rp "Please press enter when done... "
 fi
 
-if [[ "$DEPLOY_NNS_BACKEND" == "true" ]] || [[ "$DEPLOY_SNS" == "true" ]]; then
+: "Get WASM and did files.  By now it's needed for almost everything except open..."
+if [[ "$DEPLOY_NNS_BACKEND" == "true" ]] || [[ "$DEPLOY_SNS" == "true" ]] || test -n "${DEPLOY_SNS_WASM_CANISTER:-}"; then
   if [[ "$CTL_NOBUILD_NNS" == "true" ]]; then
     echo "Using exising NNS and SNS canisters"
   else
@@ -270,12 +276,6 @@ fi
 
 if [[ "$DEPLOY_NNS_BACKEND" == "true" ]]; then
   ./e2e-tests/scripts/nns-canister-install
-fi
-
-if [[ "$DEPLOY_II" == "true" ]]; then
-  dfx deploy --network "$DFX_NETWORK" internet_identity --no-wallet
-  echo "Waiting for II to be stable..."
-  sleep 4
 fi
 
 # Note: On mainnet SNS are created much later and have unpredictable canister IDs, however
@@ -307,6 +307,12 @@ if test -n "${DEPLOY_SNS_WASM_CANISTER:-}"; then
         --wasm-file "$(CANISTER="sns_$canister" jq -r '.canisters[env.CANISTER].wasm' dfx.json)" "$canister"
     done
   fi
+fi
+
+if [[ "$DEPLOY_II" == "true" ]]; then
+  dfx deploy --network "$DFX_NETWORK" internet_identity --no-wallet
+  echo "Waiting for II to be stable..."
+  sleep 4
 fi
 
 if [[ "$DEPLOY_NNS_DAPP" == "true" ]]; then
--- a/deploy.sh
+++ b/deploy.sh
@@ -66,6 +66,9 @@ help_text() {
 	--ii
 	  Create the internet_identity canister.
 
+	--sns-wasm
+	  Create or update an SNS wasm canister.
+
 	--sns
 	  Create an SNS canister set.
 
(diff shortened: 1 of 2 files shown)


dfx.json
```diff
@@ -31,31 +31,41 @@
       "build": "curl --fail -sSL \"https://github.com/dfinity/internet-identity/releases/download/release-2022-06-17/internet_identity_dev.wasm\" -o internet_identity.wasm"
     },
     "wasm_canister": {
-      "build": ["true"],
+      "build": [
+        "true"
+      ],
```
bitdivine, line 36: This file is processed with jq a lot; jq puts in these line breaks so inflates the diff.  Better to have the formatting compatible with jq by default.

deploy.sh
```diff
@@ -66,6 +66,9 @@ help_text() {
 	--ii
 	  Create the internet_identity canister.
 
+	--sns-wasm
+	  Create or update an SNS wasm canister.
+
```
bitdivine, line 71: Adding this flag and making it work is the goal of this PR.

deploy.sh
```diff
@@ -97,6 +100,7 @@ START_DFX="false"
 DEPLOY_NNS_BACKEND="false"
 DEPLOY_II="false"
 DEPLOY_SNS="false"
+DEPLOY_SNS_WASM_CANISTER=""
```
bitdivine, line 103: In this case, `""` is more useful than `"false"` as we can check false with `test -n` and there are two different true values.  See below for details!

deploy.sh
```diff
@@ -131,9 +135,14 @@ while (($# > 0)); do
...
     DEPLOY_II="true"
     ;;
+  --sns-wasm)
+    GUESS="false"
+    DEPLOY_SNS_WASM_CANISTER="true"
+    ;;
   --sns)
     GUESS="false"
     DEPLOY_SNS="true"
+    DEPLOY_SNS_WASM_CANISTER="${DEPLOY_SNS_WASM_CANISTER:-ifnotinstalled}"
```
bitdivine, line 145: If `--sns` and `--sns-wasm` are both given, the stronger `true` is used rather than `ifnotinstalled`.

deploy.sh
```diff
@@ -193,15 +202,20 @@ if [[ "$GUESS" == "true" ]]; then
   esac
 fi
```
bitdivine, line 204: Less noise:  For most runs this printout is uninteresting.

deploy.sh
```diff
@@ -263,25 +278,22 @@ if [[ "$DEPLOY_NNS_BACKEND" == "true" ]]; then
   ./e2e-tests/scripts/nns-canister-install
 fi
 
-if [[ "$DEPLOY_II" == "true" ]]; then
-  dfx deploy --network "$DFX_NETWORK" internet_identity --no-wallet
-  echo "Waiting for II to be stable..."
-  sleep 4
-fi
```
bitdivine, line 270: Moved II down to below sns-wasm, so that II and NNS-DAPP are together.  NNS_DAPP needs to come after the sns-wasm canister because it needs its canister ID.

deploy.sh
```diff
@@ -263,25 +278,22 @@ if [[ "$DEPLOY_NNS_BACKEND" == "true" ]]; then
...
-
 # Note: On mainnet SNS are created much later and have unpredictable canister IDs, however
 # until an index canister exists we need the SNS to exist at a predictable address, so we install it now.
 # Note: There may be multiple SNS canister sets; at present this can be done in a somewhat clunky way by
 # adding numbers to SNS canister names, however in fiture versions of dfx, it will be possible to have
 # several dfx.json, so we can have one dfx.json per SNS and one for the nns-dapp project, without weird names.
```
bitdivine, line 285: Here we deploy just the sns-wasm canister, not an SNS, and we support `ifnotinstalled`.

deploy.sh
```diff
@@ -363,6 +350,40 @@ if [[ "$POPULATE" == "true" ]]; then
...
 
+if [[ "$DEPLOY_SNS" == "true" ]]; then
+  echo "Checking cycle balance"
+  while dfx wallet --network "$DFX_NETWORK" balance | awk '{exit $1 >= 51.00}'; do
+    WALLET_CANISTER="$(dfx identity --network "$DFX_NETWORK" get-wallet)"
+    echo "Please add 51T cycles to this canister: $WALLET_CANISTER"
+    read -rp "Press enter when done ..."
+    echo
+  done
+
```
bitdivine, line 361: Deploying the SNS comes at the end, as the user may need to add cycles and this way they can do so via the already deployed nns-dapp.  Usually the script is run without an existing wallet and a new wallet on a testnet is pre-populated with enough cycles so interaction SHOULD not be needed with current values.  However if e.g. the cost of starting an SNS is increased this assumption could break without notice.

dfx.json
```diff
@@ -84,11 +96,11 @@
         "REDIRECT_TO_LEGACY": "prod",
         "ENABLE_NEW_SPAWN_FEATURE": true,
         "ENABLE_SNS_NEURONS": false,
-        "HOST": "https://nnsdapp.dfinity.network",
-        "OWN_URL": "https://OWN_CANISTER_ID.nnsdapp.dfinity.network",
-        "IDENTITY_SERVICE_URL": "https://qjdve-lqaaa-aaaaa-aaaeq-cai.nnsdapp.dfinity.network"
+        "HOST": "https://nnsdapp.dfinity.network"
```
bitdivine, line 99: OWN_URL and IDENTITY_CANISTER_URL already have defaults.  Hard-wiring canister IDs here just breaks deployments when the II canister ID changes.

dfx.json
```diff
@@ -171,7 +186,7 @@
       "config": {
         "RUSTC_VERSION": "1.58.1",
         "IC_CDK_OPTIMIZER_VERSION": "0.3.1",
-        "IC_COMMIT": "5b66c3f6994a43777b1161ef7f587698e1265773"
+        "IC_COMMIT": "268db05d70b3cc20475ba3ac046c0db9bd66c688"
```
bitdivine, line 189: Bump IC commit to the one used today in testnet deployments, as those served to validate this PR.

deploy.sh
lmuntaner: General question, a wallet is also a canister?

dfx.json
lmuntaner: ah, ok, I think that was me. I will take that into account.

lmuntaner (commented):
A few comments before approving if you don't mind.

deploy.sh
bitdivine: It is a string, just like true or false in the lines above.  Bash doesn't have cool things like enums.

deploy.sh
bitdivine: Yes, a cycle wallet it is just a canister with a certain API.  Dfinity provides an implementation but people are free to develop other wallets.  Any canister can hold cycles and transfer those cycles to other canisters.  A wallet is just a canister with a standard API for doing so.

bitdivine: And, yes, it should probably always be called a cycles wallet because the stock dfinity wallet canister doesn't hold ICP or other tokens.  It could, hypothetically.

deploy.sh
lmuntaner: ah, ok!

deploy.sh
```diff
@@ -263,25 +278,22 @@ if [[ "$DEPLOY_NNS_BACKEND" == "true" ]]; then
...
-if [[ "$DEPLOY_SNS" == "true" ]]; then
+if test -n "${DEPLOY_SNS_WASM_CANISTER:-}"; then
   # If the wasm canister has not been installed already, install it.
   echo Checking whether sns wasm is installed
   SNS_WASM_CANISTER_ID="$(dfx canister --network "$DFX_NETWORK" id wasm_canister 2>/dev/null || echo NOPE)"
-  if [[ "${SNS_WASM_CANISTER_ID:-}" != "NOPE" ]]; then
+  [[ "${SNS_WASM_CANISTER_ID:-}" == "NOPE" ]] || {
     echo "SNS wasm/management canister already installed at: $SNS_WASM_CANISTER_ID"
+  }
+  if [[ "${SNS_WASM_CANISTER_ID:-}" != "NOPE" ]] && [[ "$DEPLOY_SNS_WASM_CANISTER" == "ifnotinstalled" ]]; then
```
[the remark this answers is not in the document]
lmuntaner: When would the value be "ifnotinstalled"? Maybe it's this value that is confusing me.

How I read it is that `If not installed --> echo "Using existing WASM canister..."`. Which I don't understand.

deploy.sh
lmuntaner: ok, thanks!

deploy.sh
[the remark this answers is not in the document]
lmuntaner: I am approving it, but this part is still hard to read for me.

lmuntaner (approved):
Already approving, but there is a part that is not yet very clear to me:

> When would the value be "ifnotinstalled"? Maybe it's this value that is confusing me.
> 
> How I read it is that If not installed --> echo "Using existing WASM canister...". Which I don't understand.

deploy.sh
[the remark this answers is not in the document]
bitdivine: Thanks.  I'll think about a better way of explaining it and when I do I'll put it in a comment or rename the value.  There is probably a set of names or a flow or a combination of these that makes it obvious.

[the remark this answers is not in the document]
lmuntaner: thanks!

PR #1172, opened 2022-07-26
feat: init sns after sign-in on app scale level

peterpeterparker:
# Motivation

As we are going to need Snses in various screens - e.g. Neurons or Voting in the future - we can or should load Snses on for the all app after sign-in.

# Changes

- create new `app.services` and move current initialization features from cmp `App`
- move Sns initialization from `Projects` and `Neurons` to new service
- load Sns only on ENABLE_SNS_NEURONS

# Note

I wanted to use TESTNET as feature flag but it breaks test in the CI only therefore use the other existing flag ENABLE_SNS_NEURONS. I added a TODO that we shall rename it someday and make the features flag an object.


mstrasinskis (approved):
LGTM 👍

PR #1176, opened 2022-07-27
L2-836 and L2-891: Add title swap and account

lmuntaner:
# Motivation

User can see the account where ICP is transferred when participating in a swap.

User participates in a Decentralized Sale, not in a swap.

# Changes

* Change Swap reference to Decentralized Sale in user facing communication.
* New sns api function querySwapCanisterAccount.
* Use new sns api function when participating.
* New sns service to get the account which uses new sns api querySwapCanisterAccount.
* Use new sns service in the ReviewParticipate component.

# Tests

* Fix broken test.
* New test for new sns api function.
* New test for new sns service function

### DIFF
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -196,23 +196,16 @@ const wrapper = async ({
 // TODO(L2-751): remove mock data
 let mockSnsSummaries: QuerySnsSummary[] = [];
 
-export const querySwapCanisterAccount = async ({
-  rootCanisterId,
-  identity,
+export const querySwapCanisterAccount = ({
   controller,
+  swapCanisterId,
 }: {
-  rootCanisterId: Principal;
-  identity: Identity;
   controller: Principal;
-}): Promise<AccountIdentifier> => {
-  const snsWrapper: SnsWrapper = await wrapper({
-    identity,
-    rootCanisterId: rootCanisterId.toText(),
-    certified: true,
-  });
+  swapCanisterId: Principal;
+}): AccountIdentifier => {
   const principalSubaccont = SubAccount.fromPrincipal(controller);
   const accountIdentifier = AccountIdentifier.fromPrincipal({
-    principal: snsWrapper.canisterIds.swapCanisterId,
+    principal: swapCanisterId,
     subAccount: principalSubaccont,
   });
 
@@ -355,7 +348,10 @@ export const querySnsSwapState = async ({
     `Getting Sns ${rootCanisterId} swap state certified:${certified} call...`
   );
 
-  const { swapState }: SnsWrapper = await wrapper({
+  const {
+    swapState,
+    canisterIds: { swapCanisterId },
+  }: SnsWrapper = await wrapper({
     rootCanisterId,
     identity,
     certified,
@@ -372,6 +368,7 @@ export const querySnsSwapState = async ({
 
   return {
     rootCanisterId,
+    swapCanisterId,
     swap,
   };
 };
@@ -466,14 +463,16 @@ export const participateInSnsSwap = async ({
   logWithTimestamp("Participating in swap: call...");
 
   const { canister: nnsLedger } = await ledgerCanister({ identity });
-  const snsWrapper = await wrapper({
+  const {
+    canisterIds: { swapCanisterId },
+    notifyParticipation,
+  } = await wrapper({
     identity,
     rootCanisterId: rootCanisterId.toText(),
     certified: true,
   });
-  const accountIdentifier = await querySwapCanisterAccount({
-    rootCanisterId,
-    identity,
+  const accountIdentifier = querySwapCanisterAccount({
+    swapCanisterId,
     controller,
   });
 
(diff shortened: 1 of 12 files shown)


lmuntaner:
UI changes:

Screenshot 2022-07-27 at 07 55 38

lmuntaner:

frontend/src/lib/api/sns.api.ts
```diff
@@ -217,6 +217,29 @@ const wrapper = async ({
 // TODO(L2-751): remove mock data
 let mockSnsSummaries: QuerySnsSummary[] = [];
 
+export const querySwapCanisterAccount = async ({
```
peterpeterparker, line 220: so if I get it right, you mostly need this function to get the `swapCanisterId`. Have you thought about adding the information to `SnsSummary` - i.e. to the store? what you think about it?

lmuntaner: It does.

Mhm, it makes sense actually. We would avoid a call to the root canister this way.

I'll change it.

lmuntaner:

There are quite some new changes. Most of them related to the new property in SnsSummary.

frontend/src/lib/api/sns.api.ts
```diff
@@ -196,6 +196,22 @@ const wrapper = async ({
 // TODO(L2-751): remove mock data
 let mockSnsSummaries: QuerySnsSummary[] = [];
 
+export const querySwapCanisterAccount = ({
```
peterpeterparker, line 199: this can now be a `utils` I think or local function of the service

also it can probably be renamed

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
```diff
@@ -31,6 +35,13 @@
   let icpAmount: ICP;
   $: icpAmount = convertNumberToICP(amount);
 
+  let destinationAddress: AccountIdentifier | undefined;
+  $: (async () => {
+    if ($store.summary?.swapCanisterId !== undefined) {
+      destinationAddress = await getSwapAccount($store.summary?.swapCanisterId);
+    }
```
peterpeterparker, line 42: should there be a `else destinationAddress = undefined`?

just in case query+update calls issue or if the user change the id in the url

the modal probably close itself but just in case?

frontend/src/lib/services/sns.services.ts
```diff
@@ -239,6 +246,20 @@ export const routePathRootCanisterId = (path: string): string | undefined => {
...
+  swapCanisterId: Principal
+): Promise<AccountIdentifier | undefined> => {
+  try {
+    const identity = await getIdentity();
+    return querySwapCanisterAccount({
+      controller: identity.getPrincipal(),
+      swapCanisterId,
+    });
+  } catch (error) {
+    return undefined;
```
peterpeterparker, line 259: since `getIdentity` does the logout can we even simplifies and remove the try/catch and the undefined return?

```
export const getSwapAccount = async (
  swapCanisterId: Principal
): Promise<AccountIdentifier> => {
    const identity = await getIdentity();
    return querySwapCanisterAccount({
      controller: identity.getPrincipal(),
      swapCanisterId,
    });
```

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -98,6 +103,7 @@ export const concatSnsSummary = ([summary, swap]: [
 
   return {
     ...summary,
+    swapCanisterId: swap.swapCanisterId,
```
peterpeterparker, line 106: you need this line? not sure, isn't `swapCanisterId` already deconstructed in `...summary`?

frontend/src/lib/api/sns.api.ts
lmuntaner: I didn't like the idea of using util outside of api, but it's true that we already do it, so makes sense also here.

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
lmuntaner: yeah, makes sense

frontend/src/lib/utils/sns.utils.ts
lmuntaner: mhm, true, it's already added before

frontend/src/lib/services/sns.services.ts
lmuntaner: makes sense

frontend/src/lib/utils/sns.utils.ts
lmuntaner: oh, no. It's on the swap part, not the summary part.

frontend/src/lib/services/sns.services.ts
lmuntaner: done

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
lmuntaner: done

frontend/src/lib/api/sns.api.ts

frontend/src/lib/utils/sns.utils.ts
peterpeterparker: oh oki doki, thanks for checking it

peterpeterparker (approved):
LGTM, thx!

PR #1177, opened 2022-07-27
feat: handles sns swap state to display information

peterpeterparker:
# Motivation

Sns projects can be displayed in the app if there are pending, open or complete. Therefore we should align the display of the information according such state. Likewise, user can only participate ("click on button participate") if sale is open.

# Notes

All these changes have been defined and accepte by Mischa.

# Changes

- extracted component `<DateSeconds />` to display formatted dates from seconds
- display start and deadline dates according sale state
- hide participate to swap button if not open
- hide sale status if not open or comitted
- display "Start" and "End" info of sale in sale details
- remove info placeholder tests
- remove loading spinner has status can be displayed or actually not

# Screenshots of what's expected in project details

Capture d’écran 2022-07-27 à 10 22 50
Capture d’écran 2022-07-27 à 10 26 11
Capture d’écran 2022-07-27 à 10 27 16
Capture d’écran 2022-07-27 à 10 30 31

### DIFF
--- a/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
@@ -28,9 +28,9 @@
   let max_icp_e8s: bigint;
   $: ({ min_icp_e8s, max_icp_e8s } = init);
 
-  // TODO: display current commitment (info soon available in a new candid field)
   let currentCommitment: bigint;
-  $: currentCommitment = min_icp_e8s;
+  $: currentCommitment =
+    $projectDetailStore.swapCommitment?.currentCommitment ?? BigInt(0);
 
   let currentCommitmentIcp: ICP;
   $: currentCommitmentIcp = ICP.fromE8s(currentCommitment);
--- a/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
@@ -19,7 +19,7 @@
   );
 
   let swap: SnsSummarySwap;
-  // type safety validation is done in ProjectDetail component
+  // type safety validation is done in ProjectStatusSection component
   $: ({ swap } = $projectDetailStore.summary as SnsSummary);
 
   let init: SnsSwapInit;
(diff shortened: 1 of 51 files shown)


frontend/src/lib/components/project-detail/ProjectCommitment.svelte
```diff
@@ -38,12 +37,9 @@
 </script>
 
 <KeyValuePair testId="sns-project-current-commitment">
-  <InfoContextKey slot="key">
-    <svelte:fragment slot="header"
-      >{$i18n.sns_project_detail.current_commitment}</svelte:fragment
-    >
-    <p>Some details about what the current commitment means.</p>
-  </InfoContextKey>
+  <span slot="key">
```
lmuntaner, line 40: Did you check whether is `InfoContextKey` still used?

frontend/src/lib/components/project-detail/ProjectStatusSection.svelte
```diff
@@ -31,13 +32,33 @@
...
 </script>
 
-{#if swapCommitment === undefined}
-  <div class="wrapper">
-    <Spinner inline />
-  </div>
-{:else}
+{#if loadingSummary || loadingSwapState}
+  <!-- Because information might not be displayed once loaded - according the state - we do no display a spinner or skeleton -->
+{:else if displayStatus}
```
lmuntaner, line 61: Why not just one if statement?

```
{#if displayStatus}
//...
{/if}
```

lmuntaner (approved):
Minor comment, looks good to me already!

frontend/src/lib/components/project-detail/ProjectCommitment.svelte
peterpeterparker: It might not but we might/will probably use the cmp in the future, therefore did removed it (at least for now).

frontend/src/lib/components/project-detail/ProjectStatusSection.svelte
peterpeterparker: I guess I wanted to highlight the fact that we can if we want implement a spinner or skeleton but good point, doesn't make that sense. Removing it (without removing the comment).

frontend/src/lib/components/project-detail/ProjectCommitment.svelte
lmuntaner: ok, good to know

PR #1178, opened 2022-07-27
Delete the network for all IDs

bitdivine:
# Motivation
After redeploying a testnet, all wallet canisters are gone.  This is correct and expected, however if there are wallets in the user's config, they will fail to work with a confusing message.  There is  already a command that cleans local user state, for use after deploying a testnet, however it only deleted the wallet for one identity.  Now that we use several identities, that is no longer enough.

# Changes
* `--delete` now deletes the wallets for a network from all identities.

# Tests
I have run this and examined the wallet files.


peterpeterparker (approved):
Rubber stamp

PR #1179, opened 2022-07-27
refactor: load sns summaries and swaps with derived store

peterpeterparker:
# Motivation

As discussed with @lmuntaner this PR refactor the way we load the sns summaries and swaps in store.

Instead of getting the information summaries and swaps from the backend, cleaning and concatening these (match swap and summary with root id + cleaning optional candid types + sort) and applying to the stores, this pr get the information from backend, apply these to a global store and make the information available to the app through a derived store.

Think a bit the "redux selector way" rather than "preparing data and putting these in store afterwards".

# Note

I did modify the global sns/projects store but I did not modify the context store used in the detail - i.e. the context store will still get data that are prepared in advance. Open to discussion but since the context store for the detail page contains both summary, swap and commitment, it would have need to split it in two or three stores. Feels like an over complication that does not bring that much clarity in the architecture.

# Changes

- introduces a new `sns.store` that contains sns answers from backend - i.e. sns summary, swaps but also proposals
- refactor `project.store` to derive above store
- move stores from `project` to `sns` if more "sns" or abstraction "projects" related

# Side notes

This PR will probably conflicts with #1176 and #1172 and will only be merged afterwards.

### DIFF
--- a/frontend/src/lib/stores/sns.store.ts
+++ b/frontend/src/lib/stores/sns.store.ts
@@ -1,0 +1,156 @@
+import { ProposalStatus, type ProposalInfo } from "@dfinity/nns";
+import type { Principal } from "@dfinity/principal";
+import { derived, writable } from "svelte/store";
+import type { SnsSummary, SnsSwapCommitment } from "../types/sns";
+import type { QuerySnsSummary, QuerySnsSwapState } from "../types/sns.query";
+import { mapAndSortSnsQueryToSummaries } from "../utils/sns.utils";
+import { isNullish } from "../utils/utils";
+
+// ************** Proposals for Launchpad **************
+
+export type SnsProposalsStore =
+  | {
+      proposals: ProposalInfo[];
+      certified: boolean;
+    }
+  | undefined
+  | null;
+
+const initSnsProposalsStore = () => {
+  const { subscribe, set } = writable<SnsProposalsStore>(undefined);
+
+  return {
+    subscribe,
+
+    reset() {
+      set(undefined);
+    },
+
+    setLoadingState() {
+      set(null);
+    },
+
+    setProposals({
+      proposals,
+      certified,
+    }: {
+      proposals: ProposalInfo[];
+      certified: boolean;
+    }) {
+      set({
+        proposals,
+        certified,
+      });
+    },
+  };
+};
+
+const initOpenSnsProposalsStore = () =>
+  derived([snsProposalsStore], ([$snsProposalsStore]): ProposalInfo[] =>
+    isNullish($snsProposalsStore)
+      ? []
+      : $snsProposalsStore.proposals.filter(
+          ({ status }) => status === ProposalStatus.PROPOSAL_STATUS_OPEN
+        )
+  );
+
+export const snsProposalsStore = initSnsProposalsStore();
+export const openSnsProposalsStore = initOpenSnsProposalsStore();
+
+// ************** Sns summaries and swaps **************
+
+export type SnsQueryStore =
+  | {
+      response: [QuerySnsSummary[], QuerySnsSwapState[]];
+      certified: boolean;
+    }
+  | undefined
+  | null;
+
+const initSnsQueryStore = () => {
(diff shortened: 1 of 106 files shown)


frontend/src/lib/stores/sns.store.ts
```diff
@@ -0,0 +1,155 @@
...
+ */
+export const snsSummariesStore = derived(snsQueryStore, (data: SnsQueryStore) =>
+  mapAndSortSnsQueryToSummaries(data?.response ?? [[], []])
+);
+
+// ************** Sns commitment **************
+
+export type SnsSwapCommitmentsStore =
+  | {
+      swapCommitment: SnsSwapCommitment;
```
lmuntaner, line 108: That means also that we could store here the BuyerState directly along with the rootCanisterId.

And in the end remove the currentCommitment from this type, right?

The idea would be to store here what comes directly from the backend, which is not the SnsSwapCommitment type.

frontend/src/lib/stores/sns.store.ts
```diff
@@ -0,0 +1,155 @@
...
+  );
+
+export const snsProposalsStore = initSnsProposalsStore();
+export const openSnsProposalsStore = initOpenSnsProposalsStore();
+
+// ************** Sns summaries and swaps **************
+
+export type SnsQueryStore =
+  | {
+      response: [QuerySnsSummary[], QuerySnsSwapState[]];
```
lmuntaner, line 63: since they are each from a different api, would it make sense to have store for each type?

frontend/src/lib/stores/sns.store.ts
```diff
@@ -0,0 +1,155 @@
+import { ProposalStatus, type ProposalInfo } from "@dfinity/nns";
+import { derived, writable } from "svelte/store";
+import type { SnsSwapCommitment } from "../types/sns";
```
lmuntaner, line 3: ideally it would make more sense to import here only types from sns-js, right?

Not in this PR, now that we have the derived stores that will be easy.

lmuntaner (commented):
Thanks, that's very nice!

Just a couple of questions.

And thanks for waiting for the ongoing PRs.

frontend/src/lib/stores/sns.store.ts
peterpeterparker: not really because both information are fetched with a `Promise.all` - i.e. for better perf but also because it guards the update+query calls chain (difficult to explain better in a sentence).

from api

```
const summaries: (QuerySnsSummary | undefined)[] = await Promise.all(
    snsWrappers.map(({ canisterIds: { rootCanisterId } }: SnsWrapper) =>
```

therefore works well with a single store

frontend/src/lib/stores/sns.store.ts
peterpeterparker: yes ideally but not sure how you would do that since for the commitment you need to add the extra root canister id otherwise you cannot make the matching but maybe I'm missing something. like you said we can do it afterwards in another PR 👍

frontend/src/lib/stores/sns.store.ts
peterpeterparker: Not sure because it's not just one buyer, it's an array of commitment and for each we need both root canister id and the certified or not information. Maybe you can have a try in another PR afterwards? Not against at all if possible, for sure it's cool if we can mimimize the extra types we have to add

frontend/src/lib/stores/sns.store.ts
lmuntaner: Not sure why would that matter, but ok. This can easily be changed if this becomes a burden or we change how we fetch.

I like that now we abstracted this into the derived stores.

frontend/src/lib/stores/sns.store.ts
lmuntaner: yes true, it can not be just the buyerState. And this is the current user commitments actually also, not any user.

Anyway, this PR is already a huge step forward. Thanks!

lmuntaner (approved):

PR #1180, opened 2022-07-28
chore: Script to populate canister_ids.json from the testnets branch

bitdivine:
# Motivation
Changing canister IDs disrupt developer workflow.  Unfortunately canister IDs will be changing quite a bit over the next few weeks so we need a way of syncing those changes fairly seamlessly.  In short we need DNS, but that is a conversation for another day.

Testnet canister IDs are published here: https://github.com/dfinity/nns-dapp/blob/testnets/testnets/canister_ids.json

They can be copied to the local branch with e.g.:
```
git fetch && git show origin/testnets:testnets/canister_ids.json >canister_ids.json
```
However we might want not to clobber the whole local file.

This PR starts the conversation by merging with the local file.  Maybe we always want to keep the local canister ID for nnsdapp, for example.  If we specify the behaviour we want, we can modify this PR accordingly.

# Changes
* Added a script to merge the local canister_ids.json with the published testnet canister IDs.

# Tests


peterpeterparker (approved):
rubber stamp

PR #1182, opened 2022-07-28
style: attach HW values

mstrasinskis:
# Motivation

Update color of values  in HW modal.

# Changes

- added `value` class

# Tests

Only style update.


peterpeterparker (approved):
thx

PR #1183, opened 2022-07-28
Chore: Make Pending Sns Proposal

lmuntaner:
# Motivation

Create SNS proposals in testnet using pending proposals in the testnet.

# Changes

* Add swapCanisterId param to make proposal functions.
* Read snsSummariesStore in makeDummyProposals service and select a pending proposal.

# Tests

Only for testnet.

### DIFF
--- a/frontend/src/lib/services/neurons.services.ts
+++ b/frontend/src/lib/services/neurons.services.ts
@@ -1,7 +1,6 @@
 import { AnonymousIdentity, type Identity } from "@dfinity/agent";
 import { Topic, type NeuronId, type NeuronInfo } from "@dfinity/nns";
 import { Principal } from "@dfinity/principal";
-import { SnsSwapLifecycle } from "@dfinity/sns";
 import { get } from "svelte/store";
 import { makeDummyProposals as makeDummyProposalsApi } from "../api/dev.api";
 import {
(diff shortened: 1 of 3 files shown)


lmuntaner:
@mstrasinskis please review

This will make it easier to create the decentralized sale proposal.

The only "issue" is that the testnet user needs to wait until the sns are loaded in the store.

frontend/src/lib/services/neurons.services.ts
```diff
@@ -845,9 +847,18 @@ export const makeDummyProposals = async (neuronId: NeuronId): Promise<void> => {
     const identity: Identity = await getIdentityOfControllerByNeuronId(
       neuronId
     );
+    const projects = get(snsSummariesStore);
```
mstrasinskis, line 850: Probably some having an alert when projects are not loaded will improve the UX.
In case when the neuron-detail page is opened directly, will the snsSummaries loaded?

mstrasinskis (approved):
LGTM, thank you.

frontend/src/lib/services/neurons.services.ts
```diff
@@ -1,6 +1,7 @@
 import { AnonymousIdentity, type Identity } from "@dfinity/agent";
 import { Topic, type NeuronId, type NeuronInfo } from "@dfinity/nns";
 import { Principal } from "@dfinity/principal";
+import { SnsSwapLifecycle } from "@dfinity/sns";
```
peterpeterparker, line 4: not sure that's a good idea, isn't the nns-js library now loaded straight when you start the app because this references is added here?

lmuntaner: Yes, let me add lazy loading somehow then.

lmuntaner: done

lmuntaner:

frontend/src/lib/services/neurons.services.ts
```diff
@@ -847,13 +845,15 @@ export const makeDummyProposals = async (neuronId: NeuronId): Promise<void> => {
     const identity: Identity = await getIdentityOfControllerByNeuronId(
       neuronId
     );
+    const { snsSummariesStore } = await import("../stores/sns.store");
```
peterpeterparker, line 848: nice, never thought about it!

peterpeterparker (approved):
Thx

PR #1186, opened 2022-07-28
L2-867: SNS neurons list UI

lmuntaner:
# Motivation

User can see the neurons related to a project. Mock data only.

# Changes

* Implement SnsNeurons page component.
* New SnsNeuronCard.
* New set of sns-neurons.utils.ts
* Created a couple of new components used both in NeuronCard and SnsNeuronCard.
* New sns api to list neuron. Returns mock data.
* New set of stores for sns neurons. Same pattern as neurons store.
* New sns neurons service to load sns neurons into snsNeurons store. No error handling since it's all mocked data for now.

# Tests

* Test page component SnsNeurons
* Test SnsNeuronCard
* Test new utils
* Fix broken tests.
* Tests for new sns neurons stores.

### DIFF
--- a/frontend/src/lib/pages/SnsNeurons.svelte
+++ b/frontend/src/lib/pages/SnsNeurons.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-  import { onMount } from "svelte";
   import SkeletonCard from "../components/ui/SkeletonCard.svelte";
   import Value from "../components/ui/Value.svelte";
   import { authStore } from "../stores/auth.store";
@@ -8,13 +7,14 @@
   import { loadSnsNeurons } from "../services/sns-neurons.services";
   import SnsNeuronCard from "../components/sns-neurons/SnsNeuronCard.svelte";
   import type { SnsNeuron } from "@dfinity/sns";
+  import { snsProjectSelectedStore } from "../stores/projects.store";
 
   let loading = true;
-  onMount(() => {
-    loadSnsNeurons();
-    setTimeout(() => {
-      loading = false;
-    }, 1000);
+
+  snsProjectSelectedStore.subscribe(async (selectedProjectCanisterId) => {
+    loading = true;
+    await loadSnsNeurons(selectedProjectCanisterId);
+    loading = false;
   });
 
   let principalText: string = "";
--- a/frontend/src/lib/pages/SnsNeurons.svelte
+++ b/frontend/src/lib/pages/SnsNeurons.svelte
@@ -5,16 +5,25 @@
   import { authStore } from "../stores/auth.store";
   import { sortedSnsNeuronStore } from "../stores/snsNeurons.store";
   import { i18n } from "../stores/i18n";
+  import { loadSnsNeurons } from "../services/sns-neurons.services";
+  import SnsNeuronCard from "../components/sns-neurons/SnsNeuronCard.svelte";
+  import type { SnsNeuron } from "@dfinity/sns";
+
+  let loading = true;
+  onMount(() => {
+    loadSnsNeurons();
+    setTimeout(() => {
+      loading = false;
+    }, 1000);
+  });
 
   let principalText: string = "";
   $: principalText = $authStore.identity?.getPrincipal().toText() ?? "";
 
-  let loading = true;
-  onMount(() => {
-    setTimeout(() => {
-      loading = false;
-    }, 3000);
-  });
+  const goToNeuronDetails = (neuron: SnsNeuron) => () => {
+    // TODO: Go to neuron details
+    console.log("Going to details:", neuron.id);
+  };
 </script>
 
 <section data-tid="sns-neurons-body">
@@ -28,7 +37,18 @@
     <SkeletonCard />
   {:else}
     {#each $sortedSnsNeuronStore as neuron}
-      <div>{neuron.id.join("")}</div>
+      <SnsNeuronCard
+        role="link"
+        {neuron}
+        ariaLabel={$i18n.neurons.aria_label_neuron_card}
+        on:click={goToNeuronDetails(neuron)}
+      />
     {/each}
   {/if}
 </section>
+
+<style lang="scss">
+  p:last-of-type {
+    margin-bottom: var(--padding-3x);
+  }
+</style>
--- a/frontend/src/lib/pages/SnsNeurons.svelte
+++ b/frontend/src/lib/pages/SnsNeurons.svelte
@@ -1,10 +1,34 @@
 <script lang="ts">
-  import type { Principal } from "@dfinity/principal";
+  import { onMount } from "svelte";
+  import SkeletonCard from "../components/ui/SkeletonCard.svelte";
+  import Value from "../components/ui/Value.svelte";
+  import { authStore } from "../stores/auth.store";
+  import { sortedSnsNeuronStore } from "../stores/snsNeurons.store";
+  import { i18n } from "../stores/i18n";
 
-  export let rootCanisterId: Principal;
+  let principalText: string = "";
+  $: principalText = $authStore.identity?.getPrincipal().toText() ?? "";
+
+  let loading = true;
+  onMount(() => {
+    setTimeout(() => {
+      loading = false;
+    }, 3000);
+  });
 </script>
 
-<!-- TODO: https://dfinity.atlassian.net/browse/L2-867 -->
-<div data-tid="sns-neurons-body">
-  {`SNS Neurons Page for ${rootCanisterId.toText()}`}
-</div>
+<section data-tid="sns-neurons-body">
+  <p class="description">
+    {$i18n.neurons.principal_is}
+    <Value>{principalText}</Value>
+  </p>
+
+  {#if loading}
+    <SkeletonCard />
+    <SkeletonCard />
+  {:else}
+    {#each $sortedSnsNeuronStore as neuron}
+      <div>{neuron.id.join("")}</div>
+    {/each}
+  {/if}
+</section>
--- a/frontend/src/lib/pages/SnsNeurons.svelte
+++ b/frontend/src/lib/pages/SnsNeurons.svelte
@@ -2,20 +2,27 @@
   import SkeletonCard from "../components/ui/SkeletonCard.svelte";
   import Value from "../components/ui/Value.svelte";
   import { authStore } from "../stores/auth.store";
-  import { sortedSnsNeuronStore } from "../stores/snsNeurons.store";
+  import { sortedSnsNeuronStore } from "../stores/sns-neurons.store";
   import { i18n } from "../stores/i18n";
   import { loadSnsNeurons } from "../services/sns-neurons.services";
   import SnsNeuronCard from "../components/sns-neurons/SnsNeuronCard.svelte";
   import type { SnsNeuron } from "@dfinity/sns";
   import { snsProjectSelectedStore } from "../stores/projects.store";
+  import { getSnsNeuronId } from "../utils/sns-neuron.utils";
+  import type { Unsubscriber } from "svelte/store";
+  import { onDestroy } from "svelte";
 
   let loading = true;
 
-  snsProjectSelectedStore.subscribe(async (selectedProjectCanisterId) => {
-    loading = true;
-    await loadSnsNeurons(selectedProjectCanisterId);
-    loading = false;
-  });
+  const unsubscribe: Unsubscriber = snsProjectSelectedStore.subscribe(
+    async (selectedProjectCanisterId) => {
+      loading = true;
+      await loadSnsNeurons(selectedProjectCanisterId);
+      loading = false;
+    }
+  );
+
+  onDestroy(unsubscribe);
 
   let principalText: string = "";
   $: principalText = $authStore.identity?.getPrincipal().toText() ?? "";
@@ -36,7 +43,7 @@
     <SkeletonCard />
     <SkeletonCard />
   {:else}
-    {#each $sortedSnsNeuronStore as neuron}
+    {#each $sortedSnsNeuronStore as neuron (getSnsNeuronId(neuron))}
       <SnsNeuronCard
         role="link"
         {neuron}
--- a/frontend/src/lib/services/sns-neurons.services.ts
+++ b/frontend/src/lib/services/sns-neurons.services.ts
@@ -1,6 +1,6 @@
 import type { Principal } from "@dfinity/principal";
 import { querySnsNeurons } from "../api/sns.api";
-import { snsNeuronsStore } from "../stores/snsNeurons.store";
+import { snsNeuronsStore } from "../stores/sns-neurons.store";
 
 // TODO: Implement with update and query and error handling https://dfinity.atlassian.net/browse/L2-869
 export const loadSnsNeurons = async (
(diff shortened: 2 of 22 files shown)


lmuntaner:

We discussed briefly how to implement this, but I'd like you to review before I continue and add all the proper tests.

lmuntaner:
UI screenshots:

Screenshot 2022-07-28 at 17 25 06
Screenshot 2022-07-28 at 17 37 43

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -0,0 +1,10 @@
+import { querySnsNeurons } from "../api/sns.api";
+import { snsNeuronsStore } from "../stores/snsNeurons.store";
+
+export const loadSnsNeurons = async (): Promise<void> => {
+  const snsNeurons = await querySnsNeurons();
```
peterpeterparker, line 5: this will become query+update right?

frontend/src/lib/stores/snsNeurons.store.ts
```diff
@@ -0,0 +1,44 @@
+import type { SnsNeuron } from "@dfinity/sns";
+import { derived, writable, type Readable } from "svelte/store";
+import { sortSnsNeuronsByCreatedTimestamp } from "../utils/sns-neuron.utils";
+
+export interface NeuronsStore {
```
peterpeterparker, line 5: Just an idea I had, let me know what you think:

Assuming it would work out, would it be interesting to have a single neuron store for both types of neurons? UX wise it's either one or the other so it's maybe interesting to only have one set of neurons in memory?

```
export interface NeuronsStore {
  neurons?: <NeuronInfo | SnsNeuron>[];
}
```

then we can maybe have derived neuron to make easier the usage in the components?

```
export const nnsNeuron = derived([neuronsStore], here filter nns
export const snsNeuron = derived([neuronsStore], here filter sns
```

Again just an idea I had. Not sure if it's better or worth.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: yes, correct.

I didn't want to implement all the complexity of the service for now.

I will add a todo with the issue

frontend/src/lib/stores/snsNeurons.store.ts
lmuntaner: Actually, I realised that something is missing in this store. Which project the neuron belongs to.

The type could be:

```
export interface SnsNeuronsStore {
  [rootCanisterId: string]: {
    neurons: SnsNeuron[],
    certified: boolean,
  }
}
```

At the moment, differentiating between neurons is a little bit ugly. We would have the check the properties. That on top of the need for the project root canister id, then it makes it more complex to use the same store.

What do you think?

lmuntaner: Then, I can use a derived store from the selecte project and this store and return the neurons.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: added

frontend/src/lib/stores/snsNeurons.store.ts
lmuntaner: Refactored with agreed interface and two stores.

Most important, the derived store uses now the snsProjectSelectedStore to return an array of neurons.

lmuntaner:
UX Video:

https://user-images.githubusercontent.com/4550653/181750088-b15f97a4-e400-44ab-9630-085fef7a8202.mov

lmuntaner:

Sorry for the long PR, but most lines are tests...

frontend/src/lib/stores/snsNeurons.store.ts
```diff
@@ -0,0 +1,61 @@
+import type { Principal } from "@dfinity/principal";
```
peterpeterparker, line 1: can you rename `snsNeurons.store.ts` and `.spec` into `sns-neurons`?

frontend/src/lib/stores/snsNeurons.store.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { SnsNeuron } from "@dfinity/sns";
+import { derived, writable, type Readable } from "svelte/store";
+import { sortSnsNeuronsByCreatedTimestamp } from "../utils/sns-neuron.utils";
+import { snsProjectSelectedStore } from "./projects.store";
+
+export interface NeuronsStore {
+  [rootCanisterId: string]: {
```
peterpeterparker, line 8: can you add a comment why we need the `rootCanisterId` as this particular to this store?

frontend/src/lib/stores/snsNeurons.store.ts
```diff
@@ -0,0 +1,61 @@
...
+        [rootCanisterId.toText()]: {
+          neurons,
+          certified,
+        },
+      }));
+    },
+
+    // Used in tests
+    reset() {
+      set({});
```
peterpeterparker, line 46: the store does not need to be of `type | undefined | null` so that we can know when it has been initialize or not?

typically we often do:

```
reset() {
      set(undefined);
    },
    ```

File: frontend/src/lib/utils/sns-neuron.utils.ts [hunk 6]
```diff
@@ -0,0 +1,68 @@
...
+import type { SnsNeuron } from "@dfinity/sns";
+import { nowInSeconds } from "./date.utils";
+import { stateTextMapper, type StateInfo } from "./neuron.utils";
+
+export type SnsNeuronState = NeuronState;
+
+export const sortSnsNeuronsByCreatedTimestamp = (
+  neurons: SnsNeuron[]
+): SnsNeuron[] =>
+  [...neurons].sort((a, b) =>
```
peterpeterparker, line 11: nth: deconstruct 😉

haha I should check if there is an eslint rule one of these days

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -0,0 +1,68 @@
...
+    Number(b.created_timestamp_seconds - a.created_timestamp_seconds)
+  );
+
+export const getSnsNeuronState = ({
+  dissolve_state,
+}: SnsNeuron): SnsNeuronState => {
+  const dissolveState = dissolve_state[0];
```
peterpeterparker, line 18: can you add a `TODO: use upcoming fromDefinedNullable`?

@mstrasinskis had the [idea](https://github.com/dfinity/nns-dapp/pull/1168#discussion_r929068912) to create such a new util function

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -0,0 +1,68 @@
+import { NeuronState } from "@dfinity/nns";
+import type { SnsNeuron } from "@dfinity/sns";
+import { nowInSeconds } from "./date.utils";
+import { stateTextMapper, type StateInfo } from "./neuron.utils";
+
+export type SnsNeuronState = NeuronState;
```
peterpeterparker, line 6: this should not find place in the `types`?
with a small comment?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -0,0 +1,68 @@
...
+export const getSnsStateInfo = (neuron: SnsNeuron): StateInfo => {
+  const state = getSnsNeuronState(neuron);
+  return stateTextMapper[state] ?? stateTextMapper[NeuronState.UNSPECIFIED];
+};
+
+export const getSnsDissolvingTimeInSeconds = (
+  neuron: SnsNeuron
+): bigint | undefined => {
+  const neuronState = getSnsNeuronState(neuron);
+  const dissolveState = neuron.dissolve_state[0];
```
peterpeterparker, line 40: dito

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  ) {
+    return dissolveState.WhenDissolvedTimestampSeconds - BigInt(nowInSeconds());
+  }
+};
+
+export const getSnsLockedTimeInSeconds = (
+  neuron: SnsNeuron
+): bigint | undefined => {
+  const neuronState = getSnsNeuronState(neuron);
+  const dissolveState = neuron.dissolve_state[0];
```
peterpeterparker, line 54: dito (Maks was right this function will be great)

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (
+    neuronState === NeuronState.LOCKED &&
+    dissolveState !== undefined &&
+    "DissolveDelaySeconds" in dissolveState
+  ) {
+    return dissolveState.DissolveDelaySeconds - BigInt(nowInSeconds());
+  }
+};
+
+export const getSnsNeuronStake = (neuron: SnsNeuron): bigint =>
```
peterpeterparker, line 64: nth: deconstruct 😉

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -0,0 +1,68 @@
...
+  neuron.cached_neuron_stake_e8s - neuron.neuron_fees_e8s;
+
+export const getSnsNeuronId = (neuron: SnsNeuron): string =>
```
peterpeterparker, line 67: can you add a small comment, from reading the code i'm not sure what it does. without looking at the type I'm confuse about the `id[0]?.id`

btw. same Maks todo

frontend/src/lib/stores/snsNeurons.store.ts
```diff
@@ -0,0 +1,61 @@
...
+    },
+  };
+};
+
+export const snsNeuronsStore = initSnsNeuronsStore();
+
+export const sortedSnsNeuronStore: Readable<SnsNeuron[]> = derived(
```
peterpeterparker, line 53: cool 👍

frontend/src/lib/pages/SnsNeurons.svelte
```diff
@@ -1,10 +1,54 @@
...
+  <p class="description">
+    {$i18n.neurons.principal_is}
+    <Value>{principalText}</Value>
+  </p>
+
+  {#if loading}
+    <SkeletonCard />
+    <SkeletonCard />
+  {:else}
+    {#each $sortedSnsNeuronStore as neuron}
```
peterpeterparker, line 39: it maybe needs a key? maybe not but maybe it will help re-render when user will switch from his/her neurons of project Tetris to project DonkeyKong?

frontend/src/lib/pages/SnsNeurons.svelte
```diff
@@ -1,10 +1,54 @@
...
+  import { i18n } from "../stores/i18n";
+  import { loadSnsNeurons } from "../services/sns-neurons.services";
+  import SnsNeuronCard from "../components/sns-neurons/SnsNeuronCard.svelte";
+  import type { SnsNeuron } from "@dfinity/sns";
+  import { snsProjectSelectedStore } from "../stores/projects.store";
 
-  export let rootCanisterId: Principal;
+  let loading = true;
+
+  snsProjectSelectedStore.subscribe(async (selectedProjectCanisterId) => {
```
peterpeterparker, line 14: unsubscribe is missing I think

frontend/src/lib/stores/snsNeurons.store.ts
lmuntaner: sure

frontend/src/lib/stores/snsNeurons.store.ts
lmuntaner: I think that the initialization is more important at the project level. To know whether this store has been initialized or not doesn't give much information at the project level to show or not loading.

I considered to show loading more whether the rootCanisterId exists in the store or not. Then we know. But at the high level store, I am not sure that knowing whether it has been initialized or not is useful information and adds the complexity of checking whether is an object to add the key.

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: Both are the same props, I would need to rename to differentiate them. That's why I thought that it was easier to read like this.

But sure, I can do that.

frontend/src/lib/stores/snsNeurons.store.ts
lmuntaner: yes, shall we say that the pattern of file names is with `-` except for components?

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: yep

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: Since the other one was in the utils I kept the same pattern.

But no problem, let me move it there.

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: It's confusing, the id type is:

```
type Neuron {
  id: { id: number[] },
  //...
```

I will add a comment to explain it as well as Maks' comment

frontend/src/lib/pages/SnsNeurons.svelte
lmuntaner: good point, yes, makes sense

frontend/src/lib/pages/SnsNeurons.svelte
lmuntaner: yep, true!

frontend/src/lib/stores/snsNeurons.store.ts
lmuntaner: done

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: done

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: done

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: done

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: done

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: done

frontend/src/lib/pages/SnsNeurons.svelte
lmuntaner: done

lmuntaner:
@peterpeterparker ready for next round

frontend/src/lib/stores/snsNeurons.store.ts
peterpeterparker: you are right  about the project level, agree with you. I guess the fact that the neurons is an array also solved this and don't see any reason after thinking more about it. Let's go as it is, thx.

peterpeterparker (approved):
LGTM, thx

PR #1191, opened 2022-07-29
Update en.json

dubovitskaya:
# Motivation

The tooltip text incorrectly saying that maturity can still be distributed within three days.

# Changes

Fix the tooltip text saying that maturity can still be distributed within four days, not three.

# Tests

Textual change, no tests required.


peterpeterparker (approved):
Thanks!

PR #1193, opened 2022-08-01
feat: woff2 mime type support

peterpeterparker:
# Motivation

Add font `woff2` to the list of supported mime type.

# Note

Found the mime type `font/woff2` mentioned in various discussion on SO and also in this link: https://www.iana.org/assignments/media-types/media-types.xhtml#font

# Changes

- add response mime type `font/woff2` to the asset canister


bitdivine (approved):
Thanks a lot!  Die, ttf, die!  I may have loved you once but there is a slimmer format now.

peterpeterparker:
Haha yeah die ttf 🤣. At the end, the new woff2 font won't be that much smaller (like 5 or 10kb) but still, it's cool to use the most modern format particularly if we would add more fonts

PR #1194, opened 2022-08-02
build: bump sns nightly with peer agent-js

peterpeterparker:
# Motivation

Bump to the last nightly version of `sns-js` that requires `agent-js` as a peer dependencies.

# Note

`nns-js` cannot be upgraded as well. It first needs PR https://github.com/dfinity/ic-js/pull/162 because otherwise the `jest` test suite won't run.

# Changes

- bump `sns-js` nightly
- add mandatory `agent-js` dependencies to dependencies
- clean up `package-lock` by removing local path (`../../../ic-js/...`) added for development purpose

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -67,6 +67,7 @@
     "@dfinity/candid": "^0.12.0",
     "@dfinity/identity": "^0.12.0",
     "@dfinity/nns": "nightly",
+    "@dfinity/principal": "^0.12.0",
     "@dfinity/sns": "nightly",
     "@ledgerhq/hw-transport-node-hid-noevents": "^6.27.1",
     "@ledgerhq/hw-transport-webhid": "^6.27.1",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -63,6 +63,9 @@
   "dependencies": {
     "@dfinity/agent": "^0.12.0",
     "@dfinity/auth-client": "^0.12.0",
+    "@dfinity/authentication": "^0.12.0",
+    "@dfinity/candid": "^0.12.0",
+    "@dfinity/identity": "^0.12.0",
```
lmuntaner, line 68: should we also add `@dfinity/principal`? We use it directly but it's not set as dependency.

lmuntaner (approved):
Just a minor comment, looks good to me

frontend/package.json
peterpeterparker: sure, makes sense. done.

PR #1196, opened 2022-08-02
feat: read and display overall current commitment from "derived"

peterpeterparker:
# Motivation

Read and display overall current commitment from "derived" information of the swap state response instead of querying the information twice.

# Note

Following this PR, the overall current commitment is not refreshed after user participate to a swap. I'll solve this in another PR (L2-903).

# Changes

- replace `SnsSwapCommitment.currentCommitment` with mandatory `SnsSummary.derived`
- adapt API, store and component accordingly


lmuntaner (approved):
Looks good to me

PR #1198, opened 2022-08-02
chore: Bump IC to the latest version

bitdivine:
# Motivation
The SNS has a new flag to prevent uploading wasms to the wasms canister.  We need to set it to false for testing.

# Changes
* Bump IC
* Set the `sns-wasm` canister `access_controls_enabled` to `false`.

# Tests
I have deployed small12 with this change.


peterpeterparker (approved):
rubber stamp, thx

PR #1199, opened 2022-08-02
feat: user max participation check

peterpeterparker:
# Motivation

- button “Participate” should be disabled if user max commitment is reached
- if disabled, would be nice to have tooltip "Maximum commitment reached"
- the overall commitment max label should be renamed to "Maximum commitment goal"

Discussed with Mischa.

# Changes

- update max label in commitment progress bar
- extract button and modal to participate to a sale in a dedicated component
- add logic regarding disabling button and tooltip to new component
- add new reusable util function to get to know if user can participate to a swap

# Screenshots

Capture d’écran 2022-08-02 à 21 28 09


frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -0,0 +1,65 @@
...
+  {:else}
+    <Tooltip
+      id="sns-project-participate-button-tooltip"
+      text={$i18n.sns_project_detail.max_user_commitment_reached}
+    >
+      <button
+        on:click={openModal}
+        class="primary small"
+        data-tid="sns-project-participate-button"
+        disabled={true}>{$i18n.sns_project_detail.participate}</button
```
lmuntaner, line 57: just `disabled`, should be enough, right?

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -0,0 +1,65 @@
...
+      data-tid="sns-project-participate-button"
+      >{$i18n.sns_project_detail.participate}</button
+    >
+  {:else}
+    <Tooltip
+      id="sns-project-participate-button-tooltip"
+      text={$i18n.sns_project_detail.max_user_commitment_reached}
+    >
+      <button
+        on:click={openModal}
```
lmuntaner, line 54: maybe no on:click even? To avoid programatically open the modal?

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -0,0 +1,65 @@
...
+    <button
+      on:click={openModal}
+      class="primary small"
+      data-tid="sns-project-participate-button"
+      >{$i18n.sns_project_detail.participate}</button
+    >
+  {:else}
+    <Tooltip
+      id="sns-project-participate-button-tooltip"
+      text={$i18n.sns_project_detail.max_user_commitment_reached}
```
lmuntaner, line 51: Since the reasons for not participating are more than one, shouldn't we change the tooltip information per case?

If this is only when the user reached the minimum, I would create the util `userReachedMaxCommitment` instead of `canUserParticipateToSwap` to be more explicit of which rule applies here.

You already check whether the lifecycle is open before

lmuntaner (commented):
A couple of comments

frontend/src/lib/components/project-detail/ParticipateButton.svelte
peterpeterparker: sure, done

frontend/src/lib/components/project-detail/ParticipateButton.svelte
peterpeterparker: I created and use `canUserParticipateToSwap`  for convenience reason because we might use it for validation elsewhere - e.g. in the transaction modal - but in that particular case, with other ifs that are implemented around, it does effectively checks that max is not reached therefore the label matches the check.

frontend/src/lib/components/project-detail/ParticipateButton.svelte
peterpeterparker: sure, done

lmuntaner (approved):

PR #1201, opened 2022-08-03
feat: user commitment label

peterpeterparker:
# Motivation

Once the sale is committed, we should not display "your current commitment" but "your commitment" which makes more sense since it's done.

# Changes

- extract label "user commitment" and "user current commitment" to a cmp
- use cmp on launchpad and detail


lmuntaner (approved):
LGTM

PR #1203, opened 2022-08-03
feat: reload sns data after participate swap & use global data for detail page

peterpeterparker:
# Motivation

Issues:

- because we moved away from `currentCommitment` to `derived` information (#1196), the "overall current commitment" information was not updated anymore after user participate to a swap

- we have moved the loading of the sns on the app scale (#1172) but the project detail was still loading a particular sns if directly accessed. as a result the same sns project could have been queried twice (actually four times, 2 query calls, 2 update calls)

Solution:

- query sns summary and swap state after user participate to swap and update the root store that contains the fetched sns data from the backend - the one store that is the source for all derived store

- do not (re)load particular sns when the detail page is accessed but use the global store which already load these information

# Notes

- both solution are provided in a single PR because these are tight to each other.

- when it comes to reloading the data after sale participation, we might not need to reload the summary - i.e. we probably only need the reload of the swap and commitment. for simplicity reason I just reloaded summary and swap as in any case, the loading of the summary is still mocked and will be implemented soon

# Changes

- move `certified` information from global sns query store level to each data contained in store (`QuerySnsSwapState` and `QuerySnsSummary`)
- update `sns.store` to contain an object rather than an array, easier to handle
- add `updateData` to `sns-store`
- adapt consumer of `sns-store` according above changes
- modify `ProjectDetail` and `ReviewParticipate` for the above described solution
- remove `loadSnsSwapCommitment` from function `loadSnsSwapCommitment` in `sns.service`
- add `reload` fonction to project store detail context
- use relad function to well, reload the data after participation

# Screenshots

Capture d’écran 2022-08-03 à 16 45 30

### DIFF
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -269,7 +269,7 @@ export const querySnsSummary = async ({
     ].map(({ canisterIds: { rootCanisterId } }: SnsWrapper, index) => ({
       ...mockSnsSummaryList[index],
       rootCanisterId,
-      certified
+      certified,
     }));
   }
 
@@ -351,7 +351,7 @@ export const querySnsSwapState = async ({
     swapCanisterId,
     swap,
     derived,
-    certified
+    certified,
   };
 };
 
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -270,6 +270,7 @@ export const querySnsSummary = async ({
     ].map(({ canisterIds: { rootCanisterId } }: SnsWrapper, index) => ({
       ...mockSnsSummaryList[index],
       rootCanisterId,
+      certified
     }));
   }
 
@@ -352,6 +353,7 @@ export const querySnsSwapState = async ({
     swapCanisterId,
     swap,
     derived,
+    certified
   };
 };
 
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -274,6 +274,7 @@ export const querySnsSummary = async ({
   }
 
   // TODO(L2-829, L2-751): remove and replace with effective data - i.e. summary comes from sns gov canister through sns wrapper
+  console.log("Sns metadata", summary);
   return mockSnsSummaries.find(
     ({ rootCanisterId: canisterId }: QuerySnsSummary) =>
       canisterId?.toText() === rootCanisterId
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -52,7 +52,6 @@ const listSnses = async ({
 
   const SnsWasmCanister: SnsWasmCanisterCreate = await importSnsWasmCanister();
 
-  console.log({ WASM_CANISTER_ID, path: "src/lib/api/sns.api.ts" });
   const { listSnses }: SnsWasmCanister = SnsWasmCanister.create({
     canisterId: Principal.fromText(WASM_CANISTER_ID),
     agent,
@@ -275,7 +274,6 @@ export const querySnsSummary = async ({
   }
 
   // TODO(L2-829, L2-751): remove and replace with effective data - i.e. summary comes from sns gov canister through sns wrapper
-  console.log("Sns metadata", summary);
   return mockSnsSummaries.find(
     ({ rootCanisterId: canisterId }: QuerySnsSummary) =>
       canisterId?.toText() === rootCanisterId
--- a/frontend/src/lib/components/project-detail/ProjectInfoSection.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectInfoSection.svelte
@@ -10,6 +10,7 @@
     type ProjectDetailContext,
   } from "../../types/project-detail.context";
   import Spinner from "../../../lib/components/ui/Spinner.svelte";
+  import { isNullish } from "../../utils/utils";
 
   const { store: projectDetailStore } = getContext<ProjectDetailContext>(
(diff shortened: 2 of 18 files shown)


frontend/src/lib/api/sns.api.ts
```diff
@@ -52,7 +52,6 @@ const listSnses = async ({
 
   const SnsWasmCanister: SnsWasmCanisterCreate = await importSnsWasmCanister();
 
-  console.log({ WASM_CANISTER_ID, path: "src/lib/api/sns.api.ts" });
```
lmuntaner, line 55: thanks!

frontend/src/lib/services/sns.services.ts
```diff
@@ -249,31 +237,21 @@ export const participateInSwap = async ({
...
-          certified,
-        });
-        onSuccess?.(swapCommitment);
-      },
-      onError: () => {
-        // TODO: Manage errors https://dfinity.atlassian.net/browse/L2-798
-      },
-    });
+
+    onSuccess?.();
```
lmuntaner, line 253: Why do it here if it's just a callback?

In that case I guess it makes sense to do it in the component and keep it simple.

frontend/src/lib/stores/sns.store.ts
```diff
@@ -80,11 +89,42 @@ const initSnsQueryStore = () => {
...
+        summaries,
+        swaps,
+      });
+    },
+
+    updateData({
+      data: [updatedSummary, updatedSwap],
+      rootCanisterId,
+    }: {
+      data: [QuerySnsSummary | undefined, QuerySnsSwapState | undefined];
```
lmuntaner, line 103: Does `undefined` mean to remove that summary?

If so, it might make sense to add a comment about it.

frontend/src/lib/stores/sns.store.ts
```diff
@@ -80,11 +89,42 @@ const initSnsQueryStore = () => {
...
-      response: [QuerySnsSummary[], QuerySnsSwapState[]];
-      certified: boolean;
+    setData([summaries, swaps]: [QuerySnsSummary[], QuerySnsSwapState[]]) {
+      set({
+        summaries,
+        swaps,
+      });
+    },
+
+    updateData({
```
lmuntaner, line 99: pending test still right?

lmuntaner (commented):
Looks good so far, just a couple of comments

frontend/src/lib/components/project-detail/ProjectInfoSection.svelte
```diff
@@ -9,42 +9,47 @@
...
-      >
-      <span class="value" slot="value">{summary.tokenName}</span>
-    </KeyValuePair>
-    <KeyValuePair>
-      <svelte:fragment slot="key"
-        >{$i18n.sns_project_detail.token_symbol}</svelte:fragment
-      >
-      <span class="value" slot="value">{summary.symbol}</span>
-    </KeyValuePair>
+{#if summary === undefined || summary === null}
```
mstrasinskis, line 22: isNullish(summary) ?

frontend/src/lib/services/sns.services.ts
peterpeterparker: It used to be a callback and it's kind of handy to test as well - i.e. when we test the service we test that the is effectively called

But no strong feeling at all, if you rather like me to remove the callback and process the call in the component, happy to do it

frontend/src/lib/stores/sns.store.ts
peterpeterparker: yes, absolutely

frontend/src/lib/services/sns.services.ts
lmuntaner: Yeah, I think it would be cleaner outside. This is a callback done at the end. There is no reason to do it inside here instead of outside. At least for now.

mstrasinskis (approved):
LGTM

frontend/src/lib/services/sns.services.ts
peterpeterparker: Sure. Done.

frontend/src/lib/components/project-detail/ProjectInfoSection.svelte
peterpeterparker: Sure. Funny thing, I first had a variable that was assigned by `isNullish` but the linter did not interfer it and was throwing an error, that's why I implemented above code but, using directly `isNullish` in the template solve it 👍.

Done.

frontend/src/lib/stores/sns.store.ts
peterpeterparker: Sure. Done.

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
```diff
@@ -57,9 +56,10 @@
         account,
         amount: icpAmount,
         rootCanisterId: $store.summary.rootCanisterId,
-        onSuccess: (swapCommitment) => ($store.swapCommitment = swapCommitment),
       });
       if (success) {
+        reload();
```
lmuntaner, line 61: `await reload()`?

Or is it intended to be reloaded in the background?

lmuntaner (approved):
One minor comment, already approving.

frontend/src/lib/components/project-detail/ReviewParticipate.svelte
peterpeterparker: Good catch, done.

frontend/src/lib/stores/sns.store.ts
peterpeterparker: Done.

lmuntaner:
@peterpeterparker already approved on my side 😄

PR #1208, opened 2022-08-04
build: update nns, sns and agent-js

peterpeterparker:
# Motivation

Update nightly version of the lib that use agent-js v0.12.2


lmuntaner (approved):
Thanks!

PR #1210, opened 2022-08-04
feat: load sns project and token metadata from backend

peterpeterparker:
# Motivation

Load sns project (title, description, etc.) and token (name and symbol) metadata from backend.

# Note

At least for MVP, these information are mandatory. If not provided, then the corresponding project will be filter from NNS-dapp. This for code simplification and because users need information to invest in projects they trust.

# Sns-js

Corresponding `sns-js` PR is https://github.com/dfinity/ic-js/pull/167

# PRs

#1211 👉 This PR 👉 main

# Changes

- rename api `querySnsSummaries` to `queryAllSnsMetadata` and effectively implement backend call
- update sns store to contains query metadata instead of mock summaries
- add effective type nns-dapp will use for token (name and symbol) and project (title, etc.) metadata
- update `sns.utils` to filter and map mandatory data from the backend to data that can be used in the derived stores

# Tests

Most tests (except componennts in #1211) already existed therefore mostly adapted to new type structure. Two tests added in `sns.utils.spec` to test new filtering options.

### DIFF
--- a/frontend/src/lib/types/sns.ts
+++ b/frontend/src/lib/types/sns.ts
@@ -36,16 +36,6 @@ export interface SnsSummary {
   // Used to calculate the account for the participation.
   swapCanisterId: Principal;
 
-  // TODO: to be replaced with real types for real data
-
-  logo: string;
-  name: string;
-  symbol: string;
-  url: string;
-
-  tokenName: string;
-  description: string;
-
   /**
    * The metadata of the Sns project (title, description, etc.)
    */
--- a/frontend/src/lib/types/sns.ts
+++ b/frontend/src/lib/types/sns.ts
@@ -5,16 +5,17 @@ import type {
   SnsSwapDerivedState,
   SnsSwapInit,
   SnsSwapState,
+  SnsTokenMetadata,
 } from "@dfinity/sns";
 
 /**
  * Metadata are full optional in Candid files but mandatory currently in NNS-dapp
  */
 export interface SnsSummaryMetadata {
-  url: string,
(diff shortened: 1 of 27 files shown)


frontend/src/lib/types/sns.ts
```diff
@@ -27,6 +46,16 @@ export interface SnsSummary {
   tokenName: string;
   description: string;
 
+  /**
+   * The metadata of the Sns project (title, description, etc.)
+   */
+  metadata: SnsSummaryMetadata;
```
lmuntaner, line 52: Do we still need the properties above? `logo`, `name`, `symbol`, `url`, `tokenName` and `description`?

Shouldn't the components read from the `metadata`?

lmuntaner (commented):
A comment

frontend/src/lib/types/sns.ts
peterpeterparker: I split the feature in two PRs. This PR that consumes the data and another that uses the data - i.e. these properties are removed there.

> PRs
> #1211 👉 This PR 👉 main

peterpeterparker: Easier to review if split with so much changes.

lmuntaner (approved):

frontend/src/lib/types/sns.ts
lmuntaner: Normally yes, but I found it confusing because I found missing parts in one and the other.

Already approved, thanks!

PR #1212, opened 2022-08-04
Remove remote canister ID for small11

bitdivine:
# Motivation
If a remote canister Id is specified, the canister is not deployed.

# Changes
- Don't specify the wasm canister ID on small11

# Tests
- deploying...


peterpeterparker (approved):
thx!

PR #1214, opened 2022-08-04
feat: update label to replace "agree" with "proceed" as there is no agreement

peterpeterparker:
# Motivation

There is no "agreement" therefore the confirmation to participate to a decentralized sale should display the act of proceeding. 

# Changes

- update i18n label


lmuntaner (approved):

nmattia:
did we agree to proceed on this?

PR #1215, opened 2022-08-04
Proposal payload add wasm request

mstrasinskis:
# Motivation

Added new proposal type [`AddWasmRequest`](https://github.com/dfinity/ic/blob/187e933e73867efc3993572abc6344b8cedfafe5/rs/nns/sns-wasm/gen/ic_sns_wasm.pb.v1.rs#L62) support

_Made with @hpeebles_ 🤘

# Changes

- payload candids were updated to the latest blessed version 
- initial instruction for payload update (`HOWTO.md`)

# Tests

- Tested manually on currently available proposals only https://unr2c-oyaaa-aaaaa-aaa7a-cai.small06.dfinity.network/#/proposal/82


HOWTO.md
```diff
@@ -0,0 +1,10 @@
+## Update proposal payload definition
+
+- In `Cargo.toml` update the rev using the latest `Bless Replica Version`
+- In case of new proposal type
+  - Add a dependency (e.g. "ic-sns-wasm") in the Cargo.toml (because the new proposal type is executed by new canister)
```
hpeebles, line 5: It is rare that you need to add a new dependency, so I think we should specify that you only do this if needed.
eg. 'If needed, ...'

HOWTO.md
```diff
@@ -0,0 +1,10 @@
...
+  - Add new payload type (e.g. `AddWasmRequest`) in `proposals.rs`
+  - Update `transform` if needed (see `NNS function 3 - AddNNSCanister`) (use `Trimmed` suffix to define transformed version)
```
hpeebles, line 7: I think we should say something along the lines of:
- If the payload needs to be transformed for display (e.g. if it is too large), define the type into which the payload must be transformed, then implement `From<OriginalPayloadType` for this new type.
- Update the `match nns_function` expression to include the new function (use either `identity` or `transform` depending on if the payload needs to be transformed).

HOWTO.md
mstrasinskis: Updated.

HOWTO.md
mstrasinskis: Updated. Thanks

rs/src/proposals.rs
```diff
@@ -93,6 +93,8 @@ fn transform_payload_to_json(nns_function: i32, payload_bytes: &[u8]) -> Result<
     }
 
     // See full list here - https://github.com/dfinity/ic/blob/5b2647754d0c2200b645d08a6ddce32251438ed5/rs/nns/governance/gen/ic_nns_governance.pb.v1.rs#L1690
+    // identity: deserealize -> serealize to JSON
```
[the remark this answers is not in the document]
mstrasinskis: Good catch. Fixed.

bitdivine:
Notes on fixing the broken test:
- The error message is the same as the one we had in the ic repo a while ago from a crono dependency: https://app.slack.com/client/T43F9UHS5/D031GD20W01/thread/C0370Q369QW-1660743994.942279

### Local testing:

* dfx start --clean
* dfx nns install
* dfx nns import --network-mapping local=mainnet
* Set canister IDs because that last step is incomplete

### Main
Fails decoding stable memory. This is expected; we had breaking changes.  Create a new canister (by deleting the local entry in .dfx/local/canister_ids.json) and redeploy.  Then it works.

bitdivine (approved):

PR #1217, opened 2022-08-05
L2-870: SNS Add Hotkey

lmuntaner:
# Motivation

User can add a hotkey to an SNS neuron.

# Changes

* New modal AddSnsHotkeyModal. Implements same UI as adding a hotkey for NNS.
* AddSnsHotkeyButton opens AddSnsHotkeyModal.

# Tests

* Test that AddSnsHotkeyButton opens modal.
* Test the flow in AddSnsHotkeyModal

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
@@ -1,11 +1,12 @@
 <script lang="ts">
-  import type { SnsNeuron } from "@dfinity/sns";
+  import type { SnsNeuron, SnsNeuronId } from "@dfinity/sns";
   import { ICON_SIZE_LARGE } from "../../constants/style.constants";
   import IconClose from "../../icons/IconClose.svelte";
   import IconInfo from "../../icons/IconInfo.svelte";
   import IconWarning from "../../icons/IconWarning.svelte";
   import { authStore } from "../../stores/auth.store";
   import { i18n } from "../../stores/i18n";
+  import { fromNullable } from "../../utils/did.utils";
   import {
     getSnsNeuronHotkeys,
     canIdentityManageHotkeys,
@@ -16,6 +17,9 @@
   import AddSnsHotkeyButton from "./actions/AddSnsHotkeyButton.svelte";
 
   export let neuron: SnsNeuron;
+
+  let neuronId: SnsNeuronId | undefined;
+  $: neuronId = fromNullable(neuron.id);
 
   let canManageHotkeys: boolean = true;
   $: canManageHotkeys = canIdentityManageHotkeys({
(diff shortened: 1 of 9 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/183048470-0ac3ca15-37aa-406d-aca5-f01876d980ee.mov

lmuntaner:

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
```diff
@@ -17,6 +18,9 @@
 
   export let neuron: SnsNeuron;
 
+  let neuronId: SnsNeuronId | undefined;
+  $: neuronId = fromNullable(neuron.id);
```
peterpeterparker, line 22: in reality `neuron.id` if `neuron` exists is always given right? that's one of these candid optional thing that is actually always defined?

if yes, you can add a TODO to use `fromDefinedNullable`?

PR has been provided by Maks, utils comes next week (https://github.com/dfinity/ic-js/pull/171)

lmuntaner: yes, exactlu. I'll add the todo

frontend/src/lib/modals/sns/AddSnsHotkeyModal.svelte
```diff
@@ -0,0 +1,54 @@
...
+      toastsStore.error({
+        labelKey: "error.principal_not_valid",
+      });
+      return;
+    }
+    startBusy({ initiator: "add-sns-hotkey-neuron" });
+    await addHotkey({
+      neuronId,
+      hotkey: principal,
+      rootCanisterId: $snsProjectSelectedStore,
```
peterpeterparker, line 30: I was trying to think if there could be an edge case where the `$snsProjectSelectedStore` would not be the one that was previously selected and would be back NNS canister id?

if yes, then maybe we should assign / clone the `snsProjectSelectedStore` to a local variable when the component is loaded?

but that's maybe overengineered...

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -135,17 +135,16 @@ export const addHotkey = async ({
   rootCanisterId,
 }: {
   neuronId: SnsNeuronId;
-  hotkey: string;
+  hotkey: Principal;
   rootCanisterId: Principal;
 }): Promise<{ success: boolean }> => {
   try {
     const permissions = [SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_VOTE];
```
peterpeterparker, line 142: maybe not in the PR, maybe in the "remove" one => this can be extracted to a constant I guess.

frontend/src/lib/modals/sns/AddSnsHotkeyModal.svelte
lmuntaner: I also thought about it.

I will need a context when I sync the neuron after the changes: https://dfinity.atlassian.net/browse/L2-914

I could add the rootCanisterId there as well.

peterpeterparker (approved):
LgTm, thX

frontend/src/lib/modals/sns/AddSnsHotkeyModal.svelte
peterpeterparker: Sounds like a plan, thanks!

PR #1219, opened 2022-08-05
L2-705: SNS Remove Hotkey

lmuntaner:
# Motivation

User is able to remove a hotkey in an SNS neuron.

# Changes

* Add sns api function removeNeuronPermissions.
* Add sns neuron service removeHotkey which uses removeNeuronPermissions.
* Use new sns neuron service in SnsNeuronHotkeysCard.

# Tests

* Test new case in SnsNeuronHotkeysCard.
* Test new sns neuron service
* Test new sns api function.

### DIFF
--- a/frontend/src/lib/services/sns-neurons.services.ts
+++ b/frontend/src/lib/services/sns-neurons.services.ts
@@ -10,6 +10,7 @@ import {
   addNeuronPermissions,
   querySnsNeuron,
   querySnsNeurons,
+  removeNeuronPermissions,
 } from "../api/sns.api";
 import {
   snsNeuronsStore,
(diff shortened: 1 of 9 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/183106575-95ed67c1-96e5-4a10-a128-084b3546e5b2.mov

lmuntaner:

peterpeterparker (approved):
LGTM, thx

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -158,3 +159,33 @@ export const addHotkey = async ({
     return { success: false };
   }
 };
+
+export const removeHotkey = async ({
```
peterpeterparker, line 163: now that I have read the other PR about "adding hotkey", feels like this function looks really close to `addHotkey`

there is probably some potential of refactoring to avoid double code

if you are up-to-it, not against.

peterpeterparker:
Added a new (nth) comment, if you can have a look before merge it's cool

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: Yes, but it changes the sns api it uses, the hotkey is a string here, not a principal. I'd say that the pattern is very similar, but almost all the pieces are different.

This seems to me like boring code. Which is code easy to read, understand and change.

peterpeterparker: Maybe add just a TODO, I mean it's really close...

Capture d’écran 2022-08-05 à 18 05 01

peterpeterparker: oh just noticed, the i18n label for the error in remove is `sns_add_hotkey` not "remove" @lmuntaner

lmuntaner: The hotkey in one is of type Principal in the other of type string also.

But I guess I can use a similar helper like in the followees

PR #1221, opened 2022-08-08
build: integrate @dfinity/utils

peterpeterparker:
# Motivation

Last version of `@dfinity/sns` and `@dfinity/nns` have a peer dependency on  `@dfinity/utils` which we aim to use to share our utilities.

# Changes

- integrate @dfinity/utils
- use `next` instead of `nightly` version
- use existing utils

# Note

There was several todo that pointed out to the use of `fromDefinedNullable` but give the implementation that checks for `undefined` values after calling the utils, it seems that `fromNullable` is the correct utils to use to solve these todos.


frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
```diff
@@ -15,12 +14,13 @@
...
   import AddSnsHotkeyButton from "./actions/AddSnsHotkeyButton.svelte";
+  import { fromDefinedNullable } from "@dfinity/utils";
 
   export let neuron: SnsNeuron;
 
   let neuronId: SnsNeuronId | undefined;
-  // TODO: TODO to use fromDefinedNullable
-  $: neuronId = fromNullable(neuron.id);
+
+  $: neuronId = fromDefinedNullable(neuron.id);
```
mstrasinskis, line 23: Nice 😊

mstrasinskis (approved):
LGTM. Thank you!

PR #1223, opened 2022-08-09
L2-909: Use token symbol from ledger in SnsNeuronCard

lmuntaner:
# Motivation

User sees the token symbol in the SNS Neuron Card.

# Changes

* Add new derived store currentSnsTokenLabelStore to return the token of the selected SNS project.
* Use the new store in the SnsNeuronCard.

# Tests

* Test that SnsNeuronCard does not render ICP but the token name.
* Test the new derived store.


lmuntaner:
New UI:

Screenshot 2022-08-09 at 11 41 55

lmuntaner:

peterpeterparker (commented):
just a minor question(s), looks cool

frontend/src/lib/derived/sns/current-sns-token-label.store.ts
```diff
@@ -0,0 +1,12 @@
+import { derived, type Readable } from "svelte/store";
+import { snsProjectSelectedStore } from "../../stores/projects.store";
+import { snsSummariesStore } from "../../stores/sns.store";
+
+export const currentSnsTokenLabelStore: Readable<string | undefined> = derived(
```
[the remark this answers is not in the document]
lmuntaner: good point in both. That is the symbol. I will also rename.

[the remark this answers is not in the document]
lmuntaner: done!

lmuntaner:
@peterpeterparker ready for new review

I renamed it to `snsTokenSymbolSelectedStore`, I hope that's ok. I felt that Project was not neeed because we alreayd have `sns` there.

frontend/src/tests/lib/derived/sns/sns-token-symbol-selected.store.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+describe("currentSnsTokenLabelStore", () => {
+  const data = snsResponsesForLifecycle({
+    lifecycles: [SnsSwapLifecycle.Open],
+    certified: true,
+  });
+  afterEach(() => {
+    snsQueryStore.reset();
+    snsProjectSelectedStore.set(OWN_CANISTER_ID);
+  });
+  it("returns token value of current selected sns project", () => {
```
peterpeterparker, line 19: is it "return token symbol..." now?

frontend/src/tests/lib/derived/sns/sns-token-symbol-selected.store.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  it("returns undefined if selected project is NNS", () => {
+    snsQueryStore.setData(data);
+    snsProjectSelectedStore.set(OWN_CANISTER_ID);
+    const expectedToken = get(snsTokenSymbolSelectedStore);
+
+    expect(expectedToken).toBeUndefined();
+  });
+
+  it("returns undefined if current selected project has no data", () => {
```
peterpeterparker, line 46: should we also add a test "returns undefined if canister id not found"?

peterpeterparker (approved):
LGTM, thanks. Just added two small comments about the tests

frontend/src/tests/lib/derived/sns/sns-token-symbol-selected.store.spec.ts
lmuntaner: correct, I'll rewrite

frontend/src/tests/lib/derived/sns/sns-token-symbol-selected.store.spec.ts
lmuntaner: canisterId is always set, can't be undefined. It might not be found in the array of summaries, which is also tested.

```
const initSnsProjectSelectedStore = () => {
  const { subscribe, set } = writable<Principal>(OWN_CANISTER_ID);

  return {
    subscribe,
    set,
  };
};
```

peterpeterparker: i mean if `snsProjectSelectedStore.set(THIS_CANISTER_ID_DOES_NOT_EXIST);` but writing it make me realize that's an edge case which might never happens. nvm.

lmuntaner: Edge case, but it happens something similar. When the selected one is the NNS Dapp canister id, then the there is no summary that matches. Therefore, the outcome is the same in both cases. When the canister id does not exist, or the selected project is the nns dapp.

PR #1224, opened 2022-08-09
L2-899: Add Hotey label to SNS Neuron Card

lmuntaner:
# Motivation

User can easily identify SNS neurons managed by hotkeys.

# Changes

* New sns neuron util isHotkeyControllable that returns true if neuron is hotkey.
* Use new util in SnsNeuronCard.

# Tests

* Add test case in SnsNeuronCard.
* Test new sns neuron util.

### DIFF
--- a/frontend/src/lib/components/sns-neurons/SnsNeuronCard.svelte
+++ b/frontend/src/lib/components/sns-neurons/SnsNeuronCard.svelte
@@ -12,7 +12,7 @@
     getSnsNeuronStake,
     getSnsNeuronState,
     getSnsStateInfo,
-    isHotkeyControllable,
+    isUserHotkey,
   } from "../../utils/sns-neuron.utils";
   import IcpComponent from "../ic/ICP.svelte";
   import NeuronCardContainer from "../neurons/NeuronCardContainer.svelte";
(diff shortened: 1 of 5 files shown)


lmuntaner:
New UI:

Screenshot 2022-08-09 at 12 01 00

lmuntaner:

frontend/src/lib/components/sns-neurons/SnsNeuronCard.svelte
```diff
@@ -22,6 +25,12 @@
   export let cardType: CardType = "card";
   export let ariaLabel: string | undefined = undefined;
 
+  let isHotkey: boolean;
```
peterpeterparker, line 28: - is it actually `isHotkey` or `hasHotkey`?

- when I read `isHotkeyControllable`, I understand "can a hotkey be controlled" where the function actually checks if the user principal controls the neuron or if the user principal is one of the hotkey. maybe `isUserPrincipalHotkey` or better idea?

lmuntaner: Good point. I used the same wording as in the NNS Neuron, but it's true that is not very clear.

TBH, I find the whole naming confusing. The hotkeys are actually principals, but then, the label is called Hotkey, and it's on a neuron. If you then enter inside the details, yuo see the hotkeys.

I believe that we should even think of what this means for the user and also change the UI.

To me, `isUserHotkey` makes more sense. I will rename it, but we shuold also review the UI and mention this with Miscna.

peterpeterparker: Sure works for me

peterpeterparker (approved):
Lgtm, thanks

PR #1227, opened 2022-08-09
L2-914: Sync neuron after hotkey mgt

lmuntaner:
# Motivation

User sees new neuron details after adding or removing hotkeys.

# Changes

* Add context in the SnsNeuronDetail "SelectedSnsNeuronContext".
* Children components of SnsNeuronDetail read from context instead of relying on props.
* A function to reload the neuron is added in the SelectedSnsNeuronContext.
* Reload sns neuron is used after successfully removing or adding hotkeys.

# Tests

* Fix and change test components.
* Add check that the reload function is called after adding and removing hotkey.

### DIFF
--- a/frontend/src/lib/types/sns-neuron-detail.context.ts
+++ b/frontend/src/lib/types/sns-neuron-detail.context.ts
@@ -1,0 +1,19 @@
+import type { Principal } from "@dfinity/principal";
+import type { SnsNeuron } from "@dfinity/sns";
+import type { Writable } from "svelte/store";
+
+/**
+ * A store that contains the selected proposal.
+ */
+export interface SelectedSnsNeuronStore {
+  rootCanisterId: Principal | undefined;
+  neuronIdHex: string | undefined;
+  neuron: SnsNeuron | undefined;
+}
+
+export interface SelectedSnsNeuronContext {
+  store: Writable<SelectedSnsNeuronStore>;
+  reload: () => Promise<void>;
+}
+
+export const SELECTED_SNS_NEURON_CONTEXT_KEY = Symbol("selected-sns-neuron");
--- a/frontend/src/lib/types/sns-neuron-detail.context.ts
+++ b/frontend/src/lib/types/sns-neuron-detail.context.ts
@@ -6,8 +6,12 @@ import type { Writable } from "svelte/store";
  * A store that contains the selected sns neuron.
  */
 export interface SelectedSnsNeuronStore {
-  rootCanisterId: Principal | undefined;
-  neuronIdHex: string | undefined;
+  selected:
+    | {
+        rootCanisterId: Principal;
+        neuronIdHex: string;
+      }
+    | undefined;
   neuron: SnsNeuron | undefined;
 }
 
--- a/frontend/src/lib/types/sns-neuron-detail.context.ts
+++ b/frontend/src/lib/types/sns-neuron-detail.context.ts
@@ -3,7 +3,7 @@ import type { SnsNeuron } from "@dfinity/sns";
 import type { Writable } from "svelte/store";
 
 /**
- * A store that contains the selected proposal.
+ * A store that contains the selected sns neuron.
  */
 export interface SelectedSnsNeuronStore {
   rootCanisterId: Principal | undefined;
--- a/frontend/src/routes/SnsNeuronDetail.svelte
+++ b/frontend/src/routes/SnsNeuronDetail.svelte
@@ -67,8 +67,11 @@
     snsProjectSelectedStore.set(rootCanisterId);
 
     // `loadNeuron` relies on neuronId and rootCanisterId to be set in the store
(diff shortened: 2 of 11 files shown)


lmuntaner:

I haven't been able to test that all the flow works because the user does not have neurons in the SNS project. Therefore, I relied on tests to check that it "will work".

frontend/src/routes/SnsNeuronDetail.svelte
```diff
@@ -31,16 +66,10 @@
...
-      neuronIdHex: neuronIdMaybe,
-      onLoad: ({ neuron: snsNeuron }: { neuron: SnsNeuron }) => {
-        neuron = snsNeuron;
-      },
-      onError: () => {
-        console.error("Error loading neuron");
-      },
-    });
+    // `loadNeuron` relies on neuronId and rootCanisterId to be set in the store
+    $selectedSnsNeuronStore.neuronIdHex = neuronIdMaybe;
```
peterpeterparker, line 70: does that work? `$selectedSnsNeuronStore.neuronIdHex = ...` set the value?

you don't have to reset the `neuron` as well?

maybe I am missing something but to trigger a single change in the store and to reset the neuron as well, I would suggest to `set` the object - e.g.

```
selectedSnsNeuronStore.set({
     neuronIdHex: neuronIdMaybe,
     rootCanisterId,
     neuron: undefined
});
```

frontend/src/routes/SnsNeuronDetail.svelte
```diff
@@ -31,16 +66,10 @@
...
+    $selectedSnsNeuronStore.rootCanisterId = rootCanisterId;
+    loadNeuron();
```
peterpeterparker, line 72: does it helps the ui to not `await` for the promise here?

lmuntaner: no, no difference here.

frontend/src/routes/SnsNeuronDetail.svelte
lmuntaner: good point, I like it better.

frontend/src/lib/types/sns-neuron-detail.context.ts
```diff
@@ -0,0 +1,19 @@
+import type { Principal } from "@dfinity/principal";
+import type { SnsNeuron } from "@dfinity/sns";
+import type { Writable } from "svelte/store";
+
+/**
+ * A store that contains the selected proposal.
```
peterpeterparker, line 6: proposal...

frontend/src/routes/SnsNeuronDetail.svelte
lmuntaner: done

frontend/src/lib/types/sns-neuron-detail.context.ts
lmuntaner: done

frontend/src/lib/types/sns-neuron-detail.context.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export interface SelectedSnsNeuronStore {
```
peterpeterparker, line 8: maybe overenginereed, you tell me, but I ask my self if it would not be neat to have a sub-object since we need both `rootCanisterId` and `neuronIdHex` to fetch the info

something like:

```
export interface SelectedSnsNeuronStore {
     selected: {
         rootCanisterId: Principal;
         neuronIdHex: string;
     } | undefined,
     neuron: SnsNeuron | undefined;
}
```

P.S.: since a couple of times I also ask my self if it would be possible to also use derived store in these "context details pages" use cases.  like having a store for the params and the derived for the data that is selected or loaded. but not for this PR for sure or anything, just speaking out at loud, maybe someday we can have a look...

frontend/src/routes/SnsNeuronDetail.svelte
peterpeterparker: thanks for the feedback

peterpeterparker:
Generally speaking, what happens if the sns neuronId provided in the URL does not exists? The user is redirected with an error "neuron not found" or something?

Maybe already implemented, just ask because it is not displayed by the github diff view

lmuntaner:
@peterpeterparker good question.

I created an issue to work on that.

We should also manage if the root canister id doesn't exist as well: https://dfinity.atlassian.net/browse/L2-921

peterpeterparker:
Good to know, thanks for the issue 👍

frontend/src/lib/types/sns-neuron-detail.context.ts
lmuntaner: done

peterpeterparker (approved):
I might have a look to it later on when it will work out with the backend, I often understand better that pattern with real data and `console.log` (😉) but from code, looks good. Thanks 👍

PR #1230, opened 2022-08-09
L2 827 optimistic voting update

mstrasinskis:
# Motivation

Due to slow vote registration, the process was switched to optimistic UI update ([L2-827 - Proposal cast vote performance](https://dfinity.atlassian.net/browse/L2-827))

1. display voting message (yellow one) after vote registration confirmation click
2. disable current proposal action items 
  - "Adopt" and "Reject" buttons
  - disable neuron checkbox
3. update current neuron voting position: from "Cast Votes" to "My Votes" to reflect the progress (could be important for users with a huge amount of neurons)
4. after the proposal and neurons are fetched (after voting) the message is removed

# Changes

- voting using optimistic update
- show toast during update
- new toast level "running"
- expose toast ID to hide by it
- show 

# Tests

TBD

# Screenshots

https://user-images.githubusercontent.com/98811342/183778796-f52f28b0-61f1-42f5-aaad-0b28bb869ffc.mp4


frontend/src/lib/types/toast.ts
```diff
@@ -1,6 +1,6 @@
 import type { I18nSubstitutions } from "../utils/i18n.utils";
 
-export type ToastLevel = "success" | "warn" | "error";
+export type ToastLevel = "success" | "warn" | "error" | "running";
```
mstrasinskis, line 3: Name suggestions are welcome!

frontend/src/lib/components/proposal-detail/VotingCard/VotingNeuronSelect.svelte
```diff
@@ -25,53 +30,58 @@
     votingNeuronSelectStore.toggleSelection(neuronId);
 </script>
 
-<p class="headline">
-  <span>{$i18n.proposal_detail__vote.neurons}</span>
-  <span>{$i18n.proposal_detail__vote.voting_power}</span>
-</p>
-
-<ul>
-  {#each $votingNeuronSelectStore.neurons as neuron}
```
mstrasinskis, line 34: Just UI improvement to not show the semi empty UI-block between the moment when all the neurons are voted but neurons and proposal isn't yet loaded.

frontend/src/lib/components/ui/Toast.svelte
```diff
@@ -57,10 +58,14 @@
   role="dialog"
   class="toast"
   class:error={level === "error"}
-  class:warn={level === "warn"}
+  class:warn={level === "warn" || level === "running"}
```
peterpeterparker, line 61: nth: `["running", "warn"].includes(level)`

frontend/src/lib/services/proposals.services.ts
```diff
@@ -345,18 +348,69 @@ export const routePathProposalId = (
...
+      proposalId,
+    });
+    // update proposal vote state
+    votingProposal = updateProposalVote({
+      proposalInfo: votingProposal,
+      neuron: votingNeuron,
+      vote,
+    });
+
+    neuronsStore.replaceNeurons([votingNeuron]);
```
peterpeterparker, line 409: can you add a comment? I am guessing we update the stores here with an optimistic value right?

frontend/src/lib/services/proposals.services.ts
```diff
@@ -345,18 +348,69 @@ export const routePathProposalId = (
...
   neuronIds: NeuronId[];
-  proposalId: ProposalId;
+  proposalInfo: ProposalInfo;
   vote: Vote;
   reloadProposalCallback: (proposalInfo: ProposalInfo) => void;
 }): Promise<void> => {
-  startBusy({ initiator: "vote" });
-
   const identity: Identity = await getIdentity();
+  const proposalId = proposalInfo.id as bigint;
```
peterpeterparker, line 361: nth: refactor registering vote in progress in a dedicated function

frontend/src/lib/services/proposals.services.ts
```diff
@@ -345,18 +348,69 @@ export const routePathProposalId = (
...
+    vote,
+  };
+  const $definedNeuronsStore = get(definedNeuronsStore);
+
+  voteInProgressStore.add(voteInProgress);
+
+  let votingProposal: ProposalInfo = { ...proposalInfo };
+
+  const $i18n = get(i18n);
+  const toastMessage = toastsStore.show({
```
peterpeterparker, line 375: any particular reason why the toast is shown before the callback is registered and stores updated?

frontend/src/lib/services/proposals.services.ts
```diff
@@ -345,18 +348,69 @@ export const routePathProposalId = (
...
+    );
+
+    assertNonNullish(originalNeuron, `Neuron ${neuronId} not defined`);
+
+    voteInProgressStore.addSuccessfullyVotedNeuronIds({
+      proposalId,
+      successfullyVotedNeuronIds: [neuronId],
+    });
+
+    // pretend voting
```
peterpeterparker, line 396: can you improve these comments? I understand, but it's a bit a guess, that we mock the data and update the stores because with proceed with optimistic values, is that right?

frontend/src/lib/stores/toasts.store.ts
```diff
@@ -18,11 +18,14 @@ const initToastsStore = () => {
   return {
     subscribe,
 
-    show(msg: ToastMsg) {
+    show(msg: ToastMsg): symbol {
+      const id = Symbol();
```
peterpeterparker, line 22: should we add a test for this?

frontend/src/lib/stores/voting.store.ts
```diff
@@ -0,0 +1,78 @@
...
+  successfullyVotedNeuronIds: NeuronId[];
+  vote: Vote;
+}
+
+export interface VotingStore {
+  votes: VoteInProgress[];
+}
+
+/**
+ * A store that contain votes in progress data (proposals and neurons that were not confirmed by `update` calls)
```
peterpeterparker, line 17: maybe also improve comment here to explain why we need it?

frontend/src/lib/stores/voting.store.ts
```diff
@@ -0,0 +1,78 @@
...
+    votes: [],
+  });
+
+  return {
+    subscribe,
+
+    add(vote: VoteInProgress) {
+      update(({ votes }) => {
+        if (votes.find(({ proposalId }) => proposalId === vote.proposalId)) {
+          throw new Error("Simultaneous proposal voting");
```
peterpeterparker, line 30: i18n

peterpeterparker: did you test it, the error bubbled till the ui?

frontend/src/lib/stores/voting.store.ts
```diff
@@ -0,0 +1,78 @@
...
+                  ...successfullyVotedNeuronIds,
+                ])
+              ),
+            },
+          ],
+        };
+      });
+    },
+
+    remove(proposalId: ProposalId) {
```
peterpeterparker, line 70: we don't know if a `VoteInProgress` is still in progress or fully executed - is it ok to `remove` such an entity that has not been fully completed?

frontend/src/lib/stores/neurons.store.ts
```diff
@@ -50,6 +50,32 @@ const initNeuronsStore = () => {
...
+      update(({ neurons: oldNeurons, certified }: NeuronsStore) => {
+        const newNeurons = new Map(
+          neurons.map((neuron) => [neuron.neuronId, neuron])
+        );
+        const updatedNeurons = oldNeurons?.map((old) => {
+          const { neuronId } = old;
+          const newNeuron = newNeurons.get(neuronId);
+
+          if (newNeuron) {
+            newNeurons.delete(neuronId);
```
peterpeterparker, line 64: ummm not a big fan of that pattern with a `delete` within a `.map`, not particular reason, just a bit the mix of functional and imperative code feels a bit like anti-pattern.

why not just mapping the ids e.g.?

```
replaceNeurons(neurons: NeuronInfo[]) {
      update(({ neurons: oldNeurons, certified }: NeuronsStore) => {
          const newNeuronsIds = neurons.map({neuronId} => neuronId)
          return {
              neurons: [
                   ...oldNeurons.filter({neuronId} => !newNeuronIds.includes(neuronId)),
                   ..neurons.filter({neuronId} => newNeuronIds.includes(neuronId)),
```

frontend/src/lib/services/proposals.services.ts
```diff
@@ -345,18 +348,69 @@ export const routePathProposalId = (
...
+      $proposalTitle: proposalInfo.proposal?.title ?? "",
+      $proposalId: `${proposalId}`,
+    },
+  });
+  const registerVoteCallback = (neuronId: NeuronId) => {
+    const originalNeuron = $definedNeuronsStore.find(
+      ({ neuronId: id }) => id === neuronId
+    );
+
+    assertNonNullish(originalNeuron, `Neuron ${neuronId} not defined`);
```
peterpeterparker, line 389: i18n

peterpeterparker:
should any features that can change the value of existing neurons be disabled while votes are in progress?

frontend/src/lib/components/ui/Toast.svelte
mstrasinskis: Fancy look vs performance, updated 👍

peterpeterparker: 😁

frontend/src/lib/stores/voting.store.ts
mstrasinskis: It's not intended for the user. The current UI prevents this by disabling the buttons. But because it's pretty critical, the error was added to help future us with debugging.

frontend/src/lib/stores/neurons.store.ts
mstrasinskis: It was done like this before (your suggestion). The new requirement is to preserve the neuron order within the store.

frontend/src/lib/services/proposals.services.ts
mstrasinskis: > should any features that can change the value of existing neurons be disabled while votes are in progress?

Good point, I'll disable some buttons on the neuron page.

frontend/src/lib/services/proposals.services.ts
mstrasinskis: Actually, I expect the neurons to be already in place. This is more like 'just in case' assertion. I'll add a TODO to remove it as soon as the feature is tested on the `mainnet`.

frontend/src/lib/stores/voting.store.ts
mstrasinskis: Add a comment plz!

frontend/src/lib/services/proposals.services.ts
mstrasinskis: Was moved down

frontend/src/lib/stores/toasts.store.ts
mstrasinskis: Done, thanks.

frontend/src/lib/stores/neurons.store.ts
peterpeterparker: I see, thanks for the feedback. Code wise, I have to say I don't like that requirements then 😉. But makes sense.

Can you add a comment about it in the code please?

frontend/src/lib/i18n/en.json
```diff
@@ -43,7 +43,7 @@
     "rename_subaccount_type": "The account provided is not a linked account. Only linked account can be renamed.",
     "get_neurons": "There was an unexpected issue while searching for the neurons.",
     "get_known_neurons": "There was an unexpected issue while searching for the known neurons.",
-    "register_vote": "Sorry, there was an error while registering the vote. Please try again.",
+    "register_vote": "Sorry, there was an error while registering the vote (Proposal: $proposalId). Please try again.",
```
peterpeterparker, line 46: Maybe: "Sorry, there was an error while registering the vote for the proposal $proposalId"..."

Or "Sorry...your vote for the proposal $proposalId."

not sure if it's "for the proposal" or "on the proposal" though

frontend/src/lib/i18n/en.json
```diff
@@ -426,7 +426,8 @@
     "vote_status": "Neuron $neuronId has voted $vote",
     "accept_or_reject": "Do you want to adopt or reject Proposal <strong>$id</strong> &ndash; <strong>$title</strong> for <strong>$topic</strong>?",
     "cast_vote_neuronId": "Neuron ID $neuronId",
-    "cast_vote_votingPower": "Voting power of $votingPower"
+    "cast_vote_votingPower": "Voting power of $votingPower",
+    "voting_in_progress_message": "Registering the vote ($vote) for $proposalTitle ($proposalId). Please do not leave the nns-dapp until the registration is complete."
```
peterpeterparker, line 430: - instead of the proposal title, i suggest to use the proposal `Topic`. Double check with Mischa but knowing upcoming UI changes, it makes sense I think

- Not sure we should reference "nns-dapp" because actually, it seems that so far we never do. Even in the meta data.

- the vote or your vote?

- should we display "Registering your vote (yes / no)..." or actually display "Accepting proposal..." or "Rejecting proposal ..."?

So all in all, maybe:

Accepting proposal $topic ($proposalId). Please do not close the dapp until completed.

Only suggestion, if interesting maybe worth to check with Mischa

mstrasinskis: Adopting proposal $topic ($proposalId) | Rejecting proposal $topic ($proposalId)

frontend/src/lib/services/proposals.services.ts
```diff
@@ -377,104 +437,61 @@ export const registerVotes = async ({
...
-  stopBusy("vote");
+  Promise.all([reloadListNeurons(), reloadProposal()]).finally(() => {
+    // remove in progress state update call
+    voteInProgressStore.remove(voteInProgress.proposalId);
+    toastsStore.hide(toastMessage);
+  });
 };
 
-const requestRegisterVotes = async ({
+export const requestRegisterVotes = async ({
```
peterpeterparker, line 469: since we now expose the function, is it worth adding a test too? or did you?

frontend/src/lib/services/proposals.services.ts
```diff
@@ -341,31 +343,89 @@ export const routePathProposalId = (
 
 /**
  * Makes multiple registerVote calls (1 per neuronId).
- * @returns List of errors (order is preserved)
+ *
+ * In order to improve UX optimistic UI update is used: after every successful neuron vote registration (`registerVote`) we mock the data (both proposal and voted neuron) and update the stores with optimistic values.
  */
 export const registerVotes = async ({
```
peterpeterparker, line 349: can we add some test for `registerVotes` that checks that the `voteInProgressStore` new features are used?

frontend/src/lib/i18n/en.json
mstrasinskis: Just to use the same pattern as in 'in-progress' message: Sorry, there was an error while registering the vote for the proposal $topic $proposalId. Please try again. ?

peterpeterparker: Sounds good 👍

peterpeterparker (approved):
LGTM, thanks!!!

frontend/src/lib/services/proposals.services.ts
mstrasinskis: Good point. Will be done together with new utils in a separate PR.

PR #1231, opened 2022-08-10
build: cleanup, simplify config and introduce feature flags object

peterpeterparker:
# Motivation

- lots of configuration options
- too much complexity with bash variables flying around to assign config variables when only use in the frontend
- feature flags object instead of x variables

# Approach

Do NOT use a bash variable if no bash variable is needed.

# Changes

- `env.config.mjs` does not care anymore about `process.env` as these variables are processed in advance to generate `deployment-configuration.json`
- remove `ENABLE_NEW_SPAWN_FEATURE` and `ENABLE_SNS_NEURONS` from config scripts and DO NOT replace it with bash variables
- refactor `dfx.json` to group above variables to a new object `FEATURE_FLAGS`
- parse `FEATURE_FLAGS` in dapp
- set `FEATURE_FLAGS` in jest test
- set `ENABLE_SNS_NEURONS` to `false` for local configuration (this was previously done in the dapp with a `if`)

# Notes

Not in this PR but:

- Afterwards `ENABLE_NEW_SPAWN_FEATURE` can be removed, feature is in prod.
- Ultimately I think we can even get rid of `env.config.mjs`, that would even make the pipeline easier to follow


dfx.json
```diff
@@ -77,110 +67,112 @@
...
     "staging": {
       "config": {
         "BUILD": "dev",
         "FETCH_ROOT_KEY": true,
         "REDIRECT_TO_LEGACY": "prod",
-        "ENABLE_NEW_SPAWN_FEATURE": true,
-        "ENABLE_SNS_NEURONS": false,
-        "HOST": "https://nnsdapp.dfinity.network"
+        "HOST": "https://nnsdapp.dfinity.network",
+        "FEATURE_FLAGS": {
```
lmuntaner, line 88: Don't you need them in the config.sh?

frontend/env.config.mjs
```diff
@@ -52,50 +29,11 @@ const ENVIRONMENT = ROLLUP_WATCH
 // agent-js connects per default to mainnet with the anonymous identity
 const MAINNET = `https://ic0.app`;
 
-// The host that nns-js connects to
-const HOST = getRequiredEnvVar("HOST");
-
-// Canister Ids for testnet and mainnet
-const GOVERNANCE_CANISTER_ID = getRequiredEnvVar("GOVERNANCE_CANISTER_ID");
```
lmuntaner, line 59: Does this mean that the build won't fail if a required env var is not present?

dfx.json
lmuntaner: Also in the build.sh?

lmuntaner (commented):
A few comments, but looks good, thanks!

dfx.json
peterpeterparker: No. `config.sh` parse every keys defined for an environment in `dfx.json` to `deployment-config.json`.

That's why notably no bash variables are needed.

frontend/env.config.mjs
peterpeterparker: This was actually useless, it was like reading a bash variables twice.

One time to generate `deployment-config.json` and then reading it again here and comparing the value with `deployment-config.json`.

If a variable is not set it will still fails because `config.sh` expect these.

Capture d’écran 2022-08-10 à 09 44 15
Capture d’écran 2022-08-10 à 09 45 21

dfx.json
lmuntaner: thanks for the info!

bitdivine (approved):
Looks good, thanks!  happy that this is replaced by something that is more focused on how you use this.

One comment - I think one of the assignments should be unconditional.  Otherwise all fine.

config.sh
```diff
@@ -37,12 +37,14 @@ local_deployment_data="$(
...
   : "- may be deployed locally"
-  IDENTITY_SERVICE_URL="$(
+  : "- may be set as an env var"
+  LOCALLY_DEPLOYED_IDENTITY_SERVICE_URL="$(
     set -euo pipefail
     id="$(dfx canister --network "$DFX_NETWORK" id internet_identity 2>/dev/null || true)"
     : "If we have a canister ID, insert it into HOST as a subdomain."
     test -z "${id:-}" || { jq -re '.networks[env.DFX_NETWORK].config.HOST' dfx.json | sed -E "s,^(https?://)?,&${id}.,g"; }
   )"
+  test -n "${IDENTITY_SERVICE_URL:-}" || IDENTITY_SERVICE_URL="$LOCALLY_DEPLOYED_IDENTITY_SERVICE_URL"
```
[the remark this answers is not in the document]
peterpeterparker: Use case: I cannot start my local dev environment without having to set the `IDENTITY_SERVICE_URL` in `dfx.json`.

But generally speaking, I think at some point we will have anyway to have a look at the configuration because for the community I am not sure it is that trivial to start the dapp. e.g. dfx is a must to start the frontend while irl you don't need it.

I'll revert this from my PR for the time being and let's discuss afterwards, the feature flag is blocking my other PR. I go for a run, do that afterwards and we'll ping you when done.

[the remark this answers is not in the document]
peterpeterparker: Done. Reverted

PR #1222, opened 2022-08-09
feat: integrate first gix components

peterpeterparker:
# Motivation

We aim to move the UI and style to [gix-components](https://github.com/dfinity/gix-components). This PR introduces the usage of the library and of the first components.

# Changes

- add `gix-components` as a dependency 
- transpile `gix-components` for test
- introduce usage of `<Layout />` component and adapt slot and properties accordingly
- remove components that have been moved to `gix-components`
- use mixins and icons available in `gix-components`
- remove `.ttf` and replace with new `.woff2` fonts
- deprecate `<CardGrid />` for `.grid`

# Notes

- this PR does not introduces any logical changes
- `routes` were modified because `<main />` tag is not part of the `<Layout />`

PR #1235, opened 2022-08-10
feat: voting list with new grid UI 

peterpeterparker:
# Motivation

Voting list UI based on the new grid system - new design.

# Related PRs

- [x] https://github.com/dfinity/nns-dapp/pull/1231 👉 needed to add a feature flag
- [x] https://github.com/dfinity/nns-dapp/pull/1248 👉 add feature flag
- [x] https://github.com/dfinity/gix-components/pull/30 👉 needed to apply `grid` style to infinite scroll
- [x] https://github.com/dfinity/nns-dapp/pull/1222 👉 per extension to use above PR

# Note

This PR does not contain any logic or display changes - same behavior, same info displayed.

# Changes

- use new feature flag `VOTING_UI`
- remove and replace cmp `<InfiniteScroll />` and `.grid` style with `gix-components`
- move current layout to a new cmp `<ProposalsLegacy />`
- implemented new grid layout in a new cmp `<ProposalsModern />` 
- in proposal route, display either legacy or modern UI
- bump `gix-components`

# Screenshots

Capture d’écran 2022-08-16 à 10 21 00

Capture d’écran 2022-08-16 à 10 20 53


frontend/src/lib/components/proposals/ProposalsModern.svelte
```diff
@@ -0,0 +1,35 @@
...
+  export let neuronsLoaded: boolean;
+  export let loading: boolean;
+  export let nothingFound: boolean;
+  export let hidden: boolean;
+</script>
+
+<ProposalsFilters />
+
+{#if neuronsLoaded}
+  <InfiniteScroll on:nnsIntersect layout="grid">
```
lmuntaner, line 18: Not sure I like that the InfiniteScroll knows about how to render the items.

Wouldn't it make more sense that the InfiniteScroll exposes a listener on:nnsIntersect as done here and then the child knows whether it shuold look like a grid or a list?

lmuntaner (approved):
LGTM apart from nne comment, but not sure it should be addressed in this PR.

frontend/src/lib/components/proposals/ProposalsModern.svelte
peterpeterparker: The InfiniteScroll needs a [container](https://github.com/dfinity/gix-components/blob/77ca201f3d5566cf4d9c526b0a910d3ccf2340f0/src/lib/components/InfiniteScroll.svelte#L98) and items need to be direct children. One option was to extract the container outside the component but then, it would have mean that every consumers of the InfiniteScroll would have to bind and path the ref. So for usage simplicity reason we decided to have the container within the component. Per extension, because the container is in the component, we have to pass the style because the container is the grid or list.

lmuntaner: I just would have liked the other way for the Single Responsibility Principle. But it's ok.

PR #1236, opened 2022-08-10
docs: hacking and cleanup readme

peterpeterparker:
# Motivation

Add a `HACKING.md` file with some info about the various testnets and light clean up of `README.md`


HACKING.md
```diff
@@ -0,0 +1,26 @@
+# Hacking
+
+This document list a couple of useful information to develop NNS-dapp.
+
+## Environments
+
+Testnets [canister_ids.json](https://github.com/dfinity/nns-dapp/blob/testnets/testnets/canister_ids.json) provides an overview of the canister IDs that should be currently deployed and available on various test environments.
```
lmuntaner, line 7: I would say that this should be copied from here to the local file to be used for local development.

lmuntaner (approved):
Thanks, just one comment!

HACKING.md
peterpeterparker: Mmmh I don't get it but, really not against changing it. Can you rephrase the sentence as you wish?

lmuntaner: Maybe it should be added below, see my comment there.

HACKING.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Configure an environment
+
+If you wish to configure an environment for local development purpose, you can proceed as following:
```
lmuntaner, line 11: If you wish to work against a testnet, there are two options:

### Script tag

- collect...
- add a new...

### Copy Canister Ids

- Copy the remote canister_ids.json to your local version.
- Change the `DFX_NETWORK=<testnet_name>` in the `dev` script.

Example:
```
"dev": "... DFX_NETWORK=small11 npm run build:config...",
```

lmuntaner (commented):
Maybe something like this?

HACKING.md
peterpeterparker: Discussed and resolved, thanks for the inputs 👍

HACKING.md
peterpeterparker: Discussed and resolved, thanks for the inputs 👍

PR #1242, opened 2022-08-11
Fix: Force sns neuron fetch after hotkey mgt

lmuntaner:
# Motivation

If the user arrived to the sns neuron detail page from the neurons page, the neuron was loaded from the store. Reloading meant to get the same neuron from the store.

That meant that after updating the neuron, we still got the old one.

# Changes

* Add a "forceFetch" flag to the getSnsNeuron service that forces call to the api.
* Use the new flag after adding or removing hotkeys in sns neurons.

# Tests

* Change tests to check that the flag is properly set after adding or removing hotkeys.
* Check that the flag calls api instead of retrieving from the store.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
@@ -61,7 +61,7 @@
       rootCanisterId: $snsProjectSelectedStore,
     });
     if (success) {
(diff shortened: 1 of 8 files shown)


lmuntaner:

lmuntaner:
Yes, correct.

Whereas in the neuron details everything happens in the services and it's all the same store.

To be honest, I found it easier to understand and manage that scenario. Yet, we have the problem of not knowing when to show loading in the list view.

PR #1245, opened 2022-08-12
L2-817: Pending Sale Participation Canister Check

lmuntaner:
# Motivation

Manage edge case in case the client loses connection when participating in a decentralized sale.

# Changes

* Notify NNS Dapp canister that a participation will be done.
* New state in NNS Dapp canister: pending_transactions
* NNS Dapp new method add_pending_notify_swap that adds to pending_transactions
* New type of transactions to be checked by runner: ParticipateSwap.

# Tests

* Tests for the new accounts_store functionality.
* Add testcase for sns api function to participate in swap.

### DIFF
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -447,7 +447,6 @@ export const participateInSnsSwap = async ({
     swap_canister_id: swapCanisterId,
     buyer: controller,
   });
-  console.log("after da add pending notify swap");
 
   // Send amount to the ledger
   await nnsLedger.transfer({
@@ -458,7 +457,7 @@ export const participateInSnsSwap = async ({
   });
 
   // Notify participation
-  // await notifyParticipation({ buyer: controller.toText() });
+  await notifyParticipation({ buyer: controller.toText() });
 
   logWithTimestamp("Participating in swap: done");
 };
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -10,7 +10,6 @@ import type {
   SnsWrapper,
 } from "@dfinity/sns";
 import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
-import { SWAP_PARTICIPATE_MEMO } from "../constants/api.constants";
 import { HOST, WASM_CANISTER_ID } from "../constants/environment.constants";
 import {
   importInitSnsWrapper,
@@ -453,7 +452,6 @@ export const participateInSnsSwap = async ({
     amount,
     fromSubAccount,
     to: accountIdentifier,
-    memo: SWAP_PARTICIPATE_MEMO,
   });
 
   // Notify participation
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -9,6 +9,7 @@ import type {
   SnsSwapBuyerState,
   SnsWrapper,
 } from "@dfinity/sns";
+import { toNullable } from "@dfinity/utils";
 import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
 import { HOST, WASM_CANISTER_ID } from "../constants/environment.constants";
 import {
@@ -445,6 +446,7 @@ export const participateInSnsSwap = async ({
   await nnsDapp.addPendingNotifySwap({
     swap_canister_id: swapCanisterId,
     buyer: controller,
+    buyer_sub_account: toNullable(fromSubAccount),
   });
 
   // Send amount to the ledger
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -441,7 +441,7 @@ export const participateInSnsSwap = async ({
     controller,
   });
 
-  // Notify NNS Dapp to notify when transaction is checked
+  // If the client disconnects after the tranfer, the participation will still be notified.
   const { canister: nnsDapp } = await nnsDappCanister({ identity });
   await nnsDapp.addPendingNotifySwap({
     swap_canister_id: swapCanisterId,
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -10,6 +10,7 @@ import type {
   SnsWrapper,
 } from "@dfinity/sns";
 import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
+import { SWAP_PARTICIPATE_MEMO } from "../constants/api.constants";
 import { HOST, WASM_CANISTER_ID } from "../constants/environment.constants";
 import {
   importInitSnsWrapper,
@@ -28,6 +29,7 @@ import { createAgent } from "../utils/agent.utils";
 import { logWithTimestamp } from "../utils/dev.utils";
 import { getSwapCanisterAccount } from "../utils/sns.utils";
 import { ledgerCanister } from "./ledger.api";
+import { nnsDappCanister } from "./nns-dapp.api";
 
 let snsQueryWrappers: Promise<Map<QueryRootCanisterId, SnsWrapper>> | undefined;
 let snsUpdateWrappers:
@@ -439,15 +441,24 @@ export const participateInSnsSwap = async ({
     controller,
   });
 
+  // Notify NNS Dapp to notify when transaction is checked
+  const { canister: nnsDapp } = await nnsDappCanister({ identity });
+  await nnsDapp.addPendingNotifySwap({
+    swap_canister_id: swapCanisterId,
+    buyer: controller,
+  });
+  console.log("after da add pending notify swap");
+
   // Send amount to the ledger
   await nnsLedger.transfer({
     amount,
     fromSubAccount,
     to: accountIdentifier,
+    memo: SWAP_PARTICIPATE_MEMO,
   });
 
   // Notify participation
-  await notifyParticipation({ buyer: controller.toText() });
+  // await notifyParticipation({ buyer: controller.toText() });
 
   logWithTimestamp("Participating in swap: done");
 };
--- a/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
+++ b/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
@@ -14,7 +14,6 @@ import {
   CanisterNotFoundError,
   HardwareWalletAttachError,
   NameTooLongError,
-  PendingTransactionsLimitExceededError,
   ProposalPayloadNotFoundError,
   ProposalPayloadTooLargeError,
   SubAccountLimitExceededError,
@@ -346,8 +345,13 @@ export class NNSDappCanister {
     if ("Ok" in response) {
       return;
     }
-    if ("TransactionLimitExceeded" in response) {
-      throw new PendingTransactionsLimitExceededError();
+    // Edge case. `add_pending_notify_swap` uses a supported transaction
+    if ("TransactionTypeNotSupported" in response) {
+      throw new Error(
+        `Transaction type to be added not supported: ${JSON.stringify(
+          response
+        )}`
+      );
     }
     // Edge case
     throw new Error(
--- a/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
+++ b/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
@@ -14,6 +14,7 @@ import {
   CanisterNotFoundError,
   HardwareWalletAttachError,
   NameTooLongError,
+  PendingTransactionsLimitExceededError,
   ProposalPayloadNotFoundError,
   ProposalPayloadTooLargeError,
   SubAccountLimitExceededError,
@@ -24,6 +25,7 @@ import { idlFactory } from "./nns-dapp.idl";
 import type {
   AccountDetails,
   AccountIdentifierString,
+  AddPendingNotifySwapRequest,
   CanisterDetails,
   CreateSubAccountResponse,
   GetTransactionsResponse,
@@ -92,7 +94,7 @@ export class NNSDappCanister {
       throw new AccountNotFoundError("error__account.not_found");
     }
 
-    if (Ok) {
+    if (Ok !== undefined) {
       return Ok;
     }
 
@@ -134,7 +136,7 @@ export class NNSDappCanister {
       );
     }
 
-    if (Ok) {
+    if (Ok !== undefined) {
       return Ok;
     }
 
@@ -332,6 +334,24 @@ export class NNSDappCanister {
 
     throw new UnknownProposalPayloadError(
       errorText ?? (nonNullish(response) ? JSON.stringify(response) : undefined)
+    );
+  }
+
+  public async addPendingNotifySwap(
+    request: AddPendingNotifySwapRequest
+  ): Promise<void> {
+    const response = await this.certifiedService.add_pending_notify_swap(
+      request
+    );
+    if ("Ok" in response) {
+      return;
+    }
+    if ("TransactionLimitExceeded" in response) {
+      throw new PendingTransactionsLimitExceededError();
+    }
+    // Edge case
+    throw new Error(
+      `Unkonwn response for add_pending_notify_swap ${JSON.stringify(response)}`
     );
   }
 }
--- a/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
+++ b/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
@@ -355,7 +355,9 @@ export class NNSDappCanister {
     }
     // Edge case
     throw new Error(
-      `Unkonwn response for add_pending_notify_swap ${JSON.stringify(response)}`
+      `Unknonwn response for add_pending_notify_swap ${JSON.stringify(
+        response
+      )}`
     );
   }
 }
--- a/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
+++ b/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
@@ -340,14 +340,6 @@ export class NNSDappCanister {
     if ("Ok" in response) {
       return;
     }
-    // Edge case. `add_pending_notify_swap` uses a supported transaction
-    if ("TransactionTypeNotSupported" in response) {
-      throw new Error(
-        `Transaction type to be added not supported: ${JSON.stringify(
-          response
-        )}`
-      );
-    }
     // Edge case
     throw new Error(
       `Unknown response for add_pending_notify_swap ${JSON.stringify(response)}`
--- a/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
+++ b/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
@@ -27,6 +27,7 @@ import type {
   AddPendingNotifySwapRequest,
   CanisterDetails,
   CreateSubAccountResponse,
+  GetAccountResponse,
   GetTransactionsResponse,
   RegisterHardwareWalletRequest,
   RegisterHardwareWalletResponse,
@@ -86,15 +87,15 @@ export class NNSDappCanister {
   }: {
     certified: boolean;
   }): Promise<AccountDetails> {
-    const { AccountNotFound, Ok } = await this.getNNSDappService(
+    const response: GetAccountResponse = await this.getNNSDappService(
       certified
     ).get_account();
-    if (AccountNotFound === null) {
+    if ("AccountNotFound" in response) {
       throw new AccountNotFoundError("error__account.not_found");
     }
 
-    if (Ok !== undefined) {
-      return Ok;
+    if ("Ok" in response) {
+      return response.Ok;
     }
 
     // We should never reach here. Some of the previous properties should be present.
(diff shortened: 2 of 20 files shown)


rs/src/periodic_tasks_runner.rs
```diff
@@ -56,6 +59,48 @@ pub async fn run_periodic_tasks() {
...
+            .get_pending_transaction(to) {
+                Some(pending_transaction) => {
+                    match pending_transaction.pending_transaction_type {
+                        PendingTransactionType::ParticipateSwap(canister_id) => {
+                            participate_swap(canister_id, pending_transaction, block_height)
+                        }
+                    }
+                }
+                None => {
+                    Ok(true)
```
lmuntaner, line 78: @hpeebles how can I make that this part returns the same as above `participate_swap`?

rs/src/accounts_store.rs
```diff
@@ -445,6 +467,72 @@ impl AccountsStore {
...
 
+    pub fn add_pending_transaction(
+        &mut self,
+        principal: PrincipalId,
+        transaction_type: TransactionType,
+    ) -> AddPendingTransactionResponse {
+        match transaction_type {
+            TransactionType::ParticipateSwap(canister_id) => {
+                // Why the &principal.into?
+                let account_identifier = AccountIdentifier::new(canister_id.get(), Some((&principal).into()));
```
lmuntaner, line 478: @hpeebles  What does `principal.into()` do? into what? Sorry, but my editor is not very good at following types and I can't understand this part. But I saw it in other places.

rs/src/accounts_store.rs
```diff
@@ -445,6 +467,72 @@ impl AccountsStore {
...
+    ) {
+        self.multi_part_transactions_processor
+            .update_status(block_height, MultiPartTransactionStatus::Complete);
+        match self.pending_transactions.get_mut(&to) {
+            Some(transactions) => {
+                let filtered_transactions = transactions
+                    .into_iter()
+                    .filter(|t| !t.transaction_completed)
+                    .filter(|t| t.principal == principal)
+                    .filter(|t| matches!(t.transaction_type, transaction_type));
```
lmuntaner, line 527: @hpeebles is this a good way to compare two enums? Does it also compare the value inside the enum? I found the solution in Stackoverflow.

rs/src/accounts_store.rs
```diff
@@ -1282,6 +1402,7 @@ impl StableState for AccountsStore {
         Candid((
             &self.accounts,
             &self.hardware_wallets_and_sub_accounts,
+            &self.pending_transactions,
```
lmuntaner, line 1405: @hpeebles @bitdivine 

What do we need to do to support rollback? If we add this feature, there is a "PendingTransaction" done, how can then we do a rollback?

Will it ignore that part of the state if we go back to a version without this state?

rs/src/canisters/swap.rs
```diff
@@ -0,0 +1,17 @@
...
+use dfn_core::CanisterId;
+use ic_sns_swap::pb::v1::RefreshBuyerTokensRequest;
+
+pub async fn notify_swap_participation(
+    canister_id: CanisterId,
+    request: RefreshBuyerTokensRequest,
+    // TODO: Remove Cycles and NotifyError types
+    // Resposne is `record {}`
+    // https://github.com/dfinity/ic/blob/master/rs/sns/swap/canister/swap.did#L92
+) -> Result<Result<Cycles, NotifyError>, String> {
```
lmuntaner, line 13: @hpeebles I don't understand very well the types of `dfn_core::call`.

The `refresh_buyer_tokens` method returns an empty record. How can I represent that here? Do I need to map the error `map_err` still?

rs/src/accounts_store.rs
```diff
@@ -445,6 +467,72 @@ impl AccountsStore {
...
+                .into_iter()
+                .find(|t| !t.transaction_completed && t.principal == principal),
+            None => None,
+        }
+    }
+
+    pub fn complete_pending_transaction(
+        &mut self,
+        to: AccountIdentifier,
+        transaction_type: TransactionType,
```
lmuntaner, line 515: Compiler throws a warning that this variable is not use, but I use it below.

```
warning: unused variable: `transaction_type`
   --> rs/src/accounts_store.rs:521:62
    |
521 |                     .filter(|t| matches!(t.transaction_type, transaction_type));
    |                                                              ^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_transaction_type`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `transaction_type`
   --> rs/src/accounts_store.rs:509:9
    |
509 |         transaction_type: TransactionType,
    |         ^^^^^^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_transaction_type`

warning: `nns-dapp` (bin "nns-dapp" test) generated 2 warnings
```

Any idea what is the problem?

frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
```diff
@@ -334,4 +335,27 @@ export class NNSDappCanister {
...
+    if ("TransactionTypeNotSupported" in response) {
+      throw new Error(
+        `Transaction type to be added not supported: ${JSON.stringify(
+          response
+        )}`
+      );
+    }
+    // Edge case
+    throw new Error(
+      `Unkonwn response for add_pending_notify_swap ${JSON.stringify(response)}`
```
hpeebles, line 358: Typo: 'Unknown'

frontend/src/lib/api/sns.api.ts
```diff
@@ -439,11 +441,19 @@ export const participateInSnsSwap = async ({
     controller,
   });
 
+  // Notify NNS Dapp to notify when transaction is checked
```
hpeebles, line 444: I think it's worth putting a comment here explaining that this step is so that if the frontend disconnects, the NNS Dapp can still complete the process for the user.

rs/src/accounts_store.rs
hpeebles: `AccountIdentifier::new(...)` takes a principal as the first arg then a `Option<Subaccount>` as the second arg.
`principal.into()` converts the principal into a `Subaccount`.

rs/src/accounts_store.rs
```diff
@@ -445,6 +467,72 @@ impl AccountsStore {
...
+    pub fn add_pending_transaction(
+        &mut self,
+        principal: PrincipalId,
+        transaction_type: TransactionType,
+    ) -> AddPendingTransactionResponse {
+        match transaction_type {
+            TransactionType::ParticipateSwap(canister_id) => {
+                // Why the &principal.into?
+                let account_identifier = AccountIdentifier::new(canister_id.get(), Some((&principal).into()));
+                match self.pending_transactions.get_mut(&account_identifier) {
```
hpeebles, line 479: You should use the `entry` API.
`self.pending_transactions.entry(account_identifier).or_default().push(PendingTransaction...);`

rs/src/canisters/swap.rs
hpeebles: `Result<(), String>`

hpeebles: Yeah you still need to map the error

rs/src/periodic_tasks_runner.rs
```diff
@@ -56,6 +60,30 @@ pub async fn run_periodic_tasks() {
...
 }
 
+async fn handle_participate_swap(
+    block_height: BlockHeight,
+    principal: PrincipalId,
+    to: AccountIdentifier,
+    swap_canister_id: CanisterId,
+) {
+    let request = RefreshBuyerTokensRequest {
+        buyer: principal.to_string(),
```
hpeebles, line 70: The `to_string` doesn't seem right. Surely we should be sending the principal as is?

frontend/src/lib/api/sns.api.ts
lmuntaner: good point, I'll do it

frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
lmuntaner: thanks!

rs/src/canisters/swap.rs
lmuntaner: thanks!

rs/src/accounts_store.rs
lmuntaner: Is it safer or just more readable?

rs/src/periodic_tasks_runner.rs
lmuntaner: Nops...

https://github.com/dfinity/ic/blob/master/rs/sns/swap/canister/swap.did#L68

frontend/src/lib/api/sns.api.ts
```diff
@@ -10,6 +10,7 @@ import type {
   SnsWrapper,
 } from "@dfinity/sns";
 import type { SubAccountArray } from "../canisters/nns-dapp/nns-dapp.types";
+import { SWAP_PARTICIPATE_MEMO } from "../constants/api.constants";
```
hpeebles, line 13: We no longer need to set the memo so this can be removed.
I don't think there's any advantage to leaving it in.

frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
```diff
@@ -334,4 +335,29 @@ export class NNSDappCanister {
...
+    if ("TransactionTypeNotSupported" in response) {
+      throw new Error(
+        `Transaction type to be added not supported: ${JSON.stringify(
+          response
+        )}`
+      );
+    }
+    // Edge case
+    throw new Error(
+      `Unknonwn response for add_pending_notify_swap ${JSON.stringify(
```
hpeebles, line 358: Typo... it should be 'Unknown'

frontend/src/lib/canisters/nns-dapp/nns-dapp.certified.idl.js
```diff
@@ -93,10 +101,10 @@ export const idlFactory = ({ IDL }) => {
   const TransactionType = IDL.Variant({
     Burn: IDL.Null,
     Mint: IDL.Null,
-    Transfer: IDL.Null,
     StakeNeuronNotification: IDL.Null,
     TopUpCanister: CanisterId,
     CreateCanister: IDL.Null,
+    Transfer: IDL.Null,
```
hpeebles, line 107: This is missing the new transaction type.

rs/src/accounts_store.rs
```diff
@@ -445,6 +467,59 @@ impl AccountsStore {
...
+        self.multi_part_transactions_processor
+            .update_status(block_height, MultiPartTransactionStatus::Complete);
+        if let Some(transactions) = self.pending_transactions.get_mut(&to) {
+            let filtered_transactions = transactions
+                .iter_mut()
+                .filter(|t| !t.transaction_completed)
+                .filter(|t| t.principal.to_string() == principal.to_string())
+                .filter(|t| matches!(t.transaction_type, transaction_type));
+            for transaction in filtered_transactions {
+                transaction.transaction_completed = true
```
hpeebles, line 518: Is there any reason to keep these around once they are complete?
Shouldn't we simply remove them?
Otherwise they'll build up unnecessarily over time.

rs/src/accounts_store.rs
```diff
@@ -1078,6 +1155,19 @@ impl AccountsStore {
         }
     }
 
+    fn is_participate_swap_transaction(&self, to: AccountIdentifier, principal: PrincipalId) -> Option<CanisterId> {
+        match self.get_pending_transaction(to, principal) {
```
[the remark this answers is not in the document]
lmuntaner: Do you mean before all the conditional statements of `get_transaction_type`?

```
match self.get_pending_transaction(to, principal) {
    Some(pending_transaction) => {
                if let TransactionType::ParticipateSwap(canister_id) = pending_transaction.transaction_type {
                    TransactionType::ParticipateSwap(canister_id)
                } else {
                    // ALL THE IF STATMENTS?
                }
            }
```

rs/src/accounts_store.rs
```diff
@@ -445,6 +467,59 @@ impl AccountsStore {
...
+        block_height: BlockHeight,
+    ) {
+        self.multi_part_transactions_processor
+            .update_status(block_height, MultiPartTransactionStatus::Complete);
+        if let Some(transactions) = self.pending_transactions.get_mut(&to) {
+            let filtered_transactions = transactions
+                .iter_mut()
+                .filter(|t| !t.transaction_completed)
+                .filter(|t| t.principal.to_string() == principal.to_string())
+                .filter(|t| matches!(t.transaction_type, transaction_type));
```
hpeebles, line 516: Ahh the issue is that `matches!` treats the right hand arg as a pattern to pattern match against, in order to check the 2 variables are equal you should use `t.transaction_type, transaction_type`.
To do this you'll need to derive `PartialEq` on `TransactionType`

rs/src/accounts_store.rs
bitdivine: I think you want == there, not matches.  `matches` is for matching a pattern, e.g. `matches!(x, SomeEnum::variant)`.

rs/src/accounts_store.rs
lmuntaner: good point, I can't think of a reason to keep them.

I can remove the flag and just remove them once they are completed.

rs/src/accounts_store.rs
lmuntaner: THere is `PartialEq` and `Eq` in the type already.

I am using `t.transaction_type, transaction_type` already: `.filter(|t| matches!(t.transaction_type, transaction_type));`, or what do you mean?

rs/src/accounts_store.rs
```diff
@@ -445,6 +467,59 @@ impl AccountsStore {
...
+        }
+    }
+
+    // Get first pending transaction
+    pub fn get_pending_transaction(
+        &self,
+        account_identifier: AccountIdentifier,
+        principal: PrincipalId,
+    ) -> Option<&PendingTransaction> {
+        match self.pending_transactions.get(&account_identifier) {
```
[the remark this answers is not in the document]
bitdivine: ^^^ with a small tweak: I think that will give an option of an option so you need to flatten.  Change the first map to `and_then`.  Testing, in case I cam talking nonsense...!

frontend/src/lib/canisters/nns-dapp/nns-dapp.certified.idl.js
lmuntaner: oh, true! thanks!

rs/src/accounts_store.rs
lmuntaner: Well, I am comparing two enums. I wasn't sure `==` was correct, but makes sense.

rs/src/accounts_store.rs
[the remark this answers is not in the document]
bitdivine: Confirmed
```
    pub fn get_pending_transaction(
        &self,
        account_identifier: AccountIdentifier,
        principal: PrincipalId,
    ) -> Option<&PendingTransaction> {
        self.pending_transactions.get(&account_identifier).and_then(|transactions|
            transactions
                .iter()
                .find(|t| !t.transaction_completed && t.principal.to_string() == principal.to_string())
        )
    }
```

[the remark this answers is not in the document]
hpeebles: Ah yup good catch!

rs/src/accounts_store.rs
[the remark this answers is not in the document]
hpeebles: Yes exactly, but if can simply be the first `if` branch.

rs/src/accounts_store.rs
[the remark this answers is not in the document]
lmuntaner: thanks! I'll try tomorrow. I see Rust has a lot of possibilities!

rs/src/accounts_store.rs
```diff
@@ -445,6 +467,59 @@ impl AccountsStore {
...
+        principal: PrincipalId,
+        block_height: BlockHeight,
+    ) {
+        self.multi_part_transactions_processor
+            .update_status(block_height, MultiPartTransactionStatus::Complete);
+        if let Some(transactions) = self.pending_transactions.get_mut(&to) {
+            let filtered_transactions = transactions
+                .iter_mut()
+                .filter(|t| !t.transaction_completed)
+                .filter(|t| t.principal.to_string() == principal.to_string())
```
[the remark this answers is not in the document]
lmuntaner: yes, I saw that afterwads. I had a problem where the transaction was not found, but it was for another reason. I'll revert and remove the `to_string`s

rs/src/accounts_store.rs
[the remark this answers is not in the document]
lmuntaner: yes, try, a `if let` I forgot about that

frontend/src/lib/api/sns.api.ts
lmuntaner: Yes, I know. I remove it locally, forgot to push it.

rs/src/accounts_store.rs
hpeebles: Ah right then simply use `.filter(|t| t.transaction_type == transaction_type));`

rs/src/accounts_store.rs
lmuntaner: changes

frontend/src/lib/api/sns.api.ts

rs/src/accounts_store.rs
lmuntaner: done

rs/src/accounts_store.rs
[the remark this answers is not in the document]
lmuntaner: done

rs/src/accounts_store.rs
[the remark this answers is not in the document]
lmuntaner: done

frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
lmuntaner: done

frontend/src/lib/canisters/nns-dapp/nns-dapp.certified.idl.js
lmuntaner: done

rs/src/accounts_store.rs
lmuntaner: done

rs/src/accounts_store.rs
lmuntaner: done

rs/src/accounts_store.rs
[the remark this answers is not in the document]
lmuntaner: done

frontend/src/lib/api/sns.api.ts

frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
```diff
@@ -108,34 +110,28 @@ export class NNSDappCanister {
...
   }): Promise<SubAccountDetails> {
-    const {
-      AccountNotFound,
-      NameTooLong,
-      SubAccountLimitExceeded,
-      Ok,
-    }: CreateSubAccountResponse = await this.certifiedService.create_sub_account(
-      subAccountName
-    );
+    const response: CreateSubAccountResponse =
```
lmuntaner, line 113: our compiler from candid to js converts variants to:

```
export type CreateSubAccountResponse =
  | { Ok: SubAccountDetails }
  | { AccountNotFound: null }
  | { NameTooLong: null }
  | { SubAccountLimitExceeded: null };
```

We can't deconstruct that type. Instead we need to check the presence of the property.

lmuntaner:
@hpeebles @bitdivine ready for another review 💪

rs/src/accounts_store.rs
```diff
@@ -1063,6 +1130,13 @@ impl AccountsStore {
             } else {
                 TransactionType::TopUpNeuron
             }
+        } else if let Some(pending_transaction) = self.get_pending_transaction(to, *principal) {
+            if let TransactionType::ParticipateSwap(canister_id) = pending_transaction.transaction_type {
```
hpeebles, line 1134: We can simply return `pending_transaction.transaction_type`

rs/src/accounts_store.rs
hpeebles: Rolling back to the previous version won't be possible, we'd need to roll back to a version which still includes the new subtype, otherwise candid will complain when deserializing.

rs/src/accounts_store.rs
```diff
@@ -30,6 +30,7 @@ pub struct AccountsStore {
     // TODO(NNS1-720): Use AccountIdentifier directly as the key for this HashMap
     accounts: HashMap<Vec<u8>, Account>,
     hardware_wallets_and_sub_accounts: HashMap<AccountIdentifier, AccountWrapper>,
+    pending_transactions: HashMap<AccountIdentifier, Vec<PendingTransaction>>,
```
hpeebles, line 33: Could this be simplified to `HashMap<(AccountIdentifier, AccountIdentifier), TransactionType>`?
Where the tuple key is 'from' account and the 'to' account.

hpeebles: Then there is no need for the `PendingTransaction` type and when processing transactions from the ledger we can do a single lookup to get the `TransactionType`.

rs/src/accounts_store.rs
bitdivine: I don't know the answer to this.  I started poking but didn't finish, sorry.

bitdivine: I did start poking but found some unexpectedly hard corners, e.g. some code that accepts candid in binary format but not text - why???

rs/src/accounts_store.rs
lmuntaner: Can you get the principal of an AccountIdentifier? If yes, then I would say it can work and it might be better.

rs/src/accounts_store.rs
lmuntaner: yes, it's the same.

lmuntaner: you mean instaed of the if statement. Yes, of course!

lmuntaner: done

rs/src/accounts_store.rs
lmuntaner: done

hpeebles (approved):

rs/src/canisters/swap.rs
```diff
@@ -0,0 +1,12 @@
+use dfn_candid::candid;
+use dfn_core::CanisterId;
+use ic_sns_swap::pb::v1::RefreshBuyerTokensRequest;
+
+pub async fn notify_swap_participation(
+    canister_id: CanisterId,
```
nmattia, line 6: out of curiosity (and since someone who uses the code might wonder to) what canister ID is that?

rs/src/main.rs
```diff
@@ -196,6 +196,21 @@ pub fn get_proposal_payload() {
     over_async(candid_one, proposals::get_proposal_payload)
 }
 
+#[export_name = "canister_update add_pending_notify_swap"]
```
nmattia, line 199: what's the difference between this and just using `#[update]`? is this to ensure that renaming the function won't break the interface?

nmattia (approved):
Rubber stamp

rs/src/canisters/swap.rs
lmuntaner: the canister id of the swap canister

rs/src/main.rs
lmuntaner: good question, I don't know. Maybe @hpeebles can tell us? I followed the same conventions.

PR #1247, opened 2022-08-15
build: agent-js v0.13.2 with idb breaking change

peterpeterparker:
# Motivation

agent-js `v0.13.x` introduces a breaking change. It uses indexeddb instead of local storage to save the session delegation and identity.

Indexddb is awesome and it's the most convenient way to access these information from web workers which we are already using for the cron that checks the delegation and which we might use for more features in the future.

# ⚠️ Important Note ⚠️ 

- on logout, local storage is not going cleared anymore but idb instead
- we don't explicitly clear idb but defer the work to agent-js which do so
- instead of syncing the auth on local storage changes, the identity will be checked in the auth cron

#  Dependencies

fake-indexeddb (https://github.com/dumbmatter/fakeIndexedDB) has been added to mock idb for jest tests happening in the nodejs context

# PRs required

- [X] ic-js PR https://github.com/dfinity/ic-js/pull/177
- [x] agent-js PR https://github.com/dfinity/agent-js/pull/611
- [x] agent-js PR https://github.com/dfinity/agent-js/pull/612

# Changes

- in web worker, read delegation from idb instead of localstorage to check the validity of the delegation
- in web worker, also check the auth is authenticated
- on logout, do not clear local storage
- mock idb for test with new dev dependency `fake-indexeddb`
- remove auth sync on local storage changes
- bump agent-js `v0.13.2`
- bump most recent nns-js, sns-js and utils-js


frontend/package.json
```diff
@@ -16,7 +16,7 @@
     "test": "TZ=UTC jest",
     "test:watch": "TZ=UTC npm run test -- --watchAll",
     "update:next": "npm update @dfinity/nns @dfinity/sns @dfinity/utils",
-    "update:agent": "npm rm @dfinity/agent @dfinity/auth-client @dfinity/authentication @dfinity/candid @dfinity/identity @dfinity/principal && npm i @dfinity/agent @dfinity/auth-client @dfinity/authentication @dfinity/candid @dfinity/identity @dfinity/principal",
+    "update:agent": "npm rm @dfinity/agent @dfinity/auth-client @dfinity/authentication @dfinity/candid @dfinity/identity @dfinity/principal @dfinity/nns @dfinity/sns @dfinity/utils && npm i @dfinity/agent @dfinity/auth-client @dfinity/authentication @dfinity/candid @dfinity/identity @dfinity/principal @dfinity/nns@next @dfinity/sns@next @dfinity/utils@next",
```
lmuntaner, line 19: Do the utils also depend on agent, etc?

frontend/src/lib/components/common/Guard.svelte
```diff
@@ -16,10 +16,7 @@
 
 <!-- storage: on every change in local storage we sync the auth state -->
```
lmuntaner, line 17: Remove or change this comment?

frontend/src/lib/workers/auth.worker.ts
```diff
@@ -16,21 +20,47 @@ export const stopIdleTimer = () => {
...
-    return;
-  }
+/**
+ * If there is no delegation or if not valid, then delegation is not valid
+ *
+ * @returns true if delegation is valid
+ */
+const checkDelegationChain = async (): Promise<boolean> => {
+  const idbStorage: IdbStorage = new IdbStorage();
+  const delegationChain: string | null = await idbStorage.get("delegation");
```
lmuntaner, line 54: maybe put `"delegation"` in a constant with a comment of why is it so?

frontend/src/tests/lib/services/auth.services.spec.ts
```diff
@@ -13,116 +13,127 @@ import { toastsStore } from "../../../lib/stores/toasts.store";
...
 
-  it("should call auth-client logout on logout", async () => {
-    const spy = jest.spyOn(mockAuthClient, "logout");
+  describe("auth-client", () => {
+    it("agent-js should clear indexeddb auth info on logout", async () => {
+      const idbStorage = new IdbStorage();
+      await idbStorage.set("delegation", "value");
 
-    await logout({});
+      const value = await idbStorage.get("delegation");
```
lmuntaner, line 32: and use the constant here as well then

lmuntaner (approved):
LGTM, minor comments but already approving.

frontend/package.json
peterpeterparker: Yes they are tight together. If we upgrade agent-js in ic-js then we have to remove both to upgrade otherwise npm will complain if one is above the other. Like upgrading ic-js that depende on agent-js v0.15 while agent-js in nns-dapp is still on v0.14

peterpeterparker: yes notably `createDefaultAgent`

frontend/src/lib/components/common/Guard.svelte
peterpeterparker: Thx!

frontend/src/lib/workers/auth.worker.ts
peterpeterparker: create a PR for agent-js to expose the constant there: https://github.com/dfinity/agent-js/pull/616

lmuntaner: That would be even better, thanks!

PR #1250, opened 2022-08-16
L2-944: Do not allow commitments above max

lmuntaner:
# Motivation

User should not be able to commit if the amount is over the total maximum.

# Changes

* Add one more check in "validParticipation" project util.

# Tests

* Add test for new case.

### DIFF
--- a/frontend/src/lib/utils/projects.utils.ts
+++ b/frontend/src/lib/utils/projects.utils.ts
@@ -132,6 +132,13 @@ const commitmentTooLarge = ({
   summary: SnsSummary;
   amountE8s: bigint;
 }): boolean => summary.swap.init.max_participant_icp_e8s < amountE8s;
+const commitmentExceedsAmountLeft = ({
+  summary: { swap, derived },
+  amountE8s,
+}: {
+  summary: SnsSummary;
+  amountE8s: bigint;
+}): boolean => swap.init.max_icp_e8s - derived.buyer_total_icp_e8s < amountE8s;
 
 /**
  * To participate to a swap:
@@ -210,6 +217,25 @@ export const validParticipation = ({
         $commitment: formatICP({ value: totalCommitment }),
         $maxCommitment: formatICP({
           value: project.summary.swap.init.max_participant_icp_e8s,
+        }),
+      },
+    };
+  }
+  if (
+    commitmentExceedsAmountLeft({
+      summary: project.summary,
+      amountE8s: totalCommitment,
+    })
+  ) {
+    return {
+      valid: false,
+      labelKey: "error__sns.commitment_exceeds_current_allowed",
+      substitutions: {
+        $commitment: formatICP({ value: totalCommitment }),
+        $maxCommitment: formatICP({
+          value:
+            project.summary.swap.init.max_icp_e8s -
+            project.summary.derived.buyer_total_icp_e8s,
         }),
       },
     };
--- a/frontend/src/lib/utils/projects.utils.ts
+++ b/frontend/src/lib/utils/projects.utils.ts
@@ -232,7 +232,7 @@ export const validParticipation = ({
       labelKey: "error__sns.commitment_exceeds_current_allowed",
       substitutions: {
         $commitment: formatICP({ value: totalCommitment }),
-        $maxCommitment: formatICP({
+        $remainingCommitment: formatICP({
           value:
             project.summary.swap.init.max_icp_e8s -
             project.summary.derived.buyer_total_icp_e8s,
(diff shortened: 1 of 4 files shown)


lmuntaner:
UX:

Trying to increase commitment by 30 when there is only 21 left to reach maximum.

https://user-images.githubusercontent.com/4550653/184851984-4829f8d9-3dab-4f48-a302-9ee447433604.mov

lmuntaner:

frontend/src/lib/utils/projects.utils.ts
```diff
@@ -132,6 +132,13 @@ const commitmentTooLarge = ({
   summary: SnsSummary;
   amountE8s: bigint;
 }): boolean => summary.swap.init.max_participant_icp_e8s < amountE8s;
+const commitmentExceedsAmountLeft = ({
```
peterpeterparker, line 135: what you think, makes sense to add this new checks to `canUserParticipateToSwap` too?

frontend/src/lib/utils/projects.utils.ts
```diff
@@ -214,6 +221,25 @@ export const validParticipation = ({
...
+      summary: project.summary,
+      amountE8s: totalCommitment,
+    })
+  ) {
+    return {
+      valid: false,
+      labelKey: "error__sns.commitment_exceeds_current_allowed",
+      substitutions: {
+        $commitment: formatICP({ value: totalCommitment }),
+        $maxCommitment: formatICP({
```
peterpeterparker, line 235: should we rename the variable `maxCommitment` to `remainingCommitment`?

can maybe make the label easier to translate if we ever translate it

frontend/src/tests/lib/utils/projects.utils.spec.ts
```diff
@@ -454,6 +458,34 @@ describe("project-utils", () => {
...
+              max_participant_icp_e8s: maxE8s,
+            },
+          },
+        },
+      };
+      const { valid } = validParticipation({
+        project,
+        amount: ICP.fromE8s(participationE8s),
+      });
+      expect(valid).toBe(false);
```
peterpeterparker, line 487: is there also a test that checks that it returns `true` when `commitmentExceedsAmountLeft` is ok?

frontend/src/lib/utils/projects.utils.ts
lmuntaner: makes sense

frontend/src/tests/lib/utils/projects.utils.spec.ts
lmuntaner: yes, the valid participation is a participation that doesn't exceed the amount left.

frontend/src/lib/utils/projects.utils.ts
lmuntaner: no, this only happens when we have some amount that the user wants to participate with.

`canUserParticipateToSwap` doesn't have any extra amount. It's more, can the user participate with just 1ICP?

frontend/src/lib/utils/projects.utils.ts
lmuntaner: done

lmuntaner:

frontend/src/lib/utils/projects.utils.ts
peterpeterparker: thanks for the explanation. so, I was a bit confuse. can we make the name of the function more obvious or add a comment?

frontend/src/tests/lib/utils/projects.utils.spec.ts
peterpeterparker: super

frontend/src/lib/utils/projects.utils.ts
lmuntaner: There is a comment in the canUserParticipateToSwap:
```
/**
 * To participate to a swap:
 *
 * - the sale's lifecycle should be Open
 * - user commitment should not be bigger than the maximal commitment allowed per user
 * - the maximal commitment allowed per user should be defined
 */
```
Just not showed here.

Or do you mean a comment in `commitmentExceedsAmountLeft`?

peterpeterparker: i mean `commitmentExceedsAmountLeft`

lmuntaner: done

lmuntaner:

peterpeterparker (approved):
thx!

PR #1255, opened 2022-08-17
Feat: Add hotkey debug flow

lmuntaner:
# Motivation

Context: A user is the controller of a neuron. The same user adds the principal as hotkey.
Edge case: If the user then removes the hotkey, the neuron disappears.

Added a new flow in the debug store to add the user as a hotkey again.

More info here: https://forum.dfinity.org/t/8-year-locked-neuron-missing-from-neurons-tab/9619/42?u=ielashi

# Changes

* Add new LogType "ah" in the debug flow that adds the user's principal to a neuron.
* Take care of the edge case in the neuron util is "isHotKeyControllable"

# Tests

* No tests for the debug flow.
* Test for the edge case added in isHotKeyControllable


lmuntaner:

PR #1258, opened 2022-08-18
L2 939 lm add query param current selected project

lmuntaner:
# Motivation

Create a URL for sns neurons page using query params.

# Changes

# Tests


lmuntaner:
@peterpeterparker draft PR implemeting query params to select the project.

The query param is not working very nicely with the hash. It's working, but I have two issues:

* The query string is kept when navigating, even though it doesn't add anything.
* The menu items sometimes use replace (keeping the query string), sometimes not, removing the query string from the URL, but not from the routeStore. Therefore, when the user comes back to neurons it appears again.

lmuntaner:
UX video navigating:

https://user-images.githubusercontent.com/4550653/185305865-ccc47b6a-e870-4c97-a63a-1b2c46690af9.mov

lmuntaner:
CLOSING.

We will use path params `/project/:rootCanisterId/neurons` (or similar) instead of query params.

PR #1259, opened 2022-08-18
feat: voting card information reviewed and partial new layout

peterpeterparker:
# Motivation

To improve user's confidence the system data should be prioritized over the data user can input when they submit a proposal - e.g. proposal topic vs proposal title.

In addition, NNS-dapp should also display the proposal **type** which was missing so far.

Finally, the labels of the proposal meta information (topic, status, etc.) should be aligned with the dashboard.

# Changes

- review and align i18n proposal labels as those of the dashboard
- introduce proposal `type`
- add new `modern` layout to cmp `<ProposalCard />`
- review presentation of the information by giving priority on system data
- refactor few utils functions and i18n keys to make their naming closer to the Candid naming
- introduce last `gix-components`

# Screenshot

Capture d’écran 2022-08-18 à 09 06 29


frontend/src/lib/components/accounts/AccountCard.svelte
```diff
@@ -26,7 +26,7 @@
 </Card>
 
 <style lang="scss">
-  @use "../../themes/mixins/card";
+  @use "@dfinity/gix-components/styles/mixins/card";
```
lmuntaner, line 29: Not a fan that UI Frameworks expose mixins instead of components. But definitely not something to discuss in this PR.

lmuntaner (approved):
LGTM!

PR #1260, opened 2022-08-18
L2-943: Improve Increase Participation Flow

lmuntaner:
# Motivation

Improve UI when user increases participation.

Fix also the minimum to participate. If the user has already participated, the increase can be as small as wished.

# Changes 

* Check whether the user is participating in the Participate Modal and change UI slightly.
* New project util currentUserMaxCommitment to calculate the maximum participation allowed for that user for that project.
* commitmentTooSmall util takes into account current commitment.
* Fix a bug in validParticipation when user wanted to participate until max.

# Tests

* Test new participation and increasing in ParticipateSwapModal
* Test for new util currentUserMaxCommitment
* Add more tests to validParticipation to cover the bug and more test cases.

### DIFF
--- a/frontend/src/lib/components/project-detail/ParticipateScreen.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateScreen.svelte
@@ -23,7 +23,7 @@
   // TODO: Handle min and max validations inline: https://dfinity.atlassian.net/browse/L2-798
   export let minAmount: ICP;
   export let maxAmount: ICP;
-  export let increasingParticipation: boolean = false;
+  export let userHasParticipatedToSwap: boolean = false;
 
   let max: number = 0;
   $: max = maxICP({
@@ -85,7 +85,7 @@
   </div>
   <div class="wrapper info">
     <AmountInput bind:amount on:nnsMax={addMax} {max} {errorMessage} />
-    {#if increasingParticipation}
+    {#if userHasParticipatedToSwap}
       <p class="right">
         {$i18n.sns_project_detail.max_left}
         <IcpComponent singleLine icp={maxAmount} />
--- a/frontend/src/lib/components/project-detail/ParticipateScreen.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateScreen.svelte
@@ -13,6 +13,7 @@
   import { convertNumberToICP, maxICP } from "../../utils/icp.utils";
   import SelectAccountDropdown from "../accounts/SelectAccountDropdown.svelte";
   import IcpComponent from "../ic/ICP.svelte";
+  import IcpText from "../ic/ICPText.svelte";
   import AmountInput from "../ui/AmountInput.svelte";
   import KeyValuePair from "../ui/KeyValuePair.svelte";
 
@@ -92,12 +93,12 @@
       </p>
     {:else}
       <KeyValuePair>
-        <span slot="key"
-          >{$i18n.core.min} <IcpComponent singleLine icp={minAmount} /></span
-        >
-        <span slot="value"
-          >{$i18n.core.max} <IcpComponent singleLine icp={maxAmount} /></span
-        >
+        <IcpText slot="key" icp={minAmount}>
+          {$i18n.core.min}
+        </IcpText>
+        <IcpText slot="value" icp={maxAmount}>
+          {$i18n.core.max}
+        </IcpText>
       </KeyValuePair>
     {/if}
     <p class="right">
--- a/frontend/src/lib/components/project-detail/ParticipateScreen.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateScreen.svelte
@@ -23,6 +23,7 @@
   // TODO: Handle min and max validations inline: https://dfinity.atlassian.net/browse/L2-798
   export let minAmount: ICP;
   export let maxAmount: ICP;
+  export let increasingParticipation: boolean = false;
 
   let max: number = 0;
   $: max = maxICP({
@@ -84,14 +85,21 @@
   </div>
   <div class="wrapper info">
     <AmountInput bind:amount on:nnsMax={addMax} {max} {errorMessage} />
-    <KeyValuePair>
-      <span slot="key"
-        >{$i18n.core.min} <IcpComponent singleLine icp={minAmount} /></span
-      >
-      <span slot="value"
-        >{$i18n.core.max} <IcpComponent singleLine icp={maxAmount} /></span
-      >
-    </KeyValuePair>
+    {#if increasingParticipation}
+      <p class="right">
+        {$i18n.sns_project_detail.max_left}
+        <IcpComponent singleLine icp={maxAmount} />
+      </p>
+    {:else}
+      <KeyValuePair>
+        <span slot="key"
+          >{$i18n.core.min} <IcpComponent singleLine icp={minAmount} /></span
+        >
+        <span slot="value"
+          >{$i18n.core.max} <IcpComponent singleLine icp={maxAmount} /></span
+        >
+      </KeyValuePair>
+    {/if}
     <p class="right">
       <span>{$i18n.accounts.transaction_fee}</span>
       <IcpComponent singleLine icp={$mainTransactionFeeStoreAsIcp} />
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -601,6 +601,7 @@
     "completed": "Completed",
     "sale_start": "Sale Start",
     "sale_end": "Sale End",
+    "max_left": "max left",
     "max_user_commitment_reached": "Maximum commitment reached"
   },
   "sns_neuron_detail": {
--- a/frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
+++ b/frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
@@ -12,25 +12,32 @@
     type ProjectDetailContext,
   } from "../../types/project-detail.context";
   import type { SnsSwapInit } from "@dfinity/sns";
-  import { currentUserMaxCommitment } from "../../utils/projects.utils";
-  import type { SnsFullProject } from "../../stores/projects.store";
+  import {
+    currentUserMaxCommitment,
+    hasUserParticipatedToSwap,
+  } from "../../utils/projects.utils";
+  import type { SnsSummary, SnsSwapCommitment } from "../../types/sns";
 
   const { store: projectDetailStore } = getContext<ProjectDetailContext>(
     PROJECT_DETAIL_CONTEXT_KEY
   );
 
-  let project: SnsFullProject;
+  let summary: SnsSummary;
+  let swapCommitment: SnsSwapCommitment | undefined;
   // type safety validation is done in ProjectDetail component
-  $: project = $projectDetailStore as SnsFullProject;
-  let increasingParticipation: boolean = false;
-  $: increasingParticipation =
-    (project.swapCommitment?.myCommitment?.amount_icp_e8s ?? BigInt(0)) >
-    BigInt(0);
+  $: summary = $projectDetailStore.summary as SnsSummary;
+  $: swapCommitment = $projectDetailStore.swapCommitment as
+    | SnsSwapCommitment
+    | undefined;
+  let userHasParticipatedToSwap: boolean = false;
+  $: userHasParticipatedToSwap = hasUserParticipatedToSwap({
+    swapCommitment,
+  });
 
   let init: SnsSwapInit;
   $: ({
     swap: { init },
-  } = project.summary);
+  } = summary);
 
   const steps: Steps = [
     {
@@ -47,7 +54,7 @@
 
   let title: string | undefined;
   $: title =
-    currentStep?.name === "Participate" && increasingParticipation
+    currentStep?.name === "Participate" && userHasParticipatedToSwap
       ? $i18n.sns_project_detail.increase_participation
       : currentStep?.title;
 
@@ -58,11 +65,16 @@
   let amount: number | undefined;
 
   let maxCommitment: ICP;
-  $: maxCommitment = ICP.fromE8s(currentUserMaxCommitment(project));
+  $: maxCommitment = ICP.fromE8s(
+    currentUserMaxCommitment({
+      summary,
+      swapCommitment,
+    })
+  );
 
   let minCommitment: ICP;
   $: minCommitment = ICP.fromE8s(
-    increasingParticipation ? BigInt(0) : init.min_participant_icp_e8s
+    userHasParticipatedToSwap ? BigInt(0) : init.min_participant_icp_e8s
   );
 
   const goNext = () => {
@@ -74,14 +86,16 @@
 </script>
 
 <WizardModal {steps} bind:currentStep bind:this={modal} on:nnsClose>
-  <svelte:fragment slot="title">{title}</svelte:fragment>
+  <svelte:fragment slot="title"
+    >{title ?? $i18n.sns_project_detail.participate}</svelte:fragment
+  >
   {#if currentStep.name === "Participate"}
     <ParticipateScreen
       bind:selectedAccount
       bind:amount
       on:nnsNext={goNext}
       on:nnsClose
-      {increasingParticipation}
+      {userHasParticipatedToSwap}
       minAmount={minCommitment}
       maxAmount={maxCommitment}
     />
--- a/frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
+++ b/frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
@@ -23,12 +23,10 @@
   );
 
   let summary: SnsSummary;
-  let swapCommitment: SnsSwapCommitment | undefined;
+  let swapCommitment: SnsSwapCommitment | undefined | null;
   // type safety validation is done in ProjectDetail component
   $: summary = $projectDetailStore.summary as SnsSummary;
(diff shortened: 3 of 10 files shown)


lmuntaner:
New UI:

https://user-images.githubusercontent.com/4550653/185405284-b1b5f3f7-9043-46e0-92fa-7a488ffc49e6.mov

lmuntaner:

frontend/src/lib/types/project-detail.context.ts
```diff
@@ -1,10 +1,7 @@
 import type { Writable } from "svelte/store";
-import type { SnsSummary, SnsSwapCommitment } from "./sns";
+import type { SnsFullProject } from "../stores/projects.store";
 
-export interface ProjectDetailStore {
-  summary: SnsSummary | undefined | null;
-  swapCommitment: SnsSwapCommitment | undefined | null;
-}
+export type ProjectDetailStore = Partial<SnsFullProject>;
```
peterpeterparker, line 4: it isn't `Partial`:

- `rootCanisterId` is never set so instead of partial it could be probably `Omit<SnsFullProject, 'rootCanisterId'>`
- but in addition, unlike in `SnsFullProject`, `summary` can be here `undefined` or `null`
- likewise `swapCommitment` can be null

frontend/src/lib/components/project-detail/ParticipateScreen.svelte
```diff
@@ -84,14 +85,21 @@
...
-        >{$i18n.core.max} <IcpComponent singleLine icp={maxAmount} /></span
-      >
-    </KeyValuePair>
+    {#if increasingParticipation}
+      <p class="right">
+        {$i18n.sns_project_detail.max_left}
+        <IcpComponent singleLine icp={maxAmount} />
+      </p>
+    {:else}
+      <KeyValuePair>
```
peterpeterparker, line 94: don't we have already in the component the exact same piece of code for `$i18n.core.min` and `$i18n.core.max`? if yes, can we extract a cmp?

frontend/src/lib/i18n/en.json
```diff
@@ -600,6 +600,7 @@
     "completed": "Completed",
     "sale_start": "Sale Start",
     "sale_end": "Sale End",
+    "max_left": "max left",
```
peterpeterparker, line 603: with capital letters?

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
```diff
@@ -40,12 +45,26 @@
     },
   ];
 
+  let title: string | undefined;
```
peterpeterparker, line 48: edge case: `undefined` can briefly be displayed on screen. either catch `undefined` in the template or declare it as `string`

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
```diff
@@ -6,26 +6,31 @@
...
     PROJECT_DETAIL_CONTEXT_KEY
   );
 
-  let summary: SnsSummary;
+  let project: SnsFullProject;
   // type safety validation is done in ProjectDetail component
-  $: summary = $projectDetailStore.summary as SnsSummary;
+  $: project = $projectDetailStore as SnsFullProject;
+  let increasingParticipation: boolean = false;
+  $: increasingParticipation =
```
peterpeterparker, line 26: why not using `userHasParticipatedToSwap`?

frontend/src/lib/utils/projects.utils.ts
```diff
@@ -115,16 +115,36 @@ export const durationTillSwapStart = (
   return BigInt(nowInSeconds()) - start_timestamp_seconds;
 };
 
+// Returns the minimum between:
+// - user remaining commitment to reach user maximum
+// - remamining commitment to reach project maximum
+export const currentUserMaxCommitment = ({
+  summary: { swap, derived },
```
peterpeterparker, line 122: if you want some deep destruction 😁: `summary: { swap: {init: {max_icp_e8s, max_participant_icp_e8s}}, derived: {buyer_total_icp_e8s} }`

then you got

```
const remainingProjectCommitment = max_icp_e8s - buyer_total_icp_e8s;
  const remainingUserCommitment =
    max_participant_icp_e8s -
    (swapCommitment?.myCommitment?.amount_icp_e8s ?? BigInt(0));
```

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
lmuntaner: Ok, I can handle it in the template:

```
<svelte:fragment slot="title">{title ?? $i18n.sns_project_detail.participate}</svelte:fragment>
```

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
lmuntaner: Good point, sorry, I didn't see this util.

frontend/src/lib/components/project-detail/ParticipateScreen.svelte
lmuntaner: Not that I know of. I search for `$i18n.core.min` and only found it here.

frontend/src/lib/i18n/en.json
lmuntaner: not according to Mischa...

I put the screenshot in the issue: https://dfinity.atlassian.net/browse/L2-943

frontend/src/lib/types/project-detail.context.ts
lmuntaner: Ok, I'll refactor.

lmuntaner: btw, wouldn't the `null` be also applied to the SnsFullProject type? Isn't the `null` also an option in Launchpad?

lmuntaner: done the refactor. But I'd say the question of `null` in SnsFullProject remains, or how is it solved in the Launchpad?

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
lmuntaner: done

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
lmuntaner: done

lmuntaner:

frontend/src/lib/types/project-detail.context.ts
peterpeterparker: currently I think we did not needed `null` on `SnsFullProject` in the Launchpad because the stores it derives are using `null` - the loading state is checked on the stores not on the derived one.

frontend/src/lib/components/project-detail/ParticipateScreen.svelte
peterpeterparker: so, should we extract these two exact same piece of code?

frontend/src/lib/i18n/en.json
peterpeterparker: Funny, it's always capital letters except that one 🤷‍♂️. Thanks for the feedback!

frontend/src/lib/types/project-detail.context.ts
```diff
@@ -1,10 +1,10 @@
 import type { Writable } from "svelte/store";
 import type { SnsSummary, SnsSwapCommitment } from "./sns";
 
-export interface ProjectDetailStore {
+export type ProjectDetailStore = {
```
peterpeterparker, line 4: strictly out of curiosity - I'm totally fine with `type` - can you explain me the change? any advantages of using a type instead of interface here?

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
```diff
@@ -6,21 +6,33 @@
...
+    hasUserParticipatedToSwap,
+  } from "../../utils/projects.utils";
+  import type { SnsSummary, SnsSwapCommitment } from "../../types/sns";
 
   const { store: projectDetailStore } = getContext<ProjectDetailContext>(
     PROJECT_DETAIL_CONTEXT_KEY
   );
 
   let summary: SnsSummary;
+  let swapCommitment: SnsSwapCommitment | undefined;
```
peterpeterparker, line 26: just to be sure, it's correct, it cannot be `| null` here?

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -38,7 +38,7 @@
       rootCanisterId,
       onError: () => {
         // hide unproven data
-        $projectDetailStore.summary = null;
+        $projectDetailStore.summary = undefined;
```
peterpeterparker, line 41: spontaneously I would say it should go to `null` on error because that's an unexpected state. any reasons to move to `undefined`?

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -48,7 +48,7 @@
       rootCanisterId,
       onError: () => {
         // hide unproven data
-        $projectDetailStore.swapCommitment = null;
+        $projectDetailStore.swapCommitment = undefined;
```
peterpeterparker, line 51: dito

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -93,15 +96,15 @@
             ({ rootCanisterId: rootCanister }) =>
               rootCanister?.toText() === rootCanisterId
           )
-        : null;
+        : undefined;
```
peterpeterparker, line 99: here too i think `null` is important.

if store is loaded and data is not found then undefined, otherwise null because not loaded yet

but maybe there is a good reason to change?

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -93,15 +96,15 @@
...
 
     $projectDetailStore.swapCommitment =
       rootCanisterId !== undefined
         ? $snsSwapCommitmentsStore?.find(
             (item) =>
               item?.swapCommitment?.rootCanisterId?.toText() === rootCanisterId
           )?.swapCommitment
-        : null;
+        : undefined;
```
peterpeterparker, line 107: dito

frontend/src/lib/components/project-detail/ParticipateScreen.svelte
lmuntaner: Do you mean a `<span>` plus an `Icp`? component. Not really to be honest. It doesn't tell me anything if I extract a text plus an Icp component.

frontend/src/lib/types/project-detail.context.ts
lmuntaner: no, since I changed it to type to use the `Partial<SnsFullProject>`, I didn't change it back.

There is a lint rule to force all to interface or all to `type`. Not a strong opinion on whihc one, but if we want consistency we could add the rule.

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
lmuntaner: For now yes, I still don't manage the use case where the values are `null` separetely. I'll change it.

frontend/src/routes/ProjectDetail.svelte
lmuntaner: you're right. Reverting to `null`.

frontend/src/routes/ProjectDetail.svelte
lmuntaner: yep

frontend/src/routes/ProjectDetail.svelte
lmuntaner: I thought `null` meant error, which it's set in the `onError`. Here it means that it has not yet been loaded, therefore `undefined` means still loading, no?

frontend/src/lib/types/project-detail.context.ts
lmuntaner: Doesn't `null` mean error? It's another use case than `undefined` or loading, isn't it?

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
lmuntaner: done

frontend/src/routes/ProjectDetail.svelte
lmuntaner: done

frontend/src/routes/ProjectDetail.svelte
lmuntaner: done

frontend/src/routes/ProjectDetail.svelte
lmuntaner: ditto

lmuntaner:

frontend/src/routes/ProjectDetail.svelte
peterpeterparker: good question, it is initiated by:

```
const rootCanisterId = routePathRootCanisterId(path);
mapProjectDetail(rootCanisterId);
```

which set `routePathRootCanisterId` to `undefined` when the route is not the project detail page so in that case it should not load at all therefore it was set to `null`.

but I guess it still works with `undefined`?

frontend/src/lib/types/project-detail.context.ts
peterpeterparker: Not only, in our store - or at least in sns and auth stores - it's mostly `undefined` not found and `null` not initialized and "this state does or should not explicitely exist"

e.g. `snsSwapCommitmentsStore` -> `setLoadingState() { set(null);  }`

frontend/src/lib/components/project-detail/ParticipateScreen.svelte
peterpeterparker: I mean

```
<KeyValuePair>
        <span slot="key"
          >{$i18n.core.min} <IcpComponent singleLine icp={minAmount} /></span
        >
        <span slot="value"
          >{$i18n.core.max} <IcpComponent singleLine icp={maxAmount} /></span
        >
      </KeyValuePair>
      ```
      
      comes twice in the code of the same component

File: frontend/src/lib/types/project-detail.context.ts (hunk 7 above)
Comment (peterpeterparker) [reply frontend/src/lib/types/project-detail.context.ts:4]: Ah ok, thanks for the feedback. No strong opinion from my side, like I said, was really just a curiosity question

File: frontend/src/routes/ProjectDetail.svelte (hunk 11 above)
Comment (lmuntaner) [reply frontend/src/routes/ProjectDetail.svelte:99]: yes

Btw, now that I realized: Do we need the context and store in the project details when all the projects are loaded at the beginning? Shouldn't the components just read the store? And reload should happen in the service, not the component.

File: frontend/src/lib/types/project-detail.context.ts (hunk 1 above)
Comment (lmuntaner) [reply frontend/src/lib/types/project-detail.context.ts]: so the opposite, `null` is for loading state and `undefined` for error?

File: frontend/src/lib/components/project-detail/ParticipateScreen.svelte (hunk 2 above)
Comment (lmuntaner) [reply frontend/src/lib/components/project-detail/ParticipateScreen.svelte:94]: ok

Comment (lmuntaner) [reply frontend/src/lib/components/project-detail/ParticipateScreen.svelte:95]: done

**Comment by lmuntaner** [comment]

File: frontend/src/lib/types/project-detail.context.ts (hunk 1 above)
Comment (peterpeterparker) [reply frontend/src/lib/types/project-detail.context.ts]: Error is a bit grey zone I guess, not against using `undefined` as log as routing and behavior of this particular feature still works of course.

File: frontend/src/routes/ProjectDetail.svelte (hunk 11 above)
Comment (peterpeterparker) [reply frontend/src/routes/ProjectDetail.svelte:99]: The component just read the stores, not sure what you mean with reload.

We need the context and store like we do in the proposal detail page or any other details page, because user can navigate back and forth.

File: frontend/src/lib/utils/projects.utils.ts [hunk 13]
```diff
@@ -115,16 +115,39 @@ export const durationTillSwapStart = (
   return BigInt(nowInSeconds()) - start_timestamp_seconds;
 };
 
+// Returns the minimum between:
```
peterpeterparker, line 118: - small typo "remamining"
- maybe formatting the comment as following?

```
/**
 * Returns the minimum between:
 * - user remaining commitment to reach user maximum
 * - remaining commitment to reach project maximum
 */
 ```

File: frontend/src/routes/ProjectDetail.svelte [hunk 14]
```diff
@@ -62,11 +62,15 @@
...
-      loadSummary(rootCanisterId),
-      loadSwapState(rootCanisterId),
-    ]);
+    try {
+      await Promise.all([
+        loadSummary(rootCanisterId),
+        loadSwapState(rootCanisterId),
+      ]);
+    } catch (err) {
+      // TODO: Manage errors https://dfinity.atlassian.net/browse/L2-958
```
peterpeterparker, line 71: not sure we need this, both `loadSummary` and `loadSwapState` catches errors and redirect to main screen in case it happens.

peterpeterparker (approved):
Just added two last minor comments. LGTM

frontend/src/lib/utils/projects.utils.ts
lmuntaner: sure

frontend/src/routes/ProjectDetail.svelte
lmuntaner: I will try the error cases in that issue and remove any superfluous code then.

PR #1265, opened 2022-08-22
L2-956: Ignore last commitment error on notify

lmuntaner:
# Motivation

User of the last commitment received an error notification because the notify was failing. The problem is that the backend might be faster to notify than the frontend. That causes the project to change state to Committed and the subsequent notify to fail.

Catch that error specifically and ignore it.

# Changes

* Check for last commitment notify error in the participateInSwap sns service.
* Export project util commitmentExceedsAmountLeft to be used in participateInSwap.

# Tests

* Test exported project util commitmentExceedsAmountLeft
* Test the service for the new use case.

### DIFF
--- a/frontend/src/lib/services/sns.services.ts
+++ b/frontend/src/lib/services/sns.services.ts
@@ -31,7 +31,10 @@ import type { QuerySnsMetadata, QuerySnsSwapState } from "../types/sns.query";
 import { assertEnoughAccountFunds } from "../utils/accounts.utils";
 import { getLastPathDetail, isRoutePath } from "../utils/app-path.utils";
 import { toToastError } from "../utils/error.utils";
-import { validParticipation } from "../utils/projects.utils";
+import {
+  commitmentExceedsAmountLeft,
+  validParticipation,
+} from "../utils/projects.utils";
 import { getSwapCanisterAccount } from "../utils/sns.utils";
 import { getAccountIdentity, syncAccounts } from "./accounts.services";
 import { getIdentity } from "./auth.services";
@@ -251,13 +254,14 @@ export const participateInSwap = async ({
   account: Account;
 }): Promise<{ success: boolean }> => {
   let success = false;
+  let project: SnsFullProject | undefined;
   try {
     const transactionFee = get(transactionsFeesStore).main;
     assertEnoughAccountFunds({
       account,
       amountE8s: amount.toE8s() + transactionFee,
     });
-    const project = getProjectFromStore(rootCanisterId);
+    project = getProjectFromStore(rootCanisterId);
     const { valid, labelKey, substitutions } = validParticipation({
       project,
       amount,
@@ -282,6 +286,22 @@ export const participateInSwap = async ({
 
     return { success };
   } catch (error) {
+    // The last commitment might trigger this error
+    // because the backend is faster than the frontend at notifying the commitment.
+    if (
+      error !== undefined &&
+      "message" in error &&
+      error.message.includes("'open' state") &&
+      project?.summary !== undefined &&
+      // If it's the last commitment, it means that one more e8 is not a valid participation.
+      commitmentExceedsAmountLeft({
+        summary: project?.summary,
+        amountE8s: amount.toE8s() + BigInt(1),
+      })
+    ) {
+      await syncAccounts();
+      return { success: true };
+    }
     toastsStore.error(
       toToastError({
         err: error,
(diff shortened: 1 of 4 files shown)


lmuntaner:

frontend/src/lib/services/sns.services.ts
```diff
@@ -282,6 +286,20 @@ export const participateInSwap = async ({
 
     return { success };
   } catch (error) {
+    // The last commitment might trigger this error
+    // because the backend is faster than the frontend at notifying the commitment.
+    if (
```
peterpeterparker, line 291: why not adding a dedicated try / catch around `participateInSnsSwap` for that specific error?

that way you have to have to modify the `const project` in `let` and you avoid to have to add `syncAccounts` and `success:true`in that global catch

pseudo code:

```
try
   validParticipation
   
   try
       participateInSnsSwap
   catch 
       if !specific error
           throw
    syncAccount
    return true
catch               
    print error
    return false
```

lmuntaner: sure, I wasn't sure you would like a try inside another try.

lmuntaner: done

lmuntaner:

frontend/src/lib/services/sns.services.ts
```diff
@@ -251,13 +254,14 @@ export const participateInSwap = async ({
...
   let success = false;
+  let project: SnsFullProject | undefined;
   try {
     const transactionFee = get(transactionsFeesStore).main;
     assertEnoughAccountFunds({
       account,
       amountE8s: amount.toE8s() + transactionFee,
     });
-    const project = getProjectFromStore(rootCanisterId);
+    project = getProjectFromStore(rootCanisterId);
```
peterpeterparker, line 264: this change can be revered now that we move the try catch

frontend/src/lib/services/sns.services.ts
```diff
@@ -269,13 +273,30 @@ export const participateInSwap = async ({
...
+        amount,
+        controller: accountIdentity.getPrincipal(),
+        fromSubAccount:
+          "subAccount" in account ? account.subAccount : undefined,
+      });
+    } catch (error) {
+      // The last commitment might trigger this error
+      // because the backend is faster than the frontend at notifying the commitment.
+      if (
+        error.message?.includes("'open' state") !== true ||
```
peterpeterparker, line 289: - can we document the message - link to the backend code? not sure that in 6 months we will remember what is `'open' state` and why it is ok

- edge case maybe, but what if project.summary is undefined, no error then?

frontend/src/lib/services/sns.services.ts
lmuntaner: ok, yes

frontend/src/lib/services/sns.services.ts
lmuntaner: - I can try to find the backend code, and link it sure.
- If summary is undefined, then the error is thrown and not skipped. I think this is preferred since something wrong happenned.

lmuntaner: added comment

frontend/src/lib/services/sns.services.ts
lmuntaner: done

lmuntaner:

frontend/src/lib/services/sns.services.ts
peterpeterparker: > added comment

thx

>If summary is undefined, then the error is thrown and not skipped.

is it? maybe I am missing something but from the code `(project?.summary !== undefined && ...`, if summary is undefined, error is not thrown

lmuntaner: Aahh, I see the case you mean. Yes, I need to add a new case. Let me refactor the if statement, is getting hard to read.

lmuntaner: Changed to:
```
// The last commitment might trigger this error
      // because the backend is faster than the frontend at notifying the commitment.
      // Backend error line: https://github.com/dfinity/ic/blob/6ccf23ec7096b117c476bdcd34caa6fada84a3dd/rs/sns/swap/src/swap.rs#L461
      const openStateError = error.message?.includes("'open' state") === true;
      // If it's the last commitment, it means that one more e8 is not a valid participation.
      const lastCommitment =
        project?.summary !== undefined &&
        commitmentExceedsAmountLeft({
          summary: project?.summary,
          amountE8s: amount.toE8s() + BigInt(1),
        });
      if (!(openStateError && lastCommitment)) {
        throw error;
      }
```

peterpeterparker: Ah nicer, thanks

peterpeterparker (approved):
LGTM, thanks

It does the trick but for a nicer solution, thinking about it, it would be useful to reach out NNS team to ask them to not throw error in the backend but to e.g. return a variant so that we could safely check for the error. When we got 5 we should maybe reach them out about it.

lmuntaner:
Yes, totally agree that the canister should not panic in this case and return a variant. Yet, they are swamped with things and I prefer they spend time doing the refunds in the finalize_swap for example.

We need a way to request features or changes in the nns canisters. Let's also discuss this in the daily.

PR #1266, opened 2022-08-23
L2-958: Manage error project not found

lmuntaner:
# Motivation

User should be redirected to launchpad when trying to access details of a non-existing project.

# Changes

* Changes in the ProjectDetail page to manage edge case. Specifically in mapProjectDetail.

# Tests

* Provide test for not found or wrong canister id in the url of Project Detail.

### DIFF
--- a/frontend/src/routes/ProjectDetail.svelte
+++ b/frontend/src/routes/ProjectDetail.svelte
@@ -133,6 +133,7 @@
         return;
       }
 
+      // Edge case, the previous check already ensures that `routePathRootCanisterId` will return a defined value.
       const rootCanisterId = routePathRootCanisterId(path);
       if (rootCanisterId === undefined) {
         return;
@@ -164,7 +165,8 @@
   $: layoutTitleStore.set($projectDetailStore?.summary?.metadata.name ?? "");
 
   let loading: boolean;
-  $: loading = $snsSummariesStore.length === 0;
+  $: loading =
+    $snsSummariesStore.length === 0 || isNullish($snsSwapCommitmentsStore);
   let notFound: boolean;
   $: notFound = $projectDetailStore.summary === undefined;
 
--- a/frontend/src/routes/ProjectDetail.svelte
+++ b/frontend/src/routes/ProjectDetail.svelte
@@ -25,6 +25,8 @@
   import { isNullish } from "../lib/utils/utils";
   import { writable } from "svelte/store";
   import { snsSummariesStore } from "../lib/stores/sns.store";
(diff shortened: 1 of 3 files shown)


lmuntaner:
New UX:

Both for a valid canister id that is not present, and an invalid canister id. Both are the same for the user.

https://user-images.githubusercontent.com/4550653/186099861-c84f7f7e-5fb2-495f-a25a-49d1cdc7c318.mov

lmuntaner:

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -120,6 +135,9 @@
       }
 
       const rootCanisterId = routePathRootCanisterId(path);
+      if (rootCanisterId === undefined) {
```
peterpeterparker, line 138: we need that checks?

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -146,17 +164,25 @@
...
   $: layoutTitleStore.set($projectDetailStore?.summary?.metadata.name ?? "");
 
-  let loadingSummary: boolean;
-  $: loadingSummary = isNullish($projectDetailStore.summary);
-  let loadingSwapState: boolean;
-  $: loadingSwapState = isNullish($projectDetailStore.swapCommitment);
-
-  // TODO(L2-838): if error redirect to launchpad and display error there
+  let loading: boolean;
+  $: loading = $snsSummariesStore.length === 0;
```
peterpeterparker, line 168: previously the swap commitment was also considered for the "loading" check. any particular reason why not considering it anymore?

frontend/src/routes/ProjectDetail.svelte
lmuntaner: I avoided the check inside `mapProjectDetail` this way. But this is an edge case, the check `isRoutePath({ path: AppPath.ProjectDetail, routePath: path }` already makes sure that there will be such a `rootCanisterId`. Or that at least it will be defined.

frontend/src/routes/ProjectDetail.svelte
lmuntaner: I thought one was enough, but I will add it.

lmuntaner: done

frontend/src/routes/ProjectDetail.svelte
peterpeterparker: oki doki, thx for the feedback

peterpeterparker (approved):
LGTM, Thx

PR #1269, opened 2022-08-23
style: redo toast max-height

peterpeterparker:
# Motivation

Redo max-height for toast content.

# Notes

Previously height was set on the all toast, I reused same value for the content so that one line still remains displayed in a compact way. Per extension the toast max height is then a bit bigger than before but it's fine in my opinion.

# Changes

- redo max height but set to toast content

# Screenshots

Capture d’écran 2022-08-23 à 11 10 53

### DIFF
--- a/frontend/src/lib/components/ui/Toast.svelte
+++ b/frontend/src/lib/components/ui/Toast.svelte
@@ -98,9 +98,6 @@
     align-items: flex-start;
     gap: var(--padding-1_5x);
 
-    // (>=3 lines x 1rem) + top/bottom paddings
-    height: calc(8.5 * var(--padding));
-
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/ui/Toast.svelte
```diff
@@ -90,6 +90,8 @@
 </div>
 
 <style lang="scss">
+  @use "@dfinity/gix-components/styles/mixins/text";
```
mstrasinskis, line 93: Is this still needed?

peterpeterparker: Good catch, thanks

mstrasinskis (approved):
LGTM, thanks

PR #1274, opened 2022-08-24
L2-702: Project Detail skeleton

lmuntaner:
# Motivation

User has a nice loading page while project is loading.

# Changes

* New SkeletonDetails component.
* Use new SkeletonDetails component and keep grid while loading in ProjectDetail.

# Tests

None. Only substituted a Spinner for a nicer loading UI.

### DIFF
--- a/frontend/src/lib/components/ui/SkeletonDetails.svelte
+++ b/frontend/src/lib/components/ui/SkeletonDetails.svelte
@@ -1,0 +1,31 @@
+<script lang="ts">
+  import SkeletonParagraph from "./SkeletonParagraph.svelte";
+  import SkeletonTitle from "./SkeletonTitle.svelte";
+</script>
+
+<div class="loader-title">
+  <SkeletonTitle />
+</div>
+<div class="loader-content">
+  <SkeletonParagraph />
+  <SkeletonParagraph />
+  <SkeletonParagraph />
+</div>
+
+<style lang="scss">
+  @use "@dfinity/gix-components/styles/mixins/media";
+  .loader-title {
+    width: 100%;
+    margin-bottom: var(--padding-2x);
+
+    @include media.min-width(medium) {
+      width: 50%;
+    }
+  }
+
+  .loader-content {
+    display: flex;
+    flex-direction: column;
+    gap: var(--padding);
+  }
+</style>
--- a/frontend/src/routes/ProjectDetail.svelte
+++ b/frontend/src/routes/ProjectDetail.svelte
@@ -16,7 +16,6 @@
   } from "../lib/services/sns.services";
   import { isRoutePath } from "../lib/utils/app-path.utils";
   import { snsSwapCommitmentsStore } from "../lib/stores/sns.store";
-  import Spinner from "../lib/components/ui/Spinner.svelte";
   import {
     PROJECT_DETAIL_CONTEXT_KEY,
     type ProjectDetailContext,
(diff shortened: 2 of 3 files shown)


lmuntaner:
(Sorry for the lame spam in the title, haha)

UX:

gif-skeleton-details

Screenshots:

Screenshot 2022-08-24 at 08 15 46
Screenshot 2022-08-24 at 08 15 56

lmuntaner:

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -184,7 +184,14 @@
   <div class="stretch-mobile">
     <!-- notFound redirects to launchpad but we show the spinner until redirection occurs -->
     {#if loading || notFound}
-      <Spinner />
+      <div class="content-grid">
+        <div class="content-a">
+          <SkeletonDetails />
+        </div>
+        <div class="content-b">
```
peterpeterparker, line 191: not sure we should display a skeleton for the `content-b` because the `ProjectStatusSection` might actually be empty - i.e. you display a skeleton but at the end nothing is displayed. That's why I would "just" display one skeleton in `content-a`. Did not checked the question with Mischa though.

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -184,7 +184,14 @@
   <div class="stretch-mobile">
     <!-- notFound redirects to launchpad but we show the spinner until redirection occurs -->
```
peterpeterparker, line 185: "...but we show the spinner..." not a spinner anymore 😉

frontend/src/lib/components/ui/SkeletonDetails.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
+  import SkeletonParagraph from "./SkeletonParagraph.svelte";
+  import SkeletonTitle from "./SkeletonTitle.svelte";
+</script>
+
+<div class="loader-title">
```
peterpeterparker, line 6: It's probably rather a `h1` with following style to match the exact size of the title that would be displayed once loaded.

```
h1 {
      margin: 0;
      line-height: var(--line-height-standard);
    }
```

Being said, above style and structure should be extracted to a mixin or class because we will use it everywhere. Not sure if we should just add a todo or a jira task about new design...

peterpeterparker: I'll take care of this in one of my PR since I have to extract some css too.

frontend/src/routes/ProjectDetail.svelte
lmuntaner: Mhm, maybe it makes sense then that each component has its own loader. I will refactor.

frontend/src/routes/ProjectDetail.svelte
lmuntaner: thanks!

frontend/src/lib/components/ui/SkeletonDetails.svelte
lmuntaner: great, thanks!

frontend/src/routes/ProjectDetail.svelte
lmuntaner: removed comment because the logic of spinner or no spinner is in each component.

frontend/src/routes/ProjectDetail.svelte
lmuntaner: I moved the logic of loader or no loader to each component.

lmuntaner:

frontend/src/routes/ProjectDetail.svelte
peterpeterparker: good idea

peterpeterparker (approved):
thx

PR #1275, opened 2022-08-24
Fake timer for tests

mstrasinskis:
# Motivation

Optimistic (failed to reproduce the problem locally) attempt to make tests more reliable (related to [the issue](https://github.com/dfinity/nns-dapp/runs/7979317538?check_suite_focus=true)).

# Changes

- set static system time for `dissolving time` spec
- switch to `nowInSeconds` to mimic the implementation
- extended `debounce` test


peterpeterparker (approved):
Cool, thanks

PR #1277, opened 2022-08-24
L2-833: Store filters in local storage

lmuntaner:
# Motivation

Proposals filters are persisted across browser refresh.

# Changes

* New util to create a store that gets stored in local storage "writableStored".
* Use writableStored in themStore
* Use writableStored in proposalsFiltersStore.

# Tests

* Add tests for writableStored
* Mock localStorage in jest.

### DIFF
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -87,7 +87,7 @@
         "(prefers-color-scheme: dark)"
       ).matches;
       const currentTheme =
-        localStorage.nnsTheme === null
+        localStorage.nnsTheme === null || localStorage.nnsTheme === undefined
           ? undefined
           : JSON.parse(localStorage.nnsTheme);
 
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -87,11 +87,11 @@
         "(prefers-color-scheme: dark)"
       ).matches;
       const currentTheme =
-        localStorage.theme !== null
-          ? JSON.parse(localStorage.theme)
-          : undefined;
+        localStorage.nnsTheme === null
+          ? undefined
+          : JSON.parse(localStorage.nnsTheme);
 
-      // `theme` must match storeLocalStorageKey.Theme
+      // Attribute is the value of THEME_ATTRIBUTE constant
       document.documentElement.setAttribute(
         "theme",
         currentTheme ?? (isDarkPreferred ? "dark" : "light")
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -86,11 +86,15 @@
       const isDarkPreferred = window.matchMedia(
         "(prefers-color-scheme: dark)"
       ).matches;
+      const currentTheme =
+        localStorage.theme !== null
+          ? JSON.parse(localStorage.theme)
+          : undefined;
 
       // `theme` must match storeLocalStorageKey.Theme
       document.documentElement.setAttribute(
         "theme",
-        localStorage.theme ?? (isDarkPreferred ? "dark" : "light")
+        currentTheme ?? (isDarkPreferred ? "dark" : "light")
       );
     </script>
 
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -87,6 +87,7 @@
         "(prefers-color-scheme: dark)"
       ).matches;
 
+      // `theme` must match storeLocalStorageKey.Theme
       document.documentElement.setAttribute(
         "theme",
         localStorage.theme ?? (isDarkPreferred ? "dark" : "light")
--- a/frontend/src/index.html
+++ b/frontend/src/index.html
@@ -83,19 +83,26 @@
 
     <!-- Init theme as fast as possible -->
     <script>
-      const isDarkPreferred = window.matchMedia(
-        "(prefers-color-scheme: dark)"
-      ).matches;
-      const currentTheme =
-        localStorage.nnsTheme === null || localStorage.nnsTheme === undefined
-          ? undefined
-          : JSON.parse(localStorage.nnsTheme);
+      try {
+        const isDarkPreferred = window.matchMedia(
+          "(prefers-color-scheme: dark)"
+        ).matches;
+        const currentTheme =
+          localStorage.nnsTheme === null ||
+          localStorage.nnsTheme === undefined ||
+          localStorage.nnsTheme === ""
+            ? undefined
+            : JSON.parse(localStorage.nnsTheme);
 
-      // Attribute is the value of THEME_ATTRIBUTE constant
-      document.documentElement.setAttribute(
-        "theme",
-        currentTheme ?? (isDarkPreferred ? "dark" : "light")
-      );
+        // Attribute is the value of THEME_ATTRIBUTE constant
+        document.documentElement.setAttribute(
+          "theme",
+          currentTheme ?? (isDarkPreferred ? "dark" : "light")
+        );
+      } catch (error) {
+        console.log("Error initializing theme");
+        console.log(error);
+      }
     </script>
 
     <script>
--- a/frontend/src/lib/stores/proposals.store.ts
+++ b/frontend/src/lib/stores/proposals.store.ts
@@ -9,6 +9,7 @@ import type {
 } from "@dfinity/nns";
 import { writable } from "svelte/store";
 import { DEFAULT_PROPOSALS_FILTERS } from "../constants/proposals.constants";
+import { storeLocalStorageKey } from "../constants/stores.constants";
 import {
   concatenateUniqueProposals,
   excludeProposals,
@@ -16,6 +17,7 @@ import {
   replaceAndConcatenateProposals,
   replaceProposals,
 } from "../utils/proposals.utils";
+import { writableStored } from "./writable-stored";
 
 export interface ProposalsFiltersStore {
   topics: Topic[];
@@ -124,9 +126,10 @@ const initProposalsStore = () => {
  *
  */
 const initProposalsFiltersStore = () => {
-  const { subscribe, update, set } = writable<ProposalsFiltersStore>(
-    DEFAULT_PROPOSALS_FILTERS
-  );
+  const { subscribe, update, set } = writableStored<ProposalsFiltersStore>({
+    key: storeLocalStorageKey.ProposalFilters,
+    defaultValue: DEFAULT_PROPOSALS_FILTERS,
+  });
 
   return {
     subscribe,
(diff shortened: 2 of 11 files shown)


lmuntaner:

lmuntaner:
Btw, filters are also persisted after logging out. Since it's also anonymized data I thought that was ok. What do you think?

frontend/src/lib/stores/writable-stored.ts
```diff
@@ -0,0 +1,41 @@
...
+}: {
+  key: storeLocalStorageKey;
+  defaultValue: T;
+}) => {
+  const getInitialValue = (): T => {
+    // Do not break UI if local storage fails
+    try {
+      const storedValue = localStorage.getItem(key);
+      // `theme` stored "undefined" string which is not a valid JSON string.
+      if (
```
peterpeterparker, line 16: nth: the `if` can be spared since there is a `try catch`

```
try {
   return JSON.parse(storedValue) as T ?? defaultValue;
} catch (err) {
   // We ignore the errors and fallback to default value
   return defaultValue;
}
```

frontend/src/lib/stores/writable-stored.ts
```diff
@@ -0,0 +1,41 @@
...
+      }
+    } catch (error) {
+      console.error(error);
+    }
+    return defaultValue;
+  };
+
+  const store = writable<T>(getInitialValue());
+
+  store.subscribe((store: T) => {
```
peterpeterparker, line 31: can we handle unsubscription?

right now the store is use only globally but in the future if someone would use contextually it might leak

frontend/src/lib/stores/writable-stored.ts
```diff
@@ -0,0 +1,41 @@
...
+    // Do not break UI if local storage fails
```
peterpeterparker, line 32: in which case can it fails?

frontend/src/lib/stores/theme.store.ts
```diff
@@ -1,17 +1,24 @@
...
+
+  subscribe((theme: Theme) => {
+    applyTheme({ theme });
+  });
 
   return {
     subscribe,
 
     select: (theme: Theme) => {
-      applyTheme({ theme, preserve: true });
```
peterpeterparker, line 14: the `preserve` option disappeared, functionally still works the same?

lmuntaner: Logic to store in local storage it was done inside the `applyTheme`. When `preseve` was set to true.

Yet, now the logic of storing the theme is at the store level with `writableStored`. We don't need the `preserve` and `applyTheme` does only that: applying the theme.

frontend/src/lib/stores/writable-stored.ts
lmuntaner: sure, that looks cool

lmuntaner: damn, TS doesn't like that... JSON.parse expects a string...

frontend/src/lib/stores/writable-stored.ts
lmuntaner: I don't know, some edge case, but I'd rather catch it and ignore it than fail the whole dapp.

frontend/src/lib/stores/writable-stored.ts
lmuntaner: sure, I can expose it for when is needed.

lmuntaner: done

lmuntaner:

lmuntaner:

frontend/src/lib/stores/writable-stored.ts
peterpeterparker: ah damn, ok thanks for the feedback

(just for yolo)

```
try {
   return JSON.parse(storedValue ?? JSON.stringify(defaultValue)) as T;
} catch (err) {
   // We ignore the errors and fallback to default value
   return defaultValue;
}
```

frontend/src/lib/stores/theme.store.ts
peterpeterparker: Thanks for the explanation. Can it be the reason of the flash I discovered while testing?

frontend/src/lib/stores/writable-stored.ts
lmuntaner: I log the error, is not silent. I just prevent the app from crashing.

frontend/src/lib/stores/theme.store.ts
lmuntaner: It could, I'll look into that.

lmuntaner: solved

lmuntaner:

frontend/src/lib/stores/proposals.store.ts
```diff
@@ -124,9 +126,10 @@ const initProposalsStore = () => {
  *
  */
 const initProposalsFiltersStore = () => {
-  const { subscribe, update, set } = writable<ProposalsFiltersStore>(
-    DEFAULT_PROPOSALS_FILTERS
-  );
+  const { subscribe, update, set } = writableStored<ProposalsFiltersStore>({
```
peterpeterparker, line 129: not sure we need to save the key `lastAppliedFilter` of `ProposalsFiltersStore` in the local storage, what you think?

frontend/src/lib/stores/writable-stored.ts
```diff
@@ -0,0 +1,48 @@
+import { writable, type Writable } from "svelte/store";
+import type { storeLocalStorageKey } from "../constants/stores.constants";
+
+type StoredWritable<T> = Writable<T> & {
```
peterpeterparker, line 4: instead of a custom type, would `& Unsubscriber` would work too?

```
import type { Unsubscriber } from "svelte/store";
```

frontend/src/lib/stores/writable-stored.ts
peterpeterparker: With silent I meant that we will never notice it unless we face it while developing and users, particularly on mobile devices, will most probably never notice it as well

But like I said, can live with it

frontend/src/lib/utils/theme.utils.ts
```diff
@@ -2,13 +2,17 @@ import { Theme } from "../types/theme";
...
+export const THEME_ATTRIBUTE = "theme";
+
 export const initTheme = (): Theme => {
   // Jest NodeJS environment has no document
   if (isNode()) {
     return Theme.DARK;
   }
 
-  const theme: string | null = document.documentElement.getAttribute("theme");
+  // This is set in the index.html file
```
peterpeterparker, line 13: maybe `// Initial attribute value is set in the index.html`

frontend/src/lib/stores/proposals.store.ts
lmuntaner: Maybe in another future we might add the feature to skip keys in the `writableStored`. But for now, I don't see it necessary.

It increases complexity and it doesn't add that much. In the end the extra storage it's not as big. And it doesn't grow.

frontend/src/index.html
```diff
@@ -86,10 +86,15 @@
...
+      const currentTheme =
+        localStorage.nnsTheme === null
+          ? undefined
+          : JSON.parse(localStorage.nnsTheme);
 
+      // Attribute is the value of THEME_ATTRIBUTE constant
       document.documentElement.setAttribute(
         "theme",
-        localStorage.theme ?? (isDarkPreferred ? "dark" : "light")
+        currentTheme ?? (isDarkPreferred ? "dark" : "light")
```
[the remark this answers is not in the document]
lmuntaner: I think the one using the variable is more readable.

frontend/src/lib/utils/theme.utils.ts
lmuntaner: ok

frontend/src/lib/stores/writable-stored.ts
lmuntaner: yep, it works

frontend/src/lib/utils/theme.utils.ts
lmuntaner: done

frontend/src/lib/stores/writable-stored.ts
lmuntaner: done

frontend/src/index.html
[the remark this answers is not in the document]
peterpeterparker: now that we have added another clause in the `if` that "interprets" bot `null` and `undefined` into `undefined`, I have to say I kind of rather like the inline since it handles both - i.e. it isn't now much less readable and handles both case. but sure can live with it.

frontend/src/lib/stores/proposals.store.ts
peterpeterparker: works for me

frontend/src/lib/stores/writable-stored.ts
```diff
@@ -0,0 +1,48 @@
+import { writable, type Unsubscriber, type Writable } from "svelte/store";
+import type { storeLocalStorageKey } from "../constants/stores.constants";
+
+type StoredWritable<T> = Writable<T> & {
```
peterpeterparker, line 4: dumb question: store is named `writableStored` but interface `StoredWritable`. should the interface actually be named `WritableStored` for consistency reason?

frontend/src/index.html
```diff
@@ -86,10 +86,15 @@
       const isDarkPreferred = window.matchMedia(
         "(prefers-color-scheme: dark)"
       ).matches;
+      const currentTheme =
+        localStorage.nnsTheme === null || localStorage.nnsTheme === undefined
```
peterpeterparker, line 90: can you also prevent empty string `""` that would crash an error in `JSON.parse` - just in case sh*t happens? or shall we add a try catch?

peterpeterparker (approved):
LGTM, thanks. Added few minor comments if you can have a look before merge.

frontend/src/lib/stores/writable-stored.ts
lmuntaner: yes, makes sense

frontend/src/index.html
lmuntaner: good point, I'll add a try catch and also check for `""`

lmuntaner: Shame that `??` does not consider empty string as falsy:
```
> "" ?? 'hi'
''
```

frontend/src/lib/stores/writable-stored.ts
lmuntaner: done

frontend/src/index.html
lmuntaner: done

PR #1279, opened 2022-08-25
L2-898: Fix debugger

lmuntaner:
# Motivation

Debugger is not working.

# Changes

* Add an empty store for those stores that are initialized in a component.
* Add projectDetailStore to the debug store.

# Tests

Not applicable.

### DIFF
--- a/frontend/src/lib/services/debug.services.ts
+++ b/frontend/src/lib/services/debug.services.ts
@@ -13,6 +13,8 @@ import {
   anonymizeKnownNeuron,
   anonymizeNeuronInfo,
   anonymizeProposal,
+  anonymizeSnsSummary,
+  anonymizeSnsSwapCommitment,
   anonymizeTransaction,
   cutAndAnonymize,
 } from "../utils/anonymize.utils";
@@ -132,6 +134,7 @@ const anonymiseStoreState = async () => {
     transaction,
     selectedAccount,
     selectedProposal,
+    selectedProject,
   } = get(debugStore);
 
   return {
@@ -189,6 +192,12 @@ const anonymiseStoreState = async () => {
             transaction,
             account: selectedAccount?.account,
           })
+      ),
+    },
+    selectedProject: {
+      summary: await anonymizeSnsSummary(selectedProject.summary),
+      swapCommitment: await anonymizeSnsSwapCommitment(
+        selectedProject.swapCommitment
       ),
     },
   };
--- a/frontend/src/lib/utils/anonymize.utils.ts
+++ b/frontend/src/lib/utils/anonymize.utils.ts
@@ -8,10 +8,22 @@ import {
   type ProposalInfo,
 } from "@dfinity/nns";
 import type {
+  SnsSwapBuyerState,
+  SnsSwapDerivedState,
+  SnsSwapState,
+} from "@dfinity/sns";
+import type {
   CanisterDetails,
   Transaction,
 } from "../canisters/nns-dapp/nns-dapp.types";
 import type { Account } from "../types/account";
+import type {
+  SnsSummary,
+  SnsSummaryMetadata,
+  SnsSummarySwap,
+  SnsSwapCommitment,
+  SnsTokenMetadata,
+} from "../types/sns";
 import { digestText } from "./dev.utils";
 import { mapTransaction } from "./transactions.utils";
 import { isNullish, mapPromises, nonNullish } from "./utils";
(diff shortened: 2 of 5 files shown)


frontend/src/lib/services/debug.services.ts
```diff
@@ -206,7 +215,9 @@ export const generateDebugLog = async (logType: LogType) => {
   const date = new Date().toJSON().split(".")[0].replace(/:/g, "-");
 
   if (logType === LogType.ConsoleOriginalObject) {
-    console.log(date, state);
+    console.log(date);
+    // logging the sate alone makes it explorable in the console
```
peterpeterparker, line 219: - "sate"? state I guess

- not sure I understand the goal? it's just console.log, so fine for me, but just don't get the issue

frontend/src/lib/utils/anonymize.utils.ts
```diff
@@ -274,3 +286,81 @@ export const anonymizeProposal = async (
     ballots: await mapPromises(ballots, anonymizeBallot),
   };
 };
+
+const anonymiseBuyer = async ([buyer, state]) => [
```
peterpeterparker, line 290: - "anonymise"? anonymize no?

frontend/src/lib/utils/anonymize.utils.ts
```diff
@@ -274,3 +286,81 @@ export const anonymizeProposal = async (
...
+  buyer,
+  {
+    ...state,
+    amount_icp_e8s: (await anonymizeAmount(state.amount_icp_e8s)) ?? BigInt(0),
+    amount_sns_e8s: (await anonymizeAmount(state.amount_sns_e8s)) ?? BigInt(0),
+  },
+];
+
+type AnonymizedSnsSummary = {
```
peterpeterparker, line 299: you can probably reuse some Sns types with typescript utility to create the type, no?

frontend/src/lib/utils/anonymize.utils.ts
```diff
@@ -274,3 +286,81 @@ export const anonymizeProposal = async (
     ballots: await mapPromises(ballots, anonymizeBallot),
   };
 };
+
+const anonymiseBuyer = async ([buyer, state]) => [
+  buyer,
+  {
+    ...state,
+    amount_icp_e8s: (await anonymizeAmount(state.amount_icp_e8s)) ?? BigInt(0),
```
peterpeterparker, line 294: dumb question but, aren't these information public anyway?

frontend/src/lib/services/debug.services.ts
lmuntaner: Just to make it easier to explore the object withot the date next to it. the comment makes sense in the next actually. where I log the object, not the stringified version.

But maybe there was a reason for it. That's why I wanted @mstrasinskis 's review on this PR.

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: yes, sorry, we use "z" here.

frontend/src/lib/services/debug.services.ts
lmuntaner: I changed the comments.

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: I saw all the amounts being anonymized, something I wanted @mstrasinskis to review.

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: maybe, but I felt writing the type like this made it more readable.

frontend/src/routes/ProjectDetail.svelte
```diff
@@ -73,8 +74,9 @@
     summary: null,
     swapCommitment: null,
   });
+  console.log("adding to debug store");
+  debugSelectedProjectStore(projectDetailStore);
 
-  // TODO: add projectDetailStore to debug store
```
mstrasinskis, line 77: Thanks 👍

frontend/src/lib/utils/anonymize.utils.ts
peterpeterparker: if we use the utility we can easier get compilation error if the definition would ever change

frontend/src/lib/utils/anonymize.utils.ts
peterpeterparker: Probably worth to ask, not against anonymized data but if we can spare code, not against neither ;)

frontend/src/lib/utils/anonymize.utils.ts
```diff
@@ -274,3 +286,81 @@ export const anonymizeProposal = async (
...
+          ...originalSummary.swap.state,
+          buyers: anonymizedBuyers,
+        } as SnsSwapState,
+      },
+      derived: {
+        ...originalSummary.derived,
+        buyer_total_icp_e8s:
+          (await anonymizeAmount(
+            originalSummary.derived.buyer_total_icp_e8s
+          )) ?? BigInt(0),
```
mstrasinskis, line 332: Is it possible to avoid using the fallback `?? BigInt(0)`?

frontend/src/lib/services/debug.services.ts
```diff
@@ -220,6 +231,8 @@ export const generateDebugLog = async (logType: LogType) => {
       filename: `${date}_nns-local-state.json`,
     });
   } else {
-    console.log(date, stringifiedState);
+    console.log(date);
```
peterpeterparker, line 234: this is solved in https://github.com/dfinity/nns-dapp/pull/1284

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: done

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: Won't I get it also like this?

The helper expects `SnsSummary` type, if that changes, then the compilation will fail if the changes are not backwards compatible.

I can use `{ [key in keyof Required<SnsSummary>]: unknown }` like in the others. Is that preferred?

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: I would need to change the type. `buyer_total_icp_e8s` is mandatory and `anonymizeAmount` might return undefined. I can cast it, or change `anonymizeAmount`, what do you prefer?

frontend/src/lib/services/debug.services.ts
lmuntaner: I will revert then.

lmuntaner: done

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: @mstrasinskis what do you think?

lmuntaner:
@mstrasinskis please review

frontend/src/lib/utils/anonymize.utils.ts
mstrasinskis: The anonymize tools are dynamically imported. The idea is to remove connection between the data and the user in case of sharing the data. But agree, the current approach could be less imperative.

mstrasinskis (approved):
LGTM

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: ok, makes sense. So we keep it.

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: sounds good

frontend/src/lib/utils/anonymize.utils.ts
peterpeterparker: works for me

frontend/src/lib/utils/anonymize.utils.ts
lmuntaner: done

PR #1280, opened 2022-08-25
Fix/legacy proposal title layout

mstrasinskis:
# Motivation

Fix the proposal status layout (in single line).

# Changes

- updated proposal style for legacy version

## before

image

## after
image


frontend/src/lib/themes/legacy.scss
```diff
@@ -19,5 +19,10 @@ main.legacy {
       flex-wrap: wrap;
       column-gap: var(--padding);
     }
+    
+    &[data-tid="proposal-card"] .meta {
```
mstrasinskis, line 23: The abusing of the `data-tid` is not the cleanest solution, but probably enough for the temporary workaround. There is another possibility to extend the Card component with the class prop, but

peterpeterparker (approved):
Thx

frontend/src/lib/themes/legacy.scss
mstrasinskis: The style block should be removed together  with the file (because of the UI update) => no need for special treatment.

PR #1281, opened 2022-08-25
feat: script loader

peterpeterparker:
# Motivation

As in II, use a script loader for main script.

# Changes

- add a script that load `main.js` in `index.html`
- adapt csp accordingly with hashes instead of nonce


nmattia (approved):
Nice, thanks! Rubber stamp

PR #1282, opened 2022-08-25
refactor: extract i18n governance keys to dedicated file

peterpeterparker:
# Motivation

Share the same labels and descriptions for NNS-dapp and the dashboard in a single place so that both dapps can benefits automatically from the textual improvements.

# What's next

Discussed with dashboard team, the next step will be to unify and rename some keys to bring consistency in the naming of the information what can be useful for NNS-dapp, dashboard but also the community.

This will need some breaking changes in ic-js. I'll do.

# Changes

- extract proposal topics, rewards, status and types labels and descriptions to a dedicated `en.governance.js`
- adapt script - that generates `i18n` keys - and store accordingly


frontend/src/lib/stores/i18n.ts
```diff
@@ -1,7 +1,9 @@
 import { readable } from "svelte/store";
+import governance from "../i18n/en.governance.json";
 import en from "../i18n/en.json";
 
 export const i18n = readable<I18n>({
   lang: "en",
   ...en,
+  ...governance,
```
lmuntaner, line 8: We would need to be careful to not use the same keys in `en` and `governance`. Otherwise the `en` will be removed.

Does it make sense to at least add a todo in the `generate` function in `i18n.types.js` to check for duplicated keys and raise an error then?

lmuntaner (approved):
Minor comment, approving already.

I like the idea!

frontend/src/lib/stores/i18n.ts
peterpeterparker: sure we can do that, added an assertion in the script

PR #1283, opened 2022-08-25
feat: dompurify as chunk instead of lazy

peterpeterparker:
# Motivation

Instead of lazy loading dompurify as a js assets, this PR adds it as a chunked library. We are now using the sanitizer in more places, the probability that the user will have to download it is bigger, therefore the explicit lazy loading isn't that necessary anymore and we can rely on chunks. In addition, it makes the code imperative which can be easier to introduce the component `<HTML />` we briefly commented in one of my last PR.

# Changes

- replace dompurify js assets with dependency
- adapt import and sanitizer function accordingly
- update jest to make dompurify supported for nodejs tests


peterpeterparker:
what ya think @mstrasinskis ?

frontend/src/lib/utils/html.utils.ts
```diff
@@ -65,29 +66,16 @@ export const markdownToHTML = async (): Promise<(text: string) => string> => {
...
-  return convertMarkdownToHTML(sanitizeText(text ?? ""));
+  const convertMarkdownToHTML = await markdownToHTML();
+  return convertMarkdownToHTML(sanitize(text ?? ""));
 };
 
-export const sanitizeHTML = async (text: string): Promise<string> => {
-  const sanitizeText = await sanitize();
-  return sanitizeText(text);
+export const sanitize = (text: string): string => {
+  const isBrowser = typeof window !== "undefined";
```
mstrasinskis, line 79: Not against the current solution, just a reminder that there is `isNode` function available ;)

mstrasinskis (approved):
LGTM. Thank you, totally makes sense.

frontend/src/lib/utils/html.utils.ts
peterpeterparker: Ah excellent! Replaced `isBrowser` with `!isNode` thanks!

PR #1284, opened 2022-08-26
feat: console debug instead of log for debugging

peterpeterparker:
# Motivation

Hide debug information from the default browser debugger console settings.

# Note

While debugging another issue with `console.log`, I struggled distinguish my outputs from these information. Therefore as we already discussed your idea @mstrasinskis, I went on implementing it to help me debug my issue.

# Changes

- `console.debug` instead of `console.log`
- use `logWithTimestamp`  for object too
- `console.error` instead of `console.log` for polling error


mstrasinskis (approved):
LGTM. Thanks

PR #1287, opened 2022-08-26
fix: proposals infinite scroll for new design

peterpeterparker:
# Motivation

The new grid design is not compatible with the existing infinite scrolling approach. In addition, hiding cards with CSS to achieve the feature wasn't that ideal from an accessibility perspective.

# Notes

- the infinite scroll observer has to be modified to observe the last element of the "list" instead of observing an element that is displayed after x%
- to prevent issue, we add a new option to disable the infinite scroll when there is no more data to fetch
- the order of the proposals that were added to the store were not maintained between query and update results which  lead to some ui shift that had for end effect to trigger the infinite scrolling which triggered new searches with incorrect proposal id 

e.g.

query results -> store: id 10, 9, 8, 7, 6
trigger, next search index: 6
update results -> store: 8, 7, 6, 10, 9
repaint, trigger, next search index: 9
query results  -> store: : 8, 7, 6, 10, 9, 5, 4, 3, 2, 1
update results -> store: messy store

I could have fixed the issue but, for the proposal store, the order is actually not relevant and fixing it would not prevent future issue. Therefore I created a derived store that secure the order.

# Future development

Even though it is solve, as currently, the ui can become a bit shaky depending of the ui rerender. I ask my self if we should not, when update calls result matches query calls, not re-render the ui and this even though we use keyed Svelte each. Not for now but, for the future.

# PR

- [ ] need `gix-component` PR https://github.com/dfinity/gix-components/pull/45

# Changes

- new `sortedProposals` derived store
- remove cards from DOM instead of hiding it to fulfill specific client side proposal's filter too
- style voting filters container
- display spinner when more proposals need to be loaded and skeleton when proposals are fetched for the first time
- disable infinite scroll when nothing else to fetch
- refactor `replaceAndConcatenateProposals` to filter existing proposals


frontend/src/lib/services/proposals.services.ts
```diff
@@ -74,20 +74,34 @@ export const listProposals = async (): Promise<void> => {
...
+ * @param {ProposalId | undefined} params.beforeProposal Pagination starting proposal. Undefined for first results
+ * @param {(paginationOver: boolean) => void;} params.loadFinished Triggered when the loading is over. `paginationOver` equals `true` if all pages of the list have been queried.
+ */
 export const listNextProposals = async ({
   beforeProposal,
+  loadFinished,
 }: {
   beforeProposal: ProposalId | undefined;
+  loadFinished: (params: {
+    paginationOver: boolean;
```
mstrasinskis, line 89: Nice 👍

mstrasinskis (approved):
LGTM. Thank you!

PR #1288, opened 2022-08-29
Release: 2022-08-30

lmuntaner:
# Motivation

Release candidate for 30th of August of 2022

# Changes

* Changelog


lmuntaner:
Closing, we found an issue with the new agent-js version in iPhone when logging in. This needs to be fixed or we need to revert to an older version of agent-js.

PR #1290, opened 2022-08-29
feat: tooltip top position

mstrasinskis:
# Motivation

Removes redundant scrollbars and improves the toolbar position:

## Extra scrollbars
image

## Suboptimal tooltip position on mobile
image

# Changes

- add `top` position support to the `Tooltip` component

# Screenshots

## After
image

image

image


frontend/src/lib/components/neurons/NnsNeuronsFooter.svelte
```diff
@@ -37,7 +37,11 @@
         >{$i18n.neurons.merge_neurons}</button
       >
     {:else}
-      <Tooltip id="merge-neurons-info" text={$i18n.neurons.need_two_to_merge}>
+      <Tooltip
```
mstrasinskis, line 40: @peterpeterparker, not sure about adding `overflow: hidden` to the `Footer` component because it makes the footer less flexible. Currently the issue is "solved", but w/o overflow anything could come later.

peterpeterparker: We could on the other side, the `Footer` is deprecated anyway if I get it right

frontend/src/lib/components/ui/Tooltip.svelte
```diff
@@ -116,6 +118,12 @@
       white-space: nowrap;
     }
 
+    &.top {
+      bottom: unset;
+      top: calc(-1 * var(--padding));
+      transform: translate(var(--tooltip-transform-x, -50%), -100%);
```
peterpeterparker, line 124: `-100%` is correct here? not just the y position that should be fixed?

frontend/src/lib/components/neurons/NnsNeuronsFooter.svelte
peterpeterparker: So we are good with that 👍

frontend/src/lib/components/ui/Tooltip.svelte
peterpeterparker: I mismatched x and y in my early morning review 🤦‍♂️
all good thx

peterpeterparker (approved):
LGTM; thx

frontend/src/lib/components/ui/Tooltip.svelte
mstrasinskis: We need to shift it up. As discussed in slack.

PR #1294, opened 2022-08-30
L2-972: New TransactionModal component

lmuntaner:
# Motivation

Preparing for the redesign, new modal to be used in all transactions.

# Changes

* New TransactionModal component which uses Form and Review.
* Use new TransactionModal in ParticipateSwapModal.
* Move some i18n keys to accounts from project detail key.

# Tests

* Test new TransactionModal with different cases.
* Test ParticipateSwapModal only for specific cases.
* Fix broken test.


lmuntaner:
No change in UX. But just in case:

https://user-images.githubusercontent.com/4550653/187466756-f5d673c5-8b73-469a-a70d-d024e7a90ff5.mov

lmuntaner:

As agreed in https://www.notion.so/New-Transactions-Modal-3d0f7a1a315d4de6858cbb735fdc8fe0

I created a new TransactionModal to be used in all the transactions.

For now I only added the functionality needed for the ParticipateSwapModal. More will come gradually.

peterpeterparker:
can you maybe prefix the new `Modal.svelte`, `Form.svelte` and `Review.svelte` file names? Particularly because when I would perform a global search, I would find two `Modal.svelte` ( we already have one component with such a name) and with `Form` I might understand it is a `<form />` UI component.

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
```diff
@@ -75,35 +82,117 @@
...
-      {amount}
-      on:nnsBack={goBack}
-      on:nnsClose
-    />
-  {/if}
-</WizardModal>
+    <svelte:fragment slot="title"
+      >{title ?? $i18n.sns_project_detail.participate}</svelte:fragment
+    >
+    <svelte:fragment slot="additional-info-form">
```
peterpeterparker, line 126: would it make sense to extract the content of the two slots `additional-info-xxx` in separate components? From the review perspective at least, it would make it easier to read

frontend/src/lib/types/transaction.context.ts
```diff
@@ -8,6 +8,12 @@ export interface TransactionStore {
   amount: ICP | undefined;
 }
 
+export type NewTransaction = {
```
peterpeterparker, line 11: should `NewTransaction` inherits types of `TransactionStore` or the contrary? Feels like there is some dubplication there, maybe inheriting, picking types or else would be nice?

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
lmuntaner: sure, but would you mind if I create a folder for the ParticipateSwapModal then, where I add this smaller components?

It's different whey you build new components for readibility or for reusing. This could be reused, of course, but these kind of small components can explode the number of components within a folder, making the others harder to find.

frontend/src/lib/types/transaction.context.ts
lmuntaner: The idea is to remove the TransactionStore in the end. We will remove the old NewTransactionModal. Therefore, I kept it separate.

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
peterpeterparker: > sure, but would you mind if I create a folder for the ParticipateSwapModal then, where I add this smaller components?

sure I don't mind

frontend/src/lib/types/transaction.context.ts
peterpeterparker: Maybe adding a `@deprecated` comment then?

frontend/src/lib/modals/sns/ParticipateSwapModal.svelte
lmuntaner: done

lmuntaner:

I added the prefix to the files. I will do so always.

frontend/src/lib/types/transaction.context.ts
lmuntaner: it's the one being used at the moment within NewTransactionModal, so I would have to add deprecated to the whole component not just there.

This is similar to what you are doing now with the proposal page with a feature flag:

```
<main class={VOTING_UI}>
  {#if VOTING_UI === "modern"}
    <ProposalModern {neuronsReady} />
  {:else}
    <ProposalLegacy {neuronsReady} />
  {/if}
</main>
```

Yet, I will not use a feature flag, but change altogether the transaction modal in my next issue.

I don't think we need to set as deprecated everything that we are planning to substitute in the short term. It's quite some effort.

peterpeterparker: e.g. not sure the `@deprecated` annotation can be use on all Svelte component, can it be? so I assumed it cannot that's why instead I created a JIRA task and added a `TODO: to be deleted` in these components

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
```diff
@@ -0,0 +1,134 @@
...
+  import type { Account } from "../../../types/account";
+  import { replacePlaceholders } from "../../../utils/i18n.utils";
+  import { convertNumberToICP } from "../../../utils/icp.utils";
+  import { valueSpan } from "../../../utils/utils";
+  import Icp from "../../../components/ic/ICP.svelte";
+  import KeyValuePair from "../../../components/ui/KeyValuePair.svelte";
+  import { sanitize } from "../../../utils/html.utils";
+  import type { NewTransaction } from "../../../types/transaction.context";
+
+  export let sourceAccount: Account;
```
peterpeterparker, line 18: here you don't use a single property `NewTransaction` for re-usability reason? (strictly a question, I'm fine with three props)

frontend/src/lib/modals/sns/SwapModal/AdditionalInfoReview.svelte
```diff
@@ -0,0 +1,40 @@
...
+      display: grid;
+      grid-template-columns: auto 1fr;
+      gap: var(--padding-2x);
+
+      .icon {
+        color: var(--warning-emphasis);
+      }
+    }
+
+    :global(label) {
```
peterpeterparker, line 36: we already add that global style?

if yes, all good - it's refactoring
if no, is there a way to avoid it?

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
```diff
@@ -0,0 +1,145 @@
...
+    })
+  );
+
+  let minCommitment: ICP;
+  $: minCommitment = ICP.fromE8s(
+    userHasParticipatedToSwap ? BigInt(0) : init.min_participant_icp_e8s
+  );
+
+  let accepted: boolean = false;
+  const toggelAccept = () => (accepted = !accepted);
```
peterpeterparker, line 81: should this finds place in the related component where the checkbox is?

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
```diff
@@ -0,0 +1,134 @@
...
+  export let amount: number;
+  export let destinationAddress: string;
+  export let disabled: boolean;
```
peterpeterparker, line 21: would it make sense to rename this property to `disableSubmit` as well?

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
lmuntaner: Good point, I can use the type as a prop to make sure we that here and then is sent when I trigger the event.

I'll do that.

frontend/src/lib/modals/sns/SwapModal/AdditionalInfoReview.svelte
lmuntaner: this is the preferred way to change the order in the Checkbox... By default, the checkbox is at the right.

I could pass a prop instead as well.

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
lmuntaner: I still need the `accepted` here, but maybe is better. Sounds good to me.

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
lmuntaner: sure

frontend/src/lib/modals/sns/SwapModal/AdditionalInfoReview.svelte
peterpeterparker: Ideally a css variable but if not possible, then let's stick to global. was just asking because I think it's good when we can to avoid to use the global selector. Thanks for the feedback.

lmuntaner: I couldn't agree more, but we use it 56 times in 32 files at the moment.

frontend/src/lib/types/transaction.context.ts
lmuntaner: I'll do that

lmuntaner: done

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
lmuntaner: done

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
lmuntaner: done

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
lmuntaner: done

lmuntaner:
@peterpeterparker ready for another review!

peterpeterparker (approved):
LGTM, thx

PR #1295, opened 2022-08-31
fix: iOS login blocked

peterpeterparker:
# Motivation

Long story short: on iOS, click on the "Login" have no effect 👉 user cannot sign-in on iOS with the current development version which uses agent-js `v0.13.x` (mainnet uses currently `v0.12.x` which is fine).

More details and context in related agent-js issue https://github.com/dfinity/agent-js/pull/618

# Changes

- keep `authClient` object in memory to not call `authClient.create` before `authClient.login` - i.e. avoid any async callback within the same function that should ultimately call `window.open`

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -1,5 +1,5 @@
 {
-  "dfx": "0.11.1",
+  "dfx": "0.11.2",
   "canisters": {
     "nns-governance": {
       "type": "custom",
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
LGTM, thanks!

mstrasinskis (approved):
LGTM, thank you!

PR #1297, opened 2022-09-01
Fix: Max button in sale participation

lmuntaner:
# Motivation

Max button in sale participation didn't consider the remaining amount of the project.

# Changes

* Add maxAmount parameter to TransactionModal.
* Use new maxAmount when calculating the max of the transaction amount.
* Rename and refactor maxICP to use e8s to getMaxTransactionAmount
* New project util to expose the remaining amount of a project projectRemainingAmount

# Tests

* New tests for project util projectRemainingAmount
* Create new test cases for refactored util getMaxTransactionAmount

### DIFF
--- a/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
+++ b/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
@@ -4,15 +4,14 @@
   import { i18n } from "../../../stores/i18n";
   import {
     mainTransactionFeeStoreAsIcp,
-    mainTransactionFeeStore,
+    transactionsFeesStore,
   } from "../../../stores/transaction-fees.store";
   import type { Account } from "../../../types/account";
   import { InvalidAmountError } from "../../../types/neurons.errors";
   import { assertEnoughAccountFunds } from "../../../utils/accounts.utils";
   import {
     convertNumberToICP,
-    maxE8sToNumber,
-    minE8s,
+    getMaxTransactionAmount,
   } from "../../../utils/icp.utils";
   import SelectAccountDropdown from "../../../components/accounts/SelectAccountDropdown.svelte";
   import IcpComponent from "../../../components/ic/ICP.svelte";
@@ -27,12 +26,10 @@
   export let maxAmount: bigint | undefined = undefined;
 
   let max: number = 0;
-  $: max = maxE8sToNumber({
-    e8s: minE8s(
-      selectedAccount?.balance.toE8s(),
-      (maxAmount ?? BigInt(0)) + $mainTransactionFeeStoreAsIcp.toE8s()
-    ),
-    fee: $mainTransactionFeeStore,
+  $: max = getMaxTransactionAmount({
+    balance: selectedAccount?.balance.toE8s(),
+    fee: $transactionsFeesStore.main,
+    maxAmount,
   });
   const addMax = () => (amount = max);
 
--- a/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
+++ b/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
@@ -9,7 +9,11 @@
   import type { Account } from "../../../types/account";
   import { InvalidAmountError } from "../../../types/neurons.errors";
   import { assertEnoughAccountFunds } from "../../../utils/accounts.utils";
-  import { convertNumberToICP, maxICP } from "../../../utils/icp.utils";
+  import {
+    convertNumberToICP,
+    maxE8sToNumber,
+    minE8s,
+  } from "../../../utils/icp.utils";
   import SelectAccountDropdown from "../../../components/accounts/SelectAccountDropdown.svelte";
   import IcpComponent from "../../../components/ic/ICP.svelte";
   import AmountInput from "../../../components/ui/AmountInput.svelte";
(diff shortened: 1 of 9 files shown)


frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
```diff
@@ -20,10 +24,14 @@
...
   // TODO: Handle min and max validations inline: https://dfinity.atlassian.net/browse/L2-798
+  export let maxAmount: bigint | undefined = undefined;
 
   let max: number = 0;
-  $: max = maxICP({
-    icp: selectedAccount?.balance,
+  $: max = maxE8sToNumber({
+    e8s: minE8s(
+      selectedAccount?.balance.toE8s(),
+      (maxAmount ?? BigInt(0)) + $mainTransactionFeeStoreAsIcp.toE8s()
```
mstrasinskis, line 33: Not sure, could you please recheck that this value should include `$mainTransactionFeeStoreAsIcp.toE8s()`?
Probably:
```ts
e8s: minE8s(
      selectedAccount?.balance.toE8s() - $mainTransactionFeeStoreAsIcp.toE8s(),
      maxAmount ?? BigInt(0)
```

mstrasinskis (changes requested):
LGTM, only one question.

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
lmuntaner: The fee is already "removed" inside `maxE8sToNumber`.

Maybe not the best helper ever I'd say...

To be honest, it does just a subtraction and conversion to a number by dividing by E8S_PER_ICP

Maybe we don't want to hide that. What do you think?

mstrasinskis: Ah, right, the fee was already substracted. I agree, looks like `maxE8sToNumber` does more than e.g. it's sibling util `minE8s`. Would be nice either rename it or don't use fee there.

lmuntaner: Changed to the following util:

```
/**
 * Calculates the maximum amount for a transaction.
 *
 * @param balanceE8s The balance of the account in E8S.
 * @param fee The fee of the transaction in E8S.
 * @param maxAmount The maximum amount of the transaction not counting the fees.
 * @returns
 */
export const getMaxTransactionAmount = ({
  balance = BigInt(0),
  fee = BigInt(0),
  maxAmount,
}: {
  balance?: bigint;
  fee?: bigint;
  maxAmount?: bigint;
}): number => {
  if (maxAmount === undefined) {
    return Math.max(Number(balance - fee), 0) / E8S_PER_ICP;
  }
  return Math.min(Number(maxAmount), Number(balance - fee)) / E8S_PER_ICP;
};
```

Let me know what you think

lmuntaner:
@mstrasinskis ready for another review

frontend/src/lib/utils/icp.utils.ts
```diff
@@ -58,8 +58,28 @@ export const formattedTransactionFeeICP = (fee: number): string =>
...
+  maxAmount,
+}: {
+  balance?: bigint;
+  fee?: bigint;
+  maxAmount?: bigint;
+}): number => {
+  if (maxAmount === undefined) {
+    return Math.max(Number(balance - fee), 0) / E8S_PER_ICP;
+  }
+  return Math.min(Number(maxAmount), Number(balance - fee)) / E8S_PER_ICP;
```
mstrasinskis, line 81:
```suggestion
  return Math.min(Number(maxAmount), Math.max(Number(balance - fee), 0)) / E8S_PER_ICP;
```

mstrasinskis (approved):
LGTM, thank you. 2 minor suggestions.

frontend/src/lib/utils/icp.utils.ts
```diff
@@ -58,8 +58,28 @@ export const formattedTransactionFeeICP = (fee: number): string =>
...
+ * @param fee The fee of the transaction in E8S.
+ * @param maxAmount The maximum amount of the transaction not counting the fees.
+ * @returns
+ */
+export const getMaxTransactionAmount = ({
+  balance = BigInt(0),
+  fee = BigInt(0),
+  maxAmount,
+}: {
+  balance?: bigint;
```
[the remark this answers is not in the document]
lmuntaner: This is a common pattern we use. Instead of handling the `undefined` outside the function, we handle it inside.

For example:
```
export const isAccountHardwareWallet = (
  account: Account | undefined
): boolean => account?.type === "hardwareWallet";

export const getAccountFromStore = ({
  identifier,
  accountsStore: { main, subAccounts, hardwareWallets },
}: {
  identifier: string | undefined;
  accountsStore: AccountsStore;
}): Account | undefined => {
```

This avoids repeating the same code of checking `undefined` outside the function.

Not that I am a big fan of this pattern. But I am a big fan of consistency.

frontend/src/lib/utils/icp.utils.ts
lmuntaner: Good point!

lmuntaner: I will also add a test for this.

lmuntaner: done

PR #1300, opened 2022-09-03
L2-991: Upgrade nns-js with breaking changes

lmuntaner:
# Motivation

Upgrade nns-js. Fix broken changes in nns-dapp.

This is in preparation for the refactor to start using AmountToken instead of ICP.

# Changes

* Upgrade nns-js, sns-js and utils to latest next version.
* LedgerCanister and GovernanceCanister don't expect ICP anymore but bigint.

# Tests

* Fix broken tests.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -791,9 +791,9 @@
       }
     },
     "node_modules/@dfinity/nns": {
-      "version": "0.7.0-next-2022-09-01.1",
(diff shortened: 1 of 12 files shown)


peterpeterparker (approved):
LGTM, thanks

It's quite a cool idea you had to use `bigint` instead of `ICP` in ledger and governance 👍

lmuntaner:
> LGTM, thanks
> 
> It's quite a cool idea you had to use `bigint` instead of `ICP` in ledger and governance 👍

Yes, agree.

I think that the pattern we have in sns-js, to keep the types as close as possible to Candid makes it easier to maintain.

PR #1303, opened 2022-09-05
L2-991: Refactor account type

lmuntaner:
# Motivation

Refactor our codebase to be compatible with multiple tokens.

# Changes

* Change the Account type. Balance is now a TokenAmount.
* Rename "ICP" component to "AmountDisplay".

# Tests

* Fix all broken tests.


lmuntaner:

frontend/src/lib/components/accounts/AccountCard.svelte
```diff
@@ -1,17 +1,17 @@
 <script lang="ts">
   import type { Account } from "../../types/account";
   import { Card } from "@dfinity/gix-components";
-  import ICP from "../ic/ICP.svelte";
+  import AmountTokens from "../ic/AmountTokens.svelte";
```
peterpeterparker, line 4: dumb question: should it be plural? should be "Amount + Token" or "Token + Amount"?

- `TokenAmount` the class from nns-js
- `AmountTokens` the component

peterpeterparker: extra idea: should we call the component `Balance`? not sure, but it's only use to display balance?

frontend/src/lib/components/ic/AmountTokens.svelte
```diff
@@ -1,30 +1,31 @@
...
+    >{`${sign}${formatICP({ value: amount.toE8s(), detailed })}`}</span
   >
-    <span data-tid="icp-value" class="value"
-      >{`${sign}${formatICP({ value: icp.toE8s(), detailed })}`}</span
-    >
-    <span class="label">{label}</span>
-  </div>
-{/if}
+  <span class="label"
+    >{label !== undefined ? label : amount.token.symbol ?? $i18n.core.icp}</span
```
peterpeterparker, line 26: I would not fallback to ICP - i.e. if an issue happens in a Sns project - like we forget the label in a review - the value will be display in ICP what would be incorrect and can mean quite something different

frontend/src/lib/components/accounts/AccountCard.svelte
lmuntaner: It's used also to display transaction fees... not very clear with just Balance.

Maybe in singular then? `AmountToken`?

frontend/src/lib/components/ic/AmountTokens.svelte
lmuntaner: Would you fallback to nothing then?

frontend/src/lib/components/accounts/AccountCard.svelte
peterpeterparker: Then we have `TokenAmount` and `AmountToken`, don't know what's the best 🤔

frontend/src/lib/components/ic/AmountTokens.svelte
peterpeterparker: I would say, no label, no direct meaning. Or should the label be mandatory, that's possible?

lmuntaner: The label should not be mandatory because it comes from the Token.

I would have preferred to remove the label prop altogether. But it's used to show the voting power.

In another PR, I can remove the usage for voting power and always use the label from the AmountToken.

frontend/src/lib/components/accounts/AccountCard.svelte
lmuntaner: That's why I used the plural. It was easier to differentiate them...

I didn't like the same name. I realized this component was called `<Icp`, `<ICPComponet`, `<ICP`, `IcpComponent`, or the type `ICPType`, `IcpType`...

Better give it a different name.

peterpeterparker: Definitely makes sense to have a different name, not sure I'll remember plural and inverting order make the difference. Being said, don't have better idea. @mstrasinskis do you have maybe an idea of naming?

frontend/src/lib/components/ic/AmountTokens.svelte
peterpeterparker: Then let's fallback to nothing then for the time being, agree?

lmuntaner: Sounds good to me

frontend/src/lib/components/accounts/AccountCard.svelte
lmuntaner: @peterpeterparker apart from this naming, anything else to review for this PR or is it good to go?

peterpeterparker: Just the naming but happy to have a final look once we are set. Did not had an idea in the meantime unfortunatelly

lmuntaner: no, sorry. I can't think of a better name. It would be better if you can already take the final look and maybe then @mstrasinskis has a better idea for the name. This would speed up the merge if the only thing is the name.

Name can be also change later again btw.

peterpeterparker: - TokenAmountDiplay
- TokenAmountSymbol
- TokenAmountBalance
- AmountDisplay
- AmountSymbol
- Amount

lmuntaner: Do you want me to choose? Not sure what you want me to do with this list? Anything be `AmountTokens`?

peterpeterparker: Some ideas, happy to hear better solution!!! Just not a fan of `AmountTokens` because as discussed above I think it does not solve the original goal to have a distinctive name for the component.

lmuntaner: How about `AmountDisplay` then?

peterpeterparker: let's do that 👍

lmuntaner: done

lmuntaner:

frontend/src/lib/utils/icp.utils.ts
```diff
@@ -48,8 +48,14 @@ export const formatICP = ({
     .replace(/,/g, "'");
 };
 
-export const sumICPs = (...icps: ICP[]): ICP =>
-  ICP.fromE8s(icps.reduce<bigint>((acc, icp) => acc + icp.toE8s(), BigInt(0)));
+export const sumTokenAmounts = (...amountTokens: TokenAmount[]): TokenAmount =>
```
peterpeterparker, line 51: can you prevent the token being mixed?

not sure if we can solve this with typescript generic or if we should throw an error if mixed or if it should return the sum per token

frontend/src/lib/components/ic/ICPText.svelte
```diff
@@ -1,8 +1,8 @@
 <script lang="ts">
-  import type { ICP } from "@dfinity/nns";
-  import Icp from "./ICP.svelte";
+  import type { ICP, TokenAmount } from "@dfinity/nns";
+  import AmountDisplay from "./AmountDisplay.svelte";
 
-  export let icp: ICP;
+  export let amount: ICP | TokenAmount;
```
peterpeterparker, line 5: should we also add the // TODO here about removing ICP?

peterpeterparker: `// TODO: Remove ICP reference` as inn AmountDisplay

frontend/src/lib/utils/icp.utils.ts
lmuntaner: Sure, I can make sure all the token symbols are the same.

lmuntaner: done

frontend/src/lib/components/ic/ICPText.svelte
lmuntaner: done

peterpeterparker (approved):
LGTM, thanks

Before merge, just one question, do we want to rename `icp.utils` into `token.utils`?

lmuntaner:
This and many more change I prefer to do in other PRs. Step by step, hehe.

PR #1305, opened 2022-09-05
L2-999: Setup context in path to switch SNS projects

lmuntaner:
# Motivation

Setup `/u/:rootCanisterId` for neurons and projects.

# Changes

* selectedProjectStore is a derived store from routeStore. Defaults to OWN_CANISTER_ID.
* SelectProjectDropdown changes url instead of store.
* New method on routeStore "changeContext" to switch only the context id of the path `/u/<id>/neurons`
* Rename `AppPath.Neurons` to `AppPath.LegacyNeurons`. Neurons uses context.
* New app-path utils related to context: getContextDetailsFromPath, isContextPath and getContextFromPath.
* Moved derived stores of snsSortedNeuronsSgtore and selectedProjectStore to "derived" folder.

# Tests

* Fix broken tests.
* New tests for new app-path utils.
* New test for new changeContext method in routeStore.


lmuntaner:
@peterpeterparker @mstrasinskis this is a working prototype of how we could move from `/neurons` to `/context/:rootCanisterId/neurons` and similar.

I'd like your view before I continue further checking all use cases as well as adding more tests.

Thanks!

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -28,6 +30,31 @@ export const isRoutePath: ({
   new RegExp(`^${pathValidation(path)}$`).test(routePath)
 );
 
+const contextPathRegex = new RegExp(`^${AppPath.Context}/[a-zA-Z0-9-]+`);
+
+export const getContextDetailsFromPath = (
```
peterpeterparker, line 35: can you document with an example?

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -28,6 +30,31 @@ export const isRoutePath: ({
...
+    return routePath.replace(contextPathRegex, "");
+  }
+};
+
+// Don't match strictly
+export const isContextPath = (routePath: string): boolean =>
+  new RegExp(`^${AppPath.Context}/[a-zA-Z0-9-]+`).test(routePath);
+
+/**
+ * Returns true if the routePath is a path with a context prefix: `/#/context/:rootCanisterId{/...}`
```
peterpeterparker, line 48: I'm confuse about the comment. It says "Returns true...." but the function returns "string | undefined"

frontend/src/lib/stores/route.store.ts
```diff
@@ -61,6 +66,29 @@ const initRouteStore = () => {
 
       replaceHistory({ path, query });
     },
+
+    changeContext: (newContext: string) => {
```
peterpeterparker, line 70: can you please document the function?

frontend/src/lib/components/common/RouteModule.svelte
```diff
@@ -17,6 +17,9 @@
     switch (path) {
       case AppPath.Accounts:
         return (await import("../../../routes/Accounts.svelte")).default;
+      // TODO: Redirect to neurons with OWN_CANISTER_ID
```
peterpeterparker, line 20: should we reference a jira task?

frontend/src/lib/constants/routes.constants.ts
```diff
@@ -1,14 +1,17 @@
...
   ProposalDetail = "/#/proposal",
   NeuronDetail = "/#/neuron",
   CanisterDetail = "/#/canister",
   Launchpad = "/#/launchpad",
-  ProjectDetail = "/#/project",
-  SnsNeuronDetail = "/#/project/:rootCanisterId/neuron",
+  ProjectDetail = "/#/context",
+  // Not used directly
+  Context = "/#/context",
+  SnsNeuronDetail = "/#/context/:rootCanisterId/neuron",
```
peterpeterparker, line 16: shouldn't it be plural too?
or should `SnsNeuronDetail` be named `NeuronDetail`?

frontend/src/lib/constants/routes.constants.ts
```diff
@@ -1,14 +1,17 @@
...
   Proposals = "/#/proposals",
   Canisters = "/#/canisters",
   Wallet = "/#/wallet",
   ProposalDetail = "/#/proposal",
   NeuronDetail = "/#/neuron",
   CanisterDetail = "/#/canister",
   Launchpad = "/#/launchpad",
-  ProjectDetail = "/#/project",
-  SnsNeuronDetail = "/#/project/:rootCanisterId/neuron",
+  ProjectDetail = "/#/context",
```
peterpeterparker, line 13: why `/#/context` twice (`ProjectDetail` and `Context`)?

frontend/src/lib/derived/selected-project.derived.ts
```diff
@@ -0,0 +1,43 @@
+import { Principal } from "@dfinity/principal";
+import { derived } from "svelte/store";
+import { OWN_CANISTER_ID } from "../constants/canister-ids.constants";
+import { routeStore } from "../stores/route.store";
+import { getContextFromPath } from "../utils/app-path.utils";
+import { isNnsProject } from "../utils/projects.utils";
+
+/**
+ * In Neurons or ultimately in Voting screen, user can select the "universe" - e.g. display Neurons of Nns or a particular Sns
```
peterpeterparker, line 9: can you improve the documentation with what it does now - i.e. deriving the route store?

frontend/src/lib/derived/selected-project.derived.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+export const snsProjectSelectedStore = derived([routeStore], ([{ path }]) => {
+  const maybeContextId = getContextFromPath(path);
+  if (maybeContextId !== undefined) {
+    try {
+      const principal = Principal.fromText(maybeContextId);
+      return principal;
+    } catch (error) {
+      // Add execeptions, maybe bitcoin wallet?
```
peterpeterparker, line 18: add a jira task? e.g. can happens if a Sns project disappear (out of cycles) and user use a deep link

frontend/src/lib/derived/selected-project.derived.ts
```diff
@@ -0,0 +1,43 @@
...
+import { isNnsProject } from "../utils/projects.utils";
+
+/**
+ * In Neurons or ultimately in Voting screen, user can select the "universe" - e.g. display Neurons of Nns or a particular Sns
+ */
+export const snsProjectSelectedStore = derived([routeStore], ([{ path }]) => {
+  const maybeContextId = getContextFromPath(path);
+  if (maybeContextId !== undefined) {
+    try {
+      const principal = Principal.fromText(maybeContextId);
```
peterpeterparker, line 15: nth: you can spare the variable assignation

```
return Principal.fromText(maybeContextId);
`^^

File: frontend/src/lib/utils/app-path.utils.ts (hunk 2 above)
Comment (lmuntaner) [reply frontend/src/lib/utils/app-path.utils.ts:48]: yes, wrong comment. That was before. I'll update it.

File: frontend/src/lib/utils/app-path.utils.ts (hunk 1 above)
Comment (lmuntaner) [reply frontend/src/lib/utils/app-path.utils.ts:35]: yes.

File: frontend/src/lib/stores/route.store.ts (hunk 3 above)
Comment (lmuntaner) [reply frontend/src/lib/stores/route.store.ts:70]: yes

File: frontend/src/lib/components/common/RouteModule.svelte (hunk 4 above)
Comment (lmuntaner) [reply frontend/src/lib/components/common/RouteModule.svelte:20]: sure

File: frontend/src/lib/constants/routes.constants.ts (hunk 5 above)
Comment (lmuntaner) [reply frontend/src/lib/constants/routes.constants.ts:16]: Good point, probably this is the new NeuronDetail and the other one is the LegacyNeuronDetail. I'll change that.

File: frontend/src/lib/constants/routes.constants.ts (hunk 6 above)
Comment (lmuntaner) [reply frontend/src/lib/constants/routes.constants.ts:13]: I use `AppPath.Context` in the helpers to build and check the context paths. I wasn't sure it would be a page on it's own. Maybe we want to keep `/project/:rootCanisterId` or put the project details somewhere else.

One path is the context for many paths, the other is the path for the project details. They just happen to be the same at the moment, but they are conceptually different things.

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 7 above)
Comment (lmuntaner) [reply frontend/src/lib/derived/selected-project.derived.ts:9]: sure, this was the comment there already. But I'll change it.

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 8 above)
Comment (lmuntaner) [reply frontend/src/lib/derived/selected-project.derived.ts:18]: that won't be a problem here, it will still be a valid context.

then the page of wherever that is, will have to redirect, but here it's a valid context.

We don't have anything related to bktc or whatever the bitcoin wallet is called. I added the comment to understand why I am ignoring it. But at the moment there is nothing for us to do.

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 9 above)
Comment (lmuntaner) [reply frontend/src/lib/derived/selected-project.derived.ts:15]: Sure.

File: frontend/src/lib/utils/app-path.utils.ts (hunk 1 above)
Comment (lmuntaner) [reply frontend/src/lib/utils/app-path.utils.ts:46]: done

File: frontend/src/lib/utils/app-path.utils.ts (hunk 2 above)
Comment (lmuntaner) [reply frontend/src/lib/utils/app-path.utils.ts]: done

File: frontend/src/lib/stores/route.store.ts (hunk 3 above)
Comment (lmuntaner) [reply frontend/src/lib/stores/route.store.ts:82]: done

File: frontend/src/lib/constants/routes.constants.ts (hunk 5 above)
Comment (lmuntaner) [reply frontend/src/lib/constants/routes.constants.ts]: done

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 7 above)
Comment (lmuntaner) [reply frontend/src/lib/derived/selected-project.derived.ts]: done

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 9 above)
Comment (lmuntaner) [reply frontend/src/lib/derived/selected-project.derived.ts]: done

**Comment by lmuntaner** [comment]

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 8 above)
Comment (peterpeterparker) [reply frontend/src/lib/derived/selected-project.derived.ts:20]: > that won't be a problem here, it will still be a valid context.

untrue I think. if the Sns project disappear, the context does not exists therefore user is redirect to NNS without a clue why if I get it correctly.

"I just gonna check my 1 trillion Token neuron I bookmarked -> redirected to NNS dapp"

But maybe I am missing an error msg somewhere?

File: frontend/src/lib/constants/routes.constants.ts (hunk 6 above)
Comment (peterpeterparker) [reply frontend/src/lib/constants/routes.constants.ts:13]: Thanks for the answer. I would say it's probably better to keep `/project/:rootCanisterId` because if someone hits `/context/:id` we don't know if the intent was effectively to open the Sns project or that the URL was mistyped?

So if we keep `/project/` then we can redirect `/context/:id/nothing-else` or `/context/nothing-else` to root `/` or `/accounts/` (since it's the login fallback).

What do you think?

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 8 above)
Comment (lmuntaner) [reply frontend/src/lib/derived/selected-project.derived.ts:20]: Correct, the user will be redirected but that will be in the specifi page.

For example, if the user access a neuron from a canister without cycles, it will try to get the neuron and that call will fail and show the error.

Yet, the canister id is still a valid canister id, with or without cycles. Which is the only thing we are checking here. Is the context a valid canister id. That's it. The specifics of that canister will be handled in each part. In the wallet page, or the neurons or whatever.

File: frontend/src/lib/constants/routes.constants.ts (hunk 6 above)
Comment (lmuntaner) [reply frontend/src/lib/constants/routes.constants.ts:13]: I am ok with keeping `/project/:rootCanisterId` for project detail page.

But I don't think we should keep both `/project/:canisterId` and `/context/:canisterId`. At the moment, this is not released to production. Therefore, we have the possibility to set how we want it. To maintain two different ways of accessing the same thing just calls for problems, IMHO.

Comment (lmuntaner) [reply frontend/src/lib/constants/routes.constants.ts:13]: Do you want me to change only `/project/:canisterId`?

Do you have a strong argument to keep both `/context` and `/project`>

Comment (peterpeterparker) [reply frontend/src/lib/constants/routes.constants.ts:13]: As answered above, I tend to keep `/project/:rootCanisterId` and in addition I would redirect when `/context/` is not provided correctly.

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 8 above)
Comment (peterpeterparker) [reply frontend/src/lib/derived/selected-project.derived.ts:20]: If there is an explicit error then I am find with it.

File: frontend/src/lib/constants/routes.constants.ts (hunk 6 above)
Comment (lmuntaner) [reply frontend/src/lib/constants/routes.constants.ts:13]: it's not to keep, nothing has been deployed. You mean that you prefer to keep two routes for the same purpose?

I am sorry but I prefer to avoid the extra complexity.

I guess we'll have to ask @mstrasinskis

File: frontend/src/lib/derived/selected-project.derived.ts (hunk 8 above)
Comment (lmuntaner) [reply frontend/src/lib/derived/selected-project.derived.ts:20]: yes, they are in each page.

File: frontend/src/lib/constants/routes.constants.ts (hunk 6 above)
Comment (mstrasinskis) [reply frontend/src/lib/constants/routes.constants.ts:13]: @peterpeterparker, what do you think about dropping `/context` url part? So there'll be `universe` related pages and general pages.

```
Launchpad = "/#/launchpad",
...
ProjectDetail = "/#/:rootCanisterId",
NeuronDetail = "/#/:rootCanisterId/neuron",
Neurons = "/#/:rootCanisterId/neurons",
...
```

**Comment by peterpeterparker** [comment]
As discussed with @lmuntaner and @mstrasinskis :

```
https://nns.ic0.app/#/u/aaaaa-cai-aada-ccc/neurons -> neurons
https://nns.ic0.app/#/u/aaaaa-cai-aada-ccc/accounts -> accounts
https://nns.ic0.app/#/accounts -> https://nns.ic0.app/#/u/aaaaa-cai-aada-ccc/accounts -> accounts (fallback url support)
https://nns.ic0.app/#/neurons/123456 -> https://nns.ic0.app/#/u/aaaaa-cai-aada-ccc/neurons/123456 (fallback url support)
https://nns.ic0.app/#/launchpad -> launchpad
https://nns.ic0.app/#/u/bbbbb-cai-aada-ccc -> sns project detail
https://nns.ic0.app/#/u/aaaaa-cai-aada-ccc -> https://nns.ic0.app/#/u/aaaaa-cai-aada-ccc/accounts -> accounts
https://nns.ic0.app/#/u/ckbtc -> https://nns.ic0.app/#/u/ckbtc/accounts -> accounts
https://nns.ic0.app/#/something-wrong -> error page + toast
https://nns.ic0.app/#/something-wrong/whatever -> error page + toast
https://nns.ic0.app/#/u/something-wrong -> error page + toast
https://nns.ic0.app/#/u/something-wrong/whatever -> error page + toast
https://nns.ic0.app/#/unknown/cbktc -> error page 

aaaaa-cai-aada-ccc = nns
bbbbb-cai-aada-ccc = a sns project
```

File: frontend/src/lib/constants/routes.constants.ts (hunk 6 above)
Comment (lmuntaner) [reply frontend/src/lib/constants/routes.constants.ts]: done

**Comment by lmuntaner** [comment]

This only change context. There is no error handling in this PR apart from the one that it's already done in the nns dapp.

To implement the new error redirections I created the task https://dfinity.atlassian.net/browse/GIX-1011

File: frontend/src/lib/components/common/RouteModule.svelte [hunk 10]
```diff
@@ -17,6 +17,8 @@
     switch (path) {
       case AppPath.Accounts:
         return (await import("../../../routes/Accounts.svelte")).default;
+      case AppPath.LegacyNeurons:
+        return (await import("../../../routes/Neurons.svelte")).default;
```
peterpeterparker, line 21: if you remove this line that should do to, it's a `switch`

you can do the same for `NeuronDetails`

frontend/src/lib/derived/selected-project.derived.ts
```diff
@@ -0,0 +1,45 @@
+import { Principal } from "@dfinity/principal";
+import { derived } from "svelte/store";
+import { OWN_CANISTER_ID } from "../constants/canister-ids.constants";
+import { routeStore } from "../stores/route.store";
+import { getContextFromPath } from "../utils/app-path.utils";
+import { isNnsProject } from "../utils/projects.utils";
+
+/**
+ * In Neurons or ultimately in Voting screen, user can select the "context" - e.g. display Neurons of Nns or a particular Sns
+ * This "context" is represented in the path of the URL: `/#/u/<some-context>/...`
```
peterpeterparker, line 10: should we update the doc with the word "universe" instead of "context"?

frontend/src/lib/derived/selected-project.derived.ts
```diff
@@ -0,0 +1,45 @@
...
+ * The store reads the routeStore and returns the context.
+ * It defaults to NNS (OWN_CANISTER_ID) if the path is not a context path.
+ */
+export const snsProjectSelectedStore = derived([routeStore], ([{ path }]) => {
+  const maybeContextId = getContextFromPath(path);
+  if (maybeContextId !== undefined) {
+    try {
+      return Principal.fromText(maybeContextId);
+    } catch (error) {
+      // Add execeptions, maybe bitcoin wallet?
```
peterpeterparker, line 20: here I guess we can add the jira ticket number in the todo?

frontend/src/lib/pages/NnsNeuronDetail.svelte
```diff
@@ -0,0 +1,119 @@
+<script lang="ts">
```
peterpeterparker, line 1: just code moved in `NnsNeuronDetail` right - i.e. no need to review the code?

frontend/src/lib/services/neurons.services.ts
```diff
@@ -881,7 +881,10 @@ export const makeDummyProposals = async (neuronId: NeuronId): Promise<void> => {
 };
 
 export const routePathNeuronId = (path: string): NeuronId | undefined => {
-  if (!isRoutePath({ path: AppPath.NeuronDetail, routePath: path })) {
+  if (
+    !isRoutePath({ path: AppPath.LegacyNeuronDetail, routePath: path }) &&
```
peterpeterparker, line 885: it would probably be handy to modify that utils to path the `path` as array of path or something

```
!isRoutePath({ path: [AppPath.LegacyNeuronDetail, AppPath.NeuronDetail], routePath: path })
```

of course in another pr

frontend/src/lib/stores/route.store.ts
```diff
@@ -61,6 +67,50 @@ const initRouteStore = () => {
...
+     * Doesn't do anything if the current path is not a context path.
+     *
+     * When called with `bbbbb-bb`
+     * Ex: `/#/u/aaaaa-aa/neurons` becomes `/#/u/bbbbb-bb/neurons`
+     * Ex: `/#/u/aaaaa-aa/account/1234` becomes `/#/u/bbbbb-bb/account/1234`
+     * Ex: `/#/neurons` doesn't change anything
+     *
+     * @param newContext string - the new context to navigate to
+     */
+    changeContext: (newContext: string) => {
```
peterpeterparker, line 82: should we rename it `changeUniverse` or else?

frontend/src/lib/stores/route.store.ts
```diff
@@ -61,6 +67,50 @@ const initRouteStore = () => {
...
+      let newPath;
+      update((state: RouteStore) => {
+        newPath = state.path;
```
peterpeterparker, line 85: can we have this logic in `route.utils` or elsewhere - i.e in another for of pattern? feels like we mix both logic (find the value of the new context) and updating the store at the same time

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -28,6 +30,47 @@ export const isRoutePath: ({
...
+ */
+export const getContextDetailsFromPath = (
+  routePath: string
+): string | undefined => {
+  if (isContextPath(routePath)) {
+    return routePath.replace(contextPathRegex, "");
+  }
+};
+
+// Don't match strictly
```
peterpeterparker, line 54: can you provide an example of what not strictly means?

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -28,6 +30,47 @@ export const isRoutePath: ({
...
+export const isContextPath = (routePath: string): boolean =>
+  new RegExp(`^${AppPath.Context}/[a-zA-Z0-9-]+`).test(routePath);
+
+/**
+ * Returns the context if the routePath is a path with a context prefix: `/#/u/:rootCanisterId{/...}`
```
peterpeterparker, line 59: ditto about "context" vs "universe"?

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -28,6 +30,47 @@ export const isRoutePath: ({
...
+ * Returns undefined if the routePath is not a context path.
+ *
+ * Ex: `/#/u/aaaaa-aa/neurons` returns `aaaaa-aa`
```
peterpeterparker, line 62: if `/#/u/aaaaa-aa` it returns `aaaaa-aa`?

frontend/src/App.svelte
```diff
@@ -52,16 +52,17 @@
 <Guard>
   <Route path={AppPath.Authentication} />
   <PrivateRoute path={AppPath.Accounts} />
+  <PrivateRoute path={AppPath.LegacyNeurons} />
```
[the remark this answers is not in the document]
lmuntaner: Not yet, I will do that.

frontend/src/lib/components/common/RouteModule.svelte
lmuntaner: Neuron details I use a different component per path.

frontend/src/lib/derived/selected-project.derived.ts
lmuntaner: not yet, only when we have exceptions. Which we still don't.

frontend/src/lib/pages/NnsNeuronDetail.svelte
lmuntaner: correct

frontend/src/lib/stores/route.store.ts
lmuntaner: sure, I can move this.

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: sure

frontend/src/lib/components/common/RouteModule.svelte
lmuntaner: done

frontend/src/lib/derived/selected-project.derived.ts
lmuntaner: Discussed offline. We keep the concept of a context path, but we use `/u/` for url because it's nicer.

frontend/src/App.svelte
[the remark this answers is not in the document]
lmuntaner: done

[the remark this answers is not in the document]
lmuntaner: https://dfinity.atlassian.net/browse/GIX-1013

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: done

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: ditto

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -5,10 +5,12 @@ const mapper: Record<string, string> = {
   // exceptions only
   [AppPath.Wallet]: `${AppPath.Wallet}/[a-zA-Z0-9]+`,
   [AppPath.ProposalDetail]: `${AppPath.ProposalDetail}/[0-9]+`,
-  [AppPath.NeuronDetail]: `${AppPath.NeuronDetail}/[0-9]+`,
+  [AppPath.LegacyNeuronDetail]: `${AppPath.LegacyNeuronDetail}/[0-9]+`,
+  [AppPath.Neurons]: `${AppPath.Context}/[a-zA-Z0-9-]+/neurons`,
   [AppPath.CanisterDetail]: `${AppPath.CanisterDetail}/[a-zA-Z0-9-]+`,
   [AppPath.ProjectDetail]: `${AppPath.ProjectDetail}/[a-zA-Z0-9-]+`,
-  [AppPath.SnsNeuronDetail]: `${AppPath.ProjectDetail}/[a-zA-Z0-9-]+/neuron/[a-zA-Z0-9-]+`,
+  [AppPath.Context]: `${AppPath.Context}/[a-zA-Z0-9-]+`,
```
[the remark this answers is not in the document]
lmuntaner: * True, I can remove it from here.
* Discussed on slack. We keep the concept of a "context path".

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: Yes, correct.

Check the tests:
```
    it("returns values after `/#/u` parent path", () => {
      expect(getContextFromPath("/#/u/aaaaa-aa")).toBe("aaaaa-aa");
      expect(getContextFromPath("/#/u/aaaaa-aa/neurons")).toBe("aaaaa-aa");
      expect(getContextFromPath("/#/u/aaaaa-aa/accounts")).toBe("aaaaa-aa");
      expect(
        getContextFromPath("/#/u/wxns6-qiaaa-aaaaa-aaaqa-cai/neurons")
      ).toBe("wxns6-qiaaa-aaaaa-aaaqa-cai");
      expect(getContextFromPath("/#/u/123444")).toBe("123444");
    });
```

frontend/src/lib/utils/app-path.utils.ts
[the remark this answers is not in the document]
lmuntaner: done

lmuntaner:

frontend/src/lib/stores/route.store.ts
lmuntaner: as agree, we keep the concept of "context"

lmuntaner:

frontend/src/lib/components/neurons/SelectProjectDropdown.svelte
```diff
@@ -51,12 +65,18 @@
...
+    name="project"
+    bind:selectedValue={selectedCanisterId}
+    testId="select-project-dropdown"
+  >
+    {#each selectableProjects as { canisterId, name } (canisterId)}
+      <DropdownItem value={canisterId}>{name}</DropdownItem>
+    {/each}
+  </Dropdown>
+{:else}
+  <Spinner inline />
```
peterpeterparker, line 81: - does it needs a loading indication? i.e. I guess we gonna display multiple spinner on screens, is it something we want?

- is there maybe a layout shift? height of spinner != height of dropdown

frontend/src/lib/constants/routes.constants.ts
```diff
@@ -1,14 +1,17 @@
...
   ProposalDetail = "/#/proposal",
-  NeuronDetail = "/#/neuron",
+  LegacyNeuronDetail = "/#/neuron",
   CanisterDetail = "/#/canister",
   Launchpad = "/#/launchpad",
-  ProjectDetail = "/#/project",
-  SnsNeuronDetail = "/#/project/:rootCanisterId/neuron",
+  ProjectDetail = "/#/u",
+  // Not used directly
+  Context = "/#/u",
```
peterpeterparker, line 15: Shouldn't this `Context` be a separated constant rather than a value of the enum?

Since it's no effective route, it would probably be safer

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -58,5 +112,46 @@ export const getParentPathDetail = (
...
+  path,
+  newContext,
+}: {
+  path: string;
+  newContext: string;
+}) => {
+  let newPath = path;
+  // Exception
+  if (isRoutePath({ path: AppPath.LegacyNeurons, routePath: path })) {
+    newPath = `${AppPath.Context}/${newContext}/neurons`;
```
peterpeterparker, line 140: is it correct that `/neurons` is hardcoded here?

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -58,5 +112,46 @@ export const getParentPathDetail = (
...
+ */
+export const changePathContext = ({
+  path,
+  newContext,
+}: {
+  path: string;
+  newContext: string;
+}) => {
+  let newPath = path;
+  // Exception
```
peterpeterparker, line 138: not sure I get comment `// Exception`?
same line 142

frontend/src/lib/components/neurons/SelectProjectDropdown.svelte
lmuntaner: * Good point, yes, we might display multiple spinners. Not sure what's the best solution.
* Yes, there is a small layout shift as is. Shall I create an issue to improve the loading state then?

frontend/src/lib/constants/routes.constants.ts
lmuntaner: It's not a route directly, bit it's part of the routes, that's why I added it here.

But yes, I can create a context instead of part of the enum if that suits better.

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: Yes, this redirects from `/neurons` to another context. This is an edge case since the `changeContext` won't probably be called with `/neurons` the code should redirect to `/u/<OWN_CANISTER_ID>/neurons` before the user can try to change the context. But better be safe...

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: Below are execeptions the `LegacyNeurons` and `LegacyNeuronDetail`.

Let me improve the comment.

frontend/src/lib/components/neurons/SelectProjectDropdown.svelte
peterpeterparker: > Shall I create an issue to improve the loading state then?

Yes a Jira task would be welcomed. Personally not a big fan of displaying lots of spinner but, worth asking Mischa.
Likewise it would then nice to fix the layout shift.

frontend/src/lib/utils/app-path.utils.ts
peterpeterparker: then the utils function `changePathContext` is not generic.

should the function be renamed or can this "edge case" happens in other scenario too and therefore the `/neurons` should become a variable?

frontend/src/lib/components/neurons/SelectProjectDropdown.svelte
lmuntaner: Done: https://dfinity.atlassian.net/browse/GIX-1015

frontend/src/lib/constants/routes.constants.ts
lmuntaner: done

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: done

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: `changePathContext` is generic in our app.

I updated the comment:
```
/**
 * Returns a new path with the new context.
 * Doesn't do anything if the current path is not a context path.
 *
 * When called with `bbbbb-bb`
 * Ex: `/#/u/aaaaa-aa/neurons` becomes `/#/u/bbbbb-bb/neurons`
 * Ex: `/#/u/aaaaa-aa/account/1234` becomes `/#/u/bbbbb-bb/account/1234`
 * Ex: `/#/neurons` becomes `/#/u/bbbbb-bb/neurons`
 * Ex: `/#/neuron/1234` becomes `/#/u/bbbbb-bb/neuron/1234`
 * Ex: `/#/proposals` does nothing because `/#/proposals` is not a context path
 *
 * @param path string - the path to change
 * @param newContext string - the new context to navigate to
 */
```

`/neurons` is a context path. We allow to keep it that way for backwards compatibility. But it actually means `/#/u/<OWN_CANISTER_ID>/neurons`. That's why is treated separately.

lmuntaner:

frontend/src/lib/utils/app-path.utils.ts
peterpeterparker: Yes sure it works for now but, I guess we will have to implement the same legacy fallback for accounts and later for voting no? So how do we track this? e.g. how do we not forget that when we add the new routing for accounts we also have update this function with an hardcoded fallback to `/accounts`?

In that sense the method is not generic but, maybe I am missing something and we will not have to improve this particular function when we update the accounts and voting navigation? If so, sure this can be marked as resolved.

lmuntaner: That is a good question.

Unfortunately, for now we need to remember to add the exception when we deveop it.

The same way we need to add the path exception to our path validation mapper when we add a new route and it's an exception:

```
const mapper: Record<string, string> = {
  // exceptions only
  [AppPath.Wallet]: `${AppPath.Wallet}/[a-zA-Z0-9]+`,
  [AppPath.ProposalDetail]: `${AppPath.ProposalDetail}/[0-9]+`,
  [AppPath.LegacyNeuronDetail]: `${AppPath.LegacyNeuronDetail}/[0-9]+`,
  [AppPath.Neurons]: `${CONTEXT_PATH}/[a-zA-Z0-9-]+/neurons`,
  [AppPath.CanisterDetail]: `${AppPath.CanisterDetail}/[a-zA-Z0-9-]+`,
  [AppPath.ProjectDetail]: `${AppPath.ProjectDetail}/[a-zA-Z0-9-]+`,
  [AppPath.NeuronDetail]: `${AppPath.ProjectDetail}/[a-zA-Z0-9-]+/neuron/[a-zA-Z0-9-]+`,
};

const pathValidation = (path: AppPath): string => mapper[path] ?? path;
```

The good part, is that this can happen only with legacy paths, new paths should have the context from the beginning and therefore will never be exceptions. It will happen with accounts for sure and it might happen with proposals.

Accounts I will do shortly and proposals is not yet in the short term roadmap.

peterpeterparker: Then I understand correctly. So:

- I would suggest to transform the hardcoded `/neurons` into a new constant/enum value that finds place next to the `mapper` or `AppPath` list and use these in the utils. Maybe the constant/enum can even be a subset of the `mapper` or `AppPath`.

- otherwise, add a comment about updating `mapper` and utils function

lmuntaner: Ok

lmuntaner: done. I created a helper to manage exceptions. Using a generic mapper was hard because each exception might be different. For example, the legacy neuron details needs to take care also of the neuron details.

I didn't want to assume too much of the exceptions, because then it might look reusable when it's not really.

lmuntaner:

PR #1307, opened 2022-09-05
fix: proposer voting power on mobile

mstrasinskis:
# Motivation

Fix proposer voting power layout on mobile

# Changes

- make legacy selector more precise

# Screenshots

## before
image

## after
image


lmuntaner (approved):
Thanks!

PR #1308, opened 2022-09-05
feat: proposal deadline countdown

peterpeterparker:
# Motivation

As on mainnet, display the remaining time of the proposal but, introduce a real time update of the countdown.

# Changes

- display proposal deadline
- update value every minute if deadline > 1 hour from now else refresh every seconds

# Screenshots

(simulated, that's why the status remains Open at then end)

Enregistrement de l’écran 2022-09-05 à 16 22 07


frontend/src/lib/components/proposals/ProposalCountdown.svelte
```diff
@@ -0,0 +1,84 @@
...
+  export let proposalInfo: ProposalInfo;
+
+  const ZERO: bigint = BigInt(0);
+
+  let clear: NodeJS.Timeout | undefined;
+  let countdown: bigint | undefined = undefined;
+
+  const clearCountdown = () => {
+    if (clear !== undefined) {
+      clearInterval(clear);
```
mstrasinskis, line 17: nth: `clear = undefined`

mstrasinskis (approved):
LGTM, thank you

frontend/src/lib/components/proposals/ProposalCountdown.svelte
mstrasinskis: Not relevant in that case.

PR #1309, opened 2022-09-06
L2-1001: Allow pending transactions only caller

lmuntaner:
# Motivation

We only add pending transaction from the caller instead of expecting the principal as parameter.

# Changes

* Remove "buyer" from the AddPendingNotifySwapRequest and use caller instead.
* Adapt all candid and related files.

# Tests

Only interface has changed. Functionality stays the same.


bitdivine (approved):

lmuntaner:
We'll do it in another way. Closing PR.

PR #1310, opened 2022-09-06
L2-1001: Allow pending transactions only caller

lmuntaner:
# Motivation

We only add pending transaction from the caller instead of expecting the principal as parameter.

# Changes

* Add a check "check_pending_transaction_buyer" before adding pending transaction.
* Return NotAuthorized if false.
* Adapt all candid and related files to new response.

# Tests

* New test for check_pending_transaction_buyer.

### DIFF
--- a/rs/src/accounts_store.rs
+++ b/rs/src/accounts_store.rs
@@ -239,6 +239,7 @@ pub struct AddPendingNotifySwapRequest {
 #[derive(CandidType)]
 pub enum AddPendingTransactionResponse {
     Ok,
+    NotAuthorized,
 }
 
 impl AccountsStore {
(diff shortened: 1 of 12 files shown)


lmuntaner:
@bitdivine please review

rs/src/accounts_store.rs
```diff
@@ -473,6 +474,11 @@ impl AccountsStore {
         AddPendingTransactionResponse::Ok
     }
 
+    pub fn check_pending_transaction_buyer(&mut self, caller: PrincipalId, buyer: PrincipalId) -> bool {
+        // TODO: To support hardware wallets, check that the buyer is a principal of the caller's hardware walleet.
```
hpeebles, line 478: I think this is a more accurate comment -
'To support hardware wallets, check that the buyer is either the caller's principal or the principal of a hardware wallet linked to the caller's account.'

lmuntaner: sounds good!

lmuntaner: done

hpeebles (approved):

nmattia (approved):
Rubber stamp

PR #1312, opened 2022-09-06
Bump II

bitdivine:
# Motivation
We currently use a specific release of II.  We bump that release when we notice that testnets are significantly behind prod but have historically preferred having a stable II over having a recent II.  As a result, we typically test on a version of II that is about a month behind production.  This PR proposes to change that, always testing against the version of II deployed to production.  To be more precise, always testing against the latest release, which will match prod unless II has to roll back at some time.

Note that a new release of II comes about once every one or two weeks, so this is still a pretty stable target.  This PR does not use an unstable target such as head of main on II which might change every hour.

Also note that if the NNS-dapp works in CI but breaks against prod, that is not good.

# Changes
- Use the latest release of II

# Tests
See CI.


nmattia (approved):
LGTM.

Some II tests are also run against a moving release: https://github.com/dfinity/internet-identity/blob/67d426ef46b1d6275b700186f27ab9c0fb7aaf01/.github/workflows/canister-tests.yml#L216-L218

lmuntaner (approved):

PR #1314, opened 2022-09-06
L2-990: Change ICP to TokenAmount 2nd round

lmuntaner:
# Motivation

Keep removing ICP for TokenAmount.

# Changes

* Make the change necessary to use only TokenAmount type in AmountDisplay component.

# Tests

* Fix broken tests.


lmuntaner:

Second round of the refactor to remove ICP dependency.

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
```diff
@@ -37,7 +34,7 @@
 
   let balance: TokenAmount;
   $: ({ balance } = $store.selectedAccount ?? {
-    balance: TokenAmount.fromE8s({ amount: BigInt(0), token: ICPToken }),
+    balance: TokenAmount.fromE8s({ amount: BigInt(0) }),
```
peterpeterparker, line 37: should we not remove the `token: ICPToken` and actually make it mandatory in `nns-js.fromE8s`?

Even a bit more verbose, I ask my self if it is not more readable to make it mandatory. Likewise it can prevent issue when we will be develop features that are common to sns project and icp.

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
```diff
@@ -50,9 +47,11 @@
       return;
     }
 
-    const icp: ICP | undefined = convertNumberToICP(amount);
+    const icp: TokenAmount | FromStringToTokenError = TokenAmount.fromNumber({
+      amount,
+    });
 
-    if (icp === undefined) {
+    if (icp === undefined || !(icp instanceof TokenAmount)) {
```
peterpeterparker, line 54: `|| icp instanceof FromStringToTokenError` does not work?

frontend/src/lib/services/canisters.services.ts
```diff
@@ -70,7 +71,10 @@ export const createCanister = async ({
   account: Account;
 }): Promise<Principal | undefined> => {
   try {
-    const icpAmount = convertNumberToICP(amount);
+    const icpAmount = TokenAmount.fromNumber({ amount }) as TokenAmount;
+    if (!(icpAmount instanceof TokenAmount)) {
```
peterpeterparker, line 75: ditto

frontend/src/lib/stores/transaction-fees.store.ts
```diff
@@ -44,3 +44,8 @@ export const mainTransactionFeeStoreAsIcp = derived(
   transactionsFeesStore,
   ($store) => ICP.fromE8s($store.main)
 );
+
+export const mainTransactionFeeStoreAsToken = derived(
```
peterpeterparker, line 48: to be move to derived folder

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
peterpeterparker: or keep an extended class `export class ICP extends TokenAmount`?

peterpeterparker:
In previous PR we ultimately selected `AmountDisplay` as name but, the testname is `AmountTokens.spec.ts`. Can you rename the test file in this PR?

frontend/src/lib/derived/sns/sns-token-symbol-selected.store.ts
```diff
@@ -1,13 +1,13 @@
...
-export const snsTokenSymbolSelectedStore: Readable<string | undefined> =
-  derived(
-    [snsProjectSelectedStore, snsSummariesStore],
-    ([selectedRootCanisterId, summaries]) =>
-      summaries.find(
-        ({ rootCanisterId }) =>
-          rootCanisterId.toText() === selectedRootCanisterId.toText()
-      )?.token.symbol
-  );
+export const snsTokenSymbolSelectedStore: Readable<Token | undefined> = derived(
```
peterpeterparker, line 6: `Readable<Token | undefined>` is not correct yet. The type of the returned `token` is `SnsTokenMetadata`. So in `SnsSummary` the type of the token should be modified as well. Don't know why the Svelte preprocessor does not complain.

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
lmuntaner: I can remove the optional parameter. I just thought it made sense to be honest. But whatever you want.

I can do that in a couple more PRs later.

We could also use an extended class. But I would do that in NNS Dapp, not at the nns-js level.

Actually, we could also move `AmountToken` to NNS Dapp....

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
lmuntaner: no, `FromStringToTokenError` is an enum.

frontend/src/lib/stores/transaction-fees.store.ts
lmuntaner: yes, along with many others. I did it in a few other stores because I was afraid of circular dependencies.

I'd rather focus a PR to one thing.

frontend/src/lib/derived/sns/sns-token-symbol-selected.store.ts
lmuntaner: Maybe because both types match. But I'll change it.

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
lmuntaner: Another change I wanted to do is to ensure that `fromNumber` always returns an AmountToken. Not the error as well.

frontend/src/lib/stores/transaction-fees.store.ts
lmuntaner: done

frontend/src/lib/derived/sns/sns-token-symbol-selected.store.ts
lmuntaner: done

lmuntaner:

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
peterpeterparker: Even if more verbose, I think it's safer to not set a default value. I typo can slip in a review and we might end up with the wrong symbol.

Can you revert the removal of the `, token: ICPToken` in this PR and make it mandatory in nns-js?

If you want me to make it mandatory, can of course too.

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
peterpeterparker: aaaaah a pity. maybe we should have a use a variant has a return type but sure works.

frontend/src/lib/components/accounts/NewTransactionAmount.svelte
lmuntaner: There are many places where I don't set the token. Not this one.

This deserves two more PRs. One in nns-js and one more in nns-dapp. I can revert this one, but it's not the only one.

lmuntaner: done

lmuntaner:
@peterpeterparker ready for another review, I created an issue to change `TokenAmount` https://dfinity.atlassian.net/browse/L2-1005

frontend/src/lib/components/accounts/NewTransactionReview.svelte
```diff
@@ -22,7 +22,8 @@
     back,
   }: TransactionContext = context;
 
-  let amount: ICPType = $store.amount ?? ICPType.fromE8s(BigInt(0));
+  let amount: TokenAmount =
+    $store.amount ?? TokenAmount.fromE8s({ amount: BigInt(0) });
```
peterpeterparker, line 26: to anticipate the upcoming breaking change, maybe add the `token` here already?

frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
```diff
@@ -48,11 +48,13 @@
     end_timestamp_seconds: undefined,
   });
 
-  export let myCommitment: ICP | undefined = undefined;
+  export let myCommitment: TokenAmount | undefined = undefined;
   $: myCommitment =
     swapCommitment?.myCommitment === undefined
       ? undefined
-      : ICP.fromE8s(swapCommitment.myCommitment.amount_icp_e8s);
+      : TokenAmount.fromE8s({
```
peterpeterparker, line 55: ditto

frontend/src/lib/components/accounts/NewTransactionReview.svelte
peterpeterparker: and all other `.fromE8s` that have been newly added in this PR?

will make the integration of the breaking changes easier, but as you rather like

lmuntaner: I don't want to add it in this PR. One thing at a time. I will have to check all of them anyway. Therefore, adding a few in this PR is not helping me much.

peterpeterparker (approved):
LGTM. Just added one last comment, have a look before merge and process as you rather like.

PR #1316, opened 2022-09-08
Fix: Upgrade failing with new canister state

lmuntaner:
# Motivation

Upgrading the canister was failing due to the change in the state.

The problem was that when decoding, the code was expecting the new state of `pending_transactions` to be already present.

# Changes

* Remove `pending_transactions` when decoding the state and add it as a hash map when creating the new state.

# Tests

No extra test.

### DIFF
--- a/rs/src/accounts_store.rs
+++ b/rs/src/accounts_store.rs
@@ -1358,7 +1358,6 @@ impl StableState for AccountsStore {
         let (
             mut accounts,
             mut hardware_wallets_and_sub_accounts,
-            pending_transactions,
             transactions,
             neuron_accounts,
             block_height_synced_up_to,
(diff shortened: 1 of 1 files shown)


hpeebles (approved):

bitdivine:
# Testing
I tested this upgrade on a local dfx server.  tldr, it works.

There is a weird thing: When deploying, I always checked the hash and compared it with the local wasm; they didn't match.  Not on a clean install, not ever.  When a deployment succeeded both the local wasm hash and the hash reported by dfx canister info changed but they were different.  I'd really love to know what is going on there.

Tags:
* I tagged the release that is actually in prod as "prod"
* I tagged this PR as the release candidate.

Test1:
* deploy prod
* upgrade to the release candidate (this PR) - SUCCESS
* rollback to prod - fails to decode stable memory; this is expected.

Test2:
* Deploy prod
* upgrade to dfx deploy nns-dapp - FAILS to decode stable memory, this is expected

lmuntaner:
We need two more tasks:

* A rollback branch. If the upgrade succeeds but for some other reason we need to rollback, we need to branch off prod and manage the new state.
* Decode the new state. Before the next release, we need to change how to state is decoded again. Basically, we need to revert the changes of this PR.

My question is:

* Shall we merge this as it is to main or do you want to add something?
* When should we start testing? I think we should aim for more testing than we do normally.
* I would expect some fixes to this release candidate. Therefore, how do we want to do them? Shall we use a branch or we add them to main and then we rebase the release candidate branch?

bitdivine (approved):

PR #1320, opened 2022-09-09
L2-841: New transaction flow

lmuntaner:
# Motivation

Implement the new transaction flow.

This will also allow us to implement the new modal UI.

# Changes

* Change TransactionModal so that the user can select a destination or pass a selected account.
* New IcpTransactionModal to transafer ICPs.
* New SelectDestinationAddress to select a destination from current accounts with dropdown or address input.
* Extract AdressInput from Adress to be used in SelectDestinationAddress.
* Changed skipHardwareWallets prop in SelectAccountDropdown for a function to filter accounts. Used to filter the already selected account and also hardware wallets.
* Change UI of the dropdown to make it look more similar to the input.
* New prop showInfo in Input component. There was an extra space from the gap otherwise.

# Tests

* Add test cases in Transaction modal when user passes a destination or a source account initially.
* Test for IcpTransactionModal
* Test for SelectDestinationAddress
* Test for AddressInput
* New test in SelectAccountDropdown to check new filterAccounts prop.

### DIFF
--- a/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { accountsStore } from "../../stores/accounts.store";
-  import { accountsList } from "../../derived/accounts-list.derived";
+  import { accountsListStore } from "../../derived/accounts-list.derived";
   import { i18n } from "../../stores/i18n";
   import type { Account } from "../../types/account";
   import { getAccountFromStore } from "../../utils/accounts.utils";
@@ -19,7 +19,7 @@
     accountsStore: $accountsStore,
   });
 
-  $: selectableAccounts = $accountsList.filter(filterAccounts);
+  $: selectableAccounts = $accountsListStore.filter(filterAccounts);
 </script>
 
 {#if selectableAccounts.length === 0}
--- a/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -10,26 +10,16 @@
   export let selectedAccount: Account | undefined = undefined;
   export let filterAccounts: (account: Account) => boolean = () => true;
 
-  let selectedAccountIdentifier: string;
+  // In case the component is already initialized with a selectedAccount
+  // To avoid cyclical dependencies, we don't update this if `selectedAccount` changes
+  let selectedAccountIdentifier: string | undefined =
+    selectedAccount?.identifier;
   $: selectedAccount = getAccountFromStore({
     identifier: selectedAccountIdentifier,
     accountsStore: $accountsStore,
   });
 
   $: selectableAccounts = $accountsList.filter(filterAccounts);
-
-  // If the selected account is not in the list of accounts
-  // Set the selected account to the first account in the list
-  // It covers the case to select the main account by default
-  $: {
-    if (
-      selectableAccounts.find(
-        ({ identifier }) => identifier === selectedAccountIdentifier
-      ) === undefined
-    ) {
-      selectedAccountIdentifier = selectableAccounts[0]?.identifier;
-    }
-  }
 </script>
 
 {#if selectableAccounts.length === 0}
--- a/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -8,7 +8,7 @@
   import DropdownItem from "../ui/DropdownItem.svelte";
 
   export let selectedAccount: Account | undefined = undefined;
-  export let skipHardwareWallets: boolean = false;
+  export let filterAccounts: (account: Account) => boolean = () => true;
 
   let selectedAccountIdentifier: string;
   $: selectedAccount = getAccountFromStore({
@@ -16,20 +16,27 @@
     accountsStore: $accountsStore,
   });
 
-  let selectableAccounts: Account[] = [];
+  let accounts: Account[] = [];
+  $: selectableAccounts = accounts.filter(filterAccounts);
   const unsubscribe = accountsStore.subscribe(
     ({ main, subAccounts, hardwareWallets }) => {
       if (main !== undefined) {
         selectedAccountIdentifier =
           selectedAccountIdentifier ?? main.identifier;
-        selectableAccounts = [
-          main,
-          ...(subAccounts ?? []),
-          ...(skipHardwareWallets ? [] : hardwareWallets ?? []),
-        ];
+        accounts = [main, ...(subAccounts ?? []), ...(hardwareWallets ?? [])];
       }
     }
   );
+
+  $: {
+    if (
+      selectableAccounts.find(
+        ({ identifier }) => identifier === selectedAccountIdentifier
+      ) === undefined
+    ) {
+      selectedAccountIdentifier = selectableAccounts[0]?.identifier;
+    }
+  }
 
   onDestroy(unsubscribe);
 </script>
--- a/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -41,14 +41,27 @@
   onDestroy(unsubscribe);
 </script>
 
-<Dropdown
-  name="account"
-  bind:selectedValue={selectedAccountIdentifier}
-  testId="select-account-dropdown"
->
-  {#each selectableAccounts as { identifier, name } (identifier)}
-    <DropdownItem value={identifier}>
-      {name ?? $i18n.accounts.main}
+{#if selectableAccounts.length === 0}
+  <Dropdown
+    name="account"
+    disabled
+    selectedValue="no-accounts"
+    testId="select-account-dropdown"
+  >
+    <DropdownItem value="no-accounts">
+      {$i18n.accounts.no_account_select}
     </DropdownItem>
-  {/each}
-</Dropdown>
+  </Dropdown>
+{:else}
+  <Dropdown
+    name="account"
+    bind:selectedValue={selectedAccountIdentifier}
+    testId="select-account-dropdown"
+  >
+    {#each selectableAccounts as { identifier, name } (identifier)}
+      <DropdownItem value={identifier}>
+        {name ?? $i18n.accounts.main}
+      </DropdownItem>
+    {/each}
+  </Dropdown>
+{/if}
--- a/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
+++ b/frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
-  import { onDestroy } from "svelte";
   import { accountsStore } from "../../stores/accounts.store";
+  import { accountsList } from "../../derived/accounts-list.derived";
   import { i18n } from "../../stores/i18n";
   import type { Account } from "../../types/account";
   import { getAccountFromStore } from "../../utils/accounts.utils";
@@ -16,18 +16,11 @@
     accountsStore: $accountsStore,
   });
 
-  let accounts: Account[] = [];
-  $: selectableAccounts = accounts.filter(filterAccounts);
-  const unsubscribe = accountsStore.subscribe(
-    ({ main, subAccounts, hardwareWallets }) => {
-      if (main !== undefined) {
-        selectedAccountIdentifier =
-          selectedAccountIdentifier ?? main.identifier;
-        accounts = [main, ...(subAccounts ?? []), ...(hardwareWallets ?? [])];
-      }
-    }
-  );
+  $: selectableAccounts = $accountsList.filter(filterAccounts);
 
+  // If the selected account is not in the list of accounts
+  // Set the selected account to the first account in the list
+  // It covers the case to select the main account by default
   $: {
     if (
       selectableAccounts.find(
@@ -37,8 +30,6 @@
       selectedAccountIdentifier = selectableAccounts[0]?.identifier;
     }
   }
-
-  onDestroy(unsubscribe);
 </script>
 
 {#if selectableAccounts.length === 0}
--- a/frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
+++ b/frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
@@ -14,10 +14,19 @@
   export let filterAccounts: (account: Account) => boolean = () => true;
   export let showManualAddress: boolean = true;
 
+  // If the component is already initialized with a selectedDestinationAddress
+  let selectedAccount: Account | undefined = getAccountFromStore({
+    identifier: selectedDestinationAddress,
+    accountsStore: $accountsStore,
+  });
   let address: string;
   $: {
     if (!invalidAddress(address)) {
       selectedDestinationAddress = address;
+    }
+    // Keep in sync the selected destination address
+    if (selectedAccount !== undefined) {
+      selectedDestinationAddress = selectedAccount.identifier;
     }
   }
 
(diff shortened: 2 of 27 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/189340120-3cd0cb69-8876-4a61-9bd5-c8d92ac9827b.mov

https://user-images.githubusercontent.com/4550653/189340138-6fdd90f5-77b2-456d-9101-b8dd11efc45e.mov

When the account is passed. For example from the Wallet page.
Screenshot 2022-09-09 at 13 19 39

When the user has only the Main account:
Screenshot 2022-09-09 at 13 20 24

lmuntaner:

There is still pending to increase neuron stake with this new flow to be able to remove the old NewTransactionModal.

The issue is here: https://dfinity.atlassian.net/browse/GIX-1014

peterpeterparker:
The deployed branch is down so I tested locally. Here few feedback / questions about UI:

- (figure 1) instead of display a disabled dropdown when no other accounts can be selected have you thought about displayed no dropdown and no toggler?

- (figure 2) not necessary for this PR but the clickable zone in the dropdown is too narrow

- (figure 3) not sure it is still actual but in Mischa figma I think there was the word "Balance" before the display of the balance of the source account

- (figure 3) shouldn't the name of the selected account be displayed as a value color wise?

- (figure 4) when going back and forth from review to first screen, the selected account gets lost

figure 1:

Capture d’écran 2022-09-13 à 07 26 43

figure 2:

Capture d’écran 2022-09-13 à 07 28 26

figure 3:

Capture d’écran 2022-09-13 à 07 37 06

figure 4:

a

frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
```diff
@@ -8,40 +8,60 @@
...
+  export let filterAccounts: (account: Account) => boolean = () => true;
 
   let selectedAccountIdentifier: string;
   $: selectedAccount = getAccountFromStore({
     identifier: selectedAccountIdentifier,
     accountsStore: $accountsStore,
   });
 
-  let selectableAccounts: Account[] = [];
+  let accounts: Account[] = [];
```
peterpeterparker, line 19: what for it this new `accounts` variable needed?

frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
```diff
@@ -8,40 +8,60 @@
...
-          ...(skipHardwareWallets ? [] : hardwareWallets ?? []),
-        ];
+        accounts = [main, ...(subAccounts ?? []), ...(hardwareWallets ?? [])];
       }
     }
   );
 
+  $: {
+    if (
+      selectableAccounts.find(
```
peterpeterparker, line 33: not sure I get why we need this, can you add a comment?

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
```diff
@@ -0,0 +1,67 @@
...
+      <Toggle
+        bind:checked={showManualInput}
+        on:nnsToggle={onToggleManualInput}
+        ariaLabel="change"
+      />
+      <p>{$i18n.accounts.manual}</p>
+    </div>
+  </div>
+  {#if showManualInput}
+    <AddressInput bind:address on:nnsBlur={setDestination} />
```
peterpeterparker, line 45: on blur is a must or can we use input?

input ux wise would be more convenient

frontend/src/lib/components/ui/Dropdown.svelte
```diff
@@ -65,9 +68,10 @@
       // Place the caret inside the select.
       margin-left: calc(-1 * var(--padding-3x));
 
+      // To match the line-height when font-size is 16px
       :global(svg) {
-        width: 24px;
-        height: 24px;
+        width: 20px;
```
peterpeterparker, line 73: since we are already here, is it possible to add a `size` property to the icon and avoid the use of `global` style? or are we styling an icon that is not a direct child?

lmuntaner:
Thanks for testing! This is important enough to check it properly. That's why I deployed it.

* (1) I will check with Mischa what to do in that case.
* (2) This is a known issue: https://dfinity.atlassian.net/browse/GIX-892
* (3 both) I checked all the flow with Mischa before the PR. I will ask him to check again the screenshots and flows.
* (4) Ah, yes, the state is lost. Let me refactor that.

frontend/src/lib/modals/accounts/IcpTransactionModal.svelte
```diff
@@ -0,0 +1,72 @@
+<script lang="ts">
```
peterpeterparker, line 1: nothing new in `IcpTransactionModal` right, just refactoring?

frontend/src/lib/components/ui/Dropdown.svelte
lmuntaner: This is the way we style svg in the project. There are 6 other instances where we do this, that's why I did it this way:

* Logout
* Theme toggle
* Badge New
* Collpasible
* FiltersButton
* VoteConfirmationModal

I can change it here. Does that mean we should also change it everywhere? Why did the others get merged if it was not the preferred pattern?

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
```diff
@@ -78,20 +96,38 @@
...
+        <span slot="key" class="label">{$i18n.accounts.source}</span>
         <AmountDisplay
           slot="value"
           singleLine
           amount={selectedAccount?.balance}
         />
       </KeyValuePair>
     {/if}
-    <SelectAccountDropdown bind:selectedAccount skipHardwareWallets />
+    {#if canSelectSource}
```
peterpeterparker, line 107: are there tests for `if canSelectSource / canSelectDestination`?

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
lmuntaner: It is not a must, but that is when we also check whether the input is a valid address and we can set it to the variable.

Otherwise, we would have an invalid destination address.

And validating on blur is a must from Mischa that preferrs to not show the error inputs while the user is typing. But wait until the user has finished typing.

This is the current implementation, maybe we can change it in another task.

peterpeterparker: We currently use `input` (or `change`, did no check the code but not `blur`) on `mainnet` - i.e. when a user input an address or copy/paste an address in an input field, the content is validated and use can continue. This change adds an extra step where user has in addition to click elsewhere or hit tab to trigger the validation, which is not as convenient as currently on `mainnet`. So I would stick to the existing UX regarding this validation.

frontend/src/lib/components/ui/Dropdown.svelte
peterpeterparker: With "since we are already there.." I meant "since the size is updated in this PR, maybe we can have a look if we cannot spare the `:global` selector here at the same time because maybe the `:global` is not useful anymore for that case and it can be solved differently since generally speaking it's better to not use `:global` selector?"

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
lmuntaner: Where do you see that we use `input` or `change` for the address on `mainnet`?

I see on blur here: https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/components/accounts/Address.svelte#L26

```
    <InputWithError
      inputType="text"
      placeholderLabelKey="accounts.address"
      name="accounts-address"
      bind:value={address}
      minLength={ACCOUNT_ADDRESS_MIN_LENGTH}
      errorMessage={showError ? $i18n.error.address_not_valid : undefined}
      on:blur={showErrorIfAny}
    />
```

frontend/src/lib/components/ui/Dropdown.svelte
lmuntaner: I'll use a prop

frontend/src/lib/modals/accounts/IcpTransactionModal.svelte
lmuntaner: correct, nothing new

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
lmuntaner: Yes, there are tests to check what is shown:

```
  describe("when destination address is provided", () => {
    it("should not show the select destination component", async () => {
      const { queryByTestId, container } = await renderTransactionModal({
        destinationAddress: mockMainAccount.identifier,
      });

      expect(queryByTestId("select-account-dropdown")).toBeInTheDocument();
      expect(queryByTestId("select-destination")).not.toBeInTheDocument();
      expect(
        container.querySelector("input[name='amount']")
      ).toBeInTheDocument();
    });
  });

  describe("when source account is provided", () => {
    it("should not show the select account component", async () => {
      const { queryByTestId, container } = await renderTransactionModal({
        sourceAccount: mockMainAccount,
      });

      expect(queryByTestId("select-account-dropdown")).not.toBeInTheDocument();
      expect(
        container.querySelector("input[name='amount']")
      ).toBeInTheDocument();
    });
  });
```

But I didn't do all the flow with each part. Is that ok?

frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
lmuntaner: sure

frontend/src/lib/components/accounts/SelectAccountDropdown.svelte
lmuntaner: Earler, ths filter was done inside the `subscribe`, where the hardware wallets where skipped.

That meant the accounts when filtered on change of the accounts stare.

But the accounts now are filtered by a prop which is a function. And I want to filter when that prop changes as well.

Therefore, accounts just adds together all the accounts and then filters them.

I realized that this could be a derived store. I'll do that.

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
peterpeterparker: I mean this different behavior versus your PR and mainnet:

abc

lmuntaner: ok, thanks, I didn't notice, I thought you meant the error. I'll review that.

frontend/src/lib/components/ui/Dropdown.svelte
lmuntaner: I created an issue to deal with this because the icons go moved to gix-components library and solving this is not as easy as doing it now in this PR

https://dfinity.atlassian.net/browse/GIX-1016

lmuntaner:
> The deployed branch is down so I tested locally. Here few feedback / questions about UI:
> 
> * (figure 1) instead of display a disabled dropdown when no other accounts can be selected have you thought about displayed no dropdown and no toggler?
> * (figure 2) not necessary for this PR but the clickable zone in the dropdown is too narrow
> * (figure 3) not sure it is still actual but in Mischa figma I think there was the word "Balance" before the display of the balance of the source account
> * (figure 3) shouldn't the name of the selected account be displayed as a value color wise?
> * (figure 4) when going back and forth from review to first screen, the selected account gets lost

Mischa is in training this week and doesn't seem available.

figure 1 - expects his input
figure 2 - known problem with issue to fix it.
figure 3 (both) - expects his input
figure 4 - fixed, see video below

https://user-images.githubusercontent.com/4550653/190142045-cbd0d5fd-78f1-4a19-aaff-100b37a1069a.mov

@peterpeterparker can you do another review and I will fix the issues if needed when Mischa is available. He said he won't be available the whole week and I'd like to have this one merged this week.

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
```diff
@@ -0,0 +1,79 @@
...
+    selectedDestinationAddress = undefined;
+    selectedAccount = undefined;
+  };
+
+  // If the component is already initialized with a selectedDestinationAddress
+  let selectedAccount: Account | undefined = getAccountFromStore({
+    identifier: selectedDestinationAddress,
+    accountsStore: $accountsStore,
+  });
+  // Keep in sync the selected destination addres
```
peterpeterparker, line 35: typo "addres"

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
```diff
@@ -0,0 +1,79 @@
...
+  import Toggle from "../ui/Toggle.svelte";
+  import AddressInput from "./AddressInput.svelte";
+  import SelectAccountDropdown from "./SelectAccountDropdown.svelte";
+
+  export let selectedDestinationAddress: string | undefined = undefined;
+  export let filterAccounts: (account: Account) => boolean = () => true;
+  export let showManualAddress: boolean = true;
+
+  let address: string;
+  $: {
```
peterpeterparker, line 18: can you group both auto subscribe in one block? I think it would make things more readable because in both `selectedDestinationAddress` is assigned

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
```diff
@@ -0,0 +1,79 @@
...
+    .toggle {
+      display: flex;
+      flex-direction: row;
+      align-items: center;
+      gap: var(--padding);
+    }
+  }
+
+  .label {
+    color: var(--label-color);
```
peterpeterparker, line 77: do you need this? `<p class="label">` does not inherit the `.label` global style?

frontend/src/lib/components/ui/Input.svelte
```diff
@@ -116,10 +121,12 @@
 </script>
 
 <div class="input-block" class:disabled>
-  <div class="info">
-    <label for={name}><slot name="label" /></label>
-    <slot name="additional" />
-  </div>
+  {#if showInfo}
+    <div class="info">
+      <label for={name}><slot name="label" /></label>
```
peterpeterparker, line 126: `class="label"` would work?

frontend/src/lib/derived/accounts-list.derived.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * A derived store that returns the accounts as an array of accounts.
+ */
+
+import { derived } from "svelte/store";
+import { accountsStore } from "../stores/accounts.store";
+
+export const accountsList = derived(
```
peterpeterparker, line 8: should we have a test that assert that empty is returned when no main is defined?

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
lmuntaner: thanks

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
lmuntaner: I found it more readable the other way, haha. I'll put them in the same.

frontend/src/lib/modals/accounts/IcpTransactionModal.svelte
```diff
@@ -0,0 +1,72 @@
...
+      token: ICPToken,
+    }) as TokenAmount;
+
+    const { success } = await transferICP({
+      selectedAccount: sourceAccount,
+      destinationAddress,
+      amount: tokenAmount,
+    });
+
+    if (success) {
```
peterpeterparker, line 46: if not success, the toast error is thrown in the service?

frontend/src/lib/components/ui/Input.svelte
lmuntaner: I guess, I'll try

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
peterpeterparker: no no super, thanks

frontend/src/lib/derived/accounts-list.derived.ts
lmuntaner: sure

frontend/src/lib/components/accounts/SelectDestinationAddress.svelte
peterpeterparker: https://ipi2f-uqaaa-aaaad-aabza-cai.ic0.app/utility-classes/text-opacity

docs from Maks - see "CSS class"

but again, worth a try first, maybe we do a particular style here

lmuntaner:

peterpeterparker (approved):
LGTM, thanks

PR #1322, opened 2022-09-12
feat: new previous and next navigation for proposals

peterpeterparker:
# Motivation

Introduces the new "previous" and "next" proposal navigation feature pictured in the UI/UX design.

# Note

For simplicity reason, if user access the proposal with a deep link, the navigation won't be displayed as the list of proposals won't be loaded - i.e. the navigation is only available if a list of proposals have been displayed.

The navigation follows the list of proposals. e.g. if following proposal ids are displayed

1
33
598

Using previous and next will navigate through those. When optimistic voting is use, e.g.

1
33 <- voting here
598

User can still navigate to 1 or 598 while vote is in progress but once navigated, if user has selected the filter "Display Open proposal ...." and if the voting is in progress, the proposal 33 will be ignored. Same as for the display of the list.

In that way both "Previous and next" offers the same behavior as what user gets using the list.

# PRs

gix-component PRs needed:

- [x] https://github.com/dfinity/gix-components/pull/58
- [x] https://github.com/dfinity/gix-components/pull/59

# Changes

- new component `<ProposalNavigation />`
- refactor `hide` property of `<ProposalCard />` to derived stores (we do not need to perform the display/hide check on the UI side anymore since card that are not displayed are not rendered in the DOM anymore)
- fix display of proposals if no topics is selected (see small changes in `proposals.services`)

# Screenshots

Enregistrement de l’écran 2022-09-12 à 12 29 14


mstrasinskis (approved):
LGTM. Looking forward to see in prod!

PR #1323, opened 2022-09-12
feat: don't display neurons collapsible if no neurons

peterpeterparker:
# Motivation

If there are no neurons at all, the "Neurons" collapsible element can be simply not displayed.

# Changes

- display voting section only if there are neurons


mstrasinskis (approved):
LGTM. Thanks

PR #1332, opened 2022-09-15
L2-1020: Decode pending_transactions

lmuntaner:
# Motivation

Do not lose the pending_transactions when upgrading canisters.

# Changes

* When decoding, expect also the pending_transactions Hashmap.
* Use the decoded to create the AccountsStore.

# Tests

Not applicable

### DIFF
--- a/rs/src/accounts_store.rs
+++ b/rs/src/accounts_store.rs
@@ -1357,6 +1357,7 @@ impl StableState for AccountsStore {
         let (
             mut accounts,
             mut hardware_wallets_and_sub_accounts,
+            pending_transactions,
             transactions,
             neuron_accounts,
             block_height_synced_up_to,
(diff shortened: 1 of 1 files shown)


lmuntaner:
@bitdivine please review

bitdivine (approved):
Looks good to me, thanks.  How can this be tested?

lmuntaner:
@bitdivine I tested it the following way:

* I deployed the prod branch to a testnet.
* From localhost, I added a pending transaction to participate in a swap.
* The user never made the trasaction though. Therefore, the pending_transaction should stay in the state.
* I upgraded the canister to this branch. The upgrade succeeded.
* I tested with the same user, participating in the same swap. But this time, the user did only the transaction, never notifying.
* The participation went through because it used the previous pending_transaction added before the upgrade.

Does this make sense? Do you think it's test enough?

Not sure how to add unit tests or even do more than this automatically without spending days on it.

PR #1334, opened 2022-09-15
Give each workflow a 'passes' job

bitdivine:
# Motivation
It can be confusing why CI does not pass and in some cases there is little developers can do to fix it without admin rights.

# Changes
- Create one "pass" job for each workflow file - even if not currently used to gate PRs.
- Once this is merged, the pass rules can be used.  They already are to some extent but not enough.

# Tests
See CI


mstrasinskis (approved):
Rubber stamp

PR #1335, opened 2022-09-15
L2-1002: Clean Pending Transactions state

lmuntaner:
# Motivation

An attacker could overflow the memory by adding A LOT OF pending_transactions.

# Changes

* Add timestamp to the pending_transaction
* When adding a new pending transaction, remove pending transactions older than 1h.
* When adding a new pending transaction, remove the last one if the limit has been reached.

# Tests

* New tests for both new functionalities in add_pending_transaction


lmuntaner:
@hpeebles @bitdivine please review

rs/src/accounts_store.rs
```diff
@@ -495,6 +497,18 @@ impl AccountsStore {
...
+    pub fn should_prune_pending_transactions(&self) -> bool {
+        // Valid pending transactions are very short lived.
+        // If there are many, it's because it's filled with invalid pending transactions.
+        self.pending_transactions.len() > TRANSACTIONS_COUNT_LIMIT as usize
+    }
+
+    pub fn prune_pending_transactions(&mut self) {
+        // Worst case scenario a valid pending transaction is removed.
+        // The user needs to ask for a refund or notify manually if not already done.
+        self.pending_transactions = HashMap::new();
```
bitdivine, line 509: Why throw everything away?  An alternative would be to throw away random keys until we are below the limit.  This could be done at insertion time as well as then you are guaranteed never to go over the limit.
```
        // Note: key order is random so this removes a random element from the map.
        while self.should_prune_pending_transactions() {
          if let Some(key) = self.pending_transactions.keys().next().map(|key| *key) {
            self.pending_transactions.remove(&key);
          } else {
            break
          }
        }
```

bitdivine: By the way, do transactions time out or will the hash map gradually fill up with junk?  Those 1_000_000 transactions are parsed at upgrade time, which makes me nervous.  How many cycles does each map entry consume at the upgrade?

bitdivine (approved):
This works but I wonder whether it's a bit violent.  I would rather:
- Reduce the maximum number of pending transactions to something like the typical 98%tile, i.e. the number of pending transactions is below the limit 98% of the time.  So something like 1000.  Probably even less.
- If full, eject just as many entries as we need rather than throwing everything away.

This would mean that the map acts as a safety net for people who disconnect before payments are complete but doesn't dedicate so many resources to the safety net that the nns-dapp itself is put in danger.

rs/src/accounts_store.rs
```diff
@@ -495,6 +497,18 @@ impl AccountsStore {
...
 
+    pub fn should_prune_pending_transactions(&self) -> bool {
+        // Valid pending transactions are very short lived.
+        // If there are many, it's because it's filled with invalid pending transactions.
+        self.pending_transactions.len() > TRANSACTIONS_COUNT_LIMIT as usize
+    }
+
+    pub fn prune_pending_transactions(&mut self) {
+        // Worst case scenario a valid pending transaction is removed.
+        // The user needs to ask for a refund or notify manually if not already done.
```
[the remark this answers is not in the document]
lmuntaner: The frontend also notifies of the participation. Therefore, if it doesn't succeed the notify, then the user sees an error message. Yet, if only that failed, when the user refreshes, the participation might already be through.

Therefore yes, this hash map is only there to increase reliability.

rs/src/accounts_store.rs
lmuntaner: Each transaction is removed from the hash map as it is "notified". Therefore, I expect the size to be always very small.

In a valid pending_transaction, it's only some seconds that the nns-dapp holds it in state:

* User adds pending_transaction
* User makes transaction
* NNS Dapp reads transaction
* NNS Dapp checks transaction with pending_transaction
* NNS Dapp performs action of transaction (for now only "notify")
* NNS Dapp removes pending_transaction from map

What security was concerned was with an attacker filling up the state with junk. If we remove only one per heartbeat, not sure this is safe enough.

I consider that the moment that there are more than 1M pending_transactions to be due to a malicious actor. Therefore, dumping everything seems a valid option.

Because as you mentioned, this is just to make the site more reliable. It is not mandatory for the functionality.

rs/src/accounts_store.rs
```diff
@@ -30,8 +31,8 @@ pub struct AccountsStore {
     // TODO(NNS1-720): Use AccountIdentifier directly as the key for this HashMap
     accounts: HashMap<Vec<u8>, Account>,
     hardware_wallets_and_sub_accounts: HashMap<AccountIdentifier, AccountWrapper>,
-    // pending_transactions: HashMap<(from, to), TransactionType>
-    pending_transactions: HashMap<(AccountIdentifier, AccountIdentifier), TransactionType>,
+    // pending_transactions: HashMap<(from, to), (TransactionType, timestamp)>
```
bitdivine, line 34:
```suggestion
    // pending_transactions: HashMap<(from, to), (TransactionType, timestamp_ms_since_epoch)>
```

bitdivine: Does this change in type need a special deployment?  Right now we do load the old hash map but it has a different structure?

rs/src/accounts_store.rs
```diff
@@ -479,9 +491,36 @@ impl AccountsStore {
         caller == buyer
     }
 
+    fn prune_old_pending_transactions(&mut self, now_millis: u64) {
+        const HOUR_IN_MILLISECONDS: u64 = 1_000 * 60 * 60;
+        let one_hour_ago = now_millis - HOUR_IN_MILLISECONDS;
+        // Keep pending transactions of the last hour only
+        self.pending_transactions
+            .retain(|_, (_, timestamp)| *timestamp > one_hour_ago);
+    }
```
bitdivine, line 498: Looks good!

rs/src/accounts_store.rs
```diff
@@ -479,9 +491,36 @@ impl AccountsStore {
...
+            .pending_transactions
+            .iter()
+            .max_by(|(_, (_, timestamp1)), (_, (_, timestamp2))| timestamp1.cmp(timestamp2))
+        {
+            Some((k, _)) => {
+                self.remove_pending_transaction(*k);
+            }
+            None => (),
+        }
+    }
```
bitdivine, line 511: Remove last is O(n), as it scans the entire map.  When the map is full that cost will be paid for every single insertion until the heartbeat makes space.  If it is full of new records because there is some kind of denial of service attack, the heartbeat won't remove any records.  An attacker would presumably make 10,000 spurious entries quickly, then wait an hour - in that time all users will suffer relatively slow insertions as the heartbeat won't remove anything, then just before 1h is up, make another 10,000 insertions and so on.  This would maximise the cost.  That cost might be affordable but we should be aware of it.

lmuntaner: Yes, I agree. That is still a risk.

rs/src/accounts_store.rs
lmuntaner: only one pending transaction will be added. But it's ok if there is only one notify for two transactions. It updates to the full balance, not the amount per transaction.

I don't think we need a special deployment because the map is empty in mainnet. That said. I will do an upgrade test.

rs/src/accounts_store.rs
```diff
@@ -470,7 +471,16 @@ impl AccountsStore {
         to: AccountIdentifier,
         transaction_type: TransactionType,
     ) -> AddPendingTransactionResponse {
-        self.pending_transactions.insert((from, to), transaction_type);
+        // Call to `time` needs the unsafe
+        let now_millis = time_millis();
+        self.prune_old_pending_transactions(now_millis);
```
hpeebles, line 476: We could only prune the old transactions when the count > some value, say 1000.
And if after pruning the count is still above some limit we could then prune the oldest 100 or so.
Then we no longer need any other checks on the amount of pending transactions.

rs/src/accounts_store.rs
```diff
@@ -1367,7 +1411,7 @@ impl StableState for AccountsStore {
         ): (
             HashMap<Vec<u8>, Account>,
             HashMap<AccountIdentifier, AccountWrapper>,
-            HashMap<(AccountIdentifier, AccountIdentifier), TransactionType>,
+            HashMap<(AccountIdentifier, AccountIdentifier), (TransactionType, u64)>,
```
hpeebles, line 1414: Even though the map will be empty I think this may still fail due to candid first checking type compatibility before deserializing.
You may need to deserialize to the old type but then ignore that value and initialize the map to a new empty map.
But I guess you'll find out when you try upgrading a test instance.

rs/src/accounts_store.rs
lmuntaner: Doing the check makes sense. Although we might have (up to 1000) stale data. This way, we make sure that we always have up-to-date pending transactions. The trade-off is about the extra O(n) operation.

Do you mean to prune the last 100 instead of the last one?

The idea I thought was to remove the last one, to make space. And when the extra are expired, then they would be removed when pruning the old ones. This way there was no need to prune the last 100 and only the last one was needed. It would mean 1h or so of extra O(n), but I believe it's ok.

I can add the idea of checking the length, at the expense of the stale data. But I am not sure about the advantage of removing the last 100 vs the last one. That the next call will be faster?

rs/src/accounts_store.rs
lmuntaner: Yes, I'll check that on the upgrade test. I am also not sure what would happen.

rs/src/accounts_store.rs
```diff
@@ -470,7 +471,18 @@ impl AccountsStore {
         to: AccountIdentifier,
         transaction_type: TransactionType,
     ) -> AddPendingTransactionResponse {
-        self.pending_transactions.insert((from, to), transaction_type);
+        // Call to `time` needs the unsafe
```
hpeebles, line 474: You can remove this comment.

hpeebles (approved):

rs/src/accounts_store.rs
lmuntaner: yes, thanks!

PR #1339, opened 2022-09-16
feat: csp script-src fallback for older devices

peterpeterparker:
# Motivation

CSP rule `'strict-dynamic'` is supported by major modern browsers. However Safari does support it only since Marc 2022.
Therefore, to support older devices, we add `'unsafe-inline'` and `https:` to the `script-src` rule.

Note that this unsecure rule will be ignored by the modern browsers that does support `'strict-dynamic'` and therefore remain secure.

# Changes

- add `'unsafe-inline'` and `https:` to the `script-src` rule


frederikrothenberger (approved):
Thanks! 👍

PR #1343, opened 2022-09-18
feat: reset auth client after sign out

peterpeterparker:
# Motivation

Noticed a bug in Papyrs in a "signin -> signout -> signin again" as I don't reload the screen. We do not have the issue in NNS-dapp but, to prevent possible future issue, this PR reset `authClient` after `logout`.

# Changes

- reset in memory `authClient` object after sign-out


mstrasinskis (approved):
LGTM 👍👍

PR #1345, opened 2022-09-19
L2-993: User can see SNS main balance account

lmuntaner:
# Motivation

User can see the balance of the main account in an SNS project.

# Changes

* New store: snsAccountsStore.
* New derived store: snsProjectAccountsStore which combines the selectedProject and the snsAccountsStore.
* New sns api in "sns-ledger.api": getSnsAccounts. For now it returns only a Main account.
* New sns accounts service: loadSnsAccounts that users getSnsAccounts api and snsAccountsStore.
* loadSnsAccounts service is called in SnsAccounts page.
* Render AccountCard for each account in SnsAccounts and the total of tokens.

# Tests

* Test new store and derived.
* Test new sns api function
* Test new sns service
* Test new functionality in SnsAccounts.

### DIFF
--- a/frontend/src/lib/api/sns-ledger.api.ts
+++ b/frontend/src/lib/api/sns-ledger.api.ts
@@ -1,0 +1,44 @@
+import type { Identity } from "@dfinity/agent";
+import { TokenAmount } from "@dfinity/nns";
+import type { Principal } from "@dfinity/principal";
+import type { Account } from "../types/account";
+import { logWithTimestamp } from "../utils/dev.utils";
+import { mapOptionalToken } from "../utils/sns.utils";
+import { wrapper } from "./sns.api";
+
+export const getSnsAccounts = async ({
+  rootCanisterId,
+  identity,
+  certified,
+}: {
+  rootCanisterId: Principal;
+  identity: Identity;
+  certified: boolean;
+}): Promise<Account[]> => {
+  // TODO: Support subaccounts
+  logWithTimestamp("Getting sns accounts: call...");
+  const { balance, ledgerMetadata } = await wrapper({
+    identity,
+    rootCanisterId: rootCanisterId.toText(),
+    certified,
+  });
+
+  const [mainBalanceE8s, metadata] = await Promise.all([
+    balance({ owner: identity.getPrincipal() }),
+    ledgerMetadata({}),
+  ]);
+
+  const mainAccount: Account = {
+    // TODO: Implement string representation https://dfinity.atlassian.net/browse/GIX-1025
+    identifier: "sns-main-account-identifier",
+    principal: identity.getPrincipal(),
+    balance: TokenAmount.fromE8s({
+      amount: mainBalanceE8s,
+      token: mapOptionalToken(metadata),
+    }),
+    type: "main",
+  };
+
+  logWithTimestamp("Getting sns neuron: done");
+  return [mainAccount];
+};
--- a/frontend/src/lib/api/sns-ledger.api.ts
+++ b/frontend/src/lib/api/sns-ledger.api.ts
@@ -4,7 +4,7 @@ import type { Principal } from "@dfinity/principal";
 import type { Account } from "../types/account";
 import { logWithTimestamp } from "../utils/dev.utils";
 import { mapOptionalToken } from "../utils/sns.utils";
-import { wrapper } from "./sns.api";
+import { wrapper } from "./sns-wrapper.api";
 
 export const getSnsAccounts = async ({
   rootCanisterId,
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -1,8 +1,7 @@
-import type { HttpAgent, Identity } from "@dfinity/agent";
-import type { DeployedSns, ICP, SnsWasmCanister } from "@dfinity/nns";
+import type { Identity } from "@dfinity/agent";
+import type { ICP } from "@dfinity/nns";
 import { Principal } from "@dfinity/principal";
 import type {
-  InitSnsWrapper,
   SnsNeuron,
   SnsNeuronId,
   SnsNeuronPermissionType,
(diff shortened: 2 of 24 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/191020171-d3ee46b8-cf86-4dec-aad3-301c88f533c9.mov

Screenshot 2022-09-19 at 14 43 31

frontend/src/lib/api/sns-ledger.api.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const [mainBalanceE8s, metadata] = await Promise.all([
+    balance({ owner: identity.getPrincipal() }),
+    ledgerMetadata({}),
+  ]);
+
+  const mainAccount: Account = {
+    // TODO: Implement string representation https://dfinity.atlassian.net/browse/GIX-1025
+    identifier: "sns-main-account-identifier",
+    principal: identity.getPrincipal(),
```
peterpeterparker, line 34: should the principal be set here?

in the account types there is following comment:

```
// Main and HardwareWallet accounts have Principal
  principal?: Principal;
```

if it's correct, should we improve related comment to make clear that any main account needs a principal?

frontend/src/lib/api/sns.api.ts
```diff
@@ -174,7 +174,7 @@ const wrappers = async ({
   }
 };
 
-const wrapper = async ({
+export const wrapper = async ({
```
peterpeterparker, line 177: - if exported, can we add a dedicated test?

- should we refactor it in a dedicated file like `sns-wrapper.api.ts`?

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
```diff
@@ -0,0 +1,28 @@
...
+import { snsProjectSelectedStore } from "../selected-project.derived";
+
+/**
+ * Main account is put in the first position. The rest of the accounts keep the same order.
+ *
+ * @param accounts: Array of accounts
+ * @returns accounts
+ */
+const sortAccounts = (accounts: Account[]): Account[] => {
+  const nonMainAccounts: Account[] = accounts.filter(
```
peterpeterparker, line 13: naming question: aren't `nonMainAccounts` actually `subAccounts`?

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
```diff
@@ -0,0 +1,28 @@
...
+import type { Account } from "../../types/account";
+import { snsProjectSelectedStore } from "../selected-project.derived";
+
+/**
+ * Main account is put in the first position. The rest of the accounts keep the same order.
+ *
+ * @param accounts: Array of accounts
+ * @returns accounts
+ */
+const sortAccounts = (accounts: Account[]): Account[] => {
```
peterpeterparker, line 12: should it be named `sortedAccounts` instead of `sortAccounts`?

quick search we rather use `sorted...` in existing derived stores we have

peterpeterparker, line 12: is there a test about this sorting order? maybe I missed it...

frontend/src/lib/pages/NnsAccounts.svelte
```diff
@@ -23,28 +23,19 @@
...
-      ...(accounts?.subAccounts || []).map(({ balance }) => balance),
-      ...(accounts?.hardwareWallets || []).map(({ balance }) => balance)
-    );
-    totalICP = formatICP({
-      value: totalBalance.toE8s(),
-      detailed: true,
-    });
-  }
+  let totalBalance: TokenAmount | undefined;
+  $: totalBalance =
```
peterpeterparker, line 27: should we have a test for this? Asking because I guess we did not had one right?

frontend/src/lib/pages/SnsAccounts.svelte
```diff
@@ -1,17 +1,56 @@
...
-      symbol: "test",
-      name: "todo",
-    },
-  });
+  let loading: boolean = false;
+  const unsubscribe: Unsubscriber = snsOnlyProjectStore.subscribe(
+    async (selectedProjectCanisterId) => {
+      if (selectedProjectCanisterId !== undefined) {
+        loading = true;
+        await loadSnsAccounts(selectedProjectCanisterId);
```
peterpeterparker, line 20: not necessarely in this PR but, should we reload each time the sns accounts or should we keep these in memory once loaded once?

frontend/src/lib/stores/sns-accounts.store.ts
```diff
@@ -0,0 +1,62 @@
...
+      }));
+    },
+
+    // Used in tests
+    reset() {
+      set({});
+    },
+
+    resetProject(rootCanisterId: Principal) {
+      update((currentState: SnsAccountsStore) =>
```
peterpeterparker, line 53: can't it be simplified with es6? something like deconstructing the rootCanisterId and the ...rest maybe?

something like `update(({[rootCanisterId], ...rest}: SnsAccountsStore) =>` would work?

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -89,7 +89,7 @@ const mapOptionalMetadata = ({
 /**
  * Token metadata is given only if the properties NNS-dapp needs (name and symbol) are defined.
  */
-const mapOptionalToken = (
+export const mapOptionalToken = (
```
peterpeterparker, line 92: since we expose it, should we have a test?

frontend/src/lib/api/sns.api.ts
lmuntaner: sure

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
lmuntaner: I did a test for the derived store. Not for the sorting there though. But I can add it as well.

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
lmuntaner: They can also be hardware wallet accounts. Not yet supported, but maybe in the future.

frontend/src/lib/pages/NnsAccounts.svelte
lmuntaner: A test that we don't show a 0 when no account? I can add it.

frontend/src/lib/pages/SnsAccounts.svelte
lmuntaner: I would keep them in memory, the same way we don't remove the accounts when we go to neurons.

frontend/src/lib/stores/sns-accounts.store.ts
lmuntaner: I can try.

lmuntaner: no, it doesn't work. Also because rootCanisterId needs to be converted to text as well...

frontend/src/lib/api/sns-ledger.api.ts
lmuntaner: Mhm, not sure I understand what you want. The comment is indeed correct. What is missing here?

The difference with sns accounts now is that I don't rely on the nns dapp canister to store accounts. I will implement that when we implement subaccounts. But just for the main account, I can "mock" as if it was stored in the nns-dapp withuot actually storing it. And the principal if the main account is the identity's principal.

frontend/src/lib/api/sns.api.ts

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
lmuntaner: Added a test the checks that the first of the derived store is of type "main".

lmuntaner:
@peterpeterparker addressed the comments, ready for another review

frontend/src/lib/api/sns-ledger.api.ts
peterpeterparker: nvm it's a bit confusing to have a `main` account NNS and `main`account SNS while you cannot have  `main` account Sns without a main account NNS 😜

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
peterpeterparker: oki doki

frontend/src/lib/pages/SnsAccounts.svelte
peterpeterparker: In that case, it is not what's implemented. Each time the selected project canister id change, the sns accounts are reloaded from the backend.

Or I am missing something?

frontend/src/lib/pages/SnsAccounts.svelte
```diff
@@ -1,17 +1,56 @@
...
+
+  // TODO: Wallet details https://dfinity.atlassian.net/browse/GIX-995
+  const goToDetails = (account: Account) => {
+    console.log("goToDetails", account);
+  };
 </script>
 
 <section data-tid="sns-accounts-body">
   <AccountsTitle balance={totalAmountToken} />
+  {#if loading}
```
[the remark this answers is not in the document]
peterpeterparker: I mean following is the exact same code:

```
<AccountCard
        role="link"
        on:click={() => goToDetails(account)}
        showCopy
        {account}>{account.name ?? $i18n.accounts.main}</AccountCard
      >
```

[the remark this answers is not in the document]
peterpeterparker: What I meant is that the components code has probably some duplication that can be avoided by extracting a or some components without derived store or else.

frontend/src/lib/stores/sns-accounts.store.ts
peterpeterparker: thanks for giving a try 👍.

actually, just noticed, this is the exact same piece of code as in `sns-neurons.store`. Can you refactor to avoid code duplication?

frontend/src/lib/pages/SnsAccounts.svelte
lmuntaner: True, we decide to not show it because we set it to loading. Maybe we want to show it and update in the background?

frontend/src/lib/pages/SnsAccounts.svelte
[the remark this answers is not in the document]
lmuntaner: Yes, that's how a component it's used and the props.

Do you prefer I create a `LinkAccountCard` for example that expects just one account as prop and renders an `AccountCard`? Although `goToDetails` would still need to be there.

For my taste, I prefer the code repetition. But if you want I can create the extra component.

frontend/src/lib/stores/sns-accounts.store.ts
lmuntaner: maybe I can use an util, yes

lmuntaner: done

lmuntaner:

frontend/src/lib/pages/SnsAccounts.svelte
peterpeterparker: Not sure I follow. What I mean is that you said "I would keep them in memory" but implemented it does the contrary.

Not a blocker for this PR, should we add a todo `// TODO: improve loading and use in memory sns neurons or load from backend` or something like that?

Again, if I am not missing something

frontend/src/lib/pages/SnsAccounts.svelte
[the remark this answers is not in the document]
peterpeterparker: Yes I prefer for maintanability reason and because here in that particular case the UI will be exactly the same if I get it right. Also - opiniated answer - I hate duplicated code 😁.

So far I can see there is two approach. Yes either creating a card like you said or even going further and creating a generic component above the for each.

Can be done in another PR for sure if you rather like.

frontend/src/lib/utils/utils.ts
```diff
@@ -262,3 +262,18 @@ export const poll = async <T>({
...
   `<span class="value">${text}</span>`;
+
+/**
+ * Removes entries from an object given a list of keys.
+ *
+ * @param obj object to remove entries from
+ * @param keysToRemove keys to remove
+ * @returns new object with entries removed
+ */
+export const removeKeys = <T extends Record<string, unknown>>(
```
peterpeterparker, line 273: I would not be against an object as param since there is more than one

I know we still have to check if we can develop a lint rule

peterpeterparker, line 273: I am fine with `removeKeys` but just in case you like it, I was asking myself if it could not also be named `filterKeys`

frontend/src/lib/pages/SnsAccounts.svelte
lmuntaner: We load them again, but I would always load them. The question for me is whether we show the ones in memory if they are there or not.

At the moment, we keep them in memory,  but we never show them because we set `loading = true` before requesting again. And only having loaded them, we show them.

But the neurons or accounts are not removed from the memory. We just don't use the memory. I will add a comment.

frontend/src/lib/utils/utils.ts
lmuntaner: I felt it was nicer to pass two params for this utils. It's more readable the first one:
1. `removeKeys(obj, ["a"]);`
2. `removeKeys({ obj, keys: ["a"] });`

lmuntaner: I would expect `filter` to receive a function, not just remove those keys. But filter based on something.

frontend/src/lib/pages/SnsAccounts.svelte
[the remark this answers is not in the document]
lmuntaner: Maybe in another PR.

frontend/src/lib/utils/utils.ts
peterpeterparker: Second one is easier to refactor, follow our pattern and improve typesafety.

lmuntaner: done

frontend/src/lib/pages/SnsAccounts.svelte
lmuntaner: todo added

lmuntaner:

peterpeterparker (approved):
LGTM, thanks

PR #1346, opened 2022-09-19
L2-1023: Transaction UI improvements

lmuntaner:
# Motivation

UI Improvements in the Transaction flow.

# Changes

* TransactionReview: Account name is not using the label color.
* TransactionReview: Show the text "Balance" next to the balance of the source account
* TransactionForm: Hide toggle to select destination from dropdown if no accounts to choose.

# Tests

* Test that SelectDestinationAddress doesn't show toggle if no extra accounts to choose from.

### DIFF
--- a/frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
+++ b/frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
@@ -37,10 +37,13 @@
   <div class="info">
     <KeyValuePair>
       <span class="label" slot="key">{$i18n.accounts.source}</span>
-      <AmountDisplay slot="value" singleLine amount={sourceAccount.balance} />
+      <div class="balance" slot="value">
+        <span class="label">{$i18n.accounts.balance}</span>
+        <AmountDisplay singleLine amount={sourceAccount.balance} />
+      </div>
     </KeyValuePair>
     <div>
-      <p class="label">
+      <p>
         {sourceAccount.name ?? $i18n.accounts.main}
       </p>
       <p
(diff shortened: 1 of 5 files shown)


lmuntaner:
UI changes:

No toggle when no extra accounts to choose from:
Screenshot 2022-09-19 at 14 56 09

UI changes when reviewing the transaction:
Screenshot 2022-09-19 at 15 02 39

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
```diff
@@ -94,6 +97,12 @@
 <style lang="scss">
   @use "../../../themes/mixins/modal";
 
+  .balance {
+    display: flex;
+    align-items: center;
+    flex-direction: row;
+    gap: var(--padding-1_5x);
```
peterpeterparker, line 104: really nth: not sure if it's in the design or not, but according your screenshort "Balance  36.90 ICP" there are more spaces between "Balance  36..." that ".90 ICP"

if not design, have you try a lower space? `AmountDisplay` has `grid-gap: 5px;` - don't know why 5px - so maybe 5px here as well or update both with `--padding-0_5x`?

peterpeterparker (approved):
Thanks!

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
lmuntaner: I did that on purpose.

I see Balance as a label of "36.90 ICP", while ICP is more closely related to "36.90".

I used `var(--padding-1_5x);` between "Balance" and "36.90 ICP".

PR #1349, opened 2022-09-20
style: inline amount display

peterpeterparker:
# Motivation

Inline the display of all amount.

is:

123'456.65
            ICP
            
will be:

123'456.65 ICP

# Note

- in accordance with Mischa
- need it for the new tasks I have to develop

# Changes

- inline display in `AmountDisplay`
- replace JS properties in `AmountDisplay` that have style and display goals with the usage of style `class`
- replace hardcoded `700` with CSS variable for bold text

### DIFF
--- a/frontend/src/lib/components/ic/GetICPs.svelte
+++ b/frontend/src/lib/components/ic/GetICPs.svelte
@@ -137,4 +137,8 @@
 
     padding: var(--padding-2x) var(--padding);
   }
+
+  .primary {
+    min-width: calc(48px + (2 * var(--padding-2x)));
+  }
 </style>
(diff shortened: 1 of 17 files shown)


frontend/src/lib/components/ic/GetICPs.svelte
```diff
@@ -137,4 +137,8 @@
 
     padding: var(--padding-2x) var(--padding);
   }
+
+  .primary {
```
peterpeterparker, line 141: not related to this PR but since I was there and since it is a just dev/test only component, included this small style change with it.

peterpeterparker:
Ultimately not needed - old design remains old design.

PR #1351, opened 2022-09-20
L2-1024: Fix dropdown hidden content

lmuntaner:
# Motivation

Fix the hidden slice of word in the Project Dropdown selector of Accounts and Neurons.

# Changes

* Refactor the Dropdown used in the pages to its own component: SelectProjectDropdownHeader.
* Apply the width of the `select` of the dropdown as content-box not as border-box.

# Tests

Not applicable.

### DIFF
--- a/frontend/src/lib/components/ic/SelectProjectDropdownHeader.svelte
+++ b/frontend/src/lib/components/ic/SelectProjectDropdownHeader.svelte
@@ -1,0 +1,20 @@
+<script lang="ts">
+  import SelectProjectDropdown from "../neurons/SelectProjectDropdown.svelte";
+</script>
+
+<div class="dropdown-wrapper">
+  <!-- Extra div is needed to fit to the content size instead of the whole width -->
+  <div>
+    <SelectProjectDropdown />
+  </div>
+</div>
+
+<style lang="scss">
+  .dropdown-wrapper {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+
+    margin-top: var(--padding-4x);
+  }
+</style>
--- a/frontend/src/lib/components/ic/SelectProjectDropdownHeader.svelte
+++ b/frontend/src/lib/components/ic/SelectProjectDropdownHeader.svelte
@@ -3,10 +3,7 @@
 </script>
 
 <div class="dropdown-wrapper">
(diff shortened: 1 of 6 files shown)


lmuntaner:
UI of the dropdown with the longest name:

Screenshot 2022-09-20 at 16 54 37

lmuntaner:

frontend/src/lib/components/ic/SelectProjectDropdownHeader.svelte
```diff
@@ -0,0 +1,20 @@
+<script lang="ts">
+  import SelectProjectDropdown from "../neurons/SelectProjectDropdown.svelte";
+</script>
+
+<div class="dropdown-wrapper">
```
peterpeterparker, line 5: Just gave a quick try, I think you might able to solve this without any extra `div` here if you adapt the `<Dropdown />` component that contains a `div` if it's possible.

But that was just a quick try.

Capture d’écran 2022-09-20 à 17 20 40

lmuntaner: Does it work also in the TransactionModal?

I need it to be 100% width, because that is the width for transaction modal.

I could pass that as a prop, but I think this adds complexity. While an extra div does not.

peterpeterparker: Let's then do just one `div`

Capture d’écran 2022-09-20 à 18 11 07

lmuntaner: done

lmuntaner:

frontend/src/lib/components/ui/Dropdown.svelte
```diff
@@ -7,9 +7,10 @@
   export let name: string;
   export let testId: string | undefined = undefined;
   export let disabled: boolean = false;
+  export let fullWidth: boolean = false;
```
peterpeterparker, line 10: can you solve this with a CSS variable instead?

lmuntaner: done

lmuntaner:

peterpeterparker (approved):
Thank you!

PR #1357, opened 2022-09-21
L2-997: Remove ICP from Transaction Modal

lmuntaner:
# Motivation

Remove the dependency of ICP of the TransactionModal.

# Changes

* TransactionModal: Transaction fee as a prop instead of hardcoded NNS ledger fee.
* TransactionModal: Token as a prop instead of hardcoded ICP when showing it in display.
* Rename util "formatICP" to "formatToken"
* Rename css var "--icp-font-size" to "--token-font-size"

# Tests

* Add a test that the transaction fee prop is used.
* Rename also in tests.

### DIFF
--- a/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
+++ b/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
@@ -2,10 +2,6 @@
   import { createEventDispatcher } from "svelte";
   import FooterModal from "../../../modals/FooterModal.svelte";
   import { i18n } from "../../../stores/i18n";
-  import {
-    mainTransactionFeeStoreAsIcp,
-    transactionsFeesStore,
-  } from "../../../stores/transaction-fees.store";
   import type { Account } from "../../../types/account";
   import { InvalidAmountError } from "../../../types/neurons.errors";
   import {
@@ -13,15 +9,13 @@
     invalidAddress,
     isAccountHardwareWallet,
   } from "../../../utils/accounts.utils";
-  import {
-    convertNumberToICP,
-    getMaxTransactionAmount,
-  } from "../../../utils/icp.utils";
+  import { getMaxTransactionAmount } from "../../../utils/icp.utils";
   import SelectAccountDropdown from "../../../components/accounts/SelectAccountDropdown.svelte";
   import AmountDisplay from "../../../components/ic/AmountDisplay.svelte";
   import AmountInput from "../../../components/ui/AmountInput.svelte";
   import KeyValuePair from "../../../components/ui/KeyValuePair.svelte";
   import SelectDestinationAddress from "../../../components/accounts/SelectDestinationAddress.svelte";
+  import { TokenAmount, type Token } from "@dfinity/nns";
 
   // Tested in the TransactionModal
   export let selectedAccount: Account | undefined = undefined;
@@ -29,6 +23,8 @@
   export let canSelectSource: boolean;
   export let selectedDestinationAddress: string | undefined = undefined;
   export let amount: number | undefined = undefined;
+  export let token: Token;
+  export let transactionFee: TokenAmount;
   // TODO: Handle min and max validations inline: https://dfinity.atlassian.net/browse/L2-798
   export let maxAmount: bigint | undefined = undefined;
   export let skipHardwareWallets: boolean = false;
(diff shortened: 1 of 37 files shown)


lmuntaner:

Many files, but most of them are renaming. I added a comment to the ones that need close reviewing.

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
```diff
@@ -2,33 +2,29 @@
...
   import SelectDestinationAddress from "../../../components/accounts/SelectDestinationAddress.svelte";
+  import { TokenAmount, type Token } from "@dfinity/nns";
 
   // Tested in the TransactionModal
   export let selectedAccount: Account | undefined = undefined;
   export let canSelectDestination: boolean;
   export let canSelectSource: boolean;
   export let selectedDestinationAddress: string | undefined = undefined;
   export let amount: number | undefined = undefined;
+  export let token: Token;
```
lmuntaner, line 26: This is another file that changed, not just renaming.

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
```diff
@@ -5,23 +5,23 @@
...
   import { i18n } from "../../../stores/i18n";
-  import { mainTransactionFeeStoreAsToken } from "../../../derived/main-transaction-fee.derived";
   import type { Account } from "../../../types/account";
   import AmountDisplay from "../../../components/ic/AmountDisplay.svelte";
   import KeyValuePair from "../../../components/ui/KeyValuePair.svelte";
   import type { NewTransaction } from "../../../types/transaction.context";
 
   export let transaction: NewTransaction;
   export let disableSubmit: boolean;
+  export let transactionFee: TokenAmount;
```
lmuntaner, line 15: This is another file to review

frontend/src/tests/lib/modals/accounts/TransactionModal.spec.ts
```diff
@@ -131,6 +135,29 @@ describe("TransactionModal", () => {
       ).toBeInTheDocument();
     });
 
+    it("should move to the last step and render passed transaction fee", async () => {
```
lmuntaner, line 138: This is a new test.

frontend/src/tests/routes/Accounts.spec.ts
```diff
@@ -15,6 +15,12 @@ import {
   mockSnsFullProject,
 } from "../mocks/sns-projects.mock";
 
+jest.mock("../../lib/services/sns-accounts.services", () => {
```
lmuntaner, line 18: This was causing some ugly logs when running the test for the route Accounts.

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
```diff
@@ -66,10 +62,10 @@
       return;
     }
     try {
-      const icp = convertNumberToICP(amount);
+      const tokens = TokenAmount.fromNumber({ amount, token }) as TokenAmount;
```
peterpeterparker, line 65: you need the cast? should we make `fromNumber` generic `.fromNumber<TokenAmount>`?

peterpeterparker (approved):
LGTM, thanks.

Just added a question as comment.

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
lmuntaner: I'd like to make it return only `TokenAmount`. Generic not sure if it makes sense, it always returns that.

But now it uses `fromString` which doesn't always return a TokenAmount and might return someting else.

Should we maybe raise an error in `fromString` as well?

peterpeterparker: I am kind of open to any good solution that can spare a `as` cast 😉

If the case is just temporary, then maybe a `// TODO to be removed` I would say.

lmuntaner: I will remove it from everywhere where I use `fromNumber`. There is no need for an extra comment here.

I also have the issue: https://dfinity.atlassian.net/browse/GIX-1005

But I didn't add a comment everywhere where this will change.

peterpeterparker: That do too, thanks

PR #1358, opened 2022-09-22
Use current ic-admin

bitdivine:
# Motivation
There has been a breaking change in the NNS manage neuron API, so we need to be careful to use a version of ic-admin that matches the canisters.

# Changes
- When downloading NNS canisters, download ic-admin as well.
- When making a proposal with ic-admin, use the downloaded tool.

# Tests
I have used this to deploy small14


peterpeterparker (approved):
Thx.

PR #1359, opened 2022-09-22
Remove outdated patch

bitdivine:
# Motivation
The sns code has changed significantly.  One of our patches fails to apply and may well no longer be relevant.

# Changes
- Remove broken patch that allowed an SNS start time to be in the past.

# Tests
I have verified that the patchfile now applies cleanly.


peterpeterparker (approved):
Thx

PR #1363, opened 2022-09-22
fix: hide open proposals where all may neurons etc filter

peterpeterparker:
# Motivation

With the introduction of the new UI/UX of the voting we noticed that the filter *Hide "Open" proposals where all your neurons have voted or are ineligible to vote* actually did not word as expected anymore.

Also update filter text to make it more comprehensive

# Changes

- fix filter with a more comprehensive (hopefully for our future selves) approach
- update filter text to "Show only proposals you can still vote for"

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -768,9 +768,9 @@
       }
     },
     "node_modules/@dfinity/gix-components": {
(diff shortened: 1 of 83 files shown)


mstrasinskis (approved):
LGTM, thank you!

PR #1364, opened 2022-09-22
Feat/update sns did files

lmuntaner:
# Motivation

Update Sns Project types to adatp to breaking changes in SNS Swap Canister.

# Changes

* Main one is the SnsSwap. `SnsSummary.swap` has changed completely. There is now an `SnsParams` with most of the needed data that before was in Init.

# Tests

* Fixed all broken tests.

### DIFF
--- a/frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
+++ b/frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
@@ -5,21 +5,14 @@
     SnsSwapCommitment,
     SnsSummarySwap,
   } from "../../types/sns";
-  import {
-    durationTillSwapDeadline,
-    openTimeWindow,
-  } from "../../utils/projects.utils";
+  import { durationTillSwapDeadline } from "../../utils/projects.utils";
   import { TokenAmount } from "@dfinity/nns";
   import { i18n } from "../../stores/i18n";
   import { secondsToDuration } from "../../utils/date.utils";
   import AmountDisplay from "../ic/AmountDisplay.svelte";
-  import DateSeconds from "../ui/DateSeconds.svelte";
-  import {
-    SnsSwapLifecycle,
-    type SnsSwapState,
-    type SnsSwapTimeWindow,
-  } from "@dfinity/sns";
+  import { SnsSwapLifecycle } from "@dfinity/sns";
   import ProjectUserCommitmentLabel from "../project-detail/ProjectUserCommitmentLabel.svelte";
+  import { getCommitmentE8s } from "../../utils/sns.utils";
 
   export let project: SnsFullProject;
 
@@ -30,31 +23,23 @@
   let swap: SnsSummarySwap;
   $: ({ swap } = summary);
 
-  let state: SnsSwapState;
-  $: ({ state } = swap);
-
   let lifecycle: number;
-  $: lifecycle = state.lifecycle;
+  $: ({
+    swap: { lifecycle },
+  } = summary);
 
   let durationTillDeadline: bigint | undefined;
   $: durationTillDeadline = durationTillSwapDeadline(swap);
 
-  let timeWindow: SnsSwapTimeWindow | undefined;
-  $: timeWindow = openTimeWindow(swap);
-
-  let start_timestamp_seconds: bigint | undefined;
-  $: ({ start_timestamp_seconds } = timeWindow ?? {
-    start_timestamp_seconds: undefined,
-    end_timestamp_seconds: undefined,
-  });
-
-  export let myCommitment: TokenAmount | undefined = undefined;
-  $: myCommitment =
-    swapCommitment?.myCommitment === undefined
-      ? undefined
-      : TokenAmount.fromE8s({
-          amount: swapCommitment.myCommitment.amount_icp_e8s,
-        });
+  let myCommitment: TokenAmount | undefined = undefined;
+  $: {
+    const commitmentE8s = getCommitmentE8s(swapCommitment);
+    if (commitmentE8s !== undefined) {
+      myCommitment = TokenAmount.fromE8s({
+        amount: commitmentE8s,
+      });
+    }
+  }
 </script>
 
 <dl>
@@ -68,12 +53,6 @@
   {#if lifecycle === SnsSwapLifecycle.Open && durationTillDeadline !== undefined}
     <dt>{$i18n.sns_project_detail.deadline}</dt>
     <dd class="value">{secondsToDuration(durationTillDeadline)}</dd>
-  {/if}
-
-  <!-- Sale starts soon -->
-  {#if lifecycle === SnsSwapLifecycle.Pending && start_timestamp_seconds !== undefined}
-    <dt>{$i18n.sns_project_detail.sale_start}</dt>
-    <DateSeconds tagName="dd" seconds={Number(start_timestamp_seconds)} />
   {/if}
 
   {#if myCommitment !== undefined}
--- a/frontend/src/lib/types/sns.ts
+++ b/frontend/src/lib/types/sns.ts
@@ -1,8 +1,11 @@
 import type { Principal } from "@dfinity/principal";
 import type {
+  CfParticipant,
+  SnsNeuronRecipe,
+  SnsParams,
   SnsSwapBuyerState,
   SnsSwapDerivedState,
-  SnsSwapState,
+  SnsSwapInit,
 } from "@dfinity/sns";
 
 /**
@@ -24,10 +27,17 @@ export interface SnsTokenMetadata {
   symbol: string;
 }
 
-// export interface SnsSummarySwap {
-//   init: SnsSwapInit;
-//   state: ;
-// }
+export interface SnsSummarySwap {
+  neuron_recipes: Array<SnsNeuronRecipe>;
+  cf_participants: Array<CfParticipant>;
+  // We don't use it for now and keep it as the candid optional type
+  init: [] | [SnsSwapInit];
+  lifecycle: number;
+  buyers: Array<[string, SnsSwapBuyerState]>;
+  params: SnsParams;
+  // We don't use it for now and keep it as the candid optional type
+  open_sns_token_swap_proposal_id: [] | [bigint];
+}
 
 export interface SnsSummary {
   rootCanisterId: Principal;
@@ -47,7 +57,7 @@ export interface SnsSummary {
   /**
    * The initial information of the sale (min-max ICP etc.) and its current state (pending, open, committed etc.)
    */
-  swap: SnsSwapState;
+  swap: SnsSummarySwap;
   /**
    * Derived information about the sale such as the current total of ICP all buyers have invested so far
    */
--- a/frontend/src/lib/types/sns.ts
+++ b/frontend/src/lib/types/sns.ts
@@ -2,7 +2,6 @@ import type { Principal } from "@dfinity/principal";
 import type {
   SnsSwapBuyerState,
   SnsSwapDerivedState,
-  SnsSwapInit,
   SnsSwapState,
 } from "@dfinity/sns";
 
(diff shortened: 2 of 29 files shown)


lmuntaner:

Not mergeable until we update the library, but the changes should be enough. I'll review again tomorrow afer upgrading the library.

lmuntaner:

frontend/src/lib/types/sns.ts
```diff
@@ -26,8 +28,15 @@ export interface SnsTokenMetadata {
 }
 
 export interface SnsSummarySwap {
-  init: SnsSwapInit;
-  state: SnsSwapState;
+  neuron_recipes: Array<SnsNeuronRecipe>;
+  cf_participants: Array<CfParticipant>;
+  // We don't use it for now and keep it as the candid optional type
+  init: [] | [SnsSwapInit];
```
peterpeterparker, line 34: there should be no optional type in `SnsSummarySwap`

if a Sns does not provided any `Init` or `Swap` information it has to be filtered

frontend/src/lib/types/sns.ts
```diff
@@ -26,8 +28,15 @@ export interface SnsTokenMetadata {
...
+  lifecycle: number;
+  buyers: Array<[string, SnsSwapBuyerState]>;
```
peterpeterparker, line 36: `buyers` should find place in `SnsSummary` I think

`SnsSummarySwap` contains  nitial information of the sale (min-max ICP etc.) and its current state (pending, open, committed etc.)

frontend/src/lib/types/sns.ts
```diff
@@ -26,8 +28,15 @@ export interface SnsTokenMetadata {
 }
 
 export interface SnsSummarySwap {
-  init: SnsSwapInit;
-  state: SnsSwapState;
+  neuron_recipes: Array<SnsNeuronRecipe>;
```
peterpeterparker, line 31: `neuron_recipes`, `cf_participants`, `buyers` and `open_sns_token_swap_proposal_id` should find place in `SnsSummary` I think

`SnsSummarySwap` contains initial information of the sale (min-max ICP etc.) and its current state (pending, open, committed etc.)

peterpeterparker (commented):
Added two feedbacks about the main Sns type. I stop the review there - i.e. did not proceeded yet the review further as these questions are central.

frontend/src/lib/types/sns.ts
peterpeterparker: note: we should not use here Candid like optional types in the converted types (`[] | [Something]`) but Javascript like `something?: number` or `something: number | undefined`) (mapped with `fromNullable`). this spares the conversions when we are mapping or using the values in the app.

lmuntaner: What we decided was to use the same types as the response from the API. This is the new response of the get_state call. I'd rather not move anything around.

Before that data was in `summary.swap.init.max...` and now it's in `summary.swap.params.max...`.

I added the comments on why I am not changing those properties. They are not being used and this would simplify this PR. If we decided to use then, we should then take it out of the array. And that would cause 0 refactor because they are not used at the moment.

frontend/src/lib/types/sns.ts
lmuntaner: The project is filtered out if there are no params. But because the other properties are not used yet, I didn't add any filtering nor any change. I added a comment:

```
  // We don't use it for now and keep it as the candid optional type
  init: [] | [SnsSwapInit];
```

lmuntaner:
@peterpeterparker replied to the comments

frontend/src/lib/types/sns.ts
peterpeterparker: `init` should not be optional here, that's the core of the existing architecture. likewise for `params`.

frontend/src/lib/types/sns.ts
peterpeterparker: Ok works for me regarding optional or not for these particular elements (not those core information discussed in other thread).

That being said, still not convince they should find place in `SnsSummarySwap`. What's the reasoning to add these in this interface and not a level up?

e.g. one level up we have the `derived` information about the on going sale.

lmuntaner: Because like this, we keep the same structure as the candid interface (except the optional params).

This was the initial design we agreed upon. We add no transformations, except for optional params.

frontend/src/lib/types/sns.ts
lmuntaner: > Ok works for me regarding optional or not for these particular elements (not those core information discussed in other thread).

`init` is not core information anymore.

peterpeterparker: gotcha, missed that (opened ic-js now).

can we remove it then or you are against? I would rather like not to keep deadcode so that when we use it, we think at the same time the best way we can integrate it. likewise for `open_sns_token_swap_proposal_id` but if there is a good reason to have it without using these?

frontend/src/lib/types/sns.ts
peterpeterparker: Ok makes sense. Thanks for the explanation

frontend/src/lib/types/sns.ts
lmuntaner: it's not dead code, rather unused properties. We don't remove the properties of the neurons we don't use. I don't think we should do the same here.

peterpeterparker: in nns-js, sns-js absolutely yes but in nns-dapp we don't much unused declared types and fields?

lmuntaner: We do, for example in neuros or proposals. Or also before, we weren't using all the properties inside `Init` or inside `State`. We left the type as it was.

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -196,3 +192,8 @@ export const getSwapCanisterAccount = ({
 
   return accountIdentifier;
 };
+
+export const getCommitmentE8s = (
+  swapCommitment?: SnsSwapCommitment | null
+): bigint | undefined =>
+  swapCommitment?.myCommitment?.icp[0]?.amount_e8s ?? undefined;
```
peterpeterparker, line 199: nth: `fromNullable(swapCommitment?.myCommitment?.icp)?.amount_e8s ?? undefined`

frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
```diff
@@ -70,12 +55,6 @@
     <dd class="value">{secondsToDuration(durationTillDeadline)}</dd>
   {/if}
 
-  <!-- Sale starts soon -->
-  {#if lifecycle === SnsSwapLifecycle.Pending && start_timestamp_seconds !== undefined}
-    <dt>{$i18n.sns_project_detail.sale_start}</dt>
```
peterpeterparker, line 75: the label can be removed in `en.json` as well

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -144,13 +139,21 @@ export const mapAndSortSnsQueryToSummaries = ({
...
+
+      const swapData = fromNullable(swapState?.swap ?? []);
       return {
         ...rest,
         rootCanisterId,
         metadata: mapOptionalMetadata(metadata),
         token: mapOptionalToken(token),
         swapCanisterId: swapState?.swapCanisterId,
-        swap: fromNullable(swapState?.swap ?? []),
+        swap:
```
peterpeterparker, line 150: nth: extract mapping in a `mapOptionalSwap` function that has a comment

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -144,13 +139,21 @@ export const mapAndSortSnsQueryToSummaries = ({
...
+          swapData === undefined
+            ? undefined
+            : ({
+                ...swapData,
+                params: fromNullable(swapData.params),
```
peterpeterparker, line 155: if I get it right, we assign `fromNullable(swapData.params)` to `SnsSummarySwap` that has `params` mandatory - in other words we are not 100% params is set

so if `fromNullable(swapData.params) === undefined` then the all `SnsSummarySwap` should be undefined

or

this constructed object should be casted to `Partial<SnsSummarySwap>` or something and then e.g. on line 166 those that have no params should be filtered but ultimately `validSwapSummaries` should not contains params that are undefined

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -196,3 +192,8 @@ export const getSwapCanisterAccount = ({
 
   return accountIdentifier;
 };
+
+export const getCommitmentE8s = (
```
peterpeterparker, line 196: would it be worse a particular test?

frontend/src/lib/utils/sns.utils.ts
peterpeterparker: also worse a test then

frontend/src/lib/types/sns.ts
peterpeterparker: Don't like it, I think it should not be the case as explained above but, can live with it. Thanks for the feedback.

mstrasinskis (approved):
LGTM. Thanks

frontend/src/lib/utils/sns.utils.ts
lmuntaner: This is mapping to `OptionalSummary`.

Then `validSwapSummaries` removes the one where `params` is not defined.

This is what is done for the rest of the properties as well:

```
  // Only those that have valid metadata, toke, sale and derived information are - and can be - considered as valid
  const validSwapSummaries: ValidSummary[] = allSummaries.filter(
    (entry: OptionalSummary): entry is ValidSummary =>
      entry.swap !== undefined &&
      entry.swapCanisterId !== undefined &&
      entry.derived !== undefined &&
      entry.metadata !== undefined &&
      entry.token !== undefined
  );
```

I'll add a test for it, that makes sense.

frontend/src/lib/utils/sns.utils.ts
lmuntaner: sure

frontend/src/lib/utils/sns.utils.ts
peterpeterparker: Don't think it's correct.

PR code:

```
const swapData = fromNullable(swapState?.swap ?? []);
return {
        ...rest,
        rootCanisterId,
        metadata: mapOptionalMetadata(metadata),
        token: mapOptionalToken(token),
        swapCanisterId: swapState?.swapCanisterId,
        swap:
          swapData === undefined
            ? undefined
            : ({
                ...swapData,
                params: fromNullable(swapData.params),
              } as SnsSummarySwap),
        derived: fromNullable(swapState?.derived ?? []),
      };
```

which output (ignoring unrelated param in example):

```
{
 swap: undefined
}

//

or

{
 swap: {
     params: {}
 } as SnsSummarySwap
}

// or 

{
 swap: {
     params: undefined
 } as SnsSummarySwap
}
```

This particular last one is problematic I think because given the definition and our expectation, `params` cannot be null or undefined but this is overriden by the cast.

If you suppress the cast and focus on the particular object you should get ` Type 'undefined' is not assignable to type 'Params'.` because `SnsSummarySwap` contains following defintion of `params`:

```
export interface SnsSummarySwap {
    ...
    params: SnsParams;
}    
```

so later on in the test you display in your previous comment

```
onst validSwapSummaries: ValidSummary[] = allSummaries.filter(
    (entry: OptionalSummary): entry is ValidSummary =>
      entry.swap !== undefined && // <------------ we test swap yes but we do not test if params is undefined because it should be given which might not be the case
      entry.swapCanisterId !== undefined &&
      entry.derived !== undefined &&
      entry.metadata !== undefined &&
      entry.token !== undefined
  );
```

lmuntaner: fixed

frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
lmuntaner: done

frontend/src/lib/utils/sns.utils.ts
lmuntaner: done

lmuntaner:

peterpeterparker (approved):
LGTM, thanks.

PR #1365, opened 2022-09-23
GIX-1025: Sns Account Text Representation

lmuntaner:
# Motivation

User sees the string representation of the SNS Accounts.

# Changes

* New sns-accounts utils to encode and decode the accounts.
* Use the sns utils in sns-ledger api function to encode the default account.

# Tests

* Test new sns-accounts utils.

### DIFF
--- a/frontend/src/lib/utils/sns-accounts.utils.ts
+++ b/frontend/src/lib/utils/sns-accounts.utils.ts
@@ -1,0 +1,80 @@
+import { Principal } from "@dfinity/principal";
+
+interface SnsAccount {
+  owner: Principal;
+  subaccount?: Uint8Array;
+}
+
+const shrink = (bytes: Uint8Array): Uint8Array => {
+  const shrinked = Array.from(bytes);
+  while (shrinked[0] === 0) {
+    shrinked.shift();
+  }
+  return Uint8Array.from(shrinked);
+};
+
+// Reference: https://github.com/dfinity/ICRC-1/pull/55/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R238
+export const encodeSnsAccount = ({ owner, subaccount }: SnsAccount): string => {
+  if (subaccount === undefined) {
+    return owner.toText();
+  }
+
+  const subaccountBytes = shrink(subaccount);
+
+  if (subaccountBytes.length === 0) {
+    return owner.toText();
+  }
+
+  const bytes = Uint8Array.from([
+    ...owner.toUint8Array(),
+    ...subaccountBytes,
+    subaccountBytes.length,
+    parseInt("FF", 16),
+  ]);
+
+  return Principal.fromUint8Array(bytes).toText();
+};
+
+// Reference: https://github.com/dfinity/ICRC-1/pull/55/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R268
+export const decodeSnsAccount = (accountString: string): SnsAccount => {
+  const principal = Principal.fromText(accountString);
+
+  const [ff, nonZeroLength, ...restReversed] = principal
+    .toUint8Array()
+    .reverse();
+
+  if (ff !== parseInt("FF", 16)) {
+    return {
+      owner: Principal.fromText(accountString),
+    };
+  }
+
+  if (
+    nonZeroLength > 32 ||
+    nonZeroLength === 0 ||
+    nonZeroLength === undefined
+  ) {
+    throw new Error("Invalid account string");
+  }
+
+  const subaccountBytesReversed = restReversed.slice(0, nonZeroLength);
+  if (
+    subaccountBytesReversed[0] === 0 ||
+    subaccountBytesReversed.length !== nonZeroLength
+  ) {
+    throw new Error("Invalid account string");
+  }
+  while (subaccountBytesReversed.length < 32) {
+    subaccountBytesReversed.push(0);
+  }
+  const subaccount = Uint8Array.from(subaccountBytesReversed.reverse());
+
+  const principalBytes = restReversed
+    .reverse()
+    .filter((_, i) => i < restReversed.length - nonZeroLength);
+
+  return {
+    owner: Principal.fromUint8Array(Uint8Array.from(principalBytes)),
+    subaccount,
+  };
+};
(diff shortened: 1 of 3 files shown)


lmuntaner:
The string representation of the default account is the same as the user principal. This is intentional.

Screenshot 2022-09-23 at 12 24 44

frontend/src/lib/utils/sns-accounts.utils.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const shrink = (bytes: Uint8Array): Uint8Array => {
+  const shrinked = Array.from(bytes);
+  while (shrinked[0] === 0) {
+    shrinked.shift();
+  }
+  return Uint8Array.from(shrinked);
+};
+
+// Reference: https://github.com/dfinity/ICRC-1/pull/55/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R238
```
peterpeterparker, line 16: can you add comment+example of what encode and decode do?

frontend/src/lib/utils/sns-accounts.utils.ts
```diff
@@ -0,0 +1,80 @@
+import { Principal } from "@dfinity/principal";
+
+interface SnsAccount {
+  owner: Principal;
+  subaccount?: Uint8Array;
+}
+
+const shrink = (bytes: Uint8Array): Uint8Array => {
```
peterpeterparker, line 8: can you add a comment?

frontend/src/lib/utils/sns-accounts.utils.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  if (subaccountBytes.length === 0) {
+    return owner.toText();
+  }
+
+  const bytes = Uint8Array.from([
+    ...owner.toUint8Array(),
+    ...subaccountBytes,
+    subaccountBytes.length,
+    parseInt("FF", 16),
```
peterpeterparker, line 32: do we already use `"FF"`? worth a constant?

frontend/src/lib/utils/sns-accounts.utils.ts
```diff
@@ -0,0 +1,80 @@
...
+      owner: Principal.fromText(accountString),
+    };
+  }
+
+  if (
+    nonZeroLength > 32 ||
+    nonZeroLength === 0 ||
+    nonZeroLength === undefined
+  ) {
+    throw new Error("Invalid account string");
```
peterpeterparker, line 57: i18n?

frontend/src/lib/utils/sns-accounts.utils.ts
lmuntaner: sure

frontend/src/lib/utils/sns-accounts.utils.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  const subaccountBytesReversed = restReversed.slice(0, nonZeroLength);
+  if (
+    subaccountBytesReversed[0] === 0 ||
+    subaccountBytesReversed.length !== nonZeroLength
+  ) {
+    throw new Error("Invalid account string");
+  }
+  while (subaccountBytesReversed.length < 32) {
```
peterpeterparker, line 67: nth: you have nicely implemented these new functions with an immutability approach except this `while`.  maybe there is a why to solve this without pushing.

frontend/src/lib/utils/sns-accounts.utils.ts
```diff
@@ -0,0 +1,80 @@
...
+  ]);
+
+  return Principal.fromUint8Array(bytes).toText();
+};
+
+// Reference: https://github.com/dfinity/ICRC-1/pull/55/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R268
+export const decodeSnsAccount = (accountString: string): SnsAccount => {
+  const principal = Principal.fromText(accountString);
+
+  const [ff, nonZeroLength, ...restReversed] = principal
```
peterpeterparker, line 42: not sure what's the impact on performance but we reserve to reverse again later - here the principal is reversed and `restReversed` gets reversed on line 70. should we improve the perf?

frontend/src/lib/utils/sns-accounts.utils.ts
lmuntaner: `fromSnsAccount` doesn't tell me much of what the function does because this is not a class method.

I followed the same naming as they use in their documentation. I believe that will make it easier to understand and relate in the future.

I'll add a comment.

frontend/src/lib/utils/sns-accounts.utils.ts
lmuntaner: sure

peterpeterparker (commented):
Would it be interesting to have these converters in `sns-js`? Maybe these can be useful for the community too?

frontend/src/lib/utils/sns-accounts.utils.ts
lmuntaner: Not sure we want to show this error message directly.

The error message will probably be show where we use the `decode` function.

But I guess I can add it just in case.

lmuntaner:
That's a good point, I guess yes, let me move it there.

frontend/src/lib/utils/sns-accounts.utils.ts
peterpeterparker: Thanks

frontend/src/lib/utils/sns-accounts.utils.ts
peterpeterparker: Comment not relevant anymore since we move the feature to `sns-js`

peterpeterparker:
As both are questions or nth, not sure if you noticed this [comment](https://github.com/dfinity/nns-dapp/pull/1365#discussion_r979581158) and this [comment](https://github.com/dfinity/nns-dapp/pull/1365#discussion_r979582317) @lmuntaner when moving the code to `sns-js`? If yes, let's just click resolve on these...

frontend/src/lib/utils/sns-accounts.utils.ts
lmuntaner: Not sure it's worth spending time improving the performance when the maximum number of chars is about 100.

frontend/src/lib/utils/sns-accounts.utils.ts
lmuntaner: I don't think it's worth it.

lmuntaner:
@peterpeterparker sorry, forgot to reply to you. I replied them in this PR.

frontend/src/lib/utils/sns-accounts.utils.ts
peterpeterparker: if function is used only once sure, but the functions will be use often, like we will display more than one neurons

but yes can be improved later on.

lmuntaner:
Closing in favor of https://github.com/dfinity/ic-js/pull/201

PR #1367, opened 2022-09-26
Feat/Update the deploy process to account for the tip of the SNS development

DanielThurau:
# Motivation

The snsdemo branch depends on checking out a commit on the nns-dapp. These upgrades should allow for the snsdemo to work with the tip of the SNS canisters.

# Changes

- deploy.sh will deploy the ic-icrc1-archive canister to the sns-wasm canister and not require a separate call to the sns-wasm canister.

# Tests

I've tested this locally but I do not believe there are tests in place for this. If I can add tests for this, please feel free to point out and I will add them.


lmuntaner (approved):
LGTM

peterpeterparker:
Duplicated and replaced by #1374

bitdivine:
@DanielThurau Thank you.  The archive canister has already been added in `dfx nns install` but not here.  These paths will converge, eventually.

PR #1370, opened 2022-09-26
feat: neuron staked

peterpeterparker:
# Motivation

wip

# Changes

wip

# Screenshots

wip


frontend/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
```diff
@@ -53,38 +54,37 @@
...
+        >
+        <svelte:fragment slot="info">
+          {#if neuron.fullNeuron?.cachedNeuronStake !== undefined}
+            {@html sanitize(
+              replacePlaceholders($i18n.neuron_detail.voting_power_tooltip, {
                 $stake: formatToken({
                   value: neuron.fullNeuron.cachedNeuronStake,
                   detailed: true,
                 }),
+                $st4kedMaturity: formattedStakedMaturity(neuron),
```
lmuntaner, line 74: Why the `4` in `st4kedMaturity`? Is that a type or on purpose?

frontend/src/lib/components/neuron-detail/actions/StakeMaturityButton.svelte
```diff
@@ -0,0 +1,19 @@
...
+  const showModal = () => (isOpen = true);
+  const closeModal = () => (isOpen = false);
+</script>
+
+<button class="primary" on:click={showModal} data-tid="stake-maturity-button"
+  >{$i18n.neuron_detail.stake_maturity}</button
+>
+
+{#if isOpen}
+  <MergeMaturityModal on:nnsClose={closeModal} {neuron} />
```
lmuntaner, line 18: Why is this opening the `MergeMaturityModal`?

frontend/src/lib/pages/NnsNeuronDetail.svelte
```diff
@@ -98,6 +99,10 @@
       neuronId: neuron.neuronId,
       store: $voteRegistrationStore,
     });
+
+  $: layoutTitleStore.set(
+    neuronId !== undefined ? `${$i18n.core.icp} – ${neuronId}` : ""
```
lmuntaner, line 104: I guess that was Mischa, but I find it very weird to use "ICP" here...

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -153,17 +153,16 @@ export const formatVotingPower = (value: bigint): string =>
 export const hasJoinedCommunityFund = (neuron: NeuronInfo): boolean =>
   neuron.joinedCommunityFundTimestampSeconds !== undefined;
 
-export const formattedMaturity = (neuron: NeuronInfo): string => {
-  let value = neuron?.fullNeuron?.maturityE8sEquivalent;
+export const formattedMaturity = ({ fullNeuron }: NeuronInfo): string =>
```
lmuntaner, line 156: Aren't we commenting all the utils from now on?

lmuntaner (approved):
A few comments but already approving.

Yet, could you also share the screenshot when the user sees stake maturity? The current screenshots show only 0.

frontend/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
peterpeterparker: Leet speak. Our `replaceHolder` is not clever enough, if you use a replacer that uses part of another placeholder, then it fck up the results. In this case use placeholders `$stakedMaturity` with other `$stake` does not work - i.e. the second placeholder should not contains `stake`.

So had no better idea 🤷‍♂️ except using something really not related but then does not mean much. Any idea of placeholder?

frontend/src/lib/components/neuron-detail/actions/StakeMaturityButton.svelte
peterpeterparker: Because the modal implementation finds not place in this PR but in another PR that will be merged within this PR.

frontend/src/lib/pages/NnsNeuronDetail.svelte
peterpeterparker: Yep, Mischa + business requirements I guess. It is not incrorrect because here it's "just" the NNS neuron.

frontend/src/lib/utils/neuron.utils.ts
peterpeterparker: Sure, done

frontend/src/lib/components/neuron-detail/NeuronMetaInfoCard.svelte
lmuntaner: aaaahhh, I agree with the solution, but this is worth an issue and a comment.

peterpeterparker: sure

PR #1372, opened 2022-09-26
GIX-995: Sns Wallet Details Setup

lmuntaner:
# Motivation

User can visit the Sns Wallet Details page. Page itself still needs to be implemented in another PR.

# Changes

* Change current `AppPath.Wallet` for `AppPath.LegacyWallet`.
* Change `AppPath.Wallet` to be a context based path: `/u/:id/wallet/:id`.
* Add exception for `LegacyWallet` in the changeContext util.
* Create two new page components: NnsWallet and SnsWallet.
* Wallet renders NnsWallet or SnsWallet.
* New route `LegacyWallet` that redirects to context base path (when enabled) or renders NnsWallet.

# Tests

* Move tests accordingly.
* Add test for new route pages.
* Add test cases for new path.


lmuntaner:
New UX:

Notice the address bar:

https://user-images.githubusercontent.com/4550653/192239282-cf12d5d9-a461-484e-b1ae-da18e8104fe0.mov

frontend/src/lib/pages/NnsWallet.svelte
```diff
@@ -0,0 +1,160 @@
+<script lang="ts">
```
lmuntaner, line 1: This is the old `routes/Wallet.svelte`.

frontend/src/tests/lib/pages/NnsWallet.spec.ts
```diff
@@ -0,0 +1,178 @@
+/**
```
lmuntaner, line 1: This is the same test we had for `/routes/Wallet`

lmuntaner:

frontend/src/lib/pages/NnsWallet.svelte
peterpeterparker, line 1: nothing new here right?

frontend/src/lib/pages/SnsWallet.svelte
```diff
@@ -0,0 +1,2 @@
+<!-- TODO: GIX-995 Implement SNS Wallet details -->
```
peterpeterparker, line 1: you gonna implement it right after merge or do it needs a redirect in case it gets proposed on mainnet?

frontend/src/lib/services/accounts.services.ts
```diff
@@ -136,7 +136,10 @@ const transferError = ({
 export const routePathAccountIdentifier = (
   path: string | undefined
 ): { accountIdentifier: string | undefined } | undefined => {
-  if (!isRoutePath({ path: AppPath.Wallet, routePath: path })) {
+  if (
+    !isRoutePath({ path: AppPath.LegacyWallet, routePath: path }) &&
```
peterpeterparker, line 140: nth: same comment as for the neurons routing pr, it would be neat to extend this with an array of paths to compare

```
isRoutePath({ paths: [AppPath.LegacyWallet, AppPath.Wallet], ...
```

frontend/src/lib/utils/app-path.utils.ts
```diff
@@ -1,17 +1,22 @@
 import { AppPath, CONTEXT_PATH } from "../constants/routes.constants";
+import { routePathAccountIdentifier } from "../services/accounts.services";
 import { routePathNeuronId } from "../services/neurons.services";
 import { memoize } from "./optimization.utils";
 
+const PRINCIPAL_ID_REGEX = "[a-zA-Z0-9-]+";
```
peterpeterparker, line 6: nth: not sure but haven't principals a fixed length and form? if yes probably possible to improve the regex since we introduce a constant that has the particular name

something like `([a-z0-9]{5}-){10}[a-zA-Z0-9]{3}`

but again not sure on the regex

frontend/src/lib/pages/SnsWallet.svelte
lmuntaner: It's only available if you enter the URL directly. But let me add a redirect anyway.

frontend/src/lib/pages/NnsWallet.svelte
lmuntaner: Yes, nothing new, I thuoght that I added comments where the code didn't change, sorry

frontend/src/lib/utils/app-path.utils.ts
lmuntaner: I'm also not sure of the regex. But actually, we might not want to check only principal ids, I forgot that we'll have the ckbtc project as well and that might not be a principal id. Or if in the future we want to support `/u/open-chat/...`.

I'll rename the variables.

frontend/src/lib/services/accounts.services.ts
lmuntaner: yes, agree, but this will change a lot of files. It probably deserves it's own PR.

lmuntaner:

frontend/src/lib/services/accounts.services.ts
peterpeterparker: another PR sounds good, can you do it?

frontend/src/lib/pages/NnsWallet.svelte
peterpeterparker: Thanks

peterpeterparker (approved):
LGTM, thx

frontend/src/lib/services/accounts.services.ts
lmuntaner: Yes, we're blocked because we still don't have the latests SNS, therefore I can do it.

PR #1379, opened 2022-09-27
GIX-995: Transactions out of SelectedAccountStore

lmuntaner:
# Motivation

Remove "transactions" from SelectedAccountStore to make it more reusable for SNS Wallet.

# Changes

* Remove "transactions" from SelectedAccountStore.
* Add "transactions" prop in TransactionsList.
* Use a local variable in NnsWallet to track transactions.

# Tests

* Fix broken tests after refactor.

### DIFF
--- a/frontend/src/lib/components/accounts/TransactionList.svelte
+++ b/frontend/src/lib/components/accounts/TransactionList.svelte
@@ -27,9 +27,9 @@
   $: extendedTransactions = mapToSelfTransaction(transactions ?? []);
 </script>
 
-{#if account === undefined || extendedTransactions === undefined}
+{#if account === undefined || transactions === undefined}
   <SkeletonCard cardType="info" />
-{:else if extendedTransactions.length === 0}
+{:else if transactions.length === 0}
   {$i18n.wallet.no_transactions}
 {:else}
   {#each extendedTransactions as { toSelfTransaction, transaction } (`${transaction.timestamp.timestamp_nanos}${toSelfTransaction}`)}
--- a/frontend/src/lib/components/accounts/TransactionList.svelte
+++ b/frontend/src/lib/components/accounts/TransactionList.svelte
@@ -11,27 +11,28 @@
   } from "../../types/selected-account.context";
   import { mapToSelfTransaction } from "../../utils/transactions.utils";
 
+  export let transactions: Transaction[] | undefined;
+
   const { store } = getContext<SelectedAccountContext>(
     SELECTED_ACCOUNT_CONTEXT_KEY
   );
 
   let account: Account | undefined;
-  let storeTransactions: Transaction[] | undefined;
-  $: ({ account, transactions: storeTransactions } = $store);
+  $: ({ account } = $store);
 
-  let transactions: {
+  let extendedTransactions: {
     transaction: Transaction;
     toSelfTransaction: boolean;
   }[];
-  $: transactions = mapToSelfTransaction(storeTransactions ?? []);
+  $: extendedTransactions = mapToSelfTransaction(transactions ?? []);
 </script>
 
-{#if account === undefined || storeTransactions === undefined}
+{#if account === undefined || transactions === undefined}
   <SkeletonCard cardType="info" />
-{:else if storeTransactions.length === 0}
+{:else if transactions.length === 0}
   {$i18n.wallet.no_transactions}
 {:else}
-  {#each transactions as { toSelfTransaction, transaction } (`${transaction.timestamp.timestamp_nanos}${toSelfTransaction}`)}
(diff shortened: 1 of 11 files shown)


lmuntaner:

frontend/src/lib/components/accounts/TransactionList.svelte
```diff
@@ -11,27 +11,28 @@
...
+  let extendedTransactions: {
     transaction: Transaction;
     toSelfTransaction: boolean;
   }[];
-  $: transactions = mapToSelfTransaction(storeTransactions ?? []);
+  $: extendedTransactions = mapToSelfTransaction(transactions ?? []);
 </script>
 
-{#if account === undefined || storeTransactions === undefined}
+{#if account === undefined || transactions === undefined}
```
peterpeterparker, line 30: shouldn't we compare `extendedTransactions`? (I know it was like that before)

frontend/src/lib/pages/NnsWallet.svelte
```diff
@@ -37,22 +40,25 @@
...
       accountIdentifier,
-      onLoad: ({ accountIdentifier, transactions }) => {
+      onLoad: ({ accountIdentifier, transactions: loadedTransactions }) => {
         // avoid using outdated transactions
         if (accountIdentifier !== $selectedAccountStore.account?.identifier) {
           return;
         }
 
-        $selectedAccountStore.transactions = transactions;
+        transactions = loadedTransactions;
```
peterpeterparker, line 56: should we deconstruct to be sure to reevalute?

before we were assiging a new object to the store so not 100% sure

frontend/src/lib/services/debug.services.ts
```diff
@@ -185,14 +183,6 @@ const anonymiseStoreState = async () => {
     },
     selectedAccount: {
       account: await anonymizeAccount(selectedAccount?.account),
-      transactions: await mapPromises(
```
peterpeterparker, line 188: we loose the transaction in the debug information?

frontend/src/lib/pages/NnsWallet.svelte
lmuntaner: `loadedTransactions` is the array, I am reassigning it to `transactions` which is the local variable.

frontend/src/lib/services/debug.services.ts
lmuntaner: yes...

frontend/src/lib/components/accounts/TransactionList.svelte
lmuntaner: yes, good catch. If one is undefined the other is also, but it makes sense to check the one

lmuntaner:

frontend/src/lib/components/accounts/TransactionList.svelte
lmuntaner: Actually, no. It was ok before. The "old" storeTransactions, was the one in the context, which now is passed as props.

lmuntaner: Otherwise that is never undefined because extendedTransactions is an empty array if no transactions.

peterpeterparker: Oki doki, thx for checking

frontend/src/lib/pages/NnsWallet.svelte
peterpeterparker: I mean should we do `transactions = [...loadedTransactions];`?

frontend/src/lib/services/debug.services.ts
peterpeterparker: anyway to prevent this? the information is pretty interesting. we notably needed it once to debug an issue.

frontend/src/lib/pages/NnsWallet.svelte
lmuntaner: Why? `loadedTransactions` will be a new array on every call. It comes from `getAccountTransactions`, that requests transactions.

frontend/src/lib/services/debug.services.ts
lmuntaner: I can create an issue to try to add it back. I guess we could add a way to store local variables as well in the debug store. Not necessarily only svelte stores.

lmuntaner: Issue: https://dfinity.atlassian.net/browse/GIX-1043

Added a comment in NnsWallet component.

lmuntaner:

frontend/src/lib/pages/NnsWallet.svelte
peterpeterparker: Then it's ok, thanks

peterpeterparker (approved):

PR #1380, opened 2022-09-27
feat: review spawn neuron

peterpeterparker:
# Motivation

Review "Spawn neuron".

# PRs

This PR -> #1370 -> main

# Changes

- use new modal for wizard
- remove some divs to adapt content
- review wording

# Tests

No particular tests added or modified. This PR review wording and refactor/move components name.
The new wizard modal has no particular test as we did for the legacy wizard modal because it is inderectly tested.


frontend/src/lib/components/neuron-detail/NeuronConfirmActionScreen.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
```
lmuntaner, line 1: Is this just a rename from `ConfirmationActionScreen`?

frontend/src/lib/modals/LegacyWizardModal.svelte
```diff
@@ -45,5 +45,9 @@
     @include modal.section;
 
     color: var(--background-contrast);
+
+    > :global(div) {
```
lmuntaner, line 49: Isn't it dangerous to set all `div`s inside it to `height: 100%`?

But I guess we will be removing this legacy code shortly so I wouldn't do anything in this PR.

frontend/src/lib/modals/WizardModal.svelte
```diff
@@ -0,0 +1,33 @@
...
+
+  export const next = () => (stepState = stepState.next());
+  export const back = () => (stepState = stepState.back());
+  export const set = (step: number) => (stepState = stepState.set(step));
+
+  let presented = false;
+</script>
+
+<Modal on:nnsClose on:introend={() => (presented = true)}>
+  <svelte:fragment slot="title"><slot name="title" /></svelte:fragment>
```
lmuntaner, line 26: Can you do:

```
<slot slot="title" name="title />
```

lmuntaner (approved):
Minor comments, looks good to me. Already approving

frontend/src/lib/components/neuron-detail/NeuronSelectPercentage.svelte
```diff
@@ -0,0 +1,73 @@
+<script lang="ts">
```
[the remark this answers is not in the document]
peterpeterparker: Yes just rename

frontend/src/lib/components/neuron-detail/NeuronConfirmActionScreen.svelte
peterpeterparker: Just rename and ui layout (like using `p class=label`). Absolutely no functional change.

frontend/src/lib/modals/LegacyWizardModal.svelte
peterpeterparker: Definitely super ugly but, indeed it will be removed.

It's also not all div but "only" the div within a section within the legacy wizard

```
<style lang="scss">
  section {
      > :global(div) {
        height: 100%;
      }
  }
</style>
```

frontend/src/lib/modals/WizardModal.svelte
peterpeterparker: Yes that works 👍

You cannot do that in Stencil, that's why I proposed original solution. Really cool to know, thanks!

PR #1381, opened 2022-09-27
GIX-1017: Menu items keep context

lmuntaner:
# Motivation

Menu items and back buttons should keep the same context.

# Changes

* New paths.derived store with four new derived stores for neurons, neuronDetails, accounts and wallet. The idea is to read the selectedProject (as well as the feature flag for now) and return the according path.
* routes inside MenuItems is dynamic and use new paths.derived.
* new paths helper in app-path.utils to return a path interpolating the context.
* Use the new paths.derived when navigating back from Wallet and NeuronDetail pages.
* Use the new paths.derived when navigation from NeuronCard or AccountCard.

# Tests

* New tests for the new paths.derived.

### DIFF
--- a/frontend/src/lib/components/common/MenuItems.svelte
+++ b/frontend/src/lib/components/common/MenuItems.svelte
@@ -95,7 +95,7 @@
   ];
 </script>
 
-{#each routes as { context, label, href, icon, statusIcon, selected }}
+{#each routes as { context, label, href, icon, statusIcon, selected } (context)}
   <MenuItem {href} testId={`menuitem-${context}`} {selected}>
     <svelte:component this={icon} slot="icon" />
     <svelte:fragment>{$i18n.navigation[label]}</svelte:fragment>
--- a/frontend/src/lib/components/common/MenuItems.svelte
+++ b/frontend/src/lib/components/common/MenuItems.svelte
@@ -19,21 +19,28 @@
   } from "../../constants/environment.constants";
   import BadgeNew from "../ui/BadgeNew.svelte";
   import GetICPs from "../ic/GetICPs.svelte";
+  import {
+    accountsPathStore,
+    neuronsPathStore,
+  } from "../../derived/paths.derived";
 
   const baseUrl: string = baseHref();
 
   const isSelectedPath = (paths: AppPath[]): boolean =>
     isRoutePath({ paths, routePath: $routeStore.path });
 
-  const routes: {
+  let routes: {
     context: string;
+    href: string;
     selected: boolean;
     label: string;
     icon: typeof SvelteComponent;
     statusIcon?: typeof SvelteComponent;
-  }[] = [
+  }[];
+  $: routes = [
     {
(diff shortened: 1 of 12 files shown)


lmuntaner:

No UX changes.

I haven't been able to test with an SNS project because we don't have one yet.

frontend/src/lib/constants/routes.constants.ts
```diff
@@ -16,4 +18,23 @@ export enum AppPath {
   NeuronDetail = "/#/u/:rootCanisterId/neuron",
 }
 
+export const paths = {
```
peterpeterparker, line 21: do we need that constant elsewhere than in the derived store?

I can live with the constant but I have to say if used nowhere else and because it is not stricto sensu what we have for constant, I am not totally sold and I would maybe inline it in the derived store.

but if you rather like to keep the constant, as I said, can live with it but then maybe can we rename it, like `appPaths` with a comment?

frontend/src/lib/components/common/MenuItems.svelte
```diff
@@ -86,12 +97,8 @@
   ];
 </script>
 
-{#each routes as { context, label, icon, statusIcon, selected }}
-  <MenuItem
-    href={`${baseUrl}#/${context}`}
-    testId={`menuitem-${context}`}
-    {selected}
-  >
+{#each routes as { context, label, href, icon, statusIcon, selected }}
```
peterpeterparker, line 100: nth: should we specify a key (https://svelte.dev/tutorial/keyed-each-blocks) if the menu ever becomes dynamic?

frontend/src/lib/constants/routes.constants.ts
lmuntaner: For now it's only used in the derived store but I thought it would be useful in general.

Maybe in `app-path.utils.ts`?

frontend/src/lib/components/common/MenuItems.svelte
lmuntaner: sure

frontend/src/lib/constants/routes.constants.ts
lmuntaner: Moved to `app-path.utils.ts` and added a comment

lmuntaner:

peterpeterparker (approved):
Thx

lmuntaner:
# Motivation

Menu items and back buttons should keep the same context.

# Changes

* New paths.derived store with four new derived stores for neurons, neuronDetails, accounts and wallet. The idea is to read the selectedProject (as well as the feature flag for now) and return the according path.
* routes inside MenuItems is dynamic and use new paths.derived.
* new paths helper in app-path.utils to return a path interpolating the context.
* Use the new paths.derived when navigating back from Wallet and NeuronDetail pages.
* Use the new paths.derived when navigation from NeuronCard or AccountCard.

# Tests

* New tests for the new paths.derived.

PR #1385, opened 2022-09-28
GIX-1005: Improve Token Amount Class

lmuntaner:
# Motivation

Upgrade nns-js.

# Changes

* Fix usage of `TokenAmount` factory functions to pass always `token` parameter.
* Remove casting when using `fromNumber`.

# Tests

* Fixed broken tests.

### DIFF
--- a/frontend/src/lib/api/sns-ledger.api.ts
+++ b/frontend/src/lib/api/sns-ledger.api.ts
@@ -34,7 +34,7 @@ export const getSnsAccounts = async ({
     principal: identity.getPrincipal(),
     balance: TokenAmount.fromE8s({
       amount: mainBalanceE8s,
-      // TODO: Decide what we do when there is no token in an SNS project
+      // TODO: https://dfinity.atlassian.net/browse/GIX-1045
       token: mapOptionalToken(metadata) ?? ICPToken,
     }),
     type: "main",
--- a/frontend/src/lib/api/sns-ledger.api.ts
+++ b/frontend/src/lib/api/sns-ledger.api.ts
@@ -1,5 +1,5 @@
 import type { Identity } from "@dfinity/agent";
-import { TokenAmount } from "@dfinity/nns";
+import { ICPToken, TokenAmount } from "@dfinity/nns";
 import type { Principal } from "@dfinity/principal";
 import type { Account } from "../types/account";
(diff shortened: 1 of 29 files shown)


frontend/src/lib/api/sns-ledger.api.ts
```diff
@@ -34,7 +34,8 @@ export const getSnsAccounts = async ({
     principal: identity.getPrincipal(),
     balance: TokenAmount.fromE8s({
       amount: mainBalanceE8s,
-      token: mapOptionalToken(metadata),
+      // TODO: Decide what we do when there is no token in an SNS project
```
peterpeterparker, line 37: should we track this with a jira issue? sounds kind of like an essential question to me

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
```diff
@@ -92,7 +94,7 @@
         account: sourceAccount,
         // If we made it here, we know that the amount is valid
         // TODO: change fromNumber return type
```
peterpeterparker, line 96: we still need this todo or it is actually solve with this change?

frontend/src/lib/api/sns-ledger.api.ts
lmuntaner: sure, I added it as a comment https://dfinity.atlassian.net/browse/GIX-1045

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
lmuntaner: Good point, I can remove this comment

lmuntaner:
`NewTransaction` uses ICPToken as default, but we might want to change that.

Worth checking with Mischa and the rest of the team what should we do then.

frontend/src/lib/api/sns-ledger.api.ts
peterpeterparker: Thanks

peterpeterparker:
It's more a technical question than UX I would say but sure should discuss it.

peterpeterparker (approved):
LGTM, thx

PR #1394, opened 2022-09-29
feat: use main domain to serve icons and social image

peterpeterparker:
# TODO

- [ ] **This PR still need to be tested on the new testnets - waiting for the valid SSL certificates** 

# Motivation

The IC now supports crawlers therefore, technically, it should be possible to serve icons and social image from `.ic0.app` instead of `.raw.ic0.app`.

# Changes

- drop `.raw` domain in icons and social image links and therefore in csp as well


peterpeterparker:
As II task (https://github.com/dfinity/internet-identity/pull/921) this PR is paused until valid SSL is supported on the new testens

peterpeterparker:
It seems to still do not works out, therefore I close this PR.

Capture d’écran 2023-02-06 à 08 36 24

PR #1395, opened 2022-09-29
Chore: UI Improvements

lmuntaner:
# Motivation

Add consistency in labels.

# Changes

* Change h5 used as labels for `p class="label"`.
* Use KeyValuePair in ProjectTimeline
* Set color in LegacyWizardModal to background-contrast.

# Tests

Only style changes.

### DIFF
--- a/frontend/src/lib/components/accounts/HardwareWalletInfo.svelte
+++ b/frontend/src/lib/components/accounts/HardwareWalletInfo.svelte
@@ -13,25 +13,25 @@
   );
 </script>
 
-<h5>{$i18n.core.principal}</h5>
-<p class="value">{principalText}</p>
-
-<h5>{$i18n.accounts.account_identifier}</h5>
-<p class="value">{accountIdentifier}</p>
+<div class="wrapper">
+  <div>
+    <p class="label">{$i18n.core.principal}</p>
+    <p class="value">{principalText}</p>
+  </div>
+  <div>
+    <p class="label">{$i18n.accounts.account_identifier}</p>
+    <p class="value">{accountIdentifier}</p>
+  </div>
+</div>
 
 <style lang="scss">
-  h5 {
-    margin: 0;
-    align-self: flex-start;
+  .wrapper {
+    display: flex;
+    flex-direction: column;
+    gap: var(--padding);
   }
 
   p {
-    margin: 0 0 var(--padding-0_5x);
     word-wrap: break-word;
-    align-self: flex-start;
-
-    &:first-of-type {
-      margin-bottom: var(--padding-2x);
-    }
   }
 </style>
--- a/frontend/src/lib/modals/LegacyWizardModal.svelte
+++ b/frontend/src/lib/modals/LegacyWizardModal.svelte
@@ -43,5 +43,7 @@
   @use "../themes/mixins/modal";
   section {
     @include modal.section;
+
+    color: var(--background-contrast);
   }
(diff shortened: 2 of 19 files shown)


lmuntaner:
New UI:

Screenshot 2022-09-29 at 15 14 17
Screenshot 2022-09-29 at 15 14 31
Screenshot 2022-09-29 at 15 14 49
Screenshot 2022-09-29 at 15 15 11
Screenshot 2022-09-29 at 15 15 26
Screenshot 2022-09-29 at 15 15 32
Screenshot 2022-09-29 at 15 16 48
Screenshot 2022-09-29 at 15 22 07
Screenshot 2022-09-29 at 15 24 06
Screenshot 2022-09-29 at 15 25 31

lmuntaner:

frontend/src/lib/components/accounts/HardwareWalletInfo.svelte
```diff
@@ -13,25 +13,25 @@
   );
 </script>
 
-<h5>{$i18n.core.principal}</h5>
-<p class="value">{principalText}</p>
-
-<h5>{$i18n.accounts.account_identifier}</h5>
-<p class="value">{accountIdentifier}</p>
+<div class="wrapper">
```
peterpeterparker, line 16: not sure you need muchs divs and display flex.

```
<p class="label">{$i18n.core.principal}</p>
<p class="value">{principalText}</p>
<p class="label">{$i18n.accounts.account_identifier}</p>
<p class="value">{accountIdentifier}</p>

<style lang="scss">
   .space {
      margin-bottom: var(--padding);
   }
</style>
```

isn't the same?

frontend/src/lib/modals/LegacyWizardModal.svelte
```diff
@@ -43,5 +43,7 @@
   @use "../themes/mixins/modal";
   section {
     @include modal.section;
+
+    color: var(--background-contrast);
```
peterpeterparker, line 47: shouldn't this be added in the mixin `modal.section` so that other legacy modals inherits it or it is just an issue with the wizard?

frontend/src/lib/components/accounts/HardwareWalletInfo.svelte
lmuntaner: Yes, same result. To me it has more meaning the other one. There are two blocks, and then it's within blocks that we care about the space. But I guess the main target is to reduce html elements?

frontend/src/lib/modals/LegacyWizardModal.svelte
lmuntaner: We should discuss this. I don't know. It depends a lot on what we want as global color and what not.

I fixed the issue for now, but I don't know what we want to do to have this maintainable.

frontend/src/lib/components/accounts/HardwareWalletInfo.svelte
peterpeterparker: Yes, I play a lot with the DOM - notably when I develop my rich text editor - that's why as a best practice I like to keep the tree as clean and lean as possible. The smaller the dom, the faster the rendering.

I'm not sure it adds much meaning semantically speaking because a div is a generic container but if you think it does, totally fine with it.

Thanks for the feedback.

peterpeterparker (approved):
LGTM

PR #1403, opened 2022-10-04
GIX-998: Get SNS tokens from test account

lmuntaner:
# Motivation

Testnet users can get any kind of tokens.

# Changes

* Use the context to differentiate between transferring from NNS or from SNS.
* Change modal text to use token of the SNS project.

# Tests

Not applicable, this is just a feature for testnets.

### DIFF
--- a/frontend/src/lib/api/sns-wrapper.api.ts
+++ b/frontend/src/lib/api/sns-wrapper.api.ts
@@ -56,7 +56,7 @@ const listSnses = async ({
   );
 };
 
-const initSns = async ({
+export const initSns = async ({
   agent,
   rootCanisterId,
   certified,
--- a/frontend/src/lib/components/ic/GetTokens.svelte
+++ b/frontend/src/lib/components/ic/GetTokens.svelte
@@ -151,4 +151,8 @@
 
     padding: var(--padding-2x) var(--padding);
   }
+
+  button.primary {
+    min-width: var(--padding-8x);
+  }
 </style>
--- a/frontend/src/lib/components/ic/GetTokens.svelte
+++ b/frontend/src/lib/components/ic/GetTokens.svelte
@@ -18,16 +18,13 @@
 
   let inputValue: number | undefined = undefined;
 
-  const onSubmit = async ({ target }) => {
-    if (invalidForm) {
+  const onSubmit = async () => {
+    if (invalidForm || inputValue === undefined) {
       toastsError({
         labelKey: "Invalid ICPs input.",
       });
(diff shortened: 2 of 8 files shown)


lmuntaner:

frontend/src/lib/components/ic/GetTokens.svelte
```diff
@@ -92,7 +106,7 @@
     {#if transferring}
       <Spinner />
     {:else}
-      Get
+      Get tokens
```
peterpeterparker, line 109: since you are already there, can you set a `min-with` to the button so that it does not shrink when processing, when the spinner is displayed?

lmuntaner: sure

frontend/src/lib/services/dev.services.ts
```diff
@@ -19,3 +25,29 @@ export const getICPs = async (icps: number) => {
 
   await syncAccounts();
 };
+
+export const getTokens = async ({
```
peterpeterparker, line 29: can we also add a test as for `getICPs`?

frontend/src/lib/api/sns-wrapper.api.ts
```diff
@@ -56,7 +56,7 @@ const listSnses = async ({
   );
 };
 
-const initSns = async ({
+export const initSns = async ({
```
peterpeterparker, line 59: if we expose this feature, can we add a test?

frontend/src/lib/services/dev.services.ts
lmuntaner: it's only for testnet, so I don't think it's the best use of our time. Moreover the e2e test we have uses this button, therefore it's also tested there.

frontend/src/lib/api/sns-wrapper.api.ts
lmuntaner: sure

lmuntaner: done

frontend/src/lib/components/ic/GetTokens.svelte
lmuntaner: done

lmuntaner: Screenshot 2022-10-05 at 08 44 10

lmuntaner:

peterpeterparker (approved):
LGTM, thx

PR #1404, opened 2022-10-04
GIX-995: SNS Wallet details

lmuntaner:
# Motivation

User has access to the details page with main data of SNS account.

# Changes

* Implement SnsWallet to set up the context to be able to render WalletSummary.
* Encode the Sns Account in the "identifier" property.

# Tests

* Test SnsWallet page.
* Fix ugly comment in Wallet page test.

### DIFF
--- a/frontend/src/lib/pages/SnsWallet.svelte
+++ b/frontend/src/lib/pages/SnsWallet.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
   import { Spinner } from "@dfinity/gix-components";
-  import type { Principal } from "@dfinity/principal";
   import { onMount } from "svelte";
   import { onDestroy, setContext } from "svelte/internal";
   import { writable, type Unsubscriber } from "svelte/store";
--- a/frontend/src/lib/pages/SnsWallet.svelte
+++ b/frontend/src/lib/pages/SnsWallet.svelte
@@ -1,8 +1,23 @@
 <script lang="ts">
+  import { Spinner } from "@dfinity/gix-components";
+  import type { Principal } from "@dfinity/principal";
   import { onMount } from "svelte";
+  import { onDestroy, setContext } from "svelte/internal";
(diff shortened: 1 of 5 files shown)


lmuntaner:

lmuntaner:
New UI:

Screenshot 2022-10-04 at 14 32 51

frontend/src/lib/pages/SnsWallet.svelte
```diff
@@ -1,16 +1,78 @@
...
+      selectedAccountStore.update(() => ({
+        account: selectedAccount,
+      }));
+    }
+  }
 </script>
 
-<!-- TODO: GIX-995 Implement SNS Wallet details -->
-<div data-tid="sns-wallet">TO BE IMPLEMENTED</div>
+<main class="legacy" data-tid="sns-wallet">
```
peterpeterparker, line 70: should/can we implement the new ui?

lmuntaner: Since they both reuse the same code, I'd say later if we do so. Not when I am implementing the SNS accounts.

peterpeterparker: Does upgrading both to new design while implementing Sns adds an acceptable amount of work?

i.e. is upgrading design while implementing Sns significantely less work than implementing Sns while extracting legacy design and upgrading design afterwards?

Totally fine with the PR and using legacy but if there is chance to spare some effort...

peterpeterparker: Does not block this PR though, we can discuss later today

peterpeterparker (approved):
LGTM, thx

frontend/src/lib/pages/SnsWallet.svelte
lmuntaner: We could have started implementing the new design first in NNS and then implement for SNS. But the timings do not help to do that.

PR #1407, opened 2022-10-05
GIX-1041: Add feature flag sns-2

lmuntaner:
# Motivation

SNS-1 does not include sns accounts.

# Changes

* Add a new feature flag ENABLE_SNS_2
* Use the new feature flag in all sns accounts related components.

# Tests

Not applicable

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -96,6 +96,7 @@
         "IDENTITY_SERVICE_URL": "https://identity.ic0.app/",
         "FEATURE_FLAGS": {
           "ENABLE_SNS": false,
+          "ENABLE_SNS_2": false,
           "VOTING_UI": "modern"
         }
       },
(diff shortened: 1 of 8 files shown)


lmuntaner:

You can play with how it would look like here: https://tzq7c-xqaaa-aaaaa-aaamq-cai.small13.dfinity.network/#/accounts

There is some UI change when moving from a Neurons SNS project to accounts. We can try to remove this in another PR if needed. Although I hope it won't be much time before we can enable SNS_2.

peterpeterparker (approved):
LGTM

p.s.: `ENABLE_SNS_ACCOUNTS` instead of `ENABLE_SNS_2` would have work too?

lmuntaner:
> LGTM
> 
> p.s.: `ENABLE_SNS_ACCOUNTS` instead of `ENABLE_SNS_2` would have work too?

Yes, but not sure it will be more meaningful that name. We might also hide some functionality in SNS Neurons for example.

PR #1408, opened 2022-10-05
GIX-1044: Extend transaction fee store

lmuntaner:
# Motivation

Each SNS project might have a different transaction fee. Extend the current functionality to accommodate that.

# Changes

* Add "transactionFee" function to sns-ledger api.
* Extend the transactionFeeStore with a "projects" property to store different transaction fees.
* New transaction fee service to load sns transaction fee to the store.
* Load SNS transaction fees in SnsAccounts and SnsWallet.
* New derived store that returns the fee of the current project context in a TokenAmount type.

# Tests

* New test new api function.
* New test for extended transactionFeeStore functionality.
* New test for transaction fee service.
* Add calls check on SnsAccounts and SnsWallet.
* New test for new derived store.

### DIFF
--- a/frontend/src/lib/pages/SnsAccounts.svelte
+++ b/frontend/src/lib/pages/SnsAccounts.svelte
@@ -13,6 +13,7 @@
   import SkeletonCard from "../components/ui/SkeletonCard.svelte";
   import { routeStore } from "../stores/route.store";
   import { walletPathStore } from "../derived/paths.derived";
+  import { loadSnsTransactionFee } from "../services/transaction-fees.services";
 
   let loading: boolean = false;
   const unsubscribe: Unsubscriber = snsOnlyProjectStore.subscribe(
@@ -21,6 +22,8 @@
         // TODO: improve loading and use in memory sns neurons or load from backend
         loading = true;
         await loadSnsAccounts(selectedProjectCanisterId);
+        // No need to wait for the transaction fee
+        loadSnsTransactionFee(selectedProjectCanisterId);
         loading = false;
       }
     }
--- a/frontend/src/lib/pages/SnsAccounts.svelte
+++ b/frontend/src/lib/pages/SnsAccounts.svelte
@@ -4,7 +4,7 @@
   import type { Unsubscriber } from "svelte/store";
   import AccountsTitle from "../components/accounts/AccountsTitle.svelte";
   import { snsOnlyProjectStore } from "../derived/selected-project.derived";
-  import { loadSnsAccounts } from "../services/sns-accounts.services";
+  import { syncSnsAccounts } from "../services/sns-accounts.services";
   import { snsProjectAccountsStore } from "../derived/sns/sns-project-accounts.derived";
   import AccountCard from "../components/accounts/AccountCard.svelte";
   import { i18n } from "../stores/i18n";
@@ -13,7 +13,6 @@
   import SkeletonCard from "../components/ui/SkeletonCard.svelte";
   import { routeStore } from "../stores/route.store";
   import { walletPathStore } from "../derived/paths.derived";
-  import { loadSnsTransactionFee } from "../services/transaction-fees.services";
 
   let loading: boolean = false;
   const unsubscribe: Unsubscriber = snsOnlyProjectStore.subscribe(
@@ -21,9 +20,7 @@
       if (selectedProjectCanisterId !== undefined) {
         // TODO: improve loading and use in memory sns neurons or load from backend
         loading = true;
-        await loadSnsAccounts(selectedProjectCanisterId);
-        // No need to wait for the transaction fee
-        loadSnsTransactionFee(selectedProjectCanisterId);
+        await syncSnsAccounts(selectedProjectCanisterId);
         loading = false;
       }
     }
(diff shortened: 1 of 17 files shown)


lmuntaner:

It's on draft, but it's pending only the upgrade of sns-js.

frontend/src/lib/pages/SnsAccounts.svelte
```diff
@@ -21,6 +22,8 @@
         // TODO: improve loading and use in memory sns neurons or load from backend
         loading = true;
         await loadSnsAccounts(selectedProjectCanisterId);
+        // No need to wait for the transaction fee
+        loadSnsTransactionFee(selectedProjectCanisterId);
```
peterpeterparker, line 26: would be nicer to have both `loadSnsAccounts` and `loadSnsTransactionFee` in a service rather than chained in the component.

In the PR this logic is repeated twice - in two components, there is a bit of logic linked to it and it would follow the pattern (see `app.service` - `initApp`)

frontend/src/lib/services/transaction-fees.services.ts
```diff
@@ -1,15 +1,44 @@
-import { transactionFee } from "../api/ledger.api";
+import type { Principal } from "@dfinity/principal/lib/cjs";
+import { transactionFee as nnsTransactionFee } from "../api/ledger.api";
```
peterpeterparker, line 2: should we effectively rename the functions `nnsTransa...` and `snsTransa...` in the api layers?

peterpeterparker (commented):
if a sns transaction fee has already been fetched, for perfomance reason, shouldn't we first check if the fee exists in store and if so, use the value from store rather than fetch again?

likewise for nns transaction fee?

frontend/src/lib/stores/transaction-fees.store.ts
```diff
@@ -1,13 +1,22 @@
+import type { Principal } from "@dfinity/principal";
 import { derived, writable } from "svelte/store";
 import { DEFAULT_TRANSACTION_FEE_E8S } from "../constants/icp.constants";
 
 export type TransactionFeesStore = {
   // Main Ledger of IC
   main: bigint;
```
peterpeterparker, line 7: since you add the info `certified` for sns fees, should we also add it for nns?

frontend/src/lib/pages/SnsAccounts.svelte
peterpeterparker: I also literally 2min later saw a same new `loadSnsAccounts` added while reviewing @mstrasinskis new PR https://github.com/dfinity/nns-dapp/pull/1409 in `DisburseSnsNeuronModal`. You both will have to sync your PR about this.

frontend/src/lib/services/transaction-fees.services.ts
lmuntaner: Not for me, at that level we are talking about transaction fees because the context is known. It's when importing them that the context needs to be clarified.

Otherwise we should do that same also for proposals, renaming to `nnsProposals`, also renaming all types in nns-js, ...

I believe that even adding `SnsXxxx` inside sns-js was not the best choice and instead if should have been renamed inside nns-dapp.

frontend/src/lib/stores/transaction-fees.store.ts
lmuntaner: I can do that in another PR, sure.

lmuntaner:
> if a sns transaction fee has already been fetched, for perfomance reason, shouldn't we first check if the fee exists in store and if so, use the value from store rather than fetch again?
> 
> likewise for nns transaction fee?

Well, for safety reasons I preferred to fetch again. I don't believe that it changes the performance for the user because the one in the store will be used while it is refetched again. Therefore, what might only happens is that for 1-2 seconds the user sees a fee of 1 that then changes to 1.1 for example.

I believe that is better than showing 1 when it's 1.1, or showing a spinner during those 1-2 seconds.

frontend/src/lib/pages/SnsAccounts.svelte
peterpeterparker: Load on demand is fine I just mean we should have the logic tight together and not duplicated multiple times because otherwise we might have a case where we load accounts but forgot to load the fee.

Something like:

```
export const syncSnsAccounts = () => {
    loadSnsAccounts 
    updateStore
    
    // here my other question about do we want to always reload the fee?
    loadRelatedSnsFee 
    updateStore
}
```

Then use `syncSnsAccounts` in components - i.e. in `SnsAccounts.svelte`, in `SnsWallet.svelte`, in `DisburseSnsNeuronModal.svelte` and elsewhere (did not check)

Just a spontaneous thought, open to better solution.

frontend/src/lib/services/transaction-fees.services.ts
peterpeterparker: Ok works for me

peterpeterparker:
Ok let's refetch the fee, we can always add a logic later if needed

frontend/src/lib/pages/SnsAccounts.svelte
peterpeterparker: Sure or an option to "reloadFee" or not, as long as it's sync and not duplicated, I'm happy

lmuntaner: done

lmuntaner:

peterpeterparker (approved):
LGTM, thx

PR #1409, opened 2022-10-05
Fix/disburse destination

mstrasinskis:
# Motivation

- fix wrong destination address in disburse confirmation screen (screenshots)

# Changes

- new derived store `snsProjectMainAccountStore`
- load sns project accounts when not available
- DisburseButton refactoring to avoid test errors

# Tests

- snsProjectMainAccountStore
- disburse modal should load sns project accounts
- disburse button (updated because of refactoring)

# Screenshots

image
now it matches to sns project account
image


frontend/src/lib/components/neuron-detail/actions/DisburseButton.svelte
```diff
@@ -1,11 +1,12 @@
...
   import { i18n } from "../../../stores/i18n";
   import type { NeuronInfo } from "@dfinity/nns";
-  import type { SvelteComponent } from "svelte";
   import type { SnsNeuron } from "@dfinity/sns";
+  import DisburseNnsNeuronModal from "../../../modals/neurons/DisburseNnsNeuronModal.svelte";
+  import DisburseSnsNeuronModal from "../../../modals/neurons/DisburseSnsNeuronModal.svelte";
 
-  export let neuron: NeuronInfo | SnsNeuron;
-  export let modal: typeof SvelteComponent;
+  export let nnsNeuron: NeuronInfo | undefined = undefined;
```
peterpeterparker, line 8: it's not possible to keep `neuron: NeuronInfo | SnsNeuron` and detect the types to present the corresponding modal?

like a `isNnsNeuron()`?

frontend/src/lib/components/neuron-detail/actions/DisburseButton.svelte
```diff
@@ -18,10 +19,13 @@
...
 
 {#if showModal}
-  <svelte:component
-    this={modal}
-    {neuron}
-    {reloadContext}
-    on:nnsClose={closeModal}
-  />
+  {#if nnsNeuron !== undefined}
+    <DisburseNnsNeuronModal neuron={nnsNeuron} on:nnsClose={closeModal} />
```
peterpeterparker, line 23: `reloadContext` is not used anymore here, it's on purpose?

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
```diff
@@ -3,6 +3,10 @@ import { snsAccountsStore } from "../../stores/sns-accounts.store";
 import type { Account } from "../../types/account";
 import { snsProjectSelectedStore } from "../selected-project.derived";
 
+const mainAccount = (accounts: Account[]): Account | undefined => {
```
peterpeterparker, line 6: is it maybe a piece of code we have many times in the code base and therefore worth a utils if not already exists yet?

frontend/src/lib/components/neuron-detail/actions/DisburseButton.svelte
lmuntaner, line 8: I didn't realize it before, but I'd rather have `DisburseSnsNeuronButton` and `DisburseNnsNeuronButton`. But not super strong opinion.

lmuntaner (approved):
Personal preference comment, but looks good to me. Already approving.

frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
```diff
@@ -58,6 +57,27 @@
...
+    async (selectedProjectCanisterId) => {
+      if (
+        selectedProjectCanisterId === undefined ||
+        $snsProjectMainAccountStore !== undefined
+      ) {
+        return;
+      }
+
+      loading = true;
+      await loadSnsAccounts(selectedProjectCanisterId);
```
peterpeterparker, line 74: should you sync main and use the new `syncSnsAccounts` function here?

frontend/src/lib/components/neuron-detail/actions/DisburseButton.svelte
mstrasinskis: Yes, there is no context data for the neuron. All the updates are done on sevices level (store updates).

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
mstrasinskis: Added account-utils/`mainAccount` function

frontend/src/lib/components/neuron-detail/actions/DisburseButton.svelte
peterpeterparker: Sounds good 👍

frontend/src/lib/derived/sns/sns-project-accounts.derived.ts
peterpeterparker: Cool thx

frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
mstrasinskis: If you mean nns/main account then sns disburse shouldn't change it (as I understood it)

peterpeterparker: I mean @lmuntaner has add a new function in his last PR that loads the Sns accounts but also fetch the transaction fee. So my question is if it makes sense here to just load the account or if it would not also make sense to fetch the fee and therefore to use `syncSnsAccounts()` instead of `loadSnsAccounts()`?

frontend/src/lib/components/neuron-detail/actions/DisburseButton.svelte
lmuntaner: To be honest, I would prefer to have only one type and use different components.

lmuntaner: I prefer different components. The same way we have SnsNeuronCard. And then inside there we reuse the components.

I believe that is more maintainable than adding checks inside the polymorphic components.

frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
mstrasinskis: Good point, thanks.

PR #1410, opened 2022-10-06
GIX-1045: Manage when sns token not defined

lmuntaner:
# Motivation

Manage the possible situations where sns token might not be present.

# Changes

* Throw an error when loading sns accounts and token is not present.
* Cast the token in components when we know the token is present otherwise the component would not be rendered.

# Tests

* Add test to check an error is thrown when token is not present.

### DIFF
--- a/frontend/src/lib/api/sns-ledger.api.ts
+++ b/frontend/src/lib/api/sns-ledger.api.ts
@@ -1,8 +1,9 @@
 import type { Identity } from "@dfinity/agent";
-import { ICPToken, TokenAmount } from "@dfinity/nns";
+import { TokenAmount } from "@dfinity/nns";
 import type { Principal } from "@dfinity/principal";
 import { encodeSnsAccount } from "@dfinity/sns";
 import type { Account } from "../types/account";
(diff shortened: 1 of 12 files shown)


lmuntaner:

frontend/src/lib/api/sns-ledger.api.ts
```diff
@@ -30,13 +31,18 @@ export const getSnsAccounts = async ({
     ledgerMetadata({}),
   ]);
 
+  const projectToken = mapOptionalToken(metadata);
+
+  if (projectToken === undefined) {
```
peterpeterparker, line 36: maybe this can be extracted to a new `assertTokenDefined` or something like that? assuming we will reuse the logic

peterpeterparker (approved):
LGTM, just minor comment

frontend/src/lib/api/sns-ledger.api.ts
lmuntaner: Not sure we are reusing the logic. We do it here and when we create the projects. But afterwards, we have it as defined.

PR #1415, opened 2022-10-06
fix: preserve login page mobile view in landscape mode

mstrasinskis:
# Motivation

There is some overflows or hidden content on the login page in the landscape mode. A problem affects some mobile devices like iphone 11+ and pixels. 

image

# Changes

- lock the mobile view till there is not enough vertical space
- make the login main content scrollable on mobile

# Screenshots

image


peterpeterparker (approved):
LGTM, thanks

PR #1419, opened 2022-10-07
GIX-994: User transfers SNS tokens

lmuntaner:
# Motivation

User can transfer SNS tokens.

# Changes

* New modal SnsTransactionModal.
* New sns accounts service to transfer tokens.
* New sns ledger api function to transfer tokens.
* Add support for SNS accounts in helper "invalidAddress".
* accountsListStore returns the list of accounts of the context account.
* getAccountFromStore expect an array of acconts instead of the type of NNS Accounts Store.
* New SnsAccountsFooter that renders SnsTransactionModal in Accounts route.
* Moved routePathNeuronId from services to utils to avoid cycle dependency.
* Moved routePathAccountIdentifier from services to utils to avoid cycle dependency.

# Tests

* Test new modal.
* Test new sns account service.
* Test new sns ledger api function.
* Test new functionality in Accounts route to open SnsTransactionModal.
* Add test cases to invalidAddress.
* Moved according tests from services to utils.
* Fix broken tests.
* Add test cases to accountsListStore.

### DIFF
--- a/frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
+++ b/frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
@@ -2,8 +2,8 @@
   import { i18n } from "../../stores/i18n";
   import { Toolbar } from "@dfinity/gix-components";
   import Footer from "../common/Footer.svelte";
-  import IcpTransactionModal from "../../modals/accounts/IcpTransactionModal.svelte";
   import { snsProjectAccountsStore } from "../../derived/sns/sns-project-accounts.derived";
+  import SnsTransactionModal from "$lib/modals/accounts/SnsTransactionModal.svelte";
 
   // TODO: Support adding subaccounts
   let modal: "NewTransaction" | undefined = undefined;
@@ -12,7 +12,7 @@
 </script>
 
 {#if modal === "NewTransaction"}
-  <IcpTransactionModal on:nnsClose={closeModal} />
+  <SnsTransactionModal on:nnsClose={closeModal} />
 {/if}
 
 {#if $snsProjectAccountsStore !== undefined}
--- a/frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
+++ b/frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
@@ -1,0 +1,29 @@
+<script lang="ts">
+  import { i18n } from "../../stores/i18n";
+  import { Toolbar } from "@dfinity/gix-components";
+  import Footer from "../common/Footer.svelte";
+  import IcpTransactionModal from "../../modals/accounts/IcpTransactionModal.svelte";
+  import { snsProjectAccountsStore } from "../../derived/sns/sns-project-accounts.derived";
+
+  // TODO: Support adding subaccounts
+  let modal: "NewTransaction" | undefined = undefined;
+  const openNewTransaction = () => (modal = "NewTransaction");
+  const closeModal = () => (modal = undefined);
+</script>
+
+{#if modal === "NewTransaction"}
+  <IcpTransactionModal on:nnsClose={closeModal} />
+{/if}
+
+{#if $snsProjectAccountsStore !== undefined}
+  <Footer>
+    <Toolbar>
+      <button
+        class="primary full-width"
+        on:click={openNewTransaction}
+        data-tid="open-new-sns-transaction"
+        >{$i18n.accounts.new_transaction}</button
+      >
+    </Toolbar>
+  </Footer>
+{/if}
--- a/frontend/src/lib/derived/accounts-list.derived.ts
+++ b/frontend/src/lib/derived/accounts-list.derived.ts
@@ -2,8 +2,8 @@
  * A derived store that returns the accounts as an array of accounts.
  */
 
-import { isNnsProjectStore } from "$lib/selected-project.derived";
-import { snsProjectAccountsStore } from "$lib/sns/sns-project-accounts.derived";
+import { isNnsProjectStore } from "$lib/derived/selected-project.derived";
+import { snsProjectAccountsStore } from "$lib/derived/sns/sns-project-accounts.derived";
 import { accountsStore } from "$lib/stores/accounts.store";
 import { derived } from "svelte/store";
 
--- a/frontend/src/lib/derived/accounts-list.derived.ts
+++ b/frontend/src/lib/derived/accounts-list.derived.ts
@@ -2,19 +2,19 @@
  * A derived store that returns the accounts as an array of accounts.
  */
 
-import { isNnsProjectStore } from "$lib/derived/selected-project.derived";
-import { snsProjectAccountsStore } from "$lib/derived/sns/sns-project-accounts.derived";
 import { accountsStore } from "$lib/stores/accounts.store";
 import { derived } from "svelte/store";
+import { isNnsProjectStore } from "./selected-project.derived";
+import { snsProjectAccountsStore } from "./sns/sns-project-accounts.derived";
 
 export const accountsListStore = derived(
   [accountsStore, isNnsProjectStore, snsProjectAccountsStore],
-  ([{ main, subAccounts, hardwareWallets }, isNns, projectAccounts]) => {
+  ([{ main, subAccounts, hardwareWallets }, isNns, snsAccounts]) => {
     if (isNns) {
       return main === undefined
         ? []
         : [main, ...(subAccounts ?? []), ...(hardwareWallets ?? [])];
     }
-    return projectAccounts ?? [];
+    return snsAccounts ?? [];
   }
 );
--- a/frontend/src/lib/derived/accounts-list.derived.ts
+++ b/frontend/src/lib/derived/accounts-list.derived.ts
@@ -3,11 +3,15 @@
  */
 
 import { accountsStore } from "$lib/stores/accounts.store";
-import { derived } from "svelte/store";
+import type { Account } from "$lib/types/account";
+import { derived, type Readable } from "svelte/store";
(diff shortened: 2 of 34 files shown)


lmuntaner:

frontend/src/lib/derived/accounts-list.derived.ts
```diff
@@ -4,11 +4,17 @@
 
 import { accountsStore } from "$lib/stores/accounts.store";
 import { derived } from "svelte/store";
+import { isNnsProjectStore } from "./selected-project.derived";
+import { snsProjectAccountsStore } from "./sns/sns-project-accounts.derived";
 
 export const accountsListStore = derived(
```
peterpeterparker, line 10: can you add a comment about what to does the derived store?

frontend/src/lib/pages/NnsWallet.svelte
```diff
@@ -76,7 +77,7 @@
   let selectedAccount: Account | undefined;
   $: selectedAccount = getAccountFromStore({
     identifier: routeAccountIdentifier?.accountIdentifier,
-    accountsStore: $accountsStore,
+    accounts: $accountsListStore,
```
peterpeterparker, line 80: isn't this incorrect? `NnsWallet.svelte` should not be rendered with Sns accounts.

frontend/src/lib/services/accounts.services.ts
```diff
@@ -199,7 +172,7 @@ export const getAccountIdentity = async (
 ): Promise<Identity | LedgerIdentity> => {
   const account: Account | undefined = getAccountFromStore({
     identifier,
-    accountsStore: get(accountsStore),
+    accounts: get(accountsListStore),
```
peterpeterparker, line 175: dito

frontend/src/lib/utils/accounts.utils.ts
```diff
@@ -36,7 +39,16 @@ export const invalidAddress = (address: string | undefined): boolean => {
     checkAccountId(address);
     return false;
   } catch (_) {
-    return true;
+    try {
+      // It might also be an SNS address
+      decodeSnsAccount(address);
```
peterpeterparker, line 44: can we solve this without adding logic within a catch?

frontend/src/lib/utils/accounts.utils.ts
```diff
@@ -67,30 +79,16 @@ export const isAccountHardwareWallet = (
...
-  );
-
-  if (subAccount !== undefined) {
-    return subAccount;
-  }
-
-  return hardwareWallets?.find(
-    (account: Account) => account.identifier === identifier
-  );
+  return accounts.find(({ identifier: id }) => id === identifier);
```
peterpeterparker, line 91: can a sns account have the same id has a nns account?

frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
```diff
@@ -0,0 +1,29 @@
...
+  import SnsTransactionModal from "$lib/modals/accounts/SnsTransactionModal.svelte";
+
+  // TODO: Support adding subaccounts
+  let modal: "NewTransaction" | undefined = undefined;
+  const openNewTransaction = () => (modal = "NewTransaction");
+  const closeModal = () => (modal = undefined);
+</script>
+
+{#if modal === "NewTransaction"}
+  <SnsTransactionModal on:nnsClose={closeModal} />
```
peterpeterparker, line 15: no sure why this if and why there is a `..Modal` in a `..Footer` component. is this correct?

frontend/src/lib/derived/accounts-list.derived.ts
lmuntaner: sure

frontend/src/lib/pages/NnsWallet.svelte
lmuntaner: I see that `accountsListStore` is not a good name...

`accountsListStore` returns the accounts of the selected project, NNS or SNS.

I will add the comment

frontend/src/lib/utils/accounts.utils.ts
lmuntaner: Not in the nns, but two SNSs can have the same identifer.

What I am relying here is that the accounts will be already filtered by project.

One more thing to add to my comment that I don't rely how we are relying so much with the global project state.

frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
lmuntaner: Yes, that is correct.

This is how we have it before and I kept the structure:

```
// Accounts.svelte
<main>
  // iterate over accounts
  <AccountCard />
</main>

<AccountsFooter />
// Inside the footer we had the button and the modal.
```

I don't remember now whether the Footer was a component or we had the buttons there directly. But when I splitted the functionality, the buttons had the functionality of opening and closing the modal. That's why I kept it there.

frontend/src/lib/utils/accounts.utils.ts
lmuntaner: Not sure.

The helpers we use to check that raise an error. A valid NNS account is not a valid SNS account.

I could create a helper for each, `invalidNnsAddress` and `invalidSnsAddress` for example. But then, I also need to pass that as prop or create another `AddressInput`.

frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
peterpeterparker: Gotcha, thanks

frontend/src/lib/pages/NnsWallet.svelte
peterpeterparker: I got it but it is what I meant. If `accountsListStore` returns a Sns project for `NnsWallet`.

I understand `NnsWallet` is not rendered because there is a `if` above that display Nns or Sns wallet but, nevertheless, not sure `NnsWallet` should use `accountsListStore` because it is explicitely Nns.

i.e. `$accountsStore` seems more appropriate and less prone to error if someone misused the component

frontend/src/lib/utils/accounts.utils.ts
peterpeterparker: Would be neat to find a solution, adding logic in a catch doesn't sound ideal

Not in this PR but maybe worth a TODO in with low prio

frontend/src/lib/pages/NnsWallet.svelte
lmuntaner: I wanted to use this inside the SelectAccountDropdown, which can be used for NNS and SNS project.

I renamed it to `accountsSelectedProjectStore`, which returns list of accounts, NNS or SNS.

Then we have `snsProjectAccountsStore` which returns the list of accounts the selected SNS project. No NNS.

If we consider NNS as a project same Open Chat, it makes sense to me to have helpers that return NNS or SNS.

The other option that I might have is to pass the accounts list to `SelectAccountDropdown` as prop. That would also mean to pass it to TransactionModal and to TransactionForm and everywhere where we use `SelectAccountDropdown`. I would prefer that, but that means more props.

frontend/src/lib/utils/accounts.utils.ts
lmuntaner: Ok, I also think we should find a solution. This SNS vs NNS is adding a lot of complexity to this code without adding much functionality.

lmuntaner: done

frontend/src/lib/derived/accounts-list.derived.ts
lmuntaner: done

frontend/src/lib/modals/accounts/SnsTransactionModal.svelte
```diff
@@ -0,0 +1,68 @@
...
+  const dispatcher = createEventDispatcher();
+  const transfer = async ({
+    detail: { sourceAccount, amount, destinationAddress },
+  }: CustomEvent<NewTransaction>) => {
+    startBusy({
+      initiator: "accounts",
+    });
+
+    // Errors in amount already handled by TransactionModal
+    const tokenAmount = TokenAmount.fromNumber({
```
peterpeterparker, line 32: maybe this can be done in the service too? so the exception even if already handled can be handled by the service

frontend/src/lib/utils/accounts.utils.ts
peterpeterparker: thanks for the feedback

frontend/src/lib/pages/NnsWallet.svelte
peterpeterparker: Yeah really not sold. If it would be a generic component, agree but it is not, it is explicitely  a `Nns...` related component therefore think it is prone to error to error to use a store that can contains data that are not Nns.

But your call.

frontend/src/lib/modals/accounts/SnsTransactionModal.svelte
lmuntaner: Yes, it makes sense.

frontend/src/lib/pages/NnsWallet.svelte
lmuntaner: Well, it's true that I could do a derived store that just returns the NNS accounts.

frontend/src/lib/services/accounts.services.ts
lmuntaner: used also the nns specific store

frontend/src/lib/pages/NnsWallet.svelte
lmuntaner: done

PR #1423, opened 2022-10-08
FIx: Import module not found in VS Code

lmuntaner:
# Motivation

There was an error in VS Code with test files that the module was not found.

# Changes

* Add tsconfig in `tests` folder.

# Tests

Not applicable.

### DIFF
--- a/frontend/src/tests/tsconfig.json
+++ b/frontend/src/tests/tsconfig.json
@@ -1,0 +1,3 @@
+{
+  "extends": "../../tsconfig.spec.json"
+}
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -11,6 +11,7 @@
     "strict": true
   },
   "exclude": [
+    "**/*.spec.ts",
     "jest.setup.ts",
     "**/*.mock.ts",
     "**/*Test.svelte",
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -11,9 +11,10 @@
     "strict": true
   },
   "exclude": [
-    "**/*.spec.ts",
     "jest.setup.ts",
     "**/*.mock.ts",
-    "**/*Test.svelte"
+    "**/*Test.svelte",
+    "svelte/**/*",
+    "public/**/*"
   ]
 }
--- a/frontend/tsconfig.json
+++ b/frontend/tsconfig.json
@@ -14,8 +14,6 @@
     "**/*.spec.ts",
     "jest.setup.ts",
     "**/*.mock.ts",
-    "**/*Test.svelte",
-    "svelte/**/*",
-    "public/**/*"
+    "**/*Test.svelte"
   ]
 }


frontend/tsconfig.json
```diff
@@ -11,9 +11,10 @@
     "strict": true
   },
   "exclude": [
-    "**/*.spec.ts",
```
lmuntaner, line 14: This one preventing VS Code from analizing properly the test files we have and resolve imports.

frontend/tsconfig.json
```diff
@@ -11,9 +11,10 @@
...
     "jest.setup.ts",
     "**/*.mock.ts",
-    "**/*Test.svelte"
+    "**/*Test.svelte",
+    "svelte/**/*",
```
lmuntaner, line 17: These two were causing an error when opening tsconfig with VSCode:

```
Cannot write file because it would overwrite input file TS
```

It was because VSCode was also trying to parse the output files.

lmuntaner:

peterpeterparker:
`npm run test` still works?

 `strict: false` is still needed for the tests.

peterpeterparker:
This one seems the VS code related issue https://github.com/microsoft/vscode/issues/12463

peterpeterparker:
@lmuntaner found the solution [there](https://github.com/microsoft/vscode/issues/12463#issuecomment-389733188)

we can create a new file `src/tests/tsconfig.json` that extends the spec config

```
{
  "extends": "../../tsconfig.spec.json"
}
```

VS code takes the closest `tsconfig.json`

Could you try and if it works for you as well, update the PR?

lmuntaner:
> `npm run test` still works?
> 
> `strict: false` is still needed for the tests therefore the two config file.
> 
> in addition, it's also useful to exclude test from the config to avoid resolving an import that would have been wrongly added to (`import something from test` in `src`).
> 
> What's the exact issue?

This is the issue:

Screenshot 2022-10-10 at 07 47 02

Because the test tiles are excluded from tsconfig, VSCode does not resolve the imports... Not sure why, but I don't get types nor autocomplete imports and it slows down a lot development.

Yes, it's a VSCode issue.

Yes, npm test works. The CI test pass and also locally.

Let me try putting a tsconfig file inside the test folder and not excluding the test files there.

frontend/tsconfig.json
lmuntaner: revered

frontend/src/tests/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
```
lmuntaner, line 1: @peterpeterparker adding the extra tsconfig file did the trick, thanks!

lmuntaner:
@peterpeterparker added another tsconfig file, it worked, thanks!

frontend/tsconfig.json
peterpeterparker: @lmuntaner can you also revert this change?

lmuntaner: why? Doesn't it make sense to ignore the output? Or isn't that the output?

peterpeterparker: because the first one path does not exists anyway (`svelte/**/*`) and the second one is not included in first place so these two patterns add nothing.

peterpeterparker (approved):
LGTM

frontend/tsconfig.json
lmuntaner: Done.

I had an error when I was opening tsconfig but I don't have it anymore, weird.

I had `svelte` folder:

Screenshot 2022-10-10 at 08 49 36

But maybe it was from the previous configuration. I removed it.

peterpeterparker: probably the old frontend app folder when we had `ts` and `svelte` which you did not delete locally

PR #1425, opened 2022-10-10
build: polyfill buffer for dev and prod window

peterpeterparker:
# Motivation

This PR fixes the Buffer polyfill for the `window` object for dev and prod build.

# Note

Buffer is still no correctly polyfied for the web worker in dev build but, we don't use it in that form yet.

# Changes

- replace plugin with manually injection of the Buffer


lmuntaner (approved):
LGTM, thanks!

PR #1429, opened 2022-10-10
Gix components import for Popover

wackyleo459:
# Changes
- Extraction and Import of Popover
- updated package-lock


peterpeterparker:
I did not noticed that `PopoverTest` and `Popover.spec` were not moved when we moved the popover to `gix-cmp`.

- Can you please delete these two from NNS-dapp?
- And provide a PR to add these tests to `gix-cmp`?

peterpeterparker:
I have updated today a new version of gix-cmp in main so you can safely resolve the conflict `frontend/package-lock.json` with main by accepting what's come from main. the popover will be included as well.

PR #1434, opened 2022-10-11
fix: sveltekit reproducibility

bitdivine:
# Motivation
The sveltekit upgrade broke reproducibility.

# Changes
- Save the frontend docker build files so that they are available for examination.
- Use the package version number instead of the timestamp in svelte builds.
- Order manifest keys; these were in a random order.

# Tests
Builds on these machines agree:
- CI
- max's laptop
- The "gix server" ch1-spm11.ch1.dfinity.network

To test reproducibility:

- Make sure that all your files are checked in and committed, otherwise the next step will wipe your changes.
- `git clean -dfx` (should be easy to remember - deletes everything that isn't tracked so be careful!!!!!)
- `./scripts/docker-build --no-cache` to build

When the build is complete you will have all the frontend files in `out/frontend` and a file with all the frontend file shas.  Save these off somewhere.

Repeat the docker build

Compare shas.

You can also compare builds across multiple machines.

We can automate this but it's slow; maybe we don't want it on every PR.  I mean, we almost certainly don't want it on every PR.

Note: The II team does such a test on merge; perhaps we can steal some of their code.


frontend/scripts/make-reproducible
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+cd "$(dirname "$(realpath "$0")")/.."
+# shellcheck disable=SC2094 # Thi sreads the entire file into memory and then writes it out, so is correct.
```
lmuntaner, line 4: Small typo: "Thi sreads" instead of "This reads"

lmuntaner (approved):
LGTM, just a typo in a comment. Already approving, thanks!

PR #1437, opened 2022-10-12
chore: Update local canister IDs

bitdivine:
# Motivation
dfx installs NNS canisters locally.  The IDs of local canisters have changed; this provides the updated values.

# Changes
- Update/add local canister IDs

# Tests


peterpeterparker (approved):
LGTM

Note: Aren't the local canister Id reproducible if created in the same sorting order?

PR #1438, opened 2022-10-12
style: overflow proposal user detail table

peterpeterparker:
# Motivation

Catch overflow when proposer uses table in the proposal detail.

# Screenshots

Reported by @nmattia

image

Overflow

Capture d’écran 2022-10-12 à 07 51 33


lmuntaner (approved):
Didn't we say to try to avoid the use of `:global`? But I guess there was no other option?

peterpeterparker:
Excactly, no other option. Markdown is injected in the DOM therefore it can only be styled with a global selector here.

PR #1446, opened 2022-10-13
Chore: Comment on chrono version

lmuntaner:
# Motivation

When running cargo audit, it shows a vulnerability in a dependency.

But we can't update it because newer versions are not compatible with IC.

# Changes

* Add a comment on top of the chrono version to remember why we can't fix the vulnerability.


mstrasinskis (approved):
Rubber stamp. Thank you!

PR #1447, opened 2022-10-13
build: fix buffer polyfill for mainnet

peterpeterparker:
# Motivation

Same issue as https://github.com/vitejs/vite/discussions/2785#discussioncomment-1452855

Basicall the ledger needs Buffer which is a NodeJS library which we have to polyfill.

Adding a polyfill (`window.Buffer =...`) works great for local development but does not always work on mainnet.
Indeed, `_layout.js` (which contains the polyfill) might be loaded after a `page.js` therefore, we might hit the issue `start-4e5e2afa.js:1 ReferenceError: Buffer is not defined`.

# Solution

Use rollup plugin as well to inject Buffer (as we used too with rollup bundler) so that the Buffer also gets polyfied within the page

# Potential of improvement

With this PR we polyfill in to places, therefore the bundle will be a bit more heavy. We can probably spare the dev polyfill for prod build.

# Changes

- inject Buffer polyfill with Rollup for prod build

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -22,6 +22,7 @@
         "@ledgerhq/hw-transport-node-hid-noevents": "^6.27.6",
         "@ledgerhq/hw-transport-webhid": "^6.27.6",
         "@zondax/ledger-icp": "^0.6.0",
+        "buffer": "^6.0.3",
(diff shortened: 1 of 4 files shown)


lmuntaner (approved):
Thanks!

PR #1451, opened 2022-10-13
build: fix SvelteKit CSP 'stric-dynamic' issue in Firefox

peterpeterparker:
# Motivation

Using a CSP `'strict-dynamic'` with SvelteKit breaks in Firefox.

# Solution

Extracting the script content that gets injected by SvelteKit in `index.html` into a separate JS script that gets appended with a script loader.

Exact same "script loader" as the one we use to use for Rollup.

# Issue

- SvelteKit https://github.com/sveltejs/kit/issues/3558 

# Changes

- update `build.csp.mjs` to extract script
- add script loader to `app.html`

# Screenshot

Capture d’écran 2022-10-13 à 17 50 46

### DIFF
--- a/frontend/scripts/build.csp.mjs
+++ b/frontend/scripts/build.csp.mjs
@@ -32,9 +32,7 @@ const extractStartScript = () => {
     /(<script type=\"module\" data-sveltekit-hydrate[\s\S]*?>)([\s\S]*?)(<\/script>)/gm;
 
   // 1. extract SvelteKit start script to a separate main.js file
-  const m = svelteKitStartScript.exec(indexHtml);
-  const content = m[2];
-
+  const [_script, _scriptStartTag, content] = svelteKitStartScript.exec(indexHtml);
   const inlineScript = content.replace(/^\s*/gm, "");
 
   writeFileSync(
--- a/frontend/scripts/build.csp.mjs
+++ b/frontend/scripts/build.csp.mjs
@@ -32,7 +32,7 @@ const extractStartScript = () => {
     /(<script type=\"module\" data-sveltekit-hydrate[\s\S]*?>)([\s\S]*?)(<\/script>)/gm;
 
   // 1. extract SvelteKit start script to a separate main.js file
-  const [_script, _scriptStartTag, content] = svelteKitStartScript.exec(indexHtml);
+  const [_script, _scriptStartTag, content, _scriptEndTag] = svelteKitStartScript.exec(indexHtml);
   const inlineScript = content.replace(/^\s*/gm, "");
 
   writeFileSync(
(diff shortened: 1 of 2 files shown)


frontend/scripts/build.csp.mjs
```diff
@@ -8,8 +8,43 @@ import { join } from "path";
...
+  const indexHtml = readFileSync(
+    join(process.cwd(), "public", "index.html"),
+    "utf-8"
+  );
+
+  const svelteKitStartScript =
+    /(<script type=\"module\" data-sveltekit-hydrate[\s\S]*?>)([\s\S]*?)(<\/script>)/gm;
+
+  // 1. extract SvelteKit start script to a separate main.js file
+  const m = svelteKitStartScript.exec(indexHtml);
```
lmuntaner, line 35: What does `m` stand for? Maybe `result`?

frontend/scripts/build.csp.mjs
```diff
@@ -8,8 +8,43 @@ import { join } from "path";
...
+  const content = m[2];
```
lmuntaner, line 36: Why the 3rd?

lmuntaner: From MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#return_value

```
The returned array has the matched text as the first item, and then one item for each capturing group of the matched text.
```

Why do we want the third?

frontend/src/app.html
```diff
@@ -107,6 +107,13 @@
       }
     </script>
 
+    <script>
+      const loader = document.createElement("script");
```
lmuntaner, line 111: Maybe add also a comment on why we need this?

lmuntaner (approved):
I few comments to better explain what's going on and to avoid us to remove some of these in the future.

But already approving.

frontend/src/app.html
peterpeterparker: If I had a comment then it lands on the web - i.e. everyone has to fetch the comment everytime, so rather like not to add a comment within the HTML page

On the contrary, it's already documented in the CSP builder.

> - script-src and usage of 'integrity':
Ideally we would like to secure the scripts that are loaded with the 'integrity=sha256-...' hashes attributes - e.g. https://stackoverflow.com/a/68492689/5404186.
However, this is currently only supported by Chrome. Firefox issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1409200
To overcome this, we include within the index.html a first script which, when executed at app boot time, add a script that actually loads the main.js.
We generate the hash for that particular first script and set 'strict-dynamic' to trust those scripts that will be loaded per extension - the chunks used by the app.

PR #1452, opened 2022-10-14
Chore: Fix e2e running locally

lmuntaner:
# Motivation

Improve documentation of how to execute e2e tests.

Fix selectors and other minor broken e2e tests.

# Important

This does NOT fix all the e2e yet. But it's one step towards adding them to our CI.

What is still pending:

* Deploy the SNS-W canister locally. Better to wait until the next dfx release which will include an easy way to do so: `dfx nns install`.
* When making a new proposal to be tested, it should be Open and not fail immediately.
* I also added a couple more of TODOs in the code that would add nice functionality, but are not necessary for the current e2e tests.

# Changes

* Improve Readme in e2e-tests directory.
* Fix some broken e2e tests.
* Add some data-tid in nns-dapp frontend.

# Tests

All changes are in tests.

### DIFF
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -86,7 +86,7 @@ DFX_NETWORK=local ENV_OUTPUT_FILE=./e2e-tests/.env ./config.sh
 
 By default, tests will be executed (or tried to) in Firefox and Chrome.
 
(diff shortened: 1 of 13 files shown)


peterpeterparker (approved):
Rubber stamp (great initiative, thanks!)

PR #1455, opened 2022-10-17
feat: disable stake maturity button if maturity zero or less

peterpeterparker:
# Motivation

>  But I agree if maturity =0 then we can disable the button. - Product

# Changes

- disable stake maturity button is maturity is zero or less


lmuntaner (approved):
LGTM

PR #1457, opened 2022-10-17
fix: test utils should not be bundled in prod build

peterpeterparker:
# Motivation

Test utility should not be bundled in production build.

# Changes

- remove file not excluded by config and move utils in excluded file


frontend/src/tests/mocks/utils.mock.ts
```diff
@@ -1,5 +1,17 @@
+import { fireEvent } from "@testing-library/dom";
+
 export const mockWaiting = (seconds: number, value?: unknown) =>
   new Promise((resolve) => setTimeout(() => resolve(value), seconds * 1000));
 
 export const silentConsoleErrors = () =>
   jest.spyOn(console, "error").mockImplementation(jest.fn);
+
+export const clickByTestId = async (
```
lmuntaner, line 9: This doesn't look like a mock util to me. Isn't there another way to not bundle it? Does that mean we can't use helpers in utils unless they are in the mocks directory?

lmuntaner: How about using `click-by-test-id.test-utils.ts` and adding `"**/*.test-utils.ts",` to the exclude in `tsconfig.ts`?

peterpeterparker: As long as it is not bundled. Done. `utils/utils.test-utils.ts` + exclusion

frontend/src/tests/utils/utils.test-utils.ts
```diff
@@ -1,5 +1,11 @@
 import { fireEvent } from "@testing-library/dom";
 
+export const mockWaiting = (seconds: number, value?: unknown) =>
```
lmuntaner, line 3: I meant only `clickByTestId`, but this is also ok for me.

lmuntaner (approved):

frontend/src/tests/utils/utils.test-utils.ts
peterpeterparker: ok

PR #1468, opened 2022-10-20
Chore: Reduce token amount usage

lmuntaner:
# Motivation

Reduce usage of TokenAmount for non-UI related functionality and use e8s instead.

# Changes

* Remove dependency of TokenAmount in "neuron.utils", "governance.api" and "neuron.services".

# Tests

* Fix tests after refactor.

### DIFF
--- a/frontend/src/lib/stores/transaction-fees.store.ts
+++ b/frontend/src/lib/stores/transaction-fees.store.ts
@@ -68,3 +68,8 @@ export const mainTransactionFeeStore = derived(
   transactionsFeesStore,
   ($store) => Number($store.main)
 );
+
+export const mainTransactionFeeE8sStore = derived(
+  transactionsFeesStore,
+  ($store) => $store.main
+);
--- a/frontend/src/lib/stores/transaction-fees.store.ts
+++ b/frontend/src/lib/stores/transaction-fees.store.ts
@@ -64,6 +64,9 @@ const initTransactionFeesStore = () => {
 
 export const transactionsFeesStore = initTransactionFeesStore();
 
+/**
+ * @deprecated prefer mainTransactionFeeE8sStore to use e8s for amount of tokens instead of Number.
+ */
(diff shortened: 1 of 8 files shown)


lmuntaner:
@mstrasinskis @peterpeterparker as agreed in the[ FE refactors docs](https://www.notion.so/FE-Dapp-Refactors-36a53e55f52f4782bb2c39fcbf2622ab). I started one of the refactors.

Please review

frontend/src/lib/stores/transaction-fees.store.ts
```diff
@@ -68,3 +68,8 @@ export const mainTransactionFeeStore = derived(
   transactionsFeesStore,
   ($store) => Number($store.main)
 );
+
+export const mainTransactionFeeE8sStore = derived(
```
peterpeterparker, line 72: should we rename `mainTransactionFeeStore` in `mainTransactionFeeNumberStore` to prevent confusion?

peterpeterparker (approved):
LGTM, thanks

Not related but your PR make me think that I should try to check if the n notation would not be now supported now because I bumped the es target. e.g. `0n` instead of `BigInt(0)`. I should give it a try.

frontend/src/lib/stores/transaction-fees.store.ts
lmuntaner: What do you think if I set it as deprecated to force using the e8s? I believe we should avoid using numbers for amounts. At least as soon as they get out of the input element.

peterpeterparker: Sounds like a good idea to me, like it 👍

lmuntaner:
> LGTM, thanks
> 
> Not related but your PR make me think that I should try to check if the n notation would not be now supported now because I bumped the es target. e.g. `0n` instead of `BigInt(0)`. I should give it a try.

Sounds cool!

PR #1471, opened 2022-10-21
GIX-1092: Refresh SNS sale before participation

lmuntaner:
# Motivation

User might try to participate in a decentralization sale and fail because the information is stale. We only load it once at the initialization.

To avoid edge cases and bad UX where refunds need to happen, we refresh the swap state before validating the user participation.

# Changes

* New sns service "reloadSnsState" used only internally in "participateInSwap".
* New method in sns store "updateSwapState" that loads in store only the swap state.

# Tests

* A test case that participateInSwap fetches new data.
* A test case in sns store for updateSwapState.

### DIFF
--- a/frontend/src/lib/services/sns.services.ts
+++ b/frontend/src/lib/services/sns.services.ts
@@ -220,6 +220,34 @@ export const listSnsProposals = async (): Promise<void> => {
   });
 };
 
+/**
+ * Requests swap state and loads it in the store.
+ * Ignores possible undefined. This is used only to recheck the data with up-to-date information.
+ * If an error happens, we want to rely on the data that it's already in the store.
+ *
+ * @param {Principal} rootCanisterId Root canister id of the project.
+ */
+const reloadSnsState = async (rootCanisterId: Principal): Promise<void> => {
+  try {
+    const identity = await getIdentity();
+    const swapData = await querySnsSwapState({
+      rootCanisterId: rootCanisterId.toText(),
+      identity,
+      certified: true,
+    });
+    // Ignore if swap data is undefined
+    if (swapData !== undefined) {
+      snsQueryStore.updateSwapState({
+        swapData,
+        rootCanisterId: rootCanisterId.toText(),
+      });
+    }
+  } catch (err) {
+    // Ignore error
+    console.error("Error reloading swap state", err);
+  }
+};
+
 export const routePathRootCanisterId = (path: string): string | undefined => {
   if (!isRoutePath({ paths: [AppPath.ProjectDetail], routePath: path })) {
     return undefined;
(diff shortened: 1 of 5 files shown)


lmuntaner:

This was brought up by @LaraAS and @DanielThurau during their testing phase.

We are not yet handling all the edge cases, but at least we reduce the likelihood of encountering them.

frontend/src/tests/lib/stores/sns.store.spec.ts
```diff
@@ -159,32 +160,75 @@ describe("sns.store", () => {
...
+        swapData: updatedSwapData,
+        rootCanisterId: rootCanisterId,
+      });
+
+      const updatedStore = get(snsQueryStore);
+      expect(
+        updatedStore?.swaps.find(
+          (swap) => swap.rootCanisterId === rootCanisterId
+        )?.swap[0]?.lifecycle
+      ).toBe(SnsSwapLifecycle.Open);
```
mstrasinskis, line 231: nitpick: probably comparing to `updatedSwapData` would be simpler. e.g.
```ts
...?.swap[0]).toEqual(updatedSwapData);
```

mstrasinskis (approved):
LGTM, thanks! Just a couple of minor comments.

frontend/src/lib/services/sns.services.ts
```diff
@@ -220,6 +220,34 @@ export const listSnsProposals = async (): Promise<void> => {
   });
 };
 
+/**
+ * Requests swap state and loads it in the store.
+ * Ignores possible undefined. This is used only to recheck the data with up-to-date information.
+ * If an error happens, we want to rely on the data that it's already in the store.
```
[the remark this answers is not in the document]
lmuntaner: Makes sense

lmuntaner:
@peterpeterparker I don't think we can do this easily in the catch.

Even if we called error_refund, we should still be able to show an error or success notification properly. To do that, we would need to check whether the error is from stale information or just because of a second call. We already agreed that increases the complexity.

Anyway, I will add a TODO because I believe this should be improved. I'll talk with the NNS Team about this case specifically and how we can easily come up with a better solution.

frontend/src/lib/services/sns.services.ts
```diff
@@ -220,6 +220,35 @@ export const listSnsProposals = async (): Promise<void> => {
...
+const reloadSnsState = async (rootCanisterId: Principal): Promise<void> => {
+  try {
+    const identity = await getIdentity();
+    const swapData = await querySnsSwapState({
+      rootCanisterId: rootCanisterId.toText(),
+      identity,
+      certified: true,
+    });
+    // Ignore if swap data is undefined
+    if (swapData !== undefined) {
```
peterpeterparker, line 240: is this correct? above call is certified so if NNS-dapp gets undefined it is indeed undefined

frontend/src/tests/lib/stores/sns.store.spec.ts
lmuntaner: I thought this would be clearer of the change, but happy to change.

frontend/src/lib/services/sns.services.ts
lmuntaner: This was the case because the api function had also `undefined` in the return value but it was not really returning undefined never. I changed the return type of the api function and this is not needed anymore.

lmuntaner:

peterpeterparker (approved):
LGTM, thx

PR #1473, opened 2022-10-24
Payload update sample howto

mstrasinskis:
# Motivation

Extend the HOWTO with `Payload Definition Search Sample Steps` using `UpdateSnsWasmSnsSubnetIds` as a sample.


lmuntaner (approved):
Way better, thanks!

PR #1475, opened 2022-10-24
Support testnets that make use of the fixed SNS-WASM canister deployed via icos_deploy.sh

DanielThurau:
# Motivation

The SNS-WASM canister deployed via icos_deploy.sh (which use ic-nns-init) should now be used instead of a custom SNS-WASM canister. This PR includes two new propose-to-scripts that will help set up the SNS-WASM canister to work with a newly deployed testnet. 
# Changes

- The `propose-to-add-wasm-to-sns-wasm` will use the test-neuron-proposer to submit NNS Governance proposals to add the various downloaded SNS canister WASMs. These should automatically pass if no other neurons are created in the testnet.
- The `propose-to-update-sns-wasm-sns-subnet-ids` will use the test-neuron-proposer to submit an NNS Governance proposal to update the SNS-WAMS canister to be able to deploy to the newly created testnet subnets.
- The Dockerfile has been updated to support the new location of the `icp_ledger/ledger.did` which was just moved in the monorepo.
- dfx.json now points to the fixed canister id of the SNS-WASM canister (`qaa6y-5yaaa-aaaaa-aaafa-cai`). Now when deploying nns-dapp via the deploy script, the dapp will be configured to point at this canister instead of the custom one.

# Tests

- No tests exist for deploy.sh but I was able to deploy a testnet and test the SNS flow end-to-end.

# Extra

- The `nns-dapp/testnets` branch will need an equivalent update to use these new scripts. I will push the changes to the `deploy-testnet` script as an example, and future deploy scripts can imitate the usage.
- The SNS-WASM canister also needs to have a whitelisted principal id. However, this cannot be done in the deployment scripts as the wallet canister id of the user needs to be known. An example of how to do this is included below:

```
WALLET_CANISTER="$(dfx identity --network "$DFX_NETWORK" get-wallet)"
NNS_URL="$(./bin/dfx-network-subnet-dashboard --dfx-network "$DFX_NETWORK")"
ic-admin --nns-url "$NNS_URL" propose-to-update-sns-deploy-whitelist --test-neuron-proposer --added-principals "$WALLET_CANISTER"
```


lmuntaner (approved):
Thanks!

bitdivine (approved):
Thank you.  Is it worth commenting that this is ident-1?  `dfx nns install` now also provides ident-1 with a billion ICP.

DanielThurau:
@lmuntaner @bitdivine I realized that the solution wouldn't scale as you need the wallet canister id not a principal. In the meantime we also needed to start supporting the fixed canister id of the SNS-WASM canister. Please take another look at the PR.

scripts/propose-to/propose-to-add-wasm-to-sns-wasm
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
lmuntaner, line 1: When is this file used?

Is it a convenience script to create a proposal to setup an SNS?

scripts/propose-to/propose-to-update-sns-wasm-sns-subnet-ids
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Propose to populate the SNS-WASM canister with a subnet it can deploy to.
+# This requires testnet NNS Governance to be setup with the test-neuron-proposer.
+# This ic-admin command targets the fixed canister id of the SNS-WASM canister (qaa6y-5yaaa-aaaaa-aaafa-cai).
+
+SNS_SUBNET="$(ic-admin --nns-url "$NNS_URL" get-subnet-list | jq -r '.[0]')"
```
lmuntaner, line 8: When is this needed?

scripts/propose-to/propose-to-add-wasm-to-sns-wasm
DanielThurau: It's used in the testnet deployment scripts (check out an example here: https://github.com/dfinity/nns-dapp/blob/testnets/testnets/deploy-testnet#L115-L135). 

It's a script that will submit an NNS proposal to add SNS canister versions to the SNS-WASM canister that is controlled by NNS Root. In the previous deploy scripts we would use a custom SNS-WASM canister that allowed us to upload the versions without the NNS proposal. We needed to move away from that to test SNS upgrades and these scripts help.

scripts/propose-to/propose-to-update-sns-wasm-sns-subnet-ids
DanielThurau: It's sed in the testnet deployment scripts (checkout out an example here: https://github.com/dfinity/nns-dapp/blob/testnets/testnets/deploy-testnet#L115-L135).

It's a script that will submit an NNS proposal to add the testnet's subnet ids to the SNS-WASM canister so that SNSes can be deployed. In the previous deploy scripts we would use a custom SNS-WASM canister that allowed us to whitelist subnets without an NNS proposal. We need to move away from testing on the custom deployed SNS-WASM canister so that we can do full E2E tests.

scripts/propose-to/propose-to-add-wasm-to-sns-wasm
lmuntaner: Thanks for the explanation!

scripts/propose-to/propose-to-update-sns-wasm-sns-subnet-ids
lmuntaner: I can see that I was missing the testnet context, thanks!

PR #1476, opened 2022-10-25
GIX-1089: Hide SNS neurons no stake

lmuntaner:
# Motivation

The SNS Neurons list showed neurons with 0 stake after they were disbursed.

User should not see the neurons that have already been disbursed.

# Changes

* New sns neuron util `hasValidStake` that returns boolean.
* Use the new sns neuron util in "sortedSnsNeuronStore".
* Do not filter out the NNS neurons without valid stake in the neuronsStore.
* Filter out NNS neurons without valid stake in definedNeuronsStore.

# Tests

* New test for sns neuron util "hasValidStake".
* New test case for sortedSnsNeuronStore
* Tests for definedNeuronsStore.

### DIFF
--- a/frontend/src/tests/lib/derived/sorted-sns-neurons.derived.spec.ts
+++ b/frontend/src/tests/lib/derived/sorted-sns-neurons.derived.spec.ts
@@ -59,6 +59,44 @@ describe("sortedSnsNeuronStore", () => {
     ]);
   });
 
+  it("should filter out neurons with no stake nor maturity", async () => {
+    const neurons: SnsNeuron[] = [
+      {
+        ...createMockSnsNeuron({
+          stake: BigInt(0),
+          id: [1, 5, 3, 9, 1, 1, 1],
+        }),
+        created_timestamp_seconds: BigInt(1),
+        maturity_e8s_equivalent: BigInt(0),
+      },
+      {
+        ...createMockSnsNeuron({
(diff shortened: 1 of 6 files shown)


lmuntaner:
@mstrasinskis please review

frontend/src/tests/lib/derived/sorted-sns-neurons.derived.spec.ts
```diff
@@ -59,6 +59,44 @@ describe("sortedSnsNeuronStore", () => {
...
+    snsNeuronsStore.setNeurons({
+      rootCanisterId: mockPrincipal,
+      neurons,
+      certified: true,
+    });
+    routeStore.update({
+      path: `${CONTEXT_PATH}/${mockPrincipal.toText()}/neurons`,
+    });
+
+    await tick();
```
mstrasinskis, line 96: nitpick: `await waitFor(() => expect...` instead of `tick`

mstrasinskis (approved):
LGTM, thank you!

frontend/src/tests/lib/derived/sorted-sns-neurons.derived.spec.ts
lmuntaner: I thought we could use that only when we render a component because `waitFor` is imported from `@testing-library/svelte`. But I tried and it works, I'll change it, thanks!

PR #1479, opened 2022-10-26
GIX-1085: Project does not keep context

lmuntaner:
# Motivation

When the user is in the project details and clicked in Tokens or Neurons menu item was redirected to the Neurons or Tokens of that project.

That was not correct because we don't consider the project details to be part of an SNS universe, rather to be part of the NNS universe

# Changes

* MenuItems in Tokens and Neurons redirect to NNS project if they are in the project details instead of relying in the context.
* Add projectDetail to paths util

# Tests

* Add tests that menu items keep this exception.

### DIFF
--- a/frontend/src/lib/components/common/MenuItems.svelte
+++ b/frontend/src/lib/components/common/MenuItems.svelte
@@ -10,7 +10,7 @@
   import type { SvelteComponent } from "svelte";
   import { i18n } from "$lib/stores/i18n";
   import { baseHref } from "$lib/utils/route.utils";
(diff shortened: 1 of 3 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/197951547-a81f808b-882b-405e-b04b-9c20c613a868.mov

lmuntaner:

This was a bug found during the demo. It will be fixed with the SvelteKit routing but in case that's not ready by then, we should fix this.

peterpeterparker (approved):
LGTM, thx

PR #1480, opened 2022-10-26
Chore: Validate amount participation inline

lmuntaner:
# Motivation

Improve the UX by inlining the error message in sale participation.

# Changes

* New prop in TransactionModal "validateAmount".
* Use new prop validateAmount in "ParticipateSwapModal" with helper validParticipation.
* Improve error message when commitment is too large.

# Tests

* Test new prop in TransactionModal specs.

### DIFF
--- a/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
+++ b/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
@@ -30,6 +30,9 @@
   export let maxAmount: bigint | undefined = undefined;
   export let skipHardwareWallets = false;
   export let showManualAddress = true;
+  export let validateAmount: (
+    amount: number | undefined
+  ) => string | undefined = () => undefined;
 
   let filterDestinationAccounts: (account: Account) => boolean;
   $: filterDestinationAccounts = (account: Account) => {
@@ -68,7 +71,7 @@
         account: selectedAccount,
         amountE8s: tokens.toE8s() + transactionFee.toE8s(),
       });
-      errorMessage = undefined;
+      errorMessage = validateAmount(amount);
     } catch (error: unknown) {
       if (error instanceof InvalidAmountError) {
         errorMessage = $i18n.error.amount_not_valid;
--- a/frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
+++ b/frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
@@ -10,6 +10,7 @@
   import {
     currentUserMaxCommitment,
     hasUserParticipatedToSwap,
+    validParticipation,
   } from "$lib/utils/projects.utils";
   import type { SnsSummary, SnsSwapCommitment } from "$lib/types/sns";
   import TransactionModal from "$lib/modals/accounts/NewTransaction/TransactionModal.svelte";
@@ -25,6 +26,7 @@
   import AdditionalInfoReview from "./AdditionalInfoReview.svelte";
   import { OWN_CANISTER_ID } from "$lib/constants/canister-ids.constants";
   import type { WizardStep } from "@dfinity/gix-components";
+  import { replacePlaceholders, translate } from "$lib/utils/i18n.utils";
 
   const { store: projectDetailStore, reload } =
     getContext<ProjectDetailContext>(PROJECT_DETAIL_CONTEXT_KEY);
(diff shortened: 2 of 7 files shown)


lmuntaner:

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
```diff
@@ -106,6 +108,32 @@
       stopBusy("project-participate");
     }
   };
+
+  // Used for form inline validation
+  let validateAmount: (amount: number | undefined) => string | undefined;
+  $: validateAmount = (amount: number | undefined) => {
```
peterpeterparker, line 114: not sure I understand why it has to be initialized in auto subscribe?

peterpeterparker: side not, since there is a bit of logic, would it be worth having a utils and dedicated test?

can live with it in the cmp, just something that went through my mind

lmuntaner: I used the auto-subscribe in case the summary, or swap are refetched and the validation needs to also be updated.

I didn't add a test because the logic is related to this component and UI.

One part of the logic is because `validParticipation` returns the label key. Yet, the `validateAmount` should return the translated  text, not the key. I can't change that in `validateAmount` because it's also used in the service and the key used in the toast error.

The other part is the type safety checking. Which is required mostly for this component.

Yet, I agree that the `validateAmount` is ugly and hard to read. I will create a helper function inside the component to make it more readable.

lmuntaner: And we have tests for `validParticipation` and for TransactionForm with `validateAmount`.

But I can also add a test in `ParticipateSwapModal`

lmuntaner: I added a test in `ParticipateSwapModal` and simplified the function by changing the type safety. For example the amount didn't need to account for `undefined`. And a swap of `null` is the same as `undefined`.

lmuntaner:

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
peterpeterparker: With the view of the PR I did not knew `swapCommitment` was subscrable therefore did not understood why. Now it's clear. Thanks for the explanation.

peterpeterparker (approved):
LGTM, thx

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
```diff
@@ -106,6 +108,32 @@
...
+        project: {
+          rootCanisterId: summary.rootCanisterId,
+          summary,
+          swapCommitment,
+        },
+        amount: TokenAmount.fromNumber({ amount, token: ICPToken }),
+      });
+      // `validParticipation` does not return `valid` as `false` without a labelKey.
+      // But we need to check because of type safety.
+      return valid || labelKey === undefined
```
mstrasinskis, line 130: Not sure about swallowing an error in case of valid=false and no labelKey. Probably showing some general error would improve this. Not a strong opinion.

mstrasinskis (commented):
Probably later we need something more comprehensive (e.g. multiple error messages support). What do you think about having generic function in our types to simplify future refactoring? So in the `TransactionForm` it could be used like:
`export let validateAmount: ValidationFunction<number>;`

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
```diff
@@ -30,6 +30,9 @@
   export let maxAmount: bigint | undefined = undefined;
   export let skipHardwareWallets = false;
   export let showManualAddress = true;
+  export let validateAmount: (
+    amount: number | undefined
+  ) => string | undefined = () => undefined;
```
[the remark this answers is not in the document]
lmuntaner: Maybe we need a better return value.

* `undefined` means the amount is correct
* string means the error message.

Maybe:

```
interface ValidationResult {
  valid: boolean;
  labelKey: string;
  substitutions?: Record<string, string>;
}
```

frontend/src/lib/modals/sns/SwapModal/ParticipateSwapModal.svelte
lmuntaner: That's a good point.

Also a variant maybe?

```
type ValidationResult =
  { Ok: null } |
  { Err: { labelKey: string, substitutions: ... } };
```

PR #1482, opened 2022-10-27
fix: Reenable patch

bitdivine:
# Motivation
The patch with functionality such as fast maturity wass disabled when the patch got out of sync and so could not be applied.  That was fixed but the patch was not re-enabled.

# Changes
- Re-enable the patch

# Tests
- dfx has been deployed to a testnet and fast maturity works there.


lmuntaner (approved):
Thanks!

PR #1486, opened 2022-10-27
feat: Disable whitelist

bitdivine:
# Motivation
There is a temporary whitelist in place, preventing huge numbers of users from creating an SNS before we are confident that it works well in production.  This whitelist isn't really relevant for the nns-dapp development and it is an additional hurdle, so it seems sensible to disable it.

# Changes
* Disable the SNS creators whitelist in the nns-sns-wasm canister.

# Tests
* Small09 is deployed with this patch.


lmuntaner (approved):
We love patches 😍

PR #1487, opened 2022-10-28
build: remove empty unused default csp tag

peterpeterparker:
# Motivation

SvelteKit injects an empty (since we do not configure it) CSP tag that is ignored because it finds place after our CSP policy. Still cleaner to remove it.

# Changes

- remove tag before calculting script


lmuntaner (approved):
LGTM

PR #1489, opened 2022-10-28
GIX-1093: Get Transactions uses sns wrapper

lmuntaner:
# Motivation

Use sns-wrapper to interact with Index canister instead of the canister directly.

# Changes

* Remove rootCanisterId temporary checks in transactions services and api.
* Use wrapper to call getTransactions in sns-index api.

# Tests

* Add test for getTransactions api function.


peterpeterparker (approved):
LGTM, thx

PR #1490, opened 2022-10-28
feat: split project list into open and committed

mstrasinskis:
# Motivation

Display projects in two sections: "Current Launches" and "Launched".

https://dfinity.atlassian.net/browse/GIX-1102

# Changes

- expose filterProjectsStatus
- extend `Projects` component with `status: "Committed" | "Open"`
- use static skeleton count (3 as best for desktop). Because knowing projects count is not enough.
- remove `snsesCountStore`

# Tests

- updated Launchpad and Projects tests to support status filtering
- filterProjectsStatus

# Screenshots

image

Screenshot 2022-10-28 at 13 39 19

Screenshot 2022-10-28 at 14 09 38


frontend/src/lib/components/launchpad/Projects.svelte
```diff
@@ -2,49 +2,54 @@
...
   import {
     activePadProjectsStore,
     type SnsFullProject,
   } from "$lib/stores/projects.store";
+  import { SnsSwapLifecycle } from "@dfinity/sns";
+  import { filterProjectsStatus } from "$lib/utils/projects.utils";
 
-  let projects: SnsFullProject[] | undefined;
-  $: projects = $activePadProjectsStore;
+  export let status: "Committed" | "Open";
```
lmuntaner, line 17: Why not use the type `SnsSwapLifecycle`?

frontend/src/tests/lib/utils/projects.utils.spec.ts
```diff
@@ -30,15 +31,48 @@ import {
...
+            },
+            {
+              ...mockSnsFullProject,
+              summary: summaryForLifecycle(SnsSwapLifecycle.Committed),
+            },
+          ],
+          swapLifecycle: SnsSwapLifecycle.Pending,
+        })
+      ).toEqual([]);
+    }),
```
lmuntaner, line 66: is this `,` correct? Shouldn't it be a `;`?

frontend/src/lib/components/launchpad/Projects.svelte
mstrasinskis: Because only two types are supported. But why not.

frontend/src/tests/lib/utils/projects.utils.spec.ts
mstrasinskis: Good catch!

lmuntaner (approved):

PR #1492, opened 2022-10-31
build: remove https: from CSP script-src

peterpeterparker:
# Motivation

We use the `https:` in the `script-src` rules of the CSP for backwards compatibility as displayed by [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src). However we have a report from a user who cannot access NNS-dapp on his Android v10 device. We are not sure what his issue is but, using Android Simulator, we can reproduce a blank screen on OS <= Android v12. That's why, in accordance with Security team, we remove the `https:` rule that solves the issue in the simulator and which is in any case not used in modern browsers that support `strict-dynamic`.

# Changes

- remove `https:` rules from CSP `script-src`

### DIFF
--- a/frontend/scripts/build.csp.mjs
+++ b/frontend/scripts/build.csp.mjs
@@ -75,7 +75,7 @@ const extractStartScript = () => {
  *
  * - script-src and 'strict-dynamic':
  * Chrome 40+ / Firefox 31+ / Safari 15.4+ / Edge 15+ supports 'strict-dynamic'.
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):

peterpeterparker:
Deployed a test version https://qvhpv-4qaaa-aaaaa-aaagq-cai.small11.dfinity.network/ and asked the user to give it a try

peterpeterparker:
User that reported the issue confirmed that the test version which remove `https:` works on his Android v10.

PR #1493, opened 2022-10-31
feat: SnsNeuronMaturityCard

mstrasinskis:
# Motivation

Display sns neuron maturity card.
https://dfinity.atlassian.net/browse/GIX-1094

# Changes

- SnsNeuronMaturityCard component
- formattedSnsMaturity util function

# Tests

- SnsNeuronMaturityCard
- formattedSnsMaturity

# Screens
Screenshot 2022-11-02 at 15 26 59

Screenshot 2022-11-02 at 15 27 17


frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -223,3 +224,12 @@ export const isSnsNeuron = (
  */
 export const hasValidStake = (neuron: SnsNeuron): boolean =>
   neuron.cached_neuron_stake_e8s + neuron.maturity_e8s_equivalent > BigInt(0);
+
+/**
+ * Format the maturity in a value (token "currency") way.
+ * @param {SnsNeuron} neuron The neuron that contains the `maturityE8sEquivalent` formatted
+ */
+export const formattedSnsMaturity = ({ maturity_e8s_equivalent: value }: SnsNeuron): string =>
```
lmuntaner, line 232: Is `maturity_e8s_equivalent` optional??

frontend/src/lib/components/neuron-detail/SnsNeuronMaturityCard.svelte
```diff
@@ -0,0 +1,38 @@
...
+
+  let token: string;
+  $: token = $isNnsProjectStore ? ICPToken.symbol : $snsTokenSymbolSelectedStore?.symbol ?? $i18n.core.token;
+</script>
+
+<CardInfo>
+  <KeyValuePairInfo testId="maturity">
+    <h3 slot="key">{ $i18n.neuron_detail.maturity_title}</h3>
+    <svelte:fragment slot="info"
+      >{replacePlaceholders($i18n.neuron_detail.stake_maturity_tooltip, {$token: token})}</svelte:fragment
```
lmuntaner, line 28: Not sure we can add this yet. Staking maturity is not yet implemented in SNS.

Let's skip the tooltip for now.

frontend/src/lib/components/neuron-detail/SnsNeuronMaturityCard.svelte
```diff
@@ -0,0 +1,38 @@
...
+  import {getContext} from "svelte";
+
+  const { store }: SelectedSnsNeuronContext =
+          getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
+
+  let neuron: SnsNeuron | undefined | null;
+  $: neuron = $store.neuron;
+
+  let token: string;
+  $: token = $isNnsProjectStore ? ICPToken.symbol : $snsTokenSymbolSelectedStore?.symbol ?? $i18n.core.token;
```
lmuntaner, line 21: I would prefer the token to be passed as prop.

Relying in the global selected store for child components has not been a good idea so far. In the short time we have SNS I have already fixed a few bugs relating to the global sns store being access directly.

mstrasinskis: I would prefer to avoid  having two data sources (the neuron from the context store and the token from the props). How about having neuron and the token in the context store? Or in props. Both works for me.

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: Good catch!

frontend/src/lib/components/neuron-detail/SnsNeuronMaturityCard.svelte
lmuntaner: Also works to put it in the context store, makes sense!

mstrasinskis: After removing the hint message there is no need to have the token.

lmuntaner (approved):
LGTM!

PR #1498, opened 2022-11-01
GIX-1026: Manage errors in SNS Accounts pages gracefully

lmuntaner:
# Motivation

User should get feedback when something doesn't work well.
User should not get a blank page when something goes bad.

# Changes

* Try/catch when loading transactions and show a toast error in the catch.
* Make transactionFee prop as mandatory in TransactionModal.
* Show Modal with Spinner in SnsTransactionModal while transaction fee is not loaded. The fee is loaded along the accounts; therefore, it should be there by the time the user clicks to make a transaction.
* Show a toast error when there is an error loading the sns transaction fee.

# Tests

* Fix broken tests.

### DIFF
--- a/frontend/src/lib/services/sns-transactions.services.ts
+++ b/frontend/src/lib/services/sns-transactions.services.ts
@@ -1,7 +1,9 @@
 import { getTransactions } from "$lib/api/sns-index.api";
 import { DEFAULT_TRANSACTION_PAGE_LIMIT } from "$lib/constants/constants";
 import { snsTransactionsStore } from "$lib/stores/sns-transactions.store";
+import { toastsError } from "$lib/stores/toasts.store";
 import type { Account } from "$lib/types/account";
+import { toToastError } from "$lib/utils/error.utils";
 import type { Principal } from "@dfinity/principal";
 import { decodeSnsAccount } from "@dfinity/sns";
 import { getSnsAccountIdentity } from "./sns-accounts.services";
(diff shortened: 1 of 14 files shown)


lmuntaner:
New UI:

Loading modal:
Screenshot 2022-10-28 at 13 08 57

lmuntaner:

peterpeterparker:
The spinner does not seem centered.

Capture d’écran 2022-11-01 à 08 44 57

frontend/src/lib/services/sns-transactions.services.ts
```diff
@@ -13,18 +15,24 @@ export const loadAccountNextTransactions = async ({
...
+      maxResults: BigInt(DEFAULT_TRANSACTION_PAGE_LIMIT),
+      rootCanisterId,
+    });
+    snsTransactionsStore.addTransactions({
+      accountIdentifier: account.identifier,
+      rootCanisterId,
+      transactions,
+      oldestTxId,
+    });
+  } catch (err) {
```
peterpeterparker, line 33: eslint is not complaining when `err` are not typed?

frontend/src/lib/services/transaction-fees.services.ts
```diff
@@ -28,17 +29,17 @@ export const loadSnsTransactionFee = async (rootCanisterId: Principal) => {
...
-      // TODO: Manage errors gracefully https://dfinity.atlassian.net/browse/GIX-1026
-      console.error(
-        `Error loading sns transaction fee for ${rootCanisterId.toText()}`
-      );
-      console.error(error);
+      toastsError({
+        labelKey: "error.transaction_fee_not_found",
+        err,
+      });
+      console.error(err);
```
peterpeterparker, line 42: `toastsError` prints to the console I think

lmuntaner: true, no need for this.

frontend/src/lib/services/sns-transactions.services.ts
lmuntaner: no, we have a big mix of `err: unknown` and `err`. I believe that catch already comes with an implicity type.

peterpeterparker: Weird. When we migrated to SvelteKit I remembered I had to type many untyped errors because the linter was complaining about the implicit any type  🤷‍♂️.

No issue with your PR, was just a question. Thanks for the feedback.

peterpeterparker (approved):
LGTM, thx

peterpeterparker:
Note: As discussed together on slack, if we ever have another modal that needs the same "spinner loading" tricks, we could improve the Modal component to handle such case - e.g. a `slot` for loading operation or something - so that we would not had to render a modal or another while loading but totally fine for this edge case PR.

PR #1501, opened 2022-11-01
Adding back Manage Neuron Topic

skilesare:
# Motivation

Users should be able to use the manage neuron topic from the NNS

# Changes

I added Manage neuron under an "Advanced" section and added a warning that any neuron added would have control of their neuron including harvesting maturity.

# Tests

I do not have the ability to test. If someone can provide instructions on how I can run this on the local replica, I'll be happy to test.  All I changed was the display code.

### DIFF
--- a/frontend/src/lib/components/neurons/EditFollowNeurons.svelte
+++ b/frontend/src/lib/components/neurons/EditFollowNeurons.svelte
@@ -23,9 +23,9 @@
     {/each}
   </div>
  
-  <div>
+  <h3>
     {$i18n.topics_description.Advanced}
-  </div>
+  </h3>
   <div>
      <FollowTopicSection {neuron} topic={1} />
   </div>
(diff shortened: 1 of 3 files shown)


peterpeterparker:
Thanks @skilesare . Like I told you on the [forum](https://forum.dfinity.org/t/nns-dapp-ui-kit/16298/3?u=peterparker) this repo is not yet open to external contributions. The JS API layers repo - [ic-js](https://github.com/dfinity/ic-js/) - of the dapp on the contrary already welcome external contrib.

skilesare:
Hey @peterpeterparker, Now that SNS is launched, I was wondering if we might be able to get this one slotted?

peterpeterparker:
Hi @skilesare, if I remember correctly you also started a thread on the forum about this right? Can you tag me there (don't reckon which thread) so that we can start from there again?

There is currently some discussion going on about how to generally speaking improve the UI/UX of the "Following".

PR #1502, opened 2022-11-03
GIX-973: Identify SNS CF neurons

lmuntaner:
# Motivation

User can differentiate SNS neurons coming from CF investments.

# Changes

* Two new derived stores sortedSnsUserNeuronsStore and sortedSnsCFNeuronsStore.
* New sns neuron util isCommunityFund.
* Use new derived stores in SnsNeurons page.

# Tests

* Fixed SnsNeurons page test and added new test cases.
* Test for new sns neuron util isCommunityFund.
* Test for new derived stores sortedSnsUserNeuronsStore and sortedSnsCFNeuronsStore


lmuntaner:
New UI:

Screenshot 2022-11-03 at 10 00 01
Screenshot 2022-11-03 at 10 00 28

lmuntaner:

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -235,3 +235,15 @@ export const formattedSnsMaturity = (
...
+ * Returns true if the neuron comes from a Community Fund investment.
+ *
+ * A CF neuron can be identified using the source_nns_neuron_id
+ * which is the NNS neuron that joined the CF for the investment.
+ *
+ * @param {SnsNeuron} neuron
+ * @returns {boolean}
+ */
+export const isCommunityFund = ({ source_nns_neuron_id }: SnsNeuron): boolean =>
+  !isNullish(fromNullable(source_nns_neuron_id));
```
mstrasinskis, line 249: a nitpick: `nonNullish`?
Out of curiosity, should it always be one of the users neuron?

frontend/src/tests/lib/pages/SnsNeurons.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+      );
+    });
+
+    it("should render Community Fund title", async () => {
+      const { queryByTestId } = render(SnsNeurons);
+
+      await waitFor(() =>
+        expect(queryByTestId("community-fund-title")).toBeInTheDocument()
+      );
+    });
```
mstrasinskis, line 120: Probably a test for the class could be useful as an extra logic. But totally up to you.

mstrasinskis (approved):
LGTM, thanks!

frontend/src/tests/lib/pages/SnsNeurons.spec.ts
lmuntaner: the css class?

I thought we don't test UI in component tests.

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: I guess that normally yes, it should always be, otherwise this sns neuron would not be created for this user.

frontend/src/tests/lib/pages/SnsNeurons.spec.ts
lmuntaner: I could change the UI logic with a parent that uses a gap between the two lists. The test would fail, but it should not fail because the UI is the same.

That's why I'm not sure about testing CSS classes in component tests. We should use screenshot testing for that.

mstrasinskis: Thanks for the explanation.

PR #1503, opened 2022-11-04
build: update gix cmp for link _blank and menu

peterpeterparker:
# Motivation

Update gix cmp that contains an improvement to sanitize `_blank` target in links and the new menu design.


mstrasinskis (approved):
LGTM, thank you!

PR #1505, opened 2022-11-04
GIX-1668: Check SNS logo is Png

lmuntaner:
# Motivation

For security reasons, we need to check that the SNS logo from the canister is PNG.

# Changes

* Changed type of logo to `Png` instead of `string`.
* Add util to assert that an assert is of type `Png`: "isPngAsset"
* Use new util in "mapOptionalMetadata".
* Cast default logo as Png.

# Tests

* Add test for the new util "isPngAsset"

### DIFF
--- a/frontend/src/lib/types/assets.ts
+++ b/frontend/src/lib/types/assets.ts
@@ -1,0 +1,1 @@
+export type Png = string;
--- a/frontend/src/lib/types/assets.ts
+++ b/frontend/src/lib/types/assets.ts
@@ -1,1 +1,1 @@
-export type Png = string;
+export type PngDataUrl = string;
--- a/frontend/src/lib/utils/utils.ts
+++ b/frontend/src/lib/utils/utils.ts
@@ -1,4 +1,4 @@
-import type { Png } from "$lib/types/assets";
+import type { PngDataUrl } from "$lib/types/assets";
 import type { Principal } from "@dfinity/principal";
 import { errorToString } from "./error.utils";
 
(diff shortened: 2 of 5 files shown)


lmuntaner:

frontend/src/lib/types/assets.ts
```diff
@@ -0,0 +1 @@
+export type Png = string;
```
mstrasinskis, line 1: nitpick: `PngDataUrl`

frontend/src/lib/utils/utils.ts
```diff
@@ -300,3 +301,12 @@ export const keyOfOptional = <T>({
...
 }): T[keyof T] | undefined => obj?.[key as keyof T];
+
+/**
+ * Returns whether an asset is PNG or not.
+ *
+ * @param {string} src
+ * @returns boolean
+ */
+export const isPngAsset = (asset: string | undefined | Png): asset is Png =>
+  asset?.startsWith("data:image/png;base64,") ?? false;
```
mstrasinskis, line 312: What about adding the `charset` support?
```ts
nonNullish(asset) && /data:image\/png;(charset=[-a-z0-9]+;|)base64,/i.test(asset)
```

mstrasinskis (approved):
LGTM, just a couple of minor suggestions.

frontend/src/lib/utils/utils.ts
lmuntaner: I guess this is more restrictive therefore security should be happy with it.

frontend/src/lib/types/assets.ts
mstrasinskis: `data url` is/was a common name for inline images https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs. But both works for me.

lmuntaner: Oh, interesting. I didn't know that, thanks!

Then `PngDataUrl` is the most accurate. I'll change it.

PR #1509, opened 2022-11-08
feat: review static "sign-in" pages

peterpeterparker:
# Motivation

Review text of the pre-rendered "sign-in" pages with shorterm content.

Add known routes to sitemap.xml.

# Tests

No particular tests regarding static content as it aims to evolve soon.


frontend/src/lib/components/common/SignIn.svelte
```diff
@@ -49,7 +49,7 @@
 </script>
 
 <button on:click={signIn} data-tid="login-button" class="primary" {disabled}>
-  {$i18n.auth.login}
+  <slot>{$i18n.auth.login}</slot>
```
lmuntaner, line 52: Why is this needed?

lmuntaner (approved):
LGTM!

Just a minor question, but already approving.

frontend/src/lib/components/common/SignIn.svelte
peterpeterparker: It's the correct way to set a default value to a slot in Svelte right? I know it's the default way in pure HTML but now that you ask...

Need it because in case of "Voting" screen the CTA is different. Not "Sign in with II" but "Sign in to vote"

lmuntaner: Aaaahh, yes correct.

This is the first time that I see it. Cool!

lmuntaner: Yes, it's the correct way: https://svelte.dev/tutorial/slot-fallbacks

peterpeterparker: Cool, thanks for double checking!

PR #1510, opened 2022-11-08
Style/new design

peterpeterparker:
# Motivation

Small spinner is notably broken because there was a typo in the new fonts CSS sizes.

Fix in gix-cmp https://github.com/dfinity/gix-components/pull/108


peterpeterparker:
Wrong branch as source

PR #1512, opened 2022-11-08
style: new design colors

peterpeterparker:
# Motivation

wip

# Changes

wip

# Tests

wip

### DIFF
--- a/frontend/src/lib/components/common/MenuItems.svelte
+++ b/frontend/src/lib/components/common/MenuItems.svelte
@@ -1,8 +1,8 @@
 <script lang="ts">
   import {
-    IconEngineering,
-    IconHowToVote,
-    IconPsychology,
+    IconExplore,
+    IconUsers,
+    IconPassword,
     IconRocketLaunch,
     IconWallet,
     MenuItem,
@@ -57,7 +57,7 @@
         paths: [AppPath.Neurons, AppPath.Neuron],
       }),
       label: "neurons",
-      icon: IconPsychology,
+      icon: IconPassword,
     },
     {
       context: "proposals",
@@ -67,7 +67,7 @@
         paths: [AppPath.Proposals, AppPath.Proposal],
       }),
       label: "voting",
-      icon: IconHowToVote,
+      icon: IconUsers,
     },
     {
       context: "canisters",
@@ -77,7 +77,7 @@
         paths: [AppPath.Canisters, AppPath.Canister],
       }),
       label: "canisters",
-      icon: IconEngineering,
+      icon: IconExplore,
     },
     // Launchpad should not be visible on mainnet
     ...(ENABLE_SNS
--- a/frontend/src/lib/components/header/AccountMenu.svelte
+++ b/frontend/src/lib/components/header/AccountMenu.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { IconAccount, ThemeToggle, Popover } from "@dfinity/gix-components";
+  import { IconUser, ThemeToggle, Popover } from "@dfinity/gix-components";
   import Logout from "./Logout.svelte";
   import LoginIconOnly from "./LoginIconOnly.svelte";
   import { i18n } from "$lib/stores/i18n";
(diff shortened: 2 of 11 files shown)


frontend/src/app.html
```diff
@@ -82,7 +82,7 @@
       content="%sveltekit.assets%/assets/favicons/browserconfig.xml"
     />
 
-    <meta name="theme-color" content="#383c3c" />
+    <meta name="theme-color" content="#34294e" />
```
lmuntaner, line 85: Maybe a comment of which color this is?

frontend/src/lib/components/common/Layout.svelte
```diff
@@ -1,27 +1,31 @@
...
-<Layout {back} on:nnsBack={async () => await $layoutBackStore?.()}>
-  <div use:triggerDebugReport slot="title">
-    <HeaderTitle>{$layoutTitleStore}</HeaderTitle>
-  </div>
-
+<Layout>
   <MenuItems slot="menu-items" />
 
-  <AccountMenu slot="toolbar-end" />
+  <Content {back} {contrast} on:nnsBack={async () => await $layoutBackStore?.()}>
```
lmuntaner, line 20: Why the `async` and `await`?

lmuntaner (approved):
LGTM, just a couple of questions, but already approving.

frontend/src/app.html
peterpeterparker: I rather not because we don't remove html comments at build time - i.e. each users would have to fetch the comment. I know not that heavy but still, not useful neither.

Also in the manifest we have that same value too and there we cannot add a comment since json.

frontend/src/lib/components/common/Layout.svelte
peterpeterparker: SvelteKit navigation is promise based.

Note that it isn't new code, it was refactored / move from Layout to Content.

frontend/src/app.html
lmuntaner: So we have two coupled hard coded variables that we need to remember to change and keep synchronized. And we rely on our memory to do so?

frontend/src/lib/components/common/Layout.svelte
lmuntaner: What does it mean it's promise based and this `async` and `await`?

Adding `async` function doesn't change that `$layoutBackStore?.()` returns a promise.

I believe that you want to `return await` only if you want to manage the error of the last part in a `try` and `catch`. Which is not the case here.

```
on:nnsBack={async () => await $layoutBackStore?.()}
// should be the same as
on:nnsBack={() => $layoutBackStore?.()}
```

Maybe I'm missig something...

frontend/src/app.html
peterpeterparker: I mean, favicon, theme, manifest json, robots txt , sitemap, are static data that needs to be updated if anything changes. If you think we need script that populates these values, go for it.

frontend/src/lib/components/common/Layout.svelte
peterpeterparker: In this PR this has only been refactored, nothing new here.

I personally think that promises should be handled except when there is a good reason to not too. In this particular case, it guarantees that what is done underneath in the navigation by SvelteKit is for sure finished before re-rendering the ui. In that way I think we are notably sure that beforeNavigation and afterNavigation are executed as expected.

Of couse you are most welcome to edit it in another PR if you think the promise should not be catched.

lmuntaner: Not sure I am explaining myself very well here, hehe.

I am asking: Isn't `async () => await $layoutBackStore?.()` the same as `() => $layoutBackStore?.()`?

peterpeterparker: Likewise hehe, tried to explain that I am not sure it is the same.

But in any case, you know what,  the `layoutBackStore` is deprecated and I plan to remove it after this PR 😜.

mstrasinskis (approved):
LGTM, thank you!

PR #1513, opened 2022-11-08
feat: show busy screen on logout

mstrasinskis:
# Motivation

Display the busy screen on logout to hide not working UI on reload.

# Changes

- show the screen

# Tests

- added the test that the busy store should be called on logout.


peterpeterparker:
Thanks @mstrasinskis 👍

PR #1517, opened 2022-11-09
style: revert to base font 16px

peterpeterparker:
# Motivation

Font size < 16px in inputs makes iOS automatically ping zoom for a11y reason. We do not want such issue.

https://css-tricks.com/16px-or-larger-text-prevents-ios-form-zoom/

Therefore we revert to a base font size of 16px.

# PRs

- [x] https://github.com/dfinity/gix-components/pull/111


mstrasinskis (approved):
LGTM, thanks

PR #1519, opened 2022-11-10
fix: init auth worker race condition

peterpeterparker:
# Motivation

When deployed on the IC, it is possible that fetching the worker takes more time than fetching the code to sync the identity. This can lead to a race condition in the case the user proceed as following:

- user access nns-dapp
- user sign-in
- user refresh the browser

# Changes

- synchronize the auth worker not only when the auth changes but also right after the worker has been loaded

# Tests

This has been debuged locally by setting a delegation that least only 20s and tested manually in various scenario.

### DIFF
--- a/frontend/src/routes/+layout.svelte
+++ b/frontend/src/routes/+layout.svelte
@@ -25,6 +25,7 @@
 
       ready = true;
 
+      // Load app global stores data
       await initAppProxy();
   }
 
--- a/frontend/src/routes/+layout.svelte
+++ b/frontend/src/routes/+layout.svelte
@@ -5,19 +5,35 @@
   import { initWorker } from "$lib/services/worker.services";
   import { initAppProxy } from "$lib/proxy/app.services.proxy";
 
+  let ready = false;
+
   let worker: { syncAuthIdle: (auth: AuthStore) => void } | undefined;
 
-  onMount(async () => (worker = await initWorker()));
+  const syncAuth = async (auth: AuthStore) => {
+      worker?.syncAuthIdle(auth);
 
-  const unsubscribeAuth = authStore.subscribe(async (auth: AuthStore) => {
-    worker?.syncAuthIdle(auth);
+      if (!auth.identity) {
+          ready = false;
+          return;
+      }
 
-    if (!auth.identity) {
-      return;
-    }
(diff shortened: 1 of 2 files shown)


frontend/src/routes/+layout.svelte
```diff
@@ -5,20 +5,36 @@
...
-    }
+      ready = true;
 
-    await initAppProxy();
+      await initAppProxy();
+  }
+
+  onMount(async () => {
+      worker = await initWorker();
+      await syncAuth($authStore);
```
lmuntaner, line 33: Why do we need to call it here?

I'm not sure I understand the race condition. We don't want to call `initAppProxy` twice?

Is that what was causing the issue?

peterpeterparker: sure that does not appear in the PR. 

`initAppProxy` does this:

```
 const initNns: Promise<void>[] = [
    syncAccounts(),
    listNeurons(),
    loadMainTransactionFee(),
  ];

  const initSns: Promise<void>[] = ENABLE_SNS
    ? [loadSnsSummariesProxy(), loadSnsSwapCommitmentsProxy()]
    : [];

  return Promise.allSettled([Promise.all(initNns), Promise.all(initSns)]);
```

Let me know if this already solve your question? if not, happy to provide more details

lmuntaner: Maybe add a comment that we don't want to call `initAppProxy` twice then?

I agree with the changes. But I am not sure this is what it's causing the error. Nothing works when you get that error notification. Not just those init calls.

lmuntaner (approved):
Approving because the change makes sense.

Please add the comment on what we want to avoid the double call, but I believe a double call there is not the cause and it might even not be bad to do so.

peterpeterparker:
Thanks for the review. There is alread a comment, see line `20`.

> // syncAuth is triggered each time the auth changes but also when the worker is initialized to avoid race condition.
>       // As the function can be called twice with a valid identity, we use a flag to only init the data once.

peterpeterparker:
added

> // Load app global stores data

for initApp

frontend/src/routes/+layout.svelte
peterpeterparker: Answered in main thread https://github.com/dfinity/nns-dapp/pull/1519#issuecomment-1310433478

PR #1521, opened 2022-11-11
GIX-1121: Use fee from transaction data

lmuntaner:
# Motivation

In SNS transactions, use the fee from the transactions instead of the fee from the project.

# Changes

* Remove fee prop from `SnsTransactionCard`.
* Remove fee parameter from `mapSnsTransaction` and use instead the `fee` from the transaction if present.

# Tests

* Adapt tests accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2592,7 +2592,7 @@ dependencies = [
 
 [[package]]
 name = "nns-dapp"
-version = "2.0.2"
+version = "2.0.3"
 dependencies = [
  "base64 0.13.0",
  "candid",
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@dfinity/nns-dapp",
-  "version": "2.0.2",
+  "version": "2.0.3",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@dfinity/nns-dapp",
-      "version": "2.0.2",
+      "version": "2.0.3",
       "dependencies": {
         "@dfinity/agent": "^0.14.0",
         "@dfinity/auth-client": "^0.14.0",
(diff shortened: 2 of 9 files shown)


Cargo.lock
```diff
@@ -2592,7 +2592,7 @@ dependencies = [
 
 [[package]]
 name = "nns-dapp"
-version = "2.0.2"
+version = "2.0.3"
```
lmuntaner, line 2595: We need to remember to run `cargo build` when we update the version in the Release branch.

frontend/package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@dfinity/nns-dapp",
-  "version": "2.0.2",
+  "version": "2.0.3",
```
lmuntaner, line 3: We need to remember to run `npm install` after updating the version in the Release branch.

lmuntaner:

frontend/src/lib/utils/sns-transactions.utils.ts
```diff
@@ -133,6 +134,8 @@ const getTransactionInformation = (
   if (data === undefined) {
     throw new Error(`Unknown transaction type ${JSON.stringify(transaction)}`);
   }
+  // Fee is only present in "transfer" transactions.
+  const fee = fromNullable(transaction.transfer[0]?.fee ?? []);
```
peterpeterparker, line 138:
```
const fee = fromNullable(transaction.transfer)?.fee;
```

peterpeterparker (approved):
LGTM, one comment

frontend/src/lib/utils/sns-transactions.utils.ts
lmuntaner: `fee` is also nullable.

It would be 

```
fromNullable(fromNullable(transaction.transfer)?.fee ?? [])`
```

peterpeterparker: we will end up with a `fromNullables(...)` utils someday, I saw it coming ;)

thanks for the feedback

PR #1523, opened 2022-11-11
Chore: Integration Tests

lmuntaner:
# Motivation

Have tests that check the main flows of the application with different edge cases and complex states.

# Changes

* Route level tests is allowed to mock only the api level. Test a main flow like creating a subaccount.

# Tests

All changes are in tests.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,59 @@
+# Testing
+
+Some background to understand and write tests for the nns-dapp.
+
+## Frontend Testing
+
+Tests are in `./frontend/src/tests`.
+
+### Running Tests
+
+```bash
+$ cd frontend
+$ npm test
+```
+
+### Unit Tests
+
+Tests in `frontend/src/tests/lib` and `frontend/src/tests/routes` are the unit tests for each component or functions we have.
+
+A simple component test example can be [AccountBadge.spec.ts](./frontend/src/tests/lib/components/accounts/AccountBadge.spec.ts).
+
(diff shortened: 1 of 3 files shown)


lmuntaner:

My idea would be to use component testing only for specific edge cases or complex behavior.

But otherwise we would rely on this high level tests that check that all components of a flow work together. Tests at this level allows us to prepare the state in the stores how we want it, then perform some steps, and check that the proper api actions were triggered.

Rule: routes/* and pages/* are allowed to mock **only** the api layer.

WDYT?

lmuntaner:

TESTING.md
```diff
@@ -0,0 +1,59 @@
+# Testing
+
+Some background to understand and write tests for the nns-dapp.
```
peterpeterparker, line 3: Add key word "frontend"?

`Some background to...frontend`

TESTING.md
```diff
@@ -0,0 +1,59 @@
...
+Found in `frontend/src/tests/utils`.
+
+## Canister Test
+
+Tests are found in `rs/src/accounts_store/tests.rs`.
+
+### Running Tests
+
+```bash
+$ cd rs
```
peterpeterparker, line 53: not sure but I think you can run these tests from root no?

peterpeterparker (approved):
LGTM, thx

TESTING.md
lmuntaner: I believe so as well, but to be consistent and for simplicity, I added the change of directory.

PR #1524, opened 2022-11-14
style: improve login glitch

peterpeterparker:
# Motivation

Particularly with the new theme, the login screen has become glitchy because it is already presented while it still download CSS on chain.

Moving the layout from a component to a `+layout.ts` changes the way the style is bundled and therefor it becomes applied earlier and screen becomes less glitchy.

# Changes

- move `AuthLayout` to `+layout`

# Glitch

glitch

after this PR less glitchy, not perfect, but less glitchy

lessglitch


mstrasinskis (approved):
Clever, thanks!

PR #1525, opened 2022-11-14
Gix 1107 increase dissolve delay sns neuron

mstrasinskis:
# Motivation

https://dfinity.atlassian.net/browse/GIX-1107

Increase dissolve delay feature.
Switching to the sns properties will be done in a separate PR (https://dfinity.atlassian.net/browse/GIX-1120).

# Changes

- sns api `increaseDissolveDelay`
- sns service `updateDelay`
- created `IncreaseSnsDissolveDelayButton` and subcomponents (similar to `nns` implementation but with props drilling)

# Tests

Manual tests were done locally (sns-demo).

- sns governance api `increaseDissolveDelay`
- sns services `updateDelay`
- related components

# Screenshot

https://user-images.githubusercontent.com/98811342/201718240-092fb946-f3dc-4599-b534-a677f4280a61.mp4


frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
```diff
@@ -36,6 +42,14 @@
         neuron,
         identity: $authStore.identity,
       });
+
+  let allowedToDissolve: boolean;
+  $: allowedToDissolve = isNullish(neuron)
+    ? false
+    : hasPermissionToDissolve({
```
mstrasinskis, line 49: @lmuntaner, do you know, is this enough?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -65,6 +65,15 @@ export const getSnsLockedTimeInSeconds = (
   }
 };
 
+// Delay from now
+export const getSnsDissolveDelaySeconds = (
```
mstrasinskis, line 69: @lmuntaner, does this function make sense or `getSnsLockedTimeInSeconds` can be always used?

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
lmuntaner: Yes, it seems so.

https://github.com/dfinity/ic/blob/master/rs/sns/governance/src/governance.rs#L2991

Also for the buttons to stop and start dissolving.

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
```diff
@@ -0,0 +1,159 @@
...
+
+  let dissolveDelaySeconds: bigint;
+  $: dissolveDelaySeconds = getSnsLockedTimeInSeconds(neuron) ?? 0n;
+
+  const goToConfirmation = async () => {
+    dispatcher("nnsConfirmDelay");
+  };
+</script>
+
+<div class="wrapper">
```
lmuntaner, line 59: Can't we reuse the same UI component for NNS and SNS neurons?

Maybe just passing some props and slots?

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
```diff
@@ -0,0 +1,72 @@
...
+  };
+</script>
+
+<WizardModal {steps} bind:currentStep bind:this={modal} on:nnsClose>
+  <svelte:fragment slot="title">{currentStep?.title}</svelte:fragment>
+  {#if currentStep.name === "SetSnsDissolveDelay"}
+    <SetSnsDissolveDelay
+      {neuron}
+      {token}
+      cancelButtonText={$i18n.core.cancel}
```
lmuntaner, line 53: Why do we need to pass it as props? Is `SetSnsDissolveDelay` being used anywhere else?

I think that NnsDissolveDelay was, that's why we had this props. But Sns neurons might not.

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
```diff
@@ -0,0 +1,72 @@
...
+      bind:delayInSeconds
+    />
+  {/if}
+  {#if currentStep.name === "ConfirmSnsDissolveDelay"}
+    <ConfirmSnsDissolveDelay
+      confirmButtonText={$i18n.neurons.confirm_update_delay}
+      {neuron}
+      {token}
+      {delayInSeconds}
+      {reloadNeuron}
```
lmuntaner, line 67: Could you reload the neuron in the listener of `on:nnsUpdated`?

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
```diff
@@ -0,0 +1,72 @@
...
+      confirmButtonText={$i18n.neurons.update_delay}
+      {minDelayInSeconds}
+      on:nnsCancel={closeModal}
+      on:nnsConfirmDelay={goNext}
+      bind:delayInSeconds
+    />
+  {/if}
+  {#if currentStep.name === "ConfirmSnsDissolveDelay"}
+    <ConfirmSnsDissolveDelay
+      confirmButtonText={$i18n.neurons.confirm_update_delay}
```
lmuntaner, line 63: Why this prop? Similar comment as before.

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: I think this one makes sense because the delay can be either when the neuron is locked or is dissolving. Therefore it might be in different properties.

lmuntaner (commented):
A few comments, thanks!

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
mstrasinskis: Good point, the task was added - https://dfinity.atlassian.net/browse/GIX-1127
I'll do this after/during the `sns` props PR when the all differences are known.

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
mstrasinskis: The original `SetDissolveDelay` is used also in `StakeNeuronModal` so this is a praparation for the `StakeSnsNeuronModal`.

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
mstrasinskis: Because of the `busy` screen, the update neuron call should be done together with the `increase delay` call. But if the `ConfirmSnsDissolveDelay` component will be reused in another place (similar to the `nns` version) the `increase delay` call should be done from the `ConfirmSnsDissolveDelay`. Not a really strong opinion.

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
mstrasinskis: To be reused in `StakeSnsNeuronModal`.

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
lmuntaner: We might do a different way for sns neurons and also for nns neurons afterwards.

We want to set the dissolve delay in the same form where users add the amount. So we might not reuse this. Not worth preparing the props yet.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
mstrasinskis: Thanks!

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
lmuntaner: ah, yes, that makes sense.

I don't think we'll reuse it the screen in the staking process as I mentioned before. Don't prepare for it yet.

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
lmuntaner: Sounds good to me!

lmuntaner (approved):
Already approving, but I would remove the props to prepare to be used in the staking neuron flow. As I mentioned, we might do a different thing there.

There was some user feedback that they wanted to do all that in one screen when the user is staking the neuron, not afterwards.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
mstrasinskis: Permission was applied -- https://github.com/dfinity/nns-dapp/pull/1525/files#diff-c6108bd61146cc60b0bba6665538fb921c858a4100487d177d3ef4111b658833R73
The link is very useful, I’m forever indebted 🤓

frontend/src/lib/components/sns-neurons/ConfirmSnsDissolveDelay.svelte
```diff
@@ -0,0 +1,129 @@
...
+  export let token: Token;
+  export let reloadNeuron: () => Promise<void>;
+
+  const dispatcher = createEventDispatcher();
+
+  let neuronStake: bigint;
+  $: neuronStake = getSnsNeuronStake(neuron);
+
+  let neuronId: string;
+  neuronId = getSnsNeuronIdAsHexString(neuron);
```
peterpeterparker, line 34: no auto-subscription here on purpose? then it can be a `const`

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
mstrasinskis: Thank you for the explanation. Will be refactored.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
```diff
@@ -47,9 +61,16 @@
...
+          <IncreaseSnsDissolveDelayButton
+            {neuron}
+            {token}
+            reloadNeuron={reloadContext}
+          />
+        {/if}
         {#if neuronState === NeuronState.Dissolved && allowedToDisburse}
           <DisburseSnsButton {neuron} {reloadContext} />
-        {:else if neuronState === NeuronState.Dissolving || neuronState === NeuronState.Locked}
+        {:else if (neuronState === NeuronState.Dissolving || neuronState === NeuronState.Locked) && allowedToDissolve}
```
peterpeterparker, line 73: nth: `[NeuronState.Dissolving, NeuronState.Locked].includes...` in case in the future there will be more state added to the list. Maybe can even be a constant

frontend/src/lib/components/sns-neurons/ConfirmSnsDissolveDelay.svelte
```diff
@@ -0,0 +1,129 @@
...
+
+  const updateDissolveDelay = async () => {
+    startBusy({
+      initiator: "dissolve-sns-action",
+    });
+
+    let rootCanisterId: Principal | undefined = $snsOnlyProjectStore;
+
+    assertNonNullish(rootCanisterId);
```
peterpeterparker, line 43: the exception would not be catched, maybe worth to move this check inside the service to catch it?

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
```diff
@@ -0,0 +1,159 @@
...
+          $amount: valueSpan(
+            formatToken({ value: neuronStake, detailed: true })
+          ),
+          $token: token.symbol,
+        })}
+      />
+    </p>
+  </div>
+
+  {#if neuronState === NeuronState.Locked && dissolveDelaySeconds !== undefined}
```
peterpeterparker, line 79: it's ok here if the state is `disolving` or the same constant / check as above should be use?

frontend/src/lib/modals/sns/IncreaseSnsDissolveDelayModal.svelte
```diff
@@ -0,0 +1,72 @@
...
+    {
+      name: "ConfirmSnsDissolveDelay",
+      title: $i18n.neurons.confirm_dissolve_delay,
+    },
+  ];
+
+  let currentStep: WizardStep;
+  let modal: WizardModal;
+
+  let delayInSeconds = Number(getSnsLockedTimeInSeconds(neuron) ?? 0n);
```
peterpeterparker, line 33: not related to review: we can now use `0n` instead of `BigInt(0)`? works in test and prod build too? that would be neat

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -297,6 +301,40 @@ export const stopDissolving = async ({
...
+}: {
+  rootCanisterId: Principal;
+  neuron: SnsNeuron;
+  dissolveDelaySeconds: number;
+}): Promise<{ success: boolean }> => {
+  try {
+    const identity = await getNeuronIdentity();
+    const currentDissolveDelay =
+      getSnsDissolveDelaySeconds(neuron) ?? BigInt(0);
+    const additionalDissolveDelaySeconds =
```
peterpeterparker, line 317: backend throw an error if for whatever reason `additionalDissolveDelaySeconds` is a negative value?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -65,6 +65,15 @@ export const getSnsLockedTimeInSeconds = (
...
+  neuron: SnsNeuron
+): bigint | undefined => {
+  const delay =
```
peterpeterparker, line 72: can you add why there is a fallback on locked time?

frontend/src/lib/components/sns-neurons/ConfirmSnsDissolveDelay.svelte
mstrasinskis: Good catch!

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
lmuntaner: The previous is a message only for Locked. Hence, the condition makes sense.

That said, we should also show the dissolve delay when the state is dissolving. But that also mean to read another property and another message, not `$i18n.neurons.staked`.

Maybe fix it in the issue where you reuse the NNS component for this one.

Does that make sense @mstrasinskis ?

mstrasinskis: Thanks for the explanation. Added to the description of the refactoring task.

PR #1527, opened 2022-11-15
feat: public launchpad

peterpeterparker:
# Motivation

wip

# Changes

wip

# Tests

wip


frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -44,25 +45,30 @@
...
+        text={$i18n.sns_project_detail.max_user_commitment_reached}
+      >
+        <button
+          class="primary"
+          data-tid="sns-project-participate-button"
+          disabled>{$i18n.sns_project_detail.participate}</button
+        >
+      </Tooltip>
+    {/if}
+    <span slot="signin-cta">Sign in to participate</span>
```
lmuntaner, line 70: i18n key

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
```diff
@@ -74,12 +75,14 @@
     <button class="secondary" data-tid="transaction-button-back" on:click={back}
       >{$i18n.accounts.edit_transaction}</button
     >
-    <button
-      class="primary"
-      data-tid="transaction-button-execute"
-      disabled={$busy || disableSubmit}
-      on:click={submit}>{$i18n.accounts.execute}</button
-    >
+    <SignInGuard>
```
lmuntaner, line 78: Do we want to use the `SignInGuard` everywhere there is a private action by default?

Or only the public pages with a private action?

frontend/src/lib/services/$public/app.services.ts
```diff
@@ -0,0 +1,59 @@
...
+import { displayAndCleanLogoutMsg } from "$lib/services/auth.services";
+import { authStore } from "$lib/stores/auth.store";
+import { layoutAuthReady } from "$lib/stores/layout.store";
+import { toastsError } from "$lib/stores/toasts.store";
+
+/**
+ * Load the application public data that are available globally ("global stores").
+ * These data can be read by any users without being signed-in.
+ */
+export const initAppData = (): Promise<
```
lmuntaner, line 16: You already rename it when you use it; therefore, does it make sense to rename this one to `initAppPublicData` and the other one to `initAppPrivateData`?

Not a strong opinion, but since you already do so when importing, maybe it does make sense.

lmuntaner (approved):
One small change and a couple of comments, but already approving.

Thanks!

frontend/src/lib/services/$public/app.services.ts
peterpeterparker: Well yesterday you said that we should use the same function name regardless if public or private, to which I said I got no strong opinion, so I used the same function name as you said 😉

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
peterpeterparker: On every private actions that could potentially be displayed in a public page

peterpeterparker:
@lmuntaner fyi just pushed a small change: for the time being, if user is not signed in, on "Launchpad", proposals will be hidden since these are not yet public in this PR.

frontend/src/lib/components/project-detail/ParticipateButton.svelte
peterpeterparker: 🤦‍♂️

Thanks

frontend/src/lib/services/$public/app.services.ts
lmuntaner: I meant when you move one function to another place. But in this case we have the same function in public and in private. Each doing something different. Therefore, it might be confusing if they both have the same name.

listProposals does not have a listPrivateProposals and listPublicProposals for example. That's what I meant.

lmuntaner:
> @lmuntaner fyi just pushed a small change: for the time being, if user is not signed in, on "Launchpad", proposals will be hidden since these are not yet public in this PR.

SGTM

frontend/src/lib/services/$public/app.services.ts
peterpeterparker: Not against. Done.

peterpeterparker:
Last upgrade of this PR have been displayed and discussed in a call.

PR #1529, opened 2022-11-15
GIX-1116: Syns SNS Neuron Balances

lmuntaner:
# Motivation

Check sns neuron balances with the stake and claim or refresh accordingly if they don't match

# Changes

* New sns-governance api: "getNeuronBalance", "refreshNeuron", "claimNeuron".
* Rename sns neuron service "loadSnsNeurons" to "syncSnsNeurons". Add balance check there.
* Add balance check in getSnsNeuron service.
* Balance check and refreshing is done in "checkSnsNeuronBalances". No recursion, each neuron refreshed is loaded independently.
* New method in sns neurons store "addNeurons".

# Tests

* Test new cases in syncSnsNeurons and getSnsNeuron services.
* Test new sns neurons method.
* Test new sns governance apis.

### DIFF
--- a/frontend/src/lib/services/sns-neurons.services.ts
+++ b/frontend/src/lib/services/sns-neurons.services.ts
@@ -2,6 +2,7 @@ import {
   addNeuronPermissions,
   disburse as disburseApi,
   increaseDissolveDelay as increaseDissolveDelayApi,
+  refreshNeuron,
   removeNeuronPermissions,
   startDissolving as startDissolvingApi,
   stopDissolving as stopDissolvingApi,
@@ -31,12 +32,30 @@ import {
   type SnsNeuron,
   type SnsNeuronId,
 } from "@dfinity/sns";
-import { arrayOfNumberToUint8Array, fromDefinedNullable } from "@dfinity/utils";
+import {
+  arrayOfNumberToUint8Array,
+  fromDefinedNullable,
+  fromNullable,
+} from "@dfinity/utils";
 import { get } from "svelte/store";
 import { getIdentity } from "./auth.services";
+import {
+  checkSnsNeuronBalances,
+  neuronNeedsRefresh,
+} from "./sns-neurons-check-balances.services";
 import { queryAndUpdate } from "./utils.services";
 
-export const loadSnsNeurons = async (
+/**
+ * Loads sns neurons in store and checks neurons's stake against the balance of the subaccount.
+ *
+ * On update, it will check whether there are neurons that need to be refreshed or claimed.
+ * A neuron needs to be refreshed if the balance of the subaccount doesn't match the stake of the neuron.
+ * A neuron needs to be claimed if there is a subaccount with balance and no neuron.
+ *
+ * @param {Principal} rootCanisterId
+ * @returns {void}
+ */
+export const syncSnsNeurons = async (
   rootCanisterId: Principal
 ): Promise<void> => {
   return queryAndUpdate<SnsNeuron[], unknown>({
@@ -46,12 +65,19 @@ export const loadSnsNeurons = async (
         identity,
         certified,
       }),
-    onLoad: ({ response: neurons, certified }) => {
+    onLoad: async ({ response: neurons, certified }) => {
       snsNeuronsStore.setNeurons({
         rootCanisterId,
         neurons,
         certified,
       });
+
+      if (certified) {
+        checkSnsNeuronBalances({
+          rootCanisterId,
+          neurons,
+        });
+      }
     },
     onError: ({ error: err, certified }) => {
       console.error(err);
@@ -71,6 +97,26 @@ export const loadSnsNeurons = async (
       );
     },
     logMessage: "Syncing Sns Neurons",
+  });
+};
+
+const loadNeurons = async ({
+  rootCanisterId,
+  certified,
+}: {
+  rootCanisterId: Principal;
+  certified: boolean;
+}): Promise<void> => {
+  const identity = await getIdentity();
+  const neurons = await querySnsNeurons({
+    identity,
+    rootCanisterId,
+    certified,
+  });
+  snsNeuronsStore.setNeurons({
+    rootCanisterId,
+    certified,
+    neurons,
   });
 };
 
@@ -144,8 +190,26 @@ export const getSnsNeuron = async ({
         certified,
         neuronId: { id: neuronId },
       }),
-    onLoad: ({ response: neuron, certified }) => {
+    onLoad: async ({ response: neuron, certified }) => {
       onLoad({ neuron, certified });
+
+      if (certified) {
+        // Check that the neuron's stake is in sync with the subaccount's balance
+        const neuronId = fromNullable(neuron.id);
+        if (neuronId !== undefined) {
+          const identity = await getNeuronIdentity();
+          if (await neuronNeedsRefresh({ rootCanisterId, neuron, identity })) {
+            await refreshNeuron({ rootCanisterId, identity, neuronId });
+            const updatedNeuron = await querySnsNeuron({
+              identity,
+              rootCanisterId,
+              neuronId,
+              certified,
+            });
+            onLoad({ neuron: updatedNeuron, certified });
+          }
+        }
+      }
     },
     onError: ({ certified, error }) => {
       onError?.({ certified, error });
@@ -354,7 +418,7 @@ export const stakeNeuron = async ({
       identity,
       source: decodeSnsAccount(account.identifier),
     });
-    await loadSnsNeurons(rootCanisterId);
+    await loadNeurons({ rootCanisterId, certified: true });
     return { success: true };
   } catch (err) {
     toastsError(
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1653,9 +1653,9 @@
       }
     },
     "node_modules/@dfinity/cmc": {
-      "version": "0.0.3-next-2022-11-11.1",
-      "resolved": "https://registry.npmjs.org/@dfinity/cmc/-/cmc-0.0.3-next-2022-11-11.1.tgz",
-      "integrity": "sha512-/1aJJ22hzZAT/Wo2p2aTU2KKtYP6Uiw53aky4Im9QhdbW5F1tqeOR4lAhbnuD4QYxTkvx62TWRllLUKiBHKz+g==",
+      "version": "0.0.3-next-2022-11-16.2",
+      "resolved": "https://registry.npmjs.org/@dfinity/cmc/-/cmc-0.0.3-next-2022-11-16.2.tgz",
+      "integrity": "sha512-Wzmjc+4olSEhZ15wSV3Ua6o1w0V9Y886+zyhEIzPT70xTNgtawWfh0okfCx7FH3O1fQBu5GmKeYRvUlgVJBpsw==",
       "peerDependencies": {
         "@dfinity/utils": "^0.0.6-next"
       }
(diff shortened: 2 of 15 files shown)


lmuntaner:

PENDING: Tests, but the logic is already working.

lmuntaner:
@mstrasinskis ready for full review

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,226 @@ export const loadSnsNeurons = async (
...
+        rootCanisterId,
+        neuronId: currentNeuronId,
+        certified: false,
+        identity,
+      });
+      if (currentBalance > BigInt(0)) {
+        const neuron = neurons.find(
+          (neuron) =>
+            getSnsNeuronIdAsHexString(neuron) ===
+            subaccountToHexString(subaccount)
```
mstrasinskis, line 246: Just a suggestion and probably not part of this PR. Because of it's own scope, what do you think if all the sns neuron-stake related entities are prefixed with `ns` (e.g. `nsSubaccountToHexString`)?

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -10,29 +13,35 @@ import {
...
   type SnsNeuronId,
 } from "@dfinity/sns";
-import { arrayOfNumberToUint8Array } from "@dfinity/utils";
+import { arrayOfNumberToUint8Array, fromNullable } from "@dfinity/utils";
 import { get } from "svelte/store";
 import { getIdentity } from "./auth.services";
 import { queryAndUpdate } from "./utils.services";
 
-export const loadSnsNeurons = async (
+export const syncSnsNeurons = async (
```
mstrasinskis, line 44: Probably worth to add a comment that calling this function is the only way to reliable finalise transactions and get the actual neuron stakes. Just to emphasise its importance.
(Ideally also a short algorithm description from the doc because its implementation is not limited to one function)

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,226 @@ export const loadSnsNeurons = async (
...
+  const visitedSubaccounts = new Set<string>();
+  // TODO: Check neurons controlled by linked HW?
+  const identity = await getIdentity();
+  const controller = identity.getPrincipal();
+  // Visit subaccounts by order until we find a balance of 0.
+  // Therefore, we set the initial balance to some non-zero value.
+  let currentBalance = BigInt(1);
+  for (
+    let index = 0;
+    index < MAX_NEURONS_SUBACCOUNTS && currentBalance > BigInt(0);
```
mstrasinskis, line 225: Probably using `break` instead of `currentBalance > BigInt(0)` could simplify the code a bit.

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,226 @@ export const loadSnsNeurons = async (
...
+      visitedSubaccounts.add(subaccountToHexString(subaccount));
+      // Id of an sns neuron is the subaccount of the sns governance canister where the stake is
+      const currentNeuronId: SnsNeuronId = { id: subaccount };
+      currentBalance = await getNeuronBalance({
+        rootCanisterId,
+        neuronId: currentNeuronId,
+        certified: false,
+        identity,
+      });
+      if (currentBalance > BigInt(0)) {
```
mstrasinskis, line 242: nitpick: would be nice to to have more consts for readability. e.g.
```
const positiveBalance: boolean
const knownNeuron: boolean
const differentBalance: boolean
```

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,226 @@ export const loadSnsNeurons = async (
...
+            controller,
+            memo: BigInt(index),
+            subaccount,
+          });
+        } else {
+          // Check if neuron needs refresh.
+          const neuronId = fromNullable(neuron.id);
+          if (neuronId !== undefined) {
+            if (
+              await neuronNeedsRefresh({
```
mstrasinskis, line 262: Wouldn't it check the same balance as loaded before #236?
https://github.com/dfinity/nns-dapp/pull/1529/files#diff-3a1a9c5c924b5918e75d65e6497c777fa172c89b58d7a649e8ffa9b5c97f9359R236

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,226 @@ export const loadSnsNeurons = async (
...
+  });
+  snsNeuronsStore.addNeurons({
+    rootCanisterId,
+    certified,
+    neurons: [neuron],
+  });
+};
+
+/**
+ * Returns true if the neuton's stake doesn't match the balance of the subaccount.
```
mstrasinskis, line 133:
```suggestion
 * Returns true if the neuron's stake doesn't match the balance of the subaccount.
```

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,226 @@ export const loadSnsNeurons = async (
...
+          }
+        }
+      }
+    } catch (error) {
+      // Ignore the error, we do this in the background.
+      // If this fails the data might be stale but it should refreshed on the next sync.
+      console.error(error);
+    }
+  }
+  // Check unvisited neurons.
```
mstrasinskis, line 285: nitpick: Probably worth a separate function. This one is already pretty big.

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,226 @@ export const loadSnsNeurons = async (
...
+            }
+          }
+        }
+      }
+    } catch (error) {
+      // Ignore the error, we do this in the background.
+      // If this fails the data might be stale but it should refreshed on the next sync.
+      console.error(error);
+    }
+  }
```
mstrasinskis, line 284: Is this out of the current PR scope?
> if subaccount in neurons:
>      neurons.remove(subaccount)

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: That's a good idea, I'll add it.

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,201 @@ export const loadSnsNeurons = async (
   });
 };
 
+const loadNeuron = async ({
+  rootCanisterId,
```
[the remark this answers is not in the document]
lmuntaner: Sounds good. Initially I thought I might export it and use it somewhere else. That's why I didn't add it. But for now it's more of a helper function.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: 😂

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,201 @@ export const loadSnsNeurons = async (
...
+      (await getNeuronBalance({
+        rootCanisterId,
+        neuronId,
+        certified: false,
+        identity,
+      }));
+    return balance !== neuron.cached_neuron_stake_e8s;
+  }
+  // Edge case, a valid neuron should have a neuron id
+  return false;
```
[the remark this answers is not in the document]
lmuntaner: 👍

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,226 @@ export const loadSnsNeurons = async (
...
+  rootCanisterId: Principal;
+  neurons: SnsNeuron[];
+}): Promise<void> => {
+  const visitedSubaccounts = new Set<string>();
+  // TODO: Check neurons controlled by linked HW?
+  const identity = await getIdentity();
+  const controller = identity.getPrincipal();
+  // Visit subaccounts by order until we find a balance of 0.
+  // Therefore, we set the initial balance to some non-zero value.
+  let currentBalance = BigInt(1);
```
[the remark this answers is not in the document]
lmuntaner: Sounds good

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: I had it, but it obscured a little bit the condition to break the loop.

I found the logic clearer like this. A break seems like an exception, but in this loop, that's exactly when we want to stop, more than the `MAX_NEURONS_SUBACCOUNTS`.

I almost used a while loop, but we also needed the index to go up, that's why I used the for loop.

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -70,6 +86,201 @@ export const loadSnsNeurons = async (
...
+  for (
+    let index = 0;
+    index < MAX_NEURONS_SUBACCOUNTS && currentBalance > BigInt(0);
+    index++
+  ) {
+    try {
```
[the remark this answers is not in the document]
lmuntaner: Good point

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: All in for readability, sounds good.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: `subaccountToHexString` has nothing to do with neurons.

We could use this for SNS subaccounts as well.

Do you mean to diffentiate utils that are for SNS and for NNS?

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: Yes it does.

That's why I pass it as parameter:

```
const needsRefresh = await neuronNeedsRefresh({
  rootCanisterId,
  neuron,
  identity,
  // Here I pass the balance
  balanceE8s: currentBalance,
});
```

I used a helper to reuse the logic of what it means to need a refresh.

But maybe I could use an util that expects the balance, and use it here and use the service that also fetches the balance in the other places.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: This probably depends on the error.

But because this is an extra check, I think that if chekcing balances fail, we are safe to assume that the update call to get the neurons is a good fallback.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: Sounds good, let me refactor a bit.

lmuntaner:
@mstrasinskis please review. I did quite a bit of refactor to try to better explain what's going on. Let me know what you think.

frontend/src/lib/services/sns-neurons.services.ts
mstrasinskis: Thanks for the explanation

frontend/src/lib/services/sns-neurons.services.ts
mstrasinskis: I thought that it is a bit confusing because `subaccountToHexString` also has nothing to do with account subaccounts. But not a strong opinion.

lmuntaner: discussed in a call

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: Done. The algorithm refreshes the neuron and if the stake if 0 it will not show in the UI.

lmuntaner:
@mstrasinskis please review

I have refactored the code to its own file.

mstrasinskis (approved):
LGTM. Special thanks for the documentation update!

PR #1530, opened 2022-11-15
feat: public proposals

peterpeterparker:
# Motivation

Make sharing information about proposals easier, more convenient and more user friendly.

Shorten the user journey till vote.

To do so, make public information about proposals in related views public and require user to sign-in to vote or to use features that needs information about users' neurons such as "Show only proposals you can still vote for" and "Proposer modal".

# PRs

- This PR -> #1527 -> `main`

# Changes

- move `proposals.services` to `$public`
- make related query functions use new `current` identity - i.e. either query with signed-in users identity or anonymous
- new `<SignedInOnly />` component that hide totally a `slot` if user is not signed in
- use above component for features that needs neurons or user to be signed-in
- update views to not wait on neurons - spinner etc.-  to display information when user is not signed-in
- new `auth.services` function that returns either the signed-in identity or the anonymous as fallback


frontend/src/lib/components/proposal-detail/ProposalSystemInfoProposerEntry.svelte
```diff
@@ -3,10 +3,16 @@
...
   import VotingHistoryModal from "$lib/modals/neurons/VotingHistoryModal.svelte";
   import { Html, KeyValuePairInfo } from "@dfinity/gix-components";
+  import { isSignedIn } from "$lib/utils/auth.utils";
+  import { authStore } from "$lib/stores/auth.store";
 
   export let proposer: NeuronId | undefined;
 
   let modalOpen = false;
+
+  // For simplicity reason, currently we do not display the proposer details if not signed-in
```
lmuntaner, line 13: Maybe a TODO also to make it clear that we want to support that in the future?

frontend/src/tests/lib/pages/ProposalDetail.spec.ts
```diff
@@ -53,13 +52,10 @@ describe("ProposalDetail", () => {
     proposalIdText: `${mockProposals[0].id}`,
   };
 
-  it("should render loading neurons", async () => {
-    const { getByText } = render(ProposalDetail, props);
-
+  it("should render proposal detail", async () => {
```
lmuntaner, line 55: Do we want a test that the page also renders when not logged in?

frontend/src/tests/lib/services/_public/proposals.services.spec.ts
```diff
@@ -210,18 +208,24 @@ describe("proposals-services", () => {
...
+    it("should list proposals if no identity", async () => {
       const call = async () =>
         await listProposals({
           loadFinished: () => {
             // do nothing here
           },
         });
 
-      await expect(call).rejects.toThrow(Error(mockIdentityErrorMsg));
+      await expect(call).resolves;
```
lmuntaner, line 225: Don't we want to expect `queryProposals` to be called?

frontend/src/tests/lib/services/auth.services.spec.ts
```diff
@@ -144,4 +155,32 @@ describe("auth-services", () => {
       spy.mockClear();
     });
   });
+
+  describe("identity", () => {
```
lmuntaner, line 159: Is this the test for `getCurrentIdentity`?

frontend/src/tests/lib/utils/accounts.utils.spec.ts
```diff
@@ -72,7 +73,7 @@ describe("accounts-utils", () => {
       const subaccount = new Uint8Array(32).fill(0);
       subaccount[31] = 1;
       const account = {
-        owner: Principal.fromText("2vxsx-fae"),
+        owner: new AnonymousIdentity().getPrincipal(),
```
lmuntaner, line 76: Why this change?

lmuntaner (approved):
Just a few questions but already approving. Looks good!

frontend/src/lib/components/proposal-detail/ProposalSystemInfoProposerEntry.svelte
peterpeterparker: Sure.

frontend/src/tests/lib/pages/ProposalDetail.spec.ts
peterpeterparker: Sure, done

frontend/src/tests/lib/services/_public/proposals.services.spec.ts
peterpeterparker: That's covered by existing tests I would say, like "should call the canister to list proposals"

frontend/src/tests/lib/services/auth.services.spec.ts
peterpeterparker: Yep, renamed it

frontend/src/tests/lib/utils/accounts.utils.spec.ts
peterpeterparker: Replace an hardcoded value by it's correct implementation equivalent

lmuntaner: ok, thanks!

PR #1531, opened 2022-11-16
feat: use back property instead of store

peterpeterparker:
# Motivation

The `layoutBackStore` was deprecated. In addition, updating the store after all components are mounted might lead to some glitch refreshing the navigation header - updating menu and back buttons - on route changes.

# PRs

- [x] linked with PR https://github.com/dfinity/gix-components/pull/115

# Changes

- deprecate `$layoutBackStore` and replace with property defined on `+layout`

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1661,9 +1661,9 @@
       }
     },
     "node_modules/@dfinity/gix-components": {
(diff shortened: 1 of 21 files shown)


mstrasinskis (approved):
LGTM 👍

PR #1537, opened 2022-11-17
Fix: Navigation and SNS Neuron Detail issues

lmuntaner:
# Motivation

Fix some navigation issues and hides dissolve buttons depending on permissions.

# Changes

* Change the button of the SNS Disburse modal. It was "back" but that was the first screen. Switched to "Cancel".
* New sns neuron util "hasPermissionToDissolve" used in SnsNeuronMetaInfoCard to hide Dissolve button.
* Check the path from the store before raising an error if neuron or wallet not found.
* Use `accountsPathStore` and `neuronsPathStore` to maintain the universe when going back from Wallet and Neuron detail.

# Tests

* Test the new util hasPermissionToDissolve
* Add test to SnsNeuronMetaInfoCard for new cases and also disburse cases.
* Fix test SnsNeuronDetail after new check.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
@@ -15,7 +15,7 @@
     hasPermissionToDissolve,
   } from "$lib/utils/sns-neuron.utils";
   import { authStore } from "$lib/stores/auth.store";
-  import { isDefined, isNullish, nonNullish } from "$lib/utils/utils";
+  import { isNullish, nonNullish } from "$lib/utils/utils";
   import { NeuronState, type Token } from "@dfinity/nns";
   import DissolveSnsNeuronButton from "$lib/components/sns-neuron-detail/actions/DissolveSnsNeuronButton.svelte";
   import { fromDefinedNullable } from "@dfinity/utils";
--- a/frontend/__mocks__/$app/stores.ts
+++ b/frontend/__mocks__/$app/stores.ts
@@ -14,6 +14,7 @@ const initPageStoreMock = () => {
     subscribe,
 
     mock: ({
+      // routeId is the path
       routeId = undefined,
       data = { universe: OWN_CANISTER_ID_TEXT },
     }: {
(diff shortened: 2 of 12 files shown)


lmuntaner:

@peterpeterparker as discussed, this solves the issue for the user, but we should understand why `onDestroy` is not called. This promises to be a source of bugs because the components assume they are removed once it's not their page.

Let's discuss on Monday how to prioritize this.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
```diff
@@ -49,7 +58,7 @@
       <div class="buttons">
         {#if neuronState === NeuronState.Dissolved && allowedToDisburse}
           <DisburseSnsButton {neuron} {reloadContext} />
-        {:else if neuronState === NeuronState.Dissolving || neuronState === NeuronState.Locked}
+        {:else if neuronState === NeuronState.Dissolving || (neuronState === NeuronState.Locked && allowedToDissolve)}
```
mstrasinskis, line 61: nitpick: probably worth a const.

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -116,6 +116,21 @@ export const hasPermissionToDisburse = ({
     permissions: [SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_DISBURSE],
   });
 
+export const hasPermissionToDissolve = ({
```
mstrasinskis, line 119: I guess it's already in `main`.

mstrasinskis (approved):
LGTM, thanks!

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
lmuntaner: I used your canDissolve variable

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: yes, I found it when rebasing

PR #1547, opened 2022-11-21
FIX: Cache sns wrapper by identity

lmuntaner:
# Motivation

We are using different identities to interact with SNSes. For now logged in and anonymous.

Therefore, we need to cache the sns wrapper by identity so that we don't use the wrong wrapper. For example, using the anonymous identity wrapper to get the balance.

# Changes

* Change implementation of "wrappers" to cache by identity principal instead of only by certified or not.
* Map transfer errors and show a proper message according to each error.

# Tests

* No changes. Same functionality.

### DIFF
--- a/frontend/src/lib/api/sns-wrapper.api.ts
+++ b/frontend/src/lib/api/sns-wrapper.api.ts
@@ -14,7 +14,7 @@ import { Principal } from "@dfinity/principal";
 import type { InitSnsWrapper, SnsWrapper } from "@dfinity/sns";
 
 interface IdentityWrapper {
-  [principal: string]: Map<QueryRootCanisterId, SnsWrapper>;
+  [principal: string]: Promise<Map<QueryRootCanisterId, SnsWrapper>>;
 }
 const identitiesCertifiedWrappers: IdentityWrapper = {};
 const identitiesNotCertifiedWrappers: IdentityWrapper = {};
@@ -148,7 +148,9 @@ export const wrappers = async ({
   const principalText = identity.getPrincipal().toText();
   if (certified) {
     if (identitiesCertifiedWrappers[principalText] === undefined) {
-      identitiesCertifiedWrappers[principalText] = await initWrappers({
+      // the initialization of the wrappers can be called mutliple times at the same time when the app loads.
+      // We cache the promise so that if multiple calls are made then all will resolve when the first init resolve.
+      identitiesCertifiedWrappers[principalText] = initWrappers({
         identity,
         certified,
       });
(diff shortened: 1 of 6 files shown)


lmuntaner:

frontend/src/lib/api/sns-wrapper.api.ts
```diff
@@ -144,17 +145,23 @@ export const wrappers = async ({
...
-      return snsQueryWrappers;
-    default:
-      if (!snsUpdateWrappers) {
-        snsUpdateWrappers = initWrappers({ identity, certified: true });
-      }
-      return snsUpdateWrappers;
+  const principalText = identity.getPrincipal().toText();
+  if (certified) {
+    if (identitiesCertifiedWrappers[principalText] === undefined) {
+      identitiesCertifiedWrappers[principalText] = await initWrappers({
```
peterpeterparker, line 151: as war as I remember it was actually important for performance reason to not use `await` here. @mstrasinskis can confirm.

lmuntaner: Yes, earlier the cache was a Promise of a Map. Which I found weird.

I guess that made sense when we were not requesting the projects on init and therefore the sns wrapper was not needed.

But now that we load the projects on init, what's the difference between has an await here or when we get the wrapper?

```
export const wrapper = async ({
  identity,
  rootCanisterId,
  certified,
}: {
  identity: Identity;
  rootCanisterId: QueryRootCanisterId;
  certified: boolean;
}): Promise<SnsWrapper> => {
  const snsWrapper: SnsWrapper | undefined = (
    await wrappers({ identity, certified })
  ).get(rootCanisterId);
```

or

```
export const queryAllSnsMetadata = async ({
  identity,
  certified,
}: {
  certified: boolean;
  identity: Identity;
}): Promise<QuerySnsMetadata[]> => {
  logWithTimestamp(
    `Listing all deployed Sns summaries certified:${certified} call...`
  );

  const snsWrappers: SnsWrapper[] = [
    ...(
      (await wrappers({ identity, certified })) ??
      new Map<QueryRootCanisterId, SnsWrapper>()
    ).values(),
  ];
```

peterpeterparker: We definitely explicitely removed the `await` in this PR https://github.com/dfinity/nns-dapp/pull/1119/

I think it's because the initialization of the wrappers can be called mutliple times at the same time when the app loads. So it matters to initialize the wrapper but it's faster to not await because if multiple calls are made then all will resolve when the first init resolve. In addition, in that way we initialize the wrapper only once per query or update and not multiple times.

lmuntaner: ok, I'll add a note, that makes sense.

lmuntaner:

peterpeterparker (approved):
LGTM, thx!

cklim278 (commented):
Solve

PR #1549, opened 2022-11-21
feat: sync sns params

mstrasinskis:
# Motivation

Load and util sns nervous system parameters in dissolve delay process and sns voting power calculation.
https://dfinity.atlassian.net/browse/GIX-1120

# Changes

- api/`nervousSystemParameters()`
- reuse `<NeuronStateRemainingTime>` in increase delay modal
- add and apply sns neuron utils/`snsVotingPower()`
- bind sns params to increase dissolve delay process
- `snsParametersStore`
- `snsProjectParametersStore`

# Tests

All created entries are covered with tests.


frontend/src/lib/components/sns-neurons/ConfirmSnsDissolveDelay.svelte
```diff
@@ -24,6 +29,19 @@
 
   let neuronId: string;
   $: neuronId = getSnsNeuronIdAsHexString(neuron);
+
+  let snsParameters: NervousSystemParameters | undefined;
+  $: snsParameters = $snsProjectParametersStore?.parameters;
+
+  let votingPower: number | undefined;
+  $: if (neuron !== undefined && snsParameters !== undefined) {
+    votingPower = snsVotingPower({
```
lmuntaner, line 38: worth having the same api as the refactor I did?

* neuron
* newDissolveDelay

lmuntaner: plus the params used? Or maybe all params?

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
```diff
@@ -48,6 +42,44 @@
...
+      ? undefined
+      : Number(fromDefinedNullable(snsParameters?.max_dissolve_delay_seconds));
+
+  let minDissolveDelaySeconds: number | undefined;
+  $: minDissolveDelaySeconds =
+    snsParameters === undefined
+      ? undefined
+      : Number(
+          fromDefinedNullable(
+            snsParameters?.neuron_minimum_dissolve_delay_to_vote_seconds
```
lmuntaner, line 60: I would use a derived store.

frontend/src/lib/derived/sns/sns-project-parameters.derived.ts
```diff
@@ -0,0 +1,13 @@
+import { snsProjectIdSelectedStore } from "$lib/derived/selected-project.derived";
+import {
+  snsParametersStore,
+  type SnsParameters,
+} from "$lib/stores/sns-parameters.store";
+import { derived, type Readable } from "svelte/store";
+
+export const snsProjectParametersStore: Readable<SnsParameters | undefined> =
```
lmuntaner, line 8: Not convinced about this store.

As I mentioned, using this kind of stores means that the components won't be reusable in multi-token pages. Like a dashboard or a page with neurons from multiple projects.

I believe we should find a better solution.

lmuntaner: For this PR, I'd say that the neurons expect the rootCanisterId and from that and an util get the params of the project.

frontend/src/lib/services/sns-parameters.services.ts
```diff
@@ -0,0 +1,46 @@
+import { nervousSystemParameters } from "$lib/api/sns-governance.api";
+import { snsParametersStore } from "$lib/stores/sns-parameters.store";
+import { toastsError } from "$lib/stores/toasts.store";
+import { toToastError } from "$lib/utils/error.utils";
+import type { Principal } from "@dfinity/principal";
+import type { NervousSystemParameters } from "@dfinity/sns/dist/candid/sns_governance";
+import { queryAndUpdate } from "./utils.services";
+
+export const syncSnsParameters = async ({
```
lmuntaner, line 9: I would call this `loadSnsParameters`.

For my `syncXXX` means that id does more than just loading. Like checking balances.

frontend/src/lib/services/sns-parameters.services.ts
```diff
@@ -0,0 +1,46 @@
...
+  rootCanisterId,
+}: {
+  rootCanisterId: Principal;
```
lmuntaner, line 12: Why the object params instead of only one `(rootCanisterId: Principal)`?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -294,6 +298,113 @@ export const needsRefresh = ({
...
+    dissolve_state,
+    aging_since_timestamp_seconds,
+  } = neuron;
+  const votingPowerPercentageMultiplier = Number(
+    voting_power_percentage_multiplier
+  );
+
+  let agingSinceTimestampSeconds = Number(aging_since_timestamp_seconds);
+  const dissolveState = fromDefinedNullable(dissolve_state);
+  // We don't use value of `aging_since_timestamp_seconds` because the increase dissolve delay backend function updates it.
```
lmuntaner, line 365: I would ask the NNS team whether we can just use `aging_since_timestamp_seconds`.

I believe this might already be updated from the backend with the proper value.

In NNS, the ageSeconds property already gives us the value needed in the calculation.

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -294,6 +298,113 @@ export const needsRefresh = ({
...
+    } else {
+      agingSinceTimestampSeconds = nowSeconds;
+    }
+  }
+
+  const dissolveDelay = Math.min(
+    dissolveDelayInSeconds,
+    maxDissolveDelaySeconds
+  );
+  const stake = Math.max(
```
lmuntaner, line 387: You also need the staked maturity.

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -294,6 +298,113 @@ export const needsRefresh = ({
...
+    Number(getSnsNeuronStake(neuron) - neuron_fees_e8s),
+    0
+  );
+  const dissolveDelayBonus =
```
lmuntaner, line 391: I prefer the calculation we do in NNS, the backend code is not so readable IMHO:

```
votingPower = (stake + stakedMaturity) * dissolveBonus * ageBonus;
```

I believe both NNS and SNS should read the same calculation. Even share it if possible.

lmuntaner (commented):
First review, thanks!

frontend/src/lib/components/sns-neurons/ConfirmSnsDissolveDelay.svelte
mstrasinskis: Done 👍

frontend/src/lib/services/sns-parameters.services.ts
mstrasinskis: Thanks for the explanation

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: Not relevant anymore

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: After refactoring the `sns` calculation is almost the same as `nns`: 
```
sns voting power = nns_vp * voting_power_percentage_multiplier
```

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: It's not in the backend code yet, but let's be ahead.

lmuntaner: Mhm, ask them. Maybe in SNS it doesn't count? But that would not make sense...

frontend/src/lib/derived/sns/sns-project-parameters.derived.ts
mstrasinskis: The `snsProjectParametersStore` was removed 👍

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
mstrasinskis: Do you mean for `neuron_minimum_dissolve_delay_to_vote_seconds` only for to deliver a cleaner version of all sns params?

mstrasinskis:
@lmuntaner ready for another review?

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
```diff
@@ -21,13 +21,21 @@
...
+  } from "$lib/derived/selected-project.derived";
 
   const { store }: SelectedSnsNeuronContext =
     getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
 
   let neuron: SnsNeuron | undefined | null;
   $: neuron = $store.neuron;
 
+  let rootCanisterId: Principal | undefined;
+  $: rootCanisterId = $snsOnlyProjectStore;
```
lmuntaner, line 37: Why is this needed?

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
lmuntaner: I meant one for each parameter.

But maybe let's discuss and do it in another PR. This is ok for now.

frontend/src/lib/services/sns-parameters.services.ts
```diff
@@ -0,0 +1,44 @@
...
+export const loadSnsParameters = async (
+  rootCanisterId: Principal
+): Promise<void> => {
+  await queryAndUpdate<NervousSystemParameters, unknown>({
+    request: ({ certified, identity }) =>
+      nervousSystemParameters({
+        rootCanisterId,
+        identity,
+        certified,
+      }).then(),
```
lmuntaner, line 18: why this `.then()`?

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -204,7 +204,12 @@ const bonusMultiplier = ({
   amount: bigint;
   multiplier: number;
   max: number;
-}): number => 1 + multiplier * (Math.min(Number(amount), max) / max);
+}): number =>
+  1 +
+  multiplier *
+    (Math.min(Number(amount), max) /
+      // to avoid NaN
```
lmuntaner, line 211: Is this because `max` can be `undefined`?

frontend/src/tests/lib/components/sns-neurons/ConfirmSnsDissolveDelay.spec.ts
```diff
@@ -4,25 +4,62 @@
...
+        DissolveDelaySeconds: 0n,
+      },
+    ],
+  };
+
+  // freeze time
+  beforeAll(() => {
+    jest.useFakeTimers().setSystemTime(Date.now());
+
+    page.mock({ data: { universe: mockPrincipal.toText() } });
```
lmuntaner, line 43: Why is this needed? Aren't you passing everything needed as props?

frontend/src/tests/lib/components/sns-neurons/SetSnsDissolveDelay.spec.ts
```diff
@@ -4,25 +4,62 @@
...
+        DissolveDelaySeconds: 0n,
+      },
+    ],
+  };
+
+  // freeze time
+  beforeAll(() => {
+    jest.useFakeTimers().setSystemTime(Date.now());
+
+    page.mock({ data: { universe: mockPrincipal.toText() } });
```
lmuntaner, line 43: Ditto on why is this needed?

lmuntaner (commented):
A few comments, but looking very good!

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
mstrasinskis: Thanks, let's discuss this afterwards.

frontend/src/lib/utils/neuron.utils.ts
mstrasinskis: To not spoil the multiplier when the maximum is 0. For example in the backend unit test `max_neuron_age_for_age_bonus: [0n]`.

frontend/src/tests/lib/components/sns-neurons/ConfirmSnsDissolveDelay.spec.ts
mstrasinskis: Good catch, was used in prev version. Removed!

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: good point!

mstrasinskis: backend test -- https://gitlab.com/dfinity-lab/public/ic/-/blob/master/rs/sns/governance/src/neuron.rs#L764
fe test -- https://github.com/dfinity/nns-dapp/pull/1549/files#diff-ead8fd45bee5d22e63fb54b26f546cd5598bdbdb72576d3af2c48c5288feca9bR1024

frontend/src/lib/services/sns-parameters.services.ts
mstrasinskis: Good catch, redundant in that case.

lmuntaner (approved):
Thanks!

PR #1550, opened 2022-11-21
Fix: Decode II proposal payload's arg

lmuntaner:
# Motivation

The payload of the proposal when upgrading a canister has an arg property. This property is a candid blob that in the case of Internet Identity is important to deserialize to be able to properly validate the proposal.

# Changes

* Create a helper that given the value of "arg" and the canister id returns a json representation of it.
* In ProposalProposerPayloadEntry.svelte expand the object to parse any string value that is JSON.
* Util "expandObject" that parses JSON strings of an object.
* Render "null" instead of "NULL" for undefined properties in payloads. For consistency with "null" in JSON stringify.
* Json component uses "stringifyJson" util instead of it's own "stringify".
* CSS Style changes in Json component.

# Tests

* Test new use cases for ProposalProposerPayloadEntry.
* Test new util expandObject.
* Fix broken tests after some changes.

### DIFF
--- a/frontend/src/lib/components/common/Json.svelte
+++ b/frontend/src/lib/components/common/Json.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { i18n } from "$lib/stores/i18n";
-  import { bytesToHexString, isHash } from "$lib/utils/utils";
+  import { isHash, stringifyJson } from "$lib/utils/utils";
   import { isPrincipal } from "$lib/utils/utils";
 
   export let json: unknown | undefined = undefined;
@@ -27,38 +27,6 @@
     if (Array.isArray(json) && isHash(json)) return "hash";
     return typeof value;
   };
-  const stringify = (value: unknown): string | object => {
-    switch (typeof value) {
-      case "object": {
-        if (value === null) {
-          return "null";
-        }
-        // Represent Principals as strings rather than as byte arrays when serializing to JSON strings
-        if (isPrincipal(value)) {
-          const asText = value.toString();
-          if (asText !== "[object Object]") {
-            // To not stringify NOT-Principal object that contains _isPrincipal field
-            return `"${asText}"`;
-          }
-        }
-        // optimistic hash stringifying
-        if (Array.isArray(value) && isHash(value)) {
-          return bytesToHexString(value);
-        }
-        return value;
-      }
-      case "string":
-        return `"${value}"`;
-      case "bigint":
-        return value.toString();
-      case "function":
-        return "f () { ... }";
-      case "symbol":
-        return value.toString();
-      default:
-        return `${value}`;
-    }
-  };
 
   let valueType: ValueType;
   let value: unknown;
@@ -74,7 +42,7 @@
   $: {
     valueType = getValueType(json);
     isExpandable = valueType === "object";
-    value = isExpandable ? json : stringify(json);
+    value = isExpandable ? json : stringifyJson(json);
     keyLabel = `${_key}${_key.length > 0 ? ": " : ""}`;
     children = isExpandable ? Object.entries(json as object) : [];
     hasChildren = children.length > 0;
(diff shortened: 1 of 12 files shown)


lmuntaner:
@bitdivine @peterpeterparker please review

lmuntaner:
New UI:

Candid argument is parsed and shown as an object.

For II, it will show the names.

Screenshot 2022-11-28 at 14 11 10

Other times (no use case yet), we'll show numbers as property names.

lmuntaner:
New UI:

`undefined` properties are rendered as "undefined", not as NULL (agreed with Maria)
Screenshot 2022-11-28 at 14 39 28

frontend/src/lib/components/common/Json.svelte
```diff
@@ -27,38 +27,6 @@
     if (Array.isArray(json) && isHash(json)) return "hash";
     return typeof value;
   };
-  const stringify = (value: unknown): string | object => {
-    switch (typeof value) {
-      case "object": {
-        if (value === null) {
-          return "null";
-        }
-        // Represent Principals as strings rather than as byte arrays when serializing to JSON strings
```
peterpeterparker, line 36: we used to "Represent Principals as strings rather than as byte arrays when serializing to JSON strings". Aren't we loosing this feature in this PR?

frontend/src/lib/components/common/Json.svelte
```diff
@@ -27,38 +27,6 @@
...
-          return bytesToHexString(value);
-        }
-        return value;
-      }
-      case "string":
-        return `"${value}"`;
-      case "bigint":
-        return value.toString();
-      case "function":
-        return "f () { ... }";
```
peterpeterparker, line 55: dito

frontend/src/lib/components/common/Json.svelte
lmuntaner: The same logic is in the util:

```
          // Represent Principals as strings rather than as byte arrays when serializing to JSON strings
          if (isPrincipal(value)) {
            const asText = value.toString();
            // To not stringify NOT Principal instance that contains _isPrincipal field
            return asText === "[object Object]" ? value : asText;
          }
```

frontend/src/lib/components/common/Json.svelte
lmuntaner: Ditto:
```
        case "function":
          return "f () { ... }";
```

lmuntaner:

frontend/src/lib/components/common/Json.svelte
peterpeterparker: Ah cool.

In the util we don't have the check for `null` value if object as it was in `Json.svelte`. Should we add it?

```
switch (typeof value) {
      case "object": {
        if (value === null) {
          return "null";
        }
...

File: frontend/src/lib/components/proposal-detail/ProposalProposerPayloadEntry.svelte [hunk 3]
```diff
@@ -7,11 +7,17 @@
...
   import { getNnsFunctionKey } from "$lib/utils/proposals.utils";
+  import { expandObject } from "$lib/utils/utils";
 
   export let proposalId: ProposalId | undefined;
   export let proposal: Proposal | undefined;
 
   let payload: object | undefined | null;
+  let expandedPayload: object | undefined | null;
+  $: expandedPayload =
+    payload === undefined || payload === null
```
peterpeterparker, line 18: `nonNullish` ?

peterpeterparker (approved):
LGTM, thanks

frontend/src/lib/components/proposal-detail/ProposalProposerPayloadEntry.svelte
lmuntaner: sure thanks!

frontend/src/lib/components/common/Json.svelte
peterpeterparker: Coolio, thanks for the feedback

PR #1552, opened 2022-11-22
GIX-1111: Render SNS Topics

lmuntaner:
# Motivation

List the topics that the user can follow for an SNS project.

# Changes

* Rename FollowTopicSection to FollowNnsTopicSection
* New component FollowTopicSection agnostic to NNS or SNS that renders the UI of a topic.
* Implement FollowSnsNeuronsModal. Access snsFunctions and render a FollowTopicSection for each followable section.
* New sns neuron util "functionsToFollow".

# Tests

* Add test cases for FollowSnsNeuronsModal.
* Test new component FollowTopicSection

### DIFF
--- a/frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
+++ b/frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
@@ -18,8 +18,6 @@
   $: title = $i18n.follow_neurons[`topic_${topic}_title`];
   let subtitle: string;
   $: subtitle = $i18n.follow_neurons[`topic_${topic}_subtitle`];
-  let id: string;
-  $: id = Topic[topic];
 
   let showNewFolloweeModal = false;
   type FolloweeData = {
@@ -67,7 +65,7 @@
   count={followees.length}
 >
   <ul>
-    {#each followees as followee}
+    {#each followees as followee (followee.neuronId)}
       <li data-tid="current-followee-item">
         <KeyValuePair>
           <p slot="key" class="value">{followee.name ?? followee.neuronId}</p>
--- a/frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
+++ b/frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
@@ -59,11 +59,11 @@
 
 <FollowTopicSection
   on:nnsOpen={openNewFolloweeModal}
-  {title}
-  {subtitle}
   id={String(topic)}
   count={followees.length}
 >
+  <h3 slot="title">{title}</h3>
+  <p slot="subtitle" class="subtitle description">{subtitle}</p>
   <ul>
     {#each followees as followee (followee.neuronId)}
       <li data-tid="current-followee-item">
@@ -86,5 +86,14 @@
   ul {
     list-style-type: none;
     padding: 0;
+  }
+
+  h3 {
+    // Titles longer than one line had too much space with the default line-height for h3
+    line-height: normal;
+  }
+
+  .subtitle {
+    margin: 0 0 var(--padding) 0;
   }
 </style>
--- a/frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
+++ b/frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   // Tested in EditFollowNeurons.spec.ts
-  import { type NeuronId, Topic, type NeuronInfo } from "@dfinity/nns";
+  import type { NeuronId, Topic, NeuronInfo } from "@dfinity/nns";
   import NewFolloweeModal from "$lib/modals/neurons/NewFolloweeModal.svelte";
   import { removeFollowee } from "$lib/services/neurons.services";
   import { startBusy, stopBusy } from "$lib/stores/busy.store";
--- a/frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
+++ b/frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
@@ -1,0 +1,92 @@
+<script lang="ts">
+  // Tested in EditFollowNeurons.spec.ts
+  import { type NeuronId, Topic, type NeuronInfo } from "@dfinity/nns";
+  import NewFolloweeModal from "$lib/modals/neurons/NewFolloweeModal.svelte";
+  import { removeFollowee } from "$lib/services/neurons.services";
+  import { startBusy, stopBusy } from "$lib/stores/busy.store";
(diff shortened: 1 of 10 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/203274410-1dd191e7-0967-4787-b696-60952cb52b30.mov

New UI:

Screenshot 2022-11-22 at 09 32 24

lmuntaner:

Suggested order of review:

1. `FollowNnsTopicSection.svelte` (you can skip the logic) check how it renders a `FollowTopicSection.svelte`
2. `FollowTopicSection.svelte` (it's the same render as it was in FollowNnsTopicSection.
3. `FollowSnsNeuronsModal.svelte` renders a FollowNnsTopicSection per function, same as EditFollowNeurons does per Topic.
4. `sns-neurons.utils.ts` has a new util used in FollowSnsNeuronsModal to filter out functions we don't want to show in the UI.

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
```diff
@@ -0,0 +1,92 @@
...
+  export let topic: Topic;
+  export let neuron: NeuronInfo;
+
+  // TODO: Align with `en.governance.json` "topics.[topic]"
+  let title: string;
+  $: title = $i18n.follow_neurons[`topic_${topic}_title`];
+  let subtitle: string;
+  $: subtitle = $i18n.follow_neurons[`topic_${topic}_subtitle`];
+  let id: string;
+  $: id = Topic[topic];
```
peterpeterparker, line 22: `id` seems to be not used

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -289,3 +290,16 @@ export const needsRefresh = ({
...
+ *
+ * For now it filters out only the UNSPECIFIED function.
+ * https://github.com/dfinity/ic/blob/5248f11c18ca564881bbb82a4eb6915efb7ca62f/rs/sns/governance/proto/ic_sns_governance/pb/v1/governance.proto#L582
+ *
+ */
+const UNSPECIFIED_FUNCTION_ID = BigInt(0);
+export const functionsToFollow = (
+  functions: SnsNervousSystemFunction[] | undefined
+): SnsNervousSystemFunction[] | undefined =>
+  functions?.filter((f) => f.id !== UNSPECIFIED_FUNCTION_ID);
```
peterpeterparker, line 305: - should `UNSPECIFIED_FUNCTION_ID` be a constants - i.e. not find place in a utils?

- nth of nth of nth of nth: You know how I am, I would not be against deconstructing field 😉 `functions?.filter(({id}) => id !== UNSPECIFIED_FUNCTION_ID`

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
```diff
@@ -0,0 +1,92 @@
...
+    await removeFollowee({
+      neuronId: neuron.neuronId,
+      topic,
+      followee,
+    });
+    stopBusy("remove-followee");
+  };
+</script>
+
+<FollowTopicSection
```
peterpeterparker, line 62: Is there a reason why `title`, `subtitle` and `count` are properties and not slot?

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
```diff
@@ -0,0 +1,92 @@
...
+      </li>
+    {/each}
+  </ul>
+</FollowTopicSection>
+{#if showNewFolloweeModal}
+  <NewFolloweeModal {neuron} {topic} on:nnsClose={closeNewFolloweeModal} />
+{/if}
+
+<style lang="scss">
+  ul {
```
peterpeterparker, line 88: not related to your PR: we should add a CSS utility class for this in gix cmp, e.g. `<ul class="none"/>` generally available

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
lmuntaner: good catch, yes. I wonder why eslint didn't catch this.

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
```diff
@@ -0,0 +1,92 @@
...
+
+<FollowTopicSection
+  on:nnsOpen={openNewFolloweeModal}
+  {title}
+  {subtitle}
+  id={String(topic)}
+  count={followees.length}
+>
+  <ul>
+    {#each followees as followee}
```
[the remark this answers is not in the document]
lmuntaner: yes

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
lmuntaner: Count is inside a badge, and that one makes sense to always have it as number.

The other two could also be slots. But since I'm keeping exactly the same UI for NNS and SNS, I thought it was more straightforward to pass them as props. But I don't have a strong opinion on those.

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
lmuntaner: To be honest I'm not a fan of global classes on component based projects. I like my components to be as independent as possible.

But if there was such a class and since this is the projects pattern, I would use it. Yes.

I would even prefer to reset all elements globally to have 0 margin and 0 padding by default and set them when needed.

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: yes

lmuntaner:
@peterpeterparker ready for a review

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
peterpeterparker: Thanks for the feedback.

For consistancy reason, I would tend to prefer that we use `slot` for title and subtitle. Spontaneously I am thinking that we most often pass down information that are displayed only through slots. It's also closer to native HTML semantic `<p text="something"/>` vs `<p>something</p>`. But for sure your call, I can live with these as properties

`Count` no question, good point.

peterpeterparker (approved):
LGTM, thanks.

As mentionned your call regarding slot or not slots.

frontend/src/lib/components/neurons/FollowNnsTopicSection.svelte
lmuntaner: done, makes sense

PR #1555, opened 2022-11-23
Chore: Upgrade proposal payload revisions

lmuntaner:
# Motivation

A new property was added to a payload and the crypto team needs it for a proposal.

# Changes

* Upgrade the commit of registry-canister dependency to the one deployed 2022-11-23

# Tests

Not applicable.

### DIFF
--- a/rs/Cargo.toml
+++ b/rs/Cargo.toml
@@ -47,7 +47,8 @@ ic-sns-swap = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcd
 ic-sns-wasm = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcdb82486deb7aaa25d768787" }
 ledger-canister = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcdb82486deb7aaa25d768787" }
 on_wire = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcdb82486deb7aaa25d768787" }
-registry-canister = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcdb82486deb7aaa25d768787" }
+# Registry canister deploy as of 2022-11-23, ahead of Blessed Replica Version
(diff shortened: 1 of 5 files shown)


rs/Cargo.toml
```diff
@@ -47,7 +47,8 @@ ic-sns-swap = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcd
 ic-sns-wasm = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcdb82486deb7aaa25d768787" }
 ledger-canister = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcdb82486deb7aaa25d768787" }
 on_wire = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcdb82486deb7aaa25d768787" }
-registry-canister = { git = "https://github.com/dfinity/ic", rev = "a3608ed7cf64a01dcdb82486deb7aaa25d768787" }
+# Registry canister deploy as of 2022-11-23, ahead of Blessed Replica Version
+registry-canister = { git = "https://github.com/dfinity/ic", rev = "0a729806f2fbc717f2183b07efac19f24f32e717" }
```
bitdivine, line 51: The commit is also in URLs in `rs/src/proposals.rs` e.g. :
```
    // NNS function 36 - RetireReplicaVersion
    // https://github.com/dfinity/ic/blob/c2ad499466967a9a5557d737c2b9c0b9fa8ad53f/rs/registry/canister/src/mutations/do_retire_replica_version.rs#L143
    pub type RetireReplicaVersionPayload =
        registry_canister::mutations::do_retire_replica_version::RetireReplicaVersionPayload;
```
We might want to update those, although they aren't up to date in main.

bitdivine: I have confirmed that the only change in the did file is the addition of a field in two places, which matches what I understand you expect.  The field is:
```
  idkg_key_rotation_period_ms : opt nat64;
```

bitdivine (approved):

rs/Cargo.toml
lmuntaner: thanks, I'll change it.

PR #1557, opened 2022-11-23
feat: redirect user to login screen if offline

peterpeterparker:
# Motivation

Automatic sign-out might happen while user device is idle and therefore - according the error message we noticed - the network might be off.

That's why, to prevent the app to try to fetch public data and display errors, we can redirect the user to the login page which is static and will not perform.

# Changes

- if `!navigator.onLine` on automatic signout, redirect to origin
- move `Toast` to main layout so that toats can be displayed on login page too

# Screenshot

The issue we noticed.

image

Capture d’écran 2022-11-23 à 11 09 59

Capture d’écran 2022-11-23 à 11 10 33
Capture d’écran 2022-11-23 à 11 10 43


lmuntaner:
@peterpeterparker shall we close this one if it's not blocking the release?

peterpeterparker:
@lmuntaner sure was just waiting to close it with a comment. Will comment later then.

PR #1560, opened 2022-11-23
feat: wallet summary design

peterpeterparker:
# Motivation

Review design of the wallet / transactions list view for Nns and Sns.

# PRs

- [x] https://github.com/dfinity/gix-components/pull/132
- [x] https://github.com/dfinity/gix-components/pull/134

# Changes

- add style according design
- use new gix cmp `<Island />`
- move modals from sub-components to top component - outside `Island` - because new design force the `overflow` property of the content to be hidden (🙈)
- few buttons changed from `primary` to `secondary`
- two unused mixins removed
- new `<ColumnRow />` to display content as column on small device and row on larger device
- dates format modified from long (`November 24 ...`)  to (`Nov 24...`) globally as in design (reviewed with Mischa)

# Note

This PR does not contain features changes except modifying the display of the date as mentionned above.

# Screenshots

Capture d’écran 2022-11-25 à 08 11 58
Capture d’écran 2022-11-25 à 08 12 07
Capture d’écran 2022-11-25 à 08 12 10
Capture d’écran 2022-11-25 à 08 12 16
Capture d’écran 2022-11-25 à 08 12 12
Capture d’écran 2022-11-25 à 08 12 01


frontend/src/lib/modals/accounts/WalletModals.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
+  import { walletModal } from "$lib/stores/modal.store";
+  import RenameSubAccountModal from "$lib/modals/accounts/RenameSubAccountModal.svelte";
+  import HardwareWalletListNeuronsModal from "$lib/modals/accounts/HardwareWalletListNeuronsModal.svelte";
+
+  const close = () => walletModal.set(undefined);
```
lmuntaner, line 6: Should we consider having a global store for modals?

One that also accepts two open at the same time btw.

Although this might be a problem if we need to pass props to those modals.

frontend/src/tests/lib/components/accounts/WalletActionsTest.svelte
```diff
@@ -0,0 +1,35 @@
+<script lang="ts">
```
lmuntaner, line 1: It was confusing for me the name `WalletActionsTest`. How about `WalletContextTest`?

frontend/src/tests/mocks/wallet.mock.ts
```diff
@@ -0,0 +1,29 @@
...
+  SELECTED_ACCOUNT_CONTEXT_KEY,
+  type SelectedAccountContext,
+  type SelectedAccountStore,
+} from "$lib/types/selected-account.context";
+import { render } from "@testing-library/svelte";
+import type { SvelteComponent } from "svelte";
+import { writable } from "svelte/store";
+import ContextWrapperTest from "../lib/components/ContextWrapperTest.svelte";
+
+export const renderWalletActions = ({
```
lmuntaner, line 12: ditto `renderWalletContext`

lmuntaner (approved):
A few comments, but already approving. LGTM!

frontend/src/lib/modals/accounts/WalletModals.svelte
peterpeterparker: I named it `walletModal` because these are tight with the particular view since they need the context - i.e. they need to be a child of Wallet.svelte. Therefore not sure it should be a global store.

Regarding two modals why not, I have to say I personally don't have to have modals over modals in my design but no strong opinion neither. Also not my design call 😉

frontend/src/tests/lib/components/accounts/WalletActionsTest.svelte
peterpeterparker: Sure

PR #1561, opened 2022-11-23
Fix: Error message not canister controller

lmuntaner:
# Motivation

Users sees a generic error message when he or she is not the controller of the canister.

The new error message is:

```
Server returned an error:
  Code: 403 ()
  Body: Only controllers of canister wxns6-qiaaa-aaaaa-aaaqa-cai can call ic00 method canister_status
  Retrying request.
```

I believe they added the `()` and that messed up with the current parsing method.

# Changes

* Change the way we parse the error message.

# Tests

Updated tests with new error message.


lmuntaner:

peterpeterparker (approved):
LGTM, thx

PR #1562, opened 2022-11-23
Fix: Proposals not shown

lmuntaner:
# Motivation

The filter to show only the ones you can vote on was being used even though the user was not logged in.

Therefore, logged out users that had that turned on had an empty list of proposals.

# Changes

* Pass the identity to the proposals utils to check whether it's present or not before hiding proposals.
* Load proposals when auth store changes.

# Tests

* Add test cases for new scenario.
* Fix broken tests because of new API.

### DIFF
--- a/frontend/src/lib/derived/proposals.derived.ts
+++ b/frontend/src/lib/derived/proposals.derived.ts
@@ -1,3 +1,4 @@
+import { authStore } from "$lib/stores/auth.store";
 import { definedNeuronsStore } from "$lib/stores/neurons.store";
 import type {
   ProposalsFiltersStore,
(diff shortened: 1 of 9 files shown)


frontend/src/lib/derived/proposals.derived.ts
```diff
@@ -75,11 +80,24 @@ export const uiProposals: Readable<UIProposalsStore> = derived(
...
     voteRegistrationStore,
+    authStore,
   ],
-  ([{ proposals, certified }, filters, neurons, { registrations }]) => ({
+  ([
+    { proposals, certified },
+    filters,
+    neurons,
+    { registrations },
+    $authStore,
```
peterpeterparker, line 90: you can deconstruct `{identity}` here

lmuntaner: yes, done

lmuntaner:

peterpeterparker (approved):
LGTM, thanks a tone!

PR #1565, opened 2022-11-24
GIX-1113: Add SNS Neuron Followee

lmuntaner:
# Motivation

User can add followees to SNS neurons.

NOTE: Followees are still not rendered.

# Changes

* FollowsSnsNeuronModal uses a new component: FollowSnsTopicSection to render each Topic.
* FollowSnsTopicSection uses a new Component NewSnsFolloweeModal to created the new followee.
* NewSnsFolloweeModal uses a new sns-neuron service addFollowee.
* In sns-neurons.services, addFollowe uses new apis "setFollowees" and "querySndNeuron" and a new sns-neuron utils "followeesByFunction"
* In sns.api.ts, rename "querySnsNeuron" to "getSnsNeuron". This api raises an error if the neuron is not found. The one we use so far.
* In sns.api.ts, new api "querySnsNeuron" that returns `undefined` if neuron does not exist. Useful to check easily whether a followee exists.
* In sns-governance.api.ts, New api "setFollowees".

# Tests

* Test new component FollowSnsTopicSection.
* Test new modal NewSnsFolloweeModal
* Test new service addFollowee with different scenarios.
* Test new apis: "querySnsNeuron" and "setFollowees"

### DIFF
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -315,7 +315,7 @@ export const querySnsNeurons = async ({
   return neurons;
 };
 
-export const querySnsNeuron = async ({
+export const getSnsNeuron = async ({
   identity,
   rootCanisterId,
   certified,
@@ -333,6 +333,31 @@ export const querySnsNeuron = async ({
     certified,
   });
   const neuron = await getNeuron({
+    neuronId,
+  });
+
+  logWithTimestamp("Getting sns neuron: done");
+  return neuron;
+};
+
+export const querySnsNeuron = async ({
+  identity,
+  rootCanisterId,
+  certified,
+  neuronId,
+}: {
+  identity: Identity;
+  rootCanisterId: Principal;
+  certified: boolean;
+  neuronId: SnsNeuronId;
+}): Promise<SnsNeuron | undefined> => {
+  logWithTimestamp("Querying sns neuron: call...");
+  const { queryNeuron } = await wrapper({
+    identity,
+    rootCanisterId: rootCanisterId.toText(),
+    certified,
+  });
+  const neuron = await queryNeuron({
     neuronId,
   });
 
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -315,6 +315,9 @@ export const querySnsNeurons = async ({
   return neurons;
 };
 
+/**
+ * Returns the neuron or raises an error if not found.
+ */
 export const getSnsNeuron = async ({
   identity,
   rootCanisterId,
@@ -340,6 +343,9 @@ export const getSnsNeuron = async ({
   return neuron;
 };
 
+/**
+ * Returns the neuron or undefined.
+ */
 export const querySnsNeuron = async ({
   identity,
   rootCanisterId,
--- a/frontend/src/lib/components/sns-neuron-detail/FollowSnsTopicSection.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/FollowSnsTopicSection.svelte
@@ -37,11 +37,6 @@
 {/if}
 
 <style lang="scss">
-  h3 {
-    // Titles longer than one line had too much space with the default line-height for h3
-    line-height: normal;
-  }
-
   .subtitle {
     margin: 0 0 var(--padding) 0;
   }
--- a/frontend/src/lib/components/sns-neuron-detail/FollowSnsTopicSection.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/FollowSnsTopicSection.svelte
@@ -1,0 +1,48 @@
+<script lang="ts">
+  import FollowTopicSection from "$lib/components/neurons/FollowTopicSection.svelte";
+  import NewSnsFolloweeModal from "$lib/modals/sns/NewSnsFolloweeModal.svelte";
+  import type { Principal } from "@dfinity/principal";
+  import type { SnsNeuron, SnsNervousSystemFunction } from "@dfinity/sns";
+  import { fromNullable } from "@dfinity/utils";
+
+  export let neuron: SnsNeuron;
+  export let rootCanisterId: Principal;
+  export let nsFunction: SnsNervousSystemFunction;
+
+  let showModal = false;
+  const openModal = () => (showModal = true);
+  const closeModal = () => (showModal = false);
+</script>
+
+<FollowTopicSection
+  on:nnsOpen={openModal}
+  count={0}
+  id={nsFunction.id.toString()}
+>
+  <h3 slot="title">{nsFunction.name}</h3>
+  <p slot="subtitle" class="subtitle description">
+    {fromNullable(nsFunction.description)}
+  </p>
+  <!-- TODO: Render Followees https://dfinity.atlassian.net/browse/GIX-1114 -->
+  <div>{`TODO: render followees ${neuron.followees.length}`}</div>
+</FollowTopicSection>
+
+{#if showModal}
+  <NewSnsFolloweeModal
+    {rootCanisterId}
+    {neuron}
+    functionId={nsFunction.id}
+    on:nnsClose={closeModal}
+  />
+{/if}
+
+<style lang="scss">
+  h3 {
+    // Titles longer than one line had too much space with the default line-height for h3
+    line-height: normal;
+  }
+
+  .subtitle {
+    margin: 0 0 var(--padding) 0;
+  }
+</style>
--- a/frontend/src/lib/modals/sns/NewSnsFolloweeModal.svelte
+++ b/frontend/src/lib/modals/sns/NewSnsFolloweeModal.svelte
@@ -1,0 +1,71 @@
+<script lang="ts">
+  import Input from "$lib/components/ui/Input.svelte";
+  import { addFollowee } from "$lib/services/sns-neurons.services";
+  import { i18n } from "$lib/stores/i18n";
+  import {
+    SELECTED_SNS_NEURON_CONTEXT_KEY,
+    type SelectedSnsNeuronContext,
+  } from "$lib/types/sns-neuron-detail.context";
+  import { hexStringToBytes } from "$lib/utils/utils";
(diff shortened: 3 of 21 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/203725490-c9267a59-a8dd-4205-aa7b-bad0b06c9969.mov

lmuntaner:

The "changes" section in the description is also the recommended order to review them.

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -443,3 +448,97 @@ export const loadSnsNervousSystemFunctions = async (
...
+      ({ id }) => subaccountToHexString(id) === followeeHex
+    ) !== undefined
+  ) {
+    toastsError({
+      labelKey: "new_followee.already_followed",
+    });
+    return { success: false };
+  }
+  try {
+    const followeeNeuron = await querySnsNeuron({
```
lmuntaner, line 507: We should do this also for NNS.

frontend/src/lib/api/sns.api.ts
```diff
@@ -315,7 +315,7 @@ export const querySnsNeurons = async ({
   return neurons;
 };
 
-export const querySnsNeuron = async ({
+export const getSnsNeuron = async ({
```
lmuntaner, line 318: This is the one used everywhere.

frontend/src/lib/api/sns.api.ts
```diff
@@ -340,6 +340,31 @@ export const querySnsNeuron = async ({
   return neuron;
 };
 
+export const querySnsNeuron = async ({
```
lmuntaner, line 343: Used only in addFollowee service

peterpeterparker:
This button (see screenshot) should most probably not be full-width.

Capture d’écran 2022-11-25 à 07 01 12

frontend/src/lib/api/sns.api.ts
peterpeterparker: should we add a comment about the difference between `get` vs `query` neuron? Spontaneously I would not know which one I have to use if I want to info of one neuron 🤷‍♂️

frontend/src/lib/components/sns-neuron-detail/FollowSnsTopicSection.svelte
```diff
@@ -0,0 +1,48 @@
+<script lang="ts">
+  import FollowTopicSection from "$lib/components/neurons/FollowTopicSection.svelte";
+  import NewSnsFolloweeModal from "$lib/modals/sns/NewSnsFolloweeModal.svelte";
+  import type { Principal } from "@dfinity/principal";
+  import type { SnsNeuron, SnsNervousSystemFunction } from "@dfinity/sns";
+  import { fromNullable } from "@dfinity/utils";
+
+  export let neuron: SnsNeuron;
+  export let rootCanisterId: Principal;
+  export let nsFunction: SnsNervousSystemFunction;
```
peterpeterparker, line 10: Not for this PR: I really like that you prefixed here with `ns`. If you would refactor existing `functions` variables we already have in the code with `nsFunctions` that would be neat.

frontend/src/lib/components/sns-neuron-detail/FollowSnsTopicSection.svelte
```diff
@@ -0,0 +1,48 @@
...
+    {neuron}
+    functionId={nsFunction.id}
+    on:nnsClose={closeModal}
+  />
+{/if}
+
+<style lang="scss">
+  h3 {
+    // Titles longer than one line had too much space with the default line-height for h3
+    line-height: normal;
```
peterpeterparker, line 42: u can use a `var(--line-height...)` here instead

peterpeterparker: there is also mixing like `@include fonts.h3` if for example you want to apply the h3 style to another type of element like p or span

frontend/src/lib/modals/sns/NewSnsFolloweeModal.svelte
```diff
@@ -0,0 +1,69 @@
...
+    <Input
+      inputType="text"
+      autocomplete="off"
+      placeholderLabelKey="new_followee.placeholder"
+      name="new-followee-id"
+      bind:value={followeeHex}
+    >
+      <svelte:fragment slot="label">{$i18n.new_followee.label}</svelte:fragment>
+    </Input>
+    <button
```
peterpeterparker, line 60: I'm comparing with Nns, we have the notion of `isControllableByUser` and `isControllableByHotkey` there. This does not exist in Sns?

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -443,3 +448,97 @@ export const loadSnsNervousSystemFunctions = async (
...
+
+  const identity = await getNeuronIdentity();
+  const followee: SnsNeuronId = {
+    id: arrayOfNumberToUint8Array(hexStringToBytes(followeeHex)),
+  };
+
+  const topicFollowees = followeesByFunction({ neuron, functionId });
+  // Do not allow to add a neuron id who is already followed
+  if (
+    topicFollowees !== undefined &&
```
peterpeterparker, line 496: nth: `topicFollowees?.find(` would work?

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -443,3 +448,97 @@ export const loadSnsNervousSystemFunctions = async (
...
+      toastsError({
+        labelKey: "new_followee.followee_does_not_exist",
+        substitutions: {
+          $neuronId: followeeHex,
+        },
+      });
+      return { success: false };
+    }
+
+    const newFollowees: SnsNeuronId[] =
```
peterpeterparker, line 523: should we use a `Set` conversion to ensure the followee isn't twice in the array?

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -443,3 +448,97 @@ export const loadSnsNervousSystemFunctions = async (
...
+      topicFollowees === undefined ? [followee] : [...topicFollowees, followee];
+
+    await setFollowees({
+      rootCanisterId,
+      identity,
+      // We can cast it because we already checked that the neuron id is not undefined
+      neuronId: fromNullable(neuron.id) as SnsNeuronId,
```
peterpeterparker, line 530: nth: would `fromNullable<SnsNeuronId>(...` would work instead of the cast?

frontend/src/lib/utils/error.utils.ts
```diff
@@ -28,6 +29,8 @@ export const errorToString = (err?: unknown): string | undefined => {
       ? (err as string)
       : err instanceof GovernanceError
       ? (err as GovernanceError)?.detail?.error_message
+      : err instanceof SnsGovernanceError
```
peterpeterparker, line 32: you need this? i.e. `SnsGovernanceError` does not extends `Error` and per extension line `err instanceof Error` is also true for these?

peterpeterparker (approved):
LGTM, minor comments

frontend/src/lib/modals/sns/NewSnsFolloweeModal.svelte
lmuntaner: No, in sns the permissions are fine grained. The user have permission to vote, to dissolve, to submit proposal, ...

Therefore, we need to check `hasPermissionToVote`.

This modal should not open unless the user has the permission. But I can add it here as well. We have:

```
{#if allowedToManageFollows && nonNullish(neuron) && nonNullish(rootCanisterId)}
    <div class="actions">
      <FollowSnsNeuronsButton {neuron} {rootCanisterId} />
    </div>
  {/if}
```

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: `SnsNeuronId` is an object. I would need to convert that to a string, then make the set, the convert it back to the object. Not sure that makes it better...

lmuntaner:
@peterpeterparker regarding the width of the button. The current NNS modal has it full-width. But that's because there is the list of known neurons below.

It's a good point that without the list of known neurons, we should have the button like the others. I'll change it. Thanks!

frontend/src/lib/utils/error.utils.ts
lmuntaner: It does because the api is the same. But I wanted to specifically handle the type here in case we ever change the api of the SnsGovernanceError.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: no 😞

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: yes, good point!

frontend/src/lib/components/sns-neuron-detail/FollowSnsTopicSection.svelte
lmuntaner: Sounds good

frontend/src/lib/api/sns.api.ts
lmuntaner: I'll add it.

But I also wanted to check with you and maks.

We should use carefully these prefixes:

* queryXXX (undefined | item). Used for multiple as well that return empty array if no items.
* getXXX (raise error | undefined)
* loadXXX (purely loads in store)
* syncXXX (does some checks and loads in store), for example syncAccounts

What do you think?

frontend/src/lib/components/sns-neuron-detail/FollowSnsTopicSection.svelte
lmuntaner: I didn't need any rule actually. Multiple lines and everything looked good without specific rules.

I guess the new design fixed the issue I had before.

lmuntaner:
@peterpeterparker new UI:

Screenshot 2022-11-25 at 10 23 58

PR #1566, opened 2022-11-24
test: reload filters

peterpeterparker:
# Motivation

I literally dreamed of that missing test this night (😅) so let's add a test for the new proposal filter `reload()` function.

# Changes

- test `reload` store and call


lmuntaner (approved):
😂 LGTM!

PR #1573, opened 2022-11-25
feat: update neuron detail page title on neuron id intersection

peterpeterparker:
# Motivation

Following release of the new design, two users complained that the neuron detail page title repeated the neuron id that is already included in the page (source [forum](https://forum.dfinity.org/t/nns-dapp-ui-design-upgrade-and-public-proposals/16801/9)).

There is a good reason to repeast the neuron id in the page title, notably on mobile devices, because when the user scrolls, it's a good placeholder to remains the user what neuron is currently handling.

Therefore this PR make the title dynamic - i.e. if the neuron id within the content is displayed, the title will display "Neuron, if the user scrolls and the neuron id disappear from the viewport, then title is updated with neuron id. 

# Changes

- introduces new `intersection` directive that leverage the `IntersectionObserver` API
- use directive and upon intersection, display title as explained above
- rename `style.constant` to `layout.constant`

# Screenshot

gif


frontend/src/lib/directives/intersection.directives.ts
```diff
@@ -0,0 +1,46 @@
...
+  intersecting: boolean;
+}) => {
+  const $event = new CustomEvent("nnsIntersecting", {
+    detail: { intersecting },
+    bubbles: false,
+  });
+  element.dispatchEvent($event);
+};
+
+export const onIntersection = (element: HTMLElement) => {
```
lmuntaner, line 18: Cool!

lmuntaner (commented):
I love it!

I just realized the SNS neuron detail has no title if you arrive with the link, and the Neurons title if you come from the list.

Does it make sense to fix that here or in another PR?

peterpeterparker:
@lmuntaner noticed it too, would be cool to fix it in another PR. can you do so?

lmuntaner (approved):

PR #1574, opened 2022-11-25
Fix lm voting power

lmuntaner:
# Motivation

Voting power calculation doesn't match the actual value of voting power.

Calculation is used in modal to set or increase dissolve delay.

The issue was that it was missing the staked maturity.

It also includes a big refactor which I hope makes it clear how the voting power is calculated.

# Changes

* Rename votingPower neuron util to neuronVotingPower and change the params. It now expects the neuron and a new dissolve delay. The new dissolve delay is useful to show the future voting power in the increase dissolve delay modal.
* neuronVotingPower uses a new function votingPower. Which expects all the params and data necessary to calculate voting power. It users the new multiplier function.
* dissolveDelayMultiplier and ageMultiplier use a new "multiplier" function.
* new multiplier helper function.
* Moved functions around to put helpers next to each other for better context.
* Remove the voting power UI in SNS modals.

# Tests

* Add some scenarios and fix tests with new api.


frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -88,26 +90,120 @@ export const stateTextMapper: StateMapper = {
...
+  neuron,
+  newDissolveDelayInSeconds,
 }: {
-  stake: bigint;
-  dissolveDelayInSeconds: number;
-  ageSeconds: number;
-}): bigint =>
-  dissolveDelayInSeconds > SECONDS_IN_HALF_YEAR
+  neuron: NeuronInfo;
+  newDissolveDelayInSeconds?: bigint;
```
mstrasinskis, line 121: nitpick: probably a not optional parameter `dissolveDelaySeconds` (and not using the neuron.`dissolveDelaySeconds` param) could simplify the function.

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -88,26 +90,120 @@ export const stateTextMapper: StateMapper = {
...
+ * It might be useful to use it for SNS neurons.
+ *
+ * @param {VotingPowerParams}
+ * @returns {bigint}
+ */
+const votingPower = ({
+  stakeE8s,
+  dissolveDelay,
+  ageSeconds,
+  ageBonusMultiplier = AGE_MULTPIPLIER,
```
mstrasinskis, line 159: A typo in AGE_MUL**TPIP**LIER

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -88,26 +90,120 @@ export const stateTextMapper: StateMapper = {
...
+}: VotingPowerParams): bigint =>
+  dissolveDelay > minDissolveDelaySeconds
     ? BigInt(
         Math.round(
-          (Number(stake) / E8S_PER_ICP) *
-            dissolveDelayMultiplier(dissolveDelayInSeconds) *
-            ageMultiplier(ageSeconds) *
-            E8S_PER_ICP
+          Number(stakeE8s) *
+            bonusMultiplier({
```
mstrasinskis, line 169: nitpick: for readability `dissolveDelayMultiplier` * `ageMultiplier`?

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -88,26 +90,120 @@ export const stateTextMapper: StateMapper = {
...
+    (neuron.fullNeuron?.cachedNeuronStake ?? BigInt(0)) +
+    (neuron.fullNeuron?.stakedMaturityE8sEquivalent ?? BigInt(0));
+  return votingPower({
+    stakeE8s,
+    dissolveDelay,
+    ageSeconds: neuron.ageSeconds,
+  });
+};
+
+interface VotingPowerParams {
```
mstrasinskis, line 135: Isn't this a part of `votingPower` function arguments because you plan to reuse the interface?

mstrasinskis (approved):
LGTM, thanks! Just a couple of minor comments/questions.

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: Yes, that doesn't make that make sense to me. Why would the dissolveDelaySeconds go separately, but not the ageSecods or the stake or the stakedMaturity?

That's what we had before with ageSeconds and we forgot to set it and that created a bug.

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: thanks!

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: sure!

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: No, just because I wanted to make the interface of the functon more readable and easily identify what is data and what are params with comments.

It was better to have that here and not in the function signature.

PR #1575, opened 2022-11-26
refactor: debug services to directives

peterpeterparker:
# Motivation

Refactor `debug.services` to `debug.directives` because we now have more than directive.

Drop `debug.services.proxy` which had no effect in terms of bundling (both "proxy" and "services" were consumed already).


lmuntaner (approved):
Thanks!

PR #1576, opened 2022-11-26
refactor: wallet modal context

peterpeterparker:
# Motivation

The modal store that had to be introduced in #1560 because of design breaking changes works but should not be global. 
Indeed, modals have to be children of the view to inherits the context, therefore the store that opens these particular modals makes more sense within the context.

# Changes

- replace `walletModal.store` with a value within the context used by the related modal
- integrate hw account context within wallet context since that particular context had to be move at the top context as well

# Note

Note functional changes in this PR.


lmuntaner (approved):
I like it, it makes sense.

PR #1577, opened 2022-11-26
FIx: Rename action OpenSnsTokenSwap

lmuntaner:
# Motivation

Users should not see the word Swap in the nns-dapp.

We want to show also the properties of the actions that are undefined. We show them as `NULL`.

# Changes

* Include `undefined` as the type of action field value that we support.
* Check when the action is OpenSnsTokenSwap and show a different title.

# Tests

* Add test for new support for `undefined` properties in action fields.
* Add test to check that ProposalProposerActionsEntry renders `NULL` for undefined properties.

### DIFF
--- a/frontend/src/lib/components/proposal-detail/ProposalProposerActionsEntry.svelte
+++ b/frontend/src/lib/components/proposal-detail/ProposalProposerActionsEntry.svelte
@@ -6,6 +6,7 @@
   import type { Proposal } from "@dfinity/nns";
   import Json from "../common/Json.svelte";
   import { KeyValuePair } from "@dfinity/gix-components";
+  import { i18n } from "$lib/stores/i18n";
 
   export let proposal: Proposal | undefined;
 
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/proposal-detail/ProposalProposerActionsEntry.svelte
```diff
@@ -6,19 +6,25 @@
...
   import { KeyValuePair } from "@dfinity/gix-components";
+  import { i18n } from "$lib/stores/i18n";
 
   export let proposal: Proposal | undefined;
 
   let actionKey: string | undefined;
   let actionFields: [string, unknown][] = [];
   $: actionKey =
     proposal !== undefined ? proposalFirstActionKey(proposal) : undefined;
+  let title: string;
```
peterpeterparker, line 17: can you open a JIRA task and add a TODO to remove this hack once the NNS team has fixed the canisters?

frontend/src/lib/components/proposal-detail/ProposalProposerActionsEntry.svelte
```diff
@@ -28,6 +34,8 @@
       <span class="value" slot="value">
         {#if typeof value === "object"}
           <Json json={value} />
+        {:else if typeof value === "undefined"}
+          {$i18n.core.null}
```
peterpeterparker, line 38: would be nice to have a test that checks that "NULL" is rendered in the Dom

peterpeterparker (approved):
Thanks. Two comments.

frontend/src/lib/components/proposal-detail/ProposalProposerActionsEntry.svelte
lmuntaner: done

frontend/src/lib/components/proposal-detail/ProposalProposerActionsEntry.svelte
lmuntaner: done

PR #1580, opened 2022-11-28
feat: update phrasing

peterpeterparker:
# Motivation

User feedback:

- "Edit source account" is not comprehensive, "Change source account" would be more accurate (aknoledged by Mischa)
- The confirmation button of the amount of cycles when creating a canister is named "Review cycles purchase" instead of "Review canister creation"


lmuntaner (approved):
Thanks!

PR #1583, opened 2022-11-28
fix: social image

peterpeterparker:
# Motivation

In this weekend [tweet](https://twitter.com/dfinity/status/1596564526639742976) of the foundation, the social image of NNS-dapp wasn't rendered. Root cause of the issue is most probably the relative path of the social image which should be an absolute path.

# Changes

- use absolute path for `og:image` and `twitter:image`
- add non raw domain off nns-dapp to csp image (ultimately we will remove raw in #1394)


frontend/src/app.html
```diff
@@ -25,7 +25,7 @@
     <meta content="website" property="og:type" />
     <meta content="https://nns.ic0.app/" property="og:url" />
     <meta
-      content="%sveltekit.assets%/assets/social-image.png"
```
lmuntaner, line 28: What is the default of `sveltekit.assets`, the relative path?

lmuntaner (approved):
LGTM, thanks!

frontend/src/app.html
peterpeterparker: Exactly, the relative path to the assets, and if you build a sub-route, like `/wallet/` it will replace it with `../asset/` in the generated html file

PR #1584, opened 2022-11-28
fix: sync time with IC to (try to) fix invalid certificate error

peterpeterparker:
# Motivation

According SDK team we can called `syncTime` of the http-agent of `agent-js` to synchronize the time of the IC in order to fix the issue "Invalide certificate" that happens time to time - quite often on my side actually.

# Notes

- We cannot reproduce the issue locally nor on testnet, therefore this is a best guess and we shall hope it will indeed fix the issue. In worse case, it won't arm.

- It will slow down a bit boot time after sign-in as an extra call is made

# Changes

- call `agent.syncTime` in public service before first initialization of data

# Screenshot

Issue:

Capture d’écran 2022-11-28 à 07 31 09


frontend/src/lib/services/$public/app.services.ts
```diff
@@ -1,16 +1,24 @@
...
+import { syncTime } from "$lib/services/$public/agent.services";
 import { loadSnsSummaries } from "$lib/services/$public/sns.services";
 import { displayAndCleanLogoutMsg } from "$lib/services/auth.services";
 import { authStore } from "$lib/stores/auth.store";
 import { layoutAuthReady } from "$lib/stores/layout.store";
 import { toastsError } from "$lib/stores/toasts.store";
 
+export const initAppPublic = async () => {
+  // Agent-js syncTime can be called during initialization or mid-lifecycle so we do it as soon as possible.
+  await syncTime();
```
lmuntaner, line 12: I guess it's important to do it initially?

Should we do it also when we init the AppPrivate?

lmuntaner (commented):
A question

frontend/src/lib/services/$public/app.services.ts
peterpeterparker: As discussed offline, ultimately we gonna call syncTime for private data loading as well.

frontend/src/lib/services/$public/app.services.ts
```diff
@@ -1,16 +1,26 @@
...
 import { ENABLE_SNS } from "$lib/constants/environment.constants";
 import { loadSnsSummaries } from "$lib/services/$public/sns.services";
+import { syncTime } from "$lib/services/agent.services";
 import { displayAndCleanLogoutMsg } from "$lib/services/auth.services";
 import { authStore } from "$lib/stores/auth.store";
 import { layoutAuthReady } from "$lib/stores/layout.store";
 import { toastsError } from "$lib/stores/toasts.store";
 
+export const initAppPublic = async () => {
+  // We first try to sync the IC time to resolve a potential issue (see syncTime documentation).
```
lmuntaner, line 11: Thanks!

lmuntaner (approved):
LGTM, thanks!

PR #1585, opened 2022-11-28
feat: neuron details page design with islands

peterpeterparker:
# Motivation

Introduces `<Island />` - as in new design - in neuron detail pages. While the design update is trivial, it needs moving all modals as top children to avoid overflow issue.

# Changes

- use `<Island />` for styling purpose
- introduce context for nns neuron page with modal
- add modal to sns neuron page context
- move movals to `<NnsNeuronModals />` and `<SnsNeuronModals />`
- adapt properties and functions accordingly
- few files move to group sns components
- spinner in `ConfirmDisburseNeuron` replace with button displayed to avoid visual glitch

# Note

This PR does not contain any feature changes.

# Screenshots

Capture d’écran 2022-11-28 à 15 02 13

Capture d’écran 2022-11-28 à 15 02 24

Capture d’écran 2022-11-28 à 15 02 17

Capture d’écran 2022-11-28 à 15 02 22


frontend/src/lib/types/nns-neuron-detail.context.ts
```diff
@@ -0,0 +1,28 @@
+import type { NeuronInfo } from "@dfinity/nns";
+import type { Writable } from "svelte/store";
+
+export type NnsNeuronModal =
```
lmuntaner, line 4: Just out of curiosity, is there a specific reason you prefer a union of strings instead of an enum?

lmuntaner (approved):
LGTM, just a small question not blocking the PR.

frontend/src/lib/types/nns-neuron-detail.context.ts
peterpeterparker: Discussed offline

PR #1586, opened 2022-11-28
feat: canister detail design with island

peterpeterparker:
# Motivation

Implement Island in canister detail page and extract modals.

# Changes

- add `<Island />` around canister detail
- move modal a level up but as children of the detail
- enhance context store with `modal` and `selectedController` info to open the modals
- add component `CanisterDetailModals` for all the modals of that particular page

# Screenshots

Capture d’écran 2022-11-28 à 16 54 36
Capture d’écran 2022-11-28 à 16 54 41


frontend/src/lib/modals/canisters/CanisterDetailModals.svelte
```diff
@@ -0,0 +1,51 @@
...
+    store.update((data) => ({
+      ...data,
+      modal: undefined,
+      selectedController: undefined,
+    }));
+</script>
+
+{#if canisterId !== undefined}
+  {#if modal === "add-cycles"}
+    <AddCyclesModal on:nnsClose={close} />
```
lmuntaner, line 37: If you pass the details as props you won't have to do any changes if we move to the tigger by custom event.

lmuntaner (approved):
Just a comment that could lead to changes. But no necessarily in this PR. LGTM

PR #1587, opened 2022-11-28
fix: followee modal overflow

peterpeterparker:
# Motivation

Follow-up of #1585. `Followee` modal needs to be moved outside of the island as well.

# Note

As discussed, pattern is hacky and shall be improved afterwards.

# Changes

- extract modal outside of the island and pass the selected followee with the help of the context

# Screenshot

Afterwards:

Capture d’écran 2022-11-28 à 17 13 35


lmuntaner (approved):
Thanks!

PR #1591, opened 2022-11-29
feat: account title

peterpeterparker:
# Motivation

Make account title dynamic when user scroll to keep important information visible.

# Changes

- use intersection directive to display either the "Account" title or the name of the account and its balance
- fix account title not set when "Wallet" page accessed directly
- fix neuron detail test skipped

# Screenshot

Enregistrement de l’écran 2022-11-29 à 12 27 39


lmuntaner (approved):
Thanks

PR #1594, opened 2022-11-29
feat: canister detail modals with custom event

peterpeterparker:
# Motivation

Replace usage of context store to trigger the canister details modal opening, closing and passing extra data.

# Changes

- use `CustomEvent` to trigger opening of modals used in "Canister" detail page
- adapt `Intersection` types to pattern use in this PR
- remove few unused imports

### DIFF
--- a/frontend/src/lib/types/canister-detail.modal.ts
+++ b/frontend/src/lib/types/canister-detail.modal.ts
@@ -8,15 +8,15 @@ export type CanisterDetailsModalType =
 
 export interface CanisterDetailsModal<D = void> {
   type: CanisterDetailsModalType;
-  detail?: D;
+  data?: D;
 }
 
 export type CanisterDetailsModalRemoveController = Required<
-  Pick<CanisterDetailsModal<{ controller: string }>, "detail">
+  Pick<CanisterDetailsModal<{ controller: string }>, "data">
 > &
   Pick<CanisterDetailsModal, "type">;
 
 export type CanisterDetailsModalDetach = Required<
-  Pick<CanisterDetailsModal<{ canisterId: Principal }>, "detail">
+  Pick<CanisterDetailsModal<{ canisterId: Principal }>, "data">
 > &
   Pick<CanisterDetailsModal, "type">;
--- a/frontend/src/lib/types/canister-detail.modal.ts
+++ b/frontend/src/lib/types/canister-detail.modal.ts
@@ -1,0 +1,22 @@
+import type { Principal } from "@dfinity/principal";
+
+export type CanisterDetailsModalType =
+  | "add-cycles"
+  | "detach"
+  | "add-controller"
+  | "remove-controller";
+
+export interface CanisterDetailsModal<D = void> {
+  type: CanisterDetailsModalType;
+  detail?: D;
+}
+
+export type CanisterDetailsModalRemoveController = Required<
+  Pick<CanisterDetailsModal<{ controller: string }>, "detail">
+> &
+  Pick<CanisterDetailsModal, "type">;
+
+export type CanisterDetailsModalDetach = Required<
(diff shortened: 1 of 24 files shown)


frontend/src/lib/types/canister-detail.modal.ts
```diff
@@ -0,0 +1,22 @@
...
+  | "detach"
+  | "add-controller"
+  | "remove-controller";
+
+export interface CanisterDetailsModal<D = void> {
+  type: CanisterDetailsModalType;
+  detail?: D;
+}
+
+export type CanisterDetailsModalRemoveController = Required<
```
lmuntaner, line 14: Is this the same as?

```
export type CanisterDetailsModalRemoveController = CanisterDetailsModal<{ controller: string }> & {
  detail: { controller: string };
};
```

If we ever add properties to CanisterDetailsModal we would need to pick them as well.

Also, you used an interface that you never extend, just pick from it. It looks weird to me. Is there an advantage over extending it?

frontend/src/tests/lib/components/canister-detail/RemoveCanisterControllerButton.spec.ts
```diff
@@ -87,32 +87,4 @@ describe("RemoveCanisterControllerButton", () => {
     expect(removeController).toBeCalled();
     expect(reloadDetailsMock).toBeCalled();
   });
-
-  it("clear selected controller after remove", async () => {
```
lmuntaner, line 91: Why did you remove this test?

frontend/src/lib/types/canister-detail.modal.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type CanisterDetailsModalType =
+  | "add-cycles"
+  | "detach"
+  | "add-controller"
+  | "remove-controller";
+
+export interface CanisterDetailsModal<D = void> {
+  type: CanisterDetailsModalType;
+  detail?: D;
```
lmuntaner, line 11: The term `detail` is also used in the `emit`. Shall we call this `data` or `payload`?

lmuntaner (commented):
Thanks! A few comments

frontend/src/tests/lib/components/canister-detail/RemoveCanisterControllerButton.spec.ts
peterpeterparker: Because there is no reset anymore since the "controller" is ephemeral - passed through the event only and not kept in a state

frontend/src/lib/types/canister-detail.modal.ts
peterpeterparker: Nice, good idea 👍. No preference so I go for `data` since it's the first you mention but if you rather like `payload` ping me.

frontend/src/lib/types/canister-detail.modal.ts
lmuntaner: Yes, that makes sense to me. It's the usage of the generic type that I found weird. Using `Pick` instead of extending.

frontend/src/tests/lib/components/canister-detail/RemoveCanisterControllerButton.spec.ts
lmuntaner: Ah, the modal is not in the component and it doesn't open it. Ok.

That probably means that we need to test that the event is emitted then?

Or how do we test that clicking a button opens a specific modal?

lmuntaner (approved):
Thanks!

frontend/src/lib/types/canister-detail.modal.ts
peterpeterparker: Discussing in chat, modals covered in other spec tests

PR #1600, opened 2022-11-30
feat: keep track of http agent

peterpeterparker:
# Motivation

This PR was original meant to introduce the `syncTime` feature of agent-js. To do so, we would have need to keep track - keep in memory - the `HttpAgent` that have been used because the `syncTime` function has to be called after the agent is created and because we would have like to call it only once.

Ultimately, or currently, it seems we do not need the `syncTime` feature anymore - because the network seems to have been fixed. Therefore we could have close the PR.

However, keeping track of the http agent seems to improve the performance of the app. Indeed, not re-creating the http agent on each API call seems to improve the overall behavior of the app.

That's why I suggest to go on with this PR.

# Follow-up

Previous PRs:

- Introduced https://github.com/dfinity/nns-dapp/pull/1584
- Reverted https://github.com/dfinity/nns-dapp/pull/1601

# Changes

- move `createAgent` in an api
- add a global variable that keeps track of the agent related to their principal - i.e. one anonymous and one for the user
- reset variable on sign-out


frontend/src/lib/api/agent.api.ts
```diff
@@ -0,0 +1,50 @@
+import {FETCH_ROOT_KEY, HOST} from "$lib/constants/environment.constants";
+import type { HttpAgent, Identity } from "@dfinity/agent";
+import { createAgent as createAgentUtil } from "@dfinity/utils";
+import {LEDGER_CANISTER_ID} from "$lib/constants/canister-ids.constants";
+
+type PrincipalAsText = string;
+let agents: Record<PrincipalAsText, HttpAgent> = {};
+
+export const createAgent = async ({identity, host}: {
```
lmuntaner, line 9: Why wouldn't you do it in `utils`?

lmuntaner (approved):

frontend/src/lib/api/agent.api.ts
peterpeterparker: `syncTime`call the IC. it was an API but since it is move and chain with the creation of the agent that is also called by the API, it had to be moved there

mstrasinskis (approved):
LGTM, thank you!

PR #1604, opened 2022-12-01
Chore: Enable Unspecific topic in sns

lmuntaner:
# Motivation

Users can set followees for all topics using on nervous system function.

# Changes

* Remove the util functionsToFollow that filtered out the function with id 0.
* In the service loadSnsNervousSystemFunctions, change the name of the function with id 0 from "Unspecified" to "All Topics" to be more user friendly.

# Tests

* Remove test for functionsToFollow.
* Fix test that show topic with id 0.

### DIFF
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -591,6 +591,7 @@
   },
   "sns_neuron_detail": {
     "header": "Neuron",
+    "all_topic": "All topics",
     "add_hotkey_info": "In order to use this neuron to vote from the corresponding dapp, you have to add your personal dapp ID as a hotkey here. Please refer to the instructions given by the corresponding dapp.",
     "add_hotkey_tooltip": "In order to use this neuron to vote from the corresponding dapp, you have to add your personal dapp ID as a hotkey here. Please refer to the instructions given by the corresponding dapp."
   },
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -591,7 +591,7 @@
   },
   "sns_neuron_detail": {
     "header": "Neuron",
-    "all_topic": "All topics",
+    "all_topics": "All topics",
     "add_hotkey_info": "In order to use this neuron to vote from the corresponding dapp, you have to add your personal dapp ID as a hotkey here. Please refer to the instructions given by the corresponding dapp.",
     "add_hotkey_tooltip": "In order to use this neuron to vote from the corresponding dapp, you have to add your personal dapp ID as a hotkey here. Please refer to the instructions given by the corresponding dapp."
   },
--- a/frontend/src/lib/services/sns-neurons.services.ts
+++ b/frontend/src/lib/services/sns-neurons.services.ts
@@ -445,7 +445,8 @@ export const loadSnsNervousSystemFunctions = async (
         certified,
       }),
     onLoad: async ({ response: nsFunctions, certified }) => {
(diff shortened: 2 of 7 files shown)


lmuntaner:
New UI:

Screenshot 2022-12-01 at 07 57 21
Screenshot 2022-12-01 at 07 57 37

lmuntaner:

frontend/src/lib/i18n/en.json
```diff
@@ -590,6 +590,7 @@
   },
   "sns_neuron_detail": {
     "header": "Neuron",
+    "all_topic": "All topics",
```
peterpeterparker, line 593: nth: singular vs plural

all_topic
All topics

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -444,9 +445,20 @@ export const loadSnsNervousSystemFunctions = async (
         certified,
       }),
     onLoad: async ({ response: nsFunctions, certified }) => {
+      // TODO: Can we change the name in the backend?
+      const snsNervousSystemFunctions = nsFunctions.map((nsFunction) => {
```
peterpeterparker, line 449: ummm not sure that should be done when loading data, spontaneously I would have implemented it when data are rendered / displayed

any particular reason or advantages of doing this here?

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -444,9 +445,20 @@ export const loadSnsNervousSystemFunctions = async (
...
+        if (nsFunction.id === BigInt(0)) {
+          const translationKeys = get(i18n);
+          return {
+            ...nsFunction,
+            name: translationKeys.sns_neuron_detail.all_topic,
```
peterpeterparker, line 454: ummm not sure that should be done when loading data, spontaneously I would have implemented it when data are rendered / displayed

any particular reason or advantages of doing this here?

lmuntaner: I was also not sure but I thought:

- I hope to remove this and be fixed in backend. Therefore, removing in one place will be easier.
- Adding the logic of rendering a different name depending on the id will need to go in different places and maybe others where we might forget.
- If we add it in the rendering, when this changes, it also means doing more changes and maybe forgetting some changes.
- This was the solution with less repeated code.

Another solution is also to use a derived store to change it when we get the data.

frontend/src/lib/i18n/en.json
lmuntaner: done

frontend/src/lib/services/sns-neurons.services.ts
peterpeterparker: So I understand it was not implemented here because we hope to defer this to the backend (ideal solution) and to avoid for the time being the tasky task but ultimately, if not deferred to backend, there isn't much reason to do this on load rather than on rendering beside the work it needs. So I would say let's track it in Jira and see if it is implemented in the backend or not. If not, then let's refactor.

peterpeterparker (approved):
LGTM, thanks

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: Done: https://dfinity.atlassian.net/browse/GIX-1169

and added it to the comment

peterpeterparker: Thanks

PR #1607, opened 2022-12-01
Chore: Enable end-to-end test in CI

lmuntaner:
# Motivation

Enable end-to-end tests in CI.

# Changes

* Upgrade dfx version
* Fix broken e2e tests after last changes.
* Fix a UX bug discovered with CI!
* Enable e2e tests in CI
* Disable nns_canisters step because it's not needed.

# Tests

* All changes are in tests and CI pipeline.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -106,21 +106,21 @@ jobs:
         run: |
           dfx start --background --clean
 
-      - name: Get cached nns canisters
-        id: create_nns_canister_cache
-        uses: actions/cache@v3
-        with:
-          path: |
-            target/ic
-          key: "nns-canisters-${{ env.IC_COMMIT }}-${{hashFiles( 'e2e-tests/scripts/nns-canister*' )}}-${{ env.NNS_CACHING_KEY }}"
-      - name: Show canister versions
-        run: sha256sum target/ic/*.wasm
+      # - name: Get cached nns canisters
+      #   id: create_nns_canister_cache
+      #   uses: actions/cache@v3
+      #   with:
+      #     path: |
+      #       target/ic
+      #     key: "nns-canisters-${{ env.IC_COMMIT }}-${{hashFiles( 'e2e-tests/scripts/nns-canister*' )}}-${{ env.NNS_CACHING_KEY }}"
+      # - name: Show canister versions
+      #   run: sha256sum target/ic/*.wasm
       # IDENTITY_SERVICE_URL needs to match `e2e-tests/test.js`
       - name: Deploy NNS canisters
         run: |
           dfx nns install
 
-      - name: Deploy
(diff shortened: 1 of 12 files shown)


e2e-tests/README.md
```diff
@@ -104,6 +78,8 @@ First e2e is also used to populate the nns dapp:
 ./deploy.sh --populate local
 ```
 
+**PENDING: Setting cycles exchange is not working yet. Therefore, the related tests are skipped.**
```
bitdivine, line 81: The cycle exchange rate is set by `dfx nns install` - or should be!  The line to do so is here: https://github.com/dfinity/sdk/blob/master/src/dfx/src/lib/nns/install_nns.rs#L112  It s dne by proposal rather than by using an exposed API.  However we can add those tests in in another PR.  This is great progress as it is.  But I don't see tests being disabled...  I must be missing something, unless you mean the part above where the xdr rate was being set?

deploy.sh
```diff
@@ -331,8 +331,10 @@ if [[ "$DEPLOY_NNS_DAPP" == "true" ]]; then
 fi
 
 if [[ "$POPULATE" == "true" ]]; then
-  echo Setting the cycles exchange rate...
-  ./scripts/propose --to propose-xdr-icp-conversion-rate --dfx-network "$DFX_NETWORK" --jfdi
+  # echo Setting the cycles exchange rate...
+  # ./scripts/propose --to propose-xdr-icp-conversion-rate --dfx-network "$DFX_NETWORK" --jfdi
+  # TODO: Enable this by fixing ./scripts/propose
+  echo Skipping setting the cycles exchange rate...
```
bitdivine, line 337: The lies above can be deleted as the exchange rate has already been set by `dfx nns install`. Likewise setting the authorized subnets below is already done, so that too can be nuked.  

Here is the code in the implementation of dfx nns install:
https://github.com/dfinity/sdk/blob/master/src/dfx/src/lib/nns/install_nns.rs#L112

bitdivine (approved):

.github/workflows/build.yml
```diff
@@ -157,21 +108,14 @@ jobs:
...
-
       - name: prepare and run the test suite
         working-directory: e2e-tests
         run: |
           npm ci
-
           # wipe the screenshots, they will be re-committed
           rm -rf screenshots
           set -o pipefail
-          npm run test | tee -a wdio.log || true
```
lmuntaner, line 174: @bitdivine this is the line to enable the e2e test in the CI.

I should have commented it before asking for CR, sorry.

e2e-tests/README.md
lmuntaner: I'll remove this part, thanks!

deploy.sh
lmuntaner: done

PR #1613, opened 2022-12-05
Chore: Update Internet Identity Init type

lmuntaner:
# Motivation

User can see the latest "memory_migration_batch_size" argument name used when upgrading the II canister.

# Changes

* Add "memory_migration_batch_size" to InternetIdentityInit hardcoded type.

# Tests

Not applicable, only UI changes.

### DIFF
--- a/rs/src/proposals.rs
+++ b/rs/src/proposals.rs
@@ -27,7 +27,6 @@ use std::cell::RefCell;
 use std::collections::BTreeMap;
 use std::fmt::Debug;
 use std::ops::DerefMut;
-use std::str::FromStr;
 
 type Json = String;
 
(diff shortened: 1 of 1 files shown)


lmuntaner:
New UI:

Screenshot 2022-12-05 at 09 33 38

rs/src/proposals.rs
```diff
@@ -73,6 +73,7 @@ struct InternetIdentityInit {
     pub assigned_user_number_range: Option<(UserNumber, UserNumber)>,
     pub archive_module_hash: Option<[u8; 32]>,
     pub canister_creation_cycles_cost: Option<u64>,
+    pub memory_migration_batch_size: Option<u64>,
```
frederikrothenberger, line 76: It is actually an `u32`.

frederikrothenberger:
You can copy the type from here: https://github.com/dfinity/internet-identity/blob/main/src/internet_identity_interface/src/lib.rs#L174

rs/src/proposals.rs
lmuntaner: thanks!

frederikrothenberger (approved):

PR #1612, opened 2022-12-05
build: bump dev dependencies

peterpeterparker:
# Changes

- bump dev dependencies
- adapt for SvelteKit breaking changes
- few eslint warnings solved in test and type definition

PR #1615, opened 2022-12-06
fix: page store path for SvelteKit breaking changes

peterpeterparker:
# Motivation

Last version of SvelteKit (bumped in #1612) contains another breaking changes. The route id of the `$app/stores.page` store now also prefix route group with `/`.

Used to be: `(app)/neuron`
Is now:  `/(app)/neuron`

This breaks our route matcher `pathForRouteId`
# Changes

- the related constant for the group `(app)` should be adapted to `/(app)`

# Tests

We cannot detect this with unit test because we have to mock the SvelteKit store. We could have detect it with e2e tests. I detected it on the "Sns neuron detail" page because the Skeleton did not disappeared.

### DIFF
--- a/frontend/__mocks__/$app/stores.ts
+++ b/frontend/__mocks__/$app/stores.ts
@@ -1,6 +1,6 @@
 import type { Page } from "@sveltejs/kit";
 import { writable } from "svelte/store";
-import { OWN_CANISTER_ID_TEXT } from "../../src/lib/constants/canister-ids.constants";
+import { OWN_CANISTER_ID_TEXT } from "$lib/constants/canister-ids.constants";
 
 const initPageStoreMock = () => {
   const { subscribe, set } = writable<Partial<Page>>({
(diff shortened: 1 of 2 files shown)


frontend/__mocks__/$app/stores.ts
```diff
@@ -26,7 +26,7 @@ const initPageStoreMock = () => {
     }) =>
       set({
         data,
-        route: { id: `(app)${routeId}` },
+        route: { id: `/(app)${routeId}` },
```
lmuntaner, line 29: How about using `ROUTE_ID_GROUP_APP` here?

lmuntaner (changes requested):
Minor change

frontend/__mocks__/$app/stores.ts
peterpeterparker: Good idea, let me try!

peterpeterparker: Yes works, super 👍

lmuntaner (approved):

PR #1617, opened 2022-12-06
GIX-1136: Do not use universe in proposals and canisters

lmuntaner:
# Motivation

User is not confused by the universe query param in the url in proposals and canisters pages.

# Changes

* Remove universe param from "buildProposalUrl", "buildProposalsUrl",  "buildCanistersUrl" and "buildCanisterUrl".
* Remove universe param when using any of the utils above.
* "universe" param in "buildUrl" is now optional.
* Change implementation of "buildUrl" to use URLSearchParams.

# Tests

* Fix tests after refactor

### DIFF
--- a/frontend/src/lib/derived/paths.derived.ts
+++ b/frontend/src/lib/derived/paths.derived.ts
@@ -19,10 +19,10 @@ export const neuronsPathStore = derived<Readable<Page>, string>(
 
 export const proposalsPathStore = derived<Readable<Page>, string>(
   pageStore,
-  ({ universe }) => buildProposalsUrl({ universe })
+  () => buildProposalsUrl()
 );
 
 export const canistersPathStore = derived<Readable<Page>, string>(
   pageStore,
-  ({ universe }) => buildCanistersUrl({ universe })
+  () => buildCanistersUrl()
 );
--- a/frontend/src/lib/utils/navigation.utils.ts
+++ b/frontend/src/lib/utils/navigation.utils.ts
@@ -35,21 +35,32 @@ const buildUrl = ({
   params = {},
 }: {
   path: AppPath;
-  universe: string;
+  universe?: string;
   params?: Record<string, string>;
-}): string =>
-  `${path}/?${UNIVERSE_PARAM}=${universe}${Object.entries(params)
-    .map(([key, value]) => `&${key}=${value}`)
-    .join("")}`;
+}): string => {
+  const queryParams = new URLSearchParams();
+
+  if (universe !== undefined) {
+    queryParams.set(UNIVERSE_PARAM, universe);
+  }
+
+  Object.entries(params).forEach(([key, value]) => {
+    queryParams.set(key, value);
+  });
+
+  const hasQueryParams = queryParams.toString().length > 0;
+
+  return `${path}/${hasQueryParams ? "?" : ""}${queryParams.toString()}`;
+};
 
 export const buildAccountsUrl = ({ universe }: { universe: string }) =>
   buildUrl({ path: AppPath.Accounts, universe });
 export const buildNeuronsUrl = ({ universe }: { universe: string }) =>
   buildUrl({ path: AppPath.Neurons, universe });
-export const buildProposalsUrl = ({ universe }: { universe: string }) =>
-  buildUrl({ path: AppPath.Proposals, universe });
-export const buildCanistersUrl = ({ universe }: { universe: string }) =>
-  buildUrl({ path: AppPath.Canisters, universe });
+export const buildProposalsUrl = () =>
+  buildUrl({ path: AppPath.Proposals, universe: undefined });
+export const buildCanistersUrl = () =>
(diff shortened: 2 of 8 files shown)


lmuntaner:
New UX:

Focus on the address bar of the video:

https://user-images.githubusercontent.com/4550653/205853295-9baed386-cb8d-442f-8c9e-e0c62c7e35d8.mov

lmuntaner:

I consider Voting and Canister to be part of the NNS universe but without the query param to avoid confusions.

That's why when clicking to neurons and accounts from Voting or Canisters, the user goes to NNS accounts or neurons.

This was an issue raised by security. Although very minor and non-blocking. But it made sense because when sharing the url for proposals for example, it felt weird to have the universe param there. The url was ugly with it, specially being now a publich page.

peterpeterparker:
> I consider Voting and Canister to be part of the NNS

Sns Voting will come some day (rather soon than late) to NNS-dapp my understanding.

lmuntaner:
> > I consider Voting and Canister to be part of the NNS
> 
> Sns Voting will come some day (rather soon than late) to NNS-dapp my understanding.

I also think so, but until then, Voting is only NNS.

frontend/src/lib/derived/paths.derived.ts
```diff
@@ -19,10 +19,10 @@ export const neuronsPathStore = derived<Readable<Page>, string>(
 
 export const proposalsPathStore = derived<Readable<Page>, string>(
   pageStore,
-  ({ universe }) => buildProposalsUrl({ universe })
+  () => buildProposalsUrl()
```
peterpeterparker, line 22: the `proposalsPathStore` and `canistersPathStore` should be fully removed

if the store is not needed, then no derived store isn't neither

frontend/src/lib/utils/navigation.utils.ts
```diff
@@ -35,21 +35,32 @@ const buildUrl = ({
   params = {},
 }: {
   path: AppPath;
-  universe: string;
+  universe?: string;
```
peterpeterparker, line 38: not for this pr but make me think that here we should use the `Page` type

`Omit<Page, 'universe'> & Partial<Pick<Page, 'universe>>> & {params....`

frontend/src/lib/utils/navigation.utils.ts
```diff
@@ -35,21 +35,32 @@ const buildUrl = ({
...
+
+  if (universe !== undefined) {
+    queryParams.set(UNIVERSE_PARAM, universe);
+  }
+
+  Object.entries(params).forEach(([key, value]) => {
+    queryParams.set(key, value);
+  });
+
+  const hasQueryParams = queryParams.toString().length > 0;
```
peterpeterparker, line 51: is this check mandatory?

frontend/src/lib/utils/navigation.utils.ts
```diff
@@ -35,21 +35,32 @@ const buildUrl = ({
...
   path: AppPath;
-  universe: string;
+  universe?: string;
   params?: Record<string, string>;
-}): string =>
-  `${path}/?${UNIVERSE_PARAM}=${universe}${Object.entries(params)
-    .map(([key, value]) => `&${key}=${value}`)
-    .join("")}`;
+}): string => {
+  const queryParams = new URLSearchParams();
```
peterpeterparker, line 41: Just in case you would like a solution without searchparams

```
`${path}/?${Object.entries({...params, ...(universe !== undefined && {[UNIVERSE_PARAM]: universe})})
    .map(([key, value]) => `&${key}=${value}`)
    .join("")}`
```

peterpeterparker:
I guess yes but just to be sure, such a navigation as

- open Accounts
- switch to Sns project
- go to proposal
- go to Accounts -> Sns is still selected

Still works right?

lmuntaner:
> I guess yes but just to be sure, such a navigation as
> 
> * open Accounts
> * switch to Sns project
> * go to proposal
> * go to Accounts -> Sns is still selected
> 
> Still works right?

No, this is what changed. I consider going to Canisters as moving the the NNS universe. Therefore, when coming back to Accounts you are in NNS.

I'll check this with Mischa.

lmuntaner:
@peterpeterparker Mischa prefers for now to change universe:

> "I think, as long as we don’t let people vote for sns proposals in our app, I would switch back to NNS as selected universe (and see if we get bad feedback for this… " Mischa

frontend/src/lib/utils/navigation.utils.ts
lmuntaner: this added `&` in front of the query params that was not `u`, for example: `/proposal/?&proposal=123`

peterpeterparker:
As discussed offline, we gonna wait a bit to check when we will have to implement Sns proposals (because it might happens soon). As it might be easier to begin this PR from scratch, @lmuntaner closed it for now.

PR #1624, opened 2022-12-07
Fix: Hotkey label only for hotkeys

lmuntaner:
# Motivation

Users that is missing one permission is not a hotkey.

A hotkey is strictly having the permission to vote and submit proposal.

# Changes

* Change "getSnsNeuronHotkeys" to strictly check permissions.
* New util "arraysEqual"

# Tests

* Test new case.
* Test new util

### DIFF
--- a/frontend/src/lib/utils/utils.ts
+++ b/frontend/src/lib/utils/utils.ts
@@ -346,29 +346,3 @@ export const expandObject = (
     }
     return acc;
   }, {} as Record<string, unknown>);
-
-/**
- * Returns whether two arrays are equal or not. Order matters.
- *
- * Compares each element of the array with equality operator (===).
- *
- * @param a first array
- * @param b second array
- * @returns {boolean}
- */
-export const arraysEqual = ({
-  a,
-  b,
-}: {
-  a: unknown[];
-  b: unknown[];
-}): boolean => {
-  if (a === b) return true;
-  if (a == null || b == null) return false;
-  if (a.length !== b.length) return false;
-
-  for (let i = 0; i < a.length; ++i) {
-    if (a[i] !== b[i]) return false;
-  }
-  return true;
-};
--- a/frontend/src/lib/utils/utils.ts
+++ b/frontend/src/lib/utils/utils.ts
@@ -346,3 +346,29 @@ export const expandObject = (
     }
     return acc;
   }, {} as Record<string, unknown>);
+
+/**
+ * Returns whether two arrays are equal or not. Order matters.
+ *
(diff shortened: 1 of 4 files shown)


lmuntaner:

I realized this when the permissions of the controller didn't match all the permissions in the enum in ic-js. The permissions to stake maturity was included in the neuron, but not in our enum.

A hotkey is strictly having those two permissoins.

frontend/src/lib/utils/utils.ts
```diff
@@ -346,3 +346,29 @@ export const expandObject = (
...
+}: {
+  a: unknown[];
+  b: unknown[];
+}): boolean => {
+  if (a === b) return true;
+  if (a == null || b == null) return false;
+  if (a.length !== b.length) return false;
+
+  for (let i = 0; i < a.length; ++i) {
+    if (a[i] !== b[i]) return false;
```
peterpeterparker, line 371: this works only with primitive no? e.g not with objects that have keys sorted  differently.

lmuntaner: correct. I added a test that specifically checks that it returns `false`. I wanted something simple for now

peterpeterparker: someone in the future might use the utils without checking it's implementation and get some suprises.

is there a way to solve the issue without this comparison?

if not, I guess then maybe improving the doc and maybe we can find a function name that implies that it is limited?

lmuntaner: More than the comment and the test?

This is not a library, it's our code. If they add a test in they code it will fail.

I can do the comparison without an util, but I thought the preferred way was to use utils.

peterpeterparker: The comment says:

```
* Returns whether two arrays are equal or not. Order matters.
 *
 * Compares each element of the array with equality operator (===).
 ```
 
 i.e. it does not mention that it is not reliable if you use it with objects.
 
 The function name is
 
 ```
 arraysEqual
 ```
 
 Which also implies that anything can be compared.
 
If the root cause that should be solved by the PR cannot be solved differently than with this utils, I'm fine with it but the comment and name of the function should be improved in my opinion. This for maintability reason and to avoid unexpected issue in the future.

peterpeterparker: As discussed, PR provided in https://github.com/dfinity/nns-dapp/pull/1628

peterpeterparker (commented):
LGTM

peterpeterparker (approved):

PR #1626, opened 2022-12-07
fix (ci): Update formatter

bitdivine:
# Motivation
The CI formatter failed and allowed unformatted code into main.

On investigation, the formatter uses a third party github action to commit changes.  That third party action renamed an option:

> - **[BREAKING]** The `pull_strategy` input has been removed, use `pull` instead.

We had `pull_strategy` set to not pull, which is the new default, so everything carried on working fine until use of unrecognized options was made illegal (presumably by github as there was no change in the action version) and CI broke.

# Changes
- remove the use of `pull_strategy`
- put the formatting commands in shell scripts so that they are easy to run locally in the same way that they are run in CI.

# Tests
See CI

- Pushed a formatting change to `dfx.json`
- CI failed.
- Fixed CI and pushed another change
- CI formatted correctly but the commit message was "Updating frontend formatting"
- Updated message to exclude "frontend"


lmuntaner (commented):
There is still the same error in the CI:

```
Error: Error: fatal: You are not currently on a branch.
```

It's in the job Commit Formatting Changes

.github/workflows/checks.yml
```diff
@@ -59,11 +66,15 @@ jobs:
         with:
           add: .
           author_name: Formatting Committer
-          author_email: "<nobody@example.com>"
```
bitdivine, line 62: example.com is actually a real domain.  Their github stats must be amazing.

.github/workflows/checks.yml
```diff
@@ -20,6 +20,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.ref }}
+        if: ${{ github.event_name == 'pull_request' }}
+      - name: Checkout
+        uses: actions/checkout@v3
+        if: ${{ github.event_name != 'pull_request' }}
```
bitdivine, line 29: If in a PR we want to commit any formatting changes.  For that to work, we need to be on a branch.  But PRs are not on a branch by default any more.

bitdivine:
@lmuntaner 
> There is still the same error in the CI:

Sorry, I should have marked the PR as draft.  CI has fixed one malformatted file now, testing another....

.github/workflows/checks.yml
lmuntaner: Shall we remove the committing part as agreed with @peterpeterparker ?

bitdivine: Yes

.github/workflows/checks.yml
```diff
@@ -2,9 +2,6 @@ name: CI Checks
 
 on:
   push:
```
bitdivine, line 4: The "on pull request" is not needed as we no longer commit formatting changes in CI.

.github/workflows/checks.yml
```diff
@@ -21,49 +18,23 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Install rust formatter
+        run: rustup component add rustfmt
       - name: Install shfmt
         run: sudo snap install --classic shfmt
```
bitdivine, line 23: No need to install rust, as github has it installed already.

The rust version used is the default for this project, defined in the rust-toolchain.toml and cargo reads that automatically, no extra steps needed.  The rust version was actually ignored here; it is not defined in this job as you can see by the fact that there is no matrix definition above, so the version specification was just blank.

peterpeterparker (approved):
Rubber stamp

PR #1628, opened 2022-12-08
feat: hotkeys comparison

peterpeterparker:
# Motivation

PR #1624 uses a new function `arraysEqual` that is compatible with primitive only which I think is prone to introduce unexpected issue in the future if developer does not pay attention and uses it with complex objects. As discussed with @lmuntaner, this PR replaces the need of this function.

# Changes

- remove `arraysEqual`
- compare that every elements are contained within the constant and that both elements and constant have the same length


mstrasinskis (approved):
LGTM, thanks

PR #1633, opened 2022-12-12
feat: when proposal has settled, display not voted neurons

peterpeterparker:
# Motivation

When the proposal has settled, it has archived and its ballots are emptied (to spare space in the canister). Because no more ballots are delivered, querying the ineligible neurons returns that all neurons are ineligible which leads to displaying these as such which is not fully correct.

That's why this PR checks the rewards status of the proposals and if settled, display the list of not voted neurons instead of those ineligible.

# Source

Reported on the forum https://forum.dfinity.org/t/visual-error-or-malfunction-voting-neurons/17072

# PR

- [ ] https://github.com/dfinity/ic-js/pull/246

# Changes

- rename cmp `<IneligibleNeuronsCard />` to `<NotVoteNeuronsCard />`
- if proposal reward status is settled, display "not voted" neurons instead of ineligible

# Screenshot

Capture d’écran 2022-12-12 à 07 56 27


mstrasinskis (approved):
LGTM, thanks

peterpeterparker:
Will not be implemented. Solution in #1638

PR #1634, opened 2022-12-12
Follow 825 security fix claming with stake less then minimum

mstrasinskis:
# Motivation

Prevent calls to SNS governance claim neuron with stake less than min stake amount (more info [here](https://dfinity.atlassian.net/browse/FOLLOW-825)).

# Changes

- load nervouse system params on neurons page
- replace 0n with `minimum_stake_amount` in claim comparison

# Tests

- updated snsNeurons page to preload parameters


frontend/src/lib/pages/SnsNeurons.svelte
```diff
@@ -16,13 +16,18 @@
...
+  import { loadSnsParameters } from "$lib/services/sns-parameters.services";
 
   let loading = true;
 
   const unsubscribe: Unsubscriber = snsOnlyProjectStore.subscribe(
     async (selectedProjectCanisterId) => {
       if (selectedProjectCanisterId !== undefined) {
         loading = true;
+
+        // params.minimum_stake_amount needs for checking neurons balance (checkNeuronsSubaccounts)
```
peterpeterparker, line 28: can/should we extract that logic in a service and have a test that check that `loadSnsParameters` is indeed called before the calls?

frontend/src/lib/services/sns-neurons-check-balances.services.ts
```diff
@@ -287,10 +295,19 @@ export const checkSnsNeuronBalances = async ({
   // TODO: Check neurons controlled by linked HW?
   const identity = await getAuthenticatedIdentity();
 
+  // TODO: refactor using `getSnsParametersFromStore` https://dfinity.atlassian.net/browse/GIX-1178
+  const neuronMinimumStake = fromDefinedNullable(
+    (
+      get(snsParametersStore)?.[rootCanisterId.toText()]
+        ?.parameters as NervousSystemParameters
+    ).neuron_minimum_stake_e8s ?? []
```
peterpeterparker, line 303: should the fallback value be `?? [BigInt(0)]`? but yeah I guess it will be refactored anyway

peterpeterparker (approved):
LGTM, discussed in chat

frontend/src/lib/pages/SnsNeurons.svelte
mstrasinskis: Todo added

PR #1637, opened 2022-12-13
feat: update tcycles fee in static label

### LINKED ISSUE #1635: BUG: Outdated message and behaviour
When creating a new canister with the NNS UI, it shows this
image

But this is incorrect. 0.1T cycles are charged now for creating a new canister. And hence the minimum should also be reduced.

We can see that the created canister received 1.9T cycles
image

Expected behaviour:
- Reduce minimum amount to 1.1T or whatever is currently prescribed
- Update the message accordingly

peterpeterparker:
# Motivation

The T cycles fee to create a canister is 0.1 not 1 as displayed in a static hint we are displaying on the interactive modal to create canister.

Source: https://internetcomputer.org/docs/current/developer-docs/deploy/computation-and-storage-costs/

Resolves #1635

# Changes

- update i18n label with cycles


mstrasinskis (approved):
LGTM, thank you!

PR #1638, opened 2022-12-13
feat: if proposal reward status has settled hide vote pane from detail page

peterpeterparker:
# Motivation

- When the proposal reward status has settled, the `ballots` array of the proposal objects are emptied in the canister to spare memory space. 

- The `recentBallots` provided in the neurons can contain up to 100 proposal information

Above two statements could lead to a display misinterpretation in NNS-dapp. Indeed, the frontend might not be able to determine if a neuron has voted and if proposal that has settled are browsed, it might also display these with an incorrect labelling such as having a dissolve delay < 6 months while it is not the case.

Therefore, at least as shorterm solution, we have decided to hide the "Neurons" voting pane from the proposal detail page when, and only when, the reward status as settled.

# Notes

Not displaying misinterpreted information is safer that displaying these. Adding a label "we cannot compute such information" can also lead to questionning, therefore above seem the most acceptable solution.

Archiving the information in the Governance canister would need a quite certain work and resources.

# Source

This was originally reported on the forum https://forum.dfinity.org/t/visual-error-or-malfunction-voting-neurons/17072

# Changes

- do not display "Neuron" bottom sheet / collapsible if reward status is settled

### DIFF
--- a/frontend/src/lib/components/proposal-detail/ProposalVotingSection.svelte
+++ b/frontend/src/lib/components/proposal-detail/ProposalVotingSection.svelte
@@ -2,11 +2,21 @@
   import type { ProposalInfo } from "@dfinity/nns";
   import VotesResults from "./VotesResults.svelte";
   import VotingCard from "./VotingCard/VotingCard.svelte";
+  import { ProposalRewardStatus } from "@dfinity/nns";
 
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
LGTM, thanks!

PR #1640, opened 2022-12-15
cleanup: Drop canvaskit

bitdivine:
# Motivation
canvaskit is no longer needed.  Stable assets are not needed any more either.

# Changes
- Delete references to canvaskit in the rust code; there are none elsewhere.
- Remove the machinery for uploading canvaskit.
- Drop any stable assets on the upgrade after the next.
- We do NOT remove the `stable` field from assets as that might cause problems on upgrade if there are still stable assets in production canisters.

# Future
- [ ] Wait for at least one deployment, so that stable assets are flushed and there is no more persistent asset state.
- [ ] Remove the remaining stable asset code.

# Tests
There is no new functionality.  CI should suffice for existing functionality.
- [ ] Create a testnet nns-dapp; the asset hashes can be changed to the hashes of known files.
- [ ] Upload stable assets to it
- [ ] Verify that the assets can be downloaded
- [ ] Do an upgrade test
- [ ] Verify that the hashes are no longer available after upgrade.


rs/src/assets.rs
```diff
@@ -278,7 +280,13 @@ pub fn init_assets() {
 impl StableState for Assets {
     fn encode(&self) -> Vec<u8> {
         // Encode all stable assets.
```
bitdivine, line 282: Here we drop the stable assets that are already in the canister.

rs/src/assets.rs
```diff
@@ -63,6 +63,7 @@ impl Asset {
         }
     }
```
bitdivine, line 65: We don't have stable assets any more but might want to use at least some of the machinery again.

rs/src/main.rs
```diff
@@ -310,34 +309,6 @@ pub fn canister_heartbeat() {
     dfn_core::api::futures::spawn(future);
 }
```
bitdivine, line 311: This removes the method for uploading stable assets.

peterpeterparker:
> We do NOT remove the stable field from assets as that might cause problems on upgrade if there are still stable assets in production canisters.

Can you add a comment about this in the code?

PR #1641, opened 2022-12-15
Clean up clippy

bitdivine:
# Motivation
The clippy implementation is spewing warnings, some of which are easy to solve.

# Changes
- Don't use the clippy github action; it uses an unsupported nodejs and we install clippy anyway so we don't need it.
- Put the clippy linter into a script so that it is easy to run the same tests used in CI.

# Tests
There are no functional changes, so existing tests should suffice.  I have run the linter script with bad code and confirmed that it returns with a non-zero error code.


peterpeterparker (approved):
LGTM, thx

PR #1643, opened 2022-12-16
Asset compression ("gzip")

bitdivine:
# Motivation
Assets are unnecessarily large, and this slows page load time.

# Changes
* Compress all artefacts.
* Serve the artefacts with the content header specifying the compression.
* Streamline the asset signing, so that the root is computed less frequently.

# Tests


rs/src/assets.rs
```diff
@@ -213,66 +302,68 @@ pub fn hash_bytes(value: impl AsRef<[u8]>) -> Hash {
...
-                .insert(dirname.as_bytes().to_vec(), hash_bytes(&asset.bytes));
-            assets.insert(dirname, asset.clone());
-        }
-
-        asset_hashes
-            .0
-            .insert(path.as_bytes().to_vec(), hash_bytes(&asset.bytes));
-        assets.insert(path, asset);
-
-        update_root_hash(&asset_hashes);
```
bitdivine, line 241: Note regarding the old code - this was recomputing the root hash every time an asset was added.  I'm pretty sure that this is just a waste of cycles.

peterpeterparker:
Test looks good however this branch is quite outdated. Can you please upgrade it with `main` @bitdivine so that we could test/compare bundle sizes?

rs/src/assets.rs
```diff
@@ -81,12 +103,76 @@ impl Asset {
...
+            extended
+        };
+        self.0.get(&path_with_suffix)
+    }
 
-    fn get(&self, path: &str) -> Option<&Asset> {
-        self.0.get(path)
+    /// Returns the paths for which a given asset may be returned.
+    /// Note:  All these paths must be certified.
+    pub fn alternate_paths(path: &str) -> Vec<String> {
```
peterpeterparker, line 152: the comment and the function name confuse me, does it returns many "alternative paths to the provided path" or does it return "the alternative paths + the path"?

if second, maybe this function should be renamed to be more explicit

rs/src/assets.rs
```diff
@@ -81,12 +103,76 @@ impl Asset {
...
+                if path.ends_with(content_encoding.suffix()) {
+                    Some(path[0..path.len() - content_encoding.suffix().len()].to_string())
+                } else {
+                    None
+                }
+            })
+            // Add the directory, with trailing slash, as an alternative path.
+            // Note: Without the trailing slash the location of "." is the parent, so breaks resource links.
+            .flat_map(|path| {
+                if let Some(root) = path.strip_suffix("/index.html") {
```
peterpeterparker, line 166: generally speaking I think it would be a bit more future proof to extract any logic that is "trailing slash resolution" related to dedicated functions.

This implementation works as long as we use "adding" trailing slash in the frontend but, if for some reason, it would change to "removing" trailing slash, then it would not work anymore

We do not plan to do so, again just to be future proof

Reference: https://kit.svelte.dev/docs/page-options#trailingslash

rs/src/assets.rs
```diff
@@ -213,66 +302,68 @@ pub fn hash_bytes(value: impl AsRef<[u8]>) -> Hash {
...
-        dfn_core::api::print(format!("{}: {}", &name, bytes.len()));
+            let mut bytes = Vec::new();
+            entry.read_to_end(&mut bytes).unwrap();
 
-        insert_asset(name, Asset::new(bytes));
-    }
+            dfn_core::api::print(format!("{}: {}", &name, bytes.len()));
+            insert_asset_into_state(state, name, Asset::new(bytes));
+        }
+        update_root_hash(&state.asset_hashes.borrow_mut());
```
peterpeterparker, line 365: it needs to be borrowed as mutable? `&state.asset_hashes.borrow()` would not work?

bitdivine: Yes, it needs to be mutable.  All the assets are in a tree and hashes are computed at every node of the tree over the child nodes, and then finally the root node is signed.  This is then used to prove that any leaf node is valid.  As we have changed the assets, we need to update all the hashes.  Previously this recomputation was done after every asset change, now here I'm doing it once after updating all the assets.

rs/src/assets.rs
bitdivine: I replied but my comment got lost somehow after merging in main.  At the risk of ending up with two responses:  I agree that it would be good to have a clearer picture of how we handle trailing slashes.  A personal bugbear, as I often type in URLs manually, is e.g. https://nns.ic0.app/launchpad  I think it is worth syncing on what we want and then implementing it cleanly.  I think that doesn't have to be mixed in with compression though.  At least, I prefer multiple smaller PRs.

bitdivine: I agree that the index.html and trailing slash handling should be revisited.  E.g. right now, this fails: https://nns.ic0.app/launchpad  But that could be an alternate path for https://nns.ic0.app/launchpad/ and https://nns.ic0.app/launchpad/index.html  It would be good to sync n what we want this to look like and

peterpeterparker: Yes let's revisit afterwards. Please ping me when you got time!

rs/src/assets.rs
peterpeterparker: Gotcha, thanks

peterpeterparker (approved):
LGTM, thanks!

rs/src/assets.rs
bitdivine: Resolving the conversation only because it is required for merge.

PR #1647, opened 2022-12-19
chore: e2e ci for testnets

bitdivine:
# Motivation

# Changes

# Tests


.github/workflows/build.yml
```diff
@@ -1,3 +1,4 @@
+break
```
bitdivine, line 1: Don't waste CPU on this while developing.  It's another e2e test that's doing the work!

PR #1650, opened 2022-12-20
i18n: ledger upgrade text update

mstrasinskis:
# Motivation

Update the error text because the ledger app version is not yet available.

# Changes

- text update


dubovitskaya (approved):
thanks!

PR #1659, opened 2022-12-21
Gix 1175 users can split a neuron

mstrasinskis:
# Motivation

Split sns neuron feature.

- [x] https://github.com/dfinity/ic-js/pull/250

# Changes

- splitNeuron sns api
- splitNeuron service
- splitSnsNeuronButton
- `SplitSnsNeuronModal`
- `SnsNeuronMetaInfoCard` displays `splitSnsNeuronButton` when use has permission to split (behind `ENABLE_SNS_2`)
- `nextMemo` util (`NextMemoNotFoundError`)
- split related utils
- renamed Split**Nns**NeuronButton
- fix `NervousSystemParameters` source

# Tests

- related utils, services and components

# Screenshots

image

image

image


frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
```diff
@@ -21,16 +22,36 @@
...
   $: neuron = $store.neuron;
 
+  let rootCanisterId: Principal | undefined;
+  $: rootCanisterId = $store.selected?.rootCanisterId;
+
   let neuronState: NeuronState | undefined;
   $: neuronState = isNullish(neuron) ? undefined : getSnsNeuronState(neuron);
 
+  let allowedToSplit: boolean;
+  $: allowedToSplit =
```
lmuntaner, line 48: Isn't there a minimum amount to split?

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -293,6 +298,74 @@ export const removeHotkey = async ({
   }
 };
 
+export const splitNeuron = async ({
+  rootCanisterId,
+  neuronId,
+  amount,
+  transactionFee,
```
lmuntaner, line 305: Ditto that it could be read from the store.

Yet, not a strong opinion. But it would be good to do it always the same way.

In here these paremeters are used to perform checks and show errors. They are not strictly related to splitting a neuron. That's why I mentioned that the service could access that data.

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -293,6 +298,74 @@ export const removeHotkey = async ({
...
+    });
+
+    await splitNeuronApi({
+      rootCanisterId,
+      identity,
+      neuronId,
+      amount: amountAndFee,
+      memo,
+    });
+
```
lmuntaner, line 358: Don't we need to reload neurons as well here? So that they will appear in the neuron list?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -267,6 +321,43 @@ export const isSnsNeuron = (
...
+  fee,
+  neuronMinimumStake,
+}: {
+  amount: E8s;
+  fee: E8s;
+  neuronMinimumStake: E8s;
+}): boolean => amount >= neuronMinimumStake + fee;
+
+/*
+neuron_minimum_stake_e8s
```
lmuntaner, line 343: Why these three lines?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -267,6 +321,43 @@ export const isSnsNeuron = (
...
+transaction_fee_e8s
+split.amount_e8s < min_stake + transaction_fee_e8s
+
+- The amount to split minus the transfer fee is more than the minimum stake (thus the child neuron will have at least the minimum stake)
```
lmuntaner, line 347: Shouldn't this comments be in the helper `minNeuronSplittable`?

Then the comment here is: "The neuron can be split when the stake is higher than the minimum required".

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -267,6 +321,43 @@ export const isSnsNeuron = (
 export const hasValidStake = (neuron: SnsNeuron): boolean =>
   neuron.cached_neuron_stake_e8s + neuron.maturity_e8s_equivalent > BigInt(0);
 
+export const minNeuronSplittable = ({
```
lmuntaner, line 324: The two comments could go here to explain the logic of `2n * ...`

lmuntaner (commented):
Looking good! Some comments only.

frontend/src/lib/modals/sns/neurons/SnsStakeMaturityModal.svelte
```diff
@@ -32,6 +32,7 @@
     });
 
     await reloadNeuron();
+    // TODO: Checks subaccounts of identity in order to find neurons that need to be refreshed or claimed. ?
```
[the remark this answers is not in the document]
mstrasinskis: This comment is not related to this PR, just wanna discuss this w/ you to better understand, is it necessary here or not? There is no `checkSnsNeuronBalances` on the `/neuron detail` page, only on the `/neurons`.

frontend/src/lib/modals/sns/neurons/SplitSnsNeuronModal.svelte
```diff
@@ -0,0 +1,133 @@
...
+      });
+      return;
+    }
+
+    startBusy({ initiator: "split-neuron" });
+
+    const { success } = await splitNeuron({
+      rootCanisterId,
+      neuronId: fromDefinedNullable(neuron.id),
+      amount: BigInt(Math.round(amount * Number(E8S_PER_ICP))),
```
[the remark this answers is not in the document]
mstrasinskis: Thanks, saw it in your last PR. Updated 👍

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -293,6 +298,74 @@ export const removeHotkey = async ({
   }
 };
 
+export const splitNeuron = async ({
+  rootCanisterId,
+  neuronId,
+  amount,
+  transactionFee,
+  token: { symbol },
```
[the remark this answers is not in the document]
mstrasinskis: I would prefer to keep it as a prop, so it's guarded by types. Otherwise the service should handle the lack of value in the store.

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -293,6 +298,74 @@ export const removeHotkey = async ({
...
+  neuronMinimumStake,
+}: {
+  rootCanisterId: Principal;
+  neuronId: SnsNeuronId;
+  amount: E8s;
```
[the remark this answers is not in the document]
mstrasinskis: Done. `numberToE8s` applied.

frontend/src/lib/services/sns-neurons.services.ts
mstrasinskis: The neuron list calls `syncSnsNeurons` using queryAndUpdate. This should provide an updated list.

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: Good catch, relic. Removed.

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: Good point. Is moved.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
```diff
@@ -48,7 +69,7 @@
   };
 </script>
 
-{#if nonNullish(neuron) && nonNullish(neuronState)}
+{#if nonNullish(neuron) && nonNullish(neuronState) && nonNullish(parameters)}
```
peterpeterparker, line 72: Doesn't this make the layout shift? Particularly if loading the paramters is slow.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
```diff
@@ -63,6 +84,19 @@
     <SnsNeuronAge {neuron} />
 
     <SnsNeuronStateRemainingTime {neuron} inline={false} />
+
+    <div class="buttons">
+      {#if ENABLE_SNS_2}
```
peterpeterparker, line 89: nth: `{#if ENABLE_SNS_2 && allowedToSplit}`

frontend/src/lib/i18n/en.json
```diff
@@ -700,7 +699,8 @@
     "sns_remove_followee": "There was an error while unfollowing the neuron.",
     "sns_load_functions": "There was an error while loading the project topics.",
     "sns_add_hotkey": "There was an error adding the hotkey.",
-    "sns_stake_maturity": "There was an error while staking the maturity of the neuron."
+    "sns_stake_maturity": "There was an error while staking the maturity of the neuron.",
+    "sns_amount_not_enough_stake_neuron": "Sorry, the amount is too small. You need a minimum of $minimum $token to stake a neuron"
```
peterpeterparker, line 703: `.` at the end of the sentence?

frontend/src/lib/modals/sns/neurons/SnsNeuronModals.svelte
```diff
@@ -117,6 +129,20 @@
         />
       {/if}
 
+      {#if parameters !== undefined && transactionFee !== undefined}
```
peterpeterparker, line 132: `isNullish` instead of `!== undefined`?

frontend/src/lib/modals/sns/neurons/SplitSnsNeuronModal.svelte
```diff
@@ -0,0 +1,133 @@
...
+    const { success } = await splitNeuron({
+      rootCanisterId,
+      neuronId: fromDefinedNullable(neuron.id),
+      amount,
+      transactionFee,
+      neuronMinimumStake,
+      token,
+    });
+
+    await reloadNeuron();
```
peterpeterparker, line 77: we reload the neuron on error too?

frontend/src/lib/pages/SnsNeuronDetail.svelte
```diff
@@ -70,6 +91,14 @@
     }
   };
 
+  $: rootCanisterId,
+    (() => {
+      if (rootCanisterId !== undefined) {
+        loadSnsParameters(rootCanisterId);
```
peterpeterparker, line 97: why reloading params and transaction fee but not accounts?

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -295,6 +299,76 @@ export const removeHotkey = async ({
...
+    // TODO: Get identity depending on account to support HW accounts
+    const identity = await getNeuronIdentity();
+    // reload neurons (should be actual for nextMemo calculation)
+    await loadNeurons({
+      rootCanisterId,
+      certified: true,
+    });
+    const neurons = get(snsNeuronsStore)[rootCanisterId.toText()]
+      .neurons as SnsNeuron[];
+    const memo = nextMemo({
```
peterpeterparker, line 349: why is the next memo not computed by the backend?
or what happens if the next memo used is already taken when processed by the backend?

frontend/src/lib/utils/token.utils.ts
```diff
@@ -76,6 +76,16 @@ export const formattedTransactionFeeICP = (fee: number | bigint): string =>
     }).toE8s(),
   });
 
+// To make the fixed transaction fee readable, we do not display it with 8 digits but only till the last digit that is not zero
+// e.g. not 0.00010000 but 0.0001
+export const formattedTransactionFee = (fee: TokenAmount): string =>
+  formatToken({
+    value: TokenAmount.fromE8s({
+      amount: BigInt(fee.toE8s()),
```
peterpeterparker, line 84: the cast is needed? isn't `fee.toE8s()` already a `BigInt`?

frontend/src/lib/services/sns-neurons.services.ts
mstrasinskis: Here is the second attempt to discuss this w/ `nns` - https://dfinity.slack.com/archives/C0175CHJ0P6/p1672220542021449.

peterpeterparker: Thanks. Can you add a comment that there is a loop hole in this flow?

i.e. generally speaking, after the payment flow discussions, I think every time we are "forced" in the frontend to evaluate something that in our opinion should be computed in the backend to be more robust we should comment it explicitely in the code

frontend/src/lib/utils/token.utils.ts
mstrasinskis: Good catch, a relic.

frontend/src/lib/modals/sns/neurons/SplitSnsNeuronModal.svelte
mstrasinskis: The initial idea was to reload the neuron in any case because the `splitNeuron` is not atomic operation (update parent neuron, create new one). But not a strong opinion, updated.

peterpeterparker: So it's possible that a new neuron is created but the parent isn't updated, is it what you mean?

frontend/src/lib/modals/sns/neurons/SnsNeuronModals.svelte
mstrasinskis: Updated. Is there another gain except better syntax?

frontend/src/lib/pages/SnsNeuronDetail.svelte
mstrasinskis: Not sure I got the question, there is no account info on the neuron detail page.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
mstrasinskis: Sure, thanks! 😄

frontend/src/lib/modals/sns/neurons/SnsNeuronModals.svelte
peterpeterparker: Spontenously I thought only about the syntax, was only a suggestion too 😉. Cool, thanks for the change.

frontend/src/lib/pages/SnsNeuronDetail.svelte
peterpeterparker: e.g. we use the accounts on the neuron detail to increase stake maturity

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMetaInfoCard.svelte
mstrasinskis: Good catch, this is not needed anymore because parameters are provided as prop.

frontend/src/lib/modals/sns/neurons/SnsStakeMaturityModal.svelte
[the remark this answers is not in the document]
lmuntaner: Yes there is, in `getSnsNeuron` we also check whether it needs refresh. But it's a simpler check

frontend/src/lib/services/sns-neurons.services.ts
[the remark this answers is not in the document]
lmuntaner: Sounds good!

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: true, nns neuron we need it because it's called on init, but in SNS they are called when visiting the page, thanks!

frontend/src/lib/services/sns-neurons.services.ts
mstrasinskis: Thanks, comment was added.

frontend/src/lib/services/sns-neurons.services.ts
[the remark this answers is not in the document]
mstrasinskis: To have the same pattern was switched to using stores.

frontend/src/lib/services/sns-neurons.services.ts
mstrasinskis: Was switched to using stores.

frontend/src/lib/modals/sns/neurons/SplitSnsNeuronModal.svelte
mstrasinskis: Actually now I don't think it's an issue. The reload is moved in success block. Thanks for noticing.

frontend/src/lib/pages/SnsNeuronDetail.svelte
peterpeterparker: Discussed in chat, accounts are reloaded when the modal to increase Sns stake maturity is opened, therefore no need to reload the accounts there.

peterpeterparker (approved):
LGTM, thanks

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -116,7 +123,7 @@ export const syncSnsNeurons = async (
   });
 };
 
-const loadNeurons = async ({
+export const loadNeurons = async ({
```
lmuntaner, line 126: Is this one tested? It might not be tested because it was not exported.

Now that you are exporting it, it might make sense to add a test for it as well.

lmuntaner (changes requested):

frontend/src/lib/modals/sns/neurons/SplitSnsNeuronModal.svelte
```diff
@@ -0,0 +1,131 @@
...
+      neuronMinimumStake,
+    });
+
+    if (success) {
+      await reloadNeuron();
+
+      toastsSuccess({
+        labelKey: "neuron_detail.split_neuron_success",
+      });
+      close();
```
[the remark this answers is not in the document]
mstrasinskis: Thanks didn't know it's possible, never saw this issue during testing. Updated.

frontend/src/lib/services/sns-neurons.services.ts
mstrasinskis: Good point, done!

frontend/src/lib/components/sns-neuron-detail/actions/SplitSnsNeuronButton.svelte
```diff
@@ -0,0 +1,59 @@
...
+  import type { NervousSystemParameters } from "@dfinity/sns";
+  import { fromDefinedNullable } from "@dfinity/utils";
+  import { minNeuronSplittable } from "$lib/utils/sns-neuron.utils";
+  import { replacePlaceholders } from "$lib/utils/i18n.utils";
+  import type { E8s, Token } from "@dfinity/nns";
+  import { formatToken } from "$lib/utils/token.utils";
+  import Tooltip from "$lib/components/ui/Tooltip.svelte";
+
+  export let neuron: SnsNeuron;
+  export let parameters: NervousSystemParameters;
```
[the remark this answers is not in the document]
mstrasinskis: It's still an open question. Probably having a bunch of util functions could be a flexible solution.

PR #1660, opened 2022-12-22
enable staking sns neuron

mstrasinskis:
# Motivation

Enable staking sns neuron since the security fix was deployed.

# Changes

- remove flag from the button


peterpeterparker (approved):
Rubber stamp

PR #1662, opened 2022-12-23
fix: Set nns-dapp canister ID

bitdivine:
# Motivation
CI builds a wasm file for use on localhost.  Unfortunately it has the wrong own canister ID.  `dfx nns install` is not saving the canister ID of nns-dapp, so the build happens with a new and incorrect canister ID.

# Changes
- Capture the nns-dapp canister ID from stdout and save it.  That way the build is for the correct canister ID.

Note: This should be fixed upstream but the dfx release cycle is slow and this would be broken in the meantime.

# Tests
- I have verified that the own canister ID error is now gone, when used for local deployments.
- There is another error I see with the local deployment but that will need to be addressed in another PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,6 +81,7 @@ jobs:
           export FILENAME=".dfx/local/canister_ids.json"
           export DFX_NETWORK=local
           export DFX_CANISTER_NAME="nns-dapp"
+          test -f "$FILENAME" || { echo "{}" > "$FILENAME" ; }
           jq '. * {(env.DFX_CANISTER_NAME): {(env.DFX_NETWORK): (env.DFX_CANISTER_ID)}}' "$FILENAME" | grep . | sponge "$FILENAME"
           dfx canister id nns-dapp
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -76,6 +76,7 @@ jobs:
 
       - name: Set nns-dapp canister ID
         run: |
+          set -x
           sudo apt-get install -yy moreutils
           export DFX_CANISTER_ID="$(sed -nr 's,^nns-dapp +http://([-a-z0-9]+).*,\1,g;ta;b;:a;p;q' ,nns-install)"
           export FILENAME=".dfx/local/canister_ids.json"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -72,7 +72,20 @@ jobs:
       # IDENTITY_SERVICE_URL needs to match `e2e-tests/test.js`
       - name: Deploy NNS canisters
         run: |
-          dfx nns install
+          dfx nns install | tee ,nns-install
```
lmuntaner, line 75: Ths nns-dapp deployed in this step is never used in the CI, instead I deploy another nns-dapp and use that one for the e2e tests.

That's why the CI is green despite `dfx nns install` not installing a valid nns-dapp.

Do you see a reason go use the nns dapp installed in this step?

Anyway, a comment that we're not using the nns-dapp installed in this step is probably worth adding.

bitdivine: The important thing is that the canister id should be the same.  The wasm that is built and tested here is then published for use by `dfx nns install`; that will fail if the canister ID doesn't match.  I can add a comment to this effect.  It might also make sense to install dfx via `dfx nns install` so that the e2e tests don't pass unless the build is fit for publishing.

bitdivine: I have added a comment as discussed.

Note regarding deploying the nns-dapp using `dfx nns deploy`: There is a similar requirement that the II canister ID in this build must match the one provided by dfx nns deploy.  The II provided by dfx nns deploy is however old and doesn't match the one specified in dfx.json so there is a PR to deploy "our" version of II in `dfx nns deploy`.  https://github.com/dfinity/nns-dapp/pull/1679  We could deploy the nns-dapp in the same way, for e2e tests.

lmuntaner: Let me rephrase to see whether I understand.

When the CI deploys the nns-dapp with `./deploy.sh --nns-dapp local`, it will deploy it to the same canister id as the installation of `dfx nns install`.

Is that correct?

I'm not sure we want the e2e test to use the same way. We want e2e tests to run against the current branch. We deploy the new wasm if the e2e pass, but if we need the wasm to test it, that means we need to disable that and creates a circular dependecy of the CI on itself.

Moreover, IMHO, it would be good to have the same way to run e2e tests locally in a branch than in the CI.

bitdivine: A medium term goal is to make test environments like prod, by default, for both testnets and CI.  At present we have at least three different ways of deploying.  There are a number of PRs in progress trying to make testnet and local deployments the same, and also to get the production canister versions and use them in testnets, local dfx servers and CI.  It will always be possible to create different environments but by default I'm trying to make the test environment match prod.

Ok, so that aside:
> When the CI deploys the nns-dapp with ./deploy.sh --nns-dapp local, it will deploy it to the same canister id as the installation of dfx nns install.
Yes, that is the idea.  There are moves to make the nns-dapp canister ID everywhere match prod.  Going straight there has issues with the canister controller.  Plan is to make all the testnets use the same canister ID, then make one update that moves them all to the value used on mainnet.

> We deploy the new wasm if the e2e pass...

lmuntaner: Ok, I see where this is going.

I might still have some questions tomorrow. But they are more related to the upcoming changes than the ones in this PR.

Already approving.

lmuntaner (approved):

bitdivine:
This was superseded by another PR.  Closing.

PR #1668, opened 2022-12-28
GIX-1151: Identify SNS top up neuron transaction

lmuntaner:
# Motivation

Differentiate the transaction of staking and topping up SNS neurons in the Wallet page.

# Changes

* Add case in `getSnsTransactionType` for sns transactions that differentiates staking and topping up a neuron.

# Tests

* Add test for new case.

### DIFF
--- a/frontend/src/lib/utils/sns-transactions.utils.ts
+++ b/frontend/src/lib/utils/sns-transactions.utils.ts
@@ -115,7 +115,11 @@ const getSnsTransactionType = ({
   if (transfer !== undefined) {
     // A transaction to an account owned by the governance canister stakes a neuron.
     if (transfer.to.owner.toText() === governanceCanisterId?.toText()) {
-      return AccountTransactionType.StakeNeuron;
+      // Staking a neuron uses a memo, but topping up a neuron does not.
+      if (transfer.memo.length > 0) {
+        return AccountTransactionType.StakeNeuron;
+      }
+      return AccountTransactionType.TopUpNeuron;
     }
     return AccountTransactionType.Send;
   }
--- a/frontend/src/tests/lib/components/accounts/SnsTransactionCard.spec.ts
+++ b/frontend/src/tests/lib/components/accounts/SnsTransactionCard.spec.ts
@@ -80,6 +80,7 @@ describe("SnsTransactionCard", () => {
       subaccount: [Uint8Array.from([0, 0, 1])] as [Uint8Array],
     };
     const stakeNeuronTransaction = createSnstransactionWithId(toGov, from);
+    stakeNeuronTransaction.transaction.transfer[0].memo = [new Uint8Array()];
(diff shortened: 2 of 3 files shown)


lmuntaner:
New Tx:

Screenshot 2022-12-28 at 08 09 56

I found another rounding error when the transaction was sent. I'll look into it in another PR.

lmuntaner:

mstrasinskis (approved):
LGTM, thanks!

frontend/src/lib/utils/sns-transactions.utils.ts
```diff
@@ -115,7 +115,11 @@ const getSnsTransactionType = ({
   if (transfer !== undefined) {
     // A transaction to an account owned by the governance canister stakes a neuron.
     if (transfer.to.owner.toText() === governanceCanisterId?.toText()) {
-      return AccountTransactionType.StakeNeuron;
+      // Staking a neuron uses a memo, but topping up a neuron does not.
+      if (transfer.memo.length > 0) {
```
peterpeterparker, line 119: is there a way to identity the memo used to stake or topup neuron more accuratly?

it does not feel reliable for the future since memo can be used for other purpose - i.e. if another function transfer Sns tokens with a memo we might not be informed and therefore we might display such transfer as neuron related operation

lmuntaner: There is no memo when topping up a neuron.

But I understand, the problem here is that we rely a lot also in the design of how staking and topping up works.

Ideally we should be able to have the type of transaction without having to second guess data. But for now, this is how we do it for NNS and SNS as well.

peterpeterparker: I meant that memo can be use - manually or in the future - for something else than neuron transactions only.

e.g. I can transfer with dfx or else some Sns tokens from account identifier A (not in NNS-dapp) to account identifier B (an account of NNS-dapp) and for some reason I provide a memo because I feel like I am in the mood to set something in the memo field -> this transfer will be displayed as a Stake or TopUp Neuron in NNS-dapp even though it has nothing to do with neuron.

Thanks for the answer. It is what it is I guess.

lmuntaner: Before checking the memo, we check that the destination account is the governance canister.

To be set as Top Up neuron, the transfer needs to be destination governance canister and have a memo. Not just a memo.

peterpeterparker: Fck me, I focused on the lines changed and I missed the `if (transfer.to.owner.toText() === governanceCanisterId?.toText()) {`. Thanks!

lmuntaner: I guessed so, haha

PR #1669, opened 2022-12-28
Fix: Rounding error in sns transactions

lmuntaner:
# Motivation

Fix two bugs:

* Rounding issue in SNS transactions.
* Account balance was not updated after staking or topping up sns neurons.

# Changes

* Change implementation of `numberToE8s` to use `TokenAmount.fromNumber`.
* Call `loadSnsAccounts` service after staking or topping up sns neuron.

# Tests

* Fix test.
* Add test for new test cases of loadSnsAccounts call.
* Add tests for token utils `numberToE8s`.

### DIFF
--- a/frontend/src/lib/utils/token.utils.ts
+++ b/frontend/src/lib/utils/token.utils.ts
@@ -123,6 +123,15 @@ export const convertTCyclesToIcpNumber = ({
   exchangeRate: bigint;
 }): number => tCycles / (Number(exchangeRate) / NUMBER_XDR_PER_ONE_ICP);
 
+/**
+ * Returns the number of E8s for the given amount.
+ *
+ * E8s have a precision of 8 decimals. An error is thrown if the amount has more than 8 decimals.
+ *
+ * @param {number} amount
+ * @returns {bigint}
+ * @throws {Error} If the amount has more than 8 decimals.
+ */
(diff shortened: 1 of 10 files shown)


lmuntaner:

frontend/src/lib/utils/token.utils.ts
```diff
@@ -123,5 +123,17 @@ export const convertTCyclesToIcpNumber = ({
   exchangeRate: bigint;
 }): number => tCycles / (Number(exchangeRate) / NUMBER_XDR_PER_ONE_ICP);
 
+/**
+ * Returns the number of E8s for the given amount.
+ *
+ * E8s have a precision of 8 decimals. An error is thrown if the amount has more than 8 decimals.
```
mstrasinskis, line 129: Is it possible to round the number in case of more than 8 decimals to avoid error handling?

mstrasinskis (approved):
LGTM, thanks!

frontend/src/lib/utils/token.utils.ts
lmuntaner: I believe that the decision done in ic-js to return an error (then raised) is a good option: https://github.com/dfinity/ic-js/blob/main/packages/nns/src/token.ts#L11

PR #1672, opened 2022-12-30
Fix: Dissolved neuron if dissolved in the past

lmuntaner:
# Motivation

User can't disburse sns neurons because data shows "WhenDissolvedTimestampSeconds" state.

# Changes

* Check whether WhenDissolvedTimestampSeconds is in the past and set it to Unlocked state.

# Tests

* Add test with new case.

### DIFF
--- a/frontend/src/lib/utils/sns-neuron.utils.ts
+++ b/frontend/src/lib/utils/sns-neuron.utils.ts
@@ -38,7 +38,9 @@ export const getSnsNeuronState = ({
       : NeuronState.Locked;
   }
   if ("WhenDissolvedTimestampSeconds" in dissolveState) {
-    return dissolveState.WhenDissolvedTimestampSeconds < Date.now() / 1000
+    // In case `nowInSeconds` ever changes and doesn't return an integer we use Math.floor
+    return dissolveState.WhenDissolvedTimestampSeconds <
+      BigInt(Math.floor(nowInSeconds()))
       ? NeuronState.Dissolved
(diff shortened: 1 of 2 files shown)


frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -38,7 +38,9 @@ export const getSnsNeuronState = ({
       : NeuronState.Locked;
   }
   if ("WhenDissolvedTimestampSeconds" in dissolveState) {
-    return NeuronState.Dissolving;
+    return dissolveState.WhenDissolvedTimestampSeconds < Date.now() / 1000
```
mstrasinskis, line 41: nowInSeconds()?

mstrasinskis (commented):
LGTM, just one comment.

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: Thanks, done! Also transform to bigint before comparing it

mstrasinskis (approved):

peterpeterparker:
in same function:

```
if (dissolveState === undefined) {
    return NeuronState.Dissolved;
  }
```

is still accurate?

lmuntaner:
> in same function:
> 
> ```
> if (dissolveState === undefined) {
>     return NeuronState.Dissolved;
>   }
> ```
> 
> is still accurate?

Yes, that's still accurate. The initial neuron that was never locked has that state

PR #1677, opened 2023-01-02
feat: split layout and new token selector for "Accounts", "Neurons" and "Proposals" views

peterpeterparker:
# Motivation

Introduce a new navigation to select universe on desktop and build a custom dropdown for mobile.

# UX

This feature also introduces a more comprehensive name instead of the term "universe" or "project" as those are unknow for the users. Insted, the user will switch universe/project by selecting the related term "Token".

# PRs

- [x] https://github.com/dfinity/gix-components/pull/143

# Changes

## Features

- extract cmp `AccountsBalance` to display the balance in various places
- new `<Content />` and `<SplitContent />` to display either the existing content or the new design with a left column
- update gix-cmp `Card` breaking changes (few properties modified)
- update `Summary` to display the title in existing dropdown but also for new use cases
- create a `<SummaryLogo />` component to display either a nns or project logo with various size
- `SkeletonCard` can be rendered with few, more or more lines this to avoid layout shift on "Accounts" loading which would not be aligned with the new left column nav / universe selection
- adapt `framed` option of the logo as in design
- on universe project dropdown selection or in navigation click, switch selected universe
- introduces new derived stores to list the Nns + Sns projects and the total balance of the selected universe
- add various new `<SelectUniverse... />` components and modal to implement to requested feature
- use "Select Token" for call to action (not "Select universe or project")
- extract few utilities - e.g. `sumAccounts` and `buildSwitchUniverseUrl`
- switch between `Content` and `SplitContent` in "Accounts" layout
- display "My Tokens" title only on mobile devices
- update all route layout to use `<Content />` (because code has been extracted to a new component)

## Fix

- user drag of the logo image

## Refactor

- as I've added more layout components, I move the related cmp to `$lib/components/layout` and title to `$lib/components/header`

# Tests

While I provide few tests, not that I do not provide tests for those features that are related to `<svelte:window bind:innerWidth />`. For this we would need e2e that handle window size.

# Screenshots

Capture d’écran 2023-01-09 à 09 23 46
Capture d’écran 2023-01-09 à 09 23 50
Capture d’écran 2023-01-09 à 09 24 03
Capture d’écran 2023-01-09 à 09 24 05
Capture d’écran 2023-01-09 à 09 24 11
Capture d’écran 2023-01-09 à 09 24 13
Capture d’écran 2023-01-09 à 09 24 16
Capture d’écran 2023-01-09 à 09 24 21


frontend/src/lib/components/accounts/AccountsBalance.svelte
```diff
@@ -0,0 +1,30 @@
...
+          value: balance.toE8s(),
+          detailed: true,
+        })
+      : "";
+</script>
+
+{#if balance !== undefined}
+  <Tooltip
+    id="wallet-total-icp"
+    text={replacePlaceholders($i18n.accounts.current_balance_total, {
```
mstrasinskis, line 24: For reusability the token could be also provided here:
"current_balance_total": "Total balance: $amount `ICP`",

frontend/src/lib/components/login/LoginMenuItems.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { MenuItem, IconNorthEast, IconGitHub } from "@dfinity/gix-components";
```
mstrasinskis, line 2: 👍

frontend/src/lib/components/accounts/AccountsBalance.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
```
lmuntaner, line 1: This could be used to display any kind of Token, right?

Why not the name `TokenDisplay`?

`AccountBalance` makes it look as if it's related to the accounts.

frontend/src/lib/components/layout/SplitContent.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+  import { SplitContent } from "@dfinity/gix-components";
+  import Title from "$lib/components/header/Title.svelte";
+  import AccountMenu from "$lib/components/header/AccountMenu.svelte";
+  import SelectUniverseNav from "$lib/components/universe/SelectUniverseNav.svelte";
```
mstrasinskis, line 5: Probably we could refactor a bit to use a single internal name for new-tokens: `universe` or `project`. Not a call to action, just thinking aloud :)

frontend/src/lib/components/ic/AmountDisplay.svelte
```diff
@@ -23,7 +23,6 @@
   class:plus-sign={sign === "+"}
   data-tid="token-value-label"
 >
-  <slot />
```
lmuntaner, line 26: Not in this PR, but this looks very similar in responsibility to AccountBalance.

frontend/src/lib/components/ui/Logo.svelte
```diff
@@ -22,16 +23,25 @@
     height: var(--padding-3x);
 
     display: var(--logo-display);
+
+    user-select: none;
```
mstrasinskis, line 27: Is this a requirement or UX improvement?

frontend/src/lib/components/universe/SelectUniverseDropdown.svelte
```diff
@@ -0,0 +1,59 @@
+<script lang="ts">
+  import { SkeletonText, BREAKPOINT_LARGE } from "@dfinity/gix-components";
+  import { snsProjectSelectedStore } from "$lib/derived/selected-project.derived";
+  import SelectUniverseCard from "$lib/components/universe/SelectUniverseCard.svelte";
+  import { selectedProjectBalance } from "$lib/derived/selected-project-balance.derived";
+  import AmountDisplay from "$lib/components/ic/AmountDisplay.svelte";
+  import { nonNullish } from "$lib/utils/utils";
+  import SelectUniverseModal from "$lib/modals/universe/SelectUniverseModal.svelte";
+
+  export let selectedCanisterId: string;
```
lmuntaner, line 10: Not in this PR, but it would be good to understand when we want to use props or access the global store for the "selected" universe.

For example, in this component you use props and the global store.

That means that we could pass a selectedCanisterId that does not match with the summary in `snsProjectSelectedStore` and it could lead to bugs.

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
```diff
@@ -68,6 +68,17 @@ describe("ProjectCard", () => {
...
 
+    it("should be highlighted", () => {
+      const { container } = render(ProjectCard, {
+        props: {
+          project: mockSnsFullProject,
+        },
+      });
+
+      const article = container.querySelector("article.highlighted");
+      expect(article).not.toBeNull();
```
lmuntaner, line 79: Why not `.toBeInTheDocument()`

I don't have a strong preference, but I was wondering whether there was a reason to prefer this one. I've seen it many times.

frontend/src/lib/components/layout/SplitContent.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+  import { SplitContent } from "@dfinity/gix-components";
```
lmuntaner, line 2: How about renaming this component?

This is Content that it's specific to one universe. It's not about splitting any content.

Maybe `UniverseSplitContent`? `TokenedSplitContent`?

Just to understand that using this means using the Token navigation.

lmuntaner (commented):
Added some comments.

I'm still not approving because I don't think I've been involved enough in the UI part, and probably it makes sense for @mstrasinskis to take a look before merging.

Is it safe to say that the main components that should be closely reviewed are?

- SplitContent
- SelectUniverseNav
- SelectUniverseNavList
- SelectUniverseDropdown
- SelectUniverseModal
- SelectUniverseCard

mstrasinskis (approved):
LGTM, just a couple of minor comments.
Thank you for the additional small fixes/improvements 👍

frontend/src/lib/components/accounts/AccountsBalance.svelte
peterpeterparker: It's the total of the accounts. It notably contains an explicit label and tooltip related to a wallet

```
<Tooltip
    id="wallet-total-icp"
    text={replacePlaceholders($i18n.accounts.current_balance_total, {
      $amount: totalTokens,
    })}
  >
```

Therefore in its current state it makes sense to me to named it `AccountsBalance`.

frontend/src/lib/components/ic/AmountDisplay.svelte
peterpeterparker: There is no `AccountBalance` - there is `AccountsBalance` - and in comparison to `AmountDisplay` there are really different so not sure which part you are referring to?

frontend/src/lib/components/universe/SelectUniverseDropdown.svelte
peterpeterparker: That's a good point and an issue. I'll remove the exported props.

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
peterpeterparker: I am used to develop apps and libs which manipulate the DOM in vanilla JS, therefore to me checking if a node is null is common and the closest to js but, that's for sure a personal choice. I don't see neither advantages of one or the other.

frontend/src/lib/components/layout/SplitContent.svelte
peterpeterparker: A bit verbose and implicitely given by the usage in the route group `/(app)/(u)/(list)` but no strong opinion so renamed to `UniverseSplitContent`.

frontend/src/lib/components/accounts/AccountsBalance.svelte
peterpeterparker: That's actually a bug catch @mstrasinskis !!!

As the label is already `"current_balance_total": "Total balance: $amount ICP",` and used in the title that is used in `SnsAccounts.svelte`, the tooltip is currently not accurate for Sns1 on mainnet.

Will fix that here.

peterpeterparker: Same for `$i18n.accounts.current_balance_detail` used on the wallet page

frontend/src/lib/components/layout/SplitContent.svelte
peterpeterparker: At first I used "Project" but after thinking about it, I thought about the upcoming ckBtc so thought that a "Project" is implicitely a "Sns Project" and ckBtc might be something else, that's why in anticipation I thought about using already "Universe". Which is also close to the routing. Let's iteratively clean and rename.

frontend/src/lib/components/ui/Logo.svelte
peterpeterparker: It's a "David UX requirement" 😉. I was reviewing and discussing the UX with Mischa when I noticed that I was able to drag the logo, which I personaly dislike. So I told Mischa that I would like to avoid being able to drag the logo and he wasn't against nor for, so I went on.

peterpeterparker: Let me know if does not work for you.

peterpeterparker:
> Is it safe to say that the main components that should be closely reviewed are?

@lmuntaner yes

PR #1680, opened 2023-01-03
GIX-1137: Confirmation remove current user hotkey

lmuntaner:
# Motivation

User should confirm before removing itself from hotkeys. Both NNS and SNS neurons.

# Changes

* Rename NeuronHotkeysCard.svelten to NnsNeuronHotkeysCard.svelte.
* Add extra function to [Sns / Nns]NeuronHotkeysCard called "maybeRemove" to check whether it should open the modal or remove hotkey.
* New UI component ConfirmRemoveCurrentUserHotkey to share modal in Sns and Nns.
* New state `showConfirmationHotkey` in [Sns / Nns]NeuronHotkeysCard to open ConfirmRemoveCurrentUserHotkey.

# Tests

* Change tests after file rename.
* Add test case to NnsNeuronHotkeysCard.
* Add test case to SnsNeuronHotkeysCard.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
@@ -25,6 +25,10 @@
   import CardInfo from "$lib/components/ui/CardInfo.svelte";
   import Tooltip from "$lib/components/ui/Tooltip.svelte";
   import AddSnsHotkeyButton from "./actions/AddSnsHotkeyButton.svelte";
+  import { toastsShow } from "$lib/stores/toasts.store";
+  import { goto } from "$app/navigation";
+  import { neuronsPathStore } from "$lib/derived/paths.derived";
+  import ConfirmRemoveCurrentUserHotkey from "$lib/modals/neurons/ConfirmRemoveCurrentUserHotkey.svelte";
 
   const { reload, store }: SelectedSnsNeuronContext =
     getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
@@ -50,6 +54,22 @@
   let showTooltip: boolean;
   $: showTooltip = hotkeys.length > 0 && canManageHotkeys;
 
+  let currentIdentityString: string | undefined;
+  $: currentIdentityString = $authStore.identity?.getPrincipal().toText();
+
+  let showConfirmationHotkey: string | undefined;
+  const closeConfirmation = () => {
+    showConfirmationHotkey = undefined;
+  };
+  const maybeRemove = async (hotkey: string) => {
+    // Require confirmation if the user is removing itself from the hotkeys.
+    if (currentIdentityString === hotkey) {
+      showConfirmationHotkey = hotkey;
+    } else {
+      remove(hotkey);
+    }
+  };
+
   const remove = async (hotkey: string) => {
     // Edge case: Remove button is shwon only when neuron is defined
     if (neuronId === undefined) {
@@ -63,6 +83,15 @@
       hotkey,
       rootCanisterId: $snsProjectIdSelectedStore,
     });
+    // If the user removes itself from the hotkeys, it has no more access to the detail page.
+    if (currentIdentityString === hotkey && success) {
+      toastsShow({
+        level: "success",
+        labelKey: "neurons.remove_hotkey_success",
+      });
+
+      await goto($neuronsPathStore);
+    }
     if (success) {
       await reload();
     }
@@ -103,7 +132,7 @@
               <button
                 class="text"
                 aria-label={$i18n.core.remove}
-                on:click={() => remove(hotkey)}
+                on:click={() => maybeRemove(hotkey)}
                 data-tid="remove-hotkey-button"
                 ><IconClose size="18px" /></button
               >
@@ -118,6 +147,15 @@
       </div>
     {/if}
   </CardInfo>
+{/if}
+
+{#if showConfirmationHotkey !== undefined}
+  <!-- The extra const is required for TS to understand that showConfirmationHotkey is a string, not undefined -->
+  {@const hotkey = showConfirmationHotkey}
+  <ConfirmRemoveCurrentUserHotkey
+    on:nnsClose={closeConfirmation}
+    on:nnsConfirm={() => remove(hotkey)}
+  />
 {/if}
 
 <style lang="scss">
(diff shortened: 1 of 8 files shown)


lmuntaner:
New UX: (same for Nns and Sns)

https://user-images.githubusercontent.com/4550653/210385490-e90a2d14-631b-473d-9fe8-ddb5367aa847.mov

New UI:

Screenshot 2023-01-03 at 16 09 36

lmuntaner:

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
```diff
@@ -50,6 +54,22 @@
...
   $: showTooltip = hotkeys.length > 0 && canManageHotkeys;
 
+  let currentIdentityString: string | undefined;
+  $: currentIdentityString = $authStore.identity?.getPrincipal().toText();
+
+  let showConfirmationHotkey: string | undefined;
+  const closeConfirmation = () => {
+    showConfirmationHotkey = undefined;
+  };
+  const maybeRemove = async (hotkey: string) => {
```
peterpeterparker, line 64: `async` can be removed

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
```diff
@@ -50,6 +54,22 @@
...
+    // Require confirmation if the user is removing itself from the hotkeys.
+    if (currentIdentityString === hotkey) {
+      showConfirmationHotkey = hotkey;
+    } else {
+      remove(hotkey);
```
peterpeterparker, line 69: this should be `await` no?

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
```diff
@@ -50,6 +54,22 @@
   let showTooltip: boolean;
   $: showTooltip = hotkeys.length > 0 && canManageHotkeys;
 
+  let currentIdentityString: string | undefined;
+  $: currentIdentityString = $authStore.identity?.getPrincipal().toText();
+
+  let showConfirmationHotkey: string | undefined;
```
peterpeterparker, line 60: just thinking out at loud in case you would find it interesting: instead of `string | undefined` maybe we could use a type. e.g.

```
interface SelectedHotkey {
   identityString: string;
   needConfirmation: boolean
}

// instead of let currentIdentityString
let hotkey: SelectedHotkey | undefined;
```

That way we would spare the duplicated code `currentIdentityString === hotkey` with e.g. `hotkey.needConfirmation`.

Again just thinking out at loud.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
```diff
@@ -120,6 +149,15 @@
   </CardInfo>
 {/if}
 
+{#if showConfirmationHotkey !== undefined}
+  <!-- The extra const is required for TS to understand that showConfirmationHotkey is a string, not undefined -->
```
peterpeterparker, line 153: isn't this a hack? i.e. doesn't it work out because Svelte `@const` doesn't inherit the type and one day, if they do so, it will not compile anymore?

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
lmuntaner: I would still need to check that `hotkey` is not undefined: `hotkey?.needConfirmation` is undefined when `hotkey` is undefined. And we have the lint rule to check that always.

It would not spare much.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
lmuntaner: yes, thanks!

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
```diff
@@ -63,6 +83,15 @@
...
       rootCanisterId: $snsProjectIdSelectedStore,
     });
+    // If the user removes itself from the hotkeys, it has no more access to the detail page.
+    if (currentIdentityString === hotkey && success) {
+      toastsShow({
+        level: "success",
+        labelKey: "neurons.remove_hotkey_success",
+      });
+
+      await goto($neuronsPathStore);
```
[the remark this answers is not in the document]
lmuntaner: yes, it makes sense

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
lmuntaner: Yes, it's a hack.

The problem was when using `showConfirmationHotkey` in `on:nnsConfirm={() => remove(hotkey)}`.

Weirdly, TS was not smart enough to understand the if statement below when used inside the `on:nnsConfirm`.

If `@const` insherits the type, it should also take into account that `showConfirmationHotkey !== undefined`, no? Therore, it's always `string`.

lmuntaner: I get this error when `npm run check` is run:

```
Error: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'. (ts)
    on:nnsClose={closeConfirmation}
    on:nnsConfirm={() => remove(showConfirmationHotkey)}
  />
```

lmuntaner: I guess it's because the variable is used inside the function passed to `on:nnsConfirm`, instead of being used directly.

With `{@const` we are setting the value at that time, which is a string, and using that one.

It kind of makes sense.

lmuntaner:

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
peterpeterparker: Sure

frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
peterpeterparker: Yeah I'm fine with it, I just meant that we have no guarantee that in X months after updating Svelte this workaround/hack will still work.

peterpeterparker (approved):
LGTM, thx

PR #1681, opened 2023-01-03
Test local build

bitdivine:
# Motivation
We wish to update e2e tests to make them more realistic.  This PR focuses on the internet identity.
- Current e2e tests use the version of II that is bundled with dfx.  This is not the version in production, not the version used on testnets and not the version specified in the local dfx.json

# Changes
- Install the version of II specified in the local `dfx.json`.
- Update the e2e tests to match the newer II.
- Deploy II and the nns-dapp with `dfx nns install`.  This ensures that the canister IDs tested in the e2e tests match those depolyed by dfx and we don't need a separate build for dfx.  It also means that when `dfx nns install` deploys the front end canisters to the production canister IDs, our e2e tests can do the same with minimal effort.
  - Note: Having fewer different ways of deploying should increase the fidelity of testing and reduce the effort needed to maintain deployments, however developers are still free to deploy canisters in other ways.
- Use the upstream name for `internet_identity_dev.wasm` as normally `internet_identity.wasm` is a production wasm and this can be confusing.
- Test that the II and nns-dapp wasms installed for e2e tests are the ones we expect to be present.

# Tests
- Existing e2e tests should cover the changes.
- There is a new test that verifies the canister hashes after installation; this is in the install script used by the e2e tests, so the existing e2e tests will fail early if incorrect canisters are installed.


scripts/dfx-nns-deploy-custom
```diff
@@ -0,0 +1,45 @@
...
+rm -f nns-dapp.wasm nns-dapp_local.wasm
+
+: "Build nns-dapp for local use, with standard canister IDs, and inject into the wasm cache."
+"$SOURCE_DIR"/dfx-software-nd-cache
+
+: "Get our chosen version of II and inject it into the wasm cache."
+"$SOURCE_DIR"/dfx-software-ii-cache
+
+: "Deploy nns-dapp with other NNS canisters"
+dfx nns install
```
lmuntaner, line 28: Does this update canister_ids.json?

We need to create the config file for local environment before executing the e2e tests.

scripts/dfx-nns-deploy-custom
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
lmuntaner: Does that mean we could choose the same version of Governance and CMC for the local environment?

scripts/dfx-nns-deploy-custom
bitdivine: `dfx nns install` doesn't touch the local project and in particular doesn't touch the `canister_ids.json`.  `dfx nns import` does add the nns canisters as remotes in dfx.json.  Or rather it should; it actually misses out II and the nns-dapp.

Building the nns-dapp creates the config file, so the e2e tests should be fine and indeed they are passing.

scripts/dfx-nns-deploy-custom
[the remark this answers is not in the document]
lmuntaner: ok, thanks!

scripts/dfx-nns-deploy-custom
lmuntaner: Ah, that's true. Good point. thanks!

lmuntaner (approved):

PR #1683, opened 2023-01-04
feat: update canIdentityManageHotkeys (new permissions)

mstrasinskis:
# Motivation

TBD

- [ ] https://github.com/dfinity/ic-js/pull/251

# Changes

TBD

# Tests

TBD

### DIFF
--- a/frontend/src/lib/utils/sns-neuron.utils.ts
+++ b/frontend/src/lib/utils/sns-neuron.utils.ts
@@ -1,5 +1,5 @@
 import {
-  HOTKEY_PERMISSIONS,
+  HOTKEY_PERMISSIONS, HOTKEY_PERMISSIONS_V2,
   MAX_NEURONS_SUBACCOUNTS,
 } from "$lib/constants/sns-neurons.constants";
 import { NextMemoNotFoundError } from "$lib/types/sns-neurons.errors";
@@ -166,15 +166,23 @@ export const nextMemo = ({
 export const canIdentityManageHotkeys = ({
   neuron,
   identity,
+  parameters,
 }: {
   neuron: SnsNeuron;
   identity: Identity | undefined | null;
-}): boolean =>
-  hasPermissions({
+  parameters: NervousSystemParameters,
+}): boolean => {
+  const grantablePermissions = Array.from(fromDefinedNullable(parameters.neuron_grantable_permissions)?.permissions ?? []);
+  const hasManageVotingPermission = grantablePermissions.includes(SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_MANAGE_VOTING_PERMISSION);
+
+  return hasPermissions({
     neuron,
     identity,
-    permissions: HOTKEY_PERMISSIONS,
-  });
+    // fallback (before `NEURON_PERMISSION_TYPE_MANAGE_VOTING_PERMISSION` introduction)
+    // worst case scenario - showing the button that won’t work
+    permissions: hasManageVotingPermission ? HOTKEY_PERMISSIONS_V2 : HOTKEY_PERMISSIONS,
+  })
+}
 
 export const hasPermissionToDisburse = ({
   neuron,
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronHotkeysCard.svelte
@@ -29,6 +29,9 @@
   import { goto } from "$app/navigation";
   import { neuronsPathStore } from "$lib/derived/paths.derived";
   import ConfirmRemoveCurrentUserHotkey from "$lib/modals/neurons/ConfirmRemoveCurrentUserHotkey.svelte";
+  import type {NervousSystemParameters} from "@dfinity/sns";
+
+  export const parameters: NervousSystemParameters;
 
   const { reload, store }: SelectedSnsNeuronContext =
     getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
@@ -45,6 +48,7 @@
       ? canIdentityManageHotkeys({
           neuron,
           identity: $authStore.identity,
+          parameters,
         })
       : false;
   let hotkeys: string[];
--- a/frontend/src/lib/constants/sns-neurons.constants.ts
+++ b/frontend/src/lib/constants/sns-neurons.constants.ts
@@ -3,9 +3,17 @@ import { SnsNeuronPermissionType } from "@dfinity/sns";
 // Limit coming from the limit of subaccounts
 export const MAX_NEURONS_SUBACCOUNTS = 2 ** 16;
 
+/**
+ * @deprecated
+ * will be replaced w/ HOTKEY_PERMISSIONS_V2
+ */
 export const HOTKEY_PERMISSIONS = [
   SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_VOTE,
   SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_SUBMIT_PROPOSAL,
+];
+export const HOTKEY_PERMISSIONS_V2 = [
+  SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_MANAGE_PRINCIPALS,
+  SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_MANAGE_VOTING_PERMISSION,
 ];
 
 export const UNSPECIFIED_FUNCTION_ID = BigInt(0);
--- a/frontend/src/lib/pages/SnsNeuronDetail.svelte
+++ b/frontend/src/lib/pages/SnsNeuronDetail.svelte
@@ -155,7 +155,9 @@
         <SnsNeuronInfoStake />
         <SnsNeuronMaturityCard />
         <SnsNeuronFollowingCard />
-        <SnsNeuronHotkeysCard />
+        {#if nonNullish(parameters)}
+          <SnsNeuronHotkeysCard {parameters} />
+        {/if}
       {/if}
     </section>
   </main>
--- a/frontend/src/tests/lib/utils/sns-neuron.utils.spec.ts
+++ b/frontend/src/tests/lib/utils/sns-neuron.utils.spec.ts
@@ -58,7 +58,7 @@ import { mockNeuron } from "../../mocks/neurons.mock";
 import { nervousSystemFunctionMock } from "../../mocks/sns-functions.mock";
 import {
   createMockSnsNeuron,
-  mockSnsNeuron,
+  mockSnsNeuron, snsNervousSystemParametersMock,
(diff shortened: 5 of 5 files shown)


frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -166,15 +166,23 @@ export const nextMemo = ({
...
   identity,
+  parameters,
 }: {
   neuron: SnsNeuron;
   identity: Identity | undefined | null;
-}): boolean =>
-  hasPermissions({
+  parameters: NervousSystemParameters,
+}): boolean => {
+  const grantablePermissions = Array.from(fromDefinedNullable(parameters.neuron_grantable_permissions)?.permissions ?? []);
```
lmuntaner, line 175: Does it make sense to add a comment on top of `canIdentityManageHotkeys` now that the logic is not straighforward?

lmuntaner (commented):
Shouldn't we change the permissions when a principal is added as hotkey?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -166,15 +168,26 @@ export const nextMemo = ({
 export const canIdentityManageHotkeys = ({
```
lmuntaner, line 168: Great, thanks!

Could you add a comment explaining the business logic? I have the feeling we'll forget about it.

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -166,15 +168,26 @@ export const nextMemo = ({
...
+    grantableSet.has(permission)
+  );
+  const identityAllowedToManageHotkeys = hasPermissions({
     neuron,
     identity,
-    permissions: HOTKEY_PERMISSIONS,
+    permissions: MANAGE_HOTKEY_PERMISSIONS,
+    options: { any: true },
   });
+  return identityAllowedToManageHotkeys && hotkeyPermissionsGrantable;
```
lmuntaner, line 189: For SNS-1, I believe that if NEURON_PERMISSION_TYPE_MANAGE_VOTING_PERMISSION is not in neuron_grantable_permissions, then the user just needs to have the VOTING and SUBMIT_PROPOSAL. Right?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -286,12 +330,38 @@ export const hasPermissions = ({
...
+      const permissions = Array.from(permission_type);
+      // only those combinations define a hotkey
+      return (
+        comparePermissions({
+          permissions,
+          expected: HOTKEY_PERMISSIONS,
+        }) ||
+        comparePermissions({
+          permissions,
+          expected: [
```
lmuntaner, line 343: Should we set this as different `const` variables to better understand the idea?

frontend/src/lib/utils/sns-parameters.utils.ts
```diff
@@ -0,0 +1,133 @@
...
+  reward_rate_transition_duration_seconds: fromDefinedNullable(
+    rewardsParameters.reward_rate_transition_duration_seconds
+  ),
+  round_duration_seconds: fromDefinedNullable(
+    rewardsParameters.round_duration_seconds
+  ),
+});
+
+// TODO: move to ic-js (https://dfinity.atlassian.net/browse/GIX-1215)
+export const mapNervousSystemParameters = (
```
lmuntaner, line 79: Thanks!

frontend/src/lib/utils/sns-parameters.utils.ts
```diff
@@ -0,0 +1,133 @@
...
+  ),
+  reward_rate_transition_duration_seconds: fromDefinedNullable(
+    rewardsParameters.reward_rate_transition_duration_seconds
+  ),
+  round_duration_seconds: fromDefinedNullable(
+    rewardsParameters.round_duration_seconds
+  ),
+});
+
+// TODO: move to ic-js (https://dfinity.atlassian.net/browse/GIX-1215)
```
lmuntaner, line 78: Not sure I would move it to ic-js...

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -286,12 +330,38 @@ export const hasPermissions = ({
...
+            ...HOTKEY_PERMISSIONS,
+            SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_MANAGE_VOTING_PERMISSION,
+            SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_MANAGE_PRINCIPALS,
+          ],
+        }) ||
+        comparePermissions({
```
lmuntaner, line 349: Is there a test for this case?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -286,12 +330,38 @@ export const hasPermissions = ({
...
+          permissions,
+          expected: [
+            ...HOTKEY_PERMISSIONS,
+            SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_MANAGE_VOTING_PERMISSION,
+          ],
+        }) ||
+        comparePermissions({
```
lmuntaner, line 356: Ditto, is there a case for this one?

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: Is there a test for this case?

lmuntaner (commented):
Some comments, looking good!

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: Good point. Done.

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: This is not implemented. I'm going to discuss this today on the standup meeting.

lmuntaner: ok, sounds good

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: Updated.

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: Updated.

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: More tests were added, thanks 👍

frontend/src/lib/utils/sns-parameters.utils.ts
mstrasinskis: Since I have no strong opinion the extra task is removed.

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: good point, done 👍

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -163,18 +165,44 @@ export const nextMemo = ({
...
 }: {
   neuron: SnsNeuron;
   identity: Identity | undefined | null;
-}): boolean =>
-  hasPermissions({
+  parameters: NervousSystemParameters;
+}): boolean => {
+  const { neuron_grantable_permissions } =
+    mapNervousSystemParameters(parameters);
+  const isSns1 = neuron_grantable_permissions.includes(
```
mstrasinskis, line 188: Do we even need this?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -163,18 +165,44 @@ export const nextMemo = ({
...
+  const grantableSet = new Set(neuron_grantable_permissions);
+  const hotkeyPermissionsGrantable = HOTKEY_PERMISSIONS.every((permission) =>
+    grantableSet.has(permission)
+  );
+  const identityAllowedToManageHotkeys = hasPermissions({
     neuron,
     identity,
-    permissions: HOTKEY_PERMISSIONS,
+    permissions: MANAGE_HOTKEY_PERMISSIONS,
+    options: { any: true },
```
peterpeterparker, line 199: Out of curiosity, have you check if `any` is a TypeScript reserved keyword?

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -245,17 +273,27 @@ export const hasPermissionToSplit = ({
     permissions: [SnsNeuronPermissionType.NEURON_PERMISSION_TYPE_SPLIT],
   });
 
-/*
+/**
  * Returns true if the neuron contains provided permissions
+ * By default neuron should have all of `permissions`. This can be changed w/ `options.any` flag.
+ *
+ * @param id
```
peterpeterparker, line 280:
```
 * @param {Object} param
 * @param {SnsNeuronId} param.id
 etc.
 ```

File: frontend/src/lib/utils/sns-parameters.utils.ts [hunk 12]
```diff
@@ -0,0 +1,132 @@
...
+  VotingRewardsParameters,
+} from "@dfinity/sns/dist/candid/sns_governance";
+import { fromDefinedNullable } from "@dfinity/utils";
+
+interface DefaultFolloweeMap {
+  functionId: bigint;
+  followees: NeuronId[];
+}
+
+interface VotingRewardsParametersMap {
```
peterpeterparker, line 17: - is there a way to build these two interfaces using TypeScript utilities (`Pick<..>` etc.) to avoid hardcoding their declartion?

- as you are returning these shouldn't they be exported (`export interface...`)  and find places in `types` rather than within a `util`?

frontend/src/lib/utils/sns-neuron.utils.ts
mstrasinskis: The permission logic was updated so it works both for sns-1 and after-upgrade versions.

frontend/src/lib/utils/sns-parameters.utils.ts
mstrasinskis: Replaced with discussed in slack version.

frontend/src/lib/utils/sns-neuron.utils.ts
peterpeterparker: Currently used in our editor and it can throws a lint error I think.

peterpeterparker: Thanks

frontend/src/lib/utils/sns-parameters.utils.ts
peterpeterparker: Thanks, looks good 👍

peterpeterparker (approved):
LGTM, thanks!

PR #1684, opened 2023-01-05
Feat: Support stake maturity hw

lmuntaner:
# Motivation

Users with hardware wallet can turn auto-stake maturity on and off and use the endpoint stake_matyrity instead of merge_maturity with the redirection

# Changes

* Enable AutoStakeMaturity always.
* Add constant with new HW version.
* Change message of version not supported to make it more general.
* NnsStakeMaturityModal uses always service stakeMaturity.
* Add a version assert in service stakeMaturity.

# Tests

* Change test for NnsStakeMaturityModal
* Add case for stakeMaturity service.
* Add case for autoStakeMaturity service.

### DIFF
--- a/frontend/src/lib/constants/neurons.constants.ts
+++ b/frontend/src/lib/constants/neurons.constants.ts
@@ -9,6 +9,7 @@ export const SPAWN_VARIANCE_PERCENTAGE = 0.95;
 
 // HW versions
 export const MIN_VERSION_STAKE_MATURITY_WORKAROUND = "2.0.7";
+export const MIN_VERSION_STAKE_MATURITY = "2.2.1";
 
 export const DISSOLVE_DELAY_MULTIPLIER = 1;
 export const AGE_MULTIPLIER = 0.25;
--- a/frontend/src/lib/constants/neurons.constants.ts
+++ b/frontend/src/lib/constants/neurons.constants.ts
@@ -8,8 +8,10 @@ export const MAX_CONCURRENCY = 10;
 export const SPAWN_VARIANCE_PERCENTAGE = 0.95;
 
 // HW versions
+// Version published in december 2022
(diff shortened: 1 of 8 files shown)


lmuntaner:
@mstrasinskis please review

I put it in Draft because Ledger has not yet published the new version and I have not been able to test it with it yet.

I tested with the current version and as expcted users can't set auto-stake nor stake maturity.

frontend/src/lib/constants/neurons.constants.ts
```diff
@@ -8,7 +8,10 @@ export const MAX_CONCURRENCY = 10;
 export const SPAWN_VARIANCE_PERCENTAGE = 0.95;
 
 // HW versions
+// Version published in december 2022
 export const MIN_VERSION_STAKE_MATURITY_WORKAROUND = "2.0.7";
```
mstrasinskis, line 12: As discussed, using the version `2.0.8` would be safer.

mstrasinskis (changes requested):
One comment regarding the version.

frontend/src/lib/constants/neurons.constants.ts
lmuntaner: I used a new constant below `CANDID_PARSER_VERSION`.

My plan is to remove `MIN_VERSION_STAKE_MATURITY_WORKAROUND` when I remove the mergeMaturity service.

lmuntaner:
@mstrasinskis The new version (2.2.1) has been published, I tested this branch and I was able to stake maturity and set auto-stake.

I'm waiting on Bjorn to confirm forcing users to update to 2.2.1 and do not allow staking maturity with 2.0.7.

mstrasinskis (approved):
LGTM, thank you!

PR #1686, opened 2023-01-05
Refactor: ProposalCard agnostic of NNS

lmuntaner:
# Motivation

Refactor the ProposalCard into NnsProposalCard and a NNS agnostic one ProposalCard.

Preparing for adding Voting for SNS projects.

# Changes

* Split current ProposalCard into two: NnsProposalCard and ProposalCard.
* New ProposalCard expects primitive props and renders them in the UI.
* Refactored the rendering of ProposalCard to use KeyValuePair component.
* New NnsProposalCard expects a proposal, extracts the details and renders a ProposalCard.
* Refactor ProposalCountdown to Countdown and pass only the timestamp deadline in seconds to make it agnostic of NNS Proposal.

# Tests

* Tests for both splited components ProposalCard and NnsProposalCard.
* Fix other tests after change in props and renamings.
* Fix warning during test.

### DIFF
--- a/frontend/src/lib/components/proposals/ProposalCard.svelte
+++ b/frontend/src/lib/components/proposals/ProposalCard.svelte
@@ -1,84 +1,66 @@
 <script lang="ts">
-  import { Card, Value } from "@dfinity/gix-components";
-  import {
-    type ProposalInfo,
-    type NeuronId,
-    type ProposalId,
-    ProposalStatus,
-  } from "@dfinity/nns";
+  import { Card, KeyValuePair, Value } from "@dfinity/gix-components";
+  import { ProposalStatus } from "@dfinity/nns";
   import { i18n } from "$lib/stores/i18n";
-  import { mapProposalInfo } from "$lib/utils/proposals.utils";
-  import ProposalCountdown from "./ProposalCountdown.svelte";
   import { keyOfOptional } from "$lib/utils/utils";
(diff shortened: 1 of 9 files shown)


lmuntaner:

There are no UI changes.

frontend/src/lib/components/proposals/ProposalCard.svelte
```diff
@@ -91,32 +73,34 @@
...
-      margin: 0 var(--padding-2x) 0 0;
+    :global(.value) {
+      color: var(--tertiary);
     }
-
-    color: var(--value-color);
   }
 
-  .card-meta {
-    @include card.meta;
```
mstrasinskis, line 109: I think this was the only place where this class is used. Would be great to remove it from gix-components as well

mstrasinskis (approved):
LGTM, thanks!

frontend/src/lib/components/proposals/ProposalCard.svelte
lmuntaner: Yes, you're right. I'll do that.

lmuntaner: It is used, but inside a component in gix-components: https://github.com/dfinity/gix-components/blob/main/src/lib/components/Card.svelte#L137

PR #1688, opened 2023-01-06
chore: update II init type

frederikrothenberger:
# Motivation

Upgrades the II init type again.

# Changes

Copies the new type definitions from the II repo.

# Tests

There are no tests for the init type.


lmuntaner (approved):
Thanks for keeping it up to date!

PR #1690, opened 2023-01-07
GIX-1209: Setup SNS proposals list page

lmuntaner:
# Motivation

Set up the page for the SNS proposals list.

# Changes

* Rename current pages/Proposals to NnsProposals
* New route component Proposals that renders project dropdown and then chooses either NnsProposals or SnsProposals.
* Add new feature flag: `ENABLE_SNS_VOTING`

# Tests

* Fix tests after renaming and refactor.
* New test for new route Proposals.

PR #1691, opened 2023-01-10
fix: <Proposals> was created with unknown prop 'referrerPath'

peterpeterparker:
# Motivation

The `referrerPath` was not forwarded when the Sns proposals list was introduced in PR #1690.

# Changes

- fix "<Proposals> was created with unknown prop 'referrerPath'"

# Screenshot

Capture d’écran 2023-01-10 à 08 13 35


lmuntaner (approved):
Thanks

PR #1700, opened 2023-01-11
chore: Update threshold

bitdivine:
# Motivation
The threshold for ingress message size is lower than the threshold in the CI test.

# Changes
- Update (lower) the ingress message size to the one given in the error message when making a proposal.

# TODO
- Perhaps probe mainnet to detect if the threshold changes.
- Do an upgrade test; we could replicate the problem on a testnet when  installing via proposal instead of when installing directly.


mstrasinskis (approved):
LGTM, thank you!

PR #1701, opened 2023-01-12
chore: Update II init arguments for archive pull integration

frederikrothenberger:
This PR updates the II init argument to include the flag to switch the archive from push to pull.

# Motivation

The II init arguments need to be rendered correctly -> types need to be up to date.

# Changes

* Change the II init argument type.

# Tests

No tests.


lmuntaner (approved):
Thanks!

PR #1710, opened 2023-01-15
feat: load sns projects balances with queries but update on demand

peterpeterparker:
# Motivation

Load all Sns projects balances on "Account" views and display the results in the new "Token" picker navigation.

# WARNING ⚠️

Unlike any other calls of the dapp, it was requested to load these information with "query" calls only to avoid network performance issue.

 i.e. the balances that are going to be displayed won't be certified until the user select one token.

When a token is selected, then an "update" can will be performed and the related value will be updated accordingly.

# PRs

This PR -> #1677 -> `main`

# Changes

- implement effective feature by loading the Sns projects balances with queries when the "Accounts" view is loaded
- use Nns account store to display Nns balance - i.e. use query+update for Nns
- modify components - notably `ProjectBalance` - to not only display the balance of the selected project but for all projects listed in the token switcher. this apply to the desktop view as for the dropdown / modal picker
- new services functino `uncertifiedLoadSnsAccountsBalances`
- spread `certified` information of sns summaries to triggers loading of the balances only after the related queries have been fetched - i.e. do not trigger addition update calls
- add accounts store certified information for compatibility with derived stores
- improve type safety of various stores and derived stores used for this feature
- fix various minor lint warnings not related to this feature

# Screenshots

Capture d’écran 2023-01-16 à 10 27 32
Capture d’écran 2023-01-16 à 10 27 36
Capture d’écran 2023-01-16 à 10 27 38


frontend/src/lib/services/sns-accounts.services.ts
```diff
@@ -24,12 +26,19 @@ export const loadSnsAccounts = async ({
...
-    onLoad: ({ response: accounts, certified }) =>
+    onLoad: ({ response: accounts, certified }) => {
       snsAccountsStore.setAccounts({
         accounts,
         rootCanisterId,
         certified,
-      }),
+      });
+
+      snsAccountsBalanceStore.setBalance({
```
peterpeterparker, line 36: I might not need this, double checking...

frontend/src/lib/services/sns-accounts-balance.services.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @param {summaries: SnsSummary[], excludeRootCanisterIds?: RootCanisterIdText[]} params
+ * @param {SnsSummary[]} params.summaries The list of summaries - Sns projects - for which the balance of the accounts should be fetched.
+ * @param {RootCanisterIdText[] | undefined} params.excludeRootCanisterIds As the balance is also loaded by loadSnsAccounts() - to perform query and UPDATE call - this variable can be used to avoid to perform unnecessary query and per extension to override data in the balance store.
+ */
+export const uncertifiedLoadSnsAccountsBalances = async ({
+  summaries,
+  excludeRootCanisterIds = [],
+}: {
+  summaries: SnsSummary[];
```
lmuntaner, line 57: Wouldn't it be nicer as a signature to pass the list of rootCanisterIds?

frontend/src/lib/services/sns-accounts-balance.services.ts
```diff
@@ -0,0 +1,79 @@
...
+}): Promise<void> => {
+  const results: PromiseSettledResult<void>[] = await Promise.allSettled(
+    (
+      summaries.filter(
+        ({ metadataCertified, rootCanisterId }) =>
+          metadataCertified === false &&
+          !excludeRootCanisterIds.includes(rootCanisterId.toText())
+      ) ?? []
+    ).map(({ rootCanisterId }) =>
+      uncertifiedLoadSnsAccountsBalance({
```
lmuntaner, line 68: Why not using `loadSnsAccounts` and adding the "strategy" as an extra parameter there?

frontend/src/lib/services/sns-accounts-balance.services.ts
lmuntaner: This way you don't need the `excludeRootCanisterIds` param, and `uncertifiedLoadSnsAccountsBalances` in only about loading the balances.

frontend/src/lib/services/sns-accounts-balance.services.ts
```diff
@@ -0,0 +1,79 @@
...
+  excludeRootCanisterIds = [],
+}: {
+  summaries: SnsSummary[];
+  excludeRootCanisterIds?: RootCanisterIdText[];
+}): Promise<void> => {
+  const results: PromiseSettledResult<void>[] = await Promise.allSettled(
+    (
+      summaries.filter(
+        ({ metadataCertified, rootCanisterId }) =>
+          metadataCertified === false &&
```
lmuntaner, line 64: Why is this needed?

frontend/src/lib/routes/Accounts.svelte
```diff
@@ -7,6 +7,17 @@
...
   import SnsAccountsFooter from "$lib/components/accounts/SnsAccountsFooter.svelte";
+  import { snsSummariesStore } from "$lib/stores/sns.store";
+  import { uncertifiedLoadSnsAccountsBalances } from "$lib/services/sns-accounts-balance.services";
+
+  // Selected project ID on mount is excluded from load accounts balances. See documentation.
+  let projectIdSelected = $snsProjectIdSelectedStore;
+
+  $: (async () =>
+    await uncertifiedLoadSnsAccountsBalances({
+      summaries: $snsSummariesStore,
```
lmuntaner, line 18: Would it make sense to have a derived store that returns the rootCanisterIds directly and even checks the metadataCertified?

frontend/src/lib/stores/accounts.store.ts
```diff
@@ -1,26 +1,34 @@
...
 
-export interface AccountsStore {
+export interface AccountsWritableStore {
   main?: Account;
   subAccounts?: Account[];
   hardwareWallets?: Account[];
+  certified?: boolean;
+}
+
+export interface AccountsStore extends Readable<AccountsWritableStore> {
```
lmuntaner, line 11: I'm sorry, but the naming confused me.

One type `AccountsStore` is the actual object store, the one we subscribe to.

The other type is the data inside, right?

Therefore, how about `AccountsStoreData`. To me `XXXWritable` just means it's a writable store, which `AccountsStore` is?

frontend/src/lib/stores/sns-accounts-balance.store.ts
```diff
@@ -0,0 +1,65 @@
...
+  setBalance: (params: {
+    rootCanisterId: Principal;
+    balance: TokenAmount | undefined | null;
+    certified: boolean;
+  }) => void;
+  reset: () => void;
+}
+
+/**
+ * A store that contains the balance of the Sns accounts of the projects.
```
lmuntaner, line 25: Why not reusing the `SnsAccountsStore` and creating a derived store from it if you only want the balances?

frontend/src/lib/services/sns-accounts.services.ts
peterpeterparker: Actually no we need it.

frontend/src/lib/services/sns-accounts-balance.services.ts
peterpeterparker: We need both `rootCanisterIds` and `certified` from `summaries` here. 

- we need summaries because we want to load at boot time exactly one once
- we need certified because we want to chain only "query" and not "update"

We also need `excludeRootCanisterIds` because we want to avoid unecessary "query" because its 

That's what I tried to describe with the comment in the code.

So I can map the summaries to root caniter ids and certified only but not sure it brings lokt of clarity in the code.

frontend/src/lib/services/sns-accounts-balance.services.ts
peterpeterparker: - because it is insecure and I want to express the fact that it isn't
- the features does more that `loadSnsAccounts` - only "query" chain is processed, "update" chain is discarded and the current account shall not be loaded here

frontend/src/lib/services/sns-accounts-balance.services.ts
peterpeterparker: 1. To preserve the chain of how we fetch the data - i.e. we load all Sns projects with queries so here we can also do query and we do not mix using "update" calls to fetch Sns information with adding "query" calls at the top
2. To avoid to perform the queries twice - performance

Added a comment

frontend/src/lib/routes/Accounts.svelte
peterpeterparker: Duplicate of https://github.com/dfinity/nns-dapp/pull/1710#discussion_r1071157010 I guess

frontend/src/lib/stores/accounts.store.ts
peterpeterparker: I just reused existing `AccountStore`name, open to any other suggestion of naming.

Important for me is that the type declaration was improved and also this fixes Webstorm displaying me errors everywhere because it cannot interfer types.

Let me know your suggestion:

`AccountStore` -> ???
`AccountsWritableStore` -> ???

frontend/src/lib/stores/sns-accounts-balance.store.ts
peterpeterparker: The `SnsAccountsStore` should remains the same because when a user select and project, the query+update calls should be performed and the "total balance" should then become certified that way.

As described, the user sees total that are fetched with "query" but when he/she effectively select something the the data selected should become certified.

By not changing the `SnsAccountsStore` I comply with the requirements and I am also sure I don't corrupt the core of the application.

frontend/src/lib/stores/accounts.store.ts
lmuntaner: `AccountsStore` -> `AccountsStore`
`AccountsWritableStore` -> `AccountsStoreData`

For me it was `XXWritable` that was not very clear.

peterpeterparker: Sure, done

frontend/src/lib/services/sns-accounts-balance.services.ts
peterpeterparker: Discussed offline and store updated in https://github.com/dfinity/nns-dapp/pull/1719

lmuntaner (approved):
Thanks!

PR #1711, opened 2023-01-16
Fix: votingPower calculation for SNS neurons

lmuntaner:
# Motivation

The user would see an invalid simulated voting power calculation if it had some maturity or staked maturity.

# Changes

* Take into account the staked_maturity_e8s_equivalent not maturity_e8s_equivalent for SNS neuron voting power.

# Tests

* Add a test for the new case.


mstrasinskis (approved):
Thank you!

PR #1715, opened 2023-01-16
Dissolve delay input update

mstrasinskis:
# Motivation

- `SetDissolveDelay` component for nns and sns
- add input field for precise editing
- use days as dissolve delay unit
- min button

Design - https://www.figma.com/file/ei7wVYd4l8SuDQlQqVC0NV/New-NNS?node-id=0%3A1&t=MvzTS1YiVycEfMKN-0

- [x] gix-component related update https://github.com/dfinity/gix-components/pull/150

# Changes

- refactor to use new input slots
- day input component (with min/max buttons)
- refactor to use SetDissolveDelay by nns and sns modals
- daysToDuration

# Tests

- min/max buttons
- day input update
- daysToDuration

# Screenshots

## NNS
https://user-images.githubusercontent.com/98811342/212960544-59411c5b-ceaf-4728-a737-cda7cb117bc5.mov

## SNS
https://user-images.githubusercontent.com/98811342/212960576-1f0919ea-86a4-4993-a991-e703a5685fdc.mov


frontend/src/lib/components/common/MinButton.svelte
```diff
@@ -0,0 +1,25 @@
...
+  button {
+    display: flex;
+    align-items: center;
+    gap: var(--padding-0_5x);
+  }
+  .icon {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    transform: scaleX(-1);
```
peterpeterparker, line 23: you need 1px less?

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
```diff
@@ -1,106 +1,154 @@
...
-  import { valueSpan } from "$lib/utils/utils";
+  import { isDefined, valueSpan } from "$lib/utils/utils";
   import NeuronStateRemainingTime from "$lib/components/neurons/NeuronStateRemainingTime.svelte";
+  import DayInput from "$lib/components/ui/DayInput.svelte";
+  import { daysToDuration } from "$lib/utils/date.utils.js";
+  import type { NeuronState, TokenAmount } from "@dfinity/nns";
+  import Hash from "$lib/components/ui/Hash.svelte";
 
-  export let neuron: NeuronInfo;
+  export let neuronIdText: string;
```
peterpeterparker, line 16: if feel like it's a lot of properties, ask my self if it would not be neat to extract a type for the property and then descrut it internaly. just thinking out at loud

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
```diff
@@ -1,106 +1,154 @@
...
-  };
-  let neuronICP: bigint;
-  $: neuronICP = neuronStake(neuron);
+    delayInDays < secondsToDays(minProjectDelayInSeconds) ||
+    delayInDays <= minDelayInDays ||
+    delayInDays > maxDelayInDays;
+
+  const checkMinMax = () => {
+    if (delayInDays < minDelayInDays) {
+      delayInDays = minDelayInDays;
```
peterpeterparker, line 53: does the screen becomes glitchy when this happens?

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
```diff
@@ -1,106 +1,154 @@
...
+  const setMin = () => {
+    delayInDays = Math.max(
+      minDelayInDays + 1,
+      secondsToDays(minProjectDelayInSeconds)
+    );
+    checkMinMax();
+  };
+  const setMax = () => {
+    delayInDays = maxDelayInDays;
+    checkMinMax();
```
peterpeterparker, line 71: would `$: delayInDays, (() => checkMinMax())()` would also work? to avoid imperative calls and double code

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
peterpeterparker: also could / shouldn't some of these props be `<slot />`? e.g. `neuronIdText`

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
```diff
@@ -0,0 +1,48 @@
...
+    neuronStake as getNeuronStake,
+    neuronVotingPower,
+  } from "$lib/utils/neuron.utils";
+
+  import SetDissolveDelay from "$lib/components/neurons/SetDissolveDelay.svelte";
+  import { ICPToken, TokenAmount } from "@dfinity/nns";
+
+  export let delayInSeconds: number;
+  export let neuron: NeuronInfo;
+  export let cancelButtonText: string;
```
peterpeterparker, line 18: also here these two texts, `slot`?

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
```diff
@@ -1,106 +1,154 @@
...
+  $: delayInSeconds, (() => (delayInDays = secondsToDays(delayInSeconds)))();
+
+  let minDelayInDays = 0;
+  $: minDelayInDays = secondsToDays(minDelayInSeconds);
+
+  let maxDelayInDays = 0;
+  $: maxDelayInDays = secondsToDays(maxDelayInSeconds);
+
+  let votingPower: bigint;
+  $: votingPower = calculateVotingPower(delayInSeconds);
```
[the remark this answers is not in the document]
peterpeterparker: now I get it, by having a look to the consumer component. not sure it is worth a comment or not, so just in case not deleting my comment even if I get it if you think that's interesting.

frontend/src/lib/utils/date.utils.ts
```diff
@@ -49,6 +49,49 @@ export const secondsToDuration = (seconds: bigint): string => {
...
+  const dayCount = Math.ceil(days - yearCount * DAYS_IN_YEAR);
+
+  const periods = [
+    createLabel("year", BigInt(yearCount)),
+    createLabel("day", BigInt(dayCount)),
+  ];
+
+  return periods
+    .filter(({ amount }) => amount > 0)
+    .slice(0, 2)
```
peterpeterparker, line 83: just in case there is a way to make this more obvious, just wanted to point out that spontaneously don't know what's sliced here 🤷‍♂️

peterpeterparker (commented):
Few minor comments

frontend/src/lib/utils/date.utils.ts
mstrasinskis: Good catch, this is not needed anymore.

frontend/src/lib/components/common/MinButton.svelte
mstrasinskis: Discussed in slack.

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
mstrasinskis: Didn't notice any issue.

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
mstrasinskis: Not really, the idea is to avoid `checkMinMax` call during manual value entering (not w/ the range input). So the user is free to enter any value manually.

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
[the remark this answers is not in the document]
mstrasinskis: Small comment was added.

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
mstrasinskis: Generally speaking I don't see a big advantage to have slots for text only. And it makes unit testing more complicated.

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
mstrasinskis: Good point! Slot was created for the `neuronIdText` to fully preserve the previous rendering.

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
peterpeterparker: Ah ok

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
peterpeterparker: using `slot` is closer to HTML semantic and is also platform standard. we also mostly use slots rather than passing text  as properties generally speaking when we don't have to (I think). That said should not be a blocker particularly as I should have answered earlier, sorry for the delay.

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
peterpeterparker: Coolio

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
```diff
@@ -1,106 +1,155 @@
...
-    delayInSeconds < SECONDS_IN_HALF_YEAR ||
-    delayInSeconds <= minDelayInSeconds;
-  const dispatcher = createEventDispatcher();
-  const cancel = (): void => {
-    dispatcher("nnsCancel");
+    delayInDays < secondsToDays(minProjectDelayInSeconds) ||
+    delayInDays <= minDelayInDays ||
+    delayInDays > maxDelayInDays;
+
+  const checkMinMax = () => {
```
peterpeterparker, line 50: nth: actually this "update the delay" rather than checking it, so maybe `updateDelays` is a more accurate name?

frontend/src/lib/components/sns-neurons/SetSnsDissolveDelay.svelte
```diff
@@ -1,202 +1,104 @@
...
-          fromDefinedNullable(
-            snsParameters?.neuron_minimum_dissolve_delay_to_vote_seconds
+  let neuronState: NeuronState;
+  $: neuronState = getSnsNeuronState(neuron);
+
+  const calculateVotingPower = (delayInSeconds: number) =>
+    isNullish(snsParameters)
+      ? 0n
+      : BigInt(
+          Math.round(
```
peterpeterparker, line 49: nth: since I notice there is some rounding, would it be worth extracting `calculateVotingPower` in a separate `utils` to add some dedicated tests?

peterpeterparker (commented):
Just noticing there aren't much tests that tests these min / max assertion, should we add some more tests here or e2e?

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
mstrasinskis: Good point, renamed!

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
mstrasinskis: Thanks for the explanation, I'm convinced. Update.

frontend/src/lib/components/ui/Input.svelte
```diff
@@ -34,11 +34,11 @@
   {autocomplete}
   on:blur
   on:focus
-  on:input
+  on:nnsInput
   on:keydown
   {showInfo}
 >
+  <svelte:fragment slot="start"><slot name="start" /></svelte:fragment>
```
peterpeterparker, line 41: I think you can just do `<slot name="start" slot="start" />`

peterpeterparker (approved):
LGTM, thanks

Note: note sure this question is still relevant or not https://github.com/dfinity/nns-dapp/pull/1715#pullrequestreview-1275674465 maybe have a look before merge

frontend/src/lib/components/ui/Input.svelte
peterpeterparker: @lmuntaner teached me that Svelte tricks 👍

mstrasinskis:
> LGTM, thanks
> 
> Note: note sure this question is still relevant or not [#1715 (review)](https://github.com/dfinity/nns-dapp/pull/1715#pullrequestreview-1275674465) maybe have a look before merge

@peterpeterparker thanks for emphasising, tests for min/max buttons were added -- https://github.com/dfinity/nns-dapp/pull/1715/files#diff-4ab7ee43018b47284b452e07516c34c82a9f1db2e5271c58a4da8e0f17f53436R188

peterpeterparker:
Thanks

PR #1718, opened 2023-01-16
feat: Include an SNS in the CI testnet

bitdivine:
# Motivation
We need e2e tests for the SNS.  The first thing we need for that is a testnet setup that includes an SNS.

# Changes
- Run the existing e2e tests against a local dfx server that has SNS set up.
  - Note: This does NOT deploy multiple SNSs, open SNSs and so on.  We will need that for some tests.  This PR just deploys the setup needed to support SNSs, and the largest part of running SNSs is the setup.

# Tests
- There is no new functionality, so existing tests should suffice


.github/workflows/build.yml
```diff
@@ -64,16 +65,28 @@ jobs:
           cargo --version
           ic-cdk-optimizer --version
```
bitdivine, line 66: We use the snsdemo scripts to deploy an SNS.

.github/workflows/build.yml
```diff
@@ -64,16 +65,28 @@ jobs:
...
+        with:
+          repository: 'dfinity/snsdemo'
+          path: 'snsdemo'
+          # Version from Mon Jan 16 10:10:47
+          ref: 'e1d7dd4c10fb8a731a534ae2ef4b08474fea5f6b'
+
+      - name: Add SNS scripts to the path
+        run: |
+          echo "$PWD/snsdemo/bin" >> $GITHUB_PATH
+
```
bitdivine, line 79: This installs things such as dfx, jq and idl2json that we probably already have, but just to be sure that we don't have strange breakages due to something as silly as this, it seems like a good idea to explicitly ask for the snsdemo tools.

.github/workflows/build.yml
```diff
@@ -64,16 +65,28 @@ jobs:
...
+      - name: Install SNS script dependencies
         run: |
-          dfx start --background --clean
+          dfx-sns-demo-install
```
bitdivine, line 83: This is where the action happens.  SNS joins the party.

.github/workflows/build.yml
```diff
@@ -100,10 +113,6 @@ jobs:
         run: |
           DFX_NETWORK=local ENV_OUTPUT_FILE=./e2e-tests/.env ./config.sh
```
bitdivine, line 116: I think we can throw this away.  Green CI suggests so too.

.github/workflows/build.yml
```diff
@@ -64,16 +65,28 @@ jobs:
...
+          path: 'snsdemo'
+          # Version from Mon Jan 16 10:10:47
+          ref: 'e1d7dd4c10fb8a731a534ae2ef4b08474fea5f6b'
+
+      - name: Add SNS scripts to the path
+        run: |
+          echo "$PWD/snsdemo/bin" >> $GITHUB_PATH
+
+      - name: Install SNS script dependencies
         run: |
```
bitdivine, line 81: I have moved start-replica to the script that deploys the testnet.

scripts/dfx-nns-deploy-custom
```diff
@@ -24,8 +25,24 @@ rm -f nns-dapp.wasm nns-dapp_local.wasm
 : "Get our chosen version of II and inject it into the wasm cache."
 "$SOURCE_DIR"/dfx-software-ii-cache
```
bitdivine, line 27: Maybe we want just the NNS sometimes.  If the SNS is not needed, deploying without it is significantly faster.  So I have kept the old behaviour along with the new, full fat network with SNS.

.github/workflows/build.yml
lmuntaner: Do you mean the creation of the config file?

scripts/dfx-nns-deploy-custom
```diff
@@ -24,8 +25,23 @@ rm -f nns-dapp.wasm nns-dapp_local.wasm
...
-dfx nns install
+case "$DFX_NETWORK_FEATURES" in
+nns)
+  : "Deploy nns-dapp with other NNS canisters"
+  dfx start --background --clean
+  dfx nns install
+  ;;
+sns)
+  : "Deploy nns-dapp with NNS and SNS canisters"
+  dfx-sns-demo
```
lmuntaner, line 36: Can we add a comment that this comes from the snsdemo repository?

lmuntaner (approved):
Looks good to me!

Just one comment, but already approving.

Sorry that I removed one of your comments. I wanted to remove mine 😅 

But I agree, let's remove the populate.

We can even clean it from `./deploy.sh` later as well.

scripts/dfx-nns-deploy-custom
bitdivine: Done

PR #1719, opened 2023-01-16
feat: project balances use sns accounts store

peterpeterparker:
# Motivation

As discussed this PR removes using a custom store - `snsAccountsBalanceStore` - to display the balance but leverage the existing `snsAccountsStore`.

# Small Downside of the PR

If there is an error while fetching Sns accounts, the total balance in the navigation will remain displayed with a skeleton instead as disappearing as in #1710. It's an acceptable drawback in my opinion.

# PRs

This PR -> #1710 -> #1677 -> main


lmuntaner (approved):
Thanks!

PR #1720, opened 2023-01-16
feat: remove "copy" amount from token switcher and move it to wallet details

peterpeterparker:
# Motivation

We need the "copy" amount button because that's the way user can copy the effective amount with decimals. However, if there are many tokens, the "universe / token switcher" becomes crowded of "copy" buttons.

That's why this PR removes it from the switcher and add move the feature to wallet details.

# PRs

This PR -> #1677 -> main

# Screenshots

Capture d’écran 2023-01-16 à 19 46 03

Capture d’écran 2023-01-16 à 19 46 08


lmuntaner (approved):
Thanks!

PR #1723, opened 2023-01-17
GIX-1212: Add Proposal Status filter

lmuntaner:
# Motivation

User can filter proposal by decision status.

IMPORTANT: This PR includes setup of filters, button, modal and store. Not yet updating the proposals list.

# Changes

Main changes:
* SnsProposalsFilters.svelte will include all filters buttons and render each modal.
* SnsFilterStatusModal.svelte shows filters and performs the actions in the store.
* sns-filters.store manages filters for SNS projects.
* loadSnsFilters in sns-filters.services.

The idea to store the Filter type in the store is to use the same concept for all SNS filters. Recall that SNS topics is not an enum, instead is the list of nervous functions. Which have a name, id and description.

I believe this is a good idea also for NNS. (for later) This way the UI doesn't care or know where the filters come from. Whether an enum or from a backend. The UI components are more reusable. We just need to add the logic of filling up the store with the data.

Minor changes:
* New UI component FiltersWrapper.svelte to be reused in SNS and NNS filters.
* ProposalsFilters.svelte to NnsProposalsFilters.svelte

# Tests

### DIFF
--- a/frontend/src/lib/components/proposals/FiltersWrapper.svelte
+++ b/frontend/src/lib/components/proposals/FiltersWrapper.svelte
@@ -1,0 +1,28 @@
+<div class="filters" data-tid="proposals-filters">
+  <slot />
+</div>
+
+<style lang="scss">
+  .filters {
+    display: flex;
+    flex-wrap: wrap;
+    padding: 0 0 var(--padding-3x);
+
+    --checkbox-flex-direction: row-reverse;
+    --checkbox-font-size: var(--font-size-small);
+
+    :global(button) {
+      margin: var(--padding) var(--padding) 0 0;
+    }
+
+    > :global(div.checkbox) {
+      width: fit-content;
+      padding: var(--padding) calc(0.75 * var(--padding));
+      margin: var(--padding) 0 0;
+    }
+  }
+
+  :global(section > div.checkbox) {
+    margin: 0 0 var(--padding);
+  }
+</style>
--- a/frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
+++ b/frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
@@ -1,0 +1,54 @@
+<script lang="ts">
+  import FilterModal from "$lib/modals/common/FilterModal.svelte";
+  import { i18n } from "$lib/stores/i18n";
+  import { snsFiltesStore } from "$lib/stores/sns-filters.store";
+  import type { Filter } from "$lib/types/filters";
+  import type { Principal } from "@dfinity/principal";
+  import type { SnsProposalDecisionStatus } from "@dfinity/sns";
+  import { createEventDispatcher } from "svelte";
+
+  export let rootCanisterId: Principal;
+  export let filters: Filter<SnsProposalDecisionStatus>[] = [];
+
+  // This is a temporary value to be used inside the modal. It's initialized based on the prop filters;
+  let selectedFilters: SnsProposalDecisionStatus[] =
+    filters.filter(({ checked }) => checked).map(({ value }) => value) ?? [];
+  // This is a temporary value to be used inside the modal
+  let filtersValues: Filter<SnsProposalDecisionStatus>[];
+  $: filtersValues = filters.map((filter) => ({
+    ...filter,
+    checked: selectedFilters.includes(filter.value),
+  }));
+
+  const dispatch = createEventDispatcher();
+  let filter: () => void;
+  $: filter = () => {
+    snsFiltesStore.checkDecisionStatus({
+      checkedDecisionStatus: selectedFilters,
+      rootCanisterId,
+    });
+    dispatch("nnsClose");
+  };
+
+  const onChange = ({
+    detail: { filter },
+  }: CustomEvent<{
+    filter: Filter<SnsProposalDecisionStatus> | undefined;
+  }>) => {
+    if (filter === undefined) return;
+    selectedFilters = [
+      ...selectedFilters.filter((status) => status !== filter?.value),
+      // Toggle the checked value
+      ...(filter.checked ? [] : [filter.value]),
+    ];
+  };
+</script>
+
+<FilterModal
+  on:nnsClose
+  on:nnsConfirm={filter}
+  on:nnsChange={onChange}
+  filters={filtersValues}
+>
+  <span slot="title">{$i18n.voting.status}</span>
+</FilterModal>
--- a/frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
+++ b/frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
@@ -23,7 +23,7 @@
   const dispatch = createEventDispatcher();
   let filter: () => void;
   $: filter = () => {
-    snsFiltesStore.checkDecisionStatus({
+    snsFiltesStore.setCheckDecisionStatus({
       checkedDecisionStatus: selectedFilters,
       rootCanisterId,
     });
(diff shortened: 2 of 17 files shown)


lmuntaner:
New UX


https://user-images.githubusercontent.com/4550653/212920062-e322a7d8-dc0f-467f-aedd-ff7a0dcd94b1.mov

lmuntaner:

frontend/src/lib/components/proposals/FiltersWrapper.svelte
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    > :global(div.checkbox) {
+      width: fit-content;
+      padding: var(--padding) calc(0.75 * var(--padding));
+      margin: var(--padding) 0 0;
+    }
+  }
+
+  :global(section > div.checkbox) {
```
peterpeterparker, line 25: not related to your PR: a good reminder we need to fix this selector

frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
```diff
@@ -0,0 +1,54 @@
...
+    filters.filter(({ checked }) => checked).map(({ value }) => value) ?? [];
+  // This is a temporary value to be used inside the modal
+  let filtersValues: Filter<SnsProposalDecisionStatus>[];
+  $: filtersValues = filters.map((filter) => ({
+    ...filter,
+    checked: selectedFilters.includes(filter.value),
+  }));
+
+  const dispatch = createEventDispatcher();
+  let filter: () => void;
```
peterpeterparker, line 24: not sure why this is not a `const`?

frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
```diff
@@ -0,0 +1,54 @@
...
+    });
+    dispatch("nnsClose");
+  };
+
+  const onChange = ({
+    detail: { filter },
+  }: CustomEvent<{
+    filter: Filter<SnsProposalDecisionStatus> | undefined;
+  }>) => {
+    if (filter === undefined) return;
```
peterpeterparker, line 38: question: we do not have an eslint rule that enforce `{}` around single line `if`? I thought we had. Just asking in case the eslint does not pick the svelte code.

```
if (filter === undefined) {
  return;
}
```

frontend/src/lib/services/sns-filters.services.ts
```diff
@@ -0,0 +1,36 @@
+import { i18n } from "$lib/stores/i18n";
+import { snsFiltesStore } from "$lib/stores/sns-filters.store";
+import { enumValues } from "$lib/utils/enum.utils";
+import type { Principal } from "@dfinity/principal";
+import { SnsProposalDecisionStatus } from "@dfinity/sns";
+import { get } from "svelte/store";
+
+export const loadSnsFilters = async (rootCanisterId: Principal) => {
```
peterpeterparker, line 8: - I'm confuse, does this function load anything?
- it shouldn't be `async` I think
- shouldn't this be a utils rather than a service?

frontend/src/lib/services/sns-filters.services.ts
```diff
@@ -0,0 +1,36 @@
...
+  const i18nKeys = get(i18n);
+  const projectData = get(snsFiltesStore)[rootCanisterId.toText()] ?? {
```
peterpeterparker, line 10: nth: not sure the name `projectData` fits that well here

frontend/src/lib/services/sns-filters.services.ts
```diff
@@ -0,0 +1,36 @@
...
+    topics: [],
+    rewardStatus: [],
+    decisionStatus: [],
+  };
+  const mapDecisionStatus = (value: SnsProposalDecisionStatus) => {
+    return {
+      id: String(value),
+      value,
+      name: i18nKeys.sns_status[value] ?? "Unspecified",
```
peterpeterparker, line 19: "Unspecified" should also uses an i18n translation

frontend/src/lib/stores/sns-filters.store.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+interface SnsFiltersStoreData {
+  [rootCanisterId: string]: ProjectFiltersStoreData;
+}
+
+/**
+ * A store that contains the filters of the SNS proposals for each project.
+ */
+export const initSnsFiltersStore = () => {
```
peterpeterparker, line 23: If you could add a type declaration to the store that would be neat, my Webstorm would wecome it

frontend/src/lib/stores/sns-filters.store.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    setDecisionStatus({
+      rootCanisterId,
+      decisionStatus,
+    }: {
+      rootCanisterId: Principal;
+      decisionStatus: Filter<SnsProposalDecisionStatus>[];
+    }) {
+      update((currentState: SnsFiltersStoreData) => {
+        const projectFilters = currentState[rootCanisterId.toText()] || {
```
peterpeterparker, line 37: don't you want to persist the filters in the local storage as well?

frontend/src/lib/stores/sns-filters.store.ts
```diff
@@ -0,0 +1,86 @@
...
+          ...currentState,
+          [rootCanisterId.toText()]: {
+            ...projectFilters,
+            decisionStatus,
+          },
+        };
+      });
+    },
+
+    checkDecisionStatus({
```
peterpeterparker, line 53: just in case it is useful, at first I assumed this function would return a boolean - i.e. I interpreted its name as "check if the decision status".

frontend/src/lib/components/proposals/FiltersWrapper.svelte
lmuntaner: agree

frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
lmuntaner: It uses the `selectedFilters` variable that changes.

frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
lmuntaner: Sounds good to me

frontend/src/lib/services/sns-filters.services.ts
lmuntaner: Not yet, but it will load the Topics from the nervous functions.

That's why I already put it here.

Moreover, it loads data inside a store. That's why I think that a "service" makes more sense. I'd prefer "utils" to be as pure as possible.

frontend/src/lib/services/sns-filters.services.ts
lmuntaner: good point

frontend/src/lib/stores/sns-filters.store.ts
lmuntaner: Yes, and also add defaults. But not in this PR. I'll add a comment.

frontend/src/lib/stores/sns-filters.store.ts
lmuntaner: Makes sense, I'll change to `setCheckedDecisionStatus`

lmuntaner:

frontend/src/lib/modals/sns/proposals/SnsFilterStatusModal.svelte
peterpeterparker: Gotcha, thanks for the feedback

frontend/src/lib/services/sns-filters.services.ts
peterpeterparker: Ahhhh as there was no `// TODO` I did not understood there was still some work plan in this funtion therefore me being confuse and asking these questions.

peterpeterparker (approved):
LGTM, thanks

PR #1725, opened 2023-01-18
feat: display TVL (total value locked) in USD 

peterpeterparker:
# Motivation

Display the total value locked (tvl) in US dollar.

# ⚠️ Warning

This feature does not use certified data. It fetches exchange rate from Binance API (web2) and the TVL from the `.raw` domain of the governance canister.

# Future

Our plan is to use the new exchange canister (XRC) through a proxy - that caches the ICP<->USD rate - and get the TVL information with a certified or query+update calls.

# Rules

- if any of these API crashes an error, it should be silently* ignored and nothing - no TVL - should be displayed to the user
- if zero values are provided, likewise, nothing shold be displayed

* This feature is not decisive for user flow but stricly informative therefore we can/should silent the potential issues (not speaking of the fact that the endpoints are not certified)

# Notes

This feature performs the call for following reasons:

1. UI performance
2. Security

## UI performance

I personally think that it is not the job of the frontend UI to keep track and run cronjobs necesserary to fetch realtime data. I think that deferring these to workers reduce the load of the UI on the window side and therefore leads to better performance.

## Security

By executing the requests on the workers side we ensure that these are sandboxed outside the window. In addition, by doing so, we do not have to open our Content-Security-Policy to introduce exception for some web2 or raw domains.

# Changes

- new `<TotalValueLocked />` component and usage
- implementation of `rest` calls and interpretation of the results (in case of the governance)
- adds worker to fetch and re-evaluate the value every minute as cronjob
- remove ic link to spare space (kind of a duplicate of nns link to ic.org)
- few stylish adjustements on the login screen (like hover color)

# Screenshots

Capture d’écran 2023-01-18 à 14 00 48
Capture d’écran 2023-01-18 à 14 00 52
Capture d’écran 2023-01-18 à 14 01 09
Capture d’écran 2023-01-18 à 14 01 24
Capture d’écran 2023-01-18 à 14 01 29


frontend/src/lib/workers/metrics.worker.ts
```diff
@@ -0,0 +1,64 @@
...
+    case "nnsStartMetricsTimer":
+      await startMetricsTimer();
+      return;
+  }
+};
+
+let timer: NodeJS.Timeout | undefined = undefined;
+
+const startMetricsTimer = async () => {
+  const sync = async () => await syncMetrics();
```
mstrasinskis, line 29: Not sure but probably makes senes to add to the `sync` method:
```
  if (!timer) {
    return;
  }
```

And I would also put timer = `setTimeout`(sync, SYNC_METRICS_TIMER_INTERVAL); also inside the `sync` function to not request if the previous is not done. But this is just a nitpick ;)

frontend/src/tests/lib/components/metrics/TotalValueLocked.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  it("should not render TVL on load", () => {
+    const { getByTestId } = render(TotalValueLocked);
+    expect(() => getByTestId("tvl-metric")).toThrow();
+  });
+
+  it("should not render TVL if response has no metrics", async () => {
+    const { getByTestId } = render(TotalValueLocked);
+
+    metricsCallback?.({});
+
```
mstrasinskis, line 68: Probably it makes sense to add a `await tick()`? Just a guess that it will always throw right after metrics update.

frontend/src/tests/lib/rest/governance-metrics.rest.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    jest.spyOn(console, "error").mockImplementation(() => undefined)
+  );
+
+  afterAll(() => {
+    jest.clearAllMocks();
+    jest.resetAllMocks();
+  });
+
+  it("should return metrics as text", async () => {
+    const text = `# HELP governance_stable_memory_size_bytes Size of the stable memory allocated by this canister measured in bytes.
```
mstrasinskis, line 14: Maybe worth to have a constant in a mock data file.

mstrasinskis (commented):
Generally looks very good, just some questions.

frontend/src/lib/components/login/LoginFooter.svelte
```diff
@@ -1,14 +1,30 @@
...
+  import MenuMetrics from "$lib/components/common/MenuMetrics.svelte";
 
   export let presentation: "footer" | "menu" = "footer";
+
+  let innerWidth = 0;
+  let displayTvl = false;
+
+  // See comment in <LoginHeader />
+  $: displayTvl =
+    innerWidth > 0 && innerWidth <= BREAKPOINT_LARGE && presentation === "menu";
```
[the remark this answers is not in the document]
peterpeterparker: From what I noticed while debugging, `innerWidth`, at least in development, was emitted with `0` on `load`. Therefore to avoid loading the component in such state, I added this check.

frontend/src/lib/components/metrics/TotalValueLocked.svelte
```diff
@@ -0,0 +1,110 @@
...
+  export let layout: "inline" | "stacked" = "inline";
+
+  let worker:
+    | {
+        startMetricsTimer: (params: { callback: MetricsCallback }) => void;
+        stopMetricsTimer: () => void;
+      }
+    | undefined;
+
+  onMount(async () => (worker = await initMetricsWorker()));
```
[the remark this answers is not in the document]
peterpeterparker: Copy/paste of my other wip PR #1699 that uses same approach but, you are right, in the other PR it makes sense but here not. Done, move to `onMount`.

frontend/src/lib/components/metrics/TotalValueLocked.svelte
```diff
@@ -0,0 +1,110 @@
...
+      }))();
+
+  let totalNeurons: number | undefined;
+  $: totalNeurons =
+    (metricsSync?.dissolvingNeurons?.totalDissolvingNeurons ?? 0) +
+    (metricsSync?.dissolvingNeurons?.totalNotDissolvingNeurons ?? 0);
+
+  let total: number | undefined;
+  $: total =
+    ((totalNeurons ?? 0) / 1_000_00_000) *
```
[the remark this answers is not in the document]
peterpeterparker: Gosh that's the one I was looking for. Had a typo in my formatting that's why I did not find it. Thanks!

frontend/src/lib/constants/metrics.constants.ts
```diff
@@ -0,0 +1,2 @@
+// Workers
+export const SYNC_METRICS_TIMER_INTERVAL = 60000; // 1 minute
```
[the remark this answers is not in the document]
peterpeterparker: Sure

frontend/src/lib/workers/metrics.worker.ts
peterpeterparker: Good point. I added a test on the timer (if already defined) and added also a `syncInProgress` flag to perform the job only once at a time.

frontend/src/tests/lib/components/metrics/TotalValueLocked.spec.ts
peterpeterparker: Sure. Done.

frontend/src/tests/lib/rest/governance-metrics.rest.spec.ts
peterpeterparker: Why not, done

mstrasinskis (approved):
LGTM, thank you!

PR #1726, opened 2023-01-18
feat: A minimal launchpad e2e test

bitdivine:
# Motivation
An e2e test is needed for load generation.  It doesn't need to actually test anything, it just needs to load the launchpad to stress the network.

Proper e2e tests are in the pipeline for the launchpad, however stress testing doesn't cared about delicacies such as using tids to click buttons.  So the stress-testing e2e test can be used now without waiting for tids et al.

# Changes
- Add an e2e test that loads the launchpad, first logged out then logged in.

# Tests
- I have run this e2e test locally and it will be run in CI.

### DIFF
--- a/e2e-tests/specs/launchpad.e2e.ts
+++ b/e2e-tests/specs/launchpad.e2e.ts
@@ -1,0 +1,41 @@
+import { register } from "../common/register";
+import { waitForImages } from "../common/waitForImages";
+import { waitForLoad } from "../common/waitForLoad";
+import { skipUnlessBrowserIs } from "../common/test";
+
+describe("View launchpad while logged out", () => {
+  before(function () {
+    skipUnlessBrowserIs.bind(this)(["chrome", "firefox"]);
+  });
+
+  it("Views the launchpad whil elogged out", async () => {
+    await browser.url("/launchpad/");
+    await waitForLoad(browser);
+    await waitForImages(browser);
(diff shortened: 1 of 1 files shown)


e2e-tests/specs/launchpad.e2e.ts
```diff
@@ -0,0 +1,41 @@
...
+describe("View launchpad while logged out", () => {
+  before(function () {
+    skipUnlessBrowserIs.bind(this)(["chrome", "firefox"]);
+  });
+
+  it("Views the launchpad whil elogged out", async () => {
+    await browser.url("/launchpad/");
+    await waitForLoad(browser);
+    await waitForImages(browser);
+    await browser.pause(2000); // Wait a second for the back-end data to load
```
peterpeterparker, line 15: you don't want to wait more accuratly for a test-id to be rendered (with a timeout as fallback)?

bitdivine: Absolutely, for the proper e2e test that should also do such things as make sure that expected SNSs are lised.

I can put in a comment saying "placeholder for tests that the content is rendered."

peterpeterparker: Never spare a TODO 😉. Thanks for the feedback

peterpeterparker (approved):
LGTM, thx

bitdivine:
Formatting woes...

PR #1727, opened 2023-01-18
fix small09 config

bitdivine:
# Motivation
Deployments of the nns-dapp to small09 do not show SNSs because one of the canister ids is not definedin dfx.json.

And small14, actually.  How many others are there like this?  Ok, let's add all the networks and make sure they are right.

# Changes
- Set the nns-sns-wasm canister ID for small09.
- Added a script to import and fully populate a testnet.
- Imported all the common testnets

# Tests


lmuntaner (approved):
This big dfx is a little bit annoying. I wish we could split it or set this configurations elsewhere.

Not for this PR, maybe something to ask for the sdk team.

Thanks!

PR #1729, opened 2023-01-19
fix: tvl appearance in menu as block

peterpeterparker:
# Motivation

Particularly noticable on android, the TVL block makes its appearance in the menu without a fixed width which feels like a glitch.

# Changes

- use layout store from gix-component
- mitigate glitch with some animation


frontend/src/lib/components/common/MenuMetrics.svelte
```diff
@@ -1,21 +1,46 @@
...
+    }
+
+    &.open {
+      @include open;
+    }
+
+    &.sticky {
+      // On xlarge screen the menu can be always open
+      @include media.min-width(xlarge) {
+        @include open;
```
mstrasinskis, line 42: Just thinking out loud, will `transition: none;` for sticky view add any advantage...?

mstrasinskis (approved):
LGTM, thank you!

PR #1671, opened 2022-12-30
style: redo scoped styles

peterpeterparker:
# Motivation

In #1666 we have moved scoped styles to global style to workaround the build reproducibility of SvelteKit+ViteJS [#8291](https://github.com/sveltejs/kit/issues/8291).

This PR redo the changes has #1670 will solve the reproducibility issue by removing the unused vite manifest file from our build results.

# PRs

- [x] https://github.com/dfinity/nns-dapp/pull/1670

# Changes

- redo scoped style and remove related global style

PR #1731, opened 2023-01-19
style: undo scoped styles

peterpeterparker:
# Motivation

We have once again build reproducibility issue and despite the fact that we have fully remove the `vite-manifest.json` I suspect the style reproducibility we had to be the root cause.

Therefore undo the redo of the global style - i.e. undo PR #1671


mstrasinskis (approved):
LGTM, thanks 🤞

PR #1734, opened 2023-01-19
Asset reproducibility

bitdivine:
# Motivation
We want to progress:

image

This PR introduces a test but does not (yet) block PRs.  The important thing now is to decide what inputs should fully define the output of the frontend build.  The proposal here is that the output may depend on:

- code commit, including our own code and the `frontend/packages-lock.json` file.
- node version
- npm version
- tar
- xz

but should NOT depend on:
- build environment including:
  - time
  - operating system
  - docker
  - CPU

# Changes
- Extract the frontend building code from the build script.
- Run the frontend building code on:
  - different OSs (Ubuntu and Mac)
  - different versions of the OSs
  - at different times (now and with an artificial time shift into the future)

# Tests
- [ ] TODO: Verify that moving the frontend building code out doesn't change the build hash.


bitdivine:
Note: The reproducibility check is NOT blocking, so this PR can be merged while there are still reproducibility isssues.  Once the issues are resolved, this should presumably be made blocking.

build.sh
```diff
@@ -59,50 +59,7 @@ set -x
 ###################
 # frontend # (output: frontend/public/)
```
dskloetd, line 60: I think the final output of build-frontend.sh is assets.tar.xz and the frontend/public/ is more of an intermediate step, right?

.github/workflows/reproducible-assets.yaml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, ubuntu-20.04, macos-11, macos-12]
+        time: [now, future]
+    steps:
+      - name: Unbork mac
+        run: |
+          if command -v brew ; then
```
dskloetd, line 16: Should/does the step fail if brew is not available?

.github/workflows/reproducible-assets.yaml
```diff
@@ -0,0 +1,93 @@
...
+          future)
+            faketime -f "+5.5y 7m 9s" bash ./build-frontend.sh ;;
+          *)
+            ./build-frontend.sh ;;
+          esac
+          mkdir -p builds
+          cp assets.tar.xz "builds/assets_${{ matrix.os }}_${{ matrix.time }}.tar.xz"
+      - name: 'Upload assets'
+        uses: actions/upload-artifact@v3
+        if: always()
```
dskloetd, line 66: Do we try to upload the assets even if building fails?

build-frontend.sh
```diff
@@ -0,0 +1,51 @@
...
+# We use a local directory, and we don't delete it after the build, so that
+# assets can be inspected.
+tarball_dir="$TOPLEVEL/web-assets"
+rm -rf "$tarball_dir"
+echo "using $tarball_dir for tarball directory"
+cp -R "$TOPLEVEL/frontend/public/" "$tarball_dir/"
+
+# Bundle into a tight tarball
+# On macOS you need to install gtar + xz
+# brew install gnu-tar
```
lmuntaner, line 42: Why don't you install this and `xz` in the Github workflow if they are needed?

lmuntaner (approved):
I added a minor comment. Already approving.

.github/workflows/reproducible-assets.yaml
bitdivine: It does not and should not.  If brew is not availble we are either on Linux, which generally has up-to-date tools, or we are on a Mac that we cannot even fix.

.github/workflows/reproducible-assets.yaml
bitdivine: We always try.  I guess maybe it doesn't make sense to include failed attempts in the comparison but then the failed build should be visible and cause the workflow to fail.

build.sh
bitdivine: Correct.

build-frontend.sh
bitdivine: The build script isn't always run from the workflow.  In the github actions these are installed.

build.sh
dskloetd: Is it worth changing?
(Or did you change it and am I using GitHub wrong?)

dskloetd:
If I run `build-frontend.sh`, I get
```
file:///Users/dskloet/dev/nns-dapp/diff-assets/frontend/.svelte-kit/output/server/chunks/dapp.js:285
  {}.VITE_FEATURE_FLAGS.replace(/\\"/g, '"') ?? '{"ENABLE_SNS_2":false,"ENABLE_SNS_VOTING": false, "ENABLE_SNS_CACHING": false, "ENABLE_CKBTC_LEDGER": false}'
                        ^
```
but if I run `build.sh`, I get
```
Which deployment? Set DFX_NETWORK to one of:
...
```

Should build-frontend.sh do the same check? Should it never be run directly/manually? How do we make sure DFX_NETWORK is set when build-frontend.sh is run separately?

build.sh
bitdivine: I have not changed the sequence of build commands.  We can consider whether this should be changed but given that the backend consumes the tar.xz, I think that is a good point to cut control from the one build process to the other.

dskloetd: Previously, this comment block applied to just
`(cd "$TOPLEVEL/frontend" && npm ci && npm run build)`
for which it is correct.
But with your changes, this comment applies to
`"$TOPLEVEL/build-frontend.sh"`
for which I argue the comment isn't really accurate.
That's what I was trying to say.
Does that make sense?

bitdivine:
Re: Checking that DFX_NETWORK is set - that is a good idea.

Re: Running directly, yes, it is already being run directly, even manually.  You just did! :-)

dskloetd:
Actually, even when running
`$ DFX_NETWORK=mainnet ./build-frontend.sh`
I get this error:
```
file:///Users/dskloet/dev/nns-dapp/diff-assets/frontend/.svelte-kit/output/server/chunks/dapp.js:285
  {}.VITE_FEATURE_FLAGS.replace(/\\"/g, '"') ?? '{"ENABLE_SNS_2":false,"ENABLE_SNS_VOTING": false, "ENABLE_SNS_CACHING": false, "ENABLE_CKBTC_LEDGER": false}'
                        ^

TypeError: Cannot read properties of undefined (reading 'replace')
    at file:///Users/dskloet/dev/nns-dapp/diff-assets/frontend/.svelte-kit/output/server/chunks/dapp.js:285:25
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

Node.js v18.13.0
[vite-plugin-sveltekit-compile] Prerendering failed with code 1
error during build:
Error: Prerendering failed with code 1
    at ChildProcess.<anonymous> (file:///Users/dskloet/dev/nns-dapp/diff-assets/frontend/node_modules/@sveltejs/kit/src/exports/vite/index.js:551:15)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess._handle.onexit (node:internal/child_process:291:12)
```

Only after running `DFX_NETWORK=mainnet ./build.sh` (which succeeds) does `DFX_NETWORK=mainnet ./build-frontend.sh` also succeed by itself.

bitdivine:
Just a guess, did you create the configuration file?  See line 55 here: https://github.com/dfinity/nns-dapp/pull/1734/files#diff-35b54e6f328132b23ac04d6b4d2ff56d2455bfe84a5a70cf0e04cc336249a407R55

bitdivine:
The sequence is:
* Decide which network to deploy to
* assemble the URLs, canister IDs and other config for that network
* build

dskloetd:
Right, I forgot about `config.sh` but `build.sh` runs it for you.

I guess we always need the first 60 lines of build.sh. Maybe that should be extracted and we should have a version of build-frontend.sh that shouldn't be run manually and one that can be run manually, which also includes the first 60 lines of build.sh?

PR #1735, opened 2023-01-20
GIX-1173: Implement Caching Canister

lmuntaner:
# Motivation

Use caching canister to improve performance when loading SNSes

# Changes

* New feature flag: ENABLE_SNS_CACHING
* New env var but not yet connected to the build step: CACHING_CANISTER_URL
* New sns-caching api function: querySnsProjects that returns a list of SNSes.
* New sns service loadSnsProjects that tries to load with the caching canister. If there is an error, loads the old way.
* initSns app service checks the env vars and executes new loadSnsProjects or loadSnsSummaries.

# Tests

* Test new api querySnsProjects.
* Test change in initSns service.
* Test new loadSnsProjects service.


lmuntaner:

frontend/scripts/build.csp.mjs
```diff
@@ -114,11 +114,19 @@ const updateCSP = (indexHtml) => {
     );
   }
 
+  // Caching canister enabled ONLY in small12 for now
+  const enableSnsCacheCanister = process.env.VITE_DFX_NETWORK === "small12";
+  // TODO: Use env var https://dfinity.atlassian.net/browse/GIX-1245
+  const cachingCanisterUrl =
+    "https://5v72r-4aaaa-aaaaa-aabnq-cai.raw.small12.dfinity.network";
```
peterpeterparker, line 122: we should fetch the images in a certified way (not `.raw`)

frontend/scripts/build.csp.mjs
```diff
@@ -143,6 +151,10 @@ const cspConnectSrc = () => {
     process.env.VITE_LEDGER_CANISTER_URL,
   ];
 
+  if (enableSnsCacheCanister) {
+    src.push(cachingCanisterUrl);
```
peterpeterparker, line 156: ditto, certified not `.raw`

frontend/src/lib/api/sns-caching.api.ts
```diff
@@ -0,0 +1,268 @@
...
+  root: string;
+  governance: string;
+  ledger: string;
+  swap: string;
+  dapps: string[];
+  archives: [];
+  index: string;
+};
+
+export type CachedSns = {
```
peterpeterparker, line 35: - can we place the types in `types` package?

- can Max provide did files for the types so that we don't duplicate these manually?

frontend/src/lib/api/sns-caching.api.ts
```diff
@@ -0,0 +1,268 @@
...
+    derived: convertDerived(swap_state.derived),
+  },
+  icrc1_metadata: convertIcrc1Metadata(icrc1_metadata),
+  icrc1_fee: convertOptionalNumToBigInt(icrc1_fee[0]),
+});
+
+const convertDtoData = (data: CachedSnsDto[]): CachedSns[] =>
+  data.map(convertSnsData);
+
+export const querySnsProjects = async (): Promise<CachedSns[]> => {
```
peterpeterparker, line 253: there is a new package `$lib/rest` which contains `fetch` queries (for the tvl). what do you think, should we move this there?

note: not against another naming than `rest`, just thought it's nice to have a split between api and fetch

frontend/src/lib/api/sns-caching.api.ts
```diff
@@ -0,0 +1,268 @@
...
+  const response = await fetch(
+    `${CACHING_CANISTER_URL}/v1/sns/list/latest/slow.json`
+  );
+  if (!response.ok) {
+    throw new Error("Error loading SNS projects from caching canister");
+  }
+  const data: CachedSnsDto[] = await response.json();
+  try {
+    const converted = convertDtoData(data);
```
peterpeterparker, line 262: nth: webstorm would probably warns me that `converted` is not used

```
try {
   return convertDtoData(data);
   ...
```

frontend/src/lib/api/sns-caching.api.ts
```diff
@@ -0,0 +1,268 @@
...
+  icrc1_metadata: convertIcrc1Metadata(icrc1_metadata),
+  icrc1_fee: convertOptionalNumToBigInt(icrc1_fee[0]),
+});
+
+const convertDtoData = (data: CachedSnsDto[]): CachedSns[] =>
+  data.map(convertSnsData);
+
+export const querySnsProjects = async (): Promise<CachedSns[]> => {
+  const response = await fetch(
+    `${CACHING_CANISTER_URL}/v1/sns/list/latest/slow.json`
```
peterpeterparker, line 255: not related to your pr, but why is it call `slow.json`?

frontend/src/lib/services/$public/sns.services.ts
```diff
@@ -6,13 +7,61 @@ import { i18n } from "$lib/stores/i18n";
...
+        swapCanisterId: Principal.fromText(sns.canister_ids.swap_canister_id),
+        governanceCanisterId: Principal.fromText(
+          sns.canister_ids.governance_canister_id
+        ),
+        swap: toNullable(sns.swap_state.swap),
+        derived: toNullable(sns.swap_state.derived),
+      })),
+    ];
+    snsQueryStore.setData(snsQueryStoreData);
+    cachedSnses.forEach((sns) => {
```
peterpeterparker, line 41: shouldn't we set `functions` and `fee` once instead of looping and pushing canister after canister?

pushing one after the other might trigger more UI repaint than applying once all canisters therefore it might be UI wise more performant to do so. what you think?

frontend/src/lib/services/$public/sns.services.ts
```diff
@@ -6,13 +7,61 @@ import { i18n } from "$lib/stores/i18n";
...
+          rootCanisterId,
+          fee: sns.icrc1_fee,
+          certified: true,
+        });
+      }
+    });
+    // TODO: PENDING to be implemented, load SNS parameters.
+  } catch (err) {
+    // If caching canister fails, fallback to the old way
+    await loadSnsSummaries();
```
peterpeterparker, line 61: should we? in normal use case I would say yes but in high load situation, like Sns-1 launch, this feel like it might adds even some more load and therefore lead to some performance issue.

frontend/src/lib/api/sns-caching.api.ts
```diff
@@ -0,0 +1,268 @@
...
+      return [key, { Int: BigInt(value.Int[0]) }];
+    }
+    if ("Nat" in value) {
+      return [key, { Nat: BigInt(value.Nat[0]) }];
+    }
+    return [key, value];
+  });
+};
+
+const convertSnsData = ({
```
peterpeterparker, line 224: when we load Sns data currently, we ensure that all required fields are provided, notably because the conversion from proto to candid makes lot of fields as potentially nullable.

we do such filters in `mapAndSortSnsQueryToSummaries`. notably

```
// Only those that have valid metadata, toke, sale and derived information are - and can be - considered as valid
  const validSwapSummaries: ValidSummary[] = allSummaries.filter(
    (entry: OptionalSummary): entry is ValidSummary =>
      entry.swap !== undefined &&
      entry.swap.params !== undefined &&
      entry.swapCanisterId !== undefined &&
      entry.governanceCanisterId !== undefined &&
      entry.derived !== undefined &&
      entry.metadata !== undefined &&
      entry.token !== undefined
  );
```

is this also performed here?

peterpeterparker (changes requested):
Thanks!

Added few comments and questions.

Blocking it nothing related to your code, but because we should not use the `.raw` domain.

frontend/scripts/build.csp.mjs
lmuntaner: We have some issues with CORS. In a testnet it's not a problem, but from localhost it's blocked because of it. But it works with `raw`.

@bitdivine is already aware of it.

The `raw` url will not make it to mainnet but we need it (for now) for local development).

Do we have a way to identify if I'm running the frontend in localhost from build.csp.mjs? I believe that as far as that's concerned, there is no way to tell whether I build the config for small12 or for localhost pointing to small12.

Maybe I can explain you the issue and we can find a better solution?

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: I can add an endpoint in the stores, sounds good.

frontend/src/lib/api/sns-caching.api.ts
lmuntaner: Yes, this is done in the derived store `snsSummariesStore`.

What I do is to load the store with the same data as before. All the checks stay the same because they are done in that data.

frontend/src/lib/api/sns-caching.api.ts
lmuntaner: I prefer to think of it as an API. That it's a json is more implementation details. Not necessarily is important for the users of `querySnsProjects`

frontend/src/lib/api/sns-caching.api.ts
lmuntaner: A question for @bitdivine 

Also we need to make sure this doesn't change, hehe.

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: I added a comment about it.

frontend/scripts/build.csp.mjs
lmuntaner: Done. I use `raw` only for local development

frontend/src/lib/api/sns-caching.api.ts
lmuntaner: Added a comment and an issue: https://dfinity.atlassian.net/browse/GIX-1249

frontend/src/lib/api/sns-caching.api.ts
lmuntaner: done

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: done

lmuntaner:

lmuntaner:
FYI @mstrasinskis @peterpeterparker 

I deployed it to small06 where there is no caching canister (also no SNSes thought) and all works: https://qvhpv-4qaaa-aaaaa-aaagq-cai.small06.testnet.dfinity.network

frontend/src/lib/api/sns-caching.api.ts
peterpeterparker: Ok

frontend/src/lib/api/sns-caching.api.ts
peterpeterparker: Make me thing, maybe `v1` and `sns/list/latest/slow.json` are both worth two constants?

dfx.json
```diff
@@ -167,7 +167,8 @@
         "IDENTITY_SERVICE_URL": "https://identity.ic0.app/",
         "FEATURE_FLAGS": {
           "ENABLE_SNS_2": false,
-          "ENABLE_SNS_VOTING": false
+          "ENABLE_SNS_VOTING": false,
+          "ENABLE_SNS_CACHING": true
```
peterpeterparker, line 171: This isn't mainnet? It should be `false` at the time no @lmuntaner ?

lmuntaner: Good catch! Weird that "staging" I put "false" but not mainnet.

frontend/src/lib/api/sns-caching.api.ts
lmuntaner: Yes, I like that.

dfx.json
peterpeterparker: Did not noticed it in the first review neither, coolio then!

Already gonna approve, thanks for the PR and changes

peterpeterparker (approved):
LGTM, thanks

dfx.json
lmuntaner: Thanks for the review!

PR #1736, opened 2023-01-20
chore: Test full clean dockerized reproducible builds

bitdivine:
# Motivation
We have reproducibility issues.

# Changes
- Add a test that, for each of 4 operating systems (2 Ubuntu, 2 Mac):
 - Installs docker naively, as an end user would, and without a cache so that reproducibility issues are not hidden by consistent cache results.
 - Runs a full, clean, dockerized build.

# Tests
- See the test results.  Runs dying are not caused by a broken test but by the build breaking sometimes, as it does for users trying to reproduce builds.  Also, runs take up to 2 hours for Mac.


lmuntaner (commented):
Where do you compare the builds?

I don't see the code, are there changes in `./scripts/docker-build`?

bitdivine:
> Where do you compare the builds?
There is no comparison in this PR.  This PR makes the builds.

lmuntaner:
Ah, ok, I thought I was missing something. Thanks!

lmuntaner (approved):
Do we want to run it also in main?

PR #1737, opened 2023-01-20
Chore: Add created at parameter in transactions

lmuntaner:
# Motivation

Add createdAt param to all transactions to avoid that a retry from agent-js or the browser triggers a duplicated transaction.

# Changes

* Use createdAt param from nns-js and sns-js in all transactions.

# Tests

* Add test for new param.

### DIFF
--- a/frontend/src/lib/api/ledger.api.ts
+++ b/frontend/src/lib/api/ledger.api.ts
@@ -16,6 +16,7 @@ import { AccountIdentifier, LedgerCanister } from "@dfinity/nns";
  * @param {string} params.to send ICP to destination address - an account identifier
  * @param {ICP} params.amount the amount to be transferred in ICP
  * @param {number[] | undefined} params.fromSubAccount the optional subaccount that would be the source of the transaction
+ * @param {bigint | undefined} params.createdAt the optional timestamp of the transaction. Used to avoid deduplication.
  */
 export const sendICP = async ({
   identity,
@@ -23,12 +24,14 @@ export const sendICP = async ({
   amount,
   fromSubAccount,
   memo,
+  createdAt,
 }: {
   identity: Identity;
   to: string;
   amount: TokenAmount;
   fromSubAccount?: SubAccountArray | undefined;
   memo?: bigint;
+  createdAt?: bigint;
 }): Promise<BlockHeight> => {
   logWithTimestamp(`Sending icp call...`);
   const { canister } = await ledgerCanister({ identity });
@@ -38,6 +41,7 @@ export const sendICP = async ({
     amount: amount.toE8s(),
     fromSubAccount,
     memo,
+    createdAt,
   });
   logWithTimestamp(`Sending icp complete.`);
   return response;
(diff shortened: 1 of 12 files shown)


lmuntaner:

frontend/src/lib/utils/date.utils.ts
```diff
@@ -109,3 +109,4 @@ export const secondsToTime = (seconds: number): string => {
 };
 
 export const nowInSeconds = (): number => Math.round(Date.now() / 1000);
+export const nowInNanoSeconds = (): number => nowInSeconds() * 1e9;
```
mstrasinskis, line 112: Probably it would be better to use not rounded value.

mstrasinskis (approved):
LGTM, one question.

frontend/src/lib/utils/date.utils.ts
mstrasinskis: There is also `window.performance.now()` available.

mstrasinskis: But Date.now() is preferable because of testing :D

frontend/src/lib/api/ledger.api.ts
```diff
@@ -16,19 +16,22 @@ import { AccountIdentifier, LedgerCanister } from "@dfinity/nns";
...
  * @param {number[] | undefined} params.fromSubAccount the optional subaccount that would be the source of the transaction
+ * @param {bigint | undefined} params.createdAt the optional timestamp of the transaction. Used to avoid deduplication.
  */
 export const sendICP = async ({
   identity,
   to,
   amount,
   fromSubAccount,
   memo,
+  createdAt,
```
peterpeterparker, line 27: you explicitely did not wanted to set a default value in the API if not provided?

lmuntaner: Correct.

I could also see the date being set here. Not a strong opinion to be honest.

But since this is the "api", I thought that it was ok to not set it here.

Yet, reviewing it, I set it in the api for stakeNeuron...

Shall I set it here?

frontend/src/lib/utils/date.utils.ts
lmuntaner: hehe, sure!

frontend/src/lib/api/ledger.api.ts
peterpeterparker: No strong opinion neither, I was just going through the changes and had the feeling that we added the information everywhere, that's why I was asking. Works for me for sure.

lmuntaner: I think I'll change. created_at will always be set in the "api" folder. This way we avoid duplication.

frontend/src/lib/utils/date.utils.ts
lmuntaner: Actually, you're right. I should not use the rounded value. I'll use Date.now directly.

lmuntaner:
@peterpeterparker @mstrasinskis  I have kept the createdAt in the params of the api functions. But if it's not sent I populate it there.

I have removed the changes in the "services" level. Making this PR purely a change in the "api" layer. Which I believe it's a cleaner solution.

Let me know what you think.

peterpeterparker (approved):
LGTM, thanks

PR #1740, opened 2023-01-20
Gix 1219 fetch proposal

mstrasinskis:
# Motivation

Fetch sns proposal on SnsProposalDetail page. The nns approach was reused.

# Changes

- `queryProposal` api
- `loadSnsProposal` service (same pattern as nns version)
- context store `SelectedSnsProposalStore`
- `SnsProposalDetail` component that fetch sns proposal
- update `ic-js` version

# Tests

- `queryProposal`
- `loadSnsProposal`
- `SnsProposalDetail`

# Screenshots

image


frontend/src/lib/services/$public/sns-proposals.services.ts
```diff
@@ -41,3 +47,98 @@ export const loadSnsProposals = async ({
...
+  silentErrorMessages,
+  silentUpdateErrorMessages,
+  strategy = "query_and_update",
+}: {
+  rootCanisterId: Principal;
+  proposalId: bigint;
+  setProposal: (proposal: SnsProposalData) => void;
+  handleError?: (certified: boolean) => void;
+  callback?: (certified: boolean) => void;
+  silentErrorMessages?: boolean;
```
lmuntaner, line 70: Why do you prefer adding more parameters instead of having this logic in the `handleError` passed?

When I read the parameters, I'm not so sure what they do...

I have the feeling I could do the same with only `handleError` is that right?

lmuntaner (commented):
Just a small question

frontend/src/lib/services/$public/sns-proposals.services.ts
mstrasinskis: Good idea to simplify it. Initially it was done the nns way. The function was updated.

mstrasinskis:
Replaced with another PR

PR #1744, opened 2023-01-23
refactor: move wasm canister id

peterpeterparker:
# Motivation

Group wasm canister id with canister ids constants.


lmuntaner (approved):
👍

PR #1746, opened 2023-01-23
refactor: mv rs -> rs/backend

bitdivine:
# Motivation
We need space for other Rust canisters.

# Changes
* Move `rs` to `rs/backend`

# Tests
There is no change in functionality; existing tests should suffice.


lmuntaner (approved):
LGTM, align with @mstrasinskis because he is working in some changes in the backend this week.

PR #1748, opened 2023-01-23
Chore: Enable Merge HW controlled neurons

lmuntaner:
# Motivation

Users can merge neurons controlled by HWs.

# Changes

* Check HW version in mergeNeurons service.
* Change in isMergeableNeuron util, check whether neuron is controllable instead checking if it's HW controlled.
* Remove case of HW controlled neuron in getMergeableNeuronMessageKey.

# Tests

* Change test case that allows merging HW controlled neurons.
* Add new test to check that mergeNeurons checks the HW version.

### DIFF
--- a/frontend/src/lib/services/neurons.services.ts
+++ b/frontend/src/lib/services/neurons.services.ts
@@ -28,6 +28,7 @@ import {
 } from "$lib/constants/neurons.constants";
 import type { LedgerIdentity } from "$lib/identities/ledger.identity";
 import { getLedgerIdentityProxy } from "$lib/proxy/ledger.services.proxy";
+import { accountsStore } from "$lib/stores/accounts.store";
 import { startBusy, stopBusy } from "$lib/stores/busy.store";
 import { definedNeuronsStore, neuronsStore } from "$lib/stores/neurons.store";
 import {
@@ -60,6 +61,7 @@ import {
   isEnoughToStakeNeuron,
   isHotKeyControllable,
   isIdentityController,
+  isNeuronControlledByHardwareWallet,
   userAuthorizedNeuron,
   validTopUpAmount,
 } from "$lib/utils/neuron.utils";
@@ -409,9 +411,17 @@ export const mergeNeurons = async ({
         translate({ labelKey: messageKey ?? "error.governance_error" })
       );
     }
+
     const identity: Identity = await getIdentityOfControllerByNeuronId(
       targetNeuronId
     );
+    const accounts = get(accountsStore);
+    if (isNeuronControlledByHardwareWallet({ neuron: neuron2, accounts })) {
+      await assertLedgerVersion({
+        identity,
+        minVersion: CANDID_PARSER_VERSION,
+      });
+    }
 
     await mergeNeuronsApi({ sourceNeuronId, targetNeuronId, identity });
     success = true;
--- a/frontend/src/lib/services/neurons.services.ts
+++ b/frontend/src/lib/services/neurons.services.ts
@@ -399,13 +399,13 @@ export const mergeNeurons = async ({
 }): Promise<NeuronId | undefined> => {
   let success = false;
   try {
-    const { neuron: neuron1 } = await getIdentityAndNeuronHelper(
+    const { neuron: sourceNeuron } = await getIdentityAndNeuronHelper(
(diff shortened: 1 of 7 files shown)


lmuntaner:

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -466,9 +465,6 @@ const getMergeableNeuronMessageKey = ({
   if (isSpawning(neuron)) {
     return "neurons.cannot_merge_neuron_spawning";
   }
-  if (isNeuronControlledByHardwareWallet({ neuron, accounts })) {
-    return "neurons.cannot_merge_hardware_wallet";
```
peterpeterparker, line 470: you can clean the i18n key

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -815,6 +818,52 @@ describe("neurons-services", () => {
       expect(toastsShow).toHaveBeenCalled();
       expect(spyMergeNeurons).not.toHaveBeenCalled();
     });
+
+    it("should not update neuron if lower HW version than required", async () => {
```
peterpeterparker, line 822: typo? "should not update ..." shouldn't be "should not merge..."

frontend/src/lib/services/neurons.services.ts
```diff
@@ -409,9 +411,17 @@ export const mergeNeurons = async ({
         translate({ labelKey: messageKey ?? "error.governance_error" })
       );
     }
+
     const identity: Identity = await getIdentityOfControllerByNeuronId(
       targetNeuronId
     );
+    const accounts = get(accountsStore);
+    if (isNeuronControlledByHardwareWallet({ neuron: neuron2, accounts })) {
```
peterpeterparker, line 419: should we test also if `neuron1` is hw?

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -450,8 +450,7 @@ const isMergeableNeuron = ({
 }): boolean =>
   !hasJoinedCommunityFund(neuron) &&
   !isSpawning(neuron) &&
-  // Merging hardware wallet neurons is not yet supported
-  isNeuronControllableByUser({ neuron, mainAccount: accounts.main });
+  isNeuronControllable({ neuron, accounts });
```
peterpeterparker, line 453: should we test this or already covered by one?

peterpeterparker (approved):
Looks good, few minor questions so already approving

frontend/src/lib/services/neurons.services.ts
lmuntaner: To be mergeable both need to have the same controller.

"neuron2" is the one where we call "manage_neuron" on. Which is the why we use the identity from.

But I'll rename `neuron1` and `neuron2` to `sourceNeuron` and `targetNeuron` to be clear.

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: It's the test I changed from `does not allow user to select neurons` to `allows user to select neurons`.

It's inside the `when mergeable neurons by hardware wallet`.

The test to check that when the controller is no account was already in place.

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: thanks yes!

frontend/src/tests/lib/services/neurons.services.spec.ts
lmuntaner: yes, thanks!

frontend/src/lib/services/neurons.services.ts
lmuntaner: and btw, the check whether they are the same controller is above in the function `canBeMerged`.

peterpeterparker: Gotcha, thanks

frontend/src/lib/utils/neuron.utils.ts
peterpeterparker: Sounds good, thx

PR #1749, opened 2023-01-23
feat: ckBTC universe

peterpeterparker:
# Motivation

Integrate ckBTC in token switcher.

# Notes

- as we do not know what will be the future UI/UX, this PR integrates ckBTC in existing "universe" UX logic without changing it. It still relies on a single store for the selection of the universe but, because ckBTC is only available in "Accounts" views, fallback on Nns if use elsewhere

- when the fallback is applied, the URL of the browser is not updated. So if user just switch view without doing anything else, user lands back on ckBTC. Again, here we do not know the future UI/UX, so the a-ok solution

- because ckBTC has a ledger and minter but no other canister - like a root canister - I used the ledger canister id as universe identifier

- this PR renames few files and variables from "project" to "universe". That way we could use "universe" for IC / NNS, ckBTC, Sns or else but could keep the meaning "project" for "Sns project"

- in this PR no features (ledger or else are implemented). likewise, on "Accounts" under ckBTC selector the total amount of the NNS accounts is displayed

- all in all, this PR is lots of refactoring and fewer logic

# Changes

- rename and split exisiting `selectable-projects` and `selected-project` stores into "universe" and "project"
- extend `selectable-universes" with ckBTC and provide it only on "Accounts" path
- extend `selected-universe" with ckBTC and allows it only on "Accounts" path otherwise fallback to Nns
- update `<SelectUniverse... />` components to display ckBTC and be able to pick it
- dfx config `ENABLE_CKBTC_LEDGER` (false everywhere except locally)
- in routes, display Sns only if project is selected, not universe. Technically this would not be needed as the fallback is applied on the store already but, it's just cleaner

# Screenshots

Capture d’écran 2023-01-24 à 10 50 34
Capture d’écran 2023-01-24 à 10 50 39
Capture d’écran 2023-01-24 à 10 50 48
demo


frontend/src/lib/types/universe.ts
```diff
@@ -0,0 +1,6 @@
+import type { SnsSummary } from "$lib/types/sns";
+
+export interface Universe {
+  canisterId: string;
```
lmuntaner, line 4: Would it be useful to have a "type" here?

It would not add functionality, but it would make it more human readable.

```
type: "nns" | "ckBTC" | "sns";
```

frontend/src/lib/components/universe/UniverseLogo.svelte
```diff
@@ -3,18 +3,33 @@
...
+
+  let canisterId: string;
+  $: canisterId = universe.canisterId;
+
+  let ckBTC = false;
+  $: ckBTC = isNullish(summary) && isUniverseCkBTC(canisterId);
+
   let logo: string;
-  $: logo = summary?.metadata.logo ?? IC_LOGO;
+  $: logo = summary?.metadata.logo ?? (ckBTC ? CKBTC_LOGO : IC_LOGO);
```
lmuntaner, line 25: Isn't the logo in the ckBTC ledger metadata?

frontend/src/lib/components/universe/SelectUniverseCard.svelte
```diff
@@ -30,6 +32,10 @@
     currentPath: $pageStore.path,
     paths: [AppPath.Accounts, AppPath.Wallet],
   });
+
+  let ckBTC = false;
+  $: ckBTC =
+    isNullish(universe.summary) && isUniverseCkBTC(universe.canisterId);
```
lmuntaner, line 38: Why do you need the check `isNullish(universe.summary)`?

lmuntaner (approved):
A few comments but already approving, LGTM!

frontend/src/lib/components/universe/UniverseLogo.svelte
peterpeterparker: Yep, added a todo

frontend/src/lib/components/universe/SelectUniverseCard.svelte
peterpeterparker: Good catch, a relic of the development that can be removed now

frontend/src/lib/types/universe.ts
peterpeterparker: I liked the idea, so I just gave it a try but, unfortuntely noticed that if we do so, we could not remove entirely `isUniverseCkBTC`. In current form there are few places where the canister id from the url is still used. So we would end up having two "source of thruth", `isUniverseCkBTC` and the `universe.type`, which does not sound ideal. Maybe something we can try harder later in another PR?

lmuntaner: Ok, makes sense. And relying on that type to check doesn't sound bullet proof. Using the util seems better.

PR #1757, opened 2023-01-25
refactor: projects sns prefix and folder

peterpeterparker:
# Motivation

Sometimes we used `sns` as prefix and folder for the derived stores, sometimes not. This PR aligns the Sns related derived stores by moving all of them under `derived/sns` and prefixing with `sns`.

Per extension, it makes clearer the distinction between "universes" and "project term used for sns".

# Changes

No functional changes. Only refactoring.


lmuntaner (approved):
Thanks!

PR #1759, opened 2023-01-25
Fix CI required checks

bitdivine:
# Motivation
A PR was merged even though the corresponding test failed.  If a dependent test failed, the final check is skipped and historically that would prevent the PR from merging.  It seems that "needs" is now treated as a conditional, so "skipped" is treated as success.  https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks  This is not desired

# Changes
- Always run the final check, to see whether required GitHub jobs have succeeded.

# Tests
- I made a toy  job that always fails, set it as a dependency and verified that the final check was run, not skipped, and failed.
- I repeated the same with the toy job succeeding, and the final check similarly succeeded.


dskloetd:
"A PR was merged" can you link to the specific PR for reference?

lmuntaner (commented):
Could we remove the `xxx-passes` job and let the previous be the last one?

Why do we want this `xxx_passes` pattern?

.github/workflows/build.yml
```diff
@@ -139,7 +139,13 @@ jobs:
...
   build-pass:
     needs: ["build"]
+    if: ${{ always() }}
     runs-on: ubuntu-20.04
     steps:
-      - name: Build workflow passes
-        run: echo OK
+      - name: Checks workflow passes
+        run: |
+          if echo '${{ toJson(needs) }}' | jq 'to_entries[] | select(.value.result != "success")' | grep .
```
dskloetd, line 147: This is a matter of taste but I was a bit confused by `grep .` being used like that.
What do you think about this?
`if [ "$(echo '${{ toJson(needs) }}' | jq 'to_entries[] | select(.value.result != "success")' )" ]`

Feel free to answer with "We use `grep .` like this all the time. You'll get used to it."

bitdivine:
Re the XXX pattern: There are two drivers:  One is that it lets everyone in the team contribute tests and normal peer reviews decide on what tests should be blocking, rather than requiring someone with superpowers to always be available and do the right thing.  The second benefit, not originally intended but that has turned out to be valuable, is that adding tests as blocking manually has turned out to be error prone when working with matrices whereas the final check gets it right, or has so far as long as github doesn't skip it!

.github/workflows/build.yml
bitdivine: I used grep so that the failing entries are printed in the logs.  Could be useful for debugging.  I would also be OK with putting `tee /dev/stderr` after the echo so that successful tests are also printed.  As long as we can see on what basis some jobs are considered to have failed we have the information to fix or update this code if it behaves in unexpected ways.

dskloetd:
If `build` fails and `build-passes` needs `build`, doesn't that means that `build-passes` is skipped and the whole flow still passes?

.github/workflows/build.yml
dskloetd: OK, so it's doing things that make sense but it's not obvious that it's doing those things.
What do you think about
```
error_result="$(echo '${{ toJson(needs) }}' | jq 'to_entries[] | select(.value.result != "success")')"
if [ $error_result ]
then
  echo "Some steps didn't pass: $error_result"
  exit 1
else
  echo "All steps passed."
fi
```
(Please don't delay the merge on this discussion.)

lmuntaner:
> If `build` fails and `build-passes` needs `build`, doesn't that means that `build-passes` is skipped and the whole flow still passes?

I'm guessing that's why he used `if: ${{ always() }}`. That ensures the last step is not skipped. Is that right @bitdivine ?

.github/workflows/build.yml
bitdivine: Sur ething.  I think the `[ $error_result ]` won't actually work as intended as it will expand but something like `test -n "$error_result"` or `[[ "$test_result" != "" ]]` will.  In terms of clarity the last is probably best.

bitdivine: @dskloetd mentioned that `[ "$error_result" ]` will work.

bitdivine: @dskloetd We can factor out the checking logic like this: https://github.com/dfinity/nns-dapp/pull/1767/files

lmuntaner (approved):

dskloetd (approved):

PR #1760, opened 2023-01-25
GIX-1192: Handle overload in UI

lmuntaner:
# Motivation

Two main motivations:

* Handle gracefully when something is not loaded (such as accounts for the transaction modal)
* Load resources in the page that they are needed, to reduce load in the IC.

# Changes

* Use a Spinner in SelectableDropdownAccount when accounts are being loaded.
* Move "loadSnsSwapCommitments" to Launchpad instead of on init.
* "loadSnsSwapCommitments" does not query commitments again if they already loaded.

# Tests

* Add test for spinner in SelectableDropdownAccount.
* New test for loadSnsSwapCommitments.

### DIFF
--- a/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
+++ b/frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
@@ -100,16 +100,16 @@
 
 <form on:submit|preventDefault={goNext} data-tid="transaction-step-1">
   <div class="select-account">
-    {#if selectedAccount !== undefined}
-      <KeyValuePair>
-        <span slot="key" class="label">{$i18n.accounts.source}</span>
+    <KeyValuePair>
+      <span slot="key" class="label">{$i18n.accounts.source}</span>
+      {#if selectedAccount !== undefined}
         <AmountDisplay
           slot="value"
           singleLine
           amount={selectedAccount?.balance}
         />
-      </KeyValuePair>
-    {/if}
+      {/if}
+    </KeyValuePair>
 
     {#if canSelectSource}
       <SelectAccountDropdown {rootCanisterId} bind:selectedAccount />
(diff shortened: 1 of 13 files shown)


lmuntaner:
New UI:

Screenshot 2023-01-25 at 11 06 50

There is a small layout shift of 3 pixels between spinner and the normal dropdown. Due to the nature of edge case of this, I didn't find it worth it to remove totally the layout shift.

lmuntaner:

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
```diff
@@ -100,16 +100,16 @@
...
-          slot="value"
-          singleLine
-          amount={selectedAccount?.balance}
-        />
-      </KeyValuePair>
-    {/if}
+    <KeyValuePair>
+      <span slot="key" class="label">{$i18n.accounts.source}</span>
+      <!-- svelte:fragment needed to avoid warnings -->
+      <!-- Svelte issue: https://github.com/sveltejs/svelte/issues/5604 -->
```
mstrasinskis, line 106: TIL

frontend/src/lib/services/sns.services.ts
```diff
@@ -30,9 +34,35 @@ import { getAccountIdentity, syncAccounts } from "./accounts.services";
...
+ *
+ * If the commitments are already loaded, it will not reload them.
+ *
+ * We rely that the projects are already loaded to skip loading the commitments.
+ * If commitments is 0, it will load them always.
+ *
+ * Therefore, this can be called before the projects are loaded.
+ */
+export const loadSnsSwapCommitments = async (): Promise<void> => {
+  const commtimentsCanisterIds = new Set(
```
mstrasinskis, line 48: Typo in co`mmt`imentsCanisterIds

mstrasinskis (approved):
LGTM, thanks. Only a typo.

frontend/src/lib/services/sns.services.ts
lmuntaner: thanks!

frontend/src/lib/modals/accounts/NewTransaction/TransactionForm.svelte
lmuntaner: Me too, I learned it with this PR.

PR #1764, opened 2023-01-26
Release 2023 01 26

mstrasinskis:
# Motivation

Release candidate 2023-01-26.

# Changes

- changelog update
- update npm and cargo patch version


Cargo.toml
```diff
@@ -5,7 +5,7 @@ members = [
 ]
 
 [workspace.package]
-version = "2.0.13"
+version = "2.0.14"
```
lmuntaner, line 8: Oops, I didn't notice you didn't merge yours yet...

I used 2.0.14 already when I merged. But shall we keep it this way?

lmuntaner (approved):
Thanks, just a comment about the version.

I believe it's ok to keep 2.0.14, but that's the one in `main` at the moment.

Cargo.toml
mstrasinskis: Thanks, done!

PR #1765, opened 2023-01-26
Chore: Add fee in SNS transactions

lmuntaner:
# Motivation

Prepare for SNS support for hardware wallets.

To ensure that HW users signs what's really happening, we added the fee in the SNS transactions.

# Changes

* Add "fee" param to sns.api function "stakeNeuron".
* Add "fee" param to sns-ledger.api function "transfer".
* Read fee and fail if not present in sns-accounts service "snsTransferTokens".
* Read fee and fail if not present in sns-neurons service "stakeNeuron".
* Upgrade ic-js

# Tests

* Adapt tests to new requirement.

### DIFF
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -374,18 +374,28 @@ export const querySnsNeuron = async ({
   return neuron;
 };
 
+/**
+ * Stake SNS neuron.
+ *
+ * param.fee is mandatory to ensure that it's show for hardware wallets.
+ * Otherwise, the fee would not show in the device and the user would not know how much they are paying.
+ *
+ * This als adds an extra layer of safety because we show the fee before the user confirms the transaction.
+ */
 export const stakeNeuron = async ({
   controller,
   stakeE8s,
   rootCanisterId,
   identity,
   source,
+  fee,
 }: {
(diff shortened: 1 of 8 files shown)


lmuntaner:

frontend/src/lib/api/sns.api.ts
```diff
@@ -374,18 +374,28 @@ export const querySnsNeuron = async ({
...
+ *
+ * This als adds an extra layer of safety because we show the fee before the user confirms the transaction.
+ */
 export const stakeNeuron = async ({
   controller,
   stakeE8s,
   rootCanisterId,
   identity,
   source,
+  fee,
```
mstrasinskis, line 391: nitpick: `fee` to `feeE8s` to follow the same pattern as `stakeE8s`?

mstrasinskis (approved):
LGTM, thanks!

frontend/src/lib/api/sns.api.ts
lmuntaner: Good point, I believe we have a little bit of a mess.

In ic-js we already have this inconsistency.

In general, we never agreed to use `xxxxE8s` or units in all variable names. Therefore, now it's a little bit all over the place.

PR #1766, opened 2023-01-26
Additional tar flags to improve nns-dapp frontent reproducibility

dskloetd:
# Motivation

We are taking steps towards reproducible builds.
Because the frontend builds much faster than the entire canister, we separately check if the assets tarball is reproducible.
There are several potential reasons why the frontend tarball might differ:
1. The contents are different.
2. The same contents create different `.tar` archives because different metadata is included.
3. The same archive is compressed differently because different metadata is included.

# Changes

Add additional flags to the tar command to make the archive (more) reproducible.
I found these on https://reproducible-builds.org/docs/archives/
This change improves #2 from the list above.

# Tests

When inspecting the generated tarballs, I previously noticed that the same content produced different tarballs and I didn't see that after adding these flags. For the rest I'm relying on existing workflows to make sure this doesn't break anything else.

I've also noticed issues with the actual content being different but strangely in the last few runs that seems to have gone away. What I saw was that different tarballs had different css files with the same content. I suspect that different components (e.g. SignInNeuron and SignInCanister) have the same css and that the files got deduped in an inconsistent way.

### DIFF
--- a/build-frontend.sh
+++ b/build-frontend.sh
@@ -43,7 +43,7 @@ cp -R "$TOPLEVEL/frontend/public/" "$tarball_dir/"
 # brew install xz
 cd "$tarball_dir"
 
-# --mtime, --sort, --owner, --group, --numeric-owner and --pax-option are all
+# --mtime, --sort, --owner, --group, --numeric-owner and --format are all
 # there to get a tarball that's reproducible across different platforms.
 # See https://reproducible-builds.org/docs/archives/
 "$tar" cJv \
@@ -52,7 +52,6 @@ cd "$tarball_dir"
   --owner=0 \
   --group=0 \
   --numeric-owner \
-  #--pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime \
   --format=ustar \
   --exclude .last_build_id \
   -f "$TOPLEVEL/assets.tar.xz" \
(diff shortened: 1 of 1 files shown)


bitdivine (approved):
LGTM, thanks!

build-frontend.sh
```diff
@@ -43,7 +43,19 @@ cp -R "$TOPLEVEL/frontend/public/" "$tarball_dir/"
...
+# --mtime, --sort, --owner, --group, --numeric-owner and --format are all
+# there to get a tarball that's reproducible across different platforms.
+# See https://reproducible-builds.org/docs/archives/
+"$tar" cJv \
+  --mtime='2021-05-07 17:00Z' \
+  --sort=name \
+  --owner=0 \
+  --group=0 \
+  --numeric-owner \
+  --format=ustar \
```
bitdivine, line 55: Note:  We do have some funky filenames in the svelte code.  E.g.:
```
frontend/.svelte-kit/output/server/entries/pages/(app)/(u)/(detail)/neuron/_page.ts.js
```
however the ustar format limitation as described [here](https://www.gnu.org/software/tar/manual/html_node/Formats.html#ustar) doen't seem to mind weird characters.  The path length constraint seems to be fine.

dskloetd: We could go with `--pax-option=exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime` instead, if we don't want to change the format.
What do you think?

bitdivine: I think it's fine.  Just saying that I have done due diligence on that aspect, in case there are other reviewers who have similar questions.

PR #1767, opened 2023-01-26
Factor out need checking

bitdivine:
# Motivation
This is a follow up on conversations in https://github.com/dfinity/nns-dapp/pull/1759  In particular, the logic that goes at the end of each workflow was being fine-tuned.  Rather than making the changes in many places, it possibly makes sense to factor out that final logic and tweak it in one place.

Why not to merge this PR: The common code is hardly longer than the code used to call up the common code, so factoring out seems a bit pointless, however it does make sure that the common code stays in sync.  A cost of such refactoring is that the repo needs to be checked out to get the file with the common logic and run it.  The checkout takes about 1s in the gh action.  Still, overall, this change may be worthwhile.

# Changes
- Create a local saved github action that checks that all needed jobs have completed successfully.
- Incorporate some ideas from #1759 into the common code.
- Replace the final check in each workflow file with a call to the common code.

# Tests
No additional checks should be needed.


.github/actions/needs_success/action.yaml
```diff
@@ -0,0 +1,27 @@
+name: 'Needed jobs succeeded'
+description: Checks that the jobs needed by the current job succeeded.
```
dskloetd, line 2: Maybe add an explanation of how/who this is different from what GitHub does by default and why we need this instead of GitHub's default behavior?

.github/actions/needs_success/action.yaml
```diff
@@ -0,0 +1,27 @@
...
+inputs:
+  needs:
+    description: "JSON describing the needed jobs."
```
dskloetd, line 5: Based on this description, I wouldn't know what to pass. But what we pass is literally something provided by GitHub itself, right?

And is it maybe useful to give an example of the format and explain based on that what we look at? The code below that does it also isn't super trivial.

.github/workflows/build.yml
```diff
@@ -142,10 +142,7 @@ jobs:
...
-        run: |
-          if echo '${{ toJson(needs) }}' | jq 'to_entries[] | select(.value.result != "success")' | grep .
-          then echo "You shall not pass:  Some required tests did not succeed"
-               exit 1
-          else echo "Congratulations, young Frodo."
-          fi
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/needs_success
+        with:
+          needs: '${{ toJson(needs) }}'
```
dskloetd, line 148: Does it have to be JSON to pass it or could it be passed "as is"?

dskloetd (changes requested):

.github/actions/needs_success/action.yaml
bitdivine: I tried to give the formula in the description, but then, even when escaped, GitHub tried to evaluate it.  I might have to resort to interspersing the formula with English words.

.github/workflows/build.yml
dskloetd: Did you try the most obvious way and it didn't work? Or you didn't touch it at all?

dskloetd (approved):

.github/workflows/build.yml
bitdivine: Dropping the toJson here?  I looked at it briefly.  Just passing the value through requires setting a different type on the argument.  As to why I was wary of following that path: GitHub actions can and do change, though, whereas JSON is pretty stable.  Changes can come in two forms: The datatype being passed can change or the actual data in that type may change.  Using JSON rather than some proprietary type up to the whim of GitHub avoids one source of risk, and the sanity check on the parsed data reduces the other source of risk.  So I feel as if the implementation is safe.  For infrastructure I'd choose safe over cool any day.

bitdivine: As usual, I have to close this to merge but am happy to follow up if wanted.

dskloetd: But what you convert to JSON here is provided directly by GitHub and will change if GitHub changes the format of `needs`. So I don't see how this is safer.

PR #1771, opened 2023-01-27
refactor: Factor out the rust build command

bitdivine:
# Motivation
We now have two rust canisters.  They need the same commands to run so factor out the build script.

# Changes
- Move the Rust build steps into a separate script.

# Tests
- [x] The wasm created by this should be identical to the wasm on the merge-root with main. 
  - Both yield `e72857b431955309b4b0d1d8423320f32bee9a6178db46b177ef9b0d4b16e549`
  - This PR:  https://github.com/dfinity/nns-dapp/actions/runs/4025418459/jobs/6918615826
  - Main: https://github.com/dfinity/nns-dapp/actions/runs/4024511085/jobs/6916629090

### DIFF
--- a/build-rs.sh
+++ b/build-rs.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -euxo pipefail
+set -euo pipefail
 
 ###############
 # Hjelpe meg!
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Rubber stamp. Thanks!

PR #1772, opened 2023-01-27
Nns types

bitdivine:
# Motivation
We want to call the sns canisters, but to depend only on the provided wasm types, not import the types from the IC repo.

- The IC repo is large, and increases compilation time substantially.
- The traits applied to the Rust types in the ic repo are not the ones we care about.  We need parsing and serialization to JSON.  Upstream is more concerned with protobuf, that we do not need.

# Changes
- Import sns canister did files
- Use didc to generate rust types.  But:
  - did makes all fields private, so in accessible.
  - has problems with anonymous empty records
  - has no mechanism for communicating additional traits
- Use sed and patch files to give the Rust types the traits we need and to fix bugs in the conversion.

# Tests
- These types are used in the current prototypical sns cache.


rs/sns_cache/src/types/did2rs.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
```
peterpeterparker, line 1: can you please move this script elsewhere - don't think it should find place under the source code tree. maybe in `./scripts`?

peterpeterparker (commented):
commonly canisters of the community on the IC have their rust code and did files structured the following way:

```
src/declarations/canister/did_file,js and ts
src/caniser/src/rust_code
src/frontend/
```

The types provided in this PR would match this pattern but the did files wouldn't. Should we also use a `src/declarations` for the did files as well or something similar since we do not have a root `src`?

```
./frontend
./declarations
./rs
```

I would be fine moving the did files into `src/declarations/sns_{root,ledger,...}/sns_{root,ledger,...}.did`.  Do I understand your proposal correctly?  Are you OK with that as well, @mstrasinskis ?  Btw sdk code often stores did files in `$REPO/candid/${CANISTER_NAME}.did` , for example the import functionality, not just for nns.  We probably want to get an idea of what their long term plan is for that as at the moment there is some inconsistency.  Generally I think it makes sense to organize types by language, as each language has its own convention and a given language's convention might well not match whatever dfinity chooses for did files.  Just as e.g. Rust types don't fit well into this pattern.  If we add python bindings, will we need a separate pythonpath entry for each canister?  But this is mile high optimization and for the release in days/weeks I really don't care where we put the types.  Whatever is most convenient for you folks is absolutely fine by me.

rs/sns_cache/src/types/did2rs.sh
```diff
@@ -0,0 +1,80 @@
...
+  #   - It applies the canidid Deserialize trait to all the types but not other traits that we need.
+  #   - It makes almost all the types and fields private, which is not very helpful.
+  #
+  # sed:
+  #   - adds additional traits after Deserialize
+  #   - Makes structures and their fields "pub"
+  #
+  # Final tweaks are defined manually and encoded as patch files.  The changes typically include:
+  #   - Replacing the anonymous result{} type in enums with EmptyRecord.  didc produces valid rust code, but
+  #     in a form that the Candid macro cannot handle.  Using a named type works around the limit sof the macro.
```
mstrasinskis, line 67:
```suggestion
  #     in a form that the Candid macro cannot handle.  Using a named type works around the limit of the macro.
```

rs/sns_cache/src/types/did2rs.sh
bitdivine: Personally I would prefer to have all the related code together but I don't mind all that much.  Done.

bitdivine:
Re did file locations:  We will probably want to align the paths with the locations given in dfx.json, which in turn come from the sdk.  At the moment, the sdk has at least two locations for the did files.  I'm assuming that the sdk will converge on one, presumably declarations/... so I'm fine with using that path until the sdk is consistent.  The sdk may develop different conventions for internal canisters and external ones; that is already the case for some things.  Anyway, not something we need to worry about now.

bitdivine:
One e2e test timed out.

rs/sns_cache/src/types/ic_sns_root.patch
```diff
@@ -0,0 +1,22 @@
+diff --git b/rs/sns_cache/src/types/ic_sns_root.rs a/rs/sns_cache/src/types/ic_sns_root.rs
```
peterpeterparker, line 1: actually why do we need these `.patch` files?

scripts/mk_nns_types.sh
```diff
@@ -0,0 +1,31 @@
...
+	EOF
+}
+[[ "${1:-}" != "--help" ]] || {
+  print_help
+  exit 0
+}
+
+cd "$(dirname "$(realpath "$0")")"
+GIT_ROOT="$(git rev-parse --show-toplevel)"
+for CANISTER_NAME in sns_ledger sns_governance sns_root sns_swap sns_wasm; do
```
peterpeterparker, line 23: nth: can/should `sns_ledger sns_governance sns_root sns_swap sns_wasm` be replaced with a dynamic list build from `declarations` folder?

rs/sns_cache/src/types/ic_sns_root.patch
bitdivine: There is a detailed description of what the patchfiles do here: https://github.com/dfinity/nns-dapp/pull/1772/files#diff-7f41c7c1539d733325a929ada0f06532221cb458338a92eee47b85891b55db5eR74-R79

scripts/mk_nns_types.sh
bitdivine: I believe not, as otherwise if we import more canisters to the declarations folder, this could break, even if we don't want rust bindings for that new canister. All it needs is for one of the new added canisters to have an API that takes `record{}` as an argument, and this is a common case, as this would cause rust to be generated that then fails to compile or even lint, with literally hundreds of quite confusing error messages.  didc (or dfx generate, which uses the exact same API call to do the conversion) and the Candid macros are far from flawless.  If we really do want to import that new canister, we would probably have to write a patch file to fix the inaccuracies in the generated rust.

bitdivine:
- The reproducibility test has one hash that is different, but that isn't blocking and David is working on that.  All other tests pass.

peterpeterparker (approved):
LGTM, thanks

rs/sns_cache/src/types/ic_sns_root.patch
peterpeterparker: Gotcha, thanks

scripts/mk_nns_types.sh
peterpeterparker: Gotcha, thanks

PR #1773, opened 2023-01-28
Fix: Encode decode ICRC-1 subaccounts

lmuntaner:
# Motivation

Upgrade ic-js to fix issue with encoding and decoding icrc-1 subaccounts.

# Changes

* ic-js versions upgraded

# Tests

Not applicable


peterpeterparker (approved):
Thanks

PR #1776, opened 2023-01-28
Fix: Publich Launchpad

lmuntaner:
# Motivation

Non logged in users can access Launchpad.

# Changes

* Do not call loadSnsSwapCommitments if not logged in.

# Tests

* Add test for this case.

### DIFF
--- a/frontend/src/lib/pages/Launchpad.svelte
+++ b/frontend/src/lib/pages/Launchpad.svelte
@@ -6,9 +6,13 @@
   import { snsProjectsCommittedStore } from "$lib/derived/sns/sns-projects.derived";
   import { onMount } from "svelte";
   import { loadSnsSwapCommitments } from "$lib/services/sns.services";
+  import { authStore } from "$lib/stores/auth.store";
+  import { nonNullish } from "$lib/utils/utils";
 
   onMount(() => {
-    loadSnsSwapCommitments();
+    if (nonNullish($authStore.identity)) {
+      loadSnsSwapCommitments();
+    }
   });
 
   let showCommitted = false;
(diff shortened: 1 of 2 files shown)


frontend/src/lib/pages/Launchpad.svelte
```diff
@@ -6,9 +6,13 @@
   import { snsProjectsCommittedStore } from "$lib/derived/sns/sns-projects.derived";
   import { onMount } from "svelte";
   import { loadSnsSwapCommitments } from "$lib/services/sns.services";
+  import { authStore } from "$lib/stores/auth.store";
+  import { nonNullish } from "$lib/utils/utils";
 
   onMount(() => {
-    loadSnsSwapCommitments();
+    if (nonNullish($authStore.identity)) {
```
peterpeterparker, line 13: - `isSignedIn($authStore.identity)`
- I think it needs auto-suscribe because user can sign-in from that screen

(not tested) something like

```
const loadSnsSale = async () => {
    if (!isSignedIn($authStore.identity)) {
      return;
    }
    
    await loadSnsSwapCommitments();
  };

  $: $authStore.identity, (async () => await loadSnsSale());
```

lmuntaner: Isn't there a reload when the user subscribe?

peterpeterparker: No.

There is a window reload on sign out "only".

lmuntaner: no, I see it needs your code

peterpeterparker: cool, was literally doing a demo so posting it any way ;) 

demo

lmuntaner: done

peterpeterparker (approved):
LGTM, thanks

PR #1777, opened 2023-01-30
style: global style for external link

peterpeterparker:
# Motivation

Use global style for external links instead of custom non contrating colors.


lmuntaner (approved):
Yay! a couple of `global` removed!

PR #1778, opened 2023-01-30
chore: remove unused component

peterpeterparker:
# Motivation

The `<ProposalMeta />` component is unused. Its usage was removed when we implemented the new UX of the proposals.

# Changes

- remove cmp and test


mstrasinskis (approved):
LGTM 👍

PR #1779, opened 2023-01-30
feat: remove dashboard routing fallback

peterpeterparker:
# Motivation

The dashboard has migrated its links to NNS-dapp voting view to the new "universe" scheme.

Therefore we can remove the specific fallback support we still had in place for the dashboard.

# Changes

- remove old dash url fallback support for the dashboard


mstrasinskis (approved):
LGTM, thanks!

PR #1780, opened 2023-01-30
GIX-1187: Decreate update calls, setup

lmuntaner:
# Motivation

Reuse certified data and reduce number of update calls.

# Changes

Main changes:
* New service util `createCachedQueryAndUpdate` that caches responses and returns them if not stale.
* Use the new `createCachedQueryAndUpdate` in `listNeurons` service.
* Remove `listNeurons` from app.services.ts.

Minor changes:
* Derived store `isLoggedInStore`.
* New store to keep track of the usage of the caching `cachingCallsStore`.

# Tests

PENDING


lmuntaner:
@peterpeterparker @dskloetd @mstrasinskis please review

frontend/src/lib/services/utils.services.ts
```diff
@@ -22,6 +28,134 @@ export type QueryAndUpdateIdentity = "authorized" | "anonymous" | "current";
...
+          certified: boolean;
+          response: R;
+        }) => {
+          cache.set({ certified, response });
+          strategyCalls.onLoadListeners.forEach((listener) => {
+            listener({ certified, response });
+          });
+          // Reset the listeners when we get the response from the "most certified" call of the strategy
+          if (
+            (strategy === "query_and_update" && certified) ||
```
mstrasinskis, line 128: Probably having only this condition `strategy === "query_and_update" && certified` is sufficient. In other cases there is only one `onLoad` call.

frontend/src/lib/services/utils.services.ts
```diff
@@ -22,6 +28,134 @@ export type QueryAndUpdateIdentity = "authorized" | "anonymous" | "current";
...
+ * That means that the cache is typed and it should match the onLoad and onRequest parameters used in the queryAndUpdate.
+ *
+ * That also means that two calls to `createCachedQueryAndUpdate` even with the same type will return two different caches.
+ *
+ * No arguments needed. Instead, only the generic types are needed.
+ *
+ * @returns {(QueryAndUpdate<R, E>) => void} queryAndUpdate call
+ */
+export const createCachedQueryAndUpdate = <R, E>() => {
+  const cache = new ResponseCache<R>();
```
mstrasinskis, line 61: **Not a strong opinion**: to avoid some edge cases the solution could be more primitive, and store `strategy based` data, not only certified/uncertified.

frontend/src/lib/derived/is-logged-in.derived.ts
```diff
@@ -0,0 +1,7 @@
+import { authStore } from "$lib/stores/auth.store";
+import { isSignedIn } from "$lib/utils/auth.utils";
+import { derived } from "svelte/store";
+
+export const isLoggedInStore = derived(authStore, ($store) =>
+  isSignedIn($store.identity)
+);
```
dskloetd, line 7: Should this also be added to debug.derived.ts?

frontend/src/lib/pages/NnsProposalDetail.svelte
```diff
@@ -98,6 +99,9 @@
       });
 
       await findProposal();
+      if ($isLoggedInStore) {
+        await listNeurons({ resetCache: false });
```
dskloetd, line 103: I don't think you need to await here?
And is it important that listNeurons() is called only after findProposals returns?
Otherwise you can remove the await from findProposals as well.

frontend/src/lib/services/neurons.services.ts
```diff
@@ -254,15 +258,22 @@ export const stakeNeuron = async ({
  * @param {skipCheck} params.skipCheck it true, the neurons' balance won't be checked and those that are not synced won't be refreshed. It avoids possible infinite loops.
  * @param {callback} params.callback an optional callback that can be called when the data are successfully loaded (certified or not). Useful for example to close synchronously a busy spinner once all data have been fetched.
  */
```
dskloetd, line 260: This comment should apply to line 265, right?

frontend/src/lib/services/neurons.services.ts
```diff
@@ -254,15 +258,22 @@ export const stakeNeuron = async ({
...
+const queryAndUpdateNeuronsWithCache = createCachedQueryAndUpdate<
+  NeuronInfo[],
+  unknown
```
dskloetd, line 263: Maybe add a comment what `unknown` is passed to?
I was thinking just the name of the parameter would work, but it seems to be `E`?

frontend/src/lib/services/neurons.services.ts
```diff
@@ -254,15 +258,22 @@ export const stakeNeuron = async ({
...
+>();
 export const listNeurons = async ({
   callback,
   strategy = "query_and_update",
+  resetCache = true,
```
dskloetd, line 268: I don't think this should have a default value. Better be explicit about it every time.

frontend/src/lib/derived/is-logged-in.derived.ts
lmuntaner: no, only writable stores should go there. Derived stores are "duplicated" data

frontend/src/lib/pages/NnsProposalDetail.svelte
lmuntaner: Not really true, I don't need to.

frontend/src/lib/pages/NnsProposals.svelte
```diff
@@ -124,6 +126,9 @@
   );
 
   $: $authStore.identity, (() => proposalsFiltersStore.reload())();
+  $: $isLoggedInStore,
+    (async () =>
+      $isLoggedInStore && (await listNeurons({ resetCache: false })))();
```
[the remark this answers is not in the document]
lmuntaner: correct, I believe that syntax would also work, I like it.

frontend/src/lib/services/neurons.services.ts
lmuntaner: Yes, but when I created the variable `queryAndUpdateNeuronsWithCache` outside, I felt it was too far away and disconnected.

Do you think it's still better to do it above the comment?

frontend/src/lib/services/neurons.services.ts
lmuntaner: Correct, it's the error type.

I can add a comment, but the generics are kind of the function signature.

It would be nice to have named generics though. It would make them more readable

frontend/src/lib/services/neurons.services.ts
lmuntaner: It was to not touch current calls and add more changes.

But I like the idea of being explicit about the cache.

frontend/src/lib/services/utils.services.ts
lmuntaner: Do you mean to have the condition for when NOT to reset the listeners instead of when to do so?

In that case it would be `strategy === "query_and_update" && !certified`, right?

We need to reset the listeners at the end of each call for the strategy.

frontend/src/lib/derived/is-logged-in.derived.ts
```diff
@@ -0,0 +1,7 @@
+import { authStore } from "$lib/stores/auth.store";
+import { isSignedIn } from "$lib/utils/auth.utils";
+import { derived } from "svelte/store";
+
+export const isLoggedInStore = derived(authStore, ($store) =>
```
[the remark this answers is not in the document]
lmuntaner: done

frontend/src/lib/derived/is-logged-in.derived.ts
```diff
@@ -0,0 +1,7 @@
+import { authStore } from "$lib/stores/auth.store";
```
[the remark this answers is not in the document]
lmuntaner: Another day 😈

frontend/src/lib/pages/NnsProposalDetail.svelte
lmuntaner: I meant: "No, that's true, I don't need it", excuse for my english. I was going too fast.

frontend/src/lib/services/neurons.services.ts
dskloetd: I think so.

frontend/src/lib/services/neurons.services.ts
dskloetd: Right, so in the absence of named parameters, I would normally do something like this if the passed argument doesn't make it clear what the parameter is:
```
const queryAndUpdateNeuronsWithCache = createCachedQueryAndUpdate<
  NeuronInfo[],
  /* E = */ unknown
>
```

I just don't know if in this case `E` makes it clear or not.
But if not, I guess the parameter name itself should be improved rather than just the comment :).
So I guess it's been decided that `E` is clear enough.

frontend/src/lib/services/neurons.services.ts
dskloetd: You could have an separate PR that only adds the `resetCache = true` everywhere (which doesn't do anything but is true because without the existence of a cache it's as if it's always reset) if you don't want to mess up this PR with all those changes.

frontend/src/lib/services/utils.services.ts
mstrasinskis: True, my suggestion was wrong. It should be `certified || strategy ==! "query_and_update"`. But not a big win, so feel free to ignore.

mstrasinskis (approved):
LGTM, but it would be great if someone else could also review this.

frontend/src/lib/services/neurons.services.ts
lmuntaner: That looks good, specially with `unknown` generic types.

lmuntaner:
This design has a bug hard to solve nicely.

The bug is the following, if the user goes the neurons page, the list of cached, then the user goes to a neuron detail and for example starts dissolving the neuron. The neuron is changed in store, when the user comes back to the list, listNeurons uses the cache and shows the neuron that is NOT dissolving.

With the current pattern, there is no nice way to reset the cache after a neuron action.

Closing in favor of the design in #1802

PR #1781, opened 2023-01-30
feat: integrate ckBTC ledger and display main account

peterpeterparker:
# Motivation

Integrate ckBTC ledger, load related accounts balance and display main account.

# WARNING ⚠️

As for Snses, this PR introduces an insecure call to the back that is use to fetch the balance of the accounts and display such balance as information in the "universe token selector".

Upon selecting the ckBTC universe, the balance will be queried and computed with query+update calls - i.e. will be certified.

# Changes

- provide bare minimal script and docs to develop locally
- extract ICRC functionality that loads accounts from Sns to `icrc-ledger.api`
- refactor Sns to use above function and integrates it with `ckbtc-ledger.api` as well
- rename few components and stores from "project" to "universe"
- display ckBTC in token selector with static information (text and logo hosting in NNS-dapp)
- add new dedicated accounts' stores for ckBTC (accounts and accounts balance)
- add component `<CkBTCAccounts />` for the "Accounts" grid view
- refactor `goToWallet` utils to avoid duplicated code in Nns, Sns and ckBTC pages

# Screenshots

Capture d’écran 2023-01-30 à 16 27 41


HACKING.md
```diff
@@ -44,3 +49,57 @@ The `dfx` version installed locally should match the one defined in [dfx.json](h
...
+
+> Last update: 2023, Jan. 30
+
+ckBTC deployment and development in NNS-dapp are in progress.
+
+### Bitcoin network
+
+To deploy a local Bitcoin network, the [documentation](https://internetcomputer.org/docs/current/developer-docs/integrations/bitcoin/local-development) on the IC website can be followed.
+
+As ckBTC is not yet implemented in our pipeline, enabling it in `dfx.json` has bot be yet configured - i.e. needs to be manually configured:
```
lmuntaner, line 63: typo: "has bot be yet configured"

frontend/src/lib/components/universe/UniverseLogo.svelte
```diff
@@ -21,7 +20,6 @@
   let ckBTC = false;
   $: ckBTC = isUniverseCkBTC(canisterId);
 
-  // TODO: use ckBTC logo provided by ledger
```
lmuntaner, line 24: Why remove this TODO? That you are not doing it in this PR doesn't mean we will not do it in the future.

frontend/src/lib/routes/Accounts.svelte
```diff
@@ -37,16 +42,33 @@
...
     await uncertifiedLoadSnsAccountsBalances({
       rootCanisterIds: projects.map(({ rootCanisterId }) => rootCanisterId),
-      excludeRootCanisterIds: [projectIdSelected.toText()],
+      excludeRootCanisterIds: [selectedUniverseId.toText()],
     });
   };
 
-  $: (async () => await loadSnsAccountsBalances($snsProjectsCommittedStore))();
+  const loadCkBTCAccountsBalances = async () => {
+    // We load the accounts balance of ckBTC only if not already triggered by current view
```
lmuntaner, line 50: Isn't the `routes/Accounts` the current view?

Do you mean that if we are in the ckBTC accounts, they will be loaded already with query and update call in `pages/CkBTCAccounts.svelte`?

lmuntaner (approved):
LGTM, minor comments. Approving already.

HACKING.md
peterpeterparker: Thanks

frontend/src/lib/components/universe/UniverseLogo.svelte
peterpeterparker: Not in this PR, not in a shorterm future in my opinion unless required. I tried to implement it but, it would need quite some development to convert the "universe selector" to a stricly "token selector" for the accounts views and this in a clean way.

As the UX "selector / dashboard etc." will change quite in the future too and, as it will need refinement too, I informed product team that I think it is not worth the effort right now.

Therefore I removed the TODO because anyway, it probably not gonna be implemented there. Likewise, I did not open a task because I think it will not be solve specialy but rather integrated in that new UX concept.

Of course, like I said above, if it suddently becomes required, then I'll do.

Hope this make sense to you too?

frontend/src/lib/routes/Accounts.svelte
peterpeterparker: > Do you mean that if we are in the ckBTC accounts, they will be loaded already with query and update call in pages/CkBTCAccounts.svelte?

Yes. Improved the comment

frontend/src/lib/components/universe/UniverseLogo.svelte
lmuntaner: That makes sense, but I'm not sure I like the direction of the new features. I wish the requirements took this into account. We try to rely less and less in the backends, that's not a good omen.

PR #1788, opened 2023-01-31
GIX-1187: New Api Service Layer for NNS Neurons

lmuntaner:
# Motivation

We want to reuse certified data instead of doing many update calls to CI.

# Changes

* New entity: Api Service. For now only NnsNeuronsApiService.
* New folder for these entities api-services.
* New api service: NnsNeuronsApiService
* Use new api service in `listNeurons` from neuron services.
* Add temporary `useCache` flag to `listNeurons`. Default to `false`.
* Remove `listNeurons` from initial call in app.services. Instead, added call in Proposals and Neurons pages.

# Tests

PENDING to approve this new pattern.

### DIFF
--- a/frontend/src/lib/api-services/nns-neurons.api-services.ts
+++ b/frontend/src/lib/api-services/nns-neurons.api-services.ts
@@ -1,0 +1,164 @@
+import { queryNeurons, spawnNeuron } from "$lib/api/governance.api";
+import { CACHE_EXPIRATION_SECONDS } from "$lib/constants/api-services.constants";
+import { nowInSeconds } from "$lib/utils/date.utils";
+import { nonNullish } from "$lib/utils/utils";
+import type { Identity } from "@dfinity/agent";
+import type { NeuronId, NeuronInfo } from "@dfinity/nns";
+import type { Principal } from "@dfinity/principal";
+
+type CallParams = {
+  identity: Identity;
+};
+
+type QueryParmas = {
+  certified: boolean;
+} & CallParams;
+
+/**
+ * Caches two types:
+ * - Neurons
+ * - List of neuron ids for a given principal. This is needed to know whether we have ALL neurons for a given principal.
+ */
+type NnsNeuronsCache = {
+  // Store all neurons.
+  neurons: {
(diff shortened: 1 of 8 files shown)


frontend/src/lib/api-services/nns-neurons.api-services.ts
```diff
@@ -0,0 +1,164 @@
...
+} & CallParams;
+
+/**
+ * Caches two types:
+ * - Neurons
+ * - List of neuron ids for a given principal. This is needed to know whether we have ALL neurons for a given principal.
+ */
+type NnsNeuronsCache = {
+  // Store all neurons.
+  neurons: {
```
lmuntaner, line 24: I wanted to avoid storing both certified and uncertified data.

But it seems that I won't be ale to do that.

Otherwise, how do I know that a query after an update should or should not update certified data?

After merging neurons, the query call should substitute the list.

But if a query comes after an update call (for whatever weird reason), it should not change the caching.

I can rely that the cache should always be reset after an action. But it seems dangerous to me to rely on that.

Ideas are welcome.

frontend/src/lib/api-services/nns-neurons.api-services.ts
```diff
@@ -0,0 +1,180 @@
...
+    const neuronIds = this.data.list[principal.toText()];
+    if (
+      neuronIds?.ids !== undefined &&
+      (neuronIds.certified === certified || neuronIds.certified)
+    ) {
+      const neurons = neuronIds.ids
+        .map((neuronId) => this.data.neurons[String(neuronId)])
+        .filter(nonNullish)
+        .filter(({ certified: c }) => c === certified || c)
+        .filter(({ timestampSeconds }) => !this.isExpired(timestampSeconds))
```
mstrasinskis, line 72: clean if isExpired?

lmuntaner: Yes, and leave the ones that have not expired.

lmuntaner:
This was a draft to discuss the approach. Closing...

PR #1789, opened 2023-01-31
Set a default target

bitdivine:
# Motivation
Running `cargo build` by itself yields errors. Specifying the target makes this work.  However giving the target manually every time gets boring.  So set the default target.

Note: For production builds it is probably better to be very explicit about the target, so I have not removed the target specifier there.  This will be useful primarily for developers working on Rust who want to build without creating the final wasm.

# Changes
- Specify a default build target in the cargo config.

# Tests
- [x] Verify that `cargo build` runs without error.
   - Needs `touch assets.tar.xz && cargo build` but then, yes.


mstrasinskis (approved):
Rubber stamp, thanks!

bitdivine:
cargo-tests runs forever.  Is that due to something in this PR?  I see no obvious cluprit.

bitdivine:
Works well locally but this causes lots f recompilation.  What am I missing?  Not a focus for now so closing, maybe to revisit another time.

PR #1791, opened 2023-01-31
Make an nnsdapp entry in dfx.json

bitdivine:
# Motivation
When deploying infrastructure, it's helpful to use the same name for testnets everywhere.  In the ic-repo, where the static testnets are defined, our testnet is called "nnsdapp" so say the snsdemo is deployed to that name.  Alternative names such as staging and testnet can stay but "nnsdapp" is really helpful when deploying the whole stack.

# Changes
- Add nnsdapp to dfx.json

# Tests
- ~will it blend~ will it deploy?


lmuntaner (approved):
🤞

PR #1792, opened 2023-01-31
feat: ckbtc index and wallet page

peterpeterparker:
# Motivation

wip

# Changes

wip

# Tests

wip

### DIFF
--- a/frontend/src/lib/api/ckbtc-index.api.ts
+++ b/frontend/src/lib/api/ckbtc-index.api.ts
@@ -1,0 +1,56 @@
+import { createAgent } from "$lib/api/agent.api";
+import type {
+  GetTransactionsParams,
+  GetTransactionsResponse,
+} from "$lib/api/icrc-index.api";
+import { getTransactions as getIcrcTransactions } from "$lib/api/icrc-index.api";
+import { CKBTC_INDEX_CANISTER_ID } from "$lib/constants/canister-ids.constants";
+import { HOST } from "$lib/constants/environment.constants";
+import { logWithTimestamp } from "$lib/utils/dev.utils";
+import type { HttpAgent, Identity } from "@dfinity/agent";
+import { IcrcIndexCanister } from "@dfinity/ledger";
+
+export const getCkBTCTransactions = async ({
+  identity,
+  ...rest
+}: Omit<GetTransactionsParams, 'getTransactions'>): Promise<GetTransactionsResponse> => {
+  logWithTimestamp("Getting ckBTC accounts transactions: call...");
+
+  const {
+    canister: { getTransactions },
+  } = await ckBTCCanister({ identity });
+
+  const results = await getIcrcTransactions({
+    identity,
+    ...rest,
+    getTransactions,
+  });
+
+  logWithTimestamp("Getting ckBTC account transactions: done");
+
+  return results;
+};
+
+const ckBTCCanister = async ({
+  identity,
+}: {
+  identity: Identity;
+}): Promise<{
+  canister: IcrcIndexCanister;
+  agent: HttpAgent;
+}> => {
+  const agent = await createAgent({
+    identity,
+    host: HOST,
+  });
+
+  const canister = IcrcIndexCanister.create({
+    agent,
+    canisterId: CKBTC_INDEX_CANISTER_ID,
+  });
+
+  return {
+    canister,
+    agent,
+  };
+};
--- a/frontend/src/lib/api/ckbtc-index.api.ts
+++ b/frontend/src/lib/api/ckbtc-index.api.ts
@@ -13,7 +13,10 @@ import { IcrcIndexCanister } from "@dfinity/ledger";
 export const getCkBTCTransactions = async ({
   identity,
   ...rest
-}: Omit<GetTransactionsParams, 'getTransactions'>): Promise<GetTransactionsResponse> => {
+}: Omit<
+  GetTransactionsParams,
+  "getTransactions"
+>): Promise<GetTransactionsResponse> => {
   logWithTimestamp("Getting ckBTC accounts transactions: call...");
 
   const {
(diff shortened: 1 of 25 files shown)


frontend/src/lib/api/ckbtc-index.api.ts
```diff
@@ -0,0 +1,59 @@
...
+    ...rest,
+    getTransactions,
+  });
+
+  logWithTimestamp("Getting ckBTC account transactions: done");
+
+  return results;
+};
+
+const ckBTCCanister = async ({
```
lmuntaner, line 37: This is more the `ckBTCIndexCanister` right?

But I don't see it being used anywhere, right?

frontend/src/lib/api/ckbtc-index.api.ts
```diff
@@ -0,0 +1,59 @@
...
+  ...rest
+}: Omit<
+  GetTransactionsParams,
+  "getTransactions"
+>): Promise<GetTransactionsResponse> => {
+  logWithTimestamp("Getting ckBTC accounts transactions: call...");
+
+  const {
+    canister: { getTransactions },
+  } = await ckBTCCanister({ identity });
```
lmuntaner, line 24: Is this needed if you use the `getIcrcTransactions`?

frontend/src/lib/api/icrc-index.api.ts
```diff
@@ -0,0 +1,40 @@
...
+  oldestTxId?: bigint;
+}
+
+export const getTransactions = async ({
+  maxResults: max_results,
+  start,
+  account,
+  getTransactions,
+}: GetTransactionsParams): Promise<GetTransactionsResponse> => {
+  const { oldest_tx_id, ...rest } = await getTransactions({
```
lmuntaner, line 30: Is this a function that calls itself?

frontend/src/lib/api/icrc-index.api.ts
```diff
@@ -0,0 +1,40 @@
...
+    params: GetAccountTransactionsParams
+  ) => Promise<IcrcGetTransactions>;
+}
+
+export interface GetTransactionsResponse
+  extends Omit<IcrcGetTransactions, "oldest_tx_id"> {
+  oldestTxId?: bigint;
+}
+
+export const getTransactions = async ({
```
lmuntaner, line 24: Shouldn't this expect a canister id and then create the canister service and call the method to get transactions?

Maybe remove and use it for SNS in another PR?

frontend/src/lib/api/sns-index.api.ts
```diff
@@ -1,46 +1,37 @@
...
-  start,
-  maxResults,
   rootCanisterId,
-}: GetTransactionsParams): Promise<GetTransactionsResponse> => {
+  ...rest
+}: GetSnsTransactionsParams): Promise<GetTransactionsResponse> => {
   logWithTimestamp("Getting sns accounts transactions: call...");
-  const { getTransactions: getTransactionsApi } = await wrapper({
+
+  const { getTransactions } = await wrapper({
```
lmuntaner, line 22: Are you using this `getIcrcTransactions`?

lmuntaner (commented):
A few comments.

Can you check that the SNS wallet loads transactions?

peterpeterparker:
> Can you check that the SNS wallet loads transactions?
I did

frontend/src/lib/api/ckbtc-index.api.ts
peterpeterparker: Sure, renamed

frontend/src/lib/api/ckbtc-index.api.ts
peterpeterparker: Yes because Sns gets the canisters to query through the wrapper

frontend/src/lib/api/icrc-index.api.ts
peterpeterparker: No, it's the function passed as param. Assigned an alternative name to it to make this more obvious.

frontend/src/lib/api/icrc-index.api.ts
peterpeterparker: Ditto, Sns needs wrapper

frontend/src/lib/api/sns-index.api.ts
peterpeterparker: Yes

```
const { getTransactions } = await wrapper({
    identity,
    certified: true,
    rootCanisterId: rootCanisterId.toText(),
  });

  const results = await getIcrcTransactions({ <---- Icrc
    identity,
    ...rest,
    getTransactions, <--- function to call
  });
```

frontend/src/lib/api/icrc-index.api.ts
```diff
@@ -0,0 +1,40 @@
...
+  IcrcGetTransactions,
+} from "@dfinity/ledger";
+import { fromNullable } from "@dfinity/utils";
+
+export interface GetTransactionsParams {
+  identity: Identity;
+  account: IcrcAccount;
+  start?: bigint;
+  maxResults: bigint;
+  getTransactions: (
```
lmuntaner, line 14: I didn't realize it was here, this is how we do it in the `icrc-ledger.api.ts`.

I'm sorry I'm bringing this up now, but why bother even with this `icrc-x.api.ts`? We already share the interface from ledger.js, this extra file seems too much.

But maybe we can refactor once we have everything.

lmuntaner (approved):

frontend/src/lib/api/icrc-index.api.ts
peterpeterparker: The design is required. We can have a call if you wish for explanation.

Long story short:

- we cannot use only `icrc-api` because the Sns are fetched with the wrapper while ckBTC is a single canister
- we need the `icrc-api` because it avoids the duplication of code and logic - e.g. in the particular case of the ledger if avoid to duplicate the mapping of the metadata and balance to accounts

lmuntaner: I see the reason, but something doesn't resonate well with this design. I don't have a better solution for now, so it's ok.

peterpeterparker: Good then. I think this design works well. Thanks for the review.

PR #1796, opened 2023-02-01
One source of truth for wasm path

bitdivine:
# Motivation
The location of a generated wasm was specified in both `dfx.json` and `build-rs.sh`.  If we have a single source of truth, there is less scope for error when wasms are renamed (such as the `sns_aggregator` nee `sns_cache`).

# Changes
* `build-rs.sh` now gets the output path from dfx.json

# Tests
Existing tests should suffice.


dskloetd:
Can you point out to me where in `build-rs.sh` and where in `dfx.json` did it use to specify the location of `sns_aggregator`?
I don't see it specified in either so I must be missing something about the point of this change.

bitdivine:
@dskloetd  `sns_aggregator` is not in main yet, at least the vast bulk of the code is still on a feature branch.  That is why you didn't see it, and also didn't experience any pain from moving.

On the feature branch, the dfx entry is here: https://github.com/dfinity/nns-dapp/blob/sns_cache/dfx.json#L159

dskloetd:
I asked sns_aggregator because that was the example you gave. But I don't see the location of any canister specified in build-rs.sh. Can you clarify what exactly is the "second source of truth" that this PR is removing?

build-rs.sh
```diff
@@ -34,8 +34,10 @@ cargo build "${cargo_args[@]}"
 # ic-cdk-optimizer # (output: ${canister_name}.wasm)
 ####################
 echo Optimising wasm
-ic-cdk-optimizer "./target/wasm32-unknown-unknown/release/${canister_name}.wasm" -o "./${canister_name}.wasm"
-gzip -f -n "${canister_name}.wasm"
-mv "${canister_name}.wasm.gz" "${canister_name}.wasm"
-ls -sh "./${canister_name}.wasm"
-sha256sum "./${canister_name}.wasm"
+wasm_path="$(canister_name="$canister_name" jq -r '.canisters[env.canister_name].wasm' dfx.json)"
```
dskloetd, line 37: Would it make sense to have a dfx_json_lookup script to hide the jq syntax?
I'm assuming that could be used in more places.

dskloetd (approved):

build-rs.sh
bitdivine: A subcommand could be added to dfx, perhaps, similar to `dfx canister info`.  Any wider use isn't limited to our team, I'm pretty sure.  We have jq installed and use it expensively.  Another project might not and would benefit from such a `dfx` command more than we.

PR #1798, opened 2023-02-01
Fix setup

bitdivine:
# Motivation
Two bugs found in setup on a new machine:

- jq is used when getting the dfx version, but before jq is installed.
- curl is used but not necessarily already installed.

There were some other niggles in my notes but these two are straightforward.

# Changes
- Move reading the dfx version only after installing jq 
- Install curl if not already present

# Tests


scripts/setup
```diff
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
 cd "$(dirname "$0")/.."
-. <(./build-config.sh)
```
bitdivine, line 4: Reading this config requires jq, but jq is only installed later.

scripts/setup
```diff
@@ -184,6 +183,13 @@ install_wget_linux() {
   sudo apt-get install wget
 }
 
+install_curl_linux() {
+  sudo apt-get install curl
+}
+install_curl_darwin() {
+  brew install curl
+}
```
bitdivine, line 191: curl is expected but may not be installed.

lmuntaner (approved):
Thanks!

PR #1800, opened 2023-02-01
Aggregator config

bitdivine:
# Motivation
The location of an aggregator canister needs to be passed to the frontend.

# Changes
- Add the aggregator to `dfx.json`
- Provide the URL to the aggregator in the config, if there is one, else leave the location blank.

# Tests
When there is no aggregator, config has these aggregator entries:
```
$ DFX_NETWORK=local ./config.sh | grep AGG
VITE_FEATURE_FLAGS="{\"ENABLE_CKBTC_LEDGER\":true,\"ENABLE_SNS_2\":true,\"ENABLE_SNS_AGGREGATOR\":false,\"ENABLE_SNS_VOTING\":true}"
VITE_AGGREGATOR_CANISTER_URL=
```
After creating the aggregator canister:
```
$ dfx canister create sns_aggregator
Creating canister sns_aggregator...
sns_aggregator canister created with canister id: rrkah-fqaaa-aaaaa-aaaaq-cai

$ DFX_NETWORK=local ./config.sh | grep AGG
VITE_FEATURE_FLAGS="{\"ENABLE_CKBTC_LEDGER\":true,\"ENABLE_SNS_2\":true,\"ENABLE_SNS_AGGREGATOR\":false,\"ENABLE_SNS_VOTING\":true}"
VITE_AGGREGATOR_CANISTER_URL=http://rrkah-fqaaa-aaaaa-aaaaq-cai.localhost:8080
```


config.sh
```diff
@@ -99,6 +111,7 @@ fetchRootKey=$(echo "$json" | jq -r ".FETCH_ROOT_KEY")
 featureFlags=$(echo "$json" | jq -r ".FEATURE_FLAGS" | jq tostring)
 host=$(echo "$json" | jq -r ".HOST")
 identityServiceUrl=$(echo "$json" | jq -r ".IDENTITY_SERVICE_URL")
+aggregatorCanisterUrl=$(echo "$json" | jq -r '.SNS_AGGREGATOR_URL // ""')
```
bitdivine, line 114: Note: In jq, `x // y` means "use x if defined, else use y"

lmuntaner: interesting! in Typescript is `??`

lmuntaner (approved):
Thanks!

PR #1801, opened 2023-02-01
Simplify createDerivedStore in debug.derived.ts

dskloetd:
# Motivation

Small optimization.
If all we want is a Readable view on a writable store, that doesn't expose `set` and `update`, creating a derived store is overkill.
The derived store subscribes to the original and forwards the changes but this is not necessary.

# Changes

Change createDerivedStore to simply return an object with the same subscribe method and no other methods.

# Tests

Refactoring should be covered by existing tests.

### DIFF
--- a/frontend/src/lib/derived/debug.derived.ts
+++ b/frontend/src/lib/derived/debug.derived.ts
@@ -31,13 +31,13 @@ import {
   type Writable,
 } from "svelte/store";
 
-// TODO(dskloetd): Rename to createReadableStore.
-const createDerivedStore = <T>(store: Writable<T>): Readable<T> =>
-  { subscribe: store.subscribe };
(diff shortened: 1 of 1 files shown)


lmuntaner (commented):
We don't have tests for this feature.

Did you try that it worked by:
* Go to Wallet page.
* Click six times in the header.
* Enter "fo" in the prompt.
* Does the JSON file have the data of the "selectedAccount"?

dskloetd:
Actually, I forgot to bind the method to the object so I expect it won't work.
I'll pause this until I can add a test that would have caught it.

anedos-dfinity:
Should this PR be closed ?

lmuntaner:
@dskloetd what do you think?

dskloetd:
Does it matter? It's marked as Draft already.

PR #1802, opened 2023-02-02
GIX-1187: Api Service Layer Setup

lmuntaner:
# Motivation

Setup the api service layer for neurons. This will allow us to use caching in upcoming PRs.

# Changes

* New neuronsApiService object that is used in neuron.services and uses governance.api.
* The neuron services do not use the governance api anymore.
* Create types for all api function signatures to be used in neuronsApiService.

The idea is the following:

```
// Move from:
neuron service --> governance api
// To:
neuron service --> neuron api service (can provide caching) --> governance api
```

# Tests

* Change tests in neuron.services.
* Add test for all neuron api services methods.

### DIFF
--- a/frontend/src/lib/api-services/neurons.api-service.ts
+++ b/frontend/src/lib/api-services/neurons.api-service.ts
@@ -20,10 +20,10 @@ import {
   startDissolving,
   stopDissolving,
   type ApiAutoStakeMaturityParams,
-  type ApiCallNeuronParams,
   type ApiDisburseParams,
-  type ApiHotkeyCallParams,
   type ApiIncreaseDissolveDelayParams,
+  type ApiManageHotkeyParams,
+  type ApiManageNeuronParams,
   type ApiMergeMaturityParams,
   type ApiMergeNeuronsParams,
   type ApiQueryNeuronParams,
@@ -37,24 +37,24 @@ import {
 
 export const neuronsApiService = {
   // Read calls
+  queryKnownNeurons(params: ApiQueryParams) {
+    return queryKnownNeurons(params);
+  },
   queryNeuron(params: ApiQueryNeuronParams) {
     return queryNeuron(params);
   },
   queryNeurons(params: ApiQueryParams) {
     return queryNeurons(params);
   },
-  queryKnownNeurons(params: ApiQueryParams) {
-    return queryKnownNeurons(params);
-  },
 
   // Action calls
-  addHotkey(params: ApiHotkeyCallParams) {
+  addHotkey(params: ApiManageHotkeyParams) {
     return addHotkey(params);
   },
   autoStakeMaturity(params: ApiAutoStakeMaturityParams) {
     return autoStakeMaturity(params);
   },
-  claimOrRefreshNeuron(params: ApiCallNeuronParams) {
+  claimOrRefreshNeuron(params: ApiManageNeuronParams) {
     return claimOrRefreshNeuron(params);
   },
   disburse(params: ApiDisburseParams) {
@@ -63,10 +63,10 @@ export const neuronsApiService = {
   increaseDissolveDelay(params: ApiIncreaseDissolveDelayParams) {
     return increaseDissolveDelay(params);
   },
-  joinCommunityFund(params: ApiCallNeuronParams) {
+  joinCommunityFund(params: ApiManageNeuronParams) {
     return joinCommunityFund(params);
   },
-  leaveCommunityFund(params: ApiCallNeuronParams) {
+  leaveCommunityFund(params: ApiManageNeuronParams) {
     return leaveCommunityFund(params);
   },
   // @deprecated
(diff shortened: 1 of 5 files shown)


dskloetd:
> object that use used

object that *is* used?

lmuntaner:
> > object that use used
> 
> object that _is_ used?

Yes, thanks!

frontend/src/lib/api/governance.api.ts
```diff
@@ -203,15 +209,15 @@ export const spawnNeuron = async ({
   return newNeuronId;
 };
 
+export type ApiHotkeyCallParams = ApiCallNeuronParams & {
```
dskloetd, line 212: Some param types have "Call" before "Params". Some have "Call" after "Api" and some don't have "Call".
What does all this mean?

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export const neuronsApiService = {
+  // Read calls
+  queryNeuron(params: ApiQueryNeuronParams) {
+    return queryNeuron(params);
+  },
+  queryNeurons(params: ApiQueryParams) {
+    return queryNeurons(params);
+  },
+  queryKnownNeurons(params: ApiQueryParams) {
```
dskloetd, line 46: The functions below appear to be in alphabetical order. Do you want these to be as well?
'K' comes before 'N'.

frontend/src/tests/lib/api-services/neurons.api-service.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+  // Action calls
+  describe("addHotkey", () => {
+    it("should call addHotkey api", () => {
+      const params = {
+        neuronId,
+        identity: mockIdentity,
+        principal: mockPrincipal,
+      };
+      neuronsApiService.addHotkey(params);
+      expect(addHotkey).toHaveBeenCalledWith(params);
```
dskloetd, line 86: Does this expect exactly 1 call or any number of calls?

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -1,4 +1,4 @@
-import * as api from "$lib/api/governance.api";
+import { neuronsApiService } from "$lib/api-services/neurons.api-service";
```
dskloetd, line 1: What do you think about using the real neuron-api-service and expecting the same calls to the api as before?

dskloetd (changes requested):

frontend/src/lib/api-services/neurons.api-service.ts
lmuntaner: Yes, thanks!

frontend/src/lib/api/governance.api.ts
lmuntaner: Good catch, this should be `ApiHotkeyParams`.

I'll add a comment into what each means, I forgot that.

frontend/src/tests/lib/api-services/neurons.api-service.spec.ts
lmuntaner: One call api service call should be one api call. I'll add the test, thanks!

`toHaveBeenCalledWith` does not check the amount of calls

frontend/src/tests/lib/services/neurons.services.spec.ts
lmuntaner: When we add caching that won't be true.

This is more a unit test of the service. If we check the api calls, then we are testing the caching functionality at this level.

For unit tests I was thinking in testing the functionality at the api-service layer.

frontend/src/tests/lib/api-services/neurons.api-service.spec.ts
lmuntaner: done

frontend/src/lib/api/governance.api.ts
```diff
@@ -18,15 +18,38 @@ import { GovernanceCanister } from "@dfinity/nns";
 import type { Principal } from "@dfinity/principal";
 import { ledgerCanister as getLedgerCanister } from "./ledger.api";
 
+/**
+ * COMMON TYPES
```
lmuntaner, line 22: @dskloetd I added some comments on the type naming convention

lmuntaner:

frontend/src/tests/lib/services/neurons.services.spec.ts
dskloetd: Yes, it won't be true when we do caching, but then we change the tests.
In general you don't want to use mocks if you don't have to.
I think it might actually be easier to test the cache in terms of which service calls should result in which API calls than to split it.

lmuntaner: I see the value of that but we also don't test queryAndUpdate functionality from the service perspective.

Where do we test things together and where don't? How do we know what's tested where then?

I would be a little lost if there is no objective or easy way to know what types of testing we want. That also means slower PRs because it's not clear to me what is wanted by the team.

I'm not against this rule or the other, but I like objective rules as much as possible. To reduce comments and long PRs because the pattern chosen doesn't align to the project's.

Can you come up with a generic rule for this?

Or is, in general unit tests, except services where we mock the api instead of the intermediate layer?

PR #1804, opened 2023-02-02
feat: token store

peterpeterparker:
# Motivation

To integrate ckBTC transactions we need the requested ledger fee.

Currently in Sns, such fees are fetched with a dedicated `transferFee` which generate unecessary request on the backend as the fees are already provided and fetched with the `metadata` call.

Ultimately such Sns requests will be deprecated but, we need to integrate ckBTC now. In addition, as we are concern about performances, we want to avoid to implement uncessery calls for ckBTC as well.

That's why we should not query `metadata` and `balance` to load and maps our "Account" construct but, perform only what's necessary.

Therefore, this PR introduces a new `tokensStore`, adds few `TODOs` and set `deprecation` information which shall be implemented in upcoming PRs.

# Breaking Changes

As for `name` and `symbol`, `fee` becomes a mandatory field of the Icrc metadata. Sns-1 and ckBTC both provides such metadata information and FI team confirmed that an Icrc ledger metadata fee is in sync with the exposed endpoint that returns the fee.

# Changes

- mark `account.balance` as **deprecated** and to be replaced with `account.balanceE8s`
- mark `metadata` function of `icrc-ledger.api` as **deprecated**
- mark `transaction-fees.store` as **deprecated**
- mark `snsTokenSymbolSelectedStore` as **deprecated**
- introduces new `tokenStore` and related derived store to support universes with Nns
- add `getIcrcToken` to `icrc-ledger` api and usage in `sns/ckBTC-api`
- load token information in services
  - sns uncertified query account balance
  - icrc uncertified query account balance
  - `syncSnsAccounts`
  - `syncIcrcAccounts`
  - `loadSnsProjects`
- replace `loadIcrcAccounts` in cmp with `syncIcrcAccounts`
- use new `universesTokenStore` in `<UniverseAccountsBalance />`
- refactor `SnsTokenMetadata` to `IcrcTokenMetadata`
- add `fee` to `IcrcTokenMetadata`
- move `mapOptionalToken` to `icrc-tokens.utils`


frontend/src/lib/types/icrc.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Token metadata are to some extension optional and provided in Candid in a way the frontend cannot really use.
+ * That's why we have to map the data as well.
+ */
+export interface IcrcTokenMetadata {
```
lmuntaner, line 5: For NNS we can still use it, right? We'll hard code it, but we shuold use also the tokenStore, right?

frontend/src/lib/types/icrc.ts
```diff
@@ -0,0 +1,9 @@
...
+  name: string;
+  symbol: string;
+  fee: bigint;
```
lmuntaner, line 8: They also return:

```
  record {
    "icrc1:decimals";
    variant {
      Nat = 8
    }
  };
```

This could be useful as well. At the moment "8" decimals is hardcoded. Maybe we should add it to this data?

Not for this PR, but maybe a todo?

frontend/src/lib/api/icrc-ledger.api.ts
```diff
@@ -45,3 +51,21 @@ export const getIcrcMainAccount = async ({
     type: "main",
   };
 };
+
+export const getIcrcToken = async ({
+  certified,
+  metadata: ledgerMetadata,
```
lmuntaner, line 57: Could we call this `getMetadata`?

The name `metadata` seems an actual metadata object to me.

I also realize that for `balance` we also call it just that. Anyway, not a strong opinion. In case you were doing it also to be consistent and like it better with `get` or `query`...

frontend/src/lib/derived/universes-tokens.derived.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const NNS_TOKEN: TokensStoreUniverseData = {
+  token: {
+    ...ICPToken,
+    fee: BigInt(DEFAULT_TRANSACTION_FEE_E8S),
+  },
+  certified: true,
+};
+
+export const universesTokensStore = derived<[TokensStore], TokensStoreData>(
```
lmuntaner, line 23: Instead of hardcoding a value in a derived store, why don't you add this token in init?

This way, the NNS data is also in the `tokensStore`

frontend/src/lib/services/ckbtc-accounts-balance.services.ts
```diff
@@ -20,9 +16,50 @@ export const uncertifiedLoadCkBTCAccountsBalance = (): Promise<void> => {
...
+    onError: ({ error: err }) => {
+      console.error(err);
+      throw err;
+    },
     logMessage: "Syncing ckBTC Accounts Balance",
     strategy: "query",
   });
 };
+
+const loadCkBTCToken = (): Promise<void> => {
```
lmuntaner, line 28: What the difference between this `loadCkBTCToken` and the one in `ckbtc-token.services.ts`?

lmuntaner (commented):
A few comments and suggestions

frontend/src/lib/types/icrc.ts
peterpeterparker: I think so. Actually the Nns / IC ledger since last week or so now supports also Icrc1. So, actually, we could even go big an begin to use Icrc1 for the Nns / IC ledger too. It's quite fresh, so did not tried it yet but would be quite cool I think. In another PRs 😉.

frontend/src/lib/types/icrc.ts
peterpeterparker: Sure, I added a todo as following:

```
export interface IcrcTokenMetadata {
  name: string;
  symbol: string;
  fee: bigint;
  // TODO: integrate "decimals" to replace ICP_DISPLAYED_DECIMALS_DETAILED
}
```

frontend/src/lib/derived/universes-tokens.derived.ts
peterpeterparker: I thought about it too actually but ultimately, thought it was "less ugly" to derive and add a constant value rather than pushing a value in store because it implies that each `set` and `update` should be double checked to preserve the constant. By having it derived, it's always there. Makes sense? Or better argument to push it in store?

No strong opinion, like I said more "it feels less ugly" thing.

frontend/src/lib/services/ckbtc-accounts-balance.services.ts
peterpeterparker: This one performs an insecure `query` only and does not toast the error but throw it so that all errors are collected and only one toast it displayed.

The function is not documented because it is not exported and the main exported function - `uncertifiedLoadCkBTCAccountsBalance` - is documented.

frontend/src/lib/api/icrc-ledger.api.ts
peterpeterparker: Sure, done.

frontend/src/lib/types/icrc.ts
lmuntaner: Yes, and we would need to wait for the support for hardware wallet as well.

frontend/src/lib/derived/universes-tokens.derived.ts
lmuntaner: I feel that the store should not care where the information comes from.

It might come from an API, like this, or from hardcoded data. The store doesn't care and treats it all the same.

And if in the future we use ICRC-1 for ICP, it will be ready to make the call and add the data.

frontend/src/lib/services/ckbtc-accounts-balance.services.ts
lmuntaner: Maybe add a comment so that we remember why? I found it confusing.

peterpeterparker: Sure

frontend/src/lib/derived/universes-tokens.derived.ts
peterpeterparker: Implement felt a bit weird but, I can live with it so done.

lmuntaner (approved):
LGTM!

PR #1806, opened 2023-02-02
Fix comment about the output of build-frontend.sh.

dskloetd:
# Motivation

A comment in build.sh says the output of build-frontend.sh is frontend/public/ but its actual output is assets.tar.xz.

# Changes

Change the comment.

# Tests

none


bitdivine (approved):
Thanks!

PR #1807, opened 2023-02-02
Make build-frontend.sh more user friendly.

dskloetd:
# Motivation

Currently when you run `build-frontend.sh` it runs for about half a minute before failing with
```
file:///Users/dskloet/dev/nns-dapp/clean/frontend/.svelte-kit/output/server/chunks/dapp.js:285
  {}.VITE_FEATURE_FLAGS.replace(/\\"/g, '"') ?? '{"ENABLE_SNS_2":false,"ENABLE_SNS_VOTING": false, "ENABLE_SNS_AGGREGATOR": false, "ENABLE_CKBTC_LEDGER": false}'
                        ^

TypeError: Cannot read properties of undefined (reading 'replace')
    at file:///Users/dskloet/dev/nns-dapp/clean/frontend/.svelte-kit/output/server/chunks/dapp.js:285:25
    at ModuleJob.run (node:internal/modules/esm/module_job:194:25)

Node.js v18.13.0
[vite-plugin-sveltekit-compile] Prerendering failed with code 1
error during build:
Error: Prerendering failed with code 1
    at ChildProcess.<anonymous> (file:///Users/dskloet/dev/nns-dapp/clean/frontend/node_modules/@sveltejs/kit/src/exports/vite/index.js:551:15)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess._handle.onexit (node:internal/child_process:291:12)
```

This is because you need to run `. config.sh` first.

`build-frontend.sh` was extract from `build.sh` which runs `. config.sh` after checking that DFX_NETWORK is set correctly.

It would be nice if `build-frontend.sh` is as user friendly to run as `build.sh`.

# Changes

1. Extra the code from `build.sh` that verifies that `DFX_NETWORK` is set correctly and reuse it in `build-frontend.sh`.
2. Run `. config.sh` from `build-frontend.sh`
3. Remove exports of variables already exported in `config.sh`. config.sh is sourced so its exports are also exports in build.sh.

# Tests

Ran manually.


bitdivine (approved):
LGTM.  Thanks!

PR #1808, opened 2023-02-02
GIX-1267: Use "query" not signed in users

lmuntaner:
# Motivation

Not logged in users see uncertified data.

This reduces the load of update calls.

# Changes

Main change
* queryAndUpdate can't be used with other than "query" when identity is anonymous, it raises an error otherwise.

Minor change
* Show the test error when loading sns nervous functions if call fails and not certified but it's anonymous identity.

# Tests

* Add two new workflow tests to check that a user load different data in NnsProposals and NnsProposalDetails.
* Rename the file where queryAndUpdate was being tested to the same where it lives.
* Add a test case in queryAndUpdate test.
* Fix tests after change in functionality.


lmuntaner:
@peterpeterparker @mstrasinskis @dskloetd please review

I did two "integration" tests to really check that the api calls were as expected.

I liked the idea of @dskloetd to ONLY mock api calls.

Yet, I'm still mocking the authStore subscribe. I think that we can improve that if we move the logic of logging in to a service and we use the autoStore to purely store data and not do any side effects.

I believe that's a best practice as well, we don't do any side effect in any other store.

frontend/src/lib/pages/NnsProposalDetail.svelte
```diff
@@ -15,10 +15,15 @@
...
   import { goto } from "$app/navigation";
   import { authStore } from "$lib/stores/auth.store";
+  import type { QueryAndUpdateStrategy } from "$lib/services/utils.services";
+  import { isSignedIn } from "$lib/utils/auth.utils";
 
   export let proposalIdText: string | undefined | null = undefined;
   export let referrerPath: AppPath | undefined = undefined;
 
+  let strategy: QueryAndUpdateStrategy;
+  $: strategy = isSignedIn($authStore.identity) ? "query_and_update" : "query";
```
dskloetd, line 25: I don't see where the value of strategy is being used but shouldn't it be an enum or something instead of a string?

lmuntaner: That would be another change and a style decision, at the moment:

```
export type QueryAndUpdateStrategy = "query_and_update" | "query" | "update";
```

dskloetd: I didn't realize it already existed. Does TS guarantee that it can only have those 3 values? I guess then effectively it's already (like) an enum.

I'll have a look at the whole PR tomorrow.

lmuntaner: Yes, if you try passing any other string it will fail. Also, like an enum. Just a matter of style almost.

With enums you can enumerate them at run time though. Not these.

peterpeterparker: Style but also because working with enum in TypeScript isn't perfect. It adds often an unnecessary level of complexity that brings no more type checking safety than just using types. In addition, their manipulation is often difficult and end in some unlucky compromise (like applying cast). Using enum in vanilla isn't the most enjoyable experience in life neither.

So yep style but also because enum are sometimes more pain that joy. Therefore keep it simple when it can be. That said, we don't exclude these. We inherits some from Candid and sometimes, like Llorenç said, when we need iteratable "constants" too.

frontend/src/lib/pages/NnsProposalDetail.svelte
```diff
@@ -50,7 +55,8 @@
       referrerPath === AppPath.Launchpad ? AppPath.Launchpad : AppPath.Proposals
     );
 
-  const findProposal = async () => {
+  let findProposal: () => Promise<void>;
+  $: findProposal = async () => {
```
dskloetd, line 59: Do you need to update the whole function?
Won't it reference the current value of strategy regardless?

frontend/src/lib/pages/NnsProposalDetail.svelte
dskloetd, line 25: Do you need to have a constantly updating strategy variable as opposed to determining the appropriate strategy when making the loadProposal call?

frontend/src/lib/pages/NnsProposals.svelte
```diff
@@ -48,37 +53,40 @@
...
 
   let debounceFindProposals: () => void | undefined;
 
-  // We do not want to fetch the proposals twice when the component is mounting because the filter subscriber will emit a first value
-  const initDebounceFindProposals = () => {
-    debounceFindProposals = debounce(async () => await findProposals(), 250);
-  };
-
   onMount(async () => {
+    // We do not want to fetch the proposals twice when the component is mounting because the filter subscriber will emit a first value
```
dskloetd, line 85: Did you mean to move this code?
For what purpose?
Is it related to the strategy change?

frontend/src/lib/services/$public/proposals.services.ts
```diff
@@ -53,11 +53,13 @@ const handleFindProposalsError = ({
 
 export const listProposals = async ({
   loadFinished,
+  strategy,
 }: {
   loadFinished: (params: {
     paginationOver: boolean;
     certified: boolean | undefined;
   }) => void;
+  strategy?: QueryAndUpdateStrategy;
```
dskloetd, line 62: Do you sometimes not pass the strategy or could it just as well be required?

frontend/src/tests/lib/services/_public/proposals.services.spec.ts
```diff
@@ -225,6 +225,18 @@ describe("proposals-services", () => {
...
+    it("should list proposals with strategy", async () => {
+      const call = async () =>
+        await listProposals({
+          loadFinished: () => {
+            // do nothing here
+          },
+          strategy: "query",
+        });
+
+      await expect(call).resolves;
```
dskloetd, line 237: How does this verify that listProposals didn't do an update call?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+      .mockImplementation(() => Promise.resolve(proposal)),
+  };
+});
+
+jest.mock("$lib/utils/html.utils", () => ({
+  markdownToHTML: (value) => Promise.resolve(value),
+}));
+
+describe("Proposal detail page when not logged in user", () => {
+  afterEach(() => {
```
dskloetd, line 48: I think it's safer to use beforeEach.

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    jest.clearAllMocks();
+  });
+
+  describe("when logged in user", () => {
+    beforeEach(() => {
+      jest
+        .spyOn(authStore, "subscribe")
+        .mockImplementation(mockAuthStoreSubscribe);
+      neuronsStore.reset();
```
dskloetd, line 57: Why not reset the neuronsStore at the top level?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -0,0 +1,130 @@
+/**
```
dskloetd, line 1: NnsProposalDetail isn't new. Was it not tested before?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      await waitFor(() => expect(queryProposal).toHaveBeenCalledTimes(2));
+
+      await waitFor(() =>
+        expect(
+          queryByTestId("proposal-proposer-info-title")
+        ).toBeInTheDocument()
+      );
+      expect(queryByTestId("proposal-system-info-details")).toBeInTheDocument();
+      expect(queryByTestId("voting-confirmation-toolbar")).toBeInTheDocument();
```
dskloetd, line 80: Also expect that there isn't a login button?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+        },
+      });
+
+      await waitFor(() =>
+        expect(
+          queryByTestId("proposal-system-info-details")
+        ).toBeInTheDocument()
+      );
+      expect(queryByTestId("proposal-proposer-info-title")).toBeInTheDocument();
+      expect(queryByTestId("login-button")).toBeInTheDocument();
```
dskloetd, line 120: Is there also something that shouldn't be in the page that you can verify is indeed not there?

dskloetd (changes requested):

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: Sounds good

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: This is the element instead of the button, but I can add that test also.

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: yes, the `voting-confirmation-toolbar`, I'll add it

frontend/src/lib/pages/NnsProposalDetail.svelte
lmuntaner: If you use `const` not.

This is outdated anyway, the strategy is not set in the page, but it's always `query` for not logged in users.

frontend/src/lib/pages/NnsProposals.svelte
lmuntaner: Reverted.

frontend/src/lib/services/$public/proposals.services.ts
lmuntaner: Reverted

frontend/src/tests/lib/services/_public/proposals.services.spec.ts
lmuntaner: reverted

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: done

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: It was tested mocking stores and services.

The tests in `workflows` are tests for page level components where we only mock the api layer.

I started adding one before christmas and haven't added any more.

I believe these are the kind of tests you were expecting from the beginning.

lmuntaner:

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+      });
+    });
+  });
+
+  describe("when not logged in user", () => {
+    beforeEach(() => {
+      jest
+        .spyOn(authStore, "subscribe")
+        .mockImplementation((run: Subscriber<AuthStore>): (() => void) => {
+          run({ identity: undefined });
```
dskloetd, line 101: Is this what causes the user to be not logged in in this test?
Where is the equivalent that makes the user logged in in the previous test?
If it's implicit/default, I recommend making it explicit since it's essential for the test.
At my previous team we had a rule that we only rely on default values in a test, if the test doesn't care about those values. But if a value is important to the test, we set it explicitly, even if it's the same as the default value.
That makes the test easier to read because you don't need to lookup or memorize the default values to know the initial state of the test.
And it makes the test more robust against changes of the default values.
What do you think?

frontend/src/lib/services/$public/sns.services.ts
```diff
@@ -170,7 +173,7 @@ export const loadSnsNervousSystemFunctions = async (
     },
     identityType: "current",
     onError: ({ certified, error }) => {
-      if (certified) {
+      if (certified || identity.getPrincipal().isAnonymous()) {
```
dskloetd, line 176: Can you explain this change?

frontend/src/lib/services/utils.services.ts
```diff
@@ -60,6 +60,10 @@ export const queryAndUpdate = async <R, E>({
       ? getCurrentIdentity()
       : await getAuthenticatedIdentity();
 
+  const currentStrategy = identity.getPrincipal().isAnonymous()
```
dskloetd, line 63: It doesn't seem right that you can pass in a strategy and it will just be ignored in certain cases. Should the parameter be something else like a `boolean alwaysQuery` or something?

dskloetd (changes requested):
Probably not for this PR but I don't understand why we have a function called `queryAndUpdate` which might then only query or only update based on a parameter.
Why not call just `query` if you don't want `queryAndUpdate` to actually query **and** update?

lmuntaner:
> Probably not for this PR but I don't understand why we have a function called `queryAndUpdate` which might then only query or only update based on a parameter. Why not call just `query` if you don't want `queryAndUpdate` to actually query **and** update?

This way we can reuse the same code for different strategies.

In this case, we reuse the same code to query proposals when logged in (query + update) and when not logged in (query).

It's nicer than to have different services that share everything but the type of call.

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: Yes, I'll add a comment.

frontend/src/lib/services/utils.services.ts
lmuntaner: That's what I had, but since we agreed that we want to always use query for not logged in users this was the easiest way.

https://dfinity.slack.com/archives/C03HZERHBT3/p1675411741726609

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: In the previoous one we also set it:

```
jest
```

Actually, I created a helper for this one, I'll change it.

lmuntaner: In general, I agree we should not rely on default values. We try that, but it's not a written rule, nor an enforced one.

There are some contexts that are set for all tests and it's not very clear. Specifically the authoStore and identity ones.

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: Done:

```
      // If the user is not logged in, only a query is done.
      // Therefore, we want to show an error even if the error doesn't come from a certified call.
```

frontend/src/lib/services/utils.services.ts
mstrasinskis: nth: not having the default value in arguments `strategy = "query_and_update"` but apply this logic only if it's `undefined` and only then use `query_and_update` if not set. Currently this parameter is not heavily used, just to avoid future surprises. Not a strong opinion since it's a temporary solution ;)

mstrasinskis (approved):
LGTM, thanks! Optional improvement suggestion.

frontend/src/lib/services/utils.services.ts
lmuntaner: I like the idea!

So that the default is not `query_and_update` but it depends on the identity.

I'll also add in the comments that for anonymous identity, only the `query` strategy is "allowed"

frontend/src/lib/services/$public/sns.services.ts
dskloetd: Ah, ok. So if we do a query and an update, and the query fails, we don't care because the update might succeed?

Can we more explicitly only skip showing the error if it's for the query and there is also an update? For example, right now if you would pass strategy="query", you would not see the error, right?

frontend/src/lib/services/utils.services.ts
dskloetd: Instead of a comment, can we make the method actually fail with an assertion if you use "query and update" while not logged in?

dskloetd (changes requested):

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: No, you would not, but you are the one adding this error handler and the one setting the strategy.

This error handler is coupled with the strategy used here.

frontend/src/lib/services/utils.services.ts
lmuntaner: Yes, that's also a good idea.

lmuntaner: done

frontend/src/lib/services/utils.services.ts
```diff
@@ -26,12 +26,14 @@ let lastIndex = 0;
  * Depending on the strategy makes one or two requests (QUERY and UPDATE in parallel).
  * The returned promise notify when first fetched data are available.
  * Could call onLoad only once if the update response was first.
+ *
+ * If the user is not authenticated, the strategy is ignored and only QUERY is used.
```
dskloetd, line 30: This is no longer the behavior, right?

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: no, the error handler is optional

dskloetd: Do we only need this logic (not showing the error for the first query if there is a second query) for loadSnsNervousSystemFunctions and not in general?

lmuntaner: That's a good point. I caught this because one test failed. But other services might not have the test for failing and that's why I didn't see it.

Yes, ideally, yes.

lmuntaner: Done.

I didn't add tests for each of those though. This would increase the size of the PR even more.

dskloetd: Did you make a note to add the missing tests?

lmuntaner: I did now: https://dfinity.atlassian.net/jira/software/c/projects/GIX/boards/182

Thanks!

PR #1810, opened 2023-02-02
GIX-1245: Add aggregator url to build step

lmuntaner:
# Motivation

Add the sns aggregator url to the build step for nns-dapp frontend.

# Changes

* Use exposed env var VITE_AGGREGATOR_CANISTER_URL in build.csp.mjs to set correct CSP rules.
* Use VITE_AGGREGATOR_CANISTER_URL to set SNS_AGGREGATOR_CANISTER_URL

### DIFF
--- a/frontend/scripts/build.csp.mjs
+++ b/frontend/scripts/build.csp.mjs
@@ -8,12 +8,8 @@ import { findHtmlFiles } from "./build.utils.mjs";
 
 dotenv.config();
 
-// Aggregator canister enabled ONLY in small12 for now
-const enableSnsAggregatorCanister = process.env.VITE_DFX_NETWORK === "small12";
-
-// TODO: Use env var https://dfinity.atlassian.net/browse/GIX-1245
-const aggregatorCanisterUrl =
-  "https://5v72r-4aaaa-aaaaa-aabnq-cai.small12.testnet.dfinity.network";
+const aggregatorCanisterUrl = process.env.VITE_AGGREGATOR_CANISTER_URL;
+const isAggregatorCanisterUrlDefined = aggregatorCanisterUrl.length > 0;
 
 const buildCsp = (htmlFile) => {
   // 1. We extract the start script parsed by SvelteKit into the html file
@@ -127,7 +123,7 @@ const updateCSP = (indexHtml) => {
         content="default-src 'none';
         connect-src 'self' ${cspConnectSrc()};
         img-src 'self' data: https://nns.ic0.app/ https://nns.raw.ic0.app/ ${
-          enableSnsAggregatorCanister ? aggregatorCanisterUrl : ""
+          isAggregatorCanisterUrlDefined ? aggregatorCanisterUrl : ""
         };
         child-src 'self';
         manifest-src 'self';
@@ -153,7 +149,7 @@ const cspConnectSrc = () => {
     process.env.VITE_LEDGER_CANISTER_URL,
   ];
 
-  if (enableSnsAggregatorCanister) {
+  if (isAggregatorCanisterUrlDefined) {
     src.push(aggregatorCanisterUrl);
   }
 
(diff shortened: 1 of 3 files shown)


lmuntaner:
@bitdivine please review

frontend/scripts/build.csp.mjs
```diff
@@ -8,12 +8,8 @@ import { findHtmlFiles } from "./build.utils.mjs";
...
 dotenv.config();
 
-// Aggregator canister enabled ONLY in small12 for now
-const enableSnsAggregatorCanister = process.env.VITE_DFX_NETWORK === "small12";
-
-// TODO: Use env var https://dfinity.atlassian.net/browse/GIX-1245
-const aggregatorCanisterUrl =
-  "https://5v72r-4aaaa-aaaaa-aabnq-cai.small12.testnet.dfinity.network";
+const aggregatorCanisterUrl = process.env.VITE_AGGREGATOR_CANISTER_URL;
+const isAggregatorCanisterUrlDefined = aggregatorCanisterUrl.length > 0;
```
peterpeterparker, line 12: using `URL` to check if url is effectively an url would be safer

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -7,13 +7,10 @@ export const FETCH_ROOT_KEY: boolean =
...
 // TODO: Fix CORS issues
-export const SNS_AGGREGATOR_CANISTER_URL =
-  import.meta.env.SNS_AGGREGATOR_CANISTER_URL ??
-  (DFX_NETWORK === "small12"
-    ? DEV
-      ? "https://5v72r-4aaaa-aaaaa-aabnq-cai.raw.small12.testnet.dfinity.network"
-      : "https://5v72r-4aaaa-aaaaa-aabnq-cai.small12.testnet.dfinity.network"
-    : undefined);
+export const SNS_AGGREGATOR_CANISTER_URL: string | undefined =
+  (import.meta.env.VITE_AGGREGATOR_CANISTER_URL as string) === ""
```
peterpeterparker, line 11: can it be `undefined` too? not that it would really change the outcome...

maybe one of these days we should add a function `nonNullishEmpty` or something

frontend/src/lib/api/sns-aggregator.api.ts
```diff
@@ -267,6 +269,7 @@ export const querySnsProjects = async (): Promise<CachedSns[]> => {
   }
   try {
     const data: CachedSnsDto[] = await response.json();
+    logWithTimestamp("Loading SNS projects from aggregator canister completed");
```
peterpeterparker, line 272: shouldn't this be logged after `convertData` ? I think generally we do so.

peterpeterparker (approved):
LGTM, minor comments. Thanks

frontend/src/lib/constants/environment.constants.ts
lmuntaner: It can't. I would have preferred that as well. But the empty env var is an empty string.

frontend/src/lib/api/sns-aggregator.api.ts
lmuntaner: good point, yes!

frontend/scripts/build.csp.mjs
lmuntaner: Why would it be safer?

frontend/src/lib/constants/environment.constants.ts
peterpeterparker: Good to know, thanks for the feedback

frontend/scripts/build.csp.mjs
peterpeterparker: But we own the config, like I said, just said that out at loud, like spontaneous comment.

PR #1811, opened 2023-02-02
Core Rust aggregator code

bitdivine:
# Motivation
It would be useful to have the basic mplementation in main so that we can start using it in e2e tests.

# Changes
Add aggregator rust code.

# Tests
- There is one e2e test that shows what data is returned; it makes no assertions about the content.


rs/sns_aggregator/src/lib.rs
```diff
@@ -1,6 +1,49 @@
...
+{
+    ic_cdk::setup();
+    let request = call::arg_data::<(HttpRequest,)>().0;
+    let response = assets::http_request(request);
+    call::reply((response,));
+}
+
+/// Periodic function call responsible for populating the cache
+#[ic_cdk_macros::heartbeat]
+async fn heartbeat() {
```
peterpeterparker, line 37: shouldn't we use the new timer API rather than an `heartbeat`? this would stress less the network and consumes less cycles

rs/sns_aggregator/src/lib.rs
```diff
@@ -1,6 +1,49 @@
...
+    crate::upstream::update_cache().await;
+}
+
+#[ic_cdk_macros::post_upgrade]
+fn post_upgrade() {
```
peterpeterparker, line 42: we do keep the assets in stable memory on upgrade?

rs/sns_aggregator/src/state.rs
```diff
@@ -0,0 +1,108 @@
...
+impl State {
+    /// Adds an SNS into the state accessible via certfied query calls.
+    pub fn insert_sns(index: u64, upstream_data: UpstreamData) -> Result<(), anyhow::Error> {
+        Self::insert_sns_v1(index, upstream_data)
+    }
+    /// Adds pre-signed responses for the API version 1.
+    ///
+    /// - /sns/index/{index}.json <- All aggregate data about the SNS, in JSON format.
+    pub fn insert_sns_v1(index: u64, upstream_data: UpstreamData) -> Result<(), anyhow::Error> {
+        let prefix = "/v1";
```
peterpeterparker, line 45: nth: maybe extract a constant in a separate module or something?

and group all constants that are spread around various modules?

peterpeterparker (commented):
Thanks, looks good. Even though the PR is in draft, I assumed we could already have a look, hope you do not mind. Therefore, added few questions and suggestions.

rs/sns_aggregator/src/assets.rs
```diff
@@ -0,0 +1,245 @@
...
+        "txt" => Some("text/plain"),
+        _ => None,
+    })
+}
+
+/// List of recommended security headers as per https://owasp.org/www-project-secure-headers/
+/// These headers enable browser security features (like limit access to platform apis and set
+/// iFrame policies, etc.).
+/// TODO https://dfinity.atlassian.net/browse/L2-185: Add CSP and Permissions-Policy
+fn security_headers() -> Vec<HeaderField> {
```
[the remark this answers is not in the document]
bitdivine: I would actually go one step further and put this in the cdk.  Similar logic gets used over and over again in every app that serves static assets.  Now, the code here _is_ the code from the nns-dapp canister with comments and some changes.  The significant change from nns-dapp is that I have removed the stable asset flag.  Small change but it touches lots of parts in subtle ways, so th enns-dapp would still have its own custom version.  Different implementations might also want to customize how stable assets are handled.  But the code here represents simple, vanilla straightforward static asset serving and I'm sure we won't be the only users.

rs/sns_aggregator/src/assets.rs
```diff
@@ -0,0 +1,245 @@
...
+fn make_asset_certificate_header(asset_hashes: &AssetHashes, asset_name: &str) -> (String, String) {
+    let certificate = ic_cdk::api::data_certificate().unwrap_or_else(|| {
+        panic!("data certificate is only available in query calls");
+    });
+    let witness = asset_hashes.0.witness(asset_name.as_bytes());
+    let tree = labeled(LABEL_ASSETS, witness);
+    let mut serializer = serde_cbor::ser::Serializer::new(vec![]);
+    serializer.self_describe().unwrap();
+    tree.serialize(&mut serializer).unwrap_or_else(|e| {
+        panic!("failed to serialize a hash tree: {}", e);
```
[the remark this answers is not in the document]
bitdivine: Panicking when handling an http call will in fact do exactly that.  But while on the subject, in the nns-dapp we do have panics in upgrade code, which is really not good and I would like to remove them.

rs/sns_aggregator/src/assets.rs
```diff
@@ -0,0 +1,245 @@
...
+        format!(
+            "certificate=:{}:, tree=:{}:",
+            base64::encode(&certificate),
+            base64::encode(&serializer.into_inner())
+        ),
+    )
+}
+
+/// Computes the sha256 of some given bytes.
+pub fn hash_bytes(value: impl AsRef<[u8]>) -> Hash {
```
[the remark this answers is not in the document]
bitdivine: 90% of the time when hashing, it's with an IV and exactly which IV gets used and how is a gnarly subject that I would rather avoid.  I would make a library, yes, but consisting of the contents of this file and call it the asset serving library.  Does that sound OK to you?

Also on the subject of libraries, I would like to extract the NNS types as I suspect that that could be used to rid us of the dependency on the ic repo in nns-dapp, however that requires some careful work together with Maksim.

rs/sns_aggregator/src/assets.rs
```diff
@@ -0,0 +1,245 @@
...
+    });
+}
+
+fn update_root_hash(a: &AssetHashes) {
+    let prefixed_root_hash = &labeled_hash(LABEL_ASSETS, &a.0.root_hash());
+    ic_cdk::api::set_certified_data(&prefixed_root_hash[..]);
+}
+
+/// Responds to an HTTP request for an asset.
+pub fn http_request(req: HttpRequest) -> HttpResponse {
```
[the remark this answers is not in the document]
bitdivine: Is http streaming possible?  I thought there was a lot of discussion about it not being possible.  If it is, I would be interested in learning how, independent of this PR.

At present, with 1-2 SNSs we are very far from that limit so I don't see this as a blocker for the next SNS but this is absolutely something that needs to be looked into.  Pagination is 10 SNSs, each SNS metadata is about 4K (with the logo removed to another endpoint) so with 10 relatively normal SNSs we will still be far from the limit (40k, so 4% of a 1Mb limit), however the length of the SNS description is unbounded and that is a security risk that's already being looked into.  The length will eventually have to be bounded in the SNS code.  So I think that in the specific case of SNSs we will be fine, assuming reasonable limits are applied, but if we make this code into a library for reuse, other code may need that pagination.

rs/sns_aggregator/src/assets.rs
```diff
@@ -0,0 +1,245 @@
...
+            None => HttpResponse {
+                status_code: 404,
+                headers,
+                body: ByteBuf::from(format!("Asset {} not found.", request_path)),
+            },
+        }
+    })
+}
+
+/// Inserts a favicon into the certified assets, if there is not one already.
```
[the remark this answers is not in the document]
bitdivine: Good idea.

rs/sns_aggregator/src/lib.rs
```diff
@@ -1,6 +1,49 @@
...
+mod upstream;
 
+use assets::{insert_favicon, HttpRequest};
+use ic_cdk::api::call::{self};
+use state::STATE;
+use types::Icrc1Value;
+
+/// API method for basic health checks.
 #[ic_cdk_macros::query]
 fn greet(name: String) -> String {
```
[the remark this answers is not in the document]
bitdivine: Good point.

rs/sns_aggregator/src/lib.rs
bitdivine: I know about the timer but we need to agree on what our update frequency is going to be.  Once we know what sort of refresh rate we are OK with and how we will react to launches, this can be adapted accordingly.

rs/sns_aggregator/src/lib.rs
bitdivine: No, not at the moment.  Yes, this means that on upgrade, for a short time the cache will be empty.  That is fine for now.

rs/sns_aggregator/src/state.rs
bitdivine: That would not be conventional for Rust code.  Constants are generally defined in the context where they are used.  At the moment there is no use of "v1" outside this function, so it is scoped to within this function.  If it is moved up, it would be attached to a struct or module, not moved into a global constants file.  There are exceptions, but it would not normally be correct for Rust code.

rs/sns_aggregator/src/assets.rs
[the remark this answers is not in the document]
peterpeterparker: Makes sense. Maybe we can do two steps. One avoid double code in our apps and in parallel try to bring these to the cdk. Anyhow not related to your PR, thanks for the feedback.

rs/sns_aggregator/src/assets.rs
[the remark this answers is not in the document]
peterpeterparker: Gotcha

rs/sns_aggregator/src/assets.rs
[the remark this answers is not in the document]
peterpeterparker: I like to spread my code in more modules but I'm a rust newbie and regardless of it, can for sure live with as it is.

rs/sns_aggregator/src/assets.rs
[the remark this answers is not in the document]
peterpeterparker: > Is http streaming possible?

Since more than a 1 year I would say - i.e. all my personal canister streams http requests since a long time.

Agree for start we will probably fine but, nevertheless think it's safer to implement it sooner than late. Not necessary in this PR but soon. Can you add a TODO?

rs/sns_aggregator/src/lib.rs
peterpeterparker: You can already implement a timer with a short frequency and then discuss it if it's the concern. If I would be the one that pays for the cycles, I would not implement an hearbeat here.

Not blocker for your PR, maybe not for going live neither but still, kind of feel it's what would be the best for this feature.

rs/sns_aggregator/src/lib.rs
peterpeterparker: > this means that on upgrade, for a short time the cache will be empty ...

... and per extension users requesting NNS-dapp information will get unexpected errors on screen. Edge case but still.

Maybe add a TODO?

rs/sns_aggregator/src/state.rs
peterpeterparker: Gotcha, thanks for the explanation.

rs/sns_aggregator/src/assets.rs
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    /// Gets an asset, using the actual asset path, e.g.
+    /// `/foo/index.html` not `/foo/`.
+    pub fn get(&self, path: &str) -> Option<&Asset> {
+        self.0.get(path)
+    }
+}
+
+fn content_type_of(request_path: &str) -> Option<&'static str> {
```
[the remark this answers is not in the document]
bitdivine: This is just standard asset canister code.  So, can we delete it?  This canister also provides icons directly, although it is unclear whether they will be used like this.  Also, there was the idea of supporting Candid as well, so there would be one more entry to add.  I think it is good to have the machinery in place.

rs/sns_aggregator/src/lib.rs
bitdivine: Hmm.  Any experience with calling async functions from the timer, @peterpeterparker ?

bitdivine: Aha: https://github.com/dfinity/cdk-rs/issues/351

peterpeterparker: Don't have experience with it yet. If I remember correctly, they might have recently used the timer in II.

rs/sns_aggregator/src/lib.rs
bitdivine: I have opened a ticket and put it in TODO (not backlog, where it might not be done any time soon): https://dfinity.atlassian.net/browse/GIX-1291

rs/sns_aggregator/src/assets.rs
```diff
@@ -0,0 +1,245 @@
...
+}
+
+fn update_root_hash(a: &AssetHashes) {
+    let prefixed_root_hash = &labeled_hash(LABEL_ASSETS, &a.0.root_hash());
+    ic_cdk::api::set_certified_data(&prefixed_root_hash[..]);
+}
+
+/// Responds to an HTTP request for an asset.
+pub fn http_request(req: HttpRequest) -> HttpResponse {
+    let mut parts = req.url.splitn(2, '?');
```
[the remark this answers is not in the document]
bitdivine: I tested my understanding:  I added a print statement:
```
--- a/rs/sns_aggregator/src/assets.rs
+++ b/rs/sns_aggregator/src/assets.rs
@@ -198,6 +198,7 @@ fn update_root_hash(a: &AssetHashes) {
 
 /// Responds to an HTTP request for an asset.
 pub fn http_request(req: HttpRequest) -> HttpResponse {
+    ic_cdk::api::print(format!("Handling request {}", &req.url));
     let mut parts = req.url.splitn(2, '?');
```
I then deployed and requested: http://q4eej-kyaaa-aaaaa-aaaha-cai.localhost:8080/v1/sns/root/q3fc5-haaaa-aaaaa-aaahq-cai/logo.png

This printed:
```
[Canister q4eej-kyaaa-aaaaa-aaaha-cai] Handling request /v1/sns/root/q3fc5-haaaa-aaaaa-aaahq-cai/logo.png
[Canister q4eej-kyaaa-aaaaa-aaaha-cai] Handling request /favicon.ico
```
So, confirmed, logo is a misnomer.  I have corrected the description of the url field to clarify that it does not contain the first 3 parts of a normal URL (protocol, domain, port) and only path+query string.

Renaming the field is problematic because it interfaces with the replica and the candid names need to match.  I was curious how other projects handle this.  The name `url` seems to have been used everywheer I have seen in Rus, motoko and even in python: https://github.com/demergent-labs/kybra#http_request-and-http_request_update

bitdivine:
@peterpeterparker Notes on follow-up work:
- I am making a PR right now to:
  - Pass in update frequency as an init var.
  - Preserve the cache across upgrades
    - Note: Combined with the parameterized update interval, preserving artefacts can be tested by redeploying with a very long update interval and verifying that the cache is still populated.

- There are various Jira tickets in flight for items we discussed.

rs/sns_aggregator/src/assets.rs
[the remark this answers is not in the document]
peterpeterparker: Makes sense. Then I guess it's the same as for one of my other question, some day we can maybe refactor this to a utils of NNS-dapp, since same in both canister, or hopefully it will be integrated in the cdk.

rs/sns_aggregator/src/assets.rs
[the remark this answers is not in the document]
peterpeterparker: Thanks sounds good. Note that it is not exactly what I meant with my "Nice to have" comment.

What I meant is that instead of manually manipulating the url

```
let mut parts = req.url.splitn(2, '?');
```

We can use a crate named `Url` that ease working with url.

e.g.

```
let url = Url::parse(&["http://localhost", req.url);
let path = url.path();
let query = url.query();
```

But again, that was just a "nice to have" suggestion.

rs/sns_aggregator/src/upstream.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+/// Gets a list of SNSs from the nns-sns-wasm canister and puts it in the queue of SNSs to query.
+///
+/// Note: We can improve on this by filtering out SNSs that have become const.
+async fn set_list_of_sns_to_get() -> anyhow::Result<()> {
+    ic_cdk::println!("Asking for more SNSs");
+    let result: Result<(ListDeployedSnsesResponse,), (RejectionCode, std::string::String)> = ic_cdk::api::call::call(
+        CanisterId::from_text("qaa6y-5yaaa-aaaaa-aaafa-cai").expect("I don't believe it's not a valid canister ID??!"),
```
peterpeterparker, line 42: Should we also add a TODO to extract the wasm canister id to an environment varialbe in case a developer want to deploy this aggregator locally or on a testnet?

rs/sns_aggregator/src/lib.rs
```diff
@@ -1,6 +1,57 @@
...
+fn post_upgrade() {
+    // Browsers complain if they don't get pretty pictures.  So do I.
+    insert_favicon();
+    // Periodic function call responsible for populating and aggregating the cache.
+    //
+    // Note: In future we are likely to want to make the duration dynamic and
+    //       have an altogether more complicated data collection schedule.
+    set_timer_interval(
+        Duration::from_secs(1),
+        || ic_cdk::spawn(crate::upstream::update_cache()),
```
peterpeterparker, line 50: note sure we need it but do we need a lock to ensure the no other `update_cache` function is started unless previous job is finished?

peterpeterparker (commented):
LGTM, thanks!!!

Added two comments in my last review.

Don't know if someone with more experience in Rust than me should not review it as well?

Note regarding test: I understand we do not provide tests in this PR because we either gonna provide some in other PRs or it will mostly tested with E2E tests.

peterpeterparker (approved):

rs/sns_aggregator/src/upstream.rs
bitdivine: Agree.  I'd like to go a step further and make it a config var, so that it can be passed as an init arg.  https://dfinity.atlassian.net/browse/GIX-1297

rs/sns_aggregator/src/assets.rs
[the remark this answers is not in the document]
bitdivine: The problem is that the data is not a URL, although it is called `url`, so the call would not work.  Ah, I see, you would prepend the domain, then parse as a URL.  That would work.  I think that if the situation became complicated, that would make sense.  Stripping the query from the end of a path is trivial but if we started passing in query parameters, leveraging the Url library would definitely have benefit.  At the moment, we could even omit the query stripping and everything would work as intended.  Interesting.  Let me have a think about this and a look at exactly what the certificates do.  (E.g. https://nns.ic0.app/launchpad/?foo has a valid cert but https://nns.ic0.app/launchpad// does not, so there is no normalization of the asset paths but the query string doesn't matter... surprisingly to me)  If and when there is a general purpose assets library, these considerations will have to be taken into account.

rs/sns_aggregator/src/lib.rs
bitdivine: That is a bit of an open-ended question but I think it isn't needed.  The code is intended to be safe for concurrent execution, so it _should_ be fine for another to start before the current update has completed.  Due to the nature of the IC, code will never be running simultaneously but there are `awaits` when collecting data, so what can happen is that during an await, another call starts collecting data.  But even assuming simultaneous execution the code should be fine.  The data itself is stored in RefCells, which prevents one call from changing data while another call has borrowed it.  I have made the RefCells fine-grained - possibly overdone it a bit - so even if I have terrible bugs in my code it should never be possible for the contents of a refcell to be mangled by e.g. simultaneous rites or a read during another write.  A risk from having fine grained locks is some kind of deadlock condidtion.  But I only hold one lock at a time, so that is OK.  Another possible risk is inconsistent data, e.g. because call A reads some data behind a lock, then starts computing based on that, then another call changes that data, then the first call writes based on its by now out-of-date data.  E.g. one call could read the list of SNSs pick an SNS from there, then while it is collecting the SNS, another call updates the list of SNSs, maybe adding a new SNS.  There too we are fine.  It's possible to simplify by having a global write lock and locking it during updates but that can (presumably) block reads which would not be wanted.

rs/sns_aggregator/src/lib.rs
bitdivine: The timer API is working and I'm able to configure the timer on the fly for testing.  Configuration coming up in a PR.

PR #1813, opened 2023-02-02
Cleanup neurons.services.spec.ts

dskloetd:
# Motivation

I was trying to improve the listNeurons test and I noticed it didn't start with clean state.
It turns out there was a mixture of afterEach and beforeEach and together they didn't guarantee every test starts with a clean state.

# Changes

1. In the list neurons test, verify the store before (as well as after) making the call.
2. Add one global `beforeEach()` which does all the state cleanup and remove all the other clean call.

Normally I wouldn't put these things in the same PR, but (1) demonstrates well why (2) is important so I kept it together.

# Tests

only tests

### DIFF
--- a/frontend/src/tests/lib/services/neurons.services.spec.ts
+++ b/frontend/src/tests/lib/services/neurons.services.spec.ts
@@ -208,7 +208,7 @@ describe("neurons-services", () => {
   beforeEach(() => {
     spyGetNeuron.mockClear();
     jest.clearAllMocks();
-    neuronsStore.setNeurons({ neurons: [], certified: true });
+    neuronsStore.reset();
     accountsStore.reset();
     resetIdentity();
   });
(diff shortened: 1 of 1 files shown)


frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -205,15 +205,15 @@ describe("neurons-services", () => {
     .spyOn(api, "claimOrRefreshNeuron")
     .mockImplementation(() => Promise.resolve(undefined));
 
-  afterEach(() => {
+  beforeEach(() => {
     spyGetNeuron.mockClear();
     jest.clearAllMocks();
+    neuronsStore.setNeurons({ neurons: [], certified: true });
```
lmuntaner, line 211: Would `neuronsStore.reset()` work? I believe I introduced it later and didn't change the previous tests.

lmuntaner (approved):
LGTM!

frontend/src/tests/lib/services/neurons.services.spec.ts
dskloetd: Yes, works.

PR #1815, opened 2023-02-03
Use real toasts store in neurons service test.

dskloetd:
# Motivation

There is no need to mock light weight objects like Svelte stores.
This change demonstrates using the real toasts store in `neurons.services.spec.ts` instead of mocking it.

# Changes

Don't mock `$lib/stores/toasts.store` but instead let the neurons service put messages in the real store.
Subscribe to the store and have tests check the actual messages in the store.
Additionally we expect specific strings instead of just expecting there to be a toast.
Note that all expected toasts are error toasts even though the code uses a mix or toastsShow and toastsError to produce them.

# Tests

only tests

### DIFF
--- a/frontend/src/tests/lib/services/neurons.services.spec.ts
+++ b/frontend/src/tests/lib/services/neurons.services.spec.ts
@@ -57,14 +57,16 @@ const {
   topUpNeuron,
 } = services;
 
-let toasts : ToastMsg[] = [];
-toastsStore.subscribe(t => toasts = t);
+let toasts: ToastMsg[] = [];
+toastsStore.subscribe((t) => (toasts = t));
 
 const expectToastError = (contained: string) =>
-      expect(toasts).toMatchObject([{
-        level: "error",
-        text: expect.stringContaining(contained)
-      }]);
+  expect(toasts).toMatchObject([
+    {
+      level: "error",
+      text: expect.stringContaining(contained),
+    },
+  ]);
 
 let testIdentity: Identity | null = mockIdentity;
 const setNoAccountIdentity = () => (testIdentity = null);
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
I like it!

dskloetd:
Can you comment on line 522 and 816?

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -257,7 +263,7 @@ describe("neurons-services", () => {
       });
 
       expect(response).toBeUndefined();
-      expect(toastsError).toBeCalled();
+      expectToastError("minimum of 1 ICP");
```
lmuntaner, line 266: Actually, thinking about it, it would be better to import the translations.

For example:

```
import en from "../../../mocks/i18n.mock";

//...
expect(button?.innerHTML).toEqual(en.core.continue);
```

lmuntaner (changes requested):
I requested a change.

I'm sorry for this, I should take more time to review and do it more thoroughly.

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -513,7 +519,10 @@ describe("neurons-services", () => {
       };
       await toggleAutoStakeMaturity(neuron);
 
-      expect(toastsShow).toHaveBeenCalled();
+      // Is this really the right error for this situation??
```
lmuntaner, line 522: No, definitely not.

I'm guessing that the `setAccountIdentity` didn't work setting the identity for the call.

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -804,7 +813,10 @@ describe("neurons-services", () => {
         targetNeuronId: neuron2.neuronId,
       });
 
-      expect(toastsShow).toHaveBeenCalled();
+      // Is this really the right error for this situation??
```
lmuntaner, line 816: No, same as before `setAccountIdentity` didn't do what intended.

frontend/src/tests/lib/services/neurons.services.spec.ts
lmuntaner: Yes, that sounds good, thanks!

lmuntaner (approved):

PR #1817, opened 2023-02-04
More formatting and linting

bitdivine:
# Motivation
The formatting and lint commands are incomplete.  Take 10 minutes to make the situation better.

# Changes
- Added a yaml formatter (the one I use locally but it is not in CI)
- Format the YAML files
- ~Update the e2e-test lockfile, as it is out of sync with packages.json and cannot be used as is.~ actually, it's just generated with an older nodejs.
- Create a single top-level lint command, similar to the top-level fmt command.  The list of lint shell scripts is incomplete, so the top level command is necessarily also incomplete.

# Not included
These are not included in this PR:
- frontend linting (encoded in ci but lot as a local shell script)

# Tests
- See CI for the formatting update
- I have run the linter locally.  While convenient, the other lint commands should be added before it makes sense to use it in CI, in my opinion.


.github/workflows/checks.yml
```diff
@@ -16,6 +16,8 @@ jobs:
         run: rustup component add rustfmt
       - name: Install shfmt
         run: sudo snap install --classic shfmt
+      - name: Install yq
+        run: sudo snap install yq
```
bitdivine, line 20: Installs jq so that it can be used to format yaml files

e2e-tests/sns-init-assets/sns_init.yaml
```diff
@@ -70,8 +70,7 @@ min_icp_e8s: 100000000
 # principal IDs. In most use-cases, this would be the same as the original set
 # of controller(s). Must not be empty.
 #
-fallback_controller_principal_ids:
-  [hpikg-6exdt-jn33w-ndty3-fc7jc-tl2lr-buih3-cs3y7-tftkp-sfp62-gqe]
+fallback_controller_principal_ids: [hpikg-6exdt-jn33w-ndty3-fc7jc-tl2lr-buih3-cs3y7-tftkp-sfp62-gqe]
```
bitdivine, line 73: Automatic formatting.  No manual changes.

scripts/fmt
```diff
@@ -8,3 +8,4 @@ set -x
 ./scripts/fmt-json
 ./scripts/fmt-rs
 ./scripts/fmt-sh
+./scripts/fmt-yaml
```
bitdivine, line 11: Format yaml files as well.  `scripts/fmt` is used by CI to format all the code so yaml formatting will also be checked in CI.

scripts/fmt-yaml
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: Yaml formatter.  Same flags and code pattern as fmt-sh.  Note that CI does not rely on `--check` but rather formats and then checks for a git diff.

scripts/lint
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: Top level  lint command similar to the top level `scripts/fmt` command.  The top level lint command calls all existing lint shell scripts, however some lints are not present as a script so aren't included here.

scripts/setup
```diff
@@ -130,6 +130,15 @@ install_jq_linux() {
   sudo apt-get install jq
 }
```
bitdivine, line 132: Install `yq` for developers and any CI jobs needing a complete toolchain.

scripts/testnet-dfx-add
```diff
@@ -4,7 +4,7 @@ export DFX_NETWORK="$1"
 IC_REPO_DIR="${IC_REPO_DIR:-$HOME/dfn/ic}"
 
 if [[ "${DFX_NETWORK}" == "--all" ]]; then
```
bitdivine, line 6: My local shellcheck complains about the quotes.  It seems to be the same version as CI.  I wonder why the difference.  Both local and CI seem to be happy without the quotes.

sns_init.yml
```diff
@@ -1,35 +1,26 @@
-description: 'We''ve been making video game games since 2006, we''re based in Toronto,
-  Canada and launched in December 2011. And we''ve had our first video game show on
-  the internet in 2012, but didn''t quite have the excitement of The Sims.
+description: 'We''ve been making video game games since 2006, we''re based in Toronto, Canada and launched in December 2011. And we''ve had our first video game show on the internet in 2012, but didn''t quite have the excitement of The Sims.
```
bitdivine, line 1: Automatic formatting, again.

lmuntaner (commented):
One question about usage:

If --check is passed, it fails if the formatting does some change. If `--check` is not passed, then it does the format changes. Is that correct?

bitdivine:
@lmuntaner - Yes, that is correct, although given that the --check flag is not used in CI I would be a little wary of trusting it.  Maybe we should enable it.  It might have a few bugs at first but then they would get fixed.

lmuntaner (approved):
LGTM!

PR #1818, opened 2023-02-04
add benchmarklarge testnet to dfx.json

mraszyk:


lmuntaner (commented):
We add them with a script `scripts/testnet-dfx-add`.

Can you check whether it will be removed if we run the script? Or whether you can add it to the script.

Thanks!

mraszyk:
@lmuntaner Thank you for making me aware of the script! I've just run
```
IC_REPO_DIR=/home/martin/ic ./scripts/testnet-dfx-add benchmarklarge
```
which added two more lines to dfx.json and committed.

lmuntaner (approved):
Thanks!

PR #1819, opened 2023-02-06
feat: ckBTC transactions on wallet detail page

peterpeterparker:
# Motivation

Add CTA to execute a ckBTC transaction on wallet detail page.

# Changes

- add footer action to instantiate ckBTC transaction and pass selected account
- modify dispatch on success to reload account data
- duplicate properties fee and token as [requested](https://github.com/dfinity/nns-dapp/pull/1812#discussion_r1095949160)

# Screenshot

demo2

### DIFF
--- a/frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
+++ b/frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
@@ -46,7 +46,7 @@
 
     if (success) {
       toastsSuccess({ labelKey: "accounts.transaction_success" });
-      dispatcher("nnsClose");
+      dispatcher("nnsTransfer");
     }
   };
 </script>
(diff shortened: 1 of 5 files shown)


frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
```diff
@@ -46,31 +46,26 @@
 
     if (success) {
       toastsSuccess({ labelKey: "accounts.transaction_success" });
-      dispatcher("nnsClose");
+      dispatcher("nnsTransfer");
```
lmuntaner, line 49: Why did you change this event? Should we change it in all modals?

AFAIK, the modal takes care of the success and error messages. Therefore, there was no need to expose that outside.

lmuntaner (approved):
Just a minor question, but already approving. LGTM!

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
peterpeterparker: > modify dispatch on success to reload account data

To reload the account after transfer because otherwise the new balance is not reflected in screen.

In Nns we syncAccounts and therefore it works out.

In Sns wallet it currently works out because for some reason the `accountIdentifier` is refreshed after transfer and per extension the account is reloaded. Might need further inspection here but, it works out so fine so far.

lmuntaner: ok, thanks for the explanation

PR #1820, opened 2023-02-06
Chore: Add permissions to sns neuron in testnet

lmuntaner:
# Motivation

TESTNET ONLY

SNS Neuron Detail: Add functionality to add granular permissions to a principal.

This is useful to test different scenarios:
* Community Fund SNS neurons without going through all the CF process.
* Hardware Wallet controlled SNS neurons while they can't participate in a sale nor stake SNS neurons.

# Changes

* New component `SnsPermissionsCard` that displays all principals that have permissions for that neuron and which permissions.
* New modal `AddPermissionsModal` opened from `SnsPermissionsCard` to add permissions.

# Tests

No tests. This is a testnet only feature.


lmuntaner:
@mstrasinskis @peterpeterparker @dskloetd please review

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
```diff
@@ -0,0 +1,113 @@
...
+  {/if}
+
+  {#if currentStep.name === "AddPermissions"}
+    <p>{`Select permissions for ${principal?.toText()}`}</p>
+    {#each selectablePermissions as { checked, permission }}
+      <Checkbox
+        inputId={String(permission)}
+        {checked}
+        on:nnsChange={() => toggle(permission)}
+        >{permissionI18nMapper[permission]}</Checkbox
```
mstrasinskis, line 101: `permissionI18nMapper[permission] ?? permission` to not break it accidentally?

mstrasinskis (approved):
Great idea!

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
lmuntaner: sounds good!

mstrasinskis (commented):
Just thinking out loud, is it possible to use dynamic import for it...

lmuntaner:
> Just thinking out loud, is it possible to use dynamic import for it...

I reuse a lot of components. The extra complexity for the dynamic import will not even make a difference of 1Kb.

mstrasinskis:
> > Just thinking out loud, is it possible to use dynamic import for it...
> 
> I reuse a lot of components. The extra complexity for the dynamic import will not even make a difference of 1Kb.

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
```diff
@@ -0,0 +1,113 @@
...
+  {/if}
+
+  {#if currentStep.name === "AddPermissions"}
+    <p>{`Select permissions for ${principal?.toText()}`}</p>
+    {#each selectablePermissions as { checked, permission }}
+      <Checkbox
+        inputId={String(permission)}
+        {checked}
+        on:nnsChange={() => toggle(permission)}
+        >{permissionI18nMapper[permission] ?? permission}</Checkbox
```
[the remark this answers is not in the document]
lmuntaner: No, no plan on using it for non-devs.

It's a little ugly and I already had the mapper for the HW, but this will make the PR and changes even simpler. I'll use it.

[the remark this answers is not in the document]
lmuntaner: done

frontend/src/lib/components/neuron-detail/SnsPermissionsCard.svelte
```diff
@@ -0,0 +1,59 @@
...
+    getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
+
+  let neuron: SnsNeuron | undefined | null = $store.neuron;
+
+  const openModal = async () => {
+    openSnsNeuronModal({ type: "dev-add-permissions" });
+  };
+</script>
+
+<!-- ONLY FOR TESTNET. NO UNIT TESTS -->
```
dskloetd, line 25: Can you also put a message like this in the UI so the component can't be accidentally reused by someone who doesn't read this comment?

frontend/src/lib/i18n/en.json
```diff
@@ -107,6 +107,7 @@
     "fetch_transactions": "Sorry, there was an error loading the transactions for this account.",
     "transaction_data": "Sorry, there was an error with the transaction $txId.",
     "amount_not_enough_stake_sns_neuron": "Sorry, the amount is too small. You need a minimum of $amount $token to stake a $token neuron",
+    "adding_permissions": "There was an error adding permissions.",
```
dskloetd, line 110: Why translate a test-only UI? Who will use this? And if we do want to translate, what about the titles in the svelte template? Or are svelte templates already translated through a different mechanism?

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
```diff
@@ -0,0 +1,112 @@
...
+        console.error(err);
+        toastsError({ labelKey: "error.adding_permissions", err });
+      } finally {
+        stopBusy("dev-add-sns-neuron-permissions");
+      }
+    }
+  };
+</script>
+
+<!-- ONLY FOR TESTNET. NO UNIT TESTS -->
```
dskloetd, line 78: Same question: also make it obvious in the UI?

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
```diff
@@ -0,0 +1,112 @@
...
+    {
+      name: "AddPermissions",
+      title: "Select Permissions",
+    },
+  ];
+
+  let currentStep: WizardStep;
+  let modal: WizardModal;
+
+  let principal: Principal | undefined = undefined;
```
dskloetd, line 37: Would it make sense to use a name that makes it clear what this is the principal of?
It could be the current id, or the id being granted permissions, right?

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
```diff
@@ -0,0 +1,112 @@
...
+    })
+  );
+  const toggle = (permission: SnsNeuronPermissionType) => {
+    selectablePermissions = selectablePermissions.map((p) =>
+      p.permission === permission ? { ...p, checked: !p.checked } : p
+    );
+  };
+
+  const add = async () => {
+    if (principal !== undefined) {
```
dskloetd, line 53: Will the button just not do anything? Should it indicate why?

Also with
`if (principal === undefined) return;`
you can leave the rest of the function with 1 indent less.
Does the team have opinions on this style?

dskloetd (commented):
I think it would be nice if all the test-only UIs were clearly marked with a different background color or something.

frontend/src/lib/i18n/en.json
lmuntaner: The toast error service that we use expects an i18n key

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
lmuntaner: Good point, I'll add it

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
lmuntaner: No, there is no link between users and principal ids yet.

Therefore, this is just an id. We have to leave with it. We have the same for hotkeys.

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
lmuntaner: Yes, we do. I forgot because I'm the only one that wrote the document: https://www.notion.so/NNS-Dapp-Frontend-Development-Guidelines-and-Architecture-5f387639a25e4ead8bf7a066df09860c#36d1c6544d0f45548f5c9a776c198c43

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
```diff
@@ -0,0 +1,115 @@
...
+  );
+  const toggle = (permission: SnsNeuronPermissionType) => {
+    selectablePermissions = selectablePermissions.map((p) =>
+      p.permission === permission ? { ...p, checked: !p.checked } : p
+    );
+  };
+
+  const add = async () => {
+    if (principal === undefined) {
+      return;
```
dskloetd, line 54: My other question remain. Will you press a button and nothing happens?
Should there be feedback? Should the button be disabled?

dskloetd (approved):

frontend/src/lib/modals/sns/neurons/AddPermissionsModal.svelte
lmuntaner: It's ok for development features.

And it's an edge case anyway

PR #1821, opened 2023-02-06
Aggregator - preserve assets over upgrade

bitdivine:
# Motivation
We would like to preserve the cache, if possible, on upgrade so that redeploying does not cause the cache to be empty for a while and no SNSs to be shown on the NNS-dapp.

# Changes
* Move the upstream cache and downstream assets into a stable assets structure.
* Make the stable assets serializable (involves adding traits to the types contained in it.
* Add pre-upgrade and post-upgrade hooks to serialize and restore the stable memory, discarding it and starting afresh if the new schema is incompatible with the old.
* Add a did file and API methods to support testing

# Tests
* CI tests are in progress but we may need to merge before they are ready, due to a shortage of time.
* This code, less some comments and cleanup, has been deployed to farm and static testnets and appears to operate correctly there.


rs/sns_aggregator/src/types.rs
```diff
@@ -9,7 +9,7 @@ pub mod ic_sns_wasm;
 
 // Other types
 pub mod slow;
-pub mod state;
+pub mod upstream;
```
bitdivine, line 12: Module renamed for clarity but otherwise the same.

rs/sns_aggregator/src/types.rs
```diff
@@ -18,7 +18,7 @@ pub use ic_sns_ledger::{Tokens as SnsTokens, Value as Icrc1Value};
 pub use ic_sns_root::ListSnsCanistersResponse;
 pub use ic_sns_swap::GetStateResponse;
 pub use ic_sns_wasm::{DeployedSns, ListDeployedSnsesResponse, SnsCanisterIds};
-use serde::Serialize;
+pub use serde::Serialize;
```
bitdivine, line 21: Expose so that the same version of serialize is used everywhere.

rs/sns_aggregator/src/upstream.rs
```diff
@@ -4,17 +4,17 @@ use std::str::FromStr;
...
-use crate::types::state::UpstreamData;
+use crate::types::upstream::UpstreamData;
 use crate::types::{self, EmptyRecord, GetStateResponse, Icrc1Value, SnsTokens};
 use anyhow::anyhow;
 use ic_cdk::api::{call::RejectionCode, management_canister::provisional::CanisterId, time};
 
 /// Updates one part of the cache:  Either the list of SNSs or one SNS.
 pub async fn update_cache() {
-    ic_cdk::println!("Heartbeat");
+    ic_cdk::println!("Getting upstream data...");
```
bitdivine, line 14: This is no longer triggered by heartbeat.  The text shoul dactually have been changed in the PR switching to timers.

rs/sns_aggregator/src/assets.rs
```diff
@@ -212,7 +223,7 @@ pub fn http_request(req: HttpRequest) -> HttpResponse {
         let certificate_header = make_asset_certificate_header(&state.asset_hashes.borrow(), request_path);
         headers.push(certificate_header);
 
-        match state.assets.borrow().get(request_path) {
+        match state.stable.borrow().assets.borrow().get(request_path) {
```
peterpeterparker, line 226: `state.stable.borrow().assets.borrow()`

borrow + borrow is correct?

(same in few other places)

rs/sns_aggregator/src/lib.rs
```diff
@@ -9,56 +9,158 @@ mod upstream;
...
 #[ic_cdk_macros::post_upgrade]
-fn post_upgrade() {
+fn post_upgrade(config: Option<Config>) {
+    ic_cdk::api::print("Calling post_upgrade...");
+    // Make an effort to restore state.  If it doesn't work, give up.
+    STATE.with(|state| {
+        match ic_cdk::storage::stable_restore()
+            .map_err(|err| format!("Failed to retrieve stable memory: {err}"))
+            .and_then(|(bytes,): (Vec<u8>,)| {
+                serde_cbor::from_slice(&bytes[..]).map_err(|err| format!("Failed to parse stable memory: {err:?}"))
```
peterpeterparker, line 92: not related to your PR but for my knowledge: is there any performance advantages of using cbor encoding for pre and post upgrade deserialization?

I generaly use the ic_cdk `stable_restore` not sure if they do the same

e.g. https://github.com/papyrs/ic/blob/e969090b9cae3f9e50654e15bcfef3f6d687a0cd/canisters/src/storage/src/lib.rs#L81

rs/sns_aggregator/src/lib.rs
```diff
@@ -9,56 +9,158 @@ mod upstream;
...
+            }) {
+            Ok(data) => {
+                *state.stable.borrow_mut() = data;
```
peterpeterparker, line 95: should the certified data be set after upgrade or is that already done?

questions comes from the same piece of code I linked above, I generally redo the stable memory and then iterate to rebuilt the tree of the hashes and apply the certificate data.

peterpeterparker (commented):
Just few questions because I'm a noob

rs/sns_aggregator/src/lib.rs
bitdivine: Nice catch.  I planned to set the hashes with `::from(&assets)` after restoring data have failed to do so.  Done now.  I looked at the restore most closely on the farm, and there with a raw API call it was in no way obvious that the certificate was missing or invalid.  If valid, it must have been computed later, e.g. when inserting the favicon.  Curious.  Will verify that the certificates are OK before merging.

rs/sns_aggregator/src/lib.rs
peterpeterparker: Thanks a lot for the explanation!!

rs/sns_aggregator/src/lib.rs
peterpeterparker: Works for me.

peterpeterparker (approved):
LGTM, thx

PR #1823, opened 2023-02-06
Correct setup for HW version test in neurons.services.spec.ts

dskloetd:
# Motivation

In https://github.com/dfinity/nns-dapp/pull/1815 I discovered that two tests don't actually test what they're supposed to test. I didn't want to hide the fixes between many other changes so I made a separate PR.

Without the change, the error toast is actually about not being the controller of the neuron in one case, and about not merging neurons with the same ID in another case, rather than about the HW version being behind.

# Changes

Push the mock neurons to the store before calling the function under test.

Additional minor change: Move resetAccountIdentity() to beforeEach. If we rely on the tests to clean this up, then if one of those tests fails, then other tests start failing as well for confusing reasons.

# Tests

Changes are test-only. But https://github.com/dfinity/nns-dapp/pull/1815 will have additional changes to make the tests more robust.

### DIFF
--- a/frontend/src/tests/lib/services/neurons.services.spec.ts
+++ b/frontend/src/tests/lib/services/neurons.services.spec.ts
@@ -800,7 +800,10 @@ describe("neurons-services", () => {
           controller: smallerVersionIdentity.getPrincipal().toText(),
         },
       };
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!

PR #1824, opened 2023-02-06
fix: getOldestTxIdFromStore cannot read properties of undefined

peterpeterparker:
# Motivation

`getOldestTxIdFromStore` does not handle empty transactions which can lead to an "Cannot read properties of undefined (reading 'id')" issue.

# How to reproduce

Accounts -> SNS-1 -> Wallet -> No transactions -> Back -> Wallet -> Issue.

# Changes

- `accountData.transactions` can be empty => `accountData.transactions[0]?.id` instead of `accountData.transactions[0].id`

# Screenshots

Mainnet:

Capture d’écran 2023-02-06 à 14 56 04

Reproduced locally:

Capture d’écran 2023-02-06 à 14 50 31


lmuntaner (approved):
Thanks!

PR #1825, opened 2023-02-06
bump dfx version to 0.13.0-beta.3 and update ubuntu version in Dockerfile

mraszyk:
Bump dfx version to have the latest dfx features, e.g., dfx sns download and creating canisters via boundary nodes on testnets, available in SNS end-to-end tests.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -1,5 +1,5 @@
 {
-  "dfx": "0.12.1",
+  "dfx": "0.13.0-beta.3",
   "canisters": {
     "nns-governance": {
       "type": "custom",
--- a/dfx.json
+++ b/dfx.json
@@ -1,5 +1,5 @@
 {
-  "dfx": "0.13.0-beta.3",
+  "dfx": "0.13.0",
   "canisters": {
     "nns-governance": {
       "type": "custom",
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -7,7 +7,7 @@ concurrency:
   cancel-in-progress: true
 jobs:
   curl_test:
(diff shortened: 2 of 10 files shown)


lmuntaner (commented):
Aren't there any breaking changes?

Did you try deploying to a testnet and still works?

mraszyk:
> Aren't there any breaking changes?
> 
> Did you try deploying to a testnet and still works?

Looks good to me: https://qsgjb-riaaa-aaaaa-aaaga-cai.large05.testnet.dfinity.network/

dfx.json
```diff
@@ -1,5 +1,5 @@
 {
-  "dfx": "0.12.1",
```
peterpeterparker, line 2: What are the advantages or requirement to use dfx beta?

mraszyk: actually dfx is releasing soon, so I can wait a bit and update to 0.13.0 later today

mraszyk: the main motivation was to not switch between 0.12.1 and 0.13 in sns end-to-end tests (that need the latest features not available in 0.13)

peterpeterparker: That would be appreciated from my side if we can stick to official releases.

I work locally with NNS-dapp but also in other projects. 

dfx has currently no easy way to switch between versions - except reinstalling the all dfx - and I would be happy to not bump all my projects to beta.

i.e. by sticking to official releases my workflow is more smooth

That said, if there are advantages or features that are required in the beta for what you or we aim to develop, for sure this should not be a blocker.

mraszyk: I just got a confirmation that 0.13.0 releases today so I'll wait until then.

PR #1828, opened 2023-02-06
Ckbtc config

bitdivine:
# Motivation
The nns-app will be supporting ckbtc, so needs the ckbtc canister IDs.

At first we need just the index and ledger canisters, which appears to be easy because they appear to be standard icrc1 canisters.

# Changes
- Add ckbtc index and ledger canisters to dfx.json.  Source: https://forum.dfinity.org/t/ckbtc-a-canister-issued-bitcoin-twin-token-on-the-ic-1-1-backed-by-btc/17606/46?u=bitdivine
- Add code to get the ID and provided it in the standard ways to builds. 

# Tests
- I have run:
  `DFX_NETWORK=mainnet ./config.sh` and verified that it returns the values in `dfx.json` and in Manu's post.

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -65,6 +65,14 @@ local_deployment_data="$(
   export SNS_AGGREGATOR_URL
   test -n "${SNS_AGGREGATOR_URL:-}" || unset SNS_AGGREGATOR_URL
 
+  : "Try to find the ckBTC canister IDs"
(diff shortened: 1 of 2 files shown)


peterpeterparker (approved):
LGTM, thanks.

I'll provide another PR afterwards to link the IDs in JavaScript code and to activate the feature flag.

PR #1830, opened 2023-02-06
build: enable ckBTC ledger and index for mainnet

peterpeterparker:
# Motivation

Enable ckBTC ledger and index on mainnet and propose it in next version.

# Changes

- enable ckBTC on **mainnet**
- enable ckBTC per default
- integrate ckBTC canister IDs configuration and fallback to mainnet canister IDs
- update order of fields in script
- add a new utils to test if a string is not null, not undefined and not empty

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -122,8 +122,8 @@ featureFlags=$(echo "$json" | jq -r ".FEATURE_FLAGS" | jq tostring)
 host=$(echo "$json" | jq -r ".HOST")
 identityServiceUrl=$(echo "$json" | jq -r ".IDENTITY_SERVICE_URL")
 aggregatorCanisterUrl=$(echo "$json" | jq -r '.SNS_AGGREGATOR_URL // ""')
-ckbtcIndexCanisterId=$(echo "$json" | jq -r '.CKBTC_INDEX_CANISTER_ID // ""')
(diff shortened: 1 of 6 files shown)


config.sh
```diff
@@ -139,8 +139,8 @@ VITE_FEATURE_FLAGS=$featureFlags
 VITE_HOST=$host
 VITE_IDENTITY_SERVICE_URL=$identityServiceUrl
 VITE_AGGREGATOR_CANISTER_URL=${aggregatorCanisterUrl:-}
-VITE_CKBTC_INDEX_CANISTER_ID=${ckbtcIndexCanisterId:-}
-VITE_CKBTC_LEDGER_CANISTER_ID=${ckbtcLedgerCanisterId:-}" | tee "$ENV_FILE"
+VITE_CKBTC_LEDGER_CANISTER_ID=${ckbtcLedgerCanisterId:-}
```
lmuntaner, line 142: Why did yo change the order?

lmuntaner (approved):
LGTM!

config.sh
peterpeterparker: For readability consistency reason. Everywhere we use ledger then index except here where it was index then ledger

peterpeterparker: Thanks for the review!

PR #1832, opened 2023-02-07
GIX-1187: Import * as api instead of mocking individual functions.

dskloetd:
# Motivation

This makes it easier to mock individual functions on the governance api.

# Changes

Import the governance API with `* as api` instead of all the functions into the global namespace.

# Tests

test only change


lmuntaner (approved):
Thanks!

PR #1834, opened 2023-02-07
fix: force destroy wallet page

peterpeterparker:
# Motivation

For the same SvelteKit issue as for the "Login" page we noticed that on destroy of the CkBTCWallet page was not called. 
This fixes the issue for that particular page. We might want to have  a look to apply it to all routes later.

# Issues

SvelteKit issue https://github.com/sveltejs/kit/issues/5434

# Changes

- destroy layout and all children on navigation to "force" the destroy of the components


mstrasinskis (approved):
LGTM, thanks!

PR #1837, opened 2023-02-07
GIX-1187: Cache neurons response for 5 minutes in neurons api service

dskloetd:
# Motivation

We want to reduce the load on the IC.
So we cache the response of certified queries (updates) for neurons.
This first implementation is as simple as possible with the following decisions:
* We only cache 1 list of neurons at a time for a single identity.
* We only cache certified responses.
* If we have a cached response, we also use it to respond to uncertified requests.
* We only cache resolved results and not in-flight promises.
* We only cache successful results and not errors.

# Changes

Implement the neurons cache.

# Tests

Extensive tests in `neurons.api-service.spec.ts`.

### DIFF
--- a/frontend/src/lib/api-services/neurons.api-service.ts
+++ b/frontend/src/lib/api-services/neurons.api-service.ts
@@ -34,6 +34,51 @@ import {
   type ApiStakeMaturityParams,
   type ApiStakeNeuronParams,
 } from "$lib/api/governance.api";
+import type { Identity } from "@dfinity/agent";
+import type { NeuronInfo } from "@dfinity/nns";
+
+const cacheExpirationDurationMs = 5 * 60 * 1000;
+
+const nowMs = () => new Date().getTime();
+
+let cachedNeurons: NeuronInfo[] | null = null;
+
+// When the neurons were last cached.
+let cacheTimestampMs: number | null = null;
+
+// The principal of the identity for which the neurons were cached.
+let cachePrincipalText: string | null = null;
+
+export const clearCache = () => {
+  cachedNeurons = null;
+  cacheTimestampMs = null;
+  cachePrincipalText = null;
+};
+
+const hasValidCachedNeurons = (identity: Identity) => {
+  if (
+    cachedNeurons === null ||
+    cacheTimestampMs === null ||
+    cachePrincipalText === null
+  ) {
+    return false;
+  }
+  if (nowMs() - cacheTimestampMs > cacheExpirationDurationMs) {
+    clearCache();
+    return false;
+  }
+  return cachePrincipalText === identity.getPrincipal().toText();
+};
+
+const clearCacheAfter = async <R>(promise: Promise<R>) => {
+  const result: R = await promise;
+  clearCache();
+  return result;
+};
+
+export const resetNeuronsApiService = () => {
+  clearCache();
+};
 
 export const neuronsApiService = {
   // Read calls
@@ -43,61 +88,71 @@ export const neuronsApiService = {
   queryNeuron(params: ApiQueryNeuronParams) {
     return queryNeuron(params);
   },
-  queryNeurons(params: ApiQueryParams) {
-    return queryNeurons(params);
+  async queryNeurons(params: ApiQueryParams) {
+    if (hasValidCachedNeurons(params.identity)) {
+      return cachedNeurons;
+    }
+    const promise = queryNeurons(params);
+    if (!params.certified) {
+      return promise;
+    }
+    cachedNeurons = await promise;
+    cacheTimestampMs = nowMs();
+    cachePrincipalText = params.identity.getPrincipal().toText();
+    return cachedNeurons;
   },
 
   // Action calls
-  addHotkey(params: ApiManageHotkeyParams) {
-    return addHotkey(params);
+  async addHotkey(params: ApiManageHotkeyParams) {
+    return clearCacheAfter(addHotkey(params));
   },
   autoStakeMaturity(params: ApiAutoStakeMaturityParams) {
-    return autoStakeMaturity(params);
+    return clearCacheAfter(autoStakeMaturity(params));
   },
   claimOrRefreshNeuron(params: ApiManageNeuronParams) {
-    return claimOrRefreshNeuron(params);
+    return clearCacheAfter(claimOrRefreshNeuron(params));
   },
   disburse(params: ApiDisburseParams) {
-    return disburse(params);
+    return clearCacheAfter(disburse(params));
   },
   increaseDissolveDelay(params: ApiIncreaseDissolveDelayParams) {
-    return increaseDissolveDelay(params);
+    return clearCacheAfter(increaseDissolveDelay(params));
   },
   joinCommunityFund(params: ApiManageNeuronParams) {
-    return joinCommunityFund(params);
+    return clearCacheAfter(joinCommunityFund(params));
   },
   leaveCommunityFund(params: ApiManageNeuronParams) {
-    return leaveCommunityFund(params);
+    return clearCacheAfter(leaveCommunityFund(params));
   },
   // @deprecated
   mergeMaturity(params: ApiMergeMaturityParams) {
-    return mergeMaturity(params);
+    return clearCacheAfter(mergeMaturity(params));
   },
   mergeNeurons(params: ApiMergeNeuronsParams) {
-    return mergeNeurons(params);
+    return clearCacheAfter(mergeNeurons(params));
   },
   removeHotkey(params: ApiManageHotkeyParams) {
-    return removeHotkey(params);
+    return clearCacheAfter(removeHotkey(params));
   },
   setFollowees(params: ApiSetFolloweesParams) {
-    return setFollowees(params);
+    return clearCacheAfter(setFollowees(params));
   },
   spawnNeuron(params: ApiSpawnNeuronParams) {
-    return spawnNeuron(params);
+    return clearCacheAfter(spawnNeuron(params));
   },
   splitNeuron(params: ApiSplitNeuronParams) {
-    return splitNeuron(params);
+    return clearCacheAfter(splitNeuron(params));
   },
   stakeMaturity(params: ApiStakeMaturityParams) {
-    return stakeMaturity(params);
+    return clearCacheAfter(stakeMaturity(params));
   },
   stakeNeuron(params: ApiStakeNeuronParams) {
-    return stakeNeuron(params);
+    return clearCacheAfter(stakeNeuron(params));
   },
   startDissolving(params: ApiManageNeuronParams) {
-    return startDissolving(params);
+    return clearCacheAfter(startDissolving(params));
   },
   stopDissolving(params: ApiManageNeuronParams) {
-    return stopDissolving(params);
+    return clearCacheAfter(stopDissolving(params));
   },
 };
--- a/frontend/src/lib/api-services/neurons.api-service.ts
+++ b/frontend/src/lib/api-services/neurons.api-service.ts
@@ -35,44 +35,42 @@ import {
   type ApiStakeNeuronParams,
 } from "$lib/api/governance.api";
 import { SECONDS_IN_MINUTE } from "$lib/constants/constants";
+import { nowInSeconds } from "$lib/utils/date.utils";
 import { isNullish } from "$lib/utils/utils";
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Already approving, just a couple of comments, but I like it. Simple enough.

I like the tests that will allow us to easily change the implementation if needed.

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -34,6 +34,51 @@ import {
...
   type ApiStakeNeuronParams,
 } from "$lib/api/governance.api";
+import type { Identity } from "@dfinity/agent";
+import type { NeuronInfo } from "@dfinity/nns";
+
+const cacheExpirationDurationMs = 5 * 60 * 1000;
+
+const nowMs = () => new Date().getTime();
+
+let cachedNeurons: NeuronInfo[] | null = null;
```
peterpeterparker, line 44: generally we use `undefined` when a state is not yet defined and we use `null` to express the fact that the system explicitely set such a state

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -34,6 +34,51 @@ import {
...
+
+export const clearCache = () => {
+  cachedNeurons = null;
+  cacheTimestampMs = null;
+  cachePrincipalText = null;
+};
+
+const hasValidCachedNeurons = (identity: Identity) => {
+  if (
+    cachedNeurons === null ||
```
peterpeterparker, line 60: use utils `isNullish` for any comparison with null or undefined

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -34,6 +34,51 @@ import {
...
+    cacheTimestampMs === null ||
+    cachePrincipalText === null
+  ) {
+    return false;
+  }
+  if (nowMs() - cacheTimestampMs > cacheExpirationDurationMs) {
```
peterpeterparker, line 66: we have already an existing utils to get the current time

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -34,6 +34,51 @@ import {
...
+    clearCache();
```
peterpeterparker, line 67: the function is named `hasValidCachedNeurons` but does some logic - i.e. it clears the cache. can you rename the function?

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -34,6 +34,51 @@ import {
...
+// The principal of the identity for which the neurons were cached.
+let cachePrincipalText: string | null = null;
+
+export const clearCache = () => {
+  cachedNeurons = null;
+  cacheTimestampMs = null;
+  cachePrincipalText = null;
+};
+
+const hasValidCachedNeurons = (identity: Identity) => {
```
peterpeterparker, line 58: the return type is not typed

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -34,6 +34,51 @@ import {
...
+  }
+  if (nowMs() - cacheTimestampMs > cacheExpirationDurationMs) {
+    clearCache();
+    return false;
+  }
+  return cachePrincipalText === identity.getPrincipal().toText();
+};
+
+const clearCacheAfter = async <R>(promise: Promise<R>) => {
+  const result: R = await promise;
```
peterpeterparker, line 74: - not sure why the promise is not called?

`await promise` vs `await promise()`

- I'm confuse about the naming of the function - `clearCacheAfter` - it clears cache after or it calls a promise and clear cache?

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -34,6 +34,51 @@ import {
...
+  clearCache();
+  return result;
+};
+
+export const resetNeuronsApiService = () => {
```
peterpeterparker, line 79: is this only meant for test? if yes please document it explicitely

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -34,6 +34,51 @@ import {
   type ApiStakeMaturityParams,
   type ApiStakeNeuronParams,
 } from "$lib/api/governance.api";
+import type { Identity } from "@dfinity/agent";
+import type { NeuronInfo } from "@dfinity/nns";
+
+const cacheExpirationDurationMs = 5 * 60 * 1000;
```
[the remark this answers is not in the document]
peterpeterparker: May I ask why 5 min? Just some random number or something related to the fact that a session is 30 min ?

Speaking of, should it be a fix number or a percentage of the session time and yes, can you use the session duration constant to set it?

peterpeterparker (changes requested):
Few questions. Blocking only because I noticed the PR was already approved.

dskloetd (commented):
I've replied to/addressed the comments.
In some cases with general questions.
I'm still trying to understand the code style so please be gentle :-)

frontend/src/lib/api-services/neurons.api-service.ts
dskloetd: Done

frontend/src/lib/api-services/neurons.api-service.ts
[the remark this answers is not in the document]
peterpeterparker: > Why do you think it should be set as a percentage of the session time?

We do not know anything about the behavior of the user, therefore 5min is just a random number. Session duration on the contrary is given.

`AUTH_SESSION_DURATION`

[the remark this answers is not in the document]
peterpeterparker: But totally fine with any number, was meant as questions

frontend/src/lib/api-services/neurons.api-service.ts
peterpeterparker: > We have cached neurons
> We don't have cached neurons

So, why `null` and not just an array?

```
let cachedNeurons: NeuronInfo[]  = [];
```

frontend/src/lib/api-services/neurons.api-service.ts
peterpeterparker: `nowInSeconds` or `nowInBigIntNanoSeconds`. you can move your utils there if you wish.

frontend/src/lib/api-services/neurons.api-service.ts
peterpeterparker: > Done, but isn't it inferred?

Yes but when you code your function itself it do not have the return time yet so you won't have type support for the return in your editor until you finish to implement the function.

> Or is it a manual job to make sure we specify types everywhere?

Yes manually. Readability reason mostly and DX.

Thanks

frontend/src/lib/api-services/neurons.api-service.ts
dskloetd: Because an empty array indicates that we have a cached results that there are no neurons.
But we want to indicate that we don't have a cached result at all.
Granted this is also indicated by the timestamp and/or the principal being null but it seems best to keep it all consistent.

I have no put all 3 fields together in 1 type. So our reference to this object can be null but not any of the individual fields. What do you think?

frontend/src/lib/api-services/neurons.api-service.ts
dskloetd: Thanks. I've switched to using seconds instead of ms and am now using nowInSeconds.

frontend/src/lib/api-services/neurons.api-service.ts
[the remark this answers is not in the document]
dskloetd: The session time is given but also kind of arbitrary and not really related.
And the percentage itself would be completely random as well.
I asked Llorenç and he says it's up to us to pick a duration.
So while I don't really have a good answer, I'll leave it like this.

frontend/src/lib/api-services/neurons.api-service.ts
dskloetd: Thanks. Do we have written down rules for when we add them and when not? Or is it a judgement call?
`let i = 5;` instead of `let i : number = 5` would be fine, right?

peterpeterparker: we use `let i = 5`.

@lmuntaner has documented some rules in Notion and eslint valid also the typings. there are some eslint rules we did not have adusted or implemented yet for time reason.

That said, I personally think the code describe itself - i.e. we have been relatively well consistant and checking out the rest of the code describes the code style.

frontend/src/lib/api-services/neurons.api-service.ts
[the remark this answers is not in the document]
peterpeterparker: Works for me, thanks

frontend/src/lib/api-services/neurons.api-service.ts
peterpeterparker: I was also asking myself why you used three objects so that would works for me, thanks

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -43,61 +92,73 @@ export const neuronsApiService = {
   queryNeuron(params: ApiQueryNeuronParams) {
     return queryNeuron(params);
   },
-  queryNeurons(params: ApiQueryParams) {
-    return queryNeurons(params);
+  async queryNeurons(params: ApiQueryParams) {
+    if (!isNullish(neuronsCache) && hasValidCachedNeurons(params.identity)) {
```
peterpeterparker, line 96: just fyi, there is also a utils named `nonNullish` that relates to `!isNullish`

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -43,61 +92,73 @@ export const neuronsApiService = {
...
+      return neuronsCache.neurons;
+    }
+    const promise = queryNeurons(params);
+    if (!params.certified) {
```
peterpeterparker, line 100: is this correct? doesn't it mean that if non-certified data are requested no cache will be ever be used? or maybe that's the goal?

frontend/src/lib/api-services/neurons.api-service.ts
```diff
@@ -43,61 +92,73 @@ export const neuronsApiService = {
...
+      return promise;
+    }
+    neuronsCache = {
+      neurons: await promise,
+      timestampSeconds: nowInSeconds(),
+      principalText: params.identity.getPrincipal().toText(),
```
peterpeterparker, line 106: not sure this works, maybe I understand it wrong but does this mean that we maintain one single cache for one single principal, is that correct?

shouldn't we maintain a cache per identity?

```
type PrincipalAsText = string;
type NeuronsCache = Record<PrincipalAsText, {
   timestampSeconds: number;
   neurons: NeuronInfo[];
}
```

But maybe I am missing something, if that's the case sorry for the questino.

frontend/src/lib/api-services/neurons.api-service.ts
dskloetd: Done

frontend/src/lib/api-services/neurons.api-service.ts
dskloetd: As written in PR descriptiong:
* We only cache certified responses.
* If we have a cached response, we also use it to respond to uncertified requests.

frontend/src/lib/api-services/neurons.api-service.ts
dskloetd: As written in the PR description:

This first implementation is as simple as possible with the following decisions:
* We only cache 1 list of neurons at a time for a single identity.
* ...

My assumption is that most people are not constantly switching between identities so this simplification should still catch most cases.

peterpeterparker: > My assumption is that most people are not constantly switching between identities so this simplification should still catch most cases.

In that sense yes but, it also implies that to some extension when we implement functionnalies that might use various identities (anonymous or signed in), we have to remember to think at the timeline of each calls for optimal performance.

Anyhow if it decided that way, so be it for sure. Thanks for the feedback.

peterpeterparker (approved):
LGTM, thx

PR #1838, opened 2023-02-08
Comments in aggregator code + cleanup

bitdivine:
# Motivation
The aggregator code is largely uncommented.  The code is not stable but a lot of the comments can be filled in already.

# Changes
* Add a clippy for undocumented code.
* Exempt autogenerated code from the clippy.
* Document code
* Minor cleanup

# Tests
There is no change in functionality.  Tests are sorely needed, but not in this PR.

### DIFF
--- a/rs/sns_aggregator/src/lib.rs
+++ b/rs/sns_aggregator/src/lib.rs
@@ -1,5 +1,6 @@
 //! Entry points for the caching canister.
 #![warn(missing_docs)]
+#![warn(clippy::missing_docs_in_private_items)]
 pub mod assets;
 mod conversion;
 mod state;
@@ -37,6 +38,17 @@ fn http_request(/* req: HttpRequest */) /* -> HttpResponse */
     call::reply((response,));
 }
 
+/// Function called when a canister is first created IF it is created
+/// with this code.
+///
+/// Note: If the canister os created with e.g. `dfx canister create`
+///       and then deployed normally, `init(..)` is never called.
+#[ic_cdk_macros::init]
+fn init() {
+    insert_favicon();
+}
+
+/// Function called before upgrade to a new wasm.
 #[ic_cdk_macros::post_upgrade]
 fn post_upgrade() {
     // Browsers complain if they don't get pretty pictures.  So do I.
@@ -49,9 +61,4 @@ fn post_upgrade() {
         Duration::from_secs(1),
         || ic_cdk::spawn(crate::upstream::update_cache()),
     );
-}
-
-#[ic_cdk_macros::init]
-fn init() {
-    insert_favicon();
 }
--- a/rs/sns_aggregator/src/state.rs
+++ b/rs/sns_aggregator/src/state.rs
@@ -1,3 +1,4 @@
+//! The state of the canister
 use crate::assets::{insert_asset, Asset};
 use crate::convert_canister_id;
 use crate::types::slow::logo_binary;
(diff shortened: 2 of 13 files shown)


rs/sns_aggregator/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 //! Entry points for the caching canister.
 #![warn(missing_docs)]
+#![warn(clippy::missing_docs_in_private_items)]
```
bitdivine, line 3: This is the clippy that causes warnings for undocumented items.  Most types here are somehow private so the normal missing_docs clippy isn't enough.

rs/sns_aggregator/src/lib.rs
```diff
@@ -37,6 +38,17 @@ fn http_request(/* req: HttpRequest */) /* -> HttpResponse */
     call::reply((response,));
 }
```
bitdivine, line 40: Code moved up - init comes before upgrades - and commented but otherwise unchanged.

rs/sns_aggregator/src/state.rs
```diff
@@ -34,6 +35,13 @@ thread_local! {
 }
 
 impl State {
```
bitdivine, line 37: This is the constant we discussed before.  I have moved it here and documented what it does.

rs/sns_aggregator/src/types/ic_sns_governance.rs
```diff
@@ -1,7 +1,8 @@
 #![cfg_attr(rustfmt, rustfmt_skip)]
 #![allow(clippy::all)]
+#![allow(clippy::missing_docs_in_private_items)]
```
bitdivine, line 3: auto-generated code is not expected to be documented.  Furthermore we are just passing these types through from upstream to the consumer.  The upstream should have the documentation.

rs/sns_aggregator/src/types/state.rs
```diff
@@ -47,43 +49,3 @@ pub struct UpstreamData {
     /// The ledger fee, presumably a transaction fee.
     pub icrc1_fee: Nat,
 }
```
bitdivine, line 51: These types below were no longer used but had not been deleted.

scripts/did2rs.sh
```diff
@@ -58,7 +58,9 @@ cd "$GIT_ROOT"
   cat <<-EOF
 	#![cfg_attr(rustfmt, rustfmt_skip)]
 	#![allow(clippy::all)]
+	#![allow(clippy::missing_docs_in_private_items)]
 	#![allow(non_camel_case_types)]
+	#![allow(dead_code)]
```
bitdivine, line 63: The dead code annotation is already in the generated rust but presumably the line here that generates it had not been committed.

rs/sns_aggregator/src/lib.rs
```diff
@@ -37,6 +38,17 @@ fn http_request(/* req: HttpRequest */) /* -> HttpResponse */
...
 
+/// Function called when a canister is first created IF it is created
+/// with this code.
+///
+/// Note: If the canister os created with e.g. `dfx canister create`
+///       and then deployed normally, `init(..)` is never called.
```
peterpeterparker, line 45: TIL. With "deployed normally" you mean `dfx deploy ...`?

peterpeterparker (approved):
LGTM, thanks

rs/sns_aggregator/src/lib.rs
bitdivine: Yes, good point.  I have clarified that.

PR #1844, opened 2023-02-08
bump dfx to 0.13.0-downgrade.1

mraszyk:
dfx 0.13.0 is needed to be able to create canisters via BNs in SNS end-to-end tests

The change in the CI pipeline is also necessary because sns-demo installed in the previous step overrides dfx according to its dfx.json to 0.12.1


bitdivine:
Checking whether we can just bump snsdemo...
- [ ] local testnet
- [ ] remote testnet

bitdivine (approved):

PR #1840, opened 2023-02-08
this is needed to build the nns-dapp with docker on ephemeral dfx networks

mraszyk:
I have to pass the (so far) hidden `.dfx` directory to the docker build so that the canister IDs are identified by `config.sh` for ephemeral networks (e.g., "local", "https://...") in the following build runbook:
```
dfx canister create nns-dapp --network "${NETWORK}" --no-wallet
OWN_CANISTER_ID=$(dfx canister id nns-dapp --network "${NETWORK}")
export OWN_CANISTER_ID
DOCKER_BUILDKIT=1 docker build --target scratch --build-arg DFX_NETWORK="${DFX_NETWORK}" --build-arg OWN_CANISTER_ID="${OWN_CANISTER_ID}" -t "nns-dapp" -o "out" .
dfx canister install nns-dapp --network "${NETWORK}" --wasm out/nns-dapp.wasm
```

PR #1845, opened 2023-02-08
alternative local docker build of nns-dapp

mraszyk:
An alternative to #1840 to make config.sh identify the nns-dapp canister id for the "local" network in the following build runbook:
```
dfx canister create nns-dapp --no-wallet
export OWN_CANISTER_ID=$(dfx canister id nns-dapp)
DOCKER_BUILDKIT=1 docker build --target scratch --build-arg DFX_NETWORK="local" --build-arg OWN_CANISTER_ID="${OWN_CANISTER_ID}" -t "nns-dapp" -o "out" .
dfx canister install nns-dapp --wasm out/nns-dapp.wasm
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,9 +69,6 @@ jobs:
       - name: Install SNS script dependencies
         run: |
           dfx-sns-demo-install
-      - name: Update dfx
-        run: |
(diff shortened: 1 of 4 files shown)


mraszyk:
Not a good idea as one needs to clean up the wallet from the persistent wallet store before redeploying the local network.

PR #1846, opened 2023-02-09
Fix and document local development

dskloetd:
# Motivation

Our documentation about local development is missing some steps and the correct way to do things is scattered in people's heads and slack messages.

# Changes

Updated `HACKING.md` and added a script that follows the instructions to get local development running.

# Tests

Ran the script from a new user on my laptop.


bitdivine (approved):
Thanks!

HACKING.md
```diff
@@ -11,17 +11,29 @@ That is why we are providing a `./config.sh` script that generate the above envi
...
+To run the dapp against canisters deployed locally on a simulated IC network, use the steps below, or run `./scripts/dev-local.sh` which guides you through these steps
 
+- Make sure you have a clean local replica running with `dfx start --clean`. This will stay running so use a separate terminal for this.
 - Deploy the Nns backend canisters locally with `dfx nns install`
-- Once deployed, the canister IDs should be collected and the "local" network of `canister_ids.json` (**TODO**) should be updated accordingly manually
-- Run `DFX_NETWORK=local ./config.sh` to populate the `.env` file (**TODO**)
-- Start `npm run dev` in the `./frontend/` folder to serve the application
+- From the last line of output of `dfx nns install` note down the value url for `nns-dapp`
+- Run `DFX_NETWORK=local ./config.sh` to populate the `./frontend/.env` file.
+- Manually edit the `./frontend/.env` and replace `null` with the nns-dapp canister id from the url you noted down before.
```
peterpeterparker, line 20: do you know if we can improve this step? not in this PR, just generally speaking if we can spare this that's cool I think

peterpeterparker (approved):
Cool, thanks!

HACKING.md
dskloetd: I tried looking at `config.sh` to see if maybe it used to work or there is a bug that can be fixed to get the OWN_CANISTER_ID set, but I didn't see anything obvious. But `config.sh` is quite unreadable to be honest.

But I believe the value is actually always equal to "qsgjb-riaaa-aaaaa-aaaga-cai". If that's true, we could hard code it in config.sh in skip this step. But I'm not sure if that would be better?

peterpeterparker: > But I believe the value is actually always equal to "qsgjb-riaaa-aaaaa-aaaga-cai". If that's true, we could hard code it in config.sh in skip this step. But I'm not sure if that would be better?

Not sure. Technically yes, canister id are predictable. If always installed in the same order, canisters always get same ids but in case of NNS-dapp, not sure we always install  in the same order locally?

Also not sure local get's the same canister id as on mainnet and testnet.

peterpeterparker: Anyhow not related to your PR directly, thanks for the doc improvement

dskloetd: I would guess `dfx nns install` always installs in the same order, and it requires that it is the first thing that happens after `dfx start --clean`.

mainnet and testnet are likely different but this script is specifically for local development.

PR #1847, opened 2023-02-09
GIX-1187: List neurons per page

lmuntaner:
# Motivation

Reduce number of update calls.

Do not load neurons unless they will be used. To accomplish that, we load the neurons in the pages they are needed.

# Changes

* Remove `listNeurons` from `app.services.ts`.
* Add `listNeurons` call in NnsNeurons.svelte.
* Add `listNeurons` call in NnsProposals.svelte if user is signed in.
* Add `listNeurons` call in NnsProposalDetail.svelte if user is signed in.

Minor change:
* `listNeurons` relied on the order of the responses to set the neurons. I changed that.

# Tests

* Test new functionality in the page tests.
* Mock import in the routes tests.
* Test new functionality in the workflows NnsProposalDetail and NnsProposals.
* Test change in `listNeurons` service.

### DIFF
--- a/frontend/src/lib/pages/NnsNeurons.svelte
+++ b/frontend/src/lib/pages/NnsNeurons.svelte
@@ -10,11 +10,15 @@
   import { pageStore } from "$lib/derived/page.derived";
   import { buildNeuronUrl } from "$lib/utils/navigation.utils";
   import EmptyMessage from "$lib/components/ui/EmptyMessage.svelte";
-
-  // Neurons are fetch on page load. No need to do it in the route.
+  import { onMount } from "svelte";
+  import { listNeurons } from "$lib/services/neurons.services";
 
   let isLoading = false;
   $: isLoading = $neuronsStore.neurons === undefined;
+
+  onMount(() => {
+    listNeurons();
+  });
 
   const goToNeuronDetails = async (id: NeuronId) =>
     await goto(
--- a/frontend/src/lib/services/neurons.services.ts
+++ b/frontend/src/lib/services/neurons.services.ts
@@ -249,15 +249,15 @@ export const listNeurons = async ({
     request: ({ certified, identity }) =>
       neuronsApiService.queryNeurons({ certified, identity }),
     onLoad: async ({ response: neurons, certified }) => {
-      // Skip if we already have certified neurons
-      // Otherwise we depend on the order of the responses
-      if (certifiedNeuronsSet) {
+      if (!certified && certifiedNeuronsSet) {
+        // Skip if we already have certified neurons
+        // Otherwise we depend on the order of the responses
         return;
       }
       if (certified) {
         certifiedNeuronsSet = true;
-        neuronsStore.setNeurons({ neurons, certified });
       }
+      neuronsStore.setNeurons({ neurons, certified });
 
       callback?.(certified);
     },
--- a/frontend/src/lib/services/neurons.services.ts
+++ b/frontend/src/lib/services/neurons.services.ts
@@ -243,12 +243,21 @@ export const listNeurons = async ({
   callback?: (certified: boolean) => void;
   strategy?: QueryAndUpdateStrategy;
 } = {}): Promise<void> => {
+  let certifiedNeuronsSet = false;
   return queryAndUpdate<NeuronInfo[], unknown>({
     strategy,
     request: ({ certified, identity }) =>
       neuronsApiService.queryNeurons({ certified, identity }),
     onLoad: async ({ response: neurons, certified }) => {
-      neuronsStore.setNeurons({ neurons, certified });
+      // Skip if we already have certified neurons
+      // Otherwise we depend on the order of the responses
+      if (certifiedNeuronsSet) {
+        return;
+      }
+      if (certified) {
+        certifiedNeuronsSet = true;
+        neuronsStore.setNeurons({ neurons, certified });
+      }
 
       callback?.(certified);
     },
--- a/frontend/src/lib/services/neurons.services.ts
+++ b/frontend/src/lib/services/neurons.services.ts
@@ -243,20 +243,11 @@ export const listNeurons = async ({
   callback?: (certified: boolean) => void;
   strategy?: QueryAndUpdateStrategy;
 } = {}): Promise<void> => {
-  let certifiedNeuronsSet = false;
   return queryAndUpdate<NeuronInfo[], unknown>({
     strategy,
     request: ({ certified, identity }) =>
       neuronsApiService.queryNeurons({ certified, identity }),
     onLoad: async ({ response: neurons, certified }) => {
-      if (!certified && certifiedNeuronsSet) {
-        // Skip if we already have certified neurons
-        // Otherwise we depend on the order of the responses
-        return;
-      }
-      if (certified) {
-        certifiedNeuronsSet = true;
-      }
       neuronsStore.setNeurons({ neurons, certified });
 
       callback?.(certified);
--- a/frontend/src/tests/lib/pages/NnsNeurons.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsNeurons.spec.ts
@@ -9,17 +9,20 @@ import * as authServices from "$lib/services/auth.services";
 import { neuronsStore } from "$lib/stores/neurons.store";
 import { NeuronState } from "@dfinity/nns";
 import { render, waitFor } from "@testing-library/svelte";
-import { mockGetIdentity, mockIdentity } from "../../mocks/auth.store.mock";
+import { mockIdentity } from "../../mocks/auth.store.mock";
 import en from "../../mocks/i18n.mock";
 import { mockFullNeuron, mockNeuron } from "../../mocks/neurons.mock";
 
 jest.mock("$lib/api/governance.api");
 
 describe("NnsNeurons", () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+    neuronsStore.reset();
+  });
+
   describe("with enough neurons", () => {
     beforeEach(() => {
-      jest.resetAllMocks();
-      neuronsStore.reset();
       resetNeuronsApiService();
       const mockNeuron2 = {
         ...mockNeuron,
@@ -36,7 +39,7 @@ describe("NnsNeurons", () => {
       };
       jest
         .spyOn(authServices, "getAuthenticatedIdentity")
-        .mockImplementation(() => Promise.resolve(mockGetIdentity()));
+        .mockResolvedValue(mockIdentity);
       jest
         .spyOn(api, "queryNeurons")
         .mockResolvedValue([mockNeuron, spawningNeuron, mockNeuron2]);
@@ -68,12 +71,10 @@ describe("NnsNeurons", () => {
 
   describe("no neurons", () => {
     beforeEach(() => {
-      jest.clearAllMocks();
-      neuronsStore.reset();
       resetNeuronsApiService();
       jest
         .spyOn(authServices, "getAuthenticatedIdentity")
-        .mockImplementation(() => Promise.resolve(mockGetIdentity()));
+        .mockResolvedValue(mockIdentity);
       jest.spyOn(api, "queryNeurons").mockResolvedValue([]);
     });
 
@@ -88,11 +89,9 @@ describe("NnsNeurons", () => {
 
   describe("navigating", () => {
     beforeEach(() => {
-      jest.clearAllMocks();
-      neuronsStore.reset();
       jest
         .spyOn(authServices, "getAuthenticatedIdentity")
-        .mockImplementation(() => Promise.resolve(mockGetIdentity()));
+        .mockResolvedValue(mockIdentity);
       jest.spyOn(api, "queryNeurons").mockResolvedValue([]);
     });
 
--- a/frontend/src/tests/lib/pages/NnsNeurons.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsNeurons.spec.ts
@@ -19,12 +19,12 @@ jest.mock("$lib/api/governance.api");
 describe("NnsNeurons", () => {
   beforeEach(() => {
     jest.resetAllMocks();
+    resetNeuronsApiService();
     neuronsStore.reset();
   });
 
   describe("with enough neurons", () => {
     beforeEach(() => {
-      resetNeuronsApiService();
       const mockNeuron2 = {
         ...mockNeuron,
         neuronId: BigInt(223),
@@ -72,7 +72,6 @@ describe("NnsNeurons", () => {
 
   describe("no neurons", () => {
     beforeEach(() => {
-      resetNeuronsApiService();
       jest
         .spyOn(authServices, "getAuthenticatedIdentity")
         .mockResolvedValue(mockIdentity);
@@ -97,7 +96,6 @@ describe("NnsNeurons", () => {
     });
 
     it("should call query neurons twice when rendered", async () => {
-      resetNeuronsApiService();
       render(NnsNeurons);
 
       await waitFor(() =>
@@ -113,7 +111,6 @@ describe("NnsNeurons", () => {
     });
 
     it("should NOT call query neurons after being visited", async () => {
-      resetNeuronsApiService();
       render(NnsNeurons);
 
       await waitFor(() =>
--- a/frontend/src/tests/lib/pages/NnsNeurons.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsNeurons.spec.ts
@@ -2,37 +2,25 @@
  * @jest-environment jsdom
  */
 
+import { resetNeuronsApiService } from "$lib/api-services/neurons.api-service";
+import * as api from "$lib/api/governance.api";
 import NnsNeurons from "$lib/pages/NnsNeurons.svelte";
-import { authStore } from "$lib/stores/auth.store";
+import * as authServices from "$lib/services/auth.services";
 import { neuronsStore } from "$lib/stores/neurons.store";
 import { NeuronState } from "@dfinity/nns";
 import { render, waitFor } from "@testing-library/svelte";
-import { mockAuthStoreSubscribe } from "../../mocks/auth.store.mock";
+import { mockGetIdentity, mockIdentity } from "../../mocks/auth.store.mock";
 import en from "../../mocks/i18n.mock";
-import {
-  buildMockNeuronsStoreSubscribe,
-  mockFullNeuron,
-  mockNeuron,
-} from "../../mocks/neurons.mock";
+import { mockFullNeuron, mockNeuron } from "../../mocks/neurons.mock";
 
-jest.mock("$lib/services/neurons.services", () => {
-  return {
-    listNeurons: jest.fn().mockResolvedValue(undefined),
-  };
-});
+jest.mock("$lib/api/governance.api");
 
 describe("NnsNeurons", () => {
-  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  let authStoreMock: jest.MockedFunction<any>;
-
-  beforeEach(() => {
-    authStoreMock = jest
-      .spyOn(authStore, "subscribe")
-      .mockImplementation(mockAuthStoreSubscribe);
-  });
-
   describe("with enough neurons", () => {
     beforeEach(() => {
+      jest.resetAllMocks();
+      neuronsStore.reset();
+      resetNeuronsApiService();
       const mockNeuron2 = {
         ...mockNeuron,
         neuronId: BigInt(223),
@@ -40,38 +28,33 @@ describe("NnsNeurons", () => {
       const spawningNeuron = {
         ...mockNeuron,
         state: NeuronState.Spawning,
-        neuronId: BigInt(223),
+        neuronId: BigInt(224),
         fullNeuron: {
           ...mockFullNeuron,
           spawnAtTimesSeconds: BigInt(12312313),
         },
       };
       jest
-        .spyOn(neuronsStore, "subscribe")
-        .mockImplementation(
-          buildMockNeuronsStoreSubscribe([
-            mockNeuron,
-            mockNeuron2,
-            spawningNeuron,
-          ])
-        );
+        .spyOn(authServices, "getAuthenticatedIdentity")
+        .mockImplementation(() => Promise.resolve(mockGetIdentity()));
+      jest
+        .spyOn(api, "queryNeurons")
+        .mockResolvedValue([mockNeuron, spawningNeuron, mockNeuron2]);
     });
 
-    afterEach(() => jest.resetAllMocks());
-
-    it("should render spawning neurons as disabled", () => {
+    it("should render spawning neurons as disabled", async () => {
       const { queryAllByTestId } = render(NnsNeurons);
 
+      // Wait for the neurons to be loaded and rendered
+      await waitFor(() => {
+        const neuronCards = queryAllByTestId("neuron-card");
+        return expect(neuronCards.length).toBe(3);
+      });
       const neuronCards = queryAllByTestId("neuron-card");
       const disabledCards = neuronCards.filter(
         (card) => card.getAttribute("aria-disabled") === "true"
       );
       expect(disabledCards.length).toBe(1);
-    });
-
-    it("should subscribe to store", () => {
-      render(NnsNeurons);
-      expect(authStoreMock).toHaveBeenCalled();
     });
 
     it("should render a NeuronCard", async () => {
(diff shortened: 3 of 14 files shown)


lmuntaner:

frontend/src/lib/pages/NnsNeurons.svelte
```diff
@@ -10,12 +10,16 @@
...
   import EmptyMessage from "$lib/components/ui/EmptyMessage.svelte";
-
-  // Neurons are fetch on page load. No need to do it in the route.
+  import { onMount } from "svelte";
+  import { listNeurons } from "$lib/services/neurons.services";
 
   let isLoading = false;
   $: isLoading = $neuronsStore.neurons === undefined;
 
+  onMount(() => {
```
dskloetd, line 19: Why do you check isSignedIn in the other files but not here?
And why onMount() here but not in the other files?

frontend/src/lib/services/neurons.services.ts
```diff
@@ -243,11 +243,20 @@ export const listNeurons = async ({
   callback?: (certified: boolean) => void;
   strategy?: QueryAndUpdateStrategy;
 } = {}): Promise<void> => {
+  let certifiedNeuronsSet = false;
   return queryAndUpdate<NeuronInfo[], unknown>({
     strategy,
     request: ({ certified, identity }) =>
       neuronsApiService.queryNeurons({ certified, identity }),
     onLoad: async ({ response: neurons, certified }) => {
+      if (!certified && certifiedNeuronsSet) {
```
dskloetd, line 252: Aren't you duplicating logic from the api-service here?
The API service will use the cache (if present) for non-certified requests so this code here is redundant, right?

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
```diff
@@ -2,78 +2,61 @@
...
-        .spyOn(neuronsStore, "subscribe")
-        .mockImplementation(
-          buildMockNeuronsStoreSubscribe([
-            mockNeuron,
-            mockNeuron2,
-            spawningNeuron,
-          ])
-        );
+        .spyOn(authServices, "getAuthenticatedIdentity")
+        .mockImplementation(() => Promise.resolve(mockGetIdentity()));
```
dskloetd, line 39: Do you need a new mockGetIdentity() each time? Or could you use mockResolvedValue?
Also instead of `() => Promise.resolve(mockGetIdentity())` you could use `async () => mockGetIdentity()` if you think it's nicer.

dskloetd, line 39: Why does the test need mockGetIdentity() now but it didn't need it before?

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
```diff
@@ -84,16 +67,69 @@ describe("NnsNeurons", () => {
   });
 
   describe("no neurons", () => {
-    beforeAll(() => {
+    beforeEach(() => {
+      jest.clearAllMocks();
+      neuronsStore.reset();
+      resetNeuronsApiService();
```
dskloetd, line 73: Put all the shared setup in a separate `beforeEach` at the higher level.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
```diff
@@ -2,78 +2,61 @@
...
-
-    it("should render spawning neurons as disabled", () => {
+    it("should render spawning neurons as disabled", async () => {
       const { queryAllByTestId } = render(NnsNeurons);
 
+      // Wait for the neurons to be loaded and rendered
+      await waitFor(() => {
+        const neuronCards = queryAllByTestId("neuron-card");
+        return expect(neuronCards.length).toBe(3);
+      });
```
dskloetd, line 52: Why did you need this now and now before?

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
```diff
@@ -2,78 +2,61 @@
...
       const disabledCards = neuronCards.filter(
         (card) => card.getAttribute("aria-disabled") === "true"
       );
       expect(disabledCards.length).toBe(1);
     });
 
-    it("should subscribe to store", () => {
-      render(NnsNeurons);
-      expect(authStoreMock).toHaveBeenCalled();
-    });
```
dskloetd, line 75: Why remove this?

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
```diff
@@ -84,16 +67,69 @@ describe("NnsNeurons", () => {
...
+        .spyOn(authServices, "getAuthenticatedIdentity")
+        .mockImplementation(() => Promise.resolve(mockGetIdentity()));
+      jest.spyOn(api, "queryNeurons").mockResolvedValue([]);
+    });
+
+    it("should call query neurons twice when rendered", async () => {
+      resetNeuronsApiService();
+      render(NnsNeurons);
+
+      await waitFor(() =>
```
dskloetd, line 103: I read waitFor is a bad pattern. Is there something specific we can await?

frontend/src/tests/lib/pages/NnsProposalDetail.spec.ts
```diff
@@ -61,25 +65,58 @@ describe("ProposalDetail", () => {
...
+        certified: false,
+      });
+    });
   });
 
-  it("should render proposal detail if signed in", async () => {
-    authStoreMock.next({
-      identity: mockIdentity,
+  describe("logged out user", () => {
+    beforeEach(() => {
```
dskloetd, line 100: Put shared setup on the higher level?

frontend/src/tests/lib/pages/NnsProposals.spec.ts
```diff
@@ -63,6 +67,35 @@ describe("NnsProposals", () => {
 
     afterAll(() => jest.clearAllMocks());
 
+    describe("neurons", () => {
+      beforeEach(() => {
+        // TODO: Mock the canister call instead of the service
```
dskloetd, line 72: What is this referring to? Where is it mocking the service?

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -319,6 +316,15 @@ describe("neurons-services", () => {
   });
 
   describe("list neurons", () => {
+    const spyQueryNeurons = jest
+      .spyOn(api, "queryNeurons")
+      .mockImplementation(() => Promise.resolve(neurons));
```
dskloetd, line 321: mockResolvedValue?

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -319,6 +316,15 @@ describe("neurons-services", () => {
...
+
+    beforeEach(() => {
+      resetNeuronsApiService();
+      jest.clearAllMocks();
```
dskloetd, line 325: Put these at the highest level?

frontend/src/lib/pages/NnsNeurons.svelte
lmuntaner: `NnsNeurons` is rendered only when the user is logged in.

While proposals are also rendered when user is not logged in.

In the others, I want to call list neurons if the authStore changes, that's why I don't put it in the onMount.

Here, it's enough to load it when the component is rendered.

frontend/src/lib/services/neurons.services.ts
lmuntaner: Not exactly. The store does not separate data with certified and not certified.

```
{
  neurons: NeuronInfo[],
  certified: boolean,
}
```

Therefore, if the query was coming after the update, the `neuronStore.setData` was setting `certified` to `false` and changing the `neurons` for the uncertified ones.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: Same thing

lmuntaner: I was mocking `listNeurons`.

Now I'm mocking the api layer.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: No, I want it after each of the `it` tests also.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: Because I was filling the store and there was no need to wait for this. The first render already came with the cards.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: I don't understand why it was here, this is not done at this level.

Maybe it was happening in some child component and therefore it passed the test. But the NnsNeurons does not subscribe to the store.

frontend/src/tests/lib/pages/NnsProposalDetail.spec.ts
lmuntaner: I needed it being reset after each `it`

frontend/src/tests/lib/pages/NnsProposals.spec.ts
lmuntaner: Before, it was mocking the governanceCanister, but now we moved to a pattern to mock the api layer.

It's a little bit inconsistent like this. But I did't want to refactor much more.

frontend/src/tests/lib/services/neurons.services.spec.ts
lmuntaner: yes

frontend/src/tests/lib/services/neurons.services.spec.ts
lmuntaner: I needed them reset after each `it`

lmuntaner:

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
dskloetd: Sorry, maybe what I read was about `tick`. I thought it was in the Svelte tutorial but I can't find it now. Nevermind.

frontend/src/tests/lib/pages/NnsProposalDetail.spec.ts
dskloetd: After or before?
Either way, it will do that on the higher level. Just for more `it`s rather than only the ones in this block.

frontend/src/tests/lib/pages/NnsProposals.spec.ts
dskloetd: I still don't understand. Is this comment referring to something in a different file? Or is there a specific line in this file where it's mocking the service and it shouldn't?

frontend/src/tests/lib/pages/NnsProposals.spec.ts
```diff
@@ -63,6 +67,35 @@ describe("NnsProposals", () => {
...
+        const mockGovernanceCanister: MockGovernanceCanister =
+          new MockGovernanceCanister([]);
+        jest
+          .spyOn(GovernanceCanister, "create")
+          .mockImplementation((): GovernanceCanister => mockGovernanceCanister);
```
dskloetd, line 77: mockResolvedValue?

frontend/src/tests/lib/pages/NnsProposals.spec.ts
```diff
@@ -195,6 +228,28 @@ describe("NnsProposals", () => {
...
```
dskloetd, line 238: mockResolvedValue?

frontend/src/lib/services/neurons.services.ts
dskloetd: Yes, it will respond to both from the cache. But it shouldn't be a problem that you get 2 responses both from the cache, right?

lmuntaner: Not in this test, I reset the cache here.

I found the problem, but I couldn't replicate it in isolation.

I added some comment and a todo. I also did the test in the bad case that query comes after update call and it still works.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: Changed!

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: forget what I said. It can be greatly simplified:

```
      jest
        .spyOn(authServices, "getAuthenticatedIdentity")
        .mockResolvedValue(mockIdentity);
```

frontend/src/tests/lib/pages/NnsProposalDetail.spec.ts
lmuntaner: Same as before, I thought that if you put it out of the `description` the beforeEach applied to each description

frontend/src/tests/lib/pages/NnsProposals.spec.ts
lmuntaner: Below we do this:

```
const mockGovernanceCanister: MockGovernanceCanister =
          new MockGovernanceCanister([]);
jest
          .spyOn(GovernanceCanister, "create")
          .mockImplementation((): GovernanceCanister => mockGovernanceCanister);
```

Instead of mocking directly mocking the api:

```
jest.spyOn(proposalsApi, "queryProposals")...
```

I didn't want to change how it's done in this PR.

frontend/src/tests/lib/pages/NnsProposals.spec.ts
lmuntaner: yes

frontend/src/tests/lib/pages/NnsProposals.spec.ts
lmuntaner: done

lmuntaner:

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
dskloetd: You can move `resetNeuronsApiService();` as well.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
```diff
@@ -84,16 +70,65 @@ describe("NnsNeurons", () => {
...
+        })
+      );
+      expect(api.queryNeurons).toHaveBeenCalledWith({
+        identity: mockIdentity,
+        certified: false,
+      });
+
+      render(NnsNeurons);
+
+      await waitFor(() => expect(api.queryNeurons).toHaveBeenCalledTimes(2));
```
dskloetd, line 131: How doe this guarantee that it's not called more *after* this?

frontend/src/tests/lib/pages/NnsProposals.spec.ts
```diff
@@ -63,6 +67,35 @@ describe("NnsProposals", () => {
...
 
+    describe("neurons", () => {
+      beforeEach(() => {
+        // TODO: Mock the canister call instead of the service
+        const mockGovernanceCanister: MockGovernanceCanister =
+          new MockGovernanceCanister([]);
+        jest
+          .spyOn(GovernanceCanister, "create")
+          .mockReturnValue(mockGovernanceCanister);
+        jest.clearAllMocks();
```
dskloetd, line 78: It's safer to use resetAllMocks because clearAllMock doesn't reset the defined behavior. Though you'd have to do it earlier or you would reset the behavior you just defined on the previous lines.
But I would anyway suggest to put all the resets at the top level beforeEach (line 51).

frontend/src/tests/lib/pages/NnsProposals.spec.ts
```diff
@@ -195,6 +228,28 @@ describe("NnsProposals", () => {
...
+        neuronsStore.reset();
+        resetNeuronsApiService();
```
dskloetd, line 241: You don't need the resets here once you put them all on the top level.

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -319,6 +316,14 @@ describe("neurons-services", () => {
   });
 
   describe("list neurons", () => {
+    const spyQueryNeurons = jest
+      .spyOn(api, "queryNeurons")
+      .mockResolvedValue(neurons);
+
+    beforeEach(() => {
+      resetNeuronsApiService();
```
dskloetd, line 324: Put it on line 219 instead?

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -331,6 +336,24 @@ describe("neurons-services", () => {
...
+      });
+
+      expect(spyQueryNeurons).toHaveBeenCalledWith({
+        identity: mockIdentity,
+        certified: false,
+      });
+
+      await listNeurons();
+
+      expect(spyQueryNeurons).toHaveBeenCalledTimes(2);
```
dskloetd, line 354: Put a copy of this line also before line 352 to make it clear that the call on line 352 didn't change anything.

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -55,23 +59,46 @@ describe("Proposal detail page when not logged in user", () => {
 
   describe("when logged in user", () => {
     beforeEach(() => {
+      jest.clearAllMocks();
```
dskloetd, line 62: Use resetAllMock instead?

frontend/src/tests/lib/pages/NnsProposals.spec.ts
dskloetd: But I don't see a service?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -55,23 +59,46 @@ describe("Proposal detail page when not logged in user", () => {
...
+        .spyOn(governanceApi, "queryNeurons")
+        .mockImplementation(async ({ certified }) => {
+          // Mock delay in one of both calls.
+          // Otherwise there is a race condition that the onLoad of queryAndUpdate for query after certified has been called.
+          // The problem seems to be that the `certifiedDone` is set in the finally.
+          // Yet, I couldn't replicate the issue with a test in `utils.services.spec.ts`.
+          // Therefore, I delay here the QUERY to prove that it still works, even in the bad case that certified data comes before uncertified.
+          if (!certified) {
+            return new Promise((resolve) =>
+              setTimeout(() => {
```
dskloetd, line 75: It would be nicer if you return unresolved promises and explicitly resolve one before the other.

frontend/src/tests/workflows/NnsProposals.spec.ts
```diff
@@ -30,19 +33,24 @@ jest.mock("$lib/api/proposals.api", () => {
...
+
 describe("NnsProposals", () => {
   afterEach(() => {
     jest.clearAllMocks();
   });
 
   describe("when signed in user", () => {
     beforeEach(() => {
+      neuronsStore.reset();
+      resetNeuronsApiService();
```
dskloetd, line 46: Put the resets at the higher level?

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: I'm not sure how to test that.

Shall I add some `await tick()` and then do this test?

dskloetd: Is there something you can do in the component (click around like the user using the component) after which you check the number of calls?

frontend/src/tests/lib/pages/NnsProposals.spec.ts
lmuntaner: discussed in slack, comment will change

frontend/src/tests/lib/pages/NnsProposals.spec.ts
lmuntaner: yes, I'll move in the top most `beforeEach`

frontend/src/tests/lib/services/neurons.services.spec.ts
lmuntaner: Ok

frontend/src/tests/lib/services/neurons.services.spec.ts
lmuntaner: ok

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: As mentioned, there are some clash with mocks added in a global file. I'll keep it and that can be improved later

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: Not sure I understand.

Here I'm resolving one later than the other.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: Not in this list part. I added some `await tick();`s. This "simulates" the user waiting...

frontend/src/tests/workflows/NnsProposals.spec.ts
lmuntaner: done

lmuntaner:

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
```diff
@@ -2,78 +2,65 @@
...
-    authStoreMock = jest
-      .spyOn(authStore, "subscribe")
-      .mockImplementation(mockAuthStoreSubscribe);
+    jest.resetAllMocks();
+    neuronsStore.reset();
   });
 
   describe("with enough neurons", () => {
     beforeEach(() => {
+      resetNeuronsApiService();
```
dskloetd, line 27: Put this on line 23?

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
dskloetd: In the diff I see it's still here.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
```diff
@@ -84,16 +71,70 @@ describe("NnsNeurons", () => {
...
+        })
+      );
+      expect(api.queryNeurons).toHaveBeenCalledWith({
+        identity: mockIdentity,
+        certified: false,
+      });
+    });
+
+    it("should NOT call query neurons after being visited", async () => {
+      resetNeuronsApiService();
```
dskloetd, line 116: If you have this in beforeEach, you shouldn't need this here, right?

frontend/src/tests/lib/pages/NnsProposals.spec.ts
dskloetd: It still looks the same to me?

frontend/src/tests/lib/pages/NnsProposals.spec.ts
```diff
@@ -191,10 +216,29 @@ describe("NnsProposals", () => {
...
           return () => undefined;
         });
+      jest.spyOn(governanceApi, "queryNeurons").mockResolvedValue([]);
     });
 
     afterAll(() => jest.clearAllMocks());
 
+    describe("neurons", () => {
+      beforeEach(() => {
+        // TODO: Mock the canister call instead of the service
```
dskloetd, line 226: same comment here

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
dskloetd: OK, but you don't need it here because you have it on line 56 already, right?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -55,23 +59,46 @@ describe("Proposal detail page when not logged in user", () => {
 
   describe("when logged in user", () => {
     beforeEach(() => {
+      jest.clearAllMocks();
+      neuronsStore.reset();
```
dskloetd, line 63: You have this on line 57 already?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -55,23 +59,46 @@ describe("Proposal detail page when not logged in user", () => {
...
+      resetNeuronsApiService();
```
dskloetd, line 64: Move this to line 58?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
dskloetd: Make the promises available outside the implementation in the test and explicitly resolve them there in a specific order.

So like this:
```
let resolveCertifiedPromise;
let resolveUncertifiedPromise;

      jest
        .spyOn(governanceApi, "queryNeurons")
        .mockImplementation(async ({ certified }) => {
          return new Promise((resolve) => {
            if (certified) {
             resolveCertifiedPromise = resolve;
            } else {
              resolveUncertifiedPromise = resolve;
            }
          });
       });
```

In the test:
```
...
resolveCertifiedPromise();
...
resolveUncertifiedPromise();
...
```

Does this make sense?

frontend/src/tests/workflows/NnsProposals.spec.ts
dskloetd: I still see it here.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: correct, moving it

frontend/src/tests/lib/pages/NnsProposals.spec.ts
lmuntaner: I reverted some changes because test was failing and I forgot to change it back, sorry.

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: aahh, yes, I understand now.

frontend/src/tests/lib/pages/NnsNeurons.spec.ts
lmuntaner: done

frontend/src/tests/lib/pages/NnsProposals.spec.ts
lmuntaner: done

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: done

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: done

lmuntaner:
@dskloetd please review

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -47,31 +52,69 @@ jest.mock("$lib/utils/html.utils", () => ({
...
+              resolveUncertifiedPromise = resolve;
+            }
+          });
+        });
       jest
         .spyOn(authStore, "subscribe")
         .mockImplementation(mockAuthStoreSubscribe);
     });
 
     it("should render proposal with certified data", async () => {
```
dskloetd, line 90: Is the purpose of this test to test what happens if the promises resolve in the "wrong" order? Or are you piggybacking this on another test?

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
```diff
@@ -94,10 +139,40 @@ describe("Proposal detail page when not logged in user", () => {
...
+      });
+      // Make sure we have the resolve functions.
+      await waitFor(() =>
+        expect(typeof resolveCertifiedPromise).toBe("function")
+      );
+      await waitFor(() =>
+        expect(typeof resolveUncertifiedPromise).toBe("function")
+      );
+      resolveUncertifiedPromise([mockNeuron]);
+      resolveCertifiedPromise([mockNeuron]);
```
dskloetd, line 158: Does this test also rely on specific control over the order of resolution?
Otherwise would it make sense to use that specific behavior only for that test that needs it?

dskloetd (approved):
I don't want to block on the remaining comments but they do point out a real potential issue.

If there is a specific issue that needs to be tested, it needs its own separate test, otherwise it's unclear and may be removed later.

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: I'm not trying to test that here.

frontend/src/tests/workflows/NnsProposalDetail.spec.ts
lmuntaner: No, that's not the purpose of this test.

I found a test for that in `utils.services.spec.ts` already

PR #1848, opened 2023-02-09
Add logs

bitdivine:
# Motivation
In production application subnets, logs are not available.  This can make debugging issues hard.

# Changes
* Write logs to circular buffer and make the buffer available for download.
* Disable the cache restoration after upgrade, as that has a known serialization issue that spams the logs.  It's a cache, it gets repopulated, it's not a critical problem.  But it is noisy.

# Tests
* Added a test in CI
* Used on testnets, including the farm which is prod-like in th esense that it has no logs on application subnets such as the SNS subnet.


rs/sns_aggregator/src/lib.rs
```diff
@@ -103,22 +114,25 @@ fn pre_upgrade() {
 #[deny(clippy::panic)] // Panicking during upgrade is bad.
 #[ic_cdk_macros::post_upgrade]
 fn post_upgrade(config: Option<Config>) {
-    ic_cdk::api::print("Calling post_upgrade...");
+    crate::state::log("Calling post_upgrade...".to_string());
     // Make an effort to restore state.  If it doesn't work, give up.
```
bitdivine, line 118: This is a known bug.  There is no point in exercising it and spamming the logs until it is fixed. https://dfinity.atlassian.net/browse/GIX-1308

rs/sns_aggregator/src/lib.rs
```diff
@@ -48,6 +49,16 @@ fn stable_data() -> String {
     })
 }
 
+/// Get most recent log data
+#[candid_method(query)]
+#[ic_cdk_macros::query]
+fn tail_log() -> String {
```
lmuntaner, line 55: Should you limit the number of lines?

rs/sns_aggregator/src/lib.rs
```diff
@@ -103,22 +114,25 @@ fn pre_upgrade() {
...
-            .and_then(|(bytes,): (Vec<u8>,)| StableState::from_bytes(&bytes))
-        {
-            Ok(data) => {
-                *state.asset_hashes.borrow_mut() = AssetHashes::from(&*data.assets.borrow());
-                *state.stable.borrow_mut() = data;
-            }
-            Err(message) => {
-                ic_cdk::api::print(message);
+    if false {
+        // Disabled until stable serde is fixed.
```
lmuntaner, line 120: Is there an issue on our side? Or is this from another library.

Sorry, I'm not familiar with this, hehe.

lmuntaner (commented):
Why is this empty file needed `scripts/sns/aggregator/get_stable_data`?

I added a question as well

rs/sns_aggregator/src/lib.rs
bitdivine: The number of lines in the circular buffer is small (200 lines, if I recall correctly) but I could add an optional limit anyway, in case we increase the log size.

rs/sns_aggregator/src/lib.rs
bitdivine: It's elsewhere but we still need to come up with a fix ourselves, if we want to keep the cache after redeploying.  Given the upstream data, doing completely standard, uncustomized serialize + deserialize fails.

rs/sns_aggregator/src/lib.rs
bitdivine: Done.

rs/sns_aggregator/src/lib.rs
lmuntaner: Ok, good point. Thanks!

lmuntaner (approved):

PR #1849, opened 2023-02-09
Minimal landing page for the sns aggregator

bitdivine:
# Motivation
It would be useful for the sns aggregator to have a home page with basic health check information.

This PR adds a home page with nothing but a description and a link to the latest SNS info.  To be expanded and made pretty but as a question of priority, more data and pretty can come later.

# Changes
* Add an index.html

# Tests
* You can see it live here: https://s55qq-oqaaa-aaaaa-aaakq-cai.benchmarklarge.dfinity.network/ (note that the canister is being redeployed frequently so the link in the page will often show nothing.
Screenshot from 2023-02-09 14-06-16
Clicking the link goes to the JSON for the latest SNSs
Screenshot from 2023-02-09 18-10-32

### DIFF
--- a/rs/sns_aggregator/src/index.html
+++ b/rs/sns_aggregator/src/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <title>title</title>
+    <title>SNS aggregator</title>
   </head>
   <body>
 	  <h1>SNS aggregator</h1>
--- a/rs/sns_aggregator/src/index.html
+++ b/rs/sns_aggregator/src/index.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>title</title>
+  </head>
+  <body>
+	  <h1>SNS aggregator</h1>
+	  <p>The aggregator collects information on all deployed SNS and re-publishes aggegate values as certified query calls.  This data is intended for use by any canister that wishes to provide dashboards.  The data will typically be updated about once per minute.  Collecting the data from this aggregator will typically be much faster than collecting the same data directly from the SNSs.</p>
+	  <ul>
+		  <li><a href="/v1/sns/list/latest/slow.json">The latest SNSs</li>
+	  </ul>
+  </body>
+</html>
--- a/rs/sns_aggregator/src/assets.rs
+++ b/rs/sns_aggregator/src/assets.rs
@@ -268,3 +268,17 @@ pub fn insert_favicon() {
         }
     });
 }
+
+/// Insert a home page into the certified assets.
+pub fn insert_home_page() {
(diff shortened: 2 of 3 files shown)


lmuntaner (approved):

rs/sns_aggregator/src/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>title</title>
```
peterpeterparker, line 5: probably worth pimping a bit the metadata since this gonna be crawled. Notably the `<title>title</title` would deserve some love

bitdivine: Good idea.  What would you recommend for the metadata?  Apart from Title -> some pun on summing SNSs?

peterpeterparker: Here an example I literally build two days ago: https://github.com/buildwithjuno/beta-signup/blob/76d2a18fd90f29cb9d8180d1496eb795c0f016e9/src/app.html#L5

Bare minimal I would say is a miningful title.

Btw. we can also blocks robots by adding:

```
<meta name="robots" content="noindex" />
```

In the header. Not sure this canister needs to be crawled thinking about it...

We can also add a robots.txt to block all crawlers (https://github.com/buildwithjuno/beta-signup/blob/main/static/robots.txt)

peterpeterparker: Regardless if crawled or not, a meaninungfull title is useful. e.g. because it is displayed in the tab of the browser

bitdivine: Also, if it is going to be googled, it might be worth giving it a dfinity look&feel.  Background, logo, title, nothing more.  What do you think?  Is there a component for that or should I grab at least the background & font sizes from nns-dapp?

bitdivine: Let's maybe meet on Monday with pen & paper and make a sketch.  I am imagining having a list of the top 10 SNSs with name & logo.  No infinite scroll but links to the paginated responses.  The idea would be to show what is available in the API and the current health status rather than making a fancy front end.  Health status would require an agent and query calls.  It would be nice if agent.js were served from some assets canister somewhere.  That way small single-file HTML pages could import the library with one line in the header and without having to host anything more.

bitdivine: We could put our own releases in such a canister.  Why download from gh when we can host on the IC?

peterpeterparker: Good idea, we can "brainstorm" Monday what we can do without over-engineering the solution.

Meanwhile, I would suggest to just set an accurate `<title>....</title>`and merge your PR, that's already a cool stuffs 👍

bitdivine: Thank you for the review

PR #1850, opened 2023-02-09
Handle missing logo

bitdivine:
# Motivation
The sns aggregator provides a logo URL for every SNS.  This can be used directly in HTML without a custom API call.  The code that generates this URL does not, however, handle the case where the logo is not provided.

# Changes
- If the logo is not provided, provide an empty file.  Note: It can be debated whether we should return a 404, an empty file or some default image.  The priority now is to make the canister not break.  _maybe_ the 404 is the better option, as it allows fallback images, although it will print error messages in the console and red error messages shout "page is broken" which isn't nice either.  Maybe certify a pretty HTML message explaining that the SNS has no logo along with the 404 so that there are no error messages in the console and users have a good experience.  Once we have agreed, this would be a good follow-up task.  It is not a trivial matter of just changing the default text so would better be done in a separate PR.

# Tests
* See here for an SNS that has no logo but that has been scraped successfully: https://s55qq-oqaaa-aaaaa-aaakq-cai.benchmarklarge.dfinity.network/v1/sns/list/latest/slow.json
Screenshot from 2023-02-09 18-10-32

### DIFF
--- a/rs/sns_aggregator/src/state.rs
+++ b/rs/sns_aggregator/src/state.rs
@@ -105,7 +105,7 @@ impl State {
             let path = format!("{prefix}/sns/root/{root_canister_str}/logo.{LOGO_FMT}");
             let asset = Asset {
                 headers: Vec::new(),
-                bytes: logo_binary(upstream_data.meta.logo.as_ref().map(|s| s.as_str()).unwrap_or("")),
+                bytes: logo_binary(upstream_data.meta.logo.as_deref().unwrap_or("")),
             };
             insert_asset(path, asset);
         }
--- a/rs/sns_aggregator/src/state.rs
+++ b/rs/sns_aggregator/src/state.rs
@@ -105,7 +105,7 @@ impl State {
             let path = format!("{prefix}/sns/root/{root_canister_str}/logo.{LOGO_FMT}");
             let asset = Asset {
                 headers: Vec::new(),
-                bytes: logo_binary(upstream_data.meta.logo.as_ref().expect("Missing logo")),
+                bytes: logo_binary(upstream_data.meta.logo.as_ref().map(|s| s.as_str()).unwrap_or("")),
             };
             insert_asset(path, asset);
         }


lmuntaner (approved):

rs/sns_aggregator/src/state.rs
```diff
@@ -105,7 +105,7 @@ impl State {
             let path = format!("{prefix}/sns/root/{root_canister_str}/logo.{LOGO_FMT}");
             let asset = Asset {
                 headers: Vec::new(),
-                bytes: logo_binary(upstream_data.meta.logo.as_ref().expect("Missing logo")),
+                bytes: logo_binary(upstream_data.meta.logo.as_deref().unwrap_or("")),
```
peterpeterparker, line 108: shouldn't no asset be added to the memory and no key added to the json data if no logo?

bitdivine: @peterpeterparker  If I do that, there will be an uncertified 404 and red error messages in the console and that makes it look as if something is broken.  I think a better solution would be to add a certified 404 response with text explaining that the SNS has no logo.  ("Sorry.  The SNS has no logo.  Have a penguin instead: <insert picture of penguin>")  No red error messages in the console and a helpful explanation for anyone trying to find out why no logo is being served.  What do you think?

bitdivine: I am going to merge this as currently the aggregator breaks but I would like to find the "best" solution and implement it, so let's keep this discussion going, please.

bitdivine: If the visitors are geeky enough, we could have "Try another lisp.." and host the penguin under clisp.png.

peterpeterparker: For sure it should not block this PR. I probably miss something because we can add and remove asset from the certification tree so I don't see why I would broke the certification but, I probably miss something. Like I said nevermind my comment and looking forward to the "best" solution.

lmuntaner: Aren't we going to remove this png logo from the response anyway in the next version?

I thought that the idea was to return exactly the same as the SNS response.

And fix this logo thing in the source.

PR #1852, opened 2023-02-09
add cors headers to sns_aggregator

mraszyk:
Add cors headers to sns aggregator http responses so that the json file loads from localhost.


bitdivine (approved):

bitdivine:
LGTM, thank you.  I believe that Security would also be fine with this, given that the very purpose of the canister is to provide data to other web pages.

PR #1854, opened 2023-02-09
GIX-1270 - Payment flow upgrade

mstrasinskis:
# Motivation

MVP of the updated payment flow update. 
- [x] https://github.com/dfinity/ic-js/pull/280

# Changes

# Tests


declarations/sns_swap/sns_swap.did
```diff
@@ -31,26 +31,40 @@ type DerivedState = record {
 };
```
lmuntaner, line 31: I didn't know we had this file! What is this used for?

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -0,0 +1,425 @@
...
+    if ("Ok" in resultData) {
+      return {
+        rootCanisterId,
+        ticket: fromNullable(resultData.Ok.ticket),
+      };
+    }
+
+    // TODO(GIX-1271): display more details in error message
+
+    if ("Err" in resultData) {
```
lmuntaner, line 81: This is the part that we normally do in `ic-js`.

The pattern was to raise an error there instead of returning the whole response.

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -0,0 +1,425 @@
...
+    if ("Ok" in resultData) {
+      return {
+        rootCanisterId,
+        ticket: fromDefinedNullable(resultData.Ok.ticket),
+      };
+    }
+
+    // TODO(GIX-1271): display more details in error message
+
+    if ("Err" in resultData) {
```
lmuntaner, line 149: ditto

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -0,0 +1,425 @@
...
+
+/**
+ * Requests swap state and loads it in the store.
+ * Ignores possible undefined. This is used only to recheck the data with up-to-date information.
+ * This should be used only when the data is already in the store.
+ * That's why if an error happens, we want to rely on the data that it's already in the store.
+ *
+ * @param {Principal} rootCanisterId Root canister id of the project.
+ */
+const reloadSnsState = async (rootCanisterId: Principal): Promise<void> => {
```
lmuntaner, line 188: We need to check this.

@DanielThurau mentioned that using this might be a problem when there are many participations.

We could use the new endpoint `get_lifecycle` just to check that the sale has not closed at least.

Or maybe the new payment flow doesn't require this extra check. With the tickets they will have it easier to know to whom refund in case the transfer happens but the participage fails.

Whatever it's easier for you now to chage: either remove it entirely, or leave it. But we need a TODO to review this with the NNS and FI teams.

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -0,0 +1,425 @@
...
+ * |                      | SaleError::InvalidUserAmount(actual_min_participant_icp, actual_participant_icp_increment) |                                                                          | dapp can’t get it                                                                                                               |
+ * |                      | SaleError::ICPAllowanceNotAvailable                                                        |                                                                          | dapp can’t get it                                                                                                               |
+ * |                      | SaleError::CommitmentInProgress                                                            |                                                                          | dapp can’t get it                                                                                                               |
+ *
+ * @param saleTicket
+ */
+export const participateInSnsSwap = async ({
```
lmuntaner, line 354: Shouldn't this be in the `api` folder?

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -0,0 +1,425 @@
...
+export const initiateSnsSwapParticipation = async ({
+  amount,
+  rootCanisterId,
+  account,
+}: {
+  amount: TokenAmount;
+  rootCanisterId: Principal;
+  account: Account;
+}): Promise<{ success: boolean }> => {
+  console.log(
```
lmuntaner, line 228: I guess you'll remove this before merging, 😂

lmuntaner (commented):
Looking good, I made a few comments

frontend/src/lib/services/sns-sale.services.ts
mstrasinskis: Yep, updated 👍

frontend/src/lib/services/sns-sale.services.ts
mstrasinskis: Not sure, it needs a store access and uses another APIs. The decision was based on https://www.notion.so/NNS-Dapp-Frontend-Development-Guidelines-and-Architecture-5f387639a25e4ead8bf7a066df09860c?pvs=4#63001ce345dd476fbe586e0520cc9734. Let's discuss it later.

declarations/sns_swap/sns_swap.did
mstrasinskis: Me neither, I'll check this.

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -0,0 +1,377 @@
...
+    });
+    // TODO: Move the logic to the `catch` for a faster participation.
+    // At the moment we can't move it to the `catch`
+    // because it's hard for us to differentiate when the error comes from stale data or the second notify for the last participation.
+    //
+    // Reload the sale state before validating the participation.
+    // The current state might have change since it was loaded.
+    // This might prevent transferring funds that will not be accepted as participation and avoid refunds.
+
+    // TODO(sale): reload only if we don't do the transfer
```
lmuntaner, line 202: As I mentioned, I believe we should either check something else or don't do anything at all.

With the ticket the backend can handle refunds.

frontend/src/lib/services/sns-sale.services.ts
mstrasinskis: Thanks, created a task to not forget -- https://dfinity.atlassian.net/browse/GIX-1318

frontend/src/lib/services/sns-sale.services.ts
lmuntaner: Which store does it use? I don't see it.

True, the guidelines are not specific on APIs using other APIs.

In my view and how I implemented canisters or staking neurons, an api function can use other api functions. But no other concepts.

Worth rediscussing then and making it more explicit. It's not clear from the guidline: "Apis do not interact with any other concept."

frontend/src/lib/services/sns-sale.services.ts
lmuntaner: Thanks!

lmuntaner (approved):
I added a few more comments, but they don't need to be addressed in this PR.

Thanks! Already approving!

declarations/sns_swap/sns_swap.did
mstrasinskis: It is used and should be updated by the script. I'll redo the changes.

frontend/src/lib/services/sns-sale.services.ts
mstrasinskis: True, I was wrong w/ the store, but there is still the `toastsError` call. As decided, let's rediscus this later.

frontend/src/lib/services/sns-sale.services.ts
mstrasinskis: This part will be replaced w/ GIX-1318

PR #1856, opened 2023-02-09
add static testnets with url names to support multiple subnets

mraszyk:
Using, e.g., large05 results in the sns aggregator canister created on the NNS subnet. Using the BN URL as the network allows us to create the sns aggregator canister on the SNS subnet.


bitdivine (approved):

bitdivine:
Thank you.  I believe that dfx v 0.13 supports a default remote canister ID, so we will no longer need long lists repeating the same remote canister ID over and over again.  But we do still need entries for the configuration.  Maybe we can have a default configuration for testnets.

PR #1858, opened 2023-02-10
Paginated aggregations

bitdivine:
# Motivation
The sns_aggregator API includes paginated access to all SNSs.

# Changes
- Implement paginated listings of all SNS metadata (one paragraph)
- Add tests (all the rest)

# Tests
- Extended the aggregator e2e test to generate 12 SNSs and verify that the two pages (with indices 0-9 and 10-11) are present and have the correct number of items.

### DIFF
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -52,7 +52,7 @@ jobs:
           repository: 'dfinity/snsdemo'
           path: 'snsdemo'
           # Version from Thu Fep 9 with dfx 0.13.xxx
-          ref: '60a7369b864f72f1d0103462c8e979d1976ab109'
+          ref: 'b37d1ae845ed4c493aacbd59143071da8d4875d4'
       - name: Add SNS scripts to the path
         run: |
           echo "$PWD/snsdemo/bin" >> $GITHUB_PATH
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -61,7 +61,7 @@ jobs:
           dfx-sns-demo-install
       - name: Deploy NNS and SNS canisters
         run: |
-          ./scripts/deploy-snsdemo-testnet
+          ./scripts/deploy-snsdemo-testnet --verbose
       - name: Make the aggregator collect data quickly
         run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000; })'
       - name: 'Upload aggregator wasm module'
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -62,6 +62,8 @@ jobs:
       - name: Deploy NNS and SNS canisters
         run: |
           ./scripts/deploy-snsdemo-testnet
+      - name: Make the aggregator collect data quickly
+        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000; })'
       - name: 'Upload aggregator wasm module'
         uses: actions/upload-artifact@v3
         with:
@@ -70,11 +72,45 @@ jobs:
           retention-days: 3
       - name: Wait for the aggregator to get data
         run: sleep 30
-      - name: Get data from the sns aggregator
+      - name: Get the latest data from the sns aggregator
         run: |
           curl -Lf "http://q4eej-kyaaa-aaaaa-aaaha-cai.localhost:8080/v1/sns/list/latest/slow.json" | tee aggregate-1.json
-          (( $(jq length aggregate-1.json) == 1 ))  || {
-            echo ERROR: Expected to have 1 SNS in the aggregator.
+          expect=1
+          actual="$(jq length aggregate-1.json)"
+          (( expect == actual ))  || {
+            echo ERROR: Expected to have $expect SNS in the aggregator but found $actual.
+          }
+      - name: Make the aggregator collect data slowly
+        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000000; })'
+      - name: Create more SNSs
+        run: seq 11 | xargs -I{} dfx-sns-demo-mksns
+      - name: Make the aggregator collect data quickly
+        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000; })'
+      - name: Wait for the aggregator to get data
+        run: sleep 30
+      - name: Get the latest data from the sns aggregator
+        run: |
+          curl -Lf "http://q4eej-kyaaa-aaaaa-aaaha-cai.localhost:8080/v1/sns/list/latest/slow.json" | tee aggregate-1.json
+          expect=10
+          actual="$(jq length aggregate-1.json)"
+          (( expect == actual ))  || {
+            echo ERROR: Expected to have $expect SNS in the aggregator but found $actual.
+          }
+      - name: Get the first page of data from the sns aggregator
+        run: |
+          curl -Lf "http://q4eej-kyaaa-aaaaa-aaaha-cai.localhost:8080/v1/sns/list/page/0/slow.json" | tee aggregate-1.json
+          expect=10
+          actual="$(jq length aggregate-1.json)"
+          (( expect == actual ))  || {
+            echo ERROR: Expected to have $expect SNS in the aggregator but found $actual.
+          }
+      - name: Get the second page of data from the sns aggregator
+        run: |
+          curl -Lf "http://q4eej-kyaaa-aaaaa-aaaha-cai.localhost:8080/v1/sns/list/page/1/slow.json" | tee aggregate-1.json
+          expect=2
+          actual="$(jq length aggregate-1.json)"
+          (( expect == actual ))  || {
+            echo ERROR: Expected to have $expect SNS in the aggregator but found $actual.
           }
       - name: Stop replica
         run: dfx stop
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -52,7 +52,7 @@ jobs:
           repository: 'dfinity/snsdemo'
           path: 'snsdemo'
           # Version from Thu Fep 9 with dfx 0.13.xxx
-          ref: 'b37d1ae845ed4c493aacbd59143071da8d4875d4'
+          ref: 'bb6e84d52219c46f061728dc1ff4156c4393da8b'
       - name: Add SNS scripts to the path
         run: |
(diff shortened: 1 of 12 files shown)


.github/workflows/aggregator.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
           repository: 'dfinity/snsdemo'
           path: 'snsdemo'
           # Version from Thu Fep 9 with dfx 0.13.xxx
-          ref: '60a7369b864f72f1d0103462c8e979d1976ab109'
```
bitdivine, line 55: The sns_aggregator was being deployed under the default identity.  Now it is deployed under snsdemo8, which is the identity used for almost all the code here.  So now we don't need to jump between different identities any more.

.github/workflows/aggregator.yaml
```diff
@@ -61,7 +61,9 @@ jobs:
           dfx-sns-demo-install
       - name: Deploy NNS and SNS canisters
         run: |
-          ./scripts/deploy-snsdemo-testnet
+          ./scripts/deploy-snsdemo-testnet --verbose
+      - name: Make the aggregator collect data quickly
+        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000; })'
```
bitdivine, line 66: Possibly not needed, but why go slow?  In the following, I set the update interval for very fast updates before collecting data from the aggregator and slower whenever I am using the network in other ways.  Working locally, being able to change the speed is convenient.

.github/workflows/aggregator.yaml
```diff
@@ -70,11 +72,49 @@ jobs:
           retention-days: 3
       - name: Wait for the aggregator to get data
         run: sleep 30
-      - name: Get data from the sns aggregator
+      - name: Get the latest data from the sns aggregator
```
bitdivine, line 75: This test is basically unchanged; we check that there is one SNS in the list of latest SNSs.

.github/workflows/aggregator.yaml
```diff
@@ -70,11 +72,49 @@ jobs:
...
+          AGGREGATOR_CANISTER_ID="$(dfx canister id sns_aggregator)"
+          curl -Lf "http://${AGGREGATOR_CANISTER_ID}.localhost:8080/v1/sns/list/latest/slow.json" | tee aggregate-1.json
+          expect=1
+          actual="$(jq length aggregate-1.json)"
+          (( expect == actual ))  || {
+            echo ERROR: Expected to have $expect SNS in the aggregator but found $actual.
+          }
+      - name: Make the aggregator collect data slowly
+        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000000; })'
+      - name: Create more SNSs
```
bitdivine, line 86: Now we add another 11 SNSs and check that we have 12 SNSs in total in two pages of max 10 SNSs each.

rs/sns_aggregator/src/state.rs
```diff
@@ -146,6 +146,30 @@ impl State {
             };
             insert_asset(path, asset);
         }
```
bitdivine, line 148: The code for adding the current SNS to the corresponding page.

scripts/deploy-snsdemo-testnet
```diff
@@ -24,9 +24,32 @@ clap.define short=y long=nd_dir desc="Directory containing the nns-dapp source c
 source "$(clap.build)"
 set -euo pipefail
 
-: Standard dfx sns demo, with standard canisters, not local nns-dapp.
+: Standard dfx sns demo, with standard canisters, not local nns-dapp, as compiling the local dapp takes time and the canister is not needed.
 dfx-sns-demo --network "$DFX_NETWORK" ${DFX_IC_COMMIT:+--ic_commit} ${DFX_IC_COMMIT:+$DFX_IC_COMMIT} --ic_dir "$IC_REPO_DIR" --nd_dir "$ND_REPO_DIR"
```
bitdivine, line 28: Print diagnostic data and make sure that the current dfx identity as we expect.

In particular, the idetity can be wrong and the wallet is almost guaranteed to be invalid without intervention.  Wallets for the local dfx server are stored locally in each project but identities are global.  So if `dfx start --clean` is run in one project, user wallets are wiped out and recreated in that project.  But in another project, there will still be old wallets and canister IDs.

scripts/deploy-snsdemo-testnet
```diff
@@ -24,9 +24,32 @@ clap.define short=y long=nd_dir desc="Directory containing the nns-dapp source c
...
+AGGREGATOR_CANISTER_ID="$(
+  cd "$(dirname "$(command -v dfx-sns-demo)")"
+  dfx canister id sns_aggregator
+)"
+dfx-canister-set-id --canister_name sns_aggregator --canister_id "$AGGREGATOR_CANISTER_ID" --network "$DFX_NETWORK"
+dfx canister info sns_aggregator
+
 : Deploy the aggregator
-dfx deploy sns_aggregator --network "$DFX_NETWORK"
+RUSTFLAGS="--cfg feature=\"reconfigurable\"" dfx deploy sns_aggregator --network "$DFX_NETWORK"
```
bitdivine, line 55: The rustflags enable the reconfiguration API, that is otherwise not compiled.

rs/sns_aggregator/src/index.html
```diff
@@ -9,6 +9,9 @@ <h1>SNS aggregator</h1>
 	  <p>The aggregator collects information on all deployed SNS and re-publishes aggegate values as certified query calls.  This data is intended for use by any canister that wishes to provide dashboards.  The data will typically be updated about once per minute.  Collecting the data from this aggregator will typically be much faster than collecting the same data directly from the SNSs.</p>
 	  <ul>
 		  <li><a href="/v1/sns/list/latest/slow.json">The latest SNSs</li>
-	  </ul>
+          <li><a href="/v1/sns/list/page/0/slow.json">A paginated list of SNSs</li>
+          <li>Slow data for one SNS is at: /v1/sns/root/$CANISTER_ID/slow.json</li>
+          <li>The SNS logo is at: /v1/sns/root/$CANISTER_ID/logo.png</li>
+      </ul>
```
bitdivine, line 15: Advertise the pagination API on the landing page.  I have also included a description of the URLs for individual SNSs.  We plan to meet on Monday to come up with a proper design for this page.

.github/workflows/aggregator.yaml
```diff
@@ -70,11 +72,49 @@ jobs:
...
+        run: seq 11 | xargs -I{} dfx-sns-demo-mksns
+      - name: Make the aggregator collect data quickly
+        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000; })'
+      - name: Wait for the aggregator to get data
```
frederikrothenberger, line 90: Nit: Instead of always waiting 30s (which is probably too long to not cause flakiness) you could also poll on the expected result from latest with a timeout.

scripts/deploy-snsdemo-testnet
```diff
@@ -24,9 +24,32 @@ clap.define short=y long=nd_dir desc="Directory containing the nns-dapp source c
 source "$(clap.build)"
 set -euo pipefail
 
-: Standard dfx sns demo, with standard canisters, not local nns-dapp.
+: Standard dfx sns demo, with standard canisters, not local nns-dapp, as compiling the local dapp takes time and the canister is not needed.
 dfx-sns-demo --network "$DFX_NETWORK" ${DFX_IC_COMMIT:+--ic_commit} ${DFX_IC_COMMIT:+$DFX_IC_COMMIT} --ic_dir "$IC_REPO_DIR" --nd_dir "$ND_REPO_DIR"
+
+: Make sure that we are using the expected ID and that it is healthy
+[[ "$(dfx identity whoami)" == "snsdemo8" ]] || dfx identity use snsdemo8
```
frederikrothenberger, line 31: Maybe you could make `dfx` switch back to the previous identity after running the script? Otherwise this might confuse people that are not aware that this script also changes the dfx configuration.

rs/sns_aggregator/src/state.rs
```diff
@@ -146,6 +146,30 @@ impl State {
...
+                    .borrow()
+                    .sns_cache
+                    .borrow()
+                    .upstream_data
+                    .values()
+                    .skip((pagenum * State::PAGE_SIZE) as usize)
+                    .take(State::PAGE_SIZE as usize)
+                    .map(SlowSnsData::from)
+                    .collect();
+                serde_json::to_string(&slow_data).unwrap_or_default()
```
frederikrothenberger, line 165: Is there a particular reason that you fall back to the default here? Would an explicit failure (like in the previous block) not be better?

frederikrothenberger (approved):
Take my review with a grain of salt, as I'm lacking a lot of context. However, the changes make sense. Left a few suggestions.

rs/sns_aggregator/src/state.rs
bitdivine: The exceptions are all being replaced, so rather the previous block will look like this one.  Initially I wrote the code with the mentality "everything must be correct or do nothing".  In practice, partial functionality has proved to be more valuable than zero functionality.  Also, in prod we don't have access to logs, so the new policy is "never ever panic/except/unwrap but instead return an error and log the error in a circular buffer that can be read via a query call".  The no panic/expect/unwrap lints are in the feature branch and will land in main sometime soon.

.github/workflows/aggregator.yaml
bitdivine: Currently the test spends 15m installing code and SNSs and one minute sleeping.  So overall the gain would be marginal.  Development time is currently at a premium.  If the equation were different I would agree with you.

rs/sns_aggregator/src/state.rs
bitdivine: Thinking about what errors could realistically arise and what the best course of action would be.  When would serialization to string fail?  When there is insufficient RAM?

scripts/deploy-snsdemo-testnet
bitdivine: All our current test environments assume the user snsdemo8, so in the current context switching back doesn't seem to make sense.  Become snsdemo8 and never leave that state without very good reason, is the rule here!  Generally I don't like `dfx identity use` as it modifies global state.  I would prefer the current ID to be provided as an env var, so for a given environment an identity can be specified.  And env vars don't leak up, so some code low down in a subshell or subprocess somewhere cannot change the identity used after the subshell has returned.  It is also thread-safe; different processes can have the identity env var set to different values at the same time without a problem.  I would rather put effort into implementing that.

PR #1859, opened 2023-02-10
Aggregator releases

bitdivine:
# Motivation
We need to make stable, public releases of the sns aggregator.  Both for when we deploy and for other testnets.

# Changes
* Build and release the aggregator from a docker build.
* Be more precise about the nns-dapp dockerfile inputs, otherwise it is rebuilt every time even if there is only a change in the aggregator code.
* Remove some legacy code

# Tests
* See a sample release: https://github.com/dfinity/nns-dapp/releases/tag/untagged-dedc484de072526539e5
* I have verified that ./scripts/docker-build.sh produces the same nns-dapp sha256 as on main, so the refactoring has not changed the build output.


.github/workflows/docker-build.yaml
```diff
@@ -1,9 +1,6 @@
 name: Docker build
 on:
   push:
```
bitdivine, line 3: The trigger was not detecting tag pushes.  Just adding tag: to the push rules didn't help.  Not this is run on every push, like the native build.

lmuntaner: Do we want this? Isn't this going to even more delay in the CI pipeline?

Dockerfile
```diff
@@ -79,8 +80,48 @@ RUN ./build.sh
...
+COPY ./Cargo.toml /build/
+COPY ./Cargo.lock /build/
+COPY ./dfx.json /build/
+COPY --from=build_frontend /build/assets.tar.xz /build/
+WORKDIR /build
+RUN export DFX_NETWORK && ./build-backend.sh
+
+FROM builder AS build_aggregate
+SHELL ["bash", "-c"]
+COPY ./rs /build/rs
```
lmuntaner, line 112: Why do yuo need to copy again? Didn't you do it before?

lmuntaner (commented):
A couple of questions

.github/workflows/docker-build.yaml
bitdivine: No, the docker build runs in parallel to the other jobs and is not the slowest.  Indeed, if we want the pipeline to go faster, one path is to drop native builds, build in docker only and enable caching for everything, not just the builder image.  Locally that gives fast builds. :-)  But to do that in CI we need a bigger cache than that provided by github, otherwise we just thrash the cache as the 10G runs out fairly quickly with massive docker images.

This docker build, with no caching except for the builder took 8m32s: https://github.com/dfinity/nns-dapp/actions/runs/4149462846
The last native build 22m24s: https://github.com/dfinity/nns-dapp/actions/runs/4169375737

Dockerfile
bitdivine: Docker is running several images here.  Each paragraph  starting with `FROM` is a separate image.  So we have `builder` running first, installing tools such as ic-cdk-optimizer.  That is super slow but usually hits the cache as it changes only once every few months.  Then we have two paths running in parallel:  On one path we have just the aggregator, on the other we build the frontend, then the nns-dapp. All of these three jobs run in separate environments, all cloned from builder, so the `COPY ./rs ...` in one doesn't cause the files to be present in the others.

When developing locally, everything is cached, so if I make a change to the aggregator and then docker build, the builder, frontend and nns-dapp hit the cache and take maybe a second and only the aggregator is rebuilt.  If I change the nns-dapp rust code, the frontend and aggregator hit the cache and only the nns-dapp is rebuilt.  If the frontend code is changed, the aggregator hits the cache, the frontend is rebuilt, but the nns-dapp depends on the frontend via the `COPY --from=build_frontend` so everything after that COPY has to be rebuilt.

.github/workflows/docker-build.yaml
lmuntaner: Ok, noted!

lmuntaner (approved):
Thanks for the explanations! Approving!

PR #1860, opened 2023-02-11
fix: Save before computing aggregates

bitdivine:
# Motivation

This addresses a bug in the aggregator:  The aggregator would collect data for one SNS, but then compute aggregates before adding the data to the cache.  As a result it took unnecessarily long for new data to appear.

# Changes
* Move up the point at which the collected data is added to the cache.

# Tests
This has been deployed to testnets on multiple releases of the feature branch.


bitdivine:
Svelte test fails aftre not changing frontend??

rs/sns_aggregator/src/state.rs
```diff
@@ -92,6 +92,16 @@ impl State {
         let prefix = "/v1";
         let root_canister_id = convert_canister_id!(upstream_data.canister_ids.root_canister_id);
         let root_canister_str = root_canister_id.to_string();
+        // Add this to the list of values from upstream
+        STATE.with(|state| {
+            state
+                .stable
+                .borrow()
+                .sns_cache
```
lmuntaner, line 100: Not related to the change, but to the code. How does it know that when you do `sns_cache` or `upstream_data` is about a specific SNS and not another?

lmuntaner (approved):
Just on question about the code, not about the change.

Already approving

rs/sns_aggregator/src/state.rs
bitdivine: I am not quite sre I understand the question.  `sns_cache` contains the data for all upstream sns's (eventually).  The `upstream_data` is  a map keyed by root canister ID; that is how it stores data for multiple SNSs.  Is that what you meant?  I have to close the conversation in order to merge the PR but am happy to follow up if needed.

lmuntaner: That answers my question, thanks!

PR #1861, opened 2023-02-11
Collect some data early

bitdivine:
# Motivation
The aggregator update interval is configurable and can be set to a long time.  Fine, but that means that after deployment, it can take quite a while before there is any sign of life.  So, on installation, make sure that one SNS is scraped promptly.

# Changes
* Start pulling data quickly after installation.
* Slow down the default update frequency to a more prod-like value.

# Tests
CI should suffice.  Also, this has been used for deployments to testnets from a feature branch.


rs/sns_aggregator/src/lib.rs
```diff
@@ -173,6 +173,13 @@ fn setup(config: Option<Config>) {
             ic_cdk::timer::clear_timer(id);
         }
     });
+    // Get one SNS quickly, so that we don't need to wait for the normal, slow schedule before
+    // there is some sign of life.
+    for i in 0..2 {
```
lmuntaner, line 178: Why `0..2`?

rs/sns_aggregator/src/lib.rs
```diff
@@ -173,6 +173,13 @@ fn setup(config: Option<Config>) {
...
+        set_timer(Duration::from_secs(i * 4), || {
```
lmuntaner, line 179: Why 4 seconds added?

rs/sns_aggregator/src/lib.rs
```diff
@@ -173,6 +173,13 @@ fn setup(config: Option<Config>) {
             ic_cdk::timer::clear_timer(id);
         }
     });
+    // Get one SNS quickly, so that we don't need to wait for the normal, slow schedule before
+    // there is some sign of life.
```
lmuntaner, line 177: How do you know this?

lmuntaner (commented):
Just a few questions

rs/sns_aggregator/src/lib.rs
bitdivine: One to get the list of SNSs, one to get one SNS.

rs/sns_aggregator/src/lib.rs
bitdivine: The second call needs to come at least two blocks after the first.  Block speed varies but is almost invariably under 2 seconds.

rs/sns_aggregator/src/lib.rs
bitdivine: In theory because of the timer interval and in practice because I have deployed locally, to testnets and to mainnet.

rs/sns_aggregator/src/lib.rs
lmuntaner: I'd appreciate a comment with this explanation

lmuntaner (approved):
Thanks for the explanations!

I would appreciate some explanation in form of a comment. Already approving!

PR #1863, opened 2023-02-12
fix: remove error log

mstrasinskis:
# Motivation

Remove error log from test run.

# Changes

- mock console error in sns-proposals.services.spec.ts


peterpeterparker (approved):
Thanks!

PR #1864, opened 2023-02-12
build: update dev dependencies

peterpeterparker:
# Motivation

Bump all dev dependencies to stay up-to-date fix last fixes and improvements.

# Changes

- build all dev dependencies including Svelte
- adapt `enum.utils` and `<a alt />` for svelte check
- remove script `build.reproducible.mjs` because svelte does not generate `vite-manifest.json` anymore
- redo previous method to detect `isNode` to fix pre-rendering build with new libraries


frontend/package.json
```diff
@@ -6,8 +6,7 @@
     "build:csp": "node scripts/build.csp.mjs",
     "build:robots": "node scripts/build.robots.mjs",
     "build:preload": "node scripts/build.preload.mjs",
-    "build:reproducible": "node scripts/build.reproducible.mjs",
-    "build:post-process": "npm run build:robots && npm run build:csp && npm run build:reproducible",
```
lmuntaner, line 10: Why do you remove this_

lmuntaner (approved):
A question, but already approving.

frontend/package.json
peterpeterparker: See PR description:

> remove script build.reproducible.mjs because svelte does not generate vite-manifest.json anymore

No file so no need to delete it so no need of the script that deletes it

peterpeterparker: Thanks for the review!

PR #1865, opened 2023-02-12
Use the IC_COMMIT specified in nns-dapp when deploying SNSs

bitdivine:
# Motivation
In the nns-dapp repo we specify an ic-commit.  The idea is that we can deploy test environments with that commit and check that the nns-dapp code works with the ic codebase at that commit (although production always has a range of different versions, depending on what has been voted in as the latest update of each component, but never mind, there is no database that tracks the release artifacts at the current time and this is as close as we get to emulating prod at a given past or future time).  However the snsdemo is currently not told to use the commit we specify.

# Changes
- Pass the local `IC_COMMIT` to snsdemo, when deploying.
- Bump the IC_COMMIT from Oxt 2022 to Jan 2023, the default in the snsdemo.
  - Note: We would like a later commit, however firs SNS deployments need to work with that later commit

# Tests
CI should suffice


mstrasinskis (approved):
Thanks!

PR #1867, opened 2023-02-13
Script to bump ic_commit

bitdivine:
# Motivation
When we update the IC_COMMIT, we also need to update some code we derive from the upstream interfaces.

# Changes
- Add a script to bump the commit and derived code.
- Add a test to make sure that the code is in sync
- Bump the IC_COMMIT to the one currently used by the derived aggregator types.

# Tests
- There is a CI test


dfx.json
```diff
@@ -935,7 +935,7 @@
       "config": {
         "RUSTC_VERSION": "1.64.0",
```
bitdivine, line 936: Bump from Oct -> Jan (the commit currently used by the aggregator, it will have to be bumped again).

declarations/sns_swap/sns_swap.did
```diff
@@ -49,6 +49,12 @@ type FinalizeSwapResponse = record {
 type GetBuyerStateRequest = record { principal_id : opt principal };
 type GetBuyerStateResponse = record { buyer_state : opt BuyerState };
 type GetBuyersTotalResponse = record { buyers_total : nat64 };
+type GetDerivedStateResponse = record {
```
lmuntaner, line 52: Where is this used @bitdivine ?

If it's the aggregator only, could we move it inside `rs/aggregator`?

scripts/update_ic_commit
```diff
@@ -0,0 +1,55 @@
...
+  printf "%s\n" "$NEW_CONFIG" >"$SOURCE_DIR/../dfx.json"
+}
+
+: "Get the latest sns wasm and did files"
+{
+  : "Remove any existing did and wasm files from the build directory"
+  rm -f target/ic/*.{did,wasm}
+  : "Get or build the did and wasm files"
+  e2e-tests/scripts/nns-canister-build
+  : "Apparently there is a convention of putting did files here:"
```
lmuntaner, line 39: Who says it's a convention?

scripts/update_ic_commit
```diff
@@ -0,0 +1,55 @@
...
+  : "It would be nice to have a general rule rather than listing did files individually"
```
lmuntaner, line 40: Wouldn't it be better per project? So that we understand which project uses what?

Otherwise they are very decoupled and it's confusing.

lmuntaner (commented):
A few questions.

Ths nns-dapp backend also uses some ic comment in its depenencies of cargo.toml

Could that be linked as well? I'm not even sure we want, just thinking out loud.

declarations/sns_swap/sns_swap.did
bitdivine: The type or the file?  The did file is imported directly from the ic repo, complete, as it is, including types that we don't use.

There was a previous discussion about where the did files should live here https://github.com/dfinity/nns-dapp/pull/1772#pullrequestreview-1274549482  .  TLDR; I really don't care much where it lives but if there is a convention we might as well try to follow it.  The convention discussed there was actually `src/declarations`, which is indeed where the did files are if you run `dfx new foo`, not `declarations` but we don't have a `src`.  Maybe `rs/declarations` is the right interpretation of that convention.  And maybe `rs` should be renamed to `src` and we can follow the conventional layout to the full.

scripts/update_ic_commit
bitdivine: Does it make sense to have the nns-dapp targeting one version of the SNS and the aggregator a different one?  After all, we have one `dfx.json`, not one per project.  The sns canisters are listed in that dfx.json.  I believe that this is a case where we could use the technical meaning of equivocation, as the one canister would be presenting different did files depending on which canister was calling it.

scripts/update_ic_commit
bitdivine: See the conversation above.  This comment would be better stating something like: " dfx projects, by default, place did files in `/src/declarations`.  We have no src dir so place them directly in `/declarations`."

lmuntaner (approved):
Thanks for the explanations!

PR #1868, opened 2023-02-13
feat: ckBTC update balance

peterpeterparker:
# Motivation

Call `minter.update_balance` to finalize transferring bitcoin to ckBTC.

# PR

- [x] https://github.com/dfinity/ic-js/pull/289
- [x] https://github.com/dfinity/ic-js/pull/290

# Changes

- call `update_balance` and reload account on success in ckBTC "Receive" modal
- integrate related API and services from ic-js


HACKING.md
```diff
@@ -133,6 +133,26 @@ The index canister does not exist yet on mainnet but, will be proposed soon. Sam
...
+Therefore, this flag should also be manually set to `true` in [dfx.json](./dfx.json) and the `.env` should be generated.
+
+## Mint bitcoin
+
+To mint bitcoin you should set up a local bitcoin network as display in the [documentation](https://internetcomputer.org/docs/current/developer-docs/integrations/bitcoin/local-development).
+
+Once ready, 100 block can be minted to an address. The address can be found in the "ckBTC Receive modal".
+
+```bash
+./bin/bitcoin-cli -conf=$(pwd)/bitcoin.conf generatetoaddress 100 "bcrt1q286kjxmad4zmhex2dqus4t6u53z49y3shns028"
```
lmuntaner, line 147: Is this the same address than in the modal you mention above?

HACKING.md
```diff
@@ -133,6 +133,26 @@ The index canister does not exist yet on mainnet but, will be proposed soon. Sam
...
+```
+
+Once block minted, the minter should be unblocked to attribute reward. This can be done by minting more blocks to another address.
```
lmuntaner, line 150: This assumes quite some knowledge on ckBTC.

For my understanding, what does it mean "the minter should be unblocked to attribute reward"? Is that a state? What's the default?

"minting more blocks to another address"? apart from the `bcrt1q286kjxmad4zmhex2dqus4t6u53z49y3shns028` you mention above? Why would you do that?

frontend/src/lib/modals/accounts/CkBTCReceiveModal.svelte
```diff
@@ -42,7 +48,41 @@
...
+      toastsError({
+        labelKey: "error__ckbtc.update_balance",
+        err,
+      });
+    }
+
+    stopBusy("update-ckbtc-balance");
+  };
+
+  const reloadAccountAndClose = async () => {
```
lmuntaner, line 82: Shouldn't we use the busy screen also? And disable the button?

frontend/src/lib/pages/CkBTCWallet.svelte
```diff
@@ -42,8 +45,15 @@
 
   debugSelectedAccountStore(selectedAccountStore);
 
-  setContext<WalletContext>(WALLET_CONTEXT_KEY, {
+  // e.g. is called from "Receive" modal after minter.update_balance was successfully executed
```
lmuntaner, line 48: Or when the user clicks on the button if `bitcoin` right?

frontend/src/lib/api/ckbtc-minter.api.ts
```diff
@@ -3,30 +3,57 @@ import { CKBTC_MINTER_CANISTER_ID } from "$lib/constants/canister-ids.constants"
...
+export const updateBalance = async (params: {
+  identity: Identity;
+}): Promise<UpdateBalanceResponse> => {
+  logWithTimestamp("Updating ckBTC balance: call...");
+
+  const {
+    canister: { updateBalance: updateBalanceApi },
+  } = await ckBTCMinterCanister(params);
+
+  const result = await updateBalanceApi(minterParams(params));
```
lmuntaner, line 50: I should have realized this before, but wasn't the pattern to throw an error in `ic-js` if the response was an `Err` otherwise, return the data inside `Ok` or whatever prop was used?

frontend/src/lib/services/ckbtc-minter.services.ts
```diff
@@ -1,7 +1,46 @@
...
 export const getBTCAddress = async (): Promise<string> => {
   const identity = await getAuthenticatedIdentity();
   return getBTCAddressAPI({ identity });
 };
+
+export const updateBalance = async (): Promise<UpdateBalanceResult> => {
+  const identity = await getAuthenticatedIdentity();
+  const result = await updateBalanceAPI({ identity });
+
+  if ("Err" in result) {
```
lmuntaner, line 20: As I mentioned before, don't we normally have this logic in `ic-js`?

I'm sorry I didn't see it earlier and commented on it.

frontend/src/tests/lib/modals/accounts/CkBTCReceiveModal.spec.ts
```diff
@@ -11,15 +12,27 @@ import {
 import en from "../../../mocks/i18n.mock";
 import { renderModal } from "../../../mocks/modal.mock";
 
+jest.mock("$lib/services/ckbtc-minter.services", () => {
+  return {
+    updateBalance: jest.fn().mockImplementation(() => undefined),
```
lmuntaner, line 17: Ideally, we would like to start mocking only the `api` layer. Therefore, mock `updateBalance` from `$lib/api/ckbtc-minter.api.spec.ts`

lmuntaner (commented):
A few comments

HACKING.md
peterpeterparker: Yes that's an example. Added a keyword "example"

HACKING.md
peterpeterparker: That's why the chapters begins with

> ckBTC deployment and development in NNS-dapp are in progress.

> ### Bitcoin network

>To deploy a local Bitcoin network, the [documentation](https://internetcomputer.org/docs/current/developer-docs/integrations/bitcoin/local-development) on the IC website can be followed.

frontend/src/lib/pages/CkBTCWallet.svelte
peterpeterparker: Update comment as following: `// e.g. is called from "Receive" modal after user click "Done"`

frontend/src/lib/api/ckbtc-minter.api.ts
peterpeterparker: Good point I can do that.

frontend/src/tests/lib/modals/accounts/CkBTCReceiveModal.spec.ts
peterpeterparker: Not sure what would be added value here? Would you extend that in another PR?

frontend/src/lib/api/ckbtc-minter.api.ts
peterpeterparker: Done.

PR for ic-js 👉 https://github.com/dfinity/ic-js/pull/290
This PR adapted for it.

frontend/src/lib/modals/accounts/CkBTCReceiveModal.svelte
peterpeterparker: I thought about it too but ultimately did not added it because I thought the user experience would feel a bit more smooth without a spinner and backdrop because "it's just a reload in the background at the end of the day".

But not against adding a spinner at all. What do you think?

lmuntaner: I agree that maybe the busy store is too much. But at least disabling the button and showing a spinner there is needed.

Otherwise the user has no feedback that the click worked and might click twice.

peterpeterparker: I have added a spinner that's the shortest path. Good point, thanks for the input 👍

lmuntaner (approved):
LGTM, thanks!

PR #1874, opened 2023-02-14
GIX-1322: Setup featureFlags store

lmuntaner:
# Motivation

Be able to manage feature flags in mainnet.

First step: feature flags are read from a store instead of env vars.

# Changes

* New store `faetureFlagsStore` in `feature-flags.store.ts`.
* Export `FEATURE_FLAGS` in `environment.constants` instead of the individual flags.
* Read from the store when a flag wants to be read. These are most files changes.

# Tests

* Change the jest.mock import to the whole `FEATURE_FLAGS` in the setup.
* Add a test to check that store is initialized to the env vars values.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
@@ -19,7 +19,7 @@
   import SnsAutoStakeMaturity from "$lib/components/sns-neuron-detail/actions/SnsAutoStakeMaturity.svelte";
   import { isNullish } from "@dfinity/utils";
   import { authStore } from "$lib/stores/auth.store";
-  import { ENABLE_SNS_2 } from "$lib/constants/environment.constants";
+  import { featureFlagsStore } from "$lib/stores/feature-flags.store";
 
   const { store }: SelectedSnsNeuronContext =
     getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
@@ -42,7 +42,7 @@
     <h3 slot="value">{formattedTotalMaturity(neuron)}</h3>
   </KeyValuePair>
 
-  {#if hasStakedMaturity(neuron) && ENABLE_SNS_2}
+  {#if hasStakedMaturity(neuron) && $featureFlagsStore.ENABLE_SNS_2}
     <KeyValuePair testId="staked-maturity">
       <svelte:fragment slot="key">{$i18n.neurons.staked}</svelte:fragment>
 
@@ -52,13 +52,13 @@
     </KeyValuePair>
   {/if}
 
-  {#if allowedToStakeMaturity && ENABLE_SNS_2}
+  {#if allowedToStakeMaturity && $featureFlagsStore.ENABLE_SNS_2}
     <div class="actions" data-tid="stake-maturity-actions">
       <SnsStakeMaturityButton />
     </div>
   {/if}
 
-  {#if ENABLE_SNS_2}
+  {#if $featureFlagsStore.ENABLE_SNS_2}
     <SnsAutoStakeMaturity />
   {/if}
 </CardInfo>
--- a/frontend/src/lib/constants/environment.constants.ts
+++ b/frontend/src/lib/constants/environment.constants.ts
@@ -25,7 +25,7 @@ export interface FeatureFlags {
  *
  * @see feature-flags.store.ts to use feature flags
  */
-export const FEATURE_FLAGS: FeatureFlags = JSON.parse(
+export const FEATURE_FLAG_ENVIRONMENT: FeatureFlags = JSON.parse(
   import.meta.env.VITE_FEATURE_FLAGS.replace(/\\"/g, '"') ??
     '{"ENABLE_SNS_2":false, "ENABLE_SNS_VOTING": false, "ENABLE_SNS_AGGREGATOR": false, "ENABLE_CKBTC_LEDGER": true, "ENABLE_CKBTC_RECEIVE": false}'
 );
--- a/frontend/src/lib/constants/environment.constants.ts
+++ b/frontend/src/lib/constants/environment.constants.ts
@@ -12,7 +12,7 @@ export const SNS_AGGREGATOR_CANISTER_URL: string | undefined =
     ? undefined
     : (import.meta.env.VITE_AGGREGATOR_CANISTER_URL as string);
 
-export interface FEATURE_FLAGS {
+export interface FeatureFlags {
   ENABLE_SNS_2: boolean;
   ENABLE_SNS_VOTING: boolean;
   ENABLE_SNS_AGGREGATOR: boolean;
@@ -25,7 +25,7 @@ export interface FEATURE_FLAGS {
  *
  * @see feature-flags.store.ts to use feature flags
  */
-export const FEATURE_FLAGS: FEATURE_FLAGS = JSON.parse(
+export const FEATURE_FLAGS: FeatureFlags = JSON.parse(
   import.meta.env.VITE_FEATURE_FLAGS.replace(/\\"/g, '"') ??
     '{"ENABLE_SNS_2":false, "ENABLE_SNS_VOTING": false, "ENABLE_SNS_AGGREGATOR": false, "ENABLE_CKBTC_LEDGER": true, "ENABLE_CKBTC_RECEIVE": false}'
 );
--- a/frontend/src/lib/constants/environment.constants.ts
+++ b/frontend/src/lib/constants/environment.constants.ts
@@ -12,7 +12,7 @@ export const SNS_AGGREGATOR_CANISTER_URL: string | undefined =
     ? undefined
     : (import.meta.env.VITE_AGGREGATOR_CANISTER_URL as string);
 
-interface FEATURE_FLAGS {
+export interface FEATURE_FLAGS {
   ENABLE_SNS_2: boolean;
   ENABLE_SNS_VOTING: boolean;
   ENABLE_SNS_AGGREGATOR: boolean;
@@ -20,13 +20,12 @@ interface FEATURE_FLAGS {
   ENABLE_CKBTC_RECEIVE: boolean;
 }
 
-export const {
-  ENABLE_SNS_2,
-  ENABLE_SNS_VOTING,
-  ENABLE_SNS_AGGREGATOR,
-  ENABLE_CKBTC_LEDGER,
-  ENABLE_CKBTC_RECEIVE,
-}: FEATURE_FLAGS = JSON.parse(
+/**
+ * DO NOT USE DIRECTLY
+ *
+ * @see feature-flags.store.ts to use feature flags
+ */
+export const FEATURE_FLAGS: FEATURE_FLAGS = JSON.parse(
   import.meta.env.VITE_FEATURE_FLAGS.replace(/\\"/g, '"') ??
     '{"ENABLE_SNS_2":false, "ENABLE_SNS_VOTING": false, "ENABLE_SNS_AGGREGATOR": false, "ENABLE_CKBTC_LEDGER": true, "ENABLE_CKBTC_RECEIVE": false}'
 );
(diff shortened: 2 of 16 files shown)


frontend/src/lib/constants/environment.constants.ts
```diff
@@ -12,21 +12,20 @@ export const SNS_AGGREGATOR_CANISTER_URL: string | undefined =
...
-  ENABLE_SNS_2,
-  ENABLE_SNS_VOTING,
-  ENABLE_SNS_AGGREGATOR,
-  ENABLE_CKBTC_LEDGER,
-  ENABLE_CKBTC_RECEIVE,
-}: FEATURE_FLAGS = JSON.parse(
+/**
+ * DO NOT USE DIRECTLY
+ *
+ * @see feature-flags.store.ts to use feature flags
```
lmuntaner, line 26: This allows some integration with text editors:

Screenshot 2023-02-14 at 07 32 06

lmuntaner:

This is the setup, next PR will expose it to window and add the functionality to change a feature flag.

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -12,21 +12,20 @@ export const SNS_AGGREGATOR_CANISTER_URL: string | undefined =
...
+ */
+export const FEATURE_FLAGS: FEATURE_FLAGS = JSON.parse(
```
mstrasinskis, line 28: What do you think about having json.parse in the try catch block and set all the flags to false. To not break the whole app in case of wrong flags configuration?

frontend/src/lib/derived/selectable-universes.derived.ts
```diff
@@ -21,16 +22,19 @@ export const CKBTC_UNIVERSE: Universe = {
...
-  NNS_UNIVERSE,
-  ...(ENABLE_CKBTC_LEDGER ? [CKBTC_UNIVERSE] : []),
-  ...(projects?.map(({ rootCanisterId, summary }) => ({
-    canisterId: rootCanisterId.toText(),
-    summary,
-  })) ?? []),
-]);
+>(
+  [snsProjectsCommittedStore, featureFlagsStore],
+  ([projects, featueFlags]: [SnsFullProject[] | undefined, FEATURE_FLAGS]) => [
```
mstrasinskis, line 29: a typo in `featueFlags`

frontend/src/lib/derived/selected-universe.derived.ts
```diff
@@ -40,16 +41,19 @@ const pageUniverseIdStore: Readable<Principal> = derived(
...
->([pageUniverseIdStore, pageStore], ([canisterId, page]) => {
-  // ckBTC is only available on Accounts therefore we fallback to Nns if selected and user switch to another view
-  if (ENABLE_CKBTC_LEDGER && pathSupportsCkBTC(page)) {
-    return canisterId;
-  }
+>(
+  [pageUniverseIdStore, pageStore, featureFlagsStore],
+  ([canisterId, page, featueFlags]) => {
+    // ckBTC is only available on Accounts therefore we fallback to Nns if selected and user switch to another view
+    if (featueFlags.ENABLE_CKBTC_LEDGER && pathSupportsCkBTC(page)) {
```
mstrasinskis, line 50: ditto `featueFlags`

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface FeatureFlagsStore extends Readable<FEATURE_FLAGS> {
+  reset: () => void;
+}
+
+/**
+ * A store that contains the feature flags.
+ */
+export const initFeatureFlagsStore = (): FeatureFlagsStore => {
+  const defaultFlags: FEATURE_FLAGS = FEATURE_FLAGS;
```
mstrasinskis, line 14: a nitpick: it's a bit confusing that the value and the type have the same names. Probably `FeatureFlags` for the interface?

mstrasinskis (approved):
Looks great, thanks! Just a couple of suggestions.

frontend/src/lib/constants/environment.constants.ts
lmuntaner: That's a good idea. We can do that in another PR

frontend/src/lib/stores/feature-flags.store.ts
lmuntaner: Sounds good

mstrasinskis:
> This is the setup, next PR will expose it to window and add the functionality to change a feature flag.

Before exposing to the window, could you please do a quick research how risky it is from the security perspective in the modern browsers? Alternatively we could use the debug actions but it will be not so convenient as the window :)

lmuntaner:
> > This is the setup, next PR will expose it to window and add the functionality to change a feature flag.
> 
> Before exposing to the window, could you please do a quick research how risky it is from the security perspective in the modern browsers? Alternatively we could use the debug actions but it will be not so convenient as the window :)

Good point, I'll ask in prod-sec channel

frontend/src/lib/constants/environment.constants.ts
dskloetd: I disagree. We should fail fast rather than silently doing something unexpected and confusing.
It's also one of our principles:

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
```diff
@@ -42,7 +42,7 @@
     <h3 slot="value">{formattedTotalMaturity(neuron)}</h3>
   </KeyValuePair>
 
-  {#if hasStakedMaturity(neuron) && ENABLE_SNS_2}
+  {#if hasStakedMaturity(neuron) && $featureFlagsStore.ENABLE_SNS_2}
```
dskloetd, line 45: For convenience, would it be possible to have featureFlagsStore.ENABLE_SNS_2 be a derived store, which is exported as `ENABLE_SNS_2` so these lines can all stay as concise as they were before just with an extra `$`?

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -12,21 +12,20 @@ export const SNS_AGGREGATOR_CANISTER_URL: string | undefined =
...
-  ENABLE_SNS_AGGREGATOR,
-  ENABLE_CKBTC_LEDGER,
-  ENABLE_CKBTC_RECEIVE,
-}: FEATURE_FLAGS = JSON.parse(
+/**
+ * DO NOT USE DIRECTLY
+ *
+ * @see feature-flags.store.ts to use feature flags
+ */
+export const FEATURE_FLAGS: FeatureFlags = JSON.parse(
```
dskloetd, line 28: Can you change the name because this is now just 1 part of building the actual feature flags. Maybe FEATURE_FLAG_DEFAULT_VALUES or FEATURE_FLAG_ENVIRONMENT or something?

frontend/src/lib/constants/stores.constants.ts
```diff
@@ -1,4 +1,5 @@
 export enum storeLocalStorageKey {
   ProposalFilters = "nnsProposalFilters",
   Theme = "nnsTheme",
+  FeatureFlags = "nnsFeatureFlags",
```
dskloetd, line 4: FeatureFlagOverrides?

dskloetd, line 4: Doesn't this belong in the next PR?

frontend/src/lib/services/$public/app.services.ts
```diff
@@ -20,9 +19,11 @@ export const initAppPublicData = (): Promise<
   [PromiseSettledResult<void[]>, PromiseSettledResult<void[]>]
 > => {
   const initNns: Promise<void>[] = [];
+  const featureFlags = get(featureFlagsStore);
 
   const initSns: Promise<void>[] = [
-    ENABLE_SNS_AGGREGATOR && SNS_AGGREGATOR_CANISTER_URL !== undefined
+    featureFlags.ENABLE_SNS_AGGREGATOR &&
```
dskloetd, line 25: So this won't be a live update, right?

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -0,0 +1,39 @@
...
+type OverrideFeatureFlagsData = Partial<FeatureFlags>;
+export interface OverrideFeatureFlagsStore
+  extends Readable<OverrideFeatureFlagsData> {
+  reset: () => void;
+}
+
+/**
+ * A store that contains the feature flags that have been overridden by the user.
+ */
+const initOverrideFeatureFlagsStore = (): OverrideFeatureFlagsStore => {
```
dskloetd, line 18: Does this currently do anything? Based on the PR description I would expect this to be in the next PR.

dskloetd:
> > Before exposing to the window, could you please do a quick research how risky it is from the security perspective in the modern browsers? Alternatively we could use the debug actions but it will be not so convenient as the window :)
> 
> Good point, I'll ask in prod-sec channel

One risk I can think of is that you don't want to condition people that it's OK to just paste random code someone gave you into the console because with some other random code you'll get your money stolen. But this is temporary until we build a UI for it, right?

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
lmuntaner: Why would that be better? I see a lot of derived store and more work to add and then remove them

frontend/src/lib/constants/environment.constants.ts
lmuntaner: sure

frontend/src/lib/constants/stores.constants.ts
lmuntaner: yes, sounds good

frontend/src/lib/services/$public/app.services.ts
lmuntaner: Correct, this can't be a live update.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
dskloetd: More concise code is easier to read.
`{#if hasStakedMaturity(neuron) && $ENABLE_SNS_2}`
instead of 
`{#if hasStakedMaturity(neuron) && $featureFlagsStore.ENABLE_SNS_2}`

frontend/src/lib/stores/feature-flags.store.ts
lmuntaner: This sets up the store.

Yes, I could have skipped it, but it doesn't have any functionality. Therefore, I don't think it's much to review.

lmuntaner:
> > > Before exposing to the window, could you please do a quick research how risky it is from the security perspective in the modern browsers? Alternatively we could use the debug actions but it will be not so convenient as the window :)
> > 
> > 
> > Good point, I'll ask in prod-sec channel
> 
> One risk I can think of is that you don't want to condition people that it's OK to just paste random code someone gave you into the console because with some other random code you'll get your money stolen. But this is temporary until we build a UI for it, right?

If someone pastes code, they can steal money even today, there is no need to wait for a `window` exposed.

I'm not sure we'll have a UI for it, but I'd like it.

dskloetd:
> If someone pastes code, they can steal money even today, there is no need to wait for a `window` exposed.

Yes, that's why it's bad to normalize instructing people to paste code.

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: OK, that's fine. It was just confusing without explanation.

frontend/src/lib/constants/environment.constants.ts
lmuntaner: done

frontend/src/lib/constants/stores.constants.ts
lmuntaner: done

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
lmuntaner: discussed on Slack. We'll try that later.

dskloetd (approved):

frontend/src/lib/constants/stores.constants.ts
dskloetd: I meant on the left side.

lmuntaner: 🤦

PR #1875, opened 2023-02-14
feat: remove lazy load route

peterpeterparker:
# Motivation

We noticed issue with route not being destroy. Most recent issue was solve with a workaround in PR #1834.

While I am not sure what's the root cause of the issue, I tend to think that it might be linked with the fact that we lazy loaded routes with a `RouteModule` component.

Today I had a try and removed this feature.

It seems it had not a bad impact on JS chunks loading. Without it we roughly load the same number of chunks, on loading page it's even the same number of chunks.

And in addition, according my test, `onDestroy` of Wallet pages were always called.

Therefore this PR removes `RouteModule` component.

# Changes

- remove `RouteModule` component
- inegrate use of component in `+page`
- tweak `NnsProposals` component to make pre-rendering pass (list is not pre-rendered anyway)
- tweak few `back` function to make pre-rendering pass


frontend/src/lib/routes/Proposals.svelte
```diff
@@ -7,6 +7,7 @@
   import type { AppPath } from "$lib/constants/routes.constants";
   import { snsProjectSelectedStore } from "$lib/derived/sns/sns-selected-project.derived";
   import { nonNullish } from "$lib/utils/utils";
+  import { browser } from "$app/environment";
```
mstrasinskis, line 10: Will it be used here later?

mstrasinskis (approved):
LGTM, thanks! Just one question.

frontend/src/lib/routes/Proposals.svelte
peterpeterparker: Thanks no, removed. First implemention used `browser` instead of `building` and clean-up of the import missed it.

PR #1876, opened 2023-02-14
Fix: Neurons Detail Page issue

lmuntaner:
# Motivation

User sees a loading page when going directly to the neuron page deep link.

# Changes

* Call `listNeuron` on mount of NnsNeuronDetail.svelte

# Tests

* Add test to NnsNeuronDetail.svelte to check that api is called.
* Adapt test use the api mock instead of filling store direclty.

### DIFF
--- a/frontend/src/lib/pages/NnsNeuronDetail.svelte
+++ b/frontend/src/lib/pages/NnsNeuronDetail.svelte
@@ -28,12 +28,17 @@
     NnsNeuronStore,
   } from "$lib/types/nns-neuron-detail.context";
   import { NNS_NEURON_CONTEXT_KEY } from "$lib/types/nns-neuron-detail.context";
-  import { setContext } from "svelte";
+  import { onMount, setContext } from "svelte";
   import NnsNeuronModals from "$lib/modals/neurons/NnsNeuronModals.svelte";
   import NnsNeuronProposalsCard from "$lib/components/neuron-detail/NnsNeuronProposalsCard.svelte";
   import Summary from "$lib/components/summary/Summary.svelte";
+  import { listNeurons } from "$lib/services/neurons.services";
 
   export let neuronIdText: string | undefined | null;
+
+  onMount(() => {
+    listNeurons();
+  });
 
   const mapNeuronId = (
     neuronIdText: string | undefined | null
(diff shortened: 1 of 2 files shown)


frontend/src/lib/pages/NnsNeuronDetail.svelte
```diff
@@ -28,13 +28,18 @@
...
+  import { onMount, setContext } from "svelte";
   import NnsNeuronModals from "$lib/modals/neurons/NnsNeuronModals.svelte";
   import NnsNeuronProposalsCard from "$lib/components/neuron-detail/NnsNeuronProposalsCard.svelte";
   import Summary from "$lib/components/summary/Summary.svelte";
+  import { listNeurons } from "$lib/services/neurons.services";
 
   export let neuronIdText: string | undefined | null;
 
+  onMount(() => {
+    listNeurons();
```
dskloetd, line 40: Why listNeurons when we only need 1 neuron?

And how does this fix the bug?

lmuntaner: Because until now`listNeurons` has caching.

The problem was that the component was reading the store, but there was no neurons there. `listNeurons` loads the neurons in the store and then it find it.

This should be improved, but I consider this an important bug and I wanted to have a quick fix.

This is just one call also, and users don't have many neurons. Therefore, it's also not a big response.

And it has the advantage of the caching as I mentioned. If the user visits another neuron page navigating, then `listNeurons` will use the cache and no call will be done.

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
```diff
@@ -13,37 +14,36 @@ import en from "../../mocks/i18n.mock";
...
+  };
+
   const querySkeleton = (container: HTMLElement): HTMLElement | null =>
     container.querySelector('[data-tid="skeleton-card"]');
 
-  afterEach(() => {
+  beforeEach(() => {
     neuronsStore.reset();
     voteRegistrationStore.reset();
+    jest.spyOn(api, "queryNeurons").mockResolvedValue([neuron]);
```
dskloetd, line 39: Can we return multiple neurons and very that the correct one is used?
In another PR is fine.

dskloetd (approved):

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
lmuntaner: yes

lmuntaner: done

PR #1878, opened 2023-02-14
refactor: ckBTC transation modal move and call

peterpeterparker:
# Motivation

Ultimately the "Transaction" modal will / might most probably be used for ckBTC conversion too therefore we can move it - as we discussed in the footer, now worth the effort - and integrate it as the other modal.

For the same reason the "Receive" CTA can become a secondary button.

# Note

No functional changes.

# Changes

- integrate button to open modal with ckBTC wallet footer
- call modal with an event 
- style "Receive" as secondary

# Screenshot

Capture d’écran 2023-02-14 à 13 59 08

### DIFF
--- a/frontend/src/lib/pages/CkBTCWallet.svelte
+++ b/frontend/src/lib/pages/CkBTCWallet.svelte
@@ -13,7 +13,7 @@
   import { setContext } from "svelte/internal";
   import { findAccount, hasAccounts } from "$lib/utils/accounts.utils";
   import { ckBTCAccountsStore } from "$lib/stores/ckbtc-accounts.store";
-  import { isNullish, nonNullish } from "@dfinity/utils";
+  import { nonNullish } from "@dfinity/utils";
   import {
     loadCkBTCAccounts,
     syncCkBTCAccounts,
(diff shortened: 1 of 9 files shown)


frontend/src/lib/pages/CkBTCWallet.svelte
```diff
@@ -51,9 +46,13 @@
     await loadAccount();
   };
 
+  // e.g. when a function such as a transfer is called and which also reload the data and populate the stores after execution
+  const reloadAccountFromStore = () => setSelectedAccount();
```
lmuntaner, line 50: Thinking out loud, maybe not for this PR,

Could `setSelectedAccount` subscribe to changes in `ckBTCAccountsStore`?

frontend/src/tests/lib/components/accounts/CkBTCWalletFooter.spec.ts
```diff
@@ -19,6 +24,14 @@ jest.mock("$lib/services/ckbtc-minter.services", () => {
 });
 
 describe("CkBTCWalletFooter", () => {
+  beforeAll(() => {
+    jest
+      .spyOn(tokensStore, "subscribe")
+      .mockImplementation(mockTokensSubscribe(mockUniversesTokens));
+  });
+
+  afterAll(() => jest.clearAllMocks());
```
lmuntaner, line 33: How about moving both to `beforeEach`?

lmuntaner (approved):
One question and one test small refactor, but already approving.

frontend/src/tests/lib/components/accounts/CkBTCWalletFooter.spec.ts
peterpeterparker: That's a good question, I know we discusssed this today but just realising it now, moving it in beforeEach will makes the test slower. Executed once vs executed multiple times. I don't need to clean before each test, I want to clean because I used a mock and don't want to propagate it. Actually maybe I don't even need the clear? I'll try.

frontend/src/lib/pages/CkBTCWallet.svelte
peterpeterparker: Ah maybe, I keep it in my mind, good idea 👍. I go a bit further in the development of the feature and see where it goes.

frontend/src/tests/lib/components/accounts/CkBTCWalletFooter.spec.ts
peterpeterparker: Yeah can remove it. Thx.

PR #1880, opened 2023-02-15
Rm unused docker container

bitdivine:
# Motivation
The dockerfile no longer uses the "build" container. The default docker target is scratch, which uses:
```
FROM scratch AS scratch
COPY --from=build_aggregate /build/sns_aggregator.wasm /
COPY --from=build_aggregate /build/sns_aggregator_dev.wasm /
```
Some users _may_ be using docker, giving `build` explicitly as the target.  If so, now is the time to say.  We can look at your use case and figure out the best way forwards.

# Changes
- Remove the `build` target from docker.

# Tests
See CI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,24 +62,6 @@ RUN mkdir -p rs/backend/src rs/sns_aggregator/src && touch rs/backend/src/lib.rs
 COPY dfx.json dfx.json
 RUN DFX_VERSION="$(jq -cr .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
 
-# Start the second container
-FROM builder AS build
-SHELL ["bash", "-c"]
-ARG DFX_NETWORK=mainnet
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -62,24 +62,6 @@ RUN mkdir -p rs/backend/src rs/sns_aggregator/src && touch rs/backend/src/lib.rs
...
-RUN echo "OWN_CANISTER_ID: '$OWN_CANISTER_ID'"
-
-# Build
-# ... put only git-tracked files in the build directory
-COPY . /build
-WORKDIR /build
-RUN find . -type f | sed 's/^..//g' > ../build-inputs.txt
-RUN ./build.sh
-
-RUN ls -sh nns-dapp.wasm; sha256sum nns-dapp.wasm
```
bitdivine, line 81: Note: The standard method of building the nns-dapp.wasm is via another container below.

lmuntaner (approved):

PR #1881, opened 2023-02-15
Local docker builds

bitdivine:
# Motivation
@mraszyk would like local docker builds to work.  They used to a long time ago but have fallen into disuse, are no longer tested and the way we used to pass in config is completely separated from the way we pass in config now.

# Changes
- Delete old `OWN_CANISTER_ID` configuration mechanism from `scripts/docker-build` and `Dockerfile`
- Specify `OWN_CANISTER_ID` for standard local builds in dfx.json
- Dump out the config used in docker builds

# Future
I expect that we will want to be able to do docker builds with custom configurations, not just the standard values in dfx.json.  I suspect that running `config.sh` and then passing the generated config file into docker is the way to go, but let's discuss.

# Tests
Added "local" to the docker build matrix.


Dockerfile
```diff
@@ -68,9 +68,6 @@ SHELL ["bash", "-c"]
 ARG DFX_NETWORK=mainnet
 RUN echo "DFX_NETWORK: '$DFX_NETWORK'"
 
-ARG OWN_CANISTER_ID
```
bitdivine, line 71: Setting this has no effect whatsoever.

Dockerfile
```diff
@@ -124,5 +121,7 @@ RUN ./build-sns-aggregator.sh
 FROM scratch AS scratch
 COPY --from=build_nnsdapp /build/nns-dapp.wasm /
 COPY --from=build_nnsdapp /build/assets.tar.xz /
+COPY --from=build_frontend /build/deployment-config.json /
+COPY --from=build_frontend /build/frontend/.env /frontend-config.sh
```
bitdivine, line 125: Dump out the config that was used in a build.

config.sh
```diff
@@ -20,6 +20,7 @@ set -euo pipefail
 : "Move into the repository root directory"
 pushd "$(dirname "${BASH_SOURCE[0]}")"
 ENV_FILE=${ENV_OUTPUT_FILE:-$PWD/frontend/.env}
+JSON_OUT="deployment-config.json"
```
bitdivine, line 23: Dump out the JSON with all the config, so that we can also see which env vars were set by any code that sourced config.sh.

dfx.json
```diff
@@ -421,6 +421,7 @@
       "config": {
         "FETCH_ROOT_KEY": true,
         "HOST": "http://localhost:8080",
+        "OWN_CANISTER_ID": "qsgjb-riaaa-aaaaa-aaaga-cai",
```
bitdivine, line 424: This is how we currently set OWN_CANISTER_ID

scripts/docker-build
```diff
@@ -11,22 +11,7 @@ cd "$SCRIPTS_DIR/.."
 
 DFX_NETWORK=${DFX_NETWORK:-mainnet}
```
bitdivine, line 13: This mechanism of optionally passing in the canister ID takes no effect whatsoever.  We could patch it up and make it work again but it would be better to come up with a solution that matches our current needs, not a super-simple mechanism that made sense in the early days when we had only one config variable to worry about.

config.sh
```diff
@@ -173,7 +174,9 @@ VITE_AGGREGATOR_CANISTER_URL=${aggregatorCanisterUrl:-}
 VITE_CKBTC_LEDGER_CANISTER_ID=${ckbtcLedgerCanisterId:-}
 VITE_CKBTC_INDEX_CANISTER_ID=${ckbtcIndexCanisterId:-}" | tee "$ENV_FILE"
 
-echo "Config has been defined in '${ENV_FILE}'" >&2
+echo "$json" >"$JSON_OUT"
+echo "Config is available as JSON in '${JSON_OUT}'"
+echo "Frontend config has been defined in '${ENV_FILE}'" >&2
```
dskloetd, line 179: Why does this one go to stderr (&2) and not the previous line?

dskloetd (approved):

config.sh
bitdivine: Good point.  Redirecting both....

bitdivine: Done.

PR #1882, opened 2023-02-15
Specify own canister id for local docker builds

bitdivine:
# Motivation
@mraszyk  needs to be able to build the nns-dapp for local use.

This is the minimal change required to get this to work:
```
DFX_NETWORK=local ./scripts/docker-build
```

# Changes
- Set the default `OWN_CANISTER_ID` for local builds.

# Tests
I have tested the above build locally.

# Future
- Remove the old mechanism for passing information into the docker file.  It is ignored.
- Test local builds
- Agree on a good way of passing custom config into docker builds in future.

Some of this is covered by #1881

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -421,6 +421,7 @@
       "config": {
         "FETCH_ROOT_KEY": true,
         "HOST": "http://localhost:8080",
+        "OWN_CANISTER_ID": "qsgjb-riaaa-aaaaa-aaaga-cai",
         "FEATURE_FLAGS": {
           "ENABLE_SNS_2": true,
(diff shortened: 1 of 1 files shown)


mraszyk (commented):
Tricky to get config.sh find the sns_aggregator canister ID, so let's rather drop calls to config.sh from the docker build.

bitdivine:
Note: I _think_ that this can be closed now, as the dockerfile now has a way of pulling in .dfx without busting the cache with irrelevant changes.  It may be worth using the docker builds as the standard build method, just to make 100% sure.

PR #1879, opened 2023-02-14
GIX-1322: Add featureFlags store to window and method

lmuntaner:
# Motivation

We want to be able to toggle feature flags in mainnet.

# Changes

* Add `setFlag` method to `overrideFeatureFlagsStore`.
* Expose `overrideFeatureFlagsStore` to `window`.

# Tests

* Test for `setFlag` method in `overrideFeatureFlagsStore`.
* Test functionality of `overrideFeatureFlagsStore` in a component using the feature flag.

PR #1883, opened 2023-02-15
Export individual feature flags

dskloetd:
# Motivation

This makes using a feature flag more similar to how it was before https://github.com/dfinity/nns-dapp/pull/1874.
It let's you import and write `$ENABLE_SNS_2` instead of `$featureFlagsStore.ENABLE_SNS_2`.

# Changes

Export feature flags from the module as individual boolean stores.
Also add a type `FeatureKey = keyof FeatureFlags<boolean>` for convenience.

# Tests

Added a test that each defined feature flag is also exported. And updated existing tests.

### DIFF
--- a/frontend/src/lib/constants/environment.constants.ts
+++ b/frontend/src/lib/constants/environment.constants.ts
@@ -12,12 +12,12 @@ export const SNS_AGGREGATOR_CANISTER_URL: string | undefined =
     ? undefined
     : (import.meta.env.VITE_AGGREGATOR_CANISTER_URL as string);
 
-export interface FeatureFlags {
-  ENABLE_SNS_2: boolean;
-  ENABLE_SNS_VOTING: boolean;
-  ENABLE_SNS_AGGREGATOR: boolean;
-  ENABLE_CKBTC_LEDGER: boolean;
-  ENABLE_CKBTC_RECEIVE: boolean;
+export interface FeatureFlags<T> {
+  ENABLE_SNS_2: T;
+  ENABLE_SNS_VOTING: T;
+  ENABLE_SNS_AGGREGATOR: T;
+  ENABLE_CKBTC_LEDGER: T;
+  ENABLE_CKBTC_RECEIVE: T;
 }
 
 /**
@@ -25,7 +25,7 @@ export interface FeatureFlags {
  *
  * @see feature-flags.store.ts to use feature flags
  */
-export const FEATURE_FLAG_ENVIRONMENT: FeatureFlags = JSON.parse(
+export const FEATURE_FLAG_ENVIRONMENT: FeatureFlags<boolean> = JSON.parse(
   import.meta.env.VITE_FEATURE_FLAGS.replace(/\\"/g, '"') ??
     '{"ENABLE_SNS_2":false, "ENABLE_SNS_VOTING": false, "ENABLE_SNS_AGGREGATOR": false, "ENABLE_CKBTC_LEDGER": true, "ENABLE_CKBTC_RECEIVE": false}'
 );
--- a/frontend/src/lib/constants/environment.constants.ts
+++ b/frontend/src/lib/constants/environment.constants.ts
@@ -20,6 +20,8 @@ export interface FeatureFlags<T> {
   ENABLE_CKBTC_RECEIVE: T;
 }
 
+export type FeatureKey = keyof FeatureFlags<boolean>;
+
 /**
  * DO NOT USE DIRECTLY
  *
--- a/frontend/src/lib/pages/CkBTCWallet.svelte
+++ b/frontend/src/lib/pages/CkBTCWallet.svelte
@@ -30,7 +30,6 @@
   } from "$lib/derived/universes-tokens.derived";
   import CkBTCWalletFooter from "$lib/components/accounts/CkBTCWalletFooter.svelte";
   import CkBTCWalletModals from "$lib/modals/accounts/CkBTCWalletModals.svelte";
-  import { ENABLE_CKBTC_RECEIVE } from "$lib/stores/feature-flags.store";
 
   export let accountIdentifier: string | undefined | null = undefined;
 
--- a/frontend/src/lib/pages/CkBTCWallet.svelte
+++ b/frontend/src/lib/pages/CkBTCWallet.svelte
@@ -30,6 +30,7 @@
   } from "$lib/derived/universes-tokens.derived";
   import CkBTCWalletFooter from "$lib/components/accounts/CkBTCWalletFooter.svelte";
   import CkBTCWalletModals from "$lib/modals/accounts/CkBTCWalletModals.svelte";
+  import { ENABLE_CKBTC_RECEIVE } from "$lib/stores/feature-flags.store";
 
   export let accountIdentifier: string | undefined | null = undefined;
 
--- a/frontend/src/lib/stores/feature-flags.store.ts
+++ b/frontend/src/lib/stores/feature-flags.store.ts
@@ -63,9 +63,7 @@ if (browser) {
   (window as any).__featureFlagsStore = overrideFeatureFlagsStore;
 }
 
-const initFeatureFlagStore = (
-  key: FeatureKey
-): Readable<boolean> =>
+const initFeatureFlagStore = (key: FeatureKey): Readable<boolean> =>
   derived(
     overrideFeatureFlagsStore,
     ($overrideFeatureFlagsStore) =>
--- a/frontend/src/lib/stores/feature-flags.store.ts
+++ b/frontend/src/lib/stores/feature-flags.store.ts
@@ -2,6 +2,7 @@ import { browser } from "$app/environment";
 import {
   FEATURE_FLAG_ENVIRONMENT,
   type FeatureFlags,
+  type FeatureKey,
 } from "$lib/constants/environment.constants";
 import { storeLocalStorageKey } from "$lib/constants/stores.constants";
 import { derived, type Readable } from "svelte/store";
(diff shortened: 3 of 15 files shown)


frontend/src/lib/pages/CkBTCWallet.svelte
```diff
@@ -30,6 +30,7 @@
   } from "$lib/derived/universes-tokens.derived";
   import CkBTCWalletFooter from "$lib/components/accounts/CkBTCWalletFooter.svelte";
   import CkBTCWalletModals from "$lib/modals/accounts/CkBTCWalletModals.svelte";
+  import { ENABLE_CKBTC_RECEIVE } from "$lib/stores/feature-flags.store";
```
lmuntaner, line 33: Is this used here?

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -62,10 +63,28 @@ if (browser) {
...
+const initFeatureFlagsStore = (): FeatureFlags<Readable<boolean>> => {
+  let featureFlagStores: Partial<FeatureFlags<Readable<boolean>>> = {};
+  let key: FeatureKey;
+  for (key in FEATURE_FLAG_ENVIRONMENT) {
+    featureFlagStores[key] = initFeatureFlagStore(key);
+  }
+  return featureFlagStores as FeatureFlags<Readable<boolean>>;
+};
+
+export const featureFlagsStore = initFeatureFlagsStore();
```
lmuntaner, line 82: This is exported for testing purposes only, right?

Is it worth mentioning?

lmuntaner (approved):
LGTM! I added a couple of comments

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -12,20 +12,22 @@ export const SNS_AGGREGATOR_CANISTER_URL: string | undefined =
     ? undefined
     : (import.meta.env.VITE_AGGREGATOR_CANISTER_URL as string);
 
-export interface FeatureFlags {
-  ENABLE_SNS_2: boolean;
-  ENABLE_SNS_VOTING: boolean;
-  ENABLE_SNS_AGGREGATOR: boolean;
-  ENABLE_CKBTC_LEDGER: boolean;
-  ENABLE_CKBTC_RECEIVE: boolean;
+export interface FeatureFlags<T> {
```
peterpeterparker, line 15: is this a must? what's the added value of making this generic?

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -62,10 +63,28 @@ if (browser) {
...
 
-export const featureFlagsStore = derived(
-  overrideFeatureFlagsStore,
-  ($overrideFeatureFlagsStore) => ({
-    ...FEATURE_FLAG_ENVIRONMENT,
-    ...$overrideFeatureFlagsStore,
-  })
-);
+const initFeatureFlagStore = (key: FeatureKey): Readable<boolean> =>
+  derived(
```
peterpeterparker, line 67: should these `derived` stores be move the dedicated `derived` package?

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -62,10 +63,28 @@ if (browser) {
...
+    overrideFeatureFlagsStore,
+    ($overrideFeatureFlagsStore) =>
+      $overrideFeatureFlagsStore[key] ?? FEATURE_FLAG_ENVIRONMENT[key]
+  );
+
+const initFeatureFlagsStore = (): FeatureFlags<Readable<boolean>> => {
+  let featureFlagStores: Partial<FeatureFlags<Readable<boolean>>> = {};
+  let key: FeatureKey;
+  for (key in FEATURE_FLAG_ENVIRONMENT) {
```
peterpeterparker, line 76: opiniated feedback but also thinking at the existing code base:

I would not be against having this written in a more javascript functional way rather than imperative way. The code base generally follow that pattern for conversion and mapping.

i.e. not `for (something in ...)` but probably something like `Object.values().reduce()` or else.

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -62,10 +63,28 @@ if (browser) {
...
+    featureFlagStores[key] = initFeatureFlagStore(key);
+  }
+  return featureFlagStores as FeatureFlags<Readable<boolean>>;
```
peterpeterparker, line 79: can we spare this `as` and interfer the type properly?

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -33,15 +34,15 @@ const initOverrideFeatureFlagsStore = (): OverrideFeatureFlagsStore => {
   return {
     subscribe,
 
-    setFlag(flag: keyof FeatureFlags, value: boolean) {
+    setFlag(flag: FeatureKey, value: boolean) {
```
peterpeterparker, line 37: can we use object as parameter?

we don't have lint rules but we follow a rule where as soon as more than two params are available we use objects.

this for maintanability reason (and also because it's safer in js even though it's a bit slower).

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: That would be great. But can the compiler really know that after the for loop, that the map is no longer Partial (as defined on line 74)?

frontend/src/lib/constants/environment.constants.ts
dskloetd: I use it both as `FeatureFlags<boolean>` and `FeatureFlags<Readable<boolean>>` and I want to make sure they have the same keys. Do you have another suggestions?

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: Yes, but note that I'm not changing the number of parameters in this PR.
Do you want that to be included in this PR or shall I make a separate PR?
This was added in https://github.com/dfinity/nns-dapp/pull/1879

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: I don't know?
Can you agree on the answer with Llorenc and I'll just do what you decide?
Packages based on type rather than topic seems a bit odd to me to be honest, but I want to be consistent with what you do.

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: I gave it a shot but what I have now seems less type safe and not more readable.
Maybe you can improve it? Or what do you think?

frontend/src/lib/pages/CkBTCWallet.svelte
dskloetd: Sorry, I had already removed it but forgot to push.

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: Good point. I don't think we need it even in the tests. So I removed the export and updated the test.

frontend/src/lib/constants/environment.constants.ts
peterpeterparker: Thanks for the feedback, sounds good

frontend/src/lib/stores/feature-flags.store.ts
peterpeterparker: If you discussed with @lmuntaner and decided to put that derived store there for sure it's fine to me.

Beside that, yes it's by topic.

frontend/src/lib/stores/feature-flags.store.ts
peterpeterparker: what solution did you try?

frontend/src/lib/stores/feature-flags.store.ts
lmuntaner: Fine with me either place.

I believe that the derived stores that interact with more than one store should be in `derived`. If it's only convenient, it might be annoying to have them in a separate file.

frontend/src/lib/stores/feature-flags.store.ts
lmuntaner: We want to use them in the console and I thought it would be better to have it as string and boolean.

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: You can see it right here in the code:
https://github.com/dfinity/nns-dapp/pull/1883/files#diff-7ba7578b57416e031f416e704f60aa79780a0b13eba9a2c385419f4598c8b7d1R73

frontend/src/lib/stores/feature-flags.store.ts
peterpeterparker: don't have it neither spontaneoulsy, ok then

frontend/src/lib/stores/feature-flags.store.ts
peterpeterparker: Another PR sounds fine, thanks

peterpeterparker: Can we add a comment or is there already a command about the usage in the console (maybe I missed it because of the shrinked view on github)?

peterpeterparker: Makes sense then

peterpeterparker: That said, speaking out load, maybe would be nice to expose only a particular function and not a function of the store directly to the window object. Like having a cli within the console 😉

lmuntaner: I believe I can build an easy UI like I did for the SNS neuron permissions.

That's even better and security might be happy also because we don't expose anything to `window`.

I'll do that next week though.

frontend/src/lib/stores/feature-flags.store.ts
peterpeterparker: Have you try following? You can simplify this code with a single line by destructing an array instead of a record.
That said, don't know how to avoid the as there neither.

```
export const [
  ENABLE_SNS_2,
  ENABLE_SNS_VOTING,
  ENABLE_SNS_AGGREGATOR,
  ENABLE_CKBTC_LEDGER,
  ENABLE_CKBTC_RECEIVE,
] = Object.keys(FEATURE_FLAG_ENVIRONMENT).map((key) =>
  initFeatureFlagStore(key as keyof FeatureFlags<boolean>)
);
```

dskloetd: If you destructure an array instead of an object, if you get the order of the features wrong, you won't notice and they would silently swap their functionality, wouldn't they?

peterpeterparker: That's a good point. Let's keep as it. Thanks!

peterpeterparker (approved):
LGTM, thx

PR #1884, opened 2023-02-15
feat: convert ckBTC to BTC (services only)

peterpeterparker:
# Motivation

This PR introduces the JS Api and Services layers to perform a conversion of ckBTC to BTC.

# Note

These services and API are provided without UI changes in this PR.

# PRs

- [x] https://github.com/dfinity/ic-js/pull/292

# Changes

- implements APIs for `get_withdrawal_account` and `retrieve_btc`
- implement service logic that converts ckBTC to BTC
- create an interface `IcrcTransferTokensUserParams` for our internal call to transfer tokens function


frontend/src/lib/services/ckbtc-convert.services.ts
```diff
@@ -0,0 +1,143 @@
...
+    amount,
+    destinationAddress: ledgerAddress,
+    loadTransactions: false,
+  });
+
+  if (!transferSuccess) {
+    return { success: false };
+  }
+
+  // TODO(GIX-1324): how do we handle failure between these steps UX wise since
```
lmuntaner, line 73: Type "these steps UX wise since", since what? hehe

frontend/src/lib/services/ckbtc-convert.services.ts
```diff
@@ -0,0 +1,143 @@
...
+      identity,
+      address: bitcoinAddress,
+      amount: numberToE8s(amount),
+    });
+  } catch (err: unknown) {
+    toastsError(toastRetrieveBtcError(err));
+
+    return { success: false };
+  } finally {
+    await loadCkBTCAccountTransactions({
```
lmuntaner, line 86: Will this run even if `retrieveBtc` fails?

You have a `return` inside it.

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  const params = {
+    source: mockCkBTCMainAccount,
+    destinationAddress: mockCkBTCAddress,
+    amount: 1,
+  };
+
+  const convert = async () => await convertCkBTCToBtc(params);
+
+  beforeAll(() => {
```
lmuntaner, line 44: Would `beforeEach` work?

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    jest.spyOn(console, "error").mockImplementation(() => undefined);
+
+    const now = Date.now();
+    jest.useFakeTimers().setSystemTime(now);
+  });
+
+  beforeEach(() => jest.clearAllMocks());
+
+  afterAll(() => jest.clearAllTimers());
```
lmuntaner, line 57: If you set it in `beforeEach` does any other test fail?

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+
+        it("should load transactions", async () => {
+          await convert();
+
+          // We only test that the call is made here. Test should be covered by its respective service.
+          expect(loadCkBTCAccountTransactions).toBeCalled();
+        });
+      });
+
+      describe("retrieve btc fails", () => {
```
lmuntaner, line 141: Do we want to load transactions or not in this case?

It would be good to have a test to ensure that we load or not the transactions

frontend/src/lib/services/ckbtc-convert.services.ts
peterpeterparker: Since ... I speak badly english 🤦‍♂️. Thanks for the hint, corrected!

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
peterpeterparker: I guess so but it's a performance reason. These mocks are common for the all tests, so if I move them in beforeEach it will become slower for no particular reason.

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
peterpeterparker: It would work sure but same as above, kind of think it will lower the perf. Marking as resolving but for sure if you really rather like to move everything in beforeEach let me know, not that a strong opinion about this

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
peterpeterparker: There is one, the test just after this test in the same decribe

```
it("should load transactions", async () => {
          await convert();

          // We only test that the call is made here. Test should be covered by its respective service.
          expect(loadCkBTCAccountTransactions).toBeCalled();
        });
```

That said, I spared my self the test of the reload of the transactions not successing. As for success, it is already covered by another dedicated test.

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
lmuntaner: I believe that consistency is more important than this small performance gain in the unit tests.

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
lmuntaner: Yes, that's the test I was wondering.

If you think it's not important to test and you spared it, it's ok. Not a strong opinion.

frontend/src/lib/services/ckbtc-convert.services.ts
lmuntaner: You have a `return { success: false };` in the `catch` block.

If you hit that `return`, the finally will never be executed, right?

peterpeterparker: No. The `finally` is executed no matter what.

frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
peterpeterparker: With more than 2'000 tests, performance matter in my opinion, but I'll change to `beforeEach`.

peterpeterparker: Done

frontend/src/lib/services/ckbtc-convert.services.ts
lmuntaner: Oh wow, I just tried. I thought `return` was always exiting the function. Thanks!

That's very weird...

And if there is a `return` in the `finally` it overried the other returns. But if there is no `return` there, then it uses the one of the blocks.

This is pretty confusing...

Anyway, nothing to do in this PR, thanks!

lmuntaner (approved):

PR #1885, opened 2023-02-15
Chore: Restrict feature flags functionality

lmuntaner:
# Motivation

Do not allow to set any feature flags. Ideally, we would only allow the ones approved by security.

# Changes

* Rename `assertFeatureFlag` to `assertValidFeatureFlag` in `features-flags.store.ts`.
* Add list of EDITABLE_FEATURE_FLAGS.
* Check that flag is included in EDITABLE_FEATURE_FLAGS in `assertValidFeatureFlag`.

# Tests

* Test new functionality.

### DIFF
--- a/frontend/src/tests/lib/stores/feature-flags.store.spec.ts
+++ b/frontend/src/tests/lib/stores/feature-flags.store.spec.ts
@@ -9,9 +9,9 @@ import { get } from "svelte/store";
 describe("featureFlags store", () => {
   const noKey = "NO_KEY" as FeatureKey;
   const noKeyError = `Unknown feature flag: ${noKey}`;
-  const notEditable = "TEST_FLAG_NOT_EDITABLE" as FeatureKey;
+  const notEditable: FeatureKey = "TEST_FLAG_NOT_EDITABLE";
   const notEditableError = `Feature flag is not editable: ${notEditable}`;
-  const editableFlag = "TEST_FLAG_EDITABLE" as FeatureKey;
+  const editableFlag: FeatureKey = "TEST_FLAG_EDITABLE";
   beforeEach(() => {
     overrideFeatureFlagsStore.reset();
   });
--- a/frontend/src/tests/lib/stores/feature-flags.store.spec.ts
+++ b/frontend/src/tests/lib/stores/feature-flags.store.spec.ts
@@ -9,8 +9,9 @@ import { get } from "svelte/store";
 describe("featureFlags store", () => {
   const noKey = "NO_KEY" as FeatureKey;
   const noKeyError = `Unknown feature flag: ${noKey}`;
-  const noEditable = "ENABLE_SNS_VOTING" as FeatureKey;
-  const noEditableError = `Feature flag is not editable: ${noEditable}`;
+  const notEditable = "TEST_FLAG_NOT_EDITABLE" as FeatureKey;
+  const notEditableError = `Feature flag is not editable: ${notEditable}`;
+  const editableFlag = "TEST_FLAG_EDITABLE" as FeatureKey;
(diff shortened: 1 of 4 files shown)


lmuntaner:

frontend/src/tests/lib/stores/feature-flags.store.spec.ts
```diff
@@ -8,7 +8,9 @@ import { get } from "svelte/store";
 
 describe("featureFlags store", () => {
   const noKey = "NO_KEY" as FeatureKey;
-  const error = new Error(`Unknown feature flag: ${noKey}`);
+  const noKeyError = `Unknown feature flag: ${noKey}`;
+  const noEditable = "ENABLE_SNS_VOTING" as FeatureKey;
```
dskloetd, line 12: Before we hard code more feature flags into test, can we just make some custom feature flags for testing only which aren't going to be launched so we won't have to constantly change which feature flags we use for testing?

dskloetd, line 12: notEditable (missing t)

frontend/src/tests/lib/stores/feature-flags.store.spec.ts
```diff
@@ -8,7 +8,9 @@ import { get } from "svelte/store";
...
+  const noEditableError = `Feature flag is not editable: ${noEditable}`;
```
dskloetd, line 13: also missing t

dskloetd (approved):

frontend/src/tests/lib/stores/feature-flags.store.spec.ts
lmuntaner: done

frontend/src/tests/lib/stores/feature-flags.store.spec.ts
lmuntaner: thanks

frontend/src/tests/lib/stores/feature-flags.store.spec.ts
lmuntaner: thanks

frontend/src/tests/lib/stores/feature-flags.store.spec.ts
```diff
@@ -8,7 +8,10 @@ import { get } from "svelte/store";
 
 describe("featureFlags store", () => {
   const noKey = "NO_KEY" as FeatureKey;
-  const error = new Error(`Unknown feature flag: ${noKey}`);
+  const noKeyError = `Unknown feature flag: ${noKey}`;
+  const notEditable = "TEST_FLAG_NOT_EDITABLE" as FeatureKey;
+  const notEditableError = `Feature flag is not editable: ${notEditable}`;
+  const editableFlag = "TEST_FLAG_EDITABLE" as FeatureKey;
```
dskloetd, line 14: Nit: Do you need all these casts?
It seems nicer if you could just declare them as the type so that TS would complain if they are not of the correct type:
```
  const notEditable : FeatureKey = "TEST_FLAG_NOT_EDITABLE";
  const editableFlag : FeatureKey = "TEST_FLAG_EDITABLE";
```
I expect only for `noKey` do you actually need to force a cast.

dskloetd (approved):
Thanks!

frontend/src/tests/lib/stores/feature-flags.store.spec.ts
lmuntaner: You're right, probably even without the type. But cast should not be needed

lmuntaner: done

PR #1886, opened 2023-02-15
Install jest-expect-message

dskloetd:
# Motivation

It can be useful to debug a test if some expectations have a custom message conveying some state.
NPM package  `jest-expect-message` lets you pass a second (message) parameter to expect which is display when the expectation fails.
This is especially useful when an expectation inside a for loop fails because the message can indicate which iteration of the loop fails.
It's also useful when helper functions do expectation.

# Changes

Install jest-expect-message.
And use it as an example in `feature-flags.store.ts` to indicate which feature flag isn't exported if any of them isn't.

# Tests


mstrasinskis (approved):
Thank you!

PR #1888, opened 2023-02-15
Remove unused fields from MultiPartTransactionsProcessor

dskloetd:
# Motivation

This is the final step of removing unused fields `statuses` and `errors` from `MultiPartTransactionsProcessor`.
The previous PR (https://github.com/dfinity/nns-dapp/pull/2024) left some code in place to make upgrading and downgrading work.
Now it's time to clean up that code.

# Changes

1. Remove MultiPartTransactionsProcessorWithRemovedFields.
2. Stop encoding accounts_store with the old fields included.
3. Remove tests that encode between MultiPartTransactionsProcessorWithRemovedFields and MultiPartTransactionsProcessor.
4. Remove MultiPartTransactionStatus and MultiPartTransactionError which are no longer used.

# Tests
Manually upgraded and downgraded on local replica.
Manually created a subaccount and transferred some ICP on local replica.


lmuntaner (commented):
Thanks!

Did you also try redeploying in local?

peterpeterparker (changes requested):
This will break the upgrade of the canister I think.

`MultiPartTransactionsProcessor` is used in `AccountsStore` which is part of the stable memory.

dskloetd:
> This will break the upgrade of the canister I think.
> 
> `MultiPartTransactionsProcessor` is used in `AccountsStore` which is part of the stable memory.
> 
> So, I am assuming that removing some non optional variable of the state without handling the changes in the `post_upgrade` will break the deserialization.
> 
> Have you test upgrades?

Did you read the Tests section of the PR description and look at the branch linked from the Motivation?

I will still try to test deploying to local replica but if your concern is not addressed by the PR description already I'm not sure what you're referring to.

peterpeterparker:
> I will still try to test deploying to local replica but if your concern is not addressed by the PR description already I'm not sure what you're referring to.

Yes my concern will be addressed with a real test.

Let us know if deploying current NNS-dapp, using the dapp and then trying to upgrade the wasm with your changes works out.

dskloetd:
I did the following.

1. Deploy a version of nns-dapp from main.
2. Check on the replica dashboard that the hash of the nns-dapp canister is the deployed one.
3. In the app, create an account, transfer some ICP and create a neuron.
4. Deploy a version of nns-dapp from the branch with unused fields removed.
5. Check on the replica dashboard that the hash of the nns-dapp canister is the deployed one.
6. Reload (empty cache and hard reload) the page and check that the transactions and neuron are still there. 
7. Look through the local replica logs for any signs of problems. (Although I didn't see even 1 line indicating a deploy had happened.)

peterpeterparker (approved):
Thanks for the extra tests David.

Given my sometimes painful experience of migrating canisters on mainnet and because I am a Rust newbie, I would rather like to not proceed with this PR unless someone else than me confirm that indeed deleting fields in the stable memory is all fine. Just think the cleanup is not necessary worth the risk.

That said, because it has been legitimely well tested, not up to me to block this. Therefore I approve the PR because I previously blocked it and hands over the decision.

lmuntaner (approved):

dskloetd:
On Llorenç' request I'll wait 2 weeks before merging this.
I'll check back on March 3rd.

dskloetd:
I'll wait until after https://github.com/dfinity/nns-dapp/pull/2024 has been released for a while before merging this.

dskloetd:
Llorenc, can you have another look?

lmuntaner (approved):
LGTM!

PR #1889, opened 2023-02-15
Expose feature flags in the console in a way that allows for autocomplete

dskloetd:
# Motivation

Making it easier to find the correct feature flags in the console.
By having the names of the feature flags directly on the object, instead of passed as strings, the console will autocomplete those names for you.

Enabling a feature now goes like this:
`> __featureFlags.ENABLE_SNS_AGGREGATOR.overrideWith(true);`

You can also get a list of features and states like this:
```
> __featureFlags.list()
ENABLE_CKBTC_LEDGER false (override undefined default false )
ENABLE_CKBTC_RECEIVE false (override undefined default false )
ENABLE_SNS_2 true (override undefined default true )
ENABLE_SNS_AGGREGATOR false (override undefined default false )
ENABLE_SNS_VOTING true (override undefined default true )
```

# Changes

Expose the feature flags in the console with fields with the names of the features on the window._featureFlags object.

# Tests

Tested manually in the console and tests added to feature-flags.store.spec.ts

### DIFF
--- a/frontend/src/lib/stores/feature-flags.store.ts
+++ b/frontend/src/lib/stores/feature-flags.store.ts
@@ -104,7 +104,8 @@ const listFeatureFlagsToConsole = () => {
   }
 };
 
-const initConsoleInterface = (): FeatreFlagsConsoleInterface => {
+// Exported for testing.
+export const initConsoleInterface = (): FeatreFlagsConsoleInterface => {
   const consoleInterface: Partial<FeatreFlagsConsoleInterface> = {};
   let key: FeatureKey;
   for (key in FEATURE_FLAG_ENVIRONMENT) {
--- a/frontend/src/lib/stores/feature-flags.store.ts
+++ b/frontend/src/lib/stores/feature-flags.store.ts
@@ -5,7 +5,7 @@ import {
   type FeatureKey,
 } from "$lib/constants/environment.constants";
 import { storeLocalStorageKey } from "$lib/constants/stores.constants";
-import { derived, type Readable } from "svelte/store";
+import { derived, get, type Readable } from "svelte/store";
 import { writableStored } from "./writable-stored";
 
 type OverrideFeatureFlagsData = Partial<FeatureFlags<boolean>>;
@@ -67,9 +67,58 @@ const initOverrideFeatureFlagsStore = (): OverrideFeatureFlagsStore => {
 // Exported for testing purposes
 export const overrideFeatureFlagsStore = initOverrideFeatureFlagsStore();
 
+interface FeatreFlagConsoleInterface {
+  overrideWith(value: boolean): void;
+  removeOverride(): void;
+}
+
+interface FeatreFlagsConsoleInterface
+  extends FeatureFlags<FeatreFlagConsoleInterface> {
+  list(): void;
+}
+
+const initSingleFeatureConsoleInterface = (
+  key: FeatureKey
+): FeatreFlagConsoleInterface => ({
+  overrideWith: (value: boolean) =>
+    overrideFeatureFlagsStore.setFlag(key, value),
+  removeOverride: () => overrideFeatureFlagsStore.removeFlag(key),
+});
+
+const listFeatureFlagsToConsole = () => {
+  const overrideStates = get(overrideFeatureFlagsStore);
+  let key: FeatureKey;
+  for (key in FEATURE_FLAG_ENVIRONMENT) {
+    const override = overrideStates[key];
+    const defaultValue = FEATURE_FLAG_ENVIRONMENT[key];
+    const value = override ?? defaultValue;
+    console.log(
+      key,
+      value,
+      "(override",
+      override,
+      "default",
+      defaultValue,
+      ")"
+    );
+  }
+};
+
+const initConsoleInterface = (): FeatreFlagsConsoleInterface => {
+  const consoleInterface: Partial<FeatreFlagsConsoleInterface> = {};
+  let key: FeatureKey;
+  for (key in FEATURE_FLAG_ENVIRONMENT) {
+    consoleInterface[key] = initSingleFeatureConsoleInterface(key);
+  }
+  return {
+    ...consoleInterface,
+    list: listFeatureFlagsToConsole,
+  } as FeatreFlagsConsoleInterface;
+};
+
 if (browser) {
   /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  (window as any).__featureFlagsStore = overrideFeatureFlagsStore;
+  (window as any).__featureFlags = initConsoleInterface();
 }
 
 const initFeatureFlagStore = (key: FeatureKey): Readable<boolean> =>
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
😍

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -67,9 +67,59 @@ const initOverrideFeatureFlagsStore = (): OverrideFeatureFlagsStore => {
...
+      override,
+      "default",
+      defaultValue,
+      ")"
+    );
+  }
+};
+
+// Exported for testing.
+export const initConsoleInterface = (): FeatreFlagsConsoleInterface => {
```
peterpeterparker, line 108: typo `Featre...`

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -67,9 +67,59 @@ const initOverrideFeatureFlagsStore = (): OverrideFeatureFlagsStore => {
...
+  return {
+    ...consoleInterface,
+    list: listFeatureFlagsToConsole,
+  } as FeatreFlagsConsoleInterface;
+};
+
 if (browser) {
   /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  (window as any).__featureFlagsStore = overrideFeatureFlagsStore;
+  (window as any).__featureFlags = initConsoleInterface();
```
peterpeterparker, line 122:
```
(window as unknown as {__featureFlags: FeatureFlagsConsoleInterface} = initConsoleInterface();
```

would be a bit cleaner (I read once that the best practice now is to avoid `any` and use `unknown` for worth case scenario)

note we can also add some type for the `window` in `app.d.ts` I guess.

peterpeterparker: note: since it inject something in the window maybe it would be even nicer to have it in a `+layout.svelte` for separation of concern of the various layers but just thinking at loud

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -67,9 +67,59 @@ const initOverrideFeatureFlagsStore = (): OverrideFeatureFlagsStore => {
...
+});
+
+const listFeatureFlagsToConsole = () => {
+  const overrideStates = get(overrideFeatureFlagsStore);
+  let key: FeatureKey;
+  for (key in FEATURE_FLAG_ENVIRONMENT) {
+    const override = overrideStates[key];
+    const defaultValue = FEATURE_FLAG_ENVIRONMENT[key];
+    const value = override ?? defaultValue;
+    console.log(
```
peterpeterparker, line 95: just in case you would like the idea:

```
console.log(`${key} ${value} (override ${override} default ${defaultValue})`);
```

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -67,9 +67,59 @@ const initOverrideFeatureFlagsStore = (): OverrideFeatureFlagsStore => {
...
+): FeatreFlagConsoleInterface => ({
+  overrideWith: (value: boolean) =>
+    overrideFeatureFlagsStore.setFlag(key, value),
+  removeOverride: () => overrideFeatureFlagsStore.removeFlag(key),
+});
+
+const listFeatureFlagsToConsole = () => {
+  const overrideStates = get(overrideFeatureFlagsStore);
+  let key: FeatureKey;
+  for (key in FEATURE_FLAG_ENVIRONMENT) {
```
peterpeterparker, line 91: if you want to use a `for`then here `for (const key in ...)` - i.e. replace the `ley key` with `const`

frontend/src/lib/stores/feature-flags.store.ts
```diff
@@ -67,9 +67,59 @@ const initOverrideFeatureFlagsStore = (): OverrideFeatureFlagsStore => {
...
+      ")"
+    );
+  }
+};
+
+// Exported for testing.
+export const initConsoleInterface = (): FeatreFlagsConsoleInterface => {
+  const consoleInterface: Partial<FeatreFlagsConsoleInterface> = {};
+  let key: FeatureKey;
+  for (key in FEATURE_FLAG_ENVIRONMENT) {
```
peterpeterparker, line 111: ditto

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: I'm happy to try but I'm not sure what `+layout.svelte` is so some context or a pointer would be appreciated.

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: I used `let` (also in other place) because you can't specify a type inside the `for ... in` and then it complains that `key` is a string.
But for some reason, on this particular line it doesn't complain so I did change it to const here.

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: Yes, done.

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: Done

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: Here I can't because then it complains that key is a string.

frontend/src/lib/stores/feature-flags.store.ts
peterpeterparker: My editor and lint would appreciate `const`  as well.

Example screenshot.

Capture d’écran 2023-02-15 à 18 01 27

peterpeterparker: `npm run lint` as well

frontend/src/lib/stores/feature-flags.store.ts
peterpeterparker: I was just thinking out at loud like I said. I was thinking at `+layout.svelte` because it's closer, at least in my mind, to the dom and therefore to the window object but, once again, was just something that crossed my mind.

On the other side, my above [comment](https://github.com/dfinity/nns-dapp/pull/1889#discussion_r1107329674) about `any` vs `(window as unknown as {__featureFlags: FeatureFlagsConsoleInterface} = initConsoleInterface();` is maybe more relevant?

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: Actually, this line has the same problem so I had to revert back to `let key` here as well.

peterpeterparker: Cool, I guess my branch is outdated. Thanks for the feedback.

frontend/src/lib/stores/feature-flags.store.ts
dskloetd: Happy to oblige (done) but would still like to learn what `+layout.svelte` is about.

peterpeterparker: I can guide you through next week, ping me. Thanks for the update!

peterpeterparker (approved):
LGTM, thanks

PR #1892, opened 2023-02-16
Expose cycles

bitdivine:
# Motivation
We would like to monitor the cycle balance and usage of the `sns_aggregator` canister.  The balance is available externally to controllers only, so we need to make the relevant API call ourselves to get the canister info and return the value.

# Changes
- Add a method to get the standard canister info.
- Add documentation to the Notion [on-call page](https://www.notion.so/SNS-Aggregator-Change-Log-6bcd4370e5704cc08c473975c8252fdf?pvs=4#8050c6b7660442a8a1ca6252b076ca4b) for the sns-aggregator.

Note: This is the same underlying call made by a number of other canisters.  E.g. here is the call being made by the sns swap canister: https://github.com/dfinity/ic/blob/master/rs/sns/swap/src/clients.rs#L167

Note: The call changes no values, however the inner api call to get the balance still needs to be an update call.  I checked that this is still the case and tried switching it to a query call.  That didn't work.

# Tests
See:
```
$ dfx canister call --network ic sns_aggregator get_canister_status
(
  record {
    controller = principal "wgfa2-l65ry-bjl6l-idehu-56rsz-srpsc-km5fe-jgglp-p5up7-npbep-5ae";
    status = variant { running };
    freezing_threshold = 2_592_000 : nat;
    balance = vec { record { blob "\00"; 7_328_277_541_510 : nat } };
    memory_size = 13_146_832 : nat;
    cycles = 7_328_277_541_510 : nat;
    settings = record {
      controller = principal "wgfa2-l65ry-bjl6l-idehu-56rsz-srpsc-km5fe-jgglp-p5up7-npbep-5ae";
      freezing_threshold = 2_592_000 : nat;
      controllers = vec {
        principal "wgfa2-l65ry-bjl6l-idehu-56rsz-srpsc-km5fe-jgglp-p5up7-npbep-5ae";
      };
      memory_allocation = 0 : nat;
      compute_allocation = 0 : nat;
    };
    idle_cycles_burned_per_day = 351_377_760 : nat;
    module_hash = opt blob "vk\dc\aa\de\5c0A@\aa\60\c7\d3+\af*b\27\d1\e6\d0\cf_\a6\1b\c3:\f48\dd!\ae";
  },
)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -958,6 +958,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "dfn_candid"
+version = "0.8.0"
+source = "git+https://github.com/dfinity/ic?rev=d14361f9939baaeb899106b874851ad4a0ce928b#d14361f9939baaeb899106b874851ad4a0ce928b"
+dependencies = [
+ "candid",
+ "dfn_core 0.8.0 (git+https://github.com/dfinity/ic?rev=d14361f9939baaeb899106b874851ad4a0ce928b)",
+ "ic-base-types 0.8.0 (git+https://github.com/dfinity/ic?rev=d14361f9939baaeb899106b874851ad4a0ce928b)",
+ "on_wire 0.8.0 (git+https://github.com/dfinity/ic?rev=d14361f9939baaeb899106b874851ad4a0ce928b)",
+ "serde",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


lmuntaner (approved):
LGTM, thanks!

PR #1893, opened 2023-02-16
GIX-1326: Add missing param to SNS Aggregator

lmuntaner:
# Motivation

There was a missing parameter in the SNS Aggregator `decentralization_sale_open_timestamp_seconds`

# Changes

* Add missing param in SNS Aggregator.

# Tests

* Tests are not setup yet.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -792,7 +792,10 @@
         "HOST": "https://medium07.testnet.dfinity.network",
         "FEATURE_FLAGS": {
           "ENABLE_SNS_2": true,
-          "ENABLE_SNS_VOTING": true
+          "ENABLE_SNS_VOTING": true,
+          "ENABLE_SNS_AGGREGATOR": true,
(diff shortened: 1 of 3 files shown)


lmuntaner:
@bitdivine please review

bitdivine (approved):
Note: The same can be acheived more thoroughly (and with more work, as ore issues need to be taken into account) with the ic-commit update script.  That said, if you need a field quickly I have nothing against adding it like this until the next full ic commit bump.

lmuntaner:
@bitdivine we can't merge it like this. There is a CI job that checks that the types match the IC Commit. Which makes sense and it's nice.

How can I run the ic-commit update script? Or can you do it for now and explain me or have some docs no how to do it and what to take into account?

Thanks!

lmuntaner:
Closing in favor of #1903

PR #1397, opened 2022-10-02
build: sveltekit (without routing changes)

peterpeterparker:
# Motivation

Use an opiniated configuration with SvelteKit and ViteJS. This brings best practice to the dapp, improve the DX and make the project easier to jump in for the community.

# Note

This PR introduces SvelteKit but keep current custom routing system.

# Changes

- replace `rollup` build with `SvelteKit` (which uses `vitejs` and ... `rollup`)
- update all JS import updated to ` import ... from $lib/...`
- ts config and compilation set to `strict: true`
- `deployment-configuration.json` is replaced by `.env` and `HACKING.md` documented accordingly
- update script that build `index.html` to "only" parse the CSP
- remove the custom script loader from `index.html` because Sveltekit injects the script loader
- move `index.html` content to `app.html`
- move previous `routes` to `lib/routes`
- create a single `+page.svelte` that render our custom routes
- use adapter static = prerender the application
- remove `BUILD` info from `dfx.json` - not used anymore
- remove `v2/index.html`

# Tests

- npm run dev / build :white_check_mark:
- npm run test / check / lint  :white_check_mark:
- CI :white_check_mark:
- deploy on a canister https://tqtu6-byaaa-aaaaa-aaana-cai.small13.dfinity.network/  :white_check_mark:

PR #1894, opened 2023-02-16
Fix actually using the ENABLE_CKBTC_LEDGER store.

dskloetd:
# Motivation

1. `npm run lint` was accidentally removed from `npm run check` (in https://github.com/dfinity/nns-dapp/pull/1397) and therefore not run on CI.
2. So we failed to catch an unused variable and were using a store (always true) instead of the value of the store (true or false).

The specific issue is that we always consider ENABLE_CKBTC_LEDGER to be true for the purpose of determining the selectedUniverseId.

# Changes

1. Correctly use the value of the ENABLE_CKBTC_LEDGER.
2. Add a unit test where we disabled ENABLE_CKBTC_LEDGER and verify that it affects the selected universe.
3. Change `feature-flags.store.ts` to allow overriding non-editable feature flags.
4. Add unit tests to make sure that non-editable feature flags still can't be set through the console interface.

# Tests
Unit tests in `selected-universe.derived.spec.ts` and `feature-flags.store.spec.ts`.
Verified that the unit tests fail without the fixes.
No manual testing.

### DIFF
--- a/frontend/src/lib/derived/selected-universe.derived.ts
+++ b/frontend/src/lib/derived/selected-universe.derived.ts
@@ -44,9 +44,9 @@ export const selectedUniverseIdStore: Readable<Principal> = derived<
   Principal
 >(
   [pageUniverseIdStore, pageStore, ENABLE_CKBTC_LEDGER],
-  ([canisterId, page, featureFlags]) => {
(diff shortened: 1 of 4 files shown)


peterpeterparker (approved):
LGTM, thx

PR #1895, opened 2023-02-16
Migration to nns.internetcomputer.org

lmuntaner:
# Motivation

We need to move from `nns.ic0.app` to `nns.internetcomputer.org`.

The requests should move from `ic0.app` to `ic-api.io`.

# Changes

* Add `.well-known` assets to use custom domain `nns.internetcomputer.org`.
* Rename `nns.ic0.app` to `nns.internetcomputer.org`.
* Rename host from `ic0.app` to `ic-api.io`.
* Change identity provider in mainnet to `identity.internetcomputer.org`.
* Add a rule to use `identity.ic0.app` in case the user lands in `nns.ic0.app`.
* Add new urls to CSP rules.

# Tests

* Test the specific rule that should return the old identity provider url.


lmuntaner:
@mstrasinskis @peterpeterparker @bitdivine please review

I deployed it to the nnsdapp: https://xub3y-eqaaa-aaaaa-aaawq-cai.nnsdapp.testnet.dfinity.network/

lmuntaner:
@frederikrothenberger maybe it would be good if you can check that we are not missing anything for the custom domain

🙏

dfx.json
```diff
@@ -223,7 +223,7 @@
           "ENABLE_SNS_2": false,
           "ENABLE_SNS_VOTING": false,
           "ENABLE_SNS_AGGREGATOR": false,
-          "ENABLE_CKBTC_LEDGER": false
```
frederikrothenberger, line 226: Is this related to the migration?

dfx.json
```diff
@@ -223,7 +223,7 @@
...
+          "ENABLE_CKBTC_LEDGER": true
```
mstrasinskis, line 226: Should it be enabled?

frontend/scripts/build.csp.mjs
```diff
@@ -144,6 +144,8 @@ const cspConnectSrc = () => {
   // TODO: Use `URL` to check if the URL is valid and not introduce a security issue
   const src = [
     process.env.VITE_IDENTITY_SERVICE_URL,
+    // We move to internetcomputer.org, but if a user access the app with the old URL, we need to allow it
+    "https://identity.ic0.app",
```
frederikrothenberger, line 148: Shouldn't this be `https://nns.ic0.app`?

frontend/src/lib/stores/auth.store.ts
```diff
@@ -55,7 +75,7 @@ const initAuthStore = () => {
       authClient = authClient ?? (await createAuthClient());
 
       await authClient?.login({
-        identityProvider: IDENTITY_SERVICE_URL,
+        identityProvider: getIdentityProvider(),
```
frederikrothenberger, line 78: You need to start on passing in the `derivationOrigin: https://nns.ic0.app`.

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -32,7 +32,7 @@ export const {
 export const IS_TESTNET: boolean =
   DFX_NETWORK !== "mainnet" &&
   FETCH_ROOT_KEY === true &&
-  !HOST.includes(".ic0.app");
+  !HOST.includes(".ic-api.io");
```
mstrasinskis, line 35: Shouldn't both be supported for some time? Asking because of this comment https://github.com/dfinity/nns-dapp/pull/1895/files#diff-ef8663875b35cfc44789b6c8e8004e0064a2fb43f337f792126deeef55529c92R147

frederikrothenberger (commented):
I'm missing a lot of context. Left some comments. Definitely have somebody more familiar with the NNS dapp review it as well.

frontend/src/lib/rest/governance-metrics.rest.ts
```diff
@@ -1,6 +1,6 @@
 // TODO: Extract .env
 const GOVERNANCE_CANISTER_RAW_URL =
-  "https://rrkah-fqaaa-aaaaa-aaaaq-cai.raw.ic0.app";
+  "https://rrkah-fqaaa-aaaaa-aaaaq-cai.raw.ic-api.io";
```
nmattia, line 3: I think you want `icp0.io` here, `ic-api.io` (AFAICT) would only be for canister calls

frontend/src/lib/stores/auth.store.ts
```diff
@@ -16,6 +17,25 @@ export interface AuthStore {
 // @see agent-js issue [#618](https://github.com/dfinity/agent-js/pull/618)
 let authClient: AuthClient | undefined | null;
 
+const getIdentityProvider = () => {
+  const environmentUrl = IDENTITY_SERVICE_URL;
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const location = (window as any)?.location;
```
peterpeterparker, line 24: - `if (!browser) { return .... }` maybe?

- is the cast to `any` a must?

rs/backend/src/assets.rs
```diff
@@ -218,6 +218,10 @@ pub fn http_request(req: HttpRequest) -> HttpResponse {
                     if let Some(content_encoding_header) = content_encoding.header() {
                         headers.push(("Content-Encoding".to_string(), content_encoding_header.to_string()));
                     }
+                    // Assets within .well-known are used by II and should be accessible
+                    if request_path.starts_with("/.well-known") {
+                        headers.push(("Access-Control-Allow-Origin".to_string(), "*".to_string()));
```
peterpeterparker, line 223: @frederikrothenberger `*` is fine or should be more restrictive in case of nns-dapp?

frederikrothenberger: It is fine if only added to the `/.well-knonw` path.

Otherwise this needs to be either
* https://identity.ic0.app
* https://identity.internetcomputer.org
depending on which URL you are pointing to for authentication.

dfx.json
lmuntaner: Not, but needed to test the release candidate with prod-like functionality. It's enabled for `mainet`.

This is fixed in `main`, but I started this branch from `prod` tag.

dfx.json
lmuntaner: Yes, `staging` should be a copy of `mainnet`.

This is fixed in main, but I started this branch from prod tag.

frontend/scripts/build.csp.mjs
lmuntaner: No, but I should also add it.

The OWN_CANISTER_URL is that one.

frontend/src/lib/constants/environment.constants.ts
lmuntaner: yes, probably, good point

frontend/src/lib/rest/governance-metrics.rest.ts
lmuntaner: Isn't

```
await fetch(`${GOVERNANCE_CANISTER_RAW_URL}/metrics`)
```

A canister call?

frontend/src/lib/stores/auth.store.ts
lmuntaner: Do I need this for testnet? or only for mainnet?

dfx.json
mstrasinskis: Thanks for the explanation!

frontend/src/lib/stores/auth.store.ts
nmattia: Only for mainnet, I believe

frontend/src/lib/rest/governance-metrics.rest.ts
nmattia: no, that looks like a good old http fetch! although it's then turned into a canister call by the boundary nodes, but you shouldn't have to worry about that

frontend/jest-setup.ts
```diff
@@ -34,7 +34,7 @@ jest.mock("./src/lib/constants/canister-ids.constants.ts", () => ({
 
 jest.mock("./src/lib/constants/environment.constants.ts", () => ({
   DFX_NETWORK: "testnet",
-  HOST: "https://ic0.app",
+  HOST: "https://nns.internetcomputer.org",
```
nmattia, line 37: Might be wrong (no experience with the codebase) but I think you want `ic-api.io`; `HOST` seems to be the host used for api calls, not the host the assets are served from

frontend/scripts/build.csp.mjs
nmattia: I'm always a bit fuzzy on CSP issues, but do you really need II's URL in there? do you issue calls to II?

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -32,7 +34,7 @@ export const {
 export const IS_TESTNET: boolean =
   DFX_NETWORK !== "mainnet" &&
   FETCH_ROOT_KEY === true &&
-  !HOST.includes(".ic0.app");
+  !(HOST.includes(".ic-api.io") || HOST.includes(".ic0.app"));
```
nmattia, line 37:
```suggestion
  !(HOST.includes(".icp-api.io") || HOST.includes(".ic0.app"));
```

frontend/src/lib/stores/auth.store.ts
```diff
@@ -16,6 +22,22 @@ export interface AuthStore {
 // @see agent-js issue [#618](https://github.com/dfinity/agent-js/pull/618)
 let authClient: AuthClient | undefined | null;
 
+const getIdentityProvider = () => {
+  const environmentUrl = IDENTITY_SERVICE_URL;
```
nmattia, line 26: nit: why not use `IDENTITY_SERVICE_URL` directly?

frontend/scripts/build.csp.mjs
lmuntaner: I honestly haven't checked. That might be worth it's own review and maybe removal.

But in this PR I didn't want to add any other change. It was already part of the CSP; therefore, I kept it.

frontend/src/lib/stores/auth.store.ts
lmuntaner: Ok, then no. Thanks for the feedback. That's what I thought so as well.

lmuntaner: done, btw

frontend/src/lib/rest/governance-metrics.rest.ts
lmuntaner: This is a little bit confusing for me. So the old canisters will still live in `icp0.io`?

We also have the call of the SNS Aggregator, which is also an HTTP request. It uses the HOST `<id>.icp0.io`.

If SNS Aggregator doesn't work is worse than these metrics or the favicons.

I'll ask around.

frontend/jest-setup.ts
lmuntaner: yes, true. Although this is just for testing. But I'll change it.

frontend/src/lib/constants/environment.constants.ts
lmuntaner: thanks!

frontend/src/lib/stores/auth.store.ts
lmuntaner: I started with your code and refactored. This doesn't make sense anymore, I agree.

frontend/src/lib/rest/governance-metrics.rest.ts
lmuntaner: The SNS Aggregator is not yet used. Therefore, we'll ignore it for now.

I will revert these changes and use the old `ic0.app` for this call as suggested by Björn.

peterpeterparker (approved):
LGTM, thx

PR #1897, opened 2023-02-17
Use __default canister ID

bitdivine:
# Motivation
Remote canister IDs now support the `__default` network.  This removes the requirement to list every network unless it uses custom configuration.

# Changes
- Give remote canister IDs __default values.
  - Note: I have listed mainnet explicitly to make editors and reviewers aware of when they are changing production parameters.

# Tests
See CI: The `local` network takes the default values and CI passes.


lmuntaner (approved):
Thanks!

PR #1898, opened 2023-02-17
Chore: Bump snsdemo

bitdivine:
# Motivation
This allows us to use more recent versions of IC_COMMIT

# Changes
* Bump sns commit to a more recent version that supports more recent versions of IC_COMMIT.

# Tests
- See CI


lmuntaner (approved):
LGTM, will this make a PR to upgrade the aggregator types?

PR #1899, opened 2023-02-17
Allow Json expand/collapse with keyboard navigation

dskloetd:
# Motivation

I was seeing lint warnings about the Json element having a click event handler but not a key event handler.
Looking at the element I also noticed quite some code duplication.
A different element was rendered when the json is expanded than when it is collapsed.
This results in the element losing focus when expanded or collapsed, which means even if it has a key event handler, you can't press space/enter multiple times to collapse/expand multiple times.
Button elements automatically receive click events when space/enter is pressed so changing the element to a button solved the original problem.

# Changes

1. Changed the element to be a button so it receives click events on space/enter.
2. Removed code duplication so the button retains focus when rerendered.
3. Added a test to make sure the button remains focused when expanded/collapsed.
4. Lowered the number of allowed lint warnings.
The element still looks the same when rendered as a button:
image

# Tests

Manual testing + additional unit test.

### DIFF
--- a/frontend/src/lib/components/common/Json.svelte
+++ b/frontend/src/lib/components/common/Json.svelte
@@ -64,36 +64,24 @@
 </script>
 
 {#if isExpandable && hasChildren}
-  {#if collapsed}
-    <span
-      data-tid={testId}
-      class="key"
-      class:expanded={!collapsed}
-      class:collapsed
-      class:root
-      class:arrow={isExpandable && hasChildren}
-      role="button"
-      aria-label={$i18n.core.toggle}
-      tabindex="0"
-      on:click|stopPropagation={toggle}
-      >{keyLabel}
+  <button
+    data-tid={testId}
+    class="key"
+    class:expanded={!collapsed}
+    class:collapsed
+    class:root
+    class:arrow={isExpandable && hasChildren}
+    aria-label={$i18n.core.toggle}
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
Thanks!

frontend/src/lib/components/common/Json.svelte
```diff
@@ -64,36 +64,24 @@
...
+    data-tid={testId}
+    class="key"
+    class:expanded={!collapsed}
+    class:collapsed
+    class:root
+    class:arrow={isExpandable && hasChildren}
+    aria-label={$i18n.core.toggle}
+    tabindex="0"
+    on:click|stopPropagation={toggle}
+    >{keyLabel}
```
peterpeterparker, line 77: this seems incorrect, previous code displayed `keyLabel` only in certain case

peterpeterparker (commented):
can you provide a screenshot/cast of the change - i.e. open and closing children nodes still works out right?

peterpeterparker:
Looks good btw!

dskloetd:
Here you can see I can collapse/expand both with the mouse and the keyboard:
[screen-capture.webm](https://user-images.githubusercontent.com/122978264/219628810-9b810d2a-f5ee-4aa9-9b6d-eddcd59703c0.webm)

PR #1902, opened 2023-02-17
Draft: NN1-2088: retry failed sale notification in the backend

MarioDfinity:
# Motivation

The nns-dapp backend forgets about sale participation notification after sending it to the sale canister. If there is an error sending the message, the nns-dapp backend won't be able to notify the sale canister.

# Changes

Change the notification logic so that the nns-dapp backend retries notifications of sale participations that failed to be delivered.

# Tests

No test has been done at the moment.


lmuntaner (approved):
Thanks @MarioDfinity !!!

Smart solution!

There can't be an issue that it's never removed, right?

MarioDfinity:
> Thanks @MarioDfinity !!!
> 
> Smart solution!
> 
> There can't be an issue that it's never removed, right?

@lmuntaner if the sale canister is never reachable then the message could stay there forever. We could upgrade the nns-dapp in that case and change the code.
If we want to be safe I could add a (very high) number of retries after which the sale notification is forgotten. I don't know the backend enough to decide which one is the best option.

lmuntaner:
>If we want to be safe I could add a (very high) number of retries after which the sale notification is forgotten. I don't know the backend enough to decide which one is the best option.

How will you keep track of the retries?

I'm more familiar than you with the code, but not much more.

I'd say this might be a security risk. A malicious actor could force this "retry" to a canister that doesn't exist and keep our code running forever. We probably need a counter.

Maybe adding a new optional parameter to `MultiPartTransactionToBeProcessed::ParticipateSwap`?

MarioDfinity:
> > If we want to be safe I could add a (very high) number of retries after which the sale notification is forgotten. I don't know the backend enough to decide which one is the best option.
> 
> How will you keep track of the retries?
> 
> I'm more familiar than you with the code, but not much more.
> 
> I'd say this might be a security risk. A malicious actor could force this "retry" to a canister that doesn't exist and keep our code running forever. We probably need a counter.
> 
> Maybe adding a new optional parameter to `MultiPartTransactionToBeProcessed::ParticipateSwap`?

I would add two parameters to `ParticipateSwap`:

1. `number_of_retries`: the number of times this messages should be retried if sending it fails. This number is decreased every time the `ParticipateSwap` message is enqueued back.
2. `last_retry_timestamp_nanos`: the timestamp of the last retry. This will be used to check if enough time has passed since last retry. If yes then we retry once more, otherwise we just enqueue the message again.

We will use exponential backoff for retries to avoid bombarding the sale canister (/cc @michelabdalla-dfinity). I didn't think this through yet but it would be something like this: when a new `ParticipateSwap` `p` is enqueued, its `number_of_retries` is set to a constant `MAX_RETRIES` (e.g. `10`). Every time `p` is selected for a retry, we check if the message is ready to be retried, which is true if enough time since last retry has passed. The question is then how much time is "enough". We could say that this time is `t=2^(MAX_RETRIES-p.number_of_retries)` seconds so that the first retry has `p.number_of_retries=MAX_RETRIES` and `t=2^0=1` seconds, the second retry has `p.number_of_retries=MAX_RETRIES-1` and `t=2^1=2` seconds, the third retry has `p.number_of_retries=MAX_RETRIES-2` and `t=2^2=4` seconds...
A retry for `ParticipateSwap` `p` is performed if `p.last_retry_timestamp_nanos + t < current_timestamp_nanos`.

This shouldn't be too hard to add it to the backend. We just need to pass add two more parameters to `ParticipateSwap` and pass the `current_timestmap_nanos` in the function that enqueues and dequeues messages to notify.

lmuntaner:
>I would add two parameters to ParticipateSwap:

I like it!

>An open question is what to do if all notifications fail. Given that anybody can notify those messages, I'd say we just forget about the notifications in the nns-dapp backend and require manual intervention in those cases

I agree.

If you're busy, we can take it from here.

PR #1903, opened 2023-02-18
Bump ic commit 003

bitdivine:
# Motivation
The sns aggregator needs to handle newer SNSs.  The latest IC commit supported by the snsdemo is 0a5b8387 from 5 Feb.  So let's use that.

# Changes
- Run: `./scripts/update_ic_commit --ic_commit 0a5b8387179d2dec8269cc48b907ab0b334fdac1`
  - Note: This:
    - Updates the IC_COMMIT in dfx.json and so also updates the canister versions used in tests.
    - Updates the SNS .did files
    - Updates the rust code derived from the did files
- Update the patch file for dev features such as fast wait for quiet.
  - Note: The patch is applied to the IC codebase.  The patch is the same but had to be pointed at a different location, as the code had moved.
- Tweak the candid -> rust conversion that leads to a compiler error out of the box.
  - Note: the didc candid -> Rust conversion has trouble with empty enum variants.  It creates them in a valid way, but in a way that is incompatible with the CandidType macro. 

# Tests
CI should suffice


dfx.json
```diff
@@ -1031,7 +1031,7 @@
       "config": {
         "RUSTC_VERSION": "1.64.0",
         "IC_CDK_OPTIMIZER_VERSION": "0.3.1",
-        "IC_COMMIT": "3dba201865d970cfae3414ba78da18c888e2b111"
+        "IC_COMMIT": "0a5b8387179d2dec8269cc48b907ab0b334fdac1"
```
lmuntaner, line 1034: Could you try with `96325093210774910f0f702d92b282251168a534`?

This is the one I deployed to medium07 and medium06

It's from 17th of February. There are bug fixes, that's why it makes sense to use it. But you should see no candid changes.

lmuntaner (changes requested):
Thanks, just a comment to try a more updated IC_COMMIT.

dfx.json
bitdivine: I have tried with more recent IC commits but the deployment fails for them.  This is the most recent commit I can get to work reliably at the time of writing.  I hope to get more recent versions working soon but that will need many of the changes in here anyway.

lmuntaner: Understood thanks!

lmuntaner (approved):

PR #1904, opened 2023-02-20
build: remove e2e "proxy"

peterpeterparker:
# Motivation

`npm ci` in CI for `e2e-tests` is failing because it complains about missing dependencies which are inherited by `proxy` even though these are effectively installed there.

As we are not using the `proxy` anymore, this PR removes it and per extension solve the CI current issue.

# Changes

- `rm -r ./proxy && cd e2e-test && npm rm proxy`


scripts/fmt-e2e
```diff
@@ -2,6 +2,9 @@
 set -euxo pipefail
 cd "$(dirname "$(realpath "$0")")/.."
 
-cd e2e-tests
+cd proxy
+npm ci
```
bitdivine, line 6: Does the formatter also need to run on the proxy code?  (Also, out of curiosity, do you use the proxy?  I am under the impression that no-one uses the proxy but that it might be useful in future.)

bitdivine: Yes, it does look as if main is broken. :-(

peterpeterparker: It seems that the "proxy" dependency in "e2e-tests" has suddently becomes a resolution issue.
Since it is probably not use anymore we can just remove it.
Build in progress...

PR #1906, opened 2023-02-20
Enforce npm and nodejs version range

bitdivine:
# Motivation
Using an incompatible version of nodejs breaks builds, formatting and lockfiles.

# Changes
- Cause node to fail if an incompatible version of nodejs or npm is used.
  - Note: I have specified this individually in the .npmrc of each node project.  We may want to make a global entry, but must the n also remember to include it in docker builds that take just one directory and named other files.
- Specify the version of nodejs in CI.
  - Note: It would be nice to be able to read the node version from the package.json.  Ideally that should be done in both CI and in the dockerfile.  Should be possible, but in another PR.

# Tests

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -21,9 +21,9 @@
     "update:gix": "npm update @dfinity/gix-components",
     "upgrade:next": "npm rm @dfinity/nns @dfinity/cmc @dfinity/sns @dfinity/utils @dfinity/ledger @dfinity/ckbtc && npm i @dfinity/nns@next @dfinity/cmc@next @dfinity/sns@next @dfinity/utils@next @dfinity/ledger@next @dfinity/ckbtc@next"
   },
-  "engines" : {
-    "npm" : ">=8.0.0 <9.0.0",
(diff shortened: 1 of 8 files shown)


frontend/package.json
```diff
@@ -21,6 +21,10 @@
     "update:gix": "npm update @dfinity/gix-components",
     "upgrade:next": "npm rm @dfinity/nns @dfinity/cmc @dfinity/sns @dfinity/utils @dfinity/ledger @dfinity/ckbtc && npm i @dfinity/nns@next @dfinity/cmc@next @dfinity/sns@next @dfinity/utils@next @dfinity/ledger@next @dfinity/ckbtc@next"
   },
+  "engines" : {
+    "npm" : ">=8.0.0 <9.0.0",
+    "node" : ">=16.0.0 <17.0.0"
+  },
```
bitdivine, line 27: Note: The frontend `.npmrc` already sets `engine-strict=true` but doesn't specify an engine version in `package.json`.  Maybe it does so elsewhere.

peterpeterparker (approved):
LGTM

PR #1911, opened 2023-02-20
feat: ckBTC select network to "Send"

peterpeterparker:
# Motivation

- add a dropdown to "Select network" on "Send" modal
- display "Transaction fee" on `<TransactionForm />`
- few style adjustements
- rename ckBTC feature flag.

# Notes

- only above changes have been implemented, effective connection between "Send" - convert ckBTC to Bitcoin - will be implemented in another PR.

# Screenshots

Capture d’écran 2023-02-20 à 18 12 15
Capture d’écran 2023-02-20 à 18 12 21
Capture d’écran 2023-02-20 à 18 12 33
Capture d’écran 2023-02-20 à 18 12 58


frontend/src/tests/lib/components/accounts/SelectNetworkDropdown.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    ) as HTMLSelectElement | null;
+    selectElement &&
+      fireEvent.change(selectElement, {
+        target: { value: TransactionNetwork.ICP_CKBTC },
+      });
+
+    const div = container.querySelector(".container");
+
+    await waitFor(() =>
+      expect(div.classList).toContain(TransactionNetwork.ICP_CKBTC)
```
lmuntaner, line 65: Wouldn't it be better to expose the prop in `SelectNetworkDropdownTest` and test that than to rely on a class?

lmuntaner: It would also good to test that the value was another one before you trigger the `change` event.

frontend/src/tests/lib/modals/accounts/TransactionModal.spec.ts
```diff
@@ -304,6 +310,37 @@ describe("TransactionModal", () => {
...
+    });
+
+    it("should disable next button if network not selected", async () => {
+      const call = async () =>
+        await renderEnter10ICPAndNext({
+          rootCanisterId: OWN_CANISTER_ID,
+          mustSelectNetwork: true,
+        });
+
+      expect(call).rejects.toThrowError();
```
lmuntaner, line 340: Is this the same test as the title?

I assume that the previous test fails when `mustSelectNetwork` is set to `true`. But that doesn't necessarily mean that it's because the button is disabled.

lmuntaner (approved):
Looks good!

Just a few comments in the tests. But already approving.

frontend/src/tests/lib/components/accounts/SelectNetworkDropdown.spec.ts
peterpeterparker: > Wouldn't it be better to expose the prop in SelectNetworkDropdownTest and test that than to rely on a class?

How would you do that? i.e. how would read the property within the test?

> It would also good to test that the value was another one before you trigger the change event.

Not really needed I would say because there are no other changes in the test and that the default value is also covered by one, so the test here can only change from default to expected value. That said on strong opinion so added an addition except.

frontend/src/tests/lib/modals/accounts/TransactionModal.spec.ts
peterpeterparker: Here it set every single fields (amount, destination etc) so the button should be enable except if select network is not required.

frontend/src/tests/lib/components/accounts/SelectNetworkDropdown.spec.ts
lmuntaner: >How would you do that? i.e. how would read the property within the test?

```
// How to check the value of a prop in a Svelte component
// https://github.com/testing-library/svelte-testing-library/issues/117
component.$$.ctx[component.$$.props["disableInfiniteScroll"]]
```

This is in NnsProposals.spec

frontend/src/tests/lib/modals/accounts/TransactionModal.spec.ts
lmuntaner: There might be other reasons why the test fails, though. It could be a wrong setup.

If we introduce an issue with `mustSelectNetwork` where the transaction fails, we would not catch it in our test.

peterpeterparker: The `mustSelectNetwork` is not yet use for any transaction purpose in this PR. This will be implemented in another PR.

PR #1912, opened 2023-02-20
build: bump gix-cmp to fix Firefox collapsible height

peterpeterparker:
# Motivation

Build gix-cmp to fix Firefox collapsible height reported and fixed in https://github.com/dfinity/gix-components/pull/158.


mstrasinskis (approved):
Thanks!

PR #1913, opened 2023-02-20
Remove it.only from SnsNeurons.spec.ts

dskloetd:
# Motivation

Having `it.only` merged was most likely a mistake.
The other tests pass with `.only` removed.

# Changes

Replace `it.only` with just `it` in SnsNeurons.spec.ts.

# Tests

pass


lmuntaner (approved):
Oops, thanks!

PR #1916, opened 2023-02-21
Fix date formatting tests for Node 18.13.0

dskloetd:
# Motivation

With Node 18.13.0 there was an update to ICU 72 which means date formatting has a slightly different output (whitespace only). This broke some tests which have expectations on formatted dates.
I originally just updated the expected values, but this breaks the test when using a Node version < 18.13.0.
My current solution is to normalize the whitespace before comparing. This allows the test to pass both with older and newer Node versions.

# Changes

Normalize whitespace in tests where ICU 72 introduced a different whitespace separator in front of AM/PM.

# Tests

Tests only

### DIFF
--- a/frontend/src/lib/utils/format.utils.ts
+++ b/frontend/src/lib/utils/format.utils.ts
@@ -54,6 +54,3 @@ export const shortenWithMiddleEllipsis = (
     ? `${text.slice(0, splitLength)}...${text.slice(-1 * splitLength)}`
     : text;
 };
-
-export const normalizeWhitespace = (text: string): string =>
-  text.replace(/\s+/g, " ");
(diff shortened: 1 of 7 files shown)


frontend/src/lib/utils/format.utils.ts
```diff
@@ -54,3 +54,6 @@ export const shortenWithMiddleEllipsis = (
     ? `${text.slice(0, splitLength)}...${text.slice(-1 * splitLength)}`
     : text;
 };
+
+export const normalizeWhitespace = (text: string): string =>
```
peterpeterparker, line 58: 1. Can you move this utils in `utils.test-utils.ts` because we need it for test purpose only?

2. according its usage, the `text: string` should be `text: string | undefined`

peterpeterparker (approved):
Already approving, like the solution.

Added two comments about the utils, please have a look before merge.

Thx

frontend/src/lib/utils/format.utils.ts
dskloetd: Done

PR #1917, opened 2023-02-21
chore: fix webstorm resolution

peterpeterparker:
# Motivation

Webstorm does not resolve the path set by `.svelte-kit/tsconfig.json` and therefore we ends up with broken workspaces that cannot resolve dependencies and open imports.

This fix the issue providing the paths already in the `tsconfig.json` of the project.


lmuntaner (approved):
These are the same as in `.svelte-kit/tsconfig.json`?

peterpeterparker:
> These are the same as in .svelte-kit/tsconfig.json?

Yeah 😩

PR #1919, opened 2023-02-21
Clean up DisburseSnsNeuronModal.spec.ts

dskloetd:
# Motivation

My solution to GIX-1328 requires changes to this test (among others) otherwise the SNS neuron won't be available.
I decided to do a separate cleanup of the test to keep the PR that fixes GIX-1328 more readable.

# Changes

Changes to DisburseSnsNeuronModal.spec.ts:
1. Use real stores instead of mocking them.
2. Mock `sns-governance.api` instead of `sns-neurons.services`
3. Mock `getAuthenticatedIdentity` explicitly with a specific test identity, instead of relying on `frontend/jest-spy.ts`.
4. Populate `snsQueryStore` such that `snsTokenSymbolSelectedStore` (which is no longer mocked) can find the correct token metadata.
5. Add an additional convenience function to `sns-response.mock.ts` to create responses with specified principal. Reuse it for another function in the same file.

# Tests

Tests only.

### DIFF
--- a/frontend/src/tests/lib/modals/neurons/DisburseSnsNeuronModal.spec.ts
+++ b/frontend/src/tests/lib/modals/neurons/DisburseSnsNeuronModal.spec.ts
@@ -3,42 +3,37 @@
  */
 
 import * as accountsApi from "$lib/api/accounts.api";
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Nice, thanks you very much!

PR #1920, opened 2023-02-21
Clean up IncreaseSnsDissolveDelayModal.spec.ts

dskloetd:
# Motivation

My solution to GIX-1328 requires changes to this test (among others) otherwise the SNS neuron won't be available.
I decided to do a separate cleanup of the test to keep the PR that fixes GIX-1328 more readable.

# Changes

Changes to IncreaseSnsDissolveDelayModal.spec.ts:
1. Mock sns-governance.api instead of sns-neurons.services
2. Expect a specific value for the dissolve delay on the API call
3. Mock getAuthenticatedIdentity explicitly with a specific test identity, instead of relying on frontend/jest-spy.ts.
4. reset all mocks in beforeEach
5. Use `roundUpSecondsToWholeDays` instead of `daysToSeconds(secondsToDays(seconds))` because the latter sounds like it doesn't do anything.

# Tests

Tests only.

### DIFF
--- a/frontend/src/tests/lib/modals/sns/IncreaseSnsDissolveDelayModal.spec.ts
+++ b/frontend/src/tests/lib/modals/sns/IncreaseSnsDissolveDelayModal.spec.ts
@@ -2,9 +2,10 @@
  * @jest-environment jsdom
  */
 
+import * as snsGovernanceApi from "$lib/api/sns-governance.api";
 import { SECONDS_IN_YEAR } from "$lib/constants/constants";
 import IncreaseSnsDissolveDelayModal from "$lib/modals/sns/neurons/IncreaseSnsDissolveDelayModal.svelte";
-import { updateDelay } from "$lib/services/sns-neurons.services";
+import * as authServices from "$lib/services/auth.services";
 import { loadSnsParameters } from "$lib/services/sns-parameters.services";
 import { snsParametersStore } from "$lib/stores/sns-parameters.store";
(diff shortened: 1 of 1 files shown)


frontend/src/tests/lib/modals/sns/IncreaseSnsDissolveDelayModal.spec.ts
```diff
@@ -15,24 +16,23 @@ import { fromDefinedNullable } from "@dfinity/utils";
...
 
-jest.mock("$lib/services/sns-parameters.services", () => {
-  return {
-    loadSnsParameters: jest.fn().mockResolvedValue(undefined),
-  };
-});
+const testIdentity = createMockIdentity(10023);
+
+const roundUpSecondsToWholeDays = (seconds: number): number =>
+  daysToSeconds(secondsToDays(seconds));
```
mstrasinskis, line 35: Good point! Would be great to rename it not only in the test.

mstrasinskis (approved):
LGTM, thanks!

frontend/src/tests/lib/modals/sns/IncreaseSnsDissolveDelayModal.spec.ts
dskloetd: I first added it to the util but then I moved it back here because I didn't see any other instances of this combination of these functions.

PR #1923, opened 2023-02-21
Refactor dockerfile

bitdivine:
# Motivation
The dockerfile is hard to use for building images for testnets.  In particular:

- It would be useful if `./build_frontend` did not need to run `config.sh` which in turn needs access to the environment such as `.dfx` which contains a huge amount of data not relevant to the build.
- It would be nice if it were easier to switch tool versions.
- It would be useful if the build were made of smaller, more granular steps.
- At the same time, default builds for mainnet and local should work easily, as now.

These factors influence each other strongly so are a bit hard to separate into separate issues.

# Changes
To handle changes in the network configuration better:
- Create a small docker stage that pulls in .dfx and canister_ids.json (if they exist) and run config.sh with this data.
  - Note: This means that it pulls in lots of irrelevant data but as the output is unaffected by this data, downstream stages are unaffected by this random junk.  If unrelated data changes, the config stage will have to be rerun but it is small and fast, the outputr is the same as before and downstream steps all hit the cache.
- remove config.sh from other locations.

A similar pattern applies to changes in the build configuration:
- Create a base image that has just basic tools such as jq.
- Get the tool versions in a small dedicated stage
- In the existing builder stage, get versions from the tool-versioning stage.

# Tests
- Building the prod nns-dapp and aggregator wasms yields the same wasms as main.


bitdivine:
Done in other PRs

PR #1925, opened 2023-02-21
GIX-1328 Small test cleanups

dskloetd:
# Motivation

My solution to GIX-1328 requires changes to these tests (among others) otherwise the SNS neuron won't be available.
I decided to do a separate cleanup of the tests to keep the PR that fixes GIX-1328 more readable.

# Changes

1. beforeEach instead of afterEach.
2. Use real snsAccountsStore instead of mock.

# Tests

Tests only

### DIFF
--- a/frontend/src/tests/lib/pages/SnsAccounts.spec.ts
+++ b/frontend/src/tests/lib/pages/SnsAccounts.spec.ts
@@ -14,17 +14,13 @@ import type { Subscriber } from "svelte/store";
 import { mockPrincipal } from "../../mocks/auth.store.mock";
 import { mockStoreSubscribe } from "../../mocks/commont.mock";
 import en from "../../mocks/i18n.mock";
-import { mockSnsAccountsStoreSubscribe } from "../../mocks/sns-accounts.mock";
+import { mockSnsMainAccount } from "../../mocks/sns-accounts.mock";
 import {
   mockProjectSubscribe,
   mockSnsFullProject,
 } from "../../mocks/sns-projects.mock";
 
-jest.mock("$lib/services/sns-accounts.services", () => {
-  return {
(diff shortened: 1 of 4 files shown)


frontend/src/tests/lib/pages/SnsAccounts.spec.ts
```diff
@@ -14,28 +14,26 @@ import type { Subscriber } from "svelte/store";
...
     // Do nothing
   };
 
   describe("when there are accounts in the store", () => {
-    beforeAll(() => {
-      jest
-        .spyOn(snsAccountsStore, "subscribe")
-        .mockImplementation(mockSnsAccountsStoreSubscribe(mockPrincipal));
+    beforeEach(() => {
+      snsAccountsStore.setAccounts({
```
lmuntaner, line 32: Shouldn't you reset it before as well?

lmuntaner (changes requested):
One small change requested

frontend/src/tests/lib/pages/SnsAccounts.spec.ts
dskloetd: Good point!.
I assumed that setAccounts sets the entire state so reset would be redundant.
But looking at the code, it seems like setAccounts actually sets the accounts for one root canister and doesn't change the rest of the state.
I wonder if it could have a better name or I just need more domain knowledge.
https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/stores/sns-accounts.store.ts#L53

lmuntaner (approved):

frontend/src/tests/lib/pages/SnsAccounts.spec.ts
dskloetd: It could be setAccountsForProject.
But I'll just keep it in the back of my head for now.

PR #1926, opened 2023-02-21
GIX-1328 Only select selectable universes.

dskloetd:
# Motivation

It's currently possible to see uncommitted sns neurons by deep linking the root canister id.
The issue seems to be that while `selectedUniverseStore` only gives a universe from `selectableUniversesStore`, `selectedUniverseIdStore` doesn't care whether the universe is in `selectableUniversesStore`.
The solution is to make both respect selectableUniversesStore by having one derive from the other.
As a side effect we no longer need to check if the selected universe is the ckBTC universe because selectableUniversesStore already does this check.

# Changes

frontend/src/lib/derived/selected-universe.derived.ts: Make sure selectedUniverseIdStore only holds a universe id from selectableUniversesStore.

Many tests had to change because they didn't set up the selectableUniversesStore to allow the selected universe they need for the test.

# Tests

1. Set up selectableUniversesStore correctly via snsQueryStore in many tests.
2. In selected-universe.derived.spec.ts, explicitly test that a universe has to be selectable to be selected.
3. Made sure the new test fails without the fix from this PR.

### DIFF
--- a/frontend/src/lib/derived/selected-universe.derived.ts
+++ b/frontend/src/lib/derived/selected-universe.derived.ts
@@ -2,18 +2,13 @@ import {
   OWN_CANISTER_ID,
   OWN_CANISTER_ID_TEXT,
 } from "$lib/constants/canister-ids.constants";
-import { pageStore, type Page } from "$lib/derived/page.derived";
+import { pageStore } from "$lib/derived/page.derived";
 import {
   NNS_UNIVERSE,
   selectableUniversesStore,
 } from "$lib/derived/selectable-universes.derived";
-import { ENABLE_CKBTC_LEDGER } from "$lib/stores/feature-flags.store";
 import type { Universe } from "$lib/types/universe";
-import {
-  isUniverseCkBTC,
-  isUniverseNns,
-  pathSupportsCkBTC,
-} from "$lib/utils/universe.utils";
+import { isUniverseCkBTC, isUniverseNns } from "$lib/utils/universe.utils";
 import { Principal } from "@dfinity/principal";
 import { derived, type Readable } from "svelte/store";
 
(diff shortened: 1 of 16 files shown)


frontend/src/lib/derived/selected-universe.derived.ts
```diff
@@ -39,19 +34,19 @@ const pageUniverseIdStore: Readable<Principal> = derived(
...
+
 export const selectedUniverseIdStore: Readable<Principal> = derived<
-  [Readable<Principal>, Readable<Page>, Readable<boolean>],
+  Readable<Universe>,
   Principal
->(
-  [pageUniverseIdStore, pageStore, ENABLE_CKBTC_LEDGER],
-  ([canisterId, page, enableCkbtcLedger]) => {
-    // ckBTC is only available on Accounts therefore we fallback to Nns if selected and user switch to another view
-    if (enableCkbtcLedger && pathSupportsCkBTC(page)) {
```
lmuntaner, line 49: Do you cover the case where the path doesn't support ckbtc, but the `u` is set to ckBTC universe?

What should we do in that case?

This might happen if you are in Accounts, click on ckBTC, and then move to Neurons. The `u` selected might still be `ckBTC`. Therefore, neurons won't be found or an error will be raised.

frontend/src/tests/lib/derived/selected-universe.derived.spec.ts
```diff
@@ -99,5 +116,19 @@ describe("selected universe derived stores", () => {
       const $store2 = get(selectedUniverseIdStore);
       expect($store2.toText()).toEqual(OWN_CANISTER_ID.toText());
     });
+
+    it("returns OWN_CANISTER_ID if context is not a selectable universe", () => {
```
lmuntaner, line 120: Should we also add a test regarding the edge case of the path not supporting `ckBTC` but the `u` being ckBTC?

lmuntaner (commented):
Looking good, but there is an edge case that I'm not sure how it behaves and how we should treat it.

Let me know whether you encountered it and what happened.

frontend/src/tests/lib/derived/selected-universe.derived.spec.ts
dskloetd: Since that's existing functionality that I didn't change (just relied on a different mechanism to get the same functionality), I assumed it would be covered by the existing ckBTC tests in this file. But on a close look it seems like it wasn't covered. So I've added the missing tests. Thanks!

peterpeterparker (changes requested):
Tested it locally. This PR does not list ckBTC anymore.

Capture d’écran 2023-02-22 à 10 38 32
isn't ok.

peterpeterparker:
(double checking...)

peterpeterparker:
Sounds good. ckBTC is listed including correct balance.
Capture d’écran 2023-02-22 à 10 53 53
Capture d’écran 2023-02-22 à 10 53 55

peterpeterparker (approved):
LGTM according my test, thanks.

dskloetd:
> LGTM according my test, thanks.

Thanks for verifying!

frontend/src/tests/lib/derived/selected-universe.derived.spec.ts
lmuntaner: thanks!

frontend/src/lib/derived/selected-universe.derived.ts
lmuntaner: Ah!! Thanks, I forgot about that. Great, thanks for the explanation!

PR #1928, opened 2023-02-22
Fix SnsAccounts.spec.ts

dskloetd:
# Motivation

`SnsAccounts.spec.ts` has a test which tests that no amount is rendered when there are no accounts. 
But it makes no difference to the test whether there are accounts or not so the test doesn't actually test anything.

# Changes

1. Added an equivalent test which tests that an amount *is* rendered when there are accounts, to make sure the 2 situations actually differ.
2. Added a helper function to make sure the component finished loading.
3. Added a helper function to check whether an amount is rendered.
4. Used the helper functions where appropriate.

# Tests

Test only


lmuntaner (approved):
Thanks!

PR #1930, opened 2023-02-22
Fail tests if there are console.log/debug/error

dskloetd:
# Motivation

Getting log output while running `npm run test` is quite distracting and usually not intended.
If the logging is expected, the test should really mock it out and expect the behavior which means the logs don't appear in the output.

# Changes

1. Fail a test if console.log/debug/error happens during the test.
2. Only fail after the test finishes to allow seeing any other results from the test.
3. For debugging allow disabling the functionality by calling `allowLoggingInOneTestForDebugging()` from the test or beforeEach().
4. Edit existing tests that silence the console in a way that's broken by this PR to keep them silencing the console.

# Tests

Manually checked that adding console.log to a test caused it to fail and that disabling worked.


frontend/src/tests/mocks/console.mock.ts
```diff
@@ -0,0 +1,42 @@
...
+          "ForDebugging() from src/tests/mocks/console.mock in your test."
+      ).toBe(false);
+    }
+    isLoggingAllowed = false;
+  });
+};
+
+// Use this only when debugging but don't commit.
+export const allowLoggingInOneTestForDebugging = () => {
+  isLoggingAllowed = true;
```
bitdivine, line 41: Do not tests run concurrently?  Using a single global variable for all tests, as this seems to do, doesn't seem to be sound in that case.

dskloetd: If they run concurrently, they must all have their own module environment because lots of modules have global variables. Not just in test code but especially in production code.

mstrasinskis (approved):
LGTM, thanks!

PR #1932, opened 2023-02-22
Separate getting tool versions from tool installation

bitdivine:
# Broad context
I have rewritten the Dockerfile to address issues encountered recently, especially in the context of using docker builds on nns-dapp for testnets.  Docker builds play an important role for non-team members who want something that "just works" without too many knobs to configure.  That rewrite is broken down into small chunks.

# Motivation
There has been discussion about how we specify tool versions.  However changing _where_ we specify a tool version causes the tool installation to have to be rebuilt.

# Changes
- Separate getting tool versions from installing tools.

# Tests
- See CI
- I have verified that the result of `./scripts/docker-build` (with no further arguments or relevant env vars) is unaffected by this change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone &
 # Note: This can be done in the builder but is slow because unrelated changes to dfx.json can cause a rebuild.
 #
 # Note: Here we play a bit with the idea of storing config in files, one file per parameter.
-from base as tool_versions
+FROM base as tool_versions
 SHELL ["bash", "-c"]
 RUN mkdir -p config
 COPY dfx.json dfx.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,43 +5,53 @@
 # docker cp $container_id:nns-dapp.wasm nns-dapp.wasm
 # docker rm --volumes $container_id
 
-# This is the "builder", i.e. the base image used later to build the final
-# code.
-FROM ubuntu:20.04 as builder
+# Operating system with basic tools
+FROM ubuntu:20.04 as base
 SHELL ["bash", "-c"]
-
-ARG rust_version=1.64.0
-ENV NODE_VERSION=16.17.1
-
 ENV TZ=UTC
-
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone && \
     apt -yq update && \
     apt -yqq install --no-install-recommends curl ca-certificates \
         build-essential pkg-config libssl-dev llvm-dev liblmdb-dev clang cmake \
         git jq npm
 
+# Gets tool versions.
+#
+# Note: This can be done in the builder but is slow because unrelated changes to dfx.json can cause a rebuild.
+#
+# Note: Here we play a bit with the idea of storing config in files, one file per parameter.
+from base as tool_versions
+SHELL ["bash", "-c"]
+RUN mkdir -p config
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,43 +5,53 @@
...
     apt -yqq install --no-install-recommends curl ca-certificates \
         build-essential pkg-config libssl-dev llvm-dev liblmdb-dev clang cmake \
         git jq npm
 
+# Gets tool versions.
+#
+# Note: This can be done in the builder but is slow because unrelated changes to dfx.json can cause a rebuild.
+#
+# Note: Here we play a bit with the idea of storing config in files, one file per parameter.
+from base as tool_versions
```
dskloetd, line 23: uppercase FROM for consistency?

Dockerfile
```diff
@@ -5,43 +5,53 @@
...
+SHELL ["bash", "-c"]
+RUN mkdir -p config
+COPY dfx.json dfx.json
+ARG rust_version=1.64.0
+ENV NODE_VERSION=16.17.1
```
dskloetd, line 28: Why get the rust_version and NODE_VERSION by different mechanisms? (ARG vs. ENV)

Dockerfile
```diff
@@ -5,43 +5,53 @@
...
+COPY --from=tool_versions /config/*_version config/
 # Install node
 RUN npm install -g n
-RUN n "${NODE_VERSION}"
+RUN n "$(cat config/node_version)"
 RUN node --version
 RUN npm --version
-
 # Install Rust and Cargo in /opt
 ENV RUSTUP_HOME=/opt/rustup \
```
dskloetd, line 45: Would it also help to install each tool in a different stage? Then we don't need to install node again if only the rust version changed.
If we expect some tools to change more often than others, that would also affect the order that's most efficient.

Dockerfile
```diff
@@ -5,43 +5,53 @@
...
     CARGO_HOME=/opt/cargo \
     PATH=/opt/cargo/bin:$PATH
-
 RUN curl --fail https://sh.rustup.rs -sSf \
-        | sh -s -- -y --default-toolchain ${rust_version}-x86_64-unknown-linux-gnu --no-modify-path && \
-    rustup default ${rust_version}-x86_64-unknown-linux-gnu && \
+        | sh -s -- -y --default-toolchain "$(cat config/rust_version)-x86_64-unknown-linux-gnu" --no-modify-path && \
+    rustup default "$(cat config/rust_version)-x86_64-unknown-linux-gnu" && \
```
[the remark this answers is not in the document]
dskloetd: Why wouldn't it need to be specified?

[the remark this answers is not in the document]
bitdivine: The version is specified in the rust toolchain file.  As long as any rust version is installed, as long as it is recent enough to support toolchain files, rust will automatically switch to the right version and target.

Dockerfile
bitdivine: I have not changed the mechanism, as that is not the focus of this PR.

Dockerfile
bitdivine: We change tools sufficiently rarely that I decided against that.  The frequency is already accounted for in the order.

Dockerfile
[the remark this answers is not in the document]
bitdivine: This, by the way, is what I was expecting node and dfx would support as well.  dfx tries but flakily, node doesn't seem to have such a mechanism at all.

Dockerfile
bitdivine: Good point.

dskloetd (approved):

PR #1933, opened 2023-02-22
Reduce inputs to build steps

bitdivine:
# Motivation
The build steps for the nns-dapp  frontend and backend currently run ./config.sh.  For config.sh to work properly for environments other than hard-wired entries such as mainnet, it needs access to the environment such as `.dfx/`.  The environment contains a huge amount of information that is not relevant and that makes it harder to reason about the build and breaks caches.  If we put this "config extraction" into a separate step, the inputs to the build can stay small.

# Changes
- Add a separate docker stage to get build configuration
- Move ./config.sh out of the frontend build command.
- Minimize inputs to the build commands, e.g. removing env vars.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,25 +58,37 @@ RUN mkdir -p rs/backend/src rs/sns_aggregator/src && touch rs/backend/src/lib.rs
 COPY dfx.json dfx.json
 RUN DFX_VERSION="$(jq -cr .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
 
-FROM builder AS build_frontend
+# Title: Gets the deployment configuration
+# Args: Everything in the environment.  Ideally also ~/.config/dfx but that is inaccessible.
+FROM builder AS configurator
+SHELL ["bash", "-c"]
+COPY dfx.json config.sh .df[x] canister_ids.jso[n] /build/
+WORKDIR /build
 ARG DFX_NETWORK=mainnet
-RUN echo "DFX_NETWORK: '$DFX_NETWORK'"
+RUN mkdir -p frontend
+RUN ./config.sh
+
+# Title: Image to build the nns-dapp frontend.
+# Args: A file with env vars at frontend/.env created by config.sh
+FROM builder AS build_frontend
 SHELL ["bash", "-c"]
 COPY ./frontend /build/frontend
-COPY ./config.sh /build/
+COPY --from=configurator /build/frontend/.env /build/frontend/.env
 COPY ./build-frontend.sh /build/
-COPY ./dfx.json /build/
 COPY ./scripts/require-dfx-network.sh /build/scripts/
 WORKDIR /build
 RUN ( cd frontend && npm ci )
-RUN export DFX_NETWORK && . config.sh && ./build-frontend.sh
+RUN ./build-frontend.sh
 
+# Title: Image to build the nns-dapp backend.
+# Args: DFX_NETWORK env var for enabling/disabling features.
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -39,7 +39,7 @@ TOPLEVEL="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # Or do we need the exports for the backend as well?
 
 # Assemble the configuration
-. config.sh
```
bitdivine, line 42: The environment variables are not needed; the only output of the config file used now is the `frontend/.env`.  That is the only place that needs fancy config.  Unless some use case shows up in some other context, the env var exports in `config.sh` can be deleted.

build-frontend.sh
```diff
@@ -2,13 +2,12 @@
 set -euo pipefail
 TOPLEVEL="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
bitdivine, line 4: Running config.sh here requires access to the environment, which we don't want.  We reduce the required inputs to just the frontend/.env config file.

Dockerfile
```diff
@@ -58,34 +58,51 @@ RUN mkdir -p rs/backend/src rs/sns_aggregator/src && touch rs/backend/src/lib.rs
...
+RUN mkdir -p frontend
+RUN ./config.sh
+
+# Title: Image to build the nns-dapp frontend.
+# Args: A file with env vars at frontend/.env created by config.sh
+FROM builder AS build_frontend
 SHELL ["bash", "-c"]
 COPY ./frontend /build/frontend
-COPY ./config.sh /build/
+COPY --from=configurator /build/frontend/.env /build/frontend/.env
```
bitdivine, line 76: Note: If there was an irrelevant change in say `.dfx/local/canister_ids.json` then the configurator stage will rerun, but it is small and fast, the output of the configurator will be unchanged, so the input to the frontend-build step will be unchanged, so the frontend build will hit the cache, as will the backend build that follows it.

Dockerfile
```diff
@@ -58,34 +58,51 @@ RUN mkdir -p rs/backend/src rs/sns_aggregator/src && touch rs/backend/src/lib.rs
 COPY dfx.json dfx.json
 RUN DFX_VERSION="$(jq -cr .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
 
-FROM builder AS build_frontend
+# Title: Gets the deployment configuration
+# Args: Everything in the environment.  Ideally also ~/.config/dfx but that is inaccessible.
+FROM builder AS configurator
+SHELL ["bash", "-c"]
+COPY dfx.json config.sh .df[x] canister_ids.jso[n] /build/
```
dskloetd, line 65: Why are [x] and [n] between square brackets?

dskloetd, line 65: As far as I can tell it didn't previously copy canister_ids.json. How was ./config.sh able to run without it?

dskloetd (approved):

Dockerfile
bitdivine: This makes their words patterns rather than paths.  `.dfx` and `canister_ids.json` are not guaranteed to be present.  If I gave the path, absence would be an error.  If a pattern matches nothing, COPY is happy, as long as it has something else to copy.

bitdivine: `canister_ids.json` is only interesting when the network is a testnet. That scenario was not supported before.

dskloetd: Why do you want to copy them if they are not necessary?
If they are necessary only in certain cases, can you copy them only in those cases?
At the very least clarify it with a comment, because it looks rather confusing.

PR #1934, opened 2023-02-22
build: upgrade sveltekit

peterpeterparker:
# Motivation

Upgrade SvelteKit to last version - and other dev dependencies - and adapt CSP for breaking changes.

# Changes

- upgrade dev dependencies in SvelteKit
- CSP script
   - update regex to identify SvelteKit script (using attributes)
   - manipulate extracted script
      - inject sveltekit variable in `window`
      - attach script to `body`
   - remove `<link rel="preload" />` and inject these using script as we do with the automatic script
   
# Notes

No changes in the policy - i.e. `<script />` tags in `.html` are added as `'sha256-...'` to CSP, `strict-dynamic` is still use.

New is the fact that we also inject `<link preload />` with the help of script added to the html file too.


lmuntaner (approved):
I like it! Thanks!

lmuntaner:
@peterpeterparker we can update and merge this branch now that the RC is set

PR #1935, opened 2023-02-22
Add key event handler on theme toggle

dskloetd:
# Motivation

This is the last remaining lint warning.
For a11y, lint wants every element with a click handler to also have a key handler.
Buttons etc. do this automatically but div/span/etc. not.
We have a div wrapping the theme toggle on the logged-out menu on mobile, which stops propagation on click to keep the menu open to allow toggling multiple times without having to reopen the menu each time.
It's currently not possible to use the toggle with the keyboard because it's not in the tab order (and only appears on mobile) but if it were possible, we'd probably want to keep the menu op in that case as well.
So for now, adding this handler just stops lint from complaining.

# Changes

1. Stop propagation of keyboard events from the theme toggle inside the logged-out menu.
2. Limit the allowed number of lint warnings to 0.

# Tests

Manually use the theme toggle.
Ran `npm run lint` and checked that there are no warnings.

### DIFF
--- a/frontend/src/lib/components/login/LoginMenuItems.svelte
+++ b/frontend/src/lib/components/login/LoginMenuItems.svelte
@@ -26,7 +26,7 @@
   </MenuItem>
 </div>
 
-<div class="theme" on:click|stopPropagation>
+<div class="theme" on:click|stopPropagation on:keypress|stopPropagation>
   <span>{$i18n.theme.switch_theme}</span><ThemeToggle />
 </div>
 
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -13,7 +13,7 @@
     "preview": "vite preview",
     "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && npm run lint",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
-    "lint": "prettier --plugin-search-dir . --check . && eslint --max-warnings 1 .",
+    "lint": "prettier --plugin-search-dir . --check . && eslint --max-warnings 0 .",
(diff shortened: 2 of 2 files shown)


frontend/src/lib/components/login/LoginMenuItems.svelte
```diff
@@ -26,7 +26,7 @@
   </MenuItem>
 </div>
 
-<div class="theme" on:click|stopPropagation>
+<div class="theme" on:click|stopPropagation on:keypress|stopPropagation>
```
peterpeterparker, line 29: on any key? shouldn't it be scoped to "Enter" or "Space" only?

dskloetd: Is that possible here in the template or do you mean in an actually handler function?
Given that the element currently can't receive focus, is it worth it? I wouldn't be able to test it either.

dskloetd: The reason for the existing stopPropagation on the click handler is that otherwise the menu is closed here:
https://github.com/dfinity/gix-components/blob/main/src/lib/components/Menu.svelte#L16

As you can see the menu currently doesn't have a key handler.
So there is technically currently no need to stop propagation on key events, nor is there a need to propagate any key events.
I guess it would be nicest if we put a key handler on the Menu to close on Escape and then make sure we only stop propatation on Enter/Space here but I'm not sure it's worth it in the current state.

peterpeterparker: Thanks for the details, all good

peterpeterparker (approved):
LGTM, thx

PR #1936, opened 2023-02-22
GIX-1318: Remove reloadSnsState from participation flow

lmuntaner:
# Motivation

It doesn't add any value to reload the state to validate whether the user can or not participate.

This same validation is done when creating the ticket.

# Changes

* Remove call to `reloadSnsState` in the participation flow.
* Remove helper `reloadSnsState` because it's not used anymore.

# Tests

* Tests didn't check that the call was made. Therefore, no changes in the tests were necessary.


mstrasinskis (approved):
LGTM, thanks! 
P.s. Sorry for the possible merge conflicts.

PR #1937, opened 2023-02-22
npm install --save-dev @types/jest-expect-message

dskloetd:
# Motivation

Package jest-expect-message allows passing a message to expect calls.
We were sometimes getting errors such as `Expected 1 arguments, but got 2.` which surprisingly went away after running `npm run test -- --clearCache` but would reappear after some time.
After installing the types package for jest-expect-message, the errors no longer seem to appear.
The [package](https://www.npmjs.com/package/@types/jest-expect-message) is supposedly deprecated because it shouldn't be necessary but it does appear to be necessary for us.

# Changes

Ran `npm install --save-dev @types/jest-expect-message`

# Tests

I found out that after running `npm run test -- --clearCache`, the error would reappear after running `npm run test -- feature-flags.store.spec.ts` twice while making changes to the test in between.
After installing the additional package, I was no longer able to reproduce the error in this way.

I tried several other things mentioned on https://www.npmjs.com/package/jest-expect-message but none of the other things made the errors go away.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -32,6 +32,7 @@
         "@testing-library/jest-dom": "^5.16.5",
         "@testing-library/svelte": "^3.2.2",
         "@types/jest": "^29.4.0",
+        "@types/jest-expect-message": "^1.1.0",
         "@types/marked": "^4.0.8",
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Thanks!

PR #1942, opened 2023-02-23
Remove jest-expect-message

dskloetd:
# Motivation

It's causing issues where it fails with `Expected 1 arguments, but got 2.` unless you clear the jest cache.

# Changes

1. Remove package jest-expect-message.
2. Changed the 1 test the used it.

The test that was using it did so because it has an expect inside a for loop. The message was supposed to make it clear which iteration of the loop failed. Instead I now included the iteration in the expected value to achieve the same.

# Tests

Ran `npm run test`.


mstrasinskis (approved):
Thanks you!

PR #1944, opened 2023-02-23
fix: derivation origin

peterpeterparker:
# Motivation

Derivation origin should not be provided when user sign-in form `nns.ic0.app` otherwise II displays "nns.ic0.app is an alternative origin of nns.ic0.app"

# Changes

- provide derivation origin when required

# Screenshot

image


frontend/__mocks__/$app/environment.ts
```diff
@@ -1 +1,2 @@
 export const browser = false;
+export const building = false;
```
nmattia, line 2: not sure about this

nmattia (approved):
half lgtm half rubber stamp!

frontend/src/lib/utils/env.utils.ts
```diff
@@ -0,0 +1,9 @@
+import { NNS_IC_ORG_ALTERNATIVE_ORIGIN } from "$lib/constants/origin.constants";
+
+export const isNnsAlternativeOrigin = (): boolean => {
+  const {
+    location: { origin },
+  } = window;
+
+  return origin.includes(NNS_IC_ORG_ALTERNATIVE_ORIGIN);
```
nmattia, line 8: nit: this could even be `===`

frontend/__mocks__/$app/environment.ts
peterpeterparker: added it because originally I added a guard in the `isNnsAlternativeOrigin` function

```
if (building) return false
```

Ultimately removed the guard but did not remove the mock because it's not incorrect to have it there because otherwise the value is `undefined` when we run the test as the all SvelteKit environment constants are mocked.

frontend/src/lib/utils/env.utils.ts
peterpeterparker: I thought it was safer in case there would be a trailing slash but you are right, all browser provide origin without. Modified for `===`, thanks

PR #1948, opened 2023-02-23
Remove it.only from VotingCard.spec.ts

dskloetd:
# Motivation

6 tests in VotingCard.spec.ts were disabled by an accidental `it.only`.
3 of them actually don't pass but it's still better to have 4 enabled than only 1.

# Changes

Remove `it.only` and add `it.skip` to the tests that currently don't pass.

# Tests

Ran the test.


mstrasinskis (approved):
Thank you!

PR #1949, opened 2023-02-24
Feat: Retry new sale ticket

lmuntaner:
# Motivation

We want to retry and don't show error if creating the new sale ticket fails for unknown reasons. For example, replica is overloaded.

# Changes

* New helper `pollNewSaleTicket` in sns sale services.
* Use new helper `pollNewSaleTicket` instead of calling api function directly in sns service `newSaleTicket`

# Tests

* Change test that handles unknown errors and add a new one.

### DIFF
--- a/frontend/src/tests/lib/services/sns-sale.services.spec.ts
+++ b/frontend/src/tests/lib/services/sns-sale.services.spec.ts
@@ -434,6 +434,11 @@ describe("sns-api", () => {
   });
 
   describe("newSaleTicket", () => {
+    beforeEach(() => {
+      jest.clearAllTimers();
+      const now = Date.now();
+      jest.useFakeTimers().setSystemTime(now);
+    });
     it("should call newSaleTicket api", async () => {
       await newSaleTicket({
         rootCanisterId: testSnsTicket.rootCanisterId,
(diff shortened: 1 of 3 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/221108095-c012582f-9f39-43a6-b967-ecfe213084e7.mov

lmuntaner:
@mstrasinskis please review

frontend/src/tests/lib/services/sns-sale.services.spec.ts
```diff
@@ -571,25 +576,62 @@ describe("sns-api", () => {
...
+      }
+      expect(counter).toBe(retriesBeforeSuccess);
+      expect(spyOnNewSaleTicketApi).toBeCalledTimes(counter);
+    });
+
+    it("should retry unknown errors until known error", async () => {
+      spyOnNewSaleTicketApi
+        .mockRejectedValueOnce(new Error("connection error"))
+        .mockRejectedValueOnce(new Error("connection error"))
+        .mockRejectedValueOnce(new Error("connection error"))
```
mstrasinskis, line 610: Not for this pr, but interesting that if I add a couple `mockRejectedValueOnce` more, the test is still successful.

mstrasinskis (approved):
LGTM, thank you!

frontend/src/tests/lib/services/sns-sale.services.spec.ts
lmuntaner: No, the amount that I add is related to the `retriesBeforeSuccess`. I'll add a comment.

lmuntaner: Done!

PR #1950, opened 2023-02-24
feat: bitcoin network estimated fee

peterpeterparker:
# Motivation

Display the estimated Bitcoin transaction fee if user select Bitcoin network to convert ckBTC -> Bitcoin.

# Changes

- bump ic-js with new ckBTC minter API endpoint `estimated_fee`
- implement API and services that call the new endpoint `estimated_fee`
- add a new cmp `BitcoinEstimatedFee` that fetches, or not, the estimated fee according network and display, or not, the fee
- add a new cmp `BitcoinEstimatedFeeDisplay` to present the fee, if present, on the review step of the transaction modal

# Screenshots

Capture d’écran 2023-02-27 à 09 15 10
Capture d’écran 2023-02-27 à 09 15 14
Capture d’écran 2023-02-27 à 09 15 38
Capture d’écran 2023-02-27 à 09 15 41


frontend/src/lib/api/ckbtc-minter.api.ts
```diff
@@ -92,6 +93,25 @@ export const retrieveBtc = async ({
   return result;
 };
 
+export const estimateFee = async ({
```
lmuntaner, line 96: Should the name be `estimatedFee`?

`estimateFee` seems more an action to estimate the fee based on some parameters.

frontend/src/lib/components/accounts/BitcoinEstimatedFee.svelte
```diff
@@ -0,0 +1,49 @@
...
+  import { i18n } from "$lib/stores/i18n";
+  import { numberToE8s } from "$lib/utils/token.utils";
+  import { formatEstimatedFee } from "$lib/utils/bitcoin.utils";
+
+  export let amount: number | undefined = undefined;
+  export let selectedNetwork: TransactionNetwork | undefined = undefined;
+
+  export let bitcoinEstimatedFee: bigint | undefined | null = undefined;
+
+  const estimateFee = async () => {
```
lmuntaner, line 14: How about naming it `loadEstimatedFee`? this loads the fee right?

frontend/src/lib/components/accounts/BitcoinEstimatedFee.svelte
```diff
@@ -0,0 +1,49 @@
...
+  import { estimateFee as estimateFeeService } from "$lib/services/ckbtc-minter.services";
+  import { debounce, nonNullish } from "@dfinity/utils";
+  import { i18n } from "$lib/stores/i18n";
+  import { numberToE8s } from "$lib/utils/token.utils";
+  import { formatEstimatedFee } from "$lib/utils/bitcoin.utils";
+
+  export let amount: number | undefined = undefined;
+  export let selectedNetwork: TransactionNetwork | undefined = undefined;
+
+  export let bitcoinEstimatedFee: bigint | undefined | null = undefined;
```
lmuntaner, line 12: Don't we want this in a store?

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
```diff
@@ -70,11 +82,22 @@
...
   <p slot="description" class="value">
     {replacePlaceholders($i18n.accounts.ckbtc_transaction_description, {
       $token: token.symbol,
     })}
   </p>
+  <BitcoinEstimatedFee
+    slot="additional-info-form"
+    {selectedNetwork}
+    amount={userAmount}
+    bind:bitcoinEstimatedFee
```
lmuntaner, line 97: Isn't it weird that the fee is loaded in a child component, where the parent binds to it and then passes it to another child component?

Could we load the fee in this component?

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
```diff
@@ -57,7 +63,13 @@
     }
   };
 
-  // TODO(GIX-1330): display bitcoin transaction fee
+  let userAmount: number | undefined = undefined;
+  const validateAmount: ValidateAmountFn = (
```
lmuntaner, line 67: Is this hijacking the validate function for another objective? 😂

frontend/src/lib/components/accounts/BitcoinEstimatedFee.svelte
lmuntaner: I guess not because it depends on the amount, is that correct?

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
lmuntaner: Maybe we can add export `amount` from TransactionModal` so that you can bind it here?

frontend/src/lib/utils/bitcoin.utils.ts
```diff
@@ -0,0 +1,2 @@
+export const formatEstimatedFee = (bitcoinEstimatedFee: bigint): number =>
+  Number(bitcoinEstimatedFee) / 100_000_000;
```
lmuntaner, line 2: Is this E8S_PER_ICP?

Shall we have a constant to understand where it comes from and maybe why we don't use the normal one?

This depends on the "decimals" of the ledger metadata I guess, right?

lmuntaner (commented):
Looks good!

I added a few questions / comments

frontend/src/lib/api/ckbtc-minter.api.ts
peterpeterparker: It was originaly `estimatedFee` but I renamed to `estimateFee` before merge in `ic-js` because the canister function name is `estimate_fee`. So makes sense to follow with same name.

frontend/src/lib/components/accounts/BitcoinEstimatedFee.svelte
peterpeterparker: I couldn't care less so sure, done

frontend/src/lib/components/accounts/BitcoinEstimatedFee.svelte
peterpeterparker: Because of the amount and also because it is only fetched when user want to convert ckBTC to BTC and also because not sure it is a fix fee on the bitcoin network, I understood, might be wrong, that it's fluctuating.

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
peterpeterparker: Let's see in further development. As currently implemented it allows to reuse the component which you generally like 😉. I added a TODO.

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
peterpeterparker: That was my first idea but, validateAmount is a bit better for the job I think because it is triggered only under certain circumstances and therefore we can spare query+update call to the bitcoin canister API that way.

We can rename the validateAmount function, should I had a TODO to rename it?

frontend/src/lib/utils/bitcoin.utils.ts
peterpeterparker: > Is this E8S_PER_ICP?

No it isn't

> Shall we have a constant

Done

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
lmuntaner: I'm not sure about renaming, `validateAmount` needs to return an error string. Therefore, it makes sense.

I guess it's ok to use a functionality for another use case than the one intended initially.

frontend/src/lib/components/accounts/BitcoinEstimatedFee.svelte
lmuntaner: ok, thanks for the explanation!

lmuntaner (approved):
Thanks!

PR #1951, opened 2023-02-24
Feat: Use new and fewer endpoints after participation

lmuntaner:
# Motivation

Use new and fewer endpoints to load data after participation.

loadSnsSummary loaded data that didn't change after participation.

# Changes

* New sns api function `querySnsLifecycle`.
* Remove `loadSnsSummary`.
* New sns service `loadSnsLifecycle`.
* Use `loadSnsTotalCommitment` and `loadSnsLifecycle` instead of `loadSnsSummary` after participation.
* New method `updateLifecycle` in `snsQueryStore`.

# Tests

* Test for new sns api function `querySnsLifecycle`.
* Remove tests for `loadSnsSummary`.
* New test for sns service `loadSnsLifecycle`.
* Test new method `updateLifecycle` in `snsQueryStore`.

### DIFF
--- a/frontend/src/lib/pages/ProjectDetail.svelte
+++ b/frontend/src/lib/pages/ProjectDetail.svelte
@@ -6,7 +6,7 @@
   import { AppPath } from "$lib/constants/routes.constants";
   import { layoutTitleStore } from "$lib/stores/layout.store";
   import {
-    loadSnsSummary,
+    loadSnsLifecycle,
     loadSnsSwapCommitment,
     loadSnsTotalCommitment,
   } from "$lib/services/sns.services";
@@ -31,18 +31,7 @@
 
   $: if (nonNullish(rootCanisterId) && isSignedIn($authStore.identity)) {
     loadCommitment(rootCanisterId);
-    loadTotalCommitments(rootCanisterId);
   }
-
-  const loadSummary = (rootCanisterId: string) =>
-    loadSnsSummary({
-      rootCanisterId,
-      onError: () => {
-        // Set to not found
-        $projectDetailStore.summary = undefined;
-        goBack();
-      },
-    });
 
   const loadCommitment = (rootCanisterId: string) =>
     loadSnsSwapCommitment({
@@ -54,9 +43,6 @@
       },
     });
 
-  const loadTotalCommitments = (rootCanisterId: string) =>
-    loadSnsTotalCommitment({ rootCanisterId });
-
   const reload = async () => {
     if (rootCanisterId === undefined || rootCanisterId === null) {
       // We cannot reload data for an undefined rootCanisterd but we silent the error here because it most probably means that the user has already navigated away of the detail route
@@ -64,7 +50,8 @@
     }
 
     await Promise.all([
-      loadSummary(rootCanisterId),
+      loadSnsTotalCommitment({ rootCanisterId }),
+      loadSnsLifecycle({ rootCanisterId }),
       loadCommitment(rootCanisterId),
     ]);
   };
--- a/frontend/src/lib/services/sns.services.ts
+++ b/frontend/src/lib/services/sns.services.ts
@@ -19,7 +19,7 @@ import type {
   SnsGetDerivedStateResponse,
   SnsGetLifecycleResponse,
 } from "@dfinity/sns";
-import { fromNullable } from "@dfinity/utils";
+import { fromNullable, nonNullish } from "@dfinity/utils";
 import { get } from "svelte/store";
 import { getAuthenticatedIdentity } from "./auth.services";
 import { queryAndUpdate } from "./utils.services";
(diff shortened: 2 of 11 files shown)


frontend/src/lib/services/sns.services.ts
```diff
@@ -187,6 +150,39 @@ export const loadSnsTotalCommitment = async ({
...
+  queryAndUpdate<SnsGetLifecycleResponse | undefined, unknown>({
+    request: ({ certified, identity }) =>
+      querySnsLifecycle({
+        rootCanisterId,
+        identity,
+        certified,
+      }),
+    onLoad: ({ response: lifecycleResponse }) => {
+      const lifecycle = fromNullable(lifecycleResponse?.lifecycle ?? []);
+      if (lifecycle !== undefined) {
```
peterpeterparker, line 167: - `nonNullish`
- if lifecycle is undefined, would it be worth to update the store too?

frontend/src/lib/stores/sns.store.ts
```diff
@@ -216,6 +221,42 @@ const initSnsQueryStore = (): SnsQueryStore => {
         ),
       }));
     },
+
+    /**
+     * Updates only the derived state of a sale.
```
peterpeterparker, line 226: copy pasted comment and not updated?

frontend/src/lib/pages/ProjectDetail.svelte
```diff
@@ -54,17 +43,15 @@
...
-  const loadTotalCommitments = (rootCanisterId: string) =>
-    loadSnsTotalCommitment({ rootCanisterId });
-
   const reload = async () => {
     if (rootCanisterId === undefined || rootCanisterId === null) {
       // We cannot reload data for an undefined rootCanisterd but we silent the error here because it most probably means that the user has already navigated away of the detail route
       return;
     }
 
     await Promise.all([
```
peterpeterparker, line 52: nth: more and more functions in this `Promise.all`, maybe would be nice to extract it to a service so that the component calls one function and the service does the multiple calls. might also be handy for test.

frontend/src/lib/stores/sns.store.ts
```diff
@@ -216,6 +221,42 @@ const initSnsQueryStore = (): SnsQueryStore => {
...
+    }: {
+      lifecycle: SnsSwapLifecycle;
+      rootCanisterId: string;
+    }) {
+      update((data: SnsQueryStoreData) => ({
+        metadata: data?.metadata ?? [],
+        swaps: (data?.swaps ?? []).map((swapData): QuerySnsSwapState => {
+          if (swapData.rootCanisterId === rootCanisterId) {
+            const swap = fromNullable(swapData.swap);
+            if (nonNullish(swap)) {
```
peterpeterparker, line 244: if there is a lifecycle and no swap then adding the lifecycle to the store is silently ignored, is that expected? there is no risk of race condition?

frontend/src/lib/services/sns.services.ts
lmuntaner: No, I don't want to call change the store endpoint. I'd rather check it outside.

frontend/src/lib/stores/sns.store.ts
lmuntaner: Oops, yes, thanks!

frontend/src/lib/pages/ProjectDetail.svelte
lmuntaner: I was planning on adding a workflow test for the whole participation flow in another PR.

I believe that's missing and would help us understand the api calls of the flow better.

frontend/src/lib/stores/sns.store.ts
lmuntaner: If there is no swap, nothing can be changed.

This is more a type issue. `swap: [] | [Swap]`. It's theoretically optional, but it's not really optional.

Ideally, and that's my plan for the future. I'd rather store each thing separately and put them together in a derived store.

But I don't want to do this here.

lmuntaner:

frontend/src/lib/services/sns.services.ts
peterpeterparker: Ok. There is a test? Or does it need a test?

frontend/src/lib/pages/ProjectDetail.svelte
peterpeterparker: Good idea 👍

frontend/src/lib/stores/sns.store.ts
peterpeterparker: What I mean is that in

```
await Promise.all([
      loadSnsTotalCommitment({ rootCanisterId }),
     loadSnsLifecycle({ rootCanisterId }),
     loadCommitment(rootCanisterId),
    ]);
```

`loadSnsLifecycle` can finish and try to update store before `loadCommitment` does. In such case can it be a race condition? Like for example if the user refresh the project detail page?

lmuntaner: No, this doesn't load on init of the page. Only after participation. And we reload only the data that we know might change after a participation:

* User commitment
* Total project commitment
* Project lifecycle.

There is already a project in store when this is called.

This is not about loading the whole project.

peterpeterparker: Good then, thanks for the feedback.

peterpeterparker (approved):
LGTM, thx

PR #1952, opened 2023-02-24
Script to share canister ids.

dskloetd:
# Motivation

When we deploy new canisters to testnets, it's sometimes useful to share ids of multiple canisters to add to canister_ids.json.
This script lets one person do `scripts/canister_ids export medium06` and it will output something like:
```
{
  "internet_identity": {
    "medium06": "qhbym-qaaaa-aaaaa-aaafq-cai"
  },
  "nns-dapp": {
    "medium06": "qsgjb-riaaa-aaaaa-aaaga-cai"
  },
  "sns_aggregator": {
    "medium06": "5v3p4-iyaaa-aaaaa-qaaaa-cai"
  }
}
```
This can then be shared and then other people can do `scripts/canister_ids import` and paste the shared partial config to apply it to their own `canister_ids.json`.

# Changes

Add new script.

# Tests

Tested export and import manually.

### DIFF
--- a/scripts/canister_ids
+++ b/scripts/canister_ids
@@ -26,8 +26,8 @@ if [ "$command" = "export" ]; then
 		exit 1
 	fi
 
-	# This whole block can be done in a single line of jq script but it woudn't
-	# be readable.
+        # This whole block can be done in a single line of jq script but it
+        # woudn't be readable.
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
I don't know much about bash, but if it works and is a dev script. I believe it's ok to merge it, thanks!

PR #1954, opened 2023-02-25
build: web worker esm

peterpeterparker:
# Motivation

Use es modules to build web worker as well.

# Changes

- configure vite to use esm
- drop cjs for esm in few imports


peterpeterparker:
ESM web worker are not supported by Firefox, therefore we cannot rollout this change.

https://developer.mozilla.org/en-US/docs/Web/API/Worker

PR #1956, opened 2023-02-27
Script to set up testnet development

dskloetd:
# Motivation

Running development against testnet requires several steps:
You need
1. a canister_ids.json file
2. several tools installed
3. to run `DFX_NETWORK=$network ./config.sh`
4. to run `npm ci` and `npm run dev` from `frontend/`

It's easy to forget something, especially if you're new to the team.
This scripts checks all the inputs and runs all the steps so you can't forget something.

# Changes

1. Add `dev-testnet.sh` analogous to `dev-local.sh`.
2. The script needs to be run from a git repo and will work from any directory inside the repo so you don't need to `cd` into or out of `frontend/`.
3. Change to `dev-local.sh` to make the previous point work for it as well.

# Tests

Tested manually with a new user account on my Mac.

### DIFF
--- a/scripts/dev-testnet.sh
+++ b/scripts/dev-testnet.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+
+# Combine all the required steps and checks to develop against a testnet.
+# If npm dependencies and canister ids didn't change, it is faster to run
+# `npm run dev` from frontend/ but this script can be helpful to make sure
+# you don't forget anything.
+
+set -euo pipefail
+
+top_dir=$(git rev-parse --show-toplevel)
+ids_file="$top_dir/canister_ids.json"
+
+cd "$top_dir"
+
+if ! [ -f "$ids_file" ]; then
+  echo "You need a 'canister_ids.json' file in $top_dir."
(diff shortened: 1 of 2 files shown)


scripts/dev-testnet.sh
```diff
@@ -0,0 +1,75 @@
...
+need node
+
+networks=$(jq -r '[.[]|keys[]]|unique|join(" ")' canister_ids.json)
+
+usage() {
+  echo "Usage: $0 <network>"
+  echo "Available networks are: $networks"
+}
+
+if [ "$#" != "1" ]; then
```
lmuntaner, line 52: I'm not very familiar with bash, what does this do?

lmuntaner (approved):
Just a question, but already approving. LGTM, thanks!

scripts/dev-testnet.sh
dskloetd: `$#` is the number of arguments passed to the script.
This checks if the number of arguments is not 1 and then it outputs usage instructions and exits with an error code.

lmuntaner: Thanks for the explanation!

PR #1960, opened 2023-02-27
GIX-1350: Refactor and split load accounts api

lmuntaner:
# Motivation

Refactor the loadAccounts api function to enable being more specific when we load the accounts.

For example, reload a specific balance instead of getting all the accounts from nns-dapp and balance as well.

# Changes

* New ledger api `queryAccountBalance`.
* New nns-dapp api `getOrCreateAccount`.
* Move `loadAccounts` as a helper function in accounts services which uses new api functions instead of initializing the canisters directly.

# Tests

* Test new ledger api.
* Test new nns-dapp api.
* Refactor all tests that mocked accounts api `loadAccounts` function.
* Add test for new helper `loadAccounts`.


lmuntaner:

frontend/src/tests/lib/api/ledger.api.spec.ts
```diff
@@ -126,4 +130,32 @@ describe("ledger-api", () => {
...
+
+    beforeEach(() => {
+      jest
+        .spyOn(LedgerCanister, "create")
+        .mockImplementation((): LedgerCanister => ledgerMock);
+    });
+
+    it("gets accounts balance from LedgerCanister", async () => {
+      const certified = true;
+      const actualFee = await queryAccountBalance({
```
dskloetd, line 147: Should that say `actualBalance` instead of `actualFee`?

frontend/src/tests/lib/api/accounts.api.spec.ts
```diff
@@ -1,180 +1,18 @@
...
-      jest.spyOn(LedgerCanister, "create").mockReturnValue(ledgerMock);
-
-      // NNSDapp mock
-      const nnsDappMock = mock<NNSDappCanister>();
-      nnsDappMock.getAccount.mockResolvedValue(mockAccountDetails);
-      jest.spyOn(NNSDappCanister, "create").mockReturnValue(nnsDappMock);
-
-      await loadAccounts({ identity: mockIdentity, certified: true });
-
-      expect(ledgerMock.accountBalance).toBeCalled();
```
dskloetd, line 64: Did you remove all the ledgerMock expectations when moving the tests?

frontend/src/tests/lib/services/accounts.services.spec.ts
```diff
@@ -45,25 +52,132 @@ describe("accounts-services", () => {
...
+        subAccounts: [],
+        hardwareWallets: [],
+        certified: true,
+      };
       await syncAccounts();
 
-      expect(spyLoadAccounts).toHaveBeenCalled();
+      expect(getOrCreateAccountSpy).toHaveBeenCalledTimes(2);
+      expect(queryAccountBalanceSpy).toHaveBeenCalledWith({
+        identity: mockIdentity,
```
dskloetd, line 171: Nit: If you like you can define the params once and use twice with different certified spread in.

frontend/src/tests/lib/api/ledger.api.spec.ts
lmuntaner: oops, I copied from above, yes!

frontend/src/tests/lib/api/accounts.api.spec.ts
lmuntaner: In the loadAccounts test of the accounts service the api layer is mocked. In the api function I mock the creation of canisters

frontend/src/tests/lib/services/accounts.services.spec.ts
lmuntaner: I don't see suck a big improvement for now

dskloetd (approved):

frontend/src/tests/lib/services/accounts.services.spec.ts
dskloetd: Not important, but the benefit is in making it obvious what's the same and what's different, rather than necessarily making the code shorter.

PR #1961, opened 2023-02-27
GIX-1346 Add exponential backoff option to poll

dskloetd:
# Motivation

We want to use exponential backoff for sns sale retries.
This PR only adds the optional functionality to the poll function which is used by the `sns-sale-services.ts`.

# Changes

1. Adds an optional `useExponentialBackoff` parameter to `poll()`.
2. Adds missing tests to test linear retry timestamps.
3. Adds tests to test exponential retry timestamps.

# Tests

1. Unit tests.
2. Manual test by adding a fail counter before `getOpenTicketApi()` in `sns-sale.services.ts`, both with and without exponential backoff.

### DIFF
--- a/frontend/src/lib/utils/utils.ts
+++ b/frontend/src/lib/utils/utils.ts
@@ -201,6 +201,9 @@ const DEFAULT_WAIT_TIME_MS = 500;
  * @param {fn} params.fn Function to call
  * @param {shouldExit} params.shouldExit Function to check whether function should stop polling when it throws an error
  * @param {maxAttempts} params.maxAttempts Param to override the default number of times to poll.
+ * @param {counter} params.counter Param to check how many times it has polled.
+ * @param {millisecondsToWait} params.millisecondsToWait How long to wait between (the first 2, in case of exponential backoff) calls
+ * @param {useExponentialBackoff} params.useExponentialBackoff Whether to use exponential backoff instead of waiting the same time between retries
  *
  * @returns
  */
--- a/frontend/src/lib/utils/utils.ts
+++ b/frontend/src/lib/utils/utils.ts
@@ -202,7 +202,7 @@ const DEFAULT_WAIT_TIME_MS = 500;
  * @param {shouldExit} params.shouldExit Function to check whether function should stop polling when it throws an error
  * @param {maxAttempts} params.maxAttempts Param to override the default number of times to poll.
  * @param {counter} params.counter Param to check how many times it has polled.
- * @param {millisecondsToWait} params.millisecondsToWait How long to wait between (the first 2, in case of exponential backoff) calls
+ * @param {millisecondsToWait} params.millisecondsToWait How long to wait between calls, or the base for the exponential backoff if that's enabled
  * @param {useExponentialBackoff} params.useExponentialBackoff Whether to use exponential backoff instead of waiting the same time between retries
  *
  * @returns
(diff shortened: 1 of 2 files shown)


frontend/src/lib/utils/utils.ts
```diff
@@ -230,12 +235,16 @@ export const poll = async <T>({
     console.error(`Error polling: ${errorToString(error)}`);
   }
   await waitForMilliseconds(millisecondsToWait);
+  if (useExponentialBackoff) {
+    millisecondsToWait *= 2;
```
lmuntaner, line 239: This means that it will start with the `millisecondsToWait`, right?

If we pass 15, then it uses 15 and increases 15 ** 2.

frontend/src/lib/utils/utils.ts
```diff
@@ -201,6 +202,8 @@ export const DEFAULT_MAX_POLLING_ATTEMPTS = 10;
  * @param {shouldExit} params.shouldExit Function to check whether function should stop polling when it throws an error
  * @param {maxAttempts} params.maxAttempts Param to override the default number of times to poll.
  * @param {counter} params.counter Param to check how many times it has polled.
+ * @param {millisecondsToWait} params.millisecondsToWait How long to wait between (the first 2, in case of exponential backoff) calls
```
lmuntaner, line 205: It sets the base for the exponential increase. Worth adding?

frontend/src/tests/lib/utils/utils.spec.ts
```diff
@@ -263,14 +263,16 @@ describe("utils", () => {
   });
 
   describe("poll", () => {
+    describe("without timers", () => {
+    // TODO proper indent before merge. Left it like this for easy reviewing.
```
lmuntaner, line 267: Comment so that you remember to remove the comment

frontend/src/tests/lib/utils/utils.spec.ts
```diff
@@ -316,10 +333,79 @@ describe("utils", () => {
...
+        return timestamps;
+      };
+
+      it("should have a default wait time of 500ms", async () => {
+        expect(
+          await getTimestamps({
+            maxAttempts: 2,
+            useExponentialBackoff: false,
+          })
+        ).toEqual([0, 500]);
```
lmuntaner, line 386: Is it worth setting `500` as a constant and setting the value with the calculation to understand where it comes from?

It uses the default, therefore it can be confusing how we got to 500.

lmuntaner (approved):
It looks great! Thanks!

I added some minor comments, but already approving.

frontend/src/tests/lib/utils/utils.spec.ts
dskloetd: Discussed offline: this test explicitly only tests the default.
Other functionality is tested below.

frontend/src/tests/lib/utils/utils.spec.ts
```diff
@@ -316,10 +333,79 @@ describe("utils", () => {
...
+      });
+
+      it("should wait the same amount between calls", async () => {
+        expect(
+          await getTimestamps({
+            maxAttempts: 5,
+            millisecondsToWait: 1000,
+            useExponentialBackoff: false,
+          })
+        ).toEqual([0, 1000, 2000, 3000, 4000]);
```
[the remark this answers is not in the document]
dskloetd: Keeping as is, as disucced.

PR #1964, opened 2023-02-27
GIX-1346 Use exponential backoff in payments flow

dskloetd:
# Motivation

We want to improve the experience but also reduce the load caused by the sns sale under heavy load.

# Changes

1. Enable exponential backoff for retries in the sns sale.
2. Reduce the initial retry time from 15 seconds to 2 seconds.

# Tests

Tests are updated to advance timers exponentially.
Tested manually by making getOpenTicket fail several times before succeeding and looking at the console.


lmuntaner (approved):
Awesome, thanks!

PR #1965, opened 2023-02-27
Clarify that test passed if it fails because of console logs

dskloetd:
# Motivation

We make tests fail if they generate console log.
But we only cause an error after the test finishes, so the user still gets the benefit from running the test.
But this wasn't clear in the error message.

# Changes

Improve the error message.

# Tests

none


lmuntaner (approved):
Thanks!

PR #1966, opened 2023-02-27
Trailing slash

bitdivine:
# Motivation
We would like to support `/launchpad` as a path.  `/launchpad/` is already supported but users are confused when the path doesn't work without the trailing slash.

# Changes
- Factored out the list of supported path rewrites.
- Added `"" -> "/index.html"` to the list of path rewrites
- Made the path to `main.js` absolute.

# Tests
- Added `/launchpad` to the paths tested, both logged in and logged out.
- Made the `launchpad` test verify that the test has loaded, as with the existing test a 404 or HTML served as plain text would pass.

### DIFF
--- a/e2e-tests/sns-init-assets/sns_init.yaml
+++ b/e2e-tests/sns-init-assets/sns_init.yaml
@@ -70,7 +70,8 @@ min_icp_e8s: 100000000
 # principal IDs. In most use-cases, this would be the same as the original set
 # of controller(s). Must not be empty.
 #
-fallback_controller_principal_ids: [hpikg-6exdt-jn33w-ndty3-fc7jc-tl2lr-buih3-cs3y7-tftkp-sfp62-gqe]
+fallback_controller_principal_ids:
+  [hpikg-6exdt-jn33w-ndty3-fc7jc-tl2lr-buih3-cs3y7-tftkp-sfp62-gqe]
 #
 # This field sets the initial token distribution. Initially, there is only support for
 # the FractionalDeveloperVotingPower strategy. The FractionalDeveloperVotingPower token
--- a/e2e-tests/sns-init-assets/sns_init.yaml
+++ b/e2e-tests/sns-init-assets/sns_init.yaml
@@ -70,8 +70,7 @@ min_icp_e8s: 100000000
 # principal IDs. In most use-cases, this would be the same as the original set
 # of controller(s). Must not be empty.
 #
-fallback_controller_principal_ids:
-  [hpikg-6exdt-jn33w-ndty3-fc7jc-tl2lr-buih3-cs3y7-tftkp-sfp62-gqe]
+fallback_controller_principal_ids: [hpikg-6exdt-jn33w-ndty3-fc7jc-tl2lr-buih3-cs3y7-tftkp-sfp62-gqe]
 #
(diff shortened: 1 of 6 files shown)


e2e-tests/sns-init-assets/sns_init.yaml
```diff
@@ -70,7 +70,8 @@ min_icp_e8s: 100000000
 # principal IDs. In most use-cases, this would be the same as the original set
 # of controller(s). Must not be empty.
 #
-fallback_controller_principal_ids: [hpikg-6exdt-jn33w-ndty3-fc7jc-tl2lr-buih3-cs3y7-tftkp-sfp62-gqe]
+fallback_controller_principal_ids:
+  [hpikg-6exdt-jn33w-ndty3-fc7jc-tl2lr-buih3-cs3y7-tftkp-sfp62-gqe]
```
bitdivine, line 74: This gets changed by `./scripts/fmt`.  Not sure why it's different from main.  The same script seems to be run in CI.  Weird.

rs/backend/src/assets.rs
```diff
@@ -105,6 +105,8 @@ pub struct Assets(HashMap<String, Asset>);
 impl Assets {
     /// List of content encodings supported by the assets database.
     const CONTENT_ENCODINGS: [ContentEncoding; 2] = [ContentEncoding::GZip, ContentEncoding::Identity];
+    /// List of suffix changes that may be made.
+    const SUFFIX_REWRITES: [(&str, &str); 3] = [("", ""), ("/", "/index.html"), ("", "/index.html")];
```
bitdivine, line 109: The list above is used in two places below.  They used to keep the list separately.  It seems better to make sure they have the same list.

peterpeterparker: Not sure why there are two times the prefix `""`.

```
("", "")
// and
("", "/index.html")
```

Can you add a comment with examples of usage?

peterpeterparker: Note: I got it once looking at the usage, just a comment would be welcomed

rs/backend/src/assets.rs
```diff
@@ -125,16 +127,18 @@ impl Assets {
     ///   take browser capabilities into account.
     pub fn get(&self, path: &str) -> Option<(ContentEncoding, &Asset)> {
         // Note: The logic for finding an asset is the reverse of listing all asset paths.
-        for (old_suffix, new_suffix) in [("/", "/index.html")] {
+        for (old_suffix, new_suffix) in Self::SUFFIX_REWRITES {
             if let Some(root) = path.strip_suffix(old_suffix) {
                 let new_path = root.to_string() + new_suffix;
-                return self.get(&new_path);
+                if let Some(ans) = Self::CONTENT_ENCODINGS.iter().find_map(|content_encoding| {
```
peterpeterparker, line 133: what's `ans`? i.e. from reading this variable name, I don't get what it does. `ans` for `a name suffix` or else?

peterpeterparker (approved):
LGTM, minor comments

rs/backend/src/assets.rs
bitdivine: Renamed to `asset_with_encoding`.  I tend to use `ans` in functions to hold the value that will be returned by that function but I guess that is not universal.

PR #1969, opened 2023-02-28
GIX-1352: Init accounts service

lmuntaner:
# Motivation

Do not show errors from loading accounts on init.

# Changes

* Add a parameter to `syncAccounts` which is an error handler.
* Extract the current logic on error to a default error handler for `syncAccounts`.
* New `initAccounts` service that uses `syncAccounts` with an empty function as error handler.

# Tests

* Test new functionality of `syncAccounts`.
* Test for new service `initAcconts`.

### DIFF
--- a/frontend/src/lib/services/accounts.services.ts
+++ b/frontend/src/lib/services/accounts.services.ts
@@ -106,13 +106,18 @@ export const loadAccounts = async ({
   };
 };
 
+type SyncAccontsErrorHandler = (params: {
+  err: unknown;
+  certified: boolean;
+}) => void;
+
 /**
  * Default error handler for syncAccounts.
  *
  * Ignores non-certified errors.
  * Resets accountsStore and shows toast for certified errors.
  */
-const defaultErrorHandlerAccounts = ({
+const defaultErrorHandlerAccounts: SyncAccontsErrorHandler = ({
   err,
   certified,
 }: {
@@ -137,7 +142,7 @@ const defaultErrorHandlerAccounts = ({
  * Loads the account data using the ledger and the nns dapp canister.
  */
 export const syncAccounts = (
-  errorHandler: typeof defaultErrorHandlerAccounts = defaultErrorHandlerAccounts
+  errorHandler: SyncAccontsErrorHandler = defaultErrorHandlerAccounts
 ): Promise<void> => {
   return queryAndUpdate<AccountsStoreData, unknown>({
     request: (options) => loadAccounts(options),
@@ -151,7 +156,8 @@ export const syncAccounts = (
   });
 };
 
-const ignoreErrors = () => undefined;
+const ignoreErrors: SyncAccontsErrorHandler = () => undefined;
+
 /**
  * This function is called on app load to sync the accounts.
  *
--- a/frontend/src/lib/services/accounts.services.ts
+++ b/frontend/src/lib/services/accounts.services.ts
@@ -107,32 +107,65 @@ export const loadAccounts = async ({
 };
 
 /**
+ * Default error handler for syncAccounts.
+ *
+ * Ignores non-certified errors.
+ * Resets accountsStore and shows toast for certified errors.
+ */
+const defaultErrorHandlerAccounts = ({
+  err,
+  certified,
+}: {
+  err: unknown;
+  certified: boolean;
+}) => {
+  if (certified !== true) {
+    return;
+  }
+
+  // Explicitly handle only UPDATE errors
+  accountsStore.reset();
+
+  toastsError(
+    toToastError({
+      err,
+      fallbackErrorLabelKey: "error.accounts_not_found",
+    })
+  );
+};
+
+type ErrorHanlder = ({
+  err,
+  certified,
+}: {
+  err: unknown;
+  certified: boolean;
+}) => void;
+/**
  * - sync: load the account data using the ledger and the nns dapp canister itself
  */
-export const syncAccounts = (): Promise<void> => {
+export const syncAccounts = (
+  errorHandler: ErrorHanlder = defaultErrorHandlerAccounts
+): Promise<void> => {
   return queryAndUpdate<AccountsStoreData, unknown>({
     request: (options) => loadAccounts(options),
     onLoad: ({ response: accounts }) => accountsStore.set(accounts),
     onError: ({ error: err, certified }) => {
       console.error(err);
 
-      if (certified !== true) {
-        return;
-      }
-
-      // Explicitly handle only UPDATE errors
-      accountsStore.reset();
-
-      toastsError(
-        toToastError({
-          err,
-          fallbackErrorLabelKey: "error.accounts_not_found",
-        })
-      );
+      errorHandler({ err, certified });
     },
     logMessage: "Syncing Accounts",
   });
 };
+
+const ignoreErrors = () => undefined;
+/**
+ * This function is called on app load to sync the accounts.
+ *
+ * It ignores errors and does not show any toasts. Accounts will be synced again.
+ */
+export const initAccounts = () => syncAccounts(ignoreErrors);
 
 export const addSubAccount = async ({
   name,
--- a/frontend/src/lib/services/accounts.services.ts
+++ b/frontend/src/lib/services/accounts.services.ts
@@ -119,11 +119,10 @@ const defaultErrorHandlerAccounts = ({
   err: unknown;
   certified: boolean;
 }) => {
-  if (certified !== true) {
+  if (!certified) {
     return;
   }
 
-  // Explicitly handle only UPDATE errors
   accountsStore.reset();
 
   toastsError(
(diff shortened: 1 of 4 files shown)


lmuntaner:

frontend/src/lib/services/accounts.services.ts
```diff
@@ -106,34 +106,67 @@ export const loadAccounts = async ({
...
+ * Resets accountsStore and shows toast for certified errors.
+ */
+const defaultErrorHandlerAccounts = ({
+  err,
+  certified,
+}: {
+  err: unknown;
+  certified: boolean;
+}) => {
+  if (certified !== true) {
```
dskloetd, line 122: You already know certified is boolean so this can just be `if(!certified) {`, right?

frontend/src/lib/services/accounts.services.ts
```diff
@@ -106,34 +106,67 @@ export const loadAccounts = async ({
...
+    return;
+  }
+
+  // Explicitly handle only UPDATE errors
```
dskloetd, line 126: How does `accountsStore.reset();` "explicitly handle only UPDATE errors"?

frontend/src/lib/services/accounts.services.ts
```diff
@@ -106,34 +106,67 @@ export const loadAccounts = async ({
   };
 };
 
+/**
+ * Default error handler for syncAccounts.
+ *
+ * Ignores non-certified errors.
+ * Resets accountsStore and shows toast for certified errors.
+ */
+const defaultErrorHandlerAccounts = ({
```
dskloetd, line 115: Explicitly declare this as type ErrorHandler?

frontend/src/lib/services/accounts.services.ts
```diff
@@ -106,34 +106,67 @@ export const loadAccounts = async ({
...
+};
+
+type ErrorHanlder = ({
+  err,
+  certified,
+}: {
+  err: unknown;
+  certified: boolean;
+}) => void;
 /**
```
dskloetd, line 144: blank line?

frontend/src/lib/services/accounts.services.ts
```diff
@@ -106,34 +106,67 @@ export const loadAccounts = async ({
...
-        })
-      );
+      errorHandler({ err, certified });
     },
     logMessage: "Syncing Accounts",
   });
 };
 
+const ignoreErrors = () => undefined;
+/**
```
dskloetd, line 163: blank line?

frontend/src/lib/services/accounts.services.ts
```diff
@@ -106,34 +106,67 @@ export const loadAccounts = async ({
...
-          fallbackErrorLabelKey: "error.accounts_not_found",
-        })
-      );
+      errorHandler({ err, certified });
     },
     logMessage: "Syncing Accounts",
   });
 };
 
+const ignoreErrors = () => undefined;
```
dskloetd, line 162: Explicitly declare of type ErrorHandler?

frontend/src/tests/lib/services/accounts.services.spec.ts
```diff
@@ -168,6 +170,139 @@ describe("accounts-services", () => {
     });
   });
 
+  describe("initAccounts", () => {
+    beforeEach(() => {
+      toastsStore.reset();
+    });
+    it("should sync accounts", async () => {
```
dskloetd, line 177: I would expect a blank line here. Do we have any rules or practices for when to put a blank line?

frontend/src/tests/lib/services/accounts.services.spec.ts
```diff
@@ -168,6 +170,139 @@ describe("accounts-services", () => {
...
+      jest
+        .spyOn(nnsdappApi, "queryAccount")
+        .mockRejectedValue(new Error(errorTest));
+
+      await syncAccounts();
+
+      const toastsData = get(toastsStore);
+      const toast = toastsData[0];
+      expect(toast.text).toEqual(`${en.error.accounts_not_found} ${errorTest}`);
+      expect(toast.level).toEqual("error");
```
dskloetd, line 281: optional: If you like you can combine these 4 lines into 1 expect call:
```
  expect(get(toastsStore)).toMatchObject([
    {
      level: "error",
      text: `${en.error.accounts_not_found} ${errorTest}`,
    }]);
```

frontend/src/tests/lib/services/accounts.services.spec.ts
```diff
@@ -168,6 +170,139 @@ describe("accounts-services", () => {
...
+    });
+
+    it("should use handler passed", async () => {
+      const errorTest = "test";
```
dskloetd, line 285: Every time you use this you pass it to `new Error()`. Just create the new Error here?

frontend/src/lib/services/app.services.ts
```diff
@@ -1,9 +1,9 @@
-import { syncAccounts } from "./accounts.services";
+import { initAccounts } from "./accounts.services";
 
 export const initAppPrivateData = (): Promise<
   [PromiseSettledResult<void[]>]
 > => {
-  const initNns: Promise<void>[] = [syncAccounts()];
+  const initNns: Promise<void>[] = [initAccounts()];
```
dskloetd, line 6: Can you add a test that tests that app.services.ts indeed doesn't get errors on account init failure?

frontend/src/lib/services/accounts.services.ts
lmuntaner: True! thanks. I copied pasted without thinking

frontend/src/lib/services/accounts.services.ts
lmuntaner: This is more related to all the following lines. But I already added the comment at the beginning of the function. I'll remove this.

frontend/src/lib/services/accounts.services.ts
lmuntaner: true, makes sense

frontend/src/lib/services/accounts.services.ts
lmuntaner: Why?

frontend/src/lib/services/accounts.services.ts
lmuntaner: What blank line?

frontend/src/lib/services/app.services.ts
lmuntaner: yes

frontend/src/tests/lib/services/accounts.services.spec.ts
lmuntaner: no, we don't for this. I can add it

frontend/src/tests/lib/services/accounts.services.spec.ts
lmuntaner: The object has more things like an id that is a symbol that I don't want to handle how to match.

frontend/src/tests/lib/services/accounts.services.spec.ts
lmuntaner: Yes, I copied from above but here I can expect with the whole error, thanks

lmuntaner:

frontend/src/tests/lib/services/accounts.services.spec.ts
dskloetd: Yes, you don't have to. The code should work as is. Did you try it?

frontend/src/lib/services/accounts.services.ts
dskloetd: In a previous version I saw you declared a `type ErrorHandler`.
I just thought you could declare `defaultErrorHandlerAccounts` to be of that type.
But now I don't see the whole type anymore.
What happened?

frontend/src/lib/services/accounts.services.ts
```diff
@@ -107,33 +107,58 @@ export const loadAccounts = async ({
...
+      fallbackErrorLabelKey: "error.accounts_not_found",
+    })
+  );
+};
+
+/**
+ * Loads the account data using the ledger and the nns dapp canister.
+ */
+export const syncAccounts = (
+  errorHandler: typeof defaultErrorHandlerAccounts = defaultErrorHandlerAccounts
```
dskloetd, line 140: I think I liked it better when you had type `ErrorHandler`.
Did you think I asked you to remove that?

frontend/src/lib/services/accounts.services.ts
lmuntaner: Before makes sense to me. Not afterwards.

frontend/src/lib/services/accounts.services.ts
lmuntaner: ah, ok

frontend/src/lib/services/accounts.services.ts
lmuntaner: Yes

lmuntaner: I'll revert

frontend/src/lib/services/accounts.services.ts
lmuntaner: added before

lmuntaner:
@dskloetd please review

frontend/src/lib/services/accounts.services.ts
```diff
@@ -106,34 +106,65 @@ export const loadAccounts = async ({
...
+ *
+ * Ignores non-certified errors.
+ * Resets accountsStore and shows toast for certified errors.
+ */
+const defaultErrorHandlerAccounts: SyncAccontsErrorHandler = ({
+  err,
+  certified,
+}: {
+  err: unknown;
+  certified: boolean;
```
dskloetd, line 125: Is it necessary to repeat these declarations from the SyncAccontsErrorHandler?

dskloetd (approved):
Just a question. Not blocking.

frontend/src/lib/services/accounts.services.ts
lmuntaner: What do you mean?

Whether I should create a more generic type?

PR #1970, opened 2023-02-28
build: chunks group workers and api-services

peterpeterparker:
# Motivation

- fix missing `api-services` in list of grouped chunks
- add web worker to same group which now seem fine if added to the group


lmuntaner (approved):
Thanks!

PR #1971, opened 2023-02-28
GIX-1353: Poll accounts

lmuntaner:
# Motivation

Add a service to retry getting accounts if they fail.

# Changes

* New accounts service `pollAccounts`.

# Tests

* Test new accounts service `pollAccounts`.


lmuntaner:

frontend/src/lib/services/accounts.services.ts
```diff
@@ -317,3 +322,53 @@ const renameError = ({
 
   return { success: false, err: labelKey };
 };
+
+const pollLoadAccounts = async (params: {
+  identity: Identity;
+  certified: boolean;
+}): Promise<AccountsStoreData> =>
+  poll({
+    fn: () => loadAccounts(params),
```
mstrasinskis, line 331: Would be great to retry per account. Not a strong opinion and definitely not for this PR.

mstrasinskis (approved):
LGTM, thanks!

frontend/src/lib/services/accounts.services.ts
lmuntaner: Yes, I agree. But I didn't want to go into all the details on where exactly to retry or not...

It won't add much value for now anyway. All these calls are in the NNS subnet. If that's overloaded, any of these calls would need to be retried.

PR #1973, opened 2023-02-28
Gix 1349 close ticket on insufficient funds error

mstrasinskis:
# Motivation

Remove open ticket on  insufficient funds error.
https://dfinity.atlassian.net/browse/GIX-1349

# Changes

- refactor: removeOpenTicket
- refactor: replace custom CreatedInFuture handling to be covered by retry mechanism (in sync w/ Mario)

# Tests

- the notifyFailure call should be done on  insufficient funds error.
- retry transfer should also happened on `CreatedInFuture` error.
- updated some tests to be sure the `notifyFailure` call is not happened.


frontend/src/tests/lib/services/sns-sale.services.spec.ts
```diff
@@ -974,7 +979,9 @@ describe("sns-api", () => {
       const retriesUntilSuccess = 4;
       sendICPSpy
         .mockRejectedValueOnce(new Error("Connection error"))
-        .mockRejectedValueOnce(new Error("Connection error"))
+        .mockRejectedValueOnce(
+          new TxCreatedInFutureError("Created in future error")
```
lmuntaner, line 983: Can you change the title of the test to: "should poll transfer during unknown issues or `TxCreatedInFutureError`"?

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -704,36 +725,25 @@ export const participateInSnsSale = async ({
...
-    // (in theory this error should be gone in a couple of seconds)
-    if (err instanceof TxCreatedInFutureError) {
-      const retryIn = SALE_PARTICIPATION_RETRY_SECONDS;
-
-      setTimeout(() => {
-        participateInSnsSale({
+    // TODO(Maks): refactor w/ a switch
+    // if duplicated transfer, silently continue the flow
+    if (!(err instanceof TxDuplicateError)) {
+      if (err instanceof InsufficientFundsError) {
```
lmuntaner, line 731: Why do you need it inside `!(err instanceof TxDuplicateError)`?

Can't you add an `if` for `TxDuplicateError` itself?

lmuntaner (approved):
A few comments.

Already approving to not block

frontend/src/tests/lib/services/sns-sale.services.spec.ts
mstrasinskis: Good point, done!

frontend/src/lib/services/sns-sale.services.ts
mstrasinskis: Then this will be an empty if block, to just continue the flow. There is already a todo for this, thought to do it in a separate task, to avoid the merge conflicts, but probably now is a great time to do it.

PR #1975, opened 2023-02-28
fix: voting UI and cache

peterpeterparker:
# Motivation

The introduction of the neurons cache leads to an issue when the user vote. Because the list of neurons is not refreshed with the very last ballots, the voting actions ("buttons yes no") remains notably active after vote.

There is also a minor UI quirks, a line is positioned absolutely to the modal instead of parent.

Also we were not able to create dummy proposals anymore because the URL of the proposal are white listed now.


frontend/src/lib/services/vote-registration.services.ts
```diff
@@ -379,6 +380,9 @@ const updateAfterVoteRegistration = async (
       })
     );
 
+  // We need to list the neurons with the very last ballots because the user just voted
```
lmuntaner, line 383: Can you add a todo,

I believe that the layer should be called `GovernanceApiService` and all calls should go that way.

And when we vote, the neurons cache is disabled automatically.

@dskloetd FYI, a reason to call it `GovernanceApiService`

lmuntaner (approved):
I added one comment, but already approving

frontend/src/lib/services/vote-registration.services.ts
peterpeterparker: Sure done

PR #1976, opened 2023-02-28
feat: nns.ic0.app is (still) main url of NNS-dapp

peterpeterparker:
# Motivation

`nns.ic0.app` is (still) main url of NNS-dapp therefore `robots.ts`, `sitemap.xml` and other static meta tags should still use it.


lmuntaner (approved):
Thanks!

PR #1977, opened 2023-02-28
GIX-1357 Show "high load" message after ~1 minute but keep retrying

dskloetd:
# Motivation

We decide, when we keep retrying requests, to show a message after ~1 minute to indicate that there is high load.

# Changes

After a failure if at least 1 minute has passed, show a toast indicating high load.
Because of exponential backoff, this will probably happen after 62 seconds (2 + 4 + 8 + 16 + 32) + the time it took for those 6 requests to fail.

# Tests

Added tests to `src/tests/lib/utils/utils.spec.ts`.
Tested manually on testnet with smaller timeout and hardcoded error.


frontend/src/lib/utils/utils.ts
```diff
@@ -214,13 +220,17 @@ export const poll = async <T>({
   counter = 0,
   millisecondsToWait = DEFAULT_WAIT_TIME_MS,
   useExponentialBackoff = false,
+  showHighLoadMessage = true,
```
lmuntaner, line 223: How about if we use `millisBeforeHighLoadMessage`? We don't have a default value.

* `undefined` then no message
* `number` then message

frontend/src/tests/lib/utils/utils.spec.ts
```diff
@@ -340,10 +344,23 @@ describe("utils", () => {
...
+          level: "error",
+          text: `${en.error.high_load_retrying}`,
+        },
+      ];
+
       beforeEach(() => {
         jest.useFakeTimers();
       });
 
+      const advanceTime = async (): Promise<void> => {
```
lmuntaner, line 358: Not sure that will be sufficient in my issue of the `pollAccounts`

lmuntaner (changes requested):
Suggested another interface.

Let me know what you think.

frontend/src/lib/utils/utils.ts
dskloetd: The boolean is needed in the recursion to make sure we only show it once. Or are you proposing to use null for that as well? Then we need to add an explicit value to all callers, no?

frontend/src/tests/lib/utils/utils.spec.ts
dskloetd: Don't worry about it now. We'll factor out later if possible. OK?

frontend/src/lib/utils/utils.ts
lmuntaner: Why can't you set it to `undefined`?

frontend/src/tests/lib/utils/utils.spec.ts
lmuntaner: yes, sounds good.

frontend/src/lib/utils/utils.ts
dskloetd: I rewrote it to be a for-loop instead of recursion, so we don't need any additional parameters for recursion.
I made it count failures instead of millis before the "high load" message.
I made it not default.

frontend/src/lib/utils/utils.ts
```diff
@@ -194,58 +195,62 @@ export class PollingLimitExceededError extends Error {}
...
-  try {
-    return await fn();
-  } catch (error: unknown) {
-    if (shouldExit(error)) {
-      throw error;
+  for (let counter = 0; counter < maxAttempts; counter++) {
+    if (counter > 0) {
+      if (
+        nonNullish(failuresBeforeHighLoadMessage) &&
+        counter == failuresBeforeHighLoadMessage
```
lmuntaner, line 231: why `==` instead of `===`?

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -315,6 +316,7 @@ const pollNewSaleTicket = async (params: {
     shouldExit: shoulStopPollingNewTicket,
     millisecondsToWait: WAIT_FOR_TICKET_MILLIS,
     useExponentialBackoff: true,
+    failuresBeforeHighLoadMessage: 6,
```
lmuntaner, line 319: Maybe add a constant for this `6`

lmuntaner (changes requested):
I'd rather use `===` always, unless you have a strong reasone not to.

frontend/src/lib/utils/utils.ts
dskloetd: Because there's no lint check. I've changed it to ===.

frontend/src/lib/services/sns-sale.services.ts
dskloetd: Done

lmuntaner (approved):
Thanks!

PR #1979, opened 2023-02-28
Use docker as the standard build mechanism

bitdivine:
# Motivation
* Require only docker in order to deploy `nns-dapp` and `sns_aggregator` to a testnet.
  * Having docker and dfx installed (properly!) is now enough to run: `dfx deploy nns-dapp --network small09`
  * Deployments _may_ be faster if the local docker cache is warm.
* Align testnet deployments more closely with production.
  * The canonical way of building a production image is with docker.  It would be nice if testnet deployments were built in (almost) the same way.

# Changes
* Change the build command in dfx.json to `scripts/docker-build`.

Note: Deploying any of the wasms involves building all the wasms.  This is definitely a point that can be improved.  If the dockerfile is broken down into smaller chunks, this will be possible.

# Tests
- See CI for local builds.
- I have deployed sns_aggregator and nns-dapp to small09 using this code.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -17,7 +17,7 @@
       "type": "custom",
       "candid": "rs/backend/nns-dapp.did",
       "wasm": "nns-dapp.wasm",
-      "build": "./build.sh"
+      "build": "scripts/docker-build"
     },
     "internet_identity": {
       "type": "custom",
@@ -94,7 +94,7 @@
     },
     "sns_aggregator": {
       "build": [
-        "./build-sns-aggregator.sh"
+        "scripts/docker-build"
       ],
       "candid": "rs/sns_aggregator/sns_aggregator.did",
       "wasm": "sns_aggregator.wasm",
--- a/scripts/deploy-snsdemo-testnet
+++ b/scripts/deploy-snsdemo-testnet
@@ -52,4 +52,5 @@ dfx-canister-set-id --canister_name sns_aggregator --canister_id "$AGGREGATOR_CA
 dfx canister info sns_aggregator
 
 : Deploy the aggregator
-RUSTFLAGS="--cfg feature=\"reconfigurable\"" dfx deploy sns_aggregator --network "$DFX_NETWORK"
+./scripts/docker-build
+dfx canister install sns_aggregator --network "$DFX_NETWORK" --wasm sns_aggregator_dev.wasm


lmuntaner:
I agree with the PR in theory.

But I tried with Mac M1 and it's not working:

```
 => [builder  5/16] RUN npm --version                                                                                                                                     0.4s
 => [builder  6/16] RUN curl --fail https://sh.rustup.rs -sSf         | sh -s -- -y --default-toolchain "$(cat config/rust_version)-x86_64-unknown-linux-gnu" --no-modi  14.4s
 => ERROR [builder  7/16] RUN cargo install --version "$(cat config/optimizer_version)" ic-cdk-optimizer                                                                  0.3s
------
 > [builder  7/16] RUN cargo install --version "$(cat config/optimizer_version)" ic-cdk-optimizer:
#19 0.300 qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

I believe is the same reason why we can't check the hash with M1?

bitdivine:
@lmuntaner - I need to borrow an M1 again.  The last time that error came up I got the builds to work on an M1, I just had to go through certain steps when installing docker.  I was somewhat disheartened that many things did not work on M1, it felt as if one could spend a FTE just getting things to work there, but if we can narrow the dependencies to just docker then maybe the effort to get the problem resolved once and for all would be reasonable.

lmuntaner:
I agree. For example, the script to deploy SNS project doesn't work in M1 either.

Without it, it's hard to add e2e tests because we can't run e2e locally in M1.

bitdivine:
@lmuntaner - @dskloetd managed to get docker build working on M1 and the changes he needed have been merged into main.  If we can repeat the exercise with more team members, that would eventually open the door to merging this.  Please note that the very first run takes a long time on M1.  (about 1 hour)  Best done when you are not trying to get something done. My new Linux laptop is a business Excel machine, not the most powerful as as broke my nice T14, but it is still native AMD64 and does a clean build in about 20 minutes.  But with those M1 timings I am hesitant to say that Docker should be the default build tool but it is nice that they are possible and even fast if just a few stages need to be rebuilt.  It means would mean that anyone on the team could verify a release build hash locally, for example.

lmuntaner:
It's inconvenient that it takes long when we want to deploy to a testnet.

Is there a way to have a "build" entry per network? hehe.

bitdivine:
Approved in principle but in practice developers on M1 macs can't build on docker.

PR #1980, opened 2023-02-28
In poll(), don't wait after the last retry

dskloetd:
# Motivation

We have a `poll()` function which retries a call until it passes or some other condition is met.
If you reach the maximum number of retries, it returns an error but currently it still waits another timeout before doing so.

# Changes

Increase and check the counter before waiting for the timeout instead of after calling the recursion.

# Tests

Added a unit test and ran existing unit tests.


lmuntaner (approved):
Thanks!

dskloetd:
The logic got fixed by the changes in https://github.com/dfinity/nns-dapp/pull/1977 so now I'm just merging the test.

PR #1981, opened 2023-02-28
feat: display sale canister loading a bit more in a user friendly way

peterpeterparker:
# Motivation

Display a text instead of "just" disabling the "Participate" button when data is loading.

# Changes

- extract existing pattern in a `SpinnerText` component
- use component to display a loading text - reviewed by product - when data are loading

# Screenshots

connecting


lmuntaner (approved):
Perfect, thanks!

PR #1982, opened 2023-02-28
GIX-1353: Add pollAccounts to nns wallet and accounts

lmuntaner:
# Motivation

In case the accounts fail to load on init. Retry to fetch them in the nns wallet and accounts.

# Changes

* Add a call to `onMount` in NnsWallet.
* Add a call to `onMount` in NnsAccounts.

# Tests

* Add for NnsWallet and NnsAccounts tests.
* Fill accounts store in Wallet and Accounts tests.

### DIFF
--- a/frontend/src/tests/lib/pages/NnsAccounts.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsAccounts.spec.ts
@@ -2,20 +2,21 @@
  * @jest-environment jsdom
  */
 
+import * as ledgerApi from "$lib/api/ledger.api";
+import * as nnsDappApi from "$lib/api/nns-dapp.api";
 import NnsAccounts from "$lib/pages/NnsAccounts.svelte";
-import {
-  accountsStore,
-  type AccountsStoreData,
-} from "$lib/stores/accounts.store";
+import { accountsStore } from "$lib/stores/accounts.store";
 import { formatToken } from "$lib/utils/token.utils";
 import { render } from "@testing-library/svelte";
-import type { Subscriber } from "svelte/store";
 import {
-  mockAccountsStoreSubscribe,
+  mockAccountDetails,
   mockHardwareWalletAccount,
   mockMainAccount,
   mockSubAccount,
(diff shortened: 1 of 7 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/pages/NnsAccounts.spec.ts
```diff
@@ -141,5 +127,12 @@ describe("NnsAccounts", () => {
         container.querySelector(".tooltip-wrapper")
       ).not.toBeInTheDocument();
     });
+
+    it("should load accounts", () => {
+      const { container } = render(NnsAccounts, { props: { goToWallet } });
+
+      const article = container.querySelector("article");
```
dskloetd, line 134: What is "article"? How do you know it's an account?

frontend/src/tests/lib/pages/NnsWallet.spec.ts
```diff
@@ -94,14 +116,6 @@ describe("NnsWallet", () => {
       expect(titleRow).not.toBeNull();
     });
 
-    it("should hide spinner when selected account is loaded", async () => {
```
dskloetd, line 97: Why did you remove this?

frontend/src/tests/lib/pages/NnsAccounts.spec.ts
lmuntaner: We don't. I didn't want to change current tests with our new best practices.

This is a legacy test. I agree it should change.

frontend/src/tests/lib/pages/NnsWallet.spec.ts
lmuntaner: It's now called: "should show new accounts after being loaded"

I moved it to the top and instead of just checking that the spinner is not loaded, I check that a component with information is first not rendered, and then wait for it to be rendered.

dskloetd (approved):

PR #1983, opened 2023-02-28
Fix not equal to amount warning

mstrasinskis:
# Motivation

Fix visibility logic of the "actual commitment not equals to ticket amount" warning to work not only for the first participation.

# Changes

- replace (ticket.amount != finalCommitment) with (ticket.amount + userCommitment != finalCommitment)

# Tests

- updated tests for warning visibility

# Screenshots

**Increasing participation w/o warning**
Screen Recording 2023-02-28 at 23 10 17

### DIFF
--- a/frontend/src/lib/components/project-detail/ParticipateButton.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateButton.svelte
@@ -57,6 +57,7 @@
 
   let userCommitment: undefined | bigint;
   $: userCommitment =
+    // swapCommitment=null - not initialized yet
     $projectDetailStore.swapCommitment === null
       ? undefined
       : getCommitmentE8s($projectDetailStore.swapCommitment) ?? BigInt(0);
@@ -90,8 +91,6 @@
       return;
     }
 
-    assertNonNullish(userCommitment);
-
     snsTicketsStore.enablePolling(rootCanisterId);
 
     loadingTicketRootCanisterId = rootCanisterId.toText();
@@ -100,7 +99,7 @@
 
     await restoreSnsSaleParticipation({
       rootCanisterId,
-      userCommitment,
+      userCommitment: userCommitment!,
       postprocess: reload,
       updateProgress,
     });
@@ -113,7 +112,7 @@
   $: if (
     lifecycle === SnsSwapLifecycle.Open &&
     isSignedIn($authStore.identity) &&
-    userCommitment !== undefined
+    nonNullish(userCommitment)
   ) {
     updateTicket();
   }
--- a/frontend/src/lib/components/project-detail/ParticipateButton.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateButton.svelte
@@ -106,7 +106,10 @@
     });
   };
 
-  // skip ticket update if the sns is not open
+  // skip ticket update if
+  // - the sns is not open
+  // - the user is not sign in
+  // - user commitment information is not loaded
   $: if (
     lifecycle === SnsSwapLifecycle.Open &&
     isSignedIn($authStore.identity) &&
--- a/frontend/src/lib/components/project-detail/ParticipateButton.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateButton.svelte
@@ -24,7 +24,10 @@
   } from "$lib/services/sns-sale.services";
   import { isSignedIn } from "$lib/utils/auth.utils";
   import { authStore } from "$lib/stores/auth.store";
-  import { hasOpenTicketInProcess } from "$lib/utils/sns.utils";
+  import {
+    getCommitmentE8s,
+    hasOpenTicketInProcess,
+  } from "$lib/utils/sns.utils";
   import type { TicketStatus } from "$lib/types/sale";
   import type { SaleStep } from "$lib/types/sale";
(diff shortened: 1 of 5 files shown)


frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -52,6 +55,12 @@
     swapCommitment: $projectDetailStore.swapCommitment,
   });
 
+  let userCommitment: undefined | bigint;
+  $: userCommitment =
+    $projectDetailStore.swapCommitment === null
```
peterpeterparker, line 60: nth: `isNullish`

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -81,6 +90,8 @@
       return;
     }
 
+    assertNonNullish(userCommitment);
```
peterpeterparker, line 93: is this "just" to make TypeScript happy? if not, shouldn't we catch the potential error that can be thrown?

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -97,7 +109,8 @@
   // skip ticket update if the sns is not open
   $: if (
     lifecycle === SnsSwapLifecycle.Open &&
-    isSignedIn($authStore.identity)
+    isSignedIn($authStore.identity) &&
+    userCommitment !== undefined
```
peterpeterparker, line 113: nth: `nonNullish`

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -97,7 +109,8 @@
   // skip ticket update if the sns is not open
```
peterpeterparker, line 109: maybe extend comment too?

frontend/src/lib/components/project-detail/ParticipateButton.svelte
mstrasinskis: in the situation when there is no commitment 0n should be used. (null - not initialized, undefined - not found)

frontend/src/lib/components/project-detail/ParticipateButton.svelte
mstrasinskis: For the typescript good mood only. Would you prefer `userCommitment: userCommitment as bigint`?

frontend/src/lib/components/project-detail/ParticipateButton.svelte
peterpeterparker: ah makes sense. can you add a comment to avoid anyone switching to `isNullish` in the future?

we can also create a function but let's just keep it slim

frontend/src/lib/components/project-detail/ParticipateButton.svelte
mstrasinskis: Same as before, to use 0n if no commitment info.

frontend/src/lib/components/project-detail/ParticipateButton.svelte
peterpeterparker: If would prefer

```
if (isNullish(userCommitment) {
  // Typescript guard, user commitment cannot be undefined here
  return;
}
```

Or, try following?

```
`userCommitment: userCommitment!`
```

learning about`!` recently, not sure it's a best practice though

frontend/src/lib/components/project-detail/ParticipateButton.svelte
peterpeterparker: Same as before, comment would be welcomed

frontend/src/lib/components/project-detail/ParticipateButton.svelte
mstrasinskis: Good point, the current one is outdated.

frontend/src/lib/components/project-detail/ParticipateButton.svelte
mstrasinskis: TIL, thanks.
`userCommitment: userCommitment!` sounds reasonable here.

frontend/src/lib/components/project-detail/ParticipateButton.svelte
mstrasinskis: Sorry, `nonNullish` is good in this case, updated, thanks!

frontend/src/lib/components/project-detail/ParticipateButton.svelte
mstrasinskis: Comment was added 👍

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -16,15 +16,18 @@
   import Tooltip from "$lib/components/ui/Tooltip.svelte";
   import SignInGuard from "$lib/components/common/SignInGuard.svelte";
   import type { Principal } from "@dfinity/principal";
-  import { nonNullish } from "@dfinity/utils";
+  import { assertNonNullish, nonNullish } from "@dfinity/utils";
```
peterpeterparker, line 19: you can remove this import now

peterpeterparker (approved):
LGTM, thanks

PR #1984, opened 2023-03-01
Fast updates during a sale

bitdivine:
# Motivation
During a sale, the amount pledged by investors can change fairly quickly, so it behoves us to update the cache fairly frequently.

# Changes
* Add a "Fast" scheduler
  * The scheduler updates the swap data frequently while there is an open sale.
  * If there are no open sales but the date of a future sale is known, the scheduler sets a timer for the next sale and goes to sleep.
  * If there are no current or pending sales, the scheduler goes to sleep and relies on the normal updates to detect any new pending sales when they appear.

# Tests
* I have tested this manually on small09:
  * aggregator: https://tqtu6-byaaa-aaaaa-aaana-cai.small09.testnet.dfinity.network/
  * nns-dapp: https://tcvdh-niaaa-aaaaa-aaaoa-cai.small09.dfinity.network/project/?project=x2dwq-7aaaa-aaaaa-aaaxq-cai
  * Watch the logs.  If all SNSs are bought out, the fast updates cease.  If a new SNS is created, fast updates restart.


rs/sns_aggregator/src/fast_scheduler.rs
```diff
@@ -0,0 +1,244 @@
...
+        // Does the SNS have swap state?
+        let swap = swap_state.swap.as_ref().ok_or_else(|| "SNS has no swap.")?;
+        // Is a start scheduled?
+        let sale_start_seconds = swap
+            .decentralization_sale_open_timestamp_seconds
+            .ok_or_else(|| "SNS sale has no time.")?;
+        // We would want to start a bit before the sale.
+        let data_collection_start_seconds = sale_start_seconds - 10;
+        // Has the sale stage passed?
+        if swap.lifecycle > Self::LIFECYCLE_OPEN {
```
lmuntaner, line 190: lifecycle `ADOPTED` is 5, but the sale stake has not passed.

Does that mean that it will be reset anywhere else?

rs/sns_aggregator/src/fast_scheduler.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// Starts the update timer using the global state.
+    pub fn global_stop() {
+        STATE.with(|state| {
+            state.fast_scheduler.borrow_mut().stop();
+        });
+    }
+
+    /// Start collecting data now.
+    pub fn start(&mut self, timer_interval: Duration) {
+        let timer_id = set_timer_interval(timer_interval, || ic_cdk::spawn(Self::global_update_next()));
```
lmuntaner, line 141: I'm guessing this will call `global_update_next` at intervals?

lmuntaner (approved):
I'm not the best to review Rust code, hehe.

I added a couple of questions. Specially to understand whether you took into account the Adopted state as well.

But already approving, we need it merged this morning.

rs/sns_aggregator/src/fast_scheduler.rs
bitdivine: Am I correct in believing that with a 5, there will be no more purchases and, more generally, no frequent updates?  If so I _think_ that stopping frequent updates when an SNS reaches a 5 is correct, no?

rs/sns_aggregator/src/fast_scheduler.rs
bitdivine: Yes, correct.  As far as I can see there is no support for closures, so the function needs to be a global.

rs/sns_aggregator/src/fast_scheduler.rs
bitdivine: Updated.  Done.  Thank you!

PR #1989, opened 2023-03-01
GIX-1328 Test NeuronDetail with async project loading.

dskloetd:
# Motivation

A previous fix for GIX-1328 (https://github.com/dfinity/nns-dapp/pull/1926) broke sns neuron deep linking.
This PR adds a test that would have caught the breakage.
But it also introduces some new concepts to help with writing clean tests.

# Changes

1. Add `NeuronDetail.spec.ts` with a test to make sure `NeuronDetail` can be rendered while sns projects are still loading and this doesn't cause `NnsNeuronDetail` to be rendered instead of `SnsNeuronDetail`, as happened with https://github.com/dfinity/nns-dapp/pull/1926.
2. Introduce page objects for NeuronDetail, NnsNeuronDetail, SnsNeuronDetail,. and SkeletonCard.
3. Wrap components in a div (with `display: contents` not to affect rendering) with `data-tid` so page objects can refer to the whole component.
4. Add a util to block non-mocked API calls during a test to make sure all calls are expected and mocked during a test.

# Tests

The PR is about adding testing.
But I also manually clicked around to see that nns and sns neuron details still render correctly.
Also introduced the bug again to make sure the test catches it.


frontend/src/tests/lib/routes/NeuronDetail.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+const snsProps = {
+  neuronId: testSnsNeuronId,
+};
+
+const nnsProps = {
+  neuronId: testNnsNeuronId,
+};
+
+describe("NeuronDetail", () => {
+  blockAllCallsTo(blockedApiPaths);
```
lmuntaner, line 54: This relies on `jest.mock(...)` being set, is this correct?

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+      jest.mocked(snsGovernanceApi.refreshNeuron).mockResolvedValue(undefined);
+
+      page.mock({
+        data: { universe: testSnsCanisterId.toText() },
+        routeId: AppPath.Neuron,
+      });
+    });
+
+    it("should load", async () => {
+      await loadSnsProjects();
```
lmuntaner, line 108: This is what I meant. If we can render the `+page.svelte`, that will be done automatically.

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    });
+  });
+
+  describe("sns neuron", () => {
+    beforeEach(() => {
+      const [metadata, swapStates] = snsResponseFor({
+        principal: testSnsCanisterId,
+        lifecycle: SnsSwapLifecycle.Committed,
+      });
+      jest.mocked(snsApi.queryAllSnsMetadata).mockResolvedValue(metadata);
```
lmuntaner, line 86: Careful, when we use the SNS aggregator this is not called.

I guess that because the aggregator fails, the code fallsback here. But ideally we'll remove the fallback.

It makes more sense to mock the call to the `sns-aggregator.api`

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(root: Element) {
+    if (root.getAttribute("data-tid") !== tid) {
+      throw new Error(`${root} is not a NeuronDetail`);
+    }
+    this.root = root;
+  }
+
+  static under(element: HTMLElement): NeuronDetailPo | null {
+    const el = element.querySelector(`[data-tid=${tid}]`);
+    return el && new NeuronDetailPo(el);
```
lmuntaner, line 19: Does this pass our `npm check`?

I believe we don't allow the patter `XX && ...`. We want to strictly compare:

```
return nonNullish(el) ? new NeuronDetailPo(el) : null
```

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
```diff
@@ -0,0 +1,45 @@
+import { nonNullish } from "@dfinity/utils";
+import { NnsNeuronDetailPo } from "./NnsNeuronDetail.page-object";
+import { SnsNeuronDetailPo } from "./SnsNeuronDetail.page-object";
+
+const tid = "neuron-detail-component";
```
lmuntaner, line 5: How about if this is a static private property of the class?

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  hasSnsNeuronDetail(): boolean {
+    return nonNullish(this.getSnsNeuronDetail());
+  }
+
+  isContentLoaded() {
+    return (
+      this.getNnsNeuronDetail()?.isContentLoaded() ||
+      this.getSnsNeuronDetail()?.isContentLoaded() ||
+      false
```
lmuntaner, line 42: Do you need this `false`? If previous are `false` the whole is already `false`.

frontend/src/tests/page-objects/NnsNeuronDetail.page-object.ts
```diff
@@ -0,0 +1,27 @@
...
+    const el = element.querySelector(`[data-tid=${tid}]`);
+    return el && new NnsNeuronDetailPo(el);
+  }
+
+  getSkeletonCards(): SkeletonCardPo[] {
+    return SkeletonCardPo.allUnder(this.root);
+  }
+
+  isContentLoaded(): boolean {
+    return this.getSkeletonCards().length == 0;
```
lmuntaner, line 25: for our new rule: `===`, hehe

frontend/src/tests/page-objects/SkeletonCard.page-object.ts
```diff
@@ -0,0 +1,18 @@
+const tid = "skeleton-card";
```
lmuntaner, line 1: ditto

frontend/src/tests/page-objects/SnsNeuronDetail.page-object.ts
```diff
@@ -0,0 +1,27 @@
+import { SkeletonCardPo } from "./SkeletonCard.page-object";
+
+const tid = "sns-neuron-detail-component";
```
lmuntaner, line 3: ditto

lmuntaner (commented):
I added some comments.

In general, I love this approach.

Wish: "Render directly the page that triggers the same initialization as the user sees".

Questions:
* What granularity is the page object? You used a `SkeletonCard` for example, which is not a page.
* Could we define a level of granularity and test always from that level? For example, a component uses a button. We don't test the button functionality directly, but we test the component with the PO granularity, What do you think?

I really like where this is going.

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+ * @jest-environment jsdom
+ */
+
+import * as governanceApi from "$lib/api/governance.api";
+import * as snsGovernanceApi from "$lib/api/sns-governance.api";
+import * as snsLedgerApi from "$lib/api/sns-ledger.api";
+import * as snsApi from "$lib/api/sns.api";
+import { OWN_CANISTER_ID_TEXT } from "$lib/constants/canister-ids.constants";
+import { AppPath } from "$lib/constants/routes.constants";
+import NeuronDetail from "$lib/routes/NeuronDetail.svelte";
```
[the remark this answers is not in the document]
dskloetd: I'd prefer that as well but didn't know if it could work. I'll try and let you know if I run into problems.

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
dskloetd: Yes. It it will tell you if you didn't. See this line below
```
throw new Error(`You must add 'jest.mock("${modulePath}");' to your test.`);
```

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
dskloetd: I thought all featureFlags are enabled by default in tests. But in that case it should have already use the aggregator, no?

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
dskloetd: Passes, yes. I also used it in other places (I know at least 1 [already merged](https://github.com/dfinity/nns-dapp/blob/f1f97684ece3e98a026cb2384710c847b447f942/frontend/src/lib/utils/utils.ts#L298)). What would be the reason for not allowing it?

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
dskloetd: If the previous parts are both `null`, the result would also be `null`, instead of false.

dskloetd:
> Questions:
> 
> * What granularity is the page object? You used a `SkeletonCard` for example, which is not a page.

Every component! I guess they are objects on the page; not objects presenting pages. I didn't come up with the name. It's an existing design pattern.

> * Could we define a level of granularity and test always from that level? For example, a component uses a button. We don't test the button functionality directly, but we test the component with the PO granularity, What do you think?

I don't understand. But the idea is that you have your hierarchy of components on the page and an equivalent hierarchy of page objects. If you test a component you use the page object of that component **and every page object below it in the hierarchy**.

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
dskloetd: It was actually enabled but it was using the fallback after the aggregator call failed silently. Now I'm mocking the aggregator.

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
dskloetd: Let's do it in a separate PR.

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
dskloetd: What would be the benefit of that?

It means accessing it would have to be done as `NeuronDetailPo.tid` instead of `tid`. In addition to it being longer, I actually like having to change less boilerplate code when copying a page object to create a new page object.

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
dskloetd: I also couldn't find any such rule [here](https://denar90.github.io/eslint.github.io/docs/rules/) although I'm not sure what to search for other than "guard operator".

frontend/src/tests/page-objects/NnsNeuronDetail.page-object.ts
dskloetd: Done

frontend/src/tests/page-objects/SnsNeuronDetail.page-object.ts
```diff
@@ -0,0 +1,27 @@
...
+    const el = element.querySelector(`[data-tid=${tid}]`);
+    return el && new SnsNeuronDetailPo(el);
+  }
+
+  getSkeletonCards(): SkeletonCardPo[] {
+    return SkeletonCardPo.allUnder(this.root);
+  }
+
+  isContentLoaded(): boolean {
+    return this.getSkeletonCards().length == 0;
```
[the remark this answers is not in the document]
dskloetd: Done

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
lmuntaner: great, thanks!

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
lmuntaner: If it passes, good. I'm ok with the pattern

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
lmuntaner: Understood, thanks!

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
lmuntaner: But then the class is not a standalone component. It needs this extra variable, which is set in the file where it is used.

We'll need to remember that if we create a file with two page-objects, that each needs a different local variable.

dskloetd: I don't think we should create two page-objects in the same file. Yes, if we would, we'd need to use a different pattern. But we anyway always need to remember to change the value of this variable when we create a new page object. So it's not really an extra thing to remember.
So I guess it's a trade-off depending on whether we would ever have multiple page objects in the same file.
Do you think we would? I'll trust your experience with the project.

lmuntaner (approved):

PR #1995, opened 2023-03-01
fix: remove redundant error on TYPE_TICKET_EXISTS

mstrasinskis:
# Motivation

Fix wrong `TYPE_TICKET_EXISTS` handling.

# Changes

- return after notification on `TYPE_TICKET_EXISTS`

# Tests

- updated to not display an error

Untitled


lmuntaner (approved):
Thanks!

mstrasinskis:
@lmuntaner The z-index issue is not a bug it's by design of components. Let's put it outside of this pr.

PR #2000, opened 2023-03-01
Chore: Force query calls

lmuntaner:
# Motivation

Improve performance of the nns dapp by dropping update calls.

# Changes

* Add a new constant FORCE_CALL_STRATEGY
* Use the new constant in all the `queryAndUpdate` calls reviewed by security and accepted to be query calls.

# Tests

* No tests were broken. Tests checked that calls were made, but not the number of them.

### DIFF
--- a/frontend/src/lib/services/$public/proposals.services.ts
+++ b/frontend/src/lib/services/$public/proposals.services.ts
@@ -43,7 +43,11 @@ const handleFindProposalsError = ({
   identity: Identity;
 }) => {
   console.error(err);
-  if (certified || identity.getPrincipal().isAnonymous()) {
+  if (
+    certified ||
+    identity.getPrincipal().isAnonymous() ||
+    FORCE_CALL_STRATEGY === "query"
+  ) {
     proposalsStore.setProposals({ proposals: [], certified });
 
     toastsError({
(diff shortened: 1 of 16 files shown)


lmuntaner:

frontend/src/lib/services/accounts.services.ts
```diff
@@ -158,6 +159,7 @@ export const syncAccounts = (
       errorHandler({ err, certified });
     },
     logMessage: "Syncing Accounts",
+    strategy: FORCE_CALL_STRATEGY,
```
mstrasinskis, line 162: Why not make it optional parameter?

mstrasinskis: Got it!

frontend/src/lib/services/$public/proposals.services.ts
```diff
@@ -169,6 +170,7 @@ const findProposals = async ({
   let uncertifiedProposals: ProposalInfo[] | undefined;
 
   return queryAndUpdate<ProposalInfo[], unknown>({
+    strategy: FORCE_CALL_STRATEGY,
```
mstrasinskis, line 173: Not a strong opinion, but would be more good to also tweak the error handlers in every place the flag is used e.g. `if (certified || identity.getPrincipal().isAnonymous()) {...`

lmuntaner: Do you mean to show errors on query in that case?

Makes sense

lmuntaner:
@mstrasinskis ready for another review

mstrasinskis (approved):
LGTM, thank you!

lmuntaner:
This does NOT yet move the wrapper usage to not certified.

I will do that in another PR.

PR #2001, opened 2023-03-01
Minimum viable aggregator release template

bitdivine:
# Motivation
The aggregator is to be upgraded by proposal, like the NNS.  We need at least a minimal release Standard Operating Procedure.

See also the corresponding NOTION page: https://www.notion.so/SNS-Aggregator-release-SOP-d6cd1b14eb6b466cb903da6d3e76e254

# Changes
* Make a proposal template based on the nns-dapp template.

# Tests
* I have run the scripts and eyeballed the templates & command.

### DIFF
--- a/scripts/sns/aggregator/release
+++ b/scripts/sns/aggregator/release
@@ -2,14 +2,20 @@
 set -euo pipefail
 cd "$(dirname "$0")/../../.."
 
-read -rp "HSM Pin: " DFX_HSM_PIN
-
 NEURON_PATH="$HOME/.config/dfx/prod-neuron"
 if test -e "$NEURON_PATH"; then
   NEURON="$(cat "$NEURON_PATH")"
+  echo "Neuron ID: $NEURON (from $NEURON_PATH)"
 else
-  read -rp "Neuron ID: " NEURON
+  read -rp "Please enter your Neuron ID: " NEURON
+  read -rp "... Store the Neuron ID for future use? (y/N)" STORE_NEURON
+  [[ "$STORE_NEURON" != [yY] ]] || {
+    mkdir -p "$(dirname "$NEURON_PATH")"
+    printf "%s" "$NEURON" >"$NEURON_PATH"
+  }
 fi
+
+read -rp "HSM Pin: " DFX_HSM_PIN
 
 AGGREGATOR_CANISTER_ID="3r4gx-wqaaa-aaaaq-aaaia-cai"
 WASM="./release/docker/sns_aggregator.wasm"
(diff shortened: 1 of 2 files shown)


scripts/sns/aggregator/release
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+set -euo pipefail
+cd "$(dirname "$0")/../../.."
+
+read -rp "HSM Pin: " DFX_HSM_PIN
+
+NEURON_PATH="$HOME/.config/dfx/prod-neuron"
```
lmuntaner, line 7: Where and how is this set?

By dfx?

scripts/sns/aggregator/release
```diff
@@ -0,0 +1,29 @@
...
+AGGREGATOR_CANISTER_ID="3r4gx-wqaaa-aaaaq-aaaia-cai"
+WASM="./release/docker/sns_aggregator.wasm"
+SHA="$(sha256sum <"$WASM" | awk '{print $1}')"
+
+set ic-admin --pin "$DFX_HSM_PIN" --nns-url https://ic0.app --use-hsm --key-id 01 --slot 0 propose-to-change-nns-canister --proposer "$NEURON" --canister-id "$AGGREGATOR_CANISTER_ID" --mode upgrade --wasm-module-path "$WASM" --summary-file ./release/AGGREGATOR_PROPOSAL.md --wasm-module-sha256 "$SHA"
+
+echo
+echo PLEASE REVIEW THIS COMMAND:
+echo
+echo "${@}"
```
lmuntaner, line 23: Will this print the `ic-admin` command above?

lmuntaner (approved):
A couple of questions, but already approving. Thanks!

scripts/sns/aggregator/release
bitdivine: No, you have to add it yourself.  Different team members have different neuron IDs.

bitdivine: It might make sense to print a notice so that users know that they can put a neuron ID in that file.

scripts/sns/aggregator/release
bitdivine: Yes, and without the echo it will execute it below.

scripts/sns/aggregator/release
bitdivine: Done:
* If using a stored neuron, the script now prints where the neuron ID came from.
*  If no neuron is stored, the user is asked whether they would like to store the neuron ID for future use.

scripts/sns/aggregator/release
lmuntaner: ok, thanks for the explanation!

scripts/sns/aggregator/release
lmuntaner: great, thanks!

PR #2004, opened 2023-03-02
Feat: Do not poll twice

lmuntaner:
# Motivation

Make sure the polling doesn't repeat and it lasts forever.

New functionality:
* Poll won't poll twice
* Cancel pollAccounts when navigating away.

# Changes

* Add a `pollId` parameter to the `poll` util.
* Add functionality to cancel poll by `pollId`.
* Don't poll twice with the same `pollId`.
* Use the new `pollId` in `pollAccounts`
* Add a service `cancelPollAccounts`.
* Use the new service `cancelPollAccounts` when the component that triggered the `pollAccounts` is destroyed.

# Tests

* Test new functionality in the `poll` util.
* Test `cancelPollAccounts`
* Test in each page where `pollAccounts` is started, that it's cancelled when component is unmounted.

### DIFF
--- a/frontend/src/tests/lib/modals/neurons/DisburseNnsNeuronModal.spec.ts
+++ b/frontend/src/tests/lib/modals/neurons/DisburseNnsNeuronModal.spec.ts
@@ -4,9 +4,11 @@
 
 import * as ledgerApi from "$lib/api/ledger.api";
 import * as nnsDappApi from "$lib/api/nns-dapp.api";
+import { SYNC_ACCOUNTS_RETRY_SECONDS } from "$lib/constants/accounts.constants";
 import { AppPath } from "$lib/constants/routes.constants";
 import { pageStore } from "$lib/derived/page.derived";
 import DisburseNnsNeuronModal from "$lib/modals/neurons/DisburseNnsNeuronModal.svelte";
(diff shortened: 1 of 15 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/modals/neurons/DisburseNnsNeuronModal.spec.ts
```diff
@@ -32,6 +34,10 @@ jest.mock("$lib/services/neurons.services", () => {
 });
 
 describe("DisburseNnsNeuronModal", () => {
+  beforeEach(() => {
+    cancelPollAccounts();
```
dskloetd, line 38: Doesn't this cause errors in already passed tests? I wonder if it could get confusing at some point.

dskloetd (approved):
I'll follow up with a PR to remove all the await nulls.

frontend/src/tests/lib/modals/neurons/DisburseNnsNeuronModal.spec.ts
lmuntaner: Why?

I needed it when I added the last test.

Then I moved it up because I thought it made sense. All were passing without it, but they were just not testing that part.

PR #2007, opened 2023-03-02
Clean up polling tests

dskloetd:
# Motivation

https://github.com/dfinity/nns-dapp/pull/2004 added a number of `await null` to tests which use polling to make tests work without it being fulling clear what's going on.
This PR replaces all the `await null` with `advanceTime` which does not need to be called an arbitrary number of times.
I also made additional cleanups to make it clearer what's really going on in the tests.

# Changes

1. Add a test util to advance fake timers while also running all already resolved promise in order to get the full affect of advancing time.
2. Use the test util to allow removing all the `await null` statements.
3. Reset mocks to avoid one failure resulting in additional failure because of left over mock state.
4. Don't wrap expectations in `waitFor` when not necessary.
5. Instead of making additional rounds in a loop, to additional time advancement outside the loop.
4. Rename `retriesBefore...` to `callsBefore...`. "Retries" is confusing because you could think that a retry only happens after the first try so it could be 1 less than the number of calls.
3. In `sns-sale.services.spec.ts` rename `sendICPSpy` to `spyOnSendICP` for consistency with the other spies in the same file.

# Tests

Tests only


frontend/src/tests/utils/timers.test-utils.ts
```diff
@@ -0,0 +1,20 @@
+import { nonNullish } from "@dfinity/utils";
+
+const originalTimeout = setTimeout;
+
+export const runResolvedPromises = () =>
```
lmuntaner, line 5: Shall we add a comment on why and how this is used?

Otherwise it seems hard to remember what this is used for.

frontend/src/tests/utils/timers.test-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  new Promise((resolve) => originalTimeout(resolve, 0));
+
+// If `millis` is passed, advance time by that much.
+// Otherwise advance time enough to run all current timers.
+// Execute all resolved promises before and after.
+export const advanceTime = async (millis?: number): Promise<void> => {
+  // Make sure the timers are set before we advance time.
```
lmuntaner, line 12: Not for this PR, could we check this?

Maybe raise an error if timers are not mocked

lmuntaner (approved):
Approved, but added a couple of comments.

frontend/src/tests/utils/timers.test-utils.ts
dskloetd: Done

frontend/src/tests/utils/timers.test-utils.ts
dskloetd: I tried that but jest doesn't expose that information. But it does do its own check and outputs this to the console:
```
    console.warn
      A function to advance timers was called but the timers APIs are 
not replaced with fake timers. Call `jest.useFakeTimers()` in this 
test file or enable fake timers for all tests by setting 'fakeTimers': 
{'enableGlobally': true} in Jest configuration file.
      Stack Trace:
```

But that's actually not what I meant by this comment.
What I meant is that we call `await runResolvedPromises();` to make sure that our code under test calls `setTimeout` (rather than blocked on awaits) before we call `jest.runOnlyPendingTimers()` or `jest.advanceTimersByTime(millis);`.

lmuntaner: aaahh, ok, got it. Thanks for the explanation

frontend/src/tests/utils/timers.test-utils.ts
lmuntaner: Thanks

PR #2009, opened 2023-03-02
feat: cleanup before-unload and add guard to participate to sale

peterpeterparker:
# Motivation

⚠️ DO NOT MERGE.

Not sure it is a good idea, to some extension yes but to some the browser freeze while displaying the popup so not sure what happens if user transfer icp, freeze browser for 5min, and continue during a sale.

That said, the PR also do a bit of clean-up, we might continue with that later.

# Changes

# Tests


peterpeterparker:
Rename in #2771

peterpeterparker:
I guess we will not pursuie this therefore I close it.

PR #2013, opened 2023-03-02
Fix poll() when polling again after success

dskloetd:
# Motivation

Currently we only remove the pollId from currentPolls when it's canceled. But that means you can't poll again after success.
We also only hide the "We'll keep retrying" toast on success.

# Changes

1. Remove pollId from currentPolls in finally block.
2. Hide the toast in the finally block.
3. Add more test coverage.

# Tests

Several extra unit tests.


lmuntaner (approved):
Thanks!

PR #2015, opened 2023-03-02
fix: warning high-load toasts mumbo jumbo

peterpeterparker:
# Motivation

On testnet:

- fetching tvl failed so per extension all metrics were ignored (`Promise.all` and error that bubbles) . Since tvl is only fetched the first time, only the first load warning message was ignored and that's why further message (1min later) were displayeed

- on sign-in, all toasts were deleted including the warning toast therefore the warning message disappeared

- the version deployed did not behave on the ic js / svelte-wise as smooth as locally therefore the information was displayed multiple times when user navigated and refreshed screen

# Changes

In addition to above, this PR also set the correct "theme" for  the toast so that it appears "inverted" color wise.


frontend/src/lib/components/metrics/TransactionRateWarning.svelte
```diff
@@ -29,20 +31,21 @@
...
       position: "top",
       title: $i18n.metrics.nns_high_load,
       truncate: true,
       icon: TransactionRateWarningIcon,
+      theme: "inverted",
     });
   };
 
-  $: transactionRate, (() => transactionRateWarning())();
+  $: $metricsStore, (() => transactionRateWarning())();
```
peterpeterparker, line 48: I've changed this while debugging, not sure its needed but since it ultimately works out lets go with it

lmuntaner (approved):
LGTM

PR #2017, opened 2023-03-03
feat: toast resolve todo and bump gix-cmp

peterpeterparker:
# Motivation

Resolve yesterday evening todo (toast.reset -> toest.clean) and bump gix-cmp.

# Note

Because there were breaking changes in SvelteKit (adapted in gix-cmp in https://github.com/dfinity/gix-components/pull/166) this PR had to also adapt.

- gix-cmp is not flat but contains a sub-folder `dist` so Sass import and use needs to reference this sub-folder
- jest had a hard time finding the library so it needed an extra mapping configuration to find the components

### DIFF
--- a/frontend/src/lib/stores/toasts.store.ts
+++ b/frontend/src/lib/stores/toasts.store.ts
@@ -5,7 +5,6 @@ import { errorToString } from "$lib/utils/error.utils";
 import type { I18nSubstitutions } from "$lib/utils/i18n.utils";
 import { replacePlaceholders, translate } from "$lib/utils/i18n.utils";
 import { toastsStore } from "@dfinity/gix-components";
-import { get } from "svelte/store";
 
 const mapToastText = ({ labelKey, substitutions, detail }: ToastMsg): string =>
   `${replacePlaceholders(translate({ labelKey }), substitutions ?? {})}${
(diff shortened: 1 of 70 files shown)


frontend/src/lib/stores/toasts.store.ts
```diff
@@ -66,15 +65,10 @@ export const toastsError = ({
...
-  const customToasts = toasts.filter(({ level }) => level === "custom");
-
-  toastsStore.reset();
-
-  customToasts.forEach((msg) => toastsStore.show(msg));
-};
+/**
+ * We keep error and custom (used for particular notifications such as high-load messages).
+ */
+export const toastsClean = () => toastsStore.reset(["success", "warn", "info"]);
```
lmuntaner, line 71: Is there a way to ensure that ALL levels are passed?

Otherwise this might give weird bugs if we add one more level.

lmuntaner (commented):
One question / comment before approving

frontend/src/lib/stores/toasts.store.ts
peterpeterparker: I can add a test. There were none before so spared the hustle to add one but, again, can add one. Will do.

peterpeterparker: Just gonna repeat the test in the gix-cmp, for the rest it's probably more a e2e thing or less verbose to test with e2e

peterpeterparker: Add a test for `toastClean` that checks the number of toasts remaining in store after call.

lmuntaner: I meant more even a type thing somehow.

But thanks for the tests!

lmuntaner (approved):

frontend/src/lib/stores/toasts.store.ts
peterpeterparker: Oh I see, did not got it. Not sure because it's a type so would have to search. If you got an idea ping me!

peterpeterparker: Thanks for the review

PR #2021, opened 2023-03-03
Cleanup watchSnsTotalCommitment test

dskloetd:
# Motivation

https://github.com/dfinity/nns-dapp/pull/2007 cleaned up a number of tests in a certain way to remove `await null` calls.
In parallel a new similar test was added that does `await null`.
This PR cleans up that test in the same way.

# Changes

Clean up `src/tests/lib/services/sns.services.spec.ts` to remove `await null` and follow the same style as in https://github.com/dfinity/nns-dapp/pull/2007

# Tests

Test only


lmuntaner (approved):
Thanks!

PR #2022, opened 2023-03-03
Test that sns sale uses 'high load' toast after 6 failures.

dskloetd:
# Motivation

Additional test coverage for https://github.com/dfinity/nns-dapp/pull/1977 ("high load" toast after 6 retries).
The new functionality was already tested in that PR but this PR also tests that the functionality is enabled in the payment flow.

# Changes

4 new tests in `sns-sale.services.spec.ts` checking that the "high load" toast in enabled in each of the steps.

# Tests

 test only


lmuntaner (approved):
Thanks!

PR #2024, opened 2023-03-06
Remove unused fields from MultiPartTransactionsProcessor

dskloetd:
# Motivation

MultiPartTransactionsProcessor has 2 unused fields: `statuses` and `errors`.
We want to remove these fields but this has to be done carefully because there are part of stable state which is encode/decoded when the canister is upgraded.

Candid::from_bytes actually deals fine with additional data for removed fields, so removing the fields would be fine, but we want to be able to roll back to an earlier version of the canister, and in that case we would be adding the fields back.
And trying to decode bytes which do not include new fields does cause the decoding to fail.

My solution is to temporarily fork MultiPartTransactionsProcessor to have 2 versions: 1 with the unused fields removed and 1 with the unused fields intact.
We use the type with the extra fields for encoding and we use the type without the extra fields for decoding.
Let's assume we have the following 3 versions of the canister:
1. Built from current main branch with the unused fields still present.
2. Built after this PR is merged.
3. Built from current main branch with the unused fields removed and no additional code (so anything special from the PR removed again).

We can now safely go back and forth between versions 1 and 2 and we can go safely back and forth between versions 2 and 3. The only thing that fails is going straight from version 3 to version 1.

So if we stay with version 2 for one or two weeks, we should be safe to always have a rollback path to a good version.

# Changes

1. Forked MultiPartTransactionsProcessor to MultiPartTransactionsProcessorWithRemovedFields.
2. Removed `statuses` and `errors` from MultiPartTransactionsProcessor but kept them in MultiPartTransactionsProcessorWithRemovedFields.
3. When encoding AccountsStore to stable state, use MultiPartTransactionsProcessorWithRemovedFields instead of MultiPartTransactionsProcessor.
4. A test that shows that encoded MultiPartTransactionsProcessorWithRemovedFields can be decoded as MultiPartTransactionsProcessor.
5. A test that shows that encoded MultiPartTransactionsProcessor can not be decoded as MultiPartTransactionsProcessorWithRemovedFields.
6. A test that shows that AccountsStore can be encoded and decoded with the remaining field of MultiPartTransactionsProcessor (`queue`) intact.

# Tests

1. Added unit tests.
2. Deployed (on local replica) version 3 (described above) and then deployed version 1 to notice the failure.
3. Deployed version 3 and then deployed version 2 and then version 1 to notice that it works.
4. Deployed version 1 and then 2 and then 3 to make sure that works as well.

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -1,6 +1,8 @@
 use crate::constants::{MEMO_CREATE_CANISTER, MEMO_TOP_UP_CANISTER};
 use crate::metrics_encoder::MetricsEncoder;
-use crate::multi_part_transactions_processor::{MultiPartTransactionToBeProcessed, MultiPartTransactionsProcessor};
+use crate::multi_part_transactions_processor::{
+    MultiPartTransactionToBeProcessed, MultiPartTransactionsProcessor, MultiPartTransactionsProcessorWithRemovedFields,
+};
 use crate::state::StableState;
 use crate::time::time_millis;
 use crate::STATE;
@@ -1288,6 +1290,10 @@ impl AccountsStore {
 
 impl StableState for AccountsStore {
     fn encode(&self) -> Vec<u8> {
+        // Encode with removed fields intact to make rolling back to
+        // a version which still has the removed fields safe.
+        let mptp_with_removed_fields =
+            MultiPartTransactionsProcessorWithRemovedFields::from(&self.multi_part_transactions_processor);
         Candid((
             &self.accounts,
             &self.hardware_wallets_and_sub_accounts,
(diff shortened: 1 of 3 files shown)


rs/backend/src/accounts_store.rs
```diff
@@ -1288,14 +1290,18 @@ impl AccountsStore {
 
 impl StableState for AccountsStore {
     fn encode(&self) -> Vec<u8> {
+        // Encode with removed fields intact to make rolling back to
+        // a version which still has the removed fields safe.
+        let mptp_with_removed_fields =
+            MultiPartTransactionsProcessorWithRemovedFields::from(&self.multi_part_transactions_processor);
```
lmuntaner, line 1296: Is the plan to remove this code after the next release?

We will then be able to encode without the fields because decoding will also be without them. Is this correct?

lmuntaner (commented):
LGTM, just a question before approving.

Thanks for the extra tests!

rs/backend/src/accounts_store.rs
dskloetd: Yes, or after 1 or 2 weeks or something. Whenever we are very sure we won't go back to older code.

lmuntaner: Ok, sounds good to me!

lmuntaner (approved):

PR #2035, opened 2023-03-07
Add $tests as import root for "./src/tests".

dskloetd:
# Motivation

Currently we use a variable number of "../" to refer to "src/tests" but it's hard to tell if the number of "../" actually refers to "src/tests" or some deeper directory.

# Changes

Introduce $tests to refer to "src/tests" and use it in all `.ts` files under src/tests.
Apart from the changes in `frontend/jest.config.cjs` and `frontend/tsconfig.spec.json`, the changes were made mechanically as follows:
```
1. vim `grep -l '"../[^.]' frontend/src/tests/*/*.ts`
2. In vim do `:%s/from "\zs\.\.\ze\/[^.]/$tests` in every buffer.
3. vim `grep -l '"../../[^.]' frontend/src/tests/*/*/*.ts`
4. In vim do `:%s/from "\zs\.\.\/\.\.\ze\/[^.]/$tests` in every buffer.
5. vim `grep -l '"../../../[^.]' frontend/src/tests/*/*/*/*.ts`
6. In vim do `:%s/from "\zs\.\.\/\.\.\/\.\.\ze\/[^.]/$tests` in every buffer.
7. vim `grep -l '"../../../../[^.]' frontend/src/tests/*/*/*/*/*.ts`
8. In vim do `:%s/from "\zs\.\.\/\.\.\/\.\.\/\.\.\ze\/[^.]/$tests` in every buffer.
```

# Tests

`npm run test`


frontend/jest.config.cjs
```diff
@@ -22,9 +22,10 @@ module.exports = {
     "<rootDir>/node_modules/(?!(@dfinity/gix-components))",
   ],
   moduleNameMapper: {
-    "^\\$lib(.*)$": "<rootDir>/src/lib$1",
-    "^\\$routes(.*)$": "<rootDir>/src/routes$1",
-    "^\\$mocks(.*)$": "<rootDir>/__mocks__$1",
+    "^\\$lib/(.*)$": "<rootDir>/src/lib/$1",
+    "^\\$routes/(.*)$": "<rootDir>/src/routes/$1",
```
lmuntaner, line 26: Is the change needed? Why was it working before?

lmuntaner (approved):
One question, but already approving.

frontend/jest.config.cjs
dskloetd: I didn't try but it looks like before you could have written "$routesfoo/" and it would have pointed to "<rootDir>/src/routesfoo/". Now it requires the slash. Seems less brittle/confusing. And since it still worked I left it.
Sorry for not mentioning it in the PR description.

lmuntaner: Ah, ok, that makes sense!

PR #2040, opened 2023-03-07
Chore: Add flags dfx json

lmuntaner:
# Motivation

Set features flags of testnets at last as restrictive as mainnet, but not less.

Except the environments used for sns-testing `https___<network>_testnet_dfinity_network`, which require ckbtc to be disabled by default.

# Changes

* Changes in the flags in dfx.json

# Tests

Only config changes

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -191,8 +191,8 @@
         "FEATURE_FLAGS": {
           "ENABLE_SNS_2": true,
           "ENABLE_SNS_VOTING": true,
-          "ENABLE_SNS_AGGREGATOR": false,
-          "ENABLE_CKBTC": false,
+          "ENABLE_SNS_AGGREGATOR": true,
+          "ENABLE_CKBTC": true,
           "ENABLE_CKTESTBTC": false
         }
       },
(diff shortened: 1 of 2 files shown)


dfx.json
```diff
@@ -900,7 +966,10 @@
         "HOST": "https://small15.testnet.dfinity.network",
         "FEATURE_FLAGS": {
           "ENABLE_SNS_2": true,
-          "ENABLE_SNS_VOTING": true
+          "ENABLE_SNS_VOTING": true,
+          "ENABLE_SNS_AGGREGATOR": true,
+          "ENABLE_CKBTC": true,
+          "ENABLE_CKTESTBTC": false
```
dskloetd, line 972: Do we really need so much repetition in the dfx.json? Can we have default values and just specify when it's not the default value? Maybe even have the default values in the dfx.json as well?

And is it possible to have comments in json? Would be good to comment on the non-default ones explaining why they are like that.

lmuntaner: I agree that we could improve this.

Maybe we can add a default config of the flags when it's not present?

I can try that in another PR. At least for the feature flags.

lmuntaner: FYI: https://dfinity.atlassian.net/browse/GIX-1393

PR #2041, opened 2023-03-07
Debug commands to enable feature flags.

dskloetd:
# Motivation

We want to be able to enable feature flag on mobile where we can't open the JS console.

https://user-images.githubusercontent.com/122978264/223469577-0c34374d-ed90-40cf-914f-a0d3c77e0552.mov

# Changes

In the 6-click debug directive add 3 new commands:
1. "fft" for "feature-flag true" to override a feature flag to true.
2. "fff" for "feature-flag false" to override a feature flag to false.
3. "ffr" to remove a feature flag override.

# Tests

Tested manually on testnet.

### DIFF
--- a/frontend/src/lib/directives/debug.directives.ts
+++ b/frontend/src/lib/directives/debug.directives.ts
@@ -339,7 +339,7 @@ const promptFeatureFlag = (question: string): FeatureKey | null => {
     question,
     ...editableNonTestFlags.map((flag, index) => index + 1 + ": " + flag),
   ].join("\n");
-  const choice: number = prompt(message);
+  const choice: number = +prompt(message);
   if (choice >= 1 && choice <= editableNonTestFlags.length) {
     return editableNonTestFlags[choice - 1];
   }
--- a/frontend/src/lib/directives/debug.directives.ts
+++ b/frontend/src/lib/directives/debug.directives.ts
@@ -100,21 +100,21 @@ export function triggerDebugReport(node: HTMLElement) {
         }
 
         if (LogType.FeatureFlagsOverrideTrue === logType) {
-          const flag: FeatureKey = promptFeatureFlag(
+          const flag = promptFeatureFlag(
             get(i18n).feature_flags_prompt.override_true
           );
           flag && overrideFeatureFlagsStore.setFlag(flag, true);
         }
 
         if (LogType.FeatureFlagsOverrideFalse === logType) {
-          const flag: FeatureKey = promptFeatureFlag(
+          const flag = promptFeatureFlag(
             get(i18n).feature_flags_prompt.override_false
           );
           flag && overrideFeatureFlagsStore.setFlag(flag, false);
         }
 
         if (LogType.FeatureFlagsOverrideRemove === logType) {
-          const flag: FeatureKey = promptFeatureFlag(
+          const flag = promptFeatureFlag(
             get(i18n).feature_flags_prompt.remove_override
           );
           flag && overrideFeatureFlagsStore.removeFlag(flag);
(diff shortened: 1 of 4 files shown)


frontend/src/lib/directives/debug.directives.ts
```diff
@@ -91,6 +99,27 @@ export function triggerDebugReport(node: HTMLElement) {
           return;
         }
 
+        if (LogType.FeatureFlagsOverrideTrue === logType) {
+          const flag = promptFeatureFlag(
+            get(i18n).feature_flags_prompt.override_true
+          );
+          flag && overrideFeatureFlagsStore.setFlag(flag, true);
```
peterpeterparker, line 106: shouldn't you `return` after this line?

peterpeterparker: (ditto for other blocks)

frontend/src/lib/directives/debug.directives.ts
```diff
@@ -301,3 +330,20 @@ const generateDebugLog = async (logType: LogType) => {
...
   }
 };
+
+const promptFeatureFlag = (question: string): FeatureKey | null => {
+  const editableNonTestFlags = EDITABLE_FEATURE_FLAGS.filter(
+    (f) => !f.startsWith("TEST_")
+  );
+  const message = [
+    question,
+    ...editableNonTestFlags.map((flag, index) => index + 1 + ": " + flag),
```
peterpeterparker, line 340: nth: spread operator are a bit more readable - particularly in VS code or Webstorm - when it comes to formatting string while of course this is probably a bit an opiniated comment

```
...editableNonTestFlags.map((flag, index) => `${index + 1}: ${flag}`)

File: frontend/src/lib/directives/debug.directives.ts [hunk 3]
```diff
@@ -301,3 +330,20 @@ const generateDebugLog = async (logType: LogType) => {
     logWithTimestamp(stringifiedState);
   }
 };
+
+const promptFeatureFlag = (question: string): FeatureKey | null => {
+  const editableNonTestFlags = EDITABLE_FEATURE_FLAGS.filter(
+    (f) => !f.startsWith("TEST_")
```
peterpeterparker, line 336: nth: when I look at the all code base we rarely shorten variable names, we are mostly explicit. in this case we would name `f` something like `flag`

peterpeterparker (commented):
Looks great. In the video only two flags are displayed, that's just a demo thing, irl there will be more flags including ckBTC right?

frontend/src/lib/directives/debug.directives.ts
peterpeterparker: not "spread operator" 🤦‍♂️, brain freeze but I guess you get what I mean

dskloetd:
> Looks great. In the video only two flags are displayed, that's just a demo thing, irl there will be more flags including ckBTC right?

No, only [editable flags](https://github.com/dfinity/nns-dapp/blob/95df440a6776ac3be58202c7f351c02c77be25cf/frontend/src/lib/stores/feature-flags.store.ts#L32). CKBTC is not in editable flags, but I see that ENABLE_CKTESTBTC was added after I made the video.

frontend/src/lib/directives/debug.directives.ts
dskloetd: Yes! done

frontend/src/lib/directives/debug.directives.ts
dskloetd: Was gonna say did you mean template literals because I'm already using spread operator? Done.

peterpeterparker (approved):
LGTM, thanks for changes and answer 👍

PR #2042, opened 2023-03-07
Fix warning in CkBTCTransactionsList.spec.ts

dskloetd:
# Motivation

Missing parameters caused console.warn logs in the test:
```
    console.warn
      <CkBTCTransactionsList> was created with unknown prop 'canisters'
...
   console.warn
      <CkBTCTransactionsList> was created without expected prop 'indexCanisterId'
```

# Changes

Remove canisters param and add indexCanisterId param.

# Tests

`npm run test`


peterpeterparker:
Noticed it too and solved it in https://github.com/dfinity/nns-dapp/pull/2030 which is merged in main now so you can probably close this.

PR #2044, opened 2023-03-07
GIX-1381: Add total tokens supply to project

lmuntaner:
# Motivation

Add the total tokens supply to the SNS Aggregator json file.

# Changes

* Add a call to `icrc1_total_supply` in the upstream `get_sns_data`.
* Add property `icrc1_total_supply` in the Upstream struct.
* Add dependency `num-traits` to convert `Nat` to `u64`.
* Add `icrc1_total_supply` in the conversion from Upstream to SlowData. Convert to `u64` there.

# Tests

SNS Aggregator still doesn't have tests.


lmuntaner:
Deployed to medium06:

https://s24we-diaaa-aaaaa-aaaka-cai.raw.medium06.testnet.dfinity.network/v1/sns/list/latest/slow.json

rs/sns_aggregator/src/upstream.rs
```diff
@@ -127,6 +127,12 @@ async fn get_sns_data(index: u64, sns_canister_ids: DeployedSns) -> anyhow::Resu
         .map(|response: (_,)| response.0)
         .map_err(|err| anyhow!("Failed to get ledger fee: {err:?}"))?;
 
+    let icrc1_total_supply: SnsTokens = ic_cdk::api::call::call(ledger_canister_id, "icrc1_total_supply", ((),))
```
dskloetd, line 130: Does this make an actual call to another canister?
Does it need to be a whole separate call or can it piggyback on an existing call?

lmuntaner: Do you mean to do it in parallel?

The data is only available in this endpoint. We can't get it from any other from the calls.

rs/sns_aggregator/src/types/slow.rs
```diff
@@ -41,6 +44,11 @@ impl From<&UpstreamData> for SlowSnsData {
             swap_state: SlowSwapState::from(&upstream.swap_state),
             icrc1_metadata: upstream.icrc1_metadata.clone(),
             icrc1_fee: upstream.icrc1_fee.clone(),
+            icrc1_total_supply: upstream
+                .icrc1_total_supply
+                .0
+                .to_u64()
+                .expect("Error decoding total supply"),
```
peterpeterparker, line 51: rust noob question: the error bubbles properly or ends in some sort of trap?

rs/sns_aggregator/src/types/slow.rs
```diff
@@ -27,7 +27,10 @@ pub struct SlowSnsData {
     /// Ledger metadata.  The ledger keeps track of who owns how many tokens.
     pub icrc1_metadata: Vec<(String, Icrc1Value)>,
     /// The ledger fee, presumably a transaction fee.
+    /// TODO: Convert to u64 in v2 of the SNS aggregator
     pub icrc1_fee: Nat,
+    /// The ledger total tokens supply
+    pub icrc1_total_supply: u64,
```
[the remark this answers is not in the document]
lmuntaner: Good point!

I believe that the state would be dropped because it's not compatible.

I'll ask @bitdivine when he's back

rs/sns_aggregator/src/types/slow.rs
lmuntaner: I believe it bubbles. Maybe I should set a default value then?

rs/sns_aggregator/src/types/slow.rs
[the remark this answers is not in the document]
peterpeterparker: As long as the canister gets not trapped 😅.

rs/sns_aggregator/src/types/slow.rs
peterpeterparker: Good question, not sure. If we set `zero` then it's also a bit suprising to get zero as a value on the other side, if we bubble the error, then we might skip all data if one fails?

I don't have enough knowledge here.

Spontaneous though, do you need the `to_u64`? `.icrc1_total_supply.0` is already a `u128` or something no?

But don't mind me, once again I'm out context.

dskloetd: I'm also a noob, but I believe it can only bubble if your function returns Result<> and you use `?`.

rs/sns_aggregator/src/upstream.rs
dskloetd: Yeah, so we wouldn't block this change on that. But still worth asking if they can offer batching in the future, right? We don't want to do 10 cross canister calls if we can do 1, right?

rs/sns_aggregator/src/types/slow.rs
peterpeterparker: Indeed googled it, look like `except`= `unwrap_or_panic_with_this_message`

https://stackoverflow.com/questions/66362625/why-is-rusts-expect-called-expect

lmuntaner: The value is a candid `Nat` which converted to JSON was actually not very nice when it was a high number.

Instead of getting 515000000, I got an array of `[319231312, 119]`, or something like that. Weird.

That's why I converted it to integer.

rs/sns_aggregator/src/upstream.rs
lmuntaner: Correct, makes sense to me.

It's all a matter of priority.

rs/sns_aggregator/src/types/slow.rs
dskloetd: Javascript doesn't have 64 bit integers. It puts everything in double precision floats, which only hold exact integer values up to 53 bits. Are you certain this value never uses more than 53 bits?

lmuntaner: No, this value can use up to 64 bits, but not more.

In the candid agent-js, this is transformed to a bigint.

But here, I have just JSON. Shall I use better a string representation then?

peterpeterparker: > No, this value can use up to 64 bits, but not more.

Because of serde? 

Anyhow feel like I add more noise here than valuable input. Nevermind my comment, all clear.
Thanks for the answers!!!

rs/sns_aggregator/src/types/slow.rs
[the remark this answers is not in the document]
bitdivine: The stable memory upgrade do _not_ work; there is an issue with `rust_struct --[serde]->stable memory --[serde]--> Rust_struct` failing to parse, which is a little bit surprising to put it mildly.  There is a Jira ticket for this, pulled.  This does not break upgrades but  does extend downtime after upgrade, as data is unavailable until it has been re-scraped.

rs/sns_aggregator/src/types/slow.rs
bitdivine: In idl2json I made Nat convert to Javascript strings to preserve precision.  Let's open a Jira ticket to look at this  Done.  See GIX-1402.  I wouldn't block on this, though, given that the consumer isn't doing precise arithmetic with the total supply.

Serde Rust->JSON has a couple of issues that may mean that we have to fork it or write custom serializers for some fields.

bitdivine: Generally I think trapping is a bad idea in the aggregator though.  Please give me some time to look at this and find a good way forwards.

lmuntaner: Ok, thanks!

Feel free to take over the PR and do the changes yourself.

I'll need to do some changes in the nns-dapp frontend afterwards. But I'll do that in another PR after this one.

bitdivine: First of a couple of PRs to help here:  A basic downgrade-upgrade test that runs if the rust code has changed:  https://github.com/dfinity/nns-dapp/pull/2067

rs/sns_aggregator/src/types/slow.rs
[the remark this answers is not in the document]
bitdivine: Update:  The stable memory works now.

Please note that on upgrade, existing data will still be present and it will take a few minutes for new data to be populated.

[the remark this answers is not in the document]
lmuntaner: Ok, so now we can merge this as is or do you want to change how I decoded `Nat` in this PR?

Does the one in mainnet already write the stable memory?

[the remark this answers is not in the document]
lmuntaner: Discussed in a call to continue with merging this.

bitdivine (approved):

PR #2045, opened 2023-03-07
Improve console log check during tests.

dskloetd:
# Motivation

We don't want distracting logs in test output.

# Changes

1. Remove the message that says your test has passed because it may or may not have passed.
2. Explain that the log failure only happens after the test finished.
3. Don't use mocks to replace loggers because that breaks when a test does `jest.resetAllMocks()`.
4. Ban `console.warn` in addition to the other logs we ban.
5. Use `$tests` in the import path in the error message.

# Tests

`npm run test` with and without `console.warn`.


lmuntaner (approved):
Thanks!

PR #2047, opened 2023-03-08
cleanup: Remove impossible return type undefined

dskloetd:
# Motivation

snsSummariesStore is declared to hold SnsSummary[] so snsProjectsStore can't hold undefined. As a result a number of other things can't be undefined either and declare them as potentially undefined is confusing.

I recommend reviewing this PR 1 commit at a time.

# Changes

Commits:
1. snsSummariesStore is declared to hold SnsSummary[] so snsProjectsStore can't hold undefined.
2. snsProjectsStore never holds undefined, so filterActiveProjects is never called with undefined, so declare it to not take nor return undefined.
3. filterActiveProjects can't return undefined so snsProjectsActivePadStore can't hold undefined.
4. filterCommittedProjects is never called with undefined so declare it to not accept undefined.
5. filterProjectsStatus is never called with undefined so declare it to not accept and not return undefined.
6. filterProjectsStatus doesn't return undefined so filterCommittedProjects doesn't return undefined.
7. filterProjectsStatus doesn't return undefined so snsProjectsAdoptedStore can't hold undefined.
8. filterCommittedProjects doesn't return undefined so snsProjectsCommittedStore can't hold undefined.

# Tests

`npm run check` between every commit.
`npm run test`
Manual testing.

### DIFF
--- a/frontend/src/lib/components/accounts/SnsTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/SnsTransactionsList.svelte
@@ -52,7 +52,7 @@
   });
 
   let governanceCanisterId: Principal | undefined;
-  $: governanceCanisterId = $snsProjectsStore?.find(
+  $: governanceCanisterId = $snsProjectsStore.find(
     ({ rootCanisterId: currentId }) =>
       currentId.toText() === rootCanisterId.toText()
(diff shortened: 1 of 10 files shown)


lmuntaner (approved):
LGTM, thanks!

PR #2051, opened 2023-03-08
Feat: Improve merge neurons flow

lmuntaner:
# Motivation

Merge neurons is a source of confusing and bad UX for users.

While we work on a better and more sophisticated solution, this PR addresses main pain points and tries to explain better what's happening.

It uses only text to do so. No new UX is added in this PR.

# Changes

* Add explanation text in SelectNeuronsToMerge.svelte
* Add explanation text in ConfirmNeuronsMerge.svelte

# Tests

Only UI changes.

### DIFF
--- a/frontend/src/lib/components/neurons/ConfirmNeuronsMerge.svelte
+++ b/frontend/src/lib/components/neurons/ConfirmNeuronsMerge.svelte
@@ -26,16 +26,20 @@
     }
   }
 
+  let targetNeuron: NeuronInfo;
+  let sourceNeuron: NeuronInfo;
+  $: [sourceNeuron, targetNeuron] = neurons;
+
   const merge = async () => {
     startBusyNeuron({
       initiator: "merge-neurons",
-      neuronId: neurons[0].neuronId,
+      neuronId: sourceNeuron.neuronId,
     });
     // We know that neurons has 2 neurons.
     // We have a check above that closes the modal if not.
     const id = await mergeNeurons({
-      targetNeuronId: neurons[0].neuronId,
-      sourceNeuronId: neurons[1].neuronId,
+      targetNeuronId: targetNeuron.neuronId,
+      sourceNeuronId: sourceNeuron.neuronId,
     });
 
     if (id !== undefined) {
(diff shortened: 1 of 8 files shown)


lmuntaner:
New UI:

Screenshot 2023-03-08 at 11 46 00
Screenshot 2023-03-08 at 11 46 07

lmuntaner:
@dskloetd please review

lmuntaner:
@mstrasinskis please review also

frontend/src/lib/components/neurons/ConfirmNeuronsMerge.svelte
```diff
@@ -47,39 +49,37 @@
...
-  {/each}
-
-  <p class="additional-text description">
-    {$i18n.neurons.irreversible_action}
-  </p>
+  <h3>{$i18n.neurons.merge_neurons_modal_title_2}</h3>
+
+  <NnsNeuronInfo neuron={sourceNeuron} />
+
+  <h3>{$i18n.neurons.merge_neurons_modal_with}</h3>
```
dskloetd, line 59: Should that be "into" instead of "with"?

frontend/src/lib/components/neurons/ConfirmNeuronsMerge.svelte
```diff
@@ -26,16 +24,20 @@
     }
   }
 
+  let targetNeuron: NeuronInfo;
+  let sourceNeuron: NeuronInfo;
+  $: [sourceNeuron, targetNeuron] = neurons;
```
dskloetd, line 29: Did you manually confirm that this is the correct order in which they are actually merged in the backend?

frontend/src/lib/i18n/en.json
```diff
@@ -229,7 +229,7 @@
     "merge_neurons_edit_selection": "Edit Selection",
     "merge_neurons_modal_merge_button": "Merge Neurons",
     "merge_neurons_modal_title_2": "Merge neuron",
-    "merge_neurons_modal_with": "with",
+    "merge_neurons_modal_with": "into",
```
dskloetd, line 232: Not for this PR but in general you can't really construct sentence from multiple parts and expect them to translate properly. For example in some language "into" might be a suffix rather than a word in between 2 things.

dskloetd (approved):

frontend/src/lib/components/neurons/ConfirmNeuronsMerge.svelte
lmuntaner: Yes, sure

frontend/src/lib/components/neurons/ConfirmNeuronsMerge.svelte
lmuntaner: To the backend we don't pass the array, but one is the target the other is the source.

Therefore, the backend doesn't know about the order of the neurons array.

frontend/src/lib/i18n/en.json
lmuntaner: This puts together two components. It's not exactly a sentence. But yes, I understand. This should be taken into account in the next flow.

mstrasinskis (approved):
LGTM, thank you!

frontend/src/lib/components/neurons/ConfirmNeuronsMerge.svelte
dskloetd: That's not really relevant for what I'm asking. I just wanted to know that you verified that what actually happens is what you wrote in the text and not the other way around.

frontend/src/lib/i18n/en.json
dskloetd: "Merge $X into $Y" is a sentence.
That would be "X'i Y ile birleştir" in Turkish, which you can't currently accomplish.
In yet another language it might not even be possible to have that sentence without putting X after Y.

frontend/src/lib/components/neurons/ConfirmNeuronsMerge.svelte
lmuntaner: Yes, that's why I showed how I pass the neurons properly here:

```
 const id = await mergeNeurons({
      targetNeuronId: targetNeuron.neuronId,
      sourceNeuronId: sourceNeuron.neuronId,
    });
```

I could change it, and it would still work. But the sentence I added in the first screen would be wrong. But the UI of the second screen would work.

PR #2052, opened 2023-03-08
fix: input address field value blocked (edge case)

peterpeterparker:
# Motivation

Edge case: if user enters the address from one of his/her account in the input field, go to review step and come back to the form, the input field will be blocked.

Happens on mainnet as well.

# Changes

- overwrite input address with account address on in manual mode

# Screenshots

Issue:

a

Afterwards:

b


lmuntaner (approved):
Perfect, thanks!

PR #2053, opened 2023-03-08
GIX-1363: Register Vote through GovernanceApiService

lmuntaner:
# Motivation

The fix of the voting button used a test feature.

This is the fix we intended. Calls to governance canister should be done through the governance api service and this one decides when to reset or not the cache.

# Changes

* Add method "registerVote" in the GovernanceApiService. It clears the cache after it.
* Move "registerVote" api function to "governace.api". Slowly, start implementing the agreed pattern that api should be grouped by canister.
* Voting service uses governance api service instead of the api function directly. That means we can remove the call to `resetNeuronsApiService`.

# Tests

* Move test for the api function "registerVote" to the "governace.api.spec".
* Add test for registerVote in the governance api service.
* Change mocks due to moving api function to another file.

### DIFF
--- a/frontend/src/tests/lib/api-services/governance.api-service.spec.ts
+++ b/frontend/src/tests/lib/api-services/governance.api-service.spec.ts
@@ -13,7 +13,8 @@ import {
   createMockKnownNeuron,
   createMockNeuron,
 } from "$tests/mocks/neurons.mock";
-import { Topic } from "@dfinity/nns";
+import { mockProposalInfo } from "$tests/mocks/proposal.mock";
+import { Topic, Vote } from "@dfinity/nns";
 
 jest.mock("$lib/api/governance.api");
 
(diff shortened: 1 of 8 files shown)


frontend/src/tests/lib/api-services/governance.api-service.spec.ts
```diff
@@ -898,4 +899,35 @@ describe("neurons api-service", () => {
...
+
+  describe("registerVote", () => {
+    const params = {
+      neuronId,
+      identity: mockIdentity,
+      proposalId: mockProposalInfo.id,
+      vote: Vote.Yes,
+    };
+
+    it("should call stopDissolving api", () => {
```
dskloetd, line 911: Wrong copy/paste?

frontend/src/tests/lib/api/governance.api.spec.ts
```diff
@@ -629,4 +631,25 @@ describe("neurons-api", () => {
...
   });
+
+  describe("registerVote", () => {
+    const neuronId = BigInt(0);
+    const identity = mockIdentity;
+    const proposalId = BigInt(0);
+
+    it("should call the canister to cast vote neuronIds count", async () => {
+      await registerVote({
+        neuronId: neuronId,
```
dskloetd, line 642: Can be just `neuronId,`

frontend/src/tests/lib/api/governance.api.spec.ts
```diff
@@ -629,4 +631,25 @@ describe("neurons-api", () => {
       await expect(call).rejects.toThrow(error);
     });
   });
+
+  describe("registerVote", () => {
+    const neuronId = BigInt(0);
```
dskloetd, line 636: Some random non-zero number is less likely to be accidentally correct.

frontend/src/tests/lib/api/governance.api.spec.ts
```diff
@@ -629,4 +631,25 @@ describe("neurons-api", () => {
...
+    const identity = mockIdentity;
+    const proposalId = BigInt(0);
```
dskloetd, line 638: same

dskloetd (approved):
Thanks!

frontend/src/tests/lib/api-services/governance.api-service.spec.ts
lmuntaner: yes, sorry

frontend/src/tests/lib/api/governance.api.spec.ts
lmuntaner: correct, thanks!

frontend/src/tests/lib/api/governance.api.spec.ts
lmuntaner: right!

PR #2056, opened 2023-03-09
Fix: SNS Wallet page infinite loading

lmuntaner:
# Motivation

Deep link to an SNS Wallet wasn't working.

The problem was that the `load` function that sets the context store was not called after the accounts were loaded in the store.

It was called only when the identifier was there.

It worked with normal navigation because the store was already filled with the accounts. But if the user arrives with the deep link, the accounts are loaded. Therefore, the `load` function was called before they were loaded, but not after they were in the store.

# Changes

* Call `load` function also when sns accounts store changes in SnsWallet.svelte page.

# Tests

* Upcoming in another PR along with other test for deep links.

### DIFF
--- a/frontend/src/lib/pages/SnsWallet.svelte
+++ b/frontend/src/lib/pages/SnsWallet.svelte
@@ -53,7 +53,8 @@
 
   export let accountIdentifier: string | undefined | null = undefined;
 
-  const load = () => {
+  let load: () => void;
+  $: load = () => {
     if (nonNullish(accountIdentifier)) {
       const selectedAccount = $snsProjectAccountsStore?.find(
         ({ identifier }) => identifier === accountIdentifier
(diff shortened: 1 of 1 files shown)


frontend/src/lib/pages/SnsWallet.svelte
```diff
@@ -53,7 +53,8 @@
 
   export let accountIdentifier: string | undefined | null = undefined;
 
-  const load = () => {
+  let load: () => void;
+  $: load = () => {
```
peterpeterparker, line 57: We already got `$: accountIdentifier, load();` in this page, isn't `load` executed one time too much that way?

Maybe this other load call should be remove or extended?

```
const load = () => {
}

$: accountIdentifier, $snsProjectAccountsStore, load();
```

dskloetd:
Can you explain in the PR description what was wrong and why this fixes it?

Also promise to add a test asap?

frontend/src/lib/pages/SnsWallet.svelte
lmuntaner: They both do the same number of calls.

But I believe we normally use more the approach you suggest; therefore, I'll change it.

PR #2057, opened 2023-03-09
Chore: Understand qrCodeRendered

lmuntaner:
# Motivation

Understand why we needed `qrCodeRendered` in CkBTCReceiveModal.svelte

# Changes

* Add a comment explaining it.
* Skip using the env var to set the initial value.

# Tests

No new functionality.


peterpeterparker:
Note: There is no recording gif in the PR but I assume you tested visually speaking.

lmuntaner:
> Note: There is no recording gif in the PR but I assume you tested visually speaking.

Yes, sorry, the UX doesn't change.

I tried also removing the variable altogether and as you mentioned, there was a glitch.

PR #2060, opened 2023-03-09
GIX-1394: Invalid address depends on universe

lmuntaner:
# Motivation

Address formats depends on universe. Currently, we were accepting a valid ICRC in the NNS universe, and viceversa, a valid NNS ICP address in SNS universe.

This lead to a bad UX when the user got an error message when trying to send the tokens.

Solution: Use different validations depending on the universe.

# Changes

Major changes:
* Add "rootCanisterId" parameter to accounts util `invalidAddress` and use it for the validation.
* Split `invalidICPOrIcrcAddress` into two `invalidIcpAddress` and `invalidIcrcAddress`.
* Add prop "rootCanisterId" in component AddressInput.svelte and use it for the validation.
* Pass prop "rootCanisterId" in TransactionForm until it reaches the AddressInput.svelte.

Minor changes:
* Rename `Address.svelte` to `NnsAddress.svelte` to make clear where it should be used. (This component might be removed when we refactor the modals).
* Change usage of `invalidICPOrIcrcAddress` with the new helpers.

# Tests

* Test new functionality in AddressInput.svelte.
* Test `invalidAddress` new parameter.
* Test new accounts utils `invalidIcpAddress` and `invalidIcrcAddress`.


lmuntaner:
New UX:

Validation is different depending on Universe.

Below I try a valid ICP address in SNS transaction, and a valid SNS address in ICP transaction.

https://user-images.githubusercontent.com/4550653/223966036-fe2a6156-cbb0-4841-8b58-fa4de26d074c.mov

frontend/src/lib/components/accounts/SelectNetworkDropdown.svelte
```diff
@@ -36,19 +37,20 @@
...
       return;
     }
 
-    if (
-      !invalidBtcAddress({
-        address: selectedDestinationAddress,
-        network: ckTESTBTC ? BtcNetwork.Testnet : BtcNetwork.Mainnet,
-      })
-    ) {
+    const validBtcAddres = !invalidBtcAddress({
```
peterpeterparker, line 49: typo "Addres"

frontend/src/lib/utils/accounts.utils.ts
```diff
@@ -50,14 +51,16 @@ export const invalidBtcAddress = (address: BtcAddress): boolean => {
 /**
  * Is the address a valid entry to proceed with any action such as transferring ICP?
  *
- * Note: undefined is considered here as a valid address
```
peterpeterparker, line 53: this is still correct, what removing the comment?

frontend/src/lib/utils/accounts.utils.ts
```diff
@@ -50,14 +51,16 @@ export const invalidBtcAddress = (address: BtcAddress): boolean => {
...
+ * The address format depends on the universe. Therefore, the root canister ID is required.
  */
 export const invalidAddress = ({
```
peterpeterparker, line 56: not for this PR but as you suggested in the previous PR maybe we should have different components - i.e. get rid of a global `invalidAddress` and be specific in the components.

also, not for this pr, maybe we should rename/refactor to `validAddress` instead of testing for invalid?

Just thinking out at loud in both case.

peterpeterparker (approved):
LGTM,thx

frontend/src/lib/utils/accounts.utils.ts
lmuntaner: I thought about it, but it wasn't so straightforward to create multiple components.

I thought of passing the validator function to `TransactionModal`, but that added quite some props drilling and also removed all the logic of AddressInput and made it pretty dumb.

Another option was to pass the component as slot and have `IcrcAddressInput`. But I'm not sure...

What do you think? Do you like any of the ideas above?

frontend/src/lib/utils/accounts.utils.ts
peterpeterparker: Ah you are right, it's mixed in my mind. One more reason to rename it one of these days.

lmuntaner: Yes, I saw myself failing to read it properly as well

frontend/src/lib/utils/accounts.utils.ts
peterpeterparker: The other day I was thinking that we should maybe create a type for the transaction modal property and pass a single object. There are slowely lots of parameters, so thought that having one object might be easier but, did not thought about it further.

Not sure neither, both sounds interesting way. Nevertheless no rush neither, was just thinking out at loud. Thanks for the share!

lmuntaner: Sounds also interesting. I agree it has many props now. We should split it of think on how to make it nicer.

frontend/src/lib/components/accounts/SelectNetworkDropdown.svelte
lmuntaner: thansk!

PR #2061, opened 2023-03-09
Update changelog

mstrasinskis:
# Motivation

Release-2023-03-09

# Changes

- changelog update
- version update

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ members = [
 ]
 
 [workspace.package]
-version = "2.0.21"
+version = "2.0.23"
 
 [profile.release]
 lto = false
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -5,7 +5,7 @@ members = [
 ]
 
 [workspace.package]
-version = "2.0.21"
+version = "2.0.23"
```
lmuntaner, line 8: Oh, thanks! I forgot now that the `Cargo.toml` is in the root directly.

I didn't see the verison in `rs/backend` and I forgot about it.

lmuntaner (approved):
Thanks!

PR #2064, opened 2023-03-09
Upgrade instruction counter

bitdivine:
# Motivation
If the post-upgrade hook consumes too many cycles, it may fail.  Such failures are irrecoverable except by rolling back the subnet.  This is bad.  We would like to have more visibility into cycle consumption.

Cycle consumption is instruction count times a factor.  The factor depends on such things as subnet size.  As such,
cycle count is actually not useful to test in CI but it can be converted into an instruction count limit that _can_ be tested in CI.

A good first step is to capture instruction counts and make them available for examination.

# Changes
- Extend the existing `get_stats()` API method to include instruction counters in the response.
  - Define an  instruction counter struct including the count, time, and a text label.
  - Add a circular buffer with the instruction counts to stable memory.
- Add a script to get the last upgrade instruction count.

# Future
- Determine an instruction count threshold for the post upgrade hook
- Add the post-upgrade instruction count to the metrics.
- Add methods to get the cycle count.
- Take measures to  reduce the post-upgrade cycle count.
  Note: Measurements show that the overwhelming majority of cycles are consumed when unpacking and re-certifying assets.  Adding the assets could be done in a separate function call after the post-upgrade hook, but triggered by the post-upgrade.

# Tests
- In CI, upgrade and get the instructions consumed in the last upgrade.
  - Note: This test does NOT currently verify that the instruction count is below some threshold.  Once such a threshold is established, this would be the right place to enforce it.


rs/backend/src/accounts_store.rs
```diff
@@ -42,6 +43,8 @@ pub struct AccountsStore {
     hardware_wallet_accounts_count: u64,
     last_ledger_sync_timestamp_nanos: u64,
     neurons_topped_up_count: u64,
+
+    instruction_counters: VecDeque<PerformanceCounter>,
```
bitdivine, line 47: Note: `get_stats()` is advertised as returning stats about the canister as a whole but  is actually implemented on the accounts store.  This forces the instruction counters to be stored here as well.  A refactor, that moves `get_stats()` out of the accounts store would be in order.

rs/backend/src/perf.rs
```diff
@@ -0,0 +1,45 @@
+use crate::STATE;
+use candid::CandidType;
+use dfn_core::api::ic0;
+use serde::Deserialize;
+
+#[derive(CandidType, Deserialize, Clone)]
+pub struct PerformanceCounter {
+    timestamp_ns_since_epoch: u64,
+    name: String,
+    instruction_counter: u64,
```
dskloetd, line 10: Does this field ever change? If not, it's really just a count and not a counter, right?
Same for the PerformanceCounter and everything else called "counter" in this change.

rs/backend/src/perf.rs
```diff
@@ -0,0 +1,45 @@
...
+pub fn save_instruction_counter(counter: PerformanceCounter) {
+    STATE.with(|s| s.accounts_store.borrow_mut().save_instruction_counter(counter))
+}
+
+/// Records an instruction counter.
+pub fn record_instruction_counter(name: &str) {
+    STATE.with(|s| {
+        s.accounts_store
+            .borrow_mut()
+            .save_instruction_counter(PerformanceCounter::new(name))
```
dskloetd, line 43: Could these 4 lines be replaced with just `save_instruction_counter(PerformanceCounter::new(name))`?

rs/backend/src/accounts_store.rs
```diff
@@ -851,6 +854,14 @@ impl AccountsStore {
         self.multi_part_transactions_processor.push(block_height, transaction);
     }
 
+    /// Adds an instruction counter to the circular buffer.
+    pub fn save_instruction_counter(&mut self, counter: PerformanceCounter) {
+        if self.instruction_counters.len() >= 100 {
```
dskloetd, line 859: With the current code the length would never be more than 4, right?

.github/workflows/build.yml
```diff
@@ -110,6 +110,10 @@ jobs:
           rm -rf screenshots
           set -o pipefail
           SCREENSHOT=1 npm run test | tee -a wdio.log
+      - name: Get the postinstall instruction count
+        run: |
+          dfx canister install --upgrade-unchanged nns-dapp --wasm nns-dapp.wasm --mode upgrade
+          postinstall_instructions="$(scripts/backend/get_upgrade_instructions)"
```
dskloetd, line 116: Is this used somewhere or does it appear in the workflow output?

bitdivine: It is used nowhere.  We need to determine the threshold before we can enforce it.  We could print it to the output, yes, that is an idea.

rs/backend/src/accounts_store.rs
bitdivine: In the test, you mean, because we upgrade only once?  Yes.  In prod, it would keep growing with every upgrade.  We could add other counters as well.

rs/backend/src/perf.rs
bitdivine: Yes, fair enough.

rs/backend/src/perf.rs
bitdivine: Yes

rs/backend/src/accounts_store.rs
dskloetd: During every upgrade it's reset to empty on line 1369, isn't it?

bitdivine: In the current PR, yes, but only because I haven't moved the storage to where I want it to be.  That's why this PR is marked draft. :-)

rs/backend/src/main.rs
```diff
@@ -43,17 +44,24 @@ fn pre_upgrade() {
...
 
-        s.replace(new_state)
+        s.replace(new_state);
     });
 
+    perf::save_instruction_count(counter_before);
+    perf::record_instruction_count("post_upgrade after state_recovery");
     assets::init_assets();
+    perf::record_instruction_count("post_upgrade after init_assets");
+    perf::record_instruction_count("post_upgrade stop");
```
dskloetd, line 64: Should there be something between line 63 and 64?
Otherwise the difference between those 2 records is just the number of instructions it takes to record a record, right?
Seems kind of pointless?

scripts/backend/get_upgrade_instructions
```diff
@@ -0,0 +1,21 @@
...
+	      in a subnet, and that does differ significantly between subnets.
+	EOF
+}
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/../clap.bash"
+# Define options
+clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+set -euo pipefail
```
dskloetd, line 19: Put this at the top and temporarily `set +e` for the clap?
(in case there are issues with realpath again)

dskloetd (approved):

scripts/backend/get_upgrade_instructions
bitdivine: It would be nice to start pushing into clap and fixing the issues at source.  A start would be to do the set +e in clap so that the scripts that call it don't have to contort themselves to satisfy it.

rs/backend/src/main.rs
bitdivine: It makes pairing start + stop easier.  But, yes, the counts will be practically the same.  Maybe the thing to do is to support multiple labels, even if only joined by `;`.

scripts/backend/get_upgrade_instructions
dskloetd: In that case you should also do `set -e` at the end of clap, but only if it was set before.

PR #2065, opened 2023-03-09
cleanup: Remove setLoading on snsQueryStore and snsProposalsStore

dskloetd:
# Motivation

snsQueryStore and snsProposalsStore use `undefined` to mean reset and `null` to mean loading. But then when checking whether they are loading we just use isNullish. This is confusing and it's also not super clear that setLoadingState actually resets all the data.

To make this obvious we use reset() instead of setLoadingState() and expose separate stores to check whether the store is loading via isNullish.

# Changes

1. Remove setLoading from snsQueryStore and snsProposalsStore.
2. Add snsQueryStoreIsLoading and snsProposalsStoreIsLoading
3. Use `.reset()` instead of `.setLoading()`.
4. Use `snsQueryStoreIsLoading` instead of `isNullish($snsSummariesStore) || isNullish($snsQueryStore)`. [snsSummariesStore](https://github.com/dfinity/nns-dapp/blob/4588644d282a9a3bbe4df3bac35215021f277361/frontend/src/lib/stores/sns.store.ts#L271) can't actually be non-nullish and `isNullish($snsQueryStore)` is what `snsQueryStoreIsLoading` does.
5. Use $snsProposalsStoreIsLoading instead of isNullish($snsProposalsStore)

# Tests

1. Changed an existing unit test for snsQueryStoreIsLoading and added a new test for snsProposalsStoreIsLoading.
2. Manual testing.

### DIFF
--- a/frontend/src/lib/components/launchpad/Projects.svelte
+++ b/frontend/src/lib/components/launchpad/Projects.svelte
@@ -3,8 +3,7 @@
   import ProjectCard from "./ProjectCard.svelte";
   import SkeletonProjectCard from "$lib/components/ui/SkeletonProjectCard.svelte";
   import { keyOf } from "$lib/utils/utils";
-  import { isNullish } from "@dfinity/utils";
-  import { snsQueryStore, snsSummariesStore } from "$lib/stores/sns.store";
+  import { snsQueryStoreIsLoading } from "$lib/stores/sns.store";
   import {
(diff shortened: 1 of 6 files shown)


lmuntaner (approved):
LGTM!

Should we start testing the idea of extending the "subscribe" instead of fixing smaller things here and there?

PR #2067, opened 2023-03-09
Downgrade upgrade test

bitdivine:
# Motivation
We are making more changes to the nns-dapp  backend and in particular to the stable memory.  It would be good to have at least a very basic downgrade-upgrade test in CI.

Requirements:
- Fast; should not slow down CI significantly.
- Verify that there are no panics in the pre and post-upgrade hooks.

Not covered:
- Detailed checks that might be applicable to a specific upgrade, e.g. making sure that stable data is preserved or manipulated in a  certain way.

# Changes
- Add a script that performs a downgrade and then an upgrade.  Given that the current code is usually deployed when testing, a downgrade-upgrade test is quicker to implement than an upgrade-downgrade test, which would really be a downgrade-upgrade-downgrade-upgrade test.
- Perform a health check by calling the stats endpoint.  This is much faster than an e2e test.

# Tests
- No change to rust code: See how these tests perform in CI.
- Change to Rust code:  I have run the script in this PR on code that breaks that stable memory and indeed, the script fails.

### DIFF
--- a/scripts/nns-dapp/downgrade-upgrade-test
+++ b/scripts/nns-dapp/downgrade-upgrade-test
@@ -3,7 +3,7 @@ SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 PATH="$SOURCE_DIR:$PATH"
 
 help_text() {
-	cat <<-EOF
+  cat <<-EOF
 	Downgrade and then upgrade a local deployment.
 
 	Note: This deploys nns-dapp_local.wasm for the upgrade.
@@ -29,38 +29,38 @@ PROD_WASM="nns-dapp-prod_local.wasm"
 CURRENT_WASM="nns-dapp.wasm"
 
 verify_healthy() {
-	dfx canister call nns-dapp get_stats
+  dfx canister call nns-dapp get_stats
 }
 
 get_current_wasm() {
-	DFX_NETWORK=local ./scripts/docker-build
+  DFX_NETWORK=local ./scripts/docker-build
 }
 
 get_prod_wasm() {
-	curl -sL https://github.com/dfinity/nns-dapp/releases/download/prod/nns-dapp_local.wasm > "$PROD_WASM"
+  curl -sL https://github.com/dfinity/nns-dapp/releases/download/prod/nns-dapp_local.wasm >"$PROD_WASM"
 }
 
 # Installs the current build of nns-dapp
 upgrade_nnsdapp() {
-	dfx canister install --upgrade-unchanged nns-dapp --wasm "$1" --mode upgrade
+  dfx canister install --upgrade-unchanged nns-dapp --wasm "$1" --mode upgrade
 }
 
-
 test -e "$CURRENT_WASM" || {
-	echo "Please build '$CURRENT_WASM' before running this test."
-	exit 1
+  echo "Please build '$CURRENT_WASM' before running this test."
+  exit 1
 } >&2
 
 echo "Build the wasm, if it doesn't exist yet."
 test -e "$CURRENT_WASM" || {
-	echo "Building $CURRENT_WASM..."
-	DFX_NETWORK=local ./scripts/docker-build
+  echo "Building $CURRENT_WASM..."
+  DFX_NETWORK=local ./scripts/docker-build
 }
 echo "Make sure that the current wasm is the one currently installed"
-if dfx canister id nns-dapp >/dev/null 2>/dev/null
-then upgrade_nnsdapp "$CURRENT_WASM"
-else dfx canister create nns-dapp
-     dfx canister install nns-dapp --wasm "$CURRENT_WASM"
+if dfx canister id nns-dapp >/dev/null 2>/dev/null; then
+  upgrade_nnsdapp "$CURRENT_WASM"
+else
+  dfx canister create nns-dapp
+  dfx canister install nns-dapp --wasm "$CURRENT_WASM"
 fi
 echo "Make sure that the current wasm is running healthily..."
 verify_healthy
--- a/scripts/nns-dapp/downgrade-upgrade-test
+++ b/scripts/nns-dapp/downgrade-upgrade-test
@@ -1,0 +1,78 @@
+#!/usr/bin/env bash
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+
+help_text() {
+	cat <<-EOF
+	Downgrade and then upgrade a local deployment.
+
+	Note: This deploys nns-dapp_local.wasm for the upgrade.
+	      If the wasm is already present, it will be used.
+	      Please make sure that hte wasm is up to date.
+	      If the wasm is absent, it will be built.
+
+	Note: If the nns-dapp is already deployed, the deployment
+	      will be used in the downgrade and upgrade.  You may
+	      find it useful to verify that the state has been preserved.
+	      If the existing deployment is unhealthy, upgrades may fail.
+	EOF
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/../clap.bash"
+# Define options
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+set -euo pipefail
+
+PROD_WASM="nns-dapp-prod_local.wasm"
+CURRENT_WASM="nns-dapp.wasm"
+
+verify_healthy() {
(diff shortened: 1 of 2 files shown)


scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -0,0 +1,75 @@
...
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+
+help_text() {
+  cat <<-EOF
+	Downgrade and then upgrade a local deployment.
+
+	Note: This deploys nns-dapp_local.wasm for the upgrade.
+	      If the wasm is already present, it will be used.
+	      Please make sure that hte wasm is up to date.
```
dskloetd: the

dskloetd: This feels like a trap. Can the script just make sure that it's up to date?

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -0,0 +1,75 @@
...
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/../clap.bash"
+# Define options
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+set -euo pipefail
+
+PROD_WASM="nns-dapp-prod_local.wasm"
```
dskloetd, line 28: What does the "_local" here refer to?

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -0,0 +1,75 @@
...
+	Note: This deploys nns-dapp_local.wasm for the upgrade.
+	      If the wasm is already present, it will be used.
+	      Please make sure that hte wasm is up to date.
+	      If the wasm is absent, it will be built.
+
+	Note: If the nns-dapp is already deployed, the deployment
+	      will be used in the downgrade and upgrade.  You may
+	      find it useful to verify that the state has been preserved.
+	      If the existing deployment is unhealthy, upgrades may fail.
+	EOF
```
dskloetd, line 18: This help text did not make me understand what the script does until I read the code.
Can we clearly state that this
1. Deploys the locally built wasm (if it isn't deployed already)
2. Then deploys the wasm from the latest release downloaded from GitHub
3. Then deploys the locally built wasm (again)

Or if that's not correct, something similar which is correct?

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -0,0 +1,75 @@
...
+upgrade_nnsdapp() {
+  dfx canister install --upgrade-unchanged nns-dapp --wasm "$1" --mode upgrade
+}
+
+test -e "$CURRENT_WASM" || {
+  echo "Please build '$CURRENT_WASM' before running this test."
+  exit 1
+} >&2
+
+echo "Build the wasm, if it doesn't exist yet."
```
dskloetd, line 53: How can this ever happen if it exited on line 50 if "$CURRENT_WASM" doesn't exist?

dskloetd, line 53: Is this an instruction to the user or a narration of what the script does?
If the latter "Building" might be clearer.

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -0,0 +1,75 @@
...
+test -e "$CURRENT_WASM" || {
+  echo "Building $CURRENT_WASM..."
+  DFX_NETWORK=local ./scripts/docker-build
+}
+echo "Make sure that the current wasm is the one currently installed"
+if dfx canister id nns-dapp >/dev/null 2>/dev/null; then
```
dskloetd, line 59: This only checks if any canister with the name "nns-dapp" exists, right?
It has nothing to do with what's actually deployed?

bitdivine: Correct, and that is correct.

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: Good catch, the earlier check can be deleted.

bitdivine: A narration.

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: By convention, that is the name of the `NNS_NETWORK`.

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: It's not correct; I'll describe what it does in the text here.

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: No, not efficiently.

bitdivine: Done

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: Done.  Normally I would say that the description should say what code achieves, not what it does, so the new comment feels a bit wrong.  Maybe this point needs some discussion.  Why did you want to the comment to provide such a list?

scripts/nns-dapp/downgrade-upgrade-test
dskloetd: I think correctness is more important than efficiency. Maybe add a flag to skip the build for when you're confident it's correct?

scripts/nns-dapp/downgrade-upgrade-test
dskloetd: I didn't mean that the description should be pseudo code of the code or that it should be a list of 10 things.
But I didn't understand the description at all and my issue with it hasn't really gone away.
The question I wanted answered in the description is what "downgrade" and "upgrade" really mean. Which versions exactly are deployed to accomplish the "downgrade" and the "upgrade". That's what I tried to make clean in my 3 (not 10) point list.

bitdivine: The downgrade is to prod and the upgrade is to the current release.  Both are well defined.

dskloetd: It's well defined what "prod" *is* (the wasm currently running on mainnet) but unless your are literally deploying to mainnet, it's not well defined *how* you get the wasm that's currently running on mainnet.
1. Do you download it from mainnet? (is that even possible?)
2. Do you download it from somewhere where it is uploaded to at the same moment the proposal is accepted?
3. Do you build it form a given git hash? And if so where do you get that hash? And how does it make sure it's always equal to prod as accepted by proposal?

But if I understand the script correctly, it's none of those things.

"Current release" is even less clear to me. Is it the latest release candidate? Where/how do you get it?
But if I understand the script correctly, it's just whatever is in your current director. You could call this the current release if the script is run in the context of creating a new release, but it's entirely valid to run the script in any other context, isn't it?

So yes, I do think it's useful to clarify these 2 things.

bitdivine: The release procedure includes tagging the release commit as `prod`, so that is well defined.  The mechanics are not important unless debugging the mechanic, which isn't relevant in the context of the high level function description.

But "current" could perhaps be misinterpreted as "code currently on mainnet" or something like that?  HEAD in the local directory  is inaccurate as there might be uncommitted changes.  Maybe just "local build/code/wasm" is best, what do you think?

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -0,0 +1,69 @@
...
+
+PROD_WASM="nns-dapp-prod_local.wasm"
+
+help_text() {
+  cat <<-EOF
+	Tests:
+	  - downgrading the local build to prod.
+	  - upgrading prod to the local build.
+	EOF
+  # Note: If the code changes, please update the summary above to match.
```
dskloetd, line 13: I guess this referred to a previous version?

dskloetd (approved):

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: Done.

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: Yes, good catch.

PR #2069, opened 2023-03-09
GIX-1328 Only show sns neurons for committed projects

dskloetd:
# Motivation

Currently it's possible to see founder neurons before the sns project is committed by manually change a URL parameter.
The current code checks that the URL parameter represents an existing project but not that the project is committed.

# Changes

In Neurons.svelte, instead of only rendering SnsNeurons if the selected project exists, also check that it's committed.

# Tests

Unit tests and manual testing.

### DIFF
--- a/frontend/src/lib/derived/sns/sns-selected-project.derived.ts
+++ b/frontend/src/lib/derived/sns/sns-selected-project.derived.ts
@@ -1,5 +1,6 @@
 import { selectedUniverseIdStore } from "$lib/derived/selected-universe.derived";
 import {
+  snsProjectsCommittedStore,
   snsProjectsStore,
   type SnsFullProject,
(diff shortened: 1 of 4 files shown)


frontend/src/lib/derived/sns/sns-selected-project.derived.ts
```diff
@@ -29,3 +30,14 @@ export const snsProjectSelectedStore: Readable<SnsFullProject | undefined> =
           rootCanisterId.toText() === $selectedUniverseIdStore.toText()
       )
   );
+
+export const snsCommittedProjectSelectedStore: Readable<
```
lmuntaner, line 34: Great thanks!

We might want to do the same in the Accounts route. We also have SnsAccounts and NnsAccounts.

lmuntaner (approved):
LGTM, thanks!

PR #2070, opened 2023-03-10
Refactor: Move stats out of the accounts store

bitdivine:
# Motivation
Stats and metrics are currently calculated inside the accounts store.  This is limiting, given that there are interesting statistics to be had from other parts of the canister.

# Changes
* Move the stats code out of the accounts store, leaving behind just the part that populates account stats.

# Tests
This changes no externally functionality so no new tests are required.

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -1,11 +1,10 @@
 use crate::constants::{MEMO_CREATE_CANISTER, MEMO_TOP_UP_CANISTER};
-use crate::metrics_encoder::MetricsEncoder;
 use crate::multi_part_transactions_processor::{
     MultiPartTransactionToBeProcessed, MultiPartTransactionsProcessor, MultiPartTransactionsProcessorWithRemovedFields,
 };
 use crate::state::StableState;
+use crate::stats::Stats;
 use crate::time::time_millis;
-use crate::STATE;
 use candid::CandidType;
 use dfn_candid::Candid;
 use ic_base_types::{CanisterId, PrincipalId};
@@ -851,7 +850,7 @@ impl AccountsStore {
         self.multi_part_transactions_processor.push(block_height, transaction);
     }
 
-    pub fn get_stats(&self) -> Stats {
+    pub fn get_stats(&self, stats: &mut Stats) {
         let earliest_transaction = self.transactions.front();
         let latest_transaction = self.transactions.back();
         let timestamp_now_nanos = dfn_core::api::now()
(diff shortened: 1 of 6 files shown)


rs/backend/src/stats.rs
```diff
@@ -0,0 +1,71 @@
...
+use serde::Deserialize;
+
+pub fn get_stats(state: &State) -> Stats {
+    let mut ans = Stats::default();
+    // Collect values from various subcomponents
+    state.accounts_store.borrow().get_stats(&mut ans);
+    // Return all the values
+    ans
+}
+
```
bitdivine, line 15: I have made stats fields pub to allow the accounts module to populate the parts relevant to it - all parts at the time of writing.  Also, there is not much point in collecting stats if they cannot be read by other code.

rs/backend/src/stats.rs
```diff
@@ -0,0 +1,71 @@
+use crate::metrics_encoder::MetricsEncoder;
+use crate::state::State;
+use crate::STATE;
+use candid::CandidType;
+use icp_ledger::BlockIndex;
+use serde::Deserialize;
+
```
bitdivine, line 7: New function that gets stats for the whole canister.

rs/backend/src/accounts_store.rs
```diff
@@ -1508,64 +1505,5 @@ pub enum TransferResult {
     },
 }
```
bitdivine, line 1507: `encode_metrics` has moved to `stats`.

rs/backend/src/accounts_store.rs
```diff
@@ -851,7 +850,7 @@ impl AccountsStore {
         self.multi_part_transactions_processor.push(block_height, transaction);
     }
```
bitdivine, line 852: Instead of creating the stats, this now contributes to an existing stats structure.

rs/backend/src/stats.rs
```diff
@@ -0,0 +1,71 @@
...
+pub fn get_stats(state: &State) -> Stats {
+    let mut ans = Stats::default();
+    // Collect values from various subcomponents
+    state.accounts_store.borrow().get_stats(&mut ans);
```
dskloetd, line 11: Is there an existing test that fails if you remove this line?

bitdivine: No, none.  Such a test would presumably have to have an idea of what data is expected to be present.  E.g. an e2e test that created an account and then checked that that the number of accounts is non-zero.

dskloetd: I don't know much about Rust testing but I don't see why it would have to be an e2e test.
If the code is in a state that's hard to test, it might require some refactoring to make it testable but generally those type of refactorings make the code better in general.
We do want our code to be testable and tested, don't we?

bitdivine: I have added a unit test that would catch this.

dskloetd (approved):

rs/backend/src/accounts_store.rs
```diff
@@ -861,23 +860,21 @@ impl AccountsStore {
...
-            earliest_transaction_block_height: earliest_transaction.map_or(0, |t| t.block_height),
-            latest_transaction_timestamp_nanos: latest_transaction
-                .map_or(0, |t| t.timestamp.as_nanos_since_unix_epoch()),
-            latest_transaction_block_height: latest_transaction.map_or(0, |t| t.block_height),
-            seconds_since_last_ledger_sync: duration_since_last_sync.as_secs(),
-            neurons_created_count: self.neuron_accounts.len() as u64,
-            neurons_topped_up_count: self.neurons_topped_up_count,
-            transactions_to_process_queue_length: self.multi_part_transactions_processor.get_queue_length(),
-        }
+        stats.accounts_count = self.accounts.len() as u64;
```
peterpeterparker, line 863: would it be nice to extract a `trait`?

PR #2071, opened 2023-03-10
GIX-1401: Use loadBalance instead of syncAccounts

lmuntaner:
# Motivation

Continuing with the task of reducing the number of update calls by using `loadBalance` instead of `syncAccounts`.

# Changes

* Use `loadBalance` in the accounts service `transferICP`.
* Use `loadBalance` in the canister service `createCanister`.
* Use `loadBalance` in the canister service `topUpCanister`.
* Use `loadBalance` in the sns sale service `participateInSnsSale`.

# Tests

* Change expectations of calling `syncAccounts` service for only the `queryAccountBalance` api.
* Major change in `canister.services.spect.ts`. Mock only api layer.
* Use new `blockAllCallsTo` in canister and accounts services test files.

### DIFF
--- a/frontend/src/lib/services/accounts.services.ts
+++ b/frontend/src/lib/services/accounts.services.ts
@@ -37,6 +37,7 @@ import {
 } from "$lib/utils/utils";
 import type { Identity } from "@dfinity/agent";
 import { ICPToken, TokenAmount } from "@dfinity/nns";
+import { nonNullish } from "@dfinity/utils";
 import { get } from "svelte/store";
 import { getAuthenticatedIdentity } from "./auth.services";
 import { queryAndUpdate } from "./utils.services";
@@ -253,7 +254,19 @@ export const transferICP = async ({
       amount: tokenAmount,
     });
 
-    await syncAccounts();
+    // Transfer can be to one of the user's account.
+    const toAccount = findAccount({
+      identifier: to,
+      accounts: get(nnsAccountsListStore),
+    });
+    await Promise.all([
+      loadBalance({ accountIdentifier: identifier }),
+      ...[
+        nonNullish(toAccount)
+          ? loadBalance({ accountIdentifier: to })
+          : Promise.resolve(),
+      ],
+    ]);
 
     return { success: true };
   } catch (err) {
--- a/frontend/src/lib/services/accounts.services.ts
+++ b/frontend/src/lib/services/accounts.services.ts
@@ -261,11 +261,9 @@ export const transferICP = async ({
     });
     await Promise.all([
       loadBalance({ accountIdentifier: identifier }),
-      ...[
-        nonNullish(toAccount)
-          ? loadBalance({ accountIdentifier: to })
-          : Promise.resolve(),
-      ],
+      nonNullish(toAccount)
+        ? loadBalance({ accountIdentifier: to })
+        : Promise.resolve(),
     ]);
 
     return { success: true };
--- a/frontend/src/tests/lib/services/accounts.services.spec.ts
+++ b/frontend/src/tests/lib/services/accounts.services.spec.ts
@@ -526,7 +526,16 @@ describe("accounts-services", () => {
         accountIdentifier: mockSubAccount.identifier,
         certified: true,
       });
-      // 2 times for source account, 2 times for destination account
+      expect(queryAccountBalanceSpy).toHaveBeenCalledWith({
+        identity: mockIdentity,
+        accountIdentifier: sourceAccount.identifier,
+        certified: false,
+      });
+      expect(queryAccountBalanceSpy).toHaveBeenCalledWith({
+        identity: mockIdentity,
+        accountIdentifier: sourceAccount.identifier,
+        certified: true,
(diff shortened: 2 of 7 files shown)


frontend/src/lib/services/accounts.services.ts
```diff
@@ -253,7 +254,19 @@ export const transferICP = async ({
...
 
-    await syncAccounts();
+    // Transfer can be to one of the user's account.
+    const toAccount = findAccount({
+      identifier: to,
+      accounts: get(nnsAccountsListStore),
+    });
+    await Promise.all([
+      loadBalance({ accountIdentifier: identifier }),
+      ...[
```
dskloetd, line 264: You can either use an empty array instead of Promise.resolve() or remove the `...`.
So either
```
    await Promise.all([
      loadBalance({ accountIdentifier: identifier }),
      ...(nonNullish(toAccount)
          ? [loadBalance({ accountIdentifier: to })]
          : []),
]);
```
or
```
await Promise.all([
      loadBalance({ accountIdentifier: identifier }),
      nonNullish(toAccount)
        ? loadBalance({ accountIdentifier: to })
        : Promise.resolve(),
]);
```

frontend/src/tests/lib/services/accounts.services.spec.ts
```diff
@@ -459,33 +463,71 @@ describe("accounts-services", () => {
...
+        accountIdentifier: mockSubAccount.identifier,
+        certified: false,
+      });
+      expect(queryAccountBalanceSpy).toHaveBeenCalledWith({
+        identity: mockIdentity,
+        accountIdentifier: mockSubAccount.identifier,
         certified: true,
       });
+      // 2 times for source account, 2 times for destination account
+      expect(queryAccountBalanceSpy).toHaveBeenCalledTimes(4);
```
dskloetd, line 530: I only see parameter expectations for 2 of them. Can you add the other 2 as well?

frontend/src/tests/lib/services/sns-sale.services.spec.ts
```diff
@@ -917,14 +921,39 @@ describe("sns-api", () => {
...
       expect(upgradeProgressSpy).toBeCalledTimes(4);
     });
 
+    it("should update account's balance in the store", async () => {
+      snsTicketsStore.setTicket({
+        rootCanisterId: rootCanisterIdMock,
+        ticket: testTicket,
+      });
+      accountsStore.set({
+        main: mockMainAccount,
```
dskloetd, line 938: Can you also add a test that uses a subaccount since that has different logic?

frontend/src/tests/lib/services/sns-sale.services.spec.ts
```diff
@@ -917,14 +921,39 @@ describe("sns-api", () => {
...
+      const postprocessSpy = jest.fn().mockResolvedValue(undefined);
+      const upgradeProgressSpy = jest.fn().mockResolvedValue(undefined);
+
+      expect(get(accountsStore).main.balance.toE8s()).not.toEqual(
+        newBalanceE8s
+      );
+
+      await participateInSnsSale({
+        rootCanisterId: testRootCanisterId,
+        userCommitment: 0n,
```
dskloetd, line 949: Does it make sense to participate with 0?

frontend/src/lib/services/accounts.services.ts
lmuntaner: true, haha

frontend/src/tests/lib/services/accounts.services.spec.ts
lmuntaner: Those are in the previous test case. Here I'm testing that the destination is also called in case it matches an account in the store.

That the source is called is done in the previous test.

frontend/src/tests/lib/services/sns-sale.services.spec.ts
lmuntaner: Sure

frontend/src/tests/lib/services/sns-sale.services.spec.ts
lmuntaner: This is the current user commitment.

Maybe the naming is not the best.

frontend/src/tests/lib/services/sns-sale.services.spec.ts
lmuntaner: done

lmuntaner:

frontend/src/tests/lib/services/accounts.services.spec.ts
dskloetd: You are making a claim in the comment, but without expectation you don't know if that's true. The previous test is a different test.

lmuntaner: Ok, I'll add them

lmuntaner: done

dskloetd (approved):

PR #2072, opened 2023-03-10
clippy (aggregator): Prohibit panicking except where explicitly allowed

bitdivine:
# Motivation
Panicking can be anywhere between bad and disasterous if it happens in the wrong place.  Unfortunately, prohibiting panic on a function does not prohibit panics in dependencies of that function.  So, instead, we prohibit panicking everywhere and whitelist a few cases where it is safe and correct.

# Changes
In the aggregator:
- Move deny panic to the lib.rs module so that it covers the whole codebase.
- Remove panics where possible, whitelist other cases.

# Tests
This is a static compile time check.  If it compiles, it works.


rs/sns_aggregator/src/upstream.rs
```diff
@@ -38,17 +34,21 @@ pub async fn update_cache() {
     }
 }
 
+/// The NNS SNS wasm canister ID
+///
+/// This canister contains a lit of all SNS root canisters it has created.
```
lmuntaner, line 39: Typo `lit`

lmuntaner (approved):
Approved.

Bu I have some general comments:

If I understand, we can panic, only where it's explicitly allowed. Is this correct?

For the future, could you write some guidelines on where it's ok to panic and where is not ok to panic?

So that we keep consistency and don't start enabling it in places you want to specially avoid id.

Thanks!

rs/sns_aggregator/src/upstream.rs
bitdivine: Good catch!

PR #2073, opened 2023-03-11
Aggregator stable memory serialization &  deserilaization

bitdivine:
# Motivation
The aggregator stable memory serialization and deserialization does not work.  It uses serde to encode state to CBOR, but serde is broken for Nat.

# Changes
* Switch the stable memory serialization to Candid, as that works.
* Re-enable stable memory recovery in the post-install hook.

# Tests
* Add unit and e2e tests for stable memory serialization and deserialization.
* Add unit tests for Nat serde.


rs/sns_aggregator/src/state.rs
```diff
@@ -18,6 +18,9 @@ use std::cell::RefCell;
 use std::collections::VecDeque;
 use std::str::FromStr;
 
+#[cfg(test)]
+mod tests;
```
lmuntaner, line 22: I guess it's a rust question, but why do you need this here?

lmuntaner (approved):
LGTM, just a non-blocking question.

rs/sns_aggregator/src/state.rs
bitdivine: Just convention; normally tests for a module live in a sub-module so that they have access to internals.  However in this case all the fields are public so it is just convention.

lmuntaner: ok, thanks for the explanation!

PR #2076, opened 2023-03-13
Feat: Hide Increase Stake for SNS CF neurons

lmuntaner:
# Motivation

User should not be able to increase the stake of a Community Fund owned SNS neuron.

NOTE: Functionality is permitted in the backend. But the fix for now is to disable it in the frontend.

# Changes

* Check whether neuron is part of the CF in SnsNeuroInfoStake.svelte to render the SnsIncreaseStakeButton.

# Tests

* New PO SnsNeuronInforStake.
* New PO Button
* Use new PO in SnsNeuronInfoStake test file.
* Add scenario for Incrase Stake neuron belonging and not belonging to CF.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
@@ -22,7 +22,7 @@
   import DisburseSnsButton from "$lib/components/sns-neuron-detail/actions/DisburseSnsButton.svelte";
   import IncreaseSnsDissolveDelayButton from "$lib/components/sns-neuron-detail/actions/IncreaseSnsDissolveDelayButton.svelte";
   import SnsIncreaseStakeButton from "$lib/components/sns-neuron-detail/actions/SnsIncreaseStakeButton.svelte";
-  import TestIdWrapper from "../common/TestIdWrapper.svelte";
+  import TestIdWrapper from "$lib/common/TestIdWrapper.svelte";
 
   const { store }: SelectedSnsNeuronContext =
     getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
@@ -14,6 +14,7 @@
     getSnsNeuronState,
     hasPermissionToDisburse,
     hasPermissionToDissolve,
+    isCommunityFund,
   } from "$lib/utils/sns-neuron.utils";
   import { authStore } from "$lib/stores/auth.store";
   import { NeuronState } from "@dfinity/nns";
@@ -21,6 +22,7 @@
   import DisburseSnsButton from "$lib/components/sns-neuron-detail/actions/DisburseSnsButton.svelte";
   import IncreaseSnsDissolveDelayButton from "$lib/components/sns-neuron-detail/actions/IncreaseSnsDissolveDelayButton.svelte";
   import SnsIncreaseStakeButton from "$lib/components/sns-neuron-detail/actions/SnsIncreaseStakeButton.svelte";
+  import TestIdWrapper from "../common/TestIdWrapper.svelte";
 
   const { store }: SelectedSnsNeuronContext =
     getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
@@ -52,28 +54,37 @@
     nonNullish(neuronState) &&
     [NeuronState.Dissolving, NeuronState.Locked].includes(neuronState) &&
     allowedToDissolve;
+
+  let isCFNeuron = false;
+  $: isCFNeuron = nonNullish(neuron) ? isCommunityFund(neuron) : false;
 </script>
 
-{#if nonNullish(neuron) && nonNullish(neuronState)}
-  <KeyValuePair>
-    <h3 slot="key">{$i18n.neuron_detail.stake}</h3>
-    <SnsNeuronAmount {neuron} slot="value" />
-  </KeyValuePair>
+<TestIdWrapper testId="sns-neuron-info-stake">
+  {#if nonNullish(neuron) && nonNullish(neuronState)}
+    <KeyValuePair>
+      <h3 slot="key" data-tid="sns-neuron-stake">
+        {$i18n.neuron_detail.stake}
+      </h3>
+      <SnsNeuronAmount {neuron} slot="value" />
+    </KeyValuePair>
 
-  <div class="buttons">
-    {#if allowedToDissolve}
-      <IncreaseSnsDissolveDelayButton />
-    {/if}
-    <SnsIncreaseStakeButton />
-    {#if neuronState === NeuronState.Dissolved && allowedToDisburse}
-      <DisburseSnsButton />
-    {:else if canDissolve}
-      <DissolveSnsNeuronButton {neuronState} />
-    {/if}
-  </div>
+    <div class="buttons">
+      {#if allowedToDissolve}
+        <IncreaseSnsDissolveDelayButton />
+      {/if}
+      {#if !isCFNeuron}
+        <SnsIncreaseStakeButton />
+      {/if}
+      {#if neuronState === NeuronState.Dissolved && allowedToDisburse}
+        <DisburseSnsButton />
+      {:else if canDissolve}
+        <DissolveSnsNeuronButton {neuronState} />
+      {/if}
+    </div>
 
-  <Separator />
-{/if}
+    <Separator />
+  {/if}
+</TestIdWrapper>
 
 <style lang="scss">
   @use "../../themes/mixins/section";
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
@@ -22,7 +22,7 @@
   import DisburseSnsButton from "$lib/components/sns-neuron-detail/actions/DisburseSnsButton.svelte";
   import IncreaseSnsDissolveDelayButton from "$lib/components/sns-neuron-detail/actions/IncreaseSnsDissolveDelayButton.svelte";
   import SnsIncreaseStakeButton from "$lib/components/sns-neuron-detail/actions/SnsIncreaseStakeButton.svelte";
-  import TestIdWrapper from "$lib/common/TestIdWrapper.svelte";
+  import TestIdWrapper from "$lib/components/common/TestIdWrapper.svelte";
 
   const { store }: SelectedSnsNeuronContext =
     getContext<SelectedSnsNeuronContext>(SELECTED_SNS_NEURON_CONTEXT_KEY);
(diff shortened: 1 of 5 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
```diff
@@ -14,13 +14,15 @@
...
     hasPermissionToDissolve,
+    isCommunityFund,
   } from "$lib/utils/sns-neuron.utils";
   import { authStore } from "$lib/stores/auth.store";
   import { NeuronState } from "@dfinity/nns";
   import DissolveSnsNeuronButton from "$lib/components/sns-neuron-detail/actions/DissolveSnsNeuronButton.svelte";
   import DisburseSnsButton from "$lib/components/sns-neuron-detail/actions/DisburseSnsButton.svelte";
   import IncreaseSnsDissolveDelayButton from "$lib/components/sns-neuron-detail/actions/IncreaseSnsDissolveDelayButton.svelte";
   import SnsIncreaseStakeButton from "$lib/components/sns-neuron-detail/actions/SnsIncreaseStakeButton.svelte";
+  import TestIdWrapper from "../common/TestIdWrapper.svelte";
```
peterpeterparker, line 25: `$lib/common/...`

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
```diff
@@ -52,28 +54,37 @@
...
+  let isCFNeuron = false;
+  $: isCFNeuron = nonNullish(neuron) ? isCommunityFund(neuron) : false;
 </script>
 
-{#if nonNullish(neuron) && nonNullish(neuronState)}
-  <KeyValuePair>
-    <h3 slot="key">{$i18n.neuron_detail.stake}</h3>
-    <SnsNeuronAmount {neuron} slot="value" />
-  </KeyValuePair>
+<TestIdWrapper testId="sns-neuron-info-stake">
```
peterpeterparker, line 62: this adds an additional `div` in the dom for test purpose only. can you solve the test differently?

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
```diff
@@ -52,28 +54,37 @@
     nonNullish(neuronState) &&
     [NeuronState.Dissolving, NeuronState.Locked].includes(neuronState) &&
     allowedToDissolve;
+
+  let isCFNeuron = false;
```
dskloetd, line 58: I'm used to a rule that abbreviations of multiple letters (such as CF) get camelcased as if they are a single word. Do we have any such (or opposite) rule?

So this would be `isCfNeuron`.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
```diff
@@ -52,28 +54,37 @@
...
+  $: isCFNeuron = nonNullish(neuron) ? isCommunityFund(neuron) : false;
```
dskloetd, line 59: Isn't this the same as 
```
$: isCFNeuron = nonNullish(neuron) && isCommunityFund(neuron);
```
or even
```
$: isCFNeuron = neuron && isCommunityFund(neuron);
```

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronInfoStake.spec.ts
```diff
@@ -37,77 +42,115 @@ describe("SnsNeuronInfoStake", () => {
...
+  it("should render increase state button if neuron doesn't belong to the Community Fund", async () => {
+    const { container } = renderSelectedSnsNeuronContext({
+      Component: SnsNeuronInfoStake,
+      neuron: mockSnsNeuron,
+      reload: jest.fn(),
+    });
+
+    const po = SnsNeuronInfoStakePo.under(container);
+
+    expect(po.hasIncreaseStakeButton()).toBe(true);
```
dskloetd, line 138: Question:
I'm surprised the test doesn't actually click on any of these buttons.
Is there a different test that tests that the buttons actually do anything?

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-obejct.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.root = root;
+  }
+
+  static under(element: Element): SnsNeuronInfoStakePo | null {
+    const el = element.querySelector(`[data-tid=${SnsNeuronInfoStakePo.tid}]`);
+    return el && new SnsNeuronInfoStakePo(el);
+  }
+
+  isContentLoaded() {
+    return nonNullish(this.root.querySelector(`[data-tid=sns-neuron-stake]`));
```
dskloetd, line 22: Getting the stake itself seems like it could be useful as well. Have a separate method for that and call it here?

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-obejct.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  getDisburseButton() {
```
dskloetd, line 25: getDisburseButtonPo? (Po at the end)

(And fix all the other functions that return POs or check the presence of POs.)

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
lmuntaner: Not really if we want to follow the new pattern with the page objects.

@dskloetd do you have a better idea?

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
lmuntaner: thanks!

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-obejct.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  hasDisburseButton() {
+    return nonNullish(this.getDisburseButton());
+  }
+
+  getIncreaseDissolveDelayButton() {
+    return Button.under({
+      element: this.root,
+      testId: "sns-increase-dissolve-delay",
+    });
```
dskloetd, line 37: Having the special `Button.under` function doesn't really seem to make the calling code very short. How about having a `getButton(tid)` in SnsNeuronInfoStakePo so we don't have to pass `this.root` each time.
And then we can probably just use the Button constructor instead of a custum `under` function.

dskloetd (commented):
Some questions and suggestions. Let me know if you disagree with any or prefer to do them in a separate PR.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
lmuntaner: yes, thanks!

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
lmuntaner: Makes sense for me. No rule AFAIK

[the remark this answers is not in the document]
lmuntaner: sure

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronInfoStake.spec.ts
```diff
@@ -3,30 +3,35 @@
...
+} from "@dfinity/sns";
 
 describe("SnsNeuronInfoStake", () => {
-  beforeAll(() =>
+  const data = snsResponsesForLifecycle({
+    lifecycles: [SnsSwapLifecycle.Open],
+    certified: true,
+  });
+  beforeEach(() => {
+    const universe = data[0][0].rootCanisterId;
```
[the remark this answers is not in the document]
lmuntaner: That's not how `snsResponsesForLifecycle` works.

Maybe we can change that in another PR.

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronInfoStake.spec.ts
lmuntaner: Correct. There is a test for the button.

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-obejct.ts
lmuntaner: yes, thanks!

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-obejct.ts
lmuntaner: sure

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-obejct.ts
lmuntaner: sure

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
dskloetd: I don't but I also don't see a problem with it. It makes the DOM also much easier to understand for manual inspection.

peterpeterparker (approved):
LGTM, thanks

Please improve the test afterwards, as commented, adding `div` for test purpose should be avoided in my opinion because adding DOM elements for test purpose has no value for the end users and will ultimately lower the performance.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
```diff
@@ -34,35 +34,35 @@
...
     [NeuronState.Dissolving, NeuronState.Locked].includes(neuronState) &&
     allowedToDissolve;
 
-  let isCFNeuron = false;
-  $: isCFNeuron = nonNullish(neuron) ? isCommunityFund(neuron) : false;
+  let isIncreaseStakeAllowed = false;
+  $: isIncreaseStakeAllowed = nonNullish(neuron) && !isCommunityFund(neuron);
 </script>
 
 <TestIdWrapper testId="sns-neuron-info-stake">
```
dskloetd, line 62: FYI, I've been using the "-component" suffix for these tids that wrap the entire component.

dskloetd (approved):

dskloetd:
> Please improve the test afterwards, as commented, adding div for test purpose should be avoided in my opinion because adding DOM elements for test purpose has no value for the end users and will ultimately lower the performance.

The tids themselves are already only for testing. This sounds like premature optimization to hurt readability. There much bigger performance issues to solve if we care about performance.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
lmuntaner: noted for next PR if you don't mind

peterpeterparker:
> The tids themselves are already only for testing. 
I was already against so we go a step further

> This sounds like premature optimization to hurt readability. There much bigger performance issues to solve if we care about performance.
Not agree. Made my point above. Following is crap.

```
<div data-tid>
  <div data-tid>
    <div data-tid>
      <div data-tid>
        <div data-tid>
          <div data-tid>
            <div data-tid>
              <div data-tid>
                <div data-tid>
                  <div data-tid>
                    <div data-tid>
                      <div data-tid>
                        <div data-tid>
                          <div data-tid>
                            <div data-tid>
                              etc.
```

PR #2077, opened 2023-03-13
Introducing fake api modules for easier testing

dskloetd:
# Motivation

Setting up mocks for api calls during tests can be tedious and the result is often not easy to read.

A fake holds state and responds to calls in a consistent way based on the internal state, which is set up for the test.
It can also provide sane defaults for values that don't matter during the test.
This means the test can just set up the state it cares about and the fake will take care of the rest.

# Changes

1. Add partial fakes for governance.api, sns-aggregator.api and sns-ledger.api.
  * Meaning, not all of the functions on these apis are implemented and some just return empty or default values which can't yet be controlled by the test.
2. Add a fake for sns-governance.api which also fakes some functions on sns,api, which should really  be moved to sns-governance.api because they are backed by the same state.
3. Update NeuronDetail.spec.ts and Neurons.spec.ts to use the new fakes.
4. Update module.test-utils.ts to allow installing partial implementations of modules in addition to blocking calls to the functions for which no implementations are provided.
5. Update tsconfig.json to allow using "$tests" imports in fakes.
6. 

# Tests

Tests only.

### DIFF
--- a/frontend/src/tests/fakes/governance-api.fake.ts
+++ b/frontend/src/tests/fakes/governance-api.fake.ts
@@ -1,0 +1,79 @@
+import type { ApiQueryParams } from "$lib/api/governance.api";
+import { mockIdentity } from "$tests/mocks/auth.store.mock";
+import { mockNeuron } from "$tests/mocks/neurons.mock";
+import { installImplAndBlockRest } from "$tests/utils/module.test-utils";
+import type { Identity } from "@dfinity/agent";
+import type { KnownNeuron, NeuronInfo } from "@dfinity/nns";
+import { isNullish } from "@dfinity/utils";
+
+const modulePath = "$lib/api/governance.api";
+const impl = {
+  queryNeurons,
+  queryKnownNeurons,
+};
+
+//////////////////////////////////////////////
+// State and helpers for fake implementations:
+//////////////////////////////////////////////
+
+const neurons: Map<string, NeuronInfo[]> = new Map();
+
+const mapKey = (identity: Identity) => identity.getPrincipal().toText();
+
+const getNeurons = (identity: Identity) => {
+  const key = mapKey(identity);
+  let neuronList = neurons.get(key);
+  if (isNullish(neuronList)) {
+    neuronList = [];
+    neurons.set(key, neuronList);
+  }
+  return neuronList;
+};
+
+////////////////////////
+// Fake implementations:
+////////////////////////
+
+async function queryNeurons({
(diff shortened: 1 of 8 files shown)


frontend/src/tests/fakes/governance-api.fake.ts
```diff
@@ -0,0 +1,79 @@
+import type { ApiQueryParams } from "$lib/api/governance.api";
+import { mockIdentity } from "$tests/mocks/auth.store.mock";
+import { mockNeuron } from "$tests/mocks/neurons.mock";
+import { installImplAndBlockRest } from "$tests/utils/module.test-utils";
+import type { Identity } from "@dfinity/agent";
+import type { KnownNeuron, NeuronInfo } from "@dfinity/nns";
+import { isNullish } from "@dfinity/utils";
+
+const modulePath = "$lib/api/governance.api";
+const impl = {
```
lmuntaner, line 10: What do you mean by `impl`? Could you add a whole name?

frontend/src/tests/fakes/governance-api.fake.ts
```diff
@@ -0,0 +1,79 @@
...
+  queryNeurons,
+  queryKnownNeurons,
+};
+
+//////////////////////////////////////////////
+// State and helpers for fake implementations:
+//////////////////////////////////////////////
+
+const neurons: Map<string, NeuronInfo[]> = new Map();
```
lmuntaner, line 19: Could you add in a comment what the `string` represents in the map?

lmuntaner (approved):
I love this!

frontend/src/tests/fakes/governance-api.fake.ts
dskloetd: Which of these would you prefer?
1. implementation
2. fakeImplementation
3. fakeGovernanceApiImplementation
4. implementedFunctions
5. implementedGovernanceApiFunctions
6. something else?

lmuntaner: `implementedFunctions` sounds good to me

frontend/src/tests/fakes/governance-api.fake.ts
dskloetd: Done

frontend/src/tests/fakes/governance-api.fake.ts
dskloetd: Done

PR #2078, opened 2023-03-13
Release: 2023-03-13

lmuntaner:
# Motivation

Release candidate for 2023-03-13.

# Changes

* Changelog
* Update versions in cargo.toml and package.json

# Tests

Not applicable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## Proposal 112386
-* Do not allow increasing stake for CF SNS neurons.
+* Remove button to increase stake for CF SNS neurons.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Proposal 112386
+* Do not allow increasing stake for CF SNS neurons.
```
dskloetd, line 4: Should we say we don't allow it or that we remove the button?
From the thread it sounded more that we remove the button to avoid confusion rather than to really not allow it.

peterpeterparker (approved):
Thx

PR #2081, opened 2023-03-13
Faster frontend formatting?

bitdivine:
# Motivation
Formatting is surprisingly slow.  For me, on main:
```
time ./scripts/fmt
[snip]
real	0m52,322s
```

# Changes
- For e2e tests, don't format `node_modules`
- Use a parallel wrapper for prettier
- Format sh, js and rust all in parallel.

# Tests
Speedup will depend on your platform.  For me this is just over 3x as fast:
```
./scripts/fmt
[snip]
✔ Reformatted 0 / 41 files...
✔ Reformatted 0 / 707 files...

real	0m17,563s
```

Formatting just the front end:
```
Before: 0m39,819s
After: 0m14,766s
```


frontend/package.json
```diff
@@ -13,8 +13,8 @@
     "preview": "vite preview",
     "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && npm run lint",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
-    "lint": "prettier --plugin-search-dir . --check . && eslint --max-warnings 0 .",
-    "format": "prettier --plugin-search-dir . --write .",
+    "lint": "pprettier --write './**/*.{ts,tsx,json,scss}' --check . && eslint --max-warnings 0 .",
+    "format": "pprettier --write './**/*.{ts,tsx,json,scss}'",
```
bitdivine, line 17: Note: pprettier doesn't seem to support a custom plugin search dir.  But then, we don't seem to use any plugins for formatting, for other things yes but not for formatting.

e2e-tests/.prettierignore
```diff
@@ -0,0 +1 @@
+node_modules
```
bitdivine, line 1: No need to format code in node_modules.  Note: This exception is already in place for for the nns-dapp frontend.

scripts/fmt
```diff
@@ -3,9 +3,11 @@ set -euxo pipefail
 cd "$(dirname "$(realpath "$0")")/.."
 
 set -x
+time xargs -P8 -I{} bash -c "{}" <<EOF
```
bitdivine, line 6: This runs all the formatting in parallel.  Depending on how many CPU cores you have this may help, or not.  For me, this causes a drop from ~34s to ~17s.

peterpeterparker (approved):
rubber stamp

side note: some day we should maybe transform the project in a js mono repo, that way we would not have to install the same lib in two distinct node_module but just once

PR #2082, opened 2023-03-13
GIX-1401: Remove syncAcconts when not needed

lmuntaner:
# Motivation

Avoid unnecessary update calls.

When reloading an account we don't care about updating all the other accounts.

# Changes

* Use `loadBalance` in NnsWallet.svelte when reloading instead of `syncAccounts`.

# Tests

* Change test accordingly.


lmuntaner:

frontend/src/tests/lib/pages/NnsWallet.spec.ts
```diff
@@ -214,14 +214,21 @@ describe("NnsWallet", () => {
 
       await waitModalIntroEnd({ container, selector: modalToolbarSelector });
 
+      await waitFor(() =>
+        expect(accountsApi.getTransactions).toBeCalledTimes(2)
```
dskloetd, line 218: I'm curious about your thoughts but I'm thinking it doesn't really matter whether the API is called.
What matters is that the transactions and balance are updated.
Now it happens to be the case that it's not possible to update the transactions and balance without calling the API, but it is possible to call the API without updating the transactions and balance.
So shouldn't we assert that the UI is updated rather than that the API is called?

Now I will not block you because this pattern seems to be used throughout many tests but it really bothers me so I'm curious what you think.

frontend/src/tests/lib/pages/NnsWallet.spec.ts
```diff
@@ -214,14 +214,21 @@ describe("NnsWallet", () => {
...
       fireEvent.click(
         getByTestId("reload-receive-account") as HTMLButtonElement
       );
 
-      await waitFor(() => expect(spySyncAccounts).toHaveBeenCalled());
-      expect(spyGetAccountTransactions).toHaveBeenCalled();
+      await waitFor(() =>
+        expect(ledgerApi.queryAccountBalance).toBeCalledTimes(2)
+      );
+      expect(accountsApi.getTransactions).toBeCalledTimes(4);
```
dskloetd, line 231: Can the expectations be in the same order here as above?
I was a bit confused to see 2 and 0 and then 2 and 4, rather than 4 and 2.

dskloetd (approved):

frontend/src/tests/lib/pages/NnsWallet.spec.ts
lmuntaner: That's a good point.

I agree with you. Mocking the services and store layers this test didn't make sense.

But if we mock the api layer, it makes sense to prepare it in a way that the "second" call changes the value and it's rendered in the UI. Which is what we should expect.

frontend/src/tests/lib/pages/NnsWallet.spec.ts
lmuntaner: sure

PR #2083, opened 2023-03-13
feat: display ckBTC to BTC estimated time

peterpeterparker:
# Motivation

Because the transfer on BTC network is slow, display a static estimation of the time it can least.

# Changes

- display 30min estimated time on review step and in success msg

# Screenshots

Capture d’écran 2023-03-13 à 15 08 20

Capture d’écran 2023-03-13 à 15 08 25


frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
```diff
@@ -79,6 +83,15 @@
     </div>
 
     <slot name="additional-info" />
+
+    {#if networkBtc}
```
lmuntaner, line 87: Why not adding it in the `additional-info` instead?

This way the `TransactionReview` doesn't know about this specifics of ckbtc

lmuntaner (commented):
A question

frontend/src/lib/modals/accounts/NewTransaction/TransactionReview.svelte
peterpeterparker: Sure why not, will try

peterpeterparker: Done, thanks

lmuntaner (approved):
Thanks!

PR #2084, opened 2023-03-13
Remove use of non-existing user-drag property which causes a warning in svelte-check

dskloetd:
# Motivation

Running `svelte-check` which is part of `npm run check` currently gives 1 warning:
```
frontend/src/lib/components/ui/Logo.svelte:28:5
Warn: Unknown property: 'user-drag' (scss)
    -webkit-user-drag: none;
    user-drag: none;
  }
```

According to https://caniuse.com/webkit-user-drag:

> There is currently no unprefixed `user-drag` or other version of the property implemented in browsers or on a standards track as the HTML `draggable` attribute/property is the preferred solution.

# Changes

1. Remove `user-drag: none;` from Logo.svelte.
2. Add `draggable="false"` to the `` tag in Logo.svelte.
3. Add `--fail-on-warnings` to `npm run check` to prevent new warnings.

# Tests

`npm run check`
Manually tried dragging an affected logo in both Chrome and Firefox and was not able to drag it.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -11,7 +11,7 @@
     "dev": "npm run i18n && vite dev",
     "build": "npm run i18n && vite build && npm run build:post-process",
     "preview": "vite preview",
-    "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json && npm run lint",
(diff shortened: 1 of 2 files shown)


peterpeterparker (approved):
LGTM, thx

PR #2085, opened 2023-03-13
fix: display IC on canisters view

peterpeterparker:
# Motivation

If user select a Sns on "My Accounts" and then navigate to "My Canisters", the Sns name instead of IC will be displayed in the title of the page.

# Changes

- if `displayUniverse` is set to false, display IC


lmuntaner (approved):
Thanks

PR #2086, opened 2023-03-13
Fix: Message on auto stake

lmuntaner:
# Motivation

Success messages on enabling auto stake were the other way around.

It said: "Automatically staking new maturity." when disabling it.
And: "Stopped automatically staking new maturity." when enabling it.

# Changes

* Change which message to use in NnsAutoStakeMaturityModal and SnsAutoStakeMaturityModal.

# Tests

* Add tests that proper message is added to toastsStore.

### DIFF
--- a/frontend/src/lib/modals/neurons/NnsAutoStakeMaturityModal.svelte
+++ b/frontend/src/lib/modals/neurons/NnsAutoStakeMaturityModal.svelte
@@ -14,14 +14,15 @@
 
   const autoStake = async () => {
     startBusy({ initiator: "auto-stake-maturity" });
+    const labelKey = `neuron_detail.auto_stake_maturity_${
+      hasAutoStakeOn ? "off" : "on"
+    }_success`;
 
     const { success } = await toggleAutoStakeMaturity(neuron);
 
     if (success) {
       toastsSuccess({
-        labelKey: `neuron_detail.auto_stake_maturity_${
-          hasAutoStakeOn ? "off" : "on"
-        }_success`,
+        labelKey,
       });
     }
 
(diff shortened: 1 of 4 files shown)


lmuntaner:

frontend/src/lib/modals/neurons/NnsAutoStakeMaturityModal.svelte
```diff
@@ -20,7 +20,7 @@
     if (success) {
       toastsSuccess({
         labelKey: `neuron_detail.auto_stake_maturity_${
-          hasAutoStakeOn ? "on" : "off"
+          hasAutoStakeOn ? "off" : "on"
```
peterpeterparker, line 23: is it possible that we are actually hitting here a race condition? 

if yes, would it be worth to actually clone the `hasAutoStakeOn` to a tmp variable or declare the label before processing. that way we are sure the auto subscriber was not reevaluated before displaying the label?


```
let hasAutoStakeOn: boolean;
// <--- this is maybe the root cause ? -->
$: hasAutoStakeOn = hasAutoStakeMaturityOn(neuron);

  const autoStake = async () => {
    startBusy({ initiator: "auto-stake-maturity" });
    
    // Evaluate before
    const labelKey = `neuron_detail.auto_stake_maturity_${
          hasAutoStakeOn ? "off" : "on"
        }_success`

    const { success } = await toggleAutoStakeMaturity(neuron);

    if (success) {
      toastsSuccess({
        labelKey}});
    }
```

Not sure, makes sense?

lmuntaner: Yes, it makes sense. Better be safe.

lmuntaner:

peterpeterparker (approved):
LGTM, thanks

Idea of potential refactoring for the future: extract `autoStake` to a service and path `toggleAutoStakeMaturity` as parameter

PR #2088, opened 2023-03-14
Downgrade/upgrade test for aggregator

bitdivine:
# Motivation
Basic test that the canister can be downgraded and upgraded without bricking the canister.

Note: This does not catch subtle issues such as data corruption when deserializing and reserializing.  It just checks whether the canister can transition and is still alive after each transition.

Note: The test does a downgrade then an upgrade rather than the other way around as it is faster.  When testing, the new code is typically already installed and running, so downgrade+upgrade requires two installations, whereas upgrade downgrade needs four: downgrade-upgrade-downgrade-upgrade

# Changes
- Add a script to perform the downgrade-upgrade test
- Exercise the script in CI

# Tests
See the new tests in CI

### DIFF
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -164,6 +164,7 @@ jobs:
           }
       - name: Downgrade to prod and upgrade back again
         run: |
+          set -euxo pipefail
           git fetch --depth 1 origin tag aggregator-prod
           if git diff tags/aggregator-prod rs/sns_aggregator | grep -q .
           then ./scripts/sns/aggregator/downgrade-upgrade-test -w sns_aggregator.wasm
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -166,7 +166,8 @@ jobs:
         run: |
           set -euxo pipefail
           git fetch --depth 1 origin tag aggregator-prod
-          if git diff tags/aggregator-prod rs/sns_aggregator | grep -q .
+          diff="$(git diff tags/aggregator-prod rs/sns_aggregator)"
+          if test -n "${diff:-}"
           then ./scripts/sns/aggregator/downgrade-upgrade-test -w sns_aggregator.wasm
           else echo "Skipping test as there are no relevant code changes"
           fi
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -134,7 +134,7 @@ jobs:
             cat ,logs
             exit 1
           }
-      - name: Upgrade the aggregator with a slow refresh rate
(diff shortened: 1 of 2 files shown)


scripts/sns/aggregator/downgrade-upgrade-test
bitdivine, line 1: Cool, I really can comment on files.

So, this file is the downgrade-upgrade tester.

.github/workflows/aggregator.yaml
```diff
@@ -162,6 +162,13 @@ jobs:
           (( expect == actual ))  || {
             echo ERROR: Expected to have $expect SNS in the aggregator upstream data but found $actual.
           }
+      - name: Downgrade to prod and upgrade back again
+        run: |
+          git fetch --depth 1 origin tag aggregator-prod
+          if git diff tags/aggregator-prod rs/sns_aggregator | grep -q .
+          then ./scripts/sns/aggregator/downgrade-upgrade-test -w sns_aggregator.wasm
+          else echo "Skipping test as there are no relevant code changes"
+          fi
```
bitdivine, line 171: Note: Maybe the whole aggregator workflow can be skipped if there are no changes to main.

scripts/sns/aggregator/downgrade-upgrade-test
```diff
@@ -0,0 +1,69 @@
...
+if dfx canister id sns_aggregator >/dev/null 2>/dev/null; then
+  upgrade_canister "$CURRENT_WASM"
+else
+  dfx canister create sns_aggregator
+  dfx canister install sns_aggregator --wasm "$CURRENT_WASM"
+fi
+echo "Checking that the current wasm is healthy..."
+verify_healthy
+echo "Downgrading to the prod wasm..."
+upgrade_nnsdapp "$PROD_WASM"
```
lmuntaner, line 62: This says `upgrade_nnsdapp`, don't you mean `upgrade_canister`?

scripts/sns/aggregator/downgrade-upgrade-test
```diff
@@ -0,0 +1,69 @@
...
+echo "Checking that the rollback is healthy..."
+verify_healthy
+echo "Rolling forwards..."
+upgrade_nnsdapp "$CURRENT_WASM"
```
lmuntaner, line 66: ditto

lmuntaner (commented):
A couple of questions

scripts/sns/aggregator/downgrade-upgrade-test
bitdivine: Yes, thank you!

lmuntaner (approved):
Thanks!

PR #2089, opened 2023-03-14
Move some functions from sns.api to sns-governance.api

dskloetd:
# Motivation

querySnsNeurons, querySnsNeuron and getSnsNeurons are currently defined in frontend/src/lib/api/sns.api.ts.
We also have frontend/src/lib/api/sns-governance.api.ts.
Based on [sns.wrapper.ts](https://github.com/dfinity/ic-js/blob/5866c4463d0acb4bb06b520736dd058d8e0bfe5e/packages/sns/src/sns.wrapper.ts#L126), those 3 functions use only the sns governance canister and do therefore belong in sns-governance.api.ts.

In frontend/src/lib/api/sns.api.ts, we also have stakeNeuron() and increaseStakeNeuron() but because they use both the sns governance canister and the sns ledger canister, it's less obvious that they belong in sns-governance.api.ts. Let me know if you want me to move those as well.

# Changes

1. Move querySnsNeurons, querySnsNeuron and getSnsNeurons from sns.api.ts to sns-governance.api.ts.
2. Move their tests from sns.api.spec.ts to sns-governance.api.spec.ts.
3. Update callers of these functions to import them from the new location.
4. Update the tests of callers of these functions to mock them on sns-governance.api.
5. Update sns-governance-api.fake.ts to fake these functions on sns-governance.api.ts and no longer mock out sns.api.
6. Update module.test-utils.ts to throw an error if a fake implementedFunction does not have a real equivalent on the module it's supposed to fake.

# Tests

Manually checked that sns neurons are still listed and their details can be seen.


lmuntaner (approved):
Thanks!

PR #2092, opened 2023-03-14
feat: read destination address with an integrated QR code reader

peterpeterparker:
# Motivation

Scan destination address with an integrated QR code reader.

# PRs

- [x] https://github.com/dfinity/gix-components/pull/171

# Notes

Because the web worker is bundled within the gix-cmp, both `npm run dev / test` are not able to loads it. This PR amend this limitation given the fact that most important test on the QR code reader are performed with e2e test in gix-component. We might try to find a solution in further PRs, did not find despite having tried quite a lot...

# Changes

- integrate new `<QRReaderModal />` component
- add a button to open QR code in `AddressInput`
- implement a new step within `TransactionModal`


frontend/src/lib/components/accounts/NnsAddress.svelte
```diff
@@ -9,7 +9,11 @@
 
 <article>
   <form on:submit|preventDefault>
-    <AddressInput bind:address rootCanisterId={OWN_CANISTER_ID} />
+    <AddressInput
+      bind:address
+      rootCanisterId={OWN_CANISTER_ID}
+      qrCode={false}
```
lmuntaner, line 15: Do we set it as false because it will never be supported or because we don't want to release it yet?

frontend/src/tests/lib/modals/accounts/TransactionModal.spec.ts
```diff
@@ -420,4 +425,42 @@ describe("TransactionModal", () => {
       expect(() => getByTestId("close-modal")).toThrow();
     });
   });
+
+  describe("qr code", () => {
```
lmuntaner, line 429: Is it possible to add a test that the address input is filed after the qr code step?

lmuntaner (approved):
LGTM!

Just a comment on a possible test that would be great to have.

frontend/src/tests/lib/modals/accounts/TransactionModal.spec.ts
peterpeterparker: No really unfortunaly, I've tried hard too. The QR code as describe throw an error in test or dev at the moment. 

That said, the e2e test in gix-cmp does that - read qr code, return value and set value to a textarea.

frontend/src/lib/components/accounts/NnsAddress.svelte
peterpeterparker: I set `false` because the business requirement are scoped to the destination address only at the moment.

lmuntaner: Isn't that a destination address? But just for NNS...

lmuntaner: Ahhh, true.

Legacy of old modals...

PR #2094, opened 2023-03-14
Make e2e-tests/scripts/update-chromedriver work on Mac.

dskloetd:
# Motivation

e2e-tests/scripts/update-chromedriver currently doesn't work on Mac.

# Changes

1. Call `/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --version` if `google-chrome` doesn't exist.
2. Use `[[:space:]]` in sed instead of `\s`. Supposedly the latter only work on GNU sed.

# Tests

1. Used it manually on Mac.
2. Tested the part to determine chrome_major_version on my personal Linux laptop.

### DIFF
--- a/e2e-tests/scripts/update-chromedriver
+++ b/e2e-tests/scripts/update-chromedriver
@@ -4,10 +4,10 @@ set -euxo pipefail
 if command -v google-chrome; then
   chrome_version=$(google-chrome --version)
 else
-  chrome_version=$(/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --version)
+  chrome_version="$(osascript -e 'get version of application "Google Chrome"')"
 fi
 
-chrome_major_version="$(echo "$chrome_version" | sed -E 's/.*[[:space:]]([0-9]+).*/\1/g')"
+chrome_major_version="$(echo "$chrome_version" | sed -E 's/[^0-9]*([0-9]+).*/\1/g')"
 
 cd "$(dirname "$0")/.."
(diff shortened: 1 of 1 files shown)


e2e-tests/scripts/update-chromedriver
```diff
@@ -1,4 +1,13 @@
 #!/usr/bin/env bash
 set -euxo pipefail
+
+if command -v google-chrome; then
+  chrome_version=$(google-chrome --version)
+else
+  chrome_version=$(/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --version)
```
bitdivine, line 7: Try this:
```
osascript -e 'get version of application "Google Chrome"'
```
Might be better than assuming the default installation path.

bitdivine: Update:  Tested on my intel mac.  It works there.  Not tested on M1/M2.

dskloetd: Thanks. It works on my M1 mac as well.
Is this only for Mac or could this even replace the whole `if`?

dskloetd: Done

bitdivine: Only for mac.  osascript is applescript.  Played with it many years ago, it's quite a neat tool for running apple programs by remote control.  But it's also quite limited.

bitdivine (approved):

PR #2080, opened 2023-03-13
feat: ckBTC to Btc validate amount

peterpeterparker:
# Motivation

Add two validations regarding the amount the user input in the transaction wizard if ckBTC are converted to Btc.

"Continue" disabled if (one or the other):

- ckBTC fee + Btc estimated fee + amount > available funds
- amount is lower than minimal amount that can be converted by the minter

# Changes

# Tests

PR #2095, opened 2023-03-15
refactor: validateAmount parameters

peterpeterparker:
# Motivation

Instead of exposing the selected account in the transaction modal we decided to pass it as a parameters of the validation function.

Follow-up of #2080.


lmuntaner (approved):
Thanks!

PR #2096, opened 2023-03-15
chore: auth store better typed

peterpeterparker:
# Motivation

Improve `auth.store` types to avoid Webstorm to displays issue for types it cannot interfer.


lmuntaner (approved):
Thanks!

PR #2097, opened 2023-03-15
GIX-1336: Refactor SNS Swap Participation

lmuntaner:
# Motivation

Simplify the flow to participate in SNS swap.

Remove the `keepPolling` state.

# Changes

* Add cancel `cancelPollOpenTicket` to the sns swap services.
* Use `cancelPollOpenTicket` on destroy of the ParticipateButton.
* Remove `keepPolling` from the snsTicketsStore.
* Remove all usages of `keepPolling` in different utils and services. It's not needed anymore.

# Tests

* Fix tests after refactor.
* Add a test to check that `cancelPollOpenTicket` is called on destroy of ParticipateButton.

### DIFF
--- a/frontend/src/lib/components/project-detail/ParticipateButton.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateButton.svelte
@@ -139,6 +139,8 @@
       rootCanisterId,
       ticket: undefined,
     });
+    // TODO: Improve cancellatoin of actions onDestroy
+    // The polling was triggered by `restoreSnsSaleParticipation` call and needs to be canceled explicitly.
     cancelPollGetOpenTicket();
 
     // Hide toasts when moving away from the page
--- a/frontend/src/lib/components/project-detail/ParticipateButton.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateButton.svelte
@@ -19,6 +19,7 @@
   import { isNullish, nonNullish } from "@dfinity/utils";
   import { snsTicketsStore } from "$lib/stores/sns-tickets.store";
   import {
+    cancelPollOpenTicket,
     hidePollingToast,
     restoreSnsSaleParticipation,
   } from "$lib/services/sns-sale.services";
@@ -96,8 +97,6 @@
       // Typescript guard, user commitment cannot be undefined here
       return;
     }
-    snsTicketsStore.enablePolling(rootCanisterId);
-
     loadingTicketRootCanisterId = rootCanisterId.toText();
 
     const updateProgress = (step: SaleStep) => (progressStep = step);
@@ -139,8 +138,8 @@
     snsTicketsStore.setTicket({
       rootCanisterId,
       ticket: undefined,
-      keepPolling: false,
     });
+    cancelPollOpenTicket();
 
     // Hide toasts when moving away from the page
     hidePollingToast();
--- a/frontend/src/lib/components/project-detail/ParticipateButton.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateButton.svelte
@@ -19,7 +19,7 @@
   import { isNullish, nonNullish } from "@dfinity/utils";
   import { snsTicketsStore } from "$lib/stores/sns-tickets.store";
   import {
-    cancelPollOpenTicket,
+    cancelPollGetOpenTicket,
     hidePollingToast,
     restoreSnsSaleParticipation,
   } from "$lib/services/sns-sale.services";
@@ -139,7 +139,7 @@
       rootCanisterId,
       ticket: undefined,
     });
-    cancelPollOpenTicket();
+    cancelPollGetOpenTicket();
 
     // Hide toasts when moving away from the page
     hidePollingToast();
--- a/frontend/src/lib/services/sns-sale.services.ts
+++ b/frontend/src/lib/services/sns-sale.services.ts
@@ -14,7 +14,7 @@ import {
 } from "$lib/derived/sns/sns-projects.derived";
 import { loadBalance } from "$lib/services/accounts.services";
 import { getCurrentIdentity } from "$lib/services/auth.services";
-import { toastsError, toastsShow } from "$lib/stores/toasts.store";
+import { toastsError, toastsHide, toastsShow } from "$lib/stores/toasts.store";
 import { transactionsFeesStore } from "$lib/stores/transaction-fees.store";
 import type { Account } from "$lib/types/account";
 import { ApiErrorKey } from "$lib/types/api.errors";
@@ -35,7 +35,6 @@ import {
   pollingLimit,
 } from "$lib/utils/utils";
 import type { Identity } from "@dfinity/agent";
-import { toastsStore } from "@dfinity/gix-components";
 import {
   ICPToken,
   InsufficientFundsError,
@@ -77,7 +76,7 @@ import { formatToken } from "../utils/token.utils";
 let toastId: symbol | undefined;
 export const hidePollingToast = (): void => {
   if (nonNullish(toastId)) {
-    toastsStore.hide(toastId);
+    toastsHide(toastId);
     toastId = undefined;
   }
 };
@@ -103,9 +102,9 @@ const WAIT_FOR_TICKET_MILLIS = SALE_PARTICIPATION_RETRY_SECONDS * 1_000;
 // TODO: Solve problem with importing from sns.constants.ts
 const MAX_ATTEMPS_FOR_TICKET = 50;
 const SALE_FAILURES_BEFORE_HIGHlOAD_MESSAGE = 6;
-const pollOpenTicketId = Symbol("poll-open-ticket");
+const pollGetOpenTicketId = Symbol("poll-open-ticket");
 // Export for testing purposes
-export const cancelPollOpenTicket = () => cancelPoll(pollOpenTicketId);
+export const cancelPollGetOpenTicket = () => cancelPoll(pollGetOpenTicketId);
 const pollGetOpenTicket = async ({
   swapCanisterId,
   identity,
@@ -131,7 +130,7 @@ const pollGetOpenTicket = async ({
       millisecondsToWait: WAIT_FOR_TICKET_MILLIS,
       maxAttempts,
       useExponentialBackoff: true,
-      pollId: pollOpenTicketId,
+      pollId: pollGetOpenTicketId,
       failuresBeforeHighLoadMessage: SALE_FAILURES_BEFORE_HIGHlOAD_MESSAGE,
     });
   } catch (error: unknown) {
@@ -191,6 +190,7 @@ export const loadOpenTicket = async ({
   } catch (err) {
     // Don't show error if polling was cancelled
     if (pollingCancelled(err)) {
+      hidePollingToast();
       return;
     }
 
--- a/frontend/src/lib/services/sns-sale.services.ts
+++ b/frontend/src/lib/services/sns-sale.services.ts
@@ -190,7 +190,6 @@ export const loadOpenTicket = async ({
   } catch (err) {
     // Don't show error if polling was cancelled
     if (pollingCancelled(err)) {
-      hidePollingToast();
       return;
     }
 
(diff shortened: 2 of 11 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/components/project-detail/ParticipateButton.spec.ts
```diff
@@ -235,6 +237,20 @@ describe("ParticipateButton", () => {
...
+        summary: mockSnsFullProject.summary,
+        swapCommitment: mockSnsFullProject.swapCommitment as SnsSwapCommitment,
+        Component: ParticipateButton,
+      });
+
+      expect(cancelPollOpenTicket).not.toBeCalled();
+
+      unmount();
+
+      expect(cancelPollOpenTicket).toBeCalled();
```
lmuntaner, line 252: Ideally I would test the call to the api, but this PR has already enough changes and it's no so fast to review.

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -255,8 +255,8 @@ export const hasOpenTicketInProcess = ({
     return { status: "none" };
   }
 
-  // `undefined` means that we could still be polling for the ticket.
-  return { status: projectTicketData.keepPolling ? "polling" : "none" };
+  // As long as we don't have a known state, we assume we're fetching the data.
```
dskloetd, line 258: If polling was canceled, this is not true, right?
I'm trying to understand why previously we could have 2 possible states here but now only 1.

frontend/src/tests/lib/components/project-detail/ParticipateButton.spec.ts
dskloetd: Add a TODO?

frontend/src/tests/lib/components/project-detail/ProjectStatusSection.spec.ts
```diff
@@ -41,12 +45,14 @@ describe("ProjectStatusSection", () => {
...
   });
 
-  it("should render project participate button", () => {
+  it("should render project participate button", async () => {
     const { queryByTestId } = renderContextCmp({
       summary: mockSnsFullProject.summary,
       swapCommitment: mockSnsFullProject.swapCommitment as SnsSwapCommitment,
       Component: ProjectStatusSection,
     });
+    // Wait for the api call to return no ticket
```
dskloetd, line 54: What do you mean by "return no ticket". Did you mean to add an expectation on the return value?

frontend/src/tests/lib/stores/sns-ticket.store.spec.ts
```diff
@@ -23,51 +23,13 @@ describe("snsTicketsStore", () => {
...
-  });
-
   it("should set no-ticket for a project", () => {
     snsTicketsStore.setNoTicket(mockPrincipal);
 
     const $snsTicketsStore = get(snsTicketsStore);
     expect($snsTicketsStore[mockPrincipal.toText()].ticket).toBeNull();
   });
 
-  it("set no-ticket for a project keeps polling from before", () => {
```
dskloetd, line 54: Why not keep this test but test that it keeps calling the API after setNoTicket (if that's what we want)?

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -139,8 +138,8 @@
     snsTicketsStore.setTicket({
       rootCanisterId,
       ticket: undefined,
-      keepPolling: false,
     });
+    cancelPollOpenTicket();
```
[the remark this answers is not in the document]
lmuntaner: It does call `pollGetOpenTicket` through the `restoreParticipation`.

I agree the code could be improved, but this PR is about removing the `keepPolling` state.

Before it was calling setting `keepPolling: false` on destroy. Therefore, I'm still doing the same functionality, but different way.

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -77,43 +82,36 @@ export const hidePollingToast = (): void => {
...
+  return false;
+};
 
 const WAIT_FOR_TICKET_MILLIS = SALE_PARTICIPATION_RETRY_SECONDS * 1_000;
 // TODO: Solve problem with importing from sns.constants.ts
 const MAX_ATTEMPS_FOR_TICKET = 50;
 const SALE_FAILURES_BEFORE_HIGHlOAD_MESSAGE = 6;
+const pollOpenTicketId = Symbol("poll-open-ticket");
 // Export for testing purposes
+export const cancelPollOpenTicket = () => cancelPoll(pollOpenTicketId);
```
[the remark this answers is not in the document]
lmuntaner: Yes, it makes sense

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -96,8 +97,6 @@
       // Typescript guard, user commitment cannot be undefined here
       return;
     }
-    snsTicketsStore.enablePolling(rootCanisterId);
```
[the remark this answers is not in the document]
lmuntaner: Before, the store could have `keepPolling: true` and we needed it to change the state to make sure the polling was allowed.

We always want to poll when users visits the page.

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -139,8 +138,8 @@
     snsTicketsStore.setTicket({
```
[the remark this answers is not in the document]
lmuntaner: Actually, the idea right now is that the participation flow will also stop when you navigate to other pages.

Setting the ticket to `undefined` doesn't enforce stopping the flow always, it depends on where the flow is.

But yes, in another PR we can give a second thought to what should happen if the user navigates away.

For now, this is needed to have the same behavior.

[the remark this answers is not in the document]
lmuntaner: And, no, we don't need an `onMount` because we trigger the `updateTicket` on init as well, depending on some variables and data:

```
  $: if (
    lifecycle === SnsSwapLifecycle.Open &&
    isSignedIn($authStore.identity) &&
    nonNullish(userCommitment) &&
    nonNullish(swapCanisterId)
  ) {
    updateTicket(swapCanisterId);
  }
```

It is kind of a conditional `onMount`

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -191,10 +189,8 @@ export const loadOpenTicket = async ({
 
     logWithTimestamp("[sale]loadOpenTicket:", ticket);
   } catch (err) {
-    const store = get(snsTicketsStore)[rootCanisterId.toText()];
-    // Do not show errors if the user has stopped polling.
-    if (!store?.keepPolling) {
-      hidePollingToast();
+    // Don't show error if polling was cancelled
+    if (pollingCancelled(err)) {
```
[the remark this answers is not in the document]
lmuntaner: Good point, I'll take a look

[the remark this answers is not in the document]
lmuntaner: Done, that was missing. I added a test that failed and it was fixed when I added the `hidePollingToast` call

frontend/src/lib/utils/sns.utils.ts
lmuntaner: Correct. This is correct.

That would happen if this util would be called outside the ProjectDetail page.

Which is not the case.

My goal is to remove also this helper and pass down the status as props. Or use the snsParticipationProgressStore that we discussed.

This is still to be refactored though. For now, this util could cause problems if used outside the Project Detail.

frontend/src/tests/lib/components/project-detail/ParticipateButton.spec.ts
lmuntaner: Not really, the TODO is the WHOLE file. It's like adding a TODO in ALL of our files. Because in all of them we mock the services layer.

frontend/src/tests/lib/components/project-detail/ProjectStatusSection.spec.ts
lmuntaner: No, I'll add an expect on the button that is not there initially, until the call has finished. That will make it clearer.

frontend/src/tests/lib/stores/sns-ticket.store.spec.ts
lmuntaner: This is the sns ticket store, there are no api calls.

lmuntaner:

frontend/src/tests/lib/components/project-detail/ParticipateButton.spec.ts
dskloetd: But you wrote a review comment on this line.

frontend/src/tests/lib/components/project-detail/ProjectStatusSection.spec.ts
dskloetd: I still don't understand what this comment means or how it relates to the line below.

frontend/src/tests/lib/services/sns-sale.services.spec.ts
```diff
@@ -493,15 +494,47 @@ describe("sns-api", () => {
...
+          await advanceTime(retryDelay);
+          retryDelay *= 2;
+          expectedCalls += 1;
+          expect(snsSwapCanister.getOpenTicket).toBeCalledTimes(expectedCalls);
+        }
+        expect(spyOnToastsShow).toBeCalledTimes(1);
+        expect(spyOnToastsHide).not.toBeCalled();
+        cancelPollGetOpenTicket();
+
+        await advanceTime(retryDelay);
```
dskloetd, line 534: Shouldn't the toast disappear immediately rather than after retryDelay?
`await runResolvedPromises()`
should be enough, I think.

frontend/src/lib/components/project-detail/ParticipateButton.svelte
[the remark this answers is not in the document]
dskloetd: But before it was also calling `snsTicketsStore.enablePolling(rootCanisterId);` in this file so it was symmetrical and possible to understand why it was there.

Now this comes here out of the blue. Can you at least add a comment explaining that this relates to restoreParticipation doing pollGetOpenTicket?

frontend/src/lib/components/project-detail/ParticipateButton.svelte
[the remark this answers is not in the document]
lmuntaner: `cancelPollOpenTicket` will only cancel the polling of the open ticket.

But if there is an open ticket the flow will continue with a participation.

Yet, if the user navigates away, setting the ticket to `undefined` in the store might stop the flow (depends on where the flow is) because the flow reads the ticket from the store.

Therefore, the comment:

```
// remove the ticket to stop sale-participation-retry from another pages because of the non-obvious UX
```

is correct.

frontend/src/lib/services/sns-sale.services.ts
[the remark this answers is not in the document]
lmuntaner: sure

frontend/src/tests/lib/services/sns-sale.services.spec.ts
lmuntaner: Correct, that's enough.

frontend/src/tests/lib/components/project-detail/ParticipateButton.spec.ts
lmuntaner: Because I knew you'd bring it up. I wanted to let you know I was aware of it but I wasn't planning on doing it in this PR.

frontend/src/lib/components/project-detail/ParticipateButton.svelte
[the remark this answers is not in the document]
lmuntaner: sure, that makes sense

lmuntaner:

dskloetd (approved):

PR #2104, opened 2023-03-16
GIX-1381: Add total token supply to project data

lmuntaner:
# Motivation

User sees the total token supply of an SNS Swap.

# Changes

* Read new property from the SNS Aggregator: `icrc1_total_supply`. Not yet in mainnet; therefore keeping it optional.
* New store snsTotalTokenSupplyStore to keep the total supplies per project.
* Set the snsTotalTokenSupplyStore after getting the data from the SNS Aggregator in the `loadSnsProjects` service.
* Add a new KeyValuePair in the ProjectSwapDetails.svelte
* Add a new property in `ProjectDetailStore` inside `ProjectDetailContext`: `totalTokensSupply`.
* New property is set from new derived store `snsTotalSupplyTokenAmountStore` that maps total supplies to project tokens.

# Tests

* Test that the new property in the aggregator is used when present.
* Test that new store is set in loadSnsProjects if new property is present. That not set if not present.
* Test that ProjectSwapDetails renders the new data when available.
* Test the new derived store `snsTotalSupplyTokenAmountStore`.
* Test the new store snsTotalTokenSupplyStore.
* New ProjectSwapDetails page object.

### DIFF
--- a/frontend/src/lib/api/sns-aggregator.api.ts
+++ b/frontend/src/lib/api/sns-aggregator.api.ts
@@ -61,6 +61,7 @@ export type CachedSns = {
    * @deprecated we will use the icrc1_metadata.fee as source information for the fee
    */
   icrc1_fee?: bigint;
+  icrc1_total_supply?: bigint;
 };
 
 type CachedSnsMetadataDto = {
@@ -137,6 +138,7 @@ type CachedSnsDto = {
   };
   icrc1_metadata: CachedSnsTokenMetadataDto;
   icrc1_fee: [] | [number];
+  icrc1_total_supply?: number;
 };
 
 const convertOptionalNumToBigInt = (
@@ -260,6 +262,7 @@ const convertSnsData = ({
   swap_state,
   icrc1_metadata,
   icrc1_fee,
+  icrc1_total_supply,
 }: CachedSnsDto): CachedSns => ({
   index,
   canister_ids,
@@ -275,6 +278,9 @@ const convertSnsData = ({
   },
   icrc1_metadata: convertIcrc1Metadata(icrc1_metadata),
   icrc1_fee: convertOptionalNumToBigInt(icrc1_fee[0]),
+  icrc1_total_supply: nonNullish(icrc1_total_supply)
+    ? BigInt(icrc1_total_supply)
+    : undefined,
 });
 
 const convertDtoData = (data: CachedSnsDto[]): CachedSns[] =>
--- a/frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
@@ -12,6 +12,11 @@
   import type { SnsParams } from "@dfinity/sns";
   import DateSeconds from "$lib/components/ui/DateSeconds.svelte";
   import type { IcrcTokenMetadata } from "$lib/types/icrc";
+  import { snsTotalTokenSupplyStore } from "$lib/stores/sns-total-token-supply.store";
+  import type { Principal } from "@dfinity/principal";
+  import { totalTokenSupply } from "$lib/utils/sns.utils";
+  import { nonNullish } from "@dfinity/utils";
+  import TestIdWrapper from "../common/TestIdWrapper.svelte";
 
   const { store: projectDetailStore } = getContext<ProjectDetailContext>(
     PROJECT_DETAIL_CONTEXT_KEY
@@ -19,10 +24,12 @@
 
   let params: SnsParams;
   let token: IcrcTokenMetadata;
+  let rootCanisterId: Principal;
   // type safety validation is done in ProjectDetail component
   $: ({
     swap: { params },
     token,
+    rootCanisterId,
   } = $projectDetailStore.summary as SnsSummary);
 
   let minCommitmentIcp: TokenAmount;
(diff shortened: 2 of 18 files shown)


lmuntaner:
New UI:

Screenshot 2023-03-16 at 12 34 16

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -280,3 +286,23 @@ export const parseSnsSwapSaleBuyerCount = (
...
+  token: Token;
+}): TokenAmount | undefined => {
+  const totalSupply = totalSupplies[rootCanisterId.toText()];
+  if (isNullish(totalSupply)) {
+    return undefined;
+  }
+
+  return TokenAmount.fromE8s({
+    amount: totalSupply.totalSupply,
+    token,
```
dskloetd, line 306: Do we know if this is the correct token for this project?

frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
```diff
@@ -41,25 +48,40 @@
...
-</KeyValuePair>
-<KeyValuePair>
-  <span slot="key">{$i18n.sns_project_detail.sale_end} </span>
-  <DateSeconds
-    slot="value"
-    seconds={Number(params.swap_due_timestamp_seconds ?? BigInt(0))}
-    tagName="span"
-  />
-</KeyValuePair>
+<TestIdWrapper testId="project-swap-details">
```
dskloetd, line 60: I like to use the "-component" suffix on these full component tids to distinguish them from others.
What do you think?

frontend/src/tests/page-objects/ProjectSwapDetails.page-object.ts
```diff
@@ -0,0 +1,25 @@
+import { AmountDisplayPo } from "./AmountDisplay.page-object";
+
+export class ProjectSwapDetailsPo {
+  static readonly tid = "project-swap-details";
+
+  root: Element;
+
+  constructor(root: Element) {
+    if (root.getAttribute("data-tid") !== ProjectSwapDetailsPo.tid) {
+      throw new Error(`${root} is not an ProjectSwapDetailsPo`);
```
dskloetd, line 10: an => a

frontend/src/lib/api/sns-aggregator.api.ts
```diff
@@ -61,6 +61,7 @@ export type CachedSns = {
    * @deprecated we will use the icrc1_metadata.fee as source information for the fee
    */
   icrc1_fee?: bigint;
+  icrc1_total_supply?: bigint;
```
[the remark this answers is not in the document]
lmuntaner: Yes, good point. Even better, I'll create an issue and set a reminder to do so in two weeks.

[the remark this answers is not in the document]
lmuntaner: icrc1_total_supply

frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
```diff
@@ -12,17 +12,24 @@
   import type { SnsParams } from "@dfinity/sns";
   import DateSeconds from "$lib/components/ui/DateSeconds.svelte";
   import type { IcrcTokenMetadata } from "$lib/types/icrc";
+  import { snsTotalTokenSupplyStore } from "$lib/stores/sns-total-token-supply.store";
+  import type { Principal } from "@dfinity/principal";
+  import { totalTokenSupply } from "$lib/utils/sns.utils";
+  import { nonNullish } from "@dfinity/utils";
+  import TestIdWrapper from "../common/TestIdWrapper.svelte";
 
   const { store: projectDetailStore } = getContext<ProjectDetailContext>(
```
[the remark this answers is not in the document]
lmuntaner: Yes, maybe I should add the variable in the context then?

To follow the same patterns.

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -280,3 +286,23 @@ export const parseSnsSwapSaleBuyerCount = (
...
+  rootCanisterId,
+  token,
+}: {
+  totalSupplies: SnsTotalTokenSupplyStoreData;
+  rootCanisterId: Principal;
+  token: Token;
+}): TokenAmount | undefined => {
+  const totalSupply = totalSupplies[rootCanisterId.toText()];
+  if (isNullish(totalSupply)) {
+    return undefined;
```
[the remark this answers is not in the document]
lmuntaner: It's optional and it might be that is not yet loaded so we don't want to throw.

We don't have a loading state for this data and there is no point in throwing if some of the static data is not loaded.

frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
lmuntaner: Yes, you already mentioned it here. Shall we write it in the test rules?

lmuntaner:

frontend/src/lib/utils/sns.utils.ts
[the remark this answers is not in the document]
lmuntaner: not used anymore

frontend/src/lib/api/sns-aggregator.api.ts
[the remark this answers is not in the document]
lmuntaner: Searching for `icrc1_total_supply` is as fast as searcher for the comment.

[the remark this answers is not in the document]
dskloetd: Then you wouldn't find this line for example:
```
let snsTotalTokenSupply: TokenAmount | undefined | null;
```
to remove the `| undefined | null`.
If you just don't want to do it, that's fine. But that reason isn't very good :-).

The reason I have a reflex to add a TODO pointing to the ticket is that at Google we had a system that would automatically notify you if you close a ticket which still has TODOs pointing to it.
But honestly, I think it's good practice even without the system.
But if you don't want to, just say that, it's fine.

frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
```diff
@@ -23,6 +25,7 @@
   $: ({
     swap: { params },
     token,
+    rootCanisterId,
```
dskloetd, line 28: Is rootCanisterId now unused here?

frontend/src/lib/derived/sns/sns-total-supply-token-amount.derived.ts
```diff
@@ -0,0 +1,40 @@
...
+} from "$lib/stores/sns-total-token-supply.store";
+import { snsSummariesStore } from "$lib/stores/sns.store";
+import type { SnsSummary } from "$lib/types/sns";
+import { TokenAmount } from "@dfinity/nns";
+import { nonNullish } from "@dfinity/utils";
+import { derived, type Readable } from "svelte/store";
+
+export const snsTotalSupplyTokenAmountStore = derived<
+  [Readable<SnsTotalTokenSupplyStoreData>, Readable<SnsSummary[]>],
+  Record<string, TokenAmount>
```
dskloetd, line 13: Indicate in a comment that the string refers to rootCanisterId?

frontend/src/lib/pages/ProjectDetail.svelte
```diff
@@ -140,7 +142,7 @@
   };
 
   // TODO: Change to a `let` that is recalculated when the store changes
-  const setProjectStore = (rootCanisterId: string) => {
+  let setProjectStore = (rootCanisterId: string) => {
```
dskloetd, line 145: Why did you make this function non-const?

frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
lmuntaner: no, true. and linter didn't caught the unused variable, weird.

frontend/src/lib/pages/ProjectDetail.svelte
lmuntaner: I'll revert. No need for now.

lmuntaner:

PR #2109, opened 2023-03-17
GIX-1212: Filter SNS Proposals

lmuntaner:
# Motivation

User can filter SNS proposal by decision status.

# Changes

* New derived store `snsFilteredProposalsStore` that combines the snsFiltersStore and snsProposals store to return the proposals that match the current filters.
* Change in how data is loaded in SnsProposals page. Filters and nervous functions are loaded when root canister id changes. Proposals when filters change.
* New simple derived store `snsSelectedFiltersStore` that returns the selected filters in snsFiltersStore.
* Changes in service `loadSnsProposals`. Now it reads the snsSelectedFiltersStore and applies the decision filters when asking for the proposals.

Minor change:
* Change type in snsProposals store to match current pattern.
* Service `loadSnsFilters` checks if filters are already loaded and skips if so.

# Tests

* Test new derived stores `snsFilteredProposalsStore` and `snsSelectedFiltersStore`.
* Test new logic in service `loadSnsProposals`.
* Changes in SnsProposals page test to mock only the api layer.
* New test in SnsProposals page test to test new filtering functionality.
* Test new logic in `loadSnsFilters`.
* New test helper `createSnsProposal`.

### DIFF
--- a/frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
+++ b/frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
@@ -1,0 +1,57 @@
+import {
+  snsSelectedFiltersStore,
+  type SnsFiltersStoreData,
+} from "$lib/stores/sns-filters.store";
+import {
+  snsProposalsStore,
+  type SnsProposalsStore,
+  type SnsProposalsStoreData,
+} from "$lib/stores/sns-proposals.store";
+import { mapProposalInfo } from "$lib/utils/sns-proposals.utils";
+import { isNullish } from "@dfinity/utils";
+import { derived, type Readable } from "svelte/store";
+
+export const snsFilteredProposalsStore = derived<
+  [SnsProposalsStore, Readable<SnsFiltersStoreData>],
+  SnsProposalsStoreData
+>(
+  [snsProposalsStore, snsSelectedFiltersStore],
+  ([proposalsStore, selectedFilters]) => {
+    const filteredProposals = Object.entries(proposalsStore).reduce(
+      (acc, [rootCanisterIdText, projectProposals]) => {
+        const filteredProjectProposals = projectProposals.proposals.filter(
+          (proposal) => {
+            const projectSelectedFilters = selectedFilters[rootCanisterIdText];
+            // Do not filter until filters are loaded
+            if (isNullish(projectSelectedFilters)) {
+              return true;
+            }
+            const proposalData = mapProposalInfo({
+              proposalData: proposal,
+              nsFunctions: projectSelectedFilters.topics.map(
+                ({ value }) => value
+              ),
+            });
+            const statusMatch =
+              projectSelectedFilters.decisionStatus.length === 0 ||
+              projectSelectedFilters.decisionStatus
+                .map(({ value }) => value)
+                .includes(proposalData.status);
+            // TODO: Filter by reward status
+            // TODO: Filter by nervous functions
+            return statusMatch;
+          }
+        );
+        return {
+          ...acc,
+          [rootCanisterIdText]: {
+            proposals: filteredProjectProposals,
+            certified: projectProposals.certified,
+          },
+        };
+      },
+      {}
+    );
+    return filteredProposals;
+  }
+);
--- a/frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
+++ b/frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
@@ -7,7 +7,7 @@ import {
   type SnsProposalsStore,
   type SnsProposalsStoreData,
 } from "$lib/stores/sns-proposals.store";
-import { mapProposalInfo } from "$lib/utils/sns-proposals.utils";
+import { snsDecisionStatus } from "$lib/utils/sns-proposals.utils";
 import { isNullish } from "@dfinity/utils";
 import { derived, type Readable } from "svelte/store";
 
@@ -19,24 +19,18 @@ export const snsFilteredProposalsStore = derived<
   ([proposalsStore, selectedFilters]) => {
     const filteredProposals = Object.entries(proposalsStore).reduce(
       (acc, [rootCanisterIdText, projectProposals]) => {
+        const projectSelectedFilters = selectedFilters[rootCanisterIdText];
+        // Do not add proposals filters are loaded
+        if (isNullish(projectSelectedFilters)) {
+          return acc;
+        }
         const filteredProjectProposals = projectProposals.proposals.filter(
           (proposal) => {
-            const projectSelectedFilters = selectedFilters[rootCanisterIdText];
-            // Do not filter until filters are loaded
-            if (isNullish(projectSelectedFilters)) {
-              return true;
-            }
-            const proposalData = mapProposalInfo({
-              proposalData: proposal,
-              nsFunctions: projectSelectedFilters.topics.map(
-                ({ value }) => value
-              ),
-            });
             const statusMatch =
               projectSelectedFilters.decisionStatus.length === 0 ||
               projectSelectedFilters.decisionStatus
                 .map(({ value }) => value)
-                .includes(proposalData.status);
+                .includes(snsDecisionStatus(proposal));
             // TODO: Filter by reward status
             // TODO: Filter by nervous functions
             return statusMatch;
--- a/frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
+++ b/frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
@@ -20,7 +20,8 @@ export const snsFilteredProposalsStore = derived<
     const filteredProposals = Object.entries(proposalsStore).reduce(
       (acc, [rootCanisterIdText, projectProposals]) => {
         const projectSelectedFilters = selectedFilters[rootCanisterIdText];
-        // Do not add proposals filters are loaded
+        // Skip the project if there are no filters for it.
+        // This will cause the proposals to be `undefined` for a specific project.
         if (isNullish(projectSelectedFilters)) {
           return acc;
         }
--- a/frontend/src/lib/pages/SnsProposals.svelte
+++ b/frontend/src/lib/pages/SnsProposals.svelte
@@ -35,15 +35,11 @@
   let currentProjectCanisterId: Principal | undefined = undefined;
   const unsubscribeProjectIdStore: Unsubscriber = snsOnlyProjectStore.subscribe(
     async (selectedProjectCanisterId) => {
-      if (
-        nonNullish(selectedProjectCanisterId) &&
-        currentProjectCanisterId?.toText() !==
-          selectedProjectCanisterId.toText()
-      ) {
-        currentProjectCanisterId = selectedProjectCanisterId;
+      currentProjectCanisterId = selectedProjectCanisterId;
+      if (nonNullish(selectedProjectCanisterId)) {
         await Promise.all([
-          loadSnsNervousSystemFunctions(currentProjectCanisterId),
-          loadSnsFilters(currentProjectCanisterId),
+          loadSnsNervousSystemFunctions(selectedProjectCanisterId),
+          loadSnsFilters(selectedProjectCanisterId),
         ]);
       }
     }
--- a/frontend/src/lib/pages/SnsProposals.svelte
+++ b/frontend/src/lib/pages/SnsProposals.svelte
@@ -18,6 +18,10 @@
   import { loadSnsFilters } from "$lib/services/sns-filters.services";
   import { snsOnlyProjectStore } from "$lib/derived/sns/sns-selected-project.derived";
   import { ENABLE_SNS_VOTING } from "$lib/stores/feature-flags.store";
+  import { snsFiltesStore } from "$lib/stores/sns-filters.store";
+  import { nonNullish } from "@dfinity/utils";
+  import { snsFilteredProposalsStore } from "$lib/derived/sns/sns-filtered-proposals.derived";
+  import type { Principal } from "@dfinity/principal";
 
   onMount(async () => {
     // We don't render this page if not enabled, but to be safe we redirect to the NNS proposals page as well.
@@ -28,19 +32,40 @@
     }
   });
 
-  const unsubscribe: Unsubscriber = snsOnlyProjectStore.subscribe(
+  let currentProjectCanisterId: Principal | undefined = undefined;
+  const unsubscribeProjectIdStore: Unsubscriber = snsOnlyProjectStore.subscribe(
     async (selectedProjectCanisterId) => {
-      if (selectedProjectCanisterId !== undefined) {
+      if (
+        nonNullish(selectedProjectCanisterId) &&
+        currentProjectCanisterId?.toText() !==
+          selectedProjectCanisterId.toText()
+      ) {
+        currentProjectCanisterId = selectedProjectCanisterId;
         await Promise.all([
-          loadSnsProposals({ rootCanisterId: selectedProjectCanisterId }),
-          loadSnsNervousSystemFunctions(selectedProjectCanisterId),
-          loadSnsFilters(selectedProjectCanisterId),
+          loadSnsNervousSystemFunctions(currentProjectCanisterId),
+          loadSnsFilters(currentProjectCanisterId),
         ]);
       }
     }
   );
 
-  onDestroy(unsubscribe);
+  const unsubscribeFiltersStore: Unsubscriber = snsFiltesStore.subscribe(
+    async (filters) => {
+      // First call will have `filters` as `undefined`.
+      // Once we have the initial filters, we load the proposals.
+      if (
+        nonNullish(currentProjectCanisterId) &&
+        nonNullish(filters[currentProjectCanisterId.toText()])
+      ) {
+        await loadSnsProposals({ rootCanisterId: currentProjectCanisterId });
+      }
+    }
+  );
+
+  onDestroy(() => {
+    unsubscribeProjectIdStore();
+    unsubscribeFiltersStore();
+  });
 
   let loadingNextPage = false;
   let loadNextPage: () => void;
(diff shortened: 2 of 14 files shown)


lmuntaner:
New UX:

https://user-images.githubusercontent.com/4550653/225871273-5e6612b4-6f84-4b1a-8095-1ccc25e14a68.mov

lmuntaner:
@dskloetd please review

frontend/src/tests/lib/pages/SnsProposals.spec.ts
lmuntaner, line 1: I will create a page-object for this page in a following PR.

This already PR is already big enough.

frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
```diff
@@ -0,0 +1,57 @@
...
+  [SnsProposalsStore, Readable<SnsFiltersStoreData>],
+  SnsProposalsStoreData
+>(
+  [snsProposalsStore, snsSelectedFiltersStore],
+  ([proposalsStore, selectedFilters]) => {
+    const filteredProposals = Object.entries(proposalsStore).reduce(
+      (acc, [rootCanisterIdText, projectProposals]) => {
+        const filteredProjectProposals = projectProposals.proposals.filter(
+          (proposal) => {
+            const projectSelectedFilters = selectedFilters[rootCanisterIdText];
```
dskloetd, line 24: Do you need to do this again for each proposal or does it make sense to move this line outside the filter call?

frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
```diff
@@ -0,0 +1,57 @@
...
+            // Do not filter until filters are loaded
+            if (isNullish(projectSelectedFilters)) {
```
dskloetd, line 26: This can also be moved outside the filter call, right?

frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
```diff
@@ -0,0 +1,57 @@
...
+              return true;
+            }
+            const proposalData = mapProposalInfo({
```
dskloetd, line 29: What does this map function do? I looked at it but it's not documented and not obvious.
It looks like all you need is the proposal state. Can you just use `snsDecisionStatus`?

frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -28,19 +32,40 @@
     }
   });
 
-  const unsubscribe: Unsubscriber = snsOnlyProjectStore.subscribe(
+  let currentProjectCanisterId: Principal | undefined = undefined;
+  const unsubscribeProjectIdStore: Unsubscriber = snsOnlyProjectStore.subscribe(
     async (selectedProjectCanisterId) => {
-      if (selectedProjectCanisterId !== undefined) {
+      if (
+        nonNullish(selectedProjectCanisterId) &&
```
dskloetd, line 39: Do you want to do some cleanup if the selectedProjectCanisterId becomes nullish?
If it later becomes equal to currentProjectCanisterId after having been nullish, do we need to load again or we can rely on the previously loaded data?

frontend/src/lib/services/sns-filters.services.ts
```diff
@@ -3,12 +3,17 @@ import { snsFiltesStore } from "$lib/stores/sns-filters.store";
 import { enumValues } from "$lib/utils/enum.utils";
 import type { Principal } from "@dfinity/principal";
 import { SnsProposalDecisionStatus } from "@dfinity/sns";
+import { nonNullish } from "@dfinity/utils";
 import { get } from "svelte/store";
 
 // TODO: Set default filters
 export const loadSnsFilters = async (rootCanisterId: Principal) => {
```
dskloetd, line 10: What exactly is being loaded here? I don't see any api calls or anything that could load something.
Also, there is no `await` anywhere so should this even be `async`?

I don't understand what this does so I'm also not sure what the change is supposed to do.

frontend/src/lib/stores/sns-filters.store.ts
```diff
@@ -98,3 +98,20 @@ export const initSnsFiltersStore = (): ProjectFiltersStore => {
 };
 
 export const snsFiltesStore = initSnsFiltersStore();
+
+export const snsSelectedFiltersStore = derived<
+  Readable<SnsFiltersStoreData>,
+  SnsFiltersStoreData
+>(snsFiltesStore, ($snsFilters) =>
```
dskloetd, line 105: snsFiltesStore is missing an 'r'

frontend/src/tests/lib/derived/sns/sns-filtered-proposals.derived.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+  const snsProposal3: SnsProposalData = {
+    ...mockSnsProposal,
+    id: [{ id: BigInt(3) }],
+  };
+
+  beforeEach(() => {
+    snsFiltesStore.reset();
+    snsProposalsStore.reset();
+  });
+  it("should return all proposals if filter store is not loaded", () => {
```
dskloetd, line 30: Does it mean we'll make multiple calls to load proposals? One before the filter store is loaded and one after? Wouldn't it make sense to wait with loading proposals until after the filter store is loaded?

frontend/src/tests/lib/pages/SnsProposals.spec.ts
```diff
@@ -2,75 +2,98 @@
...
       beforeEach(() => {
-        snsProposalsStore.reset();
+        jest
+          .spyOn(governanceApi, "queryProposals")
+          .mockResolvedValue(proposals);
       });
 
-      it("should load proposals and nervous system functions functions", () => {
-        render(SnsProposals);
+      it("should load nervous system functions functions", async () => {
```
dskloetd, line 72: What are functions functions?

frontend/src/tests/mocks/sns-proposals.mock.ts
```diff
@@ -44,3 +49,71 @@ export const mockSnsProposal: SnsProposalData = {
...
+  proposalId: bigint;
+}): SnsProposalData => {
+  const id: [SnsProposalId] = [{ id: proposalId }];
+  switch (status) {
+    case SnsProposalDecisionStatus.PROPOSAL_DECISION_STATUS_OPEN:
+      return {
+        ...mockSnsProposal,
+        id,
+        decided_timestamp_seconds: BigInt(0),
+      };
```
dskloetd, line 81: What about this makes the proposal open?

frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
```diff
@@ -0,0 +1,57 @@
...
+import { mapProposalInfo } from "$lib/utils/sns-proposals.utils";
+import { isNullish } from "@dfinity/utils";
+import { derived, type Readable } from "svelte/store";
+
+export const snsFilteredProposalsStore = derived<
+  [SnsProposalsStore, Readable<SnsFiltersStoreData>],
+  SnsProposalsStoreData
+>(
+  [snsProposalsStore, snsSelectedFiltersStore],
+  ([proposalsStore, selectedFilters]) => {
```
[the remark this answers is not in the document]
lmuntaner: Those are set with the derived type annotation:

```
[SnsProposalsStore, Readable<SnsFiltersStoreData>]
```

That's why we add them there.

frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
lmuntaner: yes

frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
lmuntaner: I can use a helper per status. I will need to add also the rewardStatus, but that's ok.

I agree it will make it easier to follow

frontend/src/lib/services/sns-filters.services.ts
lmuntaner: Nothing is loaded at the moment because I only have the decision status filters. Those are loaded from an `enum`.

The topics filters need to come from a call, it will rely on the nervous functions being loaded.

I wanted the same pattern to load and also to use both filters, the ones coming from an `enum` and the ones coming from backend.

frontend/src/lib/stores/sns-filters.store.ts
lmuntaner: oh no, that's everywhere.

Let me do that in another PR to no pollute this one.

frontend/src/tests/lib/derived/sns/sns-filtered-proposals.derived.spec.ts
lmuntaner: That's a good point and it makes sense. I'll change this to return `undefined` then.

frontend/src/tests/lib/pages/SnsProposals.spec.ts
lmuntaner: typo

lmuntaner: NNS has types and topics for proposals.

But in SNS there are nervous functions. Those can be managed via proposals and are similar to a type or topic of proposal.

It's shown in the proposal card and there will also be a filter for them.

frontend/src/tests/mocks/sns-proposals.mock.ts
lmuntaner: The status is an entity when filtering, but it's not part of the Sns Neuron data.

In each helper `snsDecisionStatus` or `isAccepted` or `snsRewardStatus` I added a link to the backend code.

frontend/src/tests/lib/pages/SnsProposals.spec.ts
```diff
@@ -2,75 +2,98 @@
...
-
-jest.mock("$lib/services/sns-filters.services", () => {
-  return {
-    loadSnsFilters: jest.fn().mockResolvedValue(undefined),
-  };
-});
+const blockedApiPaths = ["$lib/api/sns-governance.api"];
 
 describe("SnsProposals", () => {
+  blockAllCallsTo(blockedApiPaths);
```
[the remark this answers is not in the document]
lmuntaner: done

lmuntaner:

frontend/src/lib/pages/SnsProposals.svelte
dskloetd: The way the code is here now:
1. When `selectedProjectCanisterId` becomes `null` this if skips the block below.
2. When `selectedProjectCanisterId` becomes non-null but equal to the previous `currentProjectCanisterId`, this if skips the block below again.

I'm not sure I understood your response but it should like you don't want to silently reuse the previous loaded data and want to control that from the service.
If that's what you want, then you should probably reset `currentProjectCanisterId` when `selectedProjectCanisterId` becomes null.

frontend/src/lib/services/sns-filters.services.ts
dskloetd: So what is actually changing in this file and why? Is it just that it doesn't fill the store again if it is already filled? Is it related to the rest of the PR?

frontend/src/tests/mocks/sns-proposals.mock.ts
dskloetd: I'm not sure what this means.
`createSnsProposal` is a function that creates a proposal with the given status, right?
It switches over `status` and for each case it returns a proposal with that `status`, right?
But then I don't see anything related to the particular status in any of the case blocks.
What am I missing?

frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
```diff
@@ -19,24 +19,18 @@ export const snsFilteredProposalsStore = derived<
   ([proposalsStore, selectedFilters]) => {
     const filteredProposals = Object.entries(proposalsStore).reduce(
       (acc, [rootCanisterIdText, projectProposals]) => {
+        const projectSelectedFilters = selectedFilters[rootCanisterIdText];
+        // Do not add proposals filters are loaded
```
dskloetd, line 23: I'm not sure how to parse this sentence?

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -191,12 +218,27 @@ async function getSnsNeuron({
...
+  return proposals.get(mapKey({ identity, rootCanisterId })) || [];
+}
+
 /////////////////////////////////
 // Functions to control the fake:
 /////////////////////////////////
 
 const reset = () => {
   neurons.clear();
+  proposals.clear();
```
dskloetd, line 241: Also clear nervousFunctions?

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -237,6 +279,46 @@ export const addNeuronWith = ({
...
+  const proposal: SnsProposalData = {
+    ...mockSnsProposal,
+    id: [defaultProposalId],
+    ...proposalParams,
+  };
+  proposalsList.push(proposal);
+  return proposal;
+};
+
+export const addNervousFunctionWith = ({
```
dskloetd, line 302: addNervous**System**FunctionWith?

frontend/src/lib/services/sns-filters.services.ts
lmuntaner: Yes, that's the change.

It is related because it could be called multiple times and I don't want to load them again because it would overried the current filters.

frontend/src/tests/mocks/sns-proposals.mock.ts
lmuntaner: That the status is not a property of the proposal. Instead, it's a calculation based on some properties of the proposal data.

Yet, the filters are set using this calculated property.

frontend/src/lib/derived/sns/sns-filtered-proposals.derived.ts
lmuntaner: Sorry, bery bad english.

frontend/src/lib/pages/SnsProposals.svelte
lmuntaner: Good point, I'll move the assignment outside and I can anyway call the functions because they will handle not loading the data if already there.

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -87,15 +114,15 @@ async function nervousSystemParameters({
 }
 
 async function getNervousSystemFunctions({
-  rootCanisterId: _,
-  identity: __,
-  certified: ___,
+  rootCanisterId,
+  identity,
```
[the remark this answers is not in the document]
lmuntaner: Nervous functions are the same for all users. They are defined at the SNS level.

frontend/src/tests/fakes/sns-governance-api.fake.ts
lmuntaner: yes, thanks!

frontend/src/tests/fakes/sns-governance-api.fake.ts
lmuntaner: sure, sometimes I don't use the "system" to make it shorter.

lmuntaner:

frontend/src/tests/fakes/sns-governance-api.fake.ts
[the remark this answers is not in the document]
dskloetd: Then it makes sense to not key them by identity in the fake either, right?

frontend/src/tests/mocks/sns-proposals.mock.ts
dskloetd: OK, so back to my original question (rephrased):
What about this proposal is it that makes the calculation that determines the status, determine that this one is Open?

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -30,13 +34,16 @@ const implementedFunctions = {
...
   refreshNeuron,
   claimNeuron,
+  queryProposals,
 };
 
 //////////////////////////////////////////////
 // State and helpers for fake implementations:
 //////////////////////////////////////////////
 
 const neurons: Map<string, SnsNeuron[]> = new Map();
```
dskloetd, line 44: I could swear I added a comment on this after you asked me in a PR review but it seems I only did so in governance-api and sns-ledger-api. Could you add this and also add something for the maps below?
```
// Maps a key reppresenting identity + rootCanisterId to a list of neurons for
// that identity and sns project.
```

frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -35,15 +35,11 @@
   let currentProjectCanisterId: Principal | undefined = undefined;
   const unsubscribeProjectIdStore: Unsubscriber = snsOnlyProjectStore.subscribe(
     async (selectedProjectCanisterId) => {
-      if (
-        nonNullish(selectedProjectCanisterId) &&
-        currentProjectCanisterId?.toText() !==
-          selectedProjectCanisterId.toText()
-      ) {
-        currentProjectCanisterId = selectedProjectCanisterId;
+      currentProjectCanisterId = selectedProjectCanisterId;
```
dskloetd, line 38: currentProjectCanisterId is now unused, right?
Use it or remove it?

lmuntaner: It's being used to load the proposals. See comment below

frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -28,19 +32,36 @@
...
-  onDestroy(unsubscribe);
+  const unsubscribeFiltersStore: Unsubscriber = snsFiltesStore.subscribe(
+    async (filters) => {
+      // First call will have `filters` as `undefined`.
+      // Once we have the initial filters, we load the proposals.
+      if (
+        nonNullish(currentProjectCanisterId) &&
+        nonNullish(filters[currentProjectCanisterId.toText()])
+      ) {
+        await loadSnsProposals({ rootCanisterId: currentProjectCanisterId });
```
lmuntaner, line 56: We use it here.

frontend/src/tests/fakes/sns-governance-api.fake.ts
lmuntaner: Sure

frontend/src/tests/fakes/sns-governance-api.fake.ts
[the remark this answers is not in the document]
lmuntaner: I didn't want to change the pattern. I thought it can bring more questions to change the pattern than do another one.

frontend/src/tests/mocks/sns-proposals.mock.ts
lmuntaner: In thie case, just `decided_timestamp_seconds` being 0.

You can either learn about it in the util `snsDecisionStatus` or in the backend code: https://github.com/dfinity/ic/blob/226ab04e0984367da356bbe27c90447863d33a27/rs/sns/governance/src/proposal.rs#L717

lmuntaner:

frontend/src/tests/fakes/sns-governance-api.fake.ts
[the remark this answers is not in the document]
dskloetd: There's not really a pattern.
sns-ledger-api keys accounts by identity + rootCanisterId
governance-api keys neurons by just identity
sns-aggregator-api doesn't use a map at all for CachedSnS.

I think the fake should just do what makes sense in the real world.

frontend/src/tests/mocks/sns-proposals.mock.ts
dskloetd: Thanks! I think it would be nice to have a comment on each of the cases explaining why it has/should have that status. And/or link to the corresponding code that calculates the status based on the properties.

lmuntaner: I can add a comment in the data creator, sounds good.

frontend/src/tests/fakes/sns-governance-api.fake.ts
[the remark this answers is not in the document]
lmuntaner: ok

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -91,8 +91,8 @@ const getProposals = (keyParams: KeyParams) => {
   return proposalsList;
 };
 
-const getNervousFunctions = (keyParams: KeyParams) => {
-  const key = mapKey(keyParams);
+const getNervousFunctions = (rootCanisterId: Principal) => {
+  const key = rootCanisterId.toText();
```
dskloetd, line 95: Can you fix the comment on line 49?

dskloetd (approved):

PR #2111, opened 2023-03-18
Cache rs projects separately

bitdivine:
# Motivation
If any rust code is changed, all rust code is recompiled in Docker.  Changes in the `nns-dapp` cause the aggregator to be rebuilt and vice versa.  This is unnecessary work.

# Changes
- Keep the toy implementation of every rust project created in builder to warm the cache.
- When building a rust project, copy in only the code for that project and use the placeholder for the other project.

# Tests
See docker builds in CI


lmuntaner (approved):
Thanks!

PR #2113, opened 2023-03-20
Add Playwright for e2e tests

dskloetd:
# Motivation

We want to add more e2e tests. Playwright seems nice. So let's migrate to it before we have a lot more e2e tests.

# Changes

1. Installed Playwright.
2. Added `test-e2e` script to package.json
3. Added a util to create a new user account on internet identity.
4. Added an example test which will be removed once we add a proper test.

Playwright installed a default config with a lot of options.
I left the ones I wasn't sure about unchanged, left all the explaining comments and commented some things out for potential future use.
The changes I made to the config are:

1. Added dotenv
2. testDir
3. fullyParallel
4. retries
5. reporter
6. baseUrl

# Tests

Change is tests only.


frontend/src/tests/utils/e2e.test-utils.ts
```diff
@@ -0,0 +1,26 @@
+import { expect, type BrowserContext, type Page } from "@playwright/test";
+
+export const signInWithNewUser = async ({
```
peterpeterparker, line 3: - I was thinking, is this a "utils" or should it be "helper" or something? since it holds a scenario of test, maybe we should use another package than "utils"?

- it's only an example or initial commit - i.e. not the final implementation right?

frontend/src/tests/tsconfig.json
```diff
@@ -1,3 +1,9 @@
 {
-  "extends": "../../tsconfig.spec.json"
+  "extends": "../../tsconfig.spec.json",
```
peterpeterparker, line 2: don't think this is correct. playwright should not build the test and the app.

likewise the jest test should probably not build e2e.

any particular reason to have this extend?

spontaneously I would:

- remove this extend
- use playwright config `webserver.command` to either run `npm run build && npm run preview` to run against a prod build or in development `npm run dev`
- exclude anything e2e in `tsconfig.spec.json` to not build playwright thing in jest test

peterpeterparker: Ok let's put `webserver.command` on the side for the moment.

Maybe we could just exclude anything e2e in `tsconfig.spec.json` and overwrite `include` and `exclude` in `tsconfig.json` of Playwright to be accurate.

`tsconfig.spec.json`:

```
{
  "extends": "./.svelte-kit/tsconfig.json",
  "compilerOptions": {
    "allowJs": true,
    "checkJs": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "resolveJsonModule": true,
    "skipLibCheck": true,
    "sourceMap": true,
    "strict": false,
    "paths": {
      "$lib": ["src/lib"],
      "$lib/*": ["src/lib/*"],
      "$routes": ["src/routes"],
      "$routes/*": ["src/routes/*"],
      "$mocks": ["__mocks__"],
      "$mocks/*": ["__mocks__/*"],
      "$tests": ["src/tests"],
      "$tests/*": ["src/tests/*"]
    }
  },
  "include": [
    ".svelte-kit/ambient.d.ts",
    "./vite.config.ts",
    "./src/**/*.js",
    "./src/**/*.ts",
    "./src/**/*.svelte"
  ],
  "exclude": ["./src/tests/e2e/**/*"] // <-- something like that
}
```

in `tsconfig.json` of playwright

```
{
   ...
   include: ["./src/tests/e2e/**/*", "./src/tests/utils/**/*"] // <-- something like that
```

frontend/src/tests/utils/e2e.test-utils.ts
peterpeterparker: > The example test is just an example, but logging in to internet identity we will definitely need and this seems to work so I would keep it.

👍

> Are you suggesting making a new /helpers/ directory? I'm not sure what's the difference between a helper and a util.

Yes it's what I meant. I thought about it because Playwright use "helper" in their documentation (https://playwright.dev/docs/test-typescript#tsconfigjson) that's why I thought maybe a "utils" is something that does not contain anything related for a scenario of a business feature.

But no strong opinion, it's just something that crossed my mind that's why I asking.

frontend/src/tests/tsconfig.json
dskloetd: What exactly do you mean by "`tsconfig.json` of playwright"?
[Note that](https://playwright.dev/docs/test-typescript): "Playwright **only supports** the following tsconfig options: `paths` and `baseUrl`."
So if that's correct, `include` in a playwright `tsconfig.json` will do nothing.
But maybe I misunderstood what you mean.

peterpeterparker: > Playwright only supports the following tsconfig options: paths and baseUrl.

Agree, alright then just the `tsconfig.spec.json` unless you see a a reason not too?

dskloetd: Done. The tests still pass but I'm not sure what effect it should have in order for me to verify that it's working.

peterpeterparker (approved):
LGTM, thanks

PR #2117, opened 2023-03-20
Create SNSs in parallel

bitdivine:
# Motivation
Faster CI.  There are two things that take a lot of time in the current CI.  One of those is creating many SNS for the aggregator test.

# Changes
- Use the snsdemo command that creates SNSs in parallel

# Tests
This is a change to a test; see CI for the result...

In particular, at present the aggregator test currently takes between 18 and 21 minutes.  That includes several other steps but 7m of that is creating 11 SNSs.  If successful, this PR should take a few minutes off that.  It is hard to estimate how much less as the github runners are small single core VMs and that limits the value of running things in parallel.

Result: The aggregator test took 14 minutes, so the speedup is worthwhile.

### DIFF
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -89,11 +89,7 @@ jobs:
       - name: Create more SNSs
         run: |
           cd "$( dirname "$(command -v snsdemo)" )/.."
-          seq 11 | while read -r line ; do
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
I guess there is a plan to move snsdemo to this repo.

Thanks!

PR #2119, opened 2023-03-20
Chore: Add swapCanisterId ro `participateInSnsSale`

lmuntaner:
# Motivation

Simplify the code in the Sns participation flow.

Add `swapCanisterId` as a parameter in swap service `participateInSnsSale`.

# Changes

* Add `swapCanisterId` param to `participateInSnsSale` sns swap service.
* Pass the new param where `participateInSnsSale` is used.

# Tests

* Fix tests after refactor.

### DIFF
--- a/frontend/src/lib/services/sns-sale.services.ts
+++ b/frontend/src/lib/services/sns-sale.services.ts
@@ -58,6 +58,7 @@ import type {
 } from "@dfinity/sns/dist/candid/sns_swap";
 import type { E8s } from "@dfinity/sns/dist/types/types/common";
 import {
+  assertNonNullish,
   fromDefinedNullable,
   fromNullable,
   isNullish,
@@ -466,8 +467,10 @@ export const initiateSnsSaleParticipation = async ({
     });
 
     const swapCanisterId = project?.summary.swapCanisterId;
+    // Edge case: `initiateSnsSaleParticipation` can't be called if there is no swap canister id
+    assertNonNullish(swapCanisterId);
     const ticket = get(snsTicketsStore)[rootCanisterId?.toText()]?.ticket;
-    if (nonNullish(ticket) && nonNullish(swapCanisterId)) {
+    if (nonNullish(ticket)) {
       // Step 2. to finish
       const { success } = await participateInSnsSale({
         rootCanisterId,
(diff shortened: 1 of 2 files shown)


dskloetd:
Unfinished sentence in PR description?
"Add swapCanisterId as"

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -665,18 +667,22 @@ const pollTransfer = ({
  */
 export const participateInSnsSale = async ({
   rootCanisterId,
+  swapCanisterId,
   postprocess,
   userCommitment,
   updateProgress,
   ticket,
-}: ParticipateInSnsSaleParameters & { ticket: Ticket }): Promise<{
+}: ParticipateInSnsSaleParameters & {
```
dskloetd, line 675: Based on the name, it sounds like ParticipateInSnsSaleParameters should have all the parameters for the participateInSnsSale function. Why are ticket and swapCanisterId not part of it?

frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -465,11 +465,13 @@ export const initiateSnsSaleParticipation = async ({
       amount_icp_e8s: amount.toE8s(),
     });
 
+    const swapCanisterId = project?.summary.swapCanisterId;
     const ticket = get(snsTicketsStore)[rootCanisterId?.toText()]?.ticket;
-    if (nonNullish(ticket)) {
+    if (nonNullish(ticket) && nonNullish(swapCanisterId)) {
```
dskloetd, line 470: If we don't have the project/swapCanisterId, is there some mechanism by which initiateSnsSaleParticipation is called again once we do have swapCanisterId?
Is this tested?

lmuntaner: This is more of a type test.

`initiateSnsSaleParticipation` is called when the user goes through all the flow of participation. This flow is not available unless we have all the project info.

Yet, because the project comes from a `find`, TS forces us to consider that as possibly `undefined`

lmuntaner:
> Unfinished sentence in PR description? "Add swapCanisterId as"

Done, thanks

frontend/src/lib/services/sns-sale.services.ts
lmuntaner: yes

dskloetd (approved):

PR #2122, opened 2023-03-20
Use pre-packaged test state

bitdivine:
# Motivation
Faster CI.  In particular, setting up a network with the required state is slow.  Using an existing state that has been saved instead, saves time.

# Changes
- Deploy e2e and aggregator test environments using the saved state generated by snsdemo
  - Note: This does not include ckBTC canisters, open SNS, pre-populated sns_aggregators and other features that we would like in our test environment.  That is fine, the state can be expanded.
- Delete the existing deployment code

# Tests
See CI


.github/workflows/aggregator.yaml
```diff
@@ -15,40 +15,8 @@ jobs:
...
-          set -x
-          dfx --version
-          node --version
-          npm --version
-          rustc --version
-          cargo --version
-          ic-cdk-optimizer --version
+      - name: Build the nns-dapp
+        run: ./scripts/docker-build --network local # Note: This can be optimized by not building the nns-dapp
       - name: Get SNS scripts
```
bitdivine, line 20: I think we can probably drop the sns scripts now but I'd have to check. `dfx-canister-set-id` is used from there but if we can put the whole canister_ids.json in the state, we don't need this anymore.  There might be other dependencies.

.github/workflows/aggregator.yaml
```diff
@@ -32,47 +32,31 @@ jobs:
...
       - name: Add SNS scripts to the path
         run: |
           echo "$PWD/snsdemo/bin" >> $GITHUB_PATH
-      - name: Install SNS script dependencies
+      - name: Install tools
         run: |
-          dfx-sns-demo-install
-      - name: Deploy NNS and SNS canisters
+          dfx-software-dfx-install --version "$(jq -r .dfx dfx.json)"
+          dfx-software-idl2json-install
```
bitdivine, line 43: We need fewer tools, so save time.

.github/workflows/aggregator.yaml
```diff
@@ -32,47 +32,31 @@ jobs:
         with:
           repository: 'dfinity/snsdemo'
           path: 'snsdemo'
-          # Version from Mar 17 2023
-          ref: 'b9a786a8a4e4afbfede2ecc5af0c2763ee2be774'
+          # Version from Mar 25 2023 with saved state
+          ref: '8a8403490048042ec17e06ee8f708e9fceff1bfb'
```
bitdivine, line 36: This version has tooling for saving and restoring state.

.github/workflows/aggregator.yaml
```diff
@@ -32,47 +32,31 @@ jobs:
...
-          name: sns_aggregator
-          path: sns_aggregator.wasm
-          retention-days: 3
-      - name: Wait for the aggregator to get data
-        run: sleep 30
-      - name: Get the latest data from the sns aggregator
-        run: |
-          AGGREGATOR_CANISTER_ID="$(dfx canister id sns_aggregator)"
-          curl -Lf "http://${AGGREGATOR_CANISTER_ID}.localhost:8080/v1/sns/list/latest/slow.json" | tee aggregate-1.json
-          expect=1
```
bitdivine, line 61: The installed state already has many SNS so we skip the bit where we have one SNS.

.github/workflows/aggregator.yaml
```diff
@@ -32,47 +32,31 @@ jobs:
...
         run: |
           echo "$PWD/snsdemo/bin" >> $GITHUB_PATH
-      - name: Install SNS script dependencies
+      - name: Install tools
         run: |
-          dfx-sns-demo-install
-      - name: Deploy NNS and SNS canisters
+          dfx-software-dfx-install --version "$(jq -r .dfx dfx.json)"
+          dfx-software-idl2json-install --version "v$(jq -r .defaults.build.config.IDL2JSON_VERSION dfx.json)"
+      - name: Get test environment
```
bitdivine, line 44: Here we get a shrink-wrapped environment representing our current model of mainnet rather than installing canisters and creating state.

.github/workflows/aggregator.yaml
```diff
@@ -32,47 +32,31 @@ jobs:
...
-            scripts/sns/aggregator/get_log
-          }
-      - name: Make the aggregator collect data slowly
-        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000000; fast_interval_ms = 1_000_000_000; })'
-      - name: Create more SNSs
+          curl -fL --retry 5 https://github.com/dfinity/snsdemo/releases/download/tip/state.zip > state.zip
+          dfx-state-install -f state.zip --verbose
+          dfx identity use snsdemo8
+          dfx-sns-demo-healthcheck
+      - name: Upgrade to the current aggregator
```
bitdivine, line 50: The shrink-wrapped environment probably doesn't have the aggregator or nns-dapp matching the workflow commit.  Here we want to test the aggregator so we need to install the relevant version.

.github/workflows/aggregator.yaml
```diff
@@ -32,47 +32,31 @@ jobs:
...
+          cp out/sns_aggregator_dev.wasm sns_aggregator.wasm
+          dfx canister install sns_aggregator --wasm sns_aggregator.wasm --upgrade-unchanged '(opt record { update_interval_ms = 1000; fast_interval_ms = 1_000_000_000; })'
       - name: Make the aggregator collect data quickly
-        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000; fast_interval_ms = 1_000_000_000; })'
+        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 100; fast_interval_ms = 1_000_000_000; })'
       - name: Wait for the aggregator to get data
-        run: sleep 30
+        run: sleep 120
+        # sleep time > 12 SnS & 2 block heights each + a few extra calls.
+        # TODO: The aggregator can be installed and populated in the saved state, so this sleep is not needed.
```
bitdivine, line 59: Note: Previously the aggregator was collecting data while SNSs were being installed, so only a short delay was needed.

Note: It is possible to shorten the delay, however I believe that that is not the right approach.  Rather, the aggregator should already have data in the state snapshot and the upgrade should preserve that data.

.github/workflows/aggregator.yaml
```diff
@@ -32,47 +32,31 @@ jobs:
...
-      - name: Create more SNSs
+          curl -fL --retry 5 https://github.com/dfinity/snsdemo/releases/download/tip/state.zip > state.zip
+          dfx-state-install -f state.zip --verbose
+          dfx identity use snsdemo8
+          dfx-sns-demo-healthcheck
+      - name: Upgrade to the current aggregator
         run: |
-          dfx-sns-demo-mksns-parallel --num_sns 11 --majority snsdemo8
+          cp out/sns_aggregator_dev.wasm sns_aggregator.wasm
+          dfx canister install sns_aggregator --wasm sns_aggregator.wasm --upgrade-unchanged '(opt record { update_interval_ms = 1000; fast_interval_ms = 1_000_000_000; })'
```
dskloetd, line 53: Why is the fast_interval_ms to big? Maybe add a comment to explain?

.github/workflows/aggregator.yaml
```diff
@@ -32,47 +32,31 @@ jobs:
...
       - name: Make the aggregator collect data quickly
-        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 1000; fast_interval_ms = 1_000_000_000; })'
+        run: dfx canister call sns_aggregator reconfigure '(opt record { update_interval_ms = 100; fast_interval_ms = 1_000_000_000; })'
```
dskloetd, line 55: Can we just pass 100 instead of 1000 on the previous line? Or what is the purpose of the reconfigure?

.github/workflows/build.yml
```diff
@@ -36,34 +36,38 @@ jobs:
...
+      - name: Install nns-dapp and sns_aggregator
         run: |
-          # Note: This deploys standard NNS canisters but with:
-          # - The current code for the nns-dapp
-          # - The II version specified in dfx.json
-          # - Functional SNS canisters
-          ./scripts/dfx-nns-deploy-custom --features sns
+          dfx canister install nns-dapp --wasm nns-dapp.wasm --upgrade-unchanged --mode reinstall --yes
+          dfx canister install sns_aggregator --wasm sns_aggregator.wasm --upgrade-unchanged --mode reinstall --yes
+          # Remote canisters are a terrible idea.
```
dskloetd, line 58: What is this trying to explain? Can the comment be more useful?

dskloetd (approved):

.github/workflows/build.yml
bitdivine: More useful would be to remove the concept of remote canister IDs from dfx.  But I have updated the comment to be less of a passionate objection to the endless pain that the remote canister concept inflicts and more of exactly why it is unsuitable here.

.github/workflows/aggregator.yaml
bitdivine: It doesn't need to be.  Any value will do.  I have added a comment to that effect.

.github/workflows/aggregator.yaml
bitdivine: Yes, precisely.

dskloetd: So you can remove this line, right?

dskloetd (approved):

.github/workflows/aggregator.yaml
bitdivine: No, if this line were removed the following sleep would have to be ten times as long.

PR #2124, opened 2023-03-20
Cache only docker main

bitdivine:
# Motivation
Improve docker caching

- GitHub has limited storage for caches.  If there are too many branches pushing changes to the cache, we get cache evictions and things run slowly.  Let's rather cache main only, and branches can copy from main but not write docker caches.
- On the other hand, we can cache more steps, so a given PR rebuilds only the parts of the codebase that have diverged from main.

# Changes
- Add a github action that saves a full docker build every time a commit is merged into main.
- Make other docker builds read-only

# Future
- Make more extensive use of the docker builds in CI.

# Tests
- Timing
  - Verified that bumping this PR yields:
    - Cached local docker build completes in about 30 seconds, most of which is spent downloading the cache.
      - Update: After several runs, it frequently takes around 30s but sometimes it takes 1m, doing nothing but downloading the docker layers.
    - Full mainnet docker build completes in just under 5 minutes; the builder stage hits the cache, frontend and nns-dapp must be rebuilt.
    - For a native build, installing the tooling, without building, takes just under 5 minutes.  Build is extra but hard to measure as it is not a separate step.
- Note:  I have seen e2e tests flake sometimes.  Should scrutinize them closely and fix.

# Future
- Use docker builds for the aggregator and e2e test workflows.

### DIFF
--- a/rs/backend/src/main.rs
+++ b/rs/backend/src/main.rs
@@ -1,4 +1,3 @@
-// Pointless comment
 use crate::accounts_store::{
     AccountDetails, AddPendingNotifySwapRequest, AddPendingTransactionResponse, AttachCanisterRequest,
     AttachCanisterResponse, CreateSubAccountResponse, DetachCanisterRequest, DetachCanisterResponse,


.github/workflows/docker-build.yaml
```diff
@@ -5,26 +5,7 @@ concurrency:
...
-        uses: docker/setup-buildx-action@v2
-      - name: Build base Docker image
-        uses: docker/build-push-action@v3
-        with:
-          context: .
-          file: Dockerfile
-          cache-from: type=gha,scope=cached-stage
-          cache-to: type=gha,scope=cached-stage,mode=max
-          outputs: type=cacheonly
-          target: builder
```
bitdivine, line 25: Note: docker is used again below but only with "cache-from", not "cache-to"

.github/workflows/docker-build.yaml
```diff
@@ -5,26 +5,7 @@ concurrency:
...
   assets:
-    needs: builder
```
dskloetd, line 27: Why did it need builder before and why does it no longer need builder?

.github/workflows/docker-main.yaml
```diff
@@ -0,0 +1,43 @@
...
+      # https://github.com/docker/build-push-action/issues/539
+      - name: Set up docker buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Build docker artifacts
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: Dockerfile
+          build-args: |
+            DFX_NETWORK=${{ matrix.DFX_NETWORK }}
```
dskloetd, line 39: The code removed from `.github/workflows/docker-build.yaml` did not set DFX_NETWORK. Why was it not needed there but it is needed here?

.github/workflows/docker-main.yaml
```diff
@@ -0,0 +1,43 @@
...
+          cache-from: type=gha,scope=cached-stage
+          cache-to: type=gha,scope=cached-stage,mode=max
+          outputs: type=cacheonly
```
dskloetd, line 42: Does this mean, that this flow doesn't produce anything and its only purpose is to speed up other flows through cache hits?

.github/workflows/docker-main.yaml
```diff
@@ -0,0 +1,43 @@
...
+          target: scratch
```
dskloetd, line 43: Why do you build a different stage here than in the code removed from `.github/workflows/docker-build.yaml`?

.github/workflows/docker-main.yaml
dskloetd, line 42: Looking at https://github.com/docker/build-push-action, it says "Following outputs are available" but type=cacheonly is not listed there.

.github/workflows/docker-main.yaml
bitdivine: See the PR description.  We build more so can cache more results.  If a PR changes just the aggregator code, for example, the docker build can get the nns-dapp front and backend builds from  cache.

.github/workflows/docker-main.yaml
bitdivine: I think the link in the documentation is maybe a bit misleading.  I tried getting to cacheonly just following the links,  without looking at the surrounding documentation, and indeed it is not obvious.  If you bounce up to look up to see "What is this thing I have been pointed to?" you get there but it's not as if you get pointed straight to the relevant chapter and verse of the docker bible.

.github/workflows/docker-main.yaml
dskloetd: The word "target" doesn't appear in the PR description...?

dskloetd (approved):
Thanks for the clarifications!

.github/workflows/docker-main.yaml
bitdivine: > On the other hand, we can cache more steps,
Here we are building the cache and the target defines which stages we build.

PR #2126, opened 2023-03-20
Benchmark docker build

bitdivine:
# Motivation
When trying to make something faster, it helps to have a benchmark, even if it is a very simple one.

# Changes
- Add a script that measures how long it takes for docker build to complete with various changes.

# Tests
None.  This is purely for local use.  I am using this locally.


bitdivine:
Not needed, I guess.  Not now anyway.

PR #2128, opened 2023-03-21
build: bump ic-js for ckBTC update_balance declaration

peterpeterparker:
# Motivation

Integrate PR https://github.com/dfinity/ic-js/pull/306.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -748,9 +748,9 @@
       }
     },
     "node_modules/@dfinity/ckbtc": {
-      "version": "0.0.2-next-2023-03-13.1",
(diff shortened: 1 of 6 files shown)


lmuntaner (commented):
This is a breaking change.

Does this mean we need to synchronize the deployment with another team?

peterpeterparker:
It's ckTESTBTC. They have already updated the ckTESTBTC minter on mainnet, yolo.

lmuntaner:
> It's ckTESTBTC. They have already updated the ckTESTBTC minter on mainnet, yolo.
> 
> (it's in test under feature flag)

Does that mean that current feature behind the flag doesn't work currently then?

lmuntaner (approved):

peterpeterparker:
> Does that mean that current feature behind the flag doesn't work currently then?

Probably but cannot try. We would need to get a successful result to try. No one is testing it currently but some might in upcoming days.

PR #2112, opened 2023-03-18
GIX-1336: Refactor restoring participation flow to project detail

lmuntaner:
# Motivation

Move logic to restore participation from the button component to the page component.

There is no new functionality in this PR. This is just refactoring.

# Changes

Main changes
* Move logic to restore participation from ParticipateButton to ProjectDetail.
* Refactor and organize the code in ProjectDetail.svelte to make it more understandable.
* It's worth taking a look at the ProjectDetail.svelte alone instead of the changes to better understand the flow. I'd like to simplify a bit this, but that will come in another PR.

# Tests

* In ParticipateButton, I removed all the tests regarding the restoring participation flow.
* In ProjectDetail page, I mocked apis instead of service and adapted tests to it.
* Main and biggest test: make sure that the commitment is updated if there is an open ticket in ProjectDetail page.

PR #2129, opened 2023-03-21
Chore: Refactor getCommitment

lmuntaner:
# Motivation

Simplify and refactor to simplify the PR #2112 

# Changes

* Refactor sns util `getCommitment`.
* Make `forceFetch` optional in `loadSnsSwapCommitment`.

# Tests

* Test the new `getCommitment` functionality

### DIFF
--- a/frontend/src/lib/utils/sns.utils.ts
+++ b/frontend/src/lib/utils/sns.utils.ts
@@ -195,11 +195,21 @@ export const getSwapCanisterAccount = ({
   return accountIdentifier;
 };
 
+/**
+ * Returns `undefined` if commitment is not loaded yet.
+ * Returns `BigInt(0)` if commitment is loaded but there is no ICP amount.
+ * Returns commitment e8s if commitment is defined.
+ */
 export const getCommitmentE8s = (
   swapCommitment?: SnsSwapCommitment | null
-): bigint | undefined =>
-  fromNullable(swapCommitment?.myCommitment?.icp ?? [])?.amount_e8s ??
-  undefined;
+): bigint | undefined => {
+  if (isNullish(swapCommitment) || isNullish(swapCommitment.myCommitment)) {
+    return undefined;
+  }
+  return (
+    fromNullable(swapCommitment.myCommitment.icp ?? [])?.amount_e8s ?? BigInt(0)
+  );
+};
 
 /**
  * Tests the error message against `refresh_buyer_token` canister function.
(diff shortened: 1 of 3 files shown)


frontend/src/tests/lib/utils/sns.utils.spec.ts
```diff
@@ -235,14 +235,24 @@ describe("sns-utils", () => {
...
         rootCanisterId: mockPrincipal,
         myCommitment: {
           icp: [],
         },
       };
-      expect(getCommitmentE8s(commitment)).toBeUndefined();
+      expect(getCommitmentE8s(commitment)).toEqual(BigInt(0));
+    });
+
+    it("returns commitment not loaded", () => {
```
dskloetd, line 248: Is something missing? Should it say
"returns undefined if commitment not loaded"
?

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -195,11 +195,21 @@ export const getSwapCanisterAccount = ({
...
   swapCommitment?: SnsSwapCommitment | null
-): bigint | undefined =>
-  fromNullable(swapCommitment?.myCommitment?.icp ?? [])?.amount_e8s ??
-  undefined;
+): bigint | undefined => {
+  if (isNullish(swapCommitment) || isNullish(swapCommitment.myCommitment)) {
+    return undefined;
+  }
+  return (
+    fromNullable(swapCommitment.myCommitment.icp ?? [])?.amount_e8s ?? BigInt(0)
```
dskloetd, line 210: Why wasn't returning 0 in this case needed before and why will it be needed?

frontend/src/tests/lib/utils/sns.utils.spec.ts
lmuntaner: yes, thanks

frontend/src/lib/utils/sns.utils.ts
lmuntaner: To differentiate whether the user has no commitment or it's just not loaded.

If it's not loaded, we don't want to trigger yet the restore participation.

If it's loaded, but there is no commitment, we want to restore the participation with a current commitment of 0.

The current commitment is important for validations and notifications and so on further in the flow.

lmuntaner:

dskloetd (approved):

PR #2131, opened 2023-03-21
refactor: move transaction components

peterpeterparker:
# Motivation

The transaction modal found place under `modals/accounts/NewTransaction` but some other modals - e.g. `IncreaseNeuronStakeModal` - were referencing it even though there are not "accounts" related.

`NewTransaction` folder also contained some components which we generally have under `components` and lower/uppercase.

I might also create some new types for `transaction` in upcoming PR.

# Changes

- move `modals/accounts/NewTransaction/Components.svelte` to `components/transaction/Components.svelte`
- move `modals/accounts/NewTransaction/TransactionModal.svelte` to `modals/transaction/TransactionModal.svelte`
- move tests accordingly

# Note

No functional changes, files moved around only.


lmuntaner (approved):
Thanks!

PR #2138, opened 2023-03-21
Don't use relative imports in page objects

dskloetd:
# Motivation

We want to share page objects with e2e tests.
Currently page objects in the same directory import each other via relative path imports.
I was not able to get relative imports to work with Playwright.

# Changes

Use absolute (`$tests/page-object/*`) imports from page objects.

# Tests

`npm run test`


lmuntaner (approved):
I hope the IDE picks it up.

But otherwise it won't work, right?

dskloetd:
I guess we'll find out. It's OK if one slips through. I'll fix it when necessary.

PR #2139, opened 2023-03-21
feat: reload accounts on "Finish" action of "Receive" modal

peterpeterparker:
# Motivation

Unlike on the "Wallet" page we noticed that the accounts were not reloaded we click "Finish" in the QR Receive modal on the "Accounts" page. This would be useful as user might transfer ICP and expect the accounts to be reloaded once the transaction has been made.

# Changes

- call sync ICP/Sns/ckBTC accounts on "Finish" of "Reload" modal
- rename property `reloadAccount` of the modal data into `reload`

### DIFF
--- a/frontend/src/lib/components/accounts/NnsAccountsFooter.svelte
+++ b/frontend/src/lib/components/accounts/NnsAccountsFooter.svelte
@@ -11,6 +11,7 @@
   const openNewTransaction = () => (modal = "NewTransaction");
   const closeModal = () => (modal = undefined);
 
+  // TODO: for performance reason use `loadBalance` to reload specific account
   const reload = async () => await syncAccounts();
 </script>
 
--- a/frontend/src/lib/components/accounts/NnsAccountsFooter.svelte
+++ b/frontend/src/lib/components/accounts/NnsAccountsFooter.svelte
@@ -4,7 +4,7 @@
   import Footer from "$lib/components/layout/Footer.svelte";
   import IcpTransactionModal from "$lib/modals/accounts/IcpTransactionModal.svelte";
   import { nonNullish } from "@dfinity/utils";
-  import Receive from "$lib/components/accounts/ReceiveButton.svelte";
+  import ReceiveButton from "$lib/components/accounts/ReceiveButton.svelte";
 
   let modal: "NewTransaction" | undefined = undefined;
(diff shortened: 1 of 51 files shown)


frontend/src/lib/components/accounts/NnsAccountsFooter.svelte
```diff
@@ -4,11 +4,14 @@
...
   import { nonNullish } from "@dfinity/utils";
-  import Receive from "$lib/components/accounts/ReceiveButton.svelte";
+  import ReceiveButton from "$lib/components/accounts/ReceiveButton.svelte";
+  import { syncAccounts } from "$lib/services/accounts.services";
 
   let modal: "NewTransaction" | undefined = undefined;
   const openNewTransaction = () => (modal = "NewTransaction");
   const closeModal = () => (modal = undefined);
+
+  const reload = async () => await syncAccounts();
```
lmuntaner, line 14: This could be improved by using `loadBalance` for the specific account, but not needed right now.

frontend/src/tests/lib/components/accounts/SnsAccountsFooter.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+import { testAccountsModal } from "../../../utils/accounts.test-utils";
+
+jest.mock("$lib/services/sns-accounts.services", () => {
+  return {
+    syncSnsAccounts: jest.fn().mockResolvedValue(undefined),
+  };
+});
+
+describe("SnsAccountsFooter", () => {
+  // const rootCanisterId = Principal.fromText("aaaaa-aa");
```
lmuntaner, line 31: can this be removed?

lmuntaner (approved):
Approving but I added two minor comments

frontend/src/tests/lib/components/accounts/SnsAccountsFooter.spec.ts
peterpeterparker: Definitely, thanks

frontend/src/lib/components/accounts/NnsAccountsFooter.svelte
peterpeterparker: Added a TODO

PR #2140, opened 2023-03-21
GIX-1219: Fetch SNS proposal and load it in component

lmuntaner:
# Motivation

User should see the data of an SNS proposal.

This PR: Load the proposal in the component.

# Changes

* New sns proposal service `getSnsProposalById`.
* Use the new service in page `SnsProposalDetail.svelte` to load the proposal in a local variable.
* Redirect the user if there is an error loading the proposal.

# Tests

* Test new sns proposal service.
* Add test for page SnsProposalDetail and test that user is redirected in the different scenarios.
* Add an errors map in the sns governace api fake and a function to set an error for the queryProposal api function.


lmuntaner:
@dskloetd please review

PR #2147, opened 2023-03-22
Add PageObjectElement type

dskloetd:
# Motivation

We want to reuse page objects between unit tests and e2e tests.
Currently page objects each have a root element of type `Element`, which can not be used for e2e tests.
I want to introduce an interface `PageObjectElement` which abstracts the functionality of Element and will have separate implementations for unit tests and e2e tests.
In this PR, I introduce the `PageObjectElement` but it's still equal to `Element`. This allows me to let page objects work with `PageObjectElement` while not changing anything else yet.

# Changes

1. Add `type PageObjectElement = Element`
2. Declare the root elements of page objects of type PageObjectElement instead of Element (or instead of HTMLElement in some accidental cases).

# Tests

`npm run test`
`npm run check`


lmuntaner (approved):
LGTM!

PR #2151, opened 2023-03-22
refactor: transaction modal init param

peterpeterparker:
# Motivation

Transaction modal has various properties. Some are values, some are initial parameters, such as providing the default source account.

When we will implement payment deep linking, there will be also muliple parameters (destination, token, amount). 

That's why this PR introduce an object for the initial parameters of the transaction modal. That way we can distinguish better what a parameters and also prepare the deep linking (as we will be able to pass down a single objects).

# Changes

No features change, refactoring of properties only.


lmuntaner (approved):
LGTM, thanks!

PR #2152, opened 2023-03-22
Add Jest specific implementation of PageObjectElement

dskloetd:
# Motivation

We want to share page objects between Jest unit tests and Playwright e2e tests.
So we provide an interface abstracting the element to use by page objects.
This interface will have different implementations per platform.
This PR adds a first version of the Jest implementation.

# Changes

1. Add JestPageObjectElement and use it in all the tests to create page objects.
2. Change PageObjectElement to an interface implemented by JestPageObjectElement.
3. Add just a few methods on PageObjectElement to make current tests pass.

In the future we will change the interface to be more async so that it can be compatible with e2e but this will require more changes to the tests and page objects to accommodate so it will be different PRs.

# Tests

npm run test
npm run check

### DIFF
--- a/frontend/src/tests/lib/components/project-detail/ProjectSwapDetails.spec.ts
+++ b/frontend/src/tests/lib/components/project-detail/ProjectSwapDetails.spec.ts
@@ -13,6 +13,7 @@ import {
   mockSummary,
 } from "$tests/mocks/sns-projects.mock";
 import { renderContextCmp } from "$tests/mocks/sns.mock";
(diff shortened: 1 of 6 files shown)


lmuntaner (approved):
LGTM, just a further question:

PageObject should never use HtmlElement directly, right?

Or is that ok as long as it's used only here?

dskloetd:
PageObjects should never use HTMLElements directly.
If they need to, then we'll add more functionality to PageObjectElement to expose it to both platforms.

PR #2153, opened 2023-03-22
GIX-1219: Manage errors when fetching SNS proposal data

lmuntaner:
# Motivation

User should be redirected to the SNS proposals list if there is some problem loading the specific proposal.

# Changes

Changes in SnsProposalDetail.svelte page:
* Redirect if the parameter can't be parsed to a bigint.
* Redirect if there is an error when loading the proposal

# Tests

* Add test for SnsProposalDetail.svelte page. Add test case for each of the above scenarios.
* Add error capabilities to the sns governance api fake.

### DIFF
--- a/frontend/src/lib/pages/SnsProposalDetail.svelte
+++ b/frontend/src/lib/pages/SnsProposalDetail.svelte
@@ -8,6 +8,7 @@
   import { getSnsProposalById } from "$lib/services/$public/sns-proposals.services";
   import { snsOnlyProjectStore } from "$lib/derived/sns/sns-selected-project.derived";
   import type { SnsProposalData, SnsProposalId } from "@dfinity/sns";
+  import { toastsError } from "$lib/stores/toasts.store";
 
   export let proposalIdText: string | undefined | null = undefined;
 
@@ -32,16 +33,42 @@
         getSnsProposalById({
           rootCanisterId: $snsOnlyProjectStore,
           proposalId,
+  // By setting a local variable, we avoid calling the below text when the whole store is changes.
+  let rootCanisterId = $snsOnlyProjectStore;
+
+  // TODO: Fix race condition in case the user changes the proposal before the first one hasn't loaded yet.
+  $: {
+    if (nonNullish(proposalIdText) && nonNullish(rootCanisterId)) {
+      // We can't be sure that `snsOnlyProjectStore` is the same when `handleError` is called.
+      const rootCanisterIdText = rootCanisterId.toText();
+      try {
+        const proposalId = BigInt(proposalIdText);
+        proposal = "loading";
+        getSnsProposalById({
+          rootCanisterId,
+          proposalId: { id: proposalId },
           setProposal: ({ proposal: proposalData }) => {
             proposal = proposalData;
           },
           handleError: () => {
             // TODO: redirect
+            goto(buildProposalsUrl({ universe: rootCanisterIdText }), {
+              replaceState: true,
+            });
           },
         });
       } catch (error) {
         proposal = "error";
         // TODO: Add a toast for this error and redirect
+        toastsError({
+          labelKey: "error.wrong_proposal_id",
+          substitutions: {
+            $proposalId: proposalIdText,
+          },
+        });
+        goto(buildProposalsUrl({ universe: rootCanisterIdText }), {
+          replaceState: true,
+        });
       }
     }
   }
--- a/frontend/src/lib/pages/SnsProposalDetail.svelte
+++ b/frontend/src/lib/pages/SnsProposalDetail.svelte
@@ -24,42 +24,38 @@
     }
   });
 
+  // By setting a local variable, we avoid calling the below text when the whole store is changes but the value is the same.
+  let rootCanisterIdText: undefined | string;
+
   // TODO: Fix race condition in case the user changes the proposal before the first one hasn't loaded yet.
   $: {
-    if (nonNullish(proposalIdText) && nonNullish($snsOnlyProjectStore)) {
+    if (
+      nonNullish(proposalIdText) &&
+      nonNullish($snsOnlyProjectStore) &&
+      // Call the block only if the root canister id has changed.
+      $snsOnlyProjectStore.toText() !== rootCanisterIdText
+    ) {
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/pages/SnsProposalDetail.svelte
```diff
@@ -23,9 +24,21 @@
     }
   });
 
+  // By setting a local variable, we avoid calling the below text when the whole store is changes but the value is the same.
```
dskloetd, line 27: This comment is very unclear. Can you be specific?
What changes?
What stays the same?
What "text" (?) do we need to avoid calling and why?

frontend/src/lib/pages/SnsProposalDetail.svelte
```diff
@@ -23,9 +24,21 @@
...
+  let rootCanisterIdText: undefined | string;
```
dskloetd, line 28: Could we instead have
```
$: let rootCanisterIdText = $snsOnlyProjectStore.toText();
```

And then just trigger getSnsProposalById when $rootCanisterIdText changes (and, I guess, reconstruct the Principal from the text)? Because it's a string I think it won't have the issue of the Principal that it can be equal but different.

frontend/src/lib/pages/SnsProposalDetail.svelte
```diff
@@ -23,9 +24,21 @@
...
-    if (nonNullish(proposalIdText) && nonNullish($snsOnlyProjectStore)) {
+    if (
+      nonNullish(proposalIdText) &&
+      nonNullish($snsOnlyProjectStore) &&
+      // Call the block only if the root canister id has changed.
+      $snsOnlyProjectStore.toText() !== rootCanisterIdText
+    ) {
+      rootCanisterIdText = $snsOnlyProjectStore.toText();
+      // We need this to be used in the handleError callback.
+      // Otherwise, TS doesn't believe that the value of `rootCanisterIdText` won't change.
```
dskloetd, line 40: And it's right about that, right? It actually might change.
Shall we call it something like rootCanisterIdAtTimeOfRequest?

frontend/src/lib/pages/SnsProposalDetail.svelte
```diff
@@ -23,9 +24,21 @@
...
+      const scopedId = rootCanisterIdText;
```
dskloetd, line 41: Optional: Instead you could also create the `{universe: rootCanisterIdText}` object here and use it both in line 52 and line 65.

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -235,6 +242,31 @@ async function queryProposals({
...
+  rootCanisterId,
+  certified: _,
+  proposalId,
+}: {
+  rootCanisterId: Principal;
+  identity: Identity;
+  certified: boolean;
+  proposalId: SnsProposalId;
+}): Promise<SnsProposalData | undefined> {
+  if (errorsMap.has(ImplementationsKeys.queryProposal)) {
```
dskloetd, line 256: How about still using mockRejectOnce etc. for errors?
You can still define mock behavior the old way as well.
And errors you often don't want to persist so using the "once" functions of jest is convenient.
And I expect after 1 error happens, it will automatically go back to the fake provided behavior.
And you can of course always also just use a non-existed proposal ID to trigger an error.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as snsGovernanceFake from "$tests/fakes/sns-governance-api.fake";
+import { mockAuthStoreNoIdentitySubscribe } from "$tests/mocks/auth.store.mock";
+import { mockCanisterId } from "$tests/mocks/canisters.mock";
+import { blockAllCallsTo } from "$tests/utils/module.test-utils";
+import { render, waitFor } from "@testing-library/svelte";
+import { get } from "svelte/store";
+
+jest.mock("$lib/api/sns-governance.api");
+
+const blockedApiPaths = ["$lib/api/sns-governance.api"];
```
dskloetd, line 19: You don't need to block this if you also install the same fake.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+          proposalIdText: "invalid",
+        },
+      });
+      await waitFor(() => {
+        const { path } = get(pageStore);
+        return expect(path).toEqual(AppPath.Proposals);
+      });
+    });
+
+    it("should redirect to the list of sns proposals if proposal is not found", async () => {
```
dskloetd, line 48: Can you also add a test that does not have an error?
Because as currently tested it would also pass if there is always a redirect.

frontend/src/lib/pages/SnsProposalDetail.svelte
lmuntaner: I wanted to avoit reconstructing the Principal, but sure, why not.

frontend/src/lib/pages/SnsProposalDetail.svelte
lmuntaner: sure, let me try again

frontend/src/lib/pages/SnsProposalDetail.svelte
lmuntaner: Yes, it's correct. Sure, I can rename

frontend/src/tests/fakes/sns-governance-api.fake.ts
lmuntaner: I can throw an error if not found. I believe that's how the api behaves actually.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
lmuntaner: I will do that in the next PR with the skeleton, page object and placeholder for the content.

lmuntaner:

dskloetd (approved):

PR #2154, opened 2023-03-22
Make PageObjectElement.getText() async

dskloetd:
# Motivation

We want to share page objects between unit tests and e2e tests.
In e2e tests some things have to be async because they involve making a request to the browser driver.
So for compatibility they have to be async in the platform agnostic interface.
One such things is getting the text content of an element.

# Changes

1. Make PageObjectElement.getText() async.
2. Make everything derived from it async as well.

# Tests

npm run test
npm run check

### DIFF
--- a/scripts/update_ic_commit
+++ b/scripts/update_ic_commit
@@ -32,24 +32,24 @@ source "$(clap.build)"
 
 : "Get the latest sns wasm and did files"
 {
-  : "Remove any existing did and wasm files from the build directory"
-  rm -f target/ic/*.{did,wasm}
-  : "Get or build the did and wasm files"
-  e2e-tests/scripts/nns-canister-build
-  : "Apparently there is a convention of putting did files here:"
-  : "It would be nice to have a general rule rather than listing did files individually"
-  : "and to have a way of making sure that these are in sync with the paths in dfx.json."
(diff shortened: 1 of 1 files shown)


frontend/src/tests/page-objects/AmountDisplay.page-object.ts
```diff
@@ -15,8 +15,9 @@ export class AmountDisplayPo {
     return el && new AmountDisplayPo(el);
   }
 
-  getAmount(): string | null {
-    return assertNonNullish(this.root.querySelector(`[data-tid="token-value"]`))
-      .textContent;
+  getAmount(): Promise<string> {
```
lmuntaner, line 18: `getText` doesn't return `null` now?

lmuntaner (approved):
Thanks! Approving, but I added a small question

frontend/src/tests/page-objects/AmountDisplay.page-object.ts
dskloetd: textContent/getText() was never null.
I think this method could return null before the assertNonNullish was added and the annotation wasn't updated.

lmuntaner: Ok, thanks, that makes sense

PR #2156, opened 2023-03-22
Chore: Refactor NNS Proposal cards

lmuntaner:
# Motivation

Make the cards of the Proposal page agnostic of the parent.

# Changes

* Remove margin-top from ProposalProposerActionsEntry.svelte
* Remove margin-top from ProposalProposerPayloadEntry.svelte
* Remove the component NnsProposalProposeDataSection.svelte that was just rendering two components.
* Move those two components to NnsProposal.svelte and manage the spacing there with a display flex and gap.

# Tests

* Mostly UI changes, but I added a similar test that was in the removed component to the NnsProposal.spec.


lmuntaner:
@dskloetd please review

PR #2157, opened 2023-03-22
Add Base class for page objects.

dskloetd:
# Motivation

There will be some things convenient to share between all page objects like click() and isPresent().
Having a share base class makes these easy to add.

# Changes

1. Add BasePageObject class which holds the root element.
2. Make all page objects extend BasePageObject.
3. Pass the root element to the super constructor instead of holding on to it in each page object class.

# Tests

npm run test
npm run check


lmuntaner (approved):
LGTM

PR #2158, opened 2023-03-22
fix: settings back

peterpeterparker:
# Motivation

Go to "Accounts" -> select a "Wallet" -> Go to "Settings" -> Back in the header -> No query params in URL -> error not found

# Changes

- if we navigate to settings in the dapp, then let's use the browser history api

# Tests

Would need an e2e test...

# Screenshots

Capture d’écran 2023-03-22 à 16 27 01

Capture d’écran 2023-03-22 à 16 26 56


frontend/src/routes/(app)/(nns)/settings/+layout.svelte
```diff
@@ -10,12 +10,16 @@
...
 
-  const back = (): Promise<void> =>
-    goto(
-      nonNullish(referrerPath) && referrerPath !== AppPath.Settings
-        ? referrerPath
-        : AppPath.Accounts
-    );
+  const back = async () => {
+    if (nonNullish(referrerPath)) {
+      // Referrer might be a detail page which needs query parameters therefore we use the browser API to go back here
```
lmuntaner, line 15: Typo: "to go back there".

lmuntaner (approved):
Thanks!

We were wondering how to add tests for this layout components or the whole route.

Then we could add a normal test also.

frontend/src/routes/(app)/(nns)/settings/+layout.svelte
peterpeterparker: Thx

PR #2159, opened 2023-03-22
Make presence checking on page objects async

dskloetd:
# Motivation

We want to reuse page objects for e2e tests.
Playwright uses "Locator"s which refer to specific elements.
But to be able to chain them synchronously (such set `getTab().getCard().getButton()`), they might refer to elements that don't exist. And you can only find out if they exist via a synchronous call.
For compatibility we change the Jest page objects to work the same.

# Changes

1. `JestPageObjectElement` may now hold `null` elements but its `querySelector()` will never return `null`. Instead it might return a `JestPageObjectElement` with a null element.
2. Checking if a `JestPageObjectElement` refers to an element that exists is now done async via `isPresent()`.
3. `JestPageObjectElement.querySelectorAll()` is now async. It can't be deferred because there is no way to know the size of the list in advance.
4. `PageObjectElement.getText()` now returns `null` if the object doesn't exist.
5. The `under` static function on page object classes now always returns non-null but it might refer to an element that turns out not to exist when checked for presence.
6. The `allUnder` static function on page object classes is now async. It can't be deferred because there is no way to know the size of the list in advance.
7. Add an `isPresent()` method to `BasePageObject` so every page object can be checked for presence, which would have previously been done with `nonNullish()`.
8. If a page object has an `isContentLoaded()` function that checks if a list of skeletons is empty, make sure that the page object itself is present. Otherwise the list of skeletons will be empty while the object itself isn't even there and certainly not loaded.
9. Add necessary awaits on presence checks.

# Tests

`npm run test`
`npm run check`


frontend/src/tests/page-objects/NeuronDetail.page-object.ts
```diff
@@ -11,32 +10,33 @@ export class NeuronDetailPo extends BasePageObject {
...
+    return this.getSnsNeuronDetailPo().isPresent();
   }
 
-  isContentLoaded() {
-    return (
-      this.getNnsNeuronDetailPo()?.isContentLoaded() ||
-      this.getSnsNeuronDetailPo()?.isContentLoaded() ||
-      false
-    );
+  async isContentLoaded(): Promise<boolean> {
```
lmuntaner, line 35: Didn't you say that you need to check for `isPresent` in itself first?

```
async isContentLoaded(): Promise<boolean> {
  if (await this.isPresent()) {
    const [nnsLoaded, snsLoaded] = await Promise.all([
      this.getNnsNeuronDetailPo().isContentLoaded(),
      this.getSnsNeuronDetailPo().isContentLoaded(),
    ]);
    return nnsLoaded || snsLoaded;
  }
  return false;
}
```

frontend/src/tests/page-objects/Neurons.page-object.ts
```diff
@@ -11,32 +10,31 @@ export class NeuronsPo extends BasePageObject {
...
+    return this.getSnsNeuronsPo().isPresent();
   }
 
-  isContentLoaded() {
-    return (
-      this.getNnsNeuronsPo()?.isContentLoaded() ||
-      this.getSnsNeuronsPo()?.isContentLoaded() ||
-      false
-    );
+  async isContentLoaded(): Promise<boolean> {
```
lmuntaner, line 33: ditto with `isPresent`

frontend/src/tests/page-objects/jest.page-object.ts
```diff
@@ -1,27 +1,36 @@
...
   constructor(element: Element) {
     this.element = element;
   }
 
-  querySelector(selector: string): JestPageObjectElement | null {
-    const el = this.element.querySelector(selector);
-    return el && new JestPageObjectElement(el);
+  querySelector(selector: string): JestPageObjectElement {
+    const el = this.element && this.element.querySelector(selector);
+    return new JestPageObjectElement(el);
```
lmuntaner, line 16: `el` could be `null`, right?

That means that the constructor receives `Element | null`

frontend/src/tests/page-objects/jest.page-object.ts
```diff
@@ -1,27 +1,36 @@
...
+import { isNullish, nonNullish } from "@dfinity/utils";
 
 /**
  * An implementation of the PageObjectElement interface for Jest unit tests.
  */
 export class JestPageObjectElement implements PageObjectElement {
-  private readonly element: Element;
+  private readonly element: Element | null;
 
   constructor(element: Element) {
```
lmuntaner, line 10: Shouldn't this be `Element | null`?

frontend/src/tests/page-objects/NnsNeurons.page-object.ts
```diff
@@ -10,25 +10,28 @@ export class NnsNeuronsPo extends BasePageObject {
...
+    return (
+      (await this.isPresent()) && (await this.getSkeletonCardPos()).length === 0
+    );
   }
 
-  getNeuronIds(): Promise<string[]> {
-    const cards = this.getNeuronCardPos();
+  async getNeuronIds(): Promise<string[]> {
+    const cards = await this.getNeuronCardPos();
     return Promise.all(cards.map((card) => card.getNeuronId()));
```
lmuntaner, line 35: Do you need the `Promise.all` here?

If it's an `async` function it will already return a promise.

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-object.ts
```diff
@@ -11,52 +10,57 @@ export class SnsNeuronInfoStakePo extends BasePageObject {
...
 
-  getIncreaseDissolveDelayButtonPo() {
+  getIncreaseDissolveDelayButtonPo(): ButtonPo {
     return this.getButton("sns-increase-dissolve-delay");
   }
 
-  hasIncreaseDissolveDelayButton() {
-    return nonNullish(this.getIncreaseDissolveDelayButtonPo());
+  hasIncreaseDissolveDelayButton(): Promise<boolean> {
+    return this.getIncreaseDissolveDelayButtonPo().isPresent();
```
lmuntaner, line 48: Do I understand it right that, only when checking for skeletons not to be loaded we need to also check for the `isPresent` on the same PO?

Calling `isContentLoaded` in a child PO doesn't need the `isPresent` in the current PO.

frontend/src/tests/page-objects/SnsNeurons.page-object.ts
```diff
@@ -10,25 +10,28 @@ export class SnsNeuronsPo extends BasePageObject {
...
+    return (
+      (await this.isPresent()) && (await this.getSkeletonCardPos()).length === 0
+    );
   }
 
-  getNeuronIds(): Promise<string[]> {
-    const cards = this.getNeuronCardPos();
+  async getNeuronIds(): Promise<string[]> {
+    const cards = await this.getNeuronCardPos();
     return Promise.all(cards.map((card) => card.getNeuronId()));
```
lmuntaner, line 35: Ditto `Promise.all`

frontend/src/tests/page-objects/base.page-object.ts
```diff
@@ -6,4 +6,8 @@ export class BasePageObject {
   constructor(root: PageObjectElement) {
     this.root = root;
   }
+
+  isPresent(): Promise<boolean> {
+    return this.root.isPresent();
```
lmuntaner, line 11: Can't `root` be `null`?

frontend/src/tests/page-objects/jest.page-object.ts
```diff
@@ -1,27 +1,36 @@
 import type { PageObjectElement } from "$tests/types/page-object.types";
+import { isNullish, nonNullish } from "@dfinity/utils";
 
 /**
  * An implementation of the PageObjectElement interface for Jest unit tests.
  */
 export class JestPageObjectElement implements PageObjectElement {
-  private readonly element: Element;
+  private readonly element: Element | null;
```
lmuntaner, line 8: Btw, I just realized, shouldn't this be called `root`? like in the base page object?

lmuntaner (commented):
A few questions, thanks!

frontend/src/tests/page-objects/NeuronDetail.page-object.ts
dskloetd: Only IF the only other logic is checking the absence of something (a skeleton). Because those things might be absent because this entire component is absent.
But in this case, if either NnsNeurons or SnsNeurons is loaded (and thus present), it means that this component itself must be present as well.

frontend/src/tests/page-objects/Neurons.page-object.ts
dskloetd: Answered above.

frontend/src/tests/page-objects/NnsNeurons.page-object.ts
dskloetd: Promise.all turns `Promise<string>[]` into `Promise<string[]>`. Does that answer your question?

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-object.ts
dskloetd: Correct. Because if the isContentLoaded on the child is implemented correctly, it means that the child is present. And it's not possible for the parent to be absent if the child is present.

frontend/src/tests/page-objects/SnsNeurons.page-object.ts
dskloetd: Answered above?

frontend/src/tests/page-objects/base.page-object.ts
dskloetd: Correct. PageObjectElement might hold a null element, but it will never be null itself.

frontend/src/tests/page-objects/jest.page-object.ts
dskloetd: A page object represents an entire component, of which 1 element is the root.
But this class represents an element.
The root of the component is an element of type PageObjectElement.
Does that make sense?

frontend/src/tests/page-objects/jest.page-object.ts
dskloetd: Yes, good catch! I wonder why it wasn't caught. Done.

frontend/src/tests/page-objects/jest.page-object.ts
dskloetd: Indeed.

frontend/src/tests/page-objects/NnsNeurons.page-object.ts
lmuntaner: ah, yes of course! thanks for the explanation

frontend/src/tests/page-objects/SnsNeuronInfoStake.page-object.ts
lmuntaner: Yes, sorry, I didn't get it in my first comment.

frontend/src/tests/page-objects/jest.page-object.ts
lmuntaner: Ahh, yes sorry! In the PR I sometimes lose the big picture.

This is the class used inside the PO. Got it, thanks

lmuntaner (approved):
Thanks!

PR #2160, opened 2023-03-22
Fix: Copy contents of .dfx

bitdivine:
# Motivation
Strange bug, whereby locally `./scripts/docker-build --network XXX` was not picking up the aggregator canister ID.

The dockerfile pulls in .dfx, precisely to support locally deployed, non-standard canister IDs.

On closer inspection, .dfx was being pulled in but as an empty directory.

`.*` is listed in .dockerignore.  But exempting `.dfx` in `.dockerignore` alone does not help.

In the COPY documentation, it states that if `src` is a directory, it must end in a `/`.  That did not help.

But having a separate copy command for `.dfx` does work.

# Changes
* Exempt `.dfx` from `.dockerignore`.
* Move the `.dfx` copy command to a separate step.

# Tests
* Manually tested that a local aggregator canister ID is now picked up.
* Manually tested that if I then remove `.dfx`, the locally deployed aggregator canister ID is no longer found.
How to test this in CI in a reasonable time?  It would be necessary to find  away of sharing the docker cache with the buildx action.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,6 +2,7 @@
 .*
 !.rustfmt.toml
 !.gitignore
+!.dfx/**/*
 
 build-inputs.txt
 assets.tar.xz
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Is this maybe the reason why the SNS Aggregator hash was not matching?

PR #2161, opened 2023-03-22
fix: "estimated amount received BTC" should not be displayed for ckBTC network

peterpeterparker:
# Motivation

"Estimated amount received: xxxxx BTC" should not be displayed for ckBTC.

# Changes

- use related component according network

# Screenshot

Capture d’écran 2023-03-22 à 17 37 41


lmuntaner (approved):
Thanks!

PR #2164, opened 2023-03-22
feat: do not display token symbol in labels of transactions list

peterpeterparker:
# Motivation

Remove token symbol in labels of transactions list because the information is redundant. 

# Changes

- only update labels and keep logic, who knows in the future we might have to redo this

# Screenshots

Capture d’écran 2023-03-22 à 18 25 51
Capture d’écran 2023-03-22 à 18 26 36


lmuntaner (commented):
Why do you think it might change in the future?

I'm not sure about having unused code.

peterpeterparker:
> Why do you think it might change in the future?
> I'm not sure about having unused code.

I dont' like unused code neither, agree with you, but it was decided quickly in a meeting so I would not be surprised if later someone comes and request something like "Yeah let's not display the symbol for receive but le'ts display it again for top up neuron" or something. Guts feeling.

If after a bit no one does so, we can remove the code.

lmuntaner (approved):
Sounds like a plan, I understand the feeling.

PR #2166, opened 2023-03-23
Use cdk v2

bitdivine:
# Motivation

# Changes

# Tests


lmuntaner (approved):
Thanks!

Can you add an issue and set it as TODO? We should upgrade the rest after we have the parameters working

PR #2168, opened 2023-03-23
Allow SNS governance API fake to be paused

dskloetd:
# Motivation

Sometimes we want to test a component in a loading state.
For the component to be in this state, often an API call needs to not be resolved yet.
Tests might rely on promises not resolving instantly but this is brittle.
Better is to explicitly only resolve the promise after the expectations are made.
A nice way to do this is to put a fake API in a "paused" state until the expectation have been made.

# Changes

1. Allow SNS governance API fake to be paused.
2. Pause the SNS governance API in `frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts` to make the test less brittle. Otherwise it would fail with PR #2159.

# Future

Other fakes can be done in separate PRs.

# Tests

I tested the code in PR #2159 and tested that it makes the test consistent even with many extra `await null`s.

### DIFF
--- a/frontend/src/tests/fakes/sns-governance-api.fake.ts
+++ b/frontend/src/tests/fakes/sns-governance-api.fake.ts
@@ -121,7 +121,6 @@ const wrapMaybePaused = async <T>(fn: () => Promise<T>): Promise<T> => {
   return responsePromise;
 };
 
-
 ////////////////////////
 // Fake implementations:
 ////////////////////////
@@ -150,7 +149,7 @@ function getNervousSystemFunctions({
   certified: boolean;
 }): Promise<SnsNervousSystemFunction[]> {
   return wrapMaybePaused(async () => {
-      return nervousFunctions.get(rootCanisterId.toText()) || [];
+    return nervousFunctions.get(rootCanisterId.toText()) || [];
   });
 }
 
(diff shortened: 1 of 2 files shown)


frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -270,6 +309,25 @@ const reset = () => {
...
+};
+
+export const resume = () => {
+  if (!isPaused) {
+    throw new Error("The fake wasn't paused.");
+  }
+  for (const call of pendingCalls) {
+    call();
+  }
+  pendingCalls.length = 0;
```
lmuntaner, line 336: Why didn't you set `pendingCalls = []`?

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -102,11 +107,31 @@ const getNervousFunctions = (rootCanisterId: Principal) => {
   return nervousFunctionsList;
 };
 
+/**
+ * Calls the passed function and returns its result.
+ * If the fake is paused, the function will be queued and an unresolved promise
+ * is returned which will resolve when the fake is resumed and the function
+ * called.
+ */
+const wrapMaybePaused = async <T>(fn: () => Promise<T>): Promise<T> => {
```
lmuntaner, line 116: Very cool!

lmuntaner (approved):
A question, already approving!

frontend/src/tests/fakes/sns-governance-api.fake.ts
dskloetd: `pendingCalls` is const. So I clear the existing list instead of assigning a new list.

PR #2169, opened 2023-03-23
Make it possible to pause fake governance API

dskloetd:
# Motivation

When expecting a component to be in loading state we shouldn't rely on arbitrary promises not to be resolved yet.

# Changes

1. Make it possible to pause fake governance API same as fake SNS governance API.
2. Pause the API in tests that explicitly expect a component to be in loading state.
3. Add missing `isPaused = false` in resume of fake SNS governance API.

# Tests

`npm run test`

### DIFF
--- a/frontend/src/tests/fakes/governance-api.fake.ts
+++ b/frontend/src/tests/fakes/governance-api.fake.ts
@@ -21,6 +21,11 @@ const neurons: Map<string, NeuronInfo[]> = new Map();
 
 const mapKey = (identity: Identity) => identity.getPrincipal().toText();
 
+// When the fake is paused, all calls to the fake will be queued until the fake
+// is resumed.
+let isPaused = false;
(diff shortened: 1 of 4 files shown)


lmuntaner (approved):
You need to rebase and there will be less changes, right?

dskloetd:
No, the changes that look familiar were previously in the SNS governance API and this is in the Governance API.
That's why I'm wondering if it should be a generic wrapper for all APIs.

lmuntaner:
> No, the changes that look familiar were previously in the SNS governance API and this is in the Governance API. That's why I'm wondering if it should be a generic wrapper for all APIs.

Ah, ok, true. I didn't realized it.

Yeah, we can have some helpers maybe.

lmuntaner:
@dskloetd fyi, I rebased and fixed the conflicts

PR #2172, opened 2023-03-23
Use page objects in example e2e test

dskloetd:
# Motivation

Page objects now work with both Jest unit tests and Playwright e2e tests.
The example test now demonstrates this.
In the next PR I want to write a proper e2e for the accounts tab.

# Changes

1. Add a Playwright specific implementation of `PageObjectElement`.
2. Add a click function to PageObjectElement and ButtonPo.
3. Add minimal page objects for `Accounts` and `NnsAccountsFooter`.
4. Add TestIdWrappers to the components for their page objects.

# Tests

npm run test-e2e
npm run test

### DIFF
--- a/frontend/src/tests/e2e/example.spec.ts
+++ b/frontend/src/tests/e2e/example.spec.ts
@@ -1,3 +1,5 @@
+import { AccountsPo } from "$tests/page-objects/Accounts.page-object";
+import { PlaywrightPageObjectElement } from "$tests/page-objects/playwright.page-object";
 import { signInWithNewUser } from "$tests/utils/e2e.test-utils";
 import { expect, test } from "@playwright/test";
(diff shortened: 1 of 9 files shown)


frontend/src/tests/e2e/example.spec.ts
lmuntaner, line 1: Aren't the e2e in its own directory?

lmuntaner (approved):
I'm loving this!

Just a question because I was surprised to see the e2e test inside the frontend directory.

frontend/src/tests/e2e/example.spec.ts
dskloetd: The new playwright e2e tests are/will be in `frontend/src/tests/e2e`.

PR #2174, opened 2023-03-23
Pass unsubscribe functions directly to onDestroy

dskloetd:
Note: I have done limited manual testing but I'm not 100% sure how to exercise all this code.
Please review carefully and feel free to suggest specific scenarios to test.
I'm also happy to split the PR into multiple PRs if some things need extra testing or for better reviewing.

# Motivation

When storing subscriptions to unsubscribe later it's possible to forget to unsubscribe.
Svelte stores can be used with `$` without storing subscriptions.

# Changes

For unsubscribe functions which are called only during onDestroy, use svelte `$` auto subscriptions instead.

# Tests

Ran `npm run test`.
I have done limited manual testing.


peterpeterparker (commented):
I have to say I find this refactoring a bit "ugly" and misleading.

- "ugly" just personal code style preference which of course is not a reason to reject a PR
- But misleading because reading "onDestroy" first implies expecting something that will be destroyed and not nessecary monted at the same time which isn't that obvious when you read the code the way you propose
- in Angular projects never seen a pattern where `ngOnDestroy` would subscribe and unsubscribe to any stream - i.e. don't have experience / seen such a pattern applied in any projects

So for these reasons, I'm against this personally.

If anything should be done in my point of view regarding these "subscribers", I would rather remove the subscribe and use auto subscription instead if possible.

dskloetd:
That's fair.
I wasn't 100% sure myself but when I showed it to Llorenc, he seemed to really like it so I went ahead.
Originally I had something like this in mind:
```
disposer.add(subscrbibe...
...
disposer.add(otherSubscrbibe...
...
onDestroy(() => {
  disposer.dispose();
}
```
But then you still need to remember to call dispose in onDestroy, so when I realized you can call onDestroy multiple times, it seemed like good solutions. How would you feel about a utility function called `unsubscribeOnDestroy`? It would do the same thing but the name would more clearly signal intent.

Of course it's better to use auto subscription, but the reality is that we do sometimes explicitly subscribe. And also this pattern isn't only applicable to unsubscribing but to anything that might need to be cleaned up (cancelPolling for example). 

So I would really like to establish some pattern of how to always schedule the cleanup at the same time as making the thing that needs to be cleaned up.

peterpeterparker:
We originally used `subscribe` because we did not knew about auto subscription, or I did not knew about it. Then we learned about it and we began to generalize its usage. That's why I would bet that most of the remaining `subscribe` can actually be replaced with auto subscription. That's why, if anything must be done, I would first give a shot trying to convert these to auto subscription. Then, if it remains some `subscribe`, I would search to establish a pattern but, kind of feel, guts feeling, that it might not be the case.

peterpeterparker:
Thanks for the nice feedback 👍. Beside auto subscribe note above, I have to say the `disposer` does sound interesting, it's similar to an Angular/RxJS pattern I know.

dskloetd:
I've now replaced all the onDestroy calls in this PR with auto subscriptions.

Note: I have done limited manual testing but I'm not 100% sure how to exercise all this code.
Please review carefully and feel free to suggest specific scenarios to test.
I'm also happy to split the PR into multiple PRs if some things need extra testing or for better reviewing.

lmuntaner (commented):
Just one comment, I like the pattern.

frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
```diff
@@ -62,22 +61,22 @@
...
     }
-  );
 
-  onDestroy(unsubscribe);
+    loading = true;
+    await syncSnsAccounts({ rootCanisterId: selectedProjectCanisterId });
+    loading = false;
+  };
+
+  $: syncProjectAccountIfNotAvailable($snsOnlyProjectStore);
```
peterpeterparker, line 79: not sure that change much, just sharing just in case.

when async, I do following:

```
$: (async () => await syncProject($snsOnlyProjectStore))();
```

frontend/src/lib/pages/Login.svelte
```diff
@@ -13,21 +11,21 @@
 
   let signedIn = false;
 
-  const unsubscribe: Unsubscriber = authStore.subscribe(
-    async ({ identity }: AuthStoreData) => {
-      signedIn = isSignedIn(identity);
+  const redirectWhenSignedIn = async ({
```
peterpeterparker, line 14: this one being quite a sensitive binding, would be maybe nice to have it in another PR to ease testing, what do you think?

frontend/src/routes/(app)/+layout.svelte
```diff
@@ -12,12 +12,9 @@
 
   onMount(async () => await Promise.all([initAppAuth(), initAppPublicData()]));
 
-  const unsubscribeVoteInProgress = voteRegistrationStore.subscribe(
-    ({ registrations }) =>
-      syncBeforeUnload(voteRegistrationActive(registrations))
+  $: syncBeforeUnload(
```
peterpeterparker, line 15: we should think about testing this one manually because it covers closing the browser when the action is in progress (just speaking out at loud)

frontend/src/routes/+layout.svelte
```diff
@@ -40,9 +40,7 @@
     await syncAuth($authStore);
   });
 
-  const unsubscribeAuth = authStore.subscribe(syncAuth);
-
-  onDestroy(() => unsubscribeAuth());
+  $: syncAuth($authStore);
```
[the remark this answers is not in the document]
peterpeterparker: You cannot pass parameter to a `+layout`.

Yes I think we still need the `onMount` because if I remember correctly there would be a race condition on mainnet in the initialization of the auth and worker if we remove it.

peterpeterparker (commented):
Looks good 👍. Minor comments.

frontend/src/routes/+layout.svelte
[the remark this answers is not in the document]
peterpeterparker: You cannot pass parameter to a +layout.

Yes I think we still need the onMount because if I remember correctly there would be a race condition on mainnet in the initialization of the auth and worker if we remove it.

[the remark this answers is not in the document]
lmuntaner: I meant to `syncAuth`:

```
onMount(() => {
  worker = await initWorker();
});

$: syncAuth($authStore, worker);
```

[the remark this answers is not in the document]
peterpeterparker: Yeah I understand.

Original commit history:

Capture d’écran 2023-03-29 à 07 44 00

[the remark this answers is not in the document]
peterpeterparker: I did not understand the prop though. Might work but tricky to test

frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
dskloetd: Did you mean
```
$: (async () => await syncProjectAccountIfNotAvailable($snsOnlyProjectStore))();
```
?
Because `syncProjectAccountIfNotAvailable` does more than just `syncSnsAccounts`.
If yes, what's the purpose of the more complicated syntax?
As far as I can tell it does exactly the same.

peterpeterparker: > Following comment is just for context and to answer your question, nothing about our app. Stricly answer.

Not sure in Svelte but in other languages - e.g. Stencil - handling promises on mount has an impact on the rendering of the components (function call becomes render blocker or non blocker). That's why, as a personal guideline, I always use async / await unless I explicitely want to have an impact on the rendering. Likewise, generally speaking in my code, I always handle promise unless there is a good reason to not too. That way, I know that if I did not used async / await, it was an explicit choice for a good and particular reason.

---

As I said, I was just sharing the syntax in case it would be interesting, nothing more, nothing less. So let's keep the syntax as in your PR, all good.

frontend/src/lib/pages/Login.svelte
peterpeterparker: https://github.com/dfinity/nns-dapp/pull/2202

frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
dskloetd: Thanks for clarifying!

peterpeterparker (approved):
LGTM, thanks

Since @lmuntaner also reviewed it, probably cool to merge it once it's fine for him too

frontend/src/routes/+layout.svelte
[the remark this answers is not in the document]
lmuntaner: Did you try this @dskloetd ?

I'm approving anyway because I don't want to block the merge for this only

lmuntaner (approved):
Approving, one question remaining though

dskloetd:
Thanks for the approvals but it a bit premature. I still need to look into some of the suggestions. I look at this if I'm blocked on other things so it's a bit slower than my other PRs.

frontend/src/routes/(app)/+layout.svelte
dskloetd: I'm having a bit of trouble understanding what this is about.
Is voteRegistrationStore a store of votes on proposals that have been made by the user in the UI but potentially haven't yet been committed to the blockchain?
Is voteRegistrationActive just a function that tells me whether or not there are any such pending votes?

Looking at syncBeforeUnload, it adds or removes beforeunload listeners. I'm a bit confused by this.
It looks generic but will break if it's used from more than 1 place, right?
Because if there are 2 reasons to require syncing before unload, and 1 of them becomes false, the handler is removed even though there is still 1 reason remaining.

Anyway, can you confirm my understanding is correct?

peterpeterparker: > Is voteRegistrationStore a store of votes on proposals that have been made by the user in the UI but potentially haven't yet been committed to the blockchain?

Yes.

Voting takes time so the frontend display the vote in the UI as soon as the user has voted and makes the calls on the IC in the background (displaying potential error and rollbacking UI accordingly). This is a feature users really likes because before it, voting for multiple proposals a day took ages.

While the vote is happening in the background, a toast with a spinner is displayed and in addition, we register a before unload event. That way, if user close is tab / browser, the browser displays an alert "Are you sure you want to leave?" or something.

peterpeterparker: Note: with "background" I mean "in the background on the client side - still in the browser"

dskloetd: OK, I voted on a proposal, tried to close the tab but got the confirmation dialog.
After the vote went through I was able to close the tab without confirmation dialog.

peterpeterparker: Great! Thanks for the test.

frontend/src/routes/+layout.svelte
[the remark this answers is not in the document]
dskloetd: I'm not sure what is the suggestions here?
Should I remove the line `await syncAuth($authStore);` from onMount() and then change this line to
```
$: worker, syncAuth($authStore);
```
?
And then how would I test that it's safe?

lmuntaner (approved):

PR #2176, opened 2023-03-25
Use docker build for sns aggregator test

bitdivine:
# Motivation

# Changes

# Tests

### DIFF
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -12,43 +12,26 @@ jobs:
   curl_test:
     runs-on: ubuntu-20.04
     timeout-minutes: 60
+    strategy:
+      matrix:
+        include:
+          - BUILD_NAME: "local"
+            DFX_NETWORK: "local"
     steps:
       - name: Checkout nns-dapp
         uses: actions/checkout@v3
-      - name: set-env
-        run: ./build-config.sh >> $GITHUB_ENV
-      # Cache based on the Cargo.lock
-      # The cache key is always an exact match or no match (i.e. no
-      # "restore-keys"-style matching). Because (in case of an exact match)
-      # GitHub actions won't (re-)upload the cache after the build, it means that
-      # our cache won't just grow forever.
-      - uses: actions/cache@v3
+      - name: Set up docker buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Build sns-aggregator image
+        uses: docker/build-push-action@v3
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: ${{ runner.os }}-cargo-build-${{ hashFiles('**/Cargo.lock') }}
-      # Cache the ic-cdk-optimizer
-      - uses: actions/cache@v3
-        with:
-          path: |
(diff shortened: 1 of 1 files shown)


scripts/deploy-snsdemo-testnet
```diff
@@ -52,4 +52,4 @@ dfx-canister-set-id --canister_name sns_aggregator --canister_id "$AGGREGATOR_CA
 dfx canister info sns_aggregator
 
 : Deploy the aggregator
-RUSTFLAGS="--cfg feature=\"reconfigurable\"" dfx deploy sns_aggregator --network "$DFX_NETWORK"
+dfx canister install sns_aggregator --wasm sns_aggregator.wasm --upgrade-unchanged --mode reinstall --yes
```
bitdivine, line 55: dfx deploy, as on the left, builds the wasm.  Now we use the pre-built wasm instead.

Note: This deploy script is about to be thrown away, so there is little value in making this fancy.

.github/workflows/aggregator.yaml
```diff
@@ -15,40 +15,18 @@ jobs:
...
-        run: |
-          set -x
-          dfx --version
-          node --version
-          npm --version
-          rustc --version
-          cargo --version
-          ic-cdk-optimizer --version
+          context: .
+          file: Dockerfile
```
dskloetd, line 24: Where do you specify that you build the aggregator? Or do you build everything?

scripts/deploy-snsdemo-testnet
dskloetd, line 55: Why do you need `--upgrade-unchanged`?

What does `--mode reinstall` do and why do you need it?
(I read `dfx canister install --help` but it doesn't explain this.)

dskloetd, line 55: Don't you need to either keep `--network "$DFX_NETWORK"` or `export DFX_NETWORK`?

.github/workflows/aggregator.yaml
bitdivine: This builds everything.

scripts/deploy-snsdemo-testnet
bitdivine: If that wasm is already installed, it will not be reinstalled otherwise, even if the arguments are different.  Here no argument is provided, so the default arguments are used.  If there is some existing install with weird arguments, subsequent tests could yield correspondingly weird results.  Maybe that is fine because: "When the going gets weird, the weird get pro."

bitdivine: 1. This is used exclusively for CI, which runs locally, and
2.  in about 3 days, this script can be deleted

.github/workflows/aggregator.yaml
dskloetd: Do we need everything here?

dskloetd (approved):

scripts/deploy-snsdemo-testnet
bitdivine: I would prefer to put the comment on arguments in the dfx --help, as that is the root of the problem and so will help in many more places.  But we can put a comment here for now.  DONE

dskloetd (approved):

PR #2178, opened 2023-03-25
Docker build e2e

bitdivine:
# Motivation
Faster CI.  We already have builds cached in docker.  Reusing the cache means that we don't need to rebuild unless we need to.  Furthermore, even on cache miss, a docker build takes about 4 minutes which is about as long as it takes just to install the native build tools without even starting to build.

# Changes
- docker-build the nns-dapp
- Don't build locally
- Remove unneeded tool installation and an unused env var.

# Tests
See CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,6 +78,11 @@ jobs:
       - name: Install SNS script dependencies
         run: |
           dfx-sns-demo-install
+      - name: Get node version
+        run: jq -r '"NODE_VERSION=\(.defaults.build.config.NODE_VERSION)"' dfx.json >> $GITHUB_ENV
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
       - name: Deploy NNS and SNS canisters
         run: |
           # Note: This deploys standard NNS canisters but with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,9 +100,9 @@ jobs:
           SCREENSHOT=1 npm run test |& tee -a chrome-wdio.log
       - name: Run firefox e2e tests
         working-directory: e2e-tests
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -13,15 +9,18 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-env:
-  NNS_CACHING_KEY: dv-0004
```
bitdivine, line 17: `NNS_CACHING_KEY` appears nowhere in the repo, apart from here where  it is unused.  So delete.

.github/workflows/build.yml
```diff
@@ -13,15 +9,18 @@ on:
...
+defaults:
+  run:
+    shell: bash -euxlo pipefail {0}
```
bitdivine, line 14: `run` commands don't give detailed output unless `set -x` is set at the beginning of every multiline `run` command.  Invariably the log one needs is one where `set -x` has not yet been set, so one has to add `set -x` and rerun, which costs time.  Better set it here once for all.

.github/workflows/build.yml
```diff
@@ -42,6 +41,16 @@ jobs:
...
+        with:
+          context: .
+          file: Dockerfile
+          build-args: |
+            DFX_NETWORK=local
+          cache-from: type=gha,scope=cached-stage
+          # Exports the artefacts from the final stage
+          outputs: .
       - name: Install Software
         run: |
```
bitdivine, line 55: We can actually delete a fair bit of code, but that obscures this PR so is left to another PR.

bitdivine: On the other hand this is slow (almost 4m).  And iteration with this correspondingly painful.

dskloetd (approved):

PR #2179, opened 2023-03-25
chore: Sync clap.bash

bitdivine:
# Motivation
The clap.bash argument parser is imported from the snsdemo repo.  The upstream code now has tests.  The functionality has hardly changed but it would be better to use the tested version.

# Changes
*  Update clap.bash to the latest version.

# Tests
See CI.


dskloetd:
If I understand correctly all the changes here are forked from the snsdemo repo, correct?
Do you want me to review the actual changes here or just rubber stamp the fact that clap.bash is the only changed file and CI still passes?

PR #2180, opened 2023-03-25
Set -eu by default

bitdivine:
# Motivation
`run` commands don't give detailed output unless set -x is set at the beginning of every multiline run command. Invariably the log one needs is one where set -x has not yet been set, so one has to add set -x and rerun, which costs time. Better set it at the beginning of the build once for all.

Note: This does cause several lines of startup bash commands to be printed for every `run` command.  On balance, I think it is worth it.

# Changes
-  Set `-euxo pipefail` by default for every run command in `build.yaml`

# Tests
See CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,8 +13,9 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-env:
-  NNS_CACHING_KEY: dv-0004
+defaults:
+  run:
+    shell: bash -euxlo pipefail {0}
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -13,8 +13,9 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-env:
-  NNS_CACHING_KEY: dv-0004
```
dskloetd, line 17: Is this unused?

dskloetd (approved):

.github/workflows/build.yml
bitdivine: Yes

PR #2185, opened 2023-03-27
GIX-1220: SNS proposal summary section

lmuntaner:
# Motivation

User can see the summary of an SNS proposal.

# Changes

* Implement SnsProposalSummarySection.svelte

# Tests

* New page object `ProposalProposerInfoSectionPo`
* Use new page object in test for `ProposalProposerInfoSection`.
* Test for SnsProposalSummarySection


lmuntaner:
New UI:

Screenshot 2023-03-27 at 12 50 08

Inside the bigger component:

Screenshot 2023-03-27 at 13 31 20

lmuntaner:

frontend/src/tests/lib/components/sns-proposals/SnsProposalSummarySection.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      );
+      expect(await po.getProposalTitle()).toBe(title);
+    });
+
+    it("should contain summary", async () => {
+      const { getByText } = render(SnsProposalSummarySection, {
+        props,
+      });
+
+      await waitFor(() => expect(getByText(summary)).toBeInTheDocument());
```
dskloetd, line 49: I would expect something like
```
expect(po.getSummary()).toBe(summary);
```
Would that work?
It follows the same pattern as the other tests so imposes less cognitive load on the reader.

Or does the current test test something that wouldn't be tested by my suggestion?

frontend/src/tests/lib/components/sns-proposals/SnsProposalSummarySection.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    it("should render url", async () => {
+      const { container } = render(SnsProposalSummarySection, {
+        props,
+      });
+
+      const po = ProposalProposerInfoSectionPo.under(
+        new JestPageObjectElement(container)
+      );
+      expect(await po.getUrlText()).toBe(url);
```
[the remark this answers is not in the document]
lmuntaner: The proposals have a "title", "summary" and "url".

The url is set when you create a proposal and it shuold link to a page that explains the proposal further.

frontend/src/tests/lib/components/sns-proposals/SnsProposalSummarySection.spec.ts
lmuntaner: good idea, done!

frontend/src/tests/lib/components/sns-proposals/SnsProposalSummarySection.spec.ts
```diff
@@ -31,32 +40,21 @@ describe("SnsProposalSummarySection", () => {
...
     });
 
     it("should contain summary", async () => {
-      const { getByText } = render(SnsProposalSummarySection, {
-        props,
-      });
+      const po = renderComponent(props);
 
-      await waitFor(() => expect(getByText(summary)).toBeInTheDocument());
+      await waitFor(async () =>
```
dskloetd, line 51: Does it not pass without the waitFor?
And if you make renderComponent async (and let it resolve promises)?

frontend/src/lib/components/proposal-detail/ProposalSummary.svelte
```diff
@@ -8,7 +8,7 @@
   $: showTitle = $$slots.title !== undefined;
 </script>
 
-<div class="markdown">
```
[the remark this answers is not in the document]
lmuntaner: No, it's fine to me.

We never formally agreed on what it meant the suffix `-component`. But for me it's ok if it means the parent of a component.

frontend/src/tests/lib/components/sns-proposals/SnsProposalSummarySection.spec.ts
lmuntaner: I used the same pattern with `resolvePromises` and it worked, thanks!

PR #2187, opened 2023-03-27
First bit of accounts e2e test.

dskloetd:
# Motivation

Moving to Playwright tests which reuse page objects from unit tests.
Starting with accounts tests for the 4 accounts related requirements.

# Changes

1. Add frontend/src/tests/e2e/accounts.spec.ts with just the beginnings of the e2e test, testing that there is a Main account by default.
2. Add page objects for NnsAccounts and AccountCard.
3. Add AppPo to refer to top-level page objects.

I'm not sure about AppPo but it's easy to change later so I think it makes sense to go with it for now and have another look when there is some more stuff there.
But happy to do something different if you have a suggestion.

# Tests

In 3 different terminals:
```
dfx start --clean
scripts/dev-local.sh
npm run test-e2e
```


lmuntaner (approved):
LGTM, thanks!

PR #2188, opened 2023-03-27
Accounts e2e test: Add sub account

dskloetd:
# Motivation

Migrating to the Playwright e2e tests with reused page objects.

# Changes

1. Extend the accounts e2e test to create a sub account.
2. Delete example e2e tests as the accounts test is now good enough as example.
3. Add test IDs to components that need it.
4. Add page objects for AddAccountModal, AddAccountType, AddSubAccount, TextInput, NnsAddAccount
5. Add getButton and getTextInput to BasePageObject for convenience. This requires ButtonPo and TextInputPo to not extend BasePageObject.
6. Make test ID in ButtonPo.under optional.
7. Remove getButton from SnsNeuronInfoStake.page-object.ts as it is now provided by BasePageObject.
8. Add waitFor, waitForAbsent and type, to PageObjectElement and its implementations. Although the Jest implementations still throw "not implemented". We'll implement them as we start using them.
9. 

# Tests

npm run test
npm run test-e2e
Tested manually that the add account modal renders correctly on both Mac and Android (Chrome).


frontend/src/lib/components/accounts/NnsAddAccount.svelte
```diff
@@ -10,17 +11,19 @@
   const closeModal = () => (modal = undefined);
 </script>
 
-{#if modal === "AddAccountModal"}
-  <AddAccountModal on:nnsClose={closeModal} />
-{/if}
+<TestIdWrapper testId="nns-add-account-component">
```
peterpeterparker, line 14: As I already staten in another PR, I think we should not using this component because it adds meaningless `div` in the DOM because it makes the DOM heavier and slower to manipulate.

In addition, here particularly it adds an additional div around a modal that should appears as absolute positioning and can therefore make the styling more complicated or broken at some point (if not yet the case)

frontend/src/lib/modals/accounts/AddAccountModal.svelte
```diff
@@ -85,23 +86,25 @@
   let modal: WizardModal;
 </script>
 
-<WizardModal {steps} bind:currentStep bind:this={modal} on:nnsClose>
-  <svelte:fragment slot="title"
-    >{currentStep?.title ?? $i18n.accounts.add_account}</svelte:fragment
-  >
+<TestIdWrapper testId="add-account-modal-component">
```
peterpeterparker, line 89: ditto

frontend/src/tests/page-objects/AddAccountModal.page-object.ts
```diff
@@ -0,0 +1,28 @@
+import { AddAccountTypePo } from "$tests/page-objects/AddAccountType.page-object";
+import { AddSubAccountPo } from "$tests/page-objects/AddSubAccount.page-object";
+import { BasePageObject } from "$tests/page-objects/base.page-object";
+import type { PageObjectElement } from "$tests/types/page-object.types";
+
+export class AddAccountModalPo extends BasePageObject {
```
peterpeterparker, line 6: what does the `Po` accronym mean (out of curiosity)?

frontend/src/tests/page-objects/AddAccountModal.page-object.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  private constructor(root: PageObjectElement) {
+    super(root);
+  }
+
+  static under(element: PageObjectElement): AddAccountModalPo {
+    return new AddAccountModalPo(element.byTestId(AddAccountModalPo.tid));
+  }
+
+  getAddAccountTypePo(): AddAccountTypePo {
```
peterpeterparker, line 17: if you declare these as arrow methodes these can be deconstructed by consumer

```
getAddAccountTypePo = (): AddAccountTypePo => {
```

frontend/src/tests/page-objects/AddAccountModal.page-object.ts
```diff
@@ -0,0 +1,28 @@
...
+import type { PageObjectElement } from "$tests/types/page-object.types";
+
+export class AddAccountModalPo extends BasePageObject {
+  static readonly tid = "add-account-modal-component";
+
+  private constructor(root: PageObjectElement) {
+    super(root);
+  }
+
+  static under(element: PageObjectElement): AddAccountModalPo {
```
peterpeterparker, line 13: nth: you can destruct `{byTestId}`

frontend/src/tests/page-objects/AddAccountModal.page-object.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  getAddAccountTypePo(): AddAccountTypePo {
+    return AddAccountTypePo.under(this.root);
+  }
+
+  getAddSubAccountPo(): AddSubAccountPo {
+    return AddSubAccountPo.under(this.root);
+  }
+
+  waitForClosed(): Promise<void> {
```
peterpeterparker, line 25: don't know the context, does it override a declaration of `BasePageObject`? if yes, can you use the TypeScript `override` keyword?

```
override waitForClosed = (): Promise<void> => {
```

frontend/src/tests/page-objects/Button.page-object.ts
```diff
@@ -1,21 +1,32 @@
...
     element,
     testId,
   }: {
     element: PageObjectElement;
-    testId: string;
+    testId?: string;
   }): ButtonPo {
+    if (isNullish(testId)) {
+      return new ButtonPo(element.querySelector("button"));
+    }
```
peterpeterparker, line 22: should `data-tid` be a constant in the test?

frontend/src/tests/page-objects/Input.page-object.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  static under({
+    element,
+    testId,
+  }: {
+    element: PageObjectElement;
+    testId?: string;
+  }): InputPo {
+    if (isNullish(testId)) {
```
peterpeterparker: this looks really similar to the `ButtonPo`, do you think there is a chance to avoid even a bit more code duplication?

frontend/src/tests/page-objects/NnsAccounts.page-object.ts
```diff
@@ -17,4 +18,17 @@ export class NnsAccountsPo extends BasePageObject {
     // We assume the first card is the main card.
     return AccountCardPo.under(this.root);
   }
+
+  getNnsAddAccountPo(): NnsAddAccountPo {
+    return NnsAddAccountPo.under(this.root);
+  }
+
+  async addAccount(name: string): Promise<void> {
+    await this.getNnsAddAccountPo().clickAddAccount();
```
peterpeterparker, line 27: Personal opinion so please ignore if you rather like but, I find this not particularly easily readable. Something with the repetation of `.getSomethingPo()` makes it heavy. Maybe with some destruction it can be made more smooth 🤷‍♂️

frontend/src/tests/page-objects/Button.page-object.ts
peterpeterparker: (I meant)

```
export const DATA_TID = "data-tid";
```

frontend/src/tests/page-objects/NnsAddAccount.page-object.ts
```diff
@@ -0,0 +1,23 @@
...
+  private constructor(root: PageObjectElement) {
+    super(root);
+  }
+
+  static under(element: PageObjectElement): NnsAddAccountPo {
+    return new NnsAddAccountPo(element.byTestId(NnsAddAccountPo.tid));
+  }
+
+  getAddAccountModalPo(): AddAccountModalPo {
+    return AddAccountModalPo.under(this.root);
```
lmuntaner, line 17: I would get modals at the page level.

For example, `AccountsPo.getAddAccountModalPo()` instead of the specific component where they appear.

Modals are hidden by default and adding the extra div might create styling conflicts. Moreover, a lot of modals are also rendered from it's own component that listeners to an event triggered by a button.

Check the `AccountsModals.svelte` and follow the trace for example.

frontend/src/lib/components/accounts/NnsAddAccount.svelte
lmuntaner: I agree in this specific case that the wrapper is used as parent of a Modal.

We could put the data-tid directly in the modal and get it at the page level, not from the component is being used.

Yet, I do believe that adding data-tids brings value to the user. Improved testing is in the long term beneficial for the user because there are less bugs and we're faster to develop. I believe that it's worth the performance degradation.

There are other features we could add to improve performance, like caching requests.

frontend/src/lib/modals/accounts/AddAccountModal.svelte
lmuntaner: I agree with a parent of `WizardModal`, we probably don't want that. It could lead to styling issues.

I'd rather add it in the Wizard itself as prop `testId`.

Or use a div for the content instead of a `<svelte:fragment>`.

frontend/src/tests/page-objects/playwright.page-object.ts
```diff
@@ -40,7 +40,19 @@ export class PlaywrightPageObjectElement implements PageObjectElement {
...
+
+  waitForAbsent(): Promise<void> {
+    return expect(this.locator).toHaveCount(0);
+  }
+
   click(): Promise<void> {
     return this.locator.click();
   }
+
+  type(text: string): Promise<void> {
```
lmuntaner, line 55: The name `type` is confusing to me. I don't read the verb, instead I expect it to give me the type of the value.

Could we use `write` or `typeText`.

lmuntaner (commented):
A few comments, looking good though!

frontend/src/lib/modals/accounts/AddAccountModal.svelte
peterpeterparker: > Or use a div for the content instead of a <svelte:fragment>.

No.

```
div
  div
    div
      div
        div
          div
            div
              div
                div
                div
                  div
                    div
                      div
                        div
                          div
                            div
                              div
                                div
                                div
                                  div
                                    div
                                      div
                                        div
                                          div
                                            div
                                              div
                                                div
```

Such a DOM is frontend development bad practice, heavier and slower to manipulate therefore bad for the user.

In addition, as I said above, it can lead to styling issues and/or break the styles.

We should not use div that have no runtime meaning.

frontend/src/lib/components/accounts/NnsAddAccount.svelte
peterpeterparker: > Yet, I do believe that adding data-tids ...

I'm fine with `data-tid`. I'm against puring the dom with additional HTML elements for testing purpose.

frontend/src/tests/page-objects/Input.page-object.ts
dskloetd: I've added a second base class `SimpleBasePageObject` to remove some (and in the future more) duplication between BasePageObject, ButtonPo and TextInputPo.

I also change InputPo to TextInputPo as I expect e.g. checkbox will have its own page object.

frontend/src/lib/modals/accounts/AddAccountModal.svelte
dskloetd:
```
div
  div
    div
      div
        div
          div
            div
              div
                div
                div
                  div
                    div
                      div
                        div
                          div
                            div
                              div
                                div
                                div
                                  div
                                    div
                                      div
                                        div
                                          div
                                            div
                                              div
                                                div
```

David, please try not to make straw men. It does not make your argument stronger. Let's stick to the facts and keep the conversation in a single thread.

frontend/src/tests/page-objects/NnsAddAccount.page-object.ts
dskloetd: I think it's important that the page object hierarchy reflects the component hierarchy. Otherwise it becomes a mess and hard to find anything. If the modal should not be in the button component, we can change the component itself (and then also the page object). But we shouldn't diverge the page object from the component.

It does not create styling issues. See my other comment.

frontend/src/tests/page-objects/playwright.page-object.ts
dskloetd: Changed to typeText.

frontend/src/lib/components/accounts/NnsAddAccount.svelte
peterpeterparker: > These divs have display: contents. From the spec

Buttons are notably not accessible with `display: contents` because partialy supported accross browsers

https://caniuse.com/?search=display%3A%20contents

> Why do you think it would make styling more difficult?

From experience, maintaining `z-index` accross the app is particularly tricky and implementing the `Island` component lead to issues with the modals which did not bubble anymore. So we had to move many modals and I fear that adding additional divs will/might lead to these kind of issues.

>  it should be possible to not include these wrapper divs on mainnet or to only include them if a feature flag is enabled. Would that work for you? Or do you have another alternative to suggest?

Thanks but it's better to stick to e2e tests that test same version on mainnet.

I'm fine with `data-tid`, if we continue to use these next to elements that have purposes.

frontend/src/lib/modals/accounts/AddAccountModal.svelte
peterpeterparker: https://github.com/dfinity/nns-dapp/pull/2188#discussion_r1149482296

frontend/src/tests/page-objects/AddAccountModal.page-object.ts
peterpeterparker: thx

frontend/src/tests/page-objects/AddAccountModal.page-object.ts
peterpeterparker: `byTestId` is an arrow function?

frontend/src/tests/page-objects/AddAccountModal.page-object.ts
peterpeterparker: Gotcha

frontend/src/tests/page-objects/NnsAccounts.page-object.ts
peterpeterparker: Better yes.

frontend/src/lib/components/accounts/NnsAddAccount.svelte
dskloetd: We are not applying `display: contents` to buttons. We are only applying it to divs.

If you can be more specific about the issues you fear, we can add tests to make sure they won't happen (again). But we need it to be easy to add tests or we'll forever be fighting issues in production instead of catching them early.

frontend/src/tests/page-objects/Button.page-object.ts
dskloetd: Like this?
https://github.com/dfinity/nns-dapp/blob/main/frontend/src/tests/page-objects/playwright.page-object.ts#L28

frontend/src/tests/page-objects/NnsAccounts.page-object.ts
dskloetd: I pushed some logic down to the children. Should be more readable now.

peterpeterparker: Thanks

frontend/src/lib/components/accounts/NnsAddAccount.svelte
peterpeterparker: > But we need it to be easy to add tests 

I think this is already solved by using `data-tid` without the need to add extra HTML elements.

Anyhow, you and @lmuntaner I understand are fine with this approach, so be it. I guess we will not agree on this, let's move on.

Thanks for the anwers.

frontend/src/lib/modals/accounts/AddAccountModal.svelte
```diff
@@ -85,7 +85,13 @@
   let modal: WizardModal;
 </script>
 
-<WizardModal {steps} bind:currentStep bind:this={modal} on:nnsClose>
+<WizardModal
+  testId="add-account-modal-component"
```
dskloetd, line 89: I was able to remove this one TestIdWrapper by setting the testId directly on the WizardModal.

lmuntaner: great, thanks!

frontend/src/tests/page-objects/NnsAddAccount.page-object.ts
lmuntaner: That makes sense.

Maybe the problem is that the modal is next to the button.

lmuntaner (approved):
LGTM, thanks!

PR #2189, opened 2023-03-28
Chore: Rename ProposalProposerInfoSection

lmuntaner:
# Motivation

Rename components to make clearer what they render.

# Changes

* Rename file NnsProposalProposerInfoSection.svelte to NnsProposalSummarySection.svelte
* Rename file ProposalProposerInfoSection.svelte to ProposalSummarySection.svelte
* Rename data tid to be consistent with file rename.

# Tests

* Rename testing file for each of the renamed files.
* Change data tids after change.

### DIFF
--- a/frontend/src/tests/page-objects/ProposalSummarySection.page-object.ts
+++ b/frontend/src/tests/page-objects/ProposalSummarySection.page-object.ts
@@ -1,16 +1,16 @@
 import { BasePageObject } from "$tests/page-objects/base.page-object";
 import type { PageObjectElement } from "$tests/types/page-object.types";
 
-export class ProposalProposerInfoSectionPo extends BasePageObject {
-  static readonly tid = "proposal-proposer-info-component";
+export class ProposalSummarySectionPo extends BasePageObject {
+  static readonly tid = "proposal-summary-section-component";
 
   private constructor(root: PageObjectElement) {
     super(root);
   }
 
-  static under(element: PageObjectElement): ProposalProposerInfoSectionPo {
-    return new ProposalProposerInfoSectionPo(
-      element.byTestId(ProposalProposerInfoSectionPo.tid)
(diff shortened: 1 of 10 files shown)


frontend/src/tests/page-objects/ProposalSummarySection.page-object.ts
```diff
@@ -1,16 +1,16 @@
...
+  static readonly tid = "proposal-summary-section-component";
 
   private constructor(root: PageObjectElement) {
     super(root);
   }
 
-  static under(element: PageObjectElement): ProposalProposerInfoSectionPo {
-    return new ProposalProposerInfoSectionPo(
-      element.byTestId(ProposalProposerInfoSectionPo.tid)
+  static under(element: PageObjectElement): ProposalSummarySectionPo {
```
dskloetd, line 11: Isn't there a getter on another PO that calls this and has to be renamed?

dskloetd (approved):
Thanks!

frontend/src/tests/page-objects/ProposalSummarySection.page-object.ts
lmuntaner: No, I use this PO for the other component tests.

The SnsProposalSummarySection only renders this component. Therefore, I used the same PO.

dskloetd: I'm not following what is "the other component test" and what is "this component".

But I just checked it for myself. I was just (wrongly) assuming there was already an NnsProposalPo and that it would have a getProposalProposerInfoSectionPo() which would have to be renamed.
Hence my question.
But it doesn't exist so...

Oh, but you reuse it for SNS. So I would have expected SnsProposalDetailPo to have the getter. But it also doesn't.

PR #2193, opened 2023-03-28
build: chunk qr-code to improve boot time

peterpeterparker:
# Motivation

QR code libraries are currently loaded at boot time because we deliver a single vendor chunk. I am guessing that QR code reader and generator might not be used by all clients, therefore deferring their loading to their effective usage is probably an interesting strategy.

# Note

This PR also creates a dedicated chunk for the HW wallet. If we would not do so, the bundler would place the commonjs polyfill in the QR code chunk and therefore would also loads it at boot time.

Downside is that we will load one more chunk at boot time (19 instead of 18 JS files) but, we will spare 400kb non gzipped.

# PRs

- [x] https://github.com/dfinity/gix-components/pull/179

# Changes

- add rules to manual chunking strategy

# Test

Tested manually with prod build (`npm run build && npx serve public`) both QR code features and HW wallet (Create account, send ICP) ✅

# Screenshots

Vendor.js current state:

is

Vendor.js afterwards:

Capture d’écran 2023-03-28 à 13 56 09

Boot current state:

is-boot

Boot afterwards:

after-boot

Chunk when opening a QR code function:

Capture d’écran 2023-03-28 à 13 25 11


lmuntaner (approved):
Thanks!

lmuntaner (approved):
Thanks!

PR #2194, opened 2023-03-28
fix: reset selected account on changes if not exists in selectable accounts

peterpeterparker:
# Motivation

Fix selected account reset. See screenshots and scenario.

# Changes

- if the list of selectable accounts changes and the selected account is not part of it, select the first

# Screenshots and scenario

Capture d’écran 2023-03-28 à 14 34 57
Capture d’écran 2023-03-28 à 14 35 34
Capture d’écran 2023-03-28 à 14 36 03
Capture d’écran 2023-03-28 à 14 36 55
Capture d’écran 2023-03-28 à 14 37 23
Capture d’écran 2023-03-28 à 14 38 19


peterpeterparker:
@lmuntaner when you review this, can you also test it? I'm a bit afraid of side effect and cyclic loop. Better be sure.

lmuntaner:
> @lmuntaner when you review this, can you also test it? I'm a bit afraid of side effect and cyclic loop. Better be sure.

Sorry, today I didn't have time, I'll do that tomorrow.

peterpeterparker:
No worries, I don't want to rush it in the release candidate anyway. Thanks!

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
```diff
@@ -87,6 +83,33 @@ describe("SelectAccountDropdown", () => {
       );
     });
 
+    it("should reset selected accounts on selectable list of accounts change", async () => {
+      const { component } = render(SelectAccountDropdown, {
```
lmuntaner, line 87: `render` returns a `rerender` function that it would be more useful to use.

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
```diff
@@ -87,6 +83,33 @@ describe("SelectAccountDropdown", () => {
...
+        props: {
+          ...props,
+          selectedAccount: mockHardwareWalletAccount,
+        },
+      });
+
+      expect(component.$$.ctx[component.$$.props["selectedAccount"]]).toEqual(
```
lmuntaner, line 94: Why don't you check the value of the `<select>` element?

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
```diff
@@ -87,6 +83,33 @@ describe("SelectAccountDropdown", () => {
...
+        props: {
+          ...props,
+          selectedAccount: mockHardwareWalletAccount,
+          filterAccounts: (account) => !isAccountHardwareWallet(account),
+        },
+      });
+
+      await waitFor(() =>
+        expect(
+          component2.$$.ctx[component2.$$.props["selectedAccount"]]
```
lmuntaner, line 108: Ditto on checking the `<select>` element

lmuntaner (commented):
Thanks for fixing it!

I added a few questions regarding how to test it.

I pulled the branch, tried it and it all worked well!

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
peterpeterparker: I know! That was ma original idea but it did not work out. `rerender` reset the account store here and therefore the test is not accurate. Tried to adapt the before and after each but still. Not against you giving it a try!

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
peterpeterparker: I'm interested in the binded value(s) not the one of the HTML element. It's the binded value that kept the wrong value in memory. In addition, the select binds `selectedAccountIdentifier` and I also want to ensure that the side effect resolve the `selectedAccount` once the new piece of code reset it.

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
peterpeterparker: https://github.com/dfinity/nns-dapp/pull/2194#discussion_r1151435337

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
lmuntaner: Mhm, weird, why would it do that?

Ok, I might give it a try, thanks!

peterpeterparker: > Mhm, weird, why would it do that?

Don't ask me, would be solved 😉

> Ok, I might give it a try, thanks!

Thx

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
lmuntaner: Ah, ok. Let's add a comment then on why we do this instead of the html element.

lmuntaner (approved):
Already approving, thanks!

frontend/src/tests/lib/components/accounts/SelectAccountDropdown.spec.ts
peterpeterparker: Sure, done.

PR #2200, opened 2023-03-29
Rename sns aggregator arg

bitdivine:
# Motivation
The nns-dapp is about to get arguments as well, which makes "args.did" ambiguous.

# Changes
- Rename `arg` to `sns_aggregator_arg`.

# Tests
None


lmuntaner (approved):
LGTM!

PR #2201, opened 2023-03-29
Move the nns-dapp release template

bitdivine:
# Motivation
We now have two projects that both need releases, so "PROPOSAL_TEMPLATE" is ambiguous.

# Changes
* Move the propsal template and update the relative paths to match.

TODO: Update [the release SOP](https://www.notion.so/SOP-Deploy-NNS-dapp-to-mainnet-5347d7438dc04b088c61abd0edb3abd4).

# Tests
I have run the release template script from the new location and verified that the template was populated.

### DIFF
--- a/scripts/nns-dapp/release-template
+++ b/scripts/nns-dapp/release-template
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
-cd "$(dirname "$0")"
+cd "$(dirname "$0")/../.."
 mkdir -p release/ci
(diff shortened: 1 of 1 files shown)


scripts/nns-dapp/release-template
bitdivine, line 1: Note: The equivalent script for the aggregator is at `scripts/sns/aggregator/release-template`

lmuntaner (approved):
LGTM!

@mstrasinskis FYI, we'll need to change the notion docs

PR #2202, opened 2023-03-29
Use auto subscription instead of onDestroy in Login.svelte

dskloetd:
# Motivation

When subscribing to a Svelte store, you need to make sure to unsubscribe at the appropriate time to avoid a memory leak.
When access a store via `$store` in a Svelte component, the subscribe/unsubscribe is automatic and can't be forgotten.

# Changes

1. In Login.svelte, use `$authStore` instead of subscribing/unsubscribing explicitly.

# Tests

Added logging in Login.svelte and auth.store.ts to understand when and how often is subscribed and unsubscribed and manually tested that it's the same with and without this change.

### DIFF
--- a/frontend/src/lib/pages/Login.svelte
+++ b/frontend/src/lib/pages/Login.svelte
@@ -1,6 +1,4 @@
 <script lang="ts">
-  import { onDestroy } from "svelte";
-  import type { Unsubscriber } from "svelte/store";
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
LGTM, thanks

Also tested a prod build locally manually.

PR #2203, opened 2023-03-29
Disable ENABLE_CKBTC on local by default

dskloetd:
# Motivation

It's not yet easy to run ckBTC canisters locally and without them there are lots of errors in the console when ENABLE_CKBTC is enabled.

# Changes

Change dfx.json to disable ENABLE_CKBTC on local network.

# Tests

`__featureFlags.list()` in the console to verify it's disabled by default.


peterpeterparker (approved):
LGTM

PR #2204, opened 2023-03-29
Make tid constants in page objects private and ALL CAPS.

dskloetd:
# Motivation

Minor style cleanup which was requested in https://github.com/dfinity/nns-dapp/pull/2188 and applied only to the page objects in that PR.
Most page object classes have
```
  static readonly tid = "foo";
```

# Changes

Change
```
  static readonly tid = "foo";
```
to
```
  private static readonly TID = "foo";
```

# Tests

npm run test

### DIFF
--- a/frontend/src/tests/page-objects/AccountCard.page-object.ts
+++ b/frontend/src/tests/page-objects/AccountCard.page-object.ts
@@ -2,14 +2,14 @@ import { BasePageObject } from "$tests/page-objects/base.page-object";
 import type { PageObjectElement } from "$tests/types/page-object.types";
 
 export class AccountCardPo extends BasePageObject {
-  static readonly tid = "account-card";
+  private static readonly TID = "account-card";
 
   private constructor(root: PageObjectElement) {
     super(root);
   }
 
   static under(element: PageObjectElement): AccountCardPo {
-    return new AccountCardPo(element.byTestId(AccountCardPo.tid));
+    return new AccountCardPo(element.byTestId(AccountCardPo.TID));
   }
 
   getAccountName(): Promise<string> {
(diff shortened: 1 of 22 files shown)


frontend/src/tests/page-objects/AccountCard.page-object.ts
```diff
@@ -2,14 +2,14 @@ import { BasePageObject } from "$tests/page-objects/base.page-object";
 import type { PageObjectElement } from "$tests/types/page-object.types";
 
 export class AccountCardPo extends BasePageObject {
-  static readonly tid = "account-card";
+  private static readonly TID = "account-card";
```
peterpeterparker, line 5: Would it be interesting to make `TID` a `protected` member of `BasePageObject`?

Something like:

```
export class AccountCardPo extends BasePageObject {
  private constructor(root: PageObjectElement) {
    super({root, TID: "account-card"});
  }
```

Not sure if all `Po` have a `TID`, so just an idea reading the code here in this PR.

peterpeterparker (approved):
LGTM, thanks

Just one comment in case interesting.

frontend/src/tests/page-objects/AccountCard.page-object.ts
dskloetd: Not really. In most cases the TID is only used in 1 place. So it's really just convenience to have it near the top of the file so it's easy to change if you copy a new page object file from an existing one.
But some page objects such as Button or TextInput don't have a TID and are instead identified by their tag.
TID used to be used also in the constructor to make sure the passed in root has the correct TID but that no longer works because getting the attribute is async with Playwright. So instead of the constructor is now private.

peterpeterparker: Gotcha, thanks for the feedback.

PR #2205, opened 2023-03-29
E2E: The user MUST be able to see a list of all their accounts

dskloetd:
# Motivation

Continuing with the accounts end-to-end test, we check the next requirement.

# Changes

1. Check the list of account names in the e2e test.
2. Required page object changes.
3. (not really related but somehow started causing trouble now) While signing in, make sure that we click the correct "Next" button otherwise we might click the same one a second time and get stuck on the next one.

# Tests

`npm run test-e2e`

### DIFF
--- a/frontend/src/tests/e2e/accounts.spec.ts
+++ b/frontend/src/tests/e2e/accounts.spec.ts
@@ -23,9 +23,11 @@ test("Test accounts requirements", async ({ page, context }) => {
   const subAccountName = "My second account";
   await nnsAccountsPo.addAccount(subAccountName);
 
-  // TODO:
+  // AU001: The user MUST be able to see a list of all their accounts
+  const accountNames = await nnsAccountsPo.getAccountNames();
+  expect(accountNames).toEqual([mainAccountName, subAccountName]);
 
-  // AU001:  The user MUST be able to see a list of all their accounts
+  // TODO:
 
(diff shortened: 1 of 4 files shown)


frontend/src/tests/e2e/accounts.spec.ts
```diff
@@ -23,9 +23,11 @@ test("Test accounts requirements", async ({ page, context }) => {
   const subAccountName = "My second account";
   await nnsAccountsPo.addAccount(subAccountName);
 
-  // TODO:
+  // AU001: The user MUST be able to see a list of all their accounts
+  const accountNames = await nnsAccountsPo.getAccountNames();
+  expect(accountNames).toEqual([mainAccountName, subAccountName]);
 
-  // AU001:  The user MUST be able to see a list of all their accounts
+  // TODO:
```
lmuntaner, line 30: Why this `TODO`?

lmuntaner (approved):
Thanks!

frontend/src/tests/e2e/accounts.spec.ts
dskloetd: AU004 and AU005 below still must be done.
This TODO was on line 26 before but AU001 is now done so I moved the TODO down a bit more.

lmuntaner: ok, understood, thanks!

PR #2206, opened 2023-03-29
Utility to test nns-dapp arguments

bitdivine:
# Motivation
Invalid arguments can cause deployments to fail.  So let's test them pre-deployment.

Alternatives considered:
- Parse using didc.  This would be nice as didc already exists but I was unable to get it to do what is needed.

# Changes
- Create a very minimal command line utility for parsing arguments with Rust.

# Tests
The (as yet unmerged) nns-dapp deployment scripts have code to create the argument.  I have verified that valid inputs are parsed correctly and invalid inputs cause an error.  The error message is just the standard candid parsing error.  Ee2 tests of the deployment will use this binary.


lmuntaner (approved):
Thanks!

bitdivine:
Thank you for the review.

PR #2207, opened 2023-03-29
Chore: Add more dummy SNS proposals

lmuntaner:
# Motivation

Testnet users should create more types of dummy SNS proposals for better testing.

# Changes

* Add more dummy proposals in sns-dummy.api.
* Use allSettled and log individual fails for better debugging in testnet.

# Tests

* Not applicable. Functionality only for testnet.

### DIFF
--- a/frontend/src/lib/api/dev.api.ts
+++ b/frontend/src/lib/api/dev.api.ts
@@ -148,16 +148,20 @@ export const makeSnsDummyProposals = async ({
     agent,
     canisterId: canisterIds.governanceCanisterId,
   });
-
   const { snsProposals } = await import("./sns-dummy.api");
 
   await Promise.all(
-    snsProposals.map((proposal) =>
-      canister.manageNeuron({
-        subaccount: neuronId.id,
-        command: [{ MakeProposal: proposal }],
-      })
-    )
+    snsProposals.map((proposal) => {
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd @mstrasinskis @peterpeterparker please review

frontend/src/lib/api/dev.api.ts
```diff
@@ -148,17 +148,29 @@ export const makeSnsDummyProposals = async ({
...
+            proposal.title
+          );
+          console.error(error);
+          throw error;
+        })
     )
   );
 
+  if (allCalls.some((call) => call.status === "rejected")) {
+    throw new Error();
```
dskloetd, line 172: Would it be useful to include the errors from the failed calls in the error? Or are they visible in some other way?

dskloetd (approved):
I'm OK with not testing this, but it's not true that testing is not applicable.
If the functionality is useful, it would be nice for it to not break.
And things that aren't tested tend to break eventually.

frontend/src/lib/api/dev.api.ts
lmuntaner: I console log them before.

And the service catching the error already shows an error saying that "something went wrong, please check the console for details"

PR #2208, opened 2023-03-29
E2E: Get ICP

dskloetd:
# Motivation

We want to test that it's possible to make transactions. For that we need some tokens.
So the "Get ICP" button should be usable from the end-to-end test.

# Changes

1. Add getting 20 ICP to the accounts e2e test.
2. Add required page objects.
3. Add test IDs to components.
4. Add `.first()` to the locator in in `querySelector()` in PlaywrightPageObjectElement to make the behavior similar to actual querySelector and to get the first backdrop without error.

# Tests

`npm run test-e2e -- --headed`


lmuntaner (approved):
LGTM, thanks!

PR #2210, opened 2023-03-29
chore: Update bash argument parser

bitdivine:
# Motivation
The upstream bash argument parser fixed an issue whereby a missing `print_help()` could cause an error.  The fix was implemented in https://github.com/dfinity/snsdemo/pull/99 and tested in https://github.com/dfinity/snsdemo/pull/102  Pulling in the latest code will fix that issue in our copy of the argument parser.

# Changes
* Bump the bash argument parser to the latest version in snsdemo.

# Tests
* The parser is tested upstream
* There are no breaking changes in the API
* CI should verify that there are no unintended breaking changes.


peterpeterparker (approved):
Thx

PR #2214, opened 2023-03-30
fix: settings title and visual quirks

peterpeterparker:
# Motivation

This PR fixes two things:

1. The "Settings" title should be set in the header because if you navigate from a page that sets a title (not the Accounts page) the title in the header remains the one of the previous page

2. When navigating there is a brief moment where both previous and new pages are in the DOM. The layout being what the design requires, when navigating from a page that display universe to a page with island, there could be a brief moment where the title store is updated and therefore the title might be displayed twice.

# Notes

Regarding 2. I did not find better solution for now. It does not solve the issue but, at least workaround to not display twice the title.

Note against better solution at all!

# Changes

- display "Settings" title in header
- add a CSS rule to hide one of the title if two spit panes are rendered in the DOM

Not related:

- add style `value` to display principal on settings page

# Screenshots of the issues

Capture d’écran 2023-03-30 à 20 26 33

demo

# Screenshots with fix/workaround

fix

### DIFF
--- a/frontend/src/lib/themes/global.scss
+++ b/frontend/src/lib/themes/global.scss
@@ -10,3 +10,12 @@ main.sign-in {
     margin-bottom: var(--padding-3x);
   }
 }
+
+// Workaround to not display twice the title of twi pages while navigating between those.
+// Not perfect, as navigating from start/end page to an island page will have for effect to shift the title in the toolbar.
+// But still better than displaying twice a title.
+div.split-pane + div.split-pane {
+  .toolbar {
+    display: none;
+  }
+}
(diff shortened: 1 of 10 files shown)


frontend/src/lib/themes/global.scss
```diff
@@ -10,3 +10,12 @@ main.sign-in {
     margin-bottom: var(--padding-3x);
   }
 }
+
+// Workaround to not display twice the title of twi pages while navigating between those.
```
mstrasinskis, line 14: a typo in `twi`

mstrasinskis (commented):
Every component that calls `layoutTitleStore.set` could also clean the title on destroy `onDestroy(() => layoutTitleStore.set(""))`. But really not sure if it's a better solution 😅

mstrasinskis (approved):
LGTM, thanks!

PR #2215, opened 2023-03-31
Enable canister arguments

bitdivine:
# Motivation
The canister now takes optional initial arguments.  These should be described in the did file.

Doing so also makes the API changes "live", as `dfx deploy` and similar commands ignore arguments if they are not in the API specification. So,  finally, we can start adding e2e tests.

# Changes
- Add the canister arguments to the did file.
- Provide the install arguments wherever we deploy the nns-dapp in tests.
  Note: I believe that the argument will be ignored by the prod canister, which does not take an argument.  The rollback tests van verify this.

# Tests
* There is a new test that verifies that index.html is populated with arguments.  The nns-dapp is deployed directly for this test.
* I have installed the canister by proposal in a manual test as follows; this can and should be added to CI but that requires some slightly bigger changes that I would rather put in a separate PR.
```
#!/usr/bin/env bash
set -euxo pipefail

dfx-network-stop
: Start the server with pre-packaged state
dfx-state-install --state "$(dfx cache show)/state-002.zip"

: Build config and wasms
export DFX_NETWORK=local
./scripts/docker-build
./config.sh

: Make a proposal with a majority neuron
: "... the canister needs to be controlled by root, as in prod."
dfx canister info nns-dapp
ROOT_CANISTER_ID=r7inp-6aaaa-aaaaa-aaabq-cai
dfx canister --network local update-settings nns-dapp --add-controller "$ROOT_CANISTER_ID"
dfx canister info nns-dapp

rm -fr release
mkdir -p release/ci/
cp nns-dapp.wasm release/ci/nns-dapp.wasm
./scripts/nns-dapp/release-template
./scripts/nns-dapp/release --network local

: Verify that the canister has been upgraded
dfx canister info nns-dapp
sha256sum nns-dapp.wasm
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,12 +17,6 @@ jobs:
     runs-on: ubuntu-20.04
     timeout-minutes: 60
     steps:
-      - name: Install cli htmlparser
-        run: |
-          go install github.com/ericchiang/pup@latest
-          GOPATH=$(go env GOPATH)
-          echo "$GOPATH/bin" >> $GITHUB_PATH
-          ls -l "$GOPATH/bin"
       - name: Checkout nns-dapp
         uses: actions/checkout@v3
       - name: Set up docker buildx
@@ -44,9 +38,10 @@ jobs:
           path: 'snsdemo'
           # Version from Mar 27 2023 with saved state tools
           ref: '480a86f555cc9fbee246ba0ecf8a20a4ff705dbc'
-      - name: Add SNS scripts to the path
+      - name: Add go and SNS scripts to the path
         run: |
           echo "$PWD/snsdemo/bin" >> $GITHUB_PATH
+          echo "$(go env GOPATH)/bin" >> $GITHUB_PATH
       - name: Install SNS script dependencies
         run: |
           dfx-software-dfx-install --version "$(jq -r .dfx dfx.json)"
@@ -77,7 +72,7 @@ jobs:
             # Verify that the key is non-trivial:
             # jq -e returns an error code if the value is missing
             # grep ... fails if the value is implausibly short.
-            key=key jq -re '.[0][env.key]' nns_dapp_args_in_page.json | grep -E ...
+            key="$key" jq -re '.[0][env.key]' nns_dapp_args_in_page.json | grep -E ...
           done
       - name: Basic downgrade-upgrade test
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,34 +74,34 @@ jobs:
             # grep ... fails if the value is implausibly short.
             key="$key" jq -re '.[0][env.key]' nns_dapp_args_in_page.json | grep -E ...
           done
-#       - name: Basic downgrade-upgrade test
-#         run: |
-#           git fetch --depth 1 origin tag prod
-#           if git diff tags/prod rs/backend | grep -q .
-#           then ./scripts/nns-dapp/downgrade-upgrade-test -w nns-dapp.wasm
-#           else echo "Skipping test as there are no relevant code changes"
-#           fi
-#       - name: Rename nns-dapp.wasm for upload
-#         run: cp nns-dapp.wasm nns-dapp_local.wasm
-#       - name: 'Upload nns-dapp_local wasm module'
-#         uses: actions/upload-artifact@v3
-#         with:
-#           name: nns-dapp_local
-#           path: nns-dapp_local.wasm
-#           retention-days: 3
-#       - name: Release
-#         run: |
-#           for tag in $(git tag --points-at HEAD) ; do
-#             : Creates or updates a release for the tag
-#             if gh release view "$tag"
-#             then gh release upload --repo dfinity/nns-dapp --clobber "$tag" nns-dapp_local.wasm || true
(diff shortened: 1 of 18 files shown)


config.sh
```diff
@@ -183,9 +183,9 @@ echo "$json" >"$JSON_OUT"
 
 # Creates the candid arguments passed when the canister is installed.
 cat <<EOF >"$CANDID_ARGS_FILE"
-(record{
+(opt record{
   args = vec {
-$(jq -r 'to_entries | .[] | "    record{ 0=\(.key | tojson); 1=\(.value | tostring | gsub("\"";"&quot;") | tojson) };"' "$JSON_OUT")
+$(jq -r 'to_entries | .[] | "    record{ 0=\(.key | tojson); 1=\(.value | tostring | tojson) };"' "$JSON_OUT")
```
bitdivine, line 188: Two corrections:
* The argument is optional
* Quote escaping is now upstream

rs/backend/nns-dapp.did
```diff
@@ -229,7 +229,13 @@ type HttpResponse =
         body: blob;
     };
```
bitdivine, line 231: Add the init arg to the API specification.  This is important for accuracy and also because the argument simply won't be passed otherwise.

rs/backend/src/main.rs
```diff
@@ -35,6 +35,7 @@ type Cycles = u128;
```
bitdivine, line 35: Logging - is useful!

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -37,7 +37,7 @@ get_prod_wasm() {
 
 # Installs the current build of nns-dapp
 upgrade_nnsdapp() {
-  dfx canister install --upgrade-unchanged nns-dapp --wasm "$1" --mode upgrade
+  dfx canister install --upgrade-unchanged nns-dapp --wasm "$1" --mode upgrade --argument "$(cat nns-dapp-arg.did)"
```
bitdivine, line 40: The downgrade-upgrade test, and installations in general, now need to pass the argument.

.github/workflows/build.yml
```diff
@@ -53,12 +53,26 @@ jobs:
           dfx-sns-demo-healthcheck
       - name: Install nns-dapp and sns_aggregator
         run: |
-          dfx canister install nns-dapp --wasm nns-dapp.wasm --upgrade-unchanged --mode reinstall --yes
+          echo "Create the nns-dapp install argument:"
+          DFX_NETWORK=local ./config.sh
+          echo "Install:"
+          dfx canister install nns-dapp --wasm nns-dapp.wasm --upgrade-unchanged --mode reinstall --yes --argument "$(cat nns-dapp-arg.did)"
```
bitdivine, line 59: Need to pass the install argument now.

.github/workflows/build.yml
```diff
@@ -38,9 +38,10 @@ jobs:
           path: 'snsdemo'
           # Version from Mar 27 2023 with saved state tools
           ref: '480a86f555cc9fbee246ba0ecf8a20a4ff705dbc'
-      - name: Add SNS scripts to the path
+      - name: Add go and SNS scripts to the path
```
lmuntaner, line 41: Why is `go` needed?

lmuntaner (approved):
You said that you wanted to stop using `./deploy.sh`.

Does that mean you'll do another script? Or that we should use `dfx`?

I guess we can discuss it later today.

.github/workflows/build.yml
bitdivine: To install pup, to parse html to json.  I'm sure the same can be done with node, I'm just familiar with pup and it's a short incantation.  If we assume that the meta is on one line I can also use grep.  That just felt a bit too fragile to me.

bitdivine: The nodejs equivalent one-liner would presumably be something like:
```
node 'console.log(JSON.stringify(require("some-html-parser").parse(fs.readFileSync("/dev/stdin", "utf-8")).querySelect("head meta[name="nns-dapp-vars"]").dataset))'
```
so if you know of a good parser, we can drop go.

bitdivine: Actually, `html-dom-parser` looks ok and I think I have heard nice things about it.

bitdivine: No, at least, it doesn't seem to have a nice query language.  This is as far as I got:
```
curl "http://$(dfx canister id nns-dapp).localhost:8080/" | gunzip > ,html

cat ,html | node -e 'console.log(require("html-dom-parser")(fs.readFileSync("/dev/stdin", "utf-8"))[2])'
<ref *2> Element {
  parent: null,
  prev: <ref *1> Text {
    parent: null,
    prev: ProcessingInstruction {
      parent: null,
      prev: null,
      next: [Circular *1],
      startIndex: null,
      endIndex: null,
      data: '!DOCTYPE html',

```

bitdivine: Still, I'm open to dropping the dependency if we find a decent alternative.

lmuntaner: Thanks for all the explanations and chain of thought. I belive it's a good reason to use it for now.

PR #2216, opened 2023-03-31
Fix argument autoconfig

bitdivine:
# Motivation
The argument template has two errors, due to erroneous copying from the testing branch to the small slice merged into main:
* The argument is optional, but `opt` is missing from the code in main.
* Quotes are escaped in the rust code, so the escapes are no longer needed in the template.


lmuntaner (approved):
Thanks!

PR #2219, opened 2023-03-31
Feat: Add new field to NNS payload to show arguments as hex string

lmuntaner:
# Motivation

Users want to see the arguments in the NNS Proposal payload also as hex string.

This PR adds a new field `arg_hex` to both payload types where we also decode the args.

# Changes

* Add `arg_hex` to `AddNnsCanisterProposalTrimmed` and `ChangeNnsCanisterProposalTrimmed`.

# Tests

No changes in tests.


mstrasinskis (approved):
LGTM, thank you!

PR #2221, opened 2023-03-31
feat: format-modified command

mstrasinskis:
# Motivation

To improve webstorm user experience the `format-modified` command was added. Formats only updated files (！ignores created).

# Changes

- `format-modified` command

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -15,6 +15,7 @@
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
     "lint": "pprettier --check './**/*.{ts,js,mjs,json,scss,svelte,html}' && eslint --max-warnings 0 .",
     "format": "pprettier --write './**/*.{ts,js,mjs,json,scss,svelte,html}'",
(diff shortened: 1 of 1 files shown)


frontend/package.json
```diff
@@ -15,6 +15,7 @@
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
     "lint": "pprettier --check './**/*.{ts,js,mjs,json,scss,svelte,html}' && eslint --max-warnings 0 .",
     "format": "pprettier --write './**/*.{ts,js,mjs,json,scss,svelte,html}'",
+    "format-modified": "prettier --write $(git diff --diff-filter=ACM --name-only | egrep \"\\.(ts|js|mjs|json|scss|svelte|html)$\" | cut -d'/' -f2-)",
```
peterpeterparker, line 18: `format:modified` to stick to the `:` we use in other commands?

peterpeterparker (approved):
Webstorm 🥲. Thanks

For the record, that's the Webstorm issue that bugs us since 6+ months: https://youtrack.jetbrains.com/issue/WEB-57644

frontend/package.json
mstrasinskis: Good point, thanks!

PR #2224, opened 2023-04-01
Fix: Increase Neuron Stake from SNS Neuron Detail deeplink

lmuntaner:
# Motivation

There was a bug in the SNS neuron detail page:

* User arrives to the sns neuron detail with the deep link.
* User clicks the button "Increase Stake".
* The accounts are shown as loading and are never loaded.

This PR: Adds a test to increase the sns neuron stake from the SnsNeuronDetail.svelte page and fixes the problem.

# Changes

Main change:
* Call `syncSnsAccounts` in SnsNeuronDetail.svelte instead of `loadSnsTransactionFee` (which is included in `syncSnsAccounts`.

Minor changes:
* Upgrade next dependencies to use testId in WizardModal.
* Add prop `testId` to `InputWithError`.
* Add prop `testId` to `Input`.
* Add prop `testId` to `TransactionModal`.
* Use new prop `testId` in `AmountInput` and `SnsIncreaseStakeNeuronModal`.

# Tests

* New fake for sns-api. Just for `increaseStakeNeuron`.
* Fake for sns governance api `addNeuronWith` substitutes neuron if neuron with same id already exists.
* Add test-id "amount-input" in AmountInput to use it in TransactionModal PO.
* Add test for increase neuron stake in SnsNeuronDetail.svelte page spec.
* New PO `SnsIncreaseStakeNeuronModalPo`.
* New PO `TransactionModalPo`.
* Implement `waitFor` in `JestPageObjectElement`. For that, I added an optional parameter to the constructor.
* Implement `typeText` in `JestPageObjectElement`.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -749,9 +749,9 @@
       }
     },
     "node_modules/@dfinity/ckbtc": {
-      "version": "0.0.2-next-2023-03-21.1",
-      "resolved": "https://registry.npmjs.org/@dfinity/ckbtc/-/ckbtc-0.0.2-next-2023-03-21.1.tgz",
-      "integrity": "sha512-VObk8TWyMhjQvjkWDZJUp7WPlvIk74l15hUIGCuBtKyYRySVLVPMZE8NPAKJV5Ba5Z5bin6NisqgEMKG+5YBLg==",
+      "version": "0.0.2-next-2023-03-31",
+      "resolved": "https://registry.npmjs.org/@dfinity/ckbtc/-/ckbtc-0.0.2-next-2023-03-31.tgz",
+      "integrity": "sha512-qUG0u0dUCZpi52n1m8CIKSN0wsw0JNW0H3nnl5EgBkwHnZi1qzirPV2B8zxO6IYFD/tv/ELjpX58HEzp/Y90fA==",
       "dependencies": {
         "base58-js": "^1.0.5",
         "bech32": "^2.0.0",
@@ -765,9 +765,9 @@
       }
     },
     "node_modules/@dfinity/cmc": {
-      "version": "0.0.9-next-2023-03-21.1",
-      "resolved": "https://registry.npmjs.org/@dfinity/cmc/-/cmc-0.0.9-next-2023-03-21.1.tgz",
-      "integrity": "sha512-LkhKpNY4BdvVv9elNKxmOSfFlI/CZffyTww/bvj8O7jS4dbY98JcDxnne9F6RoKqNTYaBzl/miblAYsB/kC0Bg==",
+      "version": "0.0.9-next-2023-03-31",
+      "resolved": "https://registry.npmjs.org/@dfinity/cmc/-/cmc-0.0.9-next-2023-03-31.tgz",
+      "integrity": "sha512-tETQyCCzpnkt+KXXkSHFpwWBSJZ8JVwUwVtWQz+kxKLdHxTbDJGfbVf2PMTax2tW0X1C06yaKK7L9NPBnRLC2A==",
       "peerDependencies": {
         "@dfinity/agent": "*",
         "@dfinity/candid": "*",
@@ -804,9 +804,9 @@
       }
     },
     "node_modules/@dfinity/ledger": {
-      "version": "0.0.6-next-2023-03-21.1",
-      "resolved": "https://registry.npmjs.org/@dfinity/ledger/-/ledger-0.0.6-next-2023-03-21.1.tgz",
-      "integrity": "sha512-8AD4HJW05wQWOTI4nNXIBIQeoZmr32wApdEG2X0HDknzIaqDeWRib6/FLx175DLlnEiwjarSJMvBi1ynCiDZyQ==",
+      "version": "0.0.6-next-2023-03-31",
+      "resolved": "https://registry.npmjs.org/@dfinity/ledger/-/ledger-0.0.6-next-2023-03-31.tgz",
+      "integrity": "sha512-560rqp+AVtBBlKpxk4hO8R6L9j11+F6g2gc3eC+DAtyVz3EClVTVSx0VOxnfWVJJozRLPslwRoG6lu5j/k1UwQ==",
       "peerDependencies": {
         "@dfinity/agent": "*",
         "@dfinity/candid": "*",
@@ -815,12 +815,10 @@
       }
     },
     "node_modules/@dfinity/nns": {
-      "version": "0.15.0-next-2023-03-21.1",
-      "resolved": "https://registry.npmjs.org/@dfinity/nns/-/nns-0.15.0-next-2023-03-21.1.tgz",
-      "integrity": "sha512-BkGHQXAY4JYDWS8U4MKxlxEjwWdK89n52GNBeTkRSThEQxr0VhNGVXkwzSr7BDMX312im+7Y0tUzjkNh8l/Z6w==",
+      "version": "0.15.0-next-2023-03-31",
+      "resolved": "https://registry.npmjs.org/@dfinity/nns/-/nns-0.15.0-next-2023-03-31.tgz",
+      "integrity": "sha512-dv2h+TavYyTMAIa5cXZ54q/JSdqWA0BEeBCEaUmesxmckGGTPLlsb2BYZf+9ApWLLLFdQ/yfCvwC9iaoFi7H3g==",
       "dependencies": {
-        "crc": "^4.3.2",
-        "crc-32": "^1.2.2",
         "google-protobuf": "^3.21.2",
         "js-sha256": "^0.9.0",
         "randombytes": "^2.1.0"
@@ -842,9 +840,9 @@
       }
     },
     "node_modules/@dfinity/sns": {
-      "version": "0.0.13-next-2023-03-21.1",
-      "resolved": "https://registry.npmjs.org/@dfinity/sns/-/sns-0.0.13-next-2023-03-21.1.tgz",
-      "integrity": "sha512-lV5LPFFERCws5HBBZMPngZqEh2qCUdsWq1/fcADfifNY6b2GVqcii80nOpx/0VzcTnyFTW8r2dcxKPgpJs7xEg==",
+      "version": "0.0.13-next-2023-03-31",
+      "resolved": "https://registry.npmjs.org/@dfinity/sns/-/sns-0.0.13-next-2023-03-31.tgz",
+      "integrity": "sha512-rMaUyCrvbMowEUD2S2l2uDcmITO86OkPQRifWNHshGL2n061RaRYOXcQNGvHb03W5UuX/9VKOXyB/uN3H52x/Q==",
       "dependencies": {
         "js-sha256": "^0.9.0"
       },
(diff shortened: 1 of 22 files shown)


lmuntaner:

frontend/package-lock.json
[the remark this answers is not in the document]
lmuntaner: Sure, I'll revert

frontend/src/lib/components/ui/AmountInput.svelte
```diff
@@ -19,6 +19,7 @@
   {max}
   inputType="icp"
   {errorMessage}
+  testId="amount-input"
```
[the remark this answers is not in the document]
lmuntaner: No, it makes sense to me.

I didn't create the AmountInputPo and wasn't sure to use `-component`.

Your change makes sense.

frontend/src/lib/components/ui/Input.svelte
```diff
@@ -14,13 +14,14 @@
...
   export let showInfo = true;
+  export let testId: string | undefined = undefined;
 
   let placeholder: string;
   $: placeholder = translate({ labelKey: placeholderLabelKey });
 </script>
 
 <Input
-  testId="input-ui-element"
+  testId={testId ?? "input-ui-element"}
```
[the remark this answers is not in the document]
lmuntaner: ditto. I'll revert

frontend/src/lib/components/ui/InputWithError.svelte
```diff
@@ -19,6 +19,7 @@
   export let value: string | number | undefined = undefined;
   export let placeholderLabelKey: string;
   export let showInfo = true;
+  export let testId: string | undefined = undefined;
 </script>
 
 <div class="wrapper" class:error>
```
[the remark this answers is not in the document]
lmuntaner: ditto

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -311,15 +311,22 @@ export const addNeuronWith = ({
...
+  const currentNeurons = getNeurons({ identity, rootCanisterId });
+  const index = currentNeurons.length;
   const defaultNeuronId = createNeuronId({ identity, index });
   const neuron: SnsNeuron = {
     ...mockSnsNeuron,
     id: [defaultNeuronId],
     ...neuronParams,
   };
-  neurons.push(neuron);
+  const filteredNeurons = currentNeurons.filter((currentNeuron) => {
```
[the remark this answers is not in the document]
lmuntaner: sure, no strong opinion there.

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -311,15 +311,22 @@ export const addNeuronWith = ({
...
+    const currentHexIt = snsNeuronIdToHexString(currentNeuron.id[0]);
+    const newHexIt = snsNeuronIdToHexString(neuron.id[0]);
```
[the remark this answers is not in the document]
lmuntaner: Yes

frontend/src/tests/page-objects/SnsIncreaseStakeNeuronModal.page-object.ts
```diff
@@ -0,0 +1,24 @@
+import type { PageObjectElement } from "$tests/types/page-object.types";
+import { TransactionModalPo } from "./TransactionModal.page-object";
+
+export class SnsIncreaseStakeNeuronModalPo extends TransactionModalPo {
```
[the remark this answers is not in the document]
lmuntaner: I don't see it as a stand-alone PO.

It's always used inside another `XxxModal`.

frontend/src/tests/fakes/sns-governance-api.fake.ts
```diff
@@ -303,26 +303,71 @@ const createNeuronId = ({
 // This follows the same logic to determine neuron IDs as the real code because
 // the real code will look for neurons with these IDs without even knowing if
 // they exist.
-export const addNeuronWith = ({
-  identity = mockIdentity,
+const addNeuronToList = ({
```
lmuntaner, line 306: New helper used both in `addNeuronWith` and `setNeuronWith`.

lmuntaner:

frontend/src/lib/modals/sns/neurons/SnsTransactionModal.svelte
```diff
@@ -13,6 +13,7 @@
   export let transactionFee: TokenAmount;
   export let currentStep: WizardStep;
   export let validateAmount: ValidateAmountFn = () => undefined;
+  export let testId: string | undefined = undefined;
```
dskloetd, line 16: Is this currently unused?

frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
```diff
@@ -61,6 +68,11 @@ describe("SnsNeuronDetail", () => {
     snsQueryStore.reset();
     snsQueryStore.setData(responses);
 
+    fakeSnsLedgerApi.addAccountWith({
+      identity: mockIdentity,
```
dskloetd, line 72: Doesn't it use mockIdentity by default?

frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
```diff
@@ -122,6 +136,36 @@ describe("SnsNeuronDetail", () => {
...
+      });
+      const po = await renderComponent(props);
+
+      // `neuronStake` to string formatted as expected
+      expect(await po.getStakeCardPo().getStakeAmount()).toBe("1.00");
+      const amountToStake = 20;
+      fakeSnsGovernanceApi.setNeuronWith({
+        rootCanisterId,
+        id: [validNeuronId],
+        cached_neuron_stake_e8s: numberToE8s(neuronStake + amountToStake),
```
dskloetd, line 158: Is this to simulate the increase that should have happened in the API?
If it made sense to move increaseStake from sns-api to sns-governance-api, then the API could do the proper thing.
Otherwise we should think about giving the fake sns-api access to the fake-governance-api so it can correctly increase the stake.

Also, if we can't do that, we need to add a check to the test that sns-api.increasStake has been called, right?

frontend/src/tests/page-objects/TextInput.page-object.ts
```diff
@@ -30,4 +30,9 @@ export class TextInputPo {
   typeText(text: string): Promise<void> {
     return this.root.typeText(text);
   }
+
+  // TextInputPo does not extend `PageObjectElement` therefore, we need to add the methods we need here.
```
dskloetd, line 34: Strange, I thought TextInputPo extends SimpleBasePageObject for that reason.
Can you try extending `SimpleBasePageObject` and removing this?

frontend/src/tests/page-objects/jest.page-object.ts
```diff
@@ -62,10 +93,10 @@ export class JestPageObjectElement implements PageObjectElement {
     await fireEvent.click(this.element);
   }
 
-  async typeText(_text: string): Promise<void> {
-    throw new Error("Not implemented");
-    // Not tested:
-    // fireEvent.change(this.element, {target: {value: text}});
+  async typeText(text: string): Promise<void> {
+    // Svelte generates code for listening to the `input` event
+    // https://github.com/testing-library/svelte-testing-library/issues/29#issuecomment-498055823
```
dskloetd, line 98: Do we need a comment here?
If so, it's not clear to me why.
Is it trying to explain why you didn't use the untested code that was there commented out?
Someone seeing this comment in the future won't know what was there before so it will be even less clear why there needs to be a comment.
Can you either explain what needs to be explained, or remove the comment if it's not necessary?

frontend/src/lib/modals/sns/neurons/SnsIncreaseStakeNeuronModal.svelte
```diff
@@ -108,6 +108,7 @@
     {token}
     {transactionFee}
     {governanceCanisterId}
+    testId="sns-increase-stake-neuron-modal-component"
```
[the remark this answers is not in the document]
dskloetd: I don't know the code well enough to know if it would work.
But if you can remove the if, then you can also remove the wrapper and that would be nice, yes.

frontend/src/lib/modals/sns/neurons/SnsTransactionModal.svelte
lmuntaner: Actually, yes. Now that I use the TestIdWrapper in the SnsIncreaseStakeModal this is not used anymore, thanks

frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
lmuntaner: Yes. I can remove it.

frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
lmuntaner: `increaseStake` needs to be in sns-api because it uses also the ledger canister, not just the governance.

If we start adding this functionality to the fake ones. That means that you want to also add the functionality when we stake a neuron, or increase the stake, or split a neuron.

I don't think this is a good idea. We will be duplicating a lot of code and in reality not testing the actual thing.

I prefer to test that increaseStake has been called. I'll do that.

frontend/src/tests/page-objects/jest.page-object.ts
lmuntaner: I'll explain better, yes.

frontend/src/tests/page-objects/TextInput.page-object.ts
lmuntaner: done, I thought you didn't extend before because there was some circular dependency of some kind.

frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
lmuntaner: done

lmuntaner:

frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
dskloetd: Huh, one of the reasons to use fakes is so they adjust their internal state when it makes sense in order to present a consistent state to the calling code.
In this particular case it might be challenging because they are different APIs, but when possible we *should* really prefer to update the internal state of the fake when a call would cause the state to change.

dskloetd (approved):

frontend/src/tests/page-objects/TextInput.page-object.ts
```diff
@@ -1,13 +1,8 @@
...
+import { SimpleBasePageObject } from "./simple-base.page-object";
 
-export class TextInputPo {
-  readonly root: PageObjectElement;
-
-  private constructor(root: PageObjectElement) {
-    this.root = root;
-  }
-
+export class TextInputPo extends SimpleBasePageObject {
```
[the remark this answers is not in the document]
lmuntaner: It's automatic when we use `extend`

frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
lmuntaner: In that case, we might want to have the neurons in a common place.

Isn't it better to check that the call to the proper api that changes the state was done?

Otherwise, if an endpoint changes, the test will pass and won't catch the change.

PR #2228, opened 2023-04-03
Chore: Remove SNS Aggregator fallback if less than 2

lmuntaner:
# Motivation

Remove the check on the SNS aggregator for the number of SNSs to force a fallback.

This is not necessary anymore because the SNS aggregator persists the data between upgrades.

# Changes

* Remove check on ths SNS aggregator response to trigger the fallback.

# Tests

* Remove test of the fallback

### DIFF
--- a/frontend/src/lib/services/$public/sns.services.ts
+++ b/frontend/src/lib/services/$public/sns.services.ts
@@ -27,10 +27,6 @@ export const loadSnsProjects = async (): Promise<void> => {
   try {
     const cachedSnses = await querySnsProjects();
     const identity = getCurrentIdentity();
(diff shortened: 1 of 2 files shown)


lmuntaner:

dskloetd (approved):
The PR that added the code (https://github.com/dfinity/nns-dapp/commit/f272ea68b87bf60ee9875148816cde8568f0463b) made some more changes. In particular to make sure that all the tests that would otherwise fail have at least 2 SNS projects. Do you want to keep those changes or also clean them up?

lmuntaner:
> 

No, I believe that those tests are ok. They were failing with only 1 project because of the check. But the test is probably more realistic if aggregator returns more than 1. Therefore, I'd rather keep those changes.

PR #2229, opened 2023-04-03
Refactor/votable neuron

mstrasinskis:
# Motivation

To generalise the voting mechanism `VotableNeuron` interface was introduce.

# Changes

- `VotableNeuron` interface
- `votingNeuronSelectStore` refactoring
- refactoring of related component
- util `neuronToVotingNeuron`
- `anonymizeVotingNeuron`

# Tests

TBD...

# Screenshots

## Before

https://user-images.githubusercontent.com/98811342/229456552-47028202-6664-4868-917b-7fc32ff32801.mp4

## After (should be the same)

https://user-images.githubusercontent.com/98811342/229456750-cbea7605-285b-41fb-96f5-0d305027f56e.mp4

### DIFF
--- a/frontend/src/lib/components/proposal-detail/VotingCard/VotingNeuronSelect.svelte
+++ b/frontend/src/lib/components/proposal-detail/VotingCard/VotingNeuronSelect.svelte
@@ -23,7 +23,6 @@
   $: totalNeuronsVotingPower = selectedNeuronsVotingPower({
     neurons: $votingNeuronSelectStore.neurons,
     selectedIds: $votingNeuronSelectStore.selectedIds,
-    proposal: proposalInfo,
   });
 
   let toggleContent: () => void;
--- a/frontend/src/lib/stores/proposals.store.ts
+++ b/frontend/src/lib/stores/proposals.store.ts
@@ -1,5 +1,6 @@
 import { DEFAULT_PROPOSALS_FILTERS } from "$lib/constants/proposals.constants";
 import { storeLocalStorageKey } from "$lib/constants/stores.constants";
+import type { VotingNeuron } from "$lib/types/proposals";
 import {
   concatenateUniqueProposals,
   excludeProposals,
@@ -8,8 +9,6 @@ import {
   replaceProposals,
 } from "$lib/utils/proposals.utils";
 import type {
-  NeuronId,
-  NeuronInfo,
   ProposalId,
   ProposalInfo,
   ProposalRewardStatus,
@@ -33,8 +32,9 @@ export interface ProposalsFiltersStore {
 }
 
 export interface NeuronSelectionStore {
-  neurons: NeuronInfo[];
-  selectedIds: NeuronId[];
+  neurons: VotingNeuron[];
+  // TODO: selectedNeurons OR selectedIdStrings
+  selectedIds: string[];
 }
 
 export interface ProposalsStore {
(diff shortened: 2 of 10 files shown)


frontend/src/lib/utils/proposals.utils.ts
```diff
@@ -598,3 +599,14 @@ export const voteRegistrationActive = (votes: VoteRegistration[]): boolean =>
     ({ neuronIds, successfullyVotedNeuronIds }) =>
       neuronIds.length > successfullyVotedNeuronIds.length
   ) !== undefined;
+
+export const neuronToVotingNeuron = ({
```
lmuntaner, line 603: rename to `nnsNeuronToVotingNeuron`?

frontend/src/lib/stores/proposals.store.ts
```diff
@@ -33,8 +32,9 @@ export interface ProposalsFiltersStore {
 }
 
 export interface NeuronSelectionStore {
```
lmuntaner, line 34: We reset this store when the user leaves the proposal detail page, right?

Question, if the user moves from one proposal to the next one, will this be reset? The voting power for the neurons in one proposal and the next won't be the same.

lmuntaner (commented):
Looking good! Just one question for now.

frontend/src/lib/stores/proposals.store.ts
mstrasinskis: Correct, this store will be reset (ids and voting power recalculation) `votingNeuronSelectStore.set(votableNeurons());`

frontend/src/lib/utils/proposals.utils.ts
mstrasinskis: Good point, done.

frontend/src/lib/stores/proposals.store.ts
lmuntaner: Ok.

I would expand the comment in the `initNeuronSelectStore`. To explain that this needs to be reset for each different proposal and it applies only to a specific proposal.

The other option would be to add the rootCanisterId and proposal id in the store. But I don't think it's necessary right now.

mstrasinskis: Not sure that I've got the issue. As I know there is no store state merge between different proposals.

lmuntaner: Ok

frontend/src/tests/lib/components/proposal-detail/VotingCard/VotingConfirmationToolbar.spec.ts
```diff
@@ -23,17 +24,15 @@ describe("VotingConfirmationToolbar", () => {
...
+    votingNeuronSelectStore.set([
+      nnsNeuronToVotingNeuron({ neuron, proposal: mockProposalInfo }),
+    ]);
   });
 
   it("should disable buttons if nothing is selected", async () => {
-    const { container } = render(VotingConfirmationToolbar, { props });
-    votingNeuronSelectStore.toggleSelection(neuron.neuronId);
+    const { container } = render(VotingConfirmationToolbar);
+    votingNeuronSelectStore.toggleSelection(`${neuron.neuronId}`);
```
lmuntaner, line 35: Ditto, just use a string that it's set below. But let's not use any nns neuron or proposal references in this file

frontend/src/lib/components/proposal-detail/VotingCard/VotingNeuronSelect.svelte
lmuntaner, line 1: Can this also become agnostic of proposal? Or you think we can't do this?

lmuntaner (commented):
A few comments, looking very good!

frontend/src/tests/lib/components/proposal-detail/VotingCard/VotingConfirmationToolbar.spec.ts
```diff
@@ -23,17 +24,15 @@ describe("VotingConfirmationToolbar", () => {
     votingPower,
   };
 
-  const props = {
-    proposalInfo: mockProposalInfo,
-  };
-
   beforeEach(() => {
-    votingNeuronSelectStore.set([neuron]);
+    votingNeuronSelectStore.set([
```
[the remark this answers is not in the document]
mstrasinskis: Good point, done!

lmuntaner (approved):
Thanks!

PR #2230, opened 2023-04-03
GIX-1440: NNS Dapp uses inserted env vars if possible

lmuntaner:
# Motivation

In order for the NNS Dapp wasm to be agnostic, we introduced arguments when installing the wasm that create an element in the html with all the environment configuration.

This PR adds the functionality in the frontend assets to use that configuration first.

I deployed to https://3cswx-rqaaa-aaaaa-aabyq-cai.nnsdapp.testnet.dfinity.network/accounts/?u=3cswx-rqaaa-aaaaa-aabyq-cai a wasm created for another tesnet. I get the calls to the proper canister id, set on installation.

Yet, this is not working because the CSP are not set.

# Changes

* New environment util `getEnvVars` that returns an object with the configuration. This util either reads the env vars from build or html depending on whether is local development or not.
* Use new util in all the environment variables.

# Tests

* Only build changes.


lmuntaner:
@bitdivine please take a look and feel free to try it while you try to add the CSP.

lmuntaner:

frontend/src/lib/utils/env.utils.ts
```diff
@@ -28,3 +29,31 @@ export const addRawToUrl = (urlString: string): string => {
 
   return hasTrailingSlash ? url.toString() : url.toString().slice(0, -1);
 };
+
+type DataSet = {
```
dskloetd, line 33: Can you use a more descriptive name?

frontend/src/lib/constants/canister-ids.constants.ts
```diff
@@ -1,21 +1,30 @@
...
 
-export const OWN_CANISTER_ID_TEXT = import.meta.env
-  .VITE_OWN_CANISTER_ID as string;
+const envVarsDataset = getHtmlEnvVars();
+
+console.log(envVarsDataset);
+
+export const OWN_CANISTER_ID_TEXT =
+  envVarsDataset?.ownCanisterId ??
+  (import.meta.env.VITE_OWN_CANISTER_ID as string);
```
dskloetd, line 10: Are the name consistent enough that we can do the fallback automatically instead of having to refer to 2 different names everywhere?

dskloetd, line 10: I'm not really comfortable with having a fallback for each individual variable.
I think it can make things very confusing if something is missing.
Can we somehow indicate if we intend to use one way or the other and properly fail if things aren't the way we expect?

frontend/src/lib/constants/canister-ids.constants.ts
```diff
@@ -1,21 +1,30 @@
+import { getHtmlEnvVars } from "$lib/utils/env.utils";
 import { Principal } from "@dfinity/principal";
 
-export const OWN_CANISTER_ID_TEXT = import.meta.env
-  .VITE_OWN_CANISTER_ID as string;
+const envVarsDataset = getHtmlEnvVars();
```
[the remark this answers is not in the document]
lmuntaner: I kept a similar naming as the web api: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

frontend/src/lib/constants/canister-ids.constants.ts
lmuntaner: I believe I can, yes. There is an env var when the code runs in local environment, I could use that to select the VITE instead of html ones.

frontend/src/lib/utils/env.utils.ts
lmuntaner: sure, how about `EnvironmentDataSet`?

frontend/src/lib/constants/canister-ids.constants.ts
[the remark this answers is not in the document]
dskloetd: Thanks for the link. I wasn't familiar with that.
So `dataset` is a property of a DOM element.
But once you put that data in some other format, "dataset" no longer has anything to do with it, right?
Is there any reason not to just call it `envVars`?

frontend/src/lib/utils/env.utils.ts
dskloetd: Dataset refers to the specific api to get the data from the HTML, right?
Once it's in a different format, does it still make sense to have "dataset" in the name?

lmuntaner: done

lmuntaner:

frontend/src/lib/constants/ckbtc-canister-ids.constants.ts
```diff
@@ -11,11 +14,8 @@ export const CKBTC_MINTER_CANISTER_ID = Principal.fromText(
 const MAINNET_CKBTC_LEDGER_CANISTER_ID = "mxzaz-hqaaa-aaaar-qaada-cai";
 const MAINNET_CKBTC_INDEX_CANISTER_ID = "n5wcd-faaaa-aaaar-qaaea-cai";
 
-const ENV_CKBTC_LEDGER_CANISTER_ID = import.meta.env
-  .VITE_CKBTC_LEDGER_CANISTER_ID;
-const ENV_CKBTC_INDEX_CANISTER_ID = import.meta.env
-  .VITE_CKBTC_INDEX_CANISTER_ID;
-
+const ENV_CKBTC_LEDGER_CANISTER_ID = envVarsDataset.ckbtcLedgerCanisterId ?? "";
+const ENV_CKBTC_INDEX_CANISTER_ID = envVarsDataset.ckbtcIndexCanisterId ?? "";
```
dskloetd, line 18: Why did we previously not need `?? ""` but now we do?

frontend/src/lib/constants/canister-ids.constants.ts
```diff
@@ -1,21 +1,22 @@
+import { getEnvVars } from "$lib/utils/env-vars.utils";
 import { Principal } from "@dfinity/principal";
 
-export const OWN_CANISTER_ID_TEXT = import.meta.env
-  .VITE_OWN_CANISTER_ID as string;
+const envVarsDataset = getEnvVars();
```
dskloetd, line 4: Remove "Dataset" from the name?

frontend/jest-setup.ts
```diff
@@ -22,40 +21,36 @@ global.TextEncoder = TextEncoder;
...
+      TEST_FLAG_EDITABLE: true,
+      TEST_FLAG_NOT_EDITABLE: true,
+    }),
+    fetchRootKey: "false",
+    host: "https://icp-api.io",
+    governanceCaniserId: "rrkah-fqaaa-aaaaa-aaaaq-cai",
+    identityServiceUrl: "http://localhost:8000/",
+    ledgerCanisterId: "ryjl3-tyaaa-aaaaa-aaaba-cai",
+    ownCanisterId: "qhbym-qaaaa-aaaaa-aaafq-cai",
+    ownCanisterUrl: "nns.ic0.app",
```
dskloetd, line 45: This doesn't seem to be a key of EnvironmentVars? Is there a way to have this object type checked?

dskloetd (commented):
I like this design!
Some more low level questions remaining.

frontend/src/lib/utils/env-vars.utils.ts
```diff
@@ -0,0 +1,97 @@
...
+    ) as string,
+    fetchRootKey: import.meta.env.VITE_FETCH_ROOT_KEY as string,
+    host: import.meta.env.VITE_HOST as string,
+    governanceCaniserId: import.meta.env.VITE_GOVERNANCE_CANISTER_ID as string,
+    identityServiceUrl: import.meta.env.VITE_IDENTITY_SERVICE_URL as string,
+    ledgerCanisterId: import.meta.env.VITE_LEDGER_CANISTER_ID as string,
+    ownCanisterId: import.meta.env.VITE_OWN_CANISTER_ID as string,
+    // Environments without SNS aggregator are valid
+    snsAggregatorUrl: import.meta.env.VITE_AGGREGATOR_CANISTER_URL as string,
+    wasmCanisterId: import.meta.env.VITE_WASM_CANISTER_ID,
```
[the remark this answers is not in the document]
lmuntaner: Those are to specify that the field is present. Although, since I already do the check afterwards, this might not be needed.

Let me try to remove it.

frontend/src/lib/utils/env-vars.utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  governanceCaniserId: string;
+  identityServiceUrl: string;
+  ledgerCanisterId: string;
+  ownCanisterId: string;
+  // Environments without SNS aggregator are valid
+  snsAggregatorUrl?: string;
+  wasmCanisterId: string;
+};
+
+const defaultMandatoryEnvVars: EnvironmentVars = {
```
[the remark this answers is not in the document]
lmuntaner: I started with the idea to set it as a default, and then that didn't make sense and I didn't rename it.

I'll change it.

frontend/src/lib/constants/ckbtc-canister-ids.constants.ts
lmuntaner: Before the variable was at least an empty string.

Now that we read from the HTML, the value of the dataSet might be `undefined`.

frontend/src/lib/constants/ckbtc-canister-ids.constants.ts
```diff
@@ -1,6 +1,9 @@
+import { getEnvVars } from "$lib/utils/env-vars.utils";
 import { Principal } from "@dfinity/principal";
 import { notEmptyString } from "@dfinity/utils";
 
+const envVarsDataset = getEnvVars();
```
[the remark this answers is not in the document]
lmuntaner: yes, thanks

frontend/src/lib/constants/canister-ids.constants.ts
lmuntaner: yes, thanks

frontend/jest-setup.ts
lmuntaner: Thanks, I'll remove it.

Maybe because TypeScript is using structural typing instead of nominal typing?

frontend/src/lib/utils/env-vars.utils.ts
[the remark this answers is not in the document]
lmuntaner: I was able to remove the casting

frontend/src/lib/constants/ckbtc-canister-ids.constants.ts
dskloetd: Is it now always undefined when it's not set? Or does it depend on whether we get the vars from the environment or from the HTML?
Can we make sure the API is always the same whether it's from the environment or from the HTML?
So either always `undefined` when not set or always `""` when no set?
Or is it already like that?

frontend/jest-setup.ts
dskloetd: Right, I guess adding extra fields doesn't stop it from being the type is has to be. It can only be more, not less.

frontend/src/lib/constants/ckbtc-canister-ids.constants.ts
lmuntaner: Sure, I can convert the empty strings to undefined.

lmuntaner:

frontend/src/lib/utils/env-vars.utils.ts
```diff
@@ -0,0 +1,124 @@
...
+      import.meta.env.VITE_WASM_CANISTER_ID
+    ),
+  };
+  assertEnvVars(envVars);
+  return envVars;
+};
+
+export const getEnvVars = (): EnvironmentVars => {
+  if (isBrowser && !localDev) {
+    try {
```
dskloetd, line 115: How about just checking if the element is there or returning `undefined` from `getHtmlEnvVars` if the element is not there?
Or is using errors for things that are expected a standard pattern we use in the code?
To me it doesn't look very nice when it's not necessary.

dskloetd (approved):

frontend/src/lib/utils/env-vars.utils.ts
lmuntaner: I wanted the same pattern for the element not being there and the env vars not being complete.

I "assert" that the env vars are complete and throw an error otherwise.

Also, it's nice that the return value of`getHtmlEnvVars` is `EnvironmentVars`. And not `EnvironmentVars` | `undefined`.

lmuntaner: I agree.

I'm just scared that the production release doesn't work properly with the HTML env vars, and I don't want to mess it up.

Falling back to the build env vars makes sure that the release to prod will work.

dskloetd: Then just always use the build vars until you know the html vars work.
This defensive programming is confusing and will shoot you in the foot.

PR #2231, opened 2023-04-03
Chore: Remove SNS proposal action section

lmuntaner:
# Motivation

As agreed, the NNS Dapp will rely on the `payload_text_rendering` field to render the details of the action in an SNS proposal.

This PR removes the card with the action details.

# Changes

* Don't render `SnsProposalActionSection` in the SnsProposalDetail page.
* Remove SnsProposalActionSection.svelte

# Tests

* Remove test for `SnsProposalActionSection`


lmuntaner:

frontend/src/lib/pages/SnsProposalDetail.svelte
```diff
@@ -100,7 +99,6 @@
     </div>
     <div class="content-c proposal-data-section">
       <SnsProposalSummarySection {proposal} />
-      <SnsProposalActionSection {proposal} />
       <SnsProposalPayloadSection {proposal} />
```
dskloetd, line 102: For a future PR: Should this still be called "Payload"?

lmuntaner: Yes, I believe so. Even the field is called `payload_text_rendering`.

But it's a good question. I'll mention it with product whether we should call this "action" instead.

Maybe we should introduce the "Action" concept in the Proposal page.

PR #2232, opened 2023-04-03
E2E accounts: Check that the amount was transferred

dskloetd:
# Motivation

When transferring an amount in the e2e, we want to check that the transfer was successful

# Changes

1. Check the account balances before and after transfer
2. Required supporting changes in page objects.

## Minor unrelated changes

1. Make AccountsPo and NnsAccountsFooterPo constructors private.
2. Never return null from AccountsPo.under()
3. Output available names when an account card can't be found.
4. Add utility function to sign in with an existing anchor.

# Tests

npm run test-e2e


frontend/src/tests/page-objects/NnsAccountsFooter.page-object.ts
```diff
@@ -5,7 +5,7 @@ import type { PageObjectElement } from "$tests/types/page-object.types";
 export class NnsAccountsFooterPo extends BasePageObject {
   private static readonly TID = "nns-accounts-footer-component";
 
-  constructor(root: PageObjectElement) {
+  private constructor(root: PageObjectElement) {
```
lmuntaner, line 8: ditto on removing it

lmuntaner (approved):
LGTM, just a few comments

frontend/src/tests/e2e/accounts.spec.ts
```diff
@@ -41,6 +46,14 @@ test("Test accounts requirements", async ({ page, context }) => {
     amount: 5,
   });
 
+  await appPo.goBack();
+  await nnsAccountsPo.getMainAccountCardPo().getAmountDisplayPo().waitFor();
+  //await new Promise((resolve) => setTimeout(resolve, 1 * 1000));
```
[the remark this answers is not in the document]
dskloetd: Oops. Removed

frontend/src/tests/page-objects/NnsAccountsFooter.page-object.ts
dskloetd: Reverted.

PR #2236, opened 2023-04-03
Docker args

bitdivine:
# Motivation

# Changes

# Tests


Dockerfile
```diff
@@ -77,6 +80,8 @@ WORKDIR /build
 ARG DFX_NETWORK=mainnet
 RUN mkdir -p frontend
 RUN ./config.sh
+RUN apt -yqq install --no-install-recommends xxd
```
bitdivine, line 83: The apt-install can be moved up to the other apt-installs at the beginning of the dockerfile, however that busts the docker cache, so should be done as a small self-contained PR as otherwise iterations will be super slow.

Dockerfile
```diff
@@ -77,6 +80,8 @@ WORKDIR /build
 ARG DFX_NETWORK=mainnet
 RUN mkdir -p frontend
 RUN ./config.sh
```
bitdivine, line 82: The apt-install can be moved up to the other apt-installs at the beginning of the dockerfile, however that busts the docker cache, so should be done as a small self-contained PR as otherwise iterations will be super slow.

mstrasinskis (approved):
Rubber stamp. Thanks!

PR #2237, opened 2023-04-03
Docker build

bitdivine:
# Motivation
Making docker builds the default makes deployment easier, but I'd like to see whether there are any scripts or flows that break if we make the switch before building too much on this.

# Changes
* Make docker build the default in dfx.json for both the aggregator and nns-dapp.

# Tests
See CI


dskloetd:
Fine with me but maybe more than 1 person should chime in first?

mstrasinskis (approved):
Rubber stamp.

dskloetd (approved):

PR #2238, opened 2023-04-03
Delete assets before build

bitdivine:
# Motivation
If docker-build fails and the user fails to notice, the user may believe that some old artifacts are the output and deploy or validate the wrong files.  Let's remove that risk.

# Changes
* Delete artifacts before building

# Tests
See CI


mstrasinskis (approved):
Rubber stump.

PR #2243, opened 2023-04-04
Cleanup page objects

dskloetd:
# Motivation

1. A constructor that only calls `super` with the same arguments is provided implicitly by default so no need to add one.
2. Relative imports don't work in Playwright tests. (I'm not sure why.)

# Changes

1. Remove unnecessary default constructors from page objects.
2. Replace relative imports with absolute imports in page objects.

# Tests

```
npm run test
npm run test-e2e
```


mstrasinskis (approved):
LGTM, thanks!

PR #2244, opened 2023-04-04
Split installing snapshot state from running frontend.

dskloetd:
# Motivation

There might be other use cases for using a dfx snapshot then just running frontend against it.

# Changes

1. Create `scrtips/dfx-snapshot-install` to take some of the functionality from `scripts/dev-local-state.sh`.
2. `scrtips/dfx-snapshot-install` does the following:
  * Create backup directory
  * Move existing state to the backup directory
  * Generate the restore script
3. `scripts/dev-local-state.sh` still does:
  * Check if an existing replica/script is running
  * Set a trap to restore state
  * Run the replica
  * Run the frontend

# Tests

Manually ran the scripts.


bitdivine (approved):
Thank you!

PR #2246, opened 2023-04-04
Fix build flakiness

bitdivine:
# Motivation
Sometimes docker builds fail with:
```
#52 0.714    Compiling nns-dapp v2.0.26 (/build/rs/backend)
#52 0.790 error[E0432]: unresolved import `nns_dapp::arguments`
#52 0.790  --> rs/backend/src/bin/nns-dapp-check-args.rs:3:15
#52 0.790   |
#52 0.790 3 | use nns_dapp::arguments::CanisterArguments;
#52 0.790   |               ^^^^^^^^^ could not find `arguments` in `nns_dapp`
#52 0.790 
#52 0.802 For more information about this error, try `rustc --explain E0432`.
```

The code is correct and builds succeed outside docker.

Docker differs from local builds in that it pre-compiles libraries to cache them.  If we do not pre-compile, builds work again but are slow.

The code that pre-compiles the libraries replaces the actual code with aplaceholder, but the placeholder does not include files for all the libraries and main files in the actual code.  If we put in placeholders for all lib and main files, the code works.

# Changes
* Fix the pre-warming placeholder code.

# TODO
I believe that the pre-warming code is a bit hacky and would be best spun out into a dedicated file with tests.

# Tests
Local docker builds seem to work reliably now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,7 @@ COPY Cargo.lock .
 COPY Cargo.toml .
 COPY rs/backend/Cargo.toml rs/backend/Cargo.toml
 COPY rs/sns_aggregator/Cargo.toml rs/sns_aggregator/Cargo.toml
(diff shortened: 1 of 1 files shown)


mstrasinskis (approved):
Rubber stamp. Thank you!

PR #2249, opened 2023-04-04
Install a pre-built optimizer

bitdivine:
# Motivation
Docker builds on M1 macs are hard.  One sticking point is building the optimizer, which seems to consume inordinate resources.  Given that the optimizer is now available as a pre-built download, let's use it.

# Changes
- In the dockerfile, download the optimizer instead of building it.
- Put the optimizer version in dfx, along with other versions.
- Bump the optimizer to the latest available version.

# Tests
* See CI
* And let's see whether this gets M1 mac builds further

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ ENV NODE_VERSION=16.17.1
 RUN jq -r .dfx dfx.json > config/dfx_version
 RUN jq -r '.defaults.build.config.NODE_VERSION' dfx.json > config/node_version
 RUN jq -r '.defaults.build.config.DIDC_VERSION' dfx.json > config/didc_version
-RUN printf "%s" "0.3.1" > config/optimizer_version
+RUN jq -r '.defaults.build.config.OPTIMIZER_VERSION' dfx.json > config/optimizer_version
 
 # This is the "builder", i.e. the base image used later to build the final code.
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Rubber stamp. Thanks!

Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN curl --fail https://sh.rustup.rs -sSf \
 ENV PATH=/cargo/bin:$PATH
 RUN cargo --version
 # Install IC CDK optimizer
-RUN cargo install --version "$(cat config/optimizer_version)" ic-cdk-optimizer
+RUN curl -L --fail --retry 5 "https://github.com/dfinity/cdk-rs/releases/download/$(cat config/optimizer_version)/ic-cdk-optimizer-$(cat config/optimizer_version)-ubuntu-20.04.tar.gz" | gunzip | tar -x "ic-cdk-optimizer-$(cat config/optimizer_version)-ubuntu-20.04/ic-cdk-optimizer" --to-stdout | install -m755 /dev/stdin /usr/local/bin/ic-cdk-optimizer
```
peterpeterparker, line 53: isn't the `ic-cdk-optimizer` deprecated? in II they switched to `ic-wasm` literally last week

PR #2252, opened 2023-04-05
Support .tar.xz snapshots for local development

dskloetd:
# Motivation

xz compresses replica snapshots much better than zip so we started using xz to save snapshot.

# Changes

1. Change `--state-zip / -z` flag in `dev-local-state.sh` and `dfx-snapsho-install` to `--snapshot / -s`
2. Use either `tar` or `unzip` based on the suffix of the snapshot.

# Tests

Ran `scripts/dev-local-state.sh` manually.


bitdivine (approved):
LGTM.  Thanks!

PR #2256, opened 2023-04-05
feat: display from / to BTC network on ckBTC transactions list

peterpeterparker:
# Motivation

Implement following requirements:

img

Rules efined with Mario:

```
export interface Transaction {
  burn: // -> To BTC network
  mint: // -> From BTC network
```

No other checks than the types at the moment.

# Changes

- add an optional record of descriptions that can be passed to `TransactionCard` and render matches
- pass record from ckBTC transactions list

# Notes

- There might be better solution and the types inheritance can be better (not supported by our i18n parser currently).
As it's a display information, went with the straightforward solution.

- Can only simulate it locally. There  is not testnet.

# Screenshot

Capture d’écran 2023-04-05 à 14 17 51


mstrasinskis (approved):
LGTM, thank you!

PR #2259, opened 2023-04-05
Add E2E test for SNS governance

dskloetd:
# Motivation

We don't yet have any E2E test coverage for SNS functionality.

# Changes

1. Add the beginning of an SNS governance end-to-end test. More to come but trying to keep the PRs small-ish.
2. Add page objects for SelectUniverseCard and SelectUniverseList.
3. Introduce versions of querySelector and byTestId which expect a given number of elements and will wait for that number of elements to be present.
4. Add test ID to SelectUniverseList component
5. Rename `getIcP` to `getTokens` as it works just as well for SNS tokens.

# Tests

`npm run test-e2e`


mstrasinskis (approved):
LGTM, just one question.

frontend/src/tests/e2e/sns-governance.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      .getSelectUniverseListPo()
+      .getSnsUniverseCards();
+    expect(snsUniverseCards).toHaveLength(1);
+    const snsProjectName = await snsUniverseCards[0].getName();
+
+    // Our test SNS project names are always 5 uppercase letters.
+    expect(snsProjectName).toMatch(/[A-Z]{5}/);
+
+    await snsUniverseCards[0].click();
+    await appPo.getTokens(20);
```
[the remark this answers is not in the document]
dskloetd: It's already working. The exact same flow that gives ICP gives tokens if you are in the token universe.

frontend/src/tests/e2e/sns-governance.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { AppPo } from "$tests/page-objects/App.page-object";
+import { PlaywrightPageObjectElement } from "$tests/page-objects/playwright.page-object";
+import { signInWithNewUser } from "$tests/utils/e2e.test-utils";
+import { expect, test } from "@playwright/test";
+
+test("Test SNS governance", async ({ page, context }) => {
```
lmuntaner, line 6: Why is this called "Test SNS governance"?

Shouldn't e2e tests be user flows?

frontend/src/tests/e2e/sns-governance.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const snsUniverseCards = await appPo
+    .getSelectUniverseListPo()
+    .getSnsUniverseCards();
+  expect(snsUniverseCards.length).toBeGreaterThanOrEqual(1);
+  const snsUniverseCard = snsUniverseCards[0];
+  const snsProjectName = await snsUniverseCard.getName();
+
+  // Our test SNS project names are always 5 uppercase letters.
+  expect(snsProjectName).toMatch(/[A-Z]{5}/);
```
lmuntaner, line 22: Why is this a useful check?

frontend/src/tests/page-objects/SelectUniverseList.page-object.ts
```diff
@@ -0,0 +1,51 @@
...
+  getSelectUniverseCardPos(): Promise<SelectUniverseCardPo[]> {
+    return SelectUniverseCardPo.allUnder(this.root);
+  }
+
+  isSnsName(name: string): boolean {
+    return !SelectUniverseListPo.NON_SNS_NAMES.includes(name);
+  }
+
+  async waitForSnsUniverseCards(): Promise<void> {
+    const maybeCards = SelectUniverseCardPo.countUnder({
```
lmuntaner, line 26: Why do you need to do it like this?

Why not selecting all cards and then filtering with the name?

frontend/src/tests/e2e/sns-governance.spec.ts
dskloetd: I imagine 2 (or more?) separate SNS tests:
1. One that participates in a sale of an open project.
2. One that operates on an existing project.
They have to be separate because it will not be easy to finalize a sale in the middle of a test.
This will me number 2. I used "governance" to indicate "not sale participation".
But maybe there will be even more, and in this end it will make more sense to call it "stake neuron" or something.
But probably there will be tests that require a staked neuron so in that case I think "stake neuron" will just be part of that test.
Similar to how the accounts test creates a subaccount before making a transaction.

frontend/src/tests/e2e/sns-governance.spec.ts
dskloetd: I'm not sure. I wanted to somehow verify that I got an SNS project card.
Happy to remove it if you don't think it's useful.
This is really just the set up to the test. The "meat" of the test will come in follow-up PRs.

frontend/src/tests/page-objects/SelectUniverseList.page-object.ts
dskloetd: When `waitForSnsUniverseCards` is called, the sns projects aren't loaded yet.
So there is only 1 card: "Internet Computer".
I need some way to select cards that don't exist yet and wait for them.
It doesn't necessarily have to be like this, but I can't just do `allUnder` and filter that.

frontend/src/tests/page-objects/SelectUniverseList.page-object.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  getSelectUniverseCardPos(): Promise<SelectUniverseCardPo[]> {
+    return SelectUniverseCardPo.allUnder(this.root);
+  }
+
+  isSnsName(name: string): boolean {
+    return !SelectUniverseListPo.NON_SNS_NAMES.includes(name);
+  }
+
+  async waitForSnsUniverseCards(): Promise<void> {
```
mstrasinskis, line 25: Question about the naming, does the function waits for sns cards or returns their presence?

dskloetd: It waits until at least 1 sns card exists. Since they are loaded/rendered at the same time, it means that it effectively waits for all the sns cards to appear.

At the time this function is called, the projects are still being loaded, so they don't exist yet. So if it would only check presence, it would just return false which is not useful.

lmuntaner: discussed in a call, all good, thanks for the explanation!

frontend/src/tests/e2e/sns-governance.spec.ts
lmuntaner: discussed in a call, all good, thanks for the explanation!

frontend/src/tests/e2e/sns-governance.spec.ts
lmuntaner: discussed in a call, all good, thanks for the explanation!

PR #2261, opened 2023-04-05
Test/voting results po

mstrasinskis:
# Motivation

Switch to page objects for SnsProposalVotingSection testing.

# Changes

- po / `byRole`
- po / `getAttribute`
- `SnsProposalVotingSectionPo`
- `VotesResultsPo`

# Tests

No changes


frontend/src/tests/lib/components/sns-proposals/SnsProposalVotingSection.spec.ts
```diff
@@ -12,21 +14,24 @@ describe("SnsProposalVotingSection", () => {
...
 
-    const progressbar = getByRole("progressbar");
-    expect(progressbar).toBeInTheDocument();
-    expect(progressbar.getAttribute("aria-valuemin")).toBe("0");
-    expect(progressbar.getAttribute("aria-valuemax")).toBe(`${total}`);
-    expect(progressbar.getAttribute("aria-valuenow")).toBe(`${yes}`);
+    const containerPo = new JestPageObjectElement(container);
+    const po = SnsProposalVotingSectionPo.under(containerPo);
+    const votesResultPo = await po.getVotingsResultsPo();
+    expect(await votesResultPo.isPresent()).toBeTruthy();
```
dskloetd, line 31: `await votesResultPo.isPresent()` returns a boolean so `.toBe(true)` would be more precise.

frontend/src/tests/page-objects/VotesResults.page-object.ts
```diff
@@ -0,0 +1,37 @@
...
+  private constructor(root: PageObjectElement) {
+    super(root);
+  }
+
+  static under(element: PageObjectElement): VotesResultPo {
+    return new VotesResultPo(element.byTestId(VotesResultPo.TID));
+  }
+
+  async getProgressBar(): Promise<PageObjectElement> {
+    return assertNonNullish(await this.root.byRole("progressbar"));
```
dskloetd, line 17: I'm not really familiar with the `role` attribute.
Do you think we'll use this `byRole` helper a lot?
Or should we maybe stick to `.querySelector("[role=progressbar]")` for now?
Or maybe add a `[data-tid]` to the element?
Honest questions because I'm not sure.
For reference, we had a lot of `querySelector("[data-tid=` before I decided to add `byTestId` for convenience.

frontend/src/tests/page-objects/VotesResults.page-object.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  async getProgressMinValue(): Promise<bigint> {
```
dskloetd, line 20: Does this need to be a `bigint` as opposed to just `number`?

frontend/src/tests/page-objects/VotesResults.page-object.ts
```diff
@@ -0,0 +1,37 @@
...
+    return assertNonNullish(
+      BigInt(await (await this.getProgressBar()).getAttribute("aria-valuemin"))
+    );
+  }
+
+  async getProgressMaxValue(): Promise<bigint> {
```
dskloetd, line 26: Same (bigint)

frontend/src/tests/page-objects/VotesResults.page-object.ts
```diff
@@ -0,0 +1,37 @@
...
+    return assertNonNullish(
+      BigInt(await (await this.getProgressBar()).getAttribute("aria-valuemax"))
+    );
+  }
+
+  async getProgressNowValue(): Promise<bigint> {
```
dskloetd, line 32: Same

frontend/src/tests/types/page-object.types.ts
```diff
@@ -6,11 +6,13 @@ export interface PageObjectElement {
   querySelector(selector: string): PageObjectElement;
   querySelectorAll(selector: string): Promise<PageObjectElement[]>;
   byTestId(tid: string): PageObjectElement;
+  byRole(role: string): Promise<PageObjectElement>;
```
dskloetd, line 9: Like `querySelector`, this doesn't need to return `Promise`.

frontend/src/tests/types/page-object.types.ts
```diff
@@ -6,11 +6,13 @@ export interface PageObjectElement {
...
   allByTestId(tid: string): Promise<PageObjectElement[]>;
   isPresent(): Promise<boolean>;
   waitFor(): Promise<void>;
   waitForAbsent(): Promise<void>;
   getText(): Promise<string | null>;
+  getAttribute(attribute: string): Promise<string | null>;
```
dskloetd, line 15: Can you also add implementations to `playwright.page-object.ts`?
It's fine if they `throw new Error("Not implement")` until they are used.

mstrasinskis: Done.

frontend/src/tests/page-objects/VotesResults.page-object.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  private constructor(root: PageObjectElement) {
+    super(root);
+  }
+
+  static under(element: PageObjectElement): VotesResultPo {
+    return new VotesResultPo(element.byTestId(VotesResultPo.TID));
+  }
+
+  async getProgressBar(): Promise<PageObjectElement> {
```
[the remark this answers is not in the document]
mstrasinskis: Thanks, removed.

frontend/src/tests/types/page-object.types.ts
mstrasinskis: Thanks, `byRole` was removed.

frontend/src/tests/page-objects/jest.page-object.ts
```diff
@@ -43,6 +43,10 @@ export class JestPageObjectElement implements PageObjectElement {
     return this.querySelectorAll(`[data-tid=${tid}]`);
   }
 
+  async byRole(role: string) {
```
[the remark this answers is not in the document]
mstrasinskis: `byRole` was removed.

frontend/src/tests/page-objects/VotesResults.page-object.ts
mstrasinskis: Thanks, was switched to number.

frontend/src/tests/page-objects/VotesResults.page-object.ts
mstrasinskis: Agree, there are currently not so many occurrences, let's use data-tid.

frontend/src/tests/lib/components/sns-proposals/SnsProposalVotingSection.spec.ts
mstrasinskis: Good point, true!

frontend/src/tests/lib/components/proposal-detail/VotesResults.spec.ts
```diff
@@ -25,24 +25,36 @@ describe("VotesResults", () => {
...
-    const { getByText } = renderResult;
-    expect(getByText(`${formatNumber(yes)}`)).toBeInTheDocument();
+  it('should render "Adopt" value', async () => {
+    const { container } = renderResult;
+    const votesResultPo = VotesResultPo.under(
+      new JestPageObjectElement(container)
+    );
+
+    expect(await votesResultPo.getAdoptVotingPower()).toEqual(
+      `${formatNumber(yes)}`
```
dskloetd, line 35: Better expect literal values. As the reader I still have no idea what the value is supposed to be.

frontend/src/tests/lib/components/proposal-detail/VotesResults.spec.ts
```diff
@@ -25,24 +25,36 @@ describe("VotesResults", () => {
...
-    const { getByText } = renderResult;
-    expect(getByText(`${formatNumber(no)}`)).toBeInTheDocument();
+  it('should render "Reject" value', async () => {
+    const { container } = renderResult;
+    const votesResultPo = VotesResultPo.under(
+      new JestPageObjectElement(container)
+    );
+
+    expect(await votesResultPo.getRejectVotingPower()).toEqual(
+      `${formatNumber(no)}`
```
dskloetd, line 46: Expect a literal value.

frontend/src/tests/lib/components/sns-proposals/SnsProposalVotingSection.spec.ts
```diff
@@ -27,11 +28,14 @@ describe("SnsProposalVotingSection", () => {
...
-    const votesResultPo = await po.getVotingsResultsPo();
-    expect(await votesResultPo.isPresent()).toBeTruthy();
+    const votesResultsPo = await po.getVotingsResultsPo();
 
-    expect(await votesResultPo.getProgressMinValue()).toBe(0n);
-    expect(await votesResultPo.getProgressMaxValue()).toBe(total);
-    expect(await votesResultPo.getProgressNowValue()).toBe(yes);
+    expect(await votesResultsPo.isPresent()).toBeTruthy();
+    expect(await votesResultsPo.getAdoptVotingPower()).toEqual(
+      `${formatNumber(Number(yes))}`
```
dskloetd, line 35: Can you expect a literal value?

frontend/src/tests/lib/components/sns-proposals/SnsProposalVotingSection.spec.ts
```diff
@@ -27,11 +28,14 @@ describe("SnsProposalVotingSection", () => {
...
+    );
+    expect(await votesResultsPo.getRejectVotingPower()).toEqual(
+      `${formatNumber(Number(no))}`
```
dskloetd, line 38: And here was well

frontend/src/tests/page-objects/playwright.page-object.ts
```diff
@@ -38,6 +38,10 @@ export class PlaywrightPageObjectElement implements PageObjectElement {
     return this.locator.textContent();
   }
 
+  getAttribute(attribute: string): Promise<string | null> {
```
dskloetd, line 41: Add an underscore in front of the name to indicate that it's not used: `_attribute`

frontend/src/tests/lib/components/proposal-detail/VotesResults.spec.ts
mstrasinskis: As a reader, does this knowledge gives any advantage? One argument against it is the dependency on the formatting implementation and localisation. After changes in formatting all the related tests needs to be updated.

frontend/src/tests/lib/components/proposal-detail/VotesResults.spec.ts
```diff
@@ -25,24 +25,36 @@ describe("VotesResults", () => {
...
   });
 
-  it('should render "Adopt" value', () => {
-    const { getByText } = renderResult;
-    expect(getByText(`${formatNumber(yes)}`)).toBeInTheDocument();
+  it('should render "Adopt" value', async () => {
+    const { container } = renderResult;
+    const votesResultPo = VotesResultPo.under(
+      new JestPageObjectElement(container)
+    );
```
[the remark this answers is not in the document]
mstrasinskis: Good idea, done.

frontend/src/tests/lib/components/proposal-detail/VotesResults.spec.ts
dskloetd: I have 2 reasons:
1. Tests function as documentation, explaining what the code does. If I have to decipher code to understand the test, it doesn't work well as documentation.
2. If you're using `formatNumber` wrong for some reason (maybe it needs 2 arguments, or it needs another dependency to be included or something) and it actually always returns an empty string, you wouldn't notice.

Note that "format is tested separately" doesn't really answer (2), because format might work in its own test but not here. That's why I said "using it wrong" rather than "is broken".

mstrasinskis: Thank you for the detailed explanation! Updated 👍

frontend/src/tests/lib/components/proposal-detail/VotesResults.spec.ts
```diff
@@ -25,25 +21,19 @@ describe("VotesResults", () => {
...
+    expect(await votesResultPo.getRejectVotingPower()).toEqual("3.00");
   });
 
   it("should render progressbar", async () => {
     const votesResultPo = renderComponent();
 
     expect(await votesResultPo.getProgressMinValue()).toBe(0);
-    expect(await votesResultPo.getProgressMaxValue()).toBe(yes + no);
-    expect(await votesResultPo.getProgressNowValue()).toBe(yes);
+    expect(await votesResultPo.getProgressMaxValue()).toBe(5);
```
dskloetd, line 36: By the way, if you use the same literal value in multiple places, it's fine/good to use a constant.
It was specifically the calculation that I was arguing against.
With a constant defined in the same file, it's easy to look up the value, so that's different from a calculation.

You can also have constants like
```
const yesCount = 10;
const yesFormatted = "10.00";
```
If you want to make it extra clear those are related.

dskloetd (approved):
Thanks! One more suggestion, if you're not yet tired of the review.

frontend/src/tests/lib/components/proposal-detail/VotesResults.spec.ts
mstrasinskis: Yes I do, thanks.

dskloetd (approved):
Thanks!

PR #2265, opened 2023-04-06
E2E: Verify account transactions

dskloetd:
# Motivation

We want to have end-to-end test for product requirements.
In this PR we add:
AU005: The user MUST be able to see the transactions of a specific account.

# Changes

1. In the accounts e2e test, after making the transaction, check that the transaction can be seen in the subaccount.
2. Add a test ID to TransactionList.
3. Add page objects for TransactionList and TransactionCard
4. Add functionality in page objects for NnsAccounts, NnsWallet and SkeletonCard

# Tests

npm run test-e2e

### DIFF
--- a/frontend/src/tests/page-objects/TransactionCard.page-object.ts
+++ b/frontend/src/tests/page-objects/TransactionCard.page-object.ts
@@ -1,0 +1,27 @@
+import { AmountDisplayPo } from "$tests/page-objects/AmountDisplay.page-object";
+import { BasePageObject } from "$tests/page-objects/base.page-object";
+import type { PageObjectElement } from "$tests/types/page-object.types";
+
+export class TransactionCardPo extends BasePageObject {
+  private static readonly TID = "transaction-card";
+
+  static async allUnder(
+    element: PageObjectElement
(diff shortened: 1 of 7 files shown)


frontend/src/tests/page-objects/TransactionCard.page-object.ts
```diff
@@ -0,0 +1,27 @@
...
+  static async allUnder(
+    element: PageObjectElement
+  ): Promise<TransactionCardPo[]> {
+    return Array.from(await element.allByTestId(TransactionCardPo.TID)).map(
+      (el) => new TransactionCardPo(el)
+    );
+  }
+
+  getIdentifier(): Promise<string> {
+    return this.getText("identifier");
```
lmuntaner, line 17: Shouldn't we add a data-tid for this?

lmuntaner (approved):
LGTM, just one comment but already approving

frontend/src/tests/page-objects/TransactionCard.page-object.ts
lmuntaner: oh, my bad, I didn't remember that `getText` expects the data-tid

PR #2271, opened 2023-04-11
Feat: Use only type for SNS proposals, no topic

lmuntaner:
# Motivation

Align NNS Dapp with the agreements reached within the team on how to deal with Type and Topic for SNS proposals.

Result:
* Topic does not exist yet for SNS proposals.
* The type is the name of the nervous function.

# Changes

* Change mapProposalInfo in sns proposal utils to return the type and typeDescription instead of topic and topicDescription.
* Do not render topic and type items in the SnsProposalSystemInfoSection. Render only the type item.
* Render the type as title in the SnsProposalSystemInfoSection component.

# Tests

* Adapt tests to new functionality and also change the test descriptions to match the changes.
* Remove getTopicText from SnsProposalSystemInfoSection page object.
* Change testId in SnsProposalCard and add it to the type. This is used to test that the nervous functions are properly loaded in the SnsProposals component test.

### DIFF
--- a/frontend/src/lib/components/proposals/ProposalCard.svelte
+++ b/frontend/src/lib/components/proposals/ProposalCard.svelte
@@ -10,9 +10,9 @@
   export let id: bigint | undefined;
   export let title: string | undefined;
   export let color: ProposalStatusColor | undefined;
-  export let topic: string | undefined;
+  export let topic: string | undefined = undefined;
   export let proposer: string | undefined;
-  export let type: string | undefined = undefined;
+  export let type: string | undefined;
   export let deadlineTimestampSeconds: bigint | undefined;
 </script>
 
@@ -24,7 +24,7 @@
 
     <div class="meta-data">
       {#if type !== undefined}
-        <KeyValuePair>
+        <KeyValuePair testId="proposal-type">
           <span slot="key">{$i18n.proposal_detail.type_prefix}</span>
           <span
             slot="value"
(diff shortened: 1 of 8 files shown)


lmuntaner:
New UI:

Screenshot 2023-04-11 at 09 07 52
Screenshot 2023-04-11 at 09 08 35

frontend/src/lib/components/proposals/ProposalCard.svelte
```diff
@@ -24,7 +24,7 @@
 
     <div class="meta-data">
       {#if type !== undefined}
-        <KeyValuePair>
+        <KeyValuePair testId="proposal-type">
```
lmuntaner, line 27: Used to test that nervous functions are loaded in SnsProposals spec

lmuntaner:

frontend/src/lib/components/proposals/ProposalCard.svelte
```diff
@@ -10,9 +10,9 @@
   export let id: bigint | undefined;
   export let title: string | undefined;
   export let color: ProposalStatusColor | undefined;
-  export let topic: string | undefined;
+  export let topic: string | undefined = undefined;
```
dskloetd, line 13: Is there a difference between initializing to `undefined` and not initializing to anything?

lmuntaner: Yes, if it's not initialized, then it's a mandatory prop.

Because SNS doesn't have `topic` I wanted to not force adding `topic={undefined}` when I use the ProposalCard in the SnsProposalCard.

dskloetd: Thanks!

dskloetd (approved):

PR #2272, opened 2023-04-11
feat: select account to stake neuron with standard dropdown

peterpeterparker:
# Motivation

Use standard dropdown component to select source account to stake a neuron.

# Changes

- refactor and extract common code from `TransactionForm` to `TransactionFormAccount` and `TransactionFormFee`
- use new components in stake modal
- instead of updating steps on account selection, update stake neuron modal steps with HW step after neuron is created

# Tests

Not that much particular new tests as this changes does not introduces any feature changes - i.e. functionally speaking changes were already covered.

# Screenshots

Capture d’écran 2023-04-11 à 12 35 43


peterpeterparker:
> Couldn't you use the `TransactionForm` instead of the pieces?

I followed the existing pattern which already used only pieces. If possible sure not against.

> Question, do we want `StakeNeuronModal` to use the `TransactionModal`? I'm not sure, just wondering.

There is no review step and 3-4 other steps in `StakeNeuronModal`, so not sure we can. But if there is a clean generic maybe?

> Why this change? Do you new designs on staking a neuron?

1. The display of the transaction balance and fees is about to change
2. Selecting accounts with cards is deprecated design wise since a while

In both cases, this change align the UI/UX of selecting an account to stake a neuron to the rest of the app - i.e. consistency

Thanks for the review!

peterpeterparker:
Here also a screenshot of the PR I'm working on which adds a "Available balance" line in the modal.

Capture d’écran 2023-04-11 à 13 43 15

lmuntaner:
>There is no review step and 3-4 other steps in StakeNeuronModal, so not sure we can. But if there is a clean generic maybe?

I like the idea of using the pieces for now. But forcing us to reuse the same component helps us and design to have a consistent UX.

Not for now, but we should have this as a requirement for flows with transactions maybe. This technical limitation should also be a design one to force consistency.

At the same time, I don't want this generic component to start accommodating for hundreds of cases. UX should be simple, and therefore the generic should also be simple.

PR #2273, opened 2023-04-11
Chore: Expand SnsProposalDetail test cases

lmuntaner:
# Motivation

Add missing test cases in SnsProposalDetail spec.

# Changes

* Expose Payload and Summary Section in SnsProposalDetail page object.
* Add test case for payload section specifically that renders the actual payload. It's important that we test that the payload field is rendered, not just the card.
* Add expect for SystemInfoSection and SymmarySection presence in test case for content loaded. Test on contents are done in each component test.
* Pause and resume the fake in the test for the skeleton. This way the test case is controlled.

# Tests

List of changes above.

### DIFF
--- a/frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
@@ -93,7 +93,7 @@ describe("SnsProposalDetail", () => {
       expect(await po.getSkeletonDetails().isPresent()).toBe(false);
     });
 
-    it("should render the name of the nervous function", async () => {
+    it("should render the name of the nervous function as title", async () => {
       const functionId = BigInt(12);
       const functionName = "test function";
       fakeSnsGovernanceApi.addNervousSystemFunctionWith({
--- a/frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
@@ -12,6 +12,7 @@ import { mockAuthStoreNoIdentitySubscribe } from "$tests/mocks/auth.store.mock";
 import { mockCanisterId } from "$tests/mocks/canisters.mock";
 import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
 import { SnsProposalDetailPo } from "$tests/page-objects/SnsProposalDetail.page-object";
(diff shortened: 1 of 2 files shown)


frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -75,37 +88,12 @@ describe("SnsProposalDetail", () => {
 
       fakeSnsGovernanceApi.resume();
       await waitFor(async () => expect(await po.isContentLoaded()).toBe(true));
+      expect(await po.hasSummarySection()).toBe(true);
+      expect(await po.hasSystemInfoSection()).toBe(true);
```
lmuntaner, line 92: I added this test case instead of having the same setup and expect in "should render system info content".

This test case is about rendering the content without specifics. Which covers that the sections are present.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -120,18 +108,26 @@ describe("SnsProposalDetail", () => {
...
+      const po = await renderComponent();
 
-      const po = SnsProposalDetailPo.under(
-        new JestPageObjectElement(container)
+      await waitFor(async () =>
+        expect(await po.getSystemInfoSectionTitle()).toBe(functionName)
       );
+    });
+
+    it("should render the payload", async () => {
```
lmuntaner, line 118: The payload is a special part and I'd rather have a test that the text is rendered instead of only checking that the section is present.

lmuntaner:
@dskloetd please review

dskloetd (approved):

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -40,20 +54,20 @@ describe("SnsProposalDetail", () => {
...
         props: {
           proposalIdText: proposalId.id.toString(),
         },
       });
-
       const po = SnsProposalDetailPo.under(
         new JestPageObjectElement(container)
       );
       expect(await po.getSkeletonDetails().isPresent()).toBe(true);
+      fakeSnsGovernanceApi.resume();
```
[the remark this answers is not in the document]
lmuntaner: ok, thanks!

PR #2274, opened 2023-04-11
Bundle the rest

bitdivine:
# Motivation

# Changes

# Tests

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -15,7 +15,6 @@ const config: UserConfig = {
     target: "es2020",
     rollupOptions: {
       output: {
-	chunkFileNames: "[name].js",
         manualChunks: (id) => {
           const folder = dirname(id);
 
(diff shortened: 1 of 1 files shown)


frontend/vite.config.ts
```diff
@@ -66,6 +67,8 @@ const config: UserConfig = {
           ) {
             return "dapp";
           }
+
+	  return "etc";
```
peterpeterparker, line 71: Returning `undefined`, the default, let the bundler create chunks. This is the current and expected behavior. In addition, this most probably will make SvelteKit fails at runtime.

peterpeterparker (changes requested):

frontend/vite.config.ts
bitdivine: The bundler also creates chunks like this.  Note that manually specifying names doesn't necessarily resolve this though if the hash is still in the filename and removing the hash can cause an issue whereby a user gets some files from one release and some from another.  _if_ we remove the hash we would need to replace it with something that distinguishes releases, at least.

PR #2285, opened 2023-04-12
Feat: Add Subtitle to SNS CF neurons section

lmuntaner:
# Motivation

Explain better the nature of the SNS neurons coming from the Community Fund.

# Changes

* Add a subtitle in the SNS Neurons from the CF.

# Tests

* Check that subtitle text is present.

### DIFF
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -651,6 +651,7 @@
     "header": "Neuron",
     "all_topics": "All topics",
     "community_fund_section": "Community Fund",
+    "community_fund_section_description": "You are not the controller of this SNS neuron. To facilitate bootstrapping the SNS DAO governance, the NNS treasury has temporarily granted hotkey access to you for voting and following.",
     "add_hotkey_info": "In order to use this neuron to vote from the corresponding dapp, you have to add your personal dapp ID as a hotkey here. Please refer to the instructions given by the corresponding dapp.",
     "add_hotkey_tooltip": "In order to use this neuron to vote from the corresponding dapp, you have to add your personal dapp ID as a hotkey here. Please refer to the instructions given by the corresponding dapp."
   },
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -651,7 +651,7 @@
     "header": "Neuron",
     "all_topics": "All topics",
     "community_fund_section": "Community Fund",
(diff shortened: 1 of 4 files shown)


lmuntaner:
New UI:

Screenshot 2023-04-12 at 13 51 42

peterpeterparker:
Shouldn't this informational text be implemented with an information call to action (like `KeyValuePairInfo` but without value) rather than as a static text. Wouldn't that be more accuracte UI/UX?

lmuntaner:
> dn't that be more accuracte UI/UX?

That was my first solution also, but product and Dom required the message to be more prominent than that.

More in the slack thead: https://dfinity.slack.com/archives/C03H6QEPW5D/p1681280345211209

frontend/src/lib/i18n/en.json
```diff
@@ -651,6 +651,7 @@
     "header": "Neuron",
     "all_topics": "All topics",
     "community_fund_section": "Community Fund",
+    "community_fund_section_description": "You are not the controller of this SNS neuron. To facilitate bootstrapping the SNS DAO governance, the NNS treasury has temporarily granted hotkey access to you for voting and following.",
```
dskloetd, line 654: Where did this text come from? To me it raises more questions than answers. People only get these if they participate in the community fund, right? Presumably they read about what this means before they participated? Where would they have read about it? Can we link to there or use a text that's similar to that?

> this SNS neuron

This is singular, but int he screenshot I see multiple neurons. Should the text be singular/plural depending on the number of neurons?

> To facilitate bootstrapping the SNS DAO governance

What does this mean? Do SNS DAO's really depend on the community fund for bootstrapping? Isn't the swap sale for that?

> has temporarily granted

Temporarily? So it will be revoked at some point? When?

frontend/src/lib/pages/SnsNeurons.svelte
```diff
@@ -88,12 +88,13 @@
...
     <h2
       data-tid="community-fund-title"
-      class={`bottom-margin ${
-        $sortedSnsUserNeuronsStore.length > 0 ? "top-margin" : ""
-      }`}
+      class={$sortedSnsUserNeuronsStore.length > 0 ? "top-margin" : ""}
     >
       {$i18n.sns_neuron_detail.community_fund_section}
     </h2>
+    <p data-tid="community-fund-description" class="bottom-margin">
```
dskloetd, line 95: Since the description is static (until it will be singular/plural depending on number of neurons?) I'm not sure what's the purpose of the `data-tid`.

lmuntaner: I don't need it, I'll change it.

lmuntaner:

dskloetd (approved):

frontend/src/lib/pages/SnsNeurons.svelte
lmuntaner: Yes, I had to use it. Because the text has html, I get the element and compare the text content. Instead of just getting the text directly

PR #2288, opened 2023-04-12
Implement waitForAbsent on JestPageObjectElement

dskloetd:
# Motivation

We use page objects for both e2e tests and jest unit tests.
The e2e tests and unit tests that use page objects will overlap more and more and it is good to use the same patterns in both of them.

Also, JestPageObjectElement.isPresent wasn't using the most recent state of the DOM to determine if an element is present.

# Changes

1. Add a private method `getRootAndFullSelector` to `JestPageObjectElement` to get the furthest available ancestor of the current element and the full selector to point from it to the current element.
2. Use `getRootAndFullSelector` to implement `isPresent` and update this.element.
3. Use `isPresent` to implement both `waitFor` and `waitForAbsent`.

`waitForAbsent` currently isn't used by any unit tests, but the new implementation is so straightforward that it will likely work and some testing from another branch suggests that it does.

# Tests

`npm run test`

### DIFF
--- a/frontend/src/tests/page-objects/jest.page-object.ts
+++ b/frontend/src/tests/page-objects/jest.page-object.ts
@@ -63,7 +63,10 @@ export class JestPageObjectElement implements PageObjectElement {
     // return this.querySelectorCount({ selector: `[data-tid=${tid}]`, count });
   }
 
-  private getRootAndFullSelector(): { rootElement: Element; fullSelector: string } {
+  private getRootAndFullSelector(): {
+    rootElement: Element;
+    fullSelector: string;
+  } {
     if (isNullish(this.parent)) {
       return { rootElement: this.element, fullSelector: ":scope" };
     }
@@ -75,7 +78,7 @@ export class JestPageObjectElement implements PageObjectElement {
   }
 
   async isPresent(): Promise<boolean> {
-    const {rootElement, fullSelector} = this.getRootAndFullSelector();
+    const { rootElement, fullSelector } = this.getRootAndFullSelector();
     this.element = rootElement.querySelector(fullSelector);
     return nonNullish(this.element);
   }
--- a/frontend/src/tests/page-objects/jest.page-object.ts
+++ b/frontend/src/tests/page-objects/jest.page-object.ts
@@ -63,26 +63,33 @@ export class JestPageObjectElement implements PageObjectElement {
     // return this.querySelectorCount({ selector: `[data-tid=${tid}]`, count });
   }
 
+  private getRootAndFullSelector(): { rootElement: Element; fullSelector: string } {
+    if (isNullish(this.parent)) {
+      return { rootElement: this.element, fullSelector: ":scope" };
+    }
+    const { rootElement, fullSelector } = this.parent.getRootAndFullSelector();
+    return {
+      rootElement,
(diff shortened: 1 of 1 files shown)


frontend/src/tests/page-objects/jest.page-object.ts
```diff
@@ -63,26 +63,36 @@ export class JestPageObjectElement implements PageObjectElement {
...
+    const { rootElement, fullSelector } = this.parent.getRootAndFullSelector();
+    return {
+      rootElement,
+      fullSelector: `${fullSelector} ${this.selector}`,
+    };
+  }
+
   async isPresent(): Promise<boolean> {
+    const { rootElement, fullSelector } = this.getRootAndFullSelector();
+    this.element = rootElement.querySelector(fullSelector);
```
lmuntaner, line 82: Why don't you use:

```
this.element = this.parent.querySelector(this.selector);
```

lmuntaner (commented):
A question

frontend/src/tests/page-objects/jest.page-object.ts
dskloetd: The parent might be absent as well. Or do you mean that querySelector would deal with that recursively?

dskloetd: Actually, parent.querySelector returns `JestPageObjectElement` but I need an `Element`.

lmuntaner: If the parent is absent, you'll get `this.element.querySelector(...)` in your solution as well. That won't do either, I guess.

True about return value. Maybe `this.parent.element.querySelector(this.selector)` if parent is defined.

If the parent is not there, current implementation of `waitFor` won't work either:

```
    await waitFor(() => {
      this.element = this.parent?.element.querySelector(this.selector);
      expect(this.element).not.toBeNull();
    });
```

dskloetd: By the parent being absent, I mean `parent.element === null`, not `parent === null`.
The fullSelector will point directly to the desired element from an element which is the container containing the rendered component generated by the Svelte testing library.
The current implementation of waitFor works because it waits for the parent before waiting for the element, so at that point we know the parent is present.

lmuntaner: And what's wrong the the current implementation?

How about waiting for the parent to be present here as well?

I feel that we have enough methods to implement this functionality. That's why I' not sure of this implementation, sorry.

dskloetd: We have no current implementation for `waitForAbsent`.
How do you propose we implement it?

lmuntaner: Can you use the code you had in`waitFor` for `isPresent`?

```
this.element = this.parent?.element.querySelector(this.selector);
return nonNullish(this.element);
```

lmuntaner: Ahh, this edge case is what `getRootAndFullSelector` is trying to solve. I understand now.

Thanks for the explanation. Sorry for the long feedback loop

lmuntaner (approved):

PR #2293, opened 2023-04-13
Fix: Set II release to be used

lmuntaner:
# Motivation

Deploying II from nns-dapp repo to a testnet deploys a broken II for testnets.

This PR sets the II version to be used, instead of relying in the latest version.

# Changes

* Change build step for internet_identity project in dfx.json

# Tests

Not applicable


nmattia (approved):
Rubber stamp

PR #2294, opened 2023-04-13
Feat: Add caching for known neurons calls

lmuntaner:
# Motivation

Reduce the number of calls by caching the known neurons.

# Changes

* Service to load known neuron  `listKnownNeurons` uses api service layer.
* Implemet caching for `queryKnownNeurons` in the Governance Api Sevice. Caching doesn't depend on identity because it's the same for all users.

# Tests

* Test that caching works as expected.


lmuntaner:
@dskloetd please review

dskloetd (approved):

PR #2296, opened 2023-04-13
refactor: move votingNeuronSelectStore to voting

mstrasinskis:
# Motivation

Move votingNeuronSelectStore to voting store file because it's directly related to the voting.

# Tests

- no logic changes

### DIFF
--- a/frontend/src/tests/lib/stores/proposals.store.spec.ts
+++ b/frontend/src/tests/lib/stores/proposals.store.spec.ts
@@ -3,8 +3,8 @@ import {
   proposalPayloadsStore,
   proposalsFiltersStore,
   proposalsStore,
-  votingNeuronSelectStore,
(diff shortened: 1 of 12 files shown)


frontend/src/tests/lib/stores/proposals.store.spec.ts
```diff
@@ -3,8 +3,8 @@ import {
   proposalPayloadsStore,
   proposalsFiltersStore,
   proposalsStore,
-  votingNeuronSelectStore,
 } from "$lib/stores/proposals.store";
+import { votingNeuronSelectStore } from "$lib/stores/vote-registration.store";
```
lmuntaner, line 7: Shouldn't you move the tests as well?

mstrasinskis: Good catch! Were moved.

PR #2298, opened 2023-04-13
Keep sourcemaps when building assets.tar.xz

dskloetd:
# Motivation

When debugging issues with compiled JS it can be useful to be able to decode the compiled JS.
This is what sourcemaps are for.

# Changes

1. Change vite config to output sourcemaps. The "hidden" option means that the sourcemaps are created but not served by the server.
2. Change `build-frontend.sh` to put the sourcemaps in a separate archive `sourcemaps.tar.xz`.
3. Change Dockerfile to copy out the `sourcemaps.tar.xz`.
4. Change CI workflows to upload `sourcemaps.tar.xz`
5. Ignore `sourcemaps.tar.xz` in `.gitignore` and `.dockerignore`.

# Tests

1. Built with and without these changes and made sure that assets.tar.xz still has the same sha256.
2. Downloaded [NNS mainnet frontend assets](https://github.com/dfinity/nns-dapp/suites/12218597384/artifacts/645926594) from https://github.com/dfinity/nns-dapp/actions/runs/4691620755 and checked that the sourcemaps are included.


lmuntaner (approved):
LGTM!

PR #2299, opened 2023-04-13
Don't install ii in e2e CI

dskloetd:
# Motivation

Installing internet-identity during nns-dapp CI takes time.
If we make sure internet-identity is already installed in the snapshot, we don't have to do it every time during CI.

# Changes

1. Use a snapshot tagged `testing` which has internet-identity installed. We want to use a different tag later. See also the discussion at https://github.com/dfinity/snsdemo/pull/122#pullrequestreview-1385518605
2. No longer install internet identity during the E2e workflow.

# Tests

CI


bitdivine (approved):
We might want to change the tag later but other than that this looks good.  Thanks!

dskloetd:
I've merged this since I realized `tip` is also not the tag we want to be using. So temporarily using `testing` isn't any worse.

PR #2301, opened 2023-04-14
Chore: Clean deploy.sh

lmuntaner:
# Motivation

Clean `./deploy.sh`. Most of the functionality there was outdated or not used anymore.

Current script offers only two options:

* Deploy the nns dapp to a network.
* Deploy II to a network

# Changes

* Changes in `./deploy.sh`

# Tests

Not applicable.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -7,9 +7,9 @@ help_text() {
 
 	Deploys the nns-dapp or internet identity to a network.
 
-  This file will recreate the arguments file and override it.
+	This file will recreate the arguments file and override it.
 
-  If you want to use a custom arguments file, use dfx directly.
+	If you want to use a custom arguments file, use dfx directly.
 
 	Usage:
 
@@ -20,9 +20,9 @@ help_text() {
 	./deploy.sh --nns-dapp <network>
 	  Deploys nns-dapp to the selected network. The default is "local".
 
-The available networks are the ones listed in dfx.json.
+	The available networks are the ones listed in dfx.json.
 
-EOF
+	EOF
 }
 
 #
--- a/deploy.sh
+++ b/deploy.sh
@@ -5,106 +5,29 @@ cd "$(dirname "$(realpath "$0")")" || exit
 help_text() {
   cat <<-"EOF"
 
-	Deploys the nns-dapp to a network or to local dfx:
-	- Starts dfx (optional)
-	- Installs governance canisters (optional)
-	- Installs Internet Identity (optional)
-	- Installs the NNS Dapp
-	- Opens the NNS dapp in a browser (optional)
-
-	On success the deployment is ready for development or testing:
-	- frontend:
-	    cd frontend
-	    npm ci
-	    npm run dev
-	- testing:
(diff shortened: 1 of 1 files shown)


lmuntaner:

dskloetd:
Fixes and cleanups should not be mixed in a single PR.
The fix should be small and clear but because of all the cleanup, it's hard to tell what's what.

lmuntaner:
There is no good deed that goes unpunished

deploy.sh
```diff
@@ -324,71 +66,9 @@ if [[ "$DEPLOY_NNS_DAPP" == "true" ]]; then
   # Note:  NNS dapp is the only canister provided by this repo, however dfx.json
   #        includes other canisters for testing purposes.  If testing you MAY wish
   #        to deploy these other canisters as well, but you probably don't.
+  ./config.sh
```
lmuntaner, line 69: If we migrate to `./scripts/docker-build` the args file won't be created where it's expected.

Therefore, I added this extra step here.

lmuntaner:
@dskloetd ready for the review now

deploy.sh
dskloetd: Is this a cleanup or a fix?
Does that mean that previously we relied on the args file already existing but now we don't?
Would it be confusing if someone expects their existing args file to be used but instead it's overwritten?
Would it make sense to take the file from where it is to where it is expected rather than to recreate it?
These are not leading questions. I never deploy so I don't know the answers.

dskloetd (commented):
I have rarely used or looked at this script.
I think it would be good if someone more familiar reviews it.
It's quite surprising that so much can be removed.

lmuntaner:
@bitdivine please review

deploy.sh
```diff
@@ -329,66 +75,3 @@ if [[ "$DEPLOY_NNS_DAPP" == "true" ]]; then
   OWN_CANISTER_URL="$(grep OWN_CANISTER_URL <"$CONFIG_FILE" | sed "s|VITE_OWN_CANISTER_URL=||g")"
```
bitdivine, line 75: In principle this could return two or more lines.  You could fix that with `| head -n1` or like this:
```suggestion
  OWN_CANISTER_URL="$(sed -n 's|^VITE_OWN_CANISTER_URL=||;ta;b;:a;p;q')"
```
but this is a rather hypothetical situation and I'm fine with the code as is.

The sed:
```
-n == don't print by default
s == substitute; no g is needed as we will substitute only once.
ta == if there was a match, jump to a
b == branch == jump to <tag unspecified there for just skip the remaining commands for this line>
:a == the tag to jump to
p == print
q == quit
```

bitdivine (approved):
LGTM, thanks!

deploy.sh
```diff
@@ -5,106 +5,33 @@ cd "$(dirname "$(realpath "$0")")" || exit
 help_text() {
   cat <<-"EOF"
```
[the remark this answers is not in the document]
lmuntaner: Thanks for fixing it!

PR #2303, opened 2023-04-14
GIX-1481: Render last maturity distribution date

lmuntaner:
# Motivation

We want to show the last time that maturity was distributed.

This PR implements rendering the date of the last reward in the maturity card.

# Changes

* Render another KeyValuePair in NnsNeuronMaturityCard with the latest maturity distribution date.
* Adapt styles to combine both staked maturity and latest distribution pairs.

# Tests

* Add test in NnsNeuronDetail page to check that last maturity distribution date is rendered.
* Add test in NnsNeuronMaturityCard to check that last maturity distribution date is rendered or not.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
@@ -42,13 +42,15 @@
   </KeyValuePair>
 
   {#if hasStakedMaturity(neuron)}
-    <KeyValuePair testId="staked-maturity">
-      <svelte:fragment slot="key">{$i18n.neurons.staked}</svelte:fragment>
+    <div class="details">
+      <KeyValuePair testId="staked-maturity">
+        <svelte:fragment slot="key">{$i18n.neurons.staked}</svelte:fragment>
 
-      <span slot="value" class="staked-maturity"
-        >{formattedStakedMaturity(neuron)}</span
-      >
-    </KeyValuePair>
+        <span slot="value">{formattedStakedMaturity(neuron)}</span>
+      </KeyValuePair>
+
+      <!-- TODO: Add Last maturity distribution date -->
+    </div>
   {/if}
 
   {#if allowedToStakeMaturity}
--- a/frontend/src/tests/lib/components/neuron-detail/NnsNeuronMaturityCard.spec.ts
+++ b/frontend/src/tests/lib/components/neuron-detail/NnsNeuronMaturityCard.spec.ts
@@ -22,13 +22,8 @@ import {
 import en from "$tests/mocks/i18n.mock";
 import { mockFullNeuron, mockNeuron } from "$tests/mocks/neurons.mock";
 import { mockRewardEvent } from "$tests/mocks/nns-reward-event.mock";
-// TODO: Fix prettier command. For some reason it moves the JestPageObjectElement import after the KeyValuePairPo import.
-// Yet, it doesn't happen in other files like NnsNeuronDetails.spec.ts
-// VSCode prettier plugin works fine.
-// prettier-ignore
 import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
-// prettier-ignore
-import { KeyValuePairPo } from "$tests/page-objects/KeyValuePair.page-object";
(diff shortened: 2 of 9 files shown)


lmuntaner:
New UI:

Screenshot 2023-04-14 at 11 16 57

This is when we have also staked maturity:

Screenshot 2023-04-14 at 10 52 23

lmuntaner:

frontend/src/lib/components/sns-neuron-detail/SnsNeuronMaturityCard.svelte
```diff
@@ -42,13 +42,15 @@
   </KeyValuePair>
 
   {#if hasStakedMaturity(neuron)}
-    <KeyValuePair testId="staked-maturity">
-      <svelte:fragment slot="key">{$i18n.neurons.staked}</svelte:fragment>
+    <div class="details">
```
lmuntaner, line 45: I had to also adapt this file because they both use the same CSS mixin.

And we have the rule of not allowing classes to not be used. Therefore, I had to use the same style here as in NnsNeuronMaturityCard even though it's not yet needed.

frontend/src/tests/lib/components/neuron-detail/NnsNeuronMaturityCard.spec.ts
```diff
@@ -20,6 +21,14 @@ import {
 } from "$tests/mocks/auth.store.mock";
 import en from "$tests/mocks/i18n.mock";
 import { mockFullNeuron, mockNeuron } from "$tests/mocks/neurons.mock";
+import { mockRewardEvent } from "$tests/mocks/nns-reward-event.mock";
+// TODO: Fix prettier command. For some reason it moves the JestPageObjectElement import after the KeyValuePairPo import.
```
dskloetd, line 25: Isn't that expected? Lower case `j` (of "jest.page-object") comes after upper case `K` (of "KeyValuePair.page-object") in ascii.

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
```diff
@@ -119,4 +126,17 @@ describe("NeuronDetail", () => {
 
     await waitFor(() => expect(querySkeleton(container)).not.toBeNull());
   });
+
+  it("should render last maturity distribution", async () => {
+    const { container } = render(NnsNeuronDetail, props);
+
+    await runResolvedPromises();
+
+    const keyValuePairPo = KeyValuePairPo.under({
```
dskloetd, line 135: Can you get it via the NnsNeuronDetailPo?

frontend/src/tests/lib/components/neuron-detail/NnsNeuronMaturityCard.spec.ts
lmuntaner: Yes, the weird part is that we have the same exact order in the other test `frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts`:

```
import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
import { KeyValuePairPo } from "$tests/page-objects/KeyValuePair.page-object";
```

And that one is ok.

I added myself a task to research the difference on both files.

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
lmuntaner: sure, I can add it there

lmuntaner: I'm never sure how much you want me to add as POs or not add because it's not part of the task at hand.

frontend/src/tests/lib/components/neuron-detail/NnsNeuronMaturityCard.spec.ts
lmuntaner: No errors now that I added the new POs

lmuntaner:

dskloetd (approved):

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
lmuntaner: That makes sense, but it's hard to follow sometimes where the test is already set up and you want to just add a new test.

But in general I think that it's a good rule of thumb to create the PO for X, even though it's used only in one test case.

Does that make sense?

PR #2304, opened 2023-04-14
Fix: Add SNS Aggregator URL placeholder in CSP

lmuntaner:
# Motivation

We want wasms that are agnostic to the testnet they will be installed in.

In this PR: I added a missing env var (SNS_AGGREGATOR_URL) used for the CSP rule.

# Changes

* Add placeholder `SNS_AGGREGATOR_URL` in CSP content-src rule.

# Tests

Not applicable

### DIFF
--- a/frontend/scripts/build.csp.mjs
+++ b/frontend/scripts/build.csp.mjs
@@ -8,6 +8,9 @@ import { findHtmlFiles } from "./build.utils.mjs";
 
 dotenv.config();
 
+const aggregatorCanisterUrl = process.env.VITE_AGGREGATOR_CANISTER_URL;
+const isAggregatorCanisterUrlDefined = aggregatorCanisterUrl.length > 0;
+
 const buildCsp = (htmlFile) => {
   // 1. We extract the start script parsed by SvelteKit into the html file
   const indexHTMLWithoutStartScript = extractStartScript(htmlFile);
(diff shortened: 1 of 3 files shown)


lmuntaner:

frontend/scripts/build.csp.mjs
```diff
@@ -210,6 +210,7 @@ const cspConnectSrc = () => {
     "${{HOST}}",
     "${{GOVERNANCE_CANISTER_URL}}",
     "${{LEDGER_CANISTER_URL}}",
+    "${{SNS_AGGREGATOR_URL}}",
```
lmuntaner, line 213: The code that substitutes the values will automatically search and replace this env var because it matches the key passed as argument.

dskloetd (dismissed):
I have no idea how this works but it looks safe.

lmuntaner:
@dskloetd I added a commit with minor changes after your approval. Can you review again?

frontend/scripts/build.csp.mjs
```diff
@@ -210,15 +207,12 @@ const cspConnectSrc = () => {
...
     // TODO: solve with a worker
     // Used for the metrics of OC launch
     "https://2hx64-daaaa-aaaaq-aaana-cai.raw.ic0.app",
   ];
 
-  if (isAggregatorCanisterUrlDefined) {
```
lmuntaner, line 218: We don't rely in the connect-src rule from the build step. So it's safe to remove this part.

dskloetd: Now I no longer feel qualified to approve these changes unless I understand them.
Shall I study how all of this works or is there someone who has reviewed this file previously who is in a better position to review?

lmuntaner: It makes sense. I'll ask @mstrasinskis .

I can explain them to you on Monday. Thanks!

mstrasinskis (approved):
LGTM, thanks!

PR #2306, opened 2023-04-14
Wait for aggregator

bitdivine:
# Motivation
Apparently e2e tests are flaking out because the aggregator doesn't have enough data when the e2e tests run.

# Changes
* Configure the aggregator collect data quickly.
  * Note: The local replica has an artificial delay to make timing realistic.  In the next dfx that delay is configurable, so it can be made short for fast CI tests or it can be made long to simulate usage in a stressed environment.  This would help here as well, once we upgrade.
*  Wait for the aggregator to have 12 SNS before running e2e tests

# Tests
See CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -111,10 +111,10 @@ jobs:
           npm ci
       - name: Wait for aggregator to get all SNSs
         run: |
-          for (( i=300; i>0; i-- )); do
-             num_sns="$(for ((i=0; i<3; i++)) ; do { curl -fsS "http://$(dfx canister id sns_aggregator).localhost:8080/v1/sns/list/page/$i/slow.json" | jq length ; } 2>/dev/null ; done | awk '{i+=$1}END{print i}')"
+          for (( try=300; try>0; try-- )); do
+             num_sns="$(for ((page=0; page<3; page++)) ; do { curl -fsS "http://$(dfx canister id sns_aggregator).localhost:8080/v1/sns/list/page/$page/slow.json" | jq length ; } 2>/dev/null ; done | awk '{i+=$1}END{print i}')"
              (( num_sns < 12)) || break
-             printf "\r #SNS: % 4d   Tries remaining: 4d" "$num_sns" "$i"
+             printf "\r #SNS: % 4d   Tries remaining: 4d" "$num_sns" "$try"
              sleep 1
           done
       - name: Run Playwright end-to-end tests
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -111,6 +111,7 @@ jobs:
           npm ci
       - name: Wait for aggregator to get all SNSs
         run: |
+          set +x
           for (( try=300; try>0; try-- )); do
              num_sns="$(for ((page=0; page<3; page++)) ; do { curl -fsS "http://$(dfx canister id sns_aggregator).localhost:8080/v1/sns/list/page/$page/slow.json" | jq length ; } 2>/dev/null || true ; done | awk '{i+=$1}END{print i}')"
              (( num_sns < 12)) || break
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           DFX_NETWORK=local ./config.sh
           echo "Install:"
           dfx canister install nns-dapp --wasm nns-dapp.wasm --upgrade-unchanged --mode reinstall --yes --argument "$(cat nns-dapp-arg.did)"
-          dfx canister install sns_aggregator --wasm sns_aggregator.wasm --upgrade-unchanged --mode reinstall --yes
+          dfx canister install sns_aggregator --wasm sns_aggregator.wasm --upgrade-unchanged --mode reinstall --yes --argument '(opt record { update_interval_ms = 100; fast_interval_ms = 100; })'
```
dskloetd: I found this [commit](https://github.com/dfinity/nns-dapp/commit/2206d61fbe1affbcb7915f5a0c31b604f9c26222), where you disabled the fast mode.
With [this run](https://github.com/dfinity/nns-dapp/actions/runs/4703871451/jobs/8342897927),   where it runs the full 10 minutes (300 * 2 seconds) and still only loaded 6 SNSes before continuing.
Given that today, the tests are only flaky and don't fail 100% of the time, this suggests to me that the tests are able to pass without all the SNSes being present according to that check.
Unless that run was much slower than a normal run is today without these changes?

dskloetd (commented):
Thanks!
I'm approving because it's good to confirm whether this will help or not.
But see my comment. If I understand correctly, this is either not working or it can't be the cause of the flakiness. Do you agree?

.github/workflows/build.yml
dskloetd: I was having another look at the logs of failed and passed e2e tests and I think when the test passed, it was because connecting to the aggregator failed and it used the failback:
```
dapp.a5d3df61.js:11[13:05:20.818] Loading SNS projects from aggregator canister...
start.8fd3597e.js:0Refused to connect to 'http://qvhpv-4qaaa-aaaaa-aaagq-cai.localhost:8080/v1/sns/list/latest/slow.json' because it violates the following Content Security Policy directive: "connect-src 'self' http://qhbym-qaaaa-aaaaa-aaafq-cai.localhost:8080 https://identity.ic0.app https://nns.ic0.app http://qsgjb-riaaa-aaaaa-aaaga-cai.localhost:8080 http://localhost:8080 http://rrkah-fqaaa-aaaaa-aaaaq-cai.localhost:8080 http://ryjl3-tyaaa-aaaaa-aaaba-cai.localhost:8080 https://2hx64-daaaa-aaaaq-aaana-cai.raw.ic0.app".
start.8fd3597e.js:0Refused to connect to 'http://qvhpv-4qaaa-aaaaa-aaagq-cai.localhost:8080/v1/sns/list/latest/slow.json' because it violates the document's Content Security Policy.
dapp.a5d3df61.js:11[13:05:20.846] Listing all deployed Sns summaries certified:false call...
dapp.a5d3df61.js:11[13:05:20.852] Listing all deployed Sns swap states certified:false call...
...
```

 And when the test failed it was because connecting to the aggregator worked but returned only 1 project, which wasn't the project that had tokens sent to its faucet:
 ```
 dapp.a5d3df61.js:11[14:00:49.632] Loading SNS projects from aggregator canister...
 dapp.a5d3df61.js:11[14:00:49.773] Loading SNS projects from aggregator canister completed
 ```

As far as I can tell it's trying to connect to the same URL in both cases: `http://qvhpv-4qaaa-aaaaa-aaagq-cai.localhost:8080/v1/sns/list/latest/slow.json`

I don't understand why it sometimes works and sometimes doesn't. But at least this explains why the test was flaky while the aggregator was always slow.

dskloetd (approved):

.github/workflows/build.yml
dskloetd: This may not actually help (given the current setup) because
1. Only the first of 12 projects has tokens in the faucet.
2. We assume that project will be listed first on the page.
3. We use /latests/ instead of paging when using the aggregator, which only includes the latest 10 projects and therefore excludes the project with tokens in the faucet.

I've opened a ticket to continue the investigation: https://dfinity.atlassian.net/browse/GIX-1507
I will put further updates only there and not here.

bitdivine: At least, this PR ensures that the aggregator is always in the same state when the e2e tests are run.  Whether that state is the desired state and whether the tests make appropriate assumptions, we can look into, but at least we won't have flakiness caused by the aggregator.

PR #2309, opened 2023-04-14
test: Get aggregator args & compare with deployment

bitdivine:
# Motivation
There are no tests for sns canister configuration.

# Changes
- Add a query method to get the canister config.
- Add a test to verify that the config is as specified.
  - The test passes for configuration set with `reconfigure(..)` and fails for configuration set with `dfx canister install()`.  Amusingly, reconfigure() is a custom method, seemed risky, and was exercised in tests extensively.  It is the safe standard method causing trouble.
  - Examination shows that the arguments are not passed to the canister at all, so the issue is somehow related to the way dfx tries to send the arguments.

# TODO
- Find out why dfx is not passing the argument & fix.
- Check that when canister arguments are received, they also take effect.  E.g. that if a certain update frequency is specified in the config, the `sns_aggregator` really does get data with that frequency.

# Tests
- A test is included in this PR.


.github/workflows/aggregator.yaml
bitdivine, line 1: Here we test that the provided config is actually set.  Hint: It isn't.

rs/sns_aggregator/sns_aggregator.did
bitdivine, line 1: This is the declaration of the new method to get the config.

rs/sns_aggregator/src/lib.rs
bitdivine, line 1: A method to get the config, so that we can compare it with the value provided when the canister was deployed or reconfigured.

scripts/sns/aggregator/test-config
bitdivine, line 1: Here we provide the config in various ways and check that the config has been applied.  It isn't, except when the config is provided by `reconfigure(config)`, so the other cases are disabled until the problem is fixed.

scripts/sns/aggregator/test-config
```diff
@@ -0,0 +1,75 @@
...
+  echo "Check passes"
+}
+
+(
+  printf "\n\n\n=========================================\n"
+  echo "Checking direct reinstallation"
+  EXPECTED_CONFIG='record { update_interval_ms = 1345; fast_interval_ms = 1_123_456_789; }'
+  dfx canister install sns_aggregator --wasm ./sns_aggregator.wasm --mode reinstall --argument "$(didc encode "( opt $EXPECTED_CONFIG)")" --argument-type raw --yes
+  check
+) || echo 'TODO: The argument is not provided to init(...).  Find out why and fix.'
```
dskloetd, line 36: Something I personally like to do for tests that don't yet work the way I want: Make the test fail when it starts behaving the way I want. That way I can't forget to actually make the test the way I want once the problem is fixed.

.github/workflows/aggregator.yaml
```diff
@@ -52,6 +52,9 @@ jobs:
           cp out/sns_aggregator_dev.wasm sns_aggregator.wasm
           # Note: Pretty much any argument values will do; the canister will be reconfigured.
           dfx canister install sns_aggregator --wasm sns_aggregator.wasm --upgrade-unchanged '(opt record { update_interval_ms = 1000; fast_interval_ms = 1_000_000_000; })'
+          dfx canister call sns_aggregator get_canister_config
```
dskloetd, line 55: Why is this call here? It doesn't look like we do anything with the result.

dskloetd: Can you address this?

dskloetd (approved):

PR #2310, opened 2023-04-15
Always show transaction network on review step

peterpeterparker:
# Motivation

Always show transaction network on review step (not only on ckBTC)

# Changes

- simplify enum to a single ICP network (not ICP ckbtc and ITCP cktestbtc)
- always display network on transaction review / description

# Screenshot

Capture d’écran 2023-04-15 à 08 28 58


frontend/src/lib/components/transaction/TransactionDescription.svelte
```diff
@@ -1,7 +1,6 @@
 <script lang="ts">
   import { i18n } from "$lib/stores/i18n";
-  import { nonNullish } from "@dfinity/utils";
-  import type { TransactionNetwork } from "$lib/types/transaction";
+  import { TransactionNetwork } from "$lib/types/transaction";
 
   export let destinationAddress: string;
   export let selectedNetwork: TransactionNetwork | undefined = undefined;
```
lmuntaner, line 6: How about setting default selectedNetwork `TransactionNetwork.ICP`. Would that be more accurate to what we want?

lmuntaner (approved):
Minor comment, already approving. Thanks!

frontend/src/lib/components/transaction/TransactionDescription.svelte
peterpeterparker: That would be more accurate for this component but, would need other changes in the all transaction flow.
The selected network dropdown field is only displayed on ckBTC transaction form, therefore undefined or a value.
To keep this PR slim I rather like to inherit that way here. We can review the value in the all flow afterwards if you wish.

lmuntaner: Ok, thanks for the explanation

PR #2315, opened 2023-04-16
Lint: Forbid unreachable code

dskloetd:
# Motivation

Unreachable code is usually not intended.
Not allowing it helps to avoid mistakes.

# Changes

1. Configure eslint to forbid unreachable code.
2. Fix the place where we had unreachable code.

# Tests

CI


lmuntaner (approved):
I agree, thanks!

PR #2319, opened 2023-04-17
Check wasm exports

bitdivine:
# Motivation
Observed: A cargo build failing due to a missing dependency without exiting with a non-zero code.

The created wasm was much smaller than expected and exported only one function (main) instead of all the functions in nns-dapp.  It won't pass e2e tests, but tracking down _why_ can waste time.

# Changes
- Check the list of functions exported by the nns-dapp & compare it with a list.
  - Note: `nm` doesn't seem to support wasm, so this uses `wasm-nm`.  Developers who build the backend will have to install it.

# TODO
- Instead of a fixed list, can we parse the nns-dapp .did file?
- Make a similar test for the sns_aggregator.

# Tests
* See CI. Manually re-ran the docker build and got the error condition.


rs/backend/nns-dapp-exports.txt
```diff
@@ -0,0 +1,19 @@
+canister_heartbeat
```
lmuntaner, line 1: If we add one I guess we'll need to add it here, is this correct?

lmuntaner (approved):
LGTM

PR #2322, opened 2023-04-17
Script for prebuild

bitdivine:
# Motivation
The command line to pre-build

# Changes

# Tests


bitdivine:
@dskloetd - A few days ago you were asking about putting the pre-build one-liner on multiple lines.  This is the PR I mentioned then that does the same in a general way.  I revisited it and updated it but I still don't think this is worthwhile.  Breaking the existing docker line over several lines, but keeping them in one docker step, could still be valuable.

PR #2323, opened 2023-04-17
Remove dummy wasm

bitdivine:
# Motivation
In order to cache the dependency building, the rust source code is replaced by dummy code.  However there is then a risk that the dummy wasm is mistaken for the final output.

# Changes
- Remove wasms after pre-compiling libraries.
- Touch the source code, as cargo decides whether to rebuild based on timestamps.

# Tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -110,6 +110,16 @@ COPY ./Cargo.lock /build/
 COPY ./dfx.json /build/
 COPY --from=build_frontend /build/assets.tar.xz /build/
 WORKDIR /build
+# We need to make sure that the rebuild happens if the code has changed.
+# - Docker checks whether the filesystem or command line have changed, so it will
+#   run if there are code changes and skip otherwise.  Perfect.
+# - However cargo _may_ then look at the mtime and decide that no, or only minimal,
+#   rebuilding is necessary due to the potentially recent dependency building step above.
+#   Cargo checks whether the mtime of some code is newer than its last build, like
+#   it's 1974, unlike bazel that uses checksums.
+# So we update the timestamps of the root code files.
+# Old canisters use src/main.rs, new ones use src/lib.rs.  We update the timestamps on all that exist.
+# We don't wish to update the code from main.rs to lib.rs and then have builds break.
 RUN touch --no-create rs/backend/src/main.rs rs/backend/src/lib.rs
 RUN ./build-backend.sh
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -110,6 +110,7 @@ COPY ./Cargo.lock /build/
 COPY ./dfx.json /build/
 COPY --from=build_frontend /build/assets.tar.xz /build/
 WORKDIR /build
+RUN touch --no-create rs/backend/src/main.rs rs/backend/src/lib.rs
```
lmuntaner, line 113: This forces to rebuild?

It would be good to have a comment to understand why we're creating these files here

bitdivine: This updates the timestamps of these files (if they exist, it doesn't create any that don't exist).  Cargo compares the timestamps of the build files to determine whether any need to be rebuilt; specifically it checks the mtime.  I'll add a comment with details. Given that the target wasm has been removed, there will be a rebuild, but I want to make sure that there are no cached intermediate stages with modification timestamps newer than the source files.  The mtime of the source that has been copied in _may_ be the mtime in the host system, in which case it may be older than the dependency build step.  This touch here makes 100% sure that this doesn't happen.  Using timestamps is 40+ year old technology and should really have been replaced with checksums.  This is one way in which bazel is better than cargo.

bitdivine: A further note: See that caching works as intended.  No changes yield a sub-second build, a trivial change, appending a  line to a file, takes 33 seconds.  This is with a beefy laptop, it will be at least 4 times slower on CI.
```
max@sinkpad:~/dfn/nns-dapp/branches/rm-prebuild-wasms (10:46:52)$ time chronic ./scripts/docker-build 

real	0m0,364s
user	0m0,242s
sys	0m0,082s
max@sinkpad:~/dfn/nns-dapp/branches/rm-prebuild-wasms (10:47:20)$ echo >> rs/backend/src/main.rs 
max@sinkpad:~/dfn/nns-dapp/branches/rm-prebuild-wasms (10:47:38)$ time chronic ./scripts/docker-build 

real	0m33,056s
user	0m0,293s
sys	0m0,112s
max@sinkpad:~/dfn/nns-dapp/branches/rm-prebuild-wasms (10:48:13)$ 
```

lmuntaner (approved):

PR #2328, opened 2023-04-18
Fix confirm cycles overflow style

peterpeterparker:
# Notes

Just a patch. Someday we should take the time to review the all canister creation flow and integrate the `Transaction` components.

# Screenshots

Is:

Capture d’écran 2023-04-18 à 07 11 50

After:

Capture d’écran 2023-04-18 à 07 19 06
Capture d’écran 2023-04-18 à 07 18 54


lmuntaner (approved):
I agree, thanks for the fix!

PR #2329, opened 2023-04-18
Feat: Use default config for testnets

lmuntaner:
# Motivation

Simplify dfx.json

# Changes

* Use default config for `FETCH_ROOT_KEY` and `FEATURE_FLAGS` in most testnets.

# Tests

Not applicable.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -169,7 +169,6 @@
     },
     "staging": {
       "config": {
-        "FETCH_ROOT_KEY": true,
         "HOST": "https://nnsdapp.testnet.dfinity.network",
         "FEATURE_FLAGS": {
           "ENABLE_SNS_2": false,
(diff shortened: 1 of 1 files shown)


lmuntaner:

dfx.json
```diff
@@ -362,183 +289,79 @@
     },
     "https___benchmarklarge_testnet_dfinity_network": {
       "config": {
-        "FETCH_ROOT_KEY": true,
-        "HOST": "https://benchmarklarge.testnet.dfinity.network",
-        "FEATURE_FLAGS": {
-          "ENABLE_SNS_2": true,
-          "ENABLE_SNS_VOTING": true,
-          "ENABLE_SNS_AGGREGATOR": true,
-          "ENABLE_CKBTC": false,
```
dskloetd, line 371: This is not the default value so you need to override it, right?

dfx.json
```diff
@@ -362,183 +289,79 @@
...
     },
     "https___large01_testnet_dfinity_network": {
       "config": {
-        "FETCH_ROOT_KEY": true,
-        "HOST": "https://large01.testnet.dfinity.network",
-        "FEATURE_FLAGS": {
-          "ENABLE_SNS_2": true,
-          "ENABLE_SNS_VOTING": true,
-          "ENABLE_SNS_AGGREGATOR": true,
-          "ENABLE_CKBTC": false,
```
dskloetd, line 385: This is not the default value so you need to override it, right?
And many more below. Please check that you only remove values that are the default value.

lmuntaner: Ditto

dskloetd:
This changes lots of values. I think it would be better to be explicit about all the changes that this causes.
But I never deploy testnets so if someone else thinks this is fine, I'm happy to stand back while someone else approves.
But personally I would separate changing the configurations from using the default configuration.

lmuntaner:
@mstrasinskis please review

mstrasinskis (approved):
LGTM, thanks!

bitdivine (approved):
Perfect, thank you!

PR #2340, opened 2023-04-19
E2E: Remove expect timeout

dskloetd:
# Motivation

Creating a neuron can take longer than 10 seconds.
The test expects the modal to be closed within 10 seconds and sometimes [times out](https://github.com/dfinity/nns-dapp/actions/runs/4744468055).

# Changes

Remove the 10 second timeout for expect calls.
Rely on the 60 second timeout for the entire test.

# Tests

`npm run test-e2e`

### DIFF
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -13,8 +13,9 @@ export default defineConfig({
     /**
      * Maximum time expect() should wait for the condition to be met.
      * For example in `await expect(locator).toHaveText();`
+     * 0 for no limit.
      */
-    timeout: 10000,
+    timeout: 0,
   },
   /* Run tests in files in parallel */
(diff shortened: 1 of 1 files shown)


lmuntaner (commented):
Does 0 disable the "local" timeout and relies on the test timeout?

dskloetd:
It's the `expect` timeout that's disabled. The test timeout is 60 seconds.
This is the code around the change:
```
  /* Maximum time one test can run for. */
  timeout: 60 * 1000,
  expect: {
    /**
     * Maximum time expect() should wait for the condition to be met.
     * For example in `await expect(locator).toHaveText();`
     * 0 for no limit.
     */
    timeout: 0,
  },
  ```

bitdivine (approved):

PR #2344, opened 2023-04-19
Nightly

bitdivine:
# Motivation

# Changes

# Tests


.github/workflows/nightly.yaml
```diff
@@ -10,6 +10,25 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 jobs:
```
bitdivine, line 12: This is the main part of this PR: The code that tags head of main as nightly-YYYY-mm-dd.

.github/workflows/docker-build.yaml
```diff
@@ -52,16 +52,18 @@ jobs:
       - name: Release
         run: |
           cd "${{ matrix.BUILD_NAME }}-out"
-          artefacts=(nns-dapp.wasm sns_aggregator.wasm sns_aggregator_dev.wasm assets.tar.xz)
+          daily_build_name="nns-dapp-$(git rev-parse HEAD).wasm"
+          cp nns-dapp.wasm "$daily_build_name"
+          artefacts=(nns-dapp.wasm sns_aggregator.wasm sns_aggregator_dev.wasm assets.tar.xz "$daily_build_name")
```
bitdivine, line 57: This makes a copy of the nns-dapp wasm with the git commit in the name. A copy, so that any existing code that uses the name without the hash doesn't break.  Hunting for such uses and updating them, if appropriate, can be the topic of another PR.

.github/workflows/docker-build.yaml
```diff
@@ -52,16 +52,18 @@ jobs:
...
           for tag in $(git tag --points-at HEAD) ; do
             : Creates or updates a release for the tag
             if gh release view "$tag"
             then gh release upload --repo dfinity/nns-dapp --clobber "$tag" "${artefacts[@]}" || true
             else gh release create --title "Release for tags/$tag" --draft --notes "Build artefacts from tag: $tag" "$tag" "${artefacts[@]}"
             fi
-            : If the tag is for a proposal, make it public
-            [[ "$tag" != proposal-* ]] || { echo "Making release public" ; gh release edit "$tag" --draft=false ; }
+            : If the tag is for a proposal or nightly, make it public
+            [[ "$tag" != proposal-* ]] && [[ "$tag" != nightly-* ]] || { echo "Making release public" ; gh release edit "$tag" --draft=false ; }
```
bitdivine, line 66: If the release is not public, assets attached to the release will not be publicly accessible.

dfx.json
```diff
@@ -29,7 +29,7 @@
       "type": "custom",
```
bitdivine, line 29: This is added by CI - it thinks it is running the nightly job so it is trying to make a PR to update II.  I will switch to a different branch and remove it.  That means making another PR and another approval.  I'd like to get feedback here first, though.

PR #2346, opened 2023-04-19
Nightly tag

bitdivine:
# Motivation
We wish to test a recent build (not prod) against the rest of the IC test environment.

## Requirements
* Nightly release
* No need to make a new release if there is an existing release
* The nns-dapp wasm should be available with the commit in the filename.

# Changes
* To the nightly workflow, add a job that tags the head of main with `nightly-YYY-mm-dd`
* In the docker-build job that creates releases:
  * Make a copy of the wasm that includes the commit in the name and include it in the release artefacts.
  * Make nightly releases public, so that the wasms are accessible

# Tests
* I have triggered the workflow manually by pushing to the develop branch.  It is hard to get workflows to get tags, so I have not been able to confirm that no new releases are created if there are no changes.  I have been able to confirm that the nightly job creates tags and that the tags trigger workflows.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -62,8 +62,8 @@ jobs:
             then gh release upload --repo dfinity/nns-dapp --clobber "$tag" "${artefacts[@]}" || true
             else gh release create --title "Release for tags/$tag" --draft --notes "Build artefacts from tag: $tag" "$tag" "${artefacts[@]}"
             fi
-            : If the tag is for a proposal, make it public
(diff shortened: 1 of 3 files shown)


mraszyk (approved):
Thank you!

lmuntaner (commented):
Let me see if I understand the flow of the jobs.

- Nightly job adds the `nightly-*` tag.
- Github triggers docker-build job when tags change.
- Docker-build job publishes the wasm with commit

- Person pushes `proposal-*` tag.
- Github triggers docker-build job when tags change.
- Docker-build job publishes the wasm with commit.

Is this correct?

PR #2347, opened 2023-04-20
E2E: Check SNS neuron detail

dskloetd:
# Motivation

Continue with SNS end-to-end test.

# Changes

1. Check the neuron detail in the e2e test.
2. Make the SnsNeuronDetailsPo accessible via AppPo.
3. Add convenience getStake method to SnsNeuronDetailsPo.
4. Conveniently trim the result of getTitle() in SnsNeuronDetailsPo.
5. Update existing `frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts` to use the new convenience methods as well.
6. Change relative imports that break Playwright to absolute imports.

# Tests

`npm run test e2e -- sns`
`npm run test`
`npm run check`

### DIFF
--- a/frontend/src/tests/e2e/sns-governance.spec.ts
+++ b/frontend/src/tests/e2e/sns-governance.spec.ts
@@ -40,7 +40,9 @@ test("Test SNS governance", async ({ page, context }) => {
     `You have no ${snsProjectName} neurons. Stake a neuron to vote on proposals for ${snsProjectName}.`
   );
 
-  await appPo.getNeuronsPo().getSnsNeuronsFooterPo().stakeNeuron(5);
+  const stake = 5;
+  const formattedStake = "5.00";
+  await appPo.getNeuronsPo().getSnsNeuronsFooterPo().stakeNeuron(stake);
 
   // SN001: User can see the list of neurons
   const neuronCards = await appPo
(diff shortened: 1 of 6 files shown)


frontend/src/tests/e2e/sns-governance.spec.ts
```diff
@@ -40,17 +40,24 @@ test("Test SNS governance", async ({ page, context }) => {
...
+  const neuronCard = neuronCards[0];
+  expect(await neuronCard.getStake()).toEqual(formattedStake);
 
   // SN002: User can see the details of a neuron
+  await neuronCard.click();
+  const neuronDetail = appPo.getNeuronDetailPo().getSnsNeuronDetailPo();
+  expect(await neuronDetail.getTitle()).toBe(snsProjectName);
+  expect(await neuronDetail.getStake()).toBe(formattedStake);
 
   // SN003: User can add a hotkey
```
lmuntaner, line 62: Can we next try to work the participation flow?

That flow will also help us setting up testnets to check our release candidates and develop with open SNSes

lmuntaner (approved):
Thanks!

As I mentioned, could we move to the e2e for participation.

A few new SNSes are coming and it's important we don't mess up the participation in any of the upcoming changes.

frontend/src/tests/e2e/sns-governance.spec.ts
dskloetd: Sure, I'll move to participation now.

PR #2348, opened 2023-04-20
Remove unnecessary steps from checks.yml workflow

dskloetd:
# Motivation

Cleaning up.

# Changes

1. `npm run test` and `npm run test` don't need `dfx` or `.env` so remove those steps from the `svelte-tests` and `svelte-lint` jobs.
2. Remove comment about temporarily disabling lint. It's already no longer disabled.
3. Add default `bash -euxlo pipefail {0}` to fail in a more predictable way and log what happens.

# Tests

CI


bitdivine (approved):
LGTM.  Thanks!

PR #2351, opened 2023-04-20
Fix: Deploy script also creates args for network

lmuntaner:
# Motivation

Current deploy.sh is confusing because you might set network to staging, but use arguments for medium09.

# Changes

* Add command to create config for the selected network within `./deploy.sh` script.

# Tests

I created the arguments for medium09, then deployed with `./deploy.sh --nns-dapp staging`. All good.


lmuntaner:

dskloetd (approved):

PR #2352, opened 2023-04-21
E2E: Wait for SNS neurons to be loaded.

dskloetd:
# Motivation

Avoid flakiness in the end-to-end test.
Previously I assumed that when the stake neuron modal is closed, the neurons are already loaded.
But I noticed at least once that the test failed because the neurons weren't loaded.

# Changes

Wait for sns neurons to be loaded before checking that the neuron cards are there.

# Tests

`npm run test-e2e`


lmuntaner (approved):
Thanks!

PR #2354, opened 2023-04-21
Chore: Add TVL canister id to config

lmuntaner:
# Motivation

NNS Dapp testnet tries to get data from mainnet. Because of that, errors are thrown in the console on testnet.

With errors in the console is hard to differentiate real errors and ok errors.

The data is from the TVL canister. We also might want to deploy a TVL canister to a testnet and check that all works as expected.

Therefore, this PR adds the TVL canister to the configuration.

# Changes

* Add TVL_CANISTER_ID in `config.sh`.
* Add `tvlCanisterId` key to envVars exposed in frontend
* Add parameter of canisterId to api `queryTVL` and use the `HOST` env var instead of mainnet host.
* Use env var `tvlCanisterId` in environment variable `TVL_CANISTER_ID`.
* Check that TVL_CANISTER_ID is present in service `queryTVL`.
* Convert empty string ot undefined also the env vars from HTML. Use a new helper `mapEmtpyStringsToUndefined` for that.
* Add `tvlCanisterId` to envVars from HTML and build step.

# Tests

* Add TVL_CANISTER_ID to test for `config.sh`
* Fix broken tests after changing tvl api and service
* Add `tvlCanisterId` key to jest-setup

### DIFF
--- a/frontend/src/lib/api/tvl.api.cjs.ts
+++ b/frontend/src/lib/api/tvl.api.cjs.ts
@@ -7,24 +7,29 @@ import type { Principal } from "@dfinity/principal";
 /**
  * HTTP-Agent explicit CJS import for compatibility with web worker - avoid Error [RollupError]: Unexpected token (Note that you need plugins to import files that are not JavaScript)
  */
+import { TVL_CANISTER_ID } from "$lib/constants/canister-ids.constants";
 import { HttpAgent } from "@dfinity/agent/lib/cjs/index";
+import { isNullish } from "@dfinity/utils";
 
 export const queryTVL = async ({
   identity,
   certified,
-  canisterId,
 }: {
   identity: Identity;
   certified: boolean;
-  canisterId: Principal;
-}): Promise<TvlResult> => {
-  logWithTimestamp(`Getting canister ${canisterId.toText()} TVL call...`);
+}): Promise<TvlResult | undefined> => {
+  if (isNullish(TVL_CANISTER_ID)) {
+    return undefined;
+  }
+  logWithTimestamp(`Getting canister ${TVL_CANISTER_ID.toText()} TVL call...`);
 
-  const { getTVL } = await canister({ identity, canisterId });
+  const { getTVL } = await canister({ identity, canisterId: TVL_CANISTER_ID });
 
   const result = getTVL({ certified });
 
-  logWithTimestamp(`Getting canister ${canisterId.toText()} TVL complete.`);
+  logWithTimestamp(
+    `Getting canister ${TVL_CANISTER_ID.toText()} TVL complete.`
+  );
 
   return result;
 };
--- a/frontend/src/lib/api/tvl.api.cjs.ts
+++ b/frontend/src/lib/api/tvl.api.cjs.ts
@@ -1,7 +1,6 @@
 import { TVLCanister } from "$lib/canisters/tvl/tvl.canister";
 import type { TvlResult } from "$lib/canisters/tvl/tvl.types";
-import { TVL_CANISTER_ID } from "$lib/constants/canister-ids.constants";
-import { HOST_IC0_APP } from "$lib/constants/environment.constants";
+import { HOST } from "$lib/constants/environment.constants";
 import { logWithTimestamp } from "$lib/utils/dev.utils";
 import type { Identity } from "@dfinity/agent";
 import type { Principal } from "@dfinity/principal";
@@ -13,19 +12,19 @@ import { HttpAgent } from "@dfinity/agent/lib/cjs/index";
 export const queryTVL = async ({
   identity,
   certified,
+  canisterId,
 }: {
   identity: Identity;
   certified: boolean;
+  canisterId: Principal;
 }): Promise<TvlResult> => {
-  logWithTimestamp(`Getting canister ${TVL_CANISTER_ID.toText()} TVL call...`);
+  logWithTimestamp(`Getting canister ${canisterId.toText()} TVL call...`);
 
-  const { getTVL } = await canister({ identity, canisterId: TVL_CANISTER_ID });
+  const { getTVL } = await canister({ identity, canisterId });
 
   const result = getTVL({ certified });
 
-  logWithTimestamp(
-    `Getting canister ${TVL_CANISTER_ID.toText()} TVL complete.`
-  );
+  logWithTimestamp(`Getting canister ${canisterId.toText()} TVL complete.`);
 
   return result;
 };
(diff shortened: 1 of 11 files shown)


lmuntaner:
I tested deploying to staging:

No canister id:
URL: https://my6v6-kyaaa-aaaaa-aacha-cai.nnsdapp.testnet.dfinity.network/
As expected, no errors in the console.

Defined canister id (alas wrong because we don't have TVL in staging):
URL: https://mr56c-4qaaa-aaaaa-aacgq-cai.nnsdapp.testnet.dfinity.network/
Code makes the call but gets an error because the canister doesn't have the endpoint.

Screenshot 2023-04-21 at 14 23 18

lmuntaner:
@peterpeterparker @bitdivine @dskloetd please review

frontend/src/lib/utils/env-vars.utils.ts
```diff
@@ -59,14 +75,12 @@ const getHtmlEnvVars = (): EnvironmentVars => {
       `Missing environment variables element with selector ${ENV_VARS_ELEMENT_SELECTOR} in HTML page.`
     );
   }
-  const envVars: Record<string, unknown> = dataElement.dataset ?? {};
+  const envVars: Record<string, string | undefined> =
+    mapEmtpyStringsToUndefined(dataElement.dataset ?? {});
```
dskloetd, line 79: Were empty strings converted to undefined in a different way before?
Or were they not converted?
Or did we not have empty strings before?
Is TVL_CANISTER_ID provided as empty string rather than being absent, when it's not available?

frontend/src/lib/utils/env-vars.utils.ts
```diff
@@ -49,6 +51,20 @@ function assertEnvVars(
...
+const convertEmtpyStringToUndefined = (
+  str: string | undefined
+): string | undefined => (str === "" ? undefined : str);
+
+const mapEmtpyStringsToUndefined = (obj: {
+  [key: string]: string | undefined;
+}): { [key: string]: string | undefined } => {
+  const result: { [key: string]: string | undefined } = {};
+  Object.keys(obj).forEach((key: string) => {
+    result[key] = convertEmtpyStringToUndefined(obj[key] as string);
```
dskloetd, line 63: Why is `obj[key]` cast to a `string`? According to the function signature, it can be `undefined` and you changed `convertEmtpyStringToUndefined` to accept `undefined` so this seems unnecessary.

frontend/src/lib/services/$public/tvl.service.ts
```diff
@@ -1,15 +1,21 @@
 import { queryTVL as queryTVLApi } from "$lib/api/tvl.api.cjs";
 import type { TvlResult } from "$lib/canisters/tvl/tvl.types";
+import { TVL_CANISTER_ID } from "$lib/constants/canister-ids.constants";
 import { AnonymousIdentity } from "@dfinity/agent";
+import { isNullish } from "@dfinity/utils";
 
 export const queryTVL = async (): Promise<TvlResult | undefined> => {
+  if (isNullish(TVL_CANISTER_ID)) {
```
dskloetd, line 8: Do you want to add a test for this behavior?

frontend/src/lib/api/tvl.api.cjs.ts
```diff
@@ -39,7 +38,7 @@ const canister = async ({
 }): Promise<TVLCanister> => {
   const agent = new HttpAgent({
     identity,
-    host: HOST_IC0_APP,
+    host: HOST,
```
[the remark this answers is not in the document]
lmuntaner: Before it was hardcoded to always point to mainnet. Also because the canister id was only the one in mainnet.

Now we'll enable this to be also testnet

frontend/src/lib/api/tvl.api.cjs.ts
```diff
@@ -13,19 +12,19 @@ import { HttpAgent } from "@dfinity/agent/lib/cjs/index";
 export const queryTVL = async ({
   identity,
   certified,
+  canisterId,
```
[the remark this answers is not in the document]
lmuntaner: Any canister that implements the interface of TVL would be fine.

My initial intention was to manage this in the service to keep the api simple.

But yes, maybe this logic could be in the API and it would return `undefined` if canister is not set.

frontend/src/lib/utils/env-vars.utils.ts
lmuntaner: 1. No
2. Only for build env vars, not html env vars converted empty strings to undefined. Each of the build env vars is wrapped with `convertEmtpyStringToUndefined`
3. We didn't have empty string in html env vars that's why we didn't need it.
4. Yes, the default value of env vars is empty string. They are set as `<meta ... data-tvl-canister-id ... >` which in the `element.dataSet.tvlCanisterId` is read as `""`. Not as `undefined`.

frontend/src/lib/utils/env-vars.utils.ts
lmuntaner: No need to convert, I did that at the beginning and then changed it. I forgot to change this part.

frontend/src/lib/services/$public/tvl.service.ts
lmuntaner: 1. I'll move this behavior to the queryTVL api.
2. I'll try it, but the canister id is mocked in the jest-setup. I believe that will be a bigger change than intended in the PR. Maybe later now that we have a use case.

lmuntaner: I tried the test and I would need to change the way we mock the canister ids. Maybe in another PR.

lmuntaner:

frontend/src/lib/utils/env-vars.utils.ts
dskloetd: We have other optional values, right? I'm confused why this is the first one that can be the empty string.

frontend/src/lib/services/$public/tvl.service.ts
dskloetd: Tests for new behavior should be in the same PR as the new behavior.
If you think this PR is already big (I agree, but originally decided not to say anything and review it anyway) I suggest separating making the variable available from using it.

I think concretely that would mean merging this PR just without the changes to `frontend/src/lib/api/tvl.api.cjs.ts` and including those changes in the PR with the corresponding test.

frontend/src/lib/utils/env-vars.utils.ts
lmuntaner: Yes we did, but we never tested them as such apparently.

frontend/src/lib/services/$public/tvl.service.ts
lmuntaner: done

lmuntaner:

dskloetd (approved):

PR #2355, opened 2023-04-21
Chore: Use parent data-tid in KeyValuePairInfo

lmuntaner:
# Motivation

The page object of the KeyValuePairInfo was relying in the parent to get the description because the data-tid was not added at the root.

The PR in GIX Components: https://github.com/dfinity/gix-components/pull/199 sets the test id of KeyValuePairInfo in the root component.

# Changes

* Upgrade gix-components library.
* Change the way KeyValuePairInfoPo accesses the description text. There is no need to keep a reference to the parent.

# Tests

Changes only in test implementations, not no test changes.

### DIFF
--- a/frontend/src/tests/page-objects/KeyValuePairInfo.page-object.ts
+++ b/frontend/src/tests/page-objects/KeyValuePairInfo.page-object.ts
@@ -2,28 +2,16 @@ import { BasePageObject } from "$tests/page-objects/base.page-object";
 import type { PageObjectElement } from "$tests/types/page-object.types";
 
 export class KeyValuePairInfoPo extends BasePageObject {
-  private parent: PageObjectElement;
   private testId: string;
 
-  /**
-   * KeyValuePairInfo renders two components:
-   * - a div with the key value pair and a button
-   * - a div with the description information
-   *
-   * That's why we need to keep a reference to the parent element
-   * to be able to get the description information
-   */
   constructor({
-    parent,
     element,
     testId,
   }: {
     element: PageObjectElement;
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/page-objects/KeyValuePairInfo.page-object.ts
```diff
@@ -50,6 +37,6 @@ export class KeyValuePairInfoPo extends BasePageObject {
   }
 
   getDescriptionText(): Promise<string> {
-    return this.parent.byTestId(`${this.testId}-description`).getText();
+    return this.root.byTestId(`${this.testId}-description`).getText();
```
dskloetd, line 40: It's a bit unfortunate that the testId of the child depends on the testId of the parent and because of that we still need to hold on to the testId.
How about using `.querySelector(".description")` instead?
And then we can remove `testId` as a class field.

lmuntaner: I understand the point of not depending on the parent. But the test id is different.

If we even change the style this would break. The point of data-tid is to not use the style selectors, isn't it?

Maybe we can do another change in the GIX component to remove the test-id dependency and use a static test id for the description.

dskloetd: OK.

Then you can at least simplify this to
```
return this.getText(`${this.testId}-description`);
```

dskloetd (approved):

PR #2356, opened 2023-04-21
E2E: SNS swap participation

dskloetd:
# Motivation

We want to test SNS swap participation.
This just puts the test file in place to be extended in future PRs.

# Changes

1. Add frontend/src/tests/e2e/sns-participation.spec.ts which just signs in and navigates to the launchpad. It contains TODOs for what to test.
2. Add PO functionality for navigating to the launchpad.
3. Make sure the app is loaded before we check if we need to open the menu or not.
4. Make click methods in `MenuItems.page-object.ts` return a `Promise`. It was an oversight that they didn't.

# Tests

`npm run test-e2e`

### DIFF
--- a/frontend/src/tests/page-objects/App.page-object.ts
+++ b/frontend/src/tests/page-objects/App.page-object.ts
@@ -47,17 +47,18 @@ export class AppPo extends BasePageObject {
     return this.getMenuTogglePo().click();
   }
 
-  waitForContentLoaded(): Promise<void> {
-    // Wait for an arbitrary element that is present when the content is loaded.
-    return this.getButton("account-menu").waitFor();
+  waitForHeaderLoaded(): Promise<void> {
+    return this.root.byTestId("header-component").waitFor();
   }
 
   async openMenu(): Promise<void> {
-    // If the content isn't loaded yet, it looks like the menu toggle isn't
-    // there but it's just not there *yet*.
-    await this.waitForContentLoaded();
+    // On large viewports, the menu is always open, but on smaller windows, we
+    // need to click the menu toggle to open the menu. So we check if the menu toggle is
+    // there and wait for the menu to open if necessary.
 
-    // Whether the menu needs to be opened depends on the size of the viewport.
+    // If the header isn't loaded yet, it looks like the menu toggle isn't
+    // there but it's just not there *yet*.
+    await this.waitForHeaderLoaded();
     const isTogglePresent = await this.getMenuTogglePo().isPresent();
     if (isTogglePresent) {
       const backdrop = this.getBackdropPo();
--- a/frontend/src/tests/page-objects/App.page-object.ts
+++ b/frontend/src/tests/page-objects/App.page-object.ts
@@ -47,7 +47,16 @@ export class AppPo extends BasePageObject {
     return this.getMenuTogglePo().click();
   }
 
+  waitForContentLoaded(): Promise<void> {
+    // Wait for an arbitrary element that is present when the content is loaded.
+    return this.getButton("account-menu").waitFor();
+  }
+
   async openMenu(): Promise<void> {
+    // If the content isn't loaded yet, it looks like the menu toggle isn't
+    // there but it's just not there *yet*.
+    await this.waitForContentLoaded();
+
     // Whether the menu needs to be opened depends on the size of the viewport.
     const isTogglePresent = await this.getMenuTogglePo().isPresent();
     if (isTogglePresent) {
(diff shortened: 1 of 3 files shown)


frontend/src/tests/page-objects/App.page-object.ts
```diff
@@ -47,7 +47,16 @@ export class AppPo extends BasePageObject {
     return this.getMenuTogglePo().click();
   }
 
+  waitForContentLoaded(): Promise<void> {
+    // Wait for an arbitrary element that is present when the content is loaded.
+    return this.getButton("account-menu").waitFor();
```
lmuntaner, line 52: That the menu is loaded it doesn't mean the content of the app is loaded. There could be skeletons and the menu is still available.

Does it make sense to rename this to specifically `waitForMenuLoaded`?

lmuntaner (approved):
Thanks, just a minor comment but already approving. It's more about naming.

frontend/src/tests/page-objects/App.page-object.ts
dskloetd: Hm. Your objection makes sense but I'm not sure about the solution so let me describe the problem.
1. On big screens, the menu items on the left (Tokens, Neuron staking, Launchpad, etc) are always visible.
2. On small screens you have to click the menu first.
3. I tried to make the test agnostic of this, although the test always renders the window small enough that you have to click the menu first.
4. So in order to decide if the menu has to be opened, I check if the menu button is there. This is the hamburger menu in the upper left.
5. This works on the existing tests because they always do something else first.
6. But the participation test starts with going to the launchpad so opening the menu is the first thing it does.
7. My first attempt at the test failed because, when it checked if the menu had to be opened, there was still a blank screen (and no menu button), so it assumed that the window is wide enough that the menu doesn't have to be opened. Then it waited for the Launchpad menu item to appear but it didn't because the menu had to be opened.
8. So I looked for a way to make sure that we no longer have this blank screen.
9. I can't check for the menu button because it doesn't exist if the window is large enough.
10. But I found the account menu button (in the upper right) which I'm guessing is always there.
11. Since you suggested `waitForMenuLoaded`, I guess you didn't realize that there are actually 2 different menus here.

Should I just rely on the fact that the window is always small enough that we have to open the menu and not try to be agnostic about this?

Or can we think of another better name for this?
`waitForLayoutLoaded`? `waitForHeaderLoaded`?

lmuntaner: Thanks for the explanation!

I wouldn't rely on the size of the window.

Renaming should suffice, but then it has to be specific to what it's doing. The `account-menu` is the one on the top right. Not the one you click, is that correct?

Can you want for the one you click instead? Should we agree on calling this sidebar the navigation bar? If we agree, it's easy to agree on future namings. This could be `waitForNavbarButton`?

Although, I also like `waitForHeaderLoaded`. I believe this is also good enough.

dskloetd: > The `account-menu` is the one on the top right. Not the one you click, is that correct?

Correct.

> Can you want for the one you click instead?

No, that's what I tried to explain. The menu button isn't there (and won't be there) if the window is big enough.
So I need to wait for the time that it would be there, but then it still might not be and then I won't click it.

> Should we agree on calling this sidebar the navigation bar?

The component it called `MenuItems` so that's what I'm been calling it.

> This could be `waitForNavbarButton`?

The MenuItems only appear after you click the menu button, which we don't know if it will appear. We can't wait for the MenuItems to appear before click the menu button. It's the other way around.

The menu button that I want to click (only if it's there) is in the header. So I plan to actually wait for the header to be there. That's why I added the tid to the header in https://github.com/dfinity/gix-components/pull/203. I just need to pull in the new gix components and update this PR. I'll request another review when I've done that.

lmuntaner: I like the idea of waiting for the content of the header. It makes sense.

dskloetd: Done

frontend/src/tests/page-objects/App.page-object.ts
```diff
@@ -47,8 +47,18 @@ export class AppPo extends BasePageObject {
     return this.getMenuTogglePo().click();
   }
 
+  waitForHeaderLoaded(): Promise<void> {
+    return this.root.byTestId("header-component").waitFor();
+  }
+
   async openMenu(): Promise<void> {
-    // Whether the menu needs to be opened depends on the size of the viewport.
+    // On large viewports, the menu is always open, but on smaller windows, we
```
lmuntaner, line 55: Thanks for the comment!

lmuntaner (approved):
LGTM, thanks!

PR #2358, opened 2023-04-24
Fix ic-js dist imports

peterpeterparker:
# Motivation

One token related and many Sns related types incorrectly referenced the `dist` package of ic-js libraries.

# PRs

- [x] https://github.com/dfinity/ic-js/pull/329
- [x] https://github.com/dfinity/ic-js/pull/330

# Changes

- fix import
- use Sns- prefixed import names


lmuntaner (approved):
Thanks!

PR #2361, opened 2023-04-24
E2E: See open swaps

dskloetd:
# Motivation

Continuing with SNS end-to-end testing.

# Changes

1. In `frontend/src/tests/e2e/sns-participation.spec.ts` check that there is 1 open project on the Launchpad page.
2. Add page objects for `Launchpad`, `Projects` and `ProjectCard`.
3. Fix the name of the test which was accidentally copied from "Test SNS governance".

# Tests

`npm run test-e2e`
`npm run check`
`npm run test`

PR #2363, opened 2023-04-24
Add testId to ProjectCard

dskloetd:
# Motivation

I'm adding a testId to `ProjectCard` in https://github.com/dfinity/nns-dapp/pull/2361 but it overrides the testId of the `Card` component, which causes an existing test to fail.
Because the test is unrelated to that PR, I'm updating it in this separate PR.

# Changes

1. Add testId on `Card` in `frontend/src/lib/components/launchpad/ProjectCard.svelte`.
2. Update `frontend/src/tests/lib/components/launchpad/Projects.spec.ts` to use that testId instead of the default `"card"` testId.

# Tests

`npm run test`

### DIFF
--- a/frontend/src/lib/components/launchpad/ProjectCard.svelte
+++ b/frontend/src/lib/components/launchpad/ProjectCard.svelte
@@ -37,6 +37,7 @@
 </script>
 
 <Card
+  testId="project-card-component"
   role="link"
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Thanks!

PR #2364, opened 2023-04-24
GIX-1519: Stop using deprecated endpoint in SNS Aggregator. Part 1

lmuntaner:
# Motivation

The SNS aggregator uses the deprecated endpoint `get_state`.

We want to move away from this endpoint.

This PR: Fetch swap parameters from new endpoint `get_sale_parameters`.

# Changes

* Add new property `swap_params` to `UpstreamData`.
* Fetch the data from `swap_params` when we fetch the other data.
* Add new property `swap_params` to `SlowSnsData` which is used to render the JSON.

Minor changes:
* Make `params` property public in `GetSaleParametersResponse`.

# Tests

* There is no data transformation. And we still don't have integration tests to check calls to the backend.


rs/sns_aggregator/src/types/upstream.rs
```diff
@@ -52,4 +53,6 @@ pub struct UpstreamData {
     pub icrc1_fee: Nat,
     /// The ledger total tokens supply
     pub icrc1_total_supply: Nat,
+    /// The params of the swap
+    pub swap_params: Option<GetSaleParametersResponse>,
```
lmuntaner, line 57: @bitdivine Do I need to set new fields to optional to not break the stable memory?

rs/sns_aggregator/src/types/ic_sns_swap.rs
```diff
@@ -228,7 +228,7 @@ pub struct Params {
   pub  min_icp_e8s: u64,
 }
 
-#[derive(CandidType, Deserialize, Serialize, Clone, Debug)]
+#[derive(CandidType, Deserialize, Serialize, Clone, Debug, Default)]
```
lmuntaner, line 231: @bitdivine what does this do? What kind of value will this create if we get an error in the call?

rs/sns_aggregator/src/types/ic_sns_swap.patch
```diff
@@ -58,3 +58,15 @@ index caf728e4..739c2846 100644
  
  #[derive(CandidType, Deserialize, Serialize, Clone, Debug)]
  pub struct OpenRequest {
+
+--- a/rs/sns_aggregator/src/types/ic_sns_swap.rs
```
lmuntaner, line 62: @bitdivine this is the patch I'm trying to add, but it doesn't go through.

How did you create the patch file in the first place?

bitdivine: I used diff.  In particular, I:
- Generated the files.
- Tried to compile but got errors.
- Fixed up the files where they were failing.
- Saved the diff to the patch file
- Wrote a script to try to make this process easy and reproducible.  Try running this:

```
scripts/mk_nns_patch.sh sns_swap
git diff --cached
```

This _should_ update the patch file.

rs/sns_aggregator/src/types/ic_sns_swap.rs
bitdivine: This creates a default implementation of default, which will return, in this case, `GetSaleParametersResponse{ params: None }`.  I forget why this is needed.  Comment it out and see!  Sometimes Default is "required" even when it is never needed.  Let me have a look...

rs/sns_aggregator/src/types/upstream.rs
bitdivine: I am not sure.  Breaking stable memory is not terrible, though, for the aggregator.  The stable memory will be discarded and the aggregator will re-scrape the data.  So there will be a bit more downtime but that's the worst of it.

rs/sns_aggregator/src/types/ic_sns_swap.rs
bitdivine: Still compiling.  I'll finish populating this when done.

bitdivine: Ok.  So, if `get_sale_parameters()` fails, the default response is used.  I actually think it would be better to return None in this case:
```
diff --git a/rs/sns_aggregator/src/types/ic_sns_swap.rs b/rs/sns_aggregator/src/types/ic_sns_swap.rs
index cf00f6606..321ccf5ea 100644
--- a/rs/sns_aggregator/src/types/ic_sns_swap.rs
+++ b/rs/sns_aggregator/src/types/ic_sns_swap.rs
@@ -228,7 +228,7 @@ pub struct Params {
   pub  min_icp_e8s: u64,
 }
 
-#[derive(CandidType, Deserialize, Serialize, Clone, Debug, Default)]
+#[derive(CandidType, Deserialize, Serialize, Clone, Debug)]
 pub struct GetSaleParametersResponse { params: Option<Params> }
 
 #[derive(CandidType, Deserialize, Serialize, Clone, Debug)]
diff --git a/rs/sns_aggregator/src/upstream.rs b/rs/sns_aggregator/src/upstream.rs
index ae38e14fe..ca80a34c6 100644
--- a/rs/sns_aggregator/src/upstream.rs
+++ b/rs/sns_aggregator/src/upstream.rs
@@ -131,12 +131,17 @@ async fn get_sns_data(index: u64, sns_canister_ids: DeployedSns) -> anyhow::Resu
         .map_err(|err| anyhow!("Failed to get ledger total tokens supply: {err:?}"))
         .unwrap_or_default();
 
-    let swap_params_response: GetSaleParametersResponse =
-        ic_cdk::api::call::call(swap_canister_id, "get_sale_parameters", (EmptyRecord {},))
+    let swap_params_response: Option<GetSaleParametersResponse> =
+        match ic_cdk::api::call::call(swap_canister_id, "get_sale_parameters", (EmptyRecord {},))
             .await
             .map(|response: (_,)| response.0)
-            .map_err(|err| crate::state::log(format!("Failed to get swap params: {err:?}")))
-            .unwrap_or_default();
+        {
+            Err(err) => {
+                crate::state::log(format!("Failed to get swap params: {err:?}"));
+                None
+            }
+            Ok(response) => Some(response),
+        };
 
     crate::state::log("Yay, got an SNS status".to_string());
     // If the SNS sale will open, collect data when it does.
@@ -152,7 +157,7 @@ async fn get_sns_data(index: u64, sns_canister_ids: DeployedSns) -> anyhow::Resu
         icrc1_metadata,
         icrc1_fee,
         icrc1_total_supply,
-        swap_params: Some(swap_params_response),
+        swap_params: swap_params_response,
     };
     State::insert_sns(index, slow_data)
         .map_err(|err| crate::state::log(format!("Failed to create certified assets: {err:?}")))
```
Would you like me to push that commit?

rs/sns_aggregator/src/types/upstream.rs
lmuntaner: let's keep it option to handle the case of error.

rs/sns_aggregator/src/types/ic_sns_swap.rs
lmuntaner: thanks!

bitdivine (approved):
Waiting for CI but this _looks_ good.  Thanks!

rs/sns_aggregator/src/types/ic_sns_swap.patch
lmuntaner: thanks!

PR #2368, opened 2023-04-24
Typecheck page objects

dskloetd:
# Motivation

Somehow, Playwright doesn't do typechecking.
So page objects which are only used in Playwright tests, often have type errors which are only caught when those page objects are first used in a Jest test.
Most/all page objects used by Playwright tests are (indirectly) depended on by `AppPo`.

# Changes

1. Add a dummy Jest test which depends on AppPo.
2. Fix discovered issues.

# Tests

`npm run test`


lmuntaner (approved):
Thanks!

PR #2372, opened 2023-04-25
fix: apply proposals visibility after neuron loading

mstrasinskis:
# Motivation

Fix empty proposal list after load when `only-open` filter selected.
The visibility logic is also based on the user neurons presence in the proposal ballots (see `containsUnspecifiedBallot`)

# Changes

- reapply filter after neuron store changes

### DIFF
--- a/frontend/src/lib/pages/NnsProposals.svelte
+++ b/frontend/src/lib/pages/NnsProposals.svelte
@@ -127,7 +127,7 @@
     debounceFindProposals?.();
   };
 
-  $: applyFilter($proposalsFiltersStore);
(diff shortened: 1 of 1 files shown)


lmuntaner (commented):
Thanks!

Can we add a test for this?

mstrasinskis:
> Can we add a test for this?
@lmuntaner not really, this changes only component internal state. Doing something like `jest.spyOn(proposalsStore, "setProposals");` would rely too much on the implementation.

lmuntaner (approved):
Ok, let's get this in.

Maybe we can do an e2e test at some point.

PR #2374, opened 2023-04-25
Chore: Avoid source duplication env vars testing

lmuntaner:
# Motivation

Jest setup was mocking twice some variables. For example, DFX_NETWORK was mocked in the env-vars as well as in environment.constants.

This PR splits environment.constants.ts in two: the flags that depend on the env vars, and the ones that not.

# Changes

Main changes:
* New file enviroment-flags.constants.ts
* Remove constants from environment.constants.ts

Minor changes:
* Change imports that used vars from environment.constants.ts and moved to enviroment-flags.constants.ts

# Tests

* Mock enviroment-flags.constants.ts in jest-setup.ts


lmuntaner:
@dskloetd please review

frontend/src/lib/constants/environment-flags.constants.ts
dskloetd, line 1: Do you want to rename this to `mockable.constants.ts` or something like that?

And put a comment at the top explaining something like
```
/// These are constants that can have a different value during tests but which are not set with the arguments passed to the canister.
```

dskloetd (approved):

PR #2375, opened 2023-04-25
Fix nightly tagging

bitdivine:
# Motivation
Triggering builds with tags is unreliable.  Presumably there is a race condition between adding a tag to git and triggering the run, so the list of tags known to git may or may not include the tag that triggered an event.

# Changes
- Check whether a docker-build workflow was triggered by a tag.  If so, add the tag to the list of tags for this commit, if it is not already there.

# Tests
Only time will tell for sure.  But pushing a tag does show the tag being identified in the github workflow.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -26,13 +26,13 @@ jobs:
           : Log commit and existing tags
           git rev-parse HEAD
           git tag --points-at HEAD
+          readarray -t TAGS < <(git tag --points-at HEAD)
           : Log tag that triggered this run, if any
           GH_REF="${{github.ref}}"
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -20,6 +20,21 @@ jobs:
             DFX_NETWORK: "mainnet"
     steps:
```
bitdivine, line 21: Note: Having this printout right at the beginning of the workflow makes debugging faster.  I would like to keep this in place until this matter is resolved.  However it's not a problem just adding it while debugging, in case you would prefer not to have such a debugging step in main.

bitdivine:
Discussed in Slack.  Agreed that we need to get tag builds working for all tags.

lmuntaner (approved):
Thanks!

PR #2377, opened 2023-04-25
GIX-1505: Fix prod preview

lmuntaner:
# Motivation

Fix `npm run preview` which runs a production-like environment.

# Changes

Main change:
* Logic inside `getEnvVars`. Force to use the build env vars if the port is set. This is the case for local dev server and preview.

Minor changes:
* Add `raw` to the SNS aggregator url when running from localhost as well. This is skipped if the `url` of the aggregator already includes `localhost`.
* New env util `isLocalhost`.

# Tests

* Test new env util.

I tested the three ways:
* Deploy to testnet
* Local dev server
* Preview

# Docs

Add a comment on how to run `npm run preview`.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -41,10 +41,12 @@ The [canister_ids.json] data provides the list of canister IDs available for var
 
 ## Preview
 
-To run a production like environment locally agains a replica run:
+To run a production like environment locally against a replica run:
 
 ```bash
-DFX_NETWORK=<replica-network> npm run preview
+DFX_NETWORK=<replica-network> ../config.sh
+npm run build
+npm run preview
 ```
 
 ### Configure
--- a/HACKING.md
+++ b/HACKING.md
@@ -39,6 +39,14 @@ Once you have a running replica with nns installed and a fixed
 
 The [canister_ids.json] data provides the list of canister IDs available for various test environments.
 
+## Preview
+
+To run a production like environment locally agains a replica run:
+
+```bash
+DFX_NETWORK=<replica-network> npm run preview
+```
+
 ### Configure
 
 To develop and run locally the dapp against a testnet, proceed as following:
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -9,7 +9,7 @@
     "i18n": "node --experimental-json-modules scripts/i18n.types.js",
     "dev": "npm run i18n && vite dev",
     "build": "npm run i18n && tsc --noEmit && vite build && npm run build:post-process",
-    "preview": "../config.sh && npm run build && vite preview",
+    "preview": "vite preview",
     "check": "svelte-kit sync && svelte-check --fail-on-warnings --tsconfig ./tsconfig.json && npm run lint",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
     "lint": "pprettier --check './**/*.{ts,js,mjs,json,scss,svelte,html}' && eslint --max-warnings 0 .",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -9,7 +9,7 @@
     "i18n": "node --experimental-json-modules scripts/i18n.types.js",
     "dev": "npm run i18n && vite dev",
     "build": "npm run i18n && tsc --noEmit && vite build && npm run build:post-process",
-    "preview": "vite preview",
+    "preview": "../config.sh && npm run build && vite preview",
     "check": "svelte-kit sync && svelte-check --fail-on-warnings --tsconfig ./tsconfig.json && npm run lint",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
     "lint": "pprettier --check './**/*.{ts,js,mjs,json,scss,svelte,html}' && eslint --max-warnings 0 .",
--- a/frontend/src/lib/constants/environment.constants.ts
+++ b/frontend/src/lib/constants/environment.constants.ts
@@ -1,22 +1,25 @@
 import { getEnvVars } from "$lib/utils/env-vars.utils";
-import { addRawToUrl } from "$lib/utils/env.utils";
+import { addRawToUrl, isLocalhost } from "$lib/utils/env.utils";
+import { isBrowser } from "@dfinity/auth-client/lib/cjs/storage";
 
 const envVars = getEnvVars();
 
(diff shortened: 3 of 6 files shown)


lmuntaner:

frontend/package.json
```diff
@@ -9,7 +9,7 @@
     "i18n": "node --experimental-json-modules scripts/i18n.types.js",
     "dev": "npm run i18n && vite dev",
     "build": "npm run i18n && tsc --noEmit && vite build && npm run build:post-process",
-    "preview": "vite preview",
+    "preview": "../config.sh && npm run build && vite preview",
```
dskloetd, line 12: If I manually run
`DFX_NETWORK=foo ./config.sh`
then rerunning `../config.sh` without DFX_NETWORK set will break what I intended to do, right?

frontend/src/lib/utils/env-vars.utils.ts
```diff
@@ -127,9 +127,28 @@ const getBuildEnvVars = (): EnvironmentVars => {
...
-  if (isBrowser && !localDev) {
-    return getHtmlEnvVars();
+  // There are two instances of local dev server:
+  // - The local dev server started by `npm run dev`
+  // - The local dev server started by `npm run preview`. The only way to distinguis this one is by the port.
+  // We need to check `isBrowser` to skip the check when running the build.
+  const isLocalDevServer =
+    localDev || (isBrowser && window.location.port !== "");
+  if (!isBrowser || isLocalDevServer) {
+    return getBuildEnvVars();
```
dskloetd, line 151: CI runs against `http://qsgjb-riaaa-aaaaa-aaaga-cai.localhost:8080` which is not a development server but does have a port.
I'm afraid you'll have to check against specific ports (5173 or 4173), which is what I thought we were talking about.

Also, it's very confusing that localDev doesn't actually mean the same as isLocalDevServer.

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -1,22 +1,25 @@
...
     const { hostname } = new URL(url);
-    if (hostname.includes("localhost") || hostname.includes("127.0.0.1")) {
+    if (isLocalhost(hostname)) {
       return url;
     }
 
-    if (DEV) {
+    // If the nns-dapp is running in loalhost, we need to add `raw` to the URL to avoid CORS issues.
+    if (DEV || (isBrowser && isLocalhost(window.location.hostname))) {
       return addRawToUrl(url);
```
dskloetd, line 23: Do you still need to check for DEV as well? I wouldn't be surprised if all the other checks now cover everything.

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -1,22 +1,25 @@
...
 export const FETCH_ROOT_KEY: boolean = envVars.fetchRootKey === "true";
 
 const snsAggregatorUrlEnv = envVars.snsAggregatorUrl ?? "";
 const snsAggregatorUrl = (url: string) => {
   try {
     const { hostname } = new URL(url);
-    if (hostname.includes("localhost") || hostname.includes("127.0.0.1")) {
+    if (isLocalhost(hostname)) {
       return url;
     }
```
dskloetd, line 20: loalhost => localhost

frontend/package.json
lmuntaner: Yes, another option is that `npm run preview` depends on `./config.sh` being run before.

frontend/src/lib/utils/env-vars.utils.ts
lmuntaner: `localDev` is true always when `isLocalDevServer` is true. But not the other way around.

I'll rename to `localDevelopment`. I hope that makes it clearer.

frontend/src/lib/constants/environment.constants.ts
lmuntaner: Not really, `DEV` is always true for isLocalhost`

frontend/src/lib/utils/env-vars.utils.ts
lmuntaner: done

frontend/src/lib/constants/environment.constants.ts
lmuntaner: done

frontend/src/lib/constants/environment.constants.ts
lmuntaner: done

lmuntaner:

frontend/package.json
dskloetd: Is this change required for the fix? If not, I'd prefer to keep it the way it is.

frontend/src/lib/utils/env-vars.utils.ts
```diff
@@ -127,9 +127,30 @@ const getBuildEnvVars = (): EnvironmentVars => {
...
-  if (isBrowser && !localDev) {
-    return getHtmlEnvVars();
+  // There are two instances of local dev server:
+  // - The local dev server started by `npm run dev`
+  // - The local dev server started by `npm run preview`. The only way to distinguis this one is by the port.
+  // We need to check `isBrowser` to skip the check when running the build.
+  const isDevServer =
+    localDevelopment ||
+    (isBrowser &&
+      (window.location.port === "5173" || window.location.port === "4173"));
```
dskloetd, line 151: Optional: `["5173", "4173"].includes(window.location.port)` is shorter.

frontend/package.json
lmuntaner: not required. Ok, I'll revert

lmuntaner:

HACKING.md
```diff
@@ -39,6 +39,14 @@ Once you have a running replica with nns installed and a fixed
 
 The [canister_ids.json] data provides the list of canister IDs available for various test environments.
 
+## Preview
+
+To run a production like environment locally agains a replica run:
+
+```bash
+DFX_NETWORK=<replica-network> npm run preview
```
dskloetd, line 47: With the other change reverted, should this now be?
```
DFX_NETWORK=<replica-network> ../config.sh
npm run build
npm run preview
```

dskloetd (approved):

HACKING.md
lmuntaner: yes, thanks!

PR #2380, opened 2023-04-25
Summarize insertion

bitdivine:
# Motivation
The Kibana logs are flooded with print statements about inserting assets.  We can probably survive with fewer and the mass of messages makes it harder to pick out other activity.

# Changes
- Print a count of asset insertions rather than a blow by blow narrative of every wonderful chunk that is now available.
- Print the canister install/upgrade arguments as a one-liner.  It's not as pretty when running locally but is much easier when getting the data in Kibana.  Not just easier, it turns out that a multi-line printout in Kibana logs the first line N times where N is the number of lines.  Spammy and information free.

# Tests
See a deployment log now:
```
[Canister st75y-vaaaa-aaaaa-aaalq-cai] post_upgrade with args: Some(CanisterArguments { args: [("CYCLES_MINTING_CANISTER_ID", "rkp4c-7iaaa-aaaaa-aaaca-cai"), ("DFX_NETWORK", "loc...[SNIP]
[Canister st75y-vaaaa-aaaaa-aaalq-cai] Inserting assets...
[Canister st75y-vaaaa-aaaaa-aaalq-cai] Inserted 263 assets.

```

### DIFF
--- a/rs/backend/src/assets.rs
+++ b/rs/backend/src/assets.rs
@@ -364,7 +364,7 @@ pub fn insert_asset_into_state<S: Into<String> + Clone>(state: &State, path: S, 
 ///
 /// Note: Used both in init and post_upgrade
 pub fn init_assets() {
-    dfn_core::api::print("Inserting assets...".to_string());
+    dfn_core::api::print("Inserting assets...");
     let mut num_assets = 0;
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
LGTM!

PR #2381, opened 2023-04-25
"Internetwork" should be hyphenated "Inter-network"

peterpeterparker:
# Motivation

Product request: "Internetwork" should be hyphenated "Inter-network"


lmuntaner (approved):
Thanks!

PR #2382, opened 2023-04-26
ckBTC display error as info

peterpeterparker:
# Motivation

> The "error" message when there are no new UXTOs looks to "error" like to me ("Sorry, the balance cannot be updated. There are no new UTXOs to process."). This is expected behaviour, not an error. I'm not sure users know what UXTOs are. Can we remove the (!) symbol and simply say "Updated available ckBTC : No new confirmed BTC "

# Changes

- display UTXO error as info


lmuntaner (approved):
LGTM!

PR #2383, opened 2023-04-26
GIX-1416: Set icrc1_total_supply as required

lmuntaner:
# Motivation

A few weeks ago we added a new field in the SNS aggregator. The NNS Dapp upgraded first and to handle backwards compatibility we set the field as optional.

Now that SNS aggregator was upgraded we can set the field `icrc1_total_supply` as required.

# Changes

Main change:
* Set field icrc1_total_supply required in the data from aggregator `CachedSnsMetadataDto`.

Minor change:
* No need to filter out by icrc1_total_supply not defined when setting the `snsTotalTokenSupplyStore` from SNS aggregator.

# Tests

* Add the new field in the SNS Aggregator mocks.
* Remove icrc1_total_supply specific test cases. The field is now required and expected with the mock.

### DIFF
--- a/frontend/src/tests/lib/api/sns-aggregator.api.spec.ts
+++ b/frontend/src/tests/lib/api/sns-aggregator.api.spec.ts
@@ -36,21 +36,4 @@ describe("sns-aggregator api", () => {
       expect(sns).toEqual(aggregatorSnsMock);
     });
   });
-
-  it("should include icrc1_total_supply property", async () => {
-    const mockFetch = jest.fn();
-    const totalSupply = BigInt(2000_000_000);
-    const aggregatedSnsesResponse = [
-      { ...aggregatedSnses[0], icrc1_total_supply: totalSupply },
-    ];
-    mockFetch.mockReturnValueOnce(
-      Promise.resolve({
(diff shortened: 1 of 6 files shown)


lmuntaner:

frontend/src/tests/lib/api/sns-aggregator.api.spec.ts
```diff
@@ -36,21 +36,4 @@ describe("sns-aggregator api", () => {
       expect(sns).toEqual(aggregatorSnsMock);
     });
   });
-
-  it("should include icrc1_total_supply property", async () => {
```
dskloetd, line 40: Do we need to remove this?
Shouldn't it still set the property?

dskloetd (approved):

frontend/src/tests/lib/api/sns-aggregator.api.spec.ts
lmuntaner: This was specific because the mock data didn't have the extra field. In this test I wanted to check that if the api data had the field, then the returned data also.

But now the test data used in the expect has this field

PR #2385, opened 2023-04-26
GIX-1519: Stop using deprecated endpoint in SNS Aggregator. Part 2

lmuntaner:
# Motivation

The SNS aggregator uses the deprecated endpoint `get_state`.

We want to move away from this endpoint.

This PR: Fetch swap parameters from new endpoint `get_init`.

# Changes

* Add new property `init` to `UpstreamData`.
* Fetch the data from `get_init` when we fetch the other data.
* Add new property `init` to `SlowSnsData` which is used to render the JSON.

# Tests

* There is no data transformation. And we still don't have integration tests to check calls to the backend.


lmuntaner:
@bitdivine please review

bitdivine (approved):
LGTM, thanks!

PR #2387, opened 2023-04-26
Load ckBTC address on ckBTC receive modal as well

peterpeterparker:
# Motivation

The ckBTC receive modal can also be displayed on the "Accounts" page, therefore it has to load - if not yet loaded - the ckBTC address as well.

# Changes

- extract existing loading logic in a component `BitcoinAddressLoader`
- use new component as parent in `BitcoinAddress` and `CkBTCReceiveModal`


frontend/src/lib/components/accounts/BitcoinAddressLoader.svelte
lmuntaner, line 1: This is a component that loads a store and renders nothing.

Does it make sense to create it as a service instead of a component?

The components that rely can call this service `loadBtcAddress(...)`.

Why is a component better than a service?

peterpeterparker: I originally started a service but having a component that renders a `slot` felt more accurate because I would have to use multiple stores and utils in the service - `isUniverseCkTESTBTC` and `bitcoinAddressStore` and adding `toastsError`. In addition, if I would have use a service, I would have needed to use the service in two components with the reactivity call on the `identifier` property. So all in all, felt more accurate and less verbose that way.

But really no strong opinion, if you rather like to move everything in a service, I can do so for sure.

lmuntaner: It's common to have `toastsError` and accessing multiple stores inside a service.

I believe that's the definition we used as a service.

I agree that the reactivity call might be annoying. Maybe you can use that service inside the component instead of having all the logic there though?

lmuntaner: I think that using a service follows better the patterns we have in the nns-dapp in this case.

Although I like the idea of components that have logic and return just the slot.

peterpeterparker: Like I said, no strong opinion. Should I move everything to a service?

peterpeterparker: I have moved everything to a service.

frontend/src/lib/components/accounts/BitcoinAddress.svelte
```diff
@@ -30,60 +29,39 @@
...
-  onMount(async () => await loadBtcAddress());
-
   let blockExplorerUrl: string;
   $: blockExplorerUrl = `${
     isUniverseCkTESTBTC(universeId)
       ? BITCOIN_BLOCK_EXPLORER_TESTNET_URL
       : BITCOIN_BLOCK_EXPLORER_MAINNET_URL
   }/${btcAddress ?? ""}`;
+
+  onMount(
```
lmuntaner, line 39: Don't you need also the reactivity when `universeId` or `account.identifier` changes?

lmuntaner (commented):
A few comments, thanks!

frontend/src/lib/components/accounts/BitcoinAddress.svelte
peterpeterparker: In this particular component it isn't a must, `onMount` is fine.
In the other, the modal, all parameters are deconstructed at once from the same source and only the account vary within the ui. Therefore according test looks fine but if you spotted something else?

frontend/src/tests/lib/services/ckbtc-minter.services.spec.ts
```diff
@@ -22,18 +31,27 @@ import { waitFor } from "@testing-library/svelte";
...
+        minterCanisterId: CKTESTBTC_MINTER_CANISTER_ID,
+        identifier: mockCkBTCMainAccount.identifier,
+      });
 
       await waitFor(() =>
         expect(spyGetAddress).toBeCalledWith({
           identity: mockIdentity,
-          canisterId: CKBTC_MINTER_CANISTER_ID,
+          canisterId: CKTESTBTC_MINTER_CANISTER_ID,
         })
```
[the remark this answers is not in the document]
peterpeterparker: Good point, thanks!

frontend/src/lib/components/accounts/BitcoinAddress.svelte
```diff
@@ -30,60 +29,39 @@
...
+    async () =>
```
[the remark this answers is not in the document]
peterpeterparker: We probably discussed this enough times I guess 😉.
No strong opinion, removed then.

lmuntaner (approved):
Ok, that makes sense then. Thanks for the explanation. LGTM!

PR #2390, opened 2023-04-26
Chore: No TVL call if no TVL canister id

lmuntaner:
# Motivation

TVl calls create an error on testnet where there is no TVL canister.

# Changes

* Return `undefined` and skip call If TVL_CANISTER_ID is not present in tvl api.

# Tests

* Add a test in tvl api to check new use case.

### DIFF
--- a/frontend/src/tests/lib/api/tvl.api.spec.ts
+++ b/frontend/src/tests/lib/api/tvl.api.spec.ts
@@ -1,6 +1,8 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { queryTVL } from "$lib/api/tvl.api.cjs";
 import { TVLCanister } from "$lib/canisters/tvl/tvl.canister";
 import { AnonymousIdentity } from "@dfinity/agent";
+import { Principal } from "@dfinity/principal";
 import mock from "jest-mock-extended/lib/Mock";
 
 jest.mock("@dfinity/agent", () => {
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/api/tvl.api.spec.ts
```diff
@@ -13,42 +15,72 @@ jest.mock("@dfinity/agent", () => {
...
     jest.spyOn(TVLCanister, "create").mockImplementation(() => tvlCanisterMock);
   });
 
-  afterAll(() => jest.clearAllMocks());
+  describe("with tvl canister id set", () => {
+    beforeEach(async () => {
+      // Casting to `any` is needed to change the value at runtime
+      const canisterIds = (await import(
+        "$lib/constants/canister-ids.constants"
+      )) as any;
```
dskloetd, line 43: Can you cast it to `{TVL_CANISTER_ID: string}` or something instead?

lmuntaner: The problem was that it's considered `readonly`. Let me improve the comment.

dskloetd: I still don't get it. If typescript thinks the type is 
```
{
  readonly TVL_CANISTER_ID: string,
  ...etc
}
```
then casting it to something where the field is not readonly should work, no?

lmuntaner: Oh, sorry, I didn't understand your initial comment. Yes, that's a good point. I'll change it.

lmuntaner: done

lmuntaner:

PR #2395, opened 2023-04-27
Nightly tag

bitdivine:
# Motivation
If a nightly build fails (e.g. due to a network error), the release is empty.

# Changes
- Create a tag and release only after the build has succeeded.

# Tests
- [ ] I have triggered this workflow when there is already a release for today, to verify that the workflow is skipped if the release already exists.
- [ ] I have deleted today's release and, triggered the workflow manually and verified that the assets are created.
- During testing, one docker build failed due to a networking error and the subsequent steps were indeed skipped.


dfx.json
```diff
@@ -29,7 +29,7 @@
       "type": "custom",
       "wasm": "internet_identity_dev.wasm",
       "candid": "internet_identity.did",
-      "build": "curl --fail -sSL \"https://github.com/dfinity/internet-identity/releases/download/release-2023-01-31/internet_identity_dev.wasm\" -o internet_identity_dev.wasm",
+      "build": "curl --fail -sSL \"https://github.com/dfinity/internet-identity/releases/download/release-2023-04-12/internet_identity_dev.wasm\" -o internet_identity_dev.wasm",
```
bitdivine, line 32: Added by CI.  To be removed before merging.

.github/workflows/nightly.yaml
```diff
@@ -6,6 +6,7 @@ on:
     branches:
       # Commit to the nightly branch and push to test.
       - nightly
+      - nightly-tag
```
bitdivine, line 9: For testing only.  To be removed before merging.

lmuntaner (approved):
LGTM, thanks!

PR #2400, opened 2023-04-27
Feat/sns neuron selection 2

mstrasinskis:
# Motivation

Provides ui to preselect sns neurons for voting.

# Changes

- refactor `VotingNeuronSelect` to reuse for sns
- refactor `VotingNeuronSelectList` to reuse for sns
- selection login in `SnsVotingCard`
- 

# Tests

- SnsVotingCard specs
- move and update some tests from VotingNeuronSelect to VotingNeuronSelectList
- update vote-registration-service specs because of the less toast message updates

# Screenshots
image

image

Screen Recording 2023-04-28 at 17 22 29


frontend/src/lib/components/sns-proposals/SnsVotingCard.svelte
```diff
@@ -114,7 +139,12 @@
               on:nnsConfirm={vote}
             />
           {/if}
-          TODO: add VotingNeuronSelect {fromDefinedNullable(proposal.id).id}
+
+          <VotingNeuronSelect>
+            <VotingNeuronSelectList disabled={voteRegistration !== undefined} />
+            <!--            <MyVotes {proposalInfo} />-->
```
lmuntaner, line 145: You can remove these comments, right?

frontend/src/lib/components/sns-proposals/SnsVotingCard.svelte
```diff
@@ -87,11 +79,44 @@
   let signedIn = false;
   $: signedIn = isSignedIn($authStore.identity);
 
+  const userSelectedNeurons = (): SnsNeuron[] =>
```
lmuntaner, line 82: Did you confirm that this uses the latest value of `votingNeuronSelectStore`?

I thought that creating the function with `const` didn't recalculated it. But maybe with the store it's always using the latest store.

frontend/src/lib/components/sns-proposals/SnsVotingCard.svelte
```diff
@@ -87,11 +79,44 @@
...
+        )
+      )
+      .filter(nonNullish);
+
+  const votingNeurons = () =>
+    nonNullish(snsParameters)
+      ? votableNeurons.map((neuron) =>
+          snsNeuronToVotingNeuron({
+            neuron,
+            snsParameters: snsParameters as SnsNervousSystemParameters,
```
lmuntaner, line 96: Do you need to cast even after `nonNullish`?

lmuntaner (approved):
Looking very good!!

A few comments but already approving.

Should we use the shorten version with ellipsis `...` for the neuron ids?

It's such a long number the ID that I don't think uses will care.

I believe you can use the `Hash` function or the util `shortenWithMiddleEllipsis`

frontend/src/lib/components/sns-proposals/SnsVotingCard.svelte
mstrasinskis: Unfortunately, due to TS linter.

lmuntaner: ah, because this is inside a `map`. Annoying, but it makes sense for TS to consider it as so

frontend/src/lib/components/sns-proposals/SnsVotingCard.svelte
mstrasinskis: This will be replaced w/ the next PR `sns MyVotes`.

mstrasinskis:
> I believe you can use the `Hash` function or the util `shortenWithMiddleEllipsis`
A really good point. `shortenWithMiddleEllipsis` was used because of the cut off tooltip (didn't find an easy way to fix)

frontend/src/lib/components/sns-proposals/SnsVotingCard.svelte
lmuntaner: ok, thanks for the clarification!

PR #2406, opened 2023-04-28
Make test-exports failure less confusing

dskloetd:
# Motivation

When dependencies change, the build can fail with output like this:
```
Sanity check
0a1
> canister_global_timer
9a11
> canister_update <ic-cdk internal> timer_executor
```
This is quite confusing.

# Changes

1. Explain the failure
2. Give instructions for what to do
3. Add functionality to update the `nns-dapp-exports.txt` file.

Now it looks like this:
```
Sanity check
0a1
> canister_global_timer
9a11
> canister_update <ic-cdk internal> timer_executor
Exports do not match
If this looks correct, run 'scripts/nns-dapp/test-exports --wasm nns-dapp.wasm --update-golden' to update the golden file
```

# Tests

1. Manually changed `nns-dapp-exports.txt` and ran `DFX_NETWORK=local ./build.sh` to see it fail.
2. Followed the instructions and ran the build again to see it pass.

### DIFF
--- a/scripts/nns-dapp/test-exports
+++ b/scripts/nns-dapp/test-exports
@@ -14,7 +14,22 @@ print_help() {
 source "$SOURCE_DIR/clap.bash"
 # Define options
 clap.define short=w long=wasm desc="The location of the nns-dapp wasm." variable=DFX_WASM_PATH default="nns-dapp.wasm"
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!!

PR #2414, opened 2023-05-01
Update ckBTC local deployment doc and commit

peterpeterparker:
# Motivation

Update docs information to deploy ckBTC locally manually.


HACKING.md
```diff
@@ -133,13 +139,17 @@ e.g. in `canisters`:
...
-Likewise, as the configuration is not yet automated, there are no `.env` variable that are yet generated.
+> Note: when the script is run, the network should be provided
 
-### Index
+```bash
+DFX_NETWORK=local ./scripts/ckbtc/deploy-ckbtc
+```
 
-The index canister does not exist yet on mainnet but, will be proposed soon. Same pattern will apply.
+After deployment, their respective IDs shall be collected and updated in [ckbtc-canister-ids.constants.ts](./frontend/src/lib/constants/ckbtc-canister-ids.constants.ts).
```
lmuntaner, line 150: Aren't they set in the config?

Shouldn't we add them wit the environment variables?

lmuntaner (approved):
Just a comment but already approving.

Thanks!

HACKING.md
peterpeterparker: Maybe. This part of the doc describe the "manual" deployment. I think there was some improvement on the "automatic" deployment but, not sure which part is covered or not yet. So, I just updated the manual process as it's the one I followed to manually deploy the canisters locally yesterday. For sure, later on, we should come back to this and update/remove accordingly. Thanks for the review.

PR #2415, opened 2023-05-01
ckBTC transaction description sent to BTC Network

peterpeterparker:
# Motivation

If `transaction.to.owner` is equals to minter, then display onl y `To: BTC Network`.
Per extenstion, when we display `From: BTC Network`, display only label too.

# Changes

- `descriptions` property becomes a function instead of a record
- create and use `ckBTCTransactionDescription` function to map ckBTC description 
- do not display identifier if a description is matching

# Tests

Capture d’écran 2023-05-01 à 08 32 50

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
@@ -6,13 +6,16 @@
   import IcrcTransactionCard from "./IcrcTransactionCard.svelte";
   import SkeletonCard from "../ui/SkeletonCard.svelte";
   import type { IcrcTransactionData } from "$lib/types/transaction";
+  import type { Transaction } from "$lib/types/transaction";
 
   export let account: Account;
   export let transactions: IcrcTransactionData[];
   export let loading: boolean;
   export let governanceCanisterId: Principal | undefined = undefined;
   export let completed = false;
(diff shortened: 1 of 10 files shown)


frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
```diff
@@ -6,13 +6,16 @@
...
   import type { IcrcTransactionData } from "$lib/types/transaction";
+  import type { Transaction } from "$lib/types/transaction";
 
   export let account: Account;
   export let transactions: IcrcTransactionData[];
   export let loading: boolean;
   export let governanceCanisterId: Principal | undefined = undefined;
   export let completed = false;
-  export let descriptions: Record<string, string> | undefined = undefined;
+  export let description:
```
lmuntaner, line 16: The name is not very descriptive.

How about `getTransactionDescription`?

lmuntaner (approved):
Minor comment, already approving

frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
peterpeterparker: Thanks, sure. I wasn't that a fan neither of "description".

lmuntaner: That sounds good to me also. Thanks!

PR #2418, opened 2023-05-01
Display a warning if ckBTC to BTC conversion unexpectedly failed

peterpeterparker:
# Motivation

Display a warning on login if ckBTC to BTC conversion unexpectedly failed - e.g. if user closes the browser after the transaction to the ledger was made but the minter conversion was not triggered.

# PRs

- [x] #2420
- [x] https://github.com/dfinity/gix-components/pull/208

# Changes

- save block index in local storage after transfer to ledger in conversion function
- remove block index in local storage after retrieve btc regardless if success or error as long as ui is active
- on login, if local storage contains any block storage, display a warning and clear local storage of the block indexes
- bump gix-cmp to display more lines in toast
- stringify bigint upon writing to local storage

# Screenshots

Capture d’écran 2023-05-02 à 15 39 02

PR #2420, opened 2023-05-01
Bubble Icrc transfer result block index

peterpeterparker:
# Motivation

For Pr #2418 as for other features it might be useful to know the resulting block index of an Icrc ledger `transfer`.

# Changes

- return block index as result of an Icrc ledger transfer


lmuntaner (approved):
In the api function it makes a lot of sense.

But are we going to use it in the UI so that we need to return it from the service?

peterpeterparker:
> But are we going to use it in the UI so that we need to return it from the service?

Not in the UI but, e.g. in a service `ckbtc-convert.services.convertCkBTCToBtc` that uses another service `ckbtc-accounts.services.ckBTCTransferTokens`

lmuntaner:
> > But are we going to use it in the UI so that we need to return it from the service?
> 
> Not in the UI but, e.g. in a service `ckbtc-convert.services.convertCkBTCToBtc` that uses another service `ckbtc-accounts.services.ckBTCTransferTokens`

Ok, thanks for the explanation!

PR #2422, opened 2023-05-02
Extend getIcrcAccount with owner and subaccount

peterpeterparker:
# Motivation

We need to query the balance of Icrc ledger for other accounts than `main`. This PR extends the API to supports the `owner` as an argument.


lmuntaner (approved):
Thanks!

Not for this PR, but I believe that the API layer should return only types from the api layer, not the `Account` type from the nns-dapp.

That's a legacy inconsistency that we should fix in the future.

PR #2427, opened 2023-05-02
Deploy ckBTC kyt canister

peterpeterparker:
# Motivation

Add ckBTC kyt to local deployment instruction.

# Changes

- add kyt canister to download/deploy scripts


scripts/ckbtc/deploy-ckbtc
```diff
@@ -3,16 +3,19 @@
 # Original source
 # https://github.com/dfinity/ckBTC-Minter-Frontend/blob/master/local_deploy.sh
 
-dfx canister create ckbtc_ledger --network "$DFX_NETWORK"
-dfx canister create ckbtc_minter --network "$DFX_NETWORK"
+dfx canister create ckbtc_test_ledger --network "$DFX_NETWORK"
```
lmuntaner, line 6: We removed the `_test` on purpose because that's how we call it in dfx.json

Why are you reverting the naming?

peterpeterparker: I wrongly assumed that we do not want to overwrite what deployed in `dfx.json`. I'll rever to name without `_test`, thanks for the hint.

peterpeterparker: Done.

PR #2430, opened 2023-05-02
Check ckBTC minter balance and display call to action

peterpeterparker:
# Motivation

On ckBTC accounts page, once per session, check the minter canister balance for the user withdrawal account.

Display a related button in form of a card with an information if data are loading or loaded.

# Changes

- add new `AccountType` with the type of account: `withdrawalAccount`
- new component `CkBTCWithdrawalAccount` (without meaningful content yet)
- new function `getCkBTCWithdrawalAccount` that loads withdrawal account and account and handles case in which one of the other should be skipped and values from store should be used instead
- new `loadCkBTCWithdrawalAccount` for query and update loading
- new store `ckBTCWithdrawalAccountsStore`


frontend/src/lib/services/ckbtc-accounts-loader.services.ts
```diff
@@ -27,3 +45,83 @@ export const getCkBTCAccounts = async ({
...
+    };
+  };
+
+  const store = get(ckBTCWithdrawalAccountsStore);
+  const storedWithdrawalAccount = store[universeId.toText()];
+
+  // We have to load the withdrawal account with an update call.
+  // If we never have loaded it, we return a empty account as result of the not certified (query) call to indicate we are about to load the data.
+  if (!certified && isNullish(storedWithdrawalAccount?.account.identifier)) {
+    return {
```
lmuntaner, line 93: Why do we even want to return this object?

Why not returning `undefined`?

frontend/src/lib/services/ckbtc-accounts-loader.services.ts
```diff
@@ -27,3 +45,83 @@ export const getCkBTCAccounts = async ({
 
   return [account];
 };
+
+export const getCkBTCWithdrawalAccount = async ({
```
lmuntaner, line 49: I need a map of canisters.

Maybe you can help me to understand the difference:
* getCkBTCWithdrawalAccount
* getWithdrawalAccount
* getCkBTCAccount

They are all about `getXXXAccount`. What the difference among them?

I believe only `getCkBTCWithdrawalAccount` returns of type `Account` (from nns-dapp). Is this correct?

frontend/src/lib/components/accounts/CkBTCWithdrawalAccount.svelte
```diff
@@ -0,0 +1,71 @@
...
+    (isNullish(account.balance) || isNullish(account.identifier));
+
+  let accountBalance: TokenAmount;
+  $: accountBalance =
+    account?.balance ??
+    (TokenAmount.fromString({ amount: "0", token: ICPToken }) as TokenAmount);
+</script>
+
+{#if nonNullish(account)}
+  <button
```
lmuntaner, line 60: Why is this a button if there is no click event?

frontend/src/lib/components/accounts/CkBTCWithdrawalAccount.svelte
```diff
@@ -0,0 +1,71 @@
...
+    class="card"
+    data-tid="open-restart-convert-ckbtc-to-btc"
+    disabled={isNullish(account.balance) || isNullish(account.identifier)}
+  >
+    {#if loading}
+      Loading redeeming account...
+    {:else}
+      Redeeming account <AmountDisplay singleLine amount={accountBalance} />
```
lmuntaner, line 68: I guess this is temporary?

frontend/src/lib/components/accounts/CkBTCWithdrawalAccount.svelte
```diff
@@ -0,0 +1,71 @@
...
+  import { loadCkBTCWithdrawalAccount } from "$lib/services/ckbtc-withdrawal-accounts.services";
+  import { onMount } from "svelte";
+  import {
+    type CkBTCBTCWithdrawalAccount,
+    ckBTCWithdrawalAccountsStore,
+  } from "$lib/stores/ckbtc-withdrawal-accounts.store";
+  import AmountDisplay from "$lib/components/ic/AmountDisplay.svelte";
+  import { ICPToken, TokenAmount } from "@dfinity/nns";
+
+  const reloadAccount = async () => {
```
lmuntaner, line 13: Why do you move this logic here?

Why not add:

```
await loadCkBTCWithdrawalAccount({
      universeId: $selectedCkBTCUniverseIdStore,
    });
```

Inside the `loadAccount`?

frontend/src/tests/lib/components/accounts/CkBTCWithdrawalAccount.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+
+    it("should not render a button if no account", () => {
+      const { getByTestId } = render(CkBTCWithdrawalAccount);
+
+      expect(() => getByTestId("open-restart-convert-ckbtc-to-btc")).toThrow();
+    });
+  });
+
+  describe("getCkBTCAccount", () => {
```
lmuntaner, line 58: What does this describe do?

Do you mean, when account is loaded successfully?

frontend/src/tests/lib/services/ckbtc-accounts-loader.services.spec.ts
```diff
@@ -35,4 +47,182 @@ describe("ckbtc-accounts-loader-services", () => {
...
+    });
+
+    it("should throw no minter found for incompatible universe", () => {
+      const call = () =>
+        getCkBTCWithdrawalAccount({
+          ...params,
+          universeId: OWN_CANISTER_ID,
+        });
+
+      expect(call).rejects.toThrowError();
```
lmuntaner, line 76: Can we expect the specific error?

To make sure that if we change something and the service breaks in unexpected ways, this test case won't pass

frontend/src/tests/lib/services/ckbtc-withdrawal-accounts.services.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    });
+
+    const spyGetWithdrawalAccount = jest
+      .spyOn(minterServices, "getWithdrawalAccount")
+      .mockResolvedValue({
+        owner: mockCkBTCWithdrawalIcrcAccount.owner,
+        subaccount: [mockCkBTCWithdrawalIcrcAccount.subaccount],
+      });
+
+    it("should call api.getSnsAccounts and load neurons in store", async () => {
```
lmuntaner, line 28: Should it call `getSnsAccounts` or `getCkBTCAccount`, and load neurons or account? hehe

lmuntaner (commented):
Some questions.

Cool testing thanks!

frontend/src/lib/services/ckbtc-accounts-loader.services.ts
peterpeterparker: I use `get` in the service because we opted for `get` in a service in last PR 😉

NEW in this PR:
- `getCkBTCWithdrawalAccount` added some comments in code

Existing:
- `getWithdrawalAccount` is the minter canister function of the same name. documented in ic-js
- `getCkBTCAccount` is the function that extends the `getIcrcAccount` for ckBTC. Similar to `getSnsAccount`

There are two types of account currently:

- `Account` as an NNS-dapp entity
- `IcrcAccount`


we might want to refactor our type `Account` to something else but not in this PR I would say.

frontend/src/lib/components/accounts/CkBTCWithdrawalAccount.svelte
peterpeterparker: > new component CkBTCWithdrawalAccount (without meaningful content yet)

As described. Next PR that uses the button is ready but, cannot provides it before this PR goes through.

frontend/src/lib/components/accounts/CkBTCWithdrawalAccount.svelte
peterpeterparker: Exactly, correct

frontend/src/lib/components/accounts/CkBTCWithdrawalAccount.svelte
peterpeterparker: Ultimately, in a few PRs, we will also need reload - i.e. for the all features we need `load` and `reload`.
The card display an amount of ckBTC that got stuck in the minter. From the card, user will open the transaction modal from which the user can unblock these funds and from which we will have to reload the amount that is stuck or not anymore.

frontend/src/tests/lib/services/ckbtc-accounts-loader.services.spec.ts
peterpeterparker: It does not matter here. The service should just bubbles the error to let the caller know it failed.
Specific errors are handled by the services that are used by the service we are testing.

I added a comment next to the test.

frontend/src/tests/lib/services/ckbtc-withdrawal-accounts.services.spec.ts
peterpeterparker: Thanks! Typo. `getCkBTCAccount`

frontend/src/lib/services/ckbtc-accounts-loader.services.ts
lmuntaner: Ok, the problem is that I'm not familiar with the business logic of ckbtc and I'm a little lost sometimes.

Is there any documentation you use to understand or is all by talking to people?

Thanks for the explanation btw!

frontend/src/lib/components/accounts/CkBTCWithdrawalAccount.svelte
lmuntaner: ok, thanks!

lmuntaner (approved):
Thanks!

frontend/src/lib/services/ckbtc-accounts-loader.services.ts
peterpeterparker: mostly offline discussion indeed.
ic-js is documented: https://github.com/dfinity/ic-js/tree/main/packages/ckbtc
otherwise Figma file for the UI/UX and an Excel file for the transaction fee

PR #2432, opened 2023-05-02
refactor: extract getWithdrawalAccount to minter service

peterpeterparker:
# Motivation

We will need `getWithdrawalAccount` in PR #2430 for a new feature. Therefore this PR extract the dedicated service function that handle errors in the `minter.service`. That way we can avoid the try/catch/toastError.

Refactor  only.


lmuntaner (approved):
Thanks!

PR #2433, opened 2023-05-02
Make ckBTC ledger API return a single account

peterpeterparker:
# Motivation

We will need to fetch the balance of another pseudo ckBTC account - the withdrawal account - in PR #2430.
That's why this PR makes the ckBTC ledger API returns a single account and extract the logic of fetching the main account into a new service. That way we can in the other PR add a new function that will fetch the account using the same API..

Refactor  only.

### DIFF
--- a/frontend/src/lib/services/ckbtc-accounts-loader.services.ts
+++ b/frontend/src/lib/services/ckbtc-accounts-loader.services.ts
@@ -1,0 +1,29 @@
+import { getCkBTCAccount } from "$lib/api/ckbtc-ledger.api";
+import type { Account, AccountType } from "$lib/types/account";
+import type { Identity } from "@dfinity/agent";
+import type { Principal } from "@dfinity/principal";
+
+export const loadCkBTCAccounts = async ({
+  identity,
+  certified,
+  universeId,
+}: {
+  identity: Identity;
+  certified: boolean;
+  universeId: Principal;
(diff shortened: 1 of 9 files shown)


frontend/src/lib/services/ckbtc-accounts-loader.services.ts
```diff
@@ -0,0 +1,29 @@
+import { getCkBTCAccount } from "$lib/api/ckbtc-ledger.api";
+import type { Account, AccountType } from "$lib/types/account";
+import type { Identity } from "@dfinity/agent";
+import type { Principal } from "@dfinity/principal";
+
+export const loadCkBTCAccounts = async ({
```
lmuntaner, line 6: This doesn't load anything in a svelte store. Shall we rename this `getCkBTCAccounts`?

I still like it as a service, I guess it will do multiple calls?

lmuntaner (approved):
Minor renaming suggestion.

Already approving

frontend/src/lib/services/ckbtc-accounts-loader.services.ts
peterpeterparker: I like the idea, will do tomorrow before merging it! Thanks for the share.

peterpeterparker: Done

PR #2434, opened 2023-05-03
E2E: See SNS project details

dskloetd:
# Motivation

Continuing with SNS participation end-to-end testing.

# Changes

1. In the test, click on the open project card and check some details.
2. Add needed `[data-tid]` attributes.
3. Add page objects.

# Tests

`npm run test-e2e`

### DIFF
--- a/frontend/src/tests/e2e/sns-participation.spec.ts
+++ b/frontend/src/tests/e2e/sns-participation.spec.ts
@@ -31,7 +31,13 @@ test("Test SNS participation", async ({ page, context }) => {
   expect(committedProjects.length).toBeGreaterThan(1);
 
   // D003: User can see the details of one sale
-  // TODO
+  const snsProjectName = await openProjects[0].getProjectName();
+  await openProjects[0].click();
+  const projectDetail = appPo.getProjectDetailPo();
+  await projectDetail.waitForContentLoaded();
+  const tokenSymbol = await projectDetail.getTokenSymbol();
+  expect("Project " + tokenSymbol).toBe(snsProjectName);
+  expect(await projectDetail.getStatus()).toBe("Accepting Participation");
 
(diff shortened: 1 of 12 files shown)


frontend/src/tests/e2e/sns-participation.spec.ts
```diff
@@ -31,7 +31,13 @@ test("Test SNS participation", async ({ page, context }) => {
   expect(committedProjects.length).toBeGreaterThan(1);
 
   // D003: User can see the details of one sale
-  // TODO
+  const snsProjectName = await openProjects[0].getProjectName();
+  await openProjects[0].click();
+  const projectDetail = appPo.getProjectDetailPo();
+  await projectDetail.waitForContentLoaded();
+  const tokenSymbol = await projectDetail.getTokenSymbol();
+  expect("Project " + tokenSymbol).toBe(snsProjectName);
```
lmuntaner, line 39: Projects have a name and the token has also a symbol.

`Project + {name}`. Maybe this works because the token symbol and the project name are the same?

lmuntaner (commented):
One question

frontend/src/tests/e2e/sns-participation.spec.ts
dskloetd: Yes, this works because that's how it's generated by snsdemo.

lmuntaner: Does it work because the name of the project and the symbol of the token are the same?

Should we rely on this?

dskloetd: I changed it so that now it compares the project name on on the launchpad to the project name on the project detail. And for the token symbol it just expects it to not be empty.

lmuntaner (approved):
Thanks!

PR #2437, opened 2023-05-03
Chore: Persist Sns Proposal Filters

lmuntaner:
# Motivation

Users commonly use the same filters for proposals.

Persist the selection of the filters in local storage.

# Changes

* Create the sns filters store with our `writableStored` instead of `writable` from Svelte.

# Tests

Same functionality. Maybe there could be an end-to-end test for this in the future?


mstrasinskis (approved):
LGTM

PR #2439, opened 2023-05-03
E2E: Participate in a swap

dskloetd:
# Motivation

Continue with the participation end-to-end test.

# Changes

1. Change the e2e test to get tokens, participate, and check the commitment.
2. Add necessary test IDs.
3. Add necessary functionality to page objects.
4. Add necessary new page objects.

# Tests

`npm run test-e2e -- participation`
Manually tested that the participation button and modal still look normal.


frontend/src/tests/e2e/sns-participation.spec.ts
```diff
@@ -41,7 +49,9 @@ test("Test SNS participation", async ({ page, context }) => {
   expect(await projectDetail.getStatus()).toBe("Accepting Participation");
 
   // D004: User can participate in a sale
-  // TODO
+  expect(await projectDetail.getCommitmentAmount()).toBe("0");
```
lmuntaner, line 52: This is a bug, it should not be visible.

It should return `undefined` because the component should not be there.

Can you add a todo and I'll fix it.

Thanks!

lmuntaner (approved):
Thanks!

frontend/src/tests/e2e/sns-participation.spec.ts
dskloetd: Done

PR #2440, opened 2023-05-04
Add option to select destination address methods

peterpeterparker:
# Motivation

We have the need to use the transaction form with the destination address entered as `input` only - no `dropdown` regardless of the status of the accounts.

This PR adds an option to either display the dropdown/input as currently (mode `all`) or one or the other.


lmuntaner (approved):
LGTM!

PR #2442, opened 2023-05-04
Transaction network readonly

peterpeterparker:
# Motivation

We need an option to display the selected network in the transaction form as readonly because converting ckBTC from the remaining balance of the withdrawal account should happens on the BTC network only.

# Changes

- add option `networkReadonly` to transaction init parameters
- new component that display either the network as readonly or the existing dropdown

### DIFF
--- a/frontend/src/lib/types/transaction.ts
+++ b/frontend/src/lib/types/transaction.ts
@@ -12,6 +12,7 @@ export interface TransactionInit {
   sourceAccount?: Account;
   destinationAddress?: string;
   mustSelectNetwork?: boolean;
+  networkReadonly?: boolean;
   amount?: number;
 }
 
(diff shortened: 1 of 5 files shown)


frontend/src/lib/types/transaction.ts
lmuntaner, line 1: Thanks for these explanations!

frontend/src/lib/components/transaction/TransactionFormNetwork.svelte
lmuntaner, line 1: This renders a dropdown or a network title, right?

The name is a little confusing. How about adding `Item`? `TransactionFormItemNetwork`

That this is a form item. Not specifying which one.

lmuntaner (approved):
Thanks, LGTM, just a request on one name.

frontend/src/lib/components/transaction/TransactionFormNetwork.svelte
peterpeterparker: Sure, done.

PR #2445, opened 2023-05-04
queuedStore to mitigate queryAndUpdate race condition

dskloetd:
# Motivation

There is a race condition with queryAndUpdate. If you call queryAndUpdate again before the "update" from the first queryAndUpdate responds, then when the first update does respond, it can overwrite responses from the second queryAndUpdate with old data.
This is causing end-to-end tests to be flaky.

To mitigate this I implemented a "store wrapper" which keeps the pending mutations in a queue. Then if an older mutation responds later, it is placed in the correct place in the queue and all mutations are reapplied in the correct order.
By "correct" order, I mean the order in which the original queryAndUpdate calls happened.
Whether this is actually the correct order, we can't know unless we get the block height in the response, which we currently don't.

To match responses from the same queryAndUpdate call, the caller is required to request a one-time use store and apply to it just the mutations corresponding to that queryAndUpdate.

# Changes

1. Implemented `queuedStore` which wraps a writable store and applies the queue mechanism described above.
2. Changed accountsStore to use queuedStore.
3. Updated accounts service to use the new accountsStore API.
4. Temporarily added methods for the original API of accountsStore for backwards compatibility until all existing tests have been updated.

# Tests

Added tests for queuedStore, accountsStore and accounts service.

### DIFF
--- a/frontend/src/lib/stores/queued-store.ts
+++ b/frontend/src/lib/stores/queued-store.ts
@@ -1,4 +1,6 @@
-import { derived, writable, type Readable } from "svelte/store";
+import type { QueryAndUpdateStrategy } from "$lib/services/utils.services";
+import { nonNullish } from "@dfinity/utils";
+import { writable, type Readable } from "svelte/store";
 
 // A "queued store" is Svelte store which is aware of the queryAndUpdate
 // mechanism. It keeps a queue of mutations which have been applied based on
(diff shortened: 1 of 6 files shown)


frontend/src/lib/stores/queued-store.ts
```diff
@@ -0,0 +1,192 @@
...
+  getSingleMutationStore: (
+    strategy?: QueryAndUpdateStrategy | undefined
+  ) => SingleMutationStore<StoreData>;
+  resetForTesting: () => void;
+}
+
+export const queuedStore = <StoreData>(
+  initialData: StoreData
+): QueuedStore<StoreData> => {
+  let certifiedData = initialData;
```
lmuntaner, line 63: This is only used in `updateExposedData`, is this right?

And `updateExposedData` is used only in `addMutation`.

Could we avoid having to set this variable?

Maybe by passing it as parameter? `updateExposedData(...)`?

lmuntaner (approved):
One comment, but LGTM, already approving. Thanks!!

frontend/src/lib/stores/queued-store.ts
dskloetd: I'm not sure.
Note that this value is not `const` and is updated on line 119.
Since the value needs to survive any function calls, similarly to `let mutationQueue` below, I'm not sure how we could do without.
But let me know if I missed something.
Happy to make additional changes after merging.

PR #2446, opened 2023-05-04
Add .gz to internet-identity wasm URL

dskloetd:
# Motivation

internet_identity stopped offering non-gzipped artifacts.
I'm not sure what this is used for but it seems to be (one of many reasons?) why the Nightly Publication workflow is failing.

# Changes

Add `.gz` to the internet_identity wasm URL in scripts/update-ii

# Tests

Ran the script and instead of causing an error, it updated `dfx.json`.


lmuntaner (approved):

peterpeterparker (approved):
Thx

PR #2447, opened 2023-05-04
GIX-1546: Integrate geo location service

lmuntaner:
# Motivation

SNSs projects might want to restrict participation in some countries.

To do so, the NNS Dapp needs to get the location of the user and compare that with the restriction list.

In this PR: I setup the service and store that will have the user data

# Changes

* New location api. It returns a hardcoded value for now. Once we know which service we need to use, we can implement it.
* New locationStore that has the location of the user as a country code.
* New location services that uses the api to load the store.
* New type `CountryCode` that we might want to improve. We might need to translate the country code to human readable and this type might help us. For now, this is just a string.

# Tests

* Test the new location service.
* Test the new location store.


frontend/src/lib/api/location.api.ts
```diff
@@ -0,0 +1,25 @@
...
+
+// TODO: Add to env vars?
+// const TOKEN = "XXXXXX";
+
+/**
+ * Fetches the user's country location using IPInfo API
+ *
+ * @returns {CountryCode}
+ */
+export const queryUserCountryLocation = async (): Promise<CountryCode> => {
```
peterpeterparker, line 11: what about moving `api/location.api.ts` to `rest/location.rest.ts` where we already have a service that uses fetch? or moving the existing one to api?

frontend/src/lib/api/location.api.ts
```diff
@@ -0,0 +1,25 @@
...
+  return "CH";
+  // TODO: We have two options:
+  // 1. Use the IPInfo API to get the user's country location
+  // 2. Use a service provided by the Boundary Nodes
+  // const response = await fetch(`https://ipinfo.io/json?token=${TOKEN}`);
```
peterpeterparker, line 16: if there is one, is that a token that can be exposed in the frontend code or its a the `TOKEN` that needs to be made hidden - i.e. needs to be fetched from a backend?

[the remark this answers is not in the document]
lmuntaner: Last update. We will most probably use this service. We'll need to enable it as soon as we have a token to enable the API.

frontend/src/lib/stores/location.store.ts
```diff
@@ -0,0 +1,4 @@
+import type { CountryCode } from "$lib/types/location";
+import { writable } from "svelte/store";
+
+export const locationStore = writable<CountryCode | undefined>(undefined);
```
mstrasinskis, line 4: How about using more specific name, like `userCountryLocationStore` or `userCountryCodeStore`? Not a strong opinion.

frontend/src/lib/api/location.api.ts
lmuntaner: Why do we want to differentiate between `api` and `rest`?

We also have the sns aggregator inside `api`.

frontend/src/lib/api/location.api.ts
lmuntaner: It will need to be hidden, but it can't be totally hiddne. The call needs the token and it needs to be from the frontend. Therefore, the token will be exposed in the Network tab anyway.

I'll bring it up with prod-sec again. But it seems this is the only option that is fast enough to be implemented.

frontend/src/lib/stores/location.store.ts
lmuntaner: Soudns good.

frontend/src/tests/lib/services/location.services.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    jest.clearAllMocks();
+  });
+
+  describe("loadUserLocation", () => {
+    it("should set the location store to api response", async () => {
+      const countryCode = "CH";
+      jest
+        .spyOn(locationApi, "queryUserCountryLocation")
+        .mockResolvedValue(countryCode);
+
```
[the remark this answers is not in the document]
lmuntaner: sure

frontend/src/lib/api/location.api.ts
peterpeterparker: > Why do we want to differentiate between api and rest?

Because rest do not use any agent-js / actor / candid / etc. things and are just plain json fetch queries that uses just the platform and can therefore be even move elsewhere without any dependencies

> We also have the sns aggregator inside api.

Probably did not noticed it.

For above reason, I rather like to have those types in a dedicated package. That said, if you rather like to have everything in api, no strong opinion. Just move what's in rest to api afterwards then.

frontend/src/lib/api/location.api.ts
peterpeterparker: What are the guidelines of ipinfo?

According my experience, it's the service provider that defines how or if its token must be saved in a backend or not.

frontend/src/lib/api/location.api.ts
lmuntaner: Let's discuss in the weekly call.

I'm not sure the implementation of the function should tell us where it goes, but rather the concept itself.

peterpeterparker: Like I said no strong opinion worth discussing this further. Go for refactoring then.

frontend/src/lib/api/location.api.ts
lmuntaner: The token needs to be passed as query param or bearer token

Docs: https://ipinfo.io/developers

> Your API token is used to authenticate you with our API and can be provided either as an HTTP Basic Auth username, a bearer token, or alternatively as a token URL parameter.

> Alternatively, if you do not pass in any IP address, we'll return details for the calling address. This allows you to look up your own (or a visitor to your site) IP address details without knowing the IP address in advance.

> JSONP and CORS are supported, allowing you to use ipinfo.io entirely in client-side code

They don't specify a way to store the token as far as I know

frontend/src/lib/api/location.api.ts
lmuntaner: I added it to the agenda for today.

peterpeterparker: No need, let's refactor. Thanks.

frontend/src/lib/api/location.api.ts
peterpeterparker: All good then, thanks for double checking 👍

PR #2450, opened 2023-05-05
Don't expose syncAccounts error handler

dskloetd:
# Motivation

`syncAccounts` has an errorHandler parameter but this is only used by `initAccounts` in the same file.
By limiting the functionality we export out of the file it becomes easier to refactor without breaking anything.

# Changes

1. Add `syncAccountsWithErrorHandler` (which is not exported) and called by both `initAccounts` and `syncAccounts`.
2. Remove the optional errorHandler parameter from `syncAccounts`.
3. Remove the tests that test this parameter.

# Tests

Unused functionality is removed so the corresponding tests are removed.
`initAccounts` is still tested separately as it was before.

### DIFF
--- a/frontend/src/lib/services/accounts.services.ts
+++ b/frontend/src/lib/services/accounts.services.ts
@@ -153,8 +153,8 @@ const defaultErrorHandlerAccounts: SyncAccontsErrorHandler = ({
 /**
  * Loads the account data using the ledger and the nns dapp canister.
  */
-export const syncAccounts = (
-  errorHandler: SyncAccontsErrorHandler = defaultErrorHandlerAccounts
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
LGTM, thanks!

frontend/src/lib/services/accounts.services.ts
```diff
@@ -169,14 +169,17 @@ export const syncAccounts = (
   });
 };
 
+export const syncAccounts = () =>
```
peterpeterparker, line 172: it returns a promise I think, might be worth specifying the return type?

```
export const syncAccounts = (): Promise<void> =>
```

PR #2454, opened 2023-05-08
Convert ckBTC to BTC for withdrawal account

peterpeterparker:
# Motivation

Convert ckBTC to BTC for withdrawal account.

- execute `retrieveBTC` only (no transfer to ledger) if withdrawal account is used for the transaction
- reload withdrawal account after conversion (regardless of the selected source account)

# Changes

- split existing `retrieveBTC` logic from service `convertCkBTCToBTC` to a separate function (refactoring)
- call this particular function if withdrawal account is selected
- in addition to reloading the transaction, reload the withdrawal account as well after execution
- UI wise, do not diplsay the step "transfer to ledger" in progress modal if withdrawal account is the source
- as reload is executed in the service, there is ultimately no need of the reload account function in the component therefore code refactored


lmuntaner (approved):
Thanks! LGTM!

PR #2455, opened 2023-05-08
Script to start dfx with snapshot

dskloetd:
# Motivation

We have `scripts/dev-local-state.sh` which starts a replica (`dfx start`) *and* the frontend (`npm run dev`) with a given snapshot. I originally thought this would be convenient but it hasn't been for 2 reasons:
1. `npm ci` often causes the vite server to kill itself. This then makes the script also kill the replica, which is unnecessary.
2. Some scripts (e.g. in snsdemo) run a replica and start by killing any existing running replicas. Because `dev-local-state.sh` runs the replica in the background, it doesn't notice that it's killed and then it doesn't get an opportunity to restore the state.

# Changes

Add `scripts/dfx-snapshot-start` which is just like [dev-local-state.sh](https://github.com/dfinity/nns-dapp/blob/main/scripts/dev-local-state.sh) but it runs `dfx start` in the foreground instead of the background, and doesn't run `npm run dev`.

# Tests

Ran the script manually.

### DIFF
--- a/scripts/dfx-snapshot-start
+++ b/scripts/dfx-snapshot-start
@@ -46,7 +46,7 @@ restore_backup() {
   sh "$RESTORE_BACKUP_SCRIPT_FILE"
 }
 
-# Restore the backup if the script fails or exists successfully.
+# Restore the backup if the script fails or exits successfully.
 trap restore_backup EXIT
 
 CLEAN_ARG=""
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
LGTM! One question,

Should we keep `/scripts/dev-local-state.sh` or remove it?

dskloetd:
Good question.
I'm not against removing or keeping it but I don't know if anyone uses it.
There are also still `dev-local.sh` and `dev-testnet.sh`, which both also start `npm run dev`.
I think they might be at least potentially useful as documentation for how to get a complete setup if you don't remember all the steps and want to do them separately.

PR #2456, opened 2023-05-08
No reload of the transactions for withdrawal account

peterpeterparker:
# Motivation

When ckBTC are converted to BTC from the withdrawal account, it does not make much sense to reload the related transactions as these are nowhere used in NNS-dapp.

# Changes

- on `retrieveBTC` from withdrawal account, skip reload of the transactions.

### DIFF
--- a/frontend/src/lib/services/ckbtc-accounts-loader.services.ts
+++ b/frontend/src/lib/services/ckbtc-accounts-loader.services.ts
@@ -78,6 +78,17 @@ export const getCkBTCWithdrawalAccount = async ({
     throw new Error(no_minter_defined);
   }
 
+  const store = get(ckBTCWithdrawalAccountsStore);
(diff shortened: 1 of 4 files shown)


frontend/src/lib/services/ckbtc-accounts-loader.services.ts
```diff
@@ -78,6 +78,17 @@ export const getCkBTCWithdrawalAccount = async ({
     throw new Error(no_minter_defined);
   }
 
+  const store = get(ckBTCWithdrawalAccountsStore);
```
peterpeterparker, line 81: Not related to this PR and just refactoring. I moved this inner function few lines down because it makes reading the function easier.

lmuntaner: Thanks for the explanation

PR #2458, opened 2023-05-08
Cleanup: Remove old methods from accountsStore

dskloetd:
# Motivation

https://github.com/dfinity/nns-dapp/pull/2445 introduced a new interface for the `accountsStore`.
It kept the old interface temporarily for backwards compatibility so we didn't have to update all the existing tests in the same PR.
This PR removes the old interface.

# Changes

1. Remove `set`, `setBalance` and `reset` from `accountsStre`.
2. Add `setForTesting` to be used by tests only.
3. Update tests to use `setForTesting` and `resetForTesting`.
4. (Re)move a few occasional cleanups from tests/afterEach to beforeEach.

# Tests

`npm run test`


lmuntaner (approved):
Thanks!

PR #2460, opened 2023-05-08
ckBTC minter canister id on mainnet

peterpeterparker:
# Motivation

Set the effective ckBTC minter canister id on mainnet.


lmuntaner (approved):
Thanks!

PR #2461, opened 2023-05-08
Enable conversion of ckBTC from / to BTC

peterpeterparker:
# Motivation

Enable conversion of ckBTC from / to BTC by solving various TODO that used to make various features accessible only ckTESTBTC.


lmuntaner (approved):
Thanks!

PR #2462, opened 2023-05-08
Always print asset hashes

bitdivine:
# Motivation
If a PR makes no changes to the docker build, the hash of the build artefacts is not available.  This is inconvenient when checking whether CI still produces the same outputs as local builds.

# Changes
Always print the sha256 hashes of docker outputs.

# Tests
See CI


bitdivine:
Screenshot from 2023-05-08 15-30-28

PR #2469, opened 2023-05-09
GIX-1546: Integrate geo location service

lmuntaner:
# Motivation

This PR: Integrate with two location services that return the user location.

# Changes

* Implement two services in location api function.
* Add CSP rules for both new services.

# Tests

* Add test for location api function.

### DIFF
--- a/frontend/src/lib/components/project-detail/ProjectStatusSection.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectStatusSection.svelte
@@ -12,7 +12,7 @@
     PROJECT_DETAIL_CONTEXT_KEY,
     type ProjectDetailContext,
   } from "$lib/types/project-detail.context";
-  import { isNullish } from "@dfinity/utils";
+  import { isNullish, nonNullish } from "@dfinity/utils";
   import { SnsSwapLifecycle } from "@dfinity/sns";
   import ParticipateButton from "./ParticipateButton.svelte";
   import { getCommitmentE8s } from "$lib/utils/sns.utils";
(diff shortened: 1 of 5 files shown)


frontend/src/lib/components/project-detail/ProjectStatusSection.svelte
```diff
@@ -68,7 +68,7 @@
     </div>
 
     <div class="actions content-cell-details">
-      {#if myCommitmentIcp !== undefined}
+      {#if nonNullish(myCommitmentIcp) && myCommitmentIcp.toE8s() > BigInt(0)}
```
dskloetd, line 71: Did this get mixed in from another PR?

frontend/src/tests/lib/components/project-detail/ProjectStatusSection.spec.ts
```diff
@@ -42,6 +42,31 @@ describe("ProjectStatusSection", () => {
     expect(queryByTestId("sns-project-current-commitment")).toBeInTheDocument();
   });
 
+  it("should render user commitment", () => {
```
dskloetd, line 45: This doesn't belong in this PR, right?

frontend/src/lib/components/project-detail/ProjectStatusSection.svelte
lmuntaner: No, it doesn't. I must have started this branch from the other PR.

Let me close this and open a new one. Thanks!

lmuntaner:
Branch not from `main`

PR #2470, opened 2023-05-09
GIX-1546: Integrate Geo Location Services

lmuntaner:
# Motivation

This PR: Integrate with two location services that return the user location.

# Changes

* Implement two services in location api function.
* Add CSP rules for both new services.

# Tests

* Add test for location api function.


lmuntaner:
@dskloetd please review

dskloetd (approved):

PR #2472, opened 2023-05-10
GIX-1541: Get user location if needed

lmuntaner:
# Motivation

In this PR: Load the location of the user when needed.

Note: Not fully functional yet because the backend is missing the necessary field.

# Changes

* New project util `userCountryIsNeeded`.
* Use new project util in ProjectDetail to call the service `loadUserCountry`.

# Tests

* Add tests for new project util
* Add test in ProjectDetail for new functionality.

### DIFF
--- a/frontend/src/lib/getters/sns-summary.ts
+++ b/frontend/src/lib/getters/sns-summary.ts
@@ -1,0 +1,4 @@
+import type { SnsSummary } from "$lib/types/sns";
+
+// TODO: GIX-1545 Implement this function
+export const getDenyList = (_summary: SnsSummary) => [];
--- a/frontend/src/lib/getters/sns-summary.ts
+++ b/frontend/src/lib/getters/sns-summary.ts
@@ -1,4 +1,4 @@
 import type { SnsSummary } from "$lib/types/sns";
 
 // TODO: GIX-1545 Implement this function
-export const getDenyList = (_summary: SnsSummary) => [];
+export const getDeniedCountries = (_summary: SnsSummary) => [];
--- a/frontend/src/lib/pages/ProjectDetail.svelte
+++ b/frontend/src/lib/pages/ProjectDetail.svelte
@@ -43,6 +43,8 @@
   import { browser } from "$app/environment";
   import { IS_TEST_ENV } from "$lib/constants/mockable.constants";
   import { authSignedInStore } from "$lib/derived/auth.derived";
+  import { userCountryIsNeeded } from "$lib/utils/projects.utils";
+  import { loadUserCountry } from "$lib/services/user-country.services";
 
   export let rootCanisterId: string | undefined | null;
 
@@ -159,6 +161,16 @@
         $projectDetailStore.swapCommitment = undefined;
       },
     });
+  }
+
+  let shouldLoadUserCountry = false;
+  $: shouldLoadUserCountry = userCountryIsNeeded({
+    summary: $projectDetailStore?.summary,
+    swapCommitment: $projectDetailStore?.swapCommitment,
+    loggedIn: $authSignedInStore,
+  });
+  $: if (shouldLoadUserCountry) {
+    loadUserCountry();
   }
 
   let unsubscribeWatchCommitment: () => void | undefined;
--- a/frontend/src/lib/utils/projects.utils.ts
+++ b/frontend/src/lib/utils/projects.utils.ts
@@ -1,4 +1,5 @@
 import type { SnsFullProject } from "$lib/derived/sns/sns-projects.derived";
+import { getDenyList } from "$lib/getters/sns-summary";
 import type {
   SnsSummary,
   SnsSummarySwap,
@@ -182,10 +183,7 @@ export const userCountryIsNeeded = ({
   canUserParticipateToSwap({ summary, swapCommitment }) &&
   loggedIn &&
   nonNullish(summary) &&
-  // TODO: GIX-1541 Remove unncessary checking and casting of deny_list
-  "deny_list" in summary.swap &&
-  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
-  (summary.swap.deny_list as any).length > 0;
+  getDenyList(summary).length > 0;
 
 export const hasUserParticipatedToSwap = ({
   swapCommitment,
--- a/frontend/src/lib/utils/projects.utils.ts
+++ b/frontend/src/lib/utils/projects.utils.ts
@@ -1,5 +1,5 @@
 import type { SnsFullProject } from "$lib/derived/sns/sns-projects.derived";
-import { getDenyList } from "$lib/getters/sns-summary";
+import { getDeniedCountries } from "$lib/getters/sns-summary";
 import type {
   SnsSummary,
   SnsSummarySwap,
@@ -183,7 +183,7 @@ export const userCountryIsNeeded = ({
   canUserParticipateToSwap({ summary, swapCommitment }) &&
   loggedIn &&
   nonNullish(summary) &&
-  getDenyList(summary).length > 0;
+  getDeniedCountries(summary).length > 0;
 
 export const hasUserParticipatedToSwap = ({
   swapCommitment,
--- a/frontend/src/lib/utils/projects.utils.ts
+++ b/frontend/src/lib/utils/projects.utils.ts
@@ -6,6 +6,7 @@ import type {
 } from "$lib/types/sns";
 import type { TokenAmount } from "@dfinity/nns";
 import { SnsSwapLifecycle } from "@dfinity/sns";
+import { nonNullish } from "@dfinity/utils";
 import { nowInSeconds } from "./date.utils";
 import type { I18nSubstitutions } from "./i18n.utils";
 import { getCommitmentE8s } from "./sns.utils";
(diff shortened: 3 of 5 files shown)


frontend/src/lib/pages/ProjectDetail.svelte
```diff
@@ -161,6 +163,16 @@
...
   }
 
+  let shouldLoadUserCountry = false;
+  $: shouldLoadUserCountry = userCountryIsNeeded({
+    summary: $projectDetailStore?.summary,
+    swapCommitment: $projectDetailStore?.swapCommitment,
+    loggedIn: $authSignedInStore,
+  });
+  $: if (shouldLoadUserCountry) {
+    loadUserCountry();
```
lmuntaner, line 173: This willl not be called until `deny_list` is added,

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -287,6 +295,19 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
         ).toMatch(formatToken({ value: userCommitment }));
       });
 
+      // TODO: GIX-1541 use this test to show that button is disabled if user is in a country that is not allowed to participate
+      it("should NOT load user's country", async () => {
```
lmuntaner, line 299: I want to keep this test to make sure the service is not yet used.

frontend/src/lib/utils/projects.utils.ts
```diff
@@ -158,6 +159,34 @@ export const canUserParticipateToSwap = ({
   );
 };
 
+/**
+ * Returns whether the location of the user is needed to participate to the swap.
+ *
+ * Some projects might restrict participation in the swap to users from specific countries.
+ *
+ * Yet, we don't need to get the location of all users, only those who can to participate to the swap:
```
dskloetd, line 167: "only those who can to participate to the swap"
Is there a typo there? I'm not sure how to parse this.

frontend/src/lib/utils/projects.utils.ts
```diff
@@ -158,6 +159,34 @@ export const canUserParticipateToSwap = ({
...
+  swapCommitment: SnsSwapCommitment | undefined | null;
+  loggedIn: boolean;
+}): boolean =>
+  canUserParticipateToSwap({ summary, swapCommitment }) &&
+  loggedIn &&
+  nonNullish(summary) &&
+  // TODO: GIX-1541 Remove unncessary checking and casting of deny_list
+  "deny_list" in summary.swap &&
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  (summary.swap.deny_list as any).length > 0;
```
dskloetd, line 188: Can you cast it to `string[]` instead of to `any`?

frontend/src/tests/lib/utils/projects.utils.spec.ts
```diff
@@ -272,6 +273,111 @@ describe("project-utils", () => {
...
 
+  describe("userCountryIsNeeded", () => {
+    it("country not needed if no summary or swap information", () => {
+      expect(
+        userCountryIsNeeded({
+          summary: undefined,
+          swapCommitment: undefined,
+          loggedIn: true,
+        })
+      ).toBeFalsy();
```
dskloetd, line 284: Just curious, why `toBeFalsy` instead of the stricter `toBe(false)`? Do you expect it to potentially return null or undefined?

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
dskloetd: Is it possible to also have a test where the user's country is loaded?
Otherwise it's hard to be confident that this test sets up the right conditions where the country would be loaded. If all you test is that something doesn't happen, the test also passes if the setup is completely broken and nothing happens.

frontend/src/tests/lib/utils/projects.utils.spec.ts
```diff
@@ -272,6 +273,111 @@ describe("project-utils", () => {
...
+      expect(
+        userCountryIsNeeded({
+          summary: summaryForLifecycle(SnsSwapLifecycle.Aborted),
+          swapCommitment: mockSwapCommitment,
+          loggedIn: true,
+        })
+      ).toBeFalsy();
+    });
+
+    // TODO: GIX-1541 Add a project with deny_list and test that returns true
```
dskloetd, line 342: If you do some casting, similar to the tested code, you could have a test that returns true, right?

frontend/src/lib/utils/projects.utils.ts
```diff
@@ -158,6 +159,34 @@ export const canUserParticipateToSwap = ({
...
+  loggedIn,
+}: {
+  summary: SnsSummary | undefined | null;
+  swapCommitment: SnsSwapCommitment | undefined | null;
+  loggedIn: boolean;
+}): boolean =>
+  canUserParticipateToSwap({ summary, swapCommitment }) &&
+  loggedIn &&
+  nonNullish(summary) &&
+  // TODO: GIX-1541 Remove unncessary checking and casting of deny_list
```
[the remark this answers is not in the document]
dskloetd: Looking at [SnsSummary](https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/types/sns.ts#L42), am I right to assume that the denied countries list will most likely be on the [SnsParams](https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/types/sns.ts#L37) in the `SnsSummarySwap`?

frontend/src/lib/utils/projects.utils.ts
lmuntaner: We only need the location of the users that can participate in the swap.

Does that help?

Is "swap" the problem here?

frontend/src/lib/utils/projects.utils.ts
lmuntaner: It won't be `string[]`. It's not yet clear the format.

frontend/src/tests/lib/utils/projects.utils.spec.ts
lmuntaner: No, I was thinking on making it easier to read and it's what we have also on another util.

No strong opinion.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
lmuntaner: good idea, done!

frontend/src/tests/lib/utils/projects.utils.spec.ts
lmuntaner: done

frontend/src/tests/lib/utils/projects.utils.spec.ts
lmuntaner: I used the "getter" idea and I mock it when I want to test a different value than empty string.

Let me know what you think.

frontend/src/lib/utils/projects.utils.ts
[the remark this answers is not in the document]
lmuntaner: I used the getter idea.

dskloetd (commented):
GitHub won't let me submit my review without typing something here.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -287,6 +296,32 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
...
+          rootCanisterId: Principal.fromText(rootCanisterId),
+          myCommitment: undefined,
+        } as SnsSwapCommitment);
+
+        render(ProjectDetail, props);
+
+        expect(locationApi.queryUserCountryLocation).not.toBeCalled();
+      });
+
+      it("should load user's country if no deny list", async () => {
```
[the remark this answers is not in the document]
lmuntaner: yes, sorry, thanks!

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -287,6 +296,32 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
         ).toMatch(formatToken({ value: userCommitment }));
       });
 
+      it("should not load user's country if no deny list", async () => {
```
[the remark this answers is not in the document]
lmuntaner: Which tests together?

* should not load user's country if no deny list
* should load user's country if non-empty deny list

or

* should load user's country if non-empty deny list
* should participate without user interaction if there is an open ticket.

This is already nested inside three describes, that's why I wasn't sure about nesting one further.

frontend/src/lib/getters/sns-summary.ts
```diff
@@ -0,0 +1,4 @@
+import type { SnsSummary } from "$lib/types/sns";
+
+// TODO: GIX-1545 Implement this function
+export const getDenyList = (_summary: SnsSummary) => [];
```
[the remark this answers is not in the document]
lmuntaner: sure

frontend/src/tests/lib/utils/projects.utils.spec.ts
```diff
@@ -272,6 +274,115 @@ describe("project-utils", () => {
...
+          summary: summaryForLifecycle(SnsSwapLifecycle.Aborted),
+          swapCommitment: mockSwapCommitment,
+          loggedIn: true,
+        })
+      ).toBe(false);
+    });
+
+    it("country is needed", () => {
+      // TODO: TODO: GIX-1545 Remove mock and create a summary with deny list
+      jest.spyOn(summaryGetters, "getDenyList").mockReturnValue(["US"]);
```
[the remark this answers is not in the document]
lmuntaner: Do you mean to add two describe?

The others require the deny list to be empty. If I mock it for them returning something they will fail.

I'm not sure what you mean.

Since you brought up this pattern of putting the mock in a beforeEach before. Can you describe an objective rule of when it should be there and when not?

It would be nice to be consistent in all tests also.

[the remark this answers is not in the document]
dskloetd: > The others require the deny list to be empty.

Do they? I think in every other test it will not even call `getDeniedCountries()`.
But if they accidentally do call `getDeniedCountries()` we don't want *that* to be the reason it returns false.
For example in `"country not needed if sale is not open"` we want the fact that the sale is not open to be the reason it returns false. So every other potential reason should be set to true.

I also now realized that a test seems to be missing where every other reason is true but the getDeniedCountries() list is not present. Or am I over looking it?

Basically you are testing a function like this:
`getCondition(a, b, c, d) => a && b && c && d`
Then I would expect tests where
* a, b and c are true but d is false
* a, b, and d are true but c is false
* a, c and d are true but d is false, and
* b, c and d are true but a is false

So in the beforeEach I would set a, b, c and d all to true and then in each of the tests set 1 thing to false.
Does that make sense?

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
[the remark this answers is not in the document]
lmuntaner: Ah, because they both share the setup for `getOpenTicket` and `querySnsSwapCommitment`. Got it.

frontend/src/tests/lib/utils/projects.utils.spec.ts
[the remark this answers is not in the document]
lmuntaner: Yes, that makes sense.

One thing I'm not sure is overriding mocks. I don't like the idea that a specific test overrides the setup.

But this won't apply here. So all good.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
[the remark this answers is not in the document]
lmuntaner: done

lmuntaner:

dskloetd (approved):
Thanks!

PR #2473, opened 2023-05-10
fix: reload proposals on sns switch

mstrasinskis:
# Motivation

Fixes the bug when there is no sns proposals shown on the project switch.

# Changes

- add projectId listener for reload function

# Screenshot

Screen Recording 2023-05-10 at 11 16 26

### DIFF
--- a/frontend/src/lib/pages/SnsProposals.svelte
+++ b/frontend/src/lib/pages/SnsProposals.svelte
@@ -60,7 +60,11 @@
     }
   };
 
-  $: onSnsFiltersChanged($snsFiltersStore);
+  // Fetch the proposals only on filters or project change.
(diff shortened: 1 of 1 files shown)


frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -60,7 +60,11 @@
     }
   };
 
-  $: onSnsFiltersChanged($snsFiltersStore);
+  // Fetch the proposals only on filters or project change.
+  // TODO(e2e): cover this with e2e tests.
```
lmuntaner, line 64: Or even integration tests.

lmuntaner (approved):
Thanks!

PR #2474, opened 2023-05-10
GIX-1567 Change checkbox in participation-review screen

dskloetd:
# Motivation

https://dfinity.atlassian.net/browse/GIX-1567

# Changes

1. Change the order of the label and checkbox to put the checkbox on the left.
2. Change the checkbox label text.
3. Remove the warning text.

Old:
image

New:
image

# Tests

Tested manually


lmuntaner (changes requested):
The idea was that the checkbox was aligned with the first line. Not in the middle.

Not exactly, but now I see how the screenshot in the issue was not very clear about it and it was missing in the description.

Screenshot 2023-05-10 at 11 53 41

peterpeterparker:
Maybe worth also enhancing of few pixels the gap? two columns look narrowed

dskloetd:
How about this? This doubles the gap between the checkbox and removes the `align-items: center` style.

image

dskloetd:
Mary and Artem asked me to also left align the checkbox.
It makes the hover style a bit strange but we do the same for, for example, auto stake maturity and join community fund:
```
--checkbox-padding: var(--padding) 0;
```

I used also increased the gap (already in the previous screenshot)
```
--padding: var(--padding-2x);
```
because I agreed with David that the checkbox and text were quite close together.

Now it looks like this:

image

Please review again. Thanks!

lmuntaner (approved):
LGTM! Thanks

PR #2477, opened 2023-05-10
A canister explorer type

peterpeterparker:
# Motivation

Small typo (kind of).


lmuntaner (approved):
This better be used with `replacePlaceholders` in the future, no?

Otherwise, other languages might not have it at the end.

peterpeterparker:
> This better be used with replacePlaceholders in the future, no?

Replace place holder + HTML because it parses a link would be an option indeed.

PR #2478, opened 2023-05-10
Add aggregator to the docs

bitdivine:
# Motivation
The aggregator is not mentioned in the README.  In particular, I became aware while writing an sns deploy script that the flavours of the aggregator build are not documented anywhere.

# Changes
* Add the sns aggregator to the README.
I actually expect that we would prefer to have the aggregator documentation in a  separate README but let's discuss first.

# Tests
- [x] Readability check: Do the humans reading this PR like it?


dskloetd (approved):
This file is small enough not to have to split it.
Once we have more documentation we can consider splitting.
Or did you already have more in mind?

PR #2487, opened 2023-05-11
Reload all ckBTC transactions after update balance

peterpeterparker:
# Motivation

I transferred new bitcoin on mainnet and despite the fact that load next transactions was effectivelly called after update balance, the new transactions were not displayed. I'm not sure if it's a timing issue or not that's why I though about reload all transactions after such process.

# Changes

- if account change, reload all transactions of the account

# Screenshots

Load next transactions happen but no new transaction displayed

noreload

Here locally update balance simulated to display the all reload

reload

### DIFF
--- a/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
@@ -40,7 +40,9 @@
 
     loading = true;
 
-    // We want to reload all transactions of the account
+    // We want to reload all transactions of the account from scratch.
+    // That way the skeletons will be displayed again which provides the user a visual feedback about the fact that all transactions are realoded.
+    // This is handy because the reload notably happens the "update balance" process - i.e. happens after the "busy spinner" has fade away.
     icrcTransactionsStore.resetAccount({
       universeId,
       accountIdentifier: account.identifier,
--- a/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
@@ -1,6 +1,9 @@
 <script lang="ts">
   import type { Account } from "$lib/types/account";
-  import { loadCkBTCAccountNextTransactions } from "$lib/services/ckbtc-transactions.services";
+  import {
+    loadCkBTCAccountNextTransactions,
+    loadCkBTCAccountTransactions,
+  } from "$lib/services/ckbtc-transactions.services";
   import type { IcrcTransactionData } from "$lib/types/transaction";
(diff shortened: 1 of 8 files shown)


frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
```diff
@@ -28,7 +32,32 @@
...
+  const reloadTransactions = async () => {
+    // If we are already loading transactions we do not want to double the calls
+    if (loading) {
+      return;
+    }
+
+    loading = true;
+
+    // We want to reload all transactions of the account
+    icrcTransactionsStore.resetAccount({
```
lmuntaner, line 44: Why is resetting necessary? Doesn't `loadCkBTCAccountTransactions` do the same regardless of the store data?

peterpeterparker: Let me improve the comment in the component, sorry for that.

When I developped I had to mock the transactions and I ended with duplicate transactions so I developed the reset and ultimately actually find the UX quite interesting because it displays the user something is happening after the spinner of the "update balance" has fade away.

lmuntaner: So now there is no point in storing the transactions in the store because they are never reused, right?

We could store them in a variable here and use that variable to check for next transactions.

It doesn't make sense to have a store if we reset it every time the user comes to the page.

peterpeterparker: Comment improved. Let me know if it's ok?

peterpeterparker: > It doesn't make sense to have a store if we reset it every time the user comes to the page.

1. The store is used elsewhere too.
2. It is not resetted every time the user comes to the pages. See code `if load === true return` and provided GIF

lmuntaner: I see `if loading`, but `loading = true` is set inside here also.

But yes, in the GIF it doesn't reload the transactions the second visit. So, I might be missing something.

lmuntaner (approved):

frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
peterpeterparker: > but loading = true is set inside here also.

Because `loading` is also used within `IcrcTransactionsList` to display loading information

peterpeterparker: Thanks fo rthe review!

PR #2491, opened 2023-05-11
Chore: Prepare for force query calls

lmuntaner:
# Motivation

To make sure we have a good UX when there are a lot of visitors we want to avoid update calls. There is a flag (`FORCE_CALL_STRATEGY`) that changes all query calls to "query".

In this PR: Use the flag also in SNS neurons.

# Changes

* Use `FORCE_CALL_STRATEGY` flag in sns neurons services.

# Tests

We don't have specific tests for this flag.


mstrasinskis (approved):
LGTM, thanks! We just need to be sure that this strategy will be not applied after some neuron changes.

PR #2493, opened 2023-05-11
Feat: Change user country store type

lmuntaner:
# Motivation

We need to handle three states in the country store: loading, error and success.

In this PR: We change the type of the data in the store to better manage the different states.

# Changes

* Change the type inside `userCountryStore`.
* Add two derived stores with the logic for the error and the loading state.
* New type `Country`.

# Tests

* Fix tests accordingly.
* Add tests for two new derived stores.

### DIFF
--- a/frontend/src/lib/services/sns-neurons.services.ts
+++ b/frontend/src/lib/services/sns-neurons.services.ts
@@ -76,7 +76,6 @@ export const syncSnsNeurons = async (
   rootCanisterId: Principal
 ): Promise<void> => {
   return queryAndUpdate<SnsNeuron[], unknown>({
-    strategy: FORCE_CALL_STRATEGY,
     request: ({ certified, identity }) =>
       querySnsNeurons({
         rootCanisterId,
--- a/frontend/src/lib/services/sns-neurons.services.ts
+++ b/frontend/src/lib/services/sns-neurons.services.ts
@@ -76,6 +76,7 @@ export const syncSnsNeurons = async (
   rootCanisterId: Principal
 ): Promise<void> => {
   return queryAndUpdate<SnsNeuron[], unknown>({
+    strategy: FORCE_CALL_STRATEGY,
     request: ({ certified, identity }) =>
       querySnsNeurons({
         rootCanisterId,
--- a/frontend/src/lib/services/user-country.services.ts
+++ b/frontend/src/lib/services/user-country.services.ts
@@ -1,10 +1,13 @@
 import { queryUserCountryLocation } from "$lib/api/location.api";
-import { userCountryStore } from "$lib/stores/user-country.store";
+import {
+  isUserCountryLoadedStore,
+  userCountryStore,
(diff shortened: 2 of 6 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/stores/user-country.store.ts
```diff
@@ -1,5 +1,21 @@
-import type { CountryCode } from "$lib/types/location";
-import { writable } from "svelte/store";
+import type { Country } from "$lib/types/location";
+import { derived, writable } from "svelte/store";
+
+/**
+ * - Loading: undefined
```
dskloetd, line 5: How about calling it "not loaded"? I feel like we don't really know if it's actually actively loading. Or otherwise we need a 4th state which is "not yet loading".

And shall we use an explicit value (a string `"not loaded"` or something) instead of using `undefined`. I think it might avoid some amount of confusion.

dskloetd (approved):

frontend/src/lib/stores/user-country.store.ts
lmuntaner: Sounds good.

lmuntaner: done, please review again.

I changed the derived store. I don't want a negative derived store. If we want to opposite using two negations is not readable.

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -76,7 +76,6 @@ export const syncSnsNeurons = async (
   rootCanisterId: Principal
 ): Promise<void> => {
   return queryAndUpdate<SnsNeuron[], unknown>({
-    strategy: FORCE_CALL_STRATEGY,
```
dskloetd, line 79: Did you mean to change this in this PR?

frontend/src/lib/services/user-country.services.ts
```diff
@@ -1,16 +1,20 @@
...
       return;
     }
     const countryCode = await queryUserCountryLocation();
-    userCountryStore.set(countryCode);
+    userCountryStore.set({ isoCode: countryCode });
   } catch (e: unknown) {
-    // TODO: Implement error handling
+    // Print the error to the console for debugging purposes
     console.error(e);
+    userCountryStore.set(new Error("Error loading user country"));
```
dskloetd, line 18: I think I'm missing a test where the api function (mock) throws an error and we test that the error store returns true.
If I overlooked it, can you point me to it?

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: No, I rebased and I see that it undid some changes. I'll revert this part.

frontend/src/lib/services/user-country.services.ts
lmuntaner: No, I didn't test that. Let me add it.

PR #2494, opened 2023-05-11
Include API did files in releases

bitdivine:
# Motivation
When installing the nns-dapp or sns_aggregator from a pre-built wasm, it is useful to be able to get the corresponding API did files as well.  It seems preferable to provide these in the release bundle.

# Changes
- Include the canister did files in releases.

# Tests
- Pusha  tag...


bitdivine:
Done in another PR

PR #2498, opened 2023-05-11
Replace toBeFalsy() with more specific expectations

dskloetd:
# Motivation

When using `.toBeFalsy()` it's not clear whether we actually expect `false` or another value like `0` or `undefined`.

# Changes

1. Replace `.toBeFalsy()` in `frontend/src/tests/lib/stores/accounts.store.spec.ts` with `.toBe(undefined)`.
2. Replace every other instance of `toBeFalsy()` with `.toBe(false)`.

# Tests

`npm run test`


mstrasinskis (approved):
LGTM, thank you!

mstrasinskis (approved):
LGTM, thank you!

PR #2499, opened 2023-05-11
GIX-1566 Remove min/max info from participation form

dskloetd:
# Motivation

https://dfinity.atlassian.net/browse/GIX-1566

This information is already available on the project details.
Removing it will make space for the confirmation checkbox.

# Changes

1. Remove min/max info from `frontend/src/lib/components/sale/AdditionalInfoForm.svelte`
2. Remove related props from the component.
3. Remove passing the props from the parent component.
4. Remove the data from the parent component.
5. Remove the now unused i18n labels.

# Tests

Sadly no tests failed so this was untested.
Now that it's gone it no longer needs to be tested.

I did a quick manual inspection that the form still looks fine for both initial participation and increasing participation.


lmuntaner (approved):
Thanks!! Removing code is the best!

PR #2500, opened 2023-05-11
Commit in metadata

bitdivine:
# Motivation
Requests from the nns & testing teams to publish the commit of our canisters.  In fact, it is becoming common to publish both the commit and the API did file so we cover both on both canisters.  Four with one stroke.

# Changes
* Created a new command to add metadata.  This is NOT in the canister build command so that the canister build can be cached, and commits that do not affect a canister wasm don't cause a rebuild just because the commit has changed.
* Created a test for the command
* Called the command in the dockerfile
* Created a test for deployments from the docker build

# Tests
* Tests are included.  See CI for the results.

### DIFF
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -24,6 +24,7 @@ jobs:
           file: Dockerfile
           build-args: |
             DFX_NETWORK=local
+            COMMIT=${{ github.sha }}
           cache-from: type=gha,scope=cached-stage
           # Exports the artefacts from the final stage
           outputs: ./out
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,7 @@ jobs:
           file: Dockerfile
           build-args: |
             DFX_NETWORK=local
+            COMMIT=${{ github.sha }}
           cache-from: type=gha,scope=cached-stage
           # Exports the artefacts from the final stage
           outputs: .
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,6 +83,13 @@ jobs:
             # grep ... fails if the value is implausibly short.
             key="$key" jq -re '.[0][env.key]' nns_dapp_args_in_page.json | grep -E ...
           done
+      - name: Check that metadata is present
+        run: |
+          set +x
+          for canister in nns-dapp sns_aggregator ; do
+            dfx canister metadata --network ic nns-dapp git_commit_id | tee /dev/stderr | grep -E '[a-fA-F0-9]{64}'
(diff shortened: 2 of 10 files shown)


.github/workflows/aggregator.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
           file: Dockerfile
           build-args: |
             DFX_NETWORK=local
+            COMMIT=${{ github.sha }}
```
bitdivine, line 27: We need the git commit inside docker but  don't pull .git into docker.  Instead we pass in the commit to every docker  build.

Alternative experimented with:  Pull .git in to a dedicated stage, read the commit and throw away everything else from that stage.  That fails for git worktrees and will also fail for anyone trying to build from a release, that contains a tarball of the git content but not `.git/HEAD`.  I don't know that anyone is building from releases but I do know that at least I am using git worktrees and I would rather not break my build environment, even if that means making this PR bigger.

.github/workflows/build.yml
```diff
@@ -83,6 +84,10 @@ jobs:
             # grep ... fails if the value is implausibly short.
             key="$key" jq -re '.[0][env.key]' nns_dapp_args_in_page.json | grep -E ...
           done
+      - name: Install ic-wasm
+        run: cargo install "ic-wasm@$(jq -r '.defaults.build.config.IC_WASM_VERSION' dfx.json)"
```
bitdivine, line 88: `ic-wasm` is a tool used to edit canister wasms.

.github/workflows/build.yml
```diff
@@ -83,6 +84,10 @@ jobs:
...
+      - name: Check that metadata is present
+        run: scripts/dfx-wasm-metadata-add.test
```
bitdivine, line 90: This tests that the metadata is added to the wasms.

scripts/dfx-wasm-metadata-add
```diff
@@ -0,0 +1,83 @@
...
+  echo "ERROR: There should be a wasm file at: '$DFX_CANISTER_WASM'"
+  exit 1
+} >&2
+
+: "Find the candid file"
+test -n "${DFX_CANISTER_DID:-}" || DFX_CANISTER_DID="$(jq -r '.canisters[env.DFX_CANISTER_NAME].candid // ""' dfx.json)"
+test -n "${DFX_CANISTER_DID:-}" || {
+  echo "ERROR: Unable to determine the path to the candid file."
+  echo "       Please either:"
+  echo "         Use --wasm PATH_TO_WASM"
```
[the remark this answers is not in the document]
bitdivine: Yes, thank you.  Done.

dskloetd (approved):

PR #2501, opened 2023-05-11
fix: hidden my voted sns neurons

mstrasinskis:
# Motivation

Because of the wrong neuron source, when all the neurons were voted the MyVotes block is hidden.

# Changes

- change the neuron source

# Tests

TBD

# Screenshots

Screenshot 2023-05-11 at 22 03 47


lmuntaner (approved):
Thanks!

PR #2502, opened 2023-05-12
Fix: Update dissolve delay for almost 8 years neurons

lmuntaner:
# Motivation

Neurons with dissolve delay of almost 8 years (8 years - some seconds) could not update the dissolve delay to 8 years.

The issue was raised in the forum: https://forum.dfinity.org/t/peculiar-issue-7-years-and-365-days-neuron/19910?u=diegop

# Changes

The fix is to move the range input, max and min buttons to use seconds instead of days.

Main changes:
* SetDissolveDelay component uses seconds instead of days.

# Tests

* Add a test for the specific bug in SetDissolveDelay.spec.ts
* Change tests that were using the range in days to use the range in seconds.
* New SetDissolveDelay page object.

### DIFF
--- a/frontend/src/tests/page-objects/SetDissolveDelay.page-object.ts
+++ b/frontend/src/tests/page-objects/SetDissolveDelay.page-object.ts
@@ -1,0 +1,29 @@
+import { BasePageObject } from "$tests/page-objects/base.page-object";
+import type { PageObjectElement } from "$tests/types/page-object.types";
+import { ButtonPo } from "./Button.page-object";
+
+export class SetDissolveDelayPo extends BasePageObject {
+  private static readonly TID = "set-dissolve-delay-component";
+
+  static under(element: PageObjectElement): SetDissolveDelayPo {
+    return new SetDissolveDelayPo(element.byTestId(SetDissolveDelayPo.TID));
+  }
+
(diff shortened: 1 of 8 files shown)


lmuntaner:
Bug UX:

https://github.com/dfinity/nns-dapp/assets/4550653/2463cdc0-c179-43db-9d94-49863c4d45b9

UX after fix:

https://github.com/dfinity/nns-dapp/assets/4550653/d18edee0-1254-4ba2-8f2e-45d5c0fa98dc

lmuntaner:
@mstrasinskis please review

frontend/src/tests/page-objects/SetDissolveDelay.page-object.ts
```diff
@@ -0,0 +1,29 @@
+import { BasePageObject } from "$tests/page-objects/base.page-object";
+import type { PageObjectElement } from "$tests/types/page-object.types";
+import { ButtonPo } from "./Button.page-object";
+
+export class SetDissolveDelayPo extends BasePageObject {
```
mstrasinskis, line 5: Special thanks for this :)

mstrasinskis (approved):
LGTM, thank you!

PR #2506, opened 2023-05-12
Fix sed command in scripts/did2rs.sh

dskloetd:
# Motivation

`sed` works slightly differently on Mac vs. Linux.
`sed` on Mac doesn't support `\<` and `\>` as word boundary.
`[[:<:]]` and `[[:>:]]` work on Mac but not on Linux.

# Changes

1. In the `sed` command in `scripts/did2rs.sh`, instead of using word boundaries, expect specific characters.
2. Remove the `g` modifiers because they mean "multiple times per line" but each of these substitutions happens once per line.
3. Use @ instead of / in one of the substitutions to limit escaping.

# Tests

Ran `scripts/mk_nns_patch.sh` and checked that no patch files were changed.
Without the fix, that script results in much longer patch files on Mac.


bitdivine (approved):
Works on Linux, thanks!

PR #2508, opened 2023-05-12
Verify aggregator patch files in CI

dskloetd:
# Motivation

We generate files with rust types from import `.did` files and apply some changes to them.
The changes are in patch files.
If people make manual changes, the patch files are might not be usable anymore.

# Changes

1. Change `scripts/mk_nns_patch.sh` to allow running without arguments to apply to all canisters.
2. Make sure the `git diff` command always uses the same abbreviation length for hashes to get consistent output.
3. Add `scripts/mk_nns_patch.test` that verifies that the patch files are what they would be when the update script is run.
4. Run the test from CI.
5. Remove a few `set -x` to make scripts less verbose.

# Tests

This is what a failed run looks like:
https://github.com/dfinity/nns-dapp/actions/runs/4960612014/jobs/8876323282
The output ends with:
```
FAILED: Patch files should be updated to reflect current differences.
Run 'scripts/mk_nns_patch.sh' to update patch files.
Error: Process completed with exit code 1.```

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -201,8 +201,15 @@ jobs:
           dfx-software-idl2json-install --version "v$(jq -r .defaults.build.config.IDL2JSON_VERSION dfx.json)"
       - name: Check config script
         run: bash -x ./config.test
+  sns-aggregator-patches:
+    name: SNS aggregator patches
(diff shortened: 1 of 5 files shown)


bitdivine (approved):
LGTM.  Thanks, this should be useful!

PR #2510, opened 2023-05-13
ckBTC refresh balance link in explanation paragraph

peterpeterparker:
# Motivation

We explain the user it takes 12 confirmations before BTC can be received but, we do not explain how.

This PR had a call to action to "Refresh Balance" in this explanation paragraph.

# PRs

- [x] in gix-component a PR improve the hovering color https://github.com/dfinity/gix-components/pull/213

# Screenshot

Capture d’écran 2023-05-13 à 08 32 31


frontend/src/lib/components/accounts/BitcoinAddress.svelte
```diff
@@ -45,7 +47,11 @@
 </script>
 
 <p class="description">
-  {$i18n.ckbtc.incoming_bitcoin_network}
+  {$i18n.ckbtc.incoming_bitcoin_network_part_1}
+
+  <CkBTCWalletActions inline {minterCanisterId} {reload} />
```
lmuntaner, line 52: Shouldn't we think of a way to do it with `replacePlaceholders`?

Or at least add a todo to make sure the order doesn't matter.

In other languages this doesn't need to be the best split.

lmuntaner (approved):
Just one comment, but already approving.

frontend/src/lib/components/accounts/BitcoinAddress.svelte
peterpeterparker: > Shouldn't we think of a way to do it with replacePlaceholders?

Yes for sure but, it's not part of this PR. Go for it.

peterpeterparker: Added todo

peterpeterparker:
I set auto-merge here and asked @mstrasinskis to merge and add the related gix-cmp https://github.com/dfinity/nns-dapp/pull/2510

PR #2511, opened 2023-05-13
Put URLs in canister definitions

bitdivine:
# Motivation
Network entries in dfx.json are deprecated.  One of the entries we still need is the URLs to the II canister.

# Changes
- Add the canister URL in the canister specification for identity
- Add the same for nns-dapp (for consistency)

# Tests


bitdivine:
Outdated.

PR #2512, opened 2023-05-15
Update IC commit

dskloetd:
# Motivation

We need to access the `confirmation_text` [field in the SNS init params](https://github.com/dfinity/ic/blob/master/rs/sns/swap/canister/swap.did#:~:text=confirmation_text%20:%20opt%20text;) to show it in the UI.
We will access it through the sns_aggregator so it needs to be aware of the field.
The field was added in commit [3d4752b...](https://github.com/dfinity/ic/commit/3d4752b2a717334006a5e35330a132b08f890ffc).
The first *published* commit after it, according to snsdemo/bin/dfx-software-ic-latest is 10779a35b683eb991777bfdffda94145d33759db.

# Changes

1. Updated commit, did files and rust types with `scripts/update_ic_commit --ic_commit 10779a35b683eb991777bfdffda94145d33759db`.
2. Updated patch files with `scripts/mk_nns_patch.sh`.

I did not make any manual changes.
Since I did not make manual changes, there shouldn't be any meaningful changes in the patch files but updating them will prevent too much drift in the line numbers which could eventually stop them from working.
And also to prevent people making manual changes without updating the patch files, updating them is required since https://github.com/dfinity/nns-dapp/pull/2508.

# Tests

1. I locally installed the sns_aggregator and verified that the `confirmation_text` field is present in its `slow.json` response (value with `null`).
2. I manually test that locally the Launchpad and ProjectDetail page work fine while using the sns_aggregator.

### DIFF
--- a/declarations/sns_governance/sns_governance.did
+++ b/declarations/sns_governance/sns_governance.did
@@ -28,10 +28,7 @@ type By = variant {
   NeuronId : record {};
 };
 type CanisterStatusResultV2 = record {
-  controller : principal;
   status : CanisterStatusType;
-  freezing_threshold : nat;
-  balance : vec record { vec nat8; nat };
   memory_size : nat;
   cycles : nat;
   settings : DefiniteCanisterSettingsArgs;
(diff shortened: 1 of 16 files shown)


lmuntaner:
@dskloetd could you maybe use a more latest commit. The restricted countries field should also be there.

It would make sense to add them in the same PR.

dskloetd:
It makes sense. But in this case I think it's just as easy to do a new PR so I'll do that.

PR #2513, opened 2023-05-15
Fix: Min should not increase more than max

lmuntaner:
# Motivation

Clicking on "Min" when dissolve delay was adding 1 day even thought the disssolve delay was the maximum.

The button was anyway disabled, but the user could see "8 years and 1 day"

# Changes

* Check the bound of dissolve delay when using the "Min" and "Max" buttons.

# Tests

UI fix only.


lmuntaner:
@mstrasinskis please review

mstrasinskis (approved):
LGTM, thanks!

PR #2514, opened 2023-05-15
Chore: Enable split HW controlled neuron

lmuntaner:
# Motivation

In the ICP Ledger App version 2.3.0 it's enabled to split neurons.

Therefore, we should enable the UI in NNS Dapp as well.

# Changes

* "Split Neuron" button is always present. Also HW controlled neurons.
* Neuron service splitNeuron checks the controller of the neuron and the ledger app version before splitting
* New Ledger app version constant: 2.3.0

# Tests

* Test that "Split Neuron" button is present for HW controlled neurons.
* Test that neuron service split neurons checks ledger app version before continuing.

### DIFF
--- a/frontend/src/lib/constants/neurons.constants.ts
+++ b/frontend/src/lib/constants/neurons.constants.ts
@@ -13,6 +13,8 @@ export const SPAWN_VARIANCE_PERCENTAGE = 0.95;
 export const MIN_VERSION_STAKE_MATURITY_WORKAROUND = "2.0.7";
 // Version published in January 2023
 export const CANDID_PARSER_VERSION = "2.2.1";
+// Version published in February 2023
+export const SNS_SUPPORT_VERSION = "2.3.0";
 
 export const DISSOLVE_DELAY_MULTIPLIER = 1;
 export const AGE_MULTIPLIER = 0.25;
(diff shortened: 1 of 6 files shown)


lmuntaner:
@mstrasinskis please review

frontend/src/lib/constants/neurons.constants.ts
```diff
@@ -13,6 +13,8 @@ export const SPAWN_VARIANCE_PERCENTAGE = 0.95;
 export const MIN_VERSION_STAKE_MATURITY_WORKAROUND = "2.0.7";
 // Version published in January 2023
 export const CANDID_PARSER_VERSION = "2.2.1";
+// Version published in February 2023
+export const SNS_SUPPORT_VERSION = "2.3.0";
```
mstrasinskis, line 17: Maybe add move context to the name, SNS_SUPPORT_HW_VERSION?

mstrasinskis (approved):
LGTM, thanks! One optional suggestion.

frontend/src/lib/constants/neurons.constants.ts
mstrasinskis: I see, other versions also don't contain HW so feel free to ignore this const too.

PR #2518, opened 2023-05-15
Delete unused urls

bitdivine:
# Motivation

# Changes

# Tests

### DIFF
--- a/frontend/scripts/build.csp.mjs
+++ b/frontend/scripts/build.csp.mjs
@@ -211,7 +211,6 @@ const cspConnectSrc = () => {
     "https://api.iplocation.net",
     "${{OWN_CANISTER_URL}}",
     "${{HOST}}",
-    "${{GOVERNANCE_CANISTER_URL}}",
     "${{LEDGER_CANISTER_URL}}",
     "${{SNS_AGGREGATOR_URL}}",
     // TODO: solve with a worker
--- a/frontend/scripts/build.csp.mjs
+++ b/frontend/scripts/build.csp.mjs
@@ -211,7 +211,6 @@ const cspConnectSrc = () => {
     "https://api.iplocation.net",
     "${{OWN_CANISTER_URL}}",
     "${{HOST}}",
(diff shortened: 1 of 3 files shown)


frontend/scripts/build.csp.mjs
```diff
@@ -211,8 +211,6 @@ const cspConnectSrc = () => {
     "https://api.iplocation.net",
     "${{OWN_CANISTER_URL}}",
     "${{HOST}}",
```
bitdivine, line 213: Note: This is in the CSP.  But teh CSP is now generated by the canister, so the cspConnectSrc can be deleted.  Assuming that it really is unused, it is misleading and so dangerous.

bitdivine: Update: The CSP in the frontend _is_ used (sadly) however it looks as if these entries have no effect because we make no calls to the ledger or governance canister URLs.  We cannot as those domains don't serve https.

lmuntaner (approved):
Thanks, I believe you are correct.

Those endpoints are not accessed directly, always with agent-js which uses the API endpoint.

Did you try deploying this to a testnet to make sure all works as expected? Just in case...

frontend/scripts/build.csp.mjs
lmuntaner: Worth deploying to a testnet to verify that all works as expected.

bitdivine: I agree fully.  I wanted to get your eyes on it first as you work closely with the front end and there might be some detail in the dev env that does depend on these values.

lmuntaner: Nothing, AFAIK. I also thought of removing them.

bitdivine:
Deployed to small09: https://qsgjb-riaaa-aaaaa-aaaga-cai.small09.testnet.dfinity.network/proposal/?u=qsgjb-riaaa-aaaaa-aaaga-cai&proposal=8

I have had a bit of a poke around.  Seems all good.

PR #2519, opened 2023-05-15
Chore: Add Sonics SNS Gov canister to CSP

lmuntaner:
# Motivation

We get the number of participants from the metrics endpoint of SNS governance.

To do so, we need the endpoint in our CSP rules.

# Changes

* Add new endpoint to CSP rules.

# Tests

Only build changes

### DIFF
--- a/frontend/scripts/build.csp.mjs
+++ b/frontend/scripts/build.csp.mjs
@@ -217,6 +217,8 @@ const cspConnectSrc = () => {
     // TODO: solve with a worker
     // Used for the metrics of OC launch
     "https://2hx64-daaaa-aaaaq-aaana-cai.raw.ic0.app",
+    // Used for the metrics of Sonic launch.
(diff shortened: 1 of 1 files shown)


frontend/scripts/build.csp.mjs
```diff
@@ -217,6 +217,8 @@ const cspConnectSrc = () => {
     // TODO: solve with a worker
     // Used for the metrics of OC launch
     "https://2hx64-daaaa-aaaaq-aaana-cai.raw.ic0.app",
+    // Used for the metrics of Sonic launch.
+    "https://24scz-zyaaa-aaaaq-aaapq-cai.raw.ic0.app",
```
lmuntaner, line 221: I got it from https://dashboard.internetcomputer.org/sns/23ten-uaaaa-aaaaq-aaapa-cai

PR #2525, opened 2023-05-16
Chore: Enable query updates

lmuntaner:
# Motivation

We want a good UX for the upcoming swap. Therefore, we enable using only query to request data.

# Changes

* Add the flag in two calls: known neurons and latest reward event.
* Set `FORCE_CALL_STRATEGY` to `"query"`.

# Tests

Not applicable.


lmuntaner:
@mstrasinskis please review

mstrasinskis (approved):
Thanks!

PR #2527, opened 2023-05-16
Fix: Forever loading participation button

lmuntaner:
# Motivation

Users in the phone that access the Project Detail through the Launchpad sometimes had the participation button always loading.

The problem is that the BottomSheet component triggers two mount and unmounts.

Issue to investigate this further: https://dfinity.atlassian.net/browse/GIX-1574

# Changes

* Do not cancel polling getting the open ticket when ProjectDetail is destroyed.

# Tests

There are no tests that checked that the polling was cancelled.


lmuntaner:

PR #2529, opened 2023-05-17
GIX-1545 Changes required to update ic-js

dskloetd:
# Motivation

# Changes

# Tests


dskloetd:
All checks have passed now.

lmuntaner (approved):
Thanks!

Will this work with the current SNS deployed for Sonic? Did you try this with old SNS canisters?

Do we know whether Sonic canisters have the new field when participating? Does it fail if we pass it and it's not expected?

Thanks!

dskloetd:
This PR just bumps the ic-js version and fixes problems with the type system caused by the new versions of types.
It does not try to make anything work that didn't work before.

> Will this work with the current SNS deployed for Sonic?

I don't know. I have no idea what Sonic deployed.
Given that Arshavir only merged the code to be able to create such an SNS on Wednesday, I would be surprised of Sonic is using that. But I don't know.
I guess we should have discussed with the NNS team how we make sure people don't use features that can't yet be support by NNS dapp.

> Did you try this with old SNS canisters?

Yes, I tried it with an snsdemo snapshot I created on May 10th and participated in an open swap.

> Do we know whether Sonic canisters have the new field when participating? Does it fail if we pass it and it's not expected?

I'm not sure I understand the question. But at the very least you can't yet participate if the confirmation_text is set because the frontend code does not yet pass it on refresh_buyer_token.

mstrasinskis:
> Thanks!
> 
> Will this work with the current SNS deployed for Sonic? Did you try this with old SNS canisters?
> 
> Do we know whether Sonic canisters have the new field when participating? Does it fail if we pass it and it's not expected?
> 
> Thanks!

@lmuntaner the new fields looks optional. I would expect it work. Have you noticed already an issue?

lmuntaner:
> Yes, I tried it with an snsdemo snapshot I created on May 10th and participated in an open swap.

@dskloetd @mstrasinskis if this works, then we're good.

Reading optional fields is fine. What I was wondering is whether when participating, when there is a new optional field.

If nns-dapp sends `confirmation_text: []` as parameter when participating, does that work with a canister that doesn't even expect `confirmation_text`?

And if it worked, then we're good.

dskloetd:
Yes, I'm using an old snsdemo snapshot and the confirmation_text is included in the request as `[]`.
image
https://github.com/dfinity/ic-js/blob/560d1b034d7b33a246ed366a111417604002990a/packages/sns/candid/sns_swap.d.ts#LL224C1-L228C1
```
export interface RefreshBuyerTokensRequest {
  confirmation_text: [] | [string];
  buyer: string;
}
```

PR #2530, opened 2023-05-17
Take FORCE_CALL_STRATEGY into account getCkBTCWithdrawalAccount

dskloetd:
# Motivation

During the SNS swap we rely on just query calls to be prepared for heavy load.
The getCkBTCWithdrawalAccount code assumed that there will always be an update response, even though we apply the value of FORCE_CALL_STRATEGY when making the call here:
https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/services/ckbtc-withdrawal-accounts.services.ts#:~:text=strategy:%20FORCE_CALL_STRATEGY

# Changes

Check the value of FORCE_CALL_STRATEGY before assuming there will be an update call, in `getCkBTCWithdrawalAccount`.

# Tests

Manually verified that there isn't an infinite spinner for "Checking for incomplete BTC transfers".
Will add unit test separately.


mstrasinskis (approved):
LGTM, thank you!

PR #2531, opened 2023-05-17
Feat/reload transactions on refresh click

mstrasinskis:
# Motivation

Refetch transactions on the `Refresh Balance` button click

# Changes

# Tests


frontend/src/lib/components/accounts/CkBTCWalletActions.svelte
```diff
@@ -2,17 +2,30 @@
...
-    await updateBalanceService({
-      minterCanisterId,
-      reload,
-    });
+    Promise.all([
+      updateBalanceService({
+        minterCanisterId,
+        reload,
+      }),
+      loadCkBTCAccountTransactions({
```
peterpeterparker, line 23: updating the balance update the account which per extension reload the transactions so this seems innacurate I would say spontaneoulsy

mstrasinskis:
The feature was already there.

PR #2533, opened 2023-05-17
Unit test for ckBTC FORCE_CALL_STRATEGY fix

dskloetd:
# Motivation

https://github.com/dfinity/nns-dapp/pull/2536 fixed a loading ckBTC withdrawal accounts by no longer taking `FORCE_CALL_STRATEGY` into account when loading those accounts.
But it did not add a unit test for that fix.
This PR adds the unit test.

# Changes

1. Add a util to make mockable constants mockable per individual test.
2. Add a test which sets `FORCE_CALL_STRATEGY` to `"query"` and verifies that withdrawal accounts are still loaded. (It duplicates the test above it with the only difference being `FORCE_CALL_STRATEGY = "query"`.)
3. Remove a check on `FORCE_CALL_STRATEGY` in the error handler since it's now irrelevant.

# Tests

Unit test is added.
I also verified that the test fails without the `FORCE_CALL_STRATEGY` change from https://github.com/dfinity/nns-dapp/pull/2536.

### DIFF
--- a/frontend/src/tests/utils/mockable-constants.test-utils.ts
+++ b/frontend/src/tests/utils/mockable-constants.test-utils.ts
@@ -1,0 +1,19 @@
+// Values used during unit testing for the constants in
+// frontend/src/lib/constants/mockable.constants.ts
+
+const defaultTestConstants = {
+  DEV: false,
+  ENABLE_METRICS: false,
+  FORCE_CALL_STRATEGY: undefined,
+  IS_TEST_ENV: true,
+  QR_CODE_RENDERED_DEFAULT_STATE: true,
+  ENABLE_QR_CODE_READER: false,
+};
+
+export const mockedConstants = { ...defaultTestConstants };
+
+export const resetMockedConstants = () => {
+  Object.keys(mockedConstants).forEach((key) => {
+    mockedConstants[key] = defaultTestConstants[key];
+  });
+};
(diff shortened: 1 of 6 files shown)


frontend/src/tests/utils/mockable-constants.test-utils.ts
```diff
@@ -0,0 +1,19 @@
+// Values used during unit testing for the constants in
+// frontend/src/lib/constants/mockable.constants.ts
+
+const defaultTestConstants = {
```
peterpeterparker, line 4: are these also use for e2e, that's why you move these?

if not, for maintaince reason I find easier to have all mock data set at the same place - i.e. in `jest-setup`.

dskloetd: Not for e2e tests.
My goal is that the unit tests can override the values on a test-by-test basis.
So I need a library that is used by both `jest-setup` and unit tests.
But if you prefer, I can populate the initial values from `jest-setup` into the library rather than the library having its own defaults?

peterpeterparker: Thanks for the explanation.

> But if you prefer, I can populate the initial values from jest-setup into the library 

That would be nice, thanks

peterpeterparker (approved):
Thanks

frontend/src/tests/utils/mockable-constants.test-utils.ts
dskloetd: Done

lmuntaner (approved):
Thanks!

PR #2535, opened 2023-05-17
Query and update for ckBTC withdrawal account

peterpeterparker:
# Motivation

Always use (fake) QUERY + (real) UPDATE for ckBTC to retrieve withdrawal account even if the stategy is set to query.

# Changes

- remove usage of strategy (that should have not been there in first place, copy/paste are bad)
- remove fix that handles strategy

### DIFF
--- a/frontend/src/lib/services/ckbtc-withdrawal-accounts.services.ts
+++ b/frontend/src/lib/services/ckbtc-withdrawal-accounts.services.ts
@@ -7,13 +7,16 @@ import {
 } from "$lib/stores/ckbtc-withdrawal-accounts.store";
 import type { UniverseCanisterId } from "$lib/types/universe";
 
+/**
+ * To load the withdrawal account we use QUERY+UPDATE strategy. As there is not QUERY option provided by the canister we fake a static result instead.
+ * That way we can display a spinner information in the UI while the account is loading.
+ */
 export const loadCkBTCWithdrawalAccount = async ({
   universeId,
 }: {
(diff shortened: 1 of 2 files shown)


frontend/src/lib/services/ckbtc-withdrawal-accounts.services.ts
```diff
@@ -7,13 +7,16 @@ import {
...
+ * To load the withdrawal account we use QUERY+UPDATE strategy. As there is not QUERY option provided by the canister we fake a static result instead.
+ * That way we can display a spinner information in the UI while the account is loading.
+ */
 export const loadCkBTCWithdrawalAccount = async ({
   universeId,
 }: {
   universeId: UniverseCanisterId;
 }): Promise<void> => {
   return queryAndUpdate<CkBTCBTCWithdrawalAccount, unknown>({
-    strategy: FORCE_CALL_STRATEGY,
```
dskloetd, line 16: If `"query_and_update"` is essential, I think it would be better to pass `strategy: "query_and_update"` explicitly.

peterpeterparker: good point 👍

peterpeterparker: Done

mstrasinskis (approved):
Rubber stamp!

peterpeterparker:
Changes integrated in #2536

PR #2536, opened 2023-05-17
Always reload account and transactions on "Refresh balance"

peterpeterparker:
# Motivation

We reload in any case because the user might hit "Refresh balance" in the UI thinking the feature is meant to reload the account and transactions even though it is actually meant to confirm the conversion of BTC -> ckBTC

In addition this PR integrates #2535 to reuse pseudo query and update call to load the withdrawal account.

# Changes

- always reload regardless if `updateBalance` ends in error or success


mstrasinskis (approved):
Thanks a lot! LGTM.

PR #2538, opened 2023-05-18
clean: Didc version in setup

bitdivine:
# Motivation
The didc version was not always taken from dfx.json.  While the versions were all correct, updating the version in dfx.json would have caused them to go out of sync.

# Changes
- Always get the didc version from dfx.json

# Tests
- See CI for the changes to github actions.
- I have run the local development setup script locally

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -251,11 +251,13 @@ install_idl2json() {
 }
 
 install_didc_linux() {
-  local version="$(jq -r .defaults.build.config.DIDC_VERSION dfx.json)"
+  local version
+  version="$(jq -r .defaults.build.config.DIDC_VERSION dfx.json)"
   curl -Lf "https://github.com/dfinity/candid/releases/download/${version}/didc-linux64" | install -m 755 /dev/stdin "$USER_BIN/didc"
 }
 install_didc_darwin() {
-  local version="$(jq -r .defaults.build.config.DIDC_VERSION dfx.json)"
+  local version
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Thanks!

scripts/setup
```diff
@@ -251,11 +251,15 @@ install_idl2json() {
...
-  curl -Lf https://github.com/dfinity/candid/releases/download/2022-11-17/didc-linux64 | install -m 755 /dev/stdin "$USER_BIN/didc"
+  local version
+  version="$(jq -r .defaults.build.config.DIDC_VERSION dfx.json)"
+  curl -Lf "https://github.com/dfinity/candid/releases/download/${version}/didc-linux64" | install -m 755 /dev/stdin "$USER_BIN/didc"
 }
 install_didc_darwin() {
+  local version
+  version="$(jq -r .defaults.build.config.DIDC_VERSION dfx.json)"
   tempfile="$(mktemp ,didc-XXXXXXXX)"
-  curl -Lf https://github.com/dfinity/candid/releases/download/2022-11-17/didc-macos >"$tempfile"
```
dskloetd, line 258: FYI, this is version 0.1.6 and it is the version I had installed until a few days ago. But it causes breaking changes when running `scripts/compile-idl-js` in ic-js. So I downgraded to version 0.1.4 and made `scripts/compile-idl-js` also require version 0.1.4 in [this PR](https://github.com/dfinity/ic-js/pull/336).

dskloetd (approved):

PR #2539, opened 2023-05-19
Fix: Issue with arguments in CMC proposal

lmuntaner:
# Motivation

CMC proposal had a proposal to upgrade using as argument `()`. This is not a valid argument for the nns-dapp serializer because the arguments must always be `opt <something>`. Yet, to install canisters, this is a valid argument.

We need to support this argument and not fail to show the payload.

# Changes

* Return an emtpy array json instead of an error if `Decode` fails.

# Tests

There are no tests for this. But it would be good to add a few. In another PR.


lmuntaner:
@bitdivine please review

PR #2541, opened 2023-05-19
fix: back navigation from the launchpad proposal details page

mstrasinskis:
# Motivation

Resolve the problem related to the incorrect back navigation from the launchpad proposal details page, which currently directs users to the proposals page instead of the launchpad page.

# Changes

- add `referrerPath` based back navigation

# Tests

A way to include unit tests for this specific scenario was not found.

# Screenshots
| Before | After |
|--------|--------|
| Screen Recording 2023-05-19 at 11 25 41 | Screen Recording 2023-05-19 at 11 24 16 |


peterpeterparker (approved):
Thx

PR #2542, opened 2023-05-19
Test: Add Participation Test in Unit Tests

lmuntaner:
# Motivation

Improve testing.

To easily test the participation flow under different scenarios, it's better if we can use the same page object used in e2e for unit tests.

In this PR: Add a test to ProjectDetail.svelte using the ProjectDetailPo to participate in a swap.

# Changes

Main change:
* Add new test scenario in ProjectDetail.svelte for a user to go through the participation flow. Use same as in e2e test.

Minor changes:
* Add a couple of `waitFor` in the jest object before performing some actions.

# Tests

Only changes in tests.

### DIFF
--- a/frontend/src/tests/e2e/sns-participation.spec.ts
+++ b/frontend/src/tests/e2e/sns-participation.spec.ts
@@ -50,8 +50,9 @@ test("Test SNS participation", async ({ page, context }) => {
 
   // D004: User can participate in a sale
   expect(await projectDetail.hasCommitmentAmount()).toBe(false);
-  await projectDetail.participate({ amount: 5 });
-  expect(await projectDetail.getCommitmentAmount()).toBe("5.00");
+  const formattedAmountICP = "5.00";
+  await projectDetail.participate({ amount: Number(formattedAmountICP) });
+  expect(await projectDetail.getCommitmentAmount()).toBe(formattedAmountICP);
 
   // D005: User can increase the participation in a sale
   // TODO
--- a/frontend/src/tests/e2e/sns-participation.spec.ts
+++ b/frontend/src/tests/e2e/sns-participation.spec.ts
@@ -50,9 +50,8 @@ test("Test SNS participation", async ({ page, context }) => {
 
   // D004: User can participate in a sale
   expect(await projectDetail.hasCommitmentAmount()).toBe(false);
-  const formattedAmountICP = "5.00";
-  await projectDetail.participate({ amount: Number(formattedAmountICP) });
-  expect(await projectDetail.getCommitmentAmount()).toBe(formattedAmountICP);
+  await projectDetail.participate({ amount: 5 });
+  expect(await projectDetail.getCommitmentAmount()).toBe("5.00");
 
   // D005: User can increase the participation in a sale
   // TODO
--- a/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
@@ -349,7 +349,8 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
           subAccounts: [],
           hardwareWallets: [],
         });
-        const amountICP = 5;
+        const formattedAmountICP = "5.00";
+        const amountICP = Number(formattedAmountICP);
         const amountE8s = numberToE8s(amountICP);
         const finalCommitment = {
           icp: [
@@ -398,7 +399,9 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
 
         expect(await projectDetail.hasCommitmentAmount()).toBe(false);
         await projectDetail.participate({ amount: amountICP });
-        expect(await projectDetail.getCommitmentAmount()).toBe("5.00");
+        expect(await projectDetail.getCommitmentAmount()).toBe(
+          formattedAmountICP
+        );
       });
 
       it("should participate without user interaction if there is an open ticket.", async () => {
--- a/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
@@ -4,6 +4,7 @@
 
 import * as ledgerApi from "$lib/api/ledger.api";
 import * as locationApi from "$lib/api/location.api";
+import * as nnsDappApi from "$lib/api/nns-dapp.api";
 import * as snsSaleApi from "$lib/api/sns-sale.api";
 import * as snsMetricsApi from "$lib/api/sns-swap-metrics.api";
 import * as snsApi from "$lib/api/sns.api";
@@ -13,12 +14,17 @@ import { pageStore } from "$lib/derived/page.derived";
 import * as summaryGetters from "$lib/getters/sns-summary";
 import ProjectDetail from "$lib/pages/ProjectDetail.svelte";
 import { cancelPollGetOpenTicket } from "$lib/services/sns-sale.services";
+import { accountsStore } from "$lib/stores/accounts.store";
 import { authStore } from "$lib/stores/auth.store";
 import { snsSwapMetricsStore } from "$lib/stores/sns-swap-metrics.store";
 import { snsQueryStore, snsSwapCommitmentsStore } from "$lib/stores/sns.store";
 import type { SnsSwapCommitment } from "$lib/types/sns";
-import { formatToken } from "$lib/utils/token.utils";
+import { formatToken, numberToE8s } from "$lib/utils/token.utils";
 import { page } from "$mocks/$app/stores";
+import {
+  mockAccountDetails,
+  mockMainAccount,
+} from "$tests/mocks/accounts.store.mock";
 import {
   mockAuthStoreNoIdentitySubscribe,
   mockAuthStoreSubscribe,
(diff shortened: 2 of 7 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -71,6 +80,10 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
     jest.useFakeTimers().setSystemTime(now);
 
     jest.spyOn(ledgerApi, "sendICP").mockResolvedValue(undefined);
+
+    jest
+      .spyOn(nnsDappApi, "queryAccount")
```
lmuntaner, line 85: When the modal to participate is opened, it triggers a call to `loadAccounts`. This part was not mocked here yet.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -266,7 +279,7 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
         jest.spyOn(ledgerApi, "sendICP").mockResolvedValue(BigInt(10));
         jest
           .spyOn(ledgerApi, "queryAccountBalance")
-          .mockResolvedValue(BigInt(10_000_000));
+          .mockResolvedValue(BigInt(1_000_000_000));
```
lmuntaner, line 282: I needed enough balance to participate with 5 ICP

frontend/src/tests/page-objects/ParticipateSwapModal.page-object.ts
```diff
@@ -12,9 +12,13 @@ export class ParticipateSwapModalPo extends TransactionModalPo {
 
   async participate({ amount }: { amount: number }): Promise<void> {
     const formPo = this.getTransactionFormPo();
+    // Needed for Jest
+    await formPo.waitFor();
```
dskloetd, line 16: I would like to try to avoid platform specific code in the general page objects.
Does it work if we remove this `waitFor()` here and instead put `await this.waitFor()` at the top of `JestPageObjectElement.typeText()`?

frontend/src/tests/page-objects/ParticipateSwapModal.page-object.ts
```diff
@@ -12,9 +12,13 @@ export class ParticipateSwapModalPo extends TransactionModalPo {
...
     await formPo.enterAmount(amount);
     await formPo.clickContinue();
     const review = this.getTransactionReviewPo();
+    // Needed for Jest
+    await review.waitFor();
```
dskloetd, line 21: and if so, maybe we can do something similar here?

frontend/src/tests/page-objects/ParticipateButton.page-object.ts
```diff
@@ -17,6 +17,7 @@ export class ParticipateButtonPo extends BasePageObject {
     await this.getButton().click();
     const modal = this.getParticipateSwapModalPo();
     await modal.participate({ amount });
-    await modal.waitForAbsent();
+    // Needed to pass the test in Jest environment
+    await modal.waitForAbsent(5_000);
```
dskloetd, line 21: Way is the timeout needed?

And can we isolate it to the `jest.page-object.ts`?

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -324,6 +337,65 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
...
+          } as SnsSwapCommitment)
+          .mockResolvedValue({
+            rootCanisterId: Principal.fromText(rootCanisterId),
+            myCommitment: finalCommitment,
+          } as SnsSwapCommitment);
+        jest.spyOn(snsSaleApi, "getOpenTicket").mockResolvedValue(undefined);
+        jest.spyOn(snsSaleApi, "newSaleTicket").mockResolvedValue({
+          ...testTicket,
+          amount_icp_e8s: amountE8s,
+        });
```
dskloetd, line 378: Not necessarily for this PR but it would be nice replace this wall of mock expectations with a fake.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -324,6 +337,65 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
...
+
+        await waitFor(async () =>
+          expect(await projectDetail.getStatus()).toBe(
+            "Accepting Participation"
+          )
+        );
+
+        expect(await projectDetail.hasCommitmentAmount()).toBe(false);
+        await projectDetail.participate({ amount: amountICP });
+        expect(await projectDetail.getCommitmentAmount()).toBe("5.00");
```
dskloetd, line 396: How about defining `amountIcpFormatted = "5.00"` above, near the other amount constants, so it's obvious that the value matches what's used in the participation?

frontend/src/tests/page-objects/ParticipateSwapModal.page-object.ts
lmuntaner: `waitFor` is both in Jest and Playwright:

```
// playwright.page-object.ts
  waitFor(): Promise<void> {
    return this.locator.waitFor();
  }
```

```
// jest.page-object.ts
  waitFor(): Promise<void> {
    return waitFor(async () => {
      expect(await this.isPresent()).toBe(true);
    });
  }
```

I tried and no, it's not working. It needs to wait the transaction form.

Although I also think that it should work there.

I don't see a problem having different "waitFor" in both environments. They work differently.

What's the problem with adding extra "waitFor" you think?

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
lmuntaner: yes, I agree

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
lmuntaner: Sure

frontend/src/tests/page-objects/ParticipateButton.page-object.ts
lmuntaner: I replied before. It fails. I can spend more time on it I guess.

But what's the rule of when and when not to use waits? There are already waits in other places. Even the `waitForAbsent` here.

frontend/src/tests/page-objects/ParticipateSwapModal.page-object.ts
lmuntaner: done

frontend/src/tests/page-objects/ParticipateButton.page-object.ts
lmuntaner: Moved to jest page object. I might be able to remove it from there in another PR.

dskloetd: Yeah, it's OK to have waits.
When I need waits for e2e, I don't check if maybe they aren't needed for Jest. I just add them.
So I wouldn't add a comment about Jest when adding a wait.
But this timeout looked very arbitrary and unnecessary.

And also in certain cases Playwright will automatically wait for something. For example if you try to enter text, it will wait for the input field to exist. So I thought it could be more convenient to make the JestPageObjectElement more similar to PlaywrightPageObjectElement rather than having to put Jest specific waits in places where Playwright does them automatically.

frontend/src/tests/e2e/sns-participation.spec.ts
```diff
@@ -50,8 +50,9 @@ test("Test SNS participation", async ({ page, context }) => {
 
   // D004: User can participate in a sale
   expect(await projectDetail.hasCommitmentAmount()).toBe(false);
-  await projectDetail.participate({ amount: 5 });
-  expect(await projectDetail.getCommitmentAmount()).toBe("5.00");
+  const formattedAmountICP = "5.00";
+  await projectDetail.participate({ amount: Number(formattedAmountICP) });
```
dskloetd, line 54: I honestly preferred having 2 constants (regardless of whether they are literals or named constants. I don't want to have to do a computation in my head to read what a test does.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -349,7 +349,8 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
           subAccounts: [],
           hardwareWallets: [],
         });
-        const amountICP = 5;
+        const formattedAmountICP = "5.00";
+        const amountICP = Number(formattedAmountICP);
```
dskloetd, line 353: Here I would also hard code the value instead of computing it.

frontend/src/tests/page-objects/jest.page-object.ts
```diff
@@ -89,12 +89,15 @@ export class JestPageObjectElement implements PageObjectElement {
     });
   }
 
-  waitForAbsent(timeout?: number): Promise<void> {
+  waitForAbsent(): Promise<void> {
     return waitFor(
       async () => {
-        expect(await this.isPresent()).toBe(false);
+        return expect(await this.isPresent()).toBe(false);
```
dskloetd, line 95: Good catch!

dskloetd (approved):
Thanks!

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
lmuntaner: Ok

frontend/src/tests/e2e/sns-participation.spec.ts
lmuntaner: ok, then I'll revert.

PR #2546, opened 2023-05-22
Update ParticipateSwapModal.spec.ts to use page objects

dskloetd:
# Motivation

I want to change ParticipateSwapModal to add the confirmation checkbox for geo restrictions.
I will add a test for that and would like to use page objects.
But I prefer not to mix different styles in the same test file.
So I'm first updating the test to use page objects.

# Changes

1. Use page objects in `frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts`
2. Also when checking that `initiateSnsSaleParticipation` is called, check that it wasn't called before. This wasn't actually true until I cleared the mock in `beforeEach`.
3. Add missing page objects and functionality to existing page objects.

# Tests

`npm run test`
`npm run test-e2e`

### DIFF
--- a/frontend/src/tests/page-objects/InProgress.page-object.ts
+++ b/frontend/src/tests/page-objects/InProgress.page-object.ts
@@ -1,7 +1,4 @@
-import { AmountInputPo } from "$tests/page-objects/AmountInput.page-object";
 import { BasePageObject } from "$tests/page-objects/base.page-object";
-import type { ButtonPo } from "$tests/page-objects/Button.page-object";
-import { SelectDestinationAddressPo } from "$tests/page-objects/SelectDestinationAddress.page-object";
 import type { PageObjectElement } from "$tests/types/page-object.types";
 
 export class InProgressPo extends BasePageObject {
--- a/frontend/src/tests/page-objects/TransactionFromAccount.page-object.ts
+++ b/frontend/src/tests/page-objects/TransactionFromAccount.page-object.ts
@@ -1,6 +1,5 @@
 import { BasePageObject } from "$tests/page-objects/base.page-object";
 import { DropdownPo } from "$tests/page-objects/Dropdown.page-object";
-import { TogglePo } from "$tests/page-objects/Toggle.page-object";
 import type { PageObjectElement } from "$tests/types/page-object.types";
 
 export class TransactionFromAccountPo extends BasePageObject {


frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
```diff
@@ -262,14 +238,23 @@ describe("ParticipateSwapModal", () => {
...
+      const po = new ParticipateSwapModalPo(
+        new JestPageObjectElement(container)
+      );
 
-      expect(queryByTestId("select-account-dropdown")).not.toBeInTheDocument();
+      const fromAccount = po
+        .getTransactionFormPo()
+        .getTransactionFromAccountPo();
+
+      // TODO: Relying on the accounts getting loaded async in between these
```
lmuntaner, line 250: Why is this?

The SwapModal loads accounts if they are not loaded. We want to test this here.

We could test it with `runResolvedPromises` when we migrate to mock only the API layer.

lmuntaner (approved):
LGTM! Thanks, just a couple of comments. Not for this PR though.

frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
dskloetd: But the swap modal is rendered on line 241. So the process to load accounts is set in motion there, right?
So we are just lucky that on line 253 they haven't loaded yet and we are able to verify that there?

What I'm saying is that the test should do something to delay the loading (like `fake.pause()`) before rendering and trigger the loading (like `fake.resume()`) in between line 253 and line 257.

`runResolvedPromises` is sometimes necessary but we should never rely on whatever happens during `runResolvedPromises` to not happen earlier.

Or am I really misunderstanding what is going on here?

lmuntaner: > What I'm saying is that the test should do something to delay the loading (like fake.pause()) before rendering and trigger the loading (like fake.resume()) in between line 253 and line 257.

That is also what I mean, alghough I forgot about the fake and used `runResolvedPromises` instead.

dskloetd: `runResolvedPromises` can be used to "resume" but there is no equivalent for "pause". We just have to be lucky that we don't accidentally resolve the promises in some other way, which is why I call it brittle.

PR #2549, opened 2023-05-22
GIX-1553: Disable button user restricted country

lmuntaner:
# Motivation

SNS projects can restrict swap participation based on the user's country.

# Changes

* Handle the button status "disabled-not-eligible" in the ParticipateButton.
* Implement the `getDeniedCountries` getter helper.

# Tests

* Add test case in ParticipateButton.spec

### DIFF
--- a/frontend/src/lib/components/project-detail/ParticipateButton.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateButton.svelte
@@ -89,7 +89,17 @@
                 disabled>{$i18n.sns_project_detail.participate}</button
               >
             </Tooltip>
-            <!-- TODO: GIX-1553 Add disabled-not-eligible -->
+          {:else if buttonStatus === "disabled-not-eligible"}
+            <Tooltip
+              id="sns-project-participate-button-tooltip"
+              text={$i18n.sns_project_detail.not_eligible_to_participate}
+            >
+              <button
+                class="primary"
+                data-tid="sns-project-participate-button"
+                disabled>{$i18n.sns_project_detail.participate}</button
+              >
+            </Tooltip>
           {:else}
             <!-- This is the "enabled" and "disabled-not-eligible" case -->
             <button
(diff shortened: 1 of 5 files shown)


lmuntaner:
User from restricted country:

https://github.com/dfinity/nns-dapp/assets/4550653/aae0b331-a5e6-40bd-af77-4ecd12e7162a

User from non-restricted country:

https://github.com/dfinity/nns-dapp/assets/4550653/7bac627f-5d2b-41d8-8e54-05eeb3788199

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -89,7 +89,17 @@
...
+              text={$i18n.sns_project_detail.not_eligible_to_participate}
+            >
+              <button
+                class="primary"
+                data-tid="sns-project-participate-button"
+                disabled>{$i18n.sns_project_detail.participate}</button
+              >
+            </Tooltip>
           {:else}
             <!-- This is the "enabled" and "disabled-not-eligible" case -->
```
dskloetd, line 104: This comment needs to be updated because `"disabled-not-eligible"` is now covered above.

frontend/src/tests/lib/components/project-detail/ParticipateButton.spec.ts
```diff
@@ -260,6 +260,36 @@ describe("ParticipateButton", () => {
...
+      const button = queryByTestId(
+        "sns-project-participate-button"
+      ) as HTMLButtonElement;
+      expect(button.getAttribute("disabled")).not.toBeNull();
+
+      const tooltipPo = new TooltipPo(new JestPageObjectElement(container));
+      expect(await tooltipPo.getText()).toBe(
+        "You are not eligible to participate in this swap."
+      );
+    });
```
dskloetd, line 292: Should we also have a test with a non-empty deny list, but where the user is in a different country and is allowed to participate, to make sure the button then is not disabled?

dskloetd (approved):

frontend/src/lib/components/project-detail/ParticipateButton.svelte
lmuntaner: thanks!

PR #2552, opened 2023-05-22
Argument rendering test vectors

bitdivine:
# Motivation
Argument rendering is currently imperfect.  It has no tests, though, making it hard to evaluate improvements.

# Changes
- Add tests for argument rendering.
  - The test vector includes both the current behaviour and the value that we would probably like.  A PR that makes the actual and desired values align more closely is potentially a good thing.
- Split the argument rendering function into two parts, for more comprehensive testing.

# Tests
- This PR provides almost nothing but tests.

### DIFF
--- a/rs/backend/src/proposals/tests/args.rs
+++ b/rs/backend/src/proposals/tests/args.rs
@@ -3,7 +3,6 @@ use crate::proposals::decode_arg;
 use anyhow::Context;
 use candid::parser::types::{IDLType, IDLTypes};
 use candid::{IDLArgs, IDLProg};
-use fn_error_context::context;
 use std::str::FromStr;
 
 /// Sample argument and expected corresponding output.
@@ -150,7 +149,6 @@ fn args_should_be_parsed() {
     }
 }
 
-#[context("Test vector '{}' failed", test_vector.name)]
 fn arg_should_be_parsed(test_vector: &TestVector) -> anyhow::Result<()> {
     let TestVector {
         name,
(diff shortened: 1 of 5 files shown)


rs/backend/src/proposals/tests/args.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Canister init arg.
+    ///
+    /// Provided by the proposer.  May be invalid.
+    args: &'static str,
+    /// JSON or error message
+    ///
+    /// Note: We may want to always return a str, or a str and a list of warnings.
+    ///       E.g. if the arg does not match the schema, we might show the arg with a warning.
+    #[allow(unused)]
+    json: &'static str,
```
lmuntaner, line 27: I'm not sure that I understand this field.

Do you want to change the return value of the function in the future?

rs/backend/src/proposals/tests/args.rs
```diff
@@ -0,0 +1,168 @@
...
+    // TODO: Check serialization when arguments are provided with annotation.
+    // TODO: Decide how to handle the case when the type and data don't match
+    // TODO: Handle args that are not valid candid
+    // TODO: Decide what we want to show when the type itself is invalid
+];
+
+/// Extract the args field from a did types file.
+///
+/// TODO: Move into idl2json
+fn arg_types_from_did(did: &str) -> anyhow::Result<IDLTypes> {
```
lmuntaner, line 120: This is useful for testing purposes, right?

lmuntaner (approved):
That's great thanks!

PR #2554, opened 2023-05-23
Fix: npm run build

lmuntaner:
# Motivation

`npm run build` is broken.

The problem: Code that depends on `window` was executed at build time.

Ideally: That component should not be executed at build time.

Workaround: Check whether the environment is the browser before calling the problematic code.

# Changes

* Check whether we're in the browser in `GetTokens` before getting the balance of the test account.

# Tests

This component is not tested because it's used only in testnets or loca environments.

### DIFF
--- a/frontend/src/lib/components/ic/GetTokens.svelte
+++ b/frontend/src/lib/components/ic/GetTokens.svelte
@@ -17,6 +17,7 @@
   import { ICPToken, type Token } from "@dfinity/nns";
   import { snsTokenSymbolSelectedStore } from "$lib/derived/sns/sns-token-symbol-selected.store";
   import { authSignedInStore } from "$lib/derived/auth.derived";
+  import { isBrowser } from "@dfinity/auth-client/lib/cjs/storage";
 
   let visible = false;
   let transferring = false;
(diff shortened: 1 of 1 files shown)


lmuntaner:

frontend/src/lib/components/ic/GetTokens.svelte
```diff
@@ -75,7 +76,10 @@
   let tokenBalanceE8s = 0n;
   $: selectedProjectId,
     (async () => {
-      tokenBalanceE8s = await getTestBalance(selectedProjectId);
+      // This was executed at build time and it depends on `window` to convert string to AccountIdentifier.
```
dskloetd, line 79: Why is this executed at build time?
Why did it not break the build on CI?
And why does getTestBalance depend on window?

dskloetd (approved):

frontend/src/lib/components/ic/GetTokens.svelte
lmuntaner: * I don't know. That's what we should fix long term.
* I don't know. It should have. Maybe it doesn't run it there?
* There is an util that relies on window. I'll change the comment to be more specific. `base64ToUInt8Array`

PR #2555, opened 2023-05-23
GIX-1555 Show conditions to accept if confirmation_text is present

dskloetd:
# Motivation

We want to allow SNS creators to specify conditions that have to be accepted by swap participants.
For this purpose there is a field `confirmation_text` in the swap summary.

This PR only displays the confirmation text and checkbox.
It does not yet require that the checkbox is checked or send the confirmation text in the refresh buyer tokens request.

# Changes

1. In `frontend/src/lib/modals/sns/sale/ParticipateSwapModal.svelte` pass the `confirmation_text` from the summary to the `AdditionalInfoForm` component.
2. Add a convenience getter in `frontend/src/lib/getters/sns-summary.ts`
3. Add 2 tests to verify that the text is rendered if and only if it's present in the summary.
4. Add `renderSwapModalPo` in the test, convenient for the case where the test only needs the po, and use it in existing tests.
5. Make all the params of `createSummary` optional and add `confirmationText` as one more parameter of it.
6. Add `getAdditionalInfoFormPo` to `ParticipateSwapModalPo`.

# Tests

1. Included in the PR.
2. Tested manually that the text renders only when the field is set.

### DIFF
--- a/frontend/src/lib/getters/sns-summary.ts
+++ b/frontend/src/lib/getters/sns-summary.ts
@@ -5,3 +5,8 @@ import { fromNullable } from "@dfinity/utils";
 export const getDeniedCountries = (_summary: SnsSummary): CountryCode[] =>
   fromNullable(fromNullable(_summary.swap.init)?.restricted_countries ?? [])
     ?.iso_codes ?? [];
+
+export const getConditionsToAccept = (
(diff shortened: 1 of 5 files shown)


lmuntaner (approved):
Thanks!!

PR #2556, opened 2023-05-23
GIX-1543: Allow user participate if error getting country

lmuntaner:
# Motivation

If there is an error getting the country of the user, we enable the button to participate.

# Changes

No functionality changes.

# Tests

* Add a test in the ParticipateButton and ProjectDetail with the scenario that getting the country fails.

### DIFF
--- a/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
@@ -17,6 +17,7 @@ import { accountsStore } from "$lib/stores/accounts.store";
 import { authStore } from "$lib/stores/auth.store";
 import { snsSwapMetricsStore } from "$lib/stores/sns-swap-metrics.store";
 import { snsQueryStore, snsSwapCommitmentsStore } from "$lib/stores/sns.store";
+import { userCountryStore } from "$lib/stores/user-country.store";
 import type { SnsSwapCommitment } from "$lib/types/sns";
 import { formatToken, numberToE8s } from "$lib/utils/token.utils";
 import { page } from "$mocks/$app/stores";
@@ -64,7 +65,8 @@ const blockedApiPaths = [
 
 describe("ProjectDetail", () => {
   blockAllCallsTo(blockedApiPaths);
-  const countryCode = "CH";
+  const userCountryCode = "CH";
+  const notUserCountryCode = "US";
   const newBalance = BigInt(10_000_000_000);
   const saleBuyerCount = 1_000_000;
   const rawMetricsText = `
@@ -77,6 +79,7 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
     snsQueryStore.reset();
     snsSwapCommitmentsStore.reset();
     snsSwapMetricsStore.reset();
+    userCountryStore.set("not loaded");
 
     jest.clearAllTimers();
     const now = Date.now();
@@ -91,7 +94,7 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
 
     jest
       .spyOn(locationApi, "queryUserCountryLocation")
-      .mockResolvedValue(countryCode);
+      .mockResolvedValue(userCountryCode);
 
     jest.spyOn(snsApi, "querySnsDerivedState").mockResolvedValue({
       sns_tokens_per_icp: [1],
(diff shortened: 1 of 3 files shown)


frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -349,13 +352,7 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
         });
       });
 
-      it("should participate with user interaction if there is no open ticket.", async () => {
```
lmuntaner, line 352: I moved the setup of this test to a before each inside a describe to reuse it.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -368,47 +365,128 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
...
+              myCommitment: undefined,
+            } as SnsSwapCommitment)
+            .mockResolvedValue({
+              rootCanisterId: Principal.fromText(rootCanisterId),
+              myCommitment: finalCommitment,
+            } as SnsSwapCommitment);
         });
 
-        const { container } = render(ProjectDetail, props);
+        it("when no restricted countries", async () => {
```
lmuntaner, line 399: This is the previous test.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -368,47 +365,90 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
...
-        );
+          expect(await projectDetail.hasCommitmentAmount()).toBe(false);
+          await projectDetail.participate({ amount: amountICP });
+          expect(await projectDetail.getCommitmentAmount()).toBe(
+            formattedAmountICP
+          );
+        };
+
+        it("when no restricted countries", async () => {
+          await participateInSwap();
```
lmuntaner, line 422: This is the same test as it was before.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -368,47 +365,90 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
...
+          );
 
-        await waitFor(async () =>
-          expect(await projectDetail.getStatus()).toBe(
-            "Accepting Participation"
-          )
-        );
+          await waitFor(async () =>
+            expect(
+              await projectDetail.getParticipateButton().isDisabled()
```
lmuntaner, line 410: This is the only change I did to the previous test. I thought it was more accurate to wait for the button to be enabled than to wait for the status.

lmuntaner:
@dskloetd please review

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -368,47 +365,128 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
...
-          formattedAmountICP
-        );
+          expect(await projectDetail.hasCommitmentAmount()).toBe(false);
+          await projectDetail.participate({ amount: amountICP });
+          expect(await projectDetail.getCommitmentAmount()).toBe(
+            formattedAmountICP
+          );
+        });
+
+        it("when restricted countries and user is from another country", async () => {
```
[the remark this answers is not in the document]
dskloetd: Do I see correctly that this file does not have a test where the button is disabled because the user is in a restricted country?
Will you add it in a different PR?
Or why do we need a test in this file where the button is enabled but not one where it is disabled?

dskloetd (approved):

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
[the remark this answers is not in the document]
lmuntaner: That's a good question. I wasn't sure what should and should not be here.

In the end, I decided to add at this level only the integration tests of the whole participation flow in different scenarios.

The specific edge cases where participating is not enabled are mostly in the ParticipateButton.

But I can add the test for the restricted countries in another PR as well.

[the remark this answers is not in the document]
lmuntaner: it is asymetric. I was even thinking on adding this integrations tests in another file precisely because of that.

PR #2557, opened 2023-05-23
GIX-1555 Disable Continue until swap conditions are accepted

dskloetd:
# Motivation

When an SNS confirmation text is specified, we want to require that the conditions are accepted before a user can continue participating in the swap.

# Changes

1. Export, and pass on, a `disableContinue` prop on `TransactionModal` and `TransactionForm`.
2. In `frontend/src/lib/modals/sns/sale/ParticipateSwapModal.svelte` set `disableContinue` to `true` iff `confirmation_text` is present and the checkbox is not checked.

# Tests

Added 2 tests to `frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts` for the case with and without confirmation_text.


lmuntaner (approved):
LGTM, thanks!

PR #2558, opened 2023-05-23
GIX-1555 Pass confirmation_text on notifyParticipation

dskloetd:
# Motivation

SNS creators can specify a confirmation_text that needs to be confirmed by participants.
This text needs to be included in the notifyParticipation call so the backend can verify the confirmation regardless of which client the participant uses.

# Changes

1. Make the `confirmationText` parameter in the `notifyParticipation` api non-optional.
2. Get the confirmation text from the project summary and pass it in the `notifyParticipation` call.
3. In the api test, pass `confirmationText` and verify that it is passed on to the canister.
4. In `frontend/src/tests/lib/services/sns-sale.services.spec.ts` verify that `confirmationText` is passed on whether it's set or empty and with both initial participation and restored participation.
5. Also remove `snsQueryStore.setData(` which was unused because the project store is mocked.

# Tests

Unit test + manual testing.

### DIFF
--- a/frontend/src/lib/services/sns-sale.services.ts
+++ b/frontend/src/lib/services/sns-sale.services.ts
@@ -11,6 +11,7 @@ import {
   snsProjectsStore,
   type SnsFullProject,
 } from "$lib/derived/sns/sns-projects.derived";
+import { getConditionsToAccept } from "$lib/getters/sns-summary";
 import { loadBalance } from "$lib/services/accounts.services";
 import { getCurrentIdentity } from "$lib/services/auth.services";
 import { toastsError, toastsHide, toastsShow } from "$lib/stores/toasts.store";
@@ -507,9 +508,17 @@ const pollNotifyParticipation = async ({
   identity: Identity;
 }) => {
   try {
+    const project = getProjectFromStore(rootCanisterId);
+    const confirmationText = project && getConditionsToAccept(project.summary);
+
     return await poll({
       fn: (): Promise<SnsRefreshBuyerTokensResponse> =>
-        notifyParticipation({ buyer, rootCanisterId, identity }),
+        notifyParticipation({
+          buyer,
+          rootCanisterId,
+          identity,
+          confirmationText,
+        }),
       shouldExit: isInternalRefreshBuyerTokensError,
       millisecondsToWait: WAIT_FOR_TICKET_MILLIS,
       useExponentialBackoff: true,
--- a/frontend/src/tests/lib/services/sns-sale.services.spec.ts
+++ b/frontend/src/tests/lib/services/sns-sale.services.spec.ts
@@ -36,6 +36,7 @@ import {
   mockPrincipal,
 } from "$tests/mocks/auth.store.mock";
 import {
+  createSummary,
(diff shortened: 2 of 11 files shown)


frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -507,9 +508,17 @@ const pollNotifyParticipation = async ({
   identity: Identity;
 }) => {
   try {
+    const project = getProjectFromStore(rootCanisterId);
+    const confirmationText = project && getConditionsToAccept(project.summary);
```
mstrasinskis, line 512: Probably `??` instead or `&&`?

frontend/src/tests/lib/services/sns-sale.services.spec.ts
```diff
@@ -152,23 +171,10 @@ describe("sns-api", () => {
...
 
-    jest.spyOn(snsProjectsStore, "subscribe").mockImplementation(
-      mockProjectSubscribe([
-        {
-          ...mockSnsFullProject,
-          rootCanisterId: testSnsTicket.rootCanisterId,
-        },
-      ])
-    );
+    setUpMockSnsProjectStore({ rootCanisterId: testSnsTicket.rootCanisterId });
```
mstrasinskis, line 174: Maybe move this to [the only test that uses this default value](https://github.com/dfinity/nns-dapp/pull/2558/files#diff-cf8c6617f5192b309b810bc8fdb587d91f9f94ed5dc441b82924ef0260b2634eL829)?

frontend/src/tests/lib/services/sns-sale.services.spec.ts
```diff
@@ -1066,6 +1126,62 @@ describe("sns-api", () => {
       expect(upgradeProgressSpy).toBeCalledTimes(4);
     });
 
+    it("should pass confirmation_text to notifyParticipation", async () => {
```
mstrasinskis, line 1129: nitpick: `confirmation_text ` -> `confirmationText`?

mstrasinskis (commented):
Looks good to me, just a couple of suggestions.

frontend/src/lib/services/sns-sale.services.ts
lmuntaner: Why didn't you pass it as parameter?

I see it a parameter like `amount`. That now comes from the project could change and I'm not convinced on relying in the store.

You might need to add it also in `initiateSnsSaleParticipation` then to `participateInSnsSale`.

In `restoreSnsSaleParticipation`, you might need to read it from the store and pass it also to `participateInSnsSale`.

mstrasinskis: Thanks, didn't know that about `??`.

frontend/src/tests/lib/services/sns-sale.services.spec.ts
mstrasinskis: Sorry, all's good, I didn't notice that there are more tests that needs this store.

mstrasinskis (approved):
Thank you!

PR #2559, opened 2023-05-24
Bump didc

bitdivine:
# Motivation
We are using an old `didc`.

# Changes
* Update didc version
* Make the dockerfile cache the didc installation only if it succeeds
* Adopt the following changes that come with the new didc:
  * More fields are pub - good.
  * Some generated comments have changed.
  * `Vec<u8>` is now represented as `BufBytes`.

# Tests
-Deploy to app subnet and compare with prod.
  - Ideally needs: https://github.com/dfinity/nns-dapp/pull/2711
  - Deployed manually  Compare one of the SNSs:
    - `diff <(curl https://otgyv-wyaaa-aaaak-qcgba-cai.icp0.io/v1/sns/root/67bll-riaaa-aaaaq-aaauq-cai/slow.json) <(curl https://3r4gx-wqaaa-aaaaq-aaaia-cai.ic0.app/v1/sns/root/67bll-riaaa-aaaaq-aaauq-cai/slow.json)`
    - Comparing all: `<(curl https://otgyv-wyaaa-aaaak-qcgba-cai.icp0.io/v1/sns/list/latest/slow.json | jq .) <(curl https://3r4gx-wqaaa-aaaaq-aaaia-cai.ic0.app/v1/sns/list/latest/slow.json | jq .)`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,11 +76,9 @@ COPY rs/sns_aggregator/Cargo.toml rs/sns_aggregator/Cargo.toml
 RUN mkdir -p rs/backend/src/bin rs/sns_aggregator/src && touch rs/backend/src/lib.rs rs/sns_aggregator/src/lib.rs && echo 'fn main(){}' | tee rs/backend/src/main.rs > rs/backend/src/bin/nns-dapp-check-args.rs && cargo build --target wasm32-unknown-unknown --release --package nns-dapp && rm -f target/wasm32-unknown-unknown/release/*wasm
 # Install dfx
 WORKDIR /
-RUN DFX_VERSION="$(cat config/dfx_version)" sh -c "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
-RUN dfx --version
+RUN DFX_VERSION="$(cat config/dfx_version)" sh -c "$(curl -fsSL https://sdk.dfinity.org/install.sh)" && dfx --version
 # TODO: Make didc support binstall, then use cargo binstall --no-confirm didc here.
-RUN set +x && curl -Lf --retry 5 "https://github.com/dfinity/candid/releases/download/$(cat config/didc_version)/didc-linux64" | install -m 755 /dev/stdin "/usr/local/bin/didc"
-RUN didc --version
+RUN set +x && curl -Lf --retry 5 "https://github.com/dfinity/candid/releases/download/$(cat config/didc_version)/didc-linux64" | install -m 755 /dev/stdin "/usr/local/bin/didc" && didc --version
 RUN curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash && cargo binstall -V
 RUN cargo binstall --no-confirm "wasm-nm@$(cat config/wasm_nm_version)" && command -v wasm-nm
 RUN cargo binstall --no-confirm "ic-wasm@$(cat config/ic_wasm_version)" && command -v ic-wasm
--- a/dfx.json
+++ b/dfx.json
@@ -696,7 +696,7 @@
         "IC_CDK_OPTIMIZER_VERSION": "0.3.1",
         "IDL2JSON_VERSION": "0.8.5",
         "OPTIMIZER_VERSION": "0.3.6",
-        "DIDC_VERSION": "2022-11-17",
+        "DIDC_VERSION": "2023-05-26",
         "IC_COMMIT": "26b480f775222265f4369bc485d502a140f15564"
       },
       "packtool": ""
--- a/rs/sns_aggregator/src/types/ic_sns_governance.patch
+++ b/rs/sns_aggregator/src/types/ic_sns_governance.patch
@@ -1,16 +1,26 @@
 diff --git b/rs/sns_aggregator/src/types/ic_sns_governance.rs a/rs/sns_aggregator/src/types/ic_sns_governance.rs
-index d1acc5d98..25b9ba676 100644
+index 28bc6ec98..fbcff33af 100644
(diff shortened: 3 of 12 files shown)


bitdivine:
Note: The pre-built didc in releases doesn't work on Ubuntu 20.04.  I have prepared two options to make upstream link statically.  A minimal slightly messy one and a more radical approach building everything with zig, which is cleaner but there is some problem installing zigbuild on github runners that I haven't tracked down yet.

Dockerfile
```diff
@@ -67,12 +67,10 @@ COPY rs/sns_aggregator/Cargo.toml rs/sns_aggregator/Cargo.toml
...
-RUN DFX_VERSION="$(cat config/dfx_version)" sh -c "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
-RUN dfx --version
-RUN set +x && curl -Lf --retry 5 "https://github.com/dfinity/candid/releases/download/$(cat config/didc_version)/didc-linux64" | install -m 755 /dev/stdin "/usr/local/bin/didc"
-RUN didc --version
-RUN cargo install "wasm-nm@$(cat config/wasm_nm_version)" && command -v wasm-nm
-RUN cargo install "ic-wasm@$(cat config/ic_wasm_version)"
+RUN DFX_VERSION="$(cat config/dfx_version)" sh -c "$(curl -fsSL https://sdk.dfinity.org/install.sh)" && dfx --version
+RUN set +x && curl -Lf --retry 5 "https://github.com/dfinity/candid/releases/download/$(cat config/didc_version)/didc-linux64" | install -m 755 /dev/stdin "/usr/local/bin/didc" && didc --version
+RUN cargo install "wasm-nm@$(cat config/wasm_nm_version)" && wasm-nm --version
+RUN cargo install "ic-wasm@$(cat config/ic_wasm_version)" && ic-wasm --version
```
bitdivine, line 73: In the code on the left, didc was installed, then tested in another docker step.  However if the test fails, it is unhelpful if the broken installation step is cached.  Putting the installation and test in the same docker step addresses this problem.  The same applies to other tools.

Dockerfile
```diff
@@ -76,11 +76,9 @@ COPY rs/sns_aggregator/Cargo.toml rs/sns_aggregator/Cargo.toml
 RUN mkdir -p rs/backend/src/bin rs/sns_aggregator/src && touch rs/backend/src/lib.rs rs/sns_aggregator/src/lib.rs && echo 'fn main(){}' | tee rs/backend/src/main.rs > rs/backend/src/bin/nns-dapp-check-args.rs && cargo build --target wasm32-unknown-unknown --release --package nns-dapp && rm -f target/wasm32-unknown-unknown/release/*wasm
 # Install dfx
 WORKDIR /
-RUN DFX_VERSION="$(cat config/dfx_version)" sh -c "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
-RUN dfx --version
+RUN DFX_VERSION="$(cat config/dfx_version)" sh -c "$(curl -fsSL https://sdk.dfinity.org/install.sh)" && dfx --version
 # TODO: Make didc support binstall, then use cargo binstall --no-confirm didc here.
-RUN set +x && curl -Lf --retry 5 "https://github.com/dfinity/candid/releases/download/$(cat config/didc_version)/didc-linux64" | install -m 755 /dev/stdin "/usr/local/bin/didc"
-RUN didc --version
```
bitdivine, line 83: On the left, the installation is in a separate step from checking that the installation succeeded.  That means that if installation failed, but exited zero, the failed installation is cached.  Not good.  Hence the change to put installation and check in the same step.  If the installation fails, it will not be cached.

dfx.json
```diff
@@ -696,7 +696,7 @@
         "IC_CDK_OPTIMIZER_VERSION": "0.3.1",
         "IDL2JSON_VERSION": "0.8.5",
         "OPTIMIZER_VERSION": "0.3.6",
-        "DIDC_VERSION": "2022-11-17",
```
bitdivine, line 699: The key change of this PR.

rs/sns_aggregator/src/types/ic_sns_governance.patch
```diff
@@ -1,9 +1,9 @@
 diff --git b/rs/sns_aggregator/src/types/ic_sns_governance.rs a/rs/sns_aggregator/src/types/ic_sns_governance.rs
```
bitdivine, line 1: As didc has changed, the generated rust code is slightly different, mostly in good ways.  The important thing is to check that the rust is correct.  The patch files are a function of didc+rust, interesting only if the patchfile contains changes that are not needed any more, in which case the rust can be updated to exclude the change.

rs/sns_aggregator/src/types/ic_sns_governance.rs
```diff
@@ -8,7 +8,7 @@ use crate::types::{CandidType, Deserialize, Serialize, EmptyRecord};
 use ic_cdk::api::call::CallResult;
 // This is an experimental feature to generate Rust binding from Candid.
 // You may want to manually adjust some of the types.
-// use ic_cdk::export::candid::{self, CandidType, Deserialize, Serialize, Clone, Debug};
+// use candid::{self, CandidType, Deserialize, Serialize, Clone, Debug};
 // use ic_cdk::api::call::CallResult;
```
bitdivine, line 12: Change in a generated comment doesn't change the meaning of the rust, so this is fine.

lmuntaner (approved):
LGTM

I still see it as TODO, did you deploy a version to a testnet? Just to reconfirm that all works as expected.

bitdivine:
@lmuntaner Regarding TODO, yes, deployed and the aggregator has the same data as prod.  Thank you for the reminder to remove the TODO.

PR #2561, opened 2023-05-24
Use arm64 instead of arm in the yq install URL

dskloetd:
# Motivation

`scripts/setup` uses a URL to install `yq` that gives 404.
The correct URL uses `arm64` for the architecture instead of `arm`

# Changes

When constructing the URL to download `yq`, replace `arm` with `arm64`.

# Tests

Tested manually on an M1 Mac.


bitdivine (approved):
Thanks.

PR #2562, opened 2023-05-24
Lazy load ledger and qr-code chunks

peterpeterparker:
# Motivation

Improve performance by lazy loading chunks that are required for specific actions.

# PRs

- [x] https://github.com/dfinity/ic-js/pull/337
- [x] https://github.com/dfinity/ic-js/pull/338

# Changes

- fix QR code chunks download by extracting buffer to a dedicated chunk
- lazy load ledger and proto libs
- bump ic-js next
- add `@dfinity/nns-proto`
- bump gix-cmp for compatibility reason with utils

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -1,7 +1,7 @@
 import inject from "@rollup/plugin-inject";
 import { sveltekit } from "@sveltejs/kit/vite";
 import { readFileSync } from "fs";
-import { dirname } from "path";
+import { basename, dirname } from "path";
 import { fileURLToPath } from "url";
 import type { UserConfig } from "vite";
 
@@ -19,6 +19,14 @@ const config: UserConfig = {
         manualChunks: (id) => {
           const folder = dirname(id);
 
+          // Buffer polyfill is used by various libraries (agent-js, QR code, ledger etc.). That's why we create a specific chunk.
+          if (
+            basename(folder) === "buffer" &&
+            folder.match(/node_modules/g)?.length === 1
+          ) {
+            return "buffer";
+          }
+
           if (
             ["html5-qrcode", "qr-creator"].find((lib) =>
               folder.includes(lib)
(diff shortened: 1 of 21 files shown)


frontend/vite.config.ts
```diff
@@ -19,6 +19,14 @@ const config: UserConfig = {
         manualChunks: (id) => {
           const folder = dirname(id);
 
+          // Buffer polyfill is used by various libraries (agent-js, QR code, ledger etc.). That's why we create a specific chunk.
+          if (
+            basename(folder) === "buffer" &&
```
lmuntaner, line 24: Wasn't there a problem with reproducibility and this chunk?

I see they were remove in [this commit](https://github.com/dfinity/nns-dapp/commit/7ee27433194b5cf135fefce180b96ed36cb60a5e)

frontend/vite.config.ts
```diff
@@ -28,6 +36,24 @@ const config: UserConfig = {
             return "qr";
           }
 
+          if (
+            ["@ledgerhq", "@zondax/ledger-icp"].find((lib) =>
```
lmuntaner, line 40: ditto, they were removed [here](https://github.com/dfinity/nns-dapp/commit/7ee27433194b5cf135fefce180b96ed36cb60a5e)

lmuntaner (commented):
We need to test reproducibility quite a bit before we merge this.

From the PR that changed it:

```
the circular dependency between the chunks seems to
occasionally trip up Rollup again to cause build reproducibility issues
as well as E2e test flakiness.
```

frontend/vite.config.ts
peterpeterparker: In the PR you mention, the problem was that "buffer" was included with ledger related chunk while being used for other features too (agent-js, QR code etc.)

This PR makes a particular chunk for the "buffer" which can then be used by others so reproducibility does not seem to be affected

frontend/vite.config.ts
peterpeterparker: same as https://github.com/dfinity/nns-dapp/pull/2562#discussion_r1203898047

peterpeterparker:
> We need to test reproducibility quite a bit before we merge this.

Yep, it's currently what I am doing. So far it seems that this PR reproducibility isn't worst nor better than on `main`.

frontend/vite.config.ts
lmuntaner: I'm not so sure that was exactly the issue. But it could be. Thanks for the explanation.

lmuntaner (approved):

peterpeterparker:
I ran the `test-reproducibility` script on `main` two times, it failed after around 10 runs on my M2 machine.

> The file "sha256-batch12.txt" is different from "sha256-batch11.txt".

peterpeterparker:
I also ran the same script on this PR locally. Once it failed after 9 tries but twice it failed after more than 20 tries, therefore I assume this PR does not make reproducibility worst.

> The file "sha256-batch28.txt" is different from "sha256-batch27.txt".

PR #2564, opened 2023-05-24
Move AdditionalInfoReviewPo to ParticipateSwapModalPo

dskloetd:
# Motivation

The `AdditionalInfoReview` component is specific to the `ParticipateSwapModal` component and the `TransactionReview` component is more generic.
But the `AdditionalInfoReviewPo` getter is currently defined inside `TransactionReviewPo`

# Changes

Move `getAdditionalInfoReviewPo` from `frontend/src/tests/page-objects/TransactionReview.page-object.ts` to `frontend/src/tests/page-objects/ParticipateSwapModal.page-object.ts`.

# Tests

Existing tests still pass.


lmuntaner (approved):
Good point, thanks!

PR #2565, opened 2023-05-24
chore: extend feature flags

mstrasinskis:
# Motivation

Make voting flag editable to improve testing experience.


lmuntaner (approved):
Thanks!

PR #2566, opened 2023-05-24
Style test banner

peterpeterparker:
# Motivation

Make test banner sticky and review design colors.

# Screenshot

Capture d’écran 2023-05-24 à 11 22 45
Capture d’écran 2023-05-24 à 11 23 20

### DIFF
--- a/frontend/src/lib/components/header/Banner.svelte
+++ b/frontend/src/lib/components/header/Banner.svelte
@@ -20,34 +20,11 @@
 
     localStorage.setItem(localstorageKey, "false");
   };
-
-  $: visible,
-    (() => {
-      if (!browser) {
-        return;
-      }
-
-      if (!banner) {
-        // If no banner has to be displayed, setting or removing the header offset can be skipped
-        return;
-      }
-
-      const {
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Now will be on top of the page, right? LGTM!

peterpeterparker:
> Now will be on top of the page, right?

@lmuntaner not sure what you mean?

The test banner was already at the top but, it added an offset for the content.
With the PR, there is no more offset and the banner appears over (z-index).
The offset always felt a bit weird, was my idea but, felt weird to me too.

lmuntaner:
> > Now will be on top of the page, right?
> 
> @lmuntaner not sure what you mean?
> 
> The test banner was already at the top but, it added an offset for the content. With the PR, there is no more offset and the banner appears over (z-index). The offset always felt a bit weird, was my idea but, felt weird to me too.

I meant exactly what you said. It appears over with the z-index, hehe. Sorry for the miscommunication.

peterpeterparker:
Ah cool then, all good! Thanks for the review

PR #2568, opened 2023-05-24
Fix e2e tsconfig exclude

peterpeterparker:
# Motivation

`tsconfig.spec.json` exclude `e2e` tests which lead to editor not understand `paths`.

# Changes

- extend tsconfig for e2e and reset exclusion of `e2e` path
- add requested `type` (detected once `tsconfig` properly applied)

# Screenshot

Capture d’écran 2023-05-24 à 13 52 13


dskloetd (approved):
I don't understand how this works, but the result looks positive. Thanks!

PR #2569, opened 2023-05-24
Clean up ParticipateSwapModal.spec.ts

dskloetd:
# Motivation

`ParticipateSwapModal.spec.ts` has some redundant test and some tests that don't appear to make sense anymore.

I was very confused trying to figure this out so please review carefully.

# Changes

1. Remove redundant test. See review comments.
2. Change "when user has participated" to "when user has existing swap commitment" for clarity.
3. Move cleanup to the top level `beforeEach`.
4. Resolve the `queryAccounts` results explicitly to make the test that the accounts dropdown is absent until accounts are loaded less brittle.
5. In the test "loads accounts with query only", actually verify that the corresponding update calls aren't made.

# Tests

Verified that if update calls are made, the "query only" test does indeed fail.

### DIFF
--- a/frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
+++ b/frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
@@ -194,7 +194,7 @@ describe("ParticipateSwapModal", () => {
       expect(await form.isContinueButtonEnabled()).toBe(true);
     });
 
-    describe("when user has participated", () => {
+    describe("when user has existing swap commitment", () => {
       it("should move to the last step, enable button when accepting terms and call participate in swap service", async () => {
         const po = await renderEnter10ICPAndNext(
           mockSnsFullProject.swapCommitment
--- a/frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
+++ b/frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
@@ -29,7 +29,8 @@ import {
 import { renderModalContextWrapper } from "$tests/mocks/modal.mock";
 import {
   createSummary,
-  mockSnsFullProject,
+  mockSwapCommitment,
+  createBuyersState,
 } from "$tests/mocks/sns-projects.mock";
 import { rootCanisterIdMock } from "$tests/mocks/sns.api.mock";
 import { ParticipateSwapModalPo } from "$tests/page-objects/ParticipateSwapModal.page-object";
@@ -194,10 +195,13 @@ describe("ParticipateSwapModal", () => {
       expect(await form.isContinueButtonEnabled()).toBe(true);
     });
 
-    describe("when user has existing swap commitment", () => {
+    describe("when user has non-zero swap commitment", () => {
       it("should move to the last step, enable button when accepting terms and call participate in swap service", async () => {
         const po = await renderEnter10ICPAndNext(
-          mockSnsFullProject.swapCommitment
+          {
+            ...mockSwapCommitment,
+            myCommitment: createBuyersState(BigInt(25 * 100000000)),
+          }
         );
 
         const review = po.getTransactionReviewPo();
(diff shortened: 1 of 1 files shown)


frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
```diff
@@ -193,53 +193,9 @@ describe("ParticipateSwapModal", () => {
...
-
-      await review.clickCheckbox();
-      expect(await review.isSendButtonEnabled()).toBe(true);
-
-      await sendAndExpectParticipation(review);
-    });
-  });
-
-  describe("when swapCommitment is empty", () => {
-    describe("when user has not participated", () => {
```
dskloetd, line 217: "when swapCommitment is empty" and "when user has not participated" seem to mean the same thing.
And it's already true for all tests above except "when user has participated".

frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
```diff
@@ -193,53 +193,9 @@ describe("ParticipateSwapModal", () => {
...
-      it("should move to the last step with ICP and disabled button", async () => {
```
dskloetd, line 218: This is a subset of the test on line 149.

frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
```diff
@@ -193,53 +193,9 @@ describe("ParticipateSwapModal", () => {
...
-        const po = await renderEnter10ICPAndNext();
-        const review = po.getTransactionReviewPo();
-        expect(await review.isSendButtonEnabled()).toBe(false);
-      });
-
-      it("should move to the last step, enable button when accepting terms and call participate in swap service", async () => {
```
dskloetd, line 224: This is identical to the test on line 149.

frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
```diff
@@ -193,53 +193,9 @@ describe("ParticipateSwapModal", () => {
...
-
-        await review.clickCheckbox();
-        expect(await review.isSendButtonEnabled()).toBe(true);
-
-        await sendAndExpectParticipation(review);
-      });
-    });
 
     describe("when user has participated", () => {
       it("should move to the last step, enable button when accepting terms and call participate in swap service", async () => {
```
dskloetd, line 198: This is identical to the test on line 198. In the diff it looks like this is not deleted, but it's the test on line 198 that it not deleted and this one is deleted.

frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
```diff
@@ -252,63 +208,75 @@ describe("ParticipateSwapModal", () => {
 
         await sendAndExpectParticipation(review);
       });
-
-      it("should have disabled button if no swap commitment is present", async () => {
```
dskloetd, line 256: This test doesn't do what it says. The continue button is actually enabled.

lmuntaner (approved):
LGTM! Thanks!

I added a comment to change one description which I still find confusing.

frontend/src/tests/lib/modals/sns/ParticipateSwapModal.spec.ts
```diff
@@ -193,53 +193,9 @@ describe("ParticipateSwapModal", () => {
...
-
-        await po.getAdditionalInfoReviewPo().clickCheckbox();
-        expect(await review.isSendButtonEnabled()).toBe(true);
-
-        await sendAndExpectParticipation(review);
-      });
-    });
 
-    describe("when user has participated", () => {
+    describe("when user has existing swap commitment", () => {
```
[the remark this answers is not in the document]
dskloetd: I changed "existing" to "non-zero" and also change the test to explicitly pass an amount rather than assuming that `mockSnsFullProject.swapCommitment` is non-zero.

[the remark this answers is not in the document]
lmuntaner: ok, sounds good

PR #2573, opened 2023-05-25
Split and sort filename and hash in test script

peterpeterparker:
# Motivation

In `test-reproducibility` script, print filename first and sort entries. Makes easier to compare diffs.

Also save results for manual comparison.

### DIFF
--- a/frontend/scripts/test-reproducibility
+++ b/frontend/scripts/test-reproducibility
@@ -12,6 +12,9 @@ fi
 for ((n = 0; n < 100; n++)); do
   npm run build && find public -type f -exec sha256sum {} \; | awk '{print $2 " " $1}' | sort >"$DIR/sha256-batch${n}.txt"
 
+  # Save results for manual comparison
+  cp -r public "$DIR/public-batch${n}"
+
   if ((n > 0)); then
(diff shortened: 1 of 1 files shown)


frontend/scripts/test-reproducibility
```diff
@@ -10,7 +10,10 @@ if [ ! -d "$DIR" ]; then
 fi
 
 for ((n = 0; n < 100; n++)); do
-  npm run build && find public -type f -exec sha256sum {} \; >"$DIR/sha256-batch${n}.txt"
+  npm run build && find public -type f -exec sha256sum {} \; | awk '{print $2 " " $1}' | sort >"$DIR/sha256-batch${n}.txt"
+
+  # Save results for manual comparison
+  cp -r public "$DIR/public-batch${n}"
```
dskloetd, line 16: Does it make sense to `mv` instead of `cp`?

dskloetd (approved):

frontend/scripts/test-reproducibility
peterpeterparker: Sure. I modified `cp` in `mv` and added a `rm -r tmp/...` before it.

PR #2574, opened 2023-05-25
Bump dev dependencies

peterpeterparker:
# Motivation

As far as tested, this does not solve the reproducibility but, bumping the dev dependencies should introduce a non deterministic improvement developed in Rollup.

References:

- Rollup: https://github.com/rollup/rollup/issues/4909
- Vite: https://github.com/vitejs/vite/issues/11911
- SvelteKit: https://github.com/sveltejs/kit/issues/8291

# Notes

I bumped all dev dependencies except Playwright and the types of jest and marked.


frontend/package.json
dskloetd, line 1: I assume you didn't change these manually.
Can you put the command that you used in the PR description?

dskloetd (approved):

frontend/package.json
peterpeterparker: I use a tool to check for dependencies update: https://github.com/raineorshine/npm-check-updates

Once installed locally, I can run `ncu` to check for diff and `ncu -u` to also update the package.json but, it does not filter. So often I run `ncu -u`, then manually revert those dependencies I don't want to update in `package.json` and finally run `npm i`.

It's one way to do it, not like an official way or something.

peterpeterparker: I can add it to `HACKING.md` if it can help?

dskloetd: Would be good. But would also be good to know which ones you chose to revert specifically for this PR. So that it could be reproduced if necessary. If you remember and it's not too difficult.

peterpeterparker: Will do. 

I added a note in PR description about which dev dependencies I bumped.

dskloetd: thanks!

PR #2578, opened 2023-05-25
fix: make participation button tooltip to be on top of the button

mstrasinskis:
# Motivation

Change the participation tooltip position to see 100% of its content.

# Changes

- display tooltip on top
- update the button size on mobile in tooltip

# Screenshot

# Mobile
Screen Recording 2023-05-25 at 15 14 14

## Desktop
image

### DIFF
--- a/frontend/src/lib/components/project-detail/ParticipateButton.svelte
+++ b/frontend/src/lib/components/project-detail/ParticipateButton.svelte
@@ -82,6 +82,7 @@
             <Tooltip
               id="sns-project-participate-button-tooltip"
               text={$i18n.sns_project_detail.max_user_commitment_reached}
+              top={true}
             >
               <button
                 class="primary"
@@ -93,6 +94,7 @@
             <Tooltip
               id="sns-project-participate-button-tooltip"
               text={$i18n.sns_project_detail.not_eligible_to_participate}
+              top={true}
             >
               <button
                 class="primary"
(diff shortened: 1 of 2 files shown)


frontend/src/lib/components/ui/Tooltip.svelte
```diff
@@ -138,10 +138,7 @@
     &.top {
       bottom: unset;
       top: calc(-1 * var(--padding));
-      transform: translate(
-        var(--tooltip-transform-x, var(--tooltip-transform-x-default)),
```
dskloetd, line 142: Wasn't this here for a purpose? Can it just be replaced with `50%`?

dskloetd (commented):
Do you need all these changes or would just `top={true}` mitigate the main issue?
For the other changes, I think it would be best to check with @peterpeterparker

dskloetd (commented):
Do you need all these changes or would just `top={true}` mitigate the main issue?
For the other changes, I think it would be best to check with @peterpeterparker

frontend/src/lib/components/project-detail/ParticipateButton.svelte
```diff
@@ -140,6 +142,12 @@
     justify-content: center;
     padding: var(--padding-2x);
 
+    // To make the button inside the tooltip work the same way as the button w/o a tooltip
+    :global(.tooltip-target) {
```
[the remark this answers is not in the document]
mstrasinskis: It's not so global, only for `.tooltip-target` inside `[role="toolbar"]` for this component instance. But because this is another library (gix-components), svelte doesn't have access to this class in the build time `.tooltip-target`.
image

frontend/src/lib/components/ui/Tooltip.svelte
mstrasinskis: Currently top={true} is used only for the participation button. So it's easy to test. W/o `translate` changes the tooltip is located not in the center 🤷‍♂️

mstrasinskis:
Obsolete.

PR #2583, opened 2023-05-26
Bump snsdemo commit in workflows

dskloetd:
# Motivation

`dfx` often fails to install on CI because it gets the install script from a canister which updates often.
The script that downloads the install script comes from snsdemo and has been changed to retry for longer.

# Changes

Use a newer snsdemo commit for `actions/checkout@v3` in CI workflows.

# Tests

CI


nmattia (approved):
Rubber stamp

PR #2584, opened 2023-05-26
Fix reproducibility by not explicitly importing cjs modules from agent-js in esm

peterpeterparker:
# Motivation

@dskloetd figured out that our frontend reproducibility issue started with PR #2377. After having a look at the PR we noticed it introduced an explicit import of a CJS module of agent-js within the code that is bundled with ESM. As I suspected that our issue was linked to pollyfying the Node modules for the browser, we replaced the import by inline code and as a result, the reproducibility was improved.

Using the `./script/test-reproducibility` script, on my machine, I am now able to run 100 successful check in a row.

# Changes

- remove explicit CJS `import`
- use SvelteKit `browser` information when possible
- when not (see PR for explanation) use a custom `isBrowser` information

### DIFF
--- a/frontend/src/lib/utils/env.utils.ts
+++ b/frontend/src/lib/utils/env.utils.ts
@@ -1,5 +1,17 @@
 import { NNS_IC_ORG_ALTERNATIVE_ORIGIN } from "$lib/constants/origin.constants";
 
+/**
+ * To be use when Rollup fails to resolve import "$app/environment". This can happen because we manually define chunks and SvelteKit is not entirely resilient to bundler reordering.
+ * Example of SvelteKit issue: https://github.com/sveltejs/kit/pull/9808
+ *
+ * Stacktrace of the `npm run build` error that can lead to use `isBrowser` instead of `browser`:
+ *
+ * [vite:worker] [vite]: Rollup failed to resolve import "__sveltekit/environment" from "/.../nns-dapp/frontend/node_modules/@sveltejs/kit/src/runtime/app/environment.js".
+ * This is most likely unintended because it can break your application at runtime.
+ * If you do want to externalize this module explicitly add it to `build.rollupOptions.external`
+ */
+export const isBrowser = typeof window !== "undefined";
+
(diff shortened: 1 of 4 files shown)


frontend/src/lib/utils/env.utils.ts
```diff
@@ -1,5 +1,17 @@
...
+ * To be use when Rollup fails to resolve import "$app/environment". This can happen because we manually define chunks and SvelteKit is not entirely resilient to bundler reordering.
+ * Example of SvelteKit issue: https://github.com/sveltejs/kit/pull/9808
+ *
+ * Stacktrace of the `npm run build` error that can lead to use `isBrowser` instead of `browser`:
+ *
+ * [vite:worker] [vite]: Rollup failed to resolve import "__sveltekit/environment" from "/.../nns-dapp/frontend/node_modules/@sveltejs/kit/src/runtime/app/environment.js".
+ * This is most likely unintended because it can break your application at runtime.
+ * If you do want to externalize this module explicitly add it to `build.rollupOptions.external`
+ */
+export const isBrowser = typeof window !== "undefined";
```
dskloetd, line 13: I'm just wondering, is there a benefit to using `browser` from `"$app/environment"` or would it make sense to use our own version everywhere for consistency?

dskloetd (approved):
Thanks!

frontend/src/lib/utils/env.utils.ts
peterpeterparker: The only really ultra minor benefit I can think of is that `browser` is static while `isBrowser` is evaluated at runtime.

I don't see that particular benefits. Maybe the fact that it is already in place and mocked for the test (`nns-dapp/frontend/__mocks__/$app/environment.ts`)?

No particular opinion if it should or not be replaced, happy with both.

dskloetd: Thanks for clarifying. I'll keep it in mind!

dskloetd:
Failed because of docker flakiness. I've added it to the queue again.

PR #2585, opened 2023-05-26
E2E: Check linked account after page reload

dskloetd:
# Motivation

I want to delete the old (non-Playwright) e2e tests, but before I do I want to make sure that everything they test is covered by the new Playwright e2e tests.

For the accounts related tests, the only thing the old e2e tests test that the new tests don't yet test, is that the linked account is still present after reloading the page.

# Changes

In `frontend/src/tests/e2e/accounts.spec.ts`, reload the page and check that the linked account is still there.

# Tests

`npm run test-e2e`


mstrasinskis (approved):
Great, thanks!

PR #2586, opened 2023-05-26
E2E: Remove race condition work-around

dskloetd:
# Motivation

We used to have a race condition in the accounts store when an update result for an old queryAndUpdate overwrites a query result for a newer queryAndUpdate.
This should be fixed by https://github.com/dfinity/nns-dapp/pull/2445

# Changes

Remove the work-around in frontend/src/tests/e2e/accounts.spec.ts that is no longer necessary.

# Tests

Deployed nns-dapp locally and ran:
`CI=true npm run test-e2e -- account --repeat-each 20 --workers 20`
All 20 passed.


mstrasinskis (approved):
Rubber stamp. Thanks!

PR #2587, opened 2023-05-26
Remove explicit import of lru-cache workaround

peterpeterparker:
# Motivation

Following build reproducibility improvements resulting of #2584, #2574 and #2562, I think we do not need the explicit import of `lru-cache` workaround anymore.

I ran the `./script/test-reproducibility` script which seems fine.

WDYT @dskloetd ?


dskloetd:
If I remember correctly, I was not able to reproduce the issue solved by this import locally but investigated it based on CI runs.
If we want to try this, we'll have to check with `scripts/nns-dapp/ci-consistency` after a while.
I'd rather first have a clean record following our recent effort.

peterpeterparker:
Sure works for me, let's postpone.
I'll keep the PR in a draft, ultimately I'm happy if we can remove explicit import at the root of the app.

PR #2590, opened 2023-05-26
E2E: Use test step to follow progress when headless

dskloetd:
# Motivation

If a test uses `test.step` you can see the name of the step in the terminal when running a headless e2e test.
The intended way to do this is to put all the code for a step inside a closure passed to `test.step()` but this makes the code hard to read and prevents sharing constants created in one step to another step.

# Changes

1. Add a utility function that lets you call `test.step` without passing a closure. It simulates that all the code until the next call to `step()` runs inside the passed closure.
2. Use `step()` inside existing e2e tests to make it easy to follow progress.

# Tests

Tested by running `npm run test-e2e` and observing the result in the terminal.

image

### DIFF
--- a/frontend/src/tests/utils/e2e.test-utils.ts
+++ b/frontend/src/tests/utils/e2e.test-utils.ts
@@ -1,4 +1,17 @@
-import { expect, type BrowserContext, type Page } from "@playwright/test";
+import { expect, test, type BrowserContext, type Page } from "@playwright/test";
+
+let resolvePreviousStep = () => {};
+let previousStep = undefined;
+
+export const step = async (description: string) => {
+  resolvePreviousStep();
+  await previousStep;
+  previousStep = test.step(description, () => {
+    return new Promise((resolve) => {
+      resolvePreviousStep = resolve;
+    });
+  });
+};
 
 // This is useful during debugging to speed up the tests.
 export const signInWithAnchor = async ({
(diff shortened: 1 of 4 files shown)


frontend/src/tests/utils/e2e.test-utils.ts
```diff
@@ -1,4 +1,19 @@
-import { expect, type BrowserContext, type Page } from "@playwright/test";
+import { expect, test, type BrowserContext, type Page } from "@playwright/test";
+
+let resolvePreviousStep = () => {
+  /* this function will be replaced at each step */
+};
+let previousStep = undefined;
+
+export const step = async (description: string) => {
+  resolvePreviousStep();
```
mstrasinskis, line 9: If I got it correctly, `resolvePreviousStep` call finishes the previous step. Does this mean that the last step will be not finished the same way as steps before?

dskloetd: Yes, that's correct. Playwright doesn't care. I've added a comment.

mstrasinskis: Thanks for the info!

mstrasinskis (approved):
LGTM, nice improvement!

mstrasinskis (approved):
LGTM, nice improvement!

PR #2591, opened 2023-05-26
E2E: Remove SNS neuron hotkey

dskloetd:
# Motivation

Continuing with SNS end-to-end test coverage.

# Changes

1. In `frontend/src/tests/e2e/sns-governance.spec.ts` remove the added hotkey.
2. Adding necessary `data-tid`s.
3. Add necessary page object functionality.

# Tests

`npm run test-e2e`

### DIFF
--- a/frontend/src/tests/page-objects/BusyScreen.page-object.ts
+++ b/frontend/src/tests/page-objects/BusyScreen.page-object.ts
@@ -1,0 +1,10 @@
+import { BasePageObject } from "$tests/page-objects/base.page-object";
+import type { PageObjectElement } from "$tests/types/page-object.types";
+
+export class BusyScreenPo extends BasePageObject {
+  static readonly TID = "busy";
+
+  static under(element: PageObjectElement): BusyScreenPo {
+    return new BusyScreenPo(element.byTestId(BusyScreenPo.TID));
+  }
+}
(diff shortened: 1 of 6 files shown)


frontend/src/tests/page-objects/BusyScreen.page-object.ts
```diff
@@ -0,0 +1,10 @@
+import { BasePageObject } from "$tests/page-objects/base.page-object";
+import type { PageObjectElement } from "$tests/types/page-object.types";
+
+export class BusyScreenPo extends BasePageObject {
+  static readonly TID = "busy";
```
mstrasinskis, line 5: Just out of curiosity, why sometimes `TID` is private, and sometimes w/o accessor? And do we really need it when it's being used only once?

mstrasinskis (approved):
LGTM, thank you!

frontend/src/tests/page-objects/BusyScreen.page-object.ts
dskloetd: I guess it can always be private. I'm not sure what you mean by accessor.
It could be inlined, but I always create a new page object by copying an existing one and changing a few things.
Then it's just convenient to have this at the top rather than in the middle of some function.

mstrasinskis: Thank you for the explanation.

PR #2593, opened 2023-05-29
Display spinner on navigation after sign-in

peterpeterparker:
# Motivation

The navigation between login screen to main page can take some times on testnet and mainnet when the JS chunks and other resources have to be downloaded.

Because we have to hide the content of the login mask on navigation, the UI results as being rendered empty while it is loading.

As a bare minimial improvement, we can display a spinner while it's loading.

# Screenshots

Currently - fast network

no_spinner_normal

Currently - slow network

no_spinner_slow

PR - fast network

spinner_normal_speed

PR - slow network

spinner_slow_3g


mstrasinskis (approved):
Nice! Thank you!

PR #2594, opened 2023-05-29
Remove token from "Account" and replace with e8s value

peterpeterparker:
# Motivation

We aim to integrate ICP with ICRC-1 standard. When we introduced a store for the tokens, we did set the duplicate "token" information in `Account` as a deprecated field. This PR uses the `tokensStore` or the `ICPToken` constant for Nns and resolve the deprecation by using a balance set as e8s value  (as provided by the API).

That way, in future PRs, we will be able to remove some duplicate calls in ckBTC and Sns that are use to fetch the token twice.

# Related PRs

Few related PRs were needed to implement this refactoring.

- [x] #2596
- [x] #2597
- [x] #2598
- [x] #2599

# Changes

- main change: in `Account` change `balance` from a `Token` to a `bigint`
- adapt code base and test accordingly

PR #2599, opened 2023-05-30
Pass down token property

peterpeterparker:
# Motivation

We want to use the `TokensStore` to display the token information and want to deprecate the token information within `Account`. This will be solved in #2594.

To achieve this we need to pass down the `token: Token` property to few components instead of pulling the information from the account. 

# Changes

- add `token` as property of few components
- set new property `token` statically if Nns or dynamically from `tokensStore`


frontend/src/lib/components/accounts/WalletSummary.svelte
```diff
@@ -33,14 +34,20 @@
...
+
   let detailedICP: string;
   $: detailedICP = formatToken({
     value: accountBalance.toE8s(),
     detailed: true,
   });
 
   let tokenSymbol: string;
-  $: tokenSymbol = accountBalance.token.symbol;
+  $: tokenSymbol = token?.symbol ?? "";
```
dskloetd, line 50: Since we use `ICPToken` as the fallback on line 40, shouldn't we also use the "ICP" symbol here as the fallback?

Or maybe better: can we make the `token` prop required and not have a fallback?

frontend/src/tests/lib/components/accounts/WalletSummary.spec.ts
```diff
@@ -32,6 +33,7 @@ describe("WalletSummary", () => {
         Component: WalletSummary,
         props: {
           detailedBalance,
+          token: ICPToken,
```
dskloetd, line 36: Can you add tests with different values for `token` (including `undefined`) to make sure it renders the symbol correctly?

frontend/src/tests/lib/components/accounts/AccountCard.spec.ts
```diff
@@ -7,10 +7,15 @@ import type { Account } from "$lib/types/account";
...
 import { mockMainAccount } from "$tests/mocks/accounts.store.mock";
+import type { Token } from "@dfinity/nns";
+import { ICPToken } from "@dfinity/nns";
 import { render } from "@testing-library/svelte";
 
 describe("AccountCard", () => {
-  const props: { account: Account } = { account: mockMainAccount };
+  const props: { account: Account; token: Token } = {
+    account: mockMainAccount,
+    token: ICPToken,
```
dskloetd, line 17: How about adding a test with a different token to make sure it renders correctly?

frontend/src/lib/components/accounts/WalletSummary.svelte
peterpeterparker: I kept the `token` as potentially `undefined` here because that way the check finds place only here and not in the three components that are using this component.

I will modify line 40 to not fallback to `ICPToken`. It's an adaption of the main version that fallback to "0 ICP" for any tokens which does not sound correct.

frontend/src/tests/lib/components/accounts/WalletSummary.spec.ts
peterpeterparker: Isn't this already covered by the related `SnsWallet`, "`CkBTCWallet` and `NnsWallet` test?

To some extension, this PR is a refactoring in the sense that displaying various token is an existing feature, not something new. The way we propagate the token information only is.

frontend/src/lib/components/accounts/WalletSummary.svelte
```diff
@@ -19,6 +19,7 @@
   import type { IntersectingDetail } from "$lib/types/intersection.types";
 
   export let detailedBalance = false;
+  export let token: Token | undefined;
```
dskloetd, line 22: Is the plan to also pass the balance as a prop or will it continue to be passed via the context?
If via the context, would it make sense to also pass the token via the context instead of via a prop?

frontend/src/tests/lib/components/accounts/AccountCard.spec.ts
dskloetd: Thanks for pointing it out.

frontend/src/tests/lib/components/accounts/WalletSummary.spec.ts
dskloetd: I haven't looked at those tests, but if they are well tested, I guess it's covered.

dskloetd (approved):

frontend/src/lib/components/accounts/WalletSummary.svelte
peterpeterparker: Thanks for the feedback, I have made line 40 optionaly undefined and do not display an amount if token is not defined and added a test for undefined.

frontend/src/tests/lib/components/accounts/WalletSummary.spec.ts
peterpeterparker: Thanks. I double checked and tests were partial so added some.

frontend/src/tests/lib/components/accounts/AccountCard.spec.ts
peterpeterparker: Here too it was partial, added two.

frontend/src/lib/components/accounts/WalletSummary.svelte
peterpeterparker: At the moment I don't plan to modify the way the balance, respectively the account, is passed - i.e. keep context.

I was hesitant at using context or prop for the token but, ultimately went for prop. I know Llorenç rather like prop and since both worked fine, I went for that way.

We can for sure refactor latet and move to the context too if we notice it's better.

dskloetd: I thought the plan is to pass a `bigint` instead of `TokenAmount` or `Account`. If that's the case and you touch this code again, I would prefer having both as props over one prop and the rest in the context.

PR #2605, opened 2023-05-31
refactor: rename sumTokenAmounts to sumAmountE8s

peterpeterparker:
# Motivation

Rename `sumTokenAmounts` to `sumAmountE8s` as it sums amount and not token anymore.


dskloetd (approved):
Thanks!

PR #2606, opened 2023-05-31
Remove deprecated ICRC ledger calls

peterpeterparker:
# Motivation

Because we have replaced the balance of the `Account` in Token by the inherited value in E8s (PR #2594), we can remove the calls that were added to fetch the metadata.

In other word we can spare not required query+update calls to ICRC-1 ledgers.

### DIFF
--- a/frontend/src/tests/lib/api/icrc-ledger.api.spec.ts
+++ b/frontend/src/tests/lib/api/icrc-ledger.api.spec.ts
@@ -33,7 +33,7 @@ describe("icrc-ledger api", () => {
       expect(balanceSpy).toBeCalled();
     });
 
-    it("throws an error if no token", () => {
+    it("throws an error if no balance", () => {
       const balanceSpy = async () => {
         throw new Error();
       };
--- a/frontend/src/tests/lib/api/icrc-ledger.api.spec.ts
+++ b/frontend/src/tests/lib/api/icrc-ledger.api.spec.ts
@@ -11,17 +11,14 @@ import {
 
 describe("icrc-ledger api", () => {
   describe("getIcrcMainAccount", () => {
-    const balanceSpy = jest.fn().mockResolvedValue(BigInt(10_000_000));
-
     it("returns main account with balance and project token metadata", async () => {
-      const metadataSpy = jest.fn().mockResolvedValue(mockQueryTokenResponse);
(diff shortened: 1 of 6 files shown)


frontend/src/tests/lib/api/icrc-ledger.api.spec.ts
```diff
@@ -34,11 +31,10 @@ describe("icrc-ledger api", () => {
       expect(account.type).toEqual("main");
 
       expect(balanceSpy).toBeCalled();
-      expect(metadataSpy).toBeCalled();
     });
 
     it("throws an error if no token", () => {
```
dskloetd, line 36: Should this be renamed to `"throws an error if no balance"`?

frontend/src/tests/lib/api/sns-ledger.api.spec.ts
```diff
@@ -24,6 +23,12 @@ const setMetadataSuccess = () => (metadataReturn = mockQueryTokenResponse);
 const metadataSpy = jest
   .fn()
   .mockImplementation(() => Promise.resolve(metadataReturn));
+
+let balanceReturn = Promise.resolve(mainBalance);
```
dskloetd, line 27: To avoid having tests depend on each other, I think `balanceReturn` should be reset in a top-level `beforeEach`.

frontend/src/tests/lib/api/sns-ledger.api.spec.ts
```diff
@@ -55,11 +63,11 @@ describe("sns-ledger api", () => {
       expect(main?.balanceE8s).toEqual(mainBalance);
 
       expect(balanceSpy).toBeCalled();
-      expect(metadataSpy).toBeCalled();
     });
 
     it("throws an error if no token", () => {
```
dskloetd, line 68: Should this be renamed to `"throws an error if no balance"`?

frontend/src/tests/lib/api/sns-ledger.api.spec.ts
peterpeterparker: Sure 👍

frontend/src/tests/lib/api/icrc-ledger.api.spec.ts
peterpeterparker: Yes, thx

dskloetd (approved):

PR #2611, opened 2023-05-31
Review post message types

peterpeterparker:
# Motivation

1. Review post message type definition to make more obvious which message comes from the window to the worker ("Request") and which goes the other way around ("Response")

2. Because we create each time a new worker object, it creates separate worker context. Therefore we can define multiple type declaration per response instead of a single one. That way the response data aren't optional anymore but, can be expected.

Example:

```
export interface PostMessageDataResponse {
  metrics?: MetricsSync;
  canister?: CanisterSync; // <-- optional
  authRemainingTime?: number;
}
```

can become

```
export interface PostMessageDataResponseCanister {
  canister: CanisterSync; // <--- not optional
}
```


frontend/src/lib/components/canisters/CanisterCardCycles.svelte
```diff
@@ -46,9 +46,11 @@
   let canisterSync: CanisterSync | undefined = undefined;
   // Multiple workers that sync canister information can be appended to a view.
   // postMessage being broadcasted, we filter the information that matches this canister.
```
dskloetd, line 48: Is this comment still relevant?

frontend/src/lib/workers/auth.worker.ts
```diff
@@ -10,7 +7,7 @@ import {
 } from "@dfinity/auth-client";
 import { DelegationChain, isDelegationValid } from "@dfinity/identity";
 
-onmessage = ({ data }: MessageEvent<PostMessage<PostMessageDataRequest>>) => {
+onmessage = ({ data }: MessageEvent<PostMessage<never>>) => {
```
dskloetd, line 10: Why `never` and not `PostMessageDataRequestAuth`?
If this function is never called, can we remove it?

frontend/src/lib/workers/metrics.worker.ts
```diff
@@ -7,14 +7,9 @@ import {
...
-  PostMessage,
-  PostMessageDataRequest,
-} from "$lib/types/post-messages";
-
-onmessage = async ({
-  data,
-}: MessageEvent<PostMessage<PostMessageDataRequest>>) => {
+import type { PostMessage } from "$lib/types/post-messages";
+
+onmessage = async ({ data }: MessageEvent<PostMessage<never>>) => {
```
dskloetd, line 12: same (about `never`)

frontend/src/lib/components/canisters/CanisterCardCycles.svelte
peterpeterparker: No, not relevant anymore. Thanks for the findings!

frontend/src/lib/workers/auth.worker.ts
peterpeterparker: It's to express the fact that currently it does not take any `data` in the request (two workers require no data, one does).

```
export interface PostMessage<T extends PostMessageData | never> {
  msg: PostMessageRequest | PostMessageResponse;
  data: T;
}

onmessage = ({ data }: MessageEvent<PostMessage<never>>) => {
```

I can replace with following if you rather like:

```
export interface PostMessage<T extends PostMessageData> {
  msg: PostMessageRequest | PostMessageResponse;
  data: T;
}

onmessage = ({ data }: MessageEvent<PostMessage<PostMessageData>>) => {
```

And also not against better idea!

dskloetd: Thanks for explaining. I thought `never` is supposed to only be for something that cannot happen. Like the return value of a function that always throws.
In that sense, `never` would only be correct here if the function is never called, right?
Something like `type empty = {}` would maybe be more correct here?

I might also help avoid confusion to explicitly declare the `PostMessageDataRequestAuth` type without any fields.

I'll leave it up to you if you like any of these alternatives.

dskloetd (approved):

frontend/src/lib/workers/auth.worker.ts
peterpeterparker: I'll then use the second option `PostMessageData` which equals `object`.

```
export type PostMessageData = object;
```

I used `object` because our linter actually prevents the use of empty `{}`.

> ESLint: 
> Don't use `{}` as a type. `{}` actually means "any non-nullish value".
> - If you want a type meaning "any object", you probably want `object` instead.
> - If you want a type meaning "any value", you probably want `unknown` instead.
> - If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
> - If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead.
> (@typescript-eslint/ban-types)

peterpeterparker: Done. We can for sure iterate these types any time.

dskloetd (approved):

PR #2614, opened 2023-05-31
Refactor/id based proposal navigation

mstrasinskis:
# Motivation

Modifying the ProposalNavigation component to work with IDs only, allowing it to be reused for SNS proposals.

# Changes

- move navigation to proposal component
- ProposalNavigation operate with ids only

# Tests

- ProposalNavigation functionality
- NnsProposal displays the navigation


frontend/src/tests/lib/components/proposal-detail/NnsProposal.spec.ts
```diff
@@ -75,4 +81,21 @@ describe("Proposal", () => {
...
+      .mockImplementation(
+        createMockProposalsStoreSubscribe(generateMockProposals(10))
+      );
+
+    const { container } = renderProposalModern(5n);
+    const { getNextButtonPo, getPreviousButtonPo } = ProposalNavigationPo.under(
+      new JestPageObjectElement(container)
+    );
+
+    // expect(await isPresent()).toBe(true);
```
mstrasinskis, line 97: plz remove

mstrasinskis: Done!

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,67 @@
...
-  } from "$lib/derived/proposals.derived";
-  import { onDestroy } from "svelte";
+  import { createEventDispatcher, onDestroy } from "svelte";
   import { i18n } from "$lib/stores/i18n";
-  import { goto } from "$app/navigation";
-  import { pageStore } from "$lib/derived/page.derived";
-  import { buildProposalUrl } from "$lib/utils/navigation.utils";
 
-  export let proposalInfo: ProposalInfo | undefined;
+  export let proposalIdString: string | undefined;
```
dskloetd, line 6: Given that there are many proposals, does it make sense to give this a more specific name like `currentProposalId`?

dskloetd, line 6: Does this need `String` in the name? Would people otherwise assume it's a number? Would it make sense for it to actually be a number?

If it's a choice between a less confusing API and less code, I would go with the less confusing API.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,67 @@
...
   const reset = () => {
-    nextProposal = undefined;
-    previousProposal = undefined;
+    nextIdString = undefined;
+    previousIdString = undefined;
   };
 
-  $: proposalInfo,
+  $: proposalIdString,
+    proposalIds,
```
dskloetd, line 23: I'm always hesitant when a `$` block needs these extra fields before a function and can't figure out itself when it needs to update.
I think this block could benefit from some pure helper functions so the bit in the `$` block becomes very clean.

```
const getNullableIndexOf = (item, array) => {
  if (isNullish(item) || isNullish(array)) {
    return undefined;
  }
  const index = array.indexOf(item);
  if (index < 0) {
    return undefined;
  }
  return index;
};

const getNullableElement = (array, index) => {
  if (isNullish(array) || isNullish(index) || index < 0 || index >= array.length) {
    return undefined;
  }
  return array[index];
};

$: proposalIndex = getNullableIndexOf(proposalIds, proposalIdString);
$: previousIdString = getNullableElement(proposalIds, proposalIndex - 1);
$: nextIdString = getNullableElement(proposalIds, proposalIndex + 1);
```

What do you think?

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,67 @@
...
   onDestroy(reset);
 
-  let lastProposal: boolean;
-  $: lastProposal =
-    proposalInfo !== undefined &&
-    nextProposal === undefined &&
-    previousProposal === undefined;
+  let singleProposal: boolean;
+  $: singleProposal =
+    proposalIdString !== undefined &&
```
dskloetd, line 43: `nonNullish`?

Also in the rest of the file.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,67 @@
...
-        ({ id }) =>
-          proposalInfo?.id !== undefined &&
-          id !== undefined &&
-          id < proposalInfo.id
-      );
+      previousIdString = proposalIds[index - 1];
+      nextIdString = proposalIds[index + 1];
     })();
 
   onDestroy(reset);
```
dskloetd, line 39: The variables that are cleared in the `rest` function, only exist for the life time of the component, right?
Why do we care about them when the component is destroyed?

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,67 @@
...
 
-  let lastProposal: boolean;
-  $: lastProposal =
-    proposalInfo !== undefined &&
-    nextProposal === undefined &&
-    previousProposal === undefined;
+  let singleProposal: boolean;
+  $: singleProposal =
```
dskloetd, line 42: Would `proposalIds.length === 1` be the same? That seems more intuitively what a "single proposal" is to me.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,67 @@
...
     <button
       class="ghost"
       type="button"
       aria-label={$i18n.proposal_detail.newer}
       on:click={previous}
-      class:hidden={proposalInfo !== undefined &&
-        previousProposal === undefined}
-      disabled={proposalInfo !== undefined && previousProposal === undefined}
+      class:hidden={prevDisabled}
+      disabled={prevDisabled}
```
dskloetd, line 64: Does it need to be both `disabled` and `class:hidden`?

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -2,157 +2,125 @@
...
+import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
 import { render } from "@testing-library/svelte";
-import { get } from "svelte/store";
 
 describe("ProposalNavigation", () => {
-  const props = { proposalInfo: mockProposals[0] };
+  const renderComponent = (props) => {
+    const { container, component } = render(ProposalNavigation, { props });
+    const po = ProposalNavigationPo.under(new JestPageObjectElement(container));
+    return { ...po, component };
```
dskloetd, line 14: Return the `po` itself?

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -2,157 +2,125 @@
...
+  };
 
   describe("not rendered", () => {
-    it("should not render buttons if no proposal", () => {
-      proposalsStore.setProposals({ proposals: [], certified: true });
-
-      const { getByTestId } = render(ProposalNavigation, {
-        props,
+    it("should not render the component if no proposalIdString", async () => {
+      // TODO: is it ok to use "root" here?
```
dskloetd, line 19: Shouldn't be necessary. The `po` has `isPresent()` as well.

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -2,157 +2,125 @@
...
+      const { getPreviousButtonPo, getNextButtonPo } = renderComponent({
+        proposalIdString: "1",
+        proposalIds: ["0", "1"],
       });
 
-      expect(
-        getByTestId("proposal-nav-next")?.classList.contains("hidden")
-      ).toBeTruthy();
+      expect(await getNextButtonPo().isDisabled()).toBe(true);
+      expect(await getPreviousButtonPo().isDisabled()).toBe(true);
```
dskloetd, line 85: Why is the previous button disabled here? Wouldn't it go from "1" to "0"?

frontend/src/tests/page-objects/ProposalNavigation.page-object.ts
```diff
@@ -0,0 +1,15 @@
...
+  private static readonly TID = "proposal-nav";
+
+  static under(element: PageObjectElement): ProposalNavigationPo {
+    return new ProposalNavigationPo(element.byTestId(ProposalNavigationPo.TID));
+  }
+
+  getNextButtonPo = () => this.getButton("proposal-nav-next");
+  getPreviousButtonPo = () => this.getButton("proposal-nav-previous");
+  clickNextProposal = () => this.getNextButtonPo().click();
+  clickPreviousProposal = () => this.getPreviousButtonPo().click();
```
dskloetd, line 14: Would you mind sticking to the pattern of the existing page objects?

If you think it can be improved, let's discuss that separately.

frontend/src/lib/components/proposal-detail/NnsProposal.svelte
```diff
@@ -11,13 +11,37 @@
...
+      buildProposalUrl({
+        universe: $pageStore.universe,
+        proposalId,
+      })
+    );
+  };
 </script>
 
-<ProposalNavigation proposalInfo={$store.proposal} />
+{#if nonNullish($store?.proposal?.id)}
```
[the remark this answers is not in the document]
mstrasinskis: Initially `$store?.proposal` is undefined. But you'r right, because of the current `ProposalNavigation` implementation, this `if` is redundant. Removed.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
mstrasinskis: Good catch! Did it the same way as for the neurons but it's definitely not necessary here since both nns and sns proposal ids are `bitint`s.

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
mstrasinskis: There is the missing `this` if `isPreset` is used directly. Let's check this tomorrow.

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -2,157 +2,125 @@
...
 
-      const { getByTestId } = render(ProposalNavigation, {
-        props,
-      });
+      expect(await root.isPresent()).toBe(false);
+    });
 
-      expect(() => getByTestId("proposal-nav-previous")).toThrow();
-      expect(() => getByTestId("proposal-nav-next")).toThrow();
+    it("should not render buttons if no proposalIdString in proposalIds", async () => {
```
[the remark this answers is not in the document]
mstrasinskis: Now when the `{#if}` from the the parent was removed - yes.

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
mstrasinskis: Good catch, fixed.

frontend/src/tests/page-objects/ProposalNavigation.page-object.ts
mstrasinskis: Yep, let's discuss separately. The lost `this` limits us from using all the language constructions.

mstrasinskis: Refactored, as discussed.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
mstrasinskis: The api was updated, so a separate function is not needed anymore.

frontend/src/lib/components/proposal-detail/NnsProposal.svelte
```diff
@@ -11,13 +11,22 @@
...
   import NnsProposalProposerPayloadEntry from "./NnsProposalProposerPayloadEntry.svelte";
+  import { filteredProposals } from "$lib/derived/proposals.derived";
+  import { navigateToNnsProposal } from "$lib/utils/proposals.utils";
 
   const { store } = getContext<SelectedProposalContext>(
     SELECTED_PROPOSAL_CONTEXT_KEY
   );
+
+  let proposalIds: bigint[] | undefined;
+  $: proposalIds = $filteredProposals.proposals?.map(({ id }) => id as bigint);
```
dskloetd, line 22: If you don't cast to `bigint`, the type is `type ProposalId = bigint`, right?
Maybe using `ProposalId` is appropriate?
Or would that not work for SNS?

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
dskloetd: This question remains.

frontend/src/lib/utils/proposals.utils.ts
```diff
@@ -619,7 +619,7 @@ export const nnsNeuronToVotingNeuron = ({
 });
 
 /** Navigate to the current universe proposal page */
-export const navigateToNnsProposal = (proposalId: string): Promise<void> =>
+export const navigateToNnsProposal = (proposalId: bigint): Promise<void> =>
```
dskloetd, line 622: Maybe use type `ProposalId` instead of `bigint` since that's what `buildProposalUrl` accepts?

frontend/src/tests/page-objects/ProposalNavigation.page-object.ts
```diff
@@ -8,8 +9,19 @@ export class ProposalNavigationPo extends BasePageObject {
...
-  clickPreviousProposal = () => this.getPreviousButtonPo().click();
+  getNextButtonPo(): ButtonPo {
+    return this.getButton("proposal-nav-next");
+  }
+
+  getPreviousButtonPo(): ButtonPo {
+    return this.getButton("proposal-nav-previous");
+  }
+
+  clickNextProposal(): Promise<void> {
```
dskloetd, line 20: Optional:
Since now these functions can't be disconnected from the `this` object, it's fine to use shorter names `clickNext` and `clickPrevious` since we are already in the context of the proposal navigation.

frontend/src/lib/components/proposals/NnsProposalCard.svelte
```diff
@@ -22,13 +22,7 @@
...
     mapProposalInfo(proposalInfo));
 
-  const showProposal = async () =>
-    await goto(
-      buildProposalUrl({
-        universe: $pageStore.universe,
-        proposalId: `${id}`,
-      })
-    );
+  const showProposal = async () => await navigateToNnsProposal(`${id}`);
```
dskloetd, line 25: You pass a `string`, but `navigateToNnsProposal` accepts `bigint`. Does this compile?
Should `navigateToNnsProposal` be changed to accept both `string` and `bigint` (or `ProposalId`).

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,51 @@
...
-    })();
-
-  onDestroy(reset);
-
-  let lastProposal: boolean;
-  $: lastProposal =
-    proposalInfo !== undefined &&
-    nextProposal === undefined &&
-    previousProposal === undefined;
+  export let currentProposalId: bigint | undefined;
```
dskloetd, line 5: Why is `currentProposalId` allowed to be `undefined`?

frontend/src/lib/components/proposal-detail/NnsProposal.svelte
mstrasinskis: This is because the `id` according to types could be optional (api consequences):
```ts
export interface ProposalInfo {
    id: Option<ProposalId>;
```

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
mstrasinskis: I guess the initial idea was to simplify the usage of the component. Would you prefer to move `if not undefined logic` to the client of the component?

frontend/src/lib/components/proposals/NnsProposalCard.svelte
mstrasinskis: Good catch, passing a `string` is a bug.

frontend/src/tests/page-objects/ProposalNavigation.page-object.ts
mstrasinskis: Agree. Done!

frontend/src/lib/components/proposal-detail/NnsProposal.svelte
mstrasinskis: To be honest I don't know the real reason (presumably a backward compatibility). It's `opt` in the `did` file - https://github.com/dfinity/ic-js/blob/e7669a41aec1ee270bf611e72d62abdad9c5f429/packages/nns/candid/governance.did#L387. |
But it's not expected to get no id from the request. Let's ask this tomorrow in our daily.

mstrasinskis: Bigint is used for sns compatibility.

**nns**
```ts
export type ProposalId = bigint;
```

**sns**
```ts
export interface ProposalId {
  id: bigint;
}
```

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
mstrasinskis: The `visibility` css property is used here, not `display: none`. 
```
    &.hidden {
      visibility: hidden;
      opacity: 0;
    }
```

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -2,157 +2,122 @@
...
-import { fireEvent } from "@testing-library/dom";
+import { ProposalNavigationPo } from "$tests/page-objects/ProposalNavigation.page-object";
+import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
 import { render } from "@testing-library/svelte";
-import { get } from "svelte/store";
 
 describe("ProposalNavigation", () => {
-  const props = { proposalInfo: mockProposals[0] };
+  const renderComponent = (props) => {
+    const { container, component } = render(ProposalNavigation, { props });
```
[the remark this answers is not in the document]
mstrasinskis: Thanks, removed.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
mstrasinskis: It could be undefined before the proposal is loaded.

dskloetd: Looking at `NnsProposal.svelte` where this is used:
```
...
</script>

<ProposalNavigation
  currentProposalId={$store.proposal?.id}
  {proposalIds}
  selectProposal={navigateToNnsProposal}
/>

{#if $store?.proposal !== undefined}
  <div class="content-grid" data-tid="proposal-details-grid">
    <div class="content-a content-cell-island">
```

Would it make sense to put the `<ProposalNavigation ... />` inside the `{#if`, and if we do, could we make `currentProposalId` not be `undefined` here?

I think it would simply some of the code below.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,49 @@
...
+      : proposalIds[currentProposalIndex - 1];
+
+  let nextId: bigint | undefined;
+  $: nextId =
+    currentProposalIndex === -1
+      ? undefined
+      : proposalIds[currentProposalIndex + 1];
+
+  const selectPrevious = () => {
+    if (nonNullish(previousId)) {
```
dskloetd, line 28: It's not really possible to call `selectPrevious` if `previousId` is not defined, right? (because in that case the button is hidden).

To be more explicit about that, we could change this to:
```
  const selectPrevious = () => {
    assertNonNullish(previousId);
    selectProposal(previousId);
  }
}
```
What do you think?

It could even be 
```
  const selectPrevious = () => {
    selectProposal(previousId!);
  }
}
```
but we should probably discuss with the team before we decide to use `!` instead of `assertNonNullish`.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,49 @@
...
-      })
-    );
   };
-
-  const previous = async () => {
-    // Type safety check only
-    if (previousProposal === undefined || previousProposal.id === undefined) {
-      return;
+  const selectNext = () => {
+    if (nonNullish(nextId)) {
```
dskloetd, line 33: same

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
mstrasinskis: `currentProposalId` can't be undefined anymore.

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
```diff
@@ -1,97 +1,45 @@
...
-  let lastProposal: boolean;
-  $: lastProposal =
-    proposalInfo !== undefined &&
-    nextProposal === undefined &&
-    previousProposal === undefined;
 </script>
 
-{#if $uiProposals.proposals.length > 0 && !lastProposal}
-  <div role="toolbar">
+{#if nonNullish(currentProposalId) && proposalIds.length > 1}
```
dskloetd, line 35: It's no longer possible for `currentProposalId` to be nullish, is it?

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -2,157 +2,122 @@
...
 
   describe("not rendered", () => {
-    it("should not render buttons if no proposal", () => {
-      proposalsStore.setProposals({ proposals: [], certified: true });
-
-      const { getByTestId } = render(ProposalNavigation, {
-        props,
+    it("should not render the component if no currentProposalId", async () => {
+      const po = renderComponent({
+        currentProposalId: undefined,
```
dskloetd, line 19: This is not valid, is it?

dskloetd (approved):

frontend/src/lib/components/proposal-detail/ProposalNavigation.svelte
mstrasinskis: Good catch, redundant check is removed.

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
mstrasinskis: Thanks, the test was removed.

PR #2617, opened 2023-06-01
Reduce warnings

bitdivine:
# Motivation
Clippy, in pedantic mode, produces a lot of warnings.  It would be nice to reduce them.  Pedantic warnings can be useful but interesting lints are hidden by a mass of uninteresting warnings.

# Changes
- Address about 10% of pedantic warnings.

# Tests
None.  There is no change in functionality so existing CI should suffice.


dskloetd (approved):
Thanks!

PR #2624, opened 2023-06-01
Typo and bug

bitdivine:
# Motivation
There are some minor problems in the proposal verification command.

# Changes
- Fix typo
- Correct the `cd` that does nothing.

# Tests
The command is already tested in CI.


scripts/nns-dapp/release-check
```diff
@@ -17,9 +17,9 @@ source "$SOURCE_DIR/clap.bash"
 clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="mainnet"
 # Source the output file ----------------------------------------------------------
 source "$(clap.build)"
-cd ""
+cd "$SOURCE_DIR/.."
```
bitdivine, line 20: The command on the left does nothing; it has worked only because this has always been run from the repository root.

mstrasinskis (approved):
LGTM, thanks!

PR #2630, opened 2023-06-02
Clean up neurons.services.spec.ts

dskloetd:
# Motivation

Be more confident about the test.

# Changes

In `frontend/src/tests/lib/services/neurons.services.spec.ts`:
1. Set spy implementations in `beforeEach` such that if a test change a spy implementation, it's properly reset for the next test.
2. Remove resetting spy implementation at the end of `"should call setNeuron even if the neuron doesn't have fullNeuron"` as it's now properly reset in `beforeEach`.
3. When expecting a function to be called, expect an explicit number of times.
4. When expecting a function to be called, expect specific arguments.
5. When expecting a function to be called, expect it not to be called before it should have been.
6. Consistently use `toBeCalled` instead of a mix of `toBeCalled` and `toHaveBeenCalled`.

# Tests

Test only


mstrasinskis (approved):
Nice improvement, thank you!

PR #2632, opened 2023-06-02
Controller may upload tarball of assets

bitdivine:
# Motivation
We wish to be able to deploy the nns-dapp to an application subnet.  This, however, means that it needs to be deployed in parts.  Deploying parts requires authorization.  A controller has full control over the canister anyway, so this is an easy case.

- References: https://dfinity.atlassian.net/browse/GIX-1588

# Changes
- Add an API method that allows the controller of a canister to upload an assets tarball.

# Warning
- [x] Before this can be used in CI, we need to bump the dfx version to one that uses a newer replica.  The current dfx uses a very old replica.
  - [x] Bump dfx to 0.14.1 to support is_controller
  - [x] Bump snsdemo dfx to 0.14.1
- [x] Likewise it would be prudent to confirm the replica version used on the NNS subnet.

# Tests
A test is included but it cannot be run on CI, as CI doesn't support `is_controller()`.  The testnets do support it.  Application subnets vary in which replica commit they use; theoretically we could create a canister on a very old subnet and then be unable to install assets via this path until the replica has been upgraded.  I have deployed the nns-dapp to a static testnet, though, and run the test against that deployment.

Merging this without a feature flag requires a dfx version with this commit in the replica:
```
commit c8f3c400c10adc0a44dadcc0dcffbf812a37ce32
Merge: 76ccbe7f90 8f4a04d799
Author: Dragoljub Duric <dragoljub.duric@dfinity.org>
Date:   Thu Mar 30 13:27:59 2023 +0000

    Merge branch 'EXC-1351-implement-is-controller-system-api' into 'master'
    
    feat: Implement is_controller system api
    
    Closes EXC-1351.
    [Description](https://github.com/dfinity/interface-spec/pull/141).
    
    Closes EXC-1351
    
    See merge request dfinity-lab/public/ic!11411
```


rs/backend/nns-dapp-exports.txt
```diff
@@ -10,6 +10,7 @@ canister_query get_transactions
 canister_query http_request
 canister_update <ic-cdk internal> timer_executor
 canister_update add_account
+canister_update add_assets_tar_xz
```
bitdivine, line 13: This is the new API method for adding assets in bulk, rather than individually as before.

lmuntaner (commented):
Did you check the methor `add_stable_asset`?

It does something very similar. Maybe we want to reuse it.

In another note, we can also use the fiduciary subnet to have a dapp in mainnet.

Did we do a design doc regarding the mainnet testing feature?

I don't want to ask question that have already been asked or solved.

Thanks!

bitdivine:
@lmuntaner 
Re: `add_stable_asset` yes.  There are two significant differences:
* With the new API a tarball is uploaded, so two tarballs are enough to upload all the assets.  With the existing method 196 uploads would be needed if we uploaded all the assets, and half that if we bundled some assets in the wasm and uploaded the rest.

bitdivine:
Re: Fiduciary subnet: We have a related problem of consuming too many cycles on upgrade.  If we can upload assets to mainnet after deployment, we solve that problem.  The code offered in this PR is insufficient for that, it only gets us as far as deploying to an app subnet and as controller uploading assets directly, but it moves us in the same direction as the mainnet solution.

bitdivine:
Re: Design doc:  Good idea. There was a discussion but I'm not sure it was written down.

lmuntaner (approved):
LGTM!

Let's discuss with the team how we could use this for mainnet and solve the issue with the cycles and even the payload size.

Can you do a SOP on how to deploy the canister using this method? Do we use the `./build_backend.sh`? Will that create the wasm with the backend only?

Thanks!

bitdivine:
Also, I think the code with the backend-only build hasn't been merged yet, if I recall correctly.

lmuntaner:
> Also, I think the code with the backend-only build hasn't been merged yet, if I recall correctly.

Ok, thanks, I wasn't sure I missed it.

PR #2633, opened 2023-06-02
Bump dfx to v0.14.1

bitdivine:
# Motivation
The version of dfx is out of date.

# Changes
* bump dfx to 0.14.1
* Disable tests for mac 11 that now fail.  Enable tests for the new mac 13 instead.
* Update snsdemo snapshot to one with dfx v0.14.1

# Tests
See CI


bitdivine:
That would be nice but it would man waiting, and there is a queue backed up behind this.  Can we go to 0.14.0 and then to 0.14.1, once it is released?  We have also, in the past, sometimes used a beta version but only when we really needed to.  What do you think?

bitdivine:
My impression is that whatever we do, Mac 11 is now unsupported.  I think we can drop that.  Then the only question is which x in 0.14.x.

peterpeterparker:
Sure we can go first to 14 and then upgrade. There is a tentative to relase v14.1 next week I read.

bitdivine:
I had a look at 0.14.0.  It actually still uses quite an old replica but, still, 0.14.0 is proh=gress.

PR #2635, opened 2023-06-03
One wasm to rule them all

bitdivine:
# Motivation
The wasm we have on mainnet should be identical to the one used on local networks.  Let's test that.

# Changes
- Add a test that compares wasm builds for mainnet and local.

# Notes
- There is a separate issue of whether all parts of the code are using the dynamic configuration passed in with arguments.  That is not covered in this PR.
- There is  a question of where the test should live. I am personally of the opinion that we should build the wasm once, then use that one wasm across many tests.  Why?  Why waste CPU cycles repeating work?  That means that we should not have separate e2e and testing workflows but rather one workflow with separate jobs inside it.

# Tests
See above.

### DIFF
--- a/.github/actions/release_nns_dapp/action.yaml
+++ b/.github/actions/release_nns_dapp/action.yaml
@@ -20,7 +20,7 @@ runs:
         cd "${{ inputs.assets_dir }}"
         daily_build_name="nns-dapp-$(git rev-parse HEAD).wasm"
         cp nns-dapp.wasm "$daily_build_name"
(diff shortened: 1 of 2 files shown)


mstrasinskis:
> There is a question of where the test should live. I am personally of the opinion that we should build the wasm once, then use that one wasm across many tests.

Agree, except reproducibility tests :)

mstrasinskis (approved):
Rubber stamp. Let's test it on CI too.

PR #2636, opened 2023-06-03
Use standard arg parsing for docker-build

bitdivine:
# Motivation
Consistency.  I would like to add another flag but feel as if I am reinventing the wheel when modifying the docker-build arguments.  So why not use the standard argument parser in docker-build?

# Changes
- Use standard argument parsing in `docker-build`
- Surface the `--verbose` flag in the `clap.bash` environment variables, so that it can be used to make the docker output verbose, as before.

# Tests
See CI.  I have also used it locally.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -94,7 +94,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: docker-build prints help
-        run: ./scripts/docker-build --help | grep Usage
+        run: ./scripts/docker-build --help | grep -i usage
       - name: docker-build complains about garbage
         run: |
           { ./scripts/docker-build safdashfdkj || true ; } |& grep 'ERROR: Unsupported flag:'
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -97,7 +97,12 @@ jobs:
         run: ./scripts/docker-build --help | grep -i usage
       - name: docker-build complains about garbage
         run: |
-          { ./scripts/docker-build safdashfdkj || true ; } |& grep -i error
+          { ./scripts/docker-build safdashfdkj || true ; } 2>,stderr
+          grep -qi error ,stderr || {
+            echo "ERROR: Expected to see an error message in:"
+            cat ,stderr
+            exit 1
+          }
   docker-build:
     needs: assets
     runs-on: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -94,7 +94,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: docker-build prints help
-        run: ./scripts/docker-build --help | grep Usage
+        run: ./scripts/docker-build --help | grep -i usage
```
bitdivine, line 97: `clap.bash` uses lowercase "usage".  Man puts section titles in ALL_CAPS without a colon.  Rust's clap uses ALL_CAPS: with a colon.  It might make sense for clap.bash to move to all caps.  However this test here should not care.

scripts/clap.bash
```diff
@@ -132,6 +132,7 @@ while [ \$# -ne 0 ]; do
                         usage
                         exit 0;;
 		--verbose)
+			CLAP_VERBOSE=true
```
bitdivine, line 135: If --verbose is passed, we want to know so that we can make the docker run verbose.

scripts/docker-build
```diff
@@ -1,62 +1,41 @@
...
+source "$(clap.build)"
 
 echo "DFX_NETWORK: $DFX_NETWORK"
 echo "PWD: $PWD"
 
+if [[ "${CLAP_VERBOSE:-}" == "true" ]]; then
+  PROGRESS="--progress=plain"
+else
+  PROGRESS="--progress=auto"
+fi
```
bitdivine, line 37: Arguably --progress should be surfaced as a flag, however traditionally we have used --verbose and it has worked fine.  If it ain't broke, don't fix it.  So this preserves the original behaviour.

.github/workflows/docker-build.yaml
```diff
@@ -94,10 +94,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: docker-build prints help
-        run: ./scripts/docker-build --help | grep Usage
-      - name: docker-build complains about garbage
-        run: |
-          { ./scripts/docker-build safdashfdkj || true ; } |& grep 'ERROR: Unsupported flag:'
```
bitdivine, line 100: Likewise this still produces an error, and passing `--whatever` which really is a flag, unlike `safdashfdkj`, produces an error.  However chasing the differences in error messages seems to have no benefit.  clap.bash already tests that unrecognized flags raise an error.  And non-flags are passed through to docker as positional arguments and if they are wrong, docker will throw an error.

dskloetd (approved):

PR #2638, opened 2023-06-05
Fix lint warning Type '(value: unknown) => void' is not assignable to type '() => void'.

peterpeterparker:
# Motivation

Fix in `e2e.test-utils`:

> TS2322: Type '(value: unknown) => void' is not assignable to type '() => void'.


dskloetd (approved):
Thanks!

PR #2640, opened 2023-06-05
Remove builds for different networks in CI

bitdivine:
# Motivation
We used to require different WASM builds for every network.  Now that we have one wasm for everywhere, pretty much all the builds should really be the same.  So we can clean up the branching logic in CI and use one standard build everywhere.

Exceptions:
- Verifying that the build really has the same WASMs for every network.
- Uploading the docker cache on merge to main.  Here we don't need the additional steps needed to compute more arguments.

# Changes
- Define a standard build.
- Use snsdemo scripts to install tools, instead of downloading them directly.
- Use it everywhere, removing variants and branches that we don't need any more.

# Tests
See CI


.github/actions/build_nns_dapp/action.yaml
bitdivine, line 1: This is what I would like to propose as the standard build.  I believe it covers all cases apart from populating the docker cache.

.github/workflows/aggregator.yaml
bitdivine, line 1: A standard replacement.

.github/workflows/build.yml
bitdivine, line 1: A standard replacement, except that the location of the out dir also needs to be changed.  Note that the dir is _not_ included in the uploaded artefact.

.github/workflows/build.yml
```diff
@@ -285,7 +276,7 @@ jobs:
           context: .
           file: Dockerfile
           build-args: |
-            DFX_NETWORK=mainnet
+            DFX_NETWORK=local
```
bitdivine, line 279: As the standard build uses mainnet as the target, to compare we need to build for local.

.github/workflows/docker-build.yaml
bitdivine, line 1: This workflow used to have a matrix to build for different networks.  Now we build the wasm for all networks in one go, so the matrix is not needed.

.github/workflows/nightly.yaml
bitdivine, line 1: A standard replacement.  I have added the step of recording the git log head in the standard build.

.github/actions/build_nns_dapp/action.yaml
bitdivine, line 1: This is what I would like to propose as the standard build. I believe it covers all cases apart from populating the docker cache.

.github/workflows/aggregator.yaml
bitdivine, line 1: A standard replacement.

.github/workflows/docker-build.yaml
bitdivine, line 1: This workflow used to have a matrix to build for different networks. Now we build one wasm for all networks, so the matrix is not needed.

.github/workflows/nightly.yaml
bitdivine, line 1: A standard replacement. I have added the step of recording the git log head in the standard build.

.github/actions/build_nns_dapp/action.yaml
```diff
@@ -0,0 +1,67 @@
...
+          command -v didc || {
+            echo "Failed to install didc"
+            exit 1
+          }
+        fi
+    - name: Create local arguments
+      shell: bash
+      run: |
+        set -euxo pipefail
+        DFX_NETWORK=local ./config.sh && cp nns-dapp-arg-local.did out && didc encode "$(cat out/nns-dapp-arg-local.did)" | xxd -r -p >out/nns-dapp-arg-local.bin
```
bitdivine, line 67: Note: I would prefer to create these arguments in docker, however the main goal of this PR is to establish a single standardized build, not to polish that standardized build, so I have implemented the simplest thing possible here.

.github/actions/build_nns_dapp/action.yaml
```diff
@@ -0,0 +1,67 @@
...
+      run: |
+        set -euxo pipefail
+        git log | head -n1 > out/commit.txt
+    - name: Install dfx
+      shell: bash
+      run: |
+        set -euxo pipefail
+        if command -v dfx
+        then echo "Skipping, as dfx is already installed"
+        else DFX_VERSION="$(jq -r .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
```
dskloetd, line 44: In other places we use the script from snsdemo (which retries for up to 8 minutes). Is there any reason we can't use that here?

.github/actions/build_nns_dapp/action.yaml
```diff
@@ -0,0 +1,67 @@
...
+      run: |
+        set -euxo pipefail
+        if command -v didc
+        then echo "Skipping, as didc is already installed"
+        else
+          USER_BIN="$HOME/.local/bin"
+          mkdir -p "$USER_BIN"
+          echo "$USER_BIN" >> $GITHUB_PATH
+          version="$(jq -r .defaults.build.config.DIDC_VERSION dfx.json)"
+          curl -Lf "https://github.com/dfinity/candid/releases/download/${version}/didc-linux64" | install -m 755 /dev/stdin "$USER_BIN/didc"
```
dskloetd, line 57: Don't we have a script to install `didc`?

dskloetd (approved):

dskloetd (changes requested):
Actually, is this safe given that we still have https://dfinity.atlassian.net/browse/GIX-1584 ?

bitdivine:
@dskloetd - why should it not be safe?  Here the one build is a mainnet build.  Assertion: The wasm and args are identical to the wasm on main, when this PR is up to date with main.  If that assertion is true, would you be satisfied that this is safe?

.github/actions/build_nns_dapp/action.yaml
bitdivine: We could, afaik.  I'm not convinced it has much benefit given that we are using only one of its many options.  But I don't mind.

bitdivine: Done.

.github/actions/build_nns_dapp/action.yaml
bitdivine: Done.  It does add a dependency on another repo for a one-liner but as long as it works reliably I don't mind.

bitdivine:
@dskloetd  I have copmpared the wasm hash with main.  Given that the commit is included in the metadata, it is necessary to extract the wasm before the metadata is added and compare that.  With that:

Main @ 5bf5ac78812a9cf2af599612ee49e7caea81f442
```
max@sinkpad:~/dfn/nns-dapp (53:19)$ sha256sum out/nns-dapp-nometa.wasm 
1fa88924ba2d7558ef9067cce9e56811d6ef61c7458fa57f0a9601e283ffcfda  out/nns-dapp-nometa.wasm
max@sinkpad:~/dfn/nns-dapp (53:41)$ 
```
This PR @ a2590775305f3254b18da5acfdb14b13f9929537
```
max@sinkpad:~/dfn/nns-dapp/branches/build-nns-dapp-action (9:52:22)$ sha256sum out/nns-dapp-nometa.wasm 
1fa88924ba2d7558ef9067cce9e56811d6ef61c7458fa57f0a9601e283ffcfda  out/nns-dapp-nometa.wasm
max@sinkpad:~/dfn/nns-dapp/branches/build-nns-dapp-action (9:55:35)$ 
```

.github/workflows/build.yml
```diff
@@ -70,8 +61,8 @@ jobs:
       - name: Start snapshot environment
         uses: ./.github/actions/start_dfx_snapshot
         with:
-          # Version from 2023-05-24 with longer retries to install dfx
-          snsdemo_ref: '94a006c4a4b77d7153071dee74d80939a22c835e'
+          # Version from 2023-06-06 with dfx v 0.14.1
+          snsdemo_ref: '0e61f618018db099a01b1686535fff8eff980d1c'
```
dskloetd, line 65: Can we limit the scope of the PR and not also upgrade to a different dfx version at the same time?
Also, the snapshot on the line below won't work with the new version, right?

.github/workflows/aggregator.yaml
```diff
@@ -15,26 +15,17 @@ jobs:
     steps:
       - name: Checkout nns-dapp
         uses: actions/checkout@v3
-      - name: Set up docker buildx
-        uses: docker/setup-buildx-action@v2
-      - name: Build sns-aggregator image
-        uses: docker/build-push-action@v3
+      - name: Build nns-dapp
+        uses: ./.github/actions/build_nns_dapp
```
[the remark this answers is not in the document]
bitdivine: The docker build builds both.

.github/workflows/build.yml
bitdivine: dfx has already been bumped in both `nns-dapp` and `sns_demo`.  But, sure, the snapshot could be bumped in another PR.  Here: https://github.com/dfinity/nns-dapp/pull/2680

If the snapshot doesn't work, CI will fail.

.github/workflows/aggregator.yaml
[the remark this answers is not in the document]
dskloetd: Then I think you should give the action a less misleading name.
How about naming the action `docker_build`?

bitdivine:
Bump. @dskloetd

.github/workflows/aggregator.yaml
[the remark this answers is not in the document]
dskloetd: This doesn't seem to be addressed.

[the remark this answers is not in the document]
bitdivine: I think that building the assets is the point of the action.  How it does so is an implementation detail.  I have however clarified that nns-dapp refers to the repo, not the canister.

.github/workflows/build.yml
```diff
@@ -19,38 +19,29 @@ jobs:
     steps:
       - name: Checkout nns-dapp
         uses: actions/checkout@v3
-      - name: Set up docker buildx
-        uses: docker/setup-buildx-action@v2
-      - name: Build wasms
-        uses: docker/build-push-action@v3
+      - name: Build nns-dapp
+        uses: ./.github/actions/build_nns_dapp
```
dskloetd, line 23: If you must insist on keeping the confusing name of this action, please add a comment here:
```
# .github/actions/build_nns_dapp builds all the assets, including the sns_aggregator.
```

bitdivine: The action includes that information in the description. There is no need to duplicate it.

bitdivine: I have clarified the meaning of nns-dapp in the step title, though.

dskloetd: Please add the comment.

bitdivine: Thank you for your opinion on this but I do not believe that it is appropriate or maintainable.  There is one action, it is documented.  The documentation should not be repeated with every one of the calls.

PR #2641, opened 2023-06-05
Extract interfaces types for worker services

peterpeterparker:
# Motivation

Extract interfaces types for worker services instead of duplicating types in components

# Changes

No features. Only type declarations extracted.

### DIFF
--- a/frontend/src/lib/services/$public/worker-metrics.services.ts
+++ b/frontend/src/lib/services/$public/worker-metrics.services.ts
@@ -3,12 +3,12 @@ import type { PostMessage } from "$lib/types/post-messages";
 
 export type MetricsCallback = (data: PostMessageDataResponseMetrics) => void;
 
-export interface InitMetricsWorker {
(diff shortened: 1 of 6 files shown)


frontend/src/lib/services/$public/worker-metrics.services.ts
```diff
@@ -3,7 +3,12 @@ import type { PostMessage } from "$lib/types/post-messages";
 
 export type MetricsCallback = (data: PostMessageDataResponseMetrics) => void;
 
-export const initMetricsWorker = async () => {
+export interface InitMetricsWorker {
```
dskloetd, line 6: Why do all these types start with `Init`? Are these not the workers themselves? What is "Init" in the name indicating?

peterpeterparker: Sure we can do that. Done.

dskloetd (approved):

PR #2639, opened 2023-06-05
Refresh accounts balance and transactions periodically

peterpeterparker:
# Motivation

WIP

# PRs

WIP

# Changes

WIP

PR #2642, opened 2023-06-05
Expose ledger canister id in sns summary

peterpeterparker:
# Motivation

Regardless if Sns, ckBTC or ICP we want to be able to fetch balance and transactions uses the ICRC-1 ledger.

In Snses we generally use the "wrapper" to access the canister but, in the particular crontab feature #2639 we want to avoid the use of such object and use `@dfinity/ledger` library directly regardless of the context.

That is why we need to access the ledger canister Id of a Sns project.

# Changes

- export `ledgerCanisterId` in `SnsSummary` (as we already did for swap and governance)

### DIFF
--- a/frontend/src/lib/api/sns.api.ts
+++ b/frontend/src/lib/api/sns.api.ts
@@ -143,7 +143,7 @@ export const querySnsSwapState = async ({
 
   const {
     swapState,
-    canisterIds: { swapCanisterId, governanceCanisterId },
+    canisterIds: { swapCanisterId, governanceCanisterId, ledgerCanisterId },
   }: SnsWrapper = await wrapper({
     rootCanisterId,
     identity,
@@ -160,6 +160,7 @@ export const querySnsSwapState = async ({
     rootCanisterId,
     swapCanisterId,
     governanceCanisterId,
+    ledgerCanisterId,
     swap,
     derived,
     certified,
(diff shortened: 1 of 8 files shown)


frontend/src/lib/api/sns.api.ts
dskloetd, line 1: Is there no test for this file that needs to be updated? The main change of this PR seems to be untested.

peterpeterparker: The new property is basically a clone of existing properties `governanceCanisterId` and `swapCanisterId` and used at the exact same place in code and test. So it is covered by test the exact same way those two other properties are covered.

dskloetd: If the `ledgerCanisterId` accidentally gets the same value as `governanceCanisterId` instead of its own value, would any test break?

peterpeterparker: I guess no because `swapCanisterId` has in some mock the same value as `governanceCanisterId`.

Again, same same. If you think it is worth adding some tests for it, then it's probably worth adding tests for all of these three ids.

dskloetd: I agree. They should all be tested. But when I review code, I assume that the existing code is already tested.

dskloetd (approved):

frontend/src/lib/api/sns.api.ts
peterpeterparker: Thanks for the review

PR #2644, opened 2023-06-05
Refactor Cjs import and common functions in a dedicated utils

peterpeterparker:
# Motivation

Refactor (move) Cjs dedicated import and common functions in a specific utils.

That way we spare the duplicate code and we make the functions also available for upcoming workers such as in #2639.

# Changes

- move import (and rexport)  `"@dfinity/agent/lib/cjs/index"` to `cjs.utils`
- move a common `createCanisterCjs` to `cjs.utils`
- add the `if fetch_root_key then fetch_root_key` to that create function

### DIFF
--- a/frontend/src/lib/utils/cjs.utils.ts
+++ b/frontend/src/lib/utils/cjs.utils.ts
@@ -1,0 +1,38 @@
+import { FETCH_ROOT_KEY, HOST } from "$lib/constants/environment.constants";
+import type { Identity } from "@dfinity/agent";
+import type { Principal } from "@dfinity/principal";
+
+/**
+ * HTTP-Agent explicit CJS import for compatibility with web worker - avoid Error [RollupError]: Unexpected token (Note that you need plugins to import files that are not JavaScript)
+ */
+import { HttpAgent, getManagementCanister } from "@dfinity/agent/lib/cjs/index";
+
+export {
+  HttpAgent as HttpAgentCjs,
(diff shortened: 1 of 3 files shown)


frontend/src/lib/utils/cjs.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+}: {
+  identity: Identity;
+  create: (params: CreateCanisterCjsParams) => T;
+} & Pick<CreateCanisterCjsParams, "canisterId">): Promise<T> => {
+  const agent = new HttpAgent({
+    identity,
+    host: HOST,
+  });
+
+  if (FETCH_ROOT_KEY) {
```
dskloetd, line 33: Why is this added? As far as I can tell, the code that this replaces didn't do this before.

dskloetd (approved):

frontend/src/lib/utils/cjs.utils.ts
peterpeterparker: We did not had to use the root key because this actor init function was used in a single feature so far that did not needed it, because it instantiates a canister that is not local but, in #2639 we will need local canister.

It's also standard on the IC to have such a "if fetch_root then fetch_root" in an actor initialization function.

peterpeterparker: Thanks for the review

PR #2645, opened 2023-06-06
Apple install to home screen look and feel

peterpeterparker:
# Motivation

The new release has been deployed on the mainnet, allowing me to test the installation of the dapp on the iOS home screen. However, it seems that Apple displays the icons over a black background (ignoring theming properties). Since the icon is designed as a circle, it doesn't look visually appealing.

# Screenshot

IMG_0534


mstrasinskis (approved):
Rubber stamp. Thanks!

PR #2650, opened 2023-06-06
Inherits params in worker cycles

peterpeterparker:
# Motivation

Following PR #2641 which extracts interfaces types for worker services, I noticed that I can simplify few parameters typings in `worker-cycles.services.ts`.


dskloetd (approved):
It does

dskloetd:
I don't know how that comment ended up there.

PR #2656, opened 2023-06-07
Unit test for 'Excluded Persons'

dskloetd:
# Motivation

Was missing from https://github.com/dfinity/nns-dapp/pull/2653

# Changes

1. Test that excluded countries are rendered if present and not if not.

# Tests

unit test


bitdivine (approved):
Rubber stamp

PR #2657, opened 2023-06-07
Robust docker build

bitdivine:
# Motivation
Docker builds are failing for Mac users.

# Changes
* I have added CI tests in the snsdemo repo to make sure that clap works on all mac releases, providing that a recent bash is installed.  It does.
* It remains to confirm that the local user has a recent bash installed.
* `realpath` is removed as the local version doesn't work on mac.
* Added a help text, saying how to run with a clean cache if docker fails.

# Tests
Please, everyone, test on your local machines
- [x] @dubovitskaya - Intel mac
- [ ] @dskloetd - M1 mac, fails to build
- [ ] @mstrasinskis - M1, unlikely to build.
- [x] @bitdivine on intel mac
- [x] @bitdivine on Ubuntu 20.04


scripts/docker-build
```diff
@@ -1,6 +1,19 @@
 #!/usr/bin/env bash
 set -euo pipefail
-SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+(("${BASH_VERSINFO[0]}" >= 5)) || {
+  echo "ERROR:  Please use a newer version of bash.  The minimum supported bash version is 5.  Yours appears to be '${BASH_VERSINFO[0]}'" >&2
+  exit 1
+}
+
+if test -d scripts && test -e Dockerfile; then
```
dskloetd, line 9: Does it make sense to check that $(pwd)/scripts == $(dirname $0)
Or do you intentionally support running script from a different worktree?

scripts/docker-build
```diff
@@ -16,6 +29,16 @@ print_help() {
 	  - "nns-dapp-arg-${DFX_NETWORK}.did"
 	  - "nns-dapp-arg-${DFX_NETWORK}.bin"
 
+	EOF
+  print_docker_help
+}
+
+print_docker_help() {
+  cat <<-"EOF"
+	Note: If the docker build fails, it may help to build from a clean cache:
```
dskloetd, line 38: Can we check that the build failed and output this as the last line if it did?

dskloetd:
I tried the docker build many times but it always fails on this line of `Dockerfile`:
```
RUN mkdir -p rs/backend/src/bin rs/sns_aggregator/src && touch rs/backend/src/lib.rs rs/sns_aggregator/src/lib.rs && echo 'fn main(){}' | tee rs/backend/src/main.rs > rs/backend/src/bin/nns-dapp-check-args.rs && cargo build --target wasm32-unknown-unknown --release --package nns-dapp && rm -f target/wasm32-unknown-unknown/release/*wasm
```
It always fails like this:
```
#30 2831.1    Compiling termcolor v1.2.0
#30 2831.9 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#30 2831.9 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#30 2837.4    Compiling data-encoding v2.3.3
#30 2838.3 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#30 2838.3 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#30 2838.9 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#30 2838.9 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#30 2864.7 bash: line 1:    19 Killed                  cargo build --target wasm32-unknown-unknown --release --package nns-dapp
------
executor failed running [bash -c mkdir -p rs/backend/src/bin rs/sns_aggregator/src && touch rs/backend/src/lib.rs rs/sns_aggregator/src/lib.rs && echo 'fn main(){}' | tee rs/backend/src/main.rs > rs/backend/src/bin/nns-dapp-check-args.rs && cargo build --target wasm32-unknown-unknown --release --package nns-dapp && rm -f target/wasm32-unknown-unknown/release/*wasm]: exit code: 137
+ set +x
ERROR: Docker build failed.
Note: If the docker build fails, it may help to build from a clean cache:

  ./scripts/docker-build -- --no-cache
```
except that it is always compiling something different when it fails. So it's not anything in particular that's broken but it just runs out of resources or something.

I previously made the docker build work for me with [this commit](https://github.com/dfinity/nns-dapp/commit/556959b09fd92fcdeed1e977fb635486c1cae405) so I tried if I was still able to run the docker build at that commit and I was. So then I tried binary search to find where it broke but after spending a lot of time, I realized that whether the build passed wasn't related to what it was building but to whether it was able to use the docker cache.

After running `docker system prune -a` the build also failed at that commit, although after running several times, it managed to pass and now it can pass again at that commit. But at HEAD, I wasn't able to make it pass even after running many times.

I'm wondering if we can compile the dependencies in separate steps so that if it fails at some point, at least it's able to pick up from the cache the next time and make progress.

scripts/docker-build
bitdivine: `$(pwd)/scripts == $(dirname $0)` is not usually true.  Yes, I use multiple worktrees and would like to continue doing so.

bitdivine:
I ran docker-build on an intel mac.  The build failed due to insufficient RAM.  I increased the RAM and it succeeded.  So I have added a RAM check.  The check is inside docker, as a check outside would be different depending on the host OS.  The idea is that now, if the docker RAM is insufficient, the user will get a clear error message to that effect.

README.md
```diff
@@ -18,28 +18,86 @@ can validate that we really deploy what we claim to deploy.
...
+* Get the code
+  ```sh
+  git clone https://github.com/dfinity/nns-dapp.git && cd nns-dapp
+  ```
+* If you would like to build a specific version, e.g. the commit for a specific proposal, please check it out:
+  ```sh
+  git checkout THE_COMMIT
+  ```
+* Verify that docker is running.  If not, please start it.
+  * Mac: Press apple-space and enter docker
```
dskloetd, line 84: `Cmd + space` instead of `apple-space`

scripts/docker-build
```diff
@@ -56,14 +79,26 @@ COMMIT="$(git rev-parse HEAD)"
 [ -z "$(git status --untracked-files=no --porcelain)" ] || COMMIT="${COMMIT}-dirty"
 
 set -x
-DOCKER_BUILDKIT=1 docker build \
+docker build --target check-environment .
+
+set -x
```
dskloetd, line 84: This looks redundant with line 81?

dskloetd (approved):

README.md
bitdivine: I have used the symbol just to be sure.

dskloetd: I honestly think that's more confusing. I've never seen it referred to like that. The symbol renders really small and it's hard to see what it is. I've always seen it being referred to as `Cmd`. And also please use `+` instead of `-` to indicate combining keys.

PR #2658, opened 2023-06-07
Display "Select NNS" instead of "Select Token" on proposals

peterpeterparker:
# Motivation

In universe selector, on the proposal tab, display "Select Network Nervous System" instead of "Select Token"


mstrasinskis (approved):
LGTM, thanks!

PR #2664, opened 2023-06-08
Fix Sns proposal lazy loading

peterpeterparker:
# Motivation

The Sns proposals list ends in an infinite loop. The root cause of the issue is the fact that we paginate the list of proposals but fails at providing the correct index to search next proposals - i.e. the UI ends providing always the same index and therefore the query is repeated over and over again.

Worth to note that the issue happens when filters are selected - i.e. when only a subset of the proposals are displayed.

# Changes

- resolve index (`beforeProposalId`) for the next query with list of all Sns proposals that have been queried for a universe instead of using the derived store that contains only those that are displayed

# Screenshots

Bug

bug

Fix. Note the brief glitch in the screenshot is expected with current UI. When fetching a spinner/skeleton is briefly displayed therefore until the pagination is over, these glitchs appears. This PR does not improve the UI.

fix


peterpeterparker:
> Definitely a bug, but at least on desktop, mainnet doesn't have enough SNSs to need scrolling. So, nice find, but the fix doesn't seem urgent to me. 

It's incorrect @bitdivine.

- If you use the filters you are able to display more proposals than displayed in the viewport on mainnet already
- As badly described above in summary of the PR, this does not affect only what's visible and might be useful even if there is a subset displaye

That said, if we do a patch to disable Sns voting, we indeed to not need to include it.

dskloetd:
> Does that sound right or have I missed something?

You've missed that these aren't SNSes. These are proposals.

PR #2668, opened 2023-06-08
Enable simulating merge hotkey neurons

dskloetd:
# Motivation

Simulating merging hotkey-controlled neurons is now supported by the governance canister and I tested it with a snapshot based on the latest ic commit.

# Changes

1. Remove the check that a neuron is not hardware wallet controlled.
2. Use the authenticated identity instead of the controller identity.
3. Updated the test accordingly.

# Tests

1. Unit test is updated in the PR.
2. Tested manually with the latest backend canisters that it works.


lmuntaner (approved):
Great, thanks!

PR #2669, opened 2023-06-08
Use the public URL as the staging host

bitdivine:
# Motivation
This is publicly accessible, so can be used for reviews of a release candidate:

https://ptid5-maaaa-aaaaa-aacnq-cai.nnsdapp.dfinity.network/

This is not:

https://ptid5-maaaa-aaaaa-aacnq-cai.nnsdapp.testnet.dfinity.network/

The release SOP actually states that the testnet needs to be removed and I was shown how to edit dfx to remove the testnet.

We actually never need the `testnet` in that URL.

# Changes
* Remove `testnet` from the staging URL.

# Tests
* Deployments to staging already have this change applied manually.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -186,7 +186,7 @@
     },
     "staging": {
       "config": {
-        "HOST": "https://nnsdapp.testnet.dfinity.network",
+        "HOST": "https://nnsdapp.dfinity.network",
         "FEATURE_FLAGS": {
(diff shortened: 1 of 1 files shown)


dskloetd:
Thanks! I had been wondering why it's there while there is a step in the release SOP to remove. If this gets merged, we should update the checklist.

PR #2671, opened 2023-06-12
Use fake governance API in MergeNeuronsModal.spec.ts

dskloetd:
# Motivation

I'll be making changes to this test when implementing the simulate merging neurons UI so now seems like a good time to change it from mocking the service and store to using a fake.

# Changes

In `frontend/src/tests/fakes/governance-api.fake.ts`:
1. Add `mergeNeurons`. It only adds the stakes together in the target neuron and sets the stake to 0 in the source neuron. This is very simplistic but should be good enough to verify in the test that neurons have been merged.
2. Add `getNeuron({identity, neuronId})` for use in the fake itself as well as in tests for convenience.
3. Allow setting the `stake` on neurons in the fake.
4. Add `addNeurons` to insert multiple fake neurons with a single call.

In `frontend/src/tests/lib/modals/neurons/MergeNeuronsModal.spec.ts`:
1. Create and use a test identity to make sure the authenticated identity, the controller of the neurons and the principal of the main account all align.
2. Insert neurons into the fake API instead of in the store.
3. Verify the result of the mergeNeurons call instead of only expecting `mergeNeurons` to be called on the service.

# Tests

Test only

### DIFF
--- a/frontend/src/tests/fakes/governance-api.fake.ts
+++ b/frontend/src/tests/fakes/governance-api.fake.ts
@@ -11,7 +11,7 @@ import {
 } from "$tests/utils/module.test-utils";
 import type { Identity } from "@dfinity/agent";
 import type { KnownNeuron, NeuronInfo, RewardEvent } from "@dfinity/nns";
-import { isNullish } from "@dfinity/utils";
+import { isNullish, nonNullish } from "@dfinity/utils";
 
 const modulePath = "$lib/api/governance.api";
 const fakeFunctions = {
@@ -130,6 +130,9 @@ export const addNeuronWith = ({
   }
   if (controller) {
     neuron.fullNeuron.controller = controller;
+  }
+  if (nonNullish(getNeuron({ identity, neuronId: neuron.neuronId }))) {
+    throw new Error(`A neuron with id ${neuron.neuronId} already exists`);
   }
   getNeurons(identity).push(neuron);
 };
--- a/frontend/src/tests/fakes/governance-api.fake.ts
+++ b/frontend/src/tests/fakes/governance-api.fake.ts
@@ -1,4 +1,7 @@
-import type { ApiQueryParams } from "$lib/api/governance.api";
+import type {
+  ApiMergeNeuronsParams,
+  ApiQueryParams,
+} from "$lib/api/governance.api";
 import { mockIdentity } from "$tests/mocks/auth.store.mock";
 import { mockNeuron } from "$tests/mocks/neurons.mock";
 import { mockRewardEvent } from "$tests/mocks/nns-reward-event.mock";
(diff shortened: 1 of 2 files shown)


frontend/src/tests/lib/modals/neurons/MergeNeuronsModal.spec.ts
```diff
@@ -177,21 +184,31 @@ describe("MergeNeuronsModal", () => {
...
+      const sourceNeuron = fakeGovernanceApi.getNeuron({
+        identity: testIdentity,
+        neuronId: mergeableNeuron1.neuronId,
+      });
+      const targetNeuron = fakeGovernanceApi.getNeuron({
+        identity: testIdentity,
+        neuronId: mergeableNeuron2.neuronId,
+      });
+      await runResolvedPromises();
+      expect(getStake(sourceNeuron)).toBe(BigInt(0));
```
lmuntaner, line 196: Not in this PR, but we probably need a test (not in this component) that checks that the UI has been updated.

frontend/src/tests/fakes/governance-api.fake.ts
```diff
@@ -77,16 +112,38 @@ const reset = () => {
...
+    neuron.neuronId = neuronId;
+    neuron.fullNeuron.id = neuronId;
+  }
+  if (stake) {
+    neuron.fullNeuron.cachedNeuronStake = stake;
+  }
+  if (controller) {
+    neuron.fullNeuron.controller = controller;
+  }
+  getNeurons(identity).push(neuron);
```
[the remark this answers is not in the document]
dskloetd: Good point (although not new with this PR).
I've added a check above.

I'm thinking the check should really be across all identities.
But probably the neurons shouldn't even be completely separated by identity since the same neuron can have a hotkey of another identity. I'll have to think about that more. For now this check should help.

frontend/src/tests/lib/modals/neurons/MergeNeuronsModal.spec.ts
dskloetd: I would (perhaps naively ;)) hope such a test already exists. But regardless of whether it does, it's orthogonal to this component and this test.

frontend/src/tests/fakes/governance-api.fake.ts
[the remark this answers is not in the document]
lmuntaner: Sounds good, thanks for the check!

frontend/src/tests/lib/modals/neurons/MergeNeuronsModal.spec.ts
lmuntaner: It should, I agree. And I would expect that test to also use the fake.

dskloetd: I think we'll get there when we get there.
At least I'm not trying to migrate tests unless I have to deal with them anyway.

PR #2672, opened 2023-06-12
Fix some setup issues on a new M1 installation

anedos-dfinity:
# Motivation

I tried to build outside of Docker on my M1 and required a few interventions. This is only a partial list of the issues I discovered with @dskloetd help, but I'll make another PR once I test some more.

# Changes

- `sns` binary url for Darwin was wrong
- `realpath` on my M1 exists on `/bin` so `./setup` didn't `brew install coreutils`, so I changed the condition to `sha256sum`

# Tests

Manual tests

### DIFF
--- a/e2e-tests/scripts/nns-canister-download
+++ b/e2e-tests/scripts/nns-canister-download
@@ -83,7 +83,7 @@ get_binary() {
   OS="$(uname)"
   case "$OS" in
   Darwin)
-    curl "https://download.dfinity.systems/ic/${IC_COMMIT}/nix-release/x86_64-darwin/${FILENAME}.gz" | gunzip >"$TMP_FILE"
+    curl "https://download.dfinity.systems/ic/${IC_COMMIT}/openssl-static-binaries/x86_64-darwin/${FILENAME}.gz" | gunzip >"$TMP_FILE"
     ;;
   Linux)
     curl "https://download.dfinity.systems/ic/${IC_COMMIT}/release/${FILENAME}.gz" | gunzip >"$TMP_FILE"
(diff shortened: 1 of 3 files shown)


e2e-tests/scripts/nns-canister-download
```diff
@@ -83,7 +83,7 @@ get_binary() {
   OS="$(uname)"
   case "$OS" in
   Darwin)
-    curl "https://download.dfinity.systems/ic/${IC_COMMIT}/nix-release/x86_64-darwin/${FILENAME}.gz" | gunzip >"$TMP_FILE"
+    curl "https://download.dfinity.systems/ic/${IC_COMMIT}/openssl-static-binaries/x86_64-darwin/${FILENAME}.gz" | gunzip >"$TMP_FILE"
```
dskloetd, line 86: This looks fine to me but:
1. Have you tested this or just changed it because it looks similar?
2. These are the old e2e tests that I'm not really familiar with and hope to delete once everything they test is covered by the new e2e tests.
3. Nobody seems to have issues with this so either it works fine or it's never used.

Either way is fine with me but just wanted to point this out.

dskloetd (approved):

e2e-tests/scripts/nns-canister-download
anedos-dfinity: Thanks for the feedback @dskloetd,
1. The change was driven by failing CI tests, I have to admit that I was surprised that this particular change causes the CI tests to pass, so it might be an idea to double-check (i.e., I'll revert, push,  check CI fails and re-apply the change(
2. fine with me
3. I'll double check to make sure

anedos-dfinity: Turns out the e2e tests pass regardless of the change, I propose we update it anyway to match the code change.

dskloetd: Fine with me either way. When you're ready, you can resolve the conversation and click "Merge when ready" below.

PR #2674, opened 2023-06-12
Feat: Add Min Participants to Project Detail Page

lmuntaner:
# Motivation

Add the minimum of participants needed for a Swap to pass in the Project Detail page.

# Changes

* Add one more KeyValuePair in the ProjectSwapDetails with the value of `min_participants`.

# Tests

* Add a test case for the new value rendered.

### DIFF
--- a/frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
@@ -74,9 +74,9 @@
     <span slot="key">{$i18n.sns_project_detail.total_tokens} </span>
     <AmountDisplay slot="value" amount={snsTokens} singleLine />
   </KeyValuePair>
-  <KeyValuePair>
+  <KeyValuePair testId="project-swap-min-participants">
     <span slot="key">{$i18n.sns_project_detail.min_participants} </span>
-    <span slot="value" data-tid="project-swap-min-participants"
+    <span slot="value"
       >{formatNumber(params.min_participants, {
         minFraction: 0,
         maxFraction: 0,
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
```diff
@@ -73,6 +74,15 @@
     <span slot="key">{$i18n.sns_project_detail.total_tokens} </span>
     <AmountDisplay slot="value" amount={snsTokens} singleLine />
   </KeyValuePair>
+  <KeyValuePair>
+    <span slot="key">{$i18n.sns_project_detail.min_participants} </span>
+    <span slot="value" data-tid="project-swap-min-participants"
```
dskloetd, line 79: FYI (so up to you) but there is a page object for KeyValuePair so if you add the testId there it might be more flexible.

frontend/src/tests/lib/components/project-detail/ProjectSwapDetails.spec.ts
```diff
@@ -39,6 +39,27 @@ describe("ProjectSwapDetails", () => {
     );
   });
 
+  it("should render min participants", () => {
+    const { getByTestId } = renderContextCmp({
+      summary: {
+        ...mockSummary,
```
dskloetd, line 45: How about extending `createSummary` and using that?

lmuntaner: I don't think it's worth for this field. I don't expect us to create many summaries where we need a specific min participants.

dskloetd: Requiring 10 lines to set a single field is really unfortunate, especially since we already have `createSummary` to mitigate that.
If you don't want to add this field, maybe add an argument for swap params so we can at least reduce it somewhat and keep it somewhat generic?

But honestly, I think just making this 1 case more readable is worth it. And I don't think `createSummary` would suffer from additional args.

lmuntaner: Sure, I actually like the idea that the logic to set fields is in one place.

Not necessarily about requiring 10 lines of code.

dskloetd (approved):
Thanks!

PR #2676, opened 2023-06-12
ckBTC get minter info api, services and stores

peterpeterparker:
# Motivation

No function and feature but, a PR that introduces `getMinterInfo` API, services and stores.

# PRs

- [x] https://github.com/dfinity/ic-js/pull/346

# Changes

- bump ic-js
- implement related API, services and stores


frontend/src/lib/stores/ckbtc-info.store.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * - setInfo: set the information fetched from ckBTC minter
+ * - reset: reset all information
+ * - resetUniverse: reset the information for a particular universe
+ *
+ */
+const initCkBTCInfoStore = (): CkBTCInfoStore => {
+  const initialCkBTCInfoStoreData: CkBTCInfoStoreData = {};
+
+  const { subscribe, update, set } = writable<CkBTCInfoStoreData>(
```
lmuntaner, line 41: Maybe not in this PR, but with @dskloetd we were thinking on letting each store feel itself when a client subscribes to it using the start function (second parameter).

https://svelte.dev/docs#run-time-svelte-store-writable

What do you think of the pattern in general?

lmuntaner (approved):
LGTM!

Just a comment about a new pattern that could be interesting to use.

frontend/src/lib/stores/ckbtc-info.store.ts
peterpeterparker: It's interesting. Don't know in general in which feature would you use it?

> If a function is passed as the second argument, it will be called when the number of subscribers goes from zero to one (but not from one to two, etc).

In this particular feature, the store is empty and once the user sign-in and visit a ckBTC page, then the information will be loaded from the canister and saved in store. 

So, not sure I would add the "Loading" within this "store start" function, that could introduce a circular reference between store and service. Also not sure how to handle the fact that it should be done when and only when the user has signed-in only.

peterpeterparker: (resolving conversation to merge PR but looking forward to your feedback)

lmuntaner: The idea is to avoid using services to populate stores initially.

When the user visits the page, the component subscribes to the store, which in turn triggers the second function.

This second function recieves the "set" function as argument. Therefore, it could make a request (or two) and populate itself.

The advantage is mainly: "We don't need to remember to call the service when data is needed."

Does that make sense?

peterpeterparker: Yes makes sense, then I understand it correctly. Sounds interesting, as I said above not sure about the circular dependencies and how to handle signed-in/not signed-in but, beside that, sounds interesting. Looking forward to a proof of concept / piece of code to showcase the idea.

PR #2681, opened 2023-06-13
Use ckBTC info for retrieve min amount

peterpeterparker:
# Motivation

Use new ckBTC info fetched from minter for the minimum amount that can be converted check instead of a constant.

# Note

I am not super super happy about displaying an error when the minimal amount information have not been certified yet but, I've tried several implementation and UX and it ultimately felt like the easiest and cleanest way to handle what is, at the end of the day, an edge case.

# PRs

- [x] #2682

# Changes

- remove constant `RETRIEVE_BTC_MIN_AMOUNT` and replace with new `MinterInfo` data
- enhance `assertCkBTCUserInputAmount` to use the data from the store and checks if these are provided and certified


frontend/src/lib/utils/ckbtc.utils.ts
```diff
@@ -58,4 +74,17 @@ export const assertCkBTCUserInputAmount = ({
     account: sourceAccount,
     amountE8s: amountE8s + transactionFee,
   });
+
+  // This assertion is primarily intended to handle edge cases.
+  // It serves to prevent situations where the user has entered an amount before the ckBTCInfoStore has been filled with certified data.
+  // However, considering our current user experience, the likelihood of this scenario is low.
+  if (infoCertified !== true && FORCE_CALL_STRATEGY !== "query") {
```
lmuntaner, line 81: Why not `!infoCertified`?

lmuntaner (approved):
LGTM, only a style comment. But already approving.

frontend/src/lib/utils/ckbtc.utils.ts
peterpeterparker: `infoCertified: boolean | undefined`. In the code base when something can be `boolean | undefined` whe make the comparison explicit.

PR #2682, opened 2023-06-13
Load ckBTC info

peterpeterparker:
# Motivation

Load ckBTC info when ckBTC accounts or wallet page are accessed.

# Changes

- add a new component `CkBTCInfoLoader` that loads the info for ckBTC universe
- use new cmp in ckBTC accounts and wallet page
- extend service with `minterCanisterId` parameter


frontend/src/lib/components/accounts/CkBTCInfoLoader.svelte
```diff
@@ -0,0 +1,44 @@
...
+      return;
+    }
+
+    await loadCkBTCInfo({
+      universeId: selectedCkBTCUniverseId,
+      minterCanisterId: canisters.minterCanisterId,
+    });
+  };
+
+  $: (async () =>
```
lmuntaner, line 37: This will be called when both `canisters` or `selectedCkBTCUniverseIdStore` change, right?

frontend/src/lib/components/accounts/CkBTCInfoLoader.svelte
lmuntaner, line 1: Didn't we agree in using services instead of components that render nothing?

Why can't you use a service here?

frontend/src/lib/services/ckbtc-info.services.ts
```diff
@@ -3,17 +3,19 @@ import { FORCE_CALL_STRATEGY } from "$lib/constants/mockable.constants";
...
 
 export const loadCkBTCInfo = async ({
   handleError,
   universeId,
+  minterCanisterId,
 }: {
   handleError?: () => void;
   universeId: UniverseCanisterId;
-}) => {
+} & Pick<CkBTCAdditionalCanisters, "minterCanisterId">) => {
```
lmuntaner, line 18: Why do you prefer to `Pick` one field and a union instead of adding it to the expected params?

frontend/src/tests/lib/pages/CkBTCAccounts.spec.ts
```diff
@@ -39,6 +39,12 @@ jest.mock("$lib/services/ckbtc-minter.services", () => {
   };
 });
 
+jest.mock("$lib/services/ckbtc-info.services", () => {
```
lmuntaner, line 42: We try to mock the api directly, not the service. Is it possible to do it here?

frontend/src/tests/lib/pages/CkBTCWallet.spec.ts
```diff
@@ -75,6 +75,12 @@ jest.mock("$lib/services/ckbtc-minter.services", () => {
   };
 });
 
+jest.mock("$lib/services/ckbtc-info.services", () => {
```
lmuntaner, line 78: ditto on mocking API

frontend/src/tests/lib/routes/Accounts.spec.ts
```diff
@@ -68,6 +68,12 @@ jest.mock("$lib/services/ckbtc-minter.services", () => {
   };
 });
 
+jest.mock("$lib/services/ckbtc-info.services", () => {
```
lmuntaner, line 71: ditto

lmuntaner (commented):
I added a few comments.

frontend/src/tests/lib/pages/CkBTCAccounts.spec.ts
peterpeterparker: Because there are already other services mocked in the same way in this test, it would be consistent to reuse the same approach. Additionally, I am not inclined to change the entire test.

frontend/src/tests/lib/routes/Wallet.spec.ts
```diff
@@ -35,6 +35,12 @@ jest.mock("$lib/services/ckbtc-transactions.services", () => {
   };
 });
 
+jest.mock("$lib/services/ckbtc-info.services", () => {
```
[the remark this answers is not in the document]
peterpeterparker: ditto

frontend/src/lib/services/ckbtc-info.services.ts
peterpeterparker: Because `CkBTCAdditionalCanisters` is the accurate type and we do not need the `indexCanisterId` here.

frontend/src/tests/lib/routes/Accounts.spec.ts
peterpeterparker: ditto

frontend/src/lib/components/accounts/CkBTCInfoLoader.svelte
lmuntaner: Thanks!

This is a very interesting pattern. As you mention allows for reusing some part of the code easily. But for that, we could also come up with a generic component that renders a spinner when undefined, or the slot otherwise.

That component could expect a service to call and a store value.

Yet, I'd rather use the "start" function of the store.

frontend/src/tests/lib/pages/CkBTCAccounts.spec.ts
lmuntaner: Ok, makes sense. Maybe for new test files.

lmuntaner (approved):

PR #2683, opened 2023-06-13
Fix: Use ballot voting power in SNS proposals

lmuntaner:
# Motivation

There was a bug where the voting power shows in the SNS proposals matched the voting power of the neuron, instead of the voting power of the ballot.

Voting power of a neuron for a proposal is set when the proposal is created. If the voting power of the neuron changes afterwards, it doesn't affect the voting power for earlier proposals.

# Changes

* New sns proposal util `ballotVotingPower` to get the voting power of a neuron for a specific proposal.
* Change the voting power calculation in the sns vote registration service `proposalAfterVote`.
* Change the voting power calculation in the sns neuron util `votedSnsNeuronDetails`.
* Change the voting power calculation in the sns proposal util `snsNeuronToVotingNeuron`.
* Remove the unused property `parameters` in the changed functions and in the dependencies where they are called.

# Tests

* Add two new cases in SnsVotingCard spec file to check that voting power comes from ballots.
* Edit the tests for the utils after change in calculation.
* New test for `ballotVotingPower` util.

### DIFF
--- a/frontend/src/lib/components/proposal-detail/MyVotes.svelte
+++ b/frontend/src/lib/components/proposal-detail/MyVotes.svelte
@@ -52,7 +52,7 @@
               SNS_NEURON_ID_DISPLAY_LENGTH
             )}
           </p>
-          <p class="vote-details">
+          <p class="vote-details" data-tid="my-votes-voting-power">
             <Value>{formatVotingPower(neuron.votingPower)}</Value>
             {#if voteIconMapper[neuron.vote]}
               <svelte:component this={voteIconMapper[neuron.vote]} />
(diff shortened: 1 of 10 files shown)


dskloetd:
I'm not sure how much sense it makes for me to dive into this code now given that @mstrasinskis is back. Do you want to walk me through it or shall I just let Maksims review?

lmuntaner:
> I'm not sure how much sense it makes for me to dive into this code now given that @mstrasinskis is back. Do you want to walk me through it or shall I just let Maksims review?

Yeah, I also wasn't sure to include you. But since you found the issue, I thought you might be interested in the fix.

Let's wait for @mstrasinskis

mstrasinskis (approved):
LGTM, thank you!

PR #2684, opened 2023-06-13
GIX-1584: Build env var being used in worker

lmuntaner:
# Motivation

As described in https://dfinity.atlassian.net/browse/GIX-1584 a build environment variables was being used instead of the one from the initialization.

The problem was that the workers can't access `window` and therefore do not have access to the initialization env vars.

Solution for now: Pass the parameter in the post message instead of accessing the env vars directly.

Food for thought: Should we substitute the variables in the JS files instead of setting them in HTML during canister installation and upgrade?

# Changes

* `queryCanisterDetails` expects two new optional params: `host` and `fetchRootKey`.
* PostMessage to the cycles worker includes `host` and `fetchRootKey`.
* Cycles service `initCyclesWorker` passes `host` and `fetchRootKey` in the post message to the worker.
* Cycles worker uses new params `host` and `fetchRootKey` when calling `queryCanisterDetails`.

# Tests

* New test case in `queryCanisterDetails` to check that new param is being used.

### DIFF
--- a/frontend/src/lib/api/canisters.api.cjs.ts
+++ b/frontend/src/lib/api/canisters.api.cjs.ts
@@ -2,7 +2,6 @@ import { toCanisterDetails } from "$lib/canisters/ic-management/converters";
 import type { CanisterDetails } from "$lib/canisters/ic-management/ic-management.canister.types";
 import { mapError } from "$lib/canisters/ic-management/ic-management.errors";
 import type { CanisterStatusResponse } from "$lib/canisters/ic-management/ic-management.types";
-import { FETCH_ROOT_KEY, HOST } from "$lib/constants/environment.constants";
 import { HttpAgentCjs, getManagementCanisterCjs } from "$lib/utils/cjs.utils";
 import { logWithTimestamp } from "$lib/utils/dev.utils";
 import type { Identity, ManagementCanisterRecord } from "@dfinity/agent";
@@ -11,13 +10,13 @@ import { Principal } from "@dfinity/principal";
 export const queryCanisterDetails = async ({
   identity,
   canisterId,
-  host = HOST,
-  fetchRootKey = FETCH_ROOT_KEY,
+  host,
+  fetchRootKey,
 }: {
   identity: Identity;
   canisterId: string;
-  host?: string;
-  fetchRootKey?: boolean;
+  host: string;
+  fetchRootKey: boolean;
 }): Promise<CanisterDetails> => {
   logWithTimestamp(`Getting canister ${canisterId} details call...`);
   const {
--- a/frontend/src/lib/api/canisters.api.cjs.ts
+++ b/frontend/src/lib/api/canisters.api.cjs.ts
@@ -11,14 +11,18 @@ import { Principal } from "@dfinity/principal";
 export const queryCanisterDetails = async ({
   identity,
   canisterId,
+  host = HOST,
+  fetchRootKey = FETCH_ROOT_KEY,
 }: {
   identity: Identity;
   canisterId: string;
+  host?: string;
+  fetchRootKey?: boolean;
 }): Promise<CanisterDetails> => {
   logWithTimestamp(`Getting canister ${canisterId} details call...`);
   const {
     icMgtService: { canister_status },
-  } = await canisters(identity);
+  } = await canisters({ identity, host, fetchRootKey });
 
   try {
     const canister_id = Principal.fromText(canisterId);
@@ -40,17 +44,23 @@ export const queryCanisterDetails = async ({
   }
 };
 
-const canisters = async (
-  identity: Identity
-): Promise<{
+const canisters = async ({
+  identity,
+  host,
+  fetchRootKey,
+}: {
+  identity: Identity;
+  host: string;
+  fetchRootKey: boolean;
+}): Promise<{
   icMgtService: ManagementCanisterRecord;
 }> => {
   const agent = new HttpAgentCjs({
     identity,
-    host: HOST,
+    host,
   });
 
-  if (FETCH_ROOT_KEY) {
+  if (fetchRootKey) {
     await agent.fetchRootKey();
   }
 
--- a/frontend/src/lib/services/worker-cycles.services.ts
+++ b/frontend/src/lib/services/worker-cycles.services.ts
@@ -1,17 +1,14 @@
-import type {
-  PostMessageDataRequestCycles,
-  PostMessageDataResponseCycles,
-} from "$lib/types/post-message.canister";
+import { FETCH_ROOT_KEY, HOST } from "$lib/constants/environment.constants";
+import type { PostMessageDataResponseCycles } from "$lib/types/post-message.canister";
 import type { PostMessage } from "$lib/types/post-messages";
 
 export type CyclesCallback = (data: PostMessageDataResponseCycles) => void;
 
 export interface CyclesWorker {
-  startCyclesTimer: (
-    params: {
-      callback: CyclesCallback;
-    } & PostMessageDataRequestCycles
-  ) => void;
+  startCyclesTimer: (params: {
+    callback: CyclesCallback;
+    canisterId: string;
+  }) => void;
   stopCyclesTimer: () => void;
 }
 
@@ -36,15 +33,16 @@ export const initCyclesWorker = async (): Promise<CyclesWorker> => {
   return {
     startCyclesTimer: ({
       callback,
-      ...rest
+      canisterId,
     }: {
       callback: CyclesCallback;
-    } & PostMessageDataRequestCycles) => {
+      canisterId: string;
+    }) => {
       cyclesCallback = callback;
 
       cyclesWorker.postMessage({
         msg: "nnsStartCyclesTimer",
-        data: { ...rest },
+        data: { canisterId, host: HOST, fetchRootKey: FETCH_ROOT_KEY },
       });
     },
     stopCyclesTimer: () => {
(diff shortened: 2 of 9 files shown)


lmuntaner:

frontend/src/lib/api/canisters.api.cjs.ts
```diff
@@ -11,14 +11,18 @@ import { Principal } from "@dfinity/principal";
 export const queryCanisterDetails = async ({
   identity,
   canisterId,
+  host = HOST,
```
dskloetd, line 14: Is this defaulting to the constants that should not be used?
Can we do without default value here?

frontend/src/tests/lib/api/canisters.api.cjs.spec.ts
```diff
@@ -20,16 +20,16 @@ jest.mock("@dfinity/agent/lib/cjs/index", () => {
...
-    }),
-  };
-});
+  const MockHttpAgent = jest.fn();
 
-describe("canisters-api.cjs", () => {
-  afterAll(() => jest.resetAllMocks());
+  beforeEach(async () => {
+    jest.resetAllMocks();
+    const module = await import("@dfinity/agent/lib/cjs/index");
```
dskloetd, line 27: Do you need to import this dynamically or is it the same to do
```
import * as module from "@dfinity/agent/lib/cjs/index";
```
at the top?
(I would use a different name in that case.)

frontend/src/lib/services/worker-cycles.services.ts
dskloetd, line 1: Can you add a test where you set both the environment and the html build vars and verify which ones are used?

frontend/src/lib/services/worker-cycles.services.ts
```diff
@@ -1,17 +1,14 @@
...
 export interface CyclesWorker {
-  startCyclesTimer: (
-    params: {
-      callback: CyclesCallback;
-    } & PostMessageDataRequestCycles
-  ) => void;
+  startCyclesTimer: (params: {
+    callback: CyclesCallback;
+    canisterId: string;
+  }) => void;
```
peterpeterparker, line 11: why not using the types anymore?

frontend/src/lib/types/post-messages.ts
```diff
@@ -14,6 +14,11 @@ export type PostMessageResponse =
 
 export type PostMessageData = object;
 
+export interface PostMessageFetchBase extends PostMessageData {
```
peterpeterparker, line 17: what does `FetchBase` mean? 

given the goal of the type, a term with `Actor` in it would feel more IC standard I have to say

frontend/src/lib/workers/cycles.worker.ts
```diff
@@ -33,12 +33,14 @@ onmessage = async ({
...
 const syncCanister = async ({
   identity,
-  data: { canisterId },
+  data: { canisterId, host, fetchRootKey },
 }: TimerWorkerUtilJobData<PostMessageDataRequestCycles>) => {
   try {
     const canisterInfo: CanisterDetails = await queryCanisterDetails({
       canisterId,
       identity,
+      host,
```
peterpeterparker, line 42: why not spreading `...data`?

frontend/src/lib/services/worker-cycles.services.ts
```diff
@@ -36,15 +33,16 @@ export const initCyclesWorker = async (): Promise<CyclesWorker> => {
...
       callback: CyclesCallback;
-    } & PostMessageDataRequestCycles) => {
+      canisterId: string;
+    }) => {
       cyclesCallback = callback;
 
       cyclesWorker.postMessage({
         msg: "nnsStartCyclesTimer",
-        data: { ...rest },
+        data: { canisterId, host: HOST, fetchRootKey: FETCH_ROOT_KEY },
```
peterpeterparker, line 45: why removing the spread of `...rest` here?

frontend/src/lib/api/canisters.api.cjs.ts
```diff
@@ -40,17 +44,23 @@ export const queryCanisterDetails = async ({
...
 };
 
-const canisters = async (
-  identity: Identity
-): Promise<{
+const canisters = async ({
+  identity,
+  host,
+  fetchRootKey,
+}: {
```
peterpeterparker, line 51: why not using the types `PostMessageFetchBase` here?

frontend/src/lib/api/canisters.api.cjs.ts
lmuntaner: This default works everywhere except in the worker.

Removing the default doesn't help either. The key is that the worker receives the host in the post message and uses that value. Not only that the value is passed here.

frontend/src/lib/services/worker-cycles.services.ts
lmuntaner: I can try that in another PR.

I would need to try to mock the worker environment. It might be easier to do this in an e2e test.

frontend/src/lib/workers/cycles.worker.ts
lmuntaner: yes, good point

frontend/src/tests/lib/api/canisters.api.cjs.spec.ts
lmuntaner: If I import statically, they are readonly properties and TS complains.

frontend/src/lib/services/worker-cycles.services.ts
peterpeterparker: There are various worker, some have no params, some have multiples. Spreading the values allow us to always use the same pattern, the same piece of code. Ultimately it will probably even be possible to refactor the code to avoid duplicates.

So since it's a preference and not related to this PR, I would rather like this to not change if you don't mind too much.

Unless of course there are strong arguments?

frontend/src/lib/api/canisters.api.cjs.ts
peterpeterparker: That's a good point.

There are other cjs api that also use the `PostMessage...` types already, so maybe we should stick to this pattern in this PR and in another PR, transform all APIs for consistency?

Alternatively we can also rename the `.cjs.ts` API to something like `api-worker.ts` or something as these API are only used - and should only be used by - workers?

frontend/src/lib/services/worker-cycles.services.ts
lmuntaner: Sure, makes sense. Not a strong opinion on my side.

frontend/src/lib/api/canisters.api.cjs.ts
lmuntaner: But a worker might use a normal `api`, no?

Why did you call it `cjs.ts`?

I need to do the similar for the metrics worker. I will propose a generic pattern then.

peterpeterparker: > But a worker might use a normal api, no?

Only if it does not use Agent-js.

We do not use ESM in web worker because Firefox support for such types of bundles is super recent. Even if it would, I'm not sure agent-js would work out of the box. That's why currently the only cross-browser way to use web workers with agent-js is through the usage of an explicity import to its `cjs` bundle.

From `cjs.utils.ts`:

```
/**
 * HTTP-Agent explicit CJS import for compatibility with web worker - avoid Error [RollupError]: Unexpected token (Note that you need plugins to import files that are not JavaScript)
 */
import { HttpAgent, getManagementCanister } from "@dfinity/agent/lib/cjs/index";
```

That's also why all these API and utils are prefixed with the keyword "cjs".

lmuntaner: I missed that part, thanks!

In that case, I agree to rename those apis as `api-worker` to specify they are used in the workers.

I will even make the `host` and `fetchRootKey` mandatory fields. To make sure the env vars are not imported here.

peterpeterparker: > In that case, I agree to rename those apis as api-worker to specify they are used in the workers.

Let's do that. In another PR if that work for you? Would you take care of it or should I?

> I will even make the host and fetchRootKey mandatory fields

Good idea.

lmuntaner: I wouldn't mind if you have time to do it. I'll focus on the SNS voting.

peterpeterparker: Sure, could do. As soon as I am finish with the small ckBTC task I go back to the web workers related topic anyway (#2639). Will do.

frontend/src/lib/api/canisters.api.cjs.ts
lmuntaner: done

frontend/src/lib/services/worker-cycles.services.ts
dskloetd: Unless I'm missing something, the changes in this file are not tested in this PR?
I don't think tests show go in a separate PR.
If you think the test I proposed is too complicated, please propose some way that the changes in this PR can be tested in this PR.

lmuntaner: The changes are also statically tested. The worker expects the "host" and "fetchRootKey" in the message. It doesn't import them.

But it makes sense to add some more testing.

I was able to mock the import of the worker and add a test for the cycles worker service. This way we can test that the host and fetchrootkey are sent from the service. Where `window` is available.

For more details check the last commit: https://github.com/dfinity/nns-dapp/pull/2684/commits/1d8aa400eaca08f404f817e0a13765e1d92c1c61

Let me know what you think

dskloetd (approved):

PR #2686, opened 2023-06-13
Test/po get classes

mstrasinskis:
# Motivation

Add `getClasses` function to the page object to avoid searching for a string inclusion in `class` attribute, which is not reliable. 

# Changes

- `getClasses` function
- apply it to [ProjectCard.page-object.ts](https://github.com/dfinity/nns-dapp/compare/test/po-getClasses?expand=1#diff-11e533ca8823e8103f3ee38c0eb7e2642bf814f43d5e78ee42c32a1f899c7cdd)

# Tests

- `ProjectCard.spec.ts` (calls `isHighlighted`) is still green.


frontend/src/tests/page-objects/ProjectCard.page-object.ts
```diff
@@ -19,7 +19,7 @@ export class ProjectCardPo extends BasePageObject {
   }
 
   async isHighlighted(): Promise<boolean> {
```
dskloetd, line 21: This function is no longer here since https://github.com/dfinity/nns-dapp/commit/9ffe43e1fbdb75cd0e25ecde3d7c767a65747881
Can you pull?

mstrasinskis: Thanks, updated.

dskloetd (approved):

PR #2687, opened 2023-06-13
Use ckBTC info to display the number of confirmations

peterpeterparker:
# Motivation

Use new ckBTC info fetched from minter to display the number of confirmations.

# Note

I don't like much falling back to `""` when the info parameters are not yet loaded but, it felt like the less ugly UI wise solution. Adding skeleton and spinner look meh. If you got better idea, please let me know!

# PRs

- [x] #2682

# Changes

- use `ckBTCInfoStore.min_confirmations` to display `12 confirmations` on ckBTC wallet page


frontend/src/lib/components/accounts/BitcoinAddress.svelte
```diff
@@ -45,11 +47,17 @@
...
+  $: minConfirmations =
+    $ckBTCInfoStore[universeId.toText()]?.info.min_confirmations;
+
   // TODO: incoming_bitcoin_network_part_1 comes before incoming_bitcoin_network_part_2. It would be more worth creating a component or directive that can replace placeholders of the i18n with Svelte components
 </script>
 
 <p class="description">
-  {$i18n.ckbtc.incoming_bitcoin_network_part_1}
+  {replacePlaceholders($i18n.ckbtc.incoming_bitcoin_network_part_1, {
+    $min: `${minConfirmations ?? ""}`,
```
lmuntaner, line 59: Do the confirmations change very frequently? Can we add the latest number as a fallback instead of `""`?

frontend/src/tests/lib/components/accounts/BitcoinAddress.spec.ts
```diff
@@ -112,12 +122,19 @@ describe("BitcoinAddress", () => {
...
+    it("should display a sentence info", async () => {
       const { getByText } = render(BitcoinAddress, { props });
 
-      expect(
-        getByText(en.ckbtc.incoming_bitcoin_network_part_1, { exact: false })
-      ).toBeInTheDocument();
+      await waitFor(() =>
+        expect(
+          getByText(
+            replacePlaceholders(en.ckbtc.incoming_bitcoin_network_part_1, {
```
lmuntaner, line 131: What do you think of hard coding the text here?

We started doing that in some cases. And to be honest, I find it more readable.

lmuntaner (approved):
A couple of comments. Already approving because they are no biggies.

frontend/src/lib/components/accounts/BitcoinAddress.svelte
peterpeterparker: It does not change frequently. It used to be "12" (hardcoded) and did not change since launch, therefore the least effort for a a-ok ui. Good idea though, I keep it in mind if this would becomes to change frequently (but, I doubt it will be the case).

frontend/src/tests/lib/components/accounts/BitcoinAddress.spec.ts
peterpeterparker: No particular opinion, sure care do. Done.

PR #2689, opened 2023-06-13
Use ckBTC info to display kyt fee

peterpeterparker:
# Motivation

Use new ckBTC info fetched from minter to display the kyt fee.

# PRs

- [ ] https://github.com/dfinity/ic-js/pull/348
- [ ] #2682

# Changes

- deprecate `getDepositFee`
- bump ic-js that deprecates `getDepositFee` as well
- remove fetch to above API call in component and replace with usage of the store


lmuntaner (approved):
I liked this one! Removing more lines than adding them!

PR #2690, opened 2023-06-14
Fix: Users are not able to vote in sns proposal

lmuntaner:
# Motivation

User was not able to vote on SNS proposal if was logging in directly from the SNS proposal detail page.

Problem: The neurons were not fetched.

# Changes

* The `update` call in the SnsProposalDetail page depends on the `authStore`.

# Tests

Main test:
* New test case in SnsProposalDetail to simulate the bug.

Additional changes:
* SnsVotingCard page object.
* New method `setForTesting` in `authStore`. This allows to fake logging in in unit tests.
* New methods in SnsProposalDetail page object.
* New methods in SnsProposalVotingSection page object.
* New SnsVotingCard page object.
* New ConfirmationToolBar page object.

### DIFF
--- a/frontend/src/lib/pages/SnsProposalDetail.svelte
+++ b/frontend/src/lib/pages/SnsProposalDetail.svelte
@@ -32,6 +32,7 @@
   import TestIdWrapper from "$lib/components/common/TestIdWrapper.svelte";
   import { layoutTitleStore } from "$lib/stores/layout.store";
   import { i18n } from "$lib/stores/i18n";
+  import { authStore } from "$lib/stores/auth.store";
 
   export let proposalIdText: string | undefined | null = undefined;
 
@@ -165,15 +166,15 @@
     }`
   );
 
-  // The `update` function cares about the necessary data to be refetched.
-  $: universeIdText, proposalIdText, $snsNeuronsStore, update();
-
   let proposalIds: bigint[];
   $: proposalIds = nonNullish(universeIdText)
     ? sortSnsProposalsById(
         $snsFilteredProposalsStore[universeIdText]?.proposals
       )?.map(snsProposalId) ?? []
     : [];
+
+  // The `update` function cares about the necessary data to be refetched.
+  $: universeIdText, proposalIdText, $snsNeuronsStore, $authStore, update();
 </script>
 
 <TestIdWrapper testId="sns-proposal-details-grid">
--- a/frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
@@ -289,11 +289,13 @@ describe("SnsProposalDetail", () => {
         proposalId: proposalId.id,
       });
 
+      const proposalCreatedTimestamp = 33333n;
+
       fakeSnsGovernanceApi.addProposalWith({
         identity: new AnonymousIdentity(),
         rootCanisterId,
         ...proposal,
-        proposal_creation_timestamp_seconds: 33333n,
+        proposal_creation_timestamp_seconds: proposalCreatedTimestamp,
         ballots: [
           [
             getSnsNeuronIdAsHexString(mockSnsNeuron),
@@ -310,7 +312,7 @@ describe("SnsProposalDetail", () => {
         identity: mockIdentity,
         rootCanisterId,
         id: mockSnsNeuron.id,
-        created_timestamp_seconds: 111n,
+        created_timestamp_seconds: proposalCreatedTimestamp - 100n,
         permissions: [
           {
             principal: [mockIdentity.getPrincipal()],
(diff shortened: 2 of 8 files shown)


lmuntaner:

frontend/src/lib/pages/SnsProposalDetail.svelte
```diff
@@ -149,7 +150,7 @@
   };
 
   // The `update` function cares about the necessary data to be refetched.
-  $: universeIdText, proposalIdText, $snsNeuronsStore, update();
+  $: universeIdText, proposalIdText, $snsNeuronsStore, $authStore, update();
```
dskloetd, line 153: Not for this PR but I think this is a bad pattern. Listing some variables and then calling a function without parameters. I think we should only use pure functions with these `$` blocks so Svelte can make sure all the dependencies are listened on.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -190,4 +197,73 @@ describe("SnsProposalDetail", () => {
...
+        status: SnsProposalDecisionStatus.PROPOSAL_DECISION_STATUS_OPEN,
+        rewardStatus:
+          SnsProposalRewardStatus.PROPOSAL_REWARD_STATUS_ACCEPT_VOTES,
+        proposalId: proposalId.id,
+      });
+
+      fakeSnsGovernanceApi.addProposalWith({
+        identity: new AnonymousIdentity(),
+        rootCanisterId,
+        ...proposal,
```
dskloetd, line 221: The idea of `addProposalWith` is that you can pass the fields that you care about and the rest should be taken care of. You shouldn't have to pass a full proposal. Can you pass the fields that you passed to `createSnsProposal`, here instead? Or if that doesn't work, can we make it work?

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -190,4 +197,73 @@ describe("SnsProposalDetail", () => {
...
+            },
+          ],
+        ],
+      });
+
+      fakeSnsGovernanceApi.addNeuronWith({
+        identity: mockIdentity,
+        rootCanisterId,
+        id: mockSnsNeuron.id,
+        created_timestamp_seconds: 111n,
```
dskloetd, line 239: Are all these values relevant for the test or can we let some be just whatever default `addNeuronWith` provides.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -190,4 +197,73 @@ describe("SnsProposalDetail", () => {
...
+      );
+
+      await runResolvedPromises();
+
+      expect(await po.hasVotingToolbar()).toBe(false);
+
+      authStore.setForTesting(mockIdentity);
+      await runResolvedPromises();
+
+      await waitFor(async () => expect(await po.hasVotingToolbar()).toBe(true));
```
dskloetd, line 266: Do you need `waitFor` even after `runResolvedPromises`?

frontend/src/lib/pages/SnsProposalDetail.svelte
lmuntaner: I agree. I'm always doubting which pattern to use in these cases.

It would be good to agree on one and stick to it.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
lmuntaner: The status and reward statues are not fields.

The `createSnsProposal` has logic to set the fields based on those statuses.

We could add that logic and status fields to `addProposalWith` as well.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
lmuntaner: They are relevant. That was one of the reasons it wasn't working. The created needs to be before the created of the proposal.

I can make it clearer using a variable.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
lmuntaner: yes, it works. Thanks!

lmuntaner:

dskloetd (approved):

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
lmuntaner: I agree.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -190,4 +197,76 @@ describe("SnsProposalDetail", () => {
...
+            },
+          ],
+        ],
+      });
+
+      fakeSnsGovernanceApi.addNeuronWith({
+        identity: mockIdentity,
+        rootCanisterId,
+        id: mockSnsNeuron.id,
+        created_timestamp_seconds: proposalCreatedTimestamp - 100n,
```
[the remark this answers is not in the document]
lmuntaner: sure, no problem

PR #2691, opened 2023-06-14
Feat: Change functionality empty NNS Proposal filters

lmuntaner:
# Motivation

Bring consistency in proposal filters.

Inconsistency: SNS filters showed all proposals if no filter was selected. Whereas NNS had a hack coming from flutter to not render any proposal if a filter was empty.

Solution: Along with product, we agreed to remove the legacy hack and consider no filter selected as "any". This is also the behavior of the backend endpoint.

# Changes

* Remove hack in `findProposals` proposal service that would set an empty store if a filter was empty.
* Change the behavior of the queryProposals api when the topics filter is empty.
* Change behavior of the `matchFilters` helper inside the proposal utils.

# Tests

* Add a test case for the `queryProposals` api new logic.
* Change the test case for the findProposals service.
* Change the test case for the `hide` proposal util.

### DIFF
--- a/frontend/src/lib/api/proposals.api.ts
+++ b/frontend/src/lib/api/proposals.api.ts
@@ -41,6 +41,7 @@ export const queryProposals = async ({
       limit: DEFAULT_LIST_PAGINATION_LIMIT,
       beforeProposal,
       excludeTopic:
+        // We want all topics if the filter is empty
         topics.length === 0
           ? []
           : enumsExclude<Topic>({
--- a/frontend/src/lib/api/proposals.api.ts
+++ b/frontend/src/lib/api/proposals.api.ts
@@ -40,10 +40,13 @@ export const queryProposals = async ({
     request: {
       limit: DEFAULT_LIST_PAGINATION_LIMIT,
       beforeProposal,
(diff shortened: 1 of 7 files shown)


lmuntaner:
@mstrasinskis please review

frontend/src/lib/api/proposals.api.ts
```diff
@@ -40,10 +40,13 @@ export const queryProposals = async ({
     request: {
       limit: DEFAULT_LIST_PAGINATION_LIMIT,
       beforeProposal,
-      excludeTopic: enumsExclude<Topic>({
-        obj: Topic as unknown as Topic,
-        values: topics,
-      }),
+      excludeTopic:
+        topics.length === 0
+          ? []
```
mstrasinskis, line 45: nitpick: a comment is welcome here. e.g. Request with no excluded topics if topics filter is empty

mstrasinskis (approved):
Thanks, just a minor comment.

frontend/src/lib/api/proposals.api.ts
lmuntaner: sounds good

PR #2692, opened 2023-06-14
Improve canister worker typing, add params to utils and extend TVL with params

peterpeterparker:
# Motivation

Follow-up of #2684 to pass `host` and `fetchRootKey`. 

This PR introduces a new interface to generalize the actor params and add the new options to the utils that creates a canister.

Because it extends that particular utils, per extension the TVL workers also had to be extended to support the new params.

These changes are required to use the new options with the new workers I am currently developing in #2639.

# Changes

- extend `createCanisterCjs` with `host` and `fetchRootKey`
- new interface `CanisterActorParams`
- new constant `ACTOR_PARAMS`
- new utils `mapCanisterId`
- extend TVL workers to support `host` and `fetchRootKey`

### DIFF
--- a/frontend/src/lib/api/canisters.api.cjs.ts
+++ b/frontend/src/lib/api/canisters.api.cjs.ts
@@ -13,7 +13,7 @@ export const queryCanisterDetails = async ({
   canisterId,
   host,
   fetchRootKey,
-}: CanisterActorParams): Promise<CanisterDetails> => {
+}: { canisterId: string } & CanisterActorParams): Promise<CanisterDetails> => {
   logWithTimestamp(`Getting canister ${canisterId} details call...`);
   const {
     icMgtService: { canister_status },
@@ -43,7 +43,7 @@ const canisters = async ({
   identity,
   host,
   fetchRootKey,
-}: Omit<CanisterActorParams, "canisterId">): Promise<{
+}: CanisterActorParams): Promise<{
   icMgtService: ManagementCanisterRecord;
 }> => {
   const agent = new HttpAgentCjs({
--- a/frontend/src/lib/api/canisters.api.cjs.ts
+++ b/frontend/src/lib/api/canisters.api.cjs.ts
@@ -2,29 +2,25 @@ import { toCanisterDetails } from "$lib/canisters/ic-management/converters";
 import type { CanisterDetails } from "$lib/canisters/ic-management/ic-management.canister.types";
 import { mapError } from "$lib/canisters/ic-management/ic-management.errors";
 import type { CanisterStatusResponse } from "$lib/canisters/ic-management/ic-management.types";
+import type { CanisterActorParams } from "$lib/types/canister";
+import { mapCanisterId } from "$lib/utils/canisters.utils";
 import { HttpAgentCjs, getManagementCanisterCjs } from "$lib/utils/cjs.utils";
 import { logWithTimestamp } from "$lib/utils/dev.utils";
-import type { Identity, ManagementCanisterRecord } from "@dfinity/agent";
-import { Principal } from "@dfinity/principal";
+import type { ManagementCanisterRecord } from "@dfinity/agent";
 
 export const queryCanisterDetails = async ({
   identity,
   canisterId,
   host,
   fetchRootKey,
-}: {
-  identity: Identity;
-  canisterId: string;
-  host: string;
-  fetchRootKey: boolean;
-}): Promise<CanisterDetails> => {
+}: CanisterActorParams): Promise<CanisterDetails> => {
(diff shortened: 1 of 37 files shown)


frontend/src/lib/types/canister.ts
```diff
@@ -13,3 +14,10 @@ export interface CanisterSync {
 }
 
 export type CanisterId = Principal;
+
+export interface CanisterActorParams {
+  canisterId: CanisterId | string;
```
lmuntaner, line 19: `canisterId` is specific to `queryCanisterDetails`. I'm not sure it should be part of the common type.

frontend/src/lib/types/canister.ts
```diff
@@ -13,3 +14,10 @@ export interface CanisterSync {
 }
 
 export type CanisterId = Principal;
+
+export interface CanisterActorParams {
```
lmuntaner, line 18: Why did you put it here? Isn't this type related more to worker api, than to canisters?

frontend/src/lib/utils/canisters.utils.ts
```diff
@@ -70,3 +72,6 @@ export const canisterStatusToText = (status: CanisterStatus): string => {
       return i18nObj.canister_detail.status_running;
   }
 };
+
+export const mapCanisterId = (canisterId: CanisterId | string): CanisterId =>
```
lmuntaner, line 76: thanks!

frontend/src/lib/api/canisters.api.cjs.ts
lmuntaner, line 1: I guess you'll move this file to the new package in another PR, right?

lmuntaner (commented):
Thanks LFTM, just one comment regarding the common type and the file where you put it.

frontend/src/lib/types/canister.ts
peterpeterparker: The two new jobs of the PR I'm currently developing needs a `canisterId` so do the utils `createCanisterCjs` that's why I think it is part of the definition.

But if you rather like to duplicate it in each declaration, no strong opinion. Let me know!

frontend/src/lib/types/canister.ts
peterpeterparker: Because I did not had better idea 😄. 

Where would you put it? Happy to move it elsewhere.

frontend/src/lib/api/canisters.api.cjs.ts
peterpeterparker: Yes, I want to provide renaming and moving in another PR.

frontend/src/lib/types/canister.ts
lmuntaner: Yes, I understand that there is some duplication. But to me that you need to `Omit` something to use it in the TVL is a sign that the type is not common.

frontend/src/lib/types/canister.ts
lmuntaner: How about `types/worker.ts`?

frontend/src/lib/types/canister.ts
peterpeterparker: Actually TVL has a `canisterId` too, we omit it in the parameter because it's an environment variable and not dynamic.

I can also remove the `Omit` and pass the environment variable in the call.

frontend/src/lib/types/canister.ts
peterpeterparker: Why not 👍. Done.

frontend/src/lib/types/canister.ts
peterpeterparker: I have removed the `canisterId` from the interface and spread it.

lmuntaner: Yes, we need to pass it. The point is that these apis can't access the environment variables.

frontend/src/lib/api/tvl.api.cjs.ts
```diff
@@ -1,29 +1,35 @@
 import { TVLCanister } from "$lib/canisters/tvl/tvl.canister";
 import type { TvlResult } from "$lib/canisters/tvl/tvl.types";
 import { TVL_CANISTER_ID } from "$lib/constants/canister-ids.constants";
```
lmuntaner, line 3: I realized that this also needs to be passed as a parameter.

lmuntaner (changes requested):
No env vars within worker apis

frontend/src/lib/api/tvl.api.cjs.ts
peterpeterparker: Ah gotcha. Let me do that in this PR too.

lmuntaner (approved):
LGTM, thanks!

PR #2693, opened 2023-06-14
Sns proposal navigation

mstrasinskis:
# Motivation

Add proposal navigation functionality to the sns proposal detail page.

# Changes

- add `PropsalNavigation` to detail page
- `snsProposalId` just to simplify getting proposal id as bigint (later could replace the snsProposalIdString)
- `isUIntString` to validate propsal id from url
- `buildMockSnsProposalsStoreSubscribe` for testing

# Tests

- proposal navigation should be shown on detail page
- snsProposalId
- isUIntString

# Screenshots

Screen Recording 2023-06-14 at 10 14 51

### DIFF
--- a/frontend/src/lib/pages/SnsProposalDetail.svelte
+++ b/frontend/src/lib/pages/SnsProposalDetail.svelte
@@ -1,6 +1,9 @@
 <script lang="ts">
   import { goto } from "$app/navigation";
-  import { buildProposalsUrl } from "$lib/utils/navigation.utils";
+  import {
+    buildProposalsUrl,
+    buildProposalUrl,
+  } from "$lib/utils/navigation.utils";
   import { isNullish, nonNullish } from "@dfinity/utils";
   import { getSnsProposalById } from "$lib/services/$public/sns-proposals.services";
   import type { SnsProposalData, SnsProposalId } from "@dfinity/sns";
@@ -18,10 +21,17 @@
   import { loadSnsParameters } from "$lib/services/sns-parameters.services";
   import { syncSnsNeurons } from "$lib/services/sns-neurons.services";
   import { loadSnsNervousSystemFunctions } from "$lib/services/$public/sns.services";
-  import { snsProposalIdString } from "$lib/utils/sns-proposals.utils";
+  import {
+    snsProposalId,
+    snsProposalIdString,
+    sortSnsProposalsById,
+  } from "$lib/utils/sns-proposals.utils";
   import { authSignedInStore } from "$lib/derived/auth.derived";
   import { debugSnsProposalStore } from "../derived/debug.derived";
   import { isUniverseNns } from "$lib/utils/universe.utils";
+  import { snsFilteredProposalsStore } from "$lib/derived/sns/sns-filtered-proposals.derived";
+  import { navigateToProposal } from "$lib/utils/proposals.utils";
+  import ProposalNavigation from "$lib/components/proposal-detail/ProposalNavigation.svelte";
 
   export let proposalIdText: string | undefined | null = undefined;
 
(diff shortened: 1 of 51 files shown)


frontend/src/lib/pages/SnsProposalDetail.svelte
```diff
@@ -150,8 +158,23 @@
...
+
+  let proposalIds: bigint[];
+  $: proposalIds = nonNullish(universeIdText)
+    ? sortSnsProposalsById(
+        $snsFilteredProposalsStore[universeIdText]?.proposals
+      )?.map(snsProposalId) ?? []
+    : [];
 </script>
 
+{#if nonNullish(proposalIdText) && isUIntString(proposalIdText)}
```
dskloetd, line 170: What kind of values can `proposalIdText` have? Are there cases where it's not a number?

frontend/src/lib/utils/utils.ts
```diff
@@ -390,3 +390,10 @@ export const expandObject = (
     }
     return acc;
   }, {} as Record<string, unknown>);
+
+/**
+ * Returns true only for strings that represent a positive integer.
+ * @example
+ * "0" -> true; "" -> false; "-1" -> false
+ */
+export const isUIntString = (text: string): boolean => /^\d+$/.test(text);
```
dskloetd, line 399: Should `0000` and `0009` be valid?

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
```diff
@@ -189,5 +196,45 @@ describe("SnsProposalDetail", () => {
 
       await waitFor(async () => expect(await po.isContentLoaded()).toBe(false));
     });
+
+    it("should display proposal navigation", async () => {
+      // mock the store to have 3 proposals for navigation
+      jest.spyOn(snsFilteredProposalsStore, "subscribe").mockImplementation(
```
dskloetd, line 202: Would it be too difficult to use the real store and fill the underlying store with proposals?

frontend/src/lib/pages/SnsProposalDetail.svelte
mstrasinskis: It can be undefined, null or any string from url query param `proposal`
```
/proposal/?u=oqek3-yyaaa-aaaaa-aacla-cai&proposal=462
```

frontend/src/lib/utils/utils.ts
mstrasinskis: Good question. Because of the `BigInt(x)` applied, this works and is valid, so `BigInt("00000462")` === `BigInt("462"). But from the function name PoV it's wrong. 
What do you think is more appropriate to update the function logic or to rename it?

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
mstrasinskis: This is a derived store. To make it return the needed values, more stores needs to be updated with a very detailed values.

frontend/src/lib/utils/utils.ts
mstrasinskis: But it probably depends on the language. More of them ignores leading zeros. So should be valid.

frontend/src/tests/lib/pages/SnsProposalDetail.spec.ts
lmuntaner: You would need to set the page store with:

```
page.mock({ data: { universe: rootCanisterId.toText() } });
```

Then add those proposals to the store in that same root canister id. Wouldn't that be enough?

frontend/src/lib/utils/utils.ts
mstrasinskis: This doesn't cover `BigInt("-1")` or `BigInt("") === 0n`

frontend/src/lib/pages/SnsProposalDetail.svelte
mstrasinskis: The user will be redirected to the proposal list in this case.. It's just to be sure that there will be no runtime error between page load and redirection to the proposals.
Let me find a better solution if this is not good enough.

mstrasinskis: The logic in the `if` was replaced with the existent one.

frontend/src/lib/utils/utils.ts
mstrasinskis: Thanks, the function was removed.

frontend/src/tests/page-objects/SnsProposalDetail.page-object.ts
```diff
@@ -16,6 +17,10 @@ export class SnsProposalDetailPo extends BasePageObject {
     return SkeletonDetailsPo.under(this.root);
   }
 
+  getProposalNavigation(): ProposalNavigationPo {
```
dskloetd, line 20: We usually give these getters that only get a child po, a name that ends with `Po`.
So `getProposalNavigationPo`.

This is because in some cases it can get confusing.
For example if the getter returns the PO for the `Neurons` component, then calling it `getNeurons` might suggest that it returns an array of neurons, but actually it returns a single component.
So in that case `getNeuronsPo` is clearer.
And for consistency we try to use that naming pattern everywhere.

dskloetd (approved):

PR #2696, opened 2023-06-14
Extract utils to check force call strategy

peterpeterparker:
# Motivation

Extract utils to check force call strategy and avoid code duplication.


lmuntaner (approved):
Thanks

PR #2697, opened 2023-06-14
E2E: Stake NNS neuron

dskloetd:
# Motivation

We want to replace the old wdio e2e tests with new Playwright e2e tests that use page objects.

# Changes

1. Introduce `frontend/src/tests/e2e/neurons.spec.ts` which stakes an NNS neuron.
2. Changes to components to expose test IDs.
3. Page objects.

The goal is to make the test do a lot more and also verify the staked amount etc. but this minimal test already resulted in a non-small PR so more will be added in further PRs.

# Tests

`npm run test-e2e -- neurons`


lmuntaner (approved):
LGTM, thanks!

PR #2699, opened 2023-06-15
Faster wasm network independence test

bitdivine:
# Motivation
We have a test that checks that the wasm is the same, regardless of which network we are building for.  This test is commonly the last to finish in CI:

Screenshot from 2023-06-15 09-43-37

It would be nice if CI were to finish sooner.

# Changes
* That the wasm does not depend on the network is guaranteed by the fact that the network is not an input to any of the wasm build steps in the Dockerfile.  As such, it probably suffices to perform the check only if the dockerfile has changed.
* If the test needs to be performed, use a larger runner.

# Tests
* See CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -272,15 +272,28 @@ jobs:
           rm -rf screenshots
   network_independent_wasm:
     name: "Same wasms for mainnet and local"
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
Rubber stamp

PR #2701, opened 2023-06-15
Feat: Store data utils and type

lmuntaner:
# Motivation

We want stores to hold the error, loading and success states.

This PR: Introduces a type that should be used by the stores and a helper to convert the type to the success type.

The type is already used in the useCountryStore.

The util will be used in the util from PR #2700 [swapSaleBuyerCount](https://github.com/dfinity/nns-dapp/pull/2700/files#diff-aa44a48b94ac01459ed5ebb708a264a70bbd0b1c1afc4616da02d00053131738R4)

# Changes

* New type `StoreData`.
* Use new type in `userCountryStore`.
* New store util `isStoreData`. Not used yet, only tested.

# Tests

* New test file for new store util `isStoreData`.

### DIFF
--- a/frontend/src/lib/types/store.ts
+++ b/frontend/src/lib/types/store.ts
@@ -1,0 +1,1 @@
+export type StoreData<T> = T | Error | "not loaded";
--- a/frontend/src/lib/types/store.ts
+++ b/frontend/src/lib/types/store.ts
@@ -1,1 +1,8 @@
-export type StoreData<T> = T | Error | "not loaded";
+import type { NOT_LOADED } from "$lib/constants/stores.constants";
+
+/**
+ * - Not Loaded: NOT_LOADED
+ * - Error: Error
+ * - Success: T
+ */
+export type StoreData<T> = T | Error | typeof NOT_LOADED;
--- a/frontend/src/tests/lib/utils/store.utils.spec.ts
+++ b/frontend/src/tests/lib/utils/store.utils.spec.ts
@@ -1,12 +1,14 @@
 import { isStoreData } from "$lib/utils/store.utils";
 
-describe("isStoreData", () => {
-  it("should return false if data is error or not loaded", () => {
-    expect(isStoreData("not loaded")).toBe(false);
-    expect(isStoreData(new Error())).toBe(false);
-  });
+describe("store utils", () => {
+  describe("isStoreData", () => {
+    it("should return false if data is error or not loaded", () => {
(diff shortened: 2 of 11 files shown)


lmuntaner:

frontend/src/lib/types/store.ts
```diff
@@ -0,0 +1 @@
+export type StoreData<T> = T | Error | "not loaded";
```
dskloetd, line 1: How does this work if `T` is `string` while `"not loaded"` is also a string?
Can we require that `T` is not `string`?
Or should we create a separate data type for `"not loaded"`?

frontend/src/tests/lib/utils/store.utils.spec.ts
```diff
@@ -0,0 +1,14 @@
+import { isStoreData } from "$lib/utils/store.utils";
+
+describe("store utils", () => {
+  describe("isStoreData", () => {
+    it("should return false if data is error or not loaded", () => {
+      expect(isStoreData("not loaded")).toBe(false);
+      expect(isStoreData(new Error())).toBe(false);
+    });
+
+    it("should return true if data is error or not loaded", () => {
```
dskloetd, line 10: data is NOT error NOR not loaded?

frontend/src/lib/types/store.ts
lmuntaner: It could be an edge case that the successful value is `"not loaded"` and it would be considered as not loaded.

The same way that any type we use as loading state.

I wouldn't require `T` to not be a string. As long as it's not `"not loaded"` we should be fine.

frontend/src/tests/lib/utils/store.utils.spec.ts
lmuntaner: correct, typo

frontend/src/lib/types/store.ts
lmuntaner: I added a test with a string as successful data.

lmuntaner:

frontend/src/lib/types/store.ts
dskloetd: I know it's a super edge case but it's easily avoided with something like
```
const NOT_LOADED = Symbol("not loaded");
```
right?

lmuntaner: Sounds good!

lmuntaner:

dskloetd (approved):
Thanks!

PR #2702, opened 2023-06-15
Build wasm without assets

bitdivine:
# Motivation
Deploying to an application subnet requires installing the wasm without assets, then installing assets afterwards.

# Changes
Add a new docker build stage that builds the nns-dapp wasm with an empty assets tarfile.

# Tests
None.  This should be tested in combination with the chunk uploading PR, once merged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -156,6 +156,32 @@ RUN apt-get update -yq && apt-get install -yqq --no-install-recommends file
 ARG COMMIT
 RUN scripts/dfx-wasm-metadata-add --commit "$COMMIT" --canister_name nns-dapp --verbose
 
+# Title: Image to build the nns-dapp backend.
+FROM builder AS build_nnsdapp_without_assets
+SHELL ["bash", "-c"]
+COPY ./rs/backend /build/rs/backend
+COPY ./scripts/nns-dapp/test-exports /build/scripts/nns-dapp/test-exports
+COPY ./scripts/clap.bash /build/scripts/clap.bash
+COPY ./build-backend.sh /build/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -156,6 +156,38 @@ RUN apt-get update -yq && apt-get install -yqq --no-install-recommends file
 ARG COMMIT
 RUN scripts/dfx-wasm-metadata-add --commit "$COMMIT" --canister_name nns-dapp --verbose
 
+# Title: Image to build the nns-dapp backend without assets.
+FROM builder AS build_nnsdapp_without_assets
```
lmuntaner, line 160: Will this run with the normal `docker-build`?

Dockerfile
```diff
@@ -189,6 +221,7 @@ COPY --from=configurator /build/nns-dapp-arg* /
 # Note: The frontend/.env is kept for use with test deployments only.
 COPY --from=configurator /build/frontend/.env /frontend-config.sh
 COPY --from=build_nnsdapp /build/nns-dapp.wasm /
+COPY --from=build_nnsdapp_without_assets /build/nns-dapp.wasm /nns-dapp_noassets.wasm
 COPY --from=build_nnsdapp /build/assets.tar.xz /
```
lmuntaner, line 225: This is the file needed to upload afterwards, right?

lmuntaner (approved):
Not a master at docker, but LGTM.

Just a few minor questions to understand it.

Dockerfile
bitdivine: Yes, it will.

PR #2706, opened 2023-06-15
Fix proposal navigation visibility during voting

mstrasinskis:
# Motivation

Fix the missing proposal navigation buttons during a vote registration.
Sns fix will be implemented in another pr.

# Changes

- Combine the ids for navigation component from available for voting proposal ids and currently in voting process.
- ProposalNavigation cares about the ids order and sorts them (because the combined list).

# Tests

- ProposalNavigation sorts the ids
- The issue itself could be better addressed by an end-to-end test to ensure that the navigation is functioning properly even during the registration process.

# Screenshots

Screen Recording 2023-06-15 at 15 42 07

### DIFF
--- a/frontend/src/lib/components/proposal-detail/NnsProposal.svelte
+++ b/frontend/src/lib/components/proposal-detail/NnsProposal.svelte
@@ -36,7 +36,7 @@
     voteRegistrationStore: VoteRegistrationStoreData;
   }): bigint[] => {
     const proposalIds =
-      $filteredProposals.proposals?.map(({ id }) => id as bigint) || [];
+      proposalStore.proposals?.map(({ id }) => id as bigint) || [];
     const proposalIdsInRegistrations = (
       voteRegistrationStore.registrations[OWN_CANISTER_ID_TEXT] ?? []
     ).map(({ proposalIdString }) => BigInt(proposalIdString));
--- a/frontend/src/lib/components/proposal-detail/NnsProposal.svelte
+++ b/frontend/src/lib/components/proposal-detail/NnsProposal.svelte
@@ -13,43 +13,14 @@
   import NnsProposalProposerPayloadEntry from "./NnsProposalProposerPayloadEntry.svelte";
   import { filteredProposals } from "$lib/derived/proposals.derived";
   import { navigateToProposal } from "$lib/utils/proposals.utils";
-  import {
-    voteRegistrationStore,
-    type VoteRegistrationStoreData,
-  } from "$lib/stores/vote-registration.store";
-  import { OWN_CANISTER_ID_TEXT } from "$lib/constants/canister-ids.constants";
-  import type { ProposalsStore } from "$lib/stores/proposals.store";
 
   const { store } = getContext<SelectedProposalContext>(
     SELECTED_PROPOSAL_CONTEXT_KEY
   );
 
-  // The proposal that is currently in vote registration process is not included in the filtered proposals list,
-  // since it should not be available for voting.
-  // But this shouldn't prevent user from navigating to it.
-  // Here we are adding this proposal to the list of proposals, so that the navigation block is not hidden for it.
-  const collectProposalIds = ({
-    proposalStore,
-    voteRegistrationStore,
-  }: {
-    proposalStore: ProposalsStore;
-    voteRegistrationStore: VoteRegistrationStoreData;
-  }): bigint[] => {
-    const proposalIds =
-      proposalStore.proposals?.map(({ id }) => id as bigint) || [];
-    const proposalIdsInRegistrations = (
-      voteRegistrationStore.registrations[OWN_CANISTER_ID_TEXT] ?? []
-    ).map(({ proposalIdString }) => BigInt(proposalIdString));
-
-    // return unique proposal ids
-    return Array.from(new Set([...proposalIds, ...proposalIdsInRegistrations]));
-  };
-
   let proposalIds: bigint[] | undefined;
-  $: proposalIds = collectProposalIds({
-    proposalStore: $filteredProposals,
-    voteRegistrationStore: $voteRegistrationStore,
-  });
+  $: proposalIds =
+    $filteredProposals.proposals?.map(({ id }) => id as bigint) ?? [];
 </script>
 
 {#if $store?.proposal?.id !== undefined}
--- a/frontend/src/lib/components/proposal-detail/NnsProposal.svelte
+++ b/frontend/src/lib/components/proposal-detail/NnsProposal.svelte
@@ -19,8 +19,7 @@
   );
 
   let proposalIds: bigint[] | undefined;
-  $: proposalIds =
-    $filteredProposals.proposals?.map(({ id }) => id as bigint) ?? [];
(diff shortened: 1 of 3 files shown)


frontend/src/lib/components/proposal-detail/NnsProposal.svelte
```diff
@@ -13,13 +13,43 @@
...
+    type VoteRegistrationStoreData,
+  } from "$lib/stores/vote-registration.store";
+  import { OWN_CANISTER_ID_TEXT } from "$lib/constants/canister-ids.constants";
+  import type { ProposalsStore } from "$lib/stores/proposals.store";
 
   const { store } = getContext<SelectedProposalContext>(
     SELECTED_PROPOSAL_CONTEXT_KEY
   );
 
+  // The proposal that is currently in vote registration process is not included in the filtered proposals list,
```
mstrasinskis, line 27: Here is the description of the issue.

lmuntaner (approved):
Great, thanks!

mstrasinskis:
The last commit should fix the navigation issue with the checkbox selected. 

@dskloetd, since you discovered it could you please verify the behaviour?

https://github.com/dfinity/nns-dapp/assets/98811342/0eba2848-656f-4775-bed7-6ce46fe1d075

mstrasinskis:
@dskloetd extra logic that combines 2 arrays solves the current issue of the temporary unavailable for navigation proposal.

1. start voting - proposal 469 (w/ all or part of the votable neurons)
2. click "Older" , goto 468
3. click "Newer", goto 470. Proposal 469 was skipped. (screenshot)

## prod version
https://github.com/dfinity/nns-dapp/assets/98811342/f7ac438c-5e80-4c4c-9a51-391467eaaf1a

With the updated version, proposals that are currently in the voting process are still available for navigation.

dskloetd:
* Is the issue that a proposal becomes temporarily unavailable, currently existing on mainnet? If not, why was it working before and why did it break?
* Why does the proposal disappear from the store of proposals that can be voted on while it can still be voted on? It seems to me that that needs to be fixed, rather than papered over.

mstrasinskis:
> Is the issue that a proposal becomes temporarily unavailable, currently existing on mainnet? If not, why was it working before and why did it break?

Yes, it exists on mainnet.

> Why does the proposal disappear from the store of proposals that can be voted on while it can still be voted on? It seems to me that that needs to be fixed, rather than papered over.

Probably it's an obsolete logic (to remove it from the store) to prevent user from simultaneous voting. Before the UX was different, with the loading screen during the voting process.

dskloetd:
> Probably it's an obsolete logic (to remove it from the store) to prevent user from simultaneous voting. Before the UX was different, with the loading screen during the voting process.

If the logic is obsolete, then we should fix that instead of working around it.
It the same behavior exists on mainnet, then maybe just merge the fix for the regression now and look into the proper fix for the existing issue separately?

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -61,7 +61,7 @@ describe("ProposalNavigation", () => {
     it("should disable previous button when it's selected", async () => {
       const po = renderComponent({
         currentProposalId: 1n,
-        proposalIds: [1n, 2n],
+        proposalIds: [0n, 1n],
```
dskloetd, line 64: Why this change?
Now `1n` is the last proposal so previous should be enabled and next should be disabled, right?

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -72,7 +72,7 @@ describe("ProposalNavigation", () => {
     it("should disable next when it's selected", async () => {
       const po = renderComponent({
         currentProposalId: 1n,
-        proposalIds: [0n, 1n],
+        proposalIds: [1n, 2n],
```
dskloetd, line 75: Same

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -81,33 +81,31 @@ describe("ProposalNavigation", () => {
...
-      const po = renderComponent({
-        currentProposalId: 1n,
-        proposalIds: [0n, 1n, 2n],
-        selectProposal: selectProposalSpy,
-      });
+    await po.clickNext();
 
-      await po.clickPrevious();
+    expect(selectProposalSpy).toHaveBeenCalledTimes(1);
+    expect(selectProposalSpy).toHaveBeenCalledWith(0n);
```
dskloetd, line 95: The next ID after `1n`, is `2n`, right? Why do we expect `0n`?

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -81,33 +81,31 @@ describe("ProposalNavigation", () => {
...
-
-      await po.clickNext();
-
-      expect(selectProposalSpy).toHaveBeenCalledTimes(1);
-      expect(selectProposalSpy).toHaveBeenCalledWith(2n);
+  it("should emmit next click", async () => {
+    const selectProposalSpy = jest.fn();
+    const po = renderComponent({
+      currentProposalId: 1n,
+      proposalIds: [0n, 1n, 2n],
```
dskloetd, line 88: Would it make sense to use non-consecutive IDs in the test to catch if we accidentally just return `currentId + 1` instead of the actual next ID? etc.

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
```diff
@@ -81,33 +81,31 @@ describe("ProposalNavigation", () => {
...
+      selectProposal: selectProposalSpy,
     });
+
+    await po.clickPrevious();
+
+    expect(selectProposalSpy).toHaveBeenCalledTimes(1);
+    expect(selectProposalSpy).toHaveBeenCalledWith(2n);
```
dskloetd, line 109: same

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
dskloetd: Also, please add more IDs. If there's only a single ID smaller than `1n`, the test doesn't distinguish between getting the smallest or largest ID smaller than `1n`.

mstrasinskis: There is now a separate test for non-consecutive IDs.

dskloetd (approved):
Thanks!

frontend/src/tests/lib/components/proposal-detail/ProposalNavigation.spec.ts
dskloetd: For the record. The problem was that the IDs were passed in increasing order while they needed to be passed in in decreasing order.

PR #2707, opened 2023-06-15
Correct esm import path for type

peterpeterparker:
# Motivation

Correct esm import path for a type `Principal` that referenced cjs.


lmuntaner (approved):
Thanks!

PR #2708, opened 2023-06-15
Move and rename Cjs to worker

peterpeterparker:
# Motivation

Drop `cjs` prefix and suffix for `worker`.

# Changes

- create new packages `worker-api` and `worker-services` 
- add new packaging to chunking bundling rules
- move services, api and utils to the `worker-*` packages
- rename few variables and types from `Cjs` to `Worker`
- rename `timer.util` to `timer.utils` to follow existing pattern


lmuntaner (approved):
Thanks!

PR #2710, opened 2023-06-16
Fix: steps not needs

bitdivine:
# Motivation
The code that runs on dockerfile change never runs.  This is because the wrong namespace is used for the github variable.

# Changes
- `s/needs/steps/g`

# Tests
- No run with no change to the dockerfile:
Screenshot from 2023-06-16 06-24-46

- Run with a trivial change to the dockerfile: 
Screenshot from 2023-06-16 06-33-57

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-# Use this with
+# Use this with:
 #
 # docker build . -t nns-dapp
 # container_id=$(docker create nns-dapp no-op)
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Use this with
+# Use this with:
```
bitdivine, line 1: This is a change just to exercise the action, however I think that, while trivial, it is worth keeping.

peterpeterparker (approved):
LGTM

PR #2713, opened 2023-06-16
Feat: Support derived_state field in SNS Aggregator

lmuntaner:
# Motivation

Use the new field `direct_participant_count` from derived state to display the number of participants of a swap instead of relying in the metrics raw endpoint.

In this PR: Use the SNS aggregator field derived_state which supports the latest properties.

# Changes

* Add `derived_state` to the type of the SNS Aggregator resposne.
* Convert the type of the aggregator to the type needed by the snsQueryStore.
* Update the snsQueryStore with the value in the field `derived_state` from the aggregator.
* Fix the method `updateDerivedState` of the snsQueryStore that was ignoring the new fields.

# Tests

* Add new field to the sns aggregator response mocks and converted mock.
* Test that the new field is used to update the snsQueryStore in the sns service.
* Check also the new fields in the tests for `updateDerivedState` in snsQueryStore test file

### DIFF
--- a/frontend/src/lib/api/sns-aggregator.api.ts
+++ b/frontend/src/lib/api/sns-aggregator.api.ts
@@ -10,6 +10,7 @@ import type {
 } from "@dfinity/ledger";
 import type {
   SnsFunctionType,
+  SnsGetDerivedStateResponse,
   SnsGetMetadataResponse,
   SnsNervousSystemFunction,
   SnsSwap,
@@ -61,6 +62,7 @@ export type CachedSns = {
    */
   icrc1_fee?: bigint;
   icrc1_total_supply: bigint;
+  derived_state: SnsGetDerivedStateResponse;
 };
 
 type CachedSnsMetadataDto = {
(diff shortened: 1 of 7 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/api/sns-aggregator.api.ts
```diff
@@ -116,9 +118,9 @@ type CachedSnsSwapDto = {
 type CachedSnsSwapDerivedDto = {
   buyer_total_icp_e8s: number;
   sns_tokens_per_icp: number;
-  cf_participant_count: number | undefined;
-  direct_participant_count: number | undefined;
-  cf_neuron_count: number | undefined;
+  cf_participant_count: number | undefined | null;
```
dskloetd, line 121: Are both `undefined` and `null` needed here or can it be `null` instead of `undefined`?

dskloetd (approved):

frontend/src/lib/api/sns-aggregator.api.ts
lmuntaner: Because it might be that it's not supported (undefined) or that is supported but empty (null).

PR #2716, opened 2023-06-16
Render proposal arguments with more than one field

bitdivine:
# Motivation
There was a design error, in which I was under the impression that proposal arguments were represented by IDLTypes.  I knew that proposals could have multiple arguments but was under the impression that a vector of values was itself an IDLType.  For a long time this fallacy seemed to be correct as proposals most commonly have one argument.  We recently encountered a proposal with no arguments and this broke the proposal rendering.  On investigation, there is a type called IDLArgs that is roughly equivalent to a vector of IDLTypes but that is not itself an IDLType.  It is different from an IDLType vector in that in a vector, the types of the contents are all the same whereas for IDLArgs, they can be different.  In the common case, where there is a single type, the two behave in much the same way.  This is how the false assumption passed scrutiny for so long.

A quick fix was added that handles the case where there are no arguments and tests were added for the cases where there are multiple arguments.  However in the case of multiple arguments, only the first was rendered.

The idl2json library how offers a proper solution, in which `IDLArgs` are parsed properly, so we can use that and remove the hack.  This does however change the output.  Previously the JSON output was a single JSONValue corresponding to the first argument.  Now the output is an array of JSONValues, with as many entries as there are arguments.

# Changes
- Update the idl2json library version.
- Replace the partial logic for handing variable args with the library call that handles them all correctly.

# Questions
- Does anything need to be updated with the front-end proposal rendering?

# Tests
There are existing unit tests for the conversion.  These have been updated.


rs/backend/src/proposals.rs
```diff
@@ -112,23 +111,17 @@ fn canister_arg_types(canister_id: Option<CanisterId>) -> IDLTypes {
...
-        Ok(idl_value) => {
+    // TODO: Test empty payload
+    // TODO: Test existing payloads
+    // TODO: Test muti-value payloads
+    match IDLArgs::from_bytes(arg) {
+        Ok(idl_args) => {
             let options = Idl2JsonOptions {
                 bytes_as: Some(BytesFormat::Hex),
                 long_bytes_as: None,
+                prog: Vec::new(), // These are the type definitions used in proposal payloads.  If we have them, it would be nice to use them.  Do we?
```
lmuntaner, line 122: Do you mean we could pass known argument types?

lmuntaner (commented):
Can you deploy this branch to a testnet?

Try then with proposal 181 how it looks.

And compare it with how it looks now: https://kptzy-hiaaa-aaaaa-aacsa-cai.nnsdapp.dfinity.network/proposal/?u=kptzy-hiaaa-aaaaa-aacsa-cai&proposal=181

bitdivine:
# Deployments

## staging
```
git checkout testnets canister_ids.json
DFX_NETWORK=staging ; ./scripts/docker-build --network "$DFX_NETWORK" && dfx-canister-forget -n "$DFX_NETWORK" -c nns-dapp && dfx canister create --network "$DFX_NETWORK" nns-dapp && dfx canister install --network "$DFX_NETWORK" nns-dapp --wasm out/nns-dapp.wasm --argument "$(cat out/nns-dapp-arg-${DFX_NETWORK}.did)" --mode reinstall --yes
dfx-canister-url --network "$DFX_NETWORK" nns-dapp
```

bitdivine:
@lmuntaner Here is another proposal.  It's an nns-dapp canister upgrade proposal with a difference:
- I have a used a smaller wasm so that the payload is rendered.
- I have added a second field to the init args.
- Note that here the second field is shown: https://ll6wk-6iaaa-aaaaa-aacua-cai.nnsdapp.dfinity.network/proposal/?u=kptzy-hiaaa-aaaaa-aacsa-cai&proposal=489
  
Screenshot from 2023-06-18 04-54-25
- Whereas previously it is invisible:  
Screenshot from 2023-06-18 04-55-21

lmuntaner:
This is ugly to read, but I understand that this is more accurate.

Args is a list of arg, and each arg is optional. Therefore, each arg is represented as an array.

Conclusion, args is an array of arrays. One argument is an array of one element, which is an array of one element also.

Not in this PR, but should we simplify the optionals into the argument or undefined?

lmuntaner (approved):

bitdivine:
Re: Simplifying options, here is a Jira ticket: https://dfinity.atlassian.net/browse/GIX-1621

rs/backend/src/proposals.rs
bitdivine: Yes, and known types in general.  A did file, parsed, is an IDLProg.  So if there are .did files defining types used in payloads, this _may_ be the right place to pass them in.

lmuntaner: Then we can alreay try with the Internet Identity Init type, right?

Instead of:

```
/// Best effort to determine the types of a cansiter args.
fn canister_arg_types(canister_id: Option<CanisterId>) -> IDLTypes {
    // If canister id is II
    // use InternetIdentityInit type
    let args = if canister_id == Some(IDENTITY_CANISTER_ID) {
        let idl_type = internal_candid_type_to_idl_type(&InternetIdentityInit::ty());
        vec![IDLType::OptT(Box::new(idl_type))]
    } else {
        vec![]
    };
    IDLTypes { args }
}
```

PR #2718, opened 2023-06-19
Get proposal args

bitdivine:
# Motivation
We would like to be able to examine the arguments in a proposal.

See: https://dfinity.atlassian.net/jira/software/c/projects/GIX/boards/182?modal=detail&selectedIssue=GIX-1582

# Changes
- Add a command to get the arguments from a proposal

# Example
The code currently produces this:
```
$ scripts/dfx-nns-proposal-args --network ic --proposal 123010
[
  [
    {
      "args": [
        {
          "0": "API_HOST",
          "1": "https://icp-api.io"
        },
        {
          "0": "CKBTC_INDEX_CANISTER_ID",
          "1": "n5wcd-faaaa-aaaar-qaaea-cai"
        },
====8<---------------------------------------------------------
        {
          "0": "WASM_CANISTER_ID",
          "1": "qaa6y-5yaaa-aaaaa-aaafa-cai"
        }
      ]
    }
  ]
]
Arguments hash: 4c5b9a52ce9e099af6a583601cceb4af8d3f5d2cd2602b22833b28d5123f717e  -
max@sinkpad:~/dfn/nns-dapp (2days,21:34:32)$```
```

# Tests
- There is a test that arguments can be obtained from a sample proposal on mainnet.


scripts/mk_nns_patch.sh
```diff
@@ -1,8 +1,25 @@
 #!/usr/bin/env bash
 set -euo pipefail
```
bitdivine, line 2: Note: This filter is needed because previously this code assumed that every entry under declarations was a directory called `${canister_name}` containing `${canister_name}.did`.  This is no longer the case, so we need to check the assumption before proceeding.

declarations/nns_governance/proposals.did
```diff
@@ -0,0 +1,17 @@
+// Source: Derived manually from the Rust type declaration in the ic repo:
+// 0472cf2cc6cf174f3efc9531d5947ea10cd0109f (HEAD -> rc--2023-04-04_23-01, origin/rc--2023-04-04_23-01)
+// Date:   Tue Apr 4 21:03:14 2023 +0000
```
bitdivine, line 3: Note: It would be nice to get a did file describing all proposals from upstream but as far as I can see no such thing exists.  So this contains just the one type declaration we need most.

bitdivine:
Note: This is blocked on bumping didc.

declarations/nns_governance/proposals.did
bitdivine: - [Discussion on whether a relevant did file exists (no) and how one could be make.](https://dfinity.slack.com/archives/C0175CHJ0P6/p1687276399328309?thread_ts=1687064130.422249&cid=C0175CHJ0P6)

bitdivine:
A little clarification is needed:
- Should the output, optionally, emit just the hash of the arguments?
- With some additional effort the parsing above is complete, with names for keys.  However if only the hash is needed, I would just add the minimal code needed to print the hash.

Answer:
> Maksims Strasinskis
  [10:05 AM](https://dfinity.slack.com/archives/D031GD20W01/p1687161946690589)
Good morning Max, I would prefer to see not only the hash but also the values. At least to to verify that there were provided the expected flag values.

PR #2719, opened 2023-06-19
Utils to extend JSON stringify and parse

peterpeterparker:
# Motivation

In #2639 we will pass complex objects back and forth between the UI and web workers. While the browser can serialize most data, their some types that are not supported by the browser (`bigint` or `Uint8Array`) and some that can be serialized but because they are ojbect, looses their functions during the process (for example `Principal.getText()`).

To solve the serialization issue we can stringify and parse objects when we exchange data between UI and web workers and provide special replacer and reviver.

# Changes

- new `jsonReplacer` and `jsonReviver`
- add a TODO / question for an existing replacer


frontend/src/lib/stores/writable-stored.ts
```diff
@@ -44,6 +44,7 @@ export const writableStored = <T>({
 
     // Do not break UI if local storage fails
     try {
+      // TODO: can / should we replace this replacer with json.utils.jsonReplacer? is it possible without breaking changes?
```
lmuntaner, line 47: For me yes, it makes sense.

Not necessarily in this PR

lmuntaner (approved):
Thanks!

frontend/src/lib/stores/writable-stored.ts
peterpeterparker: Me too, not sure if it is possible to do so without a breaking change?

lmuntaner: We'll need to check where the `writable` is used. Let's add a TODO.

My idea was to extend `writable` to provide versioning also. Maybe we should do that before implementing this change?

peterpeterparker: Let's do that 👍

PR #2721, opened 2023-06-19
Icrc balance and transactions api for workers

peterpeterparker:
# Motivation

In #2639 we will need to fetch Icrc balances and transactions from workers, this PR provides specific API for it.

# Notes

In the future  we might be able to refactor the app to just use the same API for any UI and workers but, thought it's a first step.

# Changes

- new `icrc-index.worker-api` and `icrc-ledger.worker-api`


lmuntaner (approved):
Thanks!

PR #2723, opened 2023-06-20
Fix: Navigate from SNS proposal to launchpad

lmuntaner:
# Motivation

There was a bug when the user navigated from the SNS Proposal detail page to the launchpad. The user was redirected to the proposals page.

The issue is a problem with Svelte transitions in components that persist across routes. In our case the `TotalValueLocked.svelte`.

The fix was to use CSS transition instead of a Svelte transition in that component.

# Changes

* Remove Svelte transition in `TotalValueLocked` and use a CSS transition instead.

# Tests

* Fix tests after refactor.

### DIFF
--- a/frontend/src/lib/components/metrics/TotalValueLocked.svelte
+++ b/frontend/src/lib/components/metrics/TotalValueLocked.svelte
@@ -18,16 +18,15 @@
     });
 </script>
 
-{#if nonNullish(total) && total > 0}
-  <div
-    class="tvl"
-    transition:fade={{ duration: 125 }}
-    class:stacked={layout === "stacked"}
-  >
-    <span>{$i18n.metrics.tvl}</span>
-    <span data-tid="tvl-metric" class="total">${format(total)}</span>
-  </div>
-{/if}
+<!-- DO NOT use a Svelte transition. It caused issues with navigation -->
+<div
+  class="tvl"
+  class:visible={nonNullish(total) && total > 0}
+  class:stacked={layout === "stacked"}
+>
+  <span>{$i18n.metrics.tvl}</span>
+  <span data-tid="tvl-metric" class="total">${format(total ?? 0)}</span>
+</div>
 
 <style lang="scss">
   @use "@dfinity/gix-components/dist/styles/mixins/fonts";
@@ -35,6 +34,15 @@
 
   .tvl {
     display: inline-block;
+
+    visibility: hidden;
+    opacity: 0;
+    transition: opacity 150ms, visibility 150ms;
+
+    &.visible {
+      visibility: visible;
+      opacity: 1;
+    }
 
     text-align: center;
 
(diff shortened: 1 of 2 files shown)


lmuntaner:
@peterpeterparker @dskloetd @mstrasinskis please review

dskloetd (approved):
This change looks fine but I have no idea how this could cause the problem so it would be great to document it in more detail in the comment and/or PR description, for future reference.

frontend/src/lib/components/metrics/TotalValueLocked.svelte
```diff
@@ -36,6 +35,15 @@
   .tvl {
     display: inline-block;
 
+    visibility: hidden;
+    opacity: 0;
+    transition: opacity 150ms, visibility 150ms;
```
peterpeterparker, line 40: - visibility cannot be animated therefore can be ommited here
- you can maybe use a standard timing function

e.g.

```
transition: opacity ease-out var(--animation-time-normal);
```

peterpeterparker (approved):
LGTM, just a small comment about the CSS

peterpeterparker:
> This change looks fine but I have no idea how this could cause the problem so it would be great to document it in more detail in the comment and/or PR description, for future reference.

@dskloetd rather than documenting it, if anyone has time, I would rather suggest to create a sample repo that replicates the issue and open one with it in SvelteKit repo.

There are few similar problems already open in Sveltekit but none have provided a simple repo, I guess this would be also be beneficial for lots of devs too.

dskloetd:
That would be great, **if anyone has time**. But explaining how it happens (if you understand) should take a lot less time. If someone opens an issue with SvelteKit then it would be good to link it here.

frontend/src/lib/components/metrics/TotalValueLocked.svelte
lmuntaner: I took that from another CSS rule, I was also confused with it. Thanks for the comment.

lmuntaner:
> There was a bug when the user navigated from the SNS Proposal detail page to the launchpad. The user was redirected to the proposals page.
> 
> The issue is a problem with Svelte transitions in components that persist across routes. In our case the `TotalValueLocked.svelte`.
> 
> The fix was to use CSS transition instead of a Svelte transition in that component.

I used the PR title and the description to explain the problem, the cause of the problem and the fix. I also added a commet in the TotalValueLocked.svelte to not revert the changes.

What else would you suggest?

dskloetd:
I have read the PR title, description and comment and I have on idea how this could have caused the issue. So I can not suggest a better explanation but I do know the explanation isn't very clear. We can talk about it later. I don't mean to block the PR if you don't fully understand the issue either.

mstrasinskis (approved):
Thank you!

PR #2724, opened 2023-06-20
Decode payment (amount) from QR code reader

peterpeterparker:
# Motivation

Extend QR code reader in transaction to extract encoded amount  that way users can proceed payment in NNS dapp.

# PRs

- [x] https://github.com/dfinity/ic-js/pull/354

# Changes

- bump ic-js to use new function `decodePayment`
- use function. if token does not match throw and error, if matches apply identifier and if provided amount

# Tests

It would need e2e test.

I tried various ways to simulate the `nnsQRCode` custom event in jest test (as we have to skip the QRCode reader in such context) but, did not make it. Of course let me know if you know how.


frontend/src/lib/modals/transaction/TransactionModal.svelte
```diff
@@ -96,6 +98,31 @@
...
+
+      if (nonNullish(paymentAmount)) {
+        amount = paymentAmount;
+      }
+
+      goForm();
+      return;
+    }
+
     selectedDestinationAddress = value;
```
lmuntaner, line 126: So the QR code will either read a single address or an encoded payment, right?

lmuntaner (approved):
LGTM, just a single question.

frontend/src/lib/modals/transaction/TransactionModal.svelte
peterpeterparker: Exactly. It tries to decode the payment (token+adress+optional amount). If it can, then it extract the info, if it cannot extract then it fallbacks to current behavior which just parse what it reads in the input field.

The input field(s) triggers then the validation.

PR #2725, opened 2023-06-20
Icrc worker store

peterpeterparker:
# Motivation

In few workers we need to hold some values in worker memory as in #2639 in which we need to keep track of ICRC balances and transactions.

That's why this PR introduces an ICRC store that can be use in worker.

# Changes

- new `IcrcWorkerStore` store
- new package `worker-store`
- add type for `IcrcAccountIdentifierText`, a `string`

### DIFF
--- a/frontend/src/lib/worker-stores/icrc.worker-store.ts
+++ b/frontend/src/lib/worker-stores/icrc.worker-store.ts
@@ -1,0 +1,40 @@
+import type { IcrcAccountIdentifierText } from "$lib/types/icrc";
+
+export interface IcrcWorkerData {
+  accountIdentifier: IcrcAccountIdentifierText;
+  certified: boolean;
+}
+
+export type IcrcWorkerState<T extends IcrcWorkerData> = Record<
+  IcrcAccountIdentifierText,
+  T
+>;
+
+export class IcrcWorkerStore<T extends IcrcWorkerData> {
+  private readonly EMPTY_STATE: IcrcWorkerState<T> = {};
+  private _state: IcrcWorkerState<T> = this.EMPTY_STATE;
+
+  update(data: T[]) {
+    this._state = {
+      ...this._state,
+      ...data.reduce(
+        (acc, { accountIdentifier, ...rest }) => ({
+          ...acc,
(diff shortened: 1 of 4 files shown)


frontend/src/lib/worker-stores/icrc.worker-store.ts
```diff
@@ -0,0 +1,40 @@
...
+  accountIdentifier: IcrcAccountIdentifierText;
+  certified: boolean;
+}
+
+export type IcrcWorkerState<T extends IcrcWorkerData> = Record<
+  IcrcAccountIdentifierText,
+  T
+>;
+
+export class IcrcWorkerStore<T extends IcrcWorkerData> {
```
lmuntaner, line 13: Why do you need to use a generic?

frontend/src/lib/worker-stores/icrc.worker-store.ts
lmuntaner, line 1: Does it need to be only for ICRC?

Could we use this store for anything we want to store per records or dictionary

More like, `dictionary.worker-store.ts` or `record.worker-store.ts`.

Instead if `accountIdentifier` we call it `key` or `id`.

lmuntaner (commented):
a couple of comments. I like the idea.

frontend/src/lib/worker-stores/icrc.worker-store.ts
peterpeterparker: Because used in two workers:

```
// Worker balance
interface BalanceData extends IcrcWorkerData {
  balance: bigint;
}

const store = new IcrcWorkerStore<BalanceData>();

// Worker transactions
const store = new IcrcWorkerStore<TransactionsData>();
```

frontend/src/lib/worker-stores/icrc.worker-store.ts
peterpeterparker: Sounds cool. I renamed the store from `Icrc` to `Dictionary` and also the `accountIdentifier` index to `key`.

WDYT?

lmuntaner (approved):
Perfect, thanks!

PR #2729, opened 2023-06-21
Use new ic-management library

peterpeterparker:
# Motivation

Introduces `@dfinity/ic-management` while minimizing the changes.

# Changes

- add `@dfinity/ic-management` and bump `next` libs
- intregrate `ICManagementCanister` from ic-js within `ICManagementCanister` of NNS dapp


frontend/src/lib/canisters/ic-management/converters.ts
```diff
@@ -1,12 +1,20 @@
...
 import type { CanisterDetails } from "./ic-management.canister.types";
 import { CanisterStatus } from "./ic-management.canister.types";
-import type {
-  CanisterStatusResponse,
-  CanisterStatus as RawCanisterStatus,
-} from "./ic-management.types";
 
-const getCanisterStatus = (status: RawCanisterStatus): CanisterStatus => {
+const getCanisterStatus = (
+  status:
```
lmuntaner, line 7: Why aren't you using `CanisterStatusResponse`?

frontend/src/lib/canisters/ic-management/ic-management.canister.ts
lmuntaner, line 1: Will you then remove this and create the canister directly in the api layer like we do in the other parts?

The transformations can happen at that level as well.

lmuntaner (approved):
LGTM, just a couple of questions.

frontend/src/lib/canisters/ic-management/converters.ts
peterpeterparker: Where? For what?

frontend/src/lib/canisters/ic-management/ic-management.canister.ts
peterpeterparker: Ultimately yes agree I think we should do it. I added a `TODO` for that but, I might priorize other tasks right now.

lmuntaner: Ok, thanks

frontend/src/lib/canisters/ic-management/converters.ts
lmuntaner: Don't we have this type in the library:

```
{ stopped: null } | { stopping: null } | { running: null }
```

peterpeterparker: Ah no, didc generates `{ stopped: null } | { stopping: null } | { running: null }` and no type for it. That was a custom type.

lmuntaner: Ah, ok, I see now it's part of the whole response, not a specific one. Thanks for the explanation!

PR #2730, opened 2023-06-21
DO NOT MERGE: New initialization in canistersStore

lmuntaner:
# Motivation

This is a draft of how the new store initialization pattern could look like.

If we like the idea, I can implement it in multiple smaller PRs.

The difference from what we discussed is in passing the identity when creating the store. This was the easiest way I found to use the identity in the start function without accessing global state.

# Changes

Main changes
* Do not initialize the store in canisters.store.ts.
* Initialize the canistersStore where it's used. In the Canisters and CanisterDetail pages as well as in the service.
* Use the start function of the stores to load the data with query and update.
* Remove the call to `listCanisters` on mounting of Canisters and CanisterDetail page.

Minor changes:
* Add a new `identity` parameter to `queryAndUpdate`.

# Tests

This is just a draft PR to discuss a new pattern.

### DIFF
--- a/frontend/src/lib/derived/debug.derived.ts
+++ b/frontend/src/lib/derived/debug.derived.ts
@@ -1,7 +1,6 @@
 import type { Transaction } from "$lib/canisters/nns-dapp/nns-dapp.types";
 import { snsProjectsStore } from "$lib/derived/sns/sns-projects.derived";
 import { accountsStore } from "$lib/stores/accounts.store";
-import { canistersStore } from "$lib/stores/canisters.store";
 import { icrcTransactionsStore } from "$lib/stores/icrc-transactions.store";
 import { knownNeuronsStore } from "$lib/stores/known-neurons.store";
 import { neuronsStore } from "$lib/stores/neurons.store";
@@ -97,12 +96,12 @@ export const debugSnsProposalStore = (
 export const initDebugStore = () =>
   derived(
     [
+      // TODO: Add canisters store
       // TODO (L2-611): anonymize wallet id and neuron ids
       busyStore,
       accountsStore,
       neuronsStore,
       knownNeuronsStore,
-      canistersStore,
       proposalsStore,
       proposalsFiltersStore,
       votingNeuronSelectStore,
@@ -129,7 +128,6 @@ export const initDebugStore = () =>
       $accountsStore,
       $neuronsStore,
       $knownNeuronsStore,
-      $canistersStore,
       $proposalsStore,
       $proposalsFiltersStore,
       $votingNeuronSelectStore,
@@ -155,7 +153,6 @@ export const initDebugStore = () =>
       accounts: $accountsStore,
       neurons: $neuronsStore,
       knownNeurons: $knownNeuronsStore,
-      canisters: $canistersStore,
       proposals: $proposalsStore,
       proposalsFilters: $proposalsFiltersStore,
       votingNeuronSelect: $votingNeuronSelectStore,
--- a/frontend/src/lib/pages/CanisterDetail.svelte
+++ b/frontend/src/lib/pages/CanisterDetail.svelte
@@ -8,7 +8,10 @@
     listCanisters,
   } from "$lib/services/canisters.services";
   import { i18n } from "$lib/stores/i18n";
-  import { canistersStore } from "$lib/stores/canisters.store";
+  import {
+    initCanistersStore,
+    type CanistersStore,
+  } from "$lib/stores/canisters.store";
   import { replacePlaceholders, translate } from "$lib/utils/i18n.utils";
   import { SkeletonText, busy, Island } from "@dfinity/gix-components";
   import SkeletonCard from "$lib/components/ui/SkeletonCard.svelte";
@@ -35,12 +38,17 @@
   import CanisterDetailModals from "$lib/modals/canisters/CanisterDetailModals.svelte";
   import { emit } from "$lib/utils/events.utils";
   import type { CanisterDetailModal } from "$lib/types/canister-detail.modal";
+  import { authStore } from "$lib/stores/auth.store";
 
   // BEGIN: loading and navigation
 
   // TODO: checking if ready is similar to what's done in <ProposalDetail /> for the neurons.
   // Therefore we can probably refactor this to generic function.
 
+  let canistersStore: undefined | CanistersStore;
+  $: canistersStore = initCanistersStore($authStore.identity);
+
+  // TODO: Remove this part
   const canistersStoreReady = (): boolean => {
     // We consider the canisters store as ready if it has been initialized once.
     if (canistersReady) {
@@ -48,7 +56,7 @@
     }
 
     // At the moment we load the stores with query only.
-    return $canistersStore.canisters !== undefined;
+    return $canistersStore?.canisters !== undefined;
   };
 
   let canistersReady = false;
--- a/frontend/src/lib/pages/CanisterDetail.svelte
+++ b/frontend/src/lib/pages/CanisterDetail.svelte
@@ -3,10 +3,7 @@
   import type { Principal } from "@dfinity/principal";
   import type { CanisterDetails as CanisterInfo } from "$lib/canisters/nns-dapp/nns-dapp.types";
   import { AppPath } from "$lib/constants/routes.constants";
-  import {
-    getCanisterDetails,
-    listCanisters,
-  } from "$lib/services/canisters.services";
+  import { getCanisterDetails } from "$lib/services/canisters.services";
   import { i18n } from "$lib/stores/i18n";
   import {
     initCanistersStore,
@@ -39,6 +36,7 @@
   import { emit } from "$lib/utils/events.utils";
   import type { CanisterDetailModal } from "$lib/types/canister-detail.modal";
   import { authStore } from "$lib/stores/auth.store";
+  import { nonNullish } from "@dfinity/utils";
 
   // BEGIN: loading and navigation
 
@@ -47,26 +45,6 @@
 
   let canistersStore: undefined | CanistersStore;
   $: canistersStore = initCanistersStore($authStore.identity);
-
-  // TODO: Remove this part
-  const canistersStoreReady = (): boolean => {
-    // We consider the canisters store as ready if it has been initialized once.
-    if (canistersReady) {
-      return true;
-    }
-
-    // At the moment we load the stores with query only.
-    return $canistersStore?.canisters !== undefined;
-  };
-
-  let canistersReady = false;
-  $: $canistersStore, (canistersReady = canistersStoreReady());
-
-  onMount(async () => {
-    if (!canistersStoreReady()) {
-      await listCanisters({ clearBeforeQuery: false });
-    }
-  });
 
   const goBack = (): Promise<void> => goto(AppPath.Canisters);
 
@@ -132,8 +110,8 @@
 
   $: canisterId,
     selectedCanister,
-    canistersReady,
     (async () => {
+      const canistersReady = nonNullish($canistersStore?.canisters);
       // When detaching, this is also executed but there is no `canisterId`.
       if (!canistersReady || canisterId === undefined) {
         return;
--- a/frontend/src/lib/pages/CanisterDetail.svelte
+++ b/frontend/src/lib/pages/CanisterDetail.svelte
@@ -1,12 +1,12 @@
 <script lang="ts">
-  import { onMount, setContext } from "svelte";
+  import { setContext } from "svelte";
   import type { Principal } from "@dfinity/principal";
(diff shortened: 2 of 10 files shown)


lmuntaner:
@dskloetd @mstrasinskis @peterpeterparker please let me know what you think

frontend/src/lib/derived/debug.derived.ts
```diff
@@ -97,12 +96,12 @@ export const debugSnsProposalStore = (
 export const initDebugStore = () =>
```
dskloetd, line 96: I guess there would need to be a debug store per identity as well?

frontend/src/lib/pages/CanisterDetail.svelte
```diff
@@ -35,30 +35,16 @@
...
-  let canistersReady = false;
-  $: $canistersStore, (canistersReady = canistersStoreReady());
-
-  onMount(async () => {
-    if (!canistersStoreReady()) {
-      await listCanisters({ clearBeforeQuery: false });
-    }
-  });
+  let canistersStore: undefined | CanistersStore;
+  $: canistersStore = initCanistersStore($authStore.identity);
```
dskloetd, line 47: The name `init...` sounds to me as if it's (always) creating a new one, which of course wouldn't be good.
I would suggest something like `getOrCreate...`.

frontend/src/lib/services/utils.services.ts
```diff
@@ -38,27 +38,30 @@ export const queryAndUpdate = async <R, E>({
...
-  const identity: Identity =
-    identityType === "anonymous"
+  const requestIdentity: Identity =
+    identity ??
+    (identityType === "anonymous"
       ? getAnonymousIdentity()
       : identityType === "current"
       ? getCurrentIdentity()
-      : await getAuthenticatedIdentity();
+      : await getAuthenticatedIdentity());
```
dskloetd, line 61: If there can only be 1 identity, and `queryAndUpdate` knows how to get it, why do we need to pass the identity to create the store? Is it just to make sure we only create the store once we know the identity and not about passing the actual value? Or am I missing something.

And would it be possible for the store to listen on the authStore and (re)load itself once the identity is known?

frontend/src/lib/stores/canisters.store.ts
```diff
@@ -1,29 +1,82 @@
...
+  if (stores[storeKey]) {
+    return stores[storeKey];
+  }
+
+  const { subscribe, set } = writable<CanistersStoreData>(
+    {
+      canisters: undefined,
+      certified: undefined,
+    },
+    loadCanistersFactory(identity)
```
dskloetd, line 68: If we anyway need to call `initCanistersStore` whenever we want to use the store, and we can't create the store ahead of time, maybe using a "start" function is overkill? We could just load the store when `initCanistersStore` is called for the first time (when `!stores[storeKey]`), right?

frontend/src/lib/pages/CanisterDetail.svelte
```diff
@@ -35,30 +35,16 @@
...
-  };
-
-  let canistersReady = false;
-  $: $canistersStore, (canistersReady = canistersStoreReady());
-
-  onMount(async () => {
-    if (!canistersStoreReady()) {
-      await listCanisters({ clearBeforeQuery: false });
-    }
-  });
```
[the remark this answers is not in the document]
lmuntaner: Because of the new pattern. There is no need to call `listCanisters` anymore.

frontend/src/lib/services/canisters.services.ts
```diff
@@ -27,15 +27,21 @@ import {
...
+import { getAuthenticatedIdentity, getCurrentIdentity } from "./auth.services";
 import { queryAndUpdate } from "./utils.services";
 
 export const listCanisters = async ({
   clearBeforeQuery,
 }: {
   clearBeforeQuery?: boolean;
 }) => {
+  const identity = getCurrentIdentity();
+  const canistersStore = initCanistersStore(identity);
```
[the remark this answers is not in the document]
lmuntaner: Good point, not really. Better to pass the identity as well.

frontend/src/lib/stores/canisters.store.ts
lmuntaner: Yes, we could. Why do you consider that an overkill?

I prefer to use the way the tool (Svelte) suggests, rather than doing it our way.

frontend/src/lib/stores/canisters.store.ts
```diff
@@ -1,29 +1,82 @@
...
   certified: boolean | undefined;
 }
 
+export interface CanistersStore extends Readable<CanistersStoreData> {
+  setCanisters(data: CanistersStoreData): void;
+}
+
+const stores: Record<string, CanistersStore> = {};
+
+const loadCanistersFactory =
```
peterpeterparker, line 21: Does this duplicates a service? 

We are loading the canisters but, are also reloading the list when a user add a new canister. How do we avoid the duplication of code?

Or did I miss something?

frontend/src/lib/stores/canisters.store.ts
```diff
@@ -1,29 +1,82 @@
...
  */
-const initCanistersStore = () => {
-  const { subscribe, set } = writable<CanistersStore>({
-    canisters: undefined,
-    certified: undefined,
-  });
+export const initCanistersStore = (
+  identity: Identity | undefined | null
+): CanistersStore | undefined => {
+  if (isNullish(identity)) {
```
peterpeterparker, line 55: Is this safe enough or isn't this weaker that current implementation? 

Maybe alright for canisters but, generally speaking in current design a service shall not be called if the appropriate identity is not in used. The `getAuthenticatedIdentity` notably validate this and calls a logout if the criteria are not met.

frontend/src/lib/stores/canisters.store.ts
```diff
@@ -1,29 +1,82 @@
...
+    return;
+  }
+  const storeKey = identity.getPrincipal().toText();
+  if (stores[storeKey]) {
```
peterpeterparker, line 59: 1. in which case for the canisters we would have multiple `storeKey`?

2. since we do no handle promises and the fact that we can potentially use `initCanistersStore` multiple times in the code base, could there be a race condition where the effective query+update calls will be duplicated because the stores might not be set every time the init is called ?

frontend/src/lib/stores/canisters.store.ts
dskloetd: Overkill is maybe not the right word. But passing a function that will then be called by the framework is less obvious than explicitly calling the function. If there's no other difference, I'd rather be explicit. But there is another difference so let's focus on that.

The purpose of the start function is that it's called every time the number of subscribers goes from 0 to 1. So the better question is if we actually want that. If you navigate to the Canisters tab for the first time, the start function is called and the store is loaded. If you then navigate away, we unsubscribe from the store, but the contents of the store is still valid, right? If you then navigate to the Canisters tab again, do we actually want to load the store again or rather rely on the previously existing content? With the start function it will load the store again.

frontend/src/lib/stores/canisters.store.ts
dskloetd: Good point! I missed this but this is the discussion we had earlier about whether the service layer should depend on the stores layer or the stores layer on the service layer or if we should have a third layer or merge the 2 layers etc.

One possibility is to pass the load function from the component (which is allowed to depend on both layers) to the store initialization but it's also kind of ugly. Having store initialization in a separate layer from store definition might make more sense.

mstrasinskis: If we reset the store after canister creation this will trigger the date load. The only issue is that it can't be so good reflected in UI as it is now.

lmuntaner: Good question.

Yes, it duplicates a service at the moment. When a new canister is created, we should reload the page.

Yet, I'm not so sure it should be duplicated. After we create a canister, we now use `listCanisters` to reload, but that doesn't need to be the case in the future. We could have an endpoint to get one canister and use that for example.

Or the error handling might be different. We might not want to reset the store the an error, rather just show a toast that the new canister was created but there was a problem getting it.

I think that loading and reloading might share some code, but they are not the same thing. Even they might share code at the moment, conceptually, they are different things and might behave different in the future. Even though the behave similar now.

dskloetd: I disagree. Even if the loading won't be identical, it's conceptually the same and so it should live in the same place and use as much as possible shared code with only the difference being in their own functions.

frontend/src/lib/stores/canisters.store.ts
lmuntaner: Good point. I'd like to have both. Rely on the previous data while fetching the new data in case something changed.

Therefore, the default behavior looks good to me.

I see that if we wanted to trigger the load only ONCE, we shouldn't use the "start" function.

The pattern allows for both use cases though. Once per store, or every time subscribers go from 0 to 1.

frontend/src/lib/stores/canisters.store.ts
peterpeterparker: > We could create a derived store from the authStore and use that one instead of authStore how I used here. The stores could also raise an error if they need the identity to not be anonymous.

Ummm don't think it would be clean to perform such an assertion and logout and window reload within a derived store.

frontend/src/lib/stores/canisters.store.ts
peterpeterparker: 1. we perform a window reload on logout so even if UX wise the user can switch more easily between identity, keeping track of this store is useless since it will be reloaded. even if there would not be a window reload, not sure it would be correct security wise to keep private data in memory after a logout.

2. `initX...` is synchronous but isn't `loadCanistersFactory` that set the initial values in store asynchronous? so isn't possible that the first synchronous code goes through without finishing the load and the second init also kicks in and start the load as well etc.?

lmuntaner: 1. We'll need to ask for a security review, I agree.
2. The load (from the start function) is called when the subscribers to the store go from 0 to 1. Therefore, for a second call, the first subscriber will need to have unsubscribed. Therefore, it's not a race condition of who calls first `initXxxx`. If the first synchronous code goes through, the store is created and the start function passed. Svelte ensures that the start function is called only when subscribers go from 0 to 1.

lmuntaner: Let me add some tests to confirm that.

peterpeterparker: 1. I mean that even does not need a security review. When you logout in any app, do you expect your data to still be saved in the browser?
2. > Svelte ensures that the start function is called only when subscribers go from 0 to 1
Cool

lmuntaner: 1. Yes, logging out makes sense. I prefer that. But switching accounts maybe not. Otherwise users won't be able to have two tabs open with different users. I do that with my gmail accounts for example.

lmuntaner: II doesn't. Agent-js, why wouldn't? We can create a new agent with a new identity.

The other account would be the HW. Not needed to be supported for II. Yet, this is just a mock that Mary is working on. Just saying that this would already help us in case we go into that direction.

peterpeterparker: Isn't agent-js saving per default one identity per domain in idb?
I guess it would be possible to overcome this by providing a custom storage interface thinking at it now

frontend/src/tests/lib/stores/canisters.store.spec.ts
lmuntaner, line 1: @peterpeterparker @dskloetd @mstrasinskis I added the testing of the store.

Let me know what you think.

dskloetd: Is this PR still DO NOT MERGE just for discussing or are you turning it into something you want to merge?

lmuntaner: My plan:

* Add the comments and open topics in the document.
* Have another call with the team to discuss and make a decision.
* Close this PR.
* Make smaller PRs with the decisions of the meeting.

Is that ok?

dskloetd: That's fine but in that case I'm not sure why you would add tests to this PR.
The decisions wouldn't really be affected by the tests in any way, right?
And it's much easier to see that a test tests the right thing in a smaller PR.

lmuntaner: @peterpeterparker had some reservations regarding race conditions. I wanted to make sure that was ok.

But feel free to skip the tests if you have no reservations on the functionality.

lmuntaner:
Closing, this was just a test. We have already implemented this pattern in some places.

PR #2731, opened 2023-06-21
Collect all Icrc transactions for web workers

peterpeterparker:
# Motivation

One of the new web workers we are developing in #2639 we periodically collect the new potential transactions that were executed for a list of accounts.

This service provide such feature.

# Changes

- new service `icrc-transactions.worker-services`
- add few corresponding types including some for the upcoming post-message of the related PR


frontend/src/tests/lib/worker-services/icrc-transactions.worker-services.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+import { mockIdentity } from "$tests/mocks/auth.store.mock";
+import { mockCanisterId } from "$tests/mocks/canisters.mock";
+import {
+  mockSnsMainAccount,
+  mockSnsSubAccount,
+} from "$tests/mocks/sns-accounts.mock";
+import { IcrcIndexCanister, type IcrcTransaction } from "@dfinity/ledger";
+import mock from "jest-mock-extended/lib/Mock";
+
+describe("transactions.worker-services", () => {
```
lmuntaner, line 14: Could you add a `describe` for each function?

That would help us identify faster which test is for which function.

Thanks!

frontend/src/tests/lib/worker-services/icrc-transactions.worker-services.spec.ts
```diff
@@ -0,0 +1,269 @@
...
+  const indexCanisterMock = mock<IcrcIndexCanister>();
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    jest
```
lmuntaner, line 20: I believe that we said that we try to mock the api layer. This is even one layer below that.

This way we can then use and create fakers to manage how to api layer works and can even stop and resume return values to do expectations in between.

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
```diff
@@ -0,0 +1,182 @@
...
+      });
+
+      return {
+        transactions: transactions.reduce((acc, value) => {
+          // Suppress duplicate transactions to provide the results
+          const alreadyExist = (): boolean =>
+            acc.find(
+              (transaction) =>
+                value.id === transaction.id &&
+                JSON.stringify(transaction, jsonReplacer) ===
```
lmuntaner, line 69: Why is the JSON comparision necessary and we can't rely on the `id`?

Maybe it's worth a comment. Maybe even a helper to understand this line better? `isSameTransaction`

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
```diff
@@ -0,0 +1,182 @@
...
+                  JSON.stringify(value, jsonReplacer)
+            ) !== undefined;
+
+          return [...acc, ...(alreadyExist() ? [] : [value])];
```
lmuntaner, line 73: If you pass the values to `alreadyExist` you can put it as an util and easily test it because it would be a pure function.

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
```diff
@@ -0,0 +1,182 @@
...
+    state,
+  });
+
+  // We compare IDs because we effectively want to sort and find the oldest transaction ID
+  const oldestTxId: IcrcTxId | undefined = [...transactions].sort(
+    ({ id: idA }, { id: idB }) => Number(idA - idB)
+  )[0]?.id;
+
+  // Did we fetch all new transactions or there were more transactions than the batch size (DEFAULT_ICRC_TRANSACTION_PAGE_LIMIT) since last time the worker fetched the transactions
+  const fetchMore = (): boolean => {
```
lmuntaner, line 112: Do we want this to be an util outsie this function which is already big?

```
fetchMore({ mostRecentTxId, oldestTxId })
```

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
```diff
@@ -0,0 +1,182 @@
...
+    const stateMostRecentTxId = state?.mostRecentTxId;
+    return (
+      nonNullish(stateMostRecentTxId) &&
+      nonNullish(oldestTxId) &&
+      oldestTxId > stateMostRecentTxId
+    );
+  };
+
+  // Two transactions can have the same Id - e.g. a transaction from/to same account.
```
lmuntaner, line 121: This comment would be useful above also.

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
```diff
@@ -0,0 +1,182 @@
...
+  // That is why we fetch the next batch of transactions starting from the same Id and not Id - 1n because otherwise there would be a chance that we might miss one.
+  // Note: when "start" is provided, getIcrcTransactions search from "start" and returns "start" included in the results.
+  const nextTxId = (oldestTxId: bigint): bigint => oldestTxId;
```
lmuntaner, line 124: Why a function and not a variable?

```
// COMMENT
const nextTxId = oldestTxId;
```

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
```diff
@@ -0,0 +1,182 @@
...
+    identity,
+    indexCanisterId,
+    accountIdentifier,
+    start,
+    fetchRootKey,
+    host,
+    state,
+  });
+
+  // We compare IDs because we effectively want to sort and find the oldest transaction ID
```
lmuntaner, line 106: Could you add in the comment why we want to find the oldest transaction id?

It's because there might be a gap betweem the transactions we already have and the oldest we just received.

Maybe something like:

```
/**
* New transactions [100, 99, 98]
* Current transactions [95, 94, 93]
* Therefore, we still need to get between 95 and 98
*/
```

lmuntaner (commented):
Thanks! A few comments. It's tricky to follow, so maybe extracting some pure functions and adding some comments will make the functions more readable.

frontend/src/tests/lib/worker-services/icrc-transactions.worker-services.spec.ts
peterpeterparker: I mean there is just one function and the text then describe the case so not sure what you mean?

frontend/src/tests/lib/worker-services/icrc-transactions.worker-services.spec.ts
peterpeterparker: Putting together this all test was a true pain in the ass so I would be happy to spare this in this PR as I do not see the value it would add.

But let me know if there is a concrete value here?

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
peterpeterparker: Yep. Solved by adding also another comment, thanks for the input.

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
peterpeterparker: Good catch. I guess I was expecting to have to write a function but ultimately not had too. I removed the function and moved the comment.

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
peterpeterparker: Sure done

frontend/src/tests/lib/worker-services/icrc-transactions.worker-services.spec.ts
lmuntaner: Oh, I didn't realize only one was exported from the services.

Ok, now I understand it. It's ok.

frontend/src/tests/lib/worker-services/icrc-transactions.worker-services.spec.ts
lmuntaner: Main value I can see short term is consistency.

But it's not that only this test is not inconsistent. We have many inconsistent tests. I guess it's ok to not change it in this PR.

But it shouldn't be a big change. It's ok.

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
lmuntaner: Ok, we don't have a generic rule, so it's also up to you.

lmuntaner (approved):
Thanks!!

frontend/src/lib/worker-services/icrc-transactions.worker-services.ts
peterpeterparker: Name was a bit confusing, I renamed to `fetchMoreTransactions`.

That plus the existing comment should hopefully makes it to understand what it does.

Thanks for the review.

PR #2733, opened 2023-06-21
Animate new transaction being pushed in the list

peterpeterparker:
# Motivation

In #2639 we will fetch the new transactions in the background web workers, that's why Artem suggested to animated the new transactions that are added to the list with a visual effect to make the user aware of the update.

I implemented it in a global way, not just for that particular process because it generally looks good in my opinion.

# Note

We know there are some issue with using Svelte animation and the routing. I deploy a test version on [https://oxfmp-vaaaa-aaaaa-aaclq-cai.nnsdapp.dfinity.network](https://oxfmp-vaaaa-aaaaa-aaclq-cai.nnsdapp.dfinity.network) and did not noticed any such issue.

# Changes

- use `animate:flip` to make the list move when a new transaction card is added
- use `transition:fade` to be fancy

# Screenshots

demo


mstrasinskis:
Are you gonna add a line to the [changelog](https://github.com/dfinity/nns-dapp/blob/main/CHANGELOG-Nns-Dapp.md)?

mstrasinskis (approved):
LGTM. Nice improvement! Tested on iOS & Android devices.

PR #2736, opened 2023-06-22
E2e nns voting

mstrasinskis:
# Motivation

e2e test for proposal voting.

# Changes

- extend neurons spec to do a basic voting
- add all required POs

# Screenshot

https://github.com/dfinity/nns-dapp/assets/98811342/fc506e40-d177-41db-ba1a-2189adbb3d74

### DIFF
--- a/frontend/src/lib/components/proposal-detail/NnsProposal.svelte
+++ b/frontend/src/lib/components/proposal-detail/NnsProposal.svelte
@@ -13,6 +13,7 @@
   import NnsProposalProposerPayloadEntry from "./NnsProposalProposerPayloadEntry.svelte";
   import { filteredProposals } from "$lib/derived/proposals.derived";
   import { navigateToProposal } from "$lib/utils/proposals.utils";
+  import TestIdWrapper from "$lib/components/common/TestIdWrapper.svelte";
 
   const { store } = getContext<SelectedProposalContext>(
     SELECTED_PROPOSAL_CONTEXT_KEY
@@ -22,40 +23,42 @@
   $: proposalIds = $filteredProposals.proposals?.map(({ id }) => id as bigint);
 </script>
 
-{#if $store?.proposal?.id !== undefined}
-  <ProposalNavigation
-    currentProposalId={$store.proposal.id}
-    {proposalIds}
-    selectProposal={navigateToProposal}
-  />
+<TestIdWrapper testId="nns-proposal-component">
+  {#if $store?.proposal?.id !== undefined}
+    <ProposalNavigation
+      currentProposalId={$store.proposal.id}
+      {proposalIds}
+      selectProposal={navigateToProposal}
+    />
 
-  <div class="content-grid" data-tid="proposal-details-grid">
-    <div class="content-a content-cell-island">
-      <ProposalSystemInfoSection proposalInfo={$store.proposal} />
-    </div>
-    <div class="content-b expand-content-b">
-      <ProposalVotingSection proposalInfo={$store.proposal} />
-    </div>
-    <div class="content-c proposal-data-section">
-      <NnsProposalSummarySection proposalInfo={$store.proposal} />
+    <div class="content-grid" data-tid="proposal-details-grid">
+      <div class="content-a content-cell-island">
+        <ProposalSystemInfoSection proposalInfo={$store.proposal} />
+      </div>
+      <div class="content-b expand-content-b">
+        <ProposalVotingSection proposalInfo={$store.proposal} />
+      </div>
+      <div class="content-c proposal-data-section">
+        <NnsProposalSummarySection proposalInfo={$store.proposal} />
 
-      <NnsProposalProposerActionsEntry proposal={$store.proposal.proposal} />
+        <NnsProposalProposerActionsEntry proposal={$store.proposal.proposal} />
 
-      <NnsProposalProposerPayloadEntry
-        proposal={$store.proposal.proposal}
-        proposalId={$store.proposalId}
-      />
-    </div>
-  </div>
-{:else}
-  <div class="content-grid" data-tid="proposal-details-grid">
-    <div class="content-a">
-      <div class="skeleton">
-        <SkeletonDetails />
+        <NnsProposalProposerPayloadEntry
+          proposal={$store.proposal.proposal}
+          proposalId={$store.proposalId}
+        />
       </div>
     </div>
-  </div>
-{/if}
+  {:else}
+    <div class="content-grid" data-tid="proposal-details-grid">
+      <div class="content-a">
+        <div class="skeleton">
+          <SkeletonDetails />
+        </div>
+      </div>
+    </div>
+  {/if}
+</TestIdWrapper>
 
 <style lang="scss">
   @use "@dfinity/gix-components/dist/styles/mixins/media";
--- a/frontend/src/tests/e2e/neurons.spec.ts
+++ b/frontend/src/tests/e2e/neurons.spec.ts
@@ -35,6 +35,7 @@ test("Test neuron management", async ({ page, context }) => {
 
   const idsAfterFirstNeuronCreation = await neuronIds(appPo);
   expect(idsAfterFirstNeuronCreation).toHaveLength(1);
+  const neuronAId = idsAfterFirstNeuronCreation[0];
 
   step("Stake neuron B (for dummy proposals creation)");
   await appPo
@@ -47,8 +48,8 @@ test("Test neuron management", async ({ page, context }) => {
   const idsAfterSecondNeuronCreation = await neuronIds(appPo);
 
   expect(idsAfterSecondNeuronCreation).toHaveLength(2);
+  expect(idsAfterSecondNeuronCreation.includes(neuronAId)).toBe(true);
 
-  const neuronAId = idsAfterFirstNeuronCreation[0];
   const neuronBId = idsAfterSecondNeuronCreation.find((id) => id !== neuronAId);
 
   step("Open neuron A details");
@@ -93,8 +94,6 @@ test("Test neuron management", async ({ page, context }) => {
   const initialAdoptVotingPower = await proposalDetails
     .getVotesResultPo()
     .getAdoptVotingPower();
-
-  expect(Number(initialAdoptVotingPower)).toBe(stake * 2);
 
   step("Vote for proposal");
   await proposalDetails.getVotingCardPo().voteYes();
--- a/frontend/src/tests/e2e/neurons.spec.ts
+++ b/frontend/src/tests/e2e/neurons.spec.ts
@@ -3,6 +3,9 @@ import { PlaywrightPageObjectElement } from "$tests/page-objects/playwright.page
 import { signInWithNewUser, step } from "$tests/utils/e2e.test-utils";
 import { expect, test } from "@playwright/test";
 
+const neuronIds = async (appPo: AppPo) =>
+  await appPo.getNeuronsPo().getNnsNeuronsPo().getNeuronIds();
+
 test("Test neuron management", async ({ page, context }) => {
   await page.goto("/");
   await expect(page).toHaveTitle("NNS Dapp");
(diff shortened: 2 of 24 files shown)


frontend/src/lib/components/proposal-detail/NnsProposal.svelte
```diff
@@ -22,40 +23,42 @@
   $: proposalIds = $filteredProposals.proposals?.map(({ id }) => id as bigint);
 </script>
 
-{#if $store?.proposal?.id !== undefined}
-  <ProposalNavigation
-    currentProposalId={$store.proposal.id}
-    {proposalIds}
-    selectProposal={navigateToProposal}
-  />
+<TestIdWrapper testId="nns-proposal-component">
```
mstrasinskis, line 26: Only the testId was added here

frontend/src/tests/page-objects/NnsNeuronDetail.page-object.ts
```diff
@@ -14,6 +14,11 @@ export class NnsNeuronDetailPo extends BasePageObject {
     return SkeletonCardPo.allUnder(this.root);
   }
 
+  async createDummyProposals(): Promise<void> {
+    await this.click("make-dummy-proposals-button");
+    await this.root.byTestId("spinner").waitForAbsent();
```
mstrasinskis, line 19: There already was a static id -> `spinner`

frontend/src/tests/page-objects/SnsProposalVotingSection.page-object.ts
```diff
@@ -16,7 +16,7 @@ export class SnsProposalVotingSectionPo extends BasePageObject {
     );
   }
 
-  async getVotingsResultsPo(): Promise<VotesResultPo> {
+  getVotingsResultsPo(): VotesResultPo {
```
mstrasinskis, line 19: A tiny refactoring sneaked in 🥷

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -18,15 +18,82 @@ test("Test neuron management", async ({ page, context }) => {
...
+  step("Open neuron B details");
+
+  await appPo.getNeuronsPo().getNnsNeuronsPo().waitForContentLoaded();
+  const ids = await appPo.getNeuronsPo().getNnsNeuronsPo().getNeuronIds();
+
+  if (ids.length === 0) {
+    throw new Error("No neuron cards found");
+  }
+
+  // goto first neuron since it should be the one we just created
```
dskloetd, line 48: Are neurons always listed in reverse order?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -18,15 +18,82 @@ test("Test neuron management", async ({ page, context }) => {
...
+  await appPo.goToNeuronDetails(ids[0]);
+
+  step("Create dummy proposal with neuron B");
+
+  // create dummy proposals
```
dskloetd, line 53: This comment is redundant. Just delete this line and the line above and the step command makes it clear what is going on here.

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -18,15 +18,82 @@ test("Test neuron management", async ({ page, context }) => {
...
+
+  await appPo.getProposalsPo().getNnsProposalListPo().waitForContentLoaded();
+
+  const proposalIds = await appPo
+    .getProposalsPo()
+    .getNnsProposalListPo()
+    .getProposalIds();
+
+  if (proposalIds.length === 0) {
+    throw new Error("No proposal cards found");
```
dskloetd, line 70: Why not `expect(proposalIds.length).toBeGreaterThan(0)`?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -18,15 +18,82 @@ test("Test neuron management", async ({ page, context }) => {
...
+
+  step("Open first proposal");
+  await (
+    await appPo.getProposalsPo().getNnsProposalListPo().getProposalCardPos()
+  )[0].click();
+
+  const proposalDetails = appPo.getProposalDetailPo().getNnsProposalPo();
+  await proposalDetails.waitForContentLoaded();
+
+  await page.waitForSelector("[data-tid='adopt']");
```
dskloetd, line 81: The point of page objects is to avoid things like this. Can we encapsulate this in a page object?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -18,15 +18,82 @@ test("Test neuron management", async ({ page, context }) => {
...
+  await proposalDetails.getVotingCardPo().voteYes();
+
+  step("Compare voting power before and after voting");
+
+  const changedAdoptVotingPower = await proposalDetails
+    .getVotesResultPo()
+    .getAdoptVotingPower();
+
+  expect(Number(changedAdoptVotingPower)).toBeGreaterThan(
+    Number(initialAdoptVotingPower)
```
dskloetd, line 97: Is the expected voting power deterministic? Can we expect a specific amount?

frontend/src/tests/page-objects/ProposalCard.page-object.ts
```diff
@@ -0,0 +1,21 @@
+import { BasePageObject } from "$tests/page-objects/base.page-object";
+import type { PageObjectElement } from "$tests/types/page-object.types";
+
+// visible proposal card
```
dskloetd, line 4: Are there also invisible proposal cards? Or what is the comment trying to clarify?

frontend/src/tests/page-objects/ProposalCard.page-object.ts
```diff
@@ -0,0 +1,21 @@
...
+      (el) => new ProposalCardPo(el)
+    );
+  }
+
+  static under(element: PageObjectElement): ProposalCardPo {
+    return new ProposalCardPo(element.byTestId(ProposalCardPo.TID));
+  }
+
+  getProposalId(): Promise<string> {
+    return this.root.byTestId("proposal-id").getText();
```
dskloetd, line 19: I think you can use
```
return this.getText("proposal-id");
```
instead

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: Yes, the `sortedNeuronStore` is used on `/neurons` page

frontend/src/tests/page-objects/ProposalCard.page-object.ts
mstrasinskis: Technically yes (https://github.com/dfinity/nns-dapp/blob/979618d82ca16d449e212b12adefe874c7f4becd/frontend/src/lib/components/proposals/ProposalCard.svelte#L19), but actually this additional info is not relevant for the test. I'll remove the comment.

frontend/src/tests/page-objects/ProposalCard.page-object.ts
mstrasinskis: Thanks, replaced.

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: Good catch. This was used for debugging only and not necessary anymore.

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: Thanks, was replaced.

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: Thanks, this comment was created before the steps were added.

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: Good point. The test was updated to test the increase with neuron voting power.

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -30,70 +33,73 @@ test("Test neuron management", async ({ page, context }) => {
     .getNnsNeuronsFooterPo()
     .stakeNeuron({ amount: stake });
 
+  const idsAfterFirstNeuronCreation = await neuronIds(appPo);
+  expect(idsAfterFirstNeuronCreation.length).toBe(1);
```
dskloetd, line 37: `expect(idsAfterFirstNeuronCreation).toHaveLength(1);`?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -30,70 +33,73 @@ test("Test neuron management", async ({ page, context }) => {
...
     .stakeNeuron({ amount: stake });
+  await appPo.getNeuronsPo().getNnsNeuronsPo().waitForContentLoaded();
 
-  step("Open neuron B details");
+  // get neurons
+  const idsAfterSecondNeuronCreation = await neuronIds(appPo);
 
-  await appPo.getNeuronsPo().getNnsNeuronsPo().waitForContentLoaded();
-  const ids = await appPo.getNeuronsPo().getNnsNeuronsPo().getNeuronIds();
+  expect(idsAfterSecondNeuronCreation.length).toBe(2);
```
dskloetd, line 49: expect(idsAfterSecondNeuronCreation).toHaveLength(2);?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -30,70 +33,73 @@ test("Test neuron management", async ({ page, context }) => {
...
 
-  if (ids.length === 0) {
-    throw new Error("No neuron cards found");
-  }
+  const neuronAId = idsAfterFirstNeuronCreation[0];
+  const neuronBId = idsAfterSecondNeuronCreation.find((id) => id !== neuronAId);
```
dskloetd, line 52: Maybe also first verify that `idsAfterSecondNeuronCreation` contains `neuronAId`?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -30,70 +33,73 @@ test("Test neuron management", async ({ page, context }) => {
...
 
-  // goto first neuron since it should be the one we just created
-  await appPo.goToNeuronDetails(ids[0]);
+  step("Open neuron A details");
+  await appPo.goToNeuronDetails(neuronAId);
 
-  step("Create dummy proposal with neuron B");
+  step("Get neuron A voting power");
+  const neuronAVotingPower = await appPo
```
dskloetd, line 58: What do we expect this to be and shall we add an explicit expectation for that?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -30,70 +33,73 @@ test("Test neuron management", async ({ page, context }) => {
...
-
   const proposalIds = await appPo
     .getProposalsPo()
     .getNnsProposalListPo()
     .getProposalIds();
 
-  if (proposalIds.length === 0) {
-    throw new Error("No proposal cards found");
-  }
+  expect(proposalIds.length).toBeGreaterThan(0);
```
dskloetd, line 82: I know I suggested this but now I'm wondering, do we know the exact number of proposals there should be at this point?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -30,70 +33,73 @@ test("Test neuron management", async ({ page, context }) => {
...
   step("Open first proposal");
   await (
     await appPo.getProposalsPo().getNnsProposalListPo().getProposalCardPos()
   )[0].click();
-
   const proposalDetails = appPo.getProposalDetailPo().getNnsProposalPo();
   await proposalDetails.waitForContentLoaded();
-
-  await page.waitForSelector("[data-tid='adopt']");
   const initialAdoptVotingPower = await proposalDetails
```
dskloetd, line 90: Can we add an expectation of what this value should be at this point in the test?

frontend/src/tests/e2e/neurons.spec.ts
```diff
@@ -30,70 +33,73 @@ test("Test neuron management", async ({ page, context }) => {
...
     .getVotesResultPo()
     .getAdoptVotingPower();
```
dskloetd, line 92: If a number is more useful, does it make sense to return a number from this function instead of casting it later?

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: Why not. Added.

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: I would avoid using exact number here (lazy loading). What we could do is navigate to proposals before dummy proposals creation and get the first (newest) proposal id. And after creation check that there are some proposals with the greater id value than before.

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: This initial voting power depends on the dummy proposal count. Initial neuron voting power = 15*2. After 1st dummy proposal creation it's (15-1)*2 etc (this will be the proposal initial adopt value). To calculate it we need to track the proposals before voting. Not sure these extra steps will bring a lot of value.

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: Done

dskloetd (approved):

frontend/src/tests/e2e/neurons.spec.ts
mstrasinskis: I'll skip this extra step to keep the test smaller/faster.

PR #2739, opened 2023-06-22
Put coreutils front of path on mac

bitdivine:
# Motivation
On M1 macs (and possibly Intel as well) coreutils tools may be installed but not front-of-path.  This can cause undefined behaviour, when ancient apple tools are invoked instead.

# Changes
* Check that tools such as date use the coreutils versions, and if not install coreutils (as before) and put the coreutils front of path.

# Tests
* None.  Can someone with an M1 mac test please?

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -66,7 +66,9 @@ mkdir -p "$USER_BIN"
 NEW_TERMINAL_NEEDED=false
 append_to_profile() {
   printf '\n%s\n' "$1" >>"$USER_PROFILE"
+  set +eu
   . "$USER_PROFILE"
+  set -eu
   NEW_TERMINAL_NEEDED=true
 }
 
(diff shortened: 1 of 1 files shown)


scripts/setup
```diff
@@ -66,7 +66,9 @@ mkdir -p "$USER_BIN"
 NEW_TERMINAL_NEEDED=false
 append_to_profile() {
   printf '\n%s\n' "$1" >>"$USER_PROFILE"
+  set +eu
```
dskloetd, line 69: I tested it. The new code worked but the script failed because `/etc/bashrc` references an unbound variable and `set -u` is inherited. I added this to work around that but maybe you have a better way?

dskloetd (approved):

scripts/setup
bitdivine: LGTM.  Thanks.

PR #2740, opened 2023-06-22
Cleanup: Sunset nns-dapp_local.wasm

bitdivine:
# Motivation
*  The nns-dapp wasm used to be different depending on which network it was deployed to.  So we had a `_local` build for use on localhost and CI.
* Then the wasm was the same for all networks, but we retained `_local` to avoid breaking workflows.
* The time has come to sunset _local.

# Changes
* Transition all remaining references to `nns-dapp_local.wasm` to `nns-dapp.wasm`

# Tests
* None.  Existing functionality should be unaffected.
* Third parties who haven't yet transitioned to the universal build will have to do so now.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,13 +23,11 @@ jobs:
         uses: ./.github/actions/build_nns_dapp
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
Thanks!

PR #2741, opened 2023-06-22
Fix: Screen Title in HW when staking two neurons

lmuntaner:
# Motivation

The user was getting "Send ICP" in the HW screen instead of "Stake Neuron" if they were staking two neurons in a row without refreshing the page.

Problem: The agent from `createAgent` is cache. Therefore it was using the identity from the first staking. Yet, the service to stake a neuron was getting a new identity and setting the flag to stake neuron in that identity.

# Changes

* Cache the ledger identity.

# Tests

* Add a test to the ledger service to check that the identity is cached.

### DIFF
--- a/frontend/src/tests/lib/services/ledger.services.spec.ts
+++ b/frontend/src/tests/lib/services/ledger.services.spec.ts
@@ -14,6 +14,7 @@ import {
   getLedgerIdentity,
   listNeuronsHardwareWallet,
   registerHardwareWallet,
+  resetIdentitiesCachedForTesting,
   showAddressAndPubKeyOnHardwareWallet,
 } from "$lib/services/ledger.services";
 import { authStore } from "$lib/stores/auth.store";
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/services/ledger.services.spec.ts
```diff
@@ -210,6 +212,18 @@ describe("ledger-services", () => {
...
+
+      expect(identity1).not.toBeNull();
+      expect(LedgerIdentity.create).toHaveBeenCalledTimes(1);
+
+      const identity2 = await getLedgerIdentity(mockLedgerIdentifier);
+
+      expect(identity2).toBe(identity1);
+      expect(LedgerIdentity.create).toHaveBeenCalledTimes(1);
+    });
+
```
dskloetd, line 226: Can you also test that if you pass a different identifier, you get a different identity and createLedgerIdentity is called again?

dskloetd (approved):

PR #2742, opened 2023-06-22
Use standard tool for getting canister URLs

bitdivine:
# Motivation
Storing network configuration in dfx.json is deprecated.  Removing the network config from dfx.json requires, among other things, using the standard logic for how to calculate URLs.

# Changes
- Update `dfx-canister-url` to the latest version.  This is a command that gets URLs using the current IC policies.
- Compute URLs in the nns-dapp config using `dfx-canister-url`.
- Update the dockerfile to include `dfx-canister-url` where needed.

# Tests
- CI should check that the parameters have not changed for mainnet, and work for localhost.
- I checked manually that the config is unchanged for `benchmarklarge`.  I used the following one-liner but it's probably not very easy to follow.  The steps are:
  - On the feature branch and on main, set canister IDs as if all the canisters had been deployed.
  - Compute the config in both branches
  - Compare the configs and verify that they are the same.
```
network=benchmarklarge && for canister_name in tvl nns-dapp sns_aggregator ckbtc_ledger ckbtc_minter ckbtc_index internet_identity ; do  ( canister_id="$(cd ../../ && dfx canister id --network mainnet "$canister_name" || echo rdmx6-jaaaa-aaaaa-aaadq-cai)" ; for dir in . ../.. ; do pushd "$dir" ; dfx-canister-set-id --canister_id "$canister_id" --canister_name "$canister_name" --network "$network" ; DFX_NETWORK="$network" ./config.sh && mv deployment-config.json "network_configs/$network.json" ; popd ; done ) ; done && diff ../../network_configs/$network.json network_configs/$network.json
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,6 +89,8 @@ RUN cargo binstall --no-confirm "ic-wasm@$(cat config/ic_wasm_version)" && comma
 FROM builder AS configurator
 SHELL ["bash", "-c"]
 COPY dfx.json config.sh canister_ids.jso[n] /build/
+COPY scripts/dfx-canister-url /build/scripts/dfx-canister-url
+COPY scripts/clap.bash /build/scripts/clap.bash
 COPY .df[x]/ /build/.dfx
 WORKDIR /build
 ARG DFX_NETWORK=mainnet
@@ -103,6 +105,8 @@ RUN didc encode "$(cat nns-dapp-arg-${DFX_NETWORK}.did)" | xxd -r -p >nns-dapp-a
 FROM builder AS mainnet_configurator
 SHELL ["bash", "-c"]
 COPY dfx.json config.sh /build/
+COPY scripts/dfx-canister-url /build/scripts/dfx-canister-url
+COPY scripts/clap.bash /build/scripts/clap.bash
 WORKDIR /build
 RUN mkdir -p frontend
 ENV DFX_NETWORK=mainnet
--- a/config.sh
+++ b/config.sh
@@ -38,30 +38,10 @@ first_not_null() {
   echo "null"
 }
 
-static_host() {
-  first_not_null \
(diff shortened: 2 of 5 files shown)


scripts/dfx-canister-url
bitdivine, line 1: `dfx-canister-url` is imported from `snsdemo`, where it has tests to confirm that it reflects the current URL logic used by the IC.

Dockerfile
bitdivine, line 1: Here we need to import `dfx-canister-url` so that it can be used by `config.sh`.

config.sh
bitdivine, line 1: Here we replace our custom logic with calls to the standard command.

dfx.json
bitdivine, line 1: The HOST entries can be removed wherever they correspond to the standard scheme.  This leaves us with less networking config in dfx.json.  We need to reach zero.

scripts/dfx-canister-url
```diff
@@ -23,6 +22,7 @@ print_help() {
 	{ "canisters": {
 	    "my_canister": {
 	      "url": { "ic": "https://my-custom-url.com" },
+	      "raw_url": { "ic": "https://my-custom-raw-url.com" },
```
mstrasinskis, line 25: Where is the raw_url being used?

mstrasinskis (approved):
Rubber stamp.

scripts/dfx-canister-url
bitdivine: Good point.  There was a discussion about raw URLs upstream, specifically about when raw URLs make sense.  E.g. should there be raw URLs for APIs?  Are there ever likely to be raw URLs for APIs?  We reached a consensus but the code hasn't been updated since that discussion.  I'll get that finished upstream.  It doesn't matter for this PR but it should be finished.

bitdivine: Thank you for the approval.  Closing this to merge but happy to discuss further if wanted.

PR #2745, opened 2023-06-23
GIX-1642: Hide unavailable buttons when vesting

lmuntaner:
# Motivation

SNS neurons for developers might have a vesting period. During this vesting period, the functionality of the neuron is limited.

UI should not show unavailable functionality. Therefore, the following buttons are hidden during the vesting period:
* Increase Dissolve Delay
* Start or Stop Dissolving
* Disburse
* Split

# Changes

* In `SnsNeuronInfoStake` use `isVesting` for `allowedToDissolve` and `allowedToDisburse`.
* In `SnsNeuronMetaInfoCard` use `isVesting` for `allowedToSplit`.

# Tests

* Extend `createMockSnsNeuron` to return neuron with vesting configuration.
* Extend `SnsNeuronMetaInfoCardPo` to check the Split button.
* Add test case for the hidden buttons during vesting in `SnsNeuronInfoStake`.
* Add test case for the hidden buttons during vesting in `SnsNeuronMetaInfoCard`.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronInfoStake.svelte
@@ -15,6 +15,7 @@
     hasPermissionToDisburse,
     hasPermissionToDissolve,
     isCommunityFund,
+    isVesting,
   } from "$lib/utils/sns-neuron.utils";
   import { authStore } from "$lib/stores/auth.store";
   import { NeuronState } from "@dfinity/nns";
@@ -39,7 +40,8 @@
     hasPermissionToDissolve({
       neuron,
       identity: $authStore.identity,
-    });
+    }) &&
+    !isVesting(neuron);
 
   let allowedToDisburse: boolean;
   $: allowedToDisburse =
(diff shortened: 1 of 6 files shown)


lmuntaner:
New UX:

This is how a vesting neuron looks like:

Screenshot 2023-06-23 at 08 26 17

lmuntaner:

dskloetd (approved):
I'm wondering if this isn't confusing. Wouldn't it be better to have the buttons disabled with a tooltip explaining why?

lmuntaner:
> I'm wondering if this isn't confusing. Wouldn't it be better to have the buttons disabled with a tooltip explaining why?

Good question, I also discussed this with @peterpeterparker 

The following is the current UX in similar situations.

Disabled buttons examples:

- Split button when the neuron's stake is less than twice the maximum.
- Spawn Neuron button when the maturity is not enough to spawn a new neuron.

Hidden buttons examples:

- Disburse button while neuron is not unlocked.
- Stop / Start dissolving when neuron is unlocked.

I thought that this situation with the vesting was more similar to the second one. It seems that our pattern is to hide buttons if functionality is not possible based on the "status" of the neuron.

PR #2746, opened 2023-06-23
Script to download wasm from CI

dskloetd:
# Motivation

We want to make the release a less manual process.
One of the steps in the release process is to download the WASM from CI.

# Changes

Add a script to download the nns-dapp WASM from CI.

# Tests

Ran the script, also with wrong commit and wrong filename to exercise the error code paths.

Added a test which checks that the downloaded file is a gzip file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -126,10 +126,6 @@ jobs:
         uses: actions/download-artifact@v3
         with:
           name: sns_aggregator_dev
-      - name: Test download CI wasm
-        run: |
-          MAIN_COMMIT="$(git ls-remote --refs https://github.com/dfinity/nns-dapp.git main | awk '{print $1}')"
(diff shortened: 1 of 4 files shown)


bitdivine:
Looks good, but can you include a test, please?  E.g. pull head of main.  Doesn't need to be in this PR.

It looks as if you plan to use this as part of the release process.  The release process for nns-dapp can also be run against a testnet or local.

bitdivine (approved):
Looks good, but can you include a test, please?  E.g. pull head of main.

It looks as if you plan to use this as part of the release process.  The release process for nns-dapp can also be run against

dskloetd:
I added `scripts/nns-dapp/download-ci-wasm.test` which tests that the downloaded file is a gzip file and added it to `.github/workflows/checks.yml`.
Please have another look.

PR #2751, opened 2023-06-26
Build Web Workers ESM

peterpeterparker:
# Motivation

Web Workers ES modules are required to build easily  #2639 and are now supported by all browsers.

# Changes

- active ESM for Web Workers
- drop explicit usage of agent-js CJS

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -109,6 +109,9 @@ const config: UserConfig = {
       },
     },
   },
+  worker: {
+    format: "es",
+  },
 };
 
 export default config;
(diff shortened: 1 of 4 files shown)


frontend/vite.config.ts
```diff
@@ -109,6 +109,9 @@ const config: UserConfig = {
       },
     },
   },
+  worker: {
+    format: "es",
```
lmuntaner, line 113: Is this related to the other changes?

lmuntaner (approved):
LGTM, just one question.

frontend/vite.config.ts
peterpeterparker: That's the Vite flag that enables ES modules for web workers.
https://vitejs.dev/config/worker-options.html#worker-format

lmuntaner: Thanks to this we don't need to cjs explicit import?

peterpeterparker: Not anymore following this PR.

lmuntaner: Ok, thanks for the explanation

PR #2754, opened 2023-06-26
Get to know Web Workers sync status in UI

peterpeterparker:
# Motivation

It will be useful in the future, notably for #2639, to display an information in the UI to the users about the status of the Web Workers synchronization. Notably displaying a small hint like a spinner when work is in progress.

This PR introduces a store that contains such information for upcoming workers of #2639 and also improve the timer utils to transmit the information

# Changes

- new `sync.store` and `sync.derived`
- add `postMessage` in timer utils to transmit the sync status to the UI

### DIFF
--- a/frontend/src/lib/derived/sync.derived.ts
+++ b/frontend/src/lib/derived/sync.derived.ts
@@ -1,0 +1,26 @@
+import {
+  syncStore,
+  type SyncStore,
+  type SyncStoreData,
+} from "$lib/stores/sync.store";
+import type { SyncState } from "$lib/types/sync";
+import { derived } from "svelte/store";
+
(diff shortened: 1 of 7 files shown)


frontend/src/lib/derived/sync.derived.ts
lmuntaner, line 1: I'm curious why this will be useful, hehe.

lmuntaner (approved):
LGTM!

frontend/src/lib/derived/sync.derived.ts
peterpeterparker: Absolutely 😉. We will display a single spinner indicator.

PR #2756, opened 2023-06-26
Balances worker

peterpeterparker:
# Motivation

A Web Worker that periodically checks for up-to-date ICRC balance of accounts.

# Notes

This introduces the worker but, not its usage.

# PRs

- [x] needs #2754

# Changes

- new `balances.worker`
- add `getIcrcAccountsBalances` utils / services for workers which execute `getIcrcBalance` for a list of accounts and for the worker
- related `initBalancesWorker` service which can be use to load and interact with the worker in a component


frontend/src/tests/lib/services/worker-balances.services.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+          msg: "nnsStartBalancesTimer";
+          data: PostMessageDataRequestBalances;
+        }) {
+          spyPostMessage(data);
+        }
+      };
+    });
+  });
+
+  it("should start worker with params", async () => {
```
lmuntaner, line 25: Coul you test also `stopBalancesTimer`

lmuntaner (approved):
LGTM, a minor comment

frontend/src/tests/lib/services/worker-balances.services.spec.ts
peterpeterparker: Definitely

PR #2759, opened 2023-06-26
Balances and transactions worker observer

peterpeterparker:
# Motivation

We are going to observe the balance and transactions of accounts in various place in the app using web workers. This is being implemented in #2639.

This PR introduces two handy components that will be use to ease the integration and communication with related workers.

# Changes

- new "observer" components for balance and transactions


lmuntaner (approved):
LGTM!

PR #2763, opened 2023-06-27
Chunking paths up-to-date

peterpeterparker:
# Motivation

I wanted to add the missing `workers-types` to chunking paths but, noticed that other path was missing and another did not exist anymore.

# Changes

- update chunking paths with actually packages sturcture


lmuntaner (approved):
Thanks!

PR #2764, opened 2023-06-27
Sync indicator

peterpeterparker:
# Motivation

Display an indicator in the toolbar that reflects the progression or error of web workers that perform synchronization of data. Will be used for transactions and balance first.

# Changes

- new component `AccountSyncIndicator`
- introduce component in header next to the menu account
- bump gix-cmp with new `IconSync`

# Screenshots

Capture d’écran 2023-06-27 à 09 19 50
Capture d’écran 2023-06-27 à 09 19 53
Capture d’écran 2023-06-27 à 09 19 58
Capture d’écran 2023-06-27 à 09 20 01


frontend/src/lib/i18n/en.json
```diff
@@ -927,5 +927,13 @@
...
     "your_session": "Your session expires in",
     "your_session_description": "All sessions last 30 minutes. After 30 minutes, you will need to authenticate again with Internet Identity."
+  },
+  "sync": {
+    "status_idle": "Synchronization is currently idle.",
+    "status_error": "An error occurred while synchronizing data.",
+    "status_in_progress": "Synchronization is in progress.",
+    "status_idle_detailed": "Synchronization is currently idle.",
+    "status_error_detailed": "An error occurred while synchronizing data. Please try reloading the page. If the problem persists, please contact support for further assistance.",
+    "status_in_progress_detailed": "Synchronization is in progress. NNS dapp is checking for the availability of new transactions."
```
lmuntaner, line 937: This might be many different things, right? Shouldn't the part that is specific to the transactions be dynamic?

frontend/src/lib/components/header/SyncIndicator.svelte
```diff
@@ -0,0 +1,88 @@
...
+        return $i18n.sync.status_idle;
+    }
+  };
+
+  const syncDescription = (): string => {
+    switch ($syncOverallStatusStore) {
+      case "error":
+        return $i18n.sync.status_error_detailed;
+      case "in_progress":
+        return $i18n.sync.status_in_progress_detailed;
```
lmuntaner, line 33: We will need more details when we apply this to other pages, right?

Any idea how we could do it?

frontend/src/lib/components/header/SyncIndicator.svelte
```diff
@@ -0,0 +1,88 @@
...
+      default:
+        return $i18n.sync.status_idle_detailed;
+    }
+  };
+
+  const syncIcon = (): typeof SvelteComponent | undefined => {
+    if ($syncOverallStatusStore === "in_progress") {
```
lmuntaner, line 40: Why not a switch here?

frontend/src/lib/components/header/SyncIndicator.svelte
```diff
@@ -0,0 +1,88 @@
...
+
+    if ($syncOverallStatusStore === "error") {
+      return IconError;
+    }
+
+    return undefined;
+  };
+
+  $: $syncOverallStatusStore,
+    (label = syncLabel()),
```
lmuntaner, line 52: How about passing the store as a parameter? This way we don't need to remember that `syncLabel` depends on the store and put it in front.

We can have a line per variables:

```
$: label = syncLabel($syncOverallStatusStore);
$: icon = syncIcon($syncOverallStatusStore);
$: description = syncDescription($syncOverallStatusStore);

lmuntaner (approved):
Thanks! LGTM, just some minor comments, but already approving.

frontend/src/lib/i18n/en.json
peterpeterparker: Originally it was suppose to be generic but, Mary by the review emitted the wish to have it specific. Because I do not want to over engineer this part at the moment, rather like to finally deliver the feature and improve the UI component in the future if there is ever the need, I decided to just update the translations.

That said, I have now renamed the component to `AccountSyncIndicator` that way it is more specific.

frontend/src/lib/components/header/SyncIndicator.svelte
peterpeterparker: Relates to https://github.com/dfinity/nns-dapp/pull/2764#discussion_r1243456592

frontend/src/lib/components/header/SyncIndicator.svelte
peterpeterparker: Good point, thanks. Switch added

frontend/src/lib/components/header/SyncIndicator.svelte
peterpeterparker: I like current pattern visually speaking but sure code style, so I don't mind changing this. Done.

PR #2767, opened 2023-06-27
Don't require a config field in the network

bitdivine:
# Motivation
The nns-dapp config script currently relies on a network config entry being present in dfx.json.  The network config in dfx.json is not always present for all networks, and is indeed deprecated so will be removed entirely.

# Changes
- Make per-network config optional.

# Tests
- The config script works exactly as before if there is a config entry, which there is for prod.
- For local deployments, CI should suffice.

# Todos


peterpeterparker (approved):
Rubber stamp

PR #2769, opened 2023-06-27
Refresh accounts balance periodically on Sns wallet

peterpeterparker:
# Motivation

On Sns wallet page, use the new workers and strategy developed in #2639 to periodically check the balance of the selected account and, in case of change, update the UI and store with the new balance value.

# Changes

- new `SnsWalletBalancesObserver` which uses the `IcrcBalancesObserver` to start, stop and interact with the related web worker
- wrap content of `SnsWallet` with this new component
- link Web Worker event with sync event to display possible error in the new sync indicator (forgot to provide this in one of my last PR)
- add a function to `sns-accounts.store` to update accounts
- bump gix-cmp with reviewed "Sync" icon

### DIFF
--- a/frontend/src/lib/components/accounts/SnsWalletBalancesObserver.svelte
+++ b/frontend/src/lib/components/accounts/SnsWalletBalancesObserver.svelte
@@ -1,0 +1,50 @@
+<script lang="ts">
+  import type { BalancesObserverData } from "$lib/types/icrc.observer";
+  import { nonNullish } from "@dfinity/utils";
+  import IcrcBalancesObserver from "$lib/components/accounts/IcrcBalancesObserver.svelte";
+  import type { BalancesCallback } from "$lib/services/worker-balances.services";
+  import { snsProjectAccountsStore } from "$lib/derived/sns/sns-project-accounts.derived";
+  import { snsAccountsStore } from "$lib/stores/sns-accounts.store";
+  import type { UniverseCanisterId } from "$lib/types/universe";
+  import type { Account } from "$lib/types/account";
+  import type { CanisterId } from "$lib/types/canister";
+
+  export let universeId: UniverseCanisterId;
+  export let ledgerCanisterId: CanisterId;
+  export let account: Account;
+
+  const callback: BalancesCallback = ({ balances }) => {
+    const accounts = balances
+      .map(({ balance, accountIdentifier }) => {
+        const selectedAccount = $snsProjectAccountsStore?.find(
+          ({ identifier }) => identifier === accountIdentifier
+        );
+
+        return nonNullish(selectedAccount)
+          ? {
+              ...selectedAccount,
+              balanceE8s: balance,
+            }
+          : undefined;
+      })
+      .filter(nonNullish);
+
+    snsAccountsStore.updateAccounts({
+      accounts,
+      rootCanisterId: universeId,
+      certified: true,
+    });
+  };
+
+  let data: BalancesObserverData;
+  $: data = {
+    account,
+    ledgerCanisterId: ledgerCanisterId.toText(),
+  };
+</script>
+
+{#if nonNullish(data)}
+  <IcrcBalancesObserver {data} {callback}>
+    <slot />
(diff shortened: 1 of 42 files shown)


frontend/src/lib/components/accounts/SnsWalletBalancesObserver.svelte
```diff
@@ -0,0 +1,50 @@
...
+  import { nonNullish } from "@dfinity/utils";
+  import IcrcBalancesObserver from "$lib/components/accounts/IcrcBalancesObserver.svelte";
+  import type { BalancesCallback } from "$lib/services/worker-balances.services";
+  import { snsProjectAccountsStore } from "$lib/derived/sns/sns-project-accounts.derived";
+  import { snsAccountsStore } from "$lib/stores/sns-accounts.store";
+  import type { UniverseCanisterId } from "$lib/types/universe";
+  import type { Account } from "$lib/types/account";
+  import type { CanisterId } from "$lib/types/canister";
+
+  export let universeId: UniverseCanisterId;
```
lmuntaner, line 12: I believe we call it `rootCanisterId` when it's about an SNS. I'd prefer to keep the same naming for consistency.

frontend/src/tests/lib/pages/SnsWallet.spec.ts
lmuntaner, line 1: Could you add a test that the SnsWallet starts a timer?

frontend/src/tests/lib/pages/SnsWallet.spec.ts
```diff
@@ -40,6 +40,32 @@ jest.mock("$lib/services/sns-transactions.services", () => {
   };
 });
 
+jest.mock("$lib/services/worker-transactions.services", () => ({
```
lmuntaner, line 43: Could you mock the `"$lib/workers/balances.worker?worker"` level instead?

Then we could test that a message is sent to the worker.

lmuntaner: Just like you do in the `SnsWalletBalancesObserver.spec.ts`

lmuntaner (commented):
A few comments, mainly on testing. Looks good!

frontend/src/lib/components/accounts/SnsWalletBalancesObserver.svelte
peterpeterparker: ok

frontend/src/tests/lib/pages/SnsWallet.spec.ts
peterpeterparker: The purpose of having implemented the `IcrcBalancesObserver` component is to defer all the work of the interaction with the worker in a single reusable component. Testing messages here would just be a duplication of code and test of this particular component. Therefore would not really adds any coverage.

But maybe I am missing something, what would be the value?

frontend/src/tests/lib/pages/SnsWallet.spec.ts
peterpeterparker: I have added a test that checks that indeed the worker is used.

frontend/src/tests/lib/pages/SnsWallet.spec.ts
lmuntaner: We're testing implementation details by mocking that level. The purpose of this component is to setup the worker.

I believe that using the service is an implementation.

lmuntaner: Ideally we could test that it calls the api every X seconds for example. But we can't use workers yet in jest and we need to mock them.

peterpeterparker: > The purpose of this component is to setup the worker.

No that's incorrect. The purpose of this component - `SnsWallet` - is to use a component - `IcrcBalancesObserver` - that isolates and takes care of any interaction with the worker.

Therefore so far do no see any value switching to moking the worker.

lmuntaner (approved):

PR #2771, opened 2023-06-28
Rename close listeners

peterpeterparker:
# Motivation

Listeners were probably copied a long time ago from Papyrs that's why they mention "sync...". 
We can rename those to be more generic.


mstrasinskis (approved):
Thanks!

PR #2773, opened 2023-06-28
feat: icrc-accounts.store update

peterpeterparker:
# Motivation

Add an `update` feature to `icrc-accounts` which can be use to update accounts when those are refreshed by a worker.


lmuntaner (approved):
Thanks!

PR #2775, opened 2023-06-28
Parse build run conslusion instead of using --status of gh

dskloetd:
# Motivation

We have a script that downloads the WASM from CI. It uses `gh` to get a successful run by using `--status success`. This was working before but at commit `8ac5bef5427b0e3b003794fd2545eb9fd92cee83` it does not find any runs. But if I remove the flag, I do see runs that are successful.

# Changes

Instead of using the `--status success` flag of `gh`. Parse the `.conclusion` field of the result to find the successful runs.

# Tests

This was found because `scripts/nns-dapp/download-ci-wasm.test` was failing so the test was working.
The test passed on this PR: https://github.com/dfinity/nns-dapp/actions/runs/5399373231/jobs/9806423017?pr=2775

# Todos


peterpeterparker (approved):
Thanks

PR #2781, opened 2023-06-28
Remove unused asset logo sns svg

peterpeterparker:
# Motivation

I don't think we use that asset anywhere anymore.


mstrasinskis (approved):
Rubber stamp. Thanks!

peterpeterparker:
Turns out it is used, good that I double checked a second time before merging it.

The way it is implemented is weak though.

@lmuntaner any particular reason why the `DEFAULT_SNS_LOGO` is a constant `string` and not imported?

currently is:

```
export const DEFAULT_SNS_LOGO = "/assets/sns-logo-default.svg";
```

following would be better:

```
import DEFAULT_SNS_LOGO from "$lib/assets/sns-logo-default.svg";
```

But maybe there is a limitation?

lmuntaner:
@peterpeterparker because the value we get from the SNS aggregator is also a string and I wanted to keep the same type.

Another idea would be to put it in the component that renders the logo, but because the logo might be used for other reasons, the default is set in the derived store.

peterpeterparker:
> because the value we get from the SNS aggregator is also a string and I wanted to keep the same type.

Well, I think that's excalty the outcome of an `import`. Vite copy the asset to the output folder and interpret the import as a string path to the resources So here's a PR for this: https://github.com/dfinity/nns-dapp/pull/2789

lmuntaner:
> > because the value we get from the SNS aggregator is also a string and I wanted to keep the same type.
> 
> Well, I think that's excalty the outcome of an `import`. Vite copy the asset to the output folder and interpret the import as a string path to the resources So here's a PR for this: #2789

That makes sense, thanks!

PR #2782, opened 2023-06-28
Refresh balances periodically on Sns accounts

peterpeterparker:
# Motivation

As for Sns wallet, we want to periodically refresh the accounts displayed on the Sns Accounts page. However, as no transactions are displayed, we only want to refresh the balances on that particular page.

# PRs

- Follow-up of #2639

# Changes

- rename `SnsWalletBalancesObserver` to `SnsBalancesObserver`
- extend property from a single account to an array of accounts
- use observer in `SnsAccounts` page

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcBalancesObserver.svelte
+++ b/frontend/src/lib/components/accounts/IcrcBalancesObserver.svelte
@@ -19,14 +19,11 @@
 
     worker = await initBalancesWorker();
 
-    const {
-      account: { identifier },
-      ledgerCanisterId,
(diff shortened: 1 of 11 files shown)


lmuntaner (approved):
Was this a bug? Using one account instead of multiple accounts?

peterpeterparker:
> Was this a bug? Using one account instead of multiple accounts?

Not a bug. More due to the fact that I first implemented in only in Sns Wallet where there is only one account.

The Web Workers was already designed for `accounts`, was really just a UI components thing.

PR #2788, opened 2023-06-29
Ignore sign-in "error" UserInterrupt

peterpeterparker:
# Motivation

Agent-js bubbles the error `UserInterrupt` when user interrupts the sign-in flow which per extension is displayed in a toast error in NNS dapp. Because this has not much value for most of the users, we are going to ignore these particular "error".

# Changes

- if `UserInterrupt` then no toast
- refactor duplicated code and solve TODO


lmuntaner (approved):
Thanks!

PR #2793, opened 2023-06-30
Address TODO: Use pop_first

bitdivine:
# Motivation
When the proposal payload cache was written, `pop_first()` existed but only in unstable rust.  Now it is available in stable, so we can close the TODO.

# Changes
* Simplify code with `pop_first()`

# Tests
* Existing tests should suffice.

# Todos


mstrasinskis (approved):
Rubber stamp. Thanks!

PR #2796, opened 2023-06-30
GIX-1656: Add renaming canister api and service

lmuntaner:
# Motivation

Users want to be able to add names to their canisters.

In this PR: I extend the nns-dapp canister to support the endpoint for renaming and create the api and service functions.

# Changes

* Extend nns dapp canister candid files and related.
* New method in nns-dapp canister class `renameCanister`.
* New canister api function `renameCanister`.
* New canister service `renameCanister`.

# Tests

* Test new nns-dapp canister method `renameCanister`.
* Test new api function `renameCanister`
* Test new service `renameCanister`

# Todos

Not yet any entry in the changelog because nothing is exposed to the user yet.

### DIFF
--- a/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
+++ b/frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
@@ -271,6 +271,39 @@ export class NNSDappCanister {
     throw new Error(`Error attaching canister ${JSON.stringify(response)}`);
   };
 
+  public renameCanister = async ({
+    name,
+    canisterId,
+  }: {
+    name: string;
+    canisterId: Principal;
+  }): Promise<void> => {
+    const response = await this.certifiedService.rename_canister({
+      name,
+      canister_id: canisterId,
+    });
+    if ("Ok" in response) {
+      return;
+    }
+    if ("NameAlreadyTaken" in response && response.NameAlreadyTaken === null) {
+      throw new CanisterNameAlreadyTakenError("error__canister.name_taken", {
+        $name: name,
+      });
+    }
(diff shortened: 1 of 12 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
```diff
@@ -271,6 +271,39 @@ export class NNSDappCanister {
...
+    canisterId: Principal;
+  }): Promise<void> => {
+    const response = await this.certifiedService.rename_canister({
+      name,
+      canister_id: canisterId,
+    });
+    if ("Ok" in response) {
+      return;
+    }
+    if ("NameAlreadyTaken" in response && response.NameAlreadyTaken === null) {
```
dskloetd, line 288: Why does `response.NameAlreadyTaken` need to be `null`? What else could it be and what would that mean?

lmuntaner: It can't be anything else at the moment. I can remove these checks according the the candid file, if they are present, they are always `null`.

dskloetd: Yeah, then I would remove these checks.
If you really want to make sure that they follow the spec, you'd also have to check that only exactly one of the variants is set and that no other fields are set, which also seems like overkill.

dskloetd (approved):

frontend/src/lib/canisters/nns-dapp/nns-dapp.canister.ts
lmuntaner: done

PR #2798, opened 2023-06-30
Set custom URL for II on the ic

bitdivine:
# Motivation
There is a currently a custom URL for II for mainnet deployments, but not for deployments to the app subnet.

When the app subnet deployment was using the mainnet config, this was not a problem.  Now that the app deployment uses its own config, we need correct values.

# Changes
- Move the custom URL to the canister config, as the network entries are deprecated.
- Add a custom URL for deployments to the ic.

# Tests
- Tested manually for the `ic` network.
- CI has a test to confirm that the mainnet config is unchanged.
- Making a full config test similar to that used to test the mainnet config requires more changes, so I would rather do this in a separate PR.

# Todos


mstrasinskis (approved):
Rubber stamp.

PR #2800, opened 2023-06-30
Consolidate tests

bitdivine:
# Motivation
`./config.sh` has a test command called, by convention, `./config.test`.  That includes a stability test.  A new stability test was created a couple of months ago that is much more maintainable and that has been running in parallel.  The original stability test can be discarded.

This is a priority now because more config tests need to be added, and the current somewhat messy situation makes this harder.

# Changes
* Remove the original stability test, calling the new stability test in its place.
* Remove one of the two github jobs.

# Tests
* CI runs the tests in this PR.

# Todos


mstrasinskis (approved):
Rubber stamp.

PR #2802, opened 2023-06-30
More details in merge neurons simulation

dskloetd:
# Motivation

We want to show the results or merging neurons before the user commits.
This PR adds more details in the `ConfirmNeuronsMerge` steps.
Here you can see what it looks like but the details are not important because this is behind a feature flag just to try it out and iterate on:

image

# Changes

1. Add new component `NnsNeuronDetailCard`
2. Use it instead of `NnsNeuronInfo` in `ConfirmNeuronsMerge`

# Tests

1. `frontend/src/tests/lib/modals/neurons/MergeNeuronsModal.spec.ts` is modified to work with the new card.
2. A new test is added for the new component.
3. A page object is added for the new component.

# Todos

### DIFF
--- a/frontend/src/tests/lib/components/neurons/NnsNeuronDetailCard.spec.ts
+++ b/frontend/src/tests/lib/components/neurons/NnsNeuronDetailCard.spec.ts
@@ -11,7 +11,6 @@ import { render } from "@testing-library/svelte";
 describe("NnsNeuronDetailCard", () => {
   const renderComponent = (neuron) => {
     const { container } = render(NnsNeuronDetailCard, { neuron });
-    //console.log('dskloetx container', container.innerHTML);
     return NnsNeuronDetailCardPo.under({
       element: new JestPageObjectElement(container),
     });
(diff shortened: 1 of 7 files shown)


frontend/src/tests/lib/components/neurons/NnsNeuronDetailCard.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    expect(await po.getMaturity()).toBe("8.00");
+  });
+
+  it("should render neuron staked maturiry", async () => {
+    const po = renderComponent({
+      ...mockNeuron,
+      fullNeuron: {
+        ...mockFullNeuron,
+        maturityE8sEquivalent: BigInt(300_000_000),
```
lmuntaner, line 86: Is this important for this test?

lmuntaner (approved):
LGTM! Just one minor comment

frontend/src/tests/lib/components/neurons/NnsNeuronDetailCard.spec.ts
dskloetd: I think it's helpful when you look at this test and the previous test that you see the same input and different outputs.
If I remove this line, I could accidentally replace `getStakedMaturity` on line 91 with `getMaturity`, and the test would still pass.

lmuntaner: Ok, got it, thanks for the explanation!

PR #2803, opened 2023-07-01
Remove load Sns transactions onMount in wallet

peterpeterparker:
# Motivation

We can spare loading the transactions in the Sns Wallet on the UI side because this operation is now executed in a Web Worker.

# Notes

Because of the [Web Worker](https://caniuse.com/?search=worker%20module) support, users using old version of Firefox on desktop would have to upgrade to upgrade their browser to access their transactions. On Firefox Android they would have to use another browser.

# Changes

- remove `loadTransactions` on mount


lmuntaner (approved):
One question, do we have a message or something for the unsupported browser? I belive we do, but I forgot.

peterpeterparker:
No we do not have message. I prepared the PR to think about it but, I think we can merge this in a bit (few weeks or more). Therefore the draft. It's good to have it in the future because performance wise it's better.

peterpeterparker:
Given how old this PR is, won't be proceeeded further.

PR #2804, opened 2023-07-01
GIX-1656: Rename canister modal

lmuntaner:
# Motivation

Users want to be able to add a name to their canisters.

In this PR: Users can add a name after the canister has been attached to the user.

# Changes

* New button RenameCanisterButton.
* New modal RenameCanisterModal.
* Render new button in CanisterDetail page.

# Tests

* Test that RenameCanisterButton opens modal.
* Test that modal calls api RenameCanisterModal.
* Test that user can rename from CanisterDetail.
* Add related page objects.

# Todos

### DIFF
--- a/frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
+++ b/frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
@@ -1,0 +1,46 @@
+<script lang="ts">
+  import { i18n } from "$lib/stores/i18n";
+  import { Modal, busy } from "@dfinity/gix-components";
+  import TextInputForm from "$lib/components/common/TextInputForm.svelte";
+  import {
+    CANISTER_DETAILS_CONTEXT_KEY,
+    type CanisterDetailsContext,
+  } from "$lib/types/canister-detail.context";
+  import { getContext } from "svelte";
+  import { isNullish } from "@dfinity/utils";
+
+  const { store }: CanisterDetailsContext = getContext<CanisterDetailsContext>(
+    CANISTER_DETAILS_CONTEXT_KEY
+  );
+
+  let currentName = $store.info?.name;
+
+  const rename = () => {
+    console.log("renaming", currentName);
+  };
+</script>
+
+<Modal on:nnsClose>
+  <svelte:fragment slot="title"
+    ><span>{$i18n.canister_detail.rename_canister}</span></svelte:fragment
+  >
+
+  <TextInputForm
+    on:nnsConfirmText={rename}
+    on:nnsClose
+    bind:text={currentName}
+    placeholderLabelKey="canister_detail.rename_canister_placeholder"
+    busy={$busy}
+    disabledConfirm={isNullish(currentName) ||
+      currentName?.length === 0 ||
+      $busy}
+  >
+    <svelte:fragment slot="label"
+      >{$i18n.canister_detail.rename_canister_title}</svelte:fragment
+    >
+    <svelte:fragment slot="cancel-text">{$i18n.core.cancel}</svelte:fragment>
+    <svelte:fragment slot="confirm-text"
+      >{$i18n.canister_detail.rename}</svelte:fragment
+    >
+  </TextInputForm>
+</Modal>
--- a/frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
+++ b/frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
@@ -6,8 +6,11 @@
     CANISTER_DETAILS_CONTEXT_KEY,
     type CanisterDetailsContext,
   } from "$lib/types/canister-detail.context";
-  import { getContext } from "svelte";
-  import { isNullish } from "@dfinity/utils";
+  import { createEventDispatcher, getContext } from "svelte";
+  import { isNullish, nonNullish } from "@dfinity/utils";
+  import { renameCanister } from "$lib/services/canisters.services";
+  import { toastsSuccess } from "$lib/stores/toasts.store";
+  import { startBusy, stopBusy } from "$lib/stores/busy.store";
 
   const { store }: CanisterDetailsContext = getContext<CanisterDetailsContext>(
     CANISTER_DETAILS_CONTEXT_KEY
@@ -15,17 +18,40 @@
 
   let currentName = $store.info?.name;
 
-  const rename = () => {
-    console.log("renaming", currentName);
+  const dispatch = createEventDispatcher();
+  const rename = async () => {
+    const canisterId = $store.info?.canister_id;
+    // For type safety reasons. If the modal is open, the canister id is set in the store of the context.
+    if (nonNullish(canisterId)) {
+      startBusy({
+        initiator: "rename-canister",
+      });
+      const { success } = await renameCanister({
+        canisterId,
+        // For type safety reasons. Button is disabled if `currentName` is nullish or ""
+        name: currentName ?? "",
+      });
+
+      stopBusy("rename-canister");
+
+      if (success) {
+        dispatch("nnsClose");
+        toastsSuccess({
+          labelKey: "canister_detail.rename_success",
+          substitutions: { $name: currentName ?? "" },
+        });
+      }
+    }
   };
 </script>
 
-<Modal on:nnsClose>
+<Modal on:nnsClose testId="rename-canister-modal-component">
   <svelte:fragment slot="title"
     ><span>{$i18n.canister_detail.rename_canister}</span></svelte:fragment
   >
 
   <TextInputForm
+    testId="rename-canister-form"
     on:nnsConfirmText={rename}
     on:nnsClose
     bind:text={currentName}
--- a/frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
+++ b/frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
@@ -1,0 +1,20 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { fireEvent, render } from "@testing-library/svelte";
+import RenameCanisterButtonTest from "./RenameCanisterButtonTest.svelte";
+
+describe("RenameCanisterButton", () => {
+  it("opens rename canister modal", async () => {
+    const { queryByTestId } = render(RenameCanisterButtonTest);
+
+    const buttonElement = queryByTestId("rename-canister-button-component");
+    expect(buttonElement).not.toBeNull();
+
+    buttonElement && (await fireEvent.click(buttonElement));
+
+    const modal = queryByTestId("rename-canister-modal-component");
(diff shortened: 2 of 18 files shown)


lmuntaner:

frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
```diff
@@ -0,0 +1,72 @@
...
+    CANISTER_DETAILS_CONTEXT_KEY
+  );
+
+  let currentName = $store.info?.name;
+
+  const dispatch = createEventDispatcher();
+  const rename = async () => {
+    const canisterId = $store.info?.canister_id;
+    // For type safety reasons. If the modal is open, the canister id is set in the store of the context.
+    if (nonNullish(canisterId)) {
```
dskloetd, line 25: How about `assertNonNullish()` instead of `if (nonNullish(`?

frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
```diff
@@ -0,0 +1,72 @@
...
+      startBusy({
+        initiator: "rename-canister",
+      });
+      const { success } = await renameCanister({
+        canisterId,
+        // For type safety reasons. Button is disabled if `currentName` is nullish or ""
```
dskloetd, line 31: `assertNonNullish`?

frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  it("opens rename canister modal", async () => {
+    const { queryByTestId } = render(RenameCanisterButtonTest);
+
+    const buttonElement = queryByTestId("rename-canister-button-component");
+    expect(buttonElement).not.toBeNull();
+
+    buttonElement && (await fireEvent.click(buttonElement));
+
+    const modal = queryByTestId("rename-canister-modal-component");
+    expect(modal).toBeInTheDocument();
```
dskloetd, line 18: Can you check that the modal is not there before clicking the button?

frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { fireEvent, render } from "@testing-library/svelte";
+import RenameCanisterButtonTest from "./RenameCanisterButtonTest.svelte";
+
+describe("RenameCanisterButton", () => {
+  it("opens rename canister modal", async () => {
+    const { queryByTestId } = render(RenameCanisterButtonTest);
```
dskloetd, line 10: This is a brand new test. Why not start out using page objects?

frontend/src/tests/lib/modals/canisters/RenameCanisterModal.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    const newName = "new name";
+    const po = renderComponent({ canisterId: mockCanisterId, name: "name" });
+
+    po.enterNewName(newName);
+
+    po.clickRenameButton();
+
+    await runResolvedPromises();
+
+    expect(canisterApi.renameCanister).toBeCalledWith({
```
dskloetd, line 45: Expect it to be called only once?

frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
dskloetd, line 18: You don't need to change it if you like it this way, but technically the button doesn't cause the modal to be opened. All the button does is dispatch an event. So I would expect this test to only test that the even is dispatched.

The actual modal is opened by the CanisterDetail component so testing that the modal is opened should be done there (and I think it is). So the RenameCanisterButtonTest component seem unnecessary as well.

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  getRenameCanisterModalPo(): RenameCanisterModalPo {
+    return RenameCanisterModalPo.under(this.root);
+  }
+
+  async renameCanister(newName: string): Promise<void> {
+    await this.getRenameCanisterModalPo().enterNewName(newName);
+    await this.getRenameCanisterModalPo().clickRenameButton();
+  }
```
dskloetd, line 31: I think it's useful to also wait for the modal to be absent. Then if the test continues, we know that the rename has happened before we continue.

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  async getCanisterTitle(): Promise<string> {
```
dskloetd, line 33: Is canister title the same as canister name?

frontend/src/tests/page-objects/RenameCanisterModal.page-object.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  getTextInputFormPo(): TextInputFormPo {
+    return TextInputFormPo.under({
+      element: this.root,
+      testId: "rename-canister-form",
+    });
+  }
+
+  enterNewName(name: string): Promise<void> {
```
dskloetd, line 22: "New" seems redundant. How about just `enterName`?

frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
lmuntaner: Shall we discuss this in general terms?

I think that now I've been following our current patterns, but maybe we want to think of a better way to handle this.

frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
lmuntaner: Good point.

Where is the point of testing implementation, though?

For example, the button component used to have the modal as well. But due to some UI redesign, we had to move modals to the page component and dispatch an event from the buttons.

I'll change it because I think it makes sense here, but it's something that I wonder sometimes.

frontend/src/tests/lib/modals/canisters/RenameCanisterModal.spec.ts
lmuntaner: sure, I forgot.

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
lmuntaner: No, this is the main title of the page. If there is no name, then the id is used.

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
lmuntaner: sure, makes sense.

frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
lmuntaner: I refactored the spec file to test the emitted event.

frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
lmuntaner: done

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
lmuntaner: done

lmuntaner:

frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
dskloetd: Yes, OK. Unfortunately David seems to be against.
https://dfinity.slack.com/archives/C03HZERHBT3/p1688373465407689?thread_ts=1688369736.584269&cid=C03HZERHBT3

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
```diff
@@ -26,8 +26,9 @@ export class CanisterDetailPo extends BasePageObject {
   }
 
   async renameCanister(newName: string): Promise<void> {
-    await this.getRenameCanisterModalPo().enterNewName(newName);
+    await this.getRenameCanisterModalPo().enterName(newName);
     await this.getRenameCanisterModalPo().clickRenameButton();
+    await this.getRenameCanisterModalPo().waitForAbsent();
```
dskloetd, line 31: Given that all the calls in this method are call on the modal PO, I think this should actually be a method on the modal PO.
It's useful to have the method here as well but it can just call the same method on the modal.

frontend/src/tests/lib/components/canister-detail/RenameCanisterButton.spec.ts
dskloetd: Well, you do want to test that the button opens the modal but the correct level to test that is at the CanisterDetail component.

The fact that the button emits an event is part of the public API of the button, so from the perspective of the button it's not an implementation detail. From the perspective of the CanisterDetail component it is an implementation detail so there you don't care about the event.

If you think this is too detailed, I think the argument could be that the button doesn't need to be tested individually because it's already tested as part of the CanisterDetail. I think that's somewhat defensible but I find it just easier to add the simple test than to think about whether it's strictly necessary.

dskloetd (approved):

frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
lmuntaner: Let's discuss this specifically. I believe there are better approaches than what we do now. But nothing that can be applied everywhere.

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
lmuntaner: aahh, ok, got it.

dskloetd: I mean that here in `frontend/src/tests/page-objects/CanisterDetail.page-object.ts`, you could have:
```
  async renameCanister(newName: string): Promise<void> {
    await this.getRenameCanisterModalPo().renameCanister(newName);
  }
 ```
 
And in `frontend/src/tests/page-objects/RenameCanisterModal.page-object.ts` you could have:
```
  async renameCanister(newName: string): Promise<void> {
    await this.enterName(newName);
    await this.clickRenameButton();
    await this.waitForAbsent();
  }
```

lmuntaner: noted, thanks!

PR #2806, opened 2023-07-03
GIX-1658: Fix task runner substitutes canister

lmuntaner:
# Motivation

User wants to add a name when creating a canister.

There was a bug in the create canister flow.

When creating a canister, the nns-dapp backend has a runner task that attached a newly created canister. This method was overriding the previous attched canister by the method `attach_canister` with an empty name.

# Changes

* Attach the canister in the runner flow only if canister_id is not alreay attached.
* Manage race condition in `attach_canister` method. The runner could be faster than the call to this request.

# Tests

* Add tests for `attach_newly_created_canister`.
* Add test for new condition in `attach_canister`.

# Todos

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -820,23 +820,30 @@ impl AccountsStore {
 
             if self.accounts.get(&account_identifier.to_vec()).is_some() {
                 let account = self.accounts.get_mut(&account_identifier.to_vec()).unwrap();
-                if account.canisters.len() >= u8::MAX as usize {
-                    return AttachCanisterResponse::CanisterLimitExceeded;
-                }
 
-                for c in account.canisters.iter() {
+                let mut index_to_remove: Option<usize> = None;
+                for (index,c) in account.canisters.iter().enumerate() {
                     if !request.name.is_empty() && c.name == request.name {
                         return AttachCanisterResponse::NameAlreadyTaken;
-                    } else if c.canister_id == request.canister_id && !c.name.is_empty() {
-                        return AttachCanisterResponse::CanisterAlreadyAttached;
+                    }
+                    // The periodic_task_runner might attach the canister before this call.
+                    // The canister attached by the periodic_task_runner has name `""`
+                    if c.canister_id == request.canister_id {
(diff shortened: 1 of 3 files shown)


lmuntaner:
@bitdivine please review

rs/backend/src/accounts_store.rs
```diff
@@ -814,34 +823,34 @@ impl AccountsStore {
                 if account.canisters.len() >= u8::MAX as usize {
                     return AttachCanisterResponse::CanisterLimitExceeded;
                 }
```
bitdivine, line 825: I think this length check needs to come after the removal, say just before the push.  Otherwise, suppose the limit is 10 and there are 9 entries.  The periodic task runner populates the 10th entry and moments later a user request with a name arrives.  My understanding, and I could be wrong, is that in this case you would like to replace the auto-save entry with the user-generated entry.  But as it is, this would fail because the auto-generated entry would have brought the count up to 10, the max, so would be rejected here.  If you first remove the auto-generated entry this code will succeed.

bitdivine (approved):
I think there is still a minor bug when the count is close to the limit but that is easy to fix.  Approving already.

rs/backend/src/accounts_store.rs
lmuntaner: Good catch! Moving it!

rs/backend/src/accounts_store.rs
```diff
@@ -814,34 +823,34 @@ impl AccountsStore {
...
                         return AttachCanisterResponse::CanisterAlreadyAttached;
                     }
                 }
+
+                // The periodic_task_runner might attach the canister before this call.
+                // The canister attached by the periodic_task_runner has name `""`
+                if let Some(index) = Self::find_canister_index(account, request.canister_id) {
+                    // Remove the previous attached canister before reattaching.
+                    account.canisters.remove(index);
+                }
```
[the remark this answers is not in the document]
lmuntaner: Rust doesn't like iterating and removing the index in the same loop.

But I stored the index to remove in a variable to avoid looping again. Thanks!

PR #2808, opened 2023-07-03
GIX-1658: Add canister name in create canister flow

lmuntaner:
# Motivation

User can create a canister with name.

# Changes

* Add step `SelectName` in `CreateCanisterModal` using the `TextInputForm` component.
* Add `name` in canister service `createCanister`.
* Render canister name if present in the `ConfirmCyclesCanister.svelte`.
* Add `required` parameter to `TextInputForm`.

# Tests

* Test new `required` parameter in `TextInputForm`.
* Refactor create canister flow test to easily test with and without name.

# Todos


lmuntaner:
New UX:

https://github.com/dfinity/nns-dapp/assets/4550653/9d1f7d6f-6d3b-4bb9-af88-3feb608a96ab

lmuntaner:

frontend/src/tests/lib/modals/canisters/CreateCanisterModal.spec.ts
```diff
@@ -45,13 +46,17 @@ describe("CreateCanisterModal", () => {
...
+  });
+
   it("should display modal", () => {
     const { container } = render(CreateCanisterModal);
 
     expect(container.querySelector("div.modal")).not.toBeNull();
   });
 
-  it("should create a canister from ICP and close modal", async () => {
+  const testCreateCanister = async (canisterName: string) => {
```
lmuntaner, line 59: This is the new helper that has the code inside the previous test `"should create a canister from ICP and close modal"`.

frontend/src/tests/lib/modals/canisters/CreateCanisterModal.spec.ts
```diff
@@ -71,6 +76,21 @@ describe("CreateCanisterModal", () => {
 
     fireEvent.click(accountCards[0]);
 
+    // Enter Name Screen
```
lmuntaner, line 79: This is a new part of this test.

frontend/src/tests/lib/modals/canisters/CreateCanisterModal.spec.ts
```diff
@@ -98,14 +119,30 @@ describe("CreateCanisterModal", () => {
       ).toBeInTheDocument()
     );
 
+    if (canisterName.length > 0) {
```
lmuntaner, line 122: This is also a new part.

frontend/src/tests/lib/modals/canisters/CreateCanisterModal.spec.ts
```diff
@@ -98,14 +119,30 @@ describe("CreateCanisterModal", () => {
...
+    }
+
     const done = jest.fn();
     component.$on("nnsClose", done);
 
     await clickByTestId(queryByTestId, "confirm-cycles-canister-button");
 
     await waitFor(() => expect(done).toBeCalled());
-    expect(createCanister).toBeCalled();
+    expect(createCanister).toBeCalledWith({
```
lmuntaner, line 132: I also checked how to service it was called.

To be done, mock the api layer instead of the service. But I didn't want to do it in this PR.

frontend/src/tests/lib/modals/canisters/CreateCanisterModal.spec.ts
```diff
@@ -138,6 +175,13 @@ describe("CreateCanisterModal", () => {
 
     fireEvent.click(accountCards[0]);
 
+    // Enter Name Screen
+    await waitFor(() =>
```
lmuntaner, line 179: This is part of the test `"should have disabled button when creating canister with less T Cycles than minimum"`

frontend/src/lib/components/common/TextInputForm.svelte
```diff
@@ -17,12 +18,14 @@
 >
   <div>
     <p class="label"><slot name="label" /></p>
+    <!-- TODO: Show (optional) error message when button is disabled -->
```
dskloetd, line 21: Why would you show an error message when the button is disabled?

Actually, shouldn't the button be disabled if the input is required but empty?
Did you forget to add that logic?

dskloetd (approved):

frontend/src/lib/components/common/TextInputForm.svelte
lmuntaner: That logic is there. But for example, if an amount is larger than the balance we show an error.

Or if the name of the canister is longer than 24 characters we disable the button.

There might be some custom logics to disable the button that need a user feedback.

PR #2810, opened 2023-07-03
Observe ckBTC balances

peterpeterparker:
# Motivation

Use the new web workers to observe the ckBTC balances in the dedicated accounts and wallet page.

# Notes

In the wallet, to refresh the balance, we have to reload the account from store. Because fetching the transactions starts when the account changes, the dapp end up doing an unnecessary call to for the transactions. Something we can try to improve in another PR.

# Changes

- new `CkBTCBalancesObserver` component
- use above component in `CkBTCAccounts` and `CkBTCWallet`

Not related: I also renamed the test `SnsWalletBalancesObserver` to `SnsBalancesObserver`, felt like we can spare a PR.

# Screenshot

demo2


lmuntaner:
Am I supposed to see a change in the balance in the recording? I don't see any...

peterpeterparker:
> Am I supposed to see a change in the balance in the recording? I don't see any...

@lmuntaner watch till the end, the loop is > 30sec

lmuntaner (approved):
Thanks! I wasn't patient enough 😂

PR #2811, opened 2023-07-03
Rm OWN_CANISTER_URL

bitdivine:
# Motivation

# Changes

# Tests

# Todos


lmuntaner (approved):
LGTM!

How about `OWN_CANISTER_ID`? Can we also remove it from the env vars?

bitdivine:
@lmuntaner Re: OWN_CANISTER_ID: Yes.  Good idea.  It is overridden by the canister but it has potential to create confusion and breakage.

PR #2812, opened 2023-07-04
GIX-1656: Apply same rules and show errors when creating and renaming canisters

lmuntaner:
# Motivation

The rules when naming canisters on creating or afterwards was not consistent and there was no feedback to the user about them

In this PR: I apply the same rules in both flows and show an error message to the user when needed.

# Changes

* Add `errorMessage` parameter to `TextInputForm`.
* New canister util `errorCanisterNameMessage` that returns error or `undefined`.
* Use new parameter with new util in `RenameCanisterModal`.
* Use new parameter with new util in `CreateCanisterModal`.
* Change the rules to disable the button in `RenameCanisterModal`.

# Tests

* Test new parameter in `TextInputForm`.
* Test new cansiter util `errorCanisterNameMessage`.
* Add a test case in `CreateCanisterModal`.
* Add a test case in `RenameCanisterModal` and change the expectation when input is empty.

# Todos

### DIFF
--- a/frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
+++ b/frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
@@ -61,7 +61,7 @@
     busy={$busy}
     disabledConfirm={$busy ||
       (nonNullish(currentName) &&
-        currentName?.length >= MAX_CANISTER_NAME_LENGTH)}
+        currentName.length > MAX_CANISTER_NAME_LENGTH)}
     errorMessage={errorCanisterNameMessage(currentName)}
     required={false}
   >
--- a/frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
+++ b/frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
@@ -7,10 +7,12 @@
     type CanisterDetailsContext,
   } from "$lib/types/canister-detail.context";
   import { createEventDispatcher, getContext } from "svelte";
-  import { isNullish, nonNullish } from "@dfinity/utils";
+  import { nonNullish } from "@dfinity/utils";
   import { renameCanister } from "$lib/services/canisters.services";
   import { toastsSuccess } from "$lib/stores/toasts.store";
   import { startBusy, stopBusy } from "$lib/stores/busy.store";
+  import { MAX_CANISTER_NAME_LENGTH } from "$lib/constants/canisters.constants";
+  import { errorCanisterNameMessage } from "$lib/utils/canisters.utils";
 
   const { store }: CanisterDetailsContext = getContext<CanisterDetailsContext>(
     CANISTER_DETAILS_CONTEXT_KEY
@@ -57,9 +59,11 @@
     bind:text={currentName}
     placeholderLabelKey="canister_detail.rename_canister_placeholder"
     busy={$busy}
-    disabledConfirm={isNullish(currentName) ||
-      currentName?.length === 0 ||
-      $busy}
+    disabledConfirm={$busy ||
+      (nonNullish(currentName) &&
+        currentName?.length >= MAX_CANISTER_NAME_LENGTH)}
+    errorMessage={errorCanisterNameMessage(currentName)}
+    required={false}
   >
     <svelte:fragment slot="label"
       >{$i18n.canister_detail.rename_canister_title}</svelte:fragment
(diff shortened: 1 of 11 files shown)


lmuntaner:
New UX:

Allow renaming to empty string:
https://github.com/dfinity/nns-dapp/assets/4550653/32a37ec0-2113-4d07-82a8-96d8109907ee

Do not allow renaming to long names:
https://github.com/dfinity/nns-dapp/assets/4550653/a059f3fd-83ac-4fc1-96f2-750e437c4f7a

Do not allow creating canisters with long names.
https://github.com/dfinity/nns-dapp/assets/4550653/51434e2e-4433-45c5-8250-c15e79eda942

lmuntaner:
@dskloetd please review

frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
```diff
@@ -57,9 +59,11 @@
     bind:text={currentName}
     placeholderLabelKey="canister_detail.rename_canister_placeholder"
     busy={$busy}
-    disabledConfirm={isNullish(currentName) ||
-      currentName?.length === 0 ||
-      $busy}
+    disabledConfirm={$busy ||
+      (nonNullish(currentName) &&
+        currentName?.length >= MAX_CANISTER_NAME_LENGTH)}
```
dskloetd, line 64: Shouldn't it be `>` instead of `>=`?
A length equal to the max length should be allowed, right?

dskloetd, line 64: After checking `nonNullish(currentName)`, you don't need the `?` in `currentName?.length`, right?

But also, `null > 24` evaluates to `false` so I guess you don't need the `nonNullish(currentName)` check unless TypeScript requires it?

frontend/src/lib/utils/canisters.utils.ts
```diff
@@ -75,3 +78,14 @@ export const canisterStatusToText = (status: CanisterStatus): string => {
 
 export const mapCanisterId = (canisterId: CanisterId | string): CanisterId =>
   typeof canisterId === "string" ? Principal.fromText(canisterId) : canisterId;
+
+export const errorCanisterNameMessage = (name: string | undefined) => {
+  if (nonNullish(name) && name?.length >= MAX_CANISTER_NAME_LENGTH) {
```
dskloetd, line 83: `>` instead of `>=`?

frontend/src/tests/lib/modals/canisters/CreateCanisterModal.spec.ts
```diff
@@ -161,6 +164,37 @@ describe("CreateCanisterModal", () => {
...
+    await waitFor(() => expect(getIcpToCyclesExchangeRate).toBeCalled());
+    // wait to update local variable with conversion rate
+    await tick();
+
+    const accountCards = queryAllByTestId("account-card");
+    expect(accountCards.length).toBe(2);
+
+    fireEvent.click(accountCards[0]);
+
+    // Enter Name Screen
```
dskloetd, line 181: Up to here it's all boiler plate test setup, right?
If this is shared between multiple tests, maybe it can we a funfction?

frontend/src/tests/lib/modals/canisters/CreateCanisterModal.spec.ts
```diff
@@ -161,6 +164,37 @@ describe("CreateCanisterModal", () => {
...
+    await waitFor(() =>
+      expect(queryByTestId("create-canister-name-form")).toBeInTheDocument()
+    );
+
+    const longName = "a".repeat(MAX_CANISTER_NAME_LENGTH + 1);
```
dskloetd, line 186: Can you also add a test that `"a".repeat(MAX_CANISTER_NAME_LENGTH)` does not cause an error?

frontend/src/lib/modals/canisters/RenameCanisterModal.svelte
lmuntaner: True, it should be `>`

lmuntaner: Discussed on slack. TS requres the check. but `?` is not necessary.

frontend/src/lib/utils/canisters.utils.ts
lmuntaner: correct

frontend/src/tests/lib/modals/canisters/CreateCanisterModal.spec.ts
lmuntaner: Ok, I'll add it case also in the util test

lmuntaner:

dskloetd (approved):

PR #2813, opened 2023-07-04
GIX-1658: Add canister name check api

lmuntaner:
# Motivation

Fail in the client if we know that the backend will fail when attaching, creating, or renaming a canister.

# Changes

* Add a check for the canister name length in canister api layer.

# Tests

* New test case for `attachCanister`, `createCanister` and `renameCanister`.

# Todos

### DIFF
--- a/frontend/src/lib/api/canisters.api.ts
+++ b/frontend/src/lib/api/canisters.api.ts
@@ -5,7 +5,10 @@ import type {
   CanisterSettings,
 } from "$lib/canisters/ic-management/ic-management.canister.types";
 import type { NNSDappCanister } from "$lib/canisters/nns-dapp/nns-dapp.canister";
-import { CanisterAlreadyAttachedError } from "$lib/canisters/nns-dapp/nns-dapp.errors";
+import {
+  CanisterAlreadyAttachedError,
+  CanisterNameTooLongError,
+} from "$lib/canisters/nns-dapp/nns-dapp.errors";
 import type {
   CanisterDetails as CanisterInfo,
   SubAccountArray,
@@ -15,6 +18,7 @@ import {
   TOP_UP_CANISTER_MEMO,
 } from "$lib/constants/api.constants";
 import { CYCLES_MINTING_CANISTER_ID } from "$lib/constants/canister-ids.constants";
+import { MAX_CANISTER_NAME_LENGTH } from "$lib/constants/canisters.constants";
 import { HOST } from "$lib/constants/environment.constants";
 import { ApiErrorKey } from "$lib/types/api.errors";
 import { nowInBigIntNanoSeconds } from "$lib/utils/date.utils";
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/api/canisters.api.spec.ts
```diff
@@ -283,6 +319,27 @@ describe("canisters-api", () => {
       expect(mockCMCCanister.notifyCreateCanister).not.toBeCalled();
       expect(mockNNSDappCanister.attachCanister).not.toBeCalled();
     });
+
+    it("should fail to create if name is longer than max", async () => {
+      const longName = "a".repeat(MAX_CANISTER_NAME_LENGTH + 1);
```
dskloetd, line 324: Can you also add a test with a name exactly `MAX_CANISTER_NAME_LENGTH` long?

frontend/src/lib/api/canisters.api.ts
dskloetd, line 1: Is this actually useful? The UI will already prevent the user from trying this, and even if the UI fails to do that, the backend will still catch it. So I don't really see the benefit of having it here as well. Am I missing something?

lmuntaner: I added a comment in one of the functions where is most important: `createCanister`. The others I added them for consistency sake:

```
// Failing fast here is specially important.
  // Otherwise the transaction will take place and attaching the canister will fail.
```

That's because the `createCanister` api does a transaction before failing to attach the canister due to the name.

Therefore, the user transfers the tokens, but the canister wouldn't be attached to the nns-dapp.

I could attach it with no name, but I thought that would be a bad UX and it would look like a bug.

Does that make sense?

frontend/src/lib/api/canisters.api.ts
```diff
@@ -25,10 +29,15 @@ import { CMCCanister, ProcessingError, type Cycles } from "@dfinity/cmc";
...
 import type { Principal } from "@dfinity/principal";
 import type { TokenAmount } from "@dfinity/utils";
-import { principalToSubAccount } from "@dfinity/utils";
+import { nonNullish, principalToSubAccount } from "@dfinity/utils";
 import { sendICP } from "./ledger.api";
 import { nnsDappCanister } from "./nns-dapp.api";
 
+// This way, TS understands that if it's invalid, then the name is a string
+type LongName = string;
+const invalidName = (name: string | undefined): name is LongName =>
```
dskloetd, line 38: Can we name the function `isNameTooLong` or something?
With `invalidName` we may want to later add checks for special characters or whatever and that could mess things up.

frontend/src/lib/api/canisters.api.ts
dskloetd: Thanks for explaining. So once we send the tokens, it's very important that everything else succeeds, otherwise you lose your tokens but have nothing to show for it? This seems very bad.
Is it possible to create the canister and attach it before sending the ICP?
If not, I wonder if we should make it more likely that everything succeeds by dropping or shortening the name. Not in this PR.

dskloetd (approved):

frontend/src/lib/api/canisters.api.ts
lmuntaner: The worst case is that it will be caught by the periodic runner of the nns-dapp and it will be attached without name.

I'll add it in the comment also.

And no, we can't attach the canister without making the transfer. To attach we need the canister id, for the canister id, we need to call notify to CMC, to notify we need the transfer to the ledger.

frontend/src/lib/api/canisters.api.ts
lmuntaner: good point, sure

frontend/src/tests/lib/api/canisters.api.spec.ts
lmuntaner: done

PR #2814, opened 2023-07-04
Bump to IC commit  from July 3

bitdivine:
# Motivation
The IC commit is a month old. Patch files need to be updated for custom IC builds and we might as well update them to the latest versions.

# Changes
- Update the IC commit to the latest published version, which is on 3 July 2023.

# Tests
Existing CI should suffice.

# Todos


declarations/sns_governance/sns_governance.did
```diff
@@ -28,10 +28,7 @@ type By = variant {
   NeuronId : record {};
 };
 type CanisterStatusResultV2 = record {
-  controller : principal;
```
bitdivine, line 31: I was not aware that it was allowed to remove a field.  How does this affect us?

lmuntaner (approved):
Thanks!

Can you add a line in the changelog of the SNS aggregator?

I believe this deserves an entry there.

PR #2818, opened 2023-07-04
E2E: Disburse NNS neuron

dskloetd:
# Motivation

I want to replace this old wdio e2e test with a Playwright e2e test so we can delete the old test:
https://github.com/dfinity/nns-dapp/blob/main/e2e-tests/specs/user-N02-neurons-disbursed.e2e.ts

I tried to follow the same steps as that test.

# Changes

1. Add `disburse.spec.ts` which does the e2e test to disburse a neuron.
2. Add necessary page objects and test IDs.
3. Allow specifying the dissolve delay when staking a neuron. Right now it only supports `"max"` and `0`. It's easy to support other numbers but the code path wouldn't be used so let's do that when we use it.

# Tests

yes

# Todos

Is it worth mentioning? It's not new coverage just the same coverage in a different way.


mstrasinskis (approved):
LGTM, thanks!

PR #2819, opened 2023-07-05
Improve reload ckBTC transactions UX

peterpeterparker:
# Motivation

When we integrated the observations of the ckBTC balances in PR #2810 we noticed that in case of updates, the transactions were fully reloaded which is UX wise suboptimal since those are are also observed.

This PR improves the UX by only reloading the list of transactions when needed.

# Changes

- imperatively `reloadTransactions` when needed instead of reloading each time the `account` state changes

# Screenshots

demo


frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
```diff
@@ -51,6 +52,11 @@
       accountIdentifier: account.identifier,
     });
 
+    // On mainnet, the ckBTC Index canister polls new transactions event two seconds. By waiting a bit more time, we optimistically try to fetch the new transaction the user just transferred.
+    const delay = (time: number) =>
+      new Promise((resolve) => setTimeout(resolve, time));
+    await delay(4000);
```
lmuntaner, line 58: Create a constant to better read why is 4?

frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
```diff
@@ -51,6 +52,11 @@
       accountIdentifier: account.identifier,
     });
 
+    // On mainnet, the ckBTC Index canister polls new transactions event two seconds. By waiting a bit more time, we optimistically try to fetch the new transaction the user just transferred.
+    const delay = (time: number) =>
```
lmuntaner, line 56: We have an `util` `waitForMilliseconds` you can use.

frontend/src/lib/pages/CkBTCWallet.svelte
```diff
@@ -56,10 +58,20 @@
...
   // e.g. when a function such as a transfer is called and which also reload the data and populate the stores after execution
-  const reloadAccountFromStore = () => setSelectedAccount();
+  const reloadAccountFromStore = () => {
+    setSelectedAccount();
+    reloadTransactions();
+  };
+
+  const reloadOnlyAccountFromStore = () => setSelectedAccount();
+
+  // We omit await. The UI displays skeletons while loading and the user can proceed with other operations during this time.
```
lmuntaner, line 73: What kind of `await` are you omitting here? I don't understand the comment, sorry.

frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
```diff
@@ -35,7 +36,7 @@
     loading = false;
   };
 
-  const reloadTransactions = async () => {
+  export const reloadTransactions = async () => {
```
lmuntaner, line 39: Why don't you move this logic to the `CkBTCWallet` instead of exposing it?

lmuntaner (commented):
Thanks, looking good, just a question.

frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
peterpeterparker: Good idea

frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
peterpeterparker: Cool

frontend/src/lib/pages/CkBTCWallet.svelte
peterpeterparker: I have hopefully now improved the comment

frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
lmuntaner: In general it seems something to avoid if possible.

But I understand the reason. Ok, thanks.

lmuntaner (approved):

PR #2820, opened 2023-07-05
Improve changelog about canister name

peterpeterparker:
# Motivation

Improve to changelog to make it clear that the "canister name" is not implemented on the protocol level.


lmuntaner (approved):
Thanks!

PR #2822, opened 2023-07-05
Use self instead of OWN_CANISTER_URL in the CSP

bitdivine:
# Motivation
The `OWN_CANISTER_URL` is fundamentally buggy in the way it interacts with the nns-dapp computing its own canister ID on the fly.  It would be good to get rid of it.

One place `OWN_CANISTER_URL` is used is in the CSP, but it isn't needed there.

# Changes
- In the CSP, replace `{{ OWN_CANISTER_URL }}` with `'self'`.

# Tests
- [x] Deploy to the app subnet.  Confirm that requests work as expected.
  - [x] Deployed via gh actions: https://github.com/dfinity/nns-dapp/actions/workflows/deploy-to-app.yaml
  - [x] Deployment loads: https://xnjld-hqaaa-aaaal-qb56q-cai.icp0.io/
  - [x] Confirmed that the CSP doesn't contain the canister ID.  `OWN_CANISTER_URL` is still further down the source code page, iin the parameter list, but that can be deleted once the e2e tests no longer rely on it.

# Todos


lmuntaner (approved):
Thanks!

PR #2823, opened 2023-07-05
Script to reorganize CHANGELOG entries

dskloetd:
# Motivation

We now update the "Unreleased" section of `CHANGELOG-Nns-Dapp.md` with each PR.
We continue doing this after we cut a release and only if the release is successful do we add a new section to the change log.
At that point we have to split the entries that were added after the release was cut from the entries that were added before.
This is a bit tedious to do manually so I made a script.

The script *dot not* remove empty sections. This still needs to be done manually.

# Changes

`scripts/nns-dapp/split-changelog` which does:

1. Take the old changelog until the end of the Unreleased section.
2. Apply the change that have been made since, but add an extra '+' at the start of new entries.
3. Remove the entries without the extra '+'.
4. Remove the extra '+' from the new entries.
5. Append the old changelog starting at the Unreleased section but replace the title with 'Proposal XXXXXX'.
6. Instead of 'XXXXXX' we use the proposal number if we can derive it from the tags on the commit.

# Tests

Added a test which executes the scripts with a specific changelog.

# Todos


scripts/nns-dapp/split-changelog
```diff
@@ -0,0 +1,89 @@
...
+# Writes stdin to the given file, but only after reading all of stdin.
+# Polyfill because `sponge` is not available everywhere.
+sponge() {
+  local tmp_file
+  tmp_file="$(mktemp)"
+  cat >"$tmp_file"
+  mv "$tmp_file" "$1"
+}
+
+CHANGELOG_BASENAME="CHANGELOG-Nns-Dapp.md"
```
lmuntaner, line 47: We will need one also for the aggregator, but in another PR

scripts/nns-dapp/split-changelog
```diff
@@ -0,0 +1,89 @@
...
+
+# Remove the entries that don't have the extra '+'.
+grep -v '^[-* ] ' "$CHANGELOG" | sed -E 's@^\+([-* ] )@* @' | sponge "$CHANGELOG"
+
+# Get proposal number from tags applied to same commit as prod tag
+PROPOSAL_NUMBER="${PROPOSAL_NUMBER:-"$(git tag --contains "$COMMIT" | grep 'proposal-[0-9]\+' | cut -d- -f2)"}"
+PROPOSAL_NUMBER="${PROPOSAL_NUMBER:-XXXXXX}"
+
+# Append the tail of the old changelog with entries before the release was cut.
+echo "## Proposal $PROPOSAL_NUMBER" >>"$CHANGELOG"
```
lmuntaner, line 83: How do we know where this gets appended? It should be appended after end of the unreleased section, right?

scripts/nns-dapp/split-changelog
```diff
@@ -0,0 +1,89 @@
...
+# Source the output file ------------------------------------------------------
+source "$(clap.build)"
+
+# The script does the following:
+# 1. Take the old changelog until the end of the Unreleased section.
+# 2. Apply the change that have been made since, but add an extra '+' at the
+#    start of new entries.
+# 3. Remove the entries without the extra '+'.
+# 4. Remove the extra '+' from the new entries.
+# 5. Append the old changelog starting at the Unreleased section but replace
```
lmuntaner, line 33: This part I don't understand: "Append ... starting at the Unreleased section".

Does that mean there is no Unreleased section anymore after this step? How are the Unreleased entries kept?

lmuntaner (approved):
Rubber stamp, I'm not very familiar with bash.

But I do have some questions about the steps.

Thanks for the comments on top of each step!

scripts/nns-dapp/split-changelog
dskloetd: I can think how to make the script work for both. But the aggregator changes so little that I wonder if it would actually be used.

scripts/nns-dapp/split-changelog
dskloetd: > This part I don't understand: "Append ... starting at the Unreleased section".

"starting at the Unreleased section" just means, excluding the introductory text at the top. So we don't get that text again after the new Unreleased section.

In step 1 we take the old change log until the end of the Unreleased section. So that includes the entire Unreleased section.

Maybe it's best if I explain it visually on a whiteboard, but you can also see in the test what the result is and that there is an Unreleased section.

scripts/nns-dapp/split-changelog
dskloetd: `>>` always appends at the end of the file.

At this point, the file is what we created on line 70. So before line 83 is executed, all we have is everything up to the end of the Unreleased section. So indeed we append after the Unreleased section.

scripts/nns-dapp/split-changelog
```diff
@@ -0,0 +1,89 @@
...
+# Define options
+clap.define long=commit desc="Entries after this commit are considered new" variable=COMMIT default="tags/release-candidate"
+clap.define long=test-changelog desc="An alternate CHANGELOG file for testing" variable=CHANGELOG default=""
+clap.define long=test-old-changelog desc="An alternate old CHANGELOG for testing" variable=OLD_CHANGELOG default=""
+clap.define long=test-proposal desc="An alternate proposal number for testing" variable=PROPOSAL_NUMBER default=""
+# Source the output file ------------------------------------------------------
+source "$(clap.build)"
+
+# The script does the following:
+# 1. Take the old changelog until the end of the Unreleased section.
```
lmuntaner, line 28: Proposal:

1. Make a copy of the old changelog file until the end of the Unreleased section.
2. Apply the change that have been made since, but add an extra '+' at the start of new entries
3. Remove the entries without the extra '+'.
4. Remove the extra '+' from the new entries.
5. Append the title "Proposal XXX" at the end of the file. The XXX is taken from the commit tag.
6. Append the old changelog file below "Proposal XXX" starting at the Unreleased section.

dskloetd: Done

PR #2825, opened 2023-07-06
Test that no new change log entries are added to past releases.

dskloetd:
# Motivation

I've seen twice that people added or tried to add an entry in the change log to an already existing release.
I suspect this is because they started the PR before the change log was updated and weren't careful enough when reviewing the results of the merge.

# Changes

Test that no changes are made to `CHANGELOG-Nns-Dapp.md` below the first `## Proposal [0-9]+` line.

We can look at doing the same for `CHANGELOG-Sns_Aggregator.md` but so far there aren't any changes there.

Also added missing entries to `checks-pass.needs` in `checks.yml`.

# Tests

This is a test.

# Todos

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -276,7 +276,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          -fetch-depth: 0
+          - fetch-depth: 0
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
Cool thanks!

PR #2829, opened 2023-07-06
E2E: Merge neurons

dskloetd:
# Motivation

I want to replace this old wdio e2e test with a Playwright e2e test so we can delete the old test:
https://github.com/dfinity/nns-dapp/blob/main/e2e-tests/specs/user-N03-neurons-merged.e2e.ts

I tried to follow the same steps as that test.

# Changes

1. Add `merge-neurons.spec.ts` which does the e2e test to merge neurons.
2. Add necessary page objects and test IDs.
3. Allow specifying the dissolve non-zero delay when staking a neuron.
4. Allow `TextInputPo` to be used for `<input type="number"` as well since the interface is the same.
5. Implement `getAttribute` and `getClasses` on `PlaywrightPageObjectElement` because it's needed for `NnsNeuronCardPo.isSelected`.

# Tests

yes

# Todos

Is it worth mentioning? It's not new coverage just the same coverage in a different way.

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -53,6 +53,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 - Set a custom URL for `internet_identity` on `ic` rather than using the default.
 - Improve Canister Detail tests by mocking the api layer instead of services.
 - Copied the newest version of clap.bash from snsdemo.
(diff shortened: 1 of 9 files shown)


mstrasinskis:
I think it worths mentioning that the test was migrated to Playwright.

mstrasinskis (approved):
Thanks!

dskloetd:
> I think it worths mentioning that the test was migrated to Playwright.

Done

PR #2830, opened 2023-07-06
Patch empty record

bitdivine:
# Motivation
The patch files applied when generating Rust from SNS .did files have grown quite large and break easily.  There are a few changes in the automatic generation that can make the patch files significantly smaller.  Here is the second:

`didc` generates invalid Rust enum variants of the form: `Name{}` when what is intended is `Name(EmptyRecord)`.  More idiomatic would be `Name` but that doesn't match the data that is sent on the wire.  The patch files contain entries to make this conversion.

# Changes
* Automate adding `EmptyRecord` in place of empty enum entries.

# Tests
- See CI.  Observe that the generated Rust is unchanged but the patch files have become shorter.

# Todos


lmuntaner (approved):
Thanks!

CHANGELOG-Sns_Aggregator.md
```diff
@@ -10,6 +10,7 @@ The SNS Aggregator is released through proposals in the Network Nervous System.
 
 ### Added
 ### Changed
+- Shortened the patch files by automating the use of EmptyRecord.
 - Shortened the patch files used when importing NNS types.
```
dskloetd, line 14: Consider merging these 2 entries into 1.

dskloetd (approved):

CHANGELOG-Sns_Aggregator.md
bitdivine: Can do in the next PR of this stream.

PR #2832, opened 2023-07-07
Fix an re-enabled change log entries test

dskloetd:
# Motivation

In https://github.com/dfinity/nns-dapp/pull/2825 I added a test to make sure new changelog entries aren't incorrectly merged into an existing release.
The test prevented merging PRs without new entries in `CHANGELOG_Nns-Dapp.md` so I disabled it.

# Changes

1. Make the test pass if there are no new entries in `CHANGELOG_Nns-Dapp.md`.
2. Enable the test again.

# Tests

It's a test.

# Todos

Entry already exists.
And not adding an entry verifies that it's allowed now.

### DIFF
--- a/scripts/past-changelog-test
+++ b/scripts/past-changelog-test
@@ -6,7 +6,11 @@ set -xeuo pipefail
 CHANGELOG="CHANGELOG-Nns-Dapp.md"
 MAIN_ANCESTOR="$(git merge-base HEAD origin/main)"
 
-LAST_ADDED_ENTRY_CONTENT="$(git diff "$MAIN_ANCESTOR" "$CHANGELOG" | grep '^+[*-]' | tail -1 | sed -e 's/^+//')"
+if ! LAST_ADDED_ENTRY_CONTENT="$(git diff "$MAIN_ANCESTOR" "$CHANGELOG" | grep '^+[*-]' | tail -1 | sed -e 's/^+//')"; then
+  echo "PASSED: No new entries were added to $CHANGELOG"
+  exit 0
+fi
+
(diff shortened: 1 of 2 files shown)


scripts/past-changelog-test
```diff
@@ -6,7 +6,11 @@ set -xeuo pipefail
 CHANGELOG="CHANGELOG-Nns-Dapp.md"
```
bitdivine, line 6: Not related to this PR but shouldn't this also apply to the aggregator?

bitdivine (approved):
LGTM, thanks.

Tested with:
```
#!/usr/bin/env bash
set -xeuo pipefail

CHANGELOG="CHANGELOG-Nns-Dapp.md"

(
  echo "==================== Should accept no change"
  : "Make sure the changelog has no changes"
  git checkout "$CHANGELOG"
  scripts/past-changelog-test
)

(
  echo "========================== Should accept changes above the last proposal entry"
  git checkout "$CHANGELOG"
  sed -i '0,/^## Proposal/s//- Change above\n&/' "$CHANGELOG"
  git diff
  scripts/past-changelog-test
  git checkout "$CHANGELOG"
)

(
  echo "============================== Should reject changesbelow the last proposal entry"
  git checkout "$CHANGELOG"
  sed -i '0,/^## Proposal.*/s//&\n- Change below/' "$CHANGELOG"
  git diff
  scripts/past-changelog-test
  git checkout "$CHANGELOG"
)
```

A key point is that changes are rejected only if they start with a - or *.  I missed that in the "how this is meant to work".

scripts/past-changelog-test
dskloetd: Probably, but when I looked at the change log for the aggregator, it didn't have any changes or releases at all.

PR #2833, opened 2023-07-07
GIX-1677: Remove VITE_OWN_CANISTER_URL from e2e tests

dskloetd:
# Motivation

VITE_OWN_CANISTER_URL should not be used by the frontend code so we want to remove it.
But it's also used in the Playwright config to determine where to connect to.
We can use `scripts/dfx-canister-url` to determine where to connect instead.

# Changes

1. In `frontend/playwright.config.ts` use environment variable `PLAYWRIGHT_BASE_URL` instead of `VITE_OWN_CANISTER_URL`.
2. Remove `dotenv` from `frontend/playwright.config.ts` because it's no longer used.
3. In `.github/workflows/build.yml` use `scripts/dfx-canister-url` to set `PLAYWRIGHT_BASE_URL`.

# Tests

Yes

# Todos

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,7 +95,7 @@ jobs:
           done
       - name: Run Playwright end-to-end tests
         working-directory: frontend
-        run: PLAYWRIGHT_BASE_URL="$(scripts/dfx-canister-url nns-dapp)" npm run test-e2e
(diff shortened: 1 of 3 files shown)


bitdivine (approved):
LGTM.  Thank you!

PR #2835, opened 2023-07-07
Fmt autogenerated rust

bitdivine:
# Motivation
The patch files applied when generating Rust from SNS .did files have grown quite large and break easily.  There are a few changes in the automatic generation that can make the patch files significantly smaller.  Here is the fourth, or forth for the FORTH programmers of Fife:

The patch files contain ~150 lines of diff due to formatting.  If we format the auto-generated code, the patch files change but in total the patch files become significantly smaller and more consistent.

# Changes
Normally it is advisable to change only one of the code or the patch files, but when formatting it is impossible to avoid changing both.

The Rust files are changed in these ways:
* Remove the header in the autogenerated rust that prevents rustfmt from formatting the files.  (See commit d732dba44eb983684ca2bceaa41840c19cf9b063)
* Format the autogenerated rust files by calling `cargo fmt` (See commit 0235deb3ee48e039a9a81a016420ff76216f16e6)
* Make all struct fields public.  (See commit 76aac4810e0e36d780c315f7bddf8122658ce664) Previously most were public, but the autogenerator missed fields in structs that were short enough to be one-liners.  This is no longer the case, so I manually checked every difference between the autogenerated files and the existing rust and added pub where appropriate.

The autogeneration is changed by:
- Formatting the autogenerated code.
  - Note: This is done after didc and again after sed, so that the sed input and output are both formatted consistently.

The patch files are updated to match.

# Tests
- See CI.
- The proof of correctness is:
  - The rust files have no semantic changes apart from making some fields pub. The remaining changes are in formatting.

# Todos

### DIFF
--- a/scripts/did2rs.sh
+++ b/scripts/did2rs.sh
@@ -43,6 +43,9 @@ DID_PATH="${GIT_ROOT}/declarations/${CANISTER_NAME}/${CANISTER_NAME}.did"
 
 cd "$GIT_ROOT"
 
+which didc
+didc --version
+
 ##########################
 # Translate candid to Rust
 ##########################
--- a/scripts/did2rs.sh
+++ b/scripts/did2rs.sh
@@ -43,9 +43,11 @@ DID_PATH="${GIT_ROOT}/declarations/${CANISTER_NAME}/${CANISTER_NAME}.did"
 
 cd "$GIT_ROOT"
 
-which didc
-didc --version
-rustfmt --version
+: "Ensure that tools are installed and working.  Rustfmt in particular can self-upgrade when called and the self-upgrade can fail."
+{
+  didc --version
+  rustfmt --version
+} >/dev/null
 
 ##########################
 # Translate candid to Rust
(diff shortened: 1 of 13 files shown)


scripts/did2rs.sh
```diff
@@ -94,14 +93,16 @@ cd "$GIT_ROOT"
   #     is not guaranteed to be correct.
   # shellcheck disable=SC2016
   didc bind "${DID_PATH}" --target rs |
+    rustfmt --edition 2021 |
     sed -E 's/^(struct|enum|type) /pub &/;
             s@^use .*@// &@;
             s/([{( ]Deserialize)([,})])/\1, Serialize, Clone, Debug\2/;
-            s/^  [a-z].*:/  pub&/;s/^( *pub ) *pub /\1/;
+            s/^    [a-z].*:/    pub&/;s/^( *pub ) *pub /\1/;
```
bitdivine, line 100: Rustfmt indents by 4 spaces instead of 2.

lmuntaner (approved):
Good idea, makes sense!

PR #2836, opened 2023-07-07
Use new dfx snapshot which has sns_aggregator pre-loaded

dskloetd:
# Motivation

Since https://github.com/dfinity/snsdemo/pull/191 `bin/dfx-snapshot-stock-make` waits for all SNSs to be loaded in the sns_aggregator before saving a snapshot.
The [latest release](https://github.com/dfinity/snsdemo/releases/tag/release-2023-07-07) is the first to have such a snapshot.

# Changes

1. Use the newest dfx snapshot from snsdemo (release-2023-07-07).
2. Remove the code to wait for the sns_aggregator because it's not necessary with the new snapshot.
3. Remove the reconfiguration to speed up the aggregator from the GitHub action that installs the snapshot: `.github/actions/start_dfx_snapshot/action.yaml`
4. Add inputs to the action to control whether to upgrade or reinstall the nns-dapp or sns_aggregator canisters.
5. Upgrade instead of reinstall the aggregator for the e2e-tests, in order to keep the pre-loaded SNSes.

# Tests

See CI

# Todos


bitdivine (approved):
LGTM.  Thanks!

PR #2841, opened 2023-07-07
Provide default values for dfx snapshot action

dskloetd:
# Motivation

We probably always want to be consistent in which snsdemo commit and snapshot release the `start_dfx_snapshot` action uses and right now we have to update those every time in multiple places.

# Changes

1. Provide default values for `snsdemo_ref` and `snapshot_url` in the `start_dfx_snapshot` action.
2. Stop passing the values for which defaults are provided.

# Tests

CI

# Todos

This is in preparation to use the new snapshot which has the aggregator preloaded.
I'll add an entry when for that PR.


bitdivine (approved):
Thank you, I think this will help.

PR #2847, opened 2023-07-10
Fix missing part sfor the automated upgrades

bitdivine:
# Motivation
The automated rust and didc upgrades don't work yet.

# Changes
* Fix the code that gets the current didc version.
* Add the slack integration (still needs a secret).

# Tests
This cannot be tested properly until we have a slack token.

# Todos

  There is  already a changelog entry for this integration.


.github/actions/slack/action.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Slack'
+description: 'A simple action that send a slack message to a channel.'
```
mstrasinskis, line 2:
```suggestion
description: 'A simple action that sends a slack message to a channel.'
```

mstrasinskis (approved):
Rubber stamp, thanks!

peterpeterparker (approved):
Rubber stamp

.github/actions/slack/action.yml
bitdivine: Even a real review!  Thanks.  Done.

PR #2850, opened 2023-07-10
Shard playwright e2e tests on CI

dskloetd:
# Motivation

End-to-end tests are quite slow and if we run more that 3 at the same time on the same runner, they get significantly slower.
A solution is to split them over multiple runners.

# Changes

1. Create a GitHub action to run a shard of the e2e tests.
2. Use the action in `build.yml` to create 2 separate shards.

# Tests

yes

# Todos


bitdivine (approved):
LGTM, thanks!

If you find that a larger runner would be useful, please shout, or change it on the workflow job.

PR #2851, opened 2023-07-10
Patch remnants

bitdivine:
# Motivation
The patch files applied when generating Rust from SNS .did files have grown quite large and break easily.  There are a few changes that can make the patch files significantly smaller.  Here is the sixth, and possibly even the last:

* The comment added by didc has changed; we can update it in our rust code.
* Some of the remaining patch changes are not needed.

# Changes
* Update the comment at the top of the generated files to match that created by didc.
* Remove changes that don't matter, e.g. the order of entries in module imports.

# Tests

# Todos


lmuntaner (approved):
Thanks!

PR #2695, opened 2023-06-14
TVL in various currencies

peterpeterparker:
# Motivation

It was requested to display the TVL in various currencies, so this PR introduces these with a random query (since the TVL is fetched once per session).

# Note

This PR does not modify the particular i18n of the display, we keep the custom display of the TVL but updates the symbol according currency.

# Changes

- update did and canister for up-to-date definition of the TVL canister
- implement new option `currency`
- add `Math.random` to worker in order to get once per session a different currency

# Screenshot

gif

PR #2853, opened 2023-07-11
Revert display of TVL in various currencies to USD only

peterpeterparker:
# Motivation

We were requested to revert the display of TVL in various currencies and display again only USD only.

# Note

In case this would change again in the future, this PR only removes the randomization of the parameters in the query and keep all the logic in place.

PR reverts what was introduced in #2695.

# Changes

- remove random currency parameter in TVL request (undefined = dollar per default)

# Todos


mstrasinskis (approved):
LGTM, thanks! Could be a nice easter egg to loop/jump-to-random currency on click.

peterpeterparker:
Haha love it! A good opportunity of easter egg missed 😉

PR #2854, opened 2023-07-11
Automated update prs

bitdivine:
# Motivation
The nns-dapp automated update messages refer to internet-identity instead of nns-dapp.

# Changes
* Use a variable for the repository name.  In particular, `GITHUB_REPOSITORY` as defined here: https://docs.github.com/en/actions/learn-github-actions/variables

# Tests
Unfortunately this can only be tested on main.

# Todos

   - There is already a changelog entry for this work.


peterpeterparker (approved):
Thx

PR #2856, opened 2023-07-11
Update ic-cdk

bitdivine:
# Motivation
Actually, the goal was to update candid to the latest version.  It seems that this is not possible yet, but it is possible to update the `ic-cdk` so we might as well do that while the candid issue is resolved.

# Changes
* Update `ic-cdk` to the latest version in the aggregator.
* Use `ic-cdk-timers` as that has been split out from `ic-cdk`.

Note: The nns-dapp does not yet make meaningful use of the CDK, so that is untouched in this PR.

# Tests
* See CI.

# Todos


rs/sns_aggregator/src/lib.rs
```diff
@@ -209,7 +209,7 @@ fn setup(config: Option<Config>) {
         let timer_id = set_timer_interval(timer_interval, || ic_cdk::spawn(crate::upstream::update_cache()));
         let old_timer = state.timer_id.replace_with(|_| Some(timer_id));
         if let Some(id) = old_timer {
-            ic_cdk::timer::clear_timer(id);
+            ic_cdk_timers::clear_timer(id);
```
peterpeterparker, line 212: out of curiosity, rust noob question, is there a particular reason in this usage why `clear_timer` here is not declared as well in the `use ic_cdk_timers::{set_timer, set_timer_interval};`?

bitdivine: No very strong reason; I use whichever seems clearer to me in a given context, which is of course subjective.  I leave in the prefix if I want to emphasize where something is coming from or if I am using lots of types in a module and it is cleaner to import the module once and use `modulename::type123` for the different types.

For an example of emphasizing where something comes from:  We use both `dfn_core` and `ic_cdk`.  They both have types such as principals and the types are incompatible, even though they represent the exact same underlying information.  So when it could get confusing I include context of which library the type is from.  The solution is to get dfn_core and the ic repo generally out of our codebase.

For an example of importing a module:  You will often see this for std.  Example:  https://doc.rust-lang.org/std/fs/fn.copy.html  Here `std::fs` is imported, then `fs::copy` is used.

bitdivine: Here, actually, I think you are right.  It would sense to make it consistent.

peterpeterparker (approved):
LGTM, thx

PR #2858, opened 2023-07-11
GIX-1683: Add "huge" to AmountDisplay

lmuntaner:
# Motivation

New neuron settings has as title the stake of the neuron in a similar display but huge font.

In this PR: Add one prop "huge" to `AmountDisplay`.

# Changes

* New prop "huge" to `AmountDisplay` that uses the `font-size-huge`

# Tests

Only UI changes.

# Todos

Not worth mentioning changes yet.

### DIFF
--- a/frontend/src/lib/components/ic/AmountDisplay.svelte
+++ b/frontend/src/lib/components/ic/AmountDisplay.svelte
@@ -9,10 +9,9 @@
   export let inline = false;
   export let singleLine = false;
   export let title = false;
-  export let huge = false;
   export let copy = false;
   export let text = false;
-  export let inheritSize = false;
+  export let size: "inherit" | "huge" | undefined = undefined;
   export let sign: "+" | "-" | "" = "";
   export let detailed: boolean | "height_decimals" = false;
 </script>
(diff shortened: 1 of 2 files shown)


lmuntaner:
New UI. Only the part in the box:

Screenshot 2023-07-11 at 13 19 11

lmuntaner:
@mstrasinskis please review

frontend/src/lib/components/ic/AmountDisplay.svelte
```diff
@@ -23,6 +24,7 @@
   class:title
   class:copy
   class:text
+  class:huge
```
mstrasinskis, line 27: What do you think about the property `size` = `inherit` | `huge`. Then there will be no questions what will happen in case of `huge=true; inheritSize=true`

mstrasinskis (approved):
LGTM, just one minor optional suggestion.

frontend/src/lib/components/ic/AmountDisplay.svelte
lmuntaner: I like it!

lmuntaner: done!

PR #2861, opened 2023-07-11
fix: add sveltekit onDestroy workaround to all detail pages

mstrasinskis:
# Motivation

Fix the broken back button after disbursing of the sns neuron.
The problem was in the svelte routing, that the disbursed neuron UI was not removed from the DOM and it overlapped the new content.

# Changes

- apply the workaround from the wallet page to all detail pages.

# Tests

Tested manually (also the DOM)

https://github.com/dfinity/nns-dapp/assets/98811342/8d1f4b1e-500e-428b-b55a-7a0262f6a915

# Todos


CHANGELOG-Nns-Dapp.md
```diff
@@ -21,6 +21,8 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Removed
 #### Fixed
 
+- back button after sns neuron disburse.
```
peterpeterparker, line 24: there were other side effects so should we rather say something like "destroy sns neuron component and page after disburse" or something?

peterpeterparker (approved):
LGTM, thx

CHANGELOG-Nns-Dapp.md
mstrasinskis: Thanks, updated 👍

PR #2862, opened 2023-07-11
Rm OWN_CANISTER_ID from config.sh and config test vectors

bitdivine:
# Motivation

# Changes

# Tests

# Todos


bitdivine:
Currently this is too hard as the local frontend developer environment uses the static config.

PR #2864, opened 2023-07-12
GIX-1683: NnsNeuronPageHeader Component

lmuntaner:
# Motivation

We are reshuffling the neuron detail page to give a better UX.

In this PR: New component for the NNS neuron detail page.

# Changes

* Upgrade gix-components library.
* New component NnsNeuronPageHeader that goes at the beginning of the new neuron detail page.

# Tests

* Test the new component `NnsNeuronPageHeader`.

# Todos

Not yet worth an entry.

### DIFF
--- a/frontend/src/lib/components/ic/AmountDisplay.svelte
+++ b/frontend/src/lib/components/ic/AmountDisplay.svelte
@@ -129,8 +129,6 @@
 
     &.huge span.value {
       font-size: var(--font-size-huge);
-      // `huge` will not span multiple lines, so we can set it to 1 to not have extra space above and below the value
-      line-height: 1;
     }
   }
 
--- a/frontend/src/lib/components/ic/AmountDisplay.svelte
+++ b/frontend/src/lib/components/ic/AmountDisplay.svelte
@@ -129,6 +129,8 @@
 
     &.huge span.value {
       font-size: var(--font-size-huge);
+      // `huge` will not span multiple lines, so we can set it to 1 to not have extra space above and below the value
+      line-height: 1;
     }
   }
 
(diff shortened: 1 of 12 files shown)


lmuntaner:
New UI:

Screenshot 2023-07-12 at 08 09 04
Screenshot 2023-07-12 at 08 09 23

peterpeterparker:
Just had a look at the screenshot and noticed that the padding is off - i.e. content is not vertically centered in header. It's expected?

252873654-2f2840a8-5e1c-4dcc-8a84-1e6e059ecef3

peterpeterparker:
If it can be interesting too, of course personal feeling but, on mobile the subtitle feels off too, like "not centered".
Would it look better by adding a padding left that is as long as the width of the copy icon, that way the neuron id would be centered?

252873662-ab9187d3-e8ad-4bf8-a1e7-786f7b17ddd4

lmuntaner:
Thanks for the review @peterpeterparker !

Very good points.

1. I'll talk with @artkorotkikh-dfinity I believe this is an issue at the component level. If we want the PageHeader to have always more spacing below than above the title, we should set it at that level. The component as you can see in the [PR](https://github.com/dfinity/gix-components/pull/231) didn't consider that. I'll convert this to Draft until we fix or agree with desing.
2. Yes, the problem is the spacing in the Copy button. I find it ugly to add a left padding, but I can't see another solution for now either.

mstrasinskis (approved):
LGTM, thanks!

lmuntaner:
Mobile alignment fixed:

Screenshot 2023-07-12 at 16 55 22

lmuntaner:
@mstrasinskis @peterpeterparker please review again.

The different spacings has not been solved. I'll prepare a document explaining it.

I split the page header into two components:

- NnsNeuronPageHeader
- NnsNeuronPageHeading

I discussed it with Artem

peterpeterparker:
> The different spacings has not been solved. I'll prepare a document explaining it.

No need to prepare a document for me. I'm fine with just CSS or no CSS. My comments were just meant to help if they would do so.

lmuntaner:
> > The different spacings has not been solved. I'll prepare a document explaining it.
> 
> No need to prepare a document for me. I'm fine with just CSS or no CSS. My comments were just meant to help if they would do so.

It helped me to understand why it wasn't matching and a few things we can do to make sure that we don't have this in other places.

I'll share it when it's ready.

PR #2865, opened 2023-07-12
Set transitions as |local to solve populating destroy

peterpeterparker:
# Motivation

Svelte transitions are executed by default in components regardless if their parent components destroy those or not. This behavior is known as the `|global` transitions.

This leads NNS dapp to various issues in which components are not destroyed and therefore, there context remains active.

Until today we solved this by applying a workaround in which we destroyed the all layout in case of navigation. While it solved the issue, this had for side effect to be a bit unpleasant for the eyes as the all page layout disappeared on navigation.

By setting all the animation to `|local` we can solve the issue and make the navigation more fluid.

# References

- documentation: https://svelte.dev/docs/v4-migration-guide#transitions-are-local-by-default
- svelte issue: https://github.com/sveltejs/svelte/issues/6686

# PRs

- [x] https://github.com/dfinity/gix-components/pull/246

# Previous issues

- #1741
- #2861

# Changes

- update gix-cmp to set `|local` to any Svelte transitions
- set `|local` to the Svelte animations implemented in NNS dapp
- remove detail page layout navigating guard

# Notes

In the login page we will continue to observe `$navigating`  because this condition was not only used to solve the issue but, also to present a spinner while the application is loading.

### DIFF
--- a/frontend/src/routes/(login)/+layout.svelte
+++ b/frontend/src/routes/(login)/+layout.svelte
@@ -14,7 +14,6 @@
   onMount(async () => await initAppAuth());
 </script>
 
-<!-- Workaround for SvelteKit issue https://github.com/sveltejs/kit/issues/5434 -->
 {#if isNullish($navigating)}
   <Layout layout="stretch">
     <Banner />
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -770,11 +770,11 @@
       }
     },
     "node_modules/@dfinity/gix-components": {
-      "version": "2.6.0-next-2023-06-29.1",
-      "resolved": "https://registry.npmjs.org/@dfinity/gix-components/-/gix-components-2.6.0-next-2023-06-29.1.tgz",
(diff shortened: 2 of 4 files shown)


frontend/src/routes/(login)/+layout.svelte
```diff
@@ -14,7 +14,6 @@
   onMount(async () => await initAppAuth());
 </script>
 
-<!-- Workaround for SvelteKit issue https://github.com/sveltejs/kit/issues/5434 -->
 {#if isNullish($navigating)}
```
mstrasinskis, line 17: Is this line still needed?

peterpeterparker: Yes, see the "Notes" in PR description

> In the login page we will continue to observe `$navigating` because this condition was not only used to solve the issue but, also to present a spinner while the application is loading.

That is why I just removed the comment. Did not felt like displaying a spinner was worth a particular component - i.e. felt like the remaining code was self explanatory.

mstrasinskis: Ah, true, thank you.

mstrasinskis (approved):
Thx!

PR #2867, opened 2023-07-12
Rm unused field

bitdivine:
# Motivation
There are two version fields for the `ic-cdk-optimizer` in `dfx.json`.  Only one is used.  Confusing.

# Changes
* Delete the unused version field.

# Tests
See CI.

# Todos


lmuntaner (approved):
Thanks!

PR #2868, opened 2023-07-12
Replace ic-cdk-optimizer with ic-wasm-shrink

bitdivine:
# Motivation
`ic-cdk-optimizer` has been deprecated for a long time.  However, as the replacement did not support our use cases we could not migrate to the proposed replacement tools.  Now two things encourage the migration:

* There is now a drop-in replacement in the form of `ic-wasm shrink`
* We cannot make some changes to the Rust code, because `ic-cdk-optimizer` does not support the corresponding opcodes.

# Changes
* Replace `ic-cdk-optimizer` with `ic-wasm shrink`

# Tests
* CI should check correctness.

# Todos

- [x] The wasm size needs to be compared before and after this change.


peterpeterparker:
Just in case useful, in addition to `shrink`, I also added an extra step for the `metadata` when I migrated Juno from `ic-cdk-optimizer` to `ic-wasm`: https://github.com/buildwithjuno/juno/pull/137/files

It's something I cloned from II. PRs https://github.com/dfinity/internet-identity/pull/1370 (ic-wasm) and https://github.com/dfinity/internet-identity/pull/1373/files (metadata)

bitdivine:
@peterpeterparker Thanks for the watchful eye.  We actually already have that, and the git commit, in the metadata-add command.  https://github.com/dfinity/nns-dapp/blob/main/scripts/dfx-wasm-metadata-add

bitdivine:
Note: I would really also like to check the cycle count but there isn't a good way of doing this fast.  I have started making a toy state but creating a realistic state and counting cycles is going to be quite a bit of work.  It is probably better to step back for a moment because something that takes a lot of work needs to pay for itself over many PRs.  Maybe it makes sense to get in touch with people who have replicated mainnet state before, for example.

CHANGELOG-Nns-Dapp.md
```diff
@@ -49,6 +49,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 
 #### Changed
 
+* Replace `ic-cdk-optimizer` with `ic-wasm shrink`.
```
peterpeterparker, line 52: shouldn't this find place in a new version in the changelog?

peterpeterparker (approved):
LGTM, thanks

scripts/setup
```diff
@@ -185,11 +185,6 @@ install_tar_linux() {
   sudo apt-get install tar
 }
 
-install_ic_cdk_optimizer() {
```
lmuntaner, line 188: Don't you need to install ic-wasm? You install it in the Dockerfile, but isn't it also needed to run `./deploy.sh` directly that doesn't use Docker?

```
cargo binstall --no-confirm "ic-wasm@0.3.6"
```

lmuntaner (approved):
LGTM, one question

CHANGELOG-Nns-Dapp.md
bitdivine: I am not sure what you mean.  Do you mean, this should be in a different section?

scripts/setup
bitdivine: Good point.  I didn't check as ic-wasm is not new but it isn't actually installed by setup.

bitdivine: I will do that in a separate PR, as there is a bit more to do than just a one-liner.

CHANGELOG-Nns-Dapp.md
peterpeterparker: I mean that this PR adds an entry in a changelog of a version that just has been executed on mainnet.

bitdivine: At the time of writing, the changelog was in the unreleased section (assuming that I checked correctly) so I just waited for the proposal to be added (~2pm) and then moved the comment up.  It might make more sense to split the changelog when the proposal is made than when it is passed.  But for now this is now correct.

e2e-tests/scripts/nns-canister.Dockerfile
```diff
@@ -7,7 +7,8 @@ RUN apt -yq update && \
     rm -rf /tmp/* /var/lib/apt/lists/* /var/tmp/*
 
 RUN curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/181b5293e73cfe16f7a79c5b3a4339bd522d31f3/install-from-binstall-release.sh | bash && cargo binstall -V
-RUN cargo binstall --no-confirm --version 0.3.2 ic-cdk-optimizer
+# Note: The version should match dfx.json, however it is expected that the old e2e tests, including this file, will be deleted soon so there is no point in adding the code to get the version.
+RUN cargo binstall --no-confirm "ic-wasm@0.3.6"
```
ggreif, line 11: this is an outdated version

PR #2871, opened 2023-07-12
Rename ledger.api|services into icp-ledger.api|services

peterpeterparker:
# Motivation

Rename `ledger.api|services` into `icp-ledger.api|services`.

Now that NNS dapp supports multiple tokens, it makes their target - the ICP ledger - more obvious. Also reflects the fact that we have `ckbtc-ledger.api|services` and `sns-ledger.api|services`.


lmuntaner (approved):
Thanks!

PR #2872, opened 2023-07-12
Print the post-upgrade cycle count

bitdivine:
# Motivation
We would like better visibility of the cycle count on upgrade.

# Changes
* Add a command to estimate the post_upgrade cycle count based on mainnet and local stats.
* Display the post-upgrade cycle estimate.
* Make the labeling and logic around the downgrade-upgrade tests cleaner and more consistent.

# Tests
The build and test workflow output includes this:

Screenshot from 2023-07-13 12-48-49

In addition, detailed statistics are printed in the workflow output itself.

Note: The stats are not as I remember, so I have checked them in detail and will be doing more profiling in the coming days.

# Todos

# Future

- Add pre-upgrade cycle counts


.github/workflows/build.yml
```diff
@@ -156,13 +156,10 @@ jobs:
       - name: Check that metadata is present
         run: |
           scripts/dfx-wasm-metadata-add.test --verbose
```
bitdivine, line 158: The title is changed for clarification as there is also a downgrade-upgrade test for the aggregator.

.github/workflows/build.yml
```diff
@@ -158,11 +158,27 @@ jobs:
...
         run: |
-          git fetch --depth 1 origin tag prod
-          if git diff tags/prod rs/backend | grep -q .
-          then ./scripts/nns-dapp/downgrade-upgrade-test -w nns-dapp.wasm.gz
-          else echo "Skipping test as there are no relevant code changes"
-          fi
+          ./scripts/nns-dapp/downgrade-upgrade-test -w nns-dapp.wasm.gz
+      - name: Count upgrade cycles
+        run: |
+          post_upgrade_instruction_count="$(dfx canister call nns-dapp get_stats | idl2json | jq -r '.performance_counts [] | select(.name == "post_upgrade after set_canister_arguments") | .instruction_count')"
```
anedos-dfinity: Is there a reason to omit `--network mainnet` in the `post_upgrade_instruction_count` command ^ ?

bitdivine: When omitted, as here, it refers to localhost.

bitdivine:
Note: Another interesting datapoint is cycle counts on the app subnet deployment.  These counts are on mainnet but are without significant state and without assets.

PR #2874, opened 2023-07-13
GIX-1725: Merge Only Locked Neurons

lmuntaner:
# Motivation

As explained [here](https://forum.dfinity.org/t/52-year-neuron-fixes/21301),  only neurons in the Locked state are mergeable until further notice.

To avoid confusion with the users, we should apply this condition on top of the rest of conditions we use to disable neurons when merging them.

# Changes

* Add a new case in neuron util helper `isMergeableNeuron`.
* Add the same case in neuron util helper `getMergeableNeuronMessageKey`.

# Tests

* Add a new teste case in `mapMergeableNeurons` which uses the helpers.
* Add the Locked state explicitly in the merging neurons tests. This was not strictly necessary because the mock neuron is already in the locked state. But I preferred to be explicit that the locked state is required for a mergeable neuron.

# Todos

### DIFF
--- a/frontend/src/lib/utils/neuron.utils.ts
+++ b/frontend/src/lib/utils/neuron.utils.ts
@@ -447,6 +447,9 @@ export const isEnoughMaturityToSpawn = ({
 export const isSpawning = (neuron: NeuronInfo): boolean =>
   neuron.state === NeuronState.Spawning;
 
+const isLocked = (neuron: NeuronInfo): boolean =>
+  neuron.state === NeuronState.Locked;
+
 // Tested with `mapMergeableNeurons`
 const isMergeableNeuron = ({
   neuron,
@@ -457,7 +460,7 @@ const isMergeableNeuron = ({
 }): boolean =>
   !hasJoinedCommunityFund(neuron) &&
   !isSpawning(neuron) &&
-  neuron.state === NeuronState.Locked &&
+  isLocked(neuron) &&
   isNeuronControllable({ neuron, accounts });
 
 const getMergeableNeuronMessageKey = ({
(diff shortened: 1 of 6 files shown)


lmuntaner:
New UX:

Screenshot 2023-07-13 at 07 35 54
Screenshot 2023-07-13 at 07 36 00

lmuntaner:

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -457,6 +457,7 @@ const isMergeableNeuron = ({
 }): boolean =>
   !hasJoinedCommunityFund(neuron) &&
   !isSpawning(neuron) &&
+  neuron.state === NeuronState.Locked &&
```
peterpeterparker, line 460:
```
Line 460: neuron.state === NeuronState.Locked &&
Line 479: if (neuron.state !== NeuronState.Locked) {
```

given the above and the fact that we have `isSpawning`, should we extract a util `isLocked` ?

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -587,6 +591,12 @@ const sameManageNeuronFollowees = (neurons: NeuronInfo[]): boolean => {
   return allHaveSameFollowees(sortedFollowees);
 };
 
+/**
+ * This function checks whether two neurons can be merged
```
peterpeterparker, line 595: two neurons or more right?

we do not check if the provide neurons are mergeable for simplicity reason? is it safe engouh?

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: Sure

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: Yes, two or more technically speaking.

I'm thinking on adding the extra check, but in another PR, it might have extra implications.

This util is about checking the conditions related to multiple neurons, not to the neurons itself. That's why I added it as a comment because when I saw the util now, it wasn't clear.

peterpeterparker: The comment is absolutely useful!

I see, sounds cool if you add extra checks in another PR(s).

Thanks!

peterpeterparker (approved):
LGTM, thx

PR #2882, opened 2023-07-14
Monitor memory usage

bitdivine:
# Motivation
We would like to have visibility into memory usage.

# Changes
* Add stable and wasm memory to the exported metrics.
  * Note: We do not make much use of stable memory yet, except in upgrades, so this will currently show the stable memory populated by the last `pre_upgrade` hook.  However we plan to use stable memory more in future.
  * Note: Given that memory size thresholds are in GiB, I have provided this data as MiB so that it is easy to read at a glance without calculation.  The stats API cannot do the same as the stats do not support f64, as it is does not support Eq.
* Add memory logging to the `pre_upgrade` hook, so we have the very latest data available if an upgrade goes wrong.
* Add wasm as the target in for `cargo build`, `cargo check` (covered by `.cargo/config -> build`) and `cargo clippy`, as the wasm memory size code compiles only when wasm is the target.

# Tests
* A test that all expected metrics are present is included in this PR.  It does not test that the values are plausible or e.g. how much memory increases with each user.
* The `get_stats` API call is exercised in `estimate-upgrade-cycles`.  This PR also re-enables the stats calls made during the downgrade-upgrade test.  No use is made of the data though, so this just confirms that the API call works, not that the data is correct.
* This branch has been deployed to an app subnet.  The metrics show the heap size: https://xnjld-hqaaa-aaaal-qb56q-cai.raw.icp0.io/metrics

# Todos

# Future
- Estimates of memory usage if a given wasm is deployed to production.

### DIFF
--- a/rs/backend/src/stats.rs
+++ b/rs/backend/src/stats.rs
@@ -3,10 +3,15 @@ use crate::perf::PerformanceCount;
 use crate::state::State;
 use crate::STATE;
 use candid::CandidType;
+use core::arch::wasm32::memory_size as wasm_memory_size;
+use ic_cdk::api::stable::stable64_size;
 use icp_ledger::BlockIndex;
 use serde::Deserialize;
 #[cfg(test)]
 mod tests;
+#[cfg(target_arch = "wasm32")]
+const WASM_PAGE_SIZE: u64 = 65536;
+const GIBIBYTE: u64 = 1 << 30;
 
 pub fn get_stats(state: &State) -> Stats {
     let mut ans = Stats::default();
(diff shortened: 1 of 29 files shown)


rs/backend/src/stats.rs
```diff
@@ -33,6 +42,11 @@ pub struct Stats {
     pub neurons_topped_up_count: u64,
     pub transactions_to_process_queue_length: u32,
     pub performance_counts: Vec<PerformanceCount>,
+    // TODO: After a transition period, these two can be required rather than being optional.
+    //       The transition period can be considered over when most deployments, including
+    //       production, CI and snsdemo populate these fields.
+    pub stable_memory_size_bytes: Option<u64>,
```
lmuntaner, line 48: Is this stored in stable memory in pre-upgrade?

Is this why you need the optional and keep backwards compatibility?

lmuntaner (approved):
LGTM! Just one comment.

rs/backend/src/stats.rs
bitdivine: These are not stored in stable memory.  Yes, it is for backwards compatibility but the backwards compatibility is to allow tests to run against new and old versions of the code.  Previously, in such cases, some tests were simply disabled for a while.  I don't believe that that is a great solution so I'm exploring what can be done without disabling tests.

bitdivine: Happy to discuss further but closing so that I can merge.

lmuntaner: Ok, sounds good! Thanks for the explanation

PR #2880, opened 2023-07-13
fix: broken onDestroy on navigation

mstrasinskis:
# Motivation

Reverting layout based fix for missing onDestroy after disbursing sns neuron ([more info](https://github.com/dfinity/nns-dapp/pull/2861))

# PRs

Revert PR #2865

# Changes

- apply the workaround from the wallet page to all detail pages.
- set local to global animation flag

# Tests

Tested manually

PR #2884, opened 2023-07-17
fix: calls to onDestroy after disburse

peterpeterparker:
# Motivation

When a Sns neuron is successfully disbursed, the modal in which the action is triggered is closed and the user is automatically redirected to the list of neurons. As a result, we noticed that the page and components `onDestroy` were not called. We fixed the issue by adding a guard at the top of the layout.

Today I debugged the issue again and figured out that the root cause of the particular issue was the fact that the navigation was executed before the dispatcher that closes the modal was triggered.

I couldn't replicate the issue with a sample repo but, by inverting both order, can solve the issue.

# Notes

This PR does not remove the guard. Only fix the particular issue in case we would remove it in the future.

# PRs

Follow-up various PR but, last is #2880

# Changes

- invert `goto` and `dispatch("nnsClose")`

### DIFF
--- a/frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
+++ b/frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
@@ -96,6 +96,8 @@
 
     stopBusy("disburse-sns-neuron");
 
+    dispatcher("nnsClose");
+
     if (success) {
       toastsSuccess({
         labelKey: "neuron_detail.disburse_success",
@@ -103,8 +105,6 @@
 
       await goto($neuronsPathStore, { replaceState: true });
     }
-
-    dispatcher("nnsClose");
   };
 </script>
 


frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
```diff
@@ -96,15 +96,15 @@
...
     stopBusy("disburse-sns-neuron");
 
+    dispatcher("nnsClose");
+
     if (success) {
       toastsSuccess({
         labelKey: "neuron_detail.disburse_success",
       });
 
       await goto($neuronsPathStore, { replaceState: true });
```
mstrasinskis, line 106: I would also remove the await part, to not block the modal close. Because when navigation is done, the modal should be destroyed, not starting the close animation. What do you think about this change?
```suggestion
      goto($neuronsPathStore, { replaceState: true });
```

mstrasinskis (approved):
Thank you!

frontend/src/lib/modals/neurons/DisburseSnsNeuronModal.svelte
peterpeterparker: Ah good to know, thanks!

PR #2888, opened 2023-07-17
Cleanup account identifier text types

peterpeterparker:
# Motivation

This cleanup of the `AccountIdentifierText` type allow us to make a visual difference between `IcpAccountIdentifierText` (current "old" account identifier text representation) and `IcrcAccountIdentifierText` (new Icrc textual representation).

Will be useful for the conversion of ICP to Icrc, it improves readability.

# Changes

- no functional changes, only types were modified

### DIFF
--- a/frontend/src/lib/api/accounts.api.ts
+++ b/frontend/src/lib/api/accounts.api.ts
@@ -64,7 +64,7 @@ export const renameSubAccount = async ({
 }: {
   newName: string;
   identity: Identity;
-  subAccountIdentifier: string;
+  subAccountIdentifier: AccountIdentifierString;
 }): Promise<void> => {
   logWithTimestamp(
     `Renaming SubAccount ${hashCode(subAccountIdentifier)} call...`
--- a/frontend/src/lib/pages/NnsWallet.svelte
+++ b/frontend/src/lib/pages/NnsWallet.svelte
@@ -43,6 +43,7 @@
   import { isNullish, nonNullish } from "@dfinity/utils";
   import ReceiveButton from "$lib/components/accounts/ReceiveButton.svelte";
   import { ICPToken } from "@dfinity/utils";
+  import type { AccountIdentifierText } from "$lib/types/account";
 
   onMount(() => {
     pollAccounts();
@@ -57,7 +58,7 @@
   let transactions: Transaction[] | undefined;
 
   const reloadTransactions = (
-    accountIdentifier: AccountIdentifierString
+    accountIdentifier: AccountIdentifierText
   ): Promise<void> =>
     getAccountTransactions({
       accountIdentifier,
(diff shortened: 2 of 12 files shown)


frontend/src/lib/services/accounts.services.ts
```diff
@@ -195,7 +198,7 @@ export const initAccounts = () => syncAccountsWithErrorHandler(ignoreErrors);
 export const loadBalance = async ({
   accountIdentifier,
 }: {
-  accountIdentifier: string;
+  accountIdentifier: AccountIdentifierText;
```
lmuntaner, line 201: This is an `IcpAccountIdentifier`, isn't it?

frontend/src/lib/types/account.ts
```diff
@@ -7,7 +11,11 @@ export type AccountType =
   | "hardwareWallet"
   | "withdrawalAccount";
 
-export type AccountIdentifierText = string;
+export type IcpAccountIdentifierText = AccountIdentifierString;
+
+export type AccountIdentifierText =
```
lmuntaner, line 16: Where do you plan using this? I believe we should try to avoid having the same function for both.

frontend/src/lib/pages/NnsWallet.svelte
```diff
@@ -57,7 +58,7 @@
   let transactions: Transaction[] | undefined;
 
   const reloadTransactions = (
-    accountIdentifier: AccountIdentifierString
+    accountIdentifier: AccountIdentifierText
```
lmuntaner, line 61: This is an `IcpAccountIdentifier`, isn't it?

frontend/src/lib/types/account.ts
peterpeterparker: What "function for both"? These are types.

We need it in #2869 to migrate ICP to ICRC.

frontend/src/lib/pages/NnsWallet.svelte
peterpeterparker: Ah yeah good catch. Same as for the other review comment, soon `AccountIdentifierText` but for now `IcpAccountIdentifierText`.

frontend/src/lib/api/accounts.api.ts
```diff
@@ -64,7 +64,7 @@ export const renameSubAccount = async ({
 }: {
   newName: string;
   identity: Identity;
-  subAccountIdentifier: string;
+  subAccountIdentifier: AccountIdentifierString;
```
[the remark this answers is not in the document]
peterpeterparker: No it's `AccountIdentifierString`. Given by the did file. `string` was historically "incorrect" here.

lmuntaner (approved):
Thanks for the explanation!

frontend/src/lib/types/account.ts
peterpeterparker: Discussed offline.

PR #2892, opened 2023-07-17
Add ic-wasm installer to setup command

bitdivine:
# Motivation
`ic-wasm` is now needed for `./deploy.sh` as it has taken over from `ic-cdk-optimizer`.  This means that `ic-wasm` needs to be installed by the base setup command.

# Changes
* Add `ic-wasm` to the setup command.

# Tests
*  I removed `ic-wasm` locally, ran setup and checked that `ic-wasm` had been installed again.

# Todos


lmuntaner (approved):
Thanks!

PR #2893, opened 2023-07-17
Check release-sop version when running

dskloetd:
# Motivation

When running `release-sop` we want to make sure to use the newest version of the checklist.

# Changes

1. After `git fetch` check that the current version is the same as the version on `origin/main`.

# Tests

Included

# Todos


scripts/nns-dapp/release-sop
```diff
@@ -238,6 +238,16 @@ verify() {
   fi
 }
 
+is_release_sop_script_newest_version() {
+  if git diff --quiet origin/main scripts/nns-dapp/release-sop; then
+    echo "yes"
+  fi
+  echo
```
mstrasinskis, line 245: Missing indentations?

mstrasinskis (approved):
Thanks, just a couple minor comments.

PR #2895, opened 2023-07-17
Fast CI runners for e2e tests

dskloetd:
# Motivation

End-to-end tests take a lot of time and we are adding more of them.
They are sharded in 2 shards, which can easily be increased but that also duplicates the setup overhead.

# Changes

Set `runs-on: ubuntu-latest-m` for Playwright e2e jobs on CI.

# Tests

When I tested with a fast runner on 1 of the 2 shards it ran in 4 minutes while the other took 6 minutes.

# Todos

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
           retention-days: 3
   test-playwright-e2e-shard-1-of-2:
     needs: build
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


bitdivine (approved):
You might want to increase the number of tests run in parallel.

dskloetd:
> You might want to increase the number of tests run in parallel.

Good idea. I'll do some experimenting with that next.

PR #2896, opened 2023-07-17
feat: aggregator html page js review

peterpeterparker:
# Motivation

More verbose and therefore a bit more heavier but, this review of the JS code of the `index.html` of the aggregator makes it a bit easier to manipulate if we want to add in the future additional information within the list of Snses.

# Changes

- create element and append those to the DOM instead of parsing HTML (`innerHTML = ...`)
- move the script to the end of the body
- wait for HTML elements to be loaded before starting the script
- add a comment and variable for the root url (useful for local development)
- size of the image through css
- image load set as `lazy`
- `article` instead of `div` to improve semantic


bitdivine (approved):
LGTM.  Thanks!

PR #2897, opened 2023-07-18
Prefix accounts.store with icp

peterpeterparker:
# Motivation

Now that we have various tokens and dedicated stores (`icrc-accounts.store`), I was thinking that prefixing `accounts.store` with `icp` - i.e. `icp-accounts.store` - makes it more explicit and easier to read.

WDYT @lmuntaner ?

# Changes

- refactoring only, no functional changes

# Notes

There is another store and service that can be prefixed later as well in separate PRs.


lmuntaner (approved):
Could we then "convert" the `icpAccountStore` into a derived store from "icrc-accounts.store"?

peterpeterparker:
Thanks for the review. Agree about the store and discussed per pm at the same time of discussing the name of the store.

PR #2898, opened 2023-07-18
Add a flag for ICP Icrc

peterpeterparker:
# Motivation

Add a flag `ENABLE_ICP_ICRC` to the configuration which can be use to activate the support for Icrc in the ICP / Nns universe.

Flag is per default set to `false` everywhere.


lmuntaner (approved):
Thanks!

PR #2903, opened 2023-07-18
Remove timeout from waitForAbsent and fix ProjectDetail test

dskloetd:
# Motivation

During the ProjectDetail test the participation modal only closed after a timeout.
We put a timeout in waitForAbsent as a work around, but the proper fix is to advance the fake timers at the right time.
This means we can't call the same `participate` function that's used by the e2e test because it does all the steps together and we need to advance time between some of the steps.
We could just always advance time in `waitForAbsent` but advancing time is only allowed when fake timers are used and there is no way to query Jest/Vitest whether fake timers are being used.

# Changes

1. In `ProjectDetail.spec.ts`, instead of calling `projectDetail.participate()`, perform the individual steps and `await advanceTime()` before expecting the modal to be closed.
2. Add convenience methods to page objects to make doing the individual steps a bit easier.
3. Remove the timeout from `waitForAbsent` in `frontend/src/tests/page-objects/jest.page-object.ts`.

# Tests

Only tests.

# Todos

I don't think it's warranted.

### DIFF
--- a/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
@@ -502,10 +502,15 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
           );
 
           expect(await projectDetail.hasCommitmentAmount()).toBe(false);
-          await projectDetail.participate({
+
+          await projectDetail.clickParticipate();
+          const modal = projectDetail.getParticipateSwapModalPo();
+          await modal.participate({
             amount: amountICP,
             acceptConditions: false,
           });
+          await advanceTime();
+          await modal.waitForAbsent();
           expect(await projectDetail.getCommitmentAmount()).toBe(
             formattedAmountICP
           );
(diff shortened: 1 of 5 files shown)


frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -502,10 +502,15 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
...
           expect(await projectDetail.hasCommitmentAmount()).toBe(false);
-          await projectDetail.participate({
+
+          await projectDetail.clickParticipate();
+          const modal = projectDetail.getParticipateSwapModalPo();
+          await modal.participate({
             amount: amountICP,
             acceptConditions: false,
           });
+          await advanceTime();
```
lmuntaner, line 512: This wasn't working for me with vitest. Did you try it? Maybe because I was advancing a specific amount of time?

lmuntaner (approved):
Thanks!

Just a comment.

frontend/src/tests/lib/pages/ProjectDetail.spec.ts
dskloetd: No I did not try with vitest. But this cleanup should have been done anyway so I see the migration to vitest as separate and it should be easier now with a proper test.

PR #2911, opened 2023-07-18
Rename account identifier in icp-ledger.api to make its icp nature explicit

peterpeterparker:
# Motivation

As we aim to introduce ICRC for ICP we will have to deal with account identifier that are either a textual representation of ICRC or "old" ICP account identifier.

The `accountBalance` endpoint of ic-js that wraps the ICP canister expect an ICP account identifier. Therefore this PR makes this more obvious by prefixing the parameter with `icp`.

It also set the function as `deprecated` as ultimately we aim to use ICRC to fetch the balance of any account.


lmuntaner (approved):
Thanks!

PR #2914, opened 2023-07-18
Wait for the main account when getting test tokens

dskloetd:
# Motivation

When getting tokens for testing, a main account needs to be already loaded or you get an error.
End-to-end tests are fast enough to be burned by this so many of them need to explicitly wait for the main account to appear before trying to get tokens.

# Changes

In the `getICPs` function, wait for the main account to be loaded if it isn't already.

# Tests

Tests still pass even though they no longer wait for the main account to be present.

# Todos

Not note worthy (?)

### DIFF
--- a/frontend/src/lib/services/dev.services.ts
+++ b/frontend/src/lib/services/dev.services.ts
@@ -1,3 +1,4 @@
+import type { Account } from "$lib/types/account";
 import {
   acquireICPTs,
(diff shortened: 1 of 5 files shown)


lmuntaner (approved):
LGTM!

I'm not sure this deserves an entry in the changelog either.

PR #2915, opened 2023-07-19
Prettier v3

peterpeterparker:
# Changes

Bump prettier v3 and plugins.

Remove `pprettier` which is incompatible and which does not seem active anymore (I opened an [issue](https://github.com/microsoft/parallel-prettier/issues/23) more than two months ago without answers).

Set `trailingComma` option to `es5` as the default has changed in v3 [release](https://prettier.io/blog/2023/07/05/3.0.0.html).

Finally format i18n file automatically generated with a prettier script tag instead of inline the call within the js script file.


frontend/package.json
```diff
@@ -7,14 +7,14 @@
...
+    "i18n": "node --experimental-json-modules scripts/i18n.types.js && prettier --write ./src/lib/types/i18n.d.ts",
     "dev": "npm run i18n && vite dev",
     "build": "npm run i18n && tsc --noEmit && vite build && npm run build:post-process",
     "preview": "vite preview",
     "check": "svelte-kit sync && svelte-check --fail-on-warnings --tsconfig ./tsconfig.json && npm run lint",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
-    "lint": "pprettier --check './**/*.{ts,js,mjs,json,scss,svelte,html}' && eslint --max-warnings 0 .",
-    "format": "pprettier --write './**/*.{ts,js,mjs,json,scss,svelte,html}'",
+    "lint": "prettier --check './**/*.{ts,js,mjs,json,scss,svelte,html}' && eslint --max-warnings 0 .",
+    "format": "prettier --write './**/*.{ts,js,mjs,json,scss,svelte,html}'",
```
bitdivine, line 17: I guess with this change, CI could get slower.  A slog as it's fast enough, that's fine though.  Actually, I checked and CI does need a speedup but this is not on the critical path.

This means that pprettier can be removed, but that can be in a separate PR.

bitdivine (approved):
LGTM, thanks!

PR #2916, opened 2023-07-19
Use short paths

peterpeterparker:
# Motivation

Use short paths for few import that used `../../../` instead of `$lib` or `$tests`.


mstrasinskis (approved):
Thanks!

PR #2917, opened 2023-07-19
ICRC Identifier in ICP load accounts

peterpeterparker:
# Motivation

Extend `icp-accounts.service` load accounts function to map ICRC account identifiers.

# Changes

- new type `ICPAccount` that extends `Account` to populate ICP and Icrc identifiers for ICP/NNS
- map ICRC or ICP identifier in `loadAccounts` function dependending if flag is on or not

### DIFF
--- a/frontend/src/lib/services/icp-accounts.services.ts
+++ b/frontend/src/lib/services/icp-accounts.services.ts
@@ -21,6 +21,7 @@ import { FORCE_CALL_STRATEGY } from "$lib/constants/mockable.constants";
 import { nnsAccountsListStore } from "$lib/derived/accounts-list.derived";
 import type { LedgerIdentity } from "$lib/identities/ledger.identity";
 import { getLedgerIdentityProxy } from "$lib/proxy/icp-ledger.services.proxy";
+import { ENABLE_ICP_ICRC } from "$lib/stores/feature-flags.store";
 import type { IcpAccountsStoreData } from "$lib/stores/icp-accounts.store";
 import {
   icpAccountsStore,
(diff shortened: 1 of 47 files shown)


frontend/src/lib/utils/accounts.utils.ts
```diff
@@ -229,3 +233,37 @@ export const sumAccounts = (
 
 export const hasAccounts = (accounts: Account[]): boolean =>
   accounts.length > 0;
+
+export const toIcpAccountIdentifier = (
```
lmuntaner, line 237: Is this used anywhere?

frontend/src/lib/services/icp-accounts.services.ts
```diff
@@ -98,8 +107,19 @@ export const loadAccounts = async ({
     (type: AccountType) =>
     async (
       account: AccountDetails | HardwareWalletAccountDetails | SubAccountDetails
-    ): Promise<Account> => ({
-      identifier: account.account_identifier,
+    ): Promise<IcpAccount> => ({
+      identifier: icrcEnabled
```
lmuntaner, line 111: Why do you think we should pass the feature flag as a parameter?

I'm not sure I understand the value of all these changes.

frontend/src/lib/utils/accounts.utils.ts
peterpeterparker: Yes, `icp-accounts.services`

frontend/src/lib/services/icp-accounts.services.ts
peterpeterparker: If we want to dynamically switch / activate / deactivate the feature flag, we will need to reactively reload the accounts, therefore we would need to subscribe to it in components.

But, not strong opinion. If you rather like to read the flag imperatively in `services` (`get(store)`) for sure I can adapt. I'm fine with both.

peterpeterparker: Discussed per PM, we go for the imperative access to the store in the services.

@lmuntaner done.

frontend/src/lib/utils/accounts.utils.ts
peterpeterparker: Ah no you are right, not in this PR. Let me provide it in another PR.

lmuntaner (approved):
Thanks for the changes! LGTM!

PR #2921, opened 2023-07-19
feat: utils to try to convert an Icrc account identifier to Icp

peterpeterparker:
# Motivation

When ICP will use Icrc, we will need to get the ICP identifier from an Icrc textual representation to query NNS dapp backend. This PR introduces a utils that tries such conversion and fallback to ICP.

# Changes

- new utils function `toIcpAccountIdentifier` that tries to either map an ICRC textual account identifier or fallback to value provided

### DIFF
--- a/frontend/src/lib/utils/accounts.utils.ts
+++ b/frontend/src/lib/utils/accounts.utils.ts
@@ -234,6 +234,13 @@ export const sumAccounts = (
 export const hasAccounts = (accounts: Account[]): boolean =>
   accounts.length > 0;
 
+/**
+ * The NNS Dapp backend accepts ICP account identifiers. The dapp now also supports Icrc textual representation, which is why we may not always know if an identifier is in ICP format or Icrc format.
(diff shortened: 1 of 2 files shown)


frontend/src/lib/utils/accounts.utils.ts
```diff
@@ -229,3 +233,37 @@ export const sumAccounts = (
 
 export const hasAccounts = (accounts: Account[]): boolean =>
   accounts.length > 0;
+
+export const toIcpAccountIdentifier = (
```
lmuntaner, line 237: Could you add a comment of what it does and doesn't.

For example, I find important that it doesn't check that it's a valid old icp account.

lmuntaner (approved):
Approving, just one minor comment

frontend/src/lib/utils/accounts.utils.ts
peterpeterparker: Sure

PR #2924, opened 2023-07-19
ICP transfer with Icrc addresses as destination

peterpeterparker:
# Motivation

Support for ICP transactions with Icrc addresses as the destination.

# Changes

- extend `transferICP` to either use existing `sendICP` or if destination address is an Icrc address `icrcTransfer`
- extend `invalidAddress` to accept Icrc for Nns, that way user can enter an Icrc address in the input field of the transaction modal


lmuntaner (approved):
LGTM! Thanks!

PR #2927, opened 2023-07-20
Fix SnsNeuronDetail.spec.ts

dskloetd:
# Motivation

`SnsNeuronDetail.spec.ts` has 2 identical test:
* "when project is an invalid canister id"
* "when neuron is not found"
Both use an invalid universe ID but the second one was supposed to use a valid universe ID but a neuron ID for a neuron that can't be found.

# Changes

In "when neuron is not found":
1. Use a correct `universe`.
2. Use an unknown neuron ID.
3. Expect `snsGovernanceApi.getSnsNeuron` to have been called twice (query + update).

In "when project is an invalid canister id", expect `snsGovernanceApi.getSnsNeuron` not to have been called to distinguish it from the other test as both tests expect a redirect to the neurons overview page.

# Tests

yes

# Todos

no

### DIFF
--- a/frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
@@ -2,6 +2,7 @@
  * @jest-environment jsdom
  */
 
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!

PR #2930, opened 2023-07-20
Fix: Dissolve Delay Modal current Delay in Dissolving SNS neuron

lmuntaner:
# Motivation

Bug: When a user tried to increase the dissolve delay of a dissolving SNS neuron, the modal didn't show the current dissolve delay. Instead, it showed 0, as if there was no dissolve delay.

# Changes

* Use also the dissolving time in the IncreaseSnsDissolveDelayModal.

# Tests

* Add a test that the delay is as expected both when neuron is locked and dissolving.

# Todos


mstrasinskis (approved):
Thank you!

PR #2932, opened 2023-07-20
Test SNS hotkey management

dskloetd:
# Motivation

There is a race condition with query and update calls when adding/removing SNS neuron hotkeys.
I want to use `queuedStore` to fix the race condition for which I will want to add unit test coverage.
Currently there is no unit test coverage in `SnsNeuronDetail.spec.ts` for adding/removing hotkeys.
So before fixing the race condition I want to improve the existing test coverage so it's easier to test the fix later.

# Changes

1. Support adding/removing SNS neuron permissions in `frontend/src/tests/fakes/sns-governance-api.fake.ts`.
2. In `frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts` test adding and removing hotkeys and also not having permission to add hotkeys.

# Tests

yes

# Todos


mstrasinskis (approved):
LGTM, thanks!

PR #2933, opened 2023-07-21
Upgrade Playwright

dskloetd:
# Motivation

Playwright 1.32 which we currently use, uses Firefox 111.
Dynamic imports from workers (which we use) are only supported from Firefox 114.

# Changes

Upgrade Playwright to version 1.36, which uses Firefox 115.

# Tests

still pass

# Todos

not worth mentioning

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -28,7 +28,7 @@
   },
   "devDependencies": {
     "@mixer/parallel-prettier": "^2.0.3",
-    "@playwright/test": "^1.32.2",
+    "@playwright/test": "^1.36",
     "@rollup/plugin-inject": "^5.0.3",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -28,7 +28,7 @@
   },
   "devDependencies": {
     "@mixer/parallel-prettier": "^2.0.3",
-    "@playwright/test": "^1.32.2",
```
mstrasinskis, line 31: Probably worth mentioning in the changelog.

mstrasinskis (approved):
Thanks!

frontend/package.json
dskloetd: Done

PR #2935, opened 2023-07-21
Cleanup: Use this.getButton() in page objects

dskloetd:
# Motivation

There is a convenience method `getButton(testId: string)` in `BasePageObject` but it's not used in many places where it could be used.

# Changes

Replace `return ButtonPo.under({ element: this.root, testId});` with `return this.getButton(testId)`.

# Tests

test-only

# Todos

not worth mentioning?


mstrasinskis (approved):
LGTM.

PR #2936, opened 2023-07-21
Dev build

bitdivine:
# Motivation
We would like to add API functions to the nns-dapp that are present only in test deployments.  For this, we need a separate test build.

This is not the first case of a wasm having build flavours.  For example, [Internet Identity has production, test and dev flavours](https://github.com/dfinity/internet-identity/releases/tag/release-2023-07-21).

# Changes
*  Add a feature flag called `test` to the nns-dapp crate.  The feature flag will be used to gate API methods available only in test builds.
* Build with `test` in the dockerfile.
* Add the test build to the list of wasms published in releases.
  * For consistency, the `noassets` build, used for testnets, is also published. The `noassets` build could also be named the `tiny` build but that is outside the scope of this PR.
  * Note: This gives us three build flavours: `production`, `test` and `tiny` aka `noassets`
  * Note: For consistency, the wasms are named: `nns-dapp_${flavour}.wasm.gz`
* Create a separate API method list for each build flavour.  The API method list is used to check that the wasm API is as expected, and we expect the APIs to diverge.

# Tests
- See a toy release: https://github.com/dfinity/nns-dapp/releases/tag/dev-build-test-tag

# Todos


scripts/nns-dapp/test-exports
bitdivine, line 1: This checks that the list of methods exported by a wasm correspond to those in a "golden file".  As we expect the APIs to diverge, we need multiple "golden files" and to check against the file for the given flavour.

Dockerfile
```diff
@@ -218,7 +218,9 @@ COPY --from=configurator /build/nns-dapp-arg* /
 # Note: The frontend/.env is kept for use with test deployments only.
 COPY --from=configurator /build/frontend/.env /frontend-config.sh
 COPY --from=build_nnsdapp /build/nns-dapp.wasm.gz /
-COPY --from=build_nnsdapp_without_assets /build/nns-dapp.wasm.gz /nns-dapp_noassets.wasm.gz
+COPY --from=build_nnsdapp /build/nns-dapp_test.wasm.gz /
+COPY --from=build_nnsdapp /build/nns-dapp_production.wasm.gz /
+COPY --from=build_nnsdapp_without_assets /build/nns-dapp_test.wasm.gz /nns-dapp_noassets.wasm.gz
```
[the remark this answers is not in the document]
lmuntaner: Naming convention makes sense to me. Thanks for the explanation!

lmuntaner (commented):
The idea looks good to me and the naming conventions also.

It seems to be a common way to do so.

AFAIK, NNS team has also a dev wasm. Did you check how they do it?

.github/actions/release_nns_dapp/template
bitdivine, line 1: We now have quite a lot of assets for every release; enough that I feel that we cannot add more without some guidance.  The release page just looks overwhelming to me.  This template provides the release summary text; it is used to highlight the most important assets, by providing a table of wasms similar to that provided by Internet Identity: https://github.com/dfinity/internet-identity/releases

The situation with too many assets should improve once we drop the `.wasm` suffix artefacts.

We will doubtless want to tinker with the template as well.  The goal here is to provide the minimum that is enough for releases not to be an embarrassment.

bitdivine:
@lmuntaner I haven't checked the nns team's dev wasms; they build on gitlab instead of github and with bazel instead of cargo so it seems unlikely that we will be able to steal much code from either their CI or their rust.  I did look at the II code, as that is much more similar to our own and they have taken the time to make their CI really nice.

lmuntaner:
> @lmuntaner I haven't checked the nns team's dev wasms; they build on gitlab instead of github and with bazel instead of cargo so it seems unlikely that we will be able to steal much code from either their CI or their rust. I did look at the II code, as that is much more similar to our own and they have taken the time to make their CI really nice.

True, good point. But it's also interesting to know that they prefer to also have a separate build to expose API for testing purposes. I believe that's what they wanted to do to add free maturity to a neuron.

bitdivine:
@lmuntaner  I had a poke around the nns repo.  They use a feature flag, "test", just as here.

They _are_ keeping the Cargo.toml up to date and have defined the feature flag there, even though the bazel builds don't use it.  I guess cargo must still be needed for something.  https://github.com/dfinity/ic/blob/master/rs/nns/governance/Cargo.toml#L96

Here is an example of it being used to change the behaviour of the wasm: https://github.com/dfinity/ic/blob/master/rs/nns/governance/src/proposals/create_service_nervous_system.rs#L16

This is just the same as the way we renable the `reconfigure()` API on dev builds of the `sns_aggregator`.

BUILD.md
```diff
@@ -94,6 +94,20 @@ Our CI also performs these steps; you can compare the SHA256 with the output the
 
 TODO: When we make a proposal, we should have a corresponding release that voters can download. E.g. https://github.com/dfinity/nns-dapp/releases/tag/release-candidate exists but it doesn't have build artefacts.
 
+### Build flavors
```
bitdivine, line 97: This describes the flavours, similar to the way Internet Identity does here: https://github.com/dfinity/internet-identity#flavors

build-backend.sh
```diff
@@ -2,12 +2,23 @@
...
 TOPLEVEL="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# Defines NNS_DAPP_BUILD_FLAVOURS:
+source "${TOPLEVEL}/scripts/nns-dapp/flavours.bash"
 
 ###############
-# backend # (output: nns-dapp.wasm.gz)
+# backend # (output: nns-dapp_test.wasm.gz, nns-dapp_production.wasm.gz and, for backwards compatibility, nns-dapp.wasm.gz)
 ###############
-echo Compiling rust package
-"$TOPLEVEL/build-rs.sh" nns-dapp
```
bitdivine, line 11: Note: I considered passing a flavour as an argument, however the build time for building one vs both is very similar so it doesn't seem worth pushing the complexity of figuring out which build they need, or both, out to the caller.

The first version of this code took the flavour as an argument to docker-build, however I reverted that. Now both flavours are built by the same docker run, just as the docker build builds the backends with and without assets on every run.

scripts/nns-dapp/test-exports
```diff
@@ -2,6 +2,8 @@
 set -euo pipefail
```
bitdivine, line 2: This checks that the list of methods exported by a wasm correspond to those in a "golden file". As we expect the APIs to diverge, we need multiple "golden files" and to check against the file for the given flavour.

.github/actions/release_nns_dapp/template
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: Here is a sample of the template below: https://github.com/dfinity/nns-dapp/releases/tag/dev-build-test-tag

lmuntaner (approved):
Thanks!

PR #2938, opened 2023-07-21
Fix race condition in removing SNS neuron hotkey

dskloetd:
# Motivation

I noticed flakiness when removing a hotkey in `frontend/src/tests/e2e/sns-governance.spec.ts`.
What happens is that we query the neuron after adding the hotkey, but the certified response only comes back after removing the hotkey, and it causes the hotkey to reappear in the UI.
We had a very similar issue when adding a subaccount, which would seem to disappear when an old certified response without the new subaccount comes back.
To fix this, I previously developed `queuedStore` to make sure the certified/non-certified responses are applied to a store in the correct order.

# Changes

1. Use `queuedStore` in `SnsNeuronDetail.svelte` to make sure that changes to the store are applied in the correct order.
2. Make changes to pausable fakes to be able to reproduce the issue in a unit test.
3. Add a unit test where we remove a hotkey and then resolve the original `loadNeuron` where the hotkey was not yet removed to make sure it does not cause the hotkey to reappear.
4. Add some commented-out code in `queued-store` to make it easy to verify that a test does expose the race condition when the `queuedStore` is not used.

In the fake:
1. Allow pausing only certain responses via a filter.
2. Only pause responses but not execution, so when resuming you get the response you would have gotten based on the previous state and not the new state.
3. Allow resolving individual paused responses and not just resuming all of them.
4. Copy a neuron before including it in a response so that when the underlying neuron is changed, the old response still reflects the old state.

# Tests

only tests

# Todos


frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
```diff
@@ -270,6 +270,56 @@ describe("SnsNeuronDetail", () => {
...
+            permission_type: Int32Array.from(HOTKEY_PERMISSIONS),
+          },
+        ],
+      });
+
+      // Delay certified responses for getSnsNeuron
+      fakeSnsGovernanceApi.pauseFor(
+        ({ functionName, args }) =>
+          functionName === "getSnsNeuron" &&
+          typeof args[0] === "object" &&
```
mstrasinskis, line 295: How about replace
```
typeof args[0] === "object" &&
"certified" in args[0] &&
args[0].certified === true
```
with
`args[0]?.certified === true`

mstrasinskis (approved):
LGTM, just one suggestion.

frontend/src/tests/lib/pages/SnsNeuronDetail.spec.ts
dskloetd: Then I get
```
src/tests/lib/pages/SnsNeuronDetail.spec.ts:295:20 - error TS2339: Property 'certified' does not exist on type 'unknown'.
```

PR #2939, opened 2023-07-21
Use Readable type when Writable is not needed

dskloetd:
# Motivation

Every `Writable` store is also a `Readable` store but not every `Readable` store is also a `Writable` store.
So if we don't use the `set` or `update` function on a store, it's more friendly to callers to accept type `Readable<T>`.
This allows functions to also accept queued stores from https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/stores/queued-store.ts

# Changes

Change type `Writable` to `Readable` in some places where a `Writable` store is not needed.

# Tests

existing tests still pass

# Todos

### DIFF
--- a/frontend/src/lib/derived/debug.derived.ts
+++ b/frontend/src/lib/derived/debug.derived.ts
@@ -31,14 +31,13 @@ import {
   readable,
   writable,
   type Readable,
-  type Writable,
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Thank you!

PR #2946, opened 2023-07-24
GIX-1686: SnsNeuronAdvancedSection actions

lmuntaner:
# Motivation

We are shuffling the data in the neuron detail to make it more user friendly.

In this PR: Add the action items in the sns neuron advanced section.

# Changes

* Render the auto stake maturity and split button in the SnsNeuronAdvancedSection.

# Tests

* Add test to check that actions are present.

# Todos

Not yet worth an entry in the changelog.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/SnsNeuronAdvancedSection.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/SnsNeuronAdvancedSection.svelte
@@ -3,16 +3,22 @@
   import { KeyValuePair, Section } from "@dfinity/gix-components";
   import { secondsToDateTime } from "$lib/utils/date.utils";
   import Hash from "../ui/Hash.svelte";
-  import type { SnsNeuron } from "@dfinity/sns";
+  import type { SnsNervousSystemParameters, SnsNeuron } from "@dfinity/sns";
   import { getSnsNeuronIdAsHexString } from "$lib/utils/sns-neuron.utils";
   import SnsNeuronAge from "../sns-neurons/SnsNeuronAge.svelte";
   import { encodeIcrcAccount, type IcrcAccount } from "@dfinity/ledger";
   import type { Principal } from "@dfinity/principal";
-  import { nonNullish } from "@dfinity/utils";
+  import { nonNullish, type Token } from "@dfinity/utils";
   import SnsNeuronVestingPeriodRemaining from "./SnsNeuronVestingPeriodRemaining.svelte";
+  import SnsAutoStakeMaturity from "./actions/SnsAutoStakeMaturity.svelte";
+  import SplitSnsNeuronButton from "./actions/SplitSnsNeuronButton.svelte";
+  import type { E8s } from "@dfinity/nns";
 
   export let governanceCanisterId: Principal | undefined;
   export let neuron: SnsNeuron;
+  export let parameters: SnsNervousSystemParameters;
+  export let transactionFee: E8s;
+  export let token: Token;
 
   let neuronAccount: IcrcAccount | undefined;
   $: neuronAccount = nonNullish(governanceCanisterId)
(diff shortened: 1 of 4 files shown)


lmuntaner:
New UI:

Screenshot 2023-07-24 at 08 32 06

lmuntaner:

dskloetd:
> Not yet worth an entry in the changelog.

Isn't that what the "Not Published" section is for? You don't need an entry for each of these PRs, but one entry about ongoing work in the neuron settings makes sense, I think. Otherwise, let's just remove the "Not Published" section, or am I misunderstanding what it is for?

dskloetd:
The checkbox in the screenshot is basically invisible. Shouldn't it have a different background color?

frontend/src/lib/components/sns-neuron-detail/SnsNeuronAdvancedSection.svelte
```diff
@@ -83,5 +91,8 @@
     flex-direction: column;
     align-items: flex-start;
     gap: var(--padding-2x);
+
+    --checkbox-padding: 0;
+    --checkbox-label-order: 1;
```
dskloetd, line 96: I really hope the new Figma process will get rid of stuff like this.

frontend/src/lib/pages/SnsNeuronDetail.svelte
```diff
@@ -156,7 +156,7 @@
           <SkeletonCard cardType="info" separator />
           <SkeletonCard cardType="info" separator />
         {:else}
-          {#if $ENABLE_NEURON_SETTINGS && nonNullish(parameters) && nonNullish(token) && nonNullish($selectedSnsNeuronStore.neuron)}
+          {#if $ENABLE_NEURON_SETTINGS && nonNullish(parameters) && nonNullish(token) && nonNullish($selectedSnsNeuronStore.neuron) && nonNullish(transactionFee)}
```
dskloetd, line 159: I guess all these checks are just to satisfy the type system?
And it doesn't work if it gets replaced with a single variable that combines all these checks? Like
```
$: areRequiredParamsPresent = nonNullish(parameters) && nonNullish(token) && nonNullish($selectedSnsNeuronStore.neuron) && nonNullish(transactionFee);
```

dskloetd (approved):

frontend/src/lib/components/sns-neuron-detail/SnsNeuronAdvancedSection.svelte
lmuntaner: This is not just Figma, but how our components work. I also don't like it to be honest.

frontend/src/lib/pages/SnsNeuronDetail.svelte
lmuntaner: Yes, I can do that. I kept adding one at a time, but I believe it makes sense now to put them all in one variable.

lmuntaner: Damn, TS doesn't like this... It doesn't understand that if the variables is true, then each type is present.

lmuntaner: Maybe we can think of a better way

frontend/src/lib/components/sns-neuron-detail/SnsNeuronAdvancedSection.svelte
dskloetd: But if we have a consistent design, we don't need to do custom configuration each time we use a checkbox. The default checkbox would just be according to the default design.

dskloetd:
> I wanted to wait until it was all finished. Which is almost there.

If we cut a release now, there won't be a change log entry but there will be unpublished work.
So I think it makes more sense to add an entry with the first of several PRs than with the last of several PRs.

frontend/src/lib/components/sns-neuron-detail/SnsNeuronAdvancedSection.svelte
lmuntaner: That's also true. But I also think that it shouldn't have padding by default. And I can also see that the design might cover checkboxes left and right. Although ideally not...

PR #2949, opened 2023-07-24
Chore: Rename ageSinceSeconds parameter

lmuntaner:
# Motivation

Rename the parameter `ageSinceSeconds` of `createMockSnsNeuron` due to misunderstanding that it's a timestamp, not a duration.

# Changes

* Rename parameter `ageSinceSeconds` to `ageSinceTimestampSeconds` in `createMockSnsNeuron`.

# Tests

Only rename, no new tests.

# Todos

Not worth a changelog entry.

### DIFF
--- a/frontend/src/tests/mocks/sns-neurons.mock.ts
+++ b/frontend/src/tests/mocks/sns-neurons.mock.ts
@@ -33,7 +33,7 @@ export const createMockSnsNeuron = ({
   whenDissolvedTimestampSeconds = BigInt(
     Math.floor(Date.now() / 1000 + 3600 * 24 * 365 * 2)
   ),
-  ageSinceSeconds = BigInt(1000),
+  ageSinceTimestampSeconds = BigInt(1000),
   stakedMaturity = BigInt(100_000_000),
   maturity = BigInt(100_000_000),
   createdTimestampSeconds = BigInt(nowInSeconds() - SECONDS_IN_DAY),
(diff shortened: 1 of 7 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/mocks/sns-neurons.mock.ts
```diff
@@ -33,7 +33,7 @@ export const createMockSnsNeuron = ({
   whenDissolvedTimestampSeconds = BigInt(
     Math.floor(Date.now() / 1000 + 3600 * 24 * 365 * 2)
   ),
-  ageSinceSeconds = BigInt(1000),
+  ageSinceTimestampSeconds = BigInt(1000),
```
dskloetd, line 36: For me the confusion mostly comes from the name starting with the word `age`.
I was going to propose something else but I noticed that the field on `SnsNeuron` starts with `aging` instead of `age`, which might actually help. How about using `aging` here as well?

dskloetd (approved):

frontend/src/tests/mocks/sns-neurons.mock.ts
lmuntaner: Sounds good!

lmuntaner: done

PR #2950, opened 2023-07-24
Canister status "endless loop"

peterpeterparker:
# Motivation

On `mainnet` - not reproducible locally and on testnet - querying the status of canisters for which the user is not a controller currently results in an "endless loop" (call finally expires after 5min).

Not sure exactly what was the root cause but, debugging the issue lead to figure out the issue can be resolved by bumping `agent-js` to last version `v0.18.1`. We were using an "older" version `v0.15.1` which received multiple updates.

# PRs

- [ ] https://github.com/dfinity/ic-js/pull/387

# Changes

- bump `agent-js`
- bump `ic-js`

# Screenshots

Call in endless loop:

image

Excepted error (user is not a controller) after 5min only:

image

image


CHANGELOG-Nns-Dapp.md
```diff
@@ -7,6 +7,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The NNS Dapp is released through proposals in the Network Nervous System. Therefore, each proposal is documented below, following the relevant changes.
 
 ## Unreleased
+
```
lmuntaner, line 10: Why this? Did you git pull? it should be set already.

lmuntaner (approved):
One comment, please take a look, already approving.

PR #2951, opened 2023-07-24
GIX-1745: Test NnsNeuronDetail with new UI

lmuntaner:
# Motivation

We are shuffling data in the neuron details page to make it more user friendly.

In this PR: Add same tests for neuron detail when new UI is enabled.

In another PR: There are two pending tests that need logic in the component and will be done in another task: https://dfinity.atlassian.net/browse/GIX-1764

# Changes

* Add the same test cases in NnsNeuronDetail for old UI in the new UI.
* Add some methods in NnsNeuronDetailPo and NnsNeuronPageHeaderPo.

# Tests

Only test changes.

# Todos

Not worth a specific changelog entry.

### DIFF
--- a/frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
@@ -137,9 +137,9 @@ describe("NeuronDetail", () => {
         canisterId: OWN_CANISTER_ID,
       });
 
-      await waitFor(async () =>
-        expect((await po.getSkeletonCardPos()).length).toBeGreaterThan(0)
-      );
+      await po.getSkeletonCardPo().waitFor();
+
+      expect(await po.getSkeletonCardPo().isPresent()).toBe(true);
     });
 
     it("should render last maturity distribution", async () => {
--- a/frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
@@ -10,12 +10,12 @@ import { overrideFeatureFlagsStore } from "$lib/stores/feature-flags.store";
 import { layoutTitleStore } from "$lib/stores/layout.store";
 import { neuronsStore } from "$lib/stores/neurons.store";
 import { voteRegistrationStore } from "$lib/stores/vote-registration.store";
+import * as fakeGovernanceApi from "$tests/fakes/governance-api.fake";
 import en from "$tests/mocks/i18n.mock";
 import { mockNeuron } from "$tests/mocks/neurons.mock";
-import { mockRewardEvent } from "$tests/mocks/nns-reward-event.mock";
 import { mockVoteRegistration } from "$tests/mocks/proposal.mock";
-import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
 import { NnsNeuronDetailPo } from "$tests/page-objects/NnsNeuronDetail.page-object";
+import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
 import { runResolvedPromises } from "$tests/utils/timers.test-utils";
 import { render, waitFor } from "@testing-library/svelte";
 import { get } from "svelte/store";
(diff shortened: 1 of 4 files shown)


frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
```diff
@@ -85,6 +85,70 @@ describe("NeuronDetail", () => {
...
+        props: {
+          neuronIdText: `${neuronId}`,
+        },
+      });
+
+      const po = NnsNeuronDetailPo.under(new JestPageObjectElement(container));
+
+      expect((await po.getSkeletonCardPos()).length).toBeGreaterThan(0);
+
+      await runResolvedPromises();
```
dskloetd, line 112: This is fragile. Some change in the component might suddenly cause everything to load at render time.
Also this prevents you from using `renderComponent` from line 45.

Can you guarantee that the skeletons are visible by returning an unresolved promise from the API call?

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
```diff
@@ -85,6 +85,70 @@ describe("NeuronDetail", () => {
...
+      expect((await po.getSkeletonCardPos()).length).toBe(0);
+
+      voteRegistrationStore.add({
+        ...mockVoteRegistration,
+        neuronIdStrings: [`${neuronId}`],
+        canisterId: OWN_CANISTER_ID,
+      });
+
+      await waitFor(async () =>
+        expect((await po.getSkeletonCardPos()).length).toBeGreaterThan(0)
```
dskloetd, line 141: If you change `getSkeletonCardPos` to `getSkeletonCardPo` (singular) and only return the first skeleton, then you can use `po.getSkeletonCardPo().waitFor()`. This is also more friendly for e2e tests.
We don't really care about the number of skeleton cards anyway because either all of them or none of them are rendered. So I've already made similar changes in other places.

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
lmuntaner: I don't see how this is fragile, can you expand?

Do you want me to return without waiting for resolvePromises? I'm not sure I understand what you're asking here.

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
lmuntaner: Sure, no strong opinion.

lmuntaner: done

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
dskloetd: `runResolvedPromises` is a work-around for the fact that sometimes promises that have already been provided with their resolved value, don't call their callbacks until we wait for a bit.
Sometimes we need to call `runResolvedPromises` to help the system move along, but we can't and shouldn't rely on the fact that certain things don't happen until we call `runResolvedPromises`. Some day a refactoring might make calling `runResolvedPromises` unnecessary in certain places and this shouldn't break any tests.

Unless I'm mistaken, this is exactly what this test is doing. It's relying on the fact that some promise inside the component has not called its callback yet and that's the only reason we are still seeing skeletons. There is no explicit reason why the skeleton is still showing on line 110. Am I right?

lmuntaner: That is correct.

I can use the pause and resume in the fake, then.

lmuntaner:

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
```diff
@@ -109,7 +112,9 @@ describe("NeuronDetail", () => {
 
       expect((await po.getSkeletonCardPos()).length).toBeGreaterThan(0);
 
-      await runResolvedPromises();
+      fakeGovernanceApi.resume();
+
+      await po.getSkeletonCardPo().waitForAbsent();
 
       expect((await po.getSkeletonCardPos()).length).toBe(0);
```
dskloetd, line 119: This is now redundant.

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
```diff
@@ -85,6 +86,74 @@ describe("NeuronDetail", () => {
...
+
+      voteRegistrationStore.add({
+        ...mockVoteRegistration,
+        neuronIdStrings: [`${neuronId}`],
+        canisterId: OWN_CANISTER_ID,
+      });
+
+      await po.getSkeletonCardPo().waitFor();
+
+      expect(await po.getSkeletonCardPo().isPresent()).toBe(true);
```
dskloetd, line 147: This is redundant given the `waitFor` above.

dskloetd (approved):

frontend/src/tests/lib/pages/NnsNeuronDetail.spec.ts
lmuntaner: true

PR #2954, opened 2023-07-24
E2E flakiness: Don't vote on proposal created by a different test

dskloetd:
# Motivation

End-to-end tests run in parallel against the same environment.
Proposals or global objects and so proposals created in one test can be seen by proposals created in another test.
This causes flakiness when a proposal gets more votes than expected because another test is voting for it at the same time.
This is likely to happen because we run the same test with Chrome and Firefox, potentially at the same time.

# Changes

1. In `frontend/src/tests/e2e/neurons.spec.ts` record the proposer ID and then choose a proposal with that proposer to vote on. This makes sure that different tests vote on different proposals.
2. In `createDummyProposal` return the neuron ID used to create the dummy proposals.
3. Page object changes.

# Tests

I was able to reliably reproduce the issue by adding a delay after creating the proposals.
Then I made sure the issue no longer reproduced with the same delay in play, before removing the delay again.

# Todos

not necessary

### DIFF
--- a/frontend/src/tests/page-objects/NnsProposalList.page-object.ts
+++ b/frontend/src/tests/page-objects/NnsProposalList.page-object.ts
@@ -1,3 +1,4 @@
+import { shortenWithMiddleEllipsis } from "$lib/utils/format.utils";
 import { ProposalCardPo } from "$tests/page-objects/ProposalCard.page-object";
 import { SkeletonCardPo } from "$tests/page-objects/SkeletonCard.page-object";
 import { BasePageObject } from "$tests/page-objects/base.page-object";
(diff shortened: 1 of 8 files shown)


frontend/src/tests/page-objects/NnsProposalList.page-object.ts
```diff
@@ -18,6 +19,21 @@ export class NnsProposalListPo extends BasePageObject {
     return ProposalCardPo.allUnder(this.root);
   }
 
+  async getFirstProposalCardPoForProposer(
+    proposer: string
+  ): Promise<ProposalCardPo> {
+    const shortProposer = shortenWithMiddleEllipsis(proposer, 5);
```
lmuntaner, line 25: Why 5?

Maybe we can reuse a constant between this test and the component?

lmuntaner (approved):
Thanks!

Just a comment but already approving.

frontend/src/tests/page-objects/NnsProposalList.page-object.ts
dskloetd: 5 because that's what's used in [ProposalCard.svelte](https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/components/proposals/ProposalCard.svelte#L50).
The default is 7 and there is more than enough space for that in the card so I'm not sure why it's not using the default.

Both the function with default 7 and its use in `ProposalCard.svelte` with value 5 were added in
https://github.com/dfinity/nns-dapp/pull/1677.

I've added a constant.

lmuntaner: thanks!

PR #2955, opened 2023-07-24
fix: Always update the index.html

bitdivine:
# Motivation
The sns aggregator index.html is not updating.  This is because the index.html is saved with stable memory and not replaced if present.

# Changes
* Always replace the index.html when requested to install the index.html, rather than checking whether it is already there.
* Ditto for the favicon.

# Tests
* Deployed this PR to the app subnet, then again with a different HTML content, then again with the original HTML content.  In every case the displayed HTML after deployment was correct.

# Todos


peterpeterparker (approved):
LGTM, thanks!

PR #2956, opened 2023-07-24
tweak: Show app subnet deployment ref

bitdivine:
# Motivation
When deploying to the app subnet often, it is easy to forget what version was last deployed.

# Changes
* Prominently display which canisters were deployed with which mode with which branch or tag and commit.

# Tests
* See a workflow run; note the body of the workflow was disabled, so the canister commits are "none".  This is expected.  https://github.com/dfinity/nns-dapp/actions/runs/5647370743

# Todos


peterpeterparker (approved):
LGTM, thanks

PR #2957, opened 2023-07-25
GIX-1764: Set neuron id in header title on scroll

lmuntaner:
# Motivation

Current neuron details sets the neuron id in the header when the user scrolls down.

We want to keep the same behavior in the new neuron page.

# Changes

* Update the layout title on intersecting the NnsPageHeader component.
* Update the layout title on intersecting the SnsPageHeader component.

# Tests

* Test that intersection updates layout title in both components.

# Todos

Not worth an entry. This maintains the same functionality as we have now.

### DIFF
--- a/frontend/src/lib/components/neuron-detail/NnsNeuronPageHeader.svelte
+++ b/frontend/src/lib/components/neuron-detail/NnsNeuronPageHeader.svelte
@@ -5,13 +5,35 @@
   import { MAX_NEURON_ID_DIGITS } from "$lib/constants/neurons.constants";
   import { NNS_UNIVERSE } from "$lib/derived/selectable-universes.derived";
   import PageHeader from "../common/PageHeader.svelte";
+  import { onIntersection } from "$lib/directives/intersection.directives";
+  import { layoutTitleStore } from "$lib/stores/layout.store";
+  import { i18n } from "$lib/stores/i18n";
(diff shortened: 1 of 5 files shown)


lmuntaner:

frontend/src/lib/components/neuron-detail/NnsNeuronPageHeader.svelte
```diff
@@ -5,13 +5,35 @@
...
+
+  const updateLayoutTitle = ($event: Event) => {
+    const {
+      detail: { intersecting },
+    } = $event as unknown as CustomEvent<IntersectingDetail>;
+
+    layoutTitleStore.set(
+      intersecting
+        ? $i18n.neuron_detail.title
+        : `${$i18n.core.icp} – ${neuron.neuronId}`
```
peterpeterparker, line 23: should it be `neuron.neuronId.toString()` as on line 38 or it is automatically casted here?

peterpeterparker (approved):
LGTM

frontend/src/lib/components/neuron-detail/NnsNeuronPageHeader.svelte
lmuntaner: it's autimatically casted. Also, this is how we do it now, so it has been working well so far.

PR #2959, opened 2023-07-25
GIX-1765: Manage permissions in NNS neuron buttons

lmuntaner:
# Motivation

Some neuron actions are available only for controllers.

In this PR: Apply same rules to show/hide neuron buttons in new NNS neuron detail.

Except one difference. The Join Neuron's fund is rendered but disabled if neuron is not controlled by user. This is similar to what we do for the auto stake checkbox. Currently, the whole card of Neurons' Fund doesn't appear if neuron is not controlled by the user.

# Changes

* Hide actions if user not controller in NnsAvailableMaturityActionItem.
* Hide split button and disable join neurons' fund checkbox if user not controller in NnsNeuronAdvancedSection.
* Hide actions if user not controller in NnsNeuronDissolveDelayActionItem.
* Hide actions if user not controller in NnsNeuronStateItemAction.
* Add `disabled` prop to JoinCommunityFundCheckbox

# Tests

* Test that buttons are hidden when user not controller in NnsAvailableMaturityActionItem.
* Test that split button is hidden and join neurons' fund checkbox is disabled when user not controller in NnsNeuronAdvancedSection.
* Test that buttons are hidden when user not controller in NnsNeuronDissolveDelayActionItem.
* Test that buttons are hidden when user not controller in NnsNeuronStateItemAction.
* Test new `disabled` prop in JoinCommunityFundCheckbox

# Todos

This is the current functionality, applied to the new UI. It doesn't deserver an entry in the changelog.

### DIFF
--- a/frontend/src/tests/lib/components/neuron-detail/NnsAvailableMaturityActionItem.spec.ts
+++ b/frontend/src/tests/lib/components/neuron-detail/NnsAvailableMaturityActionItem.spec.ts
@@ -3,6 +3,12 @@
  */
 
 import NnsAvailableMaturityActionItem from "$lib/components/neuron-detail/NnsAvailableMaturityActionItem.svelte";
+import { authStore } from "$lib/stores/auth.store";
+import {
+  mockAuthStoreSubscribe,
+  mockIdentity,
+} from "$tests/mocks/auth.store.mock";
+import { mockCanisterId } from "$tests/mocks/canisters.mock";
 import { mockNeuron } from "$tests/mocks/neurons.mock";
 import { NnsAvailableMaturityActionItemPo } from "$tests/page-objects/NnsAvailableMaturityActionItem.page-object";
 import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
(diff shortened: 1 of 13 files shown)


lmuntaner:
New UI. Example of view when user is a hotkey only:

Voting Power Section:
Screenshot 2023-07-25 at 08 48 59

Maturity Section:
Screenshot 2023-07-25 at 08 49 04

Advanced Section:
Screenshot 2023-07-25 at 08 49 14

lmuntaner:
@dskloetd please review

frontend/src/tests/lib/components/neuron-detail/NnsAvailableMaturityActionItem.spec.ts
```diff
@@ -38,9 +50,28 @@ describe("NnsAvailableMaturityActionItem", () => {
...
+      fullNeuron: {
+        ...mockNeuron.fullNeuron,
+        controller: mockCanisterId.toText(),
+      },
+    });
+
+    expect(await po.hasSpawnButton()).toBe(false);
+    expect(await po.hasStakeButton()).toBe(false);
+  });
 });
```
dskloetd, line 77: If I understand correctly, the buttons should also be rendered if the neuron is controlled by an attached hardware wallet. Can you test that as well in all the cases where it applies?

lmuntaner: Yes, correct. I'll add them.

lmuntaner:

dskloetd (approved):

PR #2960, opened 2023-07-25
Improve Ledger app version error message

dskloetd:
# Motivation

With the pervious error message I was confused because I thought I had to upgrade the Ledger firmware but the version didn't match. So I thought the error message could be clearer.

# Changes

Change the error message to mention "Internet Computer app on your Ledger device".

# Tests

Temporarily increase the required version to manually test the error message:
image

# Todos


lmuntaner (approved):
Thanks!

PR #2961, opened 2023-07-25
GIX-1745: Disburse E2e test with new neuron detail page

lmuntaner:
# Motivation

Migrate the e2e tests to the new neuron details page.

In this PR: Set the feature flag to true and use the new UI to disburse a neuron.

# Changes

* Set feature flag `ENABLE_NEURON_SETTINGS` to true in disburse e2e test.
* Change implementation of `disburseNeuron` method in NnsNeuronDetailPo to use new Section.
* Add functionality to disburseNeuron to the related POs.

# Tests

Only test changes.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -37,7 +37,6 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Changed
 
 * Upgraded to Playwright 1.36.
-* E2e disburse test uses new neuron detail UI.
 
 #### Deprecated
 #### Removed
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -37,6 +37,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Changed
 
 * Upgraded to Playwright 1.36.
+* E2e disburse test uses new neuron detail UI.
 
 #### Deprecated
 #### Removed
--- a/frontend/src/tests/e2e/disburse.spec.ts
+++ b/frontend/src/tests/e2e/disburse.spec.ts
@@ -6,9 +6,6 @@ import { expect, test } from "@playwright/test";
 test("Test disburse neuron", async ({ page, context }) => {
   await page.goto("/");
   await expect(page).toHaveTitle("NNS Dapp");
-  await page.evaluate(() =>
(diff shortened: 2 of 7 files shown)


lmuntaner:
@dskloetd please review

CHANGELOG-Nns-Dapp.md
```diff
@@ -37,6 +37,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Changed
 
 * Upgraded to Playwright 1.36.
+* E2e disburse test uses new neuron detail UI.
```
dskloetd, line 40: There is already this entry:
```
* New NNS and SNS neuron details page layout.
```
Isn't this just a part of that?

frontend/src/tests/e2e/disburse.spec.ts
```diff
@@ -13,6 +13,11 @@ test("Test disburse neuron", async ({ page, context }) => {
 
   step("Get some ICP");
   await appPo.getTokens(10);
+  // TODO: Remove once we set feature flag to true https://dfinity.atlassian.net/browse/GIX-1687
+  await page.evaluate(() =>
+    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+    (window as any).__featureFlags.ENABLE_NEURON_SETTINGS.overrideWith(true)
+  );
```
dskloetd, line 20: I think it would be useful to have a utility function for this.

dskloetd (approved):

CHANGELOG-Nns-Dapp.md
lmuntaner: I wasn't sure whether we want to note when the e2e change. But I guess it's obvious. I'll remove the changelog.

frontend/src/tests/e2e/disburse.spec.ts
lmuntaner: I thought about it, but how do you envision it?

Something like this?
```
await setFeatureFlat("ENABLE_NEURON_SETTINGS", true);
```

I'm not so sure it will add that much value and this is temporary code.

dskloetd: Yes, why would it be temporary code? We're going to continue using feature flags into the future, right?

lmuntaner: Yes, that is true.

dskloetd: Then why did you merge without it?

PR #2962, opened 2023-07-25
Proposal detail page back referrer

peterpeterparker:
# Motivation

Implementing the workaround for `onDestroy` at the top level of the `+layout` of the detail pages (PR #2880) had for side effect to make the navigation callbacks of its pages/layouts to not being called. As those are not called, the dapp is not able to determine which was the `referrer` of these pages and per extension, to effectivelly use the referrer in back action.

Long story short, on mainnet, navigation from the "Launchpad" to a "Proposal" and clicking back lead the users to "Proposals" page instead of going back to "Launchpad".

# Changes

- refactor workaround in a dedicated component `LayoutNavGuard`
- use component within the layouts
- remove top layout as unused

# Notes

I noticed a type in the CHANGELOG and since I was already there, I also improved the communication of the fix #2950 which is not related to this PR.

### DIFF
--- a/frontend/src/lib/components/layout/LayoutNavGuard.svelte
+++ b/frontend/src/lib/components/layout/LayoutNavGuard.svelte
@@ -1,9 +1,15 @@
 <script lang="ts">
   import { isNullish } from "@dfinity/utils";
   import { navigating } from "$app/stores";
+  import { Spinner } from "@dfinity/gix-components";
+
+  // To be used only when the transition between pages lasts long; otherwise, it can appear as being glitchy.
+  export let spinner = false;
 </script>
 
 <!-- Workaround for SvelteKit issue https://github.com/sveltejs/kit/issues/5434 -->
 {#if isNullish($navigating)}
   <slot />
+{:else if spinner}
+  <Spinner />
 {/if}
--- a/frontend/src/lib/components/layout/LayoutNavGuard.svelte
+++ b/frontend/src/lib/components/layout/LayoutNavGuard.svelte
@@ -4,12 +4,12 @@
   import { Spinner } from "@dfinity/gix-components";
 
   // To be used only when the transition between pages lasts long; otherwise, it can appear as being glitchy.
-  export let spinner = false;
+  export let spinnerWhileNavigating = false;
 </script>
 
 <!-- Workaround for SvelteKit issue https://github.com/sveltejs/kit/issues/5434 -->
 {#if isNullish($navigating)}
   <slot />
-{:else if spinner}
+{:else if spinnerWhileNavigating}
   <Spinner />
 {/if}
--- a/frontend/src/routes/(login)/+layout.svelte
+++ b/frontend/src/routes/(login)/+layout.svelte
@@ -2,20 +2,18 @@
   import Banner from "$lib/components/header/Banner.svelte";
   import { onMount } from "svelte";
   import { initAppAuth } from "$lib/services/$public/app.services";
-  import { Layout, ContentBackdrop, Spinner } from "@dfinity/gix-components";
+  import { Layout, ContentBackdrop } from "@dfinity/gix-components";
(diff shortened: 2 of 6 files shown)


frontend/src/routes/(login)/+layout.svelte
```diff
@@ -2,20 +2,18 @@
...
-  import { isNullish } from "@dfinity/utils";
   import Warnings from "$lib/components/warnings/Warnings.svelte";
+  import LayoutNavGuard from "$lib/components/layout/LayoutNavGuard.svelte";
 
   onMount(async () => await initAppAuth());
 </script>
 
-<!-- Workaround for SvelteKit issue https://github.com/sveltejs/kit/issues/5434 -->
-{#if isNullish($navigating)}
+<LayoutNavGuard spinner>
```
lmuntaner, line 16: Doesn't this mean that the `slot` is never rendered? When will `spinner` be false?

frontend/src/lib/components/layout/LayoutNavGuard.svelte
```diff
@@ -1,9 +1,15 @@
...
+
+  // To be used only when the transition between pages lasts long; otherwise, it can appear as being glitchy.
+  export let spinner = false;
 </script>
 
 <!-- Workaround for SvelteKit issue https://github.com/sveltejs/kit/issues/5434 -->
 {#if isNullish($navigating)}
   <slot />
+{:else if spinner}
+  <Spinner />
```
lmuntaner, line 14: So, no slot if `spinner` is passed.

frontend/src/routes/(login)/+layout.svelte
lmuntaner: Or shuold it be `spinner={nonNullish($navigating)}`?

lmuntaner (commented):
Just a couple of questions.

frontend/src/lib/components/layout/LayoutNavGuard.svelte
peterpeterparker: I can call it "displaySpinner" or "displaySpinnerWhileNavigating" if you rather like but I found it verbose, that's why I added a comment on the property.

Let me know if I should rename it and to what.

frontend/src/routes/(login)/+layout.svelte
peterpeterparker: No.

```
if not navigating then
   display slot
else 
   if we want to display a spinner during that time
        display a spinner
   end
end
```

peterpeterparker: Also note: not a new feature, code was refactored. I guess you did not catch it because of the name of the variable which we can discuss in your other comment https://github.com/dfinity/nns-dapp/pull/2962#discussion_r1273411213

frontend/src/lib/components/layout/LayoutNavGuard.svelte
lmuntaner: Ah, yes, maybe the name was the confusing part.

`spinnerWhileNavigating` is quite verbose, but I believe I would have understood the prop better if you don't mind.

frontend/src/routes/(login)/+layout.svelte
lmuntaner: Ah, I see now. Yes, I believe the name confused me a bit.

lmuntaner (approved):
I'd appreciate the rename, already approving. Thanks!

frontend/src/lib/components/layout/LayoutNavGuard.svelte
peterpeterparker: Let's do that then.

peterpeterparker: Done.

PR #2964, opened 2023-07-25
Format changelogs

peterpeterparker:
# Motivation

Since we migrated to the CHANGELOGS I often personally produce and notice inconsistency in the format of these files. 
This PR aims to add a dedicated formatter (`prettier`) to solve the issue.

# Notes

Unlike traditional repo on the IC, we do not have the JS tooling at the root of the repo. That's why the formatter use an npm cli command.

# Changes

- new script `fmt-changelogs` that uses prettier to format the CHANGELOGs markdown file
- add script to global `fmt` script

### DIFF
--- a/scripts/fmt
+++ b/scripts/fmt
@@ -10,4 +10,5 @@ time xargs -P8 -I{} bash -c "{}" <<EOF
 ./scripts/fmt-rs
 ./scripts/fmt-sh
 ./scripts/fmt-yaml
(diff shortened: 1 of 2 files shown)


bitdivine (approved):
LGTM.  Thanks!

bitdivine:
Note: We may want to disable the check that old changelogs have not been altered.  You can do that in this PR, if you wish, and we can re-enable it after the next release.

PR #2965, opened 2023-07-25
Do not display proposals title on wide screens

peterpeterparker:
# Motivation

Do not display the "Vote on Proposals" title in the page's header on wide screens to align the behavior with pages that support multiple projects.

UI consistency.

# Changes

- refactor existing code in a new component `LayoutList` to avoid duplicate code and render a title according screen width
- use new component in list layouts including in proposals


lmuntaner (approved):
I wasn't aware of this functionality. But it makes sense to keep consistency.

Thanks!

PR #2972, opened 2023-07-25
Faster formatting

bitdivine:
# Motivation
Formatting and linting shell scripts locally is slow.

So I took a time-bounded window (one CI run) to make things slightly better.  Plus time to clean up the one burst of code.

# Changes
- Support formatting only bash files named on the command line.
- Support formatting only bash files that have changed in git.
  - Note: I have _not_ specified --no-run-if-empty in xargs because this is unsupported on the native mac xargs.
    - This means that fmt-sh --modified will hang if no shell scripts have been modified.
    - If the team is happy to install an updated xargs, or has already done so, we can add --no-run-if-empty and this corner case will just work.
    - Personally, I format after modifying a file so I don't run into the corner case, and I run Linux so I could solve the problem for myself by adding `--no-run-if-empty` but that would break the command for everyone else.
- Repeat for yaml; I have not gone further as CI finished for the job I was waiting for.
- Added `--help`.
- Added short forms of the flags.

# Tests
- I have used this locally for formatting and verified that if I left quotes off a var, the linter complained.

# Todos


lmuntaner (approved):
LGTM!

PR #2973, opened 2023-07-25
feat: Check that the number of accounts etc is preserved across upgrades

bitdivine:
# Motivation
We would like to be sure that accounts and transactions are preserved across upgrades.

# Changes
- Define which data we expect not to change across upgrades and downgrades.
- In the downgrade-upgrade test, ensure that this invariant is preserved.

# Tests
This is a test - see CI for the outcome.

# Todos


lmuntaner (approved):
Thanks!

You have the same changes as in #2974 . But I guess there won't be any conflict.

PR #2974, opened 2023-07-26
fix: Use docker-build --network flag

bitdivine:
# Motivation
`docker-build` now specifies the network by means of a flag.  Some code still passes it as a command line argument, which happens to work if the relevant arguments have already been computed before but not otherwise.  This should be fixed.

# Changes
- Use the `--network` flag instead of an environment variable, whenever `docker-build` is called.

# Tests
- See CI

# Todos


lmuntaner (approved):
LGTM!

PR #2978, opened 2023-07-26
Fix: SNS Governance E2E flakyness

lmuntaner:
# Motivation

We want trustworthy e2e tests.

In this PR: Fix a flakyness in the E2E test sns-governance when getting the SNS tokens.

# Changes

* Change test id for the get tokens button. It relies now on the token symbol. This way we can wait for it in playwright.
* Getting tokens methods expect a new argument "token" to get the button.
* Use the new parameter in the sns-governance e2e test.

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/lib/components/ic/GetTokens.svelte
+++ b/frontend/src/lib/components/ic/GetTokens.svelte
@@ -93,7 +93,7 @@
   {#if $authSignedInStore}
     <button
       role="menuitem"
-      data-tid="get-icp-button"
+      data-tid={`get-${token.symbol}-button`}
       on:click|preventDefault|stopPropagation={() => (visible = true)}
       class="open"
     >
--- a/frontend/src/lib/components/ic/GetTokens.svelte
+++ b/frontend/src/lib/components/ic/GetTokens.svelte
@@ -27,6 +27,9 @@
   let selectedProjectId = OWN_CANISTER_ID;
   $: selectedProjectId = $selectedUniverseIdStore;
 
+  let isNns: boolean;
+  $: isNns = selectedProjectId.toText() === OWN_CANISTER_ID.toText();
+
   const onSubmit = async () => {
     if (invalidForm || inputValue === undefined) {
       toastsError({
@@ -39,10 +42,7 @@
 
     try {
       // Default to transfer ICPs if the test account's balance of the selected universe is 0.
-      if (
-        selectedProjectId.toText() === OWN_CANISTER_ID.toText() ||
-        tokenBalanceE8s === 0n
-      ) {
+      if (isNns || tokenBalanceE8s === 0n) {
         await getICPs(inputValue);
       } else {
         await getTokens({
@@ -93,7 +93,7 @@
   {#if $authSignedInStore}
     <button
       role="menuitem"
-      data-tid={`get-${token.symbol}-button`}
+      data-tid={`get-${isNns || tokenBalanceE8s === 0n ? "icp" : "sns"}-button`}
       on:click|preventDefault|stopPropagation={() => (visible = true)}
       class="open"
     >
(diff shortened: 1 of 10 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/components/ic/GetTokens.svelte
```diff
@@ -93,7 +93,7 @@
   {#if $authSignedInStore}
     <button
       role="menuitem"
-      data-tid="get-icp-button"
+      data-tid={`get-${token.symbol}-button`}
```
dskloetd, line 96: This is clever!
But I have one suggestion and I'm wondering what you think.
How about not using the token symbol but only `icp` or `sns-token` or something.
So not dependent on which SNS token it is.
After all there is not a button per sns token but only a single button.
Then you don't need to pass the token name to the function. We can just have 2 separate page object methods `getIcp` and `getSnsTokens`.
What do you think?

frontend/src/tests/e2e/sns-governance.spec.ts
```diff
@@ -23,8 +23,12 @@ test("Test SNS governance", async ({ page, context }) => {
   expect(snsProjectName).toMatch(/[A-Z]{5}/);
 
   await snsUniverseCard.click();
+
   step("Acquire tokens");
-  await appPo.getTokens(20);
+  await appPo.openMenu();
+  await appPo.getTokensButton(snsProjectName).waitFor();
```
dskloetd, line 29: If you pass the token name to `appPo.getTokens()` (or make it `getSnsTokens()` as I suggest above), I don't think you need this separate `waitFor()`.

frontend/src/lib/components/ic/GetTokens.svelte
lmuntaner: Yes, sounds good! Way better, otherwise getting the name of the SNS might be difficult.

frontend/src/tests/e2e/sns-governance.spec.ts
lmuntaner: Let me try!

frontend/src/lib/components/ic/GetTokens.svelte
lmuntaner: done

frontend/src/tests/e2e/sns-governance.spec.ts
lmuntaner: I had to wait for the button, but I did it inside the PO, not in the test.

frontend/src/tests/e2e/sns-governance.spec.ts
```diff
@@ -23,8 +23,9 @@ test("Test SNS governance", async ({ page, context }) => {
   expect(snsProjectName).toMatch(/[A-Z]{5}/);
 
   await snsUniverseCard.click();
+
   step("Acquire tokens");
-  await appPo.getTokens(20);
+  await appPo.getSnsTokens(1);
```
dskloetd, line 28: Below we try to stake 5 so we need more than 1.
Let's keep the 20 that was there before?

frontend/src/tests/page-objects/App.page-object.ts
```diff
@@ -134,9 +134,16 @@ export class AppPo extends BasePageObject {
     await this.getBackdropPo().waitForAbsent();
   }
 
-  async getTokens(amount: number): Promise<void> {
+  async getSnsTokens(amount: number): Promise<void> {
     await this.openMenu();
-    await this.getMenuItemsPo().getGetTokensPo().getTokens(amount);
+    await this.getMenuItemsPo().getGetTokensPo().getSnsTokensButton().waitFor();
```
dskloetd, line 139: Can you move the `getSnsTokensButton().waitFor()` inside the `getSnsTokens(amount)` function?

frontend/src/tests/e2e/sns-governance.spec.ts
lmuntaner: Yes, sorry. I was testing in an environment that was running low on tokens.

frontend/src/tests/page-objects/App.page-object.ts
lmuntaner: sure

lmuntaner:

PR #2979, opened 2023-07-26
Large user data

bitdivine:
# Motivation
We would like to verify that upgrades and downgrades work, both in the sense that they preserve account data and in that they do not consume too many cycles and so fail.

# Changes
* Add Rust code to create many toy accounts.
* Add Rust code to get a single account; this is used to verify that individual accounts are as expected.
* Expose the two new Rust methods as canister API calls in test builds.
* Add a large number of accounts to the existing nns-dapp downgrade upgrade test.

# Tests
- There is a unit test to confirm that the account creation data creates the requested number of accounts.
- The downgrade-upgrade test is the goal of this PR.

# Todos


rs/backend/Cargo.toml
```diff
@@ -59,5 +59,6 @@ anyhow = "1.0.68"
```
bitdivine, line 59: Here we define toy data as a feature and enable it for test builds.

rs/backend/nns-dapp-exports-test.txt
```diff
@@ -6,15 +6,17 @@ canister_pre_upgrade
 canister_query get_account
```
bitdivine, line 6: This is list of functions that should be exposed by the test build.  This PR proposes to add two functions:  One to create lots of toy accounts, one to get a specific toy account so that we can do in depth spot checks that accounts are preserved correctly across upgrades.

Note: I have intentionally _not_ added these methods to the .did file.  The .did file is complicated in various ways that I can elaborate on and the API methods are simple enough that signatures are not needed at this time.  I am sure that we will have to find a nice solution to the `.did` file, be it a separate file or the current file with additional methods.  Either way, the .did files would contain some hitherto internal types (!) and there would have to be big labels saying these methods and types are available for test only and may change at any time.

rs/backend/src/accounts_store/toy_data.rs
```diff
@@ -0,0 +1,93 @@
...
+                let attach_canister_request = AttachCanisterRequest {
+                    name: format!("canister_{toy_account_index}_{canister_index}"),
+                    canister_id,
+                };
+                self.attach_canister(account, attach_canister_request);
+            }
+        }
+        index_range_start
+    }
+
```
bitdivine, line 67: This lets us get a toy account by index, allowing us to get all the data for an account and check whether it has changed in an upgrade.  If the account struct changes as part of a schema migration, the upgrade check would have to take that into account by asking for the new data in the old schema.  The structure for that is not implemented in this PR.

rs/backend/src/accounts_store/toy_data.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// Creates an account sstore with the given number of test accounts.
+    #[cfg(test)]
+    pub fn with_toy_accounts(num_accounts: u64) -> AccountsStore {
+        let mut accounts_store = AccountsStore::default();
+        accounts_store.create_toy_accounts(num_accounts);
+        accounts_store
+    }
+}
+
```
bitdivine, line 87: This unit test checks that the toy accounts are created.  If we want more detailed checks, e.g. to ensure that the right number of subaccounts are created for each account, this would be the place to add it.

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -3,7 +3,8 @@ set -euo pipefail
 SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
```
bitdivine, line 3: Here we get to use the new test API methods in the downgrade upgrade test!

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -76,8 +111,15 @@ else
   dfx canister create nns-dapp
   dfx canister install nns-dapp --wasm "$CURRENT_WASM" --argument "$(cat nns-dapp-arg-local.did)"
 fi
+scripts/dfx-canister-check-wasm-hash --wasm "$CURRENT_WASM" --canister nns-dapp
 echo "Checking that the current wasm is healthy..."
 verify_healthy
+echo "Installing state"
+while (("$(dfx canister call nns-dapp get_stats | idl2json | jq -r .accounts_count)" < NUM_TOY_ACCOUNTS)); do
```
bitdivine, line 118: Note: If lots of accounts have already been created there is no need to create more and doing so can cause problems if it makes the state too large.  That is why this loop ensures that the number of requested accounts exists, rather than adding that many accounts.

.github/workflows/build.yml
```diff
@@ -35,6 +35,12 @@ jobs:
           name: nns-dapp
```
bitdivine, line 35: These changes are in #2984

rs/backend/src/accounts_store/toy_data.rs
```diff
@@ -0,0 +1,93 @@
...
+                self.create_sub_account(account, format!("sub_account_{toy_account_index}_{subaccount_index}"));
+            }
+            // Creates linked hardware wallets:
+            // Note: Successive accounts have 0, 1, 2 ... MAX_HARDWARE_WALLETS_PER_ACCOUNT-1 hardware wallets, restarting at 0.
+            for hardware_wallet_index in 0..(toy_account_index % (MAX_HARDWARE_WALLETS_PER_ACCOUNT + 1)) {
+                self.register_hardware_wallet(
+                    account,
+                    RegisterHardwareWalletRequest {
+                        name: format!("hw_wallet_{toy_account_index}_{hardware_wallet_index}"),
+                        principal: account,
```
lmuntaner, line 43: This is supposed to be the HW principal, but I guess it doesn't matter for testing purposes.

rs/backend/src/accounts_store/toy_data.rs
```diff
@@ -0,0 +1,93 @@
...
+                    },
+                );
+            }
+            // Creates pending transactions; these do not translate into neurons or similar without more steps:
+            for pending_transaction_index in 0..(toy_account_index % (MAX_PENDING_TRANSACTIONS_PER_ACCOUNT + 1)) {
```
lmuntaner, line 48: I would remove this, we don't use this anymore and it should be cleaned.

lmuntaner (commented):
A few comments. And a generic question:

Should we change the test to start with the production wasm, add the toy accounts then test the upgrade, create more toy accounts and test the rollback?

If I'm not mistaken, this tests only that the rollback will keep the data created, doesn't it?

bitdivine:
Thank you for your review, @lmuntaner 

> If I'm not mistaken, this tests only that the rollback will keep the data created, doesn't it?
This tests state preservation on every upgrade, so both in the forwards direction and back again.  The `upgrade_nnsdapp` is called twice; the second call is not easy to see as it is in a section of code that is unchanged.  To see it, expand the lines at the bottom of the downgrade upgrade test.

> Should we change the test to start with the production wasm
At present the production wasm doesn't have the ability to create lots of users but once prod has a dev flavour including this code, it could make sense.

rs/backend/src/accounts_store/toy_data.rs
bitdivine: Done.

rs/backend/src/accounts_store/toy_data.rs
bitdivine: Done

lmuntaner (approved):
Thanks for the explanation! LGTM!

PR #2984, opened 2023-07-27
Dedicated downgrade upgrade job

bitdivine:
# Motivation
The downgrade-upgrade job is about to get bigger and slower.  Bundling the downgrade-upgrade test along with lots of small cheap tests in "test-rest" is therefore no longer appropriate.

# Changes
* Move the nns-dapp downgrade-upgrade test into a dedicated job.
* Use the `_test` build so that a large state can be created by the downgrade-upgrade test.

# Tests
* See CI

# Todos


.github/workflows/build.yml
```diff
@@ -77,6 +83,30 @@ jobs:
...
+          name: nns-dapp_test
+      - name: Get sns_aggregator_dev
+        uses: actions/download-artifact@v3
+        with:
+          name: sns_aggregator_dev
+      - name: Start snapshot environment
+        uses: ./.github/actions/start_dfx_snapshot
+        with:
+          nns_dapp_wasm: 'nns-dapp_test.wasm.gz'
+          sns_aggregator_wasm: 'sns_aggregator_dev.wasm.gz'
```
dskloetd, line 105: Does the downgrade-upgrade test use the aggregator in any way?
If not, you can remove this line and remove lines 97-100.

bitdivine: Good point.  Done.

dskloetd (approved):

PR #2986, opened 2023-07-27
Noassets feature flag

bitdivine:
# Motivation
There is a TODO to consolidate the way different flavours of nns-dapp are built by generating the noassets build from a Rust feature flag.  Most of the infrastructure for flavours implemented as features was done for the `_test` flavour so this is mostly negative code.

# Changes
* Delete the Docker stage for the noassets build.
* Define a Rust feature flag that controls whether assets are included in the wasm.
* Add noassets to the list of build flavours.

# Tests
* Deployed the noassets build to an app subnet; it [was installed correctly](https://github.com/dfinity/nns-dapp/actions/runs/5680641129) and the UI, uploaded after installation, appeared as expected.

# Todos


rs/backend/Cargo.toml
```diff
@@ -58,6 +58,10 @@ maplit = "1.0.2"
 anyhow = "1.0.68"
 
 [features]
-production=[]
-test=[]
-# TODO: Define noassets as a feature flag, to keep different build flavours consistent.
+# Build flavours; these should not normally be referred to directly in code.
```
lmuntaner, line 61: Should we add that this changed in the changelog?

lmuntaner (approved):
Thanks! Already approving.

I added a comment to prevent from merging without noticing the question, but it's not a blocker.

rs/backend/Cargo.toml
bitdivine: Yes, can do.  It seemed marginal to me whether this would qualify for a changelog entry as it changes little for developers and nothing for users but I'm happy to add it.  Done.

lmuntaner: Thanks!

PR #2988, opened 2023-07-27
E2e/filter proposals by topic

mstrasinskis:
# Motivation

Filter proposals by topic playwright test (part of `user-V02-proposal-filters-work.e2e`).

# Changes

- Checkbox po
- isChecked to page-object
- FilterModal po
- NnsProposalFilters po
- proposals.spec / filter by topic

# Tests

Yes

# Todos

Will be added when the whole spec is done.


frontend/src/lib/components/proposals/NnsProposalsFilters.svelte
```diff
@@ -44,54 +45,56 @@
...
-    totalFilters={totalFiltersProposalRewardStatus}
-    activeFilters={rewards.length}
-    on:nnsFilter={() =>
-      openModal({
-        category: "rewards",
-        filters: ProposalRewardStatus,
-        selectedFilters: rewards,
-      })}>{$i18n.voting.rewards}</FiltersButton
-  >
+<TestIdWrapper testId="nns-proposals-filters-component">
```
mstrasinskis, line 48: Only `TestIdWrapper` was added.

frontend/src/lib/modals/common/FilterModal.svelte
```diff
@@ -73,8 +73,11 @@
     {#if filters}
       <div class="filters">
         {#each filters as { id, name, checked } (id)}
-          <Checkbox inputId={id} {checked} on:nnsChange={() => onChange(id)}
-            >{name}</Checkbox
+          <Checkbox
+            testId={`checkbox-${id}`}
```
mstrasinskis, line 77: So generic id `checkbox` was used because the modal is used for all filters.

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+  await appPo
+    .getProposalsPo()
+    .getNnsProposalFiltersPo()
+    .setTopicFilter([Topic.ExchangeRate]);
+  await appPo.getProposalsPo().getNnsProposalListPo().waitForContentLoaded();
+
+  expect(await getVisibleCardTopics()).toEqual(["Exchange Rate"]);
+
+  // Reset topic filter
+  await appPo.getProposalsPo().getNnsProposalFiltersPo().setTopicFilter([]);
```
mstrasinskis, line 44: `[]` by definition works as all selected.

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+  await appPo.getProposalsPo().getNnsProposalListPo().waitForContentLoaded();
+  expect((await getVisibleCardTopics()).length).toBeGreaterThanOrEqual(
```
dskloetd, line 46: If I understand correctly, we don't know exactly what proposals there are and whether all of them match the default filter. And the default filter is not the same as no filter. So we don't know if the initial visible cards include all topics and we don't know if the empty filter includes all topics. So you can't be more precise in your expectation here?

I'm not convinced of the usefulness of these imprecise expectations.

Can you do something more specific and have more specific expectations?
Maybe select all except ExchangeRate and then expect ExchangeRate not to be among the topics or something?

frontend/src/tests/page-objects/Checkbox.page-object.ts
```diff
@@ -0,0 +1,44 @@
...
+      await element.querySelectorAll(CheckboxPo.COMPONENT_SELECTOR)
+    ).map((el) => new CheckboxPo(el));
+  }
+
+  toggle(): Promise<void> {
+    return this.root.click();
+  }
+
+  getCheckboxInput(): PageObjectElement {
+    return this.root.byTestId("checkbox");
```
dskloetd, line 34: I'm thinking it might be more obvious that it's the actual `<input>` element if we do `this.root.querySelector("input[type=checkbox]")`.
What do you think?

frontend/src/tests/page-objects/NnsProposalFilters.page-object.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  clickFiltersByStatusButton(): Promise<void> {
+    return this.click("filters-by-status");
+  }
+
+  getFilterModalPo(): FilterModalPo {
+    return FilterModalPo.under(this.root);
+  }
+
+  async preselectEntriesInFilterModal(testIds: string[]): Promise<void> {
```
dskloetd, line 25: I'm trying to understand the name. Is `preselect` different from just `select`?

mstrasinskis: Renamed.

frontend/src/tests/page-objects/Checkbox.page-object.ts
mstrasinskis: Updated.

frontend/src/tests/e2e/proposals.spec.ts
mstrasinskis: Good idea, let's do it!

Jfyi, the current demo proposals are known, but it can be extended later.
image

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+  step("Open proposals list");
+  await appPo.goToProposals();
+  await appPo.getProposalsPo().getNnsProposalListPo().waitForContentLoaded();
+
+  step(`Filter proposals by Topic`);
+  const getVisibleCardTopics = () =>
+    appPo.getProposalsPo().getNnsProposalListPo().getCardTopics();
+  const initialVisibleCardTopics = await getVisibleCardTopics();
+
+  expect((await getVisibleCardTopics()).length).toBeGreaterThan(1);
```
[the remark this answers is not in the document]
mstrasinskis: The const was removed.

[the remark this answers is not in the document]
mstrasinskis: Because `createDummyProposal` creates proposals w/ 7 different topics.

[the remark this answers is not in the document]
dskloetd: If we rely on that, shall we expect it to be equal to 7?
Or we can expect it to be at least 1 just to know that there is something but not make additional assumptions?

dskloetd (approved):

mstrasinskis:
> If we rely on that, shall we expect it to be equal to 7?
Or we can expect it to be at least 1 just to know that there is something but not make additional assumptions?

After latest changing (only Exchange Rate vs w/o Exchange Rate), this expectation (`expect((await getVisibleCardTopics()).length).toBeGreaterThan(1);`) is redundant.

PR #2989, opened 2023-07-27
GIX-1687: Set ENABLE_NEURON_SETTINGS true except prod

lmuntaner:
# Motivation

To ensure that all current and new tests use the new neuron details UI.

# Changes

* Change `ENABLE_NEURON_SETTINGS` flag to true. Override to `false` in prod and staging.

# Tests

* Remove the unnecessary overriding the flag in e2e tests.

# Todos

Covered already by a changelog entry.


lmuntaner:
@dskloetd please review

dskloetd:
What is the reason for this?

lmuntaner:
> What is the reason for this?

To ensure that all current and new tests use the new neuron details UI.

PR #2993, opened 2023-07-27
refactor: Move the stable memory parsing code.

bitdivine:
# Motivation
We wish to be able to change the schema used to store data.  For this, the post_upgrade hook will need to be able to read data in the format of the last production canister and in the format of the next canister (to support rollbacks).

For this we need to determine the schema and branch on that, post upgrade.

# Changes
* Move the stable memory parsing routines into state.rs.
* Define a function that returns the schema (currently always "none").

# Tests
* CI should verify that state is preserved across downgrades and upgrades.
* The format is unchanged so cycles and memory usage should be unchanged.

# Todos


rs/backend/src/state.rs
```diff
@@ -58,3 +59,40 @@ impl StableState for State {
         })
     }
 }
+
+// Methods called on pre_upgrade and post_upgrade.
+impl State {
```
bitdivine, line 64: Here we need to determine whether the schema in stable memory is from the last schema (upgrade from an older schema), the same schema as that saved by this PR (upgrade to the same schema) or the next schema (downgrade from the next schema.  Yes, we need to deploy the parser for the next schema before we deploy the next schema if we want to be able to roll back, so schema changes take two deployments).

rs/backend/src/state.rs
```diff
@@ -58,3 +59,52 @@ impl StableState for State {
...
+    /// - The previous schema, when first migrating from the previous schema to the current schema.
+    /// - The curent schema, if upgrading without changing the schema.
+    /// - The next schema, if a new schema was deployed and we need to roll back.
+    ///
+    /// Note: Changing the schema requires at least two deployments:
+    /// - Deploy a relase with a parser for the new schema.
+    /// - Then, deploy a release that writes the new schema.
+    /// This way it is possible to roll back after deploying the new schema.
+    pub fn post_upgrade() -> Self {
+        match Self::schema_version_from_stable_memory() {
```
dskloetd, line 81: In `post_upgrade` wouldn't the schema version have to be read from stable memory rather than returned by a pure function?

rs/backend/src/state.rs
```diff
@@ -58,3 +59,52 @@ impl StableState for State {
...
+            None => Self::post_upgrade_unversioned(),
+            Some(version) => {
```
dskloetd, line 83: I must be missing something but this seems like dead code given that `schema_version_from_stable_memory()` always returns `None`.
It's not possible to be calling the `schema_version_from_stable_memory()` of a different wasm from here, right?

rs/backend/src/state.rs
```diff
@@ -58,3 +59,52 @@ impl StableState for State {
         })
     }
 }
+
+// Methods called on pre_upgrade and post_upgrade.
+impl State {
+    /// The schema version, determined by the last version that was saved to stable memory.
+    fn schema_version_from_stable_memory() -> Option<u32> {
+        None // The schema is currently unversioned.
```
dskloetd, line 67: Might we as well call this version 0 and keep the return type simple as (non-optional) `u32`?

rs/backend/src/state.rs
```diff
@@ -58,3 +59,52 @@ impl StableState for State {
...
+        }
+    }
+    /// Save any unsaved state to stable memory.
+    pub fn pre_upgrade(&self) {
+        self.pre_upgrade_unversioned()
+    }
+}
+
+// The unversionsed schema.
+impl State {
```
dskloetd, line 96: Is this the same namespace as is closed on line 93. Why not just leave it open (i.e. remove line 93-96)?

rs/backend/src/state.rs
lmuntaner: Will you add the functionality to read the version before migrating to Stable Structures?

Could you specify the steps and tasks on how you want to accomplish this in the document?

https://docs.google.com/document/d/11mpfkx6eXL4g7M8HgJyJlDpDlzR3FBKyeryXlzLqm8s/edit?usp=sharing

I added how we can't do it, and I'd like to understand your plan.

lmuntaner (commented):
I'm sorry, but I still don't fully understand the migration plan.

I added some flow at the end of the goole docs: https://docs.google.com/document/d/11mpfkx6eXL4g7M8HgJyJlDpDlzR3FBKyeryXlzLqm8s/edit?usp=sharing

The flow I added is how we shouldn't do it because canisters are not compatible. Could you explain there how you're planning to do it?

Maybe it's better a call instead of writing it down. Up to you.

rs/backend/src/state.rs
bitdivine: Yes, sure.  I already have some code but I can describe it in the document.

bitdivine: Here is how the schema version is obtained from stable memory: https://docs.google.com/document/d/11mpfkx6eXL4g7M8HgJyJlDpDlzR3FBKyeryXlzLqm8s/edit#heading=h.jss126cvym09

Note: I originally planned to use a 4 byte checksum but the canister already has SHA256 but not a CRC32.  A CRC32 _can_ be implemented in just two x86 CPU instructions but I cannot guarantee how many bytes some library is going to take and we are at 99% of memory usage.

bitdivine: Here is how roll forward/back works: https://docs.google.com/document/d/11mpfkx6eXL4g7M8HgJyJlDpDlzR3FBKyeryXlzLqm8s/edit#heading=h.qkzzpyaglm6i

Key here is that postupgrade can support multiple schemas.  As long as we have a reliable way of reading a schema version from stable memory, which we do, we can define additional postupgrade hooks for every schema we expect to want to roll forwards to or back from.

rs/backend/src/state.rs
bitdivine: We might but I would rather not.  I like the simplicity of version == first 4 bytes, confirmed by a SHA256, anything else is an error.  It's just that we start off expecting the error response! :-)

rs/backend/src/state.rs
bitdivine: Which pure function?  `Self::schema_version_from_stable_memory()` is not a pure function as it, in general, reads from stable memory, as the name implies.  I'm probably missing your point. Maybe we can jump on a call and you can explain.

rs/backend/src/state.rs
bitdivine: `schema_version_from_stable_memory()` returns None _in this PR_.  But not always.  If you wish to peer into the future, of `post_upgrade()` I recommend reading this shiny new section of the design doc: [Preupgrade and postupgrade schema support](https://docs.google.com/document/d/11mpfkx6eXL4g7M8HgJyJlDpDlzR3FBKyeryXlzLqm8s/edit#heading=h.qkzzpyaglm6i)

rs/backend/src/state.rs
bitdivine: In Rust one can have several `impl SomeStruct` paragraphs.  It helps readability and organisation to have related code together.  I expect to have one `impl` section per schema.  I considered having a separate file for every schema; at present `state.rs` is not big enough to warrant that.

rs/backend/src/state.rs
bitdivine: I do not want to declare now which PRs I want to make though.  That would be too fine-grained detail and I don't have perfect knowledge of the problems I will encounter.  Is it clear enough how this PR leads towards the schema migration? described in [Preupgrade and postupgrade schema support](https://docs.google.com/document/d/11mpfkx6eXL4g7M8HgJyJlDpDlzR3FBKyeryXlzLqm8s/edit#heading=h.qkzzpyaglm6i)?

bitdivine: If it helps, I tend to prototype where I expect to go in the next few steps, then make a small PR of the first of those steps.  I plan to test various detailed implementations of stable structures, see which seems simplest and least error prone, then pick that one.  So, for example, it seems to be common top use the first 4 or 8 bytes to store the memory length.  If it turns out that this is indeed really useful under enough scenarios, I may offset the version by 8 bytes to leave space for this length field.

lmuntaner: No need to declare how and which PRs. The layout you added in the document is enough to understand where you'll be going. I added a question there.

rs/backend/src/state.rs
dskloetd: Right, so *in this PR* this is dead code.
I guess I'm just confused by the way you decided to split up the work.
I would either expect `schema_version_from_stable_memory()` to already read from stable memory, or this branch of the code to not be added in this PR already.
But this looks safe and small enough so I'll suspend my confusion.

dskloetd (approved):

bitdivine:
I believe that all conversations are now settled either here, via the design dock or slack.

PR #2994, opened 2023-07-27
Make transactions in toy state

bitdivine:
# Motivation
We would like the downgrade-upgrade test to be realistic.  There are two large datasets in the state: The accounts an the transaction history.  The toy data already contains accounts but not transactions.  So let's add transactions.

# Changes
* Add a large number of toy transactions to the toy state.
* Update the downgrade-upgrade test to ensure that most transactions are preserved.
  * Note: Given that transactions are pruned then memory usage exceeds 1Gb, the number of transactions is not invariant but we should expect most transactions to be retained.

# Tests
* The number of transactions is no longer zero:
Screenshot from 2023-08-02 17-57-22

# Todos
- Future: Adjust the number of subaccounts, hardware accounts etc per toy account to reflect mainnet.  The ratios between test and mainnet should be close to 100%.
  - Already covered by "Increased the size of the persistent state in downgrade-upgrade tests".


scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -69,6 +73,28 @@ check_state_size() {
...
+# Verifies that the number of transactions is reasonable.  Some may be pruned, but not a lot.
+check_num_transactions() {
+  # The toy state creates 3 transactions per account; see TRANSACTIONS_PER_ACCOUNT
+  # in: rs/backend/src/accounts_store/toy_data.rs
+  toy_transactions="$((NUM_TOY_ACCOUNTS * 3))"
+  # Note: Transactions MAY be pruned if the state is large enough, however in tests
+  # we should not make the state so large that, say, half the transactions are pruned.
+  # If that many are pruned, either the test is unrealistic or we should increase the
+  # amount of storage available.
+  expected_min_transactions="$((toy_transactions / 2))"
```
lmuntaner, line 85: Should we really expect half only? At the moment, we expect that no transaction will be pruned because we haven't reached the limit.

I wonder whether we should expect that in the test.

lmuntaner (approved):
LGTM! Just a minor comment

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: The test is over the limit so this does already kick in.  I did consider noting the time pre-upgrade and the time of this check, determine the number of blocks that had passed and so determine the maximum that should have been pruned.  It is complicated by the fact that the block time in CI is not necessarily fixed.  We generally want CI to run fast and the block rate in CI can be fast.

We can reduce the size of the state so that it is in the window slightly larger than prod and slightly below 1G.  The number of subaccounts in CI is currently 8x that in production.  If that is dropped to 1.2x or so, we will probably be in that gap.

I will assume that this is non-blocking and will move to merge but I'm happy to make adjustments based on the outcome of this discussion.

PR #2996, opened 2023-07-28
GIX-1777: Communicate when No Voting Power

lmuntaner:
# Motivation

Communicate better when a neuron can't vote.

In this PR: Improve the subtitle of the new neuron details page when neuron has no voting right and also the value in the voting power section.

# Changes

* Check whether neuron can vote and show a related message in NnsNeuronPageHeader and SnsNeuronPageHeading.
* Check whether neuron can vote and show a related text in NnsNeuronVotingPowerSection and SnsNeuronVotingPowerSection.

# Tests

* Test new cases in the page heading subtitle.
* Test new cases in the voting power section.

# Todos

Covered by changelog entry related to new neuron details UI.

### DIFF
--- a/frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
+++ b/frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
@@ -6,6 +6,7 @@
   import { replacePlaceholders } from "$lib/utils/i18n.utils";
   import { i18n } from "$lib/stores/i18n";
   import PageHeading from "../common/PageHeading.svelte";
+  import { NNS_MINIMUM_DISSOLVE_DELAY_TO_VOTE } from "$lib/constants/neurons.constants";
 
   export let neuron: NeuronInfo;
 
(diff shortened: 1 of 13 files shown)


frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
```diff
@@ -14,13 +15,22 @@
     amount: neuronStake(neuron),
     token: ICPToken,
   });
+
+  // NNS neurons come with voting power but that doesn't mean they can vote.
```
dskloetd, line 19: Do you mean "The API might return a non-zero voting power even if the neuron can't vote."?
The problem is with the API/backend, right? The neuron doesn't actually have voting power (after all, it can't vote). It's just not properly indicated by the API.
Or do I see it wrong?

If I understand it correctly, I would find this comment easier to understand:
```
// The API might return a non-zero voting power even if the neuron can't vote.
```

frontend/src/lib/components/neuron-detail/NnsNeuronVotingPowerSection.svelte
```diff
@@ -16,12 +17,21 @@
     amount: neuronStake(neuron),
     token: ICPToken,
   });
+
+  // NNS neurons come with voting power but that doesn't mean they can vote.
```
dskloetd, line 21: same, if you agree

dskloetd (approved):

frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
lmuntaner: Correct. I'll change it.

frontend/src/lib/components/neuron-detail/NnsNeuronVotingPowerSection.svelte
lmuntaner: done

PR #2999, opened 2023-07-28
Fix back to accounts after transaction in wallet page

peterpeterparker:
# Motivation

Refactoring the navigation spinner in #2962 within the component that prevents SvelteKit issue [#5434](https://github.com/sveltejs/kit/issues/5434) actually also lead to the same issue of duplicating the `split-pane` in the DOM. Therefore we have to undo the refactor and keep the spinner within the `+layout` of the sign-in page.

# Reproduce

- Open the wallet (e.g. of the main account)
- Make a transaction
- Click the back button
- Where the menu button should now reappear, there is only a non-functional back button

# Changes

- revert moving spinner from login layout to navigation guard
   - remove spinner from nav component
   - redo spinner in +layout of login (same code as on mainnet)

### DIFF
--- a/frontend/src/lib/components/layout/LayoutNavGuard.svelte
+++ b/frontend/src/lib/components/layout/LayoutNavGuard.svelte
@@ -1,15 +1,9 @@
 <script lang="ts">
   import { isNullish } from "@dfinity/utils";
   import { navigating } from "$app/stores";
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
Thank you! I failed to reproduce the bug 👍

PR #2475, opened 2023-05-10
Vitest

peterpeterparker:
# Motivation

Replaces jest by vitest.

# Changes

# Tests

PR #3000, opened 2023-07-31
Expose component props setter and getter

peterpeterparker:
# Motivation

Vitest (PR #2475) requires component's properties exposed as setter and getter if props are read within the test which is the case of `<CkBTCTransactionsList />`.


mstrasinskis (approved):
Nice!

PR #3001, opened 2023-07-31
Move dispatch intersecting to events utils

peterpeterparker:
# Motivation

Vitest (PR #2475) seems to be not able to partially mock and partially not mock a module which leads the test of the intersection directive to fail. So this PR moves the dispatcher to a separate modules which is fundamently not incorrect as we already got a dedicated utility modules that is responsible for dispatching event.

# Changes

- move `dispatchIntersecting` to `events.utils`


mstrasinskis (approved):
Thanks!

PR #3003, opened 2023-07-31
GIX-1777: Hotkey and NF's tags

lmuntaner:
# Motivation

Identify hotkey neurons and those belonging to Neurons' Fund.

This is already a feature in the current neuron details page. In this PR, it adds the same functionality in the new neuron details page.

# Changes

* New UI component HeadingTag.
* Use new HeadingTag in NnsNeuronPageHeading to render a tag when neuron is part of Neurons' Fund or the user is a hotkey.
* Use new HeadingTag in SnsNeuronPageHeading to render a tag when the user is a hotkey of the neuron.

# Tests

* Test new functionality in NnsNeuronPageHeading and SnsNeuronPageHeading.

# Todos

Covered already by an entry in the changelog.

### DIFF
--- a/frontend/src/lib/components/common/HeadingTag.svelte
+++ b/frontend/src/lib/components/common/HeadingTag.svelte
@@ -1,0 +1,15 @@
+<script lang="ts">
+  export let testId: string | undefined = undefined;
+</script>
+
+<span class="description" data-tid={testId}>
+  <slot />
+</span>
+
+<style lang="scss">
+  span {
+    background-color: var(--focus-background-tint);
+    padding: var(--padding);
+    border-radius: var(--border-radius-0_5x);
+  }
+</style>
--- a/frontend/src/lib/components/common/PageHeading.svelte
+++ b/frontend/src/lib/components/common/PageHeading.svelte
@@ -2,7 +2,7 @@
   export let testId: string | undefined = undefined;
 
   let hasTags: boolean;
-  $: hasTags = !!$$slots.tag;
+  $: hasTags = $$slots.tags !== undefined;
 </script>
 
 <div class="container" data-tid={testId}>
@@ -12,7 +12,7 @@
   </h4>
   {#if hasTags}
     <div class="tags">
-      <slot name="tag" />
+      <slot name="tags" />
     </div>
   {/if}
 </div>
--- a/frontend/src/lib/components/common/PageHeading.svelte
+++ b/frontend/src/lib/components/common/PageHeading.svelte
@@ -7,6 +7,7 @@
   <h4 class="description">
     <slot name="subtitle" />
   </h4>
+  <slot name="tag" />
 </div>
 
 <style lang="scss">
--- a/frontend/src/lib/components/common/PageHeading.svelte
+++ b/frontend/src/lib/components/common/PageHeading.svelte
@@ -1,5 +1,8 @@
 <script lang="ts">
   export let testId: string | undefined = undefined;
+
+  let hasTags: boolean;
+  $: hasTags = !!$$slots.tag;
 </script>
 
 <div class="container" data-tid={testId}>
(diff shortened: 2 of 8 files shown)


lmuntaner:
New UI:

Screenshot 2023-07-31 at 12 13 49
Screenshot 2023-07-31 at 12 13 56

lmuntaner:

Still pending to be discussed with design whether we want to move the HeadingTag to the gix-components library.

frontend/src/lib/components/common/HeadingTag.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
```
peterpeterparker, line 1: Why not using the `Tag` component?
If it's a new component, why not implementing it in gix-cmp?
If it's a variation of the `Tag` component, can the `Tag` component be reworked so all tag look the same for consistency reason?

frontend/src/lib/components/common/PageHeading.svelte
```diff
@@ -7,6 +7,7 @@
   <h4 class="description">
     <slot name="subtitle" />
   </h4>
+  <slot name="tag" />
```
peterpeterparker, line 10: why do you not implement the `<div class="tags">` container here instead of passing it as a slot?

peterpeterparker (commented):
The design look like this in Figma right?

Those "tags" appear to resemble to disabled buttons, which is in my opinion confusing, so just want to be sure it's the designer's intention.

frontend/src/lib/components/common/PageHeading.svelte
lmuntaner: Good point! I'll do it. It makes sense.

frontend/src/lib/components/common/HeadingTag.svelte
lmuntaner: Good questions. I didn't see it as a tag initially when I saw the design and I didn't compare it to our current tag.

But since I named it HeadingTag, it makes a lot of sense to bring these questions to design.

lmuntaner: @peterpeterparker design is reviewing the tags and there will be a refactor and this "tag" shuold be combined with the next.

Yet, for now, we leave this tag in NNS Dapp as is.

peterpeterparker: Oki doki

frontend/src/lib/components/common/PageHeading.svelte
```diff
@@ -1,12 +1,20 @@
 <script lang="ts">
   export let testId: string | undefined = undefined;
+
+  let hasTags: boolean;
+  $: hasTags = !!$$slots.tag;
```
peterpeterparker, line 5: In the codebase we have used `$$slots.details !== undefined` so far, would it be better with `!!$$slots.tag;` - i.e. should we change the other similar checks afterwards?

frontend/src/lib/components/common/PageHeading.svelte
```diff
@@ -1,12 +1,20 @@
...
 </script>
 
 <div class="container" data-tid={testId}>
   <slot name="title" />
   <h4 class="description">
     <slot name="subtitle" />
   </h4>
+  {#if hasTags}
+    <div class="tags">
+      <slot name="tag" />
```
peterpeterparker, line 15: should we rename the slot to plural to reflect the fact that it can receive multiple tags?

peterpeterparker (approved):
LGTM, thx

frontend/src/lib/components/common/PageHeading.svelte
lmuntaner: sure, makes sense!

peterpeterparker (approved):
LGTM, thx

PR #3005, opened 2023-07-31
Improve ux by using hyperlinks for the cards

peterpeterparker:
# Motivation

Accounts, neurons, proposals, launchpad and canisters cards are button which is a bit annoying as it makes complicated for the users to work with multiple tabs. By modifying these cards to become real hyperlinks, we will allow the user to open more easily information in separate tabs.

# PRs

- [x] https://github.com/dfinity/gix-components/pull/262
- [x] https://github.com/dfinity/gix-components/pull/267
- [x] https://github.com/dfinity/gix-components/pull/266
- [x] https://github.com/dfinity/gix-components/pull/265
- [x] https://github.com/dfinity/gix-components/pull/264
- [x] https://github.com/dfinity/gix-components/pull/263

# Changes

- provide new property `href` instead of using `on:click` to navigate through the cards from pages to details page
- review `role` of the cards

# Screenshots

demo

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -771,9 +771,9 @@
       }
     },
     "node_modules/@dfinity/gix-components": {
-      "version": "2.6.0-next-2023-08-01.2",
-      "resolved": "https://registry.npmjs.org/@dfinity/gix-components/-/gix-components-2.6.0-next-2023-08-01.2.tgz",
-      "integrity": "sha512-QIUnzaItCRsRWiXbnmsZKT0mA8e1YWT+j3PwoYcDvch5+d+9w7+gYAlCJUY9L+4Nl/olQRwkAAwHEYyN7fqLrg==",
+      "version": "2.6.0-next-2023-08-02.4",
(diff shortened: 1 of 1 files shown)


frontend/src/lib/components/launchpad/ProjectCard.svelte
```diff
@@ -35,16 +34,13 @@
...
-    );
+  const buildProjectHref = (): string =>
+    `${AppPath.Project}/?project=${project.rootCanisterId.toText()}`;
 </script>
 
 <Card
   testId="project-card-component"
-  role="link"
-  on:click={showProject}
+  href={buildProjectHref()}
```
mstrasinskis, line 43: It looks like that with this approach, the href will be obsolete after the `project` prop update. If so, is it relevant for the current usage?

mstrasinskis (approved):
Thanks, just a single question.

frontend/src/lib/components/launchpad/ProjectCard.svelte
peterpeterparker: Excellent point 👍. Updated the code to use auto subscriber for those few.

PR #3006, opened 2023-08-01
Display page title in browser's tab / window title

peterpeterparker:
# Motivation

Displaying a page title within the browser's tab / window title can be particularly useful if multiple tabs are open and generally speaking a bit more user friendly.

# Note

Discussed with Artem, for this first iteration we only want to display static or public information within the tab's title. For example we do not display neuron id or account identitifer.

# Changes

- extend `layoutTitleStore` to support a title for the browser's tab and the page header
- implement new store object instead of string in existing pages
- in `Title` component render a title header tag if corrresponding information is set
- in `Title` component if no header is provided fallback on the title (which is mandatory and can be empty as currently)
- use `LayoutList` in Launch Pad and Proposals for consistency reason
- implement few titles setter that were missing

# Screenshots

Few examples:

Capture d’écran 2023-08-01 à 13 13 12
Capture d’écran 2023-08-01 à 13 13 08
Capture d’écran 2023-08-01 à 13 13 06
Capture d’écran 2023-08-01 à 13 13 02

### DIFF
--- a/frontend/src/tests/lib/components/header/Title.spec.ts
+++ b/frontend/src/tests/lib/components/header/Title.spec.ts
@@ -1,0 +1,54 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import Title from "$lib/components/header/Title.svelte";
+import { layoutTitleStore } from "$lib/stores/layout.store";
+import { render, within } from "@testing-library/svelte";
+
+describe("Title", () => {
+  beforeEach(() => layoutTitleStore.set({ title: "" }));
+
+  it("should render a title in the DOM header", () => {
+    const testTitle = "test test test";
+
+    layoutTitleStore.set({ title: testTitle });
+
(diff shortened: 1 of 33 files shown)


frontend/src/tests/lib/components/header/Title.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  beforeEach(() => layoutTitleStore.set({ title: "" }));
+
+  it("should render a title in the DOM header", () => {
+    const testTitle = "test test test";
+
+    layoutTitleStore.set({ title: testTitle });
+
+    render(Title);
+
+    const title = within(document.head).getByText(`NNS Dapp / ${testTitle}`);
```
mstrasinskis, line 19: Nice, TIL

mstrasinskis (approved):
Thank you!

PR #3007, opened 2023-08-01
Get development ICP with Icrc accounts

peterpeterparker:
# Motivation

We have a "Get ICP" feature we used for development purpose which does not transfer ICP anymore if the `ENABLE_ICP_ICRC` flag is activated.

This PR checks if the transfer destination address is an Icrc address and if so, uses the Icrc ledger definition to proceed.


mstrasinskis (approved):
Thank you!

PR #3008, opened 2023-08-01
Use standard accounts dropdown selector in canisters features

peterpeterparker:
# Motivation

For the sake of consistency and with the hope of eventually deprecating the component that renders the list of accounts for selection purposes, this PR introduces the dropdown accounts selector in the canisters' features.

Additionally, it fixes a UX quirk on the mainnet that was introduced with the addition of the "enter name" step in the creation wizard. The back action called "Change source account" now correctly forwards to the source account instead of the input field where the name can be entered.

Finally it improves the spacing and design of the modals making those more aligned and consistent with the rest of the dapp.

# Changes

- use select dropdown account instead of `NnsSelectAccount`
- add and pass down a filter to exclude hardware wallet
- style ui

# Screenshots

Capture d’écran 2023-08-07 à 16 57 29
Capture d’écran 2023-08-07 à 16 57 26
Capture d’écran 2023-08-07 à 16 57 18
Capture d’écran 2023-08-07 à 16 57 13


frontend/src/tests/lib/modals/canisters/AddCyclesModal.spec.ts
```diff
@@ -111,34 +93,46 @@ describe("AddCyclesModal", () => {
     );
   });
 
-  it("should top up a canister from ICP and close modal", async () => {
-    const { queryByTestId, queryAllByTestId, container, component } =
-      await renderModal({
-        component: AddCyclesModalTest,
-        props,
+  const selectAccount = ({ container, selectedAccount = mockMainAccount }) => {
```
lmuntaner, line 96: Don't you nee to move this to the top before using it?

lmuntaner (approved):
LGTM! Thanks! Just a minor comment

frontend/src/tests/lib/modals/canisters/AddCyclesModal.spec.ts
peterpeterparker: I did not get any warnings, maybe it's because it is declared within a `describe` context or just TypeScript that does the job? Anyway good question, moved it, thanks.

lmuntaner (approved):

PR #3016, opened 2023-08-02
Temporary disable macos-13 in reproducibility check

peterpeterparker:
# Motivation

Temporarily disable macos-13 in reproducibility check until we resolve the tar issue that leads to generating different hashes.


peterpeterparker:
Duplicate #3017.

PR #3017, opened 2023-08-02
Don't run reproducible assets against Mac OS 13

dskloetd:
# Motivation

The reproducible assets test is getting a different hash on Mac OS 13.
The full WASM build is already only reproducible on Linux, which is why we run it in Docker.

# Changes

Don't run reproducible assets against Mac OS 13

# Tests

CI

# Todos

### DIFF
--- a/.github/workflows/reproducible-assets.yaml
+++ b/.github/workflows/reproducible-assets.yaml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04, macos-12, macos-13]
+        os: [ubuntu-20.04, ubuntu-22.04, macos-12]
         time: [now, future]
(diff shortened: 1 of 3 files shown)


.github/workflows/reproducible-assets.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04, macos-12, macos-13]
+        os: [ubuntu-20.04, ubuntu-22.04, macos-12]
```
peterpeterparker, line 11: should we add a comment `# TODO: reactive macos-13`?

dskloetd: I don't know if it's important since the full WASM is already not reproducible on Mac as I mentioned in the PR description.

PR #3019, opened 2023-08-02
Invert tab title

peterpeterparker:
# Motivation

Follow best practice and invert tab titles. e.g. instead of `NNS Dapp / My Tokens` render `My Tokens / NNS Dapp`.


mstrasinskis (approved):
Nice!

PR #3020, opened 2023-08-02
Compare screenshots of the UI to check the design

peterpeterparker:
# Motivation

While we already compare screenshots in the gix-cmp to ensure that the design is the one we except, it is useful to proceed with some visual testing in NNS Dapp as well as we are making an intensive use of the components library and to ensure that the CSS style cascade we use for developing locally - when we are referencing the gix-cmp locally - is the same as when the packaged library is used.

# Changes

# Tests

# Todos


frontend/snapshots/design.spec.ts/Design-Signed-in-My-Tokens-wide-screen-1-firefox-darwin.png
dskloetd, line 1: FYI
This will fail once we have a new snapshot. I guess it's OK to update the screenshot when we have a new snapshot but I guess I should also look into having consistent names for the SNSs we generate.

.github/workflows/snapshots.yml
dskloetd, line 1: Do we need a separate workflow for this?
Or can we just add --update-snapshots to the existing npm script?

frontend/src/tests/e2e/design.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    let page: Page;
+
+    test.beforeAll(async ({ browser }) => {
+      page = await browser.newPage();
+
+      await page.goto("/");
+      await expect(page).toHaveTitle("NNS Dapp");
+
+      await waitForSignIn(page);
```
dskloetd, line 37: This seems redundant.`signInWithNewUser` will try to click on the button and it will wait if it's not there yet.

frontend/snapshots/design.spec.ts/Design-Signed-in-My-Tokens-wide-screen-1-firefox-darwin.png
dskloetd: Wait, this screenshot is generated against staging, isn't it? Does GitHub CI connect to staging? I don't see how.

peterpeterparker: This snapshot was generated for local development on MacOS. Once this PR merged, as described in "Notes", a new GitHub Actions will be made available which will allow us to generate the snapshots for the CI as rendered within the CI with linux.

Long story short, local snapshots (xxx-darwin.png) and ci snapshots (xxx-linux.png) .

And no, it won't use staging but, use local as the snapshots will be generated within the CI.

.github/workflows/snapshots.yml
peterpeterparker: Yes, we need separate GH Actions for this. Additionally, there is a separate npm script for it. Our intention is to always run the tests against existing snapshots, and we want to generate new snapshots only when we deliberately make changes, such as when we develop CSS modifications.

peterpeterparker: This is for example how you can add an e2e test in gix-cmp: https://github.com/dfinity/gix-components/blob/docs/correct-script/HACKING.md#visual-comparisons

frontend/src/tests/e2e/design.spec.ts
peterpeterparker: Ah cool. Removed. Thanks.

frontend/snapshots/design.spec.ts/Design-Signed-in-My-Tokens-wide-screen-1-firefox-darwin.png
dskloetd: I don't think I really understand how this will be used. Can you include some instruction/explanation in `frontend/src/tests/e2e/README.md`?

Also, I don't understand why you don't include correct screenshots in this PR. Can't you temporarily change the workflow so it will run already and give you the screenshots and you don't have to comment out part of the test.

peterpeterparker: > Can you include some instruction/explanation

Sure will do.

> Can't you temporarily change the workflow

Not sure how, doesn't GitHub Actions become active only once available on main?

I can provide the GitHub Actions in a separate PR though if you rather like.

.github/workflows/snapshots.yml
```diff
@@ -0,0 +1,27 @@
...
+        run: |
+          echo "This workflow should not be triggered with workflow_dispatch on main"
+          exit 1
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
+        run: npm ci
+        working-directory: ./frontend
+      - name: Run Playwright tests
+        run: npm run test-e2e:snapshots
```
dskloetd, line 19: Maybe I'm missing something, but I don't understand which environment this is running against.
If you want it to be running against the same environment as our current CI e2e tests, that needs to be set up first.
I recommend you have a look at the [test-e2e action](https://github.com/dfinity/nns-dapp/blob/main/.github/actions/test-e2e/action.yml) and the [build and test workflow](https://github.com/dfinity/nns-dapp/blob/fbd3a09ecec73e3fdb858a434c3de838acc8433d/.github/workflows/build.yml#L62).

I think it would probably make sense to add some arguments to the existing action, instead of having a separate npm script.

frontend/package.json
```diff
@@ -17,6 +17,7 @@
     "format:modified": "prettier --write $(git diff --diff-filter=ACM --name-only | egrep \"\\.(ts|js|mjs|json|scss|svelte|html)$\" | cut -d'/' -f2-)",
     "test": "TZ=UTC jest",
     "test-e2e": "tsc --noEmit -p src/tests/e2e/tsconfig.json && playwright test",
+    "test-e2e:snapshots": "playwright test --update-snapshots --reporter=list",
```
dskloetd, line 20: This only needs to run the `design.spec.ts` test, right?

dskloetd, line 20: Is `--reporter=list` necessary? It looks like that's the default we have in the config.

dskloetd:
>> Can't you temporarily change the workflow?

> Not sure how, doesn't GitHub Actions become active only once available on main?

Can you (temporarily?) change an existing workflow? The e2e tests are run in [this action](https://github.com/dfinity/nns-dapp/blob/main/.github/actions/test-e2e/action.yml). You could change
```
run: PLAYWRIGHT_BASE_URL="$(../scripts/dfx-canister-url nns-dapp)" npm run test-e2e -- $(ls src/tests/e2e/*.spec.ts | cat | split -n r/${{ inputs.shard_number }}/${{ inputs.shard_count }})
```
to
```
run: PLAYWRIGHT_BASE_URL="$(../scripts/dfx-canister-url nns-dapp)" npm run test-e2e -- --update-snapshots $(ls src/tests/e2e/*.spec.ts | cat | split -n r/${{ inputs.shard_number }}/${{ inputs.shard_count }})
```
and add steps to commit the changes?

frontend/snapshots/design.spec.ts/Design-Signed-in-My-Tokens-wide-screen-1-firefox-darwin.png
peterpeterparker: Documented in `HACKING.md`. Let me know if unclear.

frontend/package.json
peterpeterparker: Yes and no. Yes currently only `design.spec.ts` but, no as we plan to add other screenshot comparisons in the future (scoped to component or not). Therefore I rather like to not scope the command and let it generate any screenshots we would need.

peterpeterparker: It is not necessary, thanks. Removed.

peterpeterparker:
> Can you (temporarily?) change an existing workflow? The e2e tests are run in [this action](https://github.com/dfinity/nns-dapp/blob/main/.github/actions/test-e2e/action.yml). You could change

@dskloetd thanks for the input but, changing the command won't commit the screenshots in the branch right? The GitHub actions generate the screenshots but also commit those in the PR.

.github/workflows/snapshots.yml
peterpeterparker: Thanks! Absolutely.

peterpeterparker: That said we need another npm script and another gh actions (because different trigger).

peterpeterparker: I refactored the set up part of "test-e2e action" into a new "set-up-e2e action". That way we can reuse the setup easily.

peterpeterparker:
Improved and effectively implemented in #3051 by @dskloetd, therefore I close this PR.

PR #3027, opened 2023-08-03
Refactor: CommonItemActionPo

lmuntaner:
# Motivation

Share the PO among all the ItemAction components that use the CommonItemAction.

# Changes

* New PO CommonItemActionPo.
* All the others extend this one.
* Remove unnecessary `span` to access title with data-tid in all CommonItemAction implementations.
* Adapt all tests and instance creations to the new interface.
* New test for CommonItemAction.

# Tests

Only test changes.

# Todos

Covered by a changelog entry.

### DIFF
--- a/frontend/src/tests/lib/components/neuron-detail/NnsAvailableMaturityItemAction.spec.ts
+++ b/frontend/src/tests/lib/components/neuron-detail/NnsAvailableMaturityItemAction.spec.ts
@@ -30,9 +30,9 @@ describe("NnsAvailableMaturityItemAction", () => {
       },
     });
 
-    return NnsAvailableMaturityItemActionPo.under(
-      new JestPageObjectElement(container)
-    );
+    return NnsAvailableMaturityItemActionPo.under({
+      element: new JestPageObjectElement(container),
+    });
   };
 
   beforeEach(() => {
@@ -52,7 +52,7 @@ describe("NnsAvailableMaturityItemAction", () => {
     };
     const po = renderComponent(neuron);
 
-    expect(await po.getMaturity()).toBe("3.14");
+    expect(await po.getTitle()).toBe("3.14");
   });
 
   it("should render buttons", async () => {
--- a/frontend/src/tests/page-objects/NnsAvailableMaturityItemAction.page-object.ts
+++ b/frontend/src/tests/page-objects/NnsAvailableMaturityItemAction.page-object.ts
@@ -1,17 +1,17 @@
-import { BasePageObject } from "$tests/page-objects/base.page-object";
 import type { PageObjectElement } from "$tests/types/page-object.types";
+import { CommonItemActionPo } from "./CommonItemAction.page-object";
 
-export class NnsAvailableMaturityItemActionPo extends BasePageObject {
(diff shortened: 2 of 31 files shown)


frontend/src/tests/page-objects/NnsAvailableMaturityItemAction.page-object.ts
```diff
@@ -1,19 +1,19 @@
...
 import type { PageObjectElement } from "$tests/types/page-object.types";
+import { CommonItemActionPo } from "./CommonItemAction.page-object";
 
-export class NnsAvailableMaturityItemActionPo extends BasePageObject {
+export class NnsAvailableMaturityItemActionPo extends CommonItemActionPo {
   private static readonly TID = "nns-available-maturity-item-action-component";
 
-  static under(element: PageObjectElement): NnsAvailableMaturityItemActionPo {
+  static under({
+    element,
```
dskloetd, line 8: Why did you change the argument to an object while it's still only 1 argument?

(Same for the other POs.)

frontend/src/tests/lib/components/neuron-detail/NnsAvailableMaturityItemAction.spec.ts
```diff
@@ -52,7 +52,7 @@ describe("NnsAvailableMaturityItemAction", () => {
     };
     const po = renderComponent(neuron);
 
-    expect(await po.getMaturity()).toBe("3.14");
+    expect(await po.getTitle()).toBe("3.14");
```
dskloetd, line 55: I'm not so sure about having the method to get the maturity be called `getTitle()`. I might suggest adding back a `getMaturity` method which just returns `this.getTitle()` but then the benefit of the share base class becomes very limited.

I originally suggested merging the Nns and Sns page objects of the same item. I'm not so convinced about `CommonItemActionPo`.

(This comment applies to all the other `getXXX()` methods as well.)

frontend/src/tests/page-objects/NnsNeuronStateItemAction.page-object.ts
dskloetd, line 1: Should `getState()` now return `this.getTitle()` or be removed?

dskloetd (commented):
I'm not sure I like this. It looks like a big change with limited benefit and increased complexity. Is there a follow-up planned that will make it worth it?

frontend/src/tests/page-objects/NnsAvailableMaturityItemAction.page-object.ts
lmuntaner: Because it extends from `CommonItemActionPo` which expects both testId and element.

frontend/src/tests/lib/components/neuron-detail/NnsAvailableMaturityItemAction.spec.ts
lmuntaner: I agree with you that it doesn't add much value.

I don't agree that we should merge POs if they don't map to the component structure.

lmuntaner:
This PR and changes don't add as much value as we thought they would. Closing it.

PR #3028, opened 2023-08-03
Rename deleted workflows

dskloetd:
# Motivation

The GitHub Actions tab still lists workflows that we've already deleted:

image

This makes it harder than necessary to find a workflow you're looking for.
Since the workflows are listed alphabetically I want to rename the deleted workflows to something starting with "ZZZ" so that the ones we still use appear at the top and don't require clicking "Show more workflows..."

# Changes

Recreate the deleted workflow files and give the workflows a name starting with "ZZZ".

I plan to delete these files again as soon as the workflows have been renamed on GitHub.

# Tests

no

# Todos

### DIFF
--- a/.github/workflows/aggregator.yaml
+++ b/.github/workflows/aggregator.yaml
@@ -1,1 +1,3 @@
 name: ZZZ old workflow aggregator.yaml
+on:
+  workflow_dispatch:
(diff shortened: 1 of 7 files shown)


lmuntaner (approved):
Thanks!

PR #3032, opened 2023-08-03
Unify Sns/Nns StakeItemAction components 

dskloetd:
# Motivation

The HMTL in the `NnsStakeItemAction` and `SnsStakeItemAction` components is almost identical because these components serve the same purpose and we want them to look the same.
Also their page objects are almost identical.

# Changes

1. Add `IncreaseStakeButton` which is used by both `NnsIncreaseStakeButton` and `SnsIncreaseStakeButton`.
2. Add `StakeItemAction` which is used by both `NnsStakeItemAction` and `SnsStakeItemAction`. It uses the generic `IncreaseStakeButton` and controls the logic from the parent component.
3. Add `StakeItemActionPo` to replace both `NnsStakeItemActionPo` and `SnsStakeItemActionPo`.

# Tests

Tests were added for `IncreaseStakeButton` and `StakeItemAction`.

# Todos

not necessary(?)


frontend/src/tests/lib/components/neuron-detail/actions/IncreaseStakeButton.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+import { render } from "@testing-library/svelte";
+import IncreaseStakeButtonTest from "./IncreaseStakeButtonTest.svelte";
+
+describe("IncreaseStakeButton", () => {
+  const renderComponent = async (props) => {
+    const { container } = render(IncreaseStakeButtonTest, props);
+    return IncreaseStakeButtonPo.under(new JestPageObjectElement(container));
+  };
+
+  it("Dispatches event when clicked", async () => {
```
lmuntaner, line 16: Do you consider better to use a Test component than using `component.$on`?

lmuntaner (approved):
LGTM! Thanks!

Just a small question / comment

frontend/src/tests/lib/components/neuron-detail/actions/IncreaseStakeButton.spec.ts
dskloetd: Not at all! I didn't know/didn't remember `component.$on`. Thanks! Changed.

lmuntaner: great! Thanks!

PR #3033, opened 2023-08-03
E2e/propodal details

mstrasinskis:
# Motivation

Add test to validate proposal details. Replaces `wdio/user-V01-open-proposals-are-shown.e2e`

# Changes

- add testIdWrapper to ProposalSystemInfoProposerEntryComponent
- `ProposalSystemInfoSection` po
- `ProposalSystemInfoProposerEntry` po
- missing po getters
- test to validate proposal details

# Tests

yes

# Todos


frontend/src/tests/e2e/proposals.spec.ts
dskloetd, line 1: Can you also delete `user-V01-open-proposals-are-shown.e2e.ts` (if it's fully replaced)?

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -19,12 +19,74 @@ test("Test neuron voting", async ({ page, context }) => {
...
+  expect(await systemInfoSectionPo.getProposalTypeText()).toBe("Motion");
+  expect(await systemInfoSectionPo.getProposalTopicText()).toBe("Governance");
+  expect(await systemInfoSectionPo.getProposalStatusText()).toBe("Open");
+  expect(await systemInfoSectionPo.getProposalRewardText()).toBe(
+    "Accepting Votes"
+  );
+  expect(await systemInfoSectionPo.getProposalProposerNeuronIdText()).toBe(
+    proposerNeuronId
+  );
+  // voting block
```
dskloetd, line 59: What does "voting block" mean?

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -19,12 +19,74 @@ test("Test neuron voting", async ({ page, context }) => {
...
+  expect(
+    await nnsProposalPo.getVotesResultPo().getAdoptVotingPower()
+  ).toBeLessThanOrEqual(20);
```
dskloetd, line 62: Why 20?
Why less or equal?
Can we expect an exact value and make it clear why that value?
Or maybe this expectation isn't relevant for the test and we can remove it?

frontend/src/tests/page-objects/ProposalSystemInfoSection.page-object.ts
```diff
@@ -0,0 +1,51 @@
...
+      element: this.root,
+      testId: "proposal-system-info-status",
+    }).getValueText();
+  }
+
+  async getProposalRewardText(): Promise<string> {
+    return KeyValuePairPo.under({
+      element: this.root,
+      testId: "proposal-system-info-reward",
+    }).getValueText();
```
dskloetd, line 45: Optional suggestion:
You could make a helper `getKeyValuePairValueText(testId)` in this class.

frontend/src/tests/e2e/proposals.spec.ts
mstrasinskis: Good point, I also removed already replaced `user-V02-proposal-filters-work.e2e.ts`

frontend/src/tests/e2e/proposals.spec.ts
mstrasinskis: Changed to "Votes result" 👍
image

frontend/src/tests/e2e/proposals.spec.ts
mstrasinskis: Creation of a single dummy proposal costs some ICP. The neuron stake is 10 ICP. With maximum dissolve delay it's Voting Power is 20.
But all the dummy proposals are created simultaneously so we don't know the neuron voting power at the time when a specifiec proposal is created. To be more flexible I would keep 20, because this is the maximum adopt-value proposal can have after creation.

mstrasinskis:
@dskloetd ready for another review?

frontend/src/tests/e2e/proposals.spec.ts
dskloetd: So what are we actually testing here? Did we didn't suddenly gain a lot of extra voting power after creating the neuron?
I'm not sure what is the value of that in this e2e test.

mstrasinskis: The only value is that the block is rendered and the data was populated.

PR #3035, opened 2023-08-04
Refactor: Move out sns aggregator types and converters

lmuntaner:
# Motivation

We want to create an SNS aggregator store with the data returned from the aggregator.

In this PR: I moved the converters and types of the aggregator outside the api module to be reused.

Following PR:  The api layer will return the raw data and the service will convert it before putting it in the current stores

# Changes

* Move all types to `types/sns-aggregator`.
* Move all converters to `utils/sns-aggregator-converters.utils`.
* Change imports.
* Upgrade ic-js dependencies to use a direct import for `SnsGetInitResponse`.

# Tests

* Move converter test to `utils/sns-aggregator-converters.utils.spec`.
* Change imports.

# Todos

I don't think moving files is worth a changelog entry.


lmuntaner:
@dskloetd please review

frontend/src/lib/api/sns-aggregator.api.ts
```diff
@@ -3,449 +3,12 @@ import {
...
-  SnsGetLifecycleResponse,
-  SnsGetMetadataResponse,
-  SnsGetSaleParametersResponse,
-  SnsNervousSystemFunction,
-  SnsParams,
-  SnsSwap,
-  SnsSwapDerivedState,
-  SnsSwapInit,
-} from "@dfinity/sns";
-import type { GetInitResponse } from "@dfinity/sns/dist/candid/sns_swap";
```
dskloetd, line 24: Do I understand correctly that this is the line that was replaced with importing `SnsGetInitResponse` from `"@dfinity/sns"` and that that was only possible after upgrading the deps?

frontend/src/lib/types/sns-aggregator.ts
```diff
@@ -0,0 +1,181 @@
...
+  icrc1_metadata: IcrcTokenMetadataResponse;
+  /**
+   * TODO: integrate ckBTC fee
+   * @deprecated we will use the icrc1_metadata.fee as source information for the fee
+   */
+  icrc1_fee?: bigint;
+  icrc1_total_supply: bigint;
+  derived_state: SnsGetDerivedStateResponse;
+  swap_params: SnsGetSaleParametersResponse;
+  init: SnsGetInitResponse;
```
dskloetd, line 57: So this is now `SnsGetInitResponse` instead of `GetInitResponse`?
Is that the only change and everything else is moved verbatim?

frontend/src/lib/api/sns-aggregator.api.ts
lmuntaner: Correct

dskloetd (approved):

PR #3037, opened 2023-08-04
Fix: SNS Aggregator page 2 doesn't exist

lmuntaner:
# Motivation

Bug in current mainnet. The NNS Dapp sees that the first SNS aggregator page is full; therefore, it tries to get the second page. Yet, the second page fails, and the whole process of getting the info from the aggregator fails and uses the fallback.

In this PR: Ignore errors for pages higher than 0.

# Changes

* Ignore errors for pages higher than 0 in `querySnsAggregator` helper.

# Tests

* I added first a test, it failed, and then I implemented the solution.

# Todos

### DIFF
--- a/frontend/src/lib/api/sns-aggregator.api.ts
+++ b/frontend/src/lib/api/sns-aggregator.api.ts
@@ -10,20 +10,31 @@ import { convertDtoData } from "$lib/utils/sns-aggregator-converters.utils";
 const aggergatorPageUrl = (page: number) => `/sns/list/page/${page}/slow.json`;
 
 const querySnsAggregator = async (page = 0): Promise<CachedSnsDto[]> => {
-  const response = await fetch(
-    `${SNS_AGGREGATOR_CANISTER_URL}/${AGGREGATOR_CANISTER_VERSION}${aggergatorPageUrl(
-      page
-    )}`
-  );
-  if (!response.ok) {
-    throw new Error("Error loading SNS projects from aggregator canister");
+  try {
+    const response = await fetch(
+      `${SNS_AGGREGATOR_CANISTER_URL}/${AGGREGATOR_CANISTER_VERSION}${aggergatorPageUrl(
+        page
+      )}`
+    );
+    if (!response.ok) {
+      throw new Error("Error loading SNS projects from aggregator canister");
+    }
+    const data: CachedSnsDto[] = await response.json();
+    if (data.length === AGGREGATOR_PAGE_SIZE) {
+      const nextPageData = await querySnsAggregator(page + 1);
+      return [...data, ...nextPageData];
+    }
+    return data;
+  } catch (e) {
+    // If the error is after the first page, is because there are no more pages it fails
+    if (page > 0) {
+      console.error(
+        `Error loading SNS project page ${page} from aggregator canister`
+      );
+      return [];
+    }
+    throw e;
   }
-  const data: CachedSnsDto[] = await response.json();
-  if (data.length === AGGREGATOR_PAGE_SIZE) {
-    const nextPageData = await querySnsAggregator(page + 1);
-    return [...data, ...nextPageData];
-  }
-  return data;
 };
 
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/api/sns-aggregator.api.ts
```diff
@@ -447,20 +447,31 @@ export const convertDtoData = (data: CachedSnsDto[]): CachedSns[] =>
...
+    if (!response.ok) {
+      throw new Error("Error loading SNS projects from aggregator canister");
+    }
+    const data: CachedSnsDto[] = await response.json();
+    if (data.length === AGGREGATOR_PAGE_SIZE) {
+      const nextPageData = await querySnsAggregator(page + 1);
+      return [...data, ...nextPageData];
+    }
+    return data;
+  } catch (e) {
```
dskloetd, line 465: Can we distinguish between the `!response.ok` and an error in rest of the code?

I think instead of adding the `try` block and blanket catching everything, can't we just return `[]` if `!response.ok && page > 0`?

frontend/src/tests/lib/api/sns-aggregator.api.spec.ts
```diff
@@ -60,6 +60,29 @@ describe("sns-aggregator api", () => {
       expect(mockFetch).toHaveBeenCalledTimes(2);
     });
```
dskloetd, line 61: Not for this PR but a test seems to be missing that checks if it correctly concatenates > 10 projects responses.

lmuntaner: There is a test for 10 + 1. Is that what you mean?

frontend/src/lib/api/sns-aggregator.api.ts
lmuntaner: I can add that, but I tried getting the file from the console from mainnet, and it throws an error. It doesn't return a response object where I can check that.

lmuntaner:

frontend/src/lib/api/sns-aggregator.api.ts
dskloetd: Try with page 0:
```
const r = await fetch('ttps://3r4gx-wqaaa-aaaaq-aaaia-cai.icp0.io/v1/sns/list/page/0/slow.json');
```
It gives the same error. So I'm not sure you're drawing the right conclusion from that?

frontend/src/tests/lib/api/sns-aggregator.api.spec.ts
dskloetd: Yeah, it's the test right above this comment, right?
But it doesn't do anything like
```
expect(projects).toHaveLength(11);
```

It only checks that it makes the expected calls, but not that it returns the expected result.
Anyway, not for this PR.

frontend/src/lib/api/sns-aggregator.api.ts
lmuntaner: I don't get the error with page 0:

```
const r = fetch('https://3r4gx-wqaaa-aaaaq-aaaia-cai.icp0.io/v1/sns/list/page/0/slow.json');
```

This doesn't throw any error in my console.

frontend/src/tests/lib/api/sns-aggregator.api.spec.ts
lmuntaner: true, good point. I'll add that in a following PR.

frontend/src/lib/api/sns-aggregator.api.ts
dskloetd: Oops. I see what you mean now.
In that case can we put the try/catch only around the `fetch` statement?

lmuntaner: sure, sounds good

lmuntaner:
@dskloetd ready for review

dskloetd (approved):

PR #3044, opened 2023-08-04
GIX-1712: One confirmation with Ledger device

lmuntaner:
# Motivation

Improve the UX of HW users.

In this PR: Users are required to confirm the transactions only ONCE with the device. Not twice as until now.

In a nutshell, the read_state call is signed at the same time than the other request. This is possible because the request id is deterministic of the request and can be calculated before hand. We cache the read_state signature so that we can reused it in the following request call which is the read_state one.

# Changes

* LedgerIdentity uses app method "signUpdateCall" instead of "sign" that requires the blob of the read_state call also.
* Store the read_state body and signature in the LedgerIdentity to be reused in the following call.
* Do not use the new method if the transaction is to stake a neuron because there is an issue in the Ledger App. This will be solved in the next release.

# Tests

* Add a test for the LedgerIdentity to check the `transformRequest` functionality.

# Todos

### DIFF
--- a/frontend/src/lib/identities/ledger.identity.ts
+++ b/frontend/src/lib/identities/ledger.identity.ts
@@ -75,7 +75,6 @@ export class LedgerIdentity extends SignIdentity {
     return this.publicKey;
   }
 
-  // We still keep this even though it's not used directly here, because it's required by the interface.
   public override async sign(blob: ArrayBuffer): Promise<Signature> {
     const callback = async (app: LedgerApp): Promise<Signature> => {
       const responseSign: ResponseSign = await app.sign(
--- a/frontend/src/lib/identities/ledger.identity.ts
+++ b/frontend/src/lib/identities/ledger.identity.ts
@@ -10,7 +10,8 @@ import { replacePlaceholders } from "$lib/utils/i18n.utils";
 import {
   decodePublicKey,
   decodeSignature,
-  decodeSignatures,
+  decodeUpdateSignatures,
+  getRequestId,
   type RequestSignatures,
 } from "$lib/utils/ledger.utils";
 import {
@@ -18,13 +19,16 @@ import {
   SignIdentity,
   requestIdOf,
   type CallRequest,
+  type HttpAgentReadStateRequest,
   type HttpAgentRequest,
   type PublicKey,
   type ReadRequest,
   type ReadRequestType,
   type ReadStateRequest,
+  type RequestId,
   type Signature,
 } from "@dfinity/agent";
+import { Principal } from "@dfinity/principal";
 import { nonNullish } from "@dfinity/utils";
 import type Transport from "@ledgerhq/hw-transport";
 import type LedgerApp from "@zondax/ledger-icp";
@@ -43,11 +47,28 @@ type ReadStateData = {
   body: ReadStateRequest;
 };
 
+const sendersMatch = (
+  sender1: Uint8Array | Principal,
+  sender2: Uint8Array | Principal
+): boolean => {
+  if (sender1 instanceof Uint8Array && sender2 instanceof Uint8Array) {
+    return (
+      Buffer.from(sender1).toString("hex") ===
+      Buffer.from(sender2).toString("hex")
+    );
+  }
+  if (sender1 instanceof Principal && sender2 instanceof Principal) {
+    return sender1.toText() === sender2.toText();
+  }
+  return false;
+};
+
 export class LedgerIdentity extends SignIdentity {
   // TODO(L2-433): is there a better way to solve this requirements than a class variable that is set and unset?
   // A flag to signal that the next transaction to be signed will be a "stake neuron" transaction.
   private neuronStakeFlag = false;
   // Used to avoid signing the read state transaction twice.
+  // The key is the request id of the first request, not the read state request.
   // Map<requestIdHex, ReadStateData>
   private readStateMap: Map<string, ReadStateData> = new Map();
 
@@ -112,7 +133,7 @@ export class LedgerIdentity extends SignIdentity {
       // Remove the "neuron stake" flag, since we already signed the transaction.
       this.neuronStakeFlag = false;
 
-      return decodeSignatures(responseSign);
+      return decodeUpdateSignatures(responseSign);
     };
 
     return this.executeWithApp<RequestSignatures>(callback);
@@ -217,11 +238,6 @@ export class LedgerIdentity extends SignIdentity {
     return decodePublicKey(response);
   }
 
-  public clearInstanceVariablesForTesting(): void {
-    this.neuronStakeFlag = false;
-    this.readStateMap = new Map();
-  }
-
   private async executeWithApp<T>(
     callback: (app: LedgerApp) => Promise<T>
   ): Promise<T> {
@@ -246,6 +262,77 @@ export class LedgerIdentity extends SignIdentity {
     }
   }
 
+  private async getRequestFromCache(
+    request: HttpAgentReadStateRequest
+  ): Promise<Record<string, unknown> | undefined> {
+    const { body, ...fields } = request;
+    const requestId = getRequestId(body);
+    const requestIdHex = Buffer.from(requestId).toString("hex");
+    const requestData = this.readStateMap.get(requestIdHex);
+    if (nonNullish(requestData)) {
+      const { signature, body: cachedBody } = requestData;
+      // If cached data doesn't match, ignore and move on to sign the request.
+      // `ingress_expiry` doesn't need to match. TODO: Ask why.
+      if (sendersMatch(cachedBody.sender, body.sender)) {
+        return {
+          ...fields,
+          body: {
+            content: cachedBody,
+            sender_pubkey: this.publicKey.toDer(),
+            sender_sig: signature,
+          },
+        };
+      }
+    }
+    console.warn(
+      `No cached data for read state request with id ${requestIdHex}.`
+    );
+  }
+
+  private async createReadStateRequest(
+    body: CallRequest
+  ): Promise<{ readStateBody: ReadStateRequest; requestId: RequestId }> {
+    const requestId = await requestIdOf(body);
+    const readStateBody = {
+      // Can't import ReadRequestType as value from @dfinity/agent because it's const enum
+      request_type: "read_state" as ReadRequestType.ReadState,
+      // Check docs for more detais: https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-read-state
+      // Quote: "Moreover, all paths with prefix /request_status/<request_id> must refer to the same request ID <request_id>."
+      paths: [[new TextEncoder().encode("request_status"), requestId]],
+      ingress_expiry: body.ingress_expiry,
+      sender: body.sender,
+    };
+    return {
+      readStateBody,
+      requestId,
+    };
+  }
+
+  private storeData({
+    requestId,
+    signature,
+    body,
+  }: {
+    requestId: RequestId;
+    signature: Signature;
+    body: ReadStateRequest;
+  }) {
+    // Store the read state signature to be able to reuse it later.
+    const requestIdHex = Buffer.from(requestId).toString("hex");
+    this.readStateMap.set(requestIdHex, {
+      signature: signature,
+      body,
+    });
+  }
+
+  /**
+   * Convert the HttpAgentRequest body into cbor which can be signed by the Ledger Hardware Wallet.
+   * @param request - body of the HttpAgentRequest
+   */
+  private prepareCborForLedger = (
+    request: ReadRequest | CallRequest
+  ): ArrayBuffer => Cbor.encode({ content: request });
+
   /**
    * Required implementation for agent-js transformRequest.
    *
@@ -255,64 +342,39 @@ export class LedgerIdentity extends SignIdentity {
     request: HttpAgentRequest
   ): Promise<Record<string, unknown>> {
     if (
-      // Any other call will reset `readStateSignature` and `readStateBody`.
-      // Can't import Endpoint as value from @dfinity/agent because it's const enum.
+      // "read_state" is a value from Endpoint enum from @dfinity/agent but it's const enum and can't be imported as value with `isolatedModules` flag.
+      // More info: https://github.com/microsoft/TypeScript/issues/40344#issuecomment-956368612
       request.endpoint === "read_state"
     ) {
-      const { body: _body, ...fields } = request;
-      const [_, requestId] = _body.paths[0];
-      const requestIdHex = Buffer.from(requestId).toString("hex");
-      const requestData = this.readStateMap.get(requestIdHex);
-      if (nonNullish(requestData)) {
-        const { signature, body } = requestData;
-        return {
-          ...fields,
-          body: {
-            content: body,
-            sender_pubkey: this.publicKey.toDer(),
-            sender_sig: signature,
-          },
-        };
+      const cachedRequest = await this.getRequestFromCache(request);
+      if (nonNullish(cachedRequest)) {
+        return cachedRequest;
       }
     }
-
-    /**
-     * Convert the HttpAgentRequest body into cbor which can be signed by the Ledger Hardware Wallet.
-     * @param request - body of the HttpAgentRequest
-     */
-    const prepareCborForLedger = (
-      request: ReadRequest | CallRequest
-    ): ArrayBuffer => Cbor.encode({ content: request });
     const { body, ...fields } = request;
-
-    const requestId = await requestIdOf(body);
 
     let callSignature: Signature;
     // There is an issue with the Ledger App when the neuron flag is set to true and `signWithReadState`.
     // TODO: Check app version and use `signWithReadState` only if the app version has the issue fixed.
-    if (this.neuronStakeFlag || request.endpoint === "read_state") {
-      callSignature = await this.sign(prepareCborForLedger(body));
-    } else {
-      // Store the body of the read state request to be able to reuse it later.
-      const readStateBody = {
-        // Can't import ReadRequestType as value from @dfinity/agent because it's const enum
-        request_type: "read_state" as ReadRequestType.ReadState,
-        // Check docs for more detais: https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-read-state
-        paths: [[new TextEncoder().encode("request_status"), requestId]],
-        ingress_expiry: body.ingress_expiry,
-        sender: body.sender,
-      };
+    if (request.endpoint === "call" && !this.neuronStakeFlag) {
+      // If the request endpoint is a "call", then the body is a CallRequest.
+      // Reference: https://github.com/dfinity/agent-js/blob/ce772199189748f9b77c8eb3ceeb3fdd11c70b5b/packages/agent/src/agent/http/types.ts#L28
+      const callBody = body as CallRequest;
+      const { requestId, readStateBody } = await this.createReadStateRequest(
+        callBody
+      );
       const signatures = await this.signWithReadState(
-        prepareCborForLedger(body),
-        prepareCborForLedger(readStateBody)
+        this.prepareCborForLedger(body),
+        this.prepareCborForLedger(readStateBody)
       );
       callSignature = signatures.callSignature;
-      // Store the read state signature to be able to reuse it later.
-      const requestIdHex = Buffer.from(requestId).toString("hex");
-      this.readStateMap.set(requestIdHex, {
+      this.storeData({
+        requestId,
         signature: signatures.readStateSignature,
         body: readStateBody,
       });
+    } else {
+      callSignature = await this.sign(this.prepareCborForLedger(body));
     }
 
     return {
--- a/frontend/src/lib/identities/ledger.identity.ts
+++ b/frontend/src/lib/identities/ledger.identity.ts
@@ -245,8 +245,6 @@ export class LedgerIdentity extends SignIdentity {
   public override async transformRequest(
     request: HttpAgentRequest
   ): Promise<Record<string, unknown>> {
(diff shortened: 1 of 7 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -207,23 +253,74 @@ export class LedgerIdentity extends SignIdentity {
    */
   public override async transformRequest(
     request: HttpAgentRequest
-  ): Promise<unknown> {
+  ): Promise<Record<string, unknown>> {
+    if (
+      // Any other call will reset `readStateSignature` and `readStateBody`.
+      // Can't import Endpoint as value from @dfinity/agent because it's const enum.
```
dskloetd, line 259: It took me a while to understand what this comment has to do with the line below.
But I guess what you're trying to say is that `"read_state"` is `Endpoint.ReadState` but that you can't write `Endpoint.ReadState` because you can't import `Endpoint`?

Why can't you import a const enum? Is this something that can be fixed in the agent library?

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -207,23 +253,74 @@ export class LedgerIdentity extends SignIdentity {
...
+      callSignature = signatures.callSignature;
+      // Store the read state signature to be able to reuse it later.
+      const requestIdHex = Buffer.from(requestId).toString("hex");
+      this.readStateMap.set(requestIdHex, {
+        signature: signatures.readStateSignature,
+        body: readStateBody,
+      });
+    }
+
     return {
```
dskloetd, line 318: This became quite a long function. How about factoring out the "get the signature from the map" and "store the signature in the map" bits into separate functions?

frontend/src/lib/services/icp-ledger.services.ts
```diff
@@ -115,6 +115,7 @@ export const resetIdentitiesCachedForTesting = () => (identities = {});
 export const getLedgerIdentity = async (
   identifier: string
 ): Promise<LedgerIdentity> => {
+  // TODO: remove caching of the identities to avoid issues with the LedgerIdentity instance being reused.
```
dskloetd, line 118: Can you clarify what kind of issues?

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -7,31 +7,49 @@ import {
...
+  signature: Signature;
+  body: ReadStateRequest;
+};
+
 export class LedgerIdentity extends SignIdentity {
   // TODO(L2-433): is there a better way to solve this requirements than a class variable that is set and unset?
   // A flag to signal that the next transaction to be signed will be a "stake neuron" transaction.
   private neuronStakeFlag = false;
+  // Used to avoid signing the read state transaction twice.
+  // Map<requestIdHex, ReadStateData>
```
dskloetd, line 51: Can you clarify that the map is keyed by the request ID of the original call (as opposed to the request ID of the read state request)?

frontend/src/tests/lib/identities/ledger.identity.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    // Read state for first call
+    const readRequest1 = await identity.transformRequest(mockReadStateRequest1);
+    expect(mockLedgerApp.sign).toHaveBeenCalledTimes(0);
+    expect(mockLedgerApp.signUpdateCall).toHaveBeenCalledTimes(2);
+    if (typeof readRequest1.body === "object" && "paths" in readRequest1.body) {
+      expect(readRequest1.body.paths[0][1]).toEqual(requestId1);
+    }
+
+    // Read state for second call
```
dskloetd, line 183: If the order doesn't matter, maybe put these 2 blocks in reverse order?

frontend/src/tests/lib/identities/ledger.identity.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    await identity.transformRequest(mockHttpRequest1);
+    await identity.transformRequest(mockHttpRequest2);
+    expect(mockLedgerApp.signUpdateCall).toHaveBeenCalledTimes(2);
+
+    // Read state for first call
+    const readRequest1 = await identity.transformRequest(mockReadStateRequest1);
+    expect(mockLedgerApp.sign).toHaveBeenCalledTimes(0);
+    expect(mockLedgerApp.signUpdateCall).toHaveBeenCalledTimes(2);
+    if (typeof readRequest1.body === "object" && "paths" in readRequest1.body) {
+      expect(readRequest1.body.paths[0][1]).toEqual(requestId1);
```
dskloetd, line 180: Where is defined that you need `[0][1]` to get the request ID? Is there a more readable way to get it? Can we define a getter or something?

frontend/src/lib/identities/ledger.identity.ts
lmuntaner: Yes, sure.

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -176,6 +217,11 @@ export class LedgerIdentity extends SignIdentity {
     return decodePublicKey(response);
   }
 
+  public clearInstanceVariablesForTesting(): void {
```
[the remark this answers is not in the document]
lmuntaner: Actually, this is not needed. This was in the middle of the work, but I'll remove it. Thanks!

frontend/src/lib/identities/ledger.identity.ts
lmuntaner: Let me add a link to an issue I found.

Also, for more reference: https://hackmd.io/@dzearing/Sk3xV0cLs

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -207,23 +253,74 @@ export class LedgerIdentity extends SignIdentity {
...
   public override async transformRequest(
     request: HttpAgentRequest
-  ): Promise<unknown> {
+  ): Promise<Record<string, unknown>> {
+    if (
+      // Any other call will reset `readStateSignature` and `readStateBody`.
+      // Can't import Endpoint as value from @dfinity/agent because it's const enum.
+      request.endpoint === "read_state"
+    ) {
+      const { body: _body, ...fields } = request;
```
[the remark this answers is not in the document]
lmuntaner: Oh, yes, I wasn't using it before, now we do.

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -207,23 +253,74 @@ export class LedgerIdentity extends SignIdentity {
...
+      const [_, requestId] = _body.paths[0];
```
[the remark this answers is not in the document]
lmuntaner: Let me add a link.

This is defined in the protocol

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -207,23 +253,74 @@ export class LedgerIdentity extends SignIdentity {
...
+      const requestIdHex = Buffer.from(requestId).toString("hex");
+      const requestData = this.readStateMap.get(requestIdHex);
+      if (nonNullish(requestData)) {
```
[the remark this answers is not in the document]
lmuntaner: I believe the only other important asked is the `sender` and `ingress_expiry`. Let me check those also.

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -207,23 +253,74 @@ export class LedgerIdentity extends SignIdentity {
...
+        const { signature, body } = requestData;
+        return {
```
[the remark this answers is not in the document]
lmuntaner: Do we want to rely on the agent-js caching the readState?

we don't know whether the readState succeeded at this point

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -207,23 +253,74 @@ export class LedgerIdentity extends SignIdentity {
...
+    const requestId = await requestIdOf(body);
+
+    let callSignature: Signature;
+    // There is an issue with the Ledger App when the neuron flag is set to true and `signWithReadState`.
+    // TODO: Check app version and use `signWithReadState` only if the app version has the issue fixed.
+    if (this.neuronStakeFlag || request.endpoint === "read_state") {
+      callSignature = await this.sign(prepareCborForLedger(body));
+    } else {
+      // Store the body of the read state request to be able to reuse it later.
+      const readStateBody = {
```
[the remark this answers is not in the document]
lmuntaner: Good point, let me skip those as well.

frontend/src/lib/identities/ledger.identity.ts
lmuntaner: sure, makes sense!

frontend/src/lib/utils/ledger.utils.ts
```diff
@@ -77,9 +90,43 @@ export const decodeSignature = async ({
...
+  StatusReadSignatureRS,
+  returnCode,
+  errorMessage,
+}: ResponseSignUpdateCall): Promise<RequestSignatures> => {
+  await checkResponseCode(returnCode);
+  checkSignature({ signature: RequestSignatureRS, returnCode, errorMessage });
+  checkSignature({
+    signature: StatusReadSignatureRS,
+    returnCode,
+    errorMessage,
```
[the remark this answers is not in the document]
lmuntaner: I'll rename the util. This checks the return of `signUpdateCall` which returns both signatures but one `returnCode` and one `errorMessage`. I see how the name `decodeSignatures` is confusing.

frontend/src/tests/lib/identities/ledger.identity.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    const identity = await LedgerIdentity.create();
+
+    const request = await identity.transformRequest(mockHttpRequest1);
+    expect(mockLedgerApp.signUpdateCall).toHaveBeenCalledTimes(1);
+    expect(request.endpoint).toBe("call");
+
+    const readRequest = await identity.transformRequest(mockReadStateRequest1);
+    expect(mockLedgerApp.signUpdateCall).toHaveBeenCalledTimes(1);
+    expect(readRequest.endpoint).toBe("read_state");
+    identity.clearInstanceVariablesForTesting();
```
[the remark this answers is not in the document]
lmuntaner: I'll remove this line.

But I prefer to explicitly create an `identity` in each test case.

I find it ugly to create a variable outside, assign the value in the beforeEach and use it in the test cases. I don't find it readable.

frontend/src/tests/lib/identities/ledger.identity.spec.ts
lmuntaner: Good idea, I can create an util.

frontend/src/tests/lib/identities/ledger.identity.spec.ts
lmuntaner: sure

lmuntaner:

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -255,64 +342,39 @@ export class LedgerIdentity extends SignIdentity {
...
-      const requestIdHex = Buffer.from(requestId).toString("hex");
-      this.readStateMap.set(requestIdHex, {
+      this.storeData({
+        requestId,
         signature: signatures.readStateSignature,
         body: readStateBody,
       });
+    } else {
+      callSignature = await this.sign(this.prepareCborForLedger(body));
     }
```
dskloetd, line 378: Optional:
I feel like the lines from
```
if (request.endpoint === "call" && !this.neuronStakeFlag) {
```
until here should be a function because then it could return early in case of a neuron statke instead of having a really long `if` block`.
But the function name would be either not very accurate `createCallSignature` or very long `createCallSignatureAndMaybeCacheReadStateSignature`.
I still think it's worth it but I'll let you decide.

frontend/src/lib/identities/ledger.identity.ts
dskloetd: Thanks for the references. I don't fully understand but it seems that the issue is that const enums are bad.
Do we know why `@dfinity/agent` uses const enums?
Does the team know about it?

frontend/src/lib/utils/ledger.utils.ts
```diff
@@ -130,3 +130,7 @@ export const decodeSignatures = async ({
 
 const bufferToArrayBuffer = (buffer: Buffer): ArrayBuffer =>
   buffer.buffer.slice(buffer.byteOffset, buffer.byteOffset + buffer.byteLength);
+
+// Check docs for more detais: https://internetcomputer.org/docs/current/references/ic-interface-spec/#http-read-state
+// Quote: "Moreover, all paths with prefix /request_status/<request_id> must refer to the same request ID <request_id>."
+export const getRequestId = (body: ReadRequest): RequestId => body.paths[0][1];
```
dskloetd, line 136: Thanks for the reference!

The `[0][1]` is closely linked to how the request body is created. Especially the `[0]` could be something else if we request multiple states in our request. So the closer the definition of this utility function can be to where (the paths of) the request body is created, the better.

Could we have another utility function here to create the path so the connection is obvious?

frontend/src/lib/identities/ledger.identity.ts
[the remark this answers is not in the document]
dskloetd: My point was that if the format changes, or some additional field is added or whatever, we shouldn't silently ignore that. So I was hoping to check that they are byte for byte the same.

But I see you added a comment
```
`ingress_expiry` doesn't need to match. TODO: Ask why.
```
Why do you think it doesn't need to match?

frontend/src/lib/identities/ledger.identity.ts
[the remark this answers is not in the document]
dskloetd: Will the map just grow and grow?
If agent-js does cache the read state, I think it's OK to rely on it. It's already kind of a hack that relies on implementation details.

frontend/src/lib/utils/ledger.utils.ts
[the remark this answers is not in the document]
dskloetd: That doesn't change my question.
You pass the same `returnCode` and `errorMessage` to 2 different calls of `checkSignature` for different signatures.
Supposedly the error message explains what is wrong with the signature.
Why would it necessarily be the same for both signatures?

frontend/src/tests/lib/identities/ledger.identity.spec.ts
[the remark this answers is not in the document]
dskloetd: It's fine to explicitly create an identity.
But based on this resetting, I assumed that the identity is actually cached somewhere and it's not actually using a fresh one in each test.
Did I misunderstand?

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -246,6 +262,77 @@ export class LedgerIdentity extends SignIdentity {
...
+      ingress_expiry: body.ingress_expiry,
+      sender: body.sender,
+    };
+    return {
+      readStateBody,
+      requestId,
+    };
+  }
+
+  private storeData({
```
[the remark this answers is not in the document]
lmuntaner: Yes, sounds good.

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -207,23 +253,74 @@ export class LedgerIdentity extends SignIdentity {
...
+        return {
+          ...fields,
+          body: {
+            content: body,
+            sender_pubkey: this.publicKey.toDer(),
+            sender_sig: signature,
+          },
+        };
+      }
+    }
```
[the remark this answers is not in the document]
lmuntaner: Inside the `getRequestFromCache`.

The requestId is easily accessible there and it was nice to add it in the warning.

frontend/src/lib/identities/ledger.identity.ts
lmuntaner: Yes, the long or inaccurate name was what stopped me from doing it.

I might do it if you're ok with a long name then,

frontend/src/lib/utils/ledger.utils.ts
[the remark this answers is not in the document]
lmuntaner: Ah, ok, I don't know. We should ask Zondax, they developed the library that returns two signatures and one errorMessage and one returnCode.

There is nothing in the returnCode that specifies to which signature it applies.

I can add a comment to reach out to them and add a link to their code if neeeded.

frontend/src/lib/utils/ledger.utils.ts
lmuntaner: Sure

frontend/src/lib/identities/ledger.identity.ts
lmuntaner: No, I can reach out to them also.

frontend/src/lib/identities/ledger.identity.ts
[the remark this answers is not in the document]
lmuntaner: I don't know. And to be honest, I would have expected them to match because the code to create a requestId seems to use all the fields.

I'm also confused and need to look further into this.

I'll add the comment and track a todo for myself.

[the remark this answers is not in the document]
lmuntaner: Oh, I already added the TODO to ask why. I thought that was your addition

frontend/src/lib/identities/ledger.identity.ts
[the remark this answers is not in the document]
lmuntaner: Yes, until browser is refreshed. But only HW requests are cached. Each needs input from the user, so it won't grow that much.

frontend/src/tests/lib/identities/ledger.identity.spec.ts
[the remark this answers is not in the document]
lmuntaner: The caching is not in `LedgerIdentity.create`. The tests do not cache the identity

frontend/src/lib/identities/ledger.identity.ts
lmuntaner: done

frontend/src/lib/identities/ledger.identity.ts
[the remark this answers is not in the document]
lmuntaner: Ok, I understood why.

The requestId is created from the initial request. Therefore, it doesn't use the ingress_expiry.

The request that comes through `transformRequest` was created later than the initial request, that's why it has a different ingress_expiry.

Yet, the signature we have cached, was done in the body we created at the same time as the initial request. Therefore, the signature doesn't work with the body coming in `transformRequest.

I added it in the comment. Let me know if something is not clear.

lmuntaner:

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -200,30 +263,136 @@ export class LedgerIdentity extends SignIdentity {
...
   }
 
+  private async getRequestFromCache(
+    request: HttpAgentReadStateRequest
+  ): Promise<Record<string, unknown> | undefined> {
+    const { body, ...fields } = request;
+    const requestId = getRequestId(body);
+    const requestIdHex = Buffer.from(requestId).toString("hex");
+    const requestData = this.readStateMap.get(requestIdHex);
+    if (nonNullish(requestData)) {
```
dskloetd, line 273: If you change `nonNullish` to `isNullush` we can do an early return and less indent.

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -200,30 +263,136 @@ export class LedgerIdentity extends SignIdentity {
...
+
+  /**
+   * Convert the HttpAgentRequest body into cbor which can be signed by the Ledger Hardware Wallet.
+   * @param request - body of the HttpAgentRequest
+   */
+  private prepareCborForLedger = (
+    request: ReadRequest | CallRequest
+  ): ArrayBuffer => Cbor.encode({ content: request });
+
+  private async createSignAndStoreCallRequests(request: HttpAgentRequest) {
```
dskloetd, line 336: Did you mean `createSignAndStoreReadStateRequests`? (`ReadState` instead of `Call`)

lmuntaner: No, because this signs also the Call request.

I consider the "Call Requests" to be both, the Call and the Read State requests.

frontend/src/lib/identities/ledger.identity.ts
lmuntaner: sure

frontend/src/lib/identities/ledger.identity.ts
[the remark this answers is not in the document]
lmuntaner: done

frontend/src/lib/identities/ledger.identity.ts
lmuntaner: done

lmuntaner:

dskloetd (approved):

frontend/src/lib/identities/ledger.identity.ts
```diff
@@ -270,31 +255,38 @@ export class LedgerIdentity extends SignIdentity {
...
+      return;
     }
-    console.warn(
-      `No cached data for read state request with id ${requestIdHex}.`
-    );
+    const { signature, body: cachedBody } = requestData;
+    // If cached data doesn't match, ignore and move on to sign the request.
+    // The `ingress_expiry` is different in the cached in the new request because the new one is created after the first call.
+    // But the signature was done with the cached body. That's why we use the cached body.
+    const newRequest = {
```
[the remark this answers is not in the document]
lmuntaner: sure

PR #3046, opened 2023-08-07
Fix past-changelog-test line number comparison

dskloetd:
# Motivation

We have a test `past-changelog-test` to make sure no change log entries are added in existing releases.
But the way it compared line numbers, using `[[ "$x" > "$y" ]]` seems to be doing string comparison instead of number comparison.
Using `(( ))` instead o `[[ ]]` does number comparison instead.

# Changes

1. Fix `past-changelog-test` to do proper number comparison.
2. Add a test for the test.
3. Add the test to the GitHub workflow.

# Tests

Added

# Todos

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -285,6 +285,7 @@ jobs:
         with:
           fetch-depth: 0
       - run: scripts/past-changelog-test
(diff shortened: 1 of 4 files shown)


mstrasinskis (approved):
Thanks!

PR #3048, opened 2023-08-07
Feat: SNS Aggregator Store

lmuntaner:
# Motivation

Stop using deprecated get_state data.

In this PR: Store the aggregator data in its own store.

Then this data will be used in derived store `snsSummariesStore` so that we can use the new endpoints that the aggregator already collects.

# Changes

* New store: snsAggregatorStore.
* Load store in sns service `loadSnsProjects`.

# Tests

* Test new store.
* Test that service loads data in the store.

# Todos

Not worth an entry yet.

### DIFF
--- a/frontend/src/tests/lib/services/_public/sns.services.spec.ts
+++ b/frontend/src/tests/lib/services/_public/sns.services.spec.ts
@@ -105,6 +105,7 @@ describe("SNS public services", () => {
       snsQueryStore.reset();
       snsFunctionsStore.reset();
       transactionsFeesStore.reset();
+      snsAggregatorStore.reset();
       jest.clearAllMocks();
       jest
         .spyOn(authStore, "subscribe")
@@ -136,6 +137,8 @@ describe("SNS public services", () => {
       jest
         .spyOn(aggregatorApi, "querySnsProjects")
         .mockImplementation(() => Promise.resolve(aggregatorMockSnsesDataDto));
+
+      expect(get(snsAggregatorStore).data).toBeUndefined();
 
       await loadSnsProjects();
 
--- a/frontend/src/tests/lib/services/_public/sns.services.spec.ts
+++ b/frontend/src/tests/lib/services/_public/sns.services.spec.ts
@@ -9,6 +9,7 @@ import {
   loadSnsProjects,
 } from "$lib/services/$public/sns.services";
 import { authStore } from "$lib/stores/auth.store";
+import { snsAggregatorStore } from "$lib/stores/sns-aggregator.store";
 import { snsFunctionsStore } from "$lib/stores/sns-functions.store";
 import { snsTotalTokenSupplyStore } from "$lib/stores/sns-total-token-supply.store";
 import { snsQueryStore } from "$lib/stores/sns.store";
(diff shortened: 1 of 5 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/services/_public/sns.services.spec.ts
```diff
@@ -130,6 +132,16 @@ describe("SNS public services", () => {
       expect(feesStore.projects[rootCanisterId]).not.toBeUndefined();
     });
 
+    it("should load sns aggregator store", async () => {
+      jest
+        .spyOn(aggregatorApi, "querySnsProjects")
+        .mockImplementation(() => Promise.resolve(aggregatorMockSnsesDataDto));
+
+      await loadSnsProjects();
```
dskloetd, line 140: Can you check the the store is empty before this call?

frontend/src/tests/mocks/sns-aggregator.mock.ts
```diff
@@ -1,12 +1,12 @@
...
-  CachedSnsDto,
-  CachedSnsTokenMetadataDto,
-} from "$lib/types/sns-aggregator";
+import type { CachedSns, CachedSnsDto } from "$lib/types/sns-aggregator";
 import tenAggregatedSnses from "$tests/mocks/sns-aggregator.mock.json";
 import { SnsSwapLifecycle } from "@dfinity/sns";
 
+// TS is not smart enough to infer the type from the JSON file.
+export const aggregatorMockSnsesDataDto: CachedSnsDto[] =
+  tenAggregatedSnses.map((sns: unknown) => sns as CachedSnsDto);
```
dskloetd, line 8: Instead of doing a useless map (it doesn't do anything, right?), can't we just declare to TS that is is that type?
```
export const aggregatorMockSnsesDataDto = tenAggregatedSnses as CachedSnsDto[];
```

frontend/src/tests/lib/services/_public/sns.services.spec.ts
lmuntaner: sure

frontend/src/tests/mocks/sns-aggregator.mock.ts
lmuntaner: Unfortunately no, TS doesn't like it.

frontend/src/tests/lib/services/_public/sns.services.spec.ts
lmuntaner: I forgot to reset it in the beforeEach, thanks!

lmuntaner:

frontend/src/tests/mocks/sns-aggregator.mock.ts
dskloetd: I tried it and asked GPT-4 for help. It suggested the following which seems to work:
```
export const aggregatorMockSnsesDataDto = tenAggregatedSnses as unknown as CachedSnsDto[];
```

lmuntaner: haha, one point for AI

dskloetd (approved):

PR #3050, opened 2023-08-07
Allow new releases in scripts/past-changelog-test.test

dskloetd:
# Motivation

I recently added a test for `past-changelog-test` in https://github.com/dfinity/nns-dapp/pull/3046
but it fails when a PR adds a new release because in that case adding new changes in the newly created release is allowed.
`past-changelog-test` handles this correctly but `past-changelog-test.test` doesn't.

# Changes

1. In `past-changelog-test.test` when making a change below the first `## Proposal` heading in order to expect it to fail, make sure the change happens below the first `## Proposal` heading that exists in `origin/main` and not the first heading in the current commit if it adds a new `## Proposal` heading.
2. Add one more test to make sure that a new `## Proposal` heading with changes below it is allowed.

# Tests

test only

# Todos

Covered because it should have been done correctly in the previous PR.

### DIFF
--- a/scripts/past-changelog-test.test
+++ b/scripts/past-changelog-test.test
@@ -19,16 +19,25 @@ sponge() {
   mv "$tmp_file" "$1"
 }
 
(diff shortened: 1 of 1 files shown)


mstrasinskis (approved):
Thank you!

PR #3053, opened 2023-08-07
Chore: Fix mock HW account principal

lmuntaner:
# Motivation

Mock of the HW account has the same principal as the mockIdentity. This is not possible and it's better to avoid having it as default mock.

# Changes

* Change the principal in the `mockHardwareWalletAccount`.
* Fix broken test after change.

# Tests

Only test changes.

# Todos

Not worth an entry in the changelog.

### DIFF
--- a/frontend/src/tests/mocks/icp-accounts.store.mock.ts
+++ b/frontend/src/tests/mocks/icp-accounts.store.mock.ts
@@ -41,7 +41,7 @@ export const mockHardwareWalletAccount: IcpAccount = {
     "646f4d2d6fcb6fab5ba1547647526b666553467ecb5cb28c8d9ddf451c8f4c21",
   balanceE8s: 123456789010000n,
   principal: Principal.fromText(
-    "xlmdg-vkosz-ceopx-7wtgu-g3xmd-koiyc-awqaq-7modz-zf6r6-364rh-oqe"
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd please review

dskloetd (approved):

frontend/src/tests/mocks/icp-accounts.store.mock.ts
```diff
@@ -63,7 +63,7 @@ export const mockSubAccountDetails: SubAccountDetails = {
 export const mockHardwareWalletAccountDetails: HardwareWalletAccountDetails = {
   name: "ledger test",
   principal: Principal.fromText(
-    "xlmdg-vkosz-ceopx-7wtgu-g3xmd-koiyc-awqaq-7modz-zf6r6-364rh-oqe"
+    "2vtpp-r6lcd-cbfas-qbabv-wxrv5-lsrkj-c4dtb-6ets3-srlqe-xpuzf-vqe"
```
dskloetd, line 66: Make a constant for this string an use it in both places?

dskloetd (approved):

frontend/src/tests/mocks/icp-accounts.store.mock.ts
lmuntaner: sure

PR #3055, opened 2023-08-08
GIX-1751: Hide neurons fund in advanced section

lmuntaner:
# Motivation

Bug: Current functionality doesn't show the Neurons' Fund checkbox if user can't control it. There is already the tag to show that the user is part of the Neurons' Fund for informational purposes.

# Changes

* Do not render the Neurons' Fund in NnsNeuronAdvancedSection instead of disabled if user is not the controller.

# Tests

* Edit tests after change in functionality.

# Todos

Not worth an entry. This is part of the unpublished new neuron details.

### DIFF
--- a/frontend/src/tests/lib/components/neuron-detail/NnsNeuronAdvancedSection.spec.ts
+++ b/frontend/src/tests/lib/components/neuron-detail/NnsNeuronAdvancedSection.spec.ts
@@ -27,6 +27,10 @@ import { render } from "@testing-library/svelte";
 import NeuronContextActionsTest from "./NeuronContextActionsTest.svelte";
 
 describe("NnsNeuronAdvancedSection", () => {
+  const identityMainAccount = {
+    ...mockMainAccount,
+    principal: mockIdentity.getPrincipal(),
+  };
   const renderComponent = (neuron: NeuronInfo) => {
     const { container } = render(NeuronContextActionsTest, {
       props: {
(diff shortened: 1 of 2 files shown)


frontend/src/tests/lib/components/neuron-detail/NnsNeuronAdvancedSection.spec.ts
```diff
@@ -142,12 +132,10 @@ describe("NnsNeuronAdvancedSection", () => {
...
     });
 
-    expect(
-      await po.getJoinNeuronsFundCheckbox().getAttribute("disabled")
-    ).not.toBeNull();
+    expect(await po.getJoinNeuronsFundCheckbox().isPresent()).toBe(false);
   });
 
-  it("should render split button and disabled join neurons' fund if user is controlled by hardware wallet", async () => {
+  it("should render split button but not join neurons' fund if user is controlled by hardware wallet", async () => {
```
dskloetd, line 138: It's not neuron, not the user, that's controlled by the HW wallet, right?

dskloetd (approved):

frontend/src/tests/lib/components/neuron-detail/NnsNeuronAdvancedSection.spec.ts
dskloetd: I meant, it *is* the neuron.

lmuntaner: oops, yes, correct. I'll change it.

PR #3056, opened 2023-08-08
Package json license

peterpeterparker:
# Motivation

Add flags to package.json for better definition.

# Changes

`"license": "SEE LICENSE IN LICENSE.md",`

source: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#license

> If you are using a license that hasn't been assigned an SPDX identifier, or if you are using a custom license, use a string value like this one...


lmuntaner (approved):
Thanks!

PR #3057, opened 2023-08-08
GIX-1751: Allow hotkeys to manage Neurons' Fund participation

lmuntaner:
# Motivation

Allow hotkeys to manage the participation of the Neurons' Fund.

# Changes

* New util `canUserManageNeuronFundParticipation` to help with NF managenent.
* Use the new util to hide or show the checkbox in the NnsNeuronAdvancedSection.
* Use the new util to hide or show the NeuronJoinFundCard (old UI).
* Add extra text in the JoinCommunityFundModal when the user is a hotkey.
* Use the II identity in the neuron service `toggleCommunityFund`. All neurons present in UI are either controlled by the II user or it's a hotkey.

# Tests

* Test the new util `canUserManageNeuronFundParticipation`.
* Change and add test cases in NeuronJoinFundCard.
* Change and add test cases in NnsNeuronAdvancedSection.
* Add ConfirmationModalPo to add a test for JoinCommunityFundModal.
* Remove test case for neuron service `toggleCommunityFund`.

# Todos


lmuntaner:
New UX:

User is the controller:
Screenshot 2023-08-08 at 09 37 06

User is a hotkey, but neuron is NOT controlled by an attached hardware wallet
Screenshot 2023-08-08 at 09 36 35

User is a hotkey, but neuron IS controlled by an attached hardware wallet.
Screenshot 2023-08-08 at 09 36 51

New message when hotkey uses tries to join Neurons' Fund:
Screenshot 2023-08-08 at 09 36 41

lmuntaner:

dskloetd:
Add a change log entry?

frontend/src/lib/i18n/en.json
```diff
@@ -575,6 +575,8 @@
     "no_dissolve_bonus": "No dissolve delay bonus",
     "available_description": "Available",
     "join_community_fund_description": "Are you sure you want this neuron to <strong>join</strong> the neurons' fund?",
+    "join_community_fund_hw_alert_1": "Beware that the NNS Dapp doesn't support hardware wallets for SNS yet.",
+    "join_community_fund_hw_alert_2": "If this neuron is controlled by a hardware wallet, you will not be able to manage it from the NNS Dapp.",
```
dskloetd, line 579: Does "this neuron" refer to the neuron on the current page? Or does it refer to a hypothetical SNS neuron that doesn't exist yet but that the user would normally get control over later if the NF participates in a successful SNS swap?

I feel like it should be the latter but that's not clear from the text, IMO.

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -356,6 +356,24 @@ export const isHotKeyControllable = ({
   ) !== undefined &&
   fullNeuron.controller !== identity?.getPrincipal().toText();
 
+/**
+ * An identityt can manage the neurons' fund participation when one of the below is true:
+ * - User is the controller.
+ * - User is a hotkey, but a hardware wallet account is NOT the controller.
```
dskloetd, line 362: Should we clarify that a HW controlled neuron would technically work but that we don't allow it to avoid confusion because the resulting HW controlled SNS neurons wouldn't be supported by NNS dapp?

frontend/src/tests/lib/components/neuron-detail/NeuronJoinFundCard.spec.ts
```diff
@@ -48,14 +51,53 @@ describe("NeuronJoinFundCard", () => {
...
   });
 
-  it("renders no checkbox if user is not controller", () => {
+  it("renders join community fund checkbox if user is hotkey and controller not the hardware wallet", () => {
+    const neuron = {
+      ...mockNeuron,
+      fullNeuron: {
+        ...mockFullNeuron,
+        controller: "not-user-nor-hw",
+        hotKeys: [mockIdentity.getPrincipal().toText() as string],
```
dskloetd, line 60: I don't think it's possible for the identity itself to also be the hotkey, or at least it would be very unrealistic, right?
Shall we use a different principal for the hotkey to make the test more realistic/less confusing?

(I can also imagine a certain kind of bug that wouldn't be caught by using the same principal in the test.)

frontend/src/tests/lib/components/neuron-detail/NeuronJoinFundCard.spec.ts
```diff
@@ -48,14 +51,53 @@ describe("NeuronJoinFundCard", () => {
...
+    expect(queryByTestId("checkbox")).toBeInTheDocument();
+  });
+
+  it("renders no checkbox if user is hotkey and controller is the hardware wallet", () => {
+    const neuron = {
+      ...mockNeuron,
+      fullNeuron: {
+        ...mockFullNeuron,
+        controller: mockHardwareWalletAccount.principal.toText(),
+        hotKeys: [mockIdentity.getPrincipal().toText() as string],
```
dskloetd, line 79: I thought you don't want to reuse the mock identity for hardware wallets in tests?
Hence https://github.com/dfinity/nns-dapp/pull/3053

frontend/src/tests/lib/components/neuron-detail/NnsNeuronAdvancedSection.spec.ts
dskloetd, line 1: Same comments from the other test apply to this test.

frontend/src/tests/lib/services/neurons.services.spec.ts
```diff
@@ -453,33 +453,6 @@ describe("neurons-services", () => {
...
-      setNoIdentity();
-
-      await toggleCommunityFund(neuron);
-
-      expectToastError(en.error.missing_identity);
-      expect(spyJoinCommunityFund).not.toBeCalled();
-      expect(spyLeaveCommunityFund).not.toBeCalled();
-    });
-
-    it("should not update neuron if not controlled by user", async () => {
```
dskloetd, line 471: Why did you remove this test?
Should there be something else instead in the service test?

frontend/src/tests/lib/utils/neuron.utils.spec.ts
```diff
@@ -2077,4 +2079,103 @@ describe("neuron-utils", () => {
...
+        main: identityMainAccount,
+        subAccounts: [],
+        hardwareWallets: [mockHardwareWalletAccount],
+      };
+      const neuron: NeuronInfo = {
+        ...mockNeuron,
+        fullNeuron: {
+          ...mockNeuron.fullNeuron,
+          controller: mockHardwareWalletAccount.principal?.toText(),
+          hotKeys: [mockIdentity.getPrincipal().toText()],
```
dskloetd, line 2146: same as above

frontend/src/tests/page-objects/ConfirmationModal.page-object.ts
```diff
@@ -0,0 +1,16 @@
...
+export class ConfirmationModalPo extends BasePageObject {
+  private static TID = "confirmation-modal-component";
+
+  static under(element: PageObjectElement): ConfirmationModalPo {
+    return new ConfirmationModalPo(element.byTestId(ConfirmationModalPo.TID));
+  }
+
+  async getContentText(): Promise<string> {
+    return (
+      await this.root.byTestId("confirmation-modal-content").getText()
```
dskloetd, line 13: You can replace `this.root.byTestId("confirmation-modal-content").getText()` with `this.getText("confirmation-modal-content")`.

frontend/src/lib/i18n/en.json
lmuntaner: To the neuron on the curernt page.

This copy might change. We're waiting for the product copy. But fixing this can be done in another PR.

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: Yes, good point. I'll add it to this comment

frontend/src/tests/lib/components/neuron-detail/NeuronJoinFundCard.spec.ts
lmuntaner: Why not?

The identity has the principal of the II user. That's precisely the user we need as hotkey so that they neurons appear in the NNS Dapp.

frontend/src/tests/lib/components/neuron-detail/NeuronJoinFundCard.spec.ts
lmuntaner: Well, before the `mockHardwareWalletAccount.principal` was the same as `mockIdentity.getPrincipal()`.

That was messing with this test specifically.

This test needs the II user to be a hotkey, and the HW to be the controller.

Because they both had the same principal, the II user was considered the controller, not just a hotkey.

frontend/src/tests/page-objects/ConfirmationModal.page-object.ts
lmuntaner: ah, yes! I tried `this.root.getText(...)`. I forgot the method is at the PO level, not the element level.

Thanks

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -356,6 +356,28 @@ export const isHotKeyControllable = ({
   ) !== undefined &&
   fullNeuron.controller !== identity?.getPrincipal().toText();
 
+/**
+ * An identityt can manage the neurons' fund participation when one of the below is true:
```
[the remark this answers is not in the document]
lmuntaner: Thanks! Fun fact, I search for the typo and found it in other places. I'll change it in another PR.

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -356,6 +356,28 @@ export const isHotKeyControllable = ({
...
+export const canUserManageNeuronFundParticipation = ({
+  neuron,
+  identity,
+  accounts,
+}: {
+  neuron: NeuronInfo;
+  identity: Identity | null | undefined;
+  accounts: IcpAccountsStoreData;
+}): boolean =>
+  isNeuronControllableByUser({ neuron, mainAccount: accounts.main }) ||
```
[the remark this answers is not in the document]
lmuntaner: I'm not sure, the accounts are needed to return true. If the accounts are present, it means that the user is logged in.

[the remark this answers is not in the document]
peterpeterparker: Indeed there should be no accounts if not identity but, it's still a parameter therefore I'm unsure. I just had a look to other utils and we do for example that:

```
export const isUserHotkey = ({
  neuron,
  identity,
}: {
  neuron: SnsNeuron;
  identity: Identity | null | undefined;
}) =>
  identity === null || identity === undefined
    ? false
    : getSnsNeuronHotkeys(neuron).includes(identity.getPrincipal().toText());
```

So for that reason, I would add the check.

Not a super strong opinion though, more like guts feeling. If you rather like not, for sure it's fine.

frontend/src/lib/i18n/en.json
dskloetd: As discussed, I don't think it's relevant whether the current neuron is HW controlled.
What matters is:
1. NF participation can result in SNS neurons.
2. Controllers of the NF neurons get a hotkey to the SNS neurons.
3. Hotkeys of the NF neurons do not get a hotkey to the SNS neurons.
4. So only the controller of the NF neuron can manage the SNS neurons.
5. Since "you" refers to the current user, and the current user is not the controller of the NF neuron, *you* will not be able to manage the SNS neuron, regardless of what kind of hotkey this is.
6. If the controller of the NF neuron is not an NNS dapp principal (and HW principal are not NNS dapp principals), then the SNS neuron can not be managed via NNS dapp (by anyone!)

So I suggest something like the following message (instead of the 2 messages above):
Note that only the controllers of Neurons' Fund neurons get added as hotkeys to SNS neurons resulting from Neurons' Fund swap participation.
Since you are not the controller of this neuron, you will not be able to manage the resulting SNS neurons, even if you control this neuron with a hardware wallet.

frontend/src/tests/lib/components/neuron-detail/NeuronJoinFundCard.spec.ts
dskloetd: Sorry, I must have been very confused.

dskloetd (approved):

frontend/src/lib/modals/neurons/JoinCommunityFundModal.svelte
```diff
@@ -44,6 +54,14 @@
       <p>
         <Html text={$i18n.neuron_detail.join_community_fund_description} />
       </p>
+      {#if isHotkeyControlled}
```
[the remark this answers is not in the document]
lmuntaner: No, the user can land there because the controller is a hardware wallet, but not a hardware wallet attached to this user.

Maybe userA creates a neuron with HW, then adds userB as hotkey. The NNS Dapp has no way of knowing from userB perspective that the controller is a HW.

Does that make sense?

frontend/src/lib/modals/neurons/JoinCommunityFundModal.svelte
```diff
@@ -44,6 +54,14 @@
...
+        <p class="description">
```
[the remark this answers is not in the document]
lmuntaner: done, see below

frontend/src/lib/i18n/en.json
lmuntaner: done

frontend/src/lib/modals/neurons/JoinCommunityFundModal.svelte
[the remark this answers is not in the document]
peterpeterparker: Gotcha, thanks.

PR #3058, opened 2023-08-08
Fix disabled review cycles submit button

peterpeterparker:
# Motivation

When topping up a canister, on mainnet and locally, the "Review Cycles Purchase" is not disabled per default, therefore user can submit it without entering any amount which lead the wizard to an empty step.

# Changes

- if no minimal amount, request at least an amount to be entered

# Screenshots

tmp

### DIFF
--- a/frontend/src/lib/components/canisters/SelectCyclesCanister.svelte
+++ b/frontend/src/lib/components/canisters/SelectCyclesCanister.svelte
@@ -6,7 +6,6 @@
     convertTCyclesToIcpNumber,
   } from "$lib/utils/token.utils";
   import Input from "$lib/components/ui/Input.svelte";
-  import { isNullish, nonNullish } from "@dfinity/utils";
   import { areEnoughCyclesSelected } from "$lib/utils/canisters.utils";
 
   export let amount: number | undefined = undefined;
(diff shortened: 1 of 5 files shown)


frontend/src/lib/components/canisters/SelectCyclesCanister.svelte
```diff
@@ -60,8 +61,9 @@
   };
 
   let enoughCycles: boolean;
-  $: enoughCycles =
-    minimumCycles === undefined ? true : (amountCycles ?? 0) >= minimumCycles;
+  $: enoughCycles = isNullish(minimumCycles)
+    ? nonNullish(amountCycles)
```
lmuntaner, line 65: Won't this allow `0`? If the user enters `0`, is it still valid?

lmuntaner (approved):
Thanks! A question, but already approving

frontend/src/lib/components/canisters/SelectCyclesCanister.svelte
peterpeterparker: Yeah that's an excellent point, I just modified into `nonNullish` because it used to be `true` but, I think you are right.

Let's do the following?

```
let enoughCycles: boolean;
  $: enoughCycles =
    (amountCycles ?? 0) >= (minimumCycles ?? 0) && amountCycles > 0;
```

lmuntaner: Yes, that makes more sense!

Maybe create an util so that we can add some tests?

```
areEnoughCycles({ minimumCycles, amountCycles });
```

The tests would also help as documentation.

peterpeterparker: Sure, done

PR #3059, opened 2023-08-08
Ignore SIGPIPE in past-changelog-test

dskloetd:
# Motivation

`scripts/past-changelog-test` has been flaky recently, occasionally failing with exit code 141.
I believe the issue is that `grep -m 1` stops reading input as soon as it finds a match and then the process before the pipe fails because it can no longer write to the pipe.
I'm not sure why this happens only sometimes.

Some references:
https://stackoverflow.com/questions/19120263/why-exit-code-141-with-grep-q
https://unix.stackexchange.com/questions/582844/how-to-suppress-sigpipe-in-bash

# Changes

When the line with `grep -m 1` fails, ignore it if it failed with exit code 141.

# Tests

Difficult to test because I couldn't reproduce the issue (well, only once) in this script.
But it reproduces quite well with `yes | head -1` so I did some manual tests with that.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -56,6 +56,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Fixed
 
 * past-changelog-test compares lines numbers correctly.
+* Ignore SIGPIPE in scripts/past-changelog-test which caused flakiness.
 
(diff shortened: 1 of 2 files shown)


peterpeterparker (approved):
Rubber stamp, thanks

lmuntaner (approved):
Thanks!

PR #3063, opened 2023-08-09
Make it easy to skip the CI build step for quick testing

dskloetd:
# Motivation

Iterative testing on CI can be quite slow if every time you have to wait for the build step before your actual test runs.
If you're only changing your test and not your code under test it can be useful to skip the build step and reuse the artifacts from the previous build.

# Changes

In `.github/workflows/build.yml`, add a step to skip the build and reuse a previous build.
By default it's disabled, but with the code in place it's easy to enable.

When the build is skipped, the "Check that metadata is present" test will fail but everything else should pass as normal.

# Tests

With skipping disabled: https://github.com/dfinity/nns-dapp/actions/runs/5806927625
With skipping enabled: https://github.com/dfinity/nns-dapp/actions/runs/5807123115

# Todos

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Skip build for testing
         # Set to true and set a recent run_id below to reuse an existing build
         # instead of building.
-        if: true
+        if: false
         id: skip_build
         run: |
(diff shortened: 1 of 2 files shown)


lmuntaner (commented):
LGTM

lmuntaner (approved):

PR #3071, opened 2023-08-09
Chore: Test util to set SNS projects

lmuntaner:
# Motivation

Final motivation: Stop using get_state. But on the way, clean the unnecessary complexity of `snsQueryStore`.

In this PR: Create a test util that abstracts the filling and resetting of snsQueryStore in tests.

# Changes

* New test utils: resetSnsProjects, setSnsProject.
* Use new test utils in SnsAccountsFooter.spec.ts

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/tests/lib/components/accounts/SnsAccountsFooter.spec.ts
+++ b/frontend/src/tests/lib/components/accounts/SnsAccountsFooter.spec.ts
@@ -5,18 +5,16 @@
 import SnsAccountsFooter from "$lib/components/accounts/SnsAccountsFooter.svelte";
 import * as accountsServices from "$lib/services/sns-accounts.services";
 import { snsAccountsStore } from "$lib/stores/sns-accounts.store";
-import { snsQueryStore } from "$lib/stores/sns.store";
 import { transactionsFeesStore } from "$lib/stores/transaction-fees.store";
 import AccountsTest from "$tests/lib/pages/AccountsTest.svelte";
-import { mockPrincipal } from "$tests/mocks/auth.store.mock";
 import {
   modalToolbarSelector,
   waitModalIntroEnd,
 } from "$tests/mocks/modal.mock";
 import { mockSnsMainAccount } from "$tests/mocks/sns-accounts.mock";
-import { snsResponseFor } from "$tests/mocks/sns-response.mock";
+import { rootCanisterIdMock } from "$tests/mocks/sns.api.mock";
 import { testAccountsModal } from "$tests/utils/accounts.test-utils";
-import { Principal } from "@dfinity/principal";
+import { resetSnsProjects, setSnsProject } from "$tests/utils/sns.test-utils";
 import { SnsSwapLifecycle } from "@dfinity/sns";
 import { fireEvent, render, waitFor } from "@testing-library/svelte";
 import { page } from "../../../../../__mocks__/$app/stores";
@@ -28,19 +26,17 @@ jest.mock("$lib/services/sns-accounts.services", () => {
 });
 
 describe("SnsAccountsFooter", () => {
-  const responses = snsResponseFor({
-    principal: mockPrincipal,
-    lifecycle: SnsSwapLifecycle.Committed,
-  });
-
-  const rootCanisterIdText = responses[0][0].rootCanisterId;
-  const rootCanisterId = Principal.fromText(rootCanisterIdText);
+  const rootCanisterId = rootCanisterIdMock;
+  const rootCanisterIdText = rootCanisterId.toText();
 
   beforeEach(() => {
-    snsQueryStore.reset();
+    resetSnsProjects();
     snsAccountsStore.reset();
     transactionsFeesStore.reset();
-    snsQueryStore.setData(responses);
+    setSnsProject({
+      rootCanisterId,
+      lifecycle: SnsSwapLifecycle.Committed,
+    });
     transactionsFeesStore.setFee({
       rootCanisterId,
       fee: BigInt(10_000),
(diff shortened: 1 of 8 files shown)


frontend/src/tests/utils/sns.test-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+}) => {
+  const response = snsResponseFor({
+    principal: rootCanisterId,
+    lifecycle,
+    certified,
+    restrictedCountries,
+    directParticipantCount,
+    projectName,
+  });
+  snsQueryStore.setData(response);
```
dskloetd, line 29: This replaces all existing data in the store, right?
Don't we want/need to support setting more than one project?
So either adding to what's already there or specifying multiple projects in the parameters.

lmuntaner: Yes, it replaces the data.

Yes, good question. I didn't need it for this draft, but my plan is to add another function `setSnsProjects` (in plural), that has a different signature than this one (less configuration).

Something like:
```
const setSnsProjects = (params: { lifecycles: SnsSwapLifecycle[] }) => {...}
```

When multiple are needed, they don't care about rootCanisterId, just different lifecycles.

I will add the functions and extend them as I keep working on the tests.

dskloetd: OK, that's fine but I hope we won't need more than 2 functions.
This will be more generic and it might make sense to go with it right away:
```
const setSnsProjects = (params: { lifecycle: SnsSwapLifecycle }[]) => {...}
```

lmuntaner:
@dskloetd let me know what you think now

dskloetd:
Looks fine. Let's see how it looks with more tests using it.

lmuntaner:
> Looks fine. Let's see how it looks with more tests using it.

@dskloetd please review

I think this is enough changes for one PR. What do you think?

I tried adding one more "feature" to the test util. This probably still doesn't cover all use cases. But we can cover what we need with more PRs.

frontend/src/tests/lib/components/accounts/SnsAccountsFooter.spec.ts
```diff
@@ -28,19 +26,19 @@ jest.mock("$lib/services/sns-accounts.services", () => {
...
-  });
-
-  const rootCanisterIdText = responses[0][0].rootCanisterId;
-  const rootCanisterId = Principal.fromText(rootCanisterIdText);
+  const rootCanisterId = rootCanisterIdMock;
+  const rootCanisterIdText = rootCanisterId.toText();
 
   beforeEach(() => {
-    snsQueryStore.reset();
+    resetSnsProjects();
```
dskloetd, line 33: `resetSnsProjects` can be removed because `setSnsProjects` sets the full state, right?

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronInfoStake.spec.ts
```diff
@@ -29,16 +29,14 @@ import {
...
   beforeEach(() => {
-    const universe = data[0][0].rootCanisterId;
+    const rootCanisterId = rootCanisterIdMock;
     page.mock({
-      data: { universe },
+      data: { universe: rootCanisterId.toText() },
     });
 
-    snsQueryStore.setData(data);
+    resetSnsProjects();
```
dskloetd, line 38: Is this redundant as well?

frontend/src/tests/lib/derived/sns/sns-selected-project-new-tx-data.derived.spec.ts
```diff
@@ -3,48 +3,56 @@
...
 
-      const rootCanisterIdText = projectData[0][0].rootCanisterId;
-      const tokenData = projectData[0][0].token;
-      const rootCanisterId = Principal.fromText(rootCanisterIdText);
+    it("returns the data for the a new Tx from the current universe", () => {
+      const token = {
+        name: "name",
+        symbol: "symbol",
+      };
+      const tokenData = createQueryMetadataResponse(token);
```
dskloetd, line 40: Would it be too limiting to do this inside `setSnsProjects` and pass `token` instead of `tokenData`?

frontend/src/tests/lib/components/accounts/SnsAccountsFooter.spec.ts
lmuntaner: Yes, true!

frontend/src/tests/lib/derived/sns/sns-selected-project-new-tx-data.derived.spec.ts
lmuntaner: For application helpes, I'd say yes. But for test helpers I prefer to keep things as raw as possible. So that we understand what's going on exactly. Sometimes a test might be failing because some magic happened in a helper that we didn't see.

lmuntaner: Do you have a strong opinion on doing it inside?

Actually, we do something similar with the `state` of the neuron. The logic is inside the helper.

I'll keep it consistent and move it inside then.

lmuntaner: done

frontend/src/tests/lib/derived/sns/sns-selected-project-new-tx-data.derived.spec.ts
```diff
@@ -22,7 +22,6 @@ describe("selected-project-new-transaction-data derived store", () => {
 
     beforeEach(() => {
       page.mock({ data: { universe: OWN_CANISTER_ID_TEXT } });
-      resetSnsProjects();
```
dskloetd: I didn't think this one should be removed because there is no `setSnsProjects()` in the `beforeEach`.

lmuntaner:

frontend/src/tests/lib/derived/sns/sns-selected-project-new-tx-data.derived.spec.ts
lmuntaner: But I setSnsProjects in all the test cases.

I can add it in case we add one more test case. It might not fail because it uses the sns projects of a previous test, and that wouldn't be intended.

dskloetd (approved):

PR #3072, opened 2023-08-10
E2E: Replace content instead of masking

dskloetd:
# Motivation

We mask the user's wallet address before taking a screenshot because it depends on which anchor we sign in with (as well as some randomness in the snsdemo snapshot).
We also mask the SNS names and token names because they are created randomly in the snsdemo snapshot.

However, because we don't use a fixed-width font, the size of the masked area still changes when the address/name changes.

Solutions considered:
1. Use a fixed size masking rectangle with a size specified in the test. This would have to be a custom implementation because it's not supported by Playwright.
2. Always sign in with anchor 10000. This would probably work but only until the snsdemo snapshot is recreated.
3. Use a fixed width font for the masked areas (either for the actual app, or only for the e2e test).
4. Mask a parent element of which the width does not depend on the content.

Some of these might work fine but I went with:
5. Replace the contents of the relevant elements "XXXXX".

# Changes

1. Add a utility function for e2e tests to replace the content of selected elements with something else.
2. Use it in `design.spec.ts` to replace previously masked elements' content with "XXXXX".
3. Update screenshots.

# Tests

The screenshots look as expected.

# Todos

Covered by previous entry.


frontend/src/tests/utils/e2e.test-utils.ts
```diff
@@ -96,3 +96,21 @@ export const setFeatureFlag = ({
...
+  innerHtml,
+}: {
+  page: Page;
+  selectors: string[];
+  innerHtml: string;
+}) => {
+  page.evaluate(
+    ({ selectors, innerHtml }) =>
+      document.querySelectorAll(selectors.join(", ")).forEach((el) => {
+        el.innerHTML = innerHtml;
```
peterpeterparker, line 112: Perhaps we should consider limiting the scope of this utility to only replacing the `textContent`. Otherwise, if the selector isn't accurate, it could lead to replacing a "complex" element with just plain text. This might result in more significant layout changes than intended.

frontend/src/tests/e2e/design.spec.ts
```diff
@@ -38,17 +38,16 @@ test.describe("Design", () => {
...
-          page.locator(
-            '[data-tid="select-universe-card"]:not(:first-of-type) .name'
-          ),
-          page.locator(
-            '[data-tid="select-universe-card"]:not(:first-of-type) .amount .label'
-          ),
+      replaceContent({
+        page,
+        selectors: [
+          '[data-tid="identifier"]',
```
lmuntaner, line 44: Can't we use the page objects to replace the content?

This way, when the html changes or data-tid, the test doesn't need to change.

lmuntaner (commented):
Thanks for looking into it!

Just a question

frontend/src/tests/e2e/design.spec.ts
dskloetd: It's a reasonable question but I feel the use case is quite specific and I'm not sure using page objects here is over engineering.

Do you feel strongly?

frontend/src/tests/utils/e2e.test-utils.ts
dskloetd: How do you replace just the textContent of a complex element without removing its complexity?

lmuntaner (approved):
Let's merge this and work on a better solution in the meantime

dskloetd:
I'm going to merge this to stop the flakiness.
Happy to discuss other solutions, or review PRs, but I'm not convinced it's a priority.

frontend/src/tests/utils/e2e.test-utils.ts
peterpeterparker: Yeah you right at the end of the day, same same. Accurate implementation would require to test that the child of the element is stricly only a text node.

PR #3073, opened 2023-08-10
Fix: "Send ICP" instead of "Stake Neuron" in HW

lmuntaner:
# Motivation

Bug:

* Visit HW wallet page.
* Click on "Show Neurons". Finish process.
* Go to "Neurons".
* Stake a neuron from the HW account.

Expected: "Stake Neuron" appears in the ledger device.
Instead: "Send ICP" appeared in the ledger device.

Reason:

`createAgent` caches the agents per identity. On "Show Neuron" a new agent was created and cached.

Yet, when staking a neuron, a new identity was created. The neurons flag was set in the new identity, but the api layer was using the cached agent with the previous identity. And the neurons flag was not set in it.

# Changes

* Add an accountIdentifier to the `listNeuronsHardwareWallet` service.
* Ledger service `listNeuronsHardwareWallet` uses `getLedgerIdentity` which uses the cache to create and return the identity.

# Tests

* Add tests to `listNeuronsHardwareWallet` to check that identity is cached and reused if previously cached.

# Todos

### DIFF
--- a/frontend/src/lib/services/icp-ledger.services.ts
+++ b/frontend/src/lib/services/icp-ledger.services.ts
@@ -34,7 +34,7 @@ export const connectToHardwareWallet = async (
   try {
     callback({ connectionState: LedgerConnectionState.CONNECTING });
 
-    const ledgerIdentity: LedgerIdentity = await createLedgerIdentity();
+    const ledgerIdentity: LedgerIdentity = await LedgerIdentity.create();
 
     callback({
       connectionState: LedgerConnectionState.CONNECTED,
@@ -95,9 +95,6 @@ export const registerHardwareWallet = async ({
   }
 };
 
-const createLedgerIdentity = (): Promise<LedgerIdentity> =>
-  LedgerIdentity.create();
-
 // Used to cache the identities.
 // The identity is an instance of the class LedgerIdentity.
 // This identity is used in other layers and then cached. If this returns a new one it's not necessarily used.
@@ -118,7 +115,7 @@ export const getLedgerIdentity = async (
   if (identities[identifier]) {
     return identities[identifier];
   }
-  const ledgerIdentity: LedgerIdentity = await createLedgerIdentity();
+  const ledgerIdentity: LedgerIdentity = await LedgerIdentity.create();
 
   identities[identifier] = ledgerIdentity;
 
@@ -142,7 +139,7 @@ export const getLedgerIdentity = async (
 
 export const showAddressAndPubKeyOnHardwareWallet = async () => {
   try {
-    const ledgerIdentity: LedgerIdentity = await createLedgerIdentity();
+    const ledgerIdentity: LedgerIdentity = await LedgerIdentity.create();
     await ledgerIdentity.showAddressAndPubKeyOnDevice();
   } catch (err: unknown) {
     toastUnexpectedError({
@@ -166,12 +163,16 @@ const toastUnexpectedError = ({
     })
   );
 
-export const listNeuronsHardwareWallet = async (): Promise<{
+export const listNeuronsHardwareWallet = async (
+  accountIdentifier: string
+): Promise<{
   neurons: NeuronInfo[];
   err?: string;
 }> => {
   try {
-    const ledgerIdentity: LedgerIdentity = await createLedgerIdentity();
+    const ledgerIdentity: LedgerIdentity = await getLedgerIdentity(
+      accountIdentifier
+    );
     const neurons: NeuronInfo[] = await queryNeurons({
       identity: ledgerIdentity,
       certified: true,
--- a/frontend/src/tests/lib/services/icp-ledger.services.spec.ts
+++ b/frontend/src/tests/lib/services/icp-ledger.services.spec.ts
@@ -314,7 +314,7 @@ describe("icp-ledger.services", () => {
   describe("query neurons", () => {
     const mockNeurons = [mockNeuron];
 
-    beforeAll(() => {
+    beforeEach(() => {
       jest
         .spyOn(api, "queryNeurons")
         .mockImplementation(() => Promise.resolve(mockNeurons));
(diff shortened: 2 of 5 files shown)


lmuntaner:
@dskloetd please review

I solved the issue following the pattern we have, which is to pass the identifier and rely on caching the identity. Yet, I don't think this is a long-term solution.

I believe we should look into two parts:
* The idea of caching agent and identities.
* Using a flag in the identity to identify the stake neuron transaction.

frontend/src/lib/services/icp-ledger.services.ts
```diff
@@ -118,7 +115,7 @@ export const getLedgerIdentity = async (
   if (identities[identifier]) {
     return identities[identifier];
   }
-  const ledgerIdentity: LedgerIdentity = await createLedgerIdentity();
+  const ledgerIdentity: LedgerIdentity = await LedgerIdentity.create();
```
dskloetd, line 118: I'm trying to understand how, on line 126, we can expect `ledgerIdentifier` to equal `identifier`.
`identifier` is passed into `getLedgerIdentity` as an argument.
But `identifier` is not passed into `LedgerIdentity.create()`.
Then `ledgerIdentifier` is created from the result of `LedgerIdentity.create()`.
How then can `ledgerIdentifier` related in any way to the unused `identifier` parameter?

frontend/src/tests/lib/services/icp-ledger.services.spec.ts
dskloetd, line 1: Can we have a test where we test that we actually send the right request to the Ledger to show the correct display?

When we test the caching we are testing an implementation detail but what actually matters is what's displayed by the Ledger.

frontend/src/lib/services/icp-ledger.services.ts
dskloetd, line 1: This is the second time we've had issues with this recently. I think now is the time to do a proper solution rather than later.

frontend/src/lib/services/icp-ledger.services.ts
lmuntaner: This is to make sure that the connected HW is the intended hardware wallet.

`LedgerIdentity.create` connects to the device.

frontend/src/tests/lib/services/icp-ledger.services.spec.ts
lmuntaner: I don't think the effort of such a test is worth it.

I agree that the tests are not the best. But as I mentioned, I'm also not happy with the current pattern.

If we spend time with this, I'd rather spend time on refactoring than on adding such a test.

frontend/src/lib/services/icp-ledger.services.ts
dskloetd: Yes, that's what I mean.

frontend/src/lib/services/icp-ledger.services.ts
dskloetd: I really don't understand. How can the result of `principalToAccountIdentifier(ledgerIdentity.getPrincipal());` be equal to an arbitrary parameter that was passed in, while that parameter wasn't used to create `ledgerIdentity`?
It looks like magic to me and I'd like to understand it.

frontend/src/tests/lib/services/icp-ledger.services.spec.ts
dskloetd: We would need the same test with the other pattern.
The test would actually be useful because it would allow us to fix the code while keeping the same test.

frontend/src/lib/services/icp-ledger.services.ts
lmuntaner: Imagine I attach a hardware wallet to an II user. For that, I connected the device with the USB.

Next time I log in, it remembers which HW I attached.

Yet, now I connect another device with the USB cable. This check will fail becaseu the identifier of the new device doesn't match the one I connected initially.

frontend/src/tests/lib/services/icp-ledger.services.spec.ts
lmuntaner: A pattern we still don't have and we don't know how it's going to look like yet.

Still, with the other pattern. I would first spend time developing the new pattern, and then based on priorities and the value it give, I would consider implementing this test.

frontend/src/lib/services/icp-ledger.services.ts
lmuntaner: Ok, no worries. I'll ask someone else.

I don't agree with that approach.

lmuntaner:
@peterpeterparker @mstrasinskis could you please take a look?

@dskloetd thinks we should change the pattern instead of fixing the issue first.

I agree that changing the pattern is necessary. But I don't think it should block this PR from merging.

Thanks!

frontend/src/tests/lib/services/icp-ledger.services.spec.ts
dskloetd: This code is clearly fragile. That makes having a proper test, not something that just test implementation, extra important.

frontend/src/lib/services/icp-ledger.services.ts
dskloetd: But if an identity with that identifier was already cached, we just return it?
Shouldn't `getLedgerIdentity` also fail if the wrong device is attached even if we have a cached identity?

lmuntaner: Yes, good point. That's a case not covered by the code.

* User does an action with one attached device.
* User does another action with another attached device.

I'll add a todo on my list to address this. It should be fast to do.

dskloetd:
Can we spend a little bit of time to understand how difficult it is to do a proper fix before we decide not to do it properly now?

lmuntaner:
> Can we spend a little bit of time to understand how difficult it is to do a proper fix before we decide not to do it properly now?

Yes, makes sense. But let's do it in the office. This needs some brainstorming. I'll come to the office tomorrow. How about before lunch, at 11.30?

dskloetd:
Let's do it after lunch.
I already wrote down some thoughts that we can discuss:
1. In createAgent, don't cache agents for HW wallets.
2. In createAgent, only use a cached agent if the identity is identical (not just for the same principal)
3. Can we decode the request to recognize the stake transaction?
4. Can we put anything on the request, instead of on the identity, to make it as a stake transaction?
5. If I set the stake flag on a normal Send ICP transaction, I get an error. So Ledger knows it's not a stake neuron transaction? Then why do we need the flag?

lmuntaner:
> Let's do it after lunch. I already wrote down some thoughts that we can discuss:
> 
> 1. In createAgent, don't cache agents for HW wallets.

I think this is just a fix also. Not really changing patterns.

> 2. In createAgent, only use a cached agent if the identity is identical (not just for the same principal)

Similar.

> 3. Can we decode the request to recognize the stake transaction?

It's a Send ICP transaction.

We could check the destination of the transaction. Which is NNS governance. But I am not convinced that's enough, for example, topping up neurons does also a transaction to the NNS governance.

> 4. Can we put anything on the request, instead of on the identity, to make it as a stake transaction?

The memo could be the thing, but it's needed to stake the neuron. Maybe we could read this to identify the transactions. But I'm not sure we wouldn't run into issues afterwards.

> 5. If I set the stake flag on a normal Send ICP transaction, I get an error. So Ledger knows it's not a stake neuron transaction? Then why do we need the flag?

Because there is a check inside the Ledger App that when the flag is set, the destination account should be an NNS Governance subaccount.

One more idea:

Create our own Agent as well as our own Identity. Maybe we could create two different agent-js, one for staking one for the others. Then, the api layer decides to build one agent or the other. Instead of setting a flag in services.

lmuntaner:
Closing in favor of #3171

PR #3074, opened 2023-08-10
Attempt to fix flaky broken pipe in past-changelog-test

dskloetd:
# Motivation

Even with [PR ](https://github.com/dfinity/nns-dapp/pull/3059) we still see flakiness in `past-changelog-test`:
https://github.com/dfinity/nns-dapp/actions/runs/5818236522/job/15774259586?pr=3064

Here is another attempt to fix it.

# Changes

Instead of ignoring the broken pipe, avoid the broken pipe by reading all the output from `git show` by sending the remaining output to `/dev/null`.

# Tests

I tested that the following produces exit code 141:
```
seq 10000 | (head -1); echo $?
```
while the following does not:
```
seq 10000 | (head -1; cat > /dev/null); echo $?
```
So hopefully this will work for the flaky test as well.

# Todos


lmuntaner (approved):
🤞

dskloetd:
It failed again ON THIS PR with
```
++ grep -m 1 '^## Proposal [0-9]\+$'
++ git show origin/main:CHANGELOG-Nns-Dapp.md
+ LAST_RELEASE_HEADING='## Proposal 123921'
Error: Process completed with exit code 141.
```

:'(

lmuntaner:
> It failed again ON THIS PR with

😢

dskloetd:
Actually, both the original fix and the new fix might have worked.
Because the failure didn't happen in `scripts/past-changelog-test` but rather in `scripts/past-changelog-test.test`; the test of the test.
It had the same line but without the fix.

It's a bit ugly to have the same fix in both places but I'm also not sure it's worth making a whole new script just for this one `grep` statement. What do you think?

lmuntaner:
> Actually, both the original fix and the new fix might have worked. Because the failure didn't happen in `scripts/past-changelog-test` but rather in `scripts/past-changelog-test.test`; the test of the test. It had the same line but without the fix.
> 
> It's a bit ugly to have the same fix in both places but I'm also not sure it's worth making a whole new script just for this one `grep` statement. What do you think?

I agree. If this keeps failing, then we might want to review the script and test. But not for now.

PR #3077, opened 2023-08-10
Test simulate merge neuron failure

dskloetd:
# Motivation

While cleaning up the `ENABLE_SIMULATE_MERGE_NEURONS` feature flag, I noticed a test was missing for when the simulate call fails.

# Changes

1. Add a test to check that the merge result section is hidden and a toast is shown when the simulate call fails.
2. In the fake, make `mergeNeurons` and `simulateMergeNeurons` fail if not both neurons are locked. This allows for a simple and realistic way to test scenarios where these calls fail.
3. Implement `startDissolving` on the fake API to make `simulateMergeNeurons` fail.

# Tests

Test only

# Todos

not necessary

### DIFF
--- a/frontend/src/tests/fakes/governance-api.fake.ts
+++ b/frontend/src/tests/fakes/governance-api.fake.ts
@@ -1,4 +1,5 @@
 import type {
+  ApiManageNeuronParams,
   ApiMergeNeuronsParams,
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
LGTM!

PR #3080, opened 2023-08-10
Chore: Use SNS Aggregator for nervous system functions

lmuntaner:
# Motivation

No need to request the nervous functions of an SNS. The aggregator already provides the data.

# Changes

* New derived store factory `createSnsFunctionsProjectStore`. It checks first the functions store and then aggregator to return the nervous functions.
* Use the new derived store factory in `SnsNeuronFollowingCard`.
* Check the new derived store in `loadSnsNervousSystemFunctions`.

# Tests

* Test `createSnsFunctionsProjectStore`.
* The changes in `loadSnsNervousSystemFunctions` exposed a missing store reset in the SnsProposalDetail test.
* Add test case in `loadSnsNervousSystemFunctions`.

# Todos

Not worth an entry, only refactor.

### DIFF
--- a/frontend/src/lib/derived/sns-ns-functions-project.derived.ts
+++ b/frontend/src/lib/derived/sns-ns-functions-project.derived.ts
@@ -1,0 +1,57 @@
+import {
+  snsAggregatorStore,
+  type SnsAggregatorStore,
+} from "$lib/stores/sns-aggregator.store";
+import {
+  snsFunctionsStore,
+  type SnsNervousSystemFunctionsStore,
+} from "$lib/stores/sns-functions.store";
+import type { CachedSnsDto } from "$lib/types/sns-aggregator";
+import { convertNervousFuncttion } from "$lib/utils/sns-aggregator-converters.utils";
+import type { Principal } from "@dfinity/principal";
+import type { SnsNervousSystemFunction } from "@dfinity/sns";
+import { nonNullish } from "@dfinity/utils";
+import { derived, type Readable } from "svelte/store";
+
+export type SnsNervousSystemFunctionsProjectStore = Readable<
+  SnsNervousSystemFunction[] | undefined
+>;
+
+const stores: Record<string, SnsNervousSystemFunctionsProjectStore> = {};
+
+const initSnsParametersProjectStore = (
+  rootCanisterId: Principal
+): SnsNervousSystemFunctionsProjectStore =>
+  derived<
+    [SnsNervousSystemFunctionsStore, SnsAggregatorStore],
+    SnsNervousSystemFunction[] | undefined
+  >(
+    [snsFunctionsStore, snsAggregatorStore],
+    ([snsFunctions, aggregatorData]) => {
+      const rootCanisterIdText = rootCanisterId.toText();
+      if (nonNullish(snsFunctions[rootCanisterIdText])) {
+        return snsFunctions[rootCanisterIdText].nsFunctions;
+      }
+      const aggregatorProject: CachedSnsDto | undefined =
+        aggregatorData.data?.find(
+          ({ canister_ids: { root_canister_id } }) =>
+            rootCanisterIdText === root_canister_id
+        );
+      if (nonNullish(aggregatorProject)) {
+        return aggregatorProject.parameters.functions.map(
+          convertNervousFuncttion
+        );
+      }
+      return undefined;
+    }
+  );
+
+export const getOrCreateSnsParametersProjectStore = (
+  rootCanisterId: Principal
+): SnsNervousSystemFunctionsProjectStore => {
+  const key = rootCanisterId.toText();
+  if (!stores[key]) {
+    stores[key] = initSnsParametersProjectStore(rootCanisterId);
+  }
+  return stores[key];
+};
--- a/frontend/src/lib/derived/sns-ns-functions-project.derived.ts
+++ b/frontend/src/lib/derived/sns-ns-functions-project.derived.ts
@@ -17,7 +17,7 @@ export type SnsNervousSystemFunctionsProjectStore = Readable<
   SnsNervousSystemFunction[] | undefined
 >;
 
-export const createSnsParametersProjectStore = (
+export const createSnsNsFunctionsProjectStore = (
   rootCanisterId: Principal
 ): SnsNervousSystemFunctionsProjectStore =>
   derived<
--- a/frontend/src/lib/derived/sns-ns-functions-project.derived.ts
+++ b/frontend/src/lib/derived/sns-ns-functions-project.derived.ts
@@ -17,9 +17,7 @@ export type SnsNervousSystemFunctionsProjectStore = Readable<
   SnsNervousSystemFunction[] | undefined
 >;
 
-const stores: Record<string, SnsNervousSystemFunctionsProjectStore> = {};
-
-const initSnsParametersProjectStore = (
+export const createSnsParametersProjectStore = (
   rootCanisterId: Principal
 ): SnsNervousSystemFunctionsProjectStore =>
   derived<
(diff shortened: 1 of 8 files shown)


frontend/src/lib/derived/sns-ns-functions-project.derived.ts
```diff
@@ -0,0 +1,57 @@
...
+      return undefined;
+    }
+  );
+
+export const getOrCreateSnsParametersProjectStore = (
+  rootCanisterId: Principal
+): SnsNervousSystemFunctionsProjectStore => {
+  const key = rootCanisterId.toText();
+  if (!stores[key]) {
+    stores[key] = initSnsParametersProjectStore(rootCanisterId);
```
dskloetd, line 54: I'm suddenly wondering, do we really need to cache this or can we just create it every time?
How many simultaneous listeners do we expect realistically? Wouldn't it be fine and much simpler not to cache it?

lmuntaner: The idea of caching it was to avoid calling the init function multiple times.

So, I don't think it's strictly needed.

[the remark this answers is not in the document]
lmuntaner: I like the type safety of rootCanisterId as a Principal.

Creating the type `type RootCanisterIdText = string;` is not the same.

frontend/src/lib/services/$public/sns.services.ts
```diff
@@ -185,9 +186,10 @@ export const loadProposalsSnsCF = async (): Promise<void> => {
 export const loadSnsNervousSystemFunctions = async (
   rootCanisterId: Principal
 ) => {
-  const store = get(snsFunctionsStore);
-  // Avoid loading the same data multiple times if the data loaded is certified
-  if (store[rootCanisterId.toText()]?.certified) {
+  const store = getOrCreateSnsParametersProjectStore(rootCanisterId);
+  const storeData = get(store);
+  // Avoid loading the same data multiple times if the data is loaded
```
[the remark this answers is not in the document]
lmuntaner: It's a good question, but it's good as is now.

The `getOrCreateSnsParametersProjectStore` uses either the store populated in this service or the sns aggregator.

If the store data is not present, this service will call the actual endpoint to get the functions. I don't want to store that data in the sns aggregator store. Instead, I want to keep storing it in its own store.

If the data is loaded in that store and then this service is called again; then the store from `getOrCreateSnsParametersProjectStore` will read from the previous store and skip the call as expected.

frontend/src/lib/derived/sns-ns-functions-project.derived.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { Principal } from "@dfinity/principal";
+import type { SnsNervousSystemFunction } from "@dfinity/sns";
+import { nonNullish } from "@dfinity/utils";
+import { derived, type Readable } from "svelte/store";
+
+export type SnsNervousSystemFunctionsProjectStore = Readable<
+  SnsNervousSystemFunction[] | undefined
+>;
+
+export const createSnsParametersProjectStore = (
```
dskloetd, line 20: Is `SnsParameters` the same as `SnsNervousSystemFunctions`

frontend/src/lib/derived/sns-ns-functions-project.derived.ts
```diff
@@ -0,0 +1,45 @@
...
+    SnsNervousSystemFunction[] | undefined
+  >(
+    [snsFunctionsStore, snsAggregatorStore],
+    ([snsFunctions, aggregatorData]) => {
+      const rootCanisterIdText = rootCanisterId.toText();
+      if (nonNullish(snsFunctions[rootCanisterIdText])) {
+        return snsFunctions[rootCanisterIdText].nsFunctions;
+      }
+      const aggregatorProject: CachedSnsDto | undefined =
+        aggregatorData.data?.find(
```
dskloetd, line 34: Before we continue, should we consider how it would be if the aggregator data itself was also split into dynamic stores per rootCanisterId?
Is that the direction we want to go in?

frontend/src/lib/services/$public/sns.services.ts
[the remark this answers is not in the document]
dskloetd: Thanks for clarifying.

frontend/src/lib/derived/sns-ns-functions-project.derived.ts
lmuntaner: no, sorry, the sns aggregator calls the field `params`, which is what I thought they were initially I started calling this like so. I thought I renamed everything.

I want to add the parameters in another field of the SNS aggregator.

frontend/src/lib/derived/sns-ns-functions-project.derived.ts
lmuntaner: Do you mean if the sns aggregator returns an object instead of an array of SNSes?

I also don't find storing it as an object already now. But I also think there's value in storing the response as it comes.

We could add a derived store and access the data through that, instead of this one.

But to be honest, I don't think this will grow to even thuosands. We will need a better solution before that happens.

dskloetd: I mean whether there will be something like
```
getOrCreateSnsAggregatorProjectStore(rootCanisterId)
```
which returns a derived store that holds data for a single SNS derived from the full aggregator store.

lmuntaner: Ah, ok. I guess we could also do that.

It could even return the "ic-js" types instead of the raw data.

I like that. Shall I do it in this PR or another one?

dskloetd: Not in this PR. I just thought we might want to think about how it would look and if it would interfere with this code here.
Or you can do it in this PR if it would help you validate that.

lmuntaner: I'll do it in another PR. I think it's a good idea. I'll add the tests here and I'll set it for review

lmuntaner:
@dskloetd please review

This is ready to be merged.

frontend/src/tests/lib/derived/sns-ns-functions-project.derived.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  it("returns the functions from snsFunctionsStore when snsAggregatorStore has data", () => {
+    snsAggregatorStore.setData([aggregatorProject]);
+    snsFunctionsStore.setProjectFunctions({
+      rootCanisterId,
+      nsFunctions: [nervousSystemFunctionMock],
+      certified: true,
+    });
+
+    const store = createSnsNsFunctionsProjectStore(rootCanisterId);
+    expect(get(store)).toEqual([nervousSystemFunctionMock]);
```
dskloetd, line 46: Can you also check that it's different from `get(snsAggregatorStore).parameters.functions.map(convertNervousFuncttion)` to be sure we're not actually getting it from the aggregator?

lmuntaner: sure, good point

lmuntaner: done

dskloetd (approved):

frontend/src/lib/derived/sns-ns-functions-project.derived.ts
lmuntaner: I've been thinking about this idea:

```
getOrCreateSnsAggregatorProjectStore(rootCanisterId)
```

I really like it, but not here. Instead, I believe that where it will have value is to use it in the components.

As the next step of the refactor. What I mentioned of changing the usage of the data. For example, we could do `getOrCreateSnsSummary(rootCanisterId)`.

This way we might not need to change so many things at once.

dskloetd: Why not here? If it existed already, wouldn't we be using it here instead of finding the specific project here?

lmuntaner: I don't see the value of doing it for the aggregator data.

Instead, it should be for data from multiple sources.

PR #3083, opened 2023-08-11
GIX-1797: Add HW controller tag

lmuntaner:
# Motivation

Users can identify which neurons are controlled by an attached hardware wallet.

In this PR, I introduce a new neuron tag: "Hardware Wallet".

# Changes

* New util `isHotkeyFlag`.
* Use `isNeuronControlledByHardwareWallet` and `isHotkeyFlag` in NnsNeuronPageHeading to render the tags. 
* Use `isNeuronControlledByHardwareWallet` and `isHotkeyFlag` in NnsNeuronCardTitle to render the tags.

# Tests

* Add a test case for `NnsNeuronPageHeading`.
* Add a test case for `NnsNeuronCard` where `NnsNeuronCardTitle` is tested.
* Add method to POs `NnsNeuronPageHeading` and `NnsNeuronCardPo`.
* Test new util `isHotkeyFlag`.

# Todos

### DIFF
--- a/frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
+++ b/frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
@@ -5,7 +5,7 @@
   import {
     formatVotingPower,
     hasJoinedCommunityFund,
-    isHotkeyFlag,
+    isHotkeyTag,
     isNeuronControlledByHardwareWallet,
     neuronStake,
   } from "$lib/utils/neuron.utils";
@@ -33,8 +33,8 @@
   let isCommunityFund: boolean;
   $: isCommunityFund = hasJoinedCommunityFund(neuron);
 
-  let hotkeyFlag: boolean;
-  $: hotkeyFlag = isHotkeyFlag({
+  let hotkeyTag: boolean;
+  $: hotkeyTag = isHotkeyTag({
     neuron,
     identity: $authStore.identity,
     accounts: $icpAccountsStore,
@@ -64,7 +64,7 @@
         {$i18n.neurons.community_fund}
       </HeadingTag>
     {/if}
-    {#if hotkeyFlag}
+    {#if hotkeyTag}
       <HeadingTag testId="hotkey-tag">
         {$i18n.neurons.hotkey_control}
       </HeadingTag>
--- a/frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
+++ b/frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
@@ -5,7 +5,6 @@
   import {
     formatVotingPower,
     hasJoinedCommunityFund,
-    isHotKeyControllable,
     isHotkeyFlag,
     isNeuronControlledByHardwareWallet,
     neuronStake,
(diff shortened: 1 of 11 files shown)


lmuntaner:
New UI:

Screenshot 2023-08-11 at 08 27 18
Screenshot 2023-08-11 at 08 31 26

lmuntaner:
@dskloetd please review

I might create a helper to return all the tags that should be displayed. But I'd rather do that in a separate PR.

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -364,6 +364,19 @@ export const isHotKeyControllable = ({
   ) !== undefined &&
   fullNeuron.controller !== identity?.getPrincipal().toText();
 
+// All HW controlled are hotkeys, but we don't want to show two tags to the user.
+export const isHotkeyFlag = ({
```
dskloetd, line 368: What is the word `Flag` indicating here?
I think a longer name might be clearer.

frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
```diff
@@ -31,10 +33,17 @@
...
-  let isHotKeyControl: boolean;
-  $: isHotKeyControl = isHotKeyControllable({
+  let hotkeyFlag: boolean;
+  $: hotkeyFlag = isHotkeyFlag({
     neuron,
     identity: $authStore.identity,
+    accounts: $icpAccountsStore,
+  });
+
+  let isHWControlled: boolean;
```
dskloetd, line 43: Given that we want these tags to be mutually exclusive, maybe it would be clearer to have 1 function:
```
const getHotKeyTag = ({neuron, accounts}) => {
  if (...) {
    return $i18n.neurons.hardware_wallet_control;
  }
  if (...) {
    return $i18n.neurons.hotkey_control;
  }
  return null;
}
```
What do you think?

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: True, I actually meant "tag". How about `isHotkeyTag`?

frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
lmuntaner: I think they should all be coupled so that the tags are the same everywhere. There might be other couplings in the future.

That's why I explained:

> I might create a helper to return all the tags that should be displayed. But I'd rather do that in a separate PR.

I think it's better if they are all in the same helper and there is one logic for all tags.

dskloetd: I agree it's better. I didn't see your comment (I just read the PR description).
Do you think it's more difficult? It seems like it would have been easier to do that right away.

dskloetd (approved):

frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
lmuntaner: I wanted to split it in two PRs.

Otherwise I would be refactoring and adding functionality.

When I refactor, the components tests shouldn't change. It's a nice way to test the refactor.

dskloetd: But I expect that the tests will change. Because instead of
```
    expect(await po.hasHotkeyTag()).toBe(false);
    expect(await po.hasHardwareWalletTag()).toBe(true);
```
it will be something like
```
    expect(await po.getTags()).toEqual(['Hotkey']);
```
right?

PR #3085, opened 2023-08-11
Use Agent instead of HttpAgent when possible

dskloetd:
# Motivation

`HttpAgent` is an implementation of `Agent`.
I plan to add another implementation which wraps another agent.
So when we don't care what the implementation is, we should refer to the type `Agent` instead of `HttpAgent`.

# Changes

When referring to the type `HttpAgent`, refer to type `Agent` instead unless we're actually constructing an instance.

# Tests

Still pass

# Todos

not worth an entry


lmuntaner (approved):
LGTM!

PR #3088, opened 2023-08-12
Clean up sns.services.spec.ts

dskloetd:
# Motivation

While working on another PR I noticed that some tests in this file depend on other tests in the same file.

# Changes

1. Move all general cleanup to a top level `beforeEach`.
2. Remove all `afterEach`.
3. Remove an unused `.setSystemTime(now);` since no test cares about the value of `now`.
4. Add an `expect` to validate the cause of a later corresponding `expect`.

# Tests

Ran `npm run test -- src/tests/lib/services/sns.services.spec.ts --randomize` 20 times.

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #3090, opened 2023-08-14
Clean up some api tests

dskloetd:
# Motivation

These tests fail if you change the order in which tests are run:

`frontend/src/tests/lib/api-services/governance.api-service.spec.ts` because some tests set mock behavior to make a mock fail while other tests don't set mock behavior and expect the mock calls not to fail.

`frontend/src/tests/lib/api/canisters.api.spec.ts` because one group of tests did its cleanup in `beforeEach` and the other in `afterEach`.

# Changes

1. `resetAllMocks` in beforeEach.
2. Remove `afterEach`.
3. Add some expectations about the state before the test

# Tests

Tests only

# Todos

Added an entry which should cover similar PRs as well.


lmuntaner (approved):
Thanks!

PR #3092, opened 2023-08-14
Reduce dependency on jest-spy.ts from service tests

dskloetd:
# Motivation

Modules that are imported before a test starts, can't be mocked from the test because the framework has to set up the mocking before the import happens.
This means anything imported from `jest-spy.ts` or `jest-setup.ts` can't be mocked.
I want to add tests for `frontend/src/lib/api/agent.api.ts` but I can't because `jest-spy` imports its dependencies.
My goal is to remove `jest-spy.ts` completely and make tests more explicit about what they mock.

In this PR I temporarily removed the mocking of `getAuthenticatedIdentity` from `jest-spy.ts` and made all tests under `frontend/src/tests/lib/services/app.services.spec.ts` pass.

The real `getAuthenticatedIdentity` uses more `awaits` than the mocked one so in some cases we need extra `awaits` in the test before making expectations.

Some tests use `setNoIdentity` which causes the mocked `getAuthenticatedIdentity` to throw an exception, which the real `getAuthenticatedIdentity` doesn't so to keep those tests working, `getAuthenticatedIdentity` must remain mocked.

# Changes

1. Change `frontend/src/tests/mocks/auth.store.mock.ts` to use `authStore` as its underlying data.
2. Call `resetIdentity` in tests that need the identity to make sure the `authStore` is populated and the real `getAuthenticatedIdentity` (or `mockGetIdentity`) can get the identity from it.
3. Call `resetIdentity()` in `jest-spy.ts` to make sure that the `authStore` is populated and the mocked `getAuthenticatedIdentity` still has the same behavior for tests that aren't fixed yet.
4. In tests that do `setNoIdentity()`, mock `getAuthenticatedIdentity` explicitly.
5. Remove an occasional interfering `mockAuthStoreSubscribe`.
6. Add `runResolvedPromises` where necessary to work with the real `getAuthenticatedIdentity`.
7. Fix a few non-service tests that used `authStore` (without `getAuthenticatedIdentity`) and relied on it starting out as empty without explicitly emptying it.

# Tests

Still pass. With and without mocking `getAuthenticatedIdentity` in `jest-spy.ts`.

# Todos

I will add an entry when `jest-spy.ts` is fully removed.

### DIFF
--- a/frontend/src/tests/lib/components/header/AccountMenu.spec.ts
+++ b/frontend/src/tests/lib/components/header/AccountMenu.spec.ts
@@ -3,8 +3,7 @@
  */
 
 import AccountMenu from "$lib/components/header/AccountMenu.svelte";
-import { authStore } from "$lib/stores/auth.store";
-import { mockAuthStoreSubscribe } from "$tests/mocks/auth.store.mock";
+import { resetIdentity, setNoIdentity } from "$tests/mocks/auth.store.mock";
 import { fireEvent, render, waitFor } from "@testing-library/svelte";
 
 describe("AccountMenu", () => {
@@ -19,17 +18,16 @@ describe("AccountMenu", () => {
   });
 
   it("should display a sign-in button if not signed in", () => {
+    setNoIdentity();
     const { getByTestId } = render(AccountMenu);
 
     expect(getByTestId("toolbar-login")).not.toBeNull();
   });
 
   describe("signed in", () => {
-    beforeAll(() =>
-      jest
-        .spyOn(authStore, "subscribe")
-        .mockImplementation(mockAuthStoreSubscribe)
-    );
+    beforeEach(() => {
+      resetIdentity();
+    });
 
     it("should be open", async () => {
       const renderResult = render(AccountMenu);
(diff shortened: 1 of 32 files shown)


frontend/src/tests/lib/components/header/AccountMenu.spec.ts
```diff
@@ -19,17 +18,16 @@ describe("AccountMenu", () => {
   });
 
   it("should display a sign-in button if not signed in", () => {
+    setNoIdentity();
```
lmuntaner, line 21: Could you use `authStore.setForTesting`?

lmuntaner: Ignore

lmuntaner (approved):
LGTM thanks!

If we wanted to about mocking the getAuthenticatedIdentity, we would need to mock `AuthClient.create` or the authStore.logout method.

Did you think about it or not yet?

dskloetd:
> If we wanted to about mocking the getAuthenticatedIdentity, we would need to mock AuthClient.create or the authStore.logout method.
>
> Did you think about it or not yet?

I thought about it, but that results in different behavior so we'd also have to change every affect test to expect the different behavior which is a lot more work.

PR #3094, opened 2023-08-14
Fix: Nns Followee component test

lmuntaner:
# Motivation

Fix src/tests/lib/components/neuron-detail/NeuronFollowingCard/Followee.spec.ts test.

The problem was that the knownNeuronsStore was filled in one test, but never reset. Yet, most of the tests relied on that neuron to NOT be a known neuron.

The solution is to reset the knownNeuronsStore before each test.

# Changes

* New method in knownsNeuronsStore "reset".
* Reset the knownNeuronsStore beforeEach for Followee.spec.
* Remove unnecessary `waitFor` in the last test case of Followee.spec.

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/lib/stores/known-neurons.store.ts
+++ b/frontend/src/lib/stores/known-neurons.store.ts
@@ -18,7 +18,6 @@ const initKnownNeuronsStore = () => {
       set([...neurons]);
     },
 
-    // Used for testing
     reset() {
       set([]);
     },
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/stores/known-neurons.store.ts
```diff
@@ -17,6 +17,11 @@ const initKnownNeuronsStore = () => {
     setNeurons(neurons: KnownNeuron[]) {
       set([...neurons]);
     },
+
+    // Used for testing
+    reset() {
```
dskloetd, line 22: How about naming it `resetForTesting` if it should only be used in tests?
Or if it's fine to use outside tests, then just remove the comment?

lmuntaner: Comment removed

dskloetd (approved):

PR #3096, opened 2023-08-14
Fix: IncreaseNeuronStakeModal test

lmuntaner:
# Motivation

Fix src/tests/lib/modals/neurons/IncreaseNeuronStakeModal.spec.ts

The problem was that the test expected a store to be empty, but it didn't empty it before running it. When the order changed and the store was filled with data, the test failed.

# Changes

* Use beforeEach instead of beforeAll (not really needed, but we agreed this was the best practice).
* Reset icp accounts store before executing the test that expects the store to be empty.

# Tests

Only test changes

# Todos


lmuntaner:
@dskloetd please review

dskloetd (approved):
Thanks!

PR #3099, opened 2023-08-14
Stop publishing .wasm except with .gz

bitdivine:
# Motivation
Zipped `.wasm` extensions are now deprecated by `dfx`; such files must use the `.wasm.gz` suffix.  We switched to the `.wasm.gz` suffix in 394e50672ce8791934e85582b4cf07d7452474e9.  We continued supplying `.wasm` files alongside `.wasm.gz` files, to give projects that use our builds time to upgrade.  We would like to drop support for the deprecated extension now.

# Changes
* Drop `.wasm` files from releases; retain only `.wasm.gz`.
* Fix bug in the release text.
  * Note: In markdown, backticks are used to format code `like this.`  Backticks in markdown inside a shell script were not escaped, so treated as text to execute.  `shfmt` then joined into the action by changing the backticks to `$(the text to format)`.  This yielded an error message in the action:
    ```
    ./.github/actions/release_nns_dapp/template: line 24: nns-dapp: command not found
    ./.github/actions/release_nns_dapp/template: line 24: sns_aggregator: command not found
    ```
     and an incorrect release text: "This is a release of the  and  canisters."  This is now fixed.

# Tests
* See this release: https://github.com/dfinity/nns-dapp/releases/tag/untagged-56569216043fd00e1627
  - [x] Text is correct
  - [x] Wasms only as `.wasm.gz`
* We have probably found all projects that would be affected by this and notified them.  If we missed you, apologies.  Please get into contact and we will try to find a way of reaching you for future breaking changes.

# Todos


lmuntaner (approved):
Thanks!

PR #3100, opened 2023-08-14
Refactor: Use new setSnsProjects

lmuntaner:
# Motivation

In this PR: Use the new test util setSnsProjects in more tests.

# Changes

* Move Projects.spec to use new test util.
* Move sns-projects.derived to use new test util.

# Tests

Only test changes

# Todos

### DIFF
--- a/frontend/src/tests/lib/components/launchpad/Projects.spec.ts
+++ b/frontend/src/tests/lib/components/launchpad/Projects.spec.ts
@@ -3,18 +3,18 @@
  */
 
 import Projects from "$lib/components/launchpad/Projects.svelte";
-import { snsQueryStore, snsSwapCommitmentsStore } from "$lib/stores/sns.store";
+import { snsSwapCommitmentsStore } from "$lib/stores/sns.store";
 import {
   mockSnsSummaryList,
   mockSnsSwapCommitment,
 } from "$tests/mocks/sns-projects.mock";
-import { snsResponsesForLifecycle } from "$tests/mocks/sns-response.mock";
+import { resetSnsProjects, setSnsProjects } from "$tests/utils/sns.test-utils";
 import { SnsSwapLifecycle } from "@dfinity/sns";
 import { render, waitFor } from "@testing-library/svelte";
 
 describe("Projects", () => {
   beforeEach(() => {
-    snsQueryStore.reset();
+    resetSnsProjects();
     snsSwapCommitmentsStore.reset();
   });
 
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/components/launchpad/Projects.spec.ts
```diff
@@ -119,7 +107,11 @@ describe("Projects", () => {
   it("should render a message when no open projects available", () => {
     const principal = mockSnsSummaryList[0].rootCanisterId;
 
-    snsQueryStore.setData([[], []]);
+    setSnsProjects([
+      {
+        lifecycle: SnsSwapLifecycle.Committed,
```
dskloetd, line 112: Does [[], []] (which we had before) really correspond to one project with `SnsSwapLifecycle.Committed` or what's going on here?

frontend/src/tests/lib/components/launchpad/Projects.spec.ts
```diff
@@ -138,7 +130,11 @@ describe("Projects", () => {
   it("should render a message when no adopted projects available", () => {
     const principal = mockSnsSummaryList[0].rootCanisterId;
 
-    snsQueryStore.setData([[], []]);
+    setSnsProjects([
+      {
+        lifecycle: SnsSwapLifecycle.Open,
```
dskloetd, line 135: Above `[[], []]` was replaced with `SnsSwapLifecycle.Committed` and here with `SnsSwapLifecycle.Open`, is that correct?

frontend/src/tests/lib/derived/sns/sns-projects.derived.spec.ts
```diff
@@ -4,106 +4,81 @@ import {
...
     });
   });
   describe("filter projects derived", () => {
-    beforeAll(() => {
-      snsQueryStore.reset();
-    });
-
-    afterAll(() => {
-      snsQueryStore.reset();
+    beforeEach(() => {
```
dskloetd, line 37: How about putting this `beforeEach` at the top level?

frontend/src/tests/lib/components/launchpad/Projects.spec.ts
lmuntaner: No, that was setting the store to have no projects at all.

Yet, I preferred to set the store to a project that was not of the expected lifecycle. Otherwise, we wouldn't be testing the filtering properly.

frontend/src/tests/lib/components/launchpad/Projects.spec.ts
lmuntaner: No, that was setting the store to have no projects at all.

Yet, I preferred to set the store to a project that was not of the expected lifecycle. Otherwise, we wouldn't be testing the filtering properly.

frontend/src/tests/lib/derived/sns/sns-projects.derived.spec.ts
lmuntaner: Yes, it makes sense.

lmuntaner:

dskloetd (approved):

PR #3104, opened 2023-08-14
Refactor: New SnsDerivedStateStore

lmuntaner:
# Motivation

In this PR: Introduce a new Svelte store that will store the data from the derived_state call.

# Changes

* New sns derived state store creator with caching.
* Load the store when the derived state is fetched directly in "loadSnsTotalCommitment".

# Tests

* Test the new store
* Test that "loadSnsTotalCommitment" also loads the new store.

# Todos

### DIFF
--- a/frontend/src/lib/services/sns.services.ts
+++ b/frontend/src/lib/services/sns.services.ts
@@ -6,6 +6,7 @@ import {
 } from "$lib/api/sns.api";
 import { FORCE_CALL_STRATEGY } from "$lib/constants/mockable.constants";
 import { WATCH_SALE_STATE_EVERY_MILLISECONDS } from "$lib/constants/sns.constants";
+import { getOrCreateDerivedStateStore } from "$lib/stores/sns-derived-state.store";
 import {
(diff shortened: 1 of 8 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/services/sns.services.ts
```diff
@@ -160,9 +161,16 @@ export const loadSnsTotalCommitment = async ({
         identity,
         certified,
       }),
-    onLoad: ({ response: derivedState }) => {
+    onLoad: ({ response: derivedState, certified }) => {
```
dskloetd, line 164: The function is called `loadSnsTotalCommitment`.
Is total commitment the only thing in the derived state?
Would it make sense to rename?

(Sorry, GitHub won't let me comment on the line of the function declaration.)

dskloetd: That would be good. Thanks.

dskloetd (approved):

PR #3108, opened 2023-08-15
Fix: ICP Staked depends on token

lmuntaner:
# Motivation

There is a bug because the StakeItemAction always shows the description "ICP Stake".

Instead, it should show "ICP staked" or "CHAT staked" as in the Figma.

# Changes

* Change "ic_stake" message with a placeholder.
* Use the token for the message in StakeItemAction.svelte

# Tests

* Add test case in StakeItemAction.svelte
* Add test case in NnsStakeItemAction.svelte
* Add test case in SnsStakeItemAction.svelte
* Add method in StakeItemAction PO: getDescription.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -26,6 +26,8 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 
 #### Fixed
 
+* Fix wrong "ICP Staked" message in SNS neurons.
+
 #### Security
 
 #### Not Published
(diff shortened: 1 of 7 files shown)


lmuntaner:
New UI:

Screenshot 2023-08-15 at 07 35 56
Screenshot 2023-08-15 at 07 36 10

lmuntaner:

peterpeterparker:
PR looks good but, since we are already there, does it makes sense to repeat the token - including for ICP - in the message @lmuntaner ?

19.20 LEMEC
LEMEC staked

vs

19.20 LEMEC
Staked

or

19.20 LEMEC
Something short

peterpeterparker (approved):
LGTM, thanks

dskloetd (approved):
Thanks!

lmuntaner:
> PR looks good but, since we are already there, does it makes sense to repeat the token - including for ICP - in the message @lmuntaner ?

I thought about it also. I don't have a strong opinion.

I didn't feel the repetition was so bad because they have different colors. And that makes them be read separately.

If I used only "Staked", it's a broken sentence, not clear where it belongs to. "Tokens staked" doesn't seem good either.

So, from the different options. I went for the one on Figma despite the repetition.

PR #3109, opened 2023-08-15
Move simplifyJson out of Json.spec

dskloetd:
# Motivation

`NnsProposalProposerPayloadEntry.spec.ts` and `ProposalProposerPayloadEntry.spec.ts` were importing `Json.spec.ts` to get access to `simplifyJson` but this meant that when running any of these tests, it would run the tests from `Json.spec.ts` as well.
Tests shouldn't be importing other tests.

# Changes

1. Move `simplifyJson` into a separate file: `frontend/src/tests/utils/json.test-utils.ts`.
2. Import `frontend/src/tests/utils/json.test-utils.ts` instead of `Json.spec.ts`.

# Tests

only tests

# Todos

not worth it


lmuntaner (approved):
Thanks!

PR #3112, opened 2023-08-15
Fix: HW cannot participate in swap

lmuntaner:
# Motivation

We still don't allow HW to participate in the swap. Yet, there was a bug where the HW account was selectable in the modal to participate.

In case that the user selected the HW, they wouldn't participate with the HW. Instead, they would participate with the main account. Therefore, what the user selected, was ignored.

# Changes

* New variable "filterSourceAccounts" in TransactionForm using the prop "skipHardwareWallets".
* Pass the "filterSourceAccounts" to "TransactionFromAccount".
* Remove using the prop "skipHardwareWallets" when filtering the destination accounts. This was an error never seen because the TransactionForm had never "skipHardwareWallets" with selectable destination account.

# Tests

* Add methods to related POs to expose the available source accounts.
* Add a test case in ParticipateSwapModal to check that HW accounts are not available.

# Todos

### DIFF
--- a/frontend/src/lib/components/transaction/TransactionForm.svelte
+++ b/frontend/src/lib/components/transaction/TransactionForm.svelte
@@ -51,7 +51,10 @@
 
   let filterSourceAccounts: (account: Account) => boolean;
   $: filterSourceAccounts = (account: Account) => {
-    return !skipHardwareWallets || !isAccountHardwareWallet(account);
+    if (skipHardwareWallets) {
+      return !isAccountHardwareWallet(account);
+    }
+    return true;
   };
 
   let filterDestinationAccounts: (account: Account) => boolean;
--- a/frontend/src/lib/components/transaction/TransactionForm.svelte
+++ b/frontend/src/lib/components/transaction/TransactionForm.svelte
@@ -49,12 +49,14 @@
 
   export let validateAmount: ValidateAmountFn = () => undefined;
 
+  let filterSourceAccounts: (account: Account) => boolean;
+  $: filterSourceAccounts = (account: Account) => {
+    return !skipHardwareWallets || !isAccountHardwareWallet(account);
+  };
+
   let filterDestinationAccounts: (account: Account) => boolean;
   $: filterDestinationAccounts = (account: Account) => {
-    return (
-      account.identifier !== selectedAccount?.identifier ||
-      (skipHardwareWallets && isAccountHardwareWallet(account))
-    );
+    return account.identifier !== selectedAccount?.identifier;
   };
 
   let max = 0;
@@ -126,6 +128,7 @@
     {canSelectSource}
     {rootCanisterId}
     {token}
+    filterAccounts={filterSourceAccounts}
   />
 
   {#if canSelectDestination}
--- a/frontend/src/lib/components/transaction/TransactionForm.svelte
+++ b/frontend/src/lib/components/transaction/TransactionForm.svelte
@@ -51,10 +51,7 @@
 
   let filterSourceAccounts: (account: Account) => boolean;
   $: filterSourceAccounts = (account: Account) => {
-    if (skipHardwareWallets) {
-      return !isAccountHardwareWallet(account);
-    }
-    return true;
+    return !skipHardwareWallets || !isAccountHardwareWallet(account);
   };
 
   let filterDestinationAccounts: (account: Account) => boolean;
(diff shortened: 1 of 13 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/page-objects/Dropdown.page-object.ts
```diff
@@ -11,4 +11,9 @@ export class DropdownPo extends BasePageObject {
   select(option: string): Promise<void> {
     return this.root.querySelector("select").selectOption(option);
   }
+
+  async getOptions(): Promise<string[]> {
+    const options = await this.root.querySelectorAll("option");
+    return Promise.all(options.map((option) => option.getValue()));
```
dskloetd, line 17: Do you actually want to get the `value` props from the option elements, rather than the text? I would expect `getOptions()` to give me the text of the options.

frontend/src/lib/components/transaction/TransactionForm.svelte
dskloetd, line 1: I won't block the PR on it but I do think that `TransactionForm.svelte` needs its own test where we test (among other things) that the `skipHardwareWallets` prop does what it should.

frontend/src/tests/page-objects/Dropdown.page-object.ts
lmuntaner: Mhm, these are the identifiers. That's why I thought it was better than the texts.

But no strong opinions. I can change it.

frontend/src/lib/components/transaction/TransactionForm.svelte
lmuntaner: I agree, I can add a test for it in another PR

lmuntaner: Actually, that's tested in the TransactionModal. I'll add a test there.

lmuntaner: done

lmuntaner:

frontend/src/tests/lib/modals/transaction/TransactionModal.spec.ts
```diff
@@ -558,4 +564,28 @@ describe("TransactionModal", () => {
...
+    beforeEach(() => {
+      icpAccountsStore.setForTesting({
+        main: mockMainAccount,
+        subAccounts: [mockSubAccount],
+        hardwareWallets: [mockHardwareWalletAccount],
+        certified: true,
+      });
+    });
+
+    it("should not show the hardware wallet account", async () => {
```
dskloetd, line 578: I was hoping to have 2 tests, both with the same setup and only the `skipHardwareWallets` prop different.
That way we know the HW is not there because of the flag, rather than because of a wrong setup.

dskloetd (approved):

frontend/src/tests/lib/modals/transaction/TransactionModal.spec.ts
lmuntaner: Good point, let me add it

lmuntaner: Thanks to your test I discovered that I wasn't testing the prop properly because it was mocking the accounts store. I removed the mock and set the store instead.

frontend/src/lib/components/transaction/TransactionForm.svelte
```diff
@@ -51,7 +51,10 @@
 
   let filterSourceAccounts: (account: Account) => boolean;
   $: filterSourceAccounts = (account: Account) => {
-    return !skipHardwareWallets || !isAccountHardwareWallet(account);
+    if (skipHardwareWallets) {
+      return !isAccountHardwareWallet(account);
+    }
+    return true;
```
dskloetd, line 57: Did you make this change to make debugging easier? I thought it was easier to read before.

frontend/src/tests/lib/modals/transaction/TransactionModal.spec.ts
dskloetd: That's great! This is what good testing is about.

dskloetd (approved):

frontend/src/lib/components/transaction/TransactionForm.svelte
lmuntaner: Really? I changed it because I thought it was easier to read like this.

lmuntaner: reverted

PR #3113, opened 2023-08-15
Fix: Randomized test NnsStakeNeuronModal

lmuntaner:
# Motivation

NnsStakeNeuronModal.spec file was failing when randomizing unit tests.

The problem: if the case "when accounts are not loaded" was run first, it was getting the authenticated identity and failing.

# Changes

* "resetIdentity" in the beforeEach of the test.
* Remove unnecessary clearing mocks and resetting store.

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/tests/lib/modals/neurons/NnsStakeNeuronModal.spec.ts
+++ b/frontend/src/tests/lib/modals/neurons/NnsStakeNeuronModal.spec.ts
@@ -13,15 +13,10 @@ import {
   stakeNeuron,
   updateDelay,
 } from "$lib/services/neurons.services";
-import { authStore } from "$lib/stores/auth.store";
 import { icpAccountsStore } from "$lib/stores/icp-accounts.store";
 import { neuronsStore } from "$lib/stores/neurons.store";
 import { formatVotingPower } from "$lib/utils/neuron.utils";
-import {
-  mockAuthStoreSubscribe,
-  mockIdentity,
-  resetIdentity,
-} from "$tests/mocks/auth.store.mock";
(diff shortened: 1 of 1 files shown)


frontend/src/tests/lib/modals/neurons/NnsStakeNeuronModal.spec.ts
```diff
@@ -20,6 +20,7 @@ import { formatVotingPower } from "$lib/utils/neuron.utils";
 import {
   mockAuthStoreSubscribe,
```
dskloetd, line 21: Since we're setting the real `authStore`, I think we can and should remove
```
      jest
```
everywhere.
GitHub won't let me comment on those lines :(

lmuntaner: Shall I use this PR to do so?

Remember that I try to keep PRs to do one thing 😉

dskloetd: You can't remove `mockAuthStoreSubscribe` without adding `resetIdentity` so it's reasonable to say it's the same thing.
Mixing both is pretty confusing IMO but if you insist on doing it in the next PR, go ahead.

dskloetd (approved):

frontend/src/tests/lib/modals/neurons/NnsStakeNeuronModal.spec.ts
lmuntaner: Ok, good point.

dskloetd (approved):
Thanks!

PR #3114, opened 2023-08-15
Move e2e-tests/scripts to scripts/e2e-tests

dskloetd:
# Motivation

All the old wdio e2e tests have been deleted.
I would like to delete the `e2e-tests` directory but there are still scripts in it that are referenced in other places.
The scripts don't currently seem to work but fixing them is outside the scope of migrating the e2e tests.

# Changes

1. Delete `e2e-tests/scripts/update-chromedriver` because Chromedriver was only used by the wdio tests.
2. Move every other script in `e2e-tests/scripts/` to `scrtips/e2e-tests`.
3. Update paths referencing the scripts.

# Tests

The scripts don't currently seem to work but fixing them is outside the scope of migrating the e2e tests.

# Todos


dfx.json
```diff
@@ -5,7 +5,7 @@
       "type": "custom",
       "candid": "target/ic/governance.did",
       "wasm": "target/ic/governance-canister_test.wasm",
-      "build": "./e2e-tests/scripts/nns-canister-download && ./e2e-tests/scripts/nns-canister-build",
+      "build": "./scripts/e2e-tests/nns-canister-download && ./scripts/e2e-tests/nns-canister-build",
```
bitdivine, line 8: The paths don't match the location set below.  There should be another `/scripts` after `e2e-tests`.  Or change the location of the scripts to the path shown here.

dfx.json
```diff
@@ -17,7 +17,7 @@
       "type": "custom",
       "candid": "target/ic/ic-icrc1-ledger.did",
       "wasm": "target/ic/ic-icrc1-ledger.wasm",
-      "build": "./e2e-tests/scripts/nns-canister-download && ./e2e-tests/scripts/nns-canister-build",
+      "build": "./scripts/e2e-tests/nns-canister-download && ./scripts/e2e-tests/nns-canister-build",
```
bitdivine, line 20: Ditto

scripts/e2e-tests/scripts/nns-canister-install
```diff
@@ -60,7 +60,7 @@ test -x "${DOWNLOAD_DIR}/ic-nns-init" || {
 }
 
 FIRST_POSSIBLE_CANISTER_ID=rwlgt-iiaaa-aaaaa-aaaaa-cai
-if { curl -sL "$(./e2e-tests/scripts/nns-dashboard)" | grep -q "$FIRST_POSSIBLE_CANISTER_ID"; } && [[ "$UNCLEAN" != "yes" ]]; then
+if { curl -sL "$(./scripts/e2e-tests/nns-dashboard)" | grep -q "$FIRST_POSSIBLE_CANISTER_ID"; } && [[ "$UNCLEAN" != "yes" ]]; then
```
bitdivine, line 63: Ditto

scripts/e2e-tests/scripts/nns-canister-install
```diff
@@ -82,7 +82,7 @@ fi
 echo "Canister hashes:"
 sha256sum "${DOWNLOAD_DIR}"/*.wasm.gz
 
-NNS_URL="$(./e2e-tests/scripts/nns-dashboard)"
+NNS_URL="$(./scripts/e2e-tests/nns-dashboard)"
```
bitdivine, line 85: Ditto

scripts/propose
```diff
@@ -127,10 +127,10 @@ while (($# > 0)); do
 done
 
 if test -n "${IC_TESTNET:-}"; then
-  NNS_URL="$(./e2e-tests/scripts/nns-dashboard --testnet "$IC_TESTNET")"
+  NNS_URL="$(./scripts/e2e-tests/nns-dashboard --testnet "$IC_TESTNET")"
 else
   DFX_NETWORK="${DFX_NETWORK:-local}"
-  NNS_URL="$(./e2e-tests/scripts/nns-dashboard --dfx-network "${DFX_NETWORK}")"
+  NNS_URL="$(./scripts/e2e-tests/nns-dashboard --dfx-network "${DFX_NETWORK}")"
```
bitdivine, line 133: Ditto

dfx.json
dskloetd: Thanks for catching.
I intended the scripts to be in `/scripts/e2e-tests`.
The command I ran was `git mv e2e-tests/scripts scripts/e2e-tests` which I would have expected to do the right thing but apparently `git mv` does not have the same semantics as `mv`.

It should be correct now.

bitdivine (approved):

PR #3115, opened 2023-08-15
Fix: Remove unnecessary set

lmuntaner:
# Motivation

Remove unnecessary code in sns swan canisters accounts store.

# Changes

Changes in sns-swap-canisters.accounts.derived.ts
* Remove extra `Set` around the array.
* Use `[]` instead of undefined when we want an empty set.

# Tests

Not necessary.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/derived/sns-swap-canisters-accounts.derived.ts
+++ b/frontend/src/lib/derived/sns-swap-canisters-accounts.derived.ts
@@ -18,16 +18,14 @@ export const createSwapCanisterAccountsStore = (controller?: Principal) =>
     ($snsAggregatorStore) =>
       new Set(
         isNullish(controller) || isNullish($snsAggregatorStore.data)
-          ? undefined
-          : new Set(
-              $snsAggregatorStore.data.map(({ canister_ids }) =>
-                getSwapCanisterAccount({
-                  controller,
-                  swapCanisterId: Principal.fromText(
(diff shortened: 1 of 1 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/derived/sns-swap-canisters-accounts.derived.ts
```diff
@@ -18,16 +18,14 @@ export const createSwapCanisterAccountsStore = (controller?: Principal) =>
...
-          : new Set(
-              $snsAggregatorStore.data.map(({ canister_ids }) =>
-                getSwapCanisterAccount({
-                  controller,
-                  swapCanisterId: Principal.fromText(
-                    canister_ids.swap_canister_id
-                  ),
-                }).toHex()
-              )
+          ? []
```
dskloetd, line 21: I was a bit surprised by `undefined` but I tested it and it actually works. But `[]` is fine too so this is just FYI.

dskloetd (approved):
Thanks!

frontend/src/lib/derived/sns-swap-canisters-accounts.derived.ts
lmuntaner: Yes, I was also surprised that it worked. But I believe a `[]` is more readable.

PR #3119, opened 2023-08-15
Fix randomized test NnsProposalProposerPayloadEntry.spec.ts

dskloetd:
# Motivation

1. `proposalPayloadsStore.reset()` wasn't called in `beforeEach` because `beforeEach` was only passed an arrow function which returns `proposalPayloadsStore.reset` but doesn't call it.
2. This means that the second function to run wouldn't load the proposal payload but would have it cached in the store.
3. So when the component is rendered it would be correct immediately without having to wait.
4. The `waitFor()` in the second function wasn't actually working because `queryByTestId("json-wrapper")` was called outside the `waitFor()` and once we got the incomplete element, it would never pass the `expect`.
5. But this didn't matter because the store was pre-populated by the first test and thus the component rendered fast enough.

# Changes

1. Actually run `proposalPayloadsStore.reset()` in `beforeEach`.
2. Use `await runResolvedPromises();` instead of `waitFor`.
3. Drive by: Move `jest.clearAllMocks();` from `afterAll` to `beforeEach`.
4. Drive by: Add an expectation that the spy is not called before we expect it to be called.

# Tests

The tests now pass in both orders.

# Todos

covered by existing entry

### DIFF
--- a/frontend/src/tests/lib/components/proposal-detail/NnsProposalProposerPayloadEntry.spec.ts
+++ b/frontend/src/tests/lib/components/proposal-detail/NnsProposalProposerPayloadEntry.spec.ts
@@ -10,8 +10,9 @@ import {
   proposalActionNnsFunction21,
 } from "$tests/mocks/proposal.mock";
 import { simplifyJson } from "$tests/utils/json.test-utils";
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!

PR #3120, opened 2023-08-15
Fix: Randomized test sns neuron services

lmuntaner:
# Motivation

Test sns-neuron-services.spec fails when randomizing unit tests order.

The problem was that one test was mocking a service used in another test. Preventing from reaching the call to the api.

Solution: remove the mock and test that api is called instead.

# Changes

* In sns-neuron services remove mock to "loadNeurons". Instead, check that "querySnsNeurons" is called.

# Tests

Only test changes.

# Todos

Not worth an entry in the changelog.


lmuntaner:
@dskloetd please review

dskloetd (approved):

PR #3122, opened 2023-08-15
Refactor: Use new sns test util

lmuntaner:
# Motivation

In this PR: Use new test util in sns-selected-project.derived.spec and sns.services.spec

# Changes

* Change from snsQueryStore.setData to setSnsProjects in both test files.

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/tests/mocks/sns-projects.mock.ts
+++ b/frontend/src/tests/mocks/sns-projects.mock.ts
@@ -88,10 +88,6 @@ export const mockSnsSwapCommitment = (
       rootCanisterId: principal(3),
       myCommitment: undefined,
     },
-    [rootCanisterId.toText()]: {
-      rootCanisterId,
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/mocks/sns-projects.mock.ts
```diff
@@ -88,6 +88,10 @@ export const mockSnsSwapCommitment = (
       rootCanisterId: principal(3),
       myCommitment: undefined,
     },
+    [rootCanisterId.toText()]: {
```
lmuntaner, line 91: This was not strictly necessary. But it was not clear that not any root canister id would return a swap commitment mock.

I used this in a few places, so it made sense to add a fallback the root canister id.

dskloetd (approved):

PR #3124, opened 2023-08-15
Fix 'should retrieve btc' in ckbtc-convert.services.spec.ts

dskloetd:
# Motivation

The test sets mock behavior in a `describe` block, outside `beforeEach` or `it`.
This gets executed in a surprising order, because first the framework collects all the tests that should be executed by running the describe blocks, and then it runs all the tests.
So the behavior set by
```
const retrieveBtcSpy = minterCanisterMock.retrieveBtc.mockResolvedValue(ok);
```
was replaced by
```
minterCanisterMock.retrieveBtc.mockImplementation(async () => {
  throw new Error();
});
```
inside another test which ran after the `describe` block ran but before the `"should retrieve btc"` test ran.
This meant that in the `"should retrieve btc"`, btc were not actually retrieved.

The test checked that `updateProgressSpy` was called 3 times, and assumed (with comment) this was for `SEND_BTC + RELOAD + DONE`, but actually it was for `INITIALIZATION + SEND + RELOAD`.

When the mock setup is fixed, `updateProgressSpy` is actually called 4 times, with `INITIALIZATION + SEND_BTC + RELOAD + DONE`.

There are more broken tests in this file but it's tricky so I'm doing one step at a time.

# Changes

1. Define mock behavior inside `beforeEach`.

# Tests

Previously the test would fail when run in isolation but now it passes in isolation.

# Todos

covered by an existing entry

### DIFF
--- a/frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
+++ b/frontend/src/tests/lib/services/ckbtc-convert.services.spec.ts
@@ -14,6 +14,7 @@ import { bitcoinConvertBlockIndexes } from "$lib/stores/bitcoin.store";
 import { ckBTCWithdrawalAccountsStore } from "$lib/stores/ckbtc-withdrawal-accounts.store";
 import * as toastsStore from "$lib/stores/toasts.store";
 import { tokensStore } from "$lib/stores/tokens.store";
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!

PR #3125, opened 2023-08-15
Remove jest-spy.ts

dskloetd:
# Motivation

Modules imported before a test is loaded can't be mocked.
This means that `jest-spy.ts`, which is loaded before a test file is loaded, prevent us from testing certain things.
`jest-spy.ts` currently only mocks `createAgent` from `$lib/api/agent.api`, which can just be done explicitly in tests which need it.

# Changes

1. Remove `frontend/jest-spy.ts`.
2. Don't reference `frontend/jest-spy.ts` from `frontend/jest.config.cjs`.
3. In tests that need `createAgent` mocked, mock it explicitly.
4. In tests that call `createAgent` to get a mock agent, just create a mock agent directly.
5. Consistently use `import { mock } from "jest-mock-extended";` instead of sometimes `import mock from "jest-mock-extended/lib/Mock";`

# Tests

All tests still pass.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -51,6 +51,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Removed
 
 * Remove compressed `.wasm` files from releases.  Please use `.wasm.gz` instead.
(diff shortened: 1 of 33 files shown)


lmuntaner (approved):
Thanks!

PR #3127, opened 2023-08-16
GIX-1821: Use "Open" filter for lauchpad proposals

lmuntaner:
# Motivation

The launchpad renders the open proposals related to SNSes. We were fetching all the proposals of the topic and then filtering only the Open ones. We had a problem because there were big SNS proposals and the message limit was reached.

The way to solve it for the launchpad is to request only the OPEN and SNS topic.

# Changes

* Remove service "loadProposalsByTopic".
* Use queryProposals api in "loadProposalsSnsCF" with the topic and status filter.

# Tests

* Change the test of the Proposals.spec to spy on the api function and test that it's called with the proper values.

# Todos


lmuntaner:
@mstrasinskis please review

mstrasinskis (approved):
LGTM, thank you!

PR #3128, opened 2023-08-16
GIX-1822: Fail gracefully on list proposal payload size error

lmuntaner:
# Motivation

We had a problem because there were big SNS proposals and the message limit was reached. The current UX was not the best and didn't explain what the user could do.

In this PR: We catch that specific message and change it for a message that the user can understand and act upon.

# Changes

* New error util "isPayloadSizeError".
* Use new util in "handleFindProposalsError" which is used in "listProposals" service.

# Tests

* Test new error util.
* Add test cases in "listProposals" service for when api fails.

# Todos

### DIFF
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -45,6 +45,7 @@
     "sns_accounts_balance_load": "An error occurred while loading the balance of the accounts.",
     "icrc_token_load": "An error occurred while loading the token balance or metadata.",
     "list_proposals": "There was an unexpected issue while searching for the proposals.",
+    "list_proposals_payload_too_large": "The current results are too large. Please try another filter with less results.",
     "get_proposal": "There was an unexpected issue while fetching proposal $proposalId.",
     "wrong_proposal_id": "The id \"$proposalId\" is not a valid proposal id.",
     "list_canisters": "There was an unexpected issue while searching for the canisters.",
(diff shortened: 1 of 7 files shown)


lmuntaner:
New UX:

https://github.com/dfinity/nns-dapp/assets/4550653/3176f574-e8fa-4777-a25d-7543eab681e1

New message:

Screenshot 2023-08-16 at 10 48 29

lmuntaner:
@mstrasinskis please review

frontend/src/lib/i18n/en.json
```diff
@@ -45,6 +45,7 @@
     "sns_accounts_balance_load": "An error occurred while loading the balance of the accounts.",
     "icrc_token_load": "An error occurred while loading the token balance or metadata.",
     "list_proposals": "There was an unexpected issue while searching for the proposals.",
+    "list_proposals_payload_too_large": "The current results are too large. Please try another filter with less results.",
```
mstrasinskis, line 48: nitpick:
What do you think about providing a bit more details here, like: The current proposals response is too large.

And maybe use plural for filter: eg. Please adjust proposal filters to get less results.

mstrasinskis (approved):
Thanks, just an optional comment.

frontend/src/lib/i18n/en.json
lmuntaner: Sounds good!

PR #3129, opened 2023-08-16
GIX-1820: Identify Swap Refund Transaction

lmuntaner:
# Motivation

The transactions list in the ICP wallet identifies the swap participation transaction. Therefore, we should also identify the refunds.

# Changes

* Add a transaction type in `AccountTransactionType`.
* Add a check in `transactionType` util to identify swap refunds.

# Tests

* Test new case in `transactionType`.
* Test new case in `mapNnsTransaction`.

# Todos

### DIFF
--- a/frontend/src/lib/utils/transactions.utils.ts
+++ b/frontend/src/lib/utils/transactions.utils.ts
@@ -38,6 +38,13 @@ export const transactionType = ({
     }
   }
 
+  if ("Receive" in transfer) {
+    const { from } = transfer.Receive;
+    if (swapCanisterAccounts.has(from)) {
+      return AccountTransactionType.RefundSwap;
+    }
+  }
+
   if ("Transfer" in transaction_type[0]) {
     return AccountTransactionType.Send;
   } else if ("StakeNeuron" in transaction_type[0]) {
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -465,6 +465,7 @@
     "topUpNeuron": "Top-up Neuron",
     "createCanister": "Create Canister",
     "topUpCanister": "Top-up Canister",
+    "refundSwap": "Swap Refund",
     "participateSwap": "Decentralization Swap"
   },
   "ckbtc_transaction_names": {
--- a/frontend/src/lib/types/i18n.d.ts
+++ b/frontend/src/lib/types/i18n.d.ts
@@ -481,6 +481,7 @@ interface I18nTransaction_names {
   topUpNeuron: string;
   createCanister: string;
   topUpCanister: string;
+  refundSwap: string;
   participateSwap: string;
 }
 
--- a/frontend/src/lib/types/transaction.ts
+++ b/frontend/src/lib/types/transaction.ts
@@ -63,6 +63,7 @@ export enum AccountTransactionType {
   CreateCanister = "createCanister",
   TopUpCanister = "topUpCanister",
   ParticipateSwap = "participateSwap",
+  RefundSwap = "refundSwap",
 }
 
 export interface Transaction {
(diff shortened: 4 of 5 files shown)


lmuntaner:
New UI:

Screenshot 2023-08-16 at 11 25 28

lmuntaner:
@dskloetd please review

frontend/src/lib/utils/transactions.utils.ts
```diff
@@ -38,6 +38,13 @@ export const transactionType = ({
     }
   }
 
+  if ("Receive" in transfer) {
+    const { from } = transfer.Receive;
+    if (swapCanisterAccounts.has(from)) {
+      return AccountTransactionType.RefundSwap;
+    }
+  }
+
```
dskloetd, line 47: More about the previous PR than this one but I just noticed `ParticipateSwap` below on line 65.
Is that not working or is that for a different scenario than line 37?

frontend/src/lib/utils/transactions.utils.ts
```diff
@@ -38,6 +38,13 @@ export const transactionType = ({
     }
   }
 
+  if ("Receive" in transfer) {
```
dskloetd, line 41: What if it's a receive but not a refund?
I don't see that case handled anywhere but I guess it's not resulting in an error?
What am I missing?
Just wondering if this should be grouped with the normal receive.

lmuntaner: Those are handled below. The transactions already come with a type mapped in the backend.

The refund and participation override the backend mapping.

Below could be more approapiate a switch.

frontend/src/lib/utils/transactions.utils.ts
lmuntaner: The NNS Dapp was able to identify Participate Swap for the first SNS that was launched. That's because a mechanism was added to add the participation as a "Pending Participation" and then the canister could refresh the tokens.

This was removed for a flow relying only in the client.

We haven't removed ALL the code, and removing one entry in the enum of the backend means migrating. And we haven't prioritized migrating this enum.

frontend/src/lib/utils/transactions.utils.ts
dskloetd: Yeah, I looked below but I don't see it handled there. That's why I am confused.
I see `Send`, `StakeNeuron`, etc. but no `Receive`.

lmuntaner: Receive is not a transaction type:

```
export interface Transaction {
  transaction_type: [] | [TransactionType];
  memo: bigint;
  timestamp: Timestamp;
  block_height: BlockHeight;
  transfer: Transfer;
}
export type TransactionType =
  | { Burn: null }
  | { Mint: null }
  | { StakeNeuronNotification: null }
  | { TopUpCanister: CanisterId }
  | { ParticipateSwap: CanisterId }
  | { CreateCanister: null }
  | { Transfer: null }
  | { TopUpNeuron: null }
  | { StakeNeuron: null };
export type Transfer =
  | { Burn: { amount: ICPTs } }
  | { Mint: { amount: ICPTs } }
  | { Send: Send }
  | { Receive: Receive };
```

Instead, it's a type "Transfer".

"Receive" is a property of the field "transfer" of type "Transfer".

I understand your confusion...

dskloetd: So if in my transaction list there is a transaction where I received some ICP, on which line of this function will it return the transaction type for that transaction?

dskloetd: So when it's a "receive" transaction it gets transaction type `Send`?
Wow, that's super confusion. Thanks for clarifying.

dskloetd (approved):

frontend/src/lib/utils/transactions.utils.ts
lmuntaner: I know...

Hopefully we'll move the the ICRC Index canister soon and we can remove all this complexity from the backend.

PR #3130, opened 2023-08-16
Test pagination

bitdivine:
# Motivation
We are about to change the pagination; it would be good to be sure that the pagination is as expected.

# Changes
* Add a test that the pages are present and contain the expected SNS indices.
* Update the `nns-sns-wasm` .did file path to use the file we have committed in the repo.
  * Note: This means that the .did file is available and can be used without building, unlike the previous code.
  * Note: The existing path was built when building custom sns canisters for e2e tests, however the old e2e tests with the custom builds may be deleted soon.
  * Note: There are probbaly other .did paths that should be updated, but only this update is needed for the API call made in the test.

# Tests
* This PR contains a test and the test is exercised in CI.
* I have run this test against mainnet and the app subnet aggregator:
  ```
   max@sinkpad:~/dfn/nns-dapp/branches/test-pagination (2:03:06)$ dfx-canister-set-id --canister_name sns_aggregator --canister_id otgyv-wyaaa-aaaak-qcgba-cai --network ic
   max@sinkpad:~/dfn/nns-dapp/branches/test-pagination (2:03:18)$ scripts/sns/aggregator/test-pagination --network ic
   max@sinkpad:~/dfn/nns-dapp/branches/test-pagination (2:03:22)$ scripts/sns/aggregator/test-pagination --network mainnet
   max@sinkpad:~/dfn/nns-dapp/branches/test-pagination (2:03:26)$ 
  ```
  (I have deleted stdout to keep the text short)

# Todos

### DIFF
--- a/scripts/sns/aggregator/test-pagination
+++ b/scripts/sns/aggregator/test-pagination
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+# shellcheck disable=SC2317
+print_help() {
+  cat <<-EOF
+
+	Tests that SNSs can be obtained without error from the aggregator.
+
+	Prerequisites:
+	 * The aggregator has been installed.
+	 * The aggregator has had time to populate its cache, so all data should be present and correct.
+	 * No new SNSs are being added, as this could cause a race condition when checking
+	  the number of SNSs against the "latest" endpoint.
+	EOF
+}
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/../../clap.bash"
+# Define options
+clap.define short=n long=network desc="The dfx network to use" variable=DFX_NETWORK default="local"
+clap.define short=s long=num desc="The number of SNSs" variable=NUM_SNS default=""
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+EXIT=0
+
+AGGREGATOR_CANISTER_URL="$(dfx-canister-url --network "$DFX_NETWORK" sns_aggregator)"
+curl --fail -sSL "$AGGREGATOR_CANISTER_URL" >/dev/null || {
+  echo "ERROR: SNS aggregator not found at $AGGREGATOR_CANISTER_URL"
+  exit 1
+} >&2
+[[ "" != "${NUM_SNS:-}" ]] || NUM_SNS="$(curl --fail -sSL "${AGGREGATOR_CANISTER_URL}/v1/sns/list/latest/slow.json" | jq '.[0].index +1' 2>/dev/null || echo 0)"
+
+(
+  printf "\n\n\n=========================================\n"
+  printf "%s\n" "Test that:" \
+    " * Successive pages can be downloaded until we get a page that is not full." \
+    " * These pages provide the same number of SNS suggested by the 'latest' API."
+  PAGE_SIZE=10
+  PAGE_FILE="$(mktemp aggregator-last-page-XXXXX.json)"
+  for ((page = 0, sns_count = 0; ; page++)); do
+    # Get the page
+    curl -Lf "${AGGREGATOR_CANISTER_URL}/v1/sns/list/page/${page}/slow.json" -o "$PAGE_FILE"
+    # Verify that the expected number of SNSs is prseent.
+    expect=$((sns_count + PAGE_SIZE > NUM_SNS ? NUM_SNS - sns_count : PAGE_SIZE))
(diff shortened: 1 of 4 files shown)


scripts/sns/aggregator/test-pagination
```diff
@@ -0,0 +1,75 @@
...
+}
+sns_from_wasm_canister() {
+  dfx canister call nns-sns-wasm list_deployed_snses --network "$DFX_NETWORK" '(record {})' | idl2json | jq '.instances | length'
+}
+[[ "" != "${NUM_SNS:-}" ]] || NUM_SNS="$(sns_from_wasm_canister || sns_from_latest)"
+
+(
+  printf "\n\n\n=========================================\n"
+  printf "%s\n" "Test that:" \
+    " * Successive pages can be downloaded until we get a page that is not full." \
```
bitdivine, line 45: Note: This line above is not true until the code has been updated; I will delete it from this PR and add it when the code is updated.

scripts/sns/aggregator/test-pagination
```diff
@@ -0,0 +1,75 @@
...
+    " * These pages provide the same number of SNS suggested by the 'latest' API."
+  PAGE_SIZE=10
+  for ((page = 0, sns_count = 0; ; page++)); do
+    # Get the page
+    PAGE_FILE="$(mktemp aggregator-last-page-XXXXX.json)"
+    curl -Lf "${AGGREGATOR_CANISTER_URL}/v1/sns/list/page/${page}/slow.json" -o "$PAGE_FILE"
+    # Verify that the expected number of SNSs is prseent.
```
lmuntaner, line 52: typo "prseent"

scripts/sns/aggregator/test-pagination
```diff
@@ -0,0 +1,75 @@
...
+} >&2
+
+# Get the number of SNSs
+sns_from_latest() {
+  curl --fail -sSL "${AGGREGATOR_CANISTER_URL}/v1/sns/list/latest/slow.json" | jq '.[0].index +1' 2>/dev/null || echo 0
+}
+sns_from_wasm_canister() {
+  dfx canister call nns-sns-wasm list_deployed_snses --network "$DFX_NETWORK" '(record {})' | idl2json | jq '.instances | length'
+}
+[[ "" != "${NUM_SNS:-}" ]] || NUM_SNS="$(sns_from_wasm_canister || sns_from_latest)"
```
lmuntaner, line 40: Do I understand correctly, that here you are setting the variable `NUM_SNS` that has the total amount of SNS?

scripts/sns/aggregator/test-pagination
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+
+# shellcheck disable=SC2317
+print_help() {
+  cat <<-EOF
+
+	Tests that SNSs can be obtained without error from the aggregator.
```
lmuntaner, line 9: Ideally we can test this with different number of SNSes.

For example, what happens if there are 5, or 10, or 20, or 23.

It's not just that we can get the number of SNSes the test aggregator has, right?

bitdivine: For that we would install 5 SNS, run this, then 10, run this and so on.

scripts/sns/aggregator/test-pagination
```diff
@@ -0,0 +1,75 @@
...
+
+# Get and check the canister URL.
+AGGREGATOR_CANISTER_URL="$(dfx-canister-url --network "$DFX_NETWORK" sns_aggregator)"
+curl --fail -sSL "$AGGREGATOR_CANISTER_URL" >/dev/null || {
+  echo "ERROR: SNS aggregator not found at $AGGREGATOR_CANISTER_URL"
+  exit 1
+} >&2
+
+# Get the number of SNSs
+sns_from_latest() {
```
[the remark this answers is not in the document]
bitdivine: Yes, but one can learn the number of SNSs by looking at the index of the latest SNS and adding 1.

scripts/sns/aggregator/test-pagination
bitdivine: Correct.  Based on that number, we know how which pages should be available and how many SNSs should be in each.

scripts/sns/aggregator/test-pagination
bitdivine: Thank you.  Fixed.

scripts/sns/aggregator/test-pagination
lmuntaner: Sounds good! Thanks for the explanation!

lmuntaner (approved):
Thanks!

PR #3132, opened 2023-08-16
GIX-1823: New sns lifecycle store

lmuntaner:
# Motivation

In this PR: Introduce a new Svelte store that will store the data from the get_lifecycle call.

This store is needed to store the latest lifecycle after the user's participation.

# Changes

* New sns lifecycle store creator with caching.
* Load the store when the derived state is fetched directly in "loadSnsLifecycle".

# Tests

* Test the new store
* Test that "loadSnsTotalCommitment" also loads the new store.

# Todos

### DIFF
--- a/frontend/src/lib/stores/sns-lifecycle.store.ts
+++ b/frontend/src/lib/stores/sns-lifecycle.store.ts
@@ -1,0 +1,56 @@
+import type { Principal } from "@dfinity/principal";
+import type { SnsGetLifecycleResponse } from "@dfinity/sns";
+import { nonNullish } from "@dfinity/utils";
+import { writable, type Readable } from "svelte/store";
+
+interface SnsLifecycleData {
+  data: SnsGetLifecycleResponse;
+  certified: boolean;
+}
+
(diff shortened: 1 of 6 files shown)


frontend/src/lib/stores/sns-lifecycle.store.ts
```diff
@@ -0,0 +1,56 @@
...
+export interface SnsLifecycleStore
+  extends Readable<SnsLifecycleData | undefined> {
+  setData: (data: SnsLifecycleData) => void;
+  reset: () => void;
+}
+
+let stores: Map<string, SnsLifecycleStore> = new Map();
+
+export const resetLiefcycleStoresForTesting = () => {
+  stores = new Map();
```
dskloetd, line 20: If you do `stores.clear()` you can make `stores` const.

dskloetd (approved):

frontend/src/lib/stores/sns-lifecycle.store.ts
lmuntaner: good point!

PR #3133, opened 2023-08-16
Add tests for createAgent

dskloetd:
# Motivation

I want add a wrapper for `Agent` which provides a different `Identity` instance and use that to wrap agents cached in the agent API.
But before I do, I'm adding some test coverage to the agent API because currently there is none.

# Changes

Add tests for `createAgent` from `agent.api.ts`.

# Tests

yes

# Todos

will add when I add the agent wrapper


lmuntaner (approved):
Thanks!

PR #3134, opened 2023-08-17
SNS aggregator always supplies an partial last page

bitdivine:
# Motivation
We wish to change the logic for getting all pages:
* From: Get pages indexed 0, 1, 2, ... until there is a 404
* To: Get pages until there is an incomplete page

This requires the caller to know how many SNSs are in a complete page and also stops prematurely if the cache is still missing some entries, as the cache is filled from the latest index to the earliest.  However this logic seems to be convenient for the client.

# Changes
* Always provide an incomplete last page.

# Tests
* The test that checks the pagination has been updated and is now also run in the upgrade/downgrade test.
  * Note: I would like to test with no SNSs as well but `dfx stop ; dfx start --clean` just fails too often.

# Todos

### DIFF
--- a/rs/sns_aggregator/src/state.rs
+++ b/rs/sns_aggregator/src/state.rs
@@ -116,6 +116,11 @@ impl StableState {
 thread_local! {
     /// Single global container for state
     pub static STATE: State = State::default();
+
+    /// The invariant that the last page is not full has been established.
+    ///
+    /// Note: Once the invariant is established, it must be maintained by checking whenever the max index is updated.
+    pub static LAST_PAGE_NOT_FULL: RefCell<bool> = RefCell::new(false);
 }
 
 /// Log to console and store for retrieval by query calls.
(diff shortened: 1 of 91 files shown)


rs/sns_aggregator/src/state.rs
```diff
@@ -230,16 +233,43 @@ impl State {
...
+    fn slow_data_asset_v1(slow_data: &[SlowSnsData]) -> Asset {
+        let json_data = serde_json::to_string(&slow_data).unwrap_or_default();
+        Asset {
+            headers: Vec::new(),
+            bytes: json_data.into_bytes(),
+        }
+    }
+    /// If the last page is full, create an empty next page.
+    fn ensure_last_page_is_not_full_v1(state: &State) {
+        let (last_page, entries) = {
```
lmuntaner, line 256: Could you rename to `(last_page, last_page_entries)`. It's clearer to me what "entries" is this way. Thanks!

lmuntaner (approved):
LGTM! Thanks!

Just a minor comment of renaming. But already approving.

PR #3135, opened 2023-08-17
GIX-1824: Fix Canisters.spec

lmuntaner:
# Motivation

Some test cases in Canisters.spec relied on the page to be rendered in previous unit tests to pass.

# Changes

* Render the page in the specific test cases instead or relying in the previous renders.
* Clear all mocks to make sure the call belong to the test.

# Tests

Only test changes

# Todos

### DIFF
--- a/frontend/src/tests/lib/pages/Canisters.spec.ts
+++ b/frontend/src/tests/lib/pages/Canisters.spec.ts
@@ -56,11 +56,13 @@ describe("Canisters", () => {
   });
 
   it("should subscribe to store", () => {
+    expect(authStoreMock).not.toHaveBeenCalled();
     render(Canisters);
     expect(authStoreMock).toHaveBeenCalled();
   });
 
   it("should load canisters", () => {
+    expect(listCanisters).not.toHaveBeenCalled();
     render(Canisters);
     expect(listCanisters).toHaveBeenCalled();
   });
--- a/frontend/src/tests/lib/pages/Canisters.spec.ts
+++ b/frontend/src/tests/lib/pages/Canisters.spec.ts
@@ -54,10 +54,15 @@ describe("Canisters", () => {
     expect(getByText(en.core.ic)).toBeInTheDocument();
   });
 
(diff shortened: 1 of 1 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/pages/Canisters.spec.ts
```diff
@@ -54,10 +55,15 @@ describe("Canisters", () => {
     expect(getByText(en.core.ic)).toBeInTheDocument();
   });
 
-  it("should subscribe to store", () =>
-    expect(authStoreMock).toHaveBeenCalled());
+  it("should subscribe to store", () => {
+    render(Canisters);
+    expect(authStoreMock).toHaveBeenCalled();
```
dskloetd, line 60: Can you add an expectation that it's not yet called before the render call?

frontend/src/tests/lib/pages/Canisters.spec.ts
```diff
@@ -54,10 +55,15 @@ describe("Canisters", () => {
...
+  });
 
-  it("should load canisters", () => expect(listCanisters).toHaveBeenCalled());
+  it("should load canisters", () => {
+    render(Canisters);
+    expect(listCanisters).toHaveBeenCalled();
```
dskloetd, line 65: same

lmuntaner:

frontend/src/tests/lib/pages/Canisters.spec.ts
dskloetd: Yes, I think that's ok. Especially if they are individual improvements and not changes across the file that you need to see all at the same time to understand.

dskloetd (approved):

frontend/src/tests/lib/pages/Canisters.spec.ts
lmuntaner: Noted!

PR #3138, opened 2023-08-17
GIX-1823: Converter agg data to SnsSummary

lmuntaner:
# Motivation

In this PR: Introduce the converter from sns aggreagator data to SnsSummary.

The SnsSummary type is used in the `snsSummariesStore` to return a list of SnsSummary.

After this PR, we can enable using the aggregator data instead of snsQueryStore.

Then, we can remove snsQueryStore.

# Changes

* New sns aggregator converter `convertDtoToSnsSummary`. Some of the utils are similar to utils in sns.utils.ts. I didn't use those on purpose because I want to remove them along the snsQueryStore.

# Tests

* Add a test for the new converter `convertDtoToSnsSummary`.

# Todos

Not yet worth an entry.

### DIFF
--- a/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
+++ b/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
@@ -372,7 +372,7 @@ const convertDtoToTokenMetadata = (
 
 const convertDtoToSnsSummarySwap = (
   swap: CachedSnsSwapDto
-): SnsSummarySwap | undefined => ({
+): SnsSummarySwap => ({
   ...convertSwap(swap),
   decentralization_sale_open_timestamp_seconds: convertOptionalNumToBigInt(
     swap.decentralization_sale_open_timestamp_seconds
@@ -384,16 +384,13 @@ const convertDtoToSnsSwapDerivedState = (
   data: CachedSnsSwapDerivedDto
 ): SnsSwapDerivedState => convertDerived(data);
 
-const isValidSummary = (entry: Partial<SnsSummary>): entry is SnsSummary =>
-  entry.swap !== undefined &&
-  entry.swap.params !== undefined &&
-  entry.swapCanisterId !== undefined &&
-  entry.governanceCanisterId !== undefined &&
-  entry.ledgerCanisterId !== undefined &&
-  entry.indexCanisterId !== undefined &&
-  entry.derived !== undefined &&
-  entry.metadata !== undefined &&
-  entry.token !== undefined;
+type PartialSummary = Omit<SnsSummary, "metadata" | "token"> & {
+  metadata?: SnsSummaryMetadata;
+  token?: IcrcTokenMetadata;
+};
+
+const isValidSummary = (entry: PartialSummary): entry is SnsSummary =>
+  entry.metadata !== undefined && entry.token !== undefined;
 
 export const convertDtoToSnsSummary = ({
   canister_ids: {
@@ -408,7 +405,7 @@ export const convertDtoToSnsSummary = ({
   swap_state,
   derived_state,
 }: CachedSnsDto): SnsSummary | undefined => {
-  const partialSummary: Partial<SnsSummary> = {
+  const partialSummary: PartialSummary = {
     rootCanisterId: Principal.from(root_canister_id),
     swapCanisterId: Principal.from(swap_canister_id),
     governanceCanisterId: Principal.from(governance_canister_id),
--- a/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
+++ b/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
@@ -1,5 +1,12 @@
+import DEFAULT_SNS_LOGO from "$lib/assets/sns-logo-default.svg";
 import { SNS_AGGREGATOR_CANISTER_URL } from "$lib/constants/environment.constants";
 import { AGGREGATOR_CANISTER_VERSION } from "$lib/constants/sns.constants";
+import type { IcrcTokenMetadata } from "$lib/types/icrc";
+import type {
+  SnsSummary,
+  SnsSummaryMetadata,
+  SnsSummarySwap,
+} from "$lib/types/sns";
 import type {
   CachedFunctionTypeDto,
   CachedLifecycleResponseDto,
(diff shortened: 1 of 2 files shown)


frontend/src/lib/utils/sns-aggregator-converters.utils.ts
```diff
@@ -334,3 +338,84 @@ const convertSnsData = ({
 
 export const convertDtoData = (data: CachedSnsDto[]): CachedSns[] =>
   data.map(convertSnsData);
+
+/**
+ * Metadata is given only if all its properties are defined.
```
dskloetd, line 343: Is it normal for some properties not to be defined or is this defensive programming?
Why or when would some properties not be defined?

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
```diff
@@ -1,11 +1,144 @@
...
   aggregatorSnsMockDto,
 } from "$tests/mocks/sns-aggregator.mock";
+import { Principal } from "@dfinity/principal";
 
 describe("sns aggregator converters utils", () => {
-  it("converts aggregator types to ic-js types", () => {
-    expect(convertDtoData([aggregatorSnsMockDto])).toEqual([aggregatorSnsMock]);
+  describe("convertDtoData", () => {
+    it("converts aggregator types to ic-js types", () => {
+      expect(convertDtoData([aggregatorSnsMockDto])).toEqual([
```
dskloetd, line 14: I'm not that comfortable relying on some arbitrary mock to give us good coverage.
How about having an explicit value here in the test?
It can currently be the same as the mock but if the mock changes for some other reason at least we don't lose coverage.
We can even do both.

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
```diff
@@ -1,11 +1,144 @@
...
+    it("returns sns summary from aggregator data", () => {
+      const {
+        canister_ids: {
+          root_canister_id,
+          swap_canister_id,
+          governance_canister_id,
+          ledger_canister_id,
+          index_canister_id,
+        },
+      } = aggregatorSnsMockDto;
```
dskloetd, line 30: Same. Let's not rely on mock data from another file.

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
dskloetd, line 1: We now have 2 ways of converting the same data, right?
1. Through snsQueryStore
2. Not through snsQueryStore

Do you think it would make sense to at some point enable both conversions simultaneously (e.g. through a feature flag) and check that both conversions result in exactly the same data and log some errors if not?
Then we can be confident that nothing which currently exists in prod data is missed.

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
```diff
@@ -324,7 +328,7 @@ const convertSnsData = ({
   icrc1_total_supply: BigInt(icrc1_total_supply),
   derived_state: convertDerivedToResponse(derived_state),
   swap_params: {
-    params: convertSwapParams(swap_params.params),
+    params: [convertSwapParams(swap_params.params)],
```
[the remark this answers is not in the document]
lmuntaner: Not here, it was only in the "swap_state" that was optional.

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
lmuntaner: I want to ask the NNS Team how much they are enforced and why they are optional.

From the type, they are optional, that's why I do this here.

What would you do if the type is optional, but they can never be so? Throw an error instead of ignoring the data?

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
```diff
@@ -334,3 +338,84 @@ const convertSnsData = ({
...
+  ...convertSwap(swap),
+  decentralization_sale_open_timestamp_seconds: convertOptionalNumToBigInt(
+    swap.decentralization_sale_open_timestamp_seconds
+  ),
+  params: convertSwapParams(swap.params),
+});
+
+const convertDtoToSnsSwapDerivedState = (
+  data: CachedSnsSwapDerivedDto
+): SnsSwapDerivedState => convertDerived(data);
```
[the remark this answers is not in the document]
lmuntaner: True, I started setting the signatures and returns I wanted and I didn't realize this one was not doing anything.

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
lmuntaner: I was thinking the same thing and I wasn't sure about it.

I believe that using a feature flag and test it in mainnet is safer. But at the same time, I can't think why it wouldn't work in testnet and not in mainnet.

But because the SNS aggregator is not typed, let's be safe. I'll use a feature flag for `snsSummariesStore`.

lmuntaner: In another PR though. None is the code in this PR is used in production or even testnet at the moment.

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
lmuntaner: This is not part of this PR. But I guess I can do it in another PR.

I just wrapped this test in a describe.

lmuntaner: Moreover, `convertDtoData` is temporary. So, I'm not sure it makes sense to spend much time on it.

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
lmuntaner: sure

lmuntaner:

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
dskloetd: I would prefer to throw an error but I would need to really understand the situation.
It's also not clear to me why we drop all the metadata if just some of the metadata is missing.
But we can go with this and see what the NNS team answers.

dskloetd (approved):

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
lmuntaner: Accepting partial metadata makes sense to me. That would mean changing a few things though.

Let's better understand it and then we decide.

PR #3143, opened 2023-08-18
Delete e2e-tests/

dskloetd:
# Motivation

Wdio e2e tests have been replaced by Playwright e2e tests.
Scripts that we want to keep have been moved to `scripts/e2e-tests/`.
Time to delete the remaining files in `e2e-tests/`.

# Changes

`git rm -r e2e-tests`

# Tests

CI still passes

# Todos


bitdivine (approved):
Thank you.

I would not be surprised if there is something, somewhere that uses some of this but when that situation arises, we can deal with it.  The .pem files used to be used by the NNS team but I _think_ that is history.

PR #3144, opened 2023-08-18
GIX-1823: Use one store for sns derived state

lmuntaner:
# Motivation

We want to use the snsDerivedStateStore in the derived store `snsSummariesStore`. To do that, we can't use dynamic creationg of the store. To allow dynamic creation of the store, we would need to also create dynamically `snsSummariesStore`. This is a bigger refactor than expected and not sure it's worth the effort now.

Therfore, I implemented the same pattern of storing all project data in one single store using the root canister id as key in the snsDerivedStateStore.

# Changes

* `snsDerivedStateStore` doesn't create store dynamically. It's one store to rule them all.
* Refactor where this store was created.

# Tests

* Test new capabilities and remove old ones.
* Change test after refactor.

# Todos

Not worth a changelog entry.

### DIFF
--- a/frontend/src/lib/stores/sns-derived-state.store.ts
+++ b/frontend/src/lib/stores/sns-derived-state.store.ts
@@ -3,14 +3,22 @@ import type { SnsGetDerivedStateResponse } from "@dfinity/sns";
 import { nonNullish } from "@dfinity/utils";
 import { writable, type Readable } from "svelte/store";
 
-interface SnsDerivedStateData {
+interface SnsDerivedStateProjectData {
   derivedState: SnsGetDerivedStateResponse;
   certified: boolean;
 }
 
+interface SnsDerivedStateData {
+  [rootCanisterId: string]: SnsDerivedStateProjectData;
+}
+
 export interface SnsDerivedStateStore
   extends Readable<SnsDerivedStateData | undefined> {
-  setDerivedState: (data: SnsDerivedStateData) => void;
+  setDerivedState: (params: {
+    data: SnsGetDerivedStateResponse;
+    certified: boolean;
+    rootCanisterId: Principal;
+  }) => void;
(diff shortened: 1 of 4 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/stores/sns-derived-state.store.ts
```diff
@@ -3,14 +3,22 @@ import type { SnsGetDerivedStateResponse } from "@dfinity/sns";
...
+  [rootCanisterId: string]: SnsDerivedStateProjectData;
+}
+
 export interface SnsDerivedStateStore
   extends Readable<SnsDerivedStateData | undefined> {
-  setDerivedState: (data: SnsDerivedStateData) => void;
+  setDerivedState: (params: {
+    data: SnsGetDerivedStateResponse;
+    certified: boolean;
+    rootCanisterId: Principal;
```
dskloetd, line 20: Can you put `rootCanisterId` before `data`?
I was confused by not seeing the "key" type first while expecting a key/value map.

lmuntaner: sure

frontend/src/lib/stores/sns-derived-state.store.ts
```diff
@@ -52,7 +74,7 @@ export const getOrCreateDerivedStateStore = (
     return existingStore;
   }
 
-  const newStore = createSnsDerivedStateStore();
+  const newStore = initSnsDerivedStateStore();
```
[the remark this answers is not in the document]
lmuntaner: I need to remove this. The `getOrCreateDerivedStateStore` is not used anymore.

lmuntaner:

dskloetd (approved):

PR #3145, opened 2023-08-18
GIX-1823: Use one snsLifecycleStore

lmuntaner:
# Motivation

Similar to #3144 

We want to use the snsLifecycleStore in the derived store snsSummariesStore. To do that, we can't use dynamic creation of the store. To allow dynamic creation of the store, we would need to also create dynamically snsSummariesStore. This is a bigger refactor than expected and not sure it's worth the effort now.

Therefore, I implemented the same pattern of storing all project data in one single store using the root canister id as key in the snsLifecycleStore.

# Changes

* Change snsLifecycleStore to use only one store to rule them all.
* Change usage of the store.

# Tests

* Change test cases for snsLifecycleStore.
* Change related tests after refactor.

# Todos

Not worth a changelog entry


lmuntaner:
@dskloetd please review

PR #3147, opened 2023-08-18
Fix: Do not use real time in tests

lmuntaner:
# Motivation

Two days ago, I had a unit test failure because a test was relying on real time instead of setting the system time.

In this PR: I looked at all the tests that were using real time and changed it.

# Changes

* Set time for NnsNeuronCard.spec.
* Set time and clarify dependencies of creation times with a variable in SnsVotingCard.spec.
* Set time in NnsProposalCard.spec
* Set time in sns-neuron.utils.spec

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/tests/lib/components/proposal-detail/VotingCard/SnsVotingCard.spec.ts
+++ b/frontend/src/tests/lib/components/proposal-detail/VotingCard/SnsVotingCard.spec.ts
@@ -4,6 +4,7 @@
 
 import * as snsGovernanceApi from "$lib/api/sns-governance.api";
 import SnsVotingCard from "$lib/components/sns-proposals/SnsVotingCard.svelte";
+import { SECONDS_IN_DAY } from "$lib/constants/constants";
 import { authStore } from "$lib/stores/auth.store";
 import { snsNeuronsStore } from "$lib/stores/sns-neurons.store";
 import { snsParametersStore } from "$lib/stores/sns-parameters.store";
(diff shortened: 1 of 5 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/components/proposal-detail/VotingCard/SnsVotingCard.spec.ts
```diff
@@ -63,13 +65,15 @@ describe("SnsVotingCard", () => {
...
+  const neuronCreatedAt = BigInt(nowInSeconds - SECONDS_IN_DAY * 2);
+  const proposalCreatedAt = BigInt(nowInSeconds - SECONDS_IN_DAY);
   const testProposal: SnsProposalData = {
     ...createSnsProposal({
       proposalId: 123n,
       status: SnsProposalDecisionStatus.PROPOSAL_DECISION_STATUS_OPEN,
       rewardStatus: SnsProposalRewardStatus.PROPOSAL_REWARD_STATUS_ACCEPT_VOTES,
       ballots: testBallots,
-      createdAt: BigInt(Math.round(Date.now() / 1000)),
+      createdAt: proposalCreatedAt,
```
dskloetd, line 76: Previously the proposal was created now, but now it's created yesterday.
Was this change necessary?

dskloetd: Thanks for clarifying.

dskloetd (approved):

PR #3149, opened 2023-08-18
Refactor: Create principal id from index

lmuntaner:
# Motivation

Have an easier way to create multiple principals.

# Changes

* Reimplement the function `principal` from `sns-projects.mock` to create a principal based on a number.
* Remove check in `setSnsProjects` for number of projects to set based on the previous limitation of `principal`.

# Tests

* Only test changes.

# Todos

### DIFF
--- a/frontend/src/tests/mocks/sns-projects.mock.ts
+++ b/frontend/src/tests/mocks/sns-projects.mock.ts
@@ -26,7 +26,7 @@ import {
   type SnsTransferableAmount,
 } from "@dfinity/sns";
 import type { Token } from "@dfinity/utils";
-import { nonNullish, toNullable } from "@dfinity/utils";
+import { hexStringToUint8Array, nonNullish, toNullable } from "@dfinity/utils";
 import type { Subscriber } from "svelte/store";
 
 export const mockProjectSubscribe =
@@ -38,23 +38,7 @@ export const mockProjectSubscribe =
   };
 
 export const principal = (index: number): Principal =>
-  [
-    Principal.fromText(
-      "2vtpp-r6lcd-cbfas-qbabv-wxrv5-lsrkj-c4dtb-6ets3-srlqe-xpuzf-vqe"
-    ),
-    Principal.fromText(
-      "nv24n-kslcc-636yn-hazy3-t2zgj-fsrkg-2uhfm-vumlm-vqolw-6ciai-tae"
-    ),
-    Principal.fromText(
-      "2lwez-knpss-xe26y-sqpx3-7m5ev-gbqwb-ogdk4-af53j-r7fed-k5df4-uqe"
-    ),
-    Principal.fromText(
-      "vxi5c-ydsws-tmett-fndw6-7qwga-thtxc-epwtj-st3wy-jc464-muowb-eqe"
-    ),
-    Principal.fromText(
-      "4etav-nasrq-uvswa-iqsll-6spts-ryhsl-e4yf6-xtycj-4sxvp-ciay5-yae"
-    ),
-  ][index];
+  Principal.fromUint8Array(hexStringToUint8Array(index.toString(16) + "0x01"));
 
 export const createTransferableAmount = (
   amount: bigint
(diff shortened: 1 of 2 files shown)


frontend/src/tests/mocks/sns-projects.mock.ts
```diff
@@ -37,24 +37,9 @@ export const mockProjectSubscribe =
...
-      "2lwez-knpss-xe26y-sqpx3-7m5ev-gbqwb-ogdk4-af53j-r7fed-k5df4-uqe"
-    ),
-    Principal.fromText(
-      "vxi5c-ydsws-tmett-fndw6-7qwga-thtxc-epwtj-st3wy-jc464-muowb-eqe"
-    ),
-    Principal.fromText(
-      "4etav-nasrq-uvswa-iqsll-6spts-ryhsl-e4yf6-xtycj-4sxvp-ciay5-yae"
-    ),
-  ][index];
+  Principal.fromUint8Array(hexStringToUint8Array(index.toString(16) + "0x01"));
```
dskloetd, line 42: `index.toString(16) + "0x01"`
is not a valid hex string.
This only works because `Uint8Array.from()` converts `NaN` to  `0` and because we only care that we get some principal and not what exactly.

But why not just use `Principal.fromHex(index.toString(16))`?

lmuntaner: I wanted to follow the documentation and set an opaque principal. But I guess it doesn't really matter.

dskloetd: Which documentation?

lmuntaner:

frontend/src/tests/mocks/sns-projects.mock.ts
dskloetd: Thanks for the link!

`0x` is just a prefix which indicates that the digits following it should be interpreted as hexadecimal. For example if you type `0x23` into python, you get `35` because 2 * 16 + 3 = 35.

So in the documentation, where they write `0x01`, that's really just the number 1. I guess the only reason they wrote it as `0x01` is to make it clear that it should be interpreted as binary data, rather than just a number or a character. And because a hexadecimal digit is 4 bits, 1 byte is 2 digits. Hence `01` for 1 byte, and `0x01` including the prefix.

But when you pass a string to a function that only accepts hexadecimal, it's already clear that the string should be interpreted as hexadecimal and it's likely that the function doesn't expect the `"0x"` prefix. And indeed, `hexStringToUint8Array` ([code](https://github.com/dfinity/ic-js/blob/e5679f7f7773ed7e389ee8488c765e0ad7490fdc/packages/utils/src/utils/arrays.utils.ts#L52)) does not.

Additionally, when you concatenate hexadecimal strings and the `0x` prefix of the second string ends up in the middle of the string, you get something meaningless.

So to get the `0x01` byte at the end, as suggested by the documentation, you could have appended `"01"` instead of `"0x01"`.

But I also discovered a bug in the implementation of both `hexStringToUint8Array` and `Principal.fromHex` which affects both the code you had before and my suggested code. Because while it looks like the code should work with an odd number, of digits, the way it does that, means that `"101"` and `"1001"` result in the same binary data (and thus principal). Because `"101"` gets split into `"10"` and `"1"`, which results in `[16, 1]`, and `"1001"` gets split into `"10"` and `"01"`, which also results in `[16, 1]`.

So with your previous implementation `principal(1)` and `principal(16)` would be the same, but with my current implementation, `principal(257)` and `principal(4097)` would be the same. The solution is to prepend `"0"` if the hex string has odd length, either in `hexStringToUint8Array`/`Principal.fromHex` or before calling either.

So I would propose the following code (with or without the `"01"` suffix) here until those utility functions are fixed:
```
export const principal = (index: number): Principal => {
  let hexString = index.toString(16) + "01";
  if (hexString.length % 2 == 1) {
    hexString = "0" + hexString;
  }
  return Principal.fromHex(hexString);
}
```

lmuntaner: Thanks for the explanation!! Done!

PR #3150, opened 2023-08-18
detailed min participation

dskloetd:
# Motivation

Modclub set `"min_participant_icp_e8s":100000278` which we round in the UI to 1.00 ICP.
Then when you try to participate with 1 ICP you get an error saying
"Sorry, the amount is too small. You need a minimum of 1.00 ICP to participate in this swap."

# Changes

Call `formatToken` with `detailed: true`. This renders all significant digits if amount >= 0.01 (if I understand the code correctly).

# Tests

no

# Todos

### DIFF
--- a/frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
@@ -85,7 +85,7 @@
   </KeyValuePair>
   <KeyValuePair>
     <span slot="key">{$i18n.sns_project_detail.min_commitment} </span>
-    <AmountDisplay slot="value" amount={minCommitmentIcp} singleLine />
+    <AmountDisplay slot="value" amount={minCommitmentIcp} detailed singleLine />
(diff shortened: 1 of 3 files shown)


frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
```diff
@@ -85,7 +85,7 @@
   </KeyValuePair>
   <KeyValuePair>
     <span slot="key">{$i18n.sns_project_detail.min_commitment} </span>
-    <AmountDisplay slot="value" amount={minCommitmentIcp} singleLine />
+    <AmountDisplay slot="value" amount={minCommitmentIcp} detailed singleLine />
```
lmuntaner, line 88: This will make it inconsistent with the other tokens

lmuntaner: ah, ok!

lmuntaner (approved):

PR #3156, opened 2023-08-21
Page object fire input event

mstrasinskis:
# Motivation

Adds fire input event to the page object. Will be used to update percentage selector range.

# Tests

No logic, just a svelte.fireEvent

# Todos

Not worth mentioning.

### DIFF
--- a/frontend/src/tests/page-objects/jest.page-object.ts
+++ b/frontend/src/tests/page-objects/jest.page-object.ts
@@ -139,6 +139,11 @@ export class JestPageObjectElement implements PageObjectElement {
     await fireEvent.click(this.element);
   }
 
+  async input(value: string): Promise<void> {
+    await this.waitFor();
+    await fireEvent.input(this.element, { target: { value } });
+  }
+
   async typeText(text: string): Promise<void> {
     await this.waitFor();
     // Svelte generates code for listening to the `input` event, not the `change` event in input fields.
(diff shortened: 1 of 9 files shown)


frontend/src/tests/page-objects/jest.page-object.ts
```diff
@@ -139,6 +139,11 @@ export class JestPageObjectElement implements PageObjectElement {
     await fireEvent.click(this.element);
   }
 
+  async input(value: string): Promise<void> {
+    await this.waitFor();
+    await fireEvent.input(this.element, { target: { value } });
```
dskloetd, line 144: This looks identical to the `typeText` method below. Can you use that?

mstrasinskis: It will look wrong to use `typeText` for the input type="range". How about renaming `typeText` to `input` to have the same pattern with the `click` method?

dskloetd: I don't know input type="range". Can you give me some more context?

mstrasinskis: As discussed, the `input` function was reused in the `typeText`.

frontend/src/tests/page-objects/playwright.page-object.ts
```diff
@@ -100,6 +100,11 @@ export class PlaywrightPageObjectElement implements PageObjectElement {
     return this.locator.click();
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
dskloetd, line 103: Please change `value` to `_value` instead to indicate that the variable isn't used.

mstrasinskis: Good point, updated!

dskloetd (approved):

PR #3154, opened 2023-08-21
Aggregator favicon, spacing, open color and text clamp

peterpeterparker:
# Changes

- replace empty favicon with IC favicon
- improve article spacing and alignment (row-gap, align-self etc.)
- clamp Snses titles on three lines
- improve render performance (`append(...)` instead of x `appendChild`)
- new background color for status "open"

# Screenshots

Capture d’écran 2023-08-21 à 07 22 24

PR #3157, opened 2023-08-21
Fix changelogs

peterpeterparker:
# Motivation

Changelogs formatter introduced in PR #3154 did not preserve correctly the history. This PR redo the correct entries and apply format.


dskloetd:
Like this it's hard to tell from the diff if it's correct. I prefer if you revert the entire PR (https://github.com/dfinity/nns-dapp/pull/2964). Then we can discuss what we want as a team and if we indeed what the changes, we can carefully review them from the original.

peterpeterparker:
👉 https://github.com/dfinity/nns-dapp/pull/3158

PR #3159, opened 2023-08-21
Build/format changelogs

bitdivine:
# Motivation
The prettier formatter can be applied to markdown files.

# Changes
*  Apply the markdown formatting command to all .md files.
   * The changelogs are _not_ formatted as some people had trouble merging changes.  We need to make it easy for users to do it right.
   * The command fails for markdown files under the frontend directory, as the npm packages there interfere.  We can skip those until we know how to fix them.  By definition, for a quick win we don't  want to get bogged down in the last 5%.
* Add the standard flags for `./scripts/fmt-XXX` scripts, including `--check` and `--help`.

# Tests
* See CI
* Note that in the PR, the formatted text, e.g. the table in `BUILD.md`, is indeed noticeably  more readable.

# Todos

### DIFF
--- a/.github/actions/slack/README.md
+++ b/.github/actions/slack/README.md
@@ -2,10 +2,9 @@
 
 A GitHub Action for sending slack notifications. For more information, see slack's [documentation](https://api.slack.com/messaging/webhooks). In particular, note that the channel and workspace are set by the webhook URL, not directly by this action.
 
-
 ## Usage
 
-``` yaml
+```yaml
 name: My Action
 
 on:
--- a/scripts/fmt-md
+++ b/scripts/fmt-md
@@ -16,8 +16,9 @@ list_files() {
 
 # Selects eligible files; filenames are read from stdin, one line per filename.
 # TODO: Formatting fails in the frontend directory, due to some npx or prettier magic.  Fix.
+# TODO: Remove the union merge rule for one week, to help PRs that struggle to merge correctly.
 filter() {
-  grep -E '[.]md$' | grep -v frontend
+  grep -E '[.]md$' | grep -v frontend | grep -v CHANGELOG
 }
 
 case "${1:-}" in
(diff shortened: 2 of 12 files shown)


.github/actions/slack/README.md
```diff
@@ -2,10 +2,9 @@
 
 A GitHub Action for sending slack notifications. For more information, see slack's [documentation](https://api.slack.com/messaging/webhooks). In particular, note that the channel and workspace are set by the webhook URL, not directly by this action.
 
-
```
dskloetd, line 5: I'm not convinced that this is a good idea.
Markdown files should be readable as text files.
If people put extra blank lines, they probably did because they found it more readable.
I don't really see the benefit of having one strict style for markdown files.
Also I like `*` as bullet points instead of `-`.

Anyway, all that to basically say (as I said before) that I think this should be discussed with the team first.

scripts/fmt-md
```diff
@@ -0,0 +1,35 @@
...
+
+# Selects eligible files; filenames are read from stdin, one line per filename.
+# TODO: Formatting fails in the frontend directory, due to some npx or prettier magic.  Fix.
+# TODO: Remove the union merge rule for one week, to help PRs that struggle to merge correctly.
+filter() {
+  grep -E '[.]md$' | grep -v frontend | grep -v CHANGELOG
+}
+
+case "${1:-}" in
+--help) print_help && exit 0 ;;
```
dskloetd, line 25: Any reason not to use clap?

bitdivine: Consistency.  All the lint and fmt commands parse manually.  At some stage it would probably make sense to update all of them, but that is not the focus of this PR.  There will probably be some breakages where the formatter is called and clap is not present; those can be fixed.

.github/actions/slack/README.md
bitdivine: We discussed this at the team meeting on 23 Aug 2023 and agreed to apply formatting to non-changelog files.

dskloetd (approved):

PR #3160, opened 2023-08-21
Chore: Add Launchpad test with prod data

lmuntaner:
# Motivation

We had a bug because the type of the aggregator in ths NNS Dapp was not correct and when it tried to convert a certain field it failed.

In this PR, I try to add a test that renders the Launchpad with prod data from the aggregator. I then check that there are committed projects and no errors.

# Changes

* New test  in workflows/Launchpad/ProdLaunchpad.spec.ts.
* New test component in workflows/Launchpad/LaunchpadWithLayout.svelte. This is used to trigger the same flows as the svelte app. In this scenario, the call to the SNS aggregator.

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
+++ b/frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
@@ -1,0 +1,56 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import * as proposalsApi from "$lib/api/proposals.api";
+import { authStore } from "$lib/stores/auth.store";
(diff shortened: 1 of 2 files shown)


frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
dskloetd, line 1: Did you try reintroducing the bug and seeing if the test fails?

lmuntaner: Yes, I did and the test failed.

dskloetd (approved):

PR #3161, opened 2023-08-21
Fix: SnsProposals.spec

lmuntaner:
# Motivation

The SnsProposals.spec was failing when being randomized.

The problem was that filters set in one tests were interfering with other tests. The solution is to reset the snsFiltersStore in the beforeEach.

# Changes

* Add a `snsFiltersStore.reset();` in the beforeEach of SnsProposals.spec.

# Tests

Only test changes

# Todos


dskloetd (approved):
Thanks!

PR #3162, opened 2023-08-21
E2E: Wait for content to load after going to the neurons tab

dskloetd:
# Motivation

Some [flakiness was observed](https://github.com/dfinity/nns-dapp/actions/runs/5925864493/job/16066262925?pr=3155) in the e2e test that disburses a neuron.
What happened is that after going to the neurons tab, no neuron cards were visible.
When I tried manually going from the accounts tab to the neurons tab, I noticed that the accounts tab was still there for a short time before the neurons tab renders.

# Changes

In `AppPo.goToNeurons()` wait for the neurons tab to be loaded before continuing.

# Tests

I wasn't able to reproduce the flakiness but the tests still pass with this change as well and it seems like it should help.

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #3163, opened 2023-08-21
Improve ProjectSwapDetails.spec.ts

dskloetd:
# Motivation

I wanted to add unit tests for https://github.com/dfinity/nns-dapp/pull/3150 but I noticed some issues in `ProjectSwapDetails.spec.ts` so I wanted to fix those first.

`ProjectSwapDetails.spec.ts` expects token values to appear in a certain order in the UI and it queries all of them with the same query and then uses a magic index into that list.
Because of this it also relies on different values being different otherwise we wouldn't notice of some swapped places in the order.

# Changes

1. Use page objects in all the tests in `ProjectSwapDetails.spec.ts`.
2. Set specific values on the input and expect those specific values to be rendered instead of depending on what happens to be in the mock data.
3. Add necessary test IDs.
4. Add getters on `ProjectSwapDetails.page-object.ts`.
5. Allow more customization in `createSummary`.

# Tests

Apart from added test ID in component, the changes are test only.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectSwapDetails.svelte
@@ -70,7 +70,7 @@
       <AmountDisplay slot="value" amount={snsTotalTokenSupply} singleLine />
     </KeyValuePair>
   {/if}
-  <KeyValuePair>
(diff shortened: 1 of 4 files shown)


lmuntaner (approved):
Thanks!

PR #3164, opened 2023-08-21
Fix: ProjectDetail.spec randomization

lmuntaner:
# Motivation

ProjectDetail.spec failed when tests were randomized.

The problem is hard to explain. When `"should participate without user interaction if there is an open ticket."` was called first, then `"successful participation"` failed.

They failed because the swap participation on reload return no participation. As if no participation went through and it didn't show in the UI.

This participation seemed to come from the mock set in `"should participate without user interaction if there is an open ticket."`. Because when I commented it, the test `"successful participation"` passed successfully.

Why that was, I don't know.

In this PR, I discovered a problem in the test `"should participate without user interaction if there is an open ticket."`. The problem doesn't seem to have relation with the problem, but it seems to fix it 🤷 .

The problem is that it was mocking only one call for no swap commitment. Instead of the two calls (query + update) that the code does.

Why this fixes it, it's a mystery to me. But I believe that the change was still worth it.

# Changes

* Add a mock once for `querySnsSwapCommitment` in `"should participate without user interaction if there is an open ticket."`.

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
+++ b/frontend/src/tests/lib/pages/ProjectDetail.spec.ts
@@ -538,6 +538,12 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
         };
         jest
           .spyOn(snsApi, "querySnsSwapCommitment")
+          // Query call
+          .mockResolvedValueOnce({
+            rootCanisterId,
+            myCommitment: initialCommitment,
+          } as SnsSwapCommitment)
+          // Update call
           .mockResolvedValueOnce({
             rootCanisterId,
             myCommitment: initialCommitment,


frontend/src/tests/lib/pages/ProjectDetail.spec.ts
```diff
@@ -538,6 +538,12 @@ sale_buyer_count ${saleBuyerCount} 1677707139456
         };
         jest
           .spyOn(snsApi, "querySnsSwapCommitment")
+          // Query call
+          .mockResolvedValueOnce({
```
dskloetd, line 542: Can you add expectations that `querySnsSwapCommitment` starts out being called 0 times and ends up being called 3 times?
That way we make sure that adding this behavior doesn't change the spy for other tests since the "once" behaviors are all used up inside this test.

I did a little bit of experimentation and I noticed that without this extra behavior, `querySnsSwapCommitment` is actually only called twice. I'm guessing that the `waitFor` on line 565 finishes early because the update call already receives the final commitment and we never finish going through sale. I'm guessing that results in leaving behind some state for other test.
And ideally we would reset that state between every test. Is there some ticket store or something that we could reset?

If you don't want to spend more time on investigating, can you at least add the 2 expectations I asked in the first paragraph? Then I'll approve.

lmuntaner: I added a reset to the snsTicketsStore. It doesn't fix the test, but I believe it makes sense.

I added three expects. One that querySnsSwapCommitment is not called at the beginning, to be called twice in the middle, and three times at the end.

lmuntaner:

dskloetd (approved):

PR #3166, opened 2023-08-21
Refactor, removing a rarely used legacy migration out of the main flow

bitdivine:
# Motivation
The `add_account()` function is obscured by a long and rarely used migration for early accounts.  This is awkward and the lack of clarity is potentially dangerous as we migrate the data yet again.

# Changes
* Move the transaction migration into a separate function.

# Tests
* There is already a test for the migration, so existing CI should suffice.

# Todos

  - Not necessary, IMO.

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -334,13 +334,12 @@ impl AccountsStore {
     // without storing each user's principal).
     pub fn add_account(&mut self, caller: PrincipalId) -> bool {
         let account_identifier = AccountIdentifier::from(caller);
-        if let Some(account) = self.accounts.get(&account_identifier.to_vec()) {
+        if let Some(account) = self.accounts.get_mut(&account_identifier.to_vec()) {
             if account.principal.is_none() {
                 // This is an old account that needs a one-off fix to set the principal and update the transactions.
                 let mut account = account.clone();
                 account.principal = Some(caller);
                 self.fix_transactions_for_early_user(&account, caller);
-                self.accounts.insert(account_identifier.to_vec(), account);
             }
             false
         } else {
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -50,6 +50,9 @@ enum AccountWrapper {
 
 #[derive(CandidType, Deserialize, Debug, Eq, PartialEq, Clone)]
 pub struct Account {
+    /// The user principal.
+    ///
+    /// Note: The principal was not stored for early users.  When early users log in, we discover their principal and set this field.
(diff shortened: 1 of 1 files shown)


rs/backend/src/accounts_store.rs
```diff
@@ -331,86 +334,92 @@ impl AccountsStore {
     // without storing each user's principal).
```
bitdivine, line 334: Here we move the big block of code that migrates the transactions into a separate function.

rs/backend/src/accounts_store.rs
```diff
@@ -331,86 +334,92 @@ impl AccountsStore {
...
+    fn fix_transactions_for_early_user(&mut self, account: &Account, caller: PrincipalId) {
+        let canister_ids: Vec<dfn_core::CanisterId> = account.canisters.iter().map(|c| c.canister_id).collect();
+        let transactions: Vec<TransactionIndex> = account.get_all_transactions_linked_to_principal_sorted();
+
+        // Now that we know the principal we can set the transaction types. The
+        // transactions must be sorted since some transaction types can only be
+        // determined based on earlier transactions (eg. we can only detect
+        // TopUpNeuron transactions that happen after StakeNeuron transactions).
+        for transaction_index in transactions {
+            let transaction = self.get_transaction(transaction_index).unwrap();
```
bitdivine, line 366: Note: Transactions _will_ eventually be pruned, in which case that unwrap will panic.  We can update the logic so that the transaction is updated only if it has not been pruned, but preferably not in this PR.  I don't like refactoring that does not change the semantics of the code with changes to the behaviour.  What do you think?

rs/backend/src/accounts_store.rs
```diff
@@ -331,86 +334,91 @@ impl AccountsStore {
...
-            let mut transactions: Vec<TransactionIndex> = vec![];
-            let account = self.accounts.get_mut(&account_identifier.to_vec()).unwrap();
+        if let Some(account) = self.accounts.get_mut(&account_identifier.to_vec()) {
             if account.principal.is_none() {
+                // This is an old account that needs a one-off fix to set the principal and update the transactions.
+                let mut account = account.clone();
                 account.principal = Some(caller);
-                canister_ids = account.canisters.iter().map(|c| c.canister_id).collect();
-                transactions = account.get_all_transactions_linked_to_principal_sorted();
-            }
```
bitdivine, line 342: Note: The for loop below does nothing if the `if` above is false, because then transactions is empty.  Thus we can move the for loop into the `if` clause.

Then, note that the vectors `canister_ids` and `transactions` are used only in the if clause, so we can move them into the if clause as well.  Then we have all the code for fixing up the transactions inside the if clause, and can move it out into a separate function.

PR #3167, opened 2023-08-22
Fix: Flaky test ProjectSwapDetails.spec

lmuntaner:
# Motivation

ProjectSwapDetails.spec failed because the white space what not the same.

# Changes

* Normalize the white space in the test of ProjectSwapDetails.spec.

# Tests

Only test changes

# Todos

Not necessary.

### DIFF
--- a/frontend/src/tests/lib/components/project-detail/ProjectSwapDetails.spec.ts
+++ b/frontend/src/tests/lib/components/project-detail/ProjectSwapDetails.spec.ts
@@ -15,7 +15,6 @@ import {
 import { renderContextCmp } from "$tests/mocks/sns.mock";
 import { ProjectSwapDetailsPo } from "$tests/page-objects/ProjectSwapDetails.page-object";
 import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
-import { normalizeWhitespace } from "$tests/utils/utils.test-utils";
 import { TokenAmount } from "@dfinity/utils";
 
 describe("ProjectSwapDetails", () => {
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/components/project-detail/ProjectSwapDetails.spec.ts
```diff
@@ -80,7 +81,9 @@ describe("ProjectSwapDetails", () => {
       swapCommitment: mockSnsFullProject.swapCommitment as SnsSwapCommitment,
     });
 
-    expect(await po.getSaleEnd()).toEqual("Oct 4, 2023 3:00 PM");
+    expect(normalizeWhitespace(await po.getSaleEnd())).toEqual(
```
dskloetd, line 84: Some getters in `ProjectSwapDetailsPo` already do `trim()` on the result before returning.
Do you think it makes sense for `normalizeWhitespace` to go inside `getSaleEnd()`?

Also, as far as I remember, the kind of whitespace used only depends on the Node version.
Have you seen this fail on CI or are you maybe using a different Node version when you see this fail?

dskloetd (approved):

frontend/src/tests/lib/components/project-detail/ProjectSwapDetails.spec.ts
```diff
@@ -80,7 +80,7 @@ describe("ProjectSwapDetails", () => {
       swapCommitment: mockSnsFullProject.swapCommitment as SnsSwapCommitment,
     });
 
-    expect(await po.getSaleEnd()).toEqual("Oct 4, 2023 3:00 PM");
+    expect(await po.getSaleEnd()).toEqual("Oct 4, 2023 3:00 PM");
```
lmuntaner, line 83: The change is using a normal whitespace

frontend/src/tests/lib/components/project-detail/ProjectSwapDetails.spec.ts
lmuntaner: done

dskloetd (approved):
Thanks!

PR #3168, opened 2023-08-22
Fix: More SNS aggregator optional types

lmuntaner:
# Motivation

We realized that some of our SNS aggregator types were not correct. It seems that the changes were not exhaustive and there are more fields that can still be optional.

I based my findings on the actual types of the canisters. Where the response said `[] | [T]`, I changed it to `null | T`.

# Changes

* Change value of some fields in SNS aggregator to make them optional.
* Fix converters after change.

# Tests

* Test the new functionality when converting the summary that might return undefined if swap params are not defined.
* Add a test for `convertNervousFuncttion` (typo will be addressed in another PR)

# Todos

### DIFF
--- a/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
+++ b/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
@@ -100,7 +100,7 @@ export const convertNervousFuncttion = ({
 });
 
 const convertSwapInitParams = (
-  init: CachedSwapInitParamsDto
+  init: CachedSwapInitParamsDto | null
 ): [SnsSwapInit] | [] =>
   nonNullish(init)
     ? toNullable({
@@ -225,7 +225,7 @@ const convertSwap = ({
     open_sns_token_swap_proposal_id !== undefined
       ? toNullable(convertOptionalNumToBigInt(open_sns_token_swap_proposal_id))
       : [],
-  init: nonNullish(init) ? convertSwapInitParams(init) : [],
+  init: convertSwapInitParams(init),
   params: isNullish(params) ? [] : [convertSwapParams(params)],
 });
 
--- a/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
+++ b/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
@@ -94,7 +94,9 @@ export const convertNervousFuncttion = ({
   id: BigInt(id),
   name: name,
   description: toNullable(description),
-  function_type: toNullable(convertFunctionType(function_type)),
+  function_type: nonNullish(function_type)
+    ? toNullable(convertFunctionType(function_type))
+    : [],
 });
 
 const convertSwapInitParams = (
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
```diff
@@ -223,7 +225,7 @@ const convertSwap = ({
     open_sns_token_swap_proposal_id !== undefined
       ? toNullable(convertOptionalNumToBigInt(open_sns_token_swap_proposal_id))
       : [],
-  init: convertSwapInitParams(init),
+  init: nonNullish(init) ? convertSwapInitParams(init) : [],
```
dskloetd, line 228: Looking at `convertSwapInitParams`, it already has logic to check if `init` is nullish and to return `[SnsSwapInit] | []`.

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
```diff
@@ -349,21 +351,29 @@ const convertDtoToTokenMetadata = (
...
   swap: CachedSnsSwapDto
-): SnsSummarySwap => ({
-  ...convertSwap(swap),
-  decentralization_sale_open_timestamp_seconds: convertOptionalNumToBigInt(
-    swap.decentralization_sale_open_timestamp_seconds
-  ),
-  params: convertSwapParams(swap.params),
-});
+): SnsSummarySwap | undefined => {
+  if (isNullish(swap.params)) {
```
dskloetd, line 355: Is it not possible to still convert the rest of `swap` if only `swap.params` are `null`?

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
```diff
@@ -377,5 +378,57 @@ describe("sns aggregator converters utils", () => {
       };
       expect(convertDtoToSnsSummary(aggregatorMissingMetadata)).toBeUndefined();
     });
+
+    it("returns undefined if a swap params required field is missing", () => {
+      const aggregatorMissingMetadata: CachedSnsDto = {
```
dskloetd, line 383: Did you mean `aggregatorMissingSwapParams` instead of `aggregatorMissingMetadata`?

lmuntaner: yes, sorry, I copied from above. Good catch!

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
lmuntaner: Good point! TS was complaining because even though checking for nullish, the type of the function argument doesn't consider that option. I'll change it.

lmuntaner:

dskloetd (approved):

PR #3169, opened 2023-08-22
GIX-1834: New Visitor Landing Page

lmuntaner:
# Motivation

New landing pages when the user is not logged in.

Remove the current login page and redirect instead to accounts.

# Changes

* Use new PageBanner in SignIn pages.
* New component EmptyCards used in SignIn pages.

# Tests

* Only UI changes.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -28,6 +28,8 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 * Review checkboxes vertical alignment, border contrast on dark mode and remove hover background colors
 * Launchpad proposal requests only Open proposals of the SNS topic.
 * When reusing cached agents, use the current identity instead of the one in the cached agent.
+* New page UI for Accounts, Neurons, Canister and Settings when not logged in.
+* Remove login page and redirect to accounts instead.
 
 #### Deprecated
 #### Removed
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -28,7 +28,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 * Review checkboxes vertical alignment, border contrast on dark mode and remove hover background colors
 * Launchpad proposal requests only Open proposals of the SNS topic.
 * When reusing cached agents, use the current identity instead of the one in the cached agent.
-* New page UI for Accounts, Neurons, Canister and Settings when not logged in.
+* New landing pages for Accounts, Neurons, Canister and Settings when not logged in.
 * Remove login page and redirect to accounts instead.
 
 #### Deprecated
--- a/frontend/src/lib/components/common/EmptyCards.svelte
+++ b/frontend/src/lib/components/common/EmptyCards.svelte
@@ -18,11 +18,6 @@
   <Card disabled>
     <div class="card-item" />
   </Card>
-  <article class="card-dashed">
-    <div class="card-item">
-      <div class="select" />
-    </div>
-  </article>
 </div>
 
 <style lang="scss">
@@ -30,12 +25,12 @@
 
   .card-grid {
     // Only show one card for mobile
-    :global(.card):nth-child(-n + 4) {
+    :global(.card):nth-child(-n + 3) {
       display: none;
     }
 
     @include media.min-width(medium) {
-      :global(.card):nth-child(-n + 4) {
+      :global(.card):nth-child(-n + 3) {
         display: block;
       }
     }
@@ -49,20 +44,5 @@
     align-items: center;
     justify-content: center;
     flex-direction: column;
-  }
-
-  .card-dashed {
-    border: 4px dashed var(--input-background);
-    border-radius: var(--border-radius);
-
-    padding: calc(var(--padding-2x) - var(--card-border-size));
-    .select {
-      width: 40%;
-      height: calc(var(--line-height-standard) * 1rem);
-
-      border-radius: var(--border-radius);
-
-      background-color: var(--input-background);
-    }
   }
 </style>
--- a/frontend/src/lib/components/common/EmptyCards.svelte
+++ b/frontend/src/lib/components/common/EmptyCards.svelte
@@ -18,11 +18,11 @@
   <Card disabled>
     <div class="card-item" />
   </Card>
-  <button class="card" disabled>
+  <article class="card-dashed">
     <div class="card-item">
       <div class="select" />
     </div>
-  </button>
+  </article>
 </div>
 
 <style lang="scss">
@@ -51,12 +51,18 @@
     flex-direction: column;
   }
 
-  .select {
-    width: 40%;
-    height: calc(var(--line-height-standard) * 1rem);
-
+  .card-dashed {
+    border: 4px dashed var(--input-background);
     border-radius: var(--border-radius);
 
-    background-color: var(--card-background);
+    padding: calc(var(--padding-2x) - var(--card-border-size));
+    .select {
+      width: 40%;
+      height: calc(var(--line-height-standard) * 1rem);
+
+      border-radius: var(--border-radius);
+
+      background-color: var(--input-background);
+    }
   }
 </style>
--- a/frontend/src/lib/components/common/EmptyCards.svelte
+++ b/frontend/src/lib/components/common/EmptyCards.svelte
@@ -1,0 +1,62 @@
+<script>
+  import { Card } from "@dfinity/gix-components";
+</script>
+
+<div class="card-grid">
+  <Card disabled>
+    <div class="card-item" />
+  </Card>
+  <Card disabled>
+    <div class="card-item" />
+  </Card>
+  <Card disabled>
+    <div class="card-item" />
+  </Card>
+  <Card disabled>
+    <div class="card-item" />
+  </Card>
+  <Card disabled>
+    <div class="card-item" />
+  </Card>
+  <button class="card" disabled>
+    <div class="card-item">
+      <div class="select" />
+    </div>
+  </button>
+</div>
+
+<style lang="scss">
+  @use "@dfinity/gix-components/dist/styles/mixins/media";
+
+  .card-grid {
+    // Only show one card for mobile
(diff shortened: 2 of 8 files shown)


lmuntaner:
New UI:

Screenshot 2023-08-22 at 10 11 14
Screenshot 2023-08-22 at 10 11 21
Screenshot 2023-08-22 at 10 12 06
Screenshot 2023-08-22 at 10 12 19
Screenshot 2023-08-22 at 10 12 54

lmuntaner:

frontend/src/routes/(login)/+page.svelte
```diff
@@ -1,5 +1,9 @@
 <script lang="ts">
-  import Login from "$lib/pages/Login.svelte";
-</script>
+  import { goto } from "$app/navigation";
+  import { AppPath } from "$lib/constants/routes.constants";
+  import { onMount } from "svelte";
 
-<Login />
+  onMount(() => {
+    goto(AppPath.Accounts);
```
peterpeterparker, line 7: This would push to entries in the browser history. Even if we would do a replace history I don't think this is optimal from a technical and ux perspective.

Spontaneously I would suggest for example to move `routes/(login)/+page.svelte` to `routes/(app)/+page.svelte` and within that page replicate the same code as in `routes/(app)/accounts/+page.svelte`.

There might be better options, just a spontaneous though.

frontend/src/lib/components/common/EmptyCards.svelte
```diff
@@ -0,0 +1,62 @@
+<script>
+  import { Card } from "@dfinity/gix-components";
+</script>
+
+<div class="card-grid">
```
peterpeterparker, line 5: Dumb question: what's these empty cards? Something will be displayed in it?

frontend/src/lib/components/common/EmptyCards.svelte
```diff
@@ -0,0 +1,62 @@
...
+  </button>
+</div>
+
+<style lang="scss">
+  @use "@dfinity/gix-components/dist/styles/mixins/media";
+
+  .card-grid {
+    // Only show one card for mobile
+    :global(.card):nth-child(-n + 4) {
+      display: none;
```
peterpeterparker, line 34: would it be more shiny to use opacity and an animation or a hard display is better?

frontend/src/routes/(login)/+page.svelte
lmuntaner: Ideally, we would do a 300 from the backend, right?

Having the same page in different paths seems confusing for the users.

What is the problem with pushing to browser history?

frontend/src/lib/components/common/EmptyCards.svelte
lmuntaner: It's part of the design: https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=197-2867&mode=design&t=15XhUiPhpYSCjBRZ-0

frontend/src/routes/(login)/+page.svelte
peterpeterparker: Both might need the same `+layout.ts` so maybe we would have to create a new `(subgroup)`.

frontend/src/lib/components/common/EmptyCards.svelte
lmuntaner: I don't think that animating the transition when users change the screen size is worth it.

No opacity still uses them for the flow. The last one is special and needs to be in the second place.

I didn't like the idea of using `global` either.

Shall I use JS to detect the window size and change how many are displayed instead?

frontend/src/routes/(login)/+page.svelte
lmuntaner: Which both?

The `(login)` has no layout now. The redirection UX is fast. There is no glitch.

What's the problem with it being pushed to the browser history?

I'll change it to `repalceState: true`.

peterpeterparker: > Ideally, we would do a 300 from the backend, right?

300?

> What is the problem with pushing to browser history?

If you push then you have two entries in your browser history

If you replace the history, then the browser history is ok but, then it's unperformant because you run code for no valid reason while app's boot time should be as fast as possible.

Having a redirect upfront and not using the root, then maybe more a personal opinion, but feels not state of the art.

> Having the same page in different paths seems confusing for the users.

Do the user know about the different paths? If confusing, then should we rather redirect /accounts/ to root?

frontend/src/lib/components/common/EmptyCards.svelte
peterpeterparker: No strong opinion about js or css, I'm fine with the solution. My question strictly had the animation for topic so if it's fine without for sure it's fine. Thanks for the answer.

frontend/src/lib/components/common/EmptyCards.svelte
peterpeterparker: Ah ok. Interesting.

Side note: two colors of your implementation are not in line with the design

262284982-d75b6b0b-5ad8-4d58-a7dc-f68e41c4fe6e

frontend/src/routes/(login)/+page.svelte
lmuntaner: I like the idea of moving all accounts to `/` and redirecting from `/accounts` to `/`.

Maybe it's good to redirect now and it might be temporary. Mary and Artem are working on a dashboard like where you have all your assets. That dashboard could be added in `/` and then we ditch `/accounts`.

frontend/src/lib/components/common/EmptyCards.svelte
lmuntaner: That's correct. But I used the same element `button` with class `card` that you used for the AddAccountButton.

That's what this represents, so it makes sense that it follows that rather than the design.

peterpeterparker: I would rather follow the design in the case. These cards are basically "skeleton / ghost" cards. By having different colors for the last element it feels different and can be interpreted as a call to action, particularly as it's rendering is sharp in comparison to the cards. i.e. it feels like it's a button card but, the text in the middle is missing somehow you know what I mean?

Of course if you already double checked with Artem or if you rather like it that way, none of my business and then good.

frontend/src/routes/(login)/+page.svelte
lmuntaner: Thanks, I might duplicate the code.

Can you approve https://github.com/dfinity/gix-components/pull/279 so that I can continue with this PR then?

frontend/src/lib/components/common/EmptyCards.svelte
lmuntaner: Thanks for the input. I'll check with Artem.

frontend/src/routes/(login)/+page.svelte
peterpeterparker: Sure done.

lmuntaner: I'll do the duplicated code in another PR, that will be cleaner.

lmuntaner:
@peterpeterparker ready for another review.

frontend/src/routes/(login)/+page.svelte
peterpeterparker: I actually really do not mind if you push everything here I have to say. Or make a PR from this PR?

I'm just thinking, since the login page and routing is tricky, that it would be cool if we deploy a version on testnet before merging anything. You know shit happens, just to be sure it works well even with the sw and deployed on a canister. Or maybe I'm too paranoid?

frontend/src/lib/components/common/EmptyCards.svelte
peterpeterparker: What Artem said? Or you mean you want to merge it as it and iterate if necessary afterwards?

peterpeterparker: Thanks

CHANGELOG-Nns-Dapp.md
```diff
@@ -28,6 +28,8 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 * Review checkboxes vertical alignment, border contrast on dark mode and remove hover background colors
 * Launchpad proposal requests only Open proposals of the SNS topic.
 * When reusing cached agents, use the current identity instead of the one in the cached agent.
+* New page UI for Accounts, Neurons, Canister and Settings when not logged in.
+* Remove login page and redirect to accounts instead.
```
lmuntaner, line 32: I forgot to add it in the other PR...

lmuntaner:
@peterpeterparker ready for a new review

CHANGELOG-Nns-Dapp.md
```diff
@@ -28,6 +28,8 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 * Review checkboxes vertical alignment, border contrast on dark mode and remove hover background colors
 * Launchpad proposal requests only Open proposals of the SNS topic.
 * When reusing cached agents, use the current identity instead of the one in the cached agent.
+* New page UI for Accounts, Neurons, Canister and Settings when not logged in.
```
peterpeterparker, line 31: "New landing page" is maybe less technical that "New page UI"?

peterpeterparker (approved):
LGTM, thanks

CHANGELOG-Nns-Dapp.md
peterpeterparker: @lmuntaner but this won't stay in any case no? i.e. we will not propose a version on mainnet with current redirect right?

CHANGELOG-Nns-Dapp.md
lmuntaner: Yes, good point.

CHANGELOG-Nns-Dapp.md
lmuntaner: My idea was to do so, yes.

lmuntaner: How about this https://github.com/dfinity/nns-dapp/pull/3184 ?

PR #3170, opened 2023-08-22
Rustdoc - fix

bitdivine:
# Motivation
Rust code documentation can be parsed and displayed in a nice, readable way by rustdoc.

The sns aggregator documentation is available in this way, but unfortunately most of the nns-dapp code is inaccessible, making it unnecessarily to read.

# Changes
* Make the nns-dapp code accessible to the nns-dapp lib, which is the entry point for rustdoc.
  * Note: `rustdoc` _does_ have limited support for documentation of binaries, however it is not good.
* Fix rustdoc links
* Test the rustdoc in CI.
* Add two utility scripts to make it easy to generate documentation without retyping a bunch of flags.
  * Note: `rustdoc` flags such as `--open` are passed through the script.
* Performed a quick pass checking spelling and formatting.  E.g. code `should be in backticks`.

# Tests
*  `cargo doc` is now tested in CI.

# Todos

### DIFF
--- a/rs/backend/src/lib.rs
+++ b/rs/backend/src/lib.rs
@@ -1,4 +1,8 @@
 //! Library for <https://nns.ic0.app>
+//!
+//! Note: The types are accessed by the nns-dapp directly from main.rs, not via this library.
+//!
+//! Note: This library is for utilities and tools that need access to the types.
 pub mod accounts_store;
 pub mod arguments;
 pub mod assets;
--- a/rs/backend/src/lib.rs
+++ b/rs/backend/src/lib.rs
@@ -1,1 +1,13 @@
+//! Library for <https://nns.ic0.app>
+pub mod accounts_store;
 pub mod arguments;
+pub mod assets;
+pub mod constants;
+pub mod metrics_encoder;
+pub mod multi_part_transactions_processor;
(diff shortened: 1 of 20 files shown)


rs/backend/src/lib.rs
```diff
@@ -1 +1,13 @@
...
+pub mod accounts_store;
 pub mod arguments;
+pub mod assets;
+pub mod constants;
+pub mod metrics_encoder;
+pub mod multi_part_transactions_processor;
+pub mod perf;
+pub mod state;
+pub mod stats;
+pub mod time;
```
dskloetd, line 11: What are these changes for?
They don't seem to be documentation changes.

dskloetd: Thanks for clarifying.
It would be good to have a comment explaining that.
But there actually is a comment at the top saying:
```
//! Library for <https://nns.ic0.app>
```
which suggests rather that the library is use for the WASM.

bitdivine: Added a comment.  I left the title as is, as I think the expanded title is no clearer.  Those who are interested in details will read the comment below it.

dskloetd (approved):

PR #3171, opened 2023-08-22
IdentityAgentWrapper to use cached agent with specified identity

dskloetd:
# Motivation

We cache created agents for performance (according to https://github.com/dfinity/nns-dapp/pull/1600).
But the agent keeps a reference to the identity used to create the agent which means the identity is implicitly cached as well.
This has been causing problems with hardware wallet interactions where we need to keep state on the identity to indicate that a ledger transfer is for the purpose of staking a neuron.

# Changes

1. Add a wrapper agent which uses a newly specified identity instead of the identity used to create the agent originally.
2. When returning an agent from the cache, wrap it in the wrapper such that the current identity is used instead of the previously cached one.

# Tests

1. Method calls are forwarded to the underlying agent.
2. The identity can be invalidated.
3. Method calls on different wrappers result in calls on the same underlying agent with different identities.

# Todos

### DIFF
--- a/frontend/src/tests/lib/api/agent.api.spec.ts
+++ b/frontend/src/tests/lib/api/agent.api.spec.ts
@@ -324,21 +324,15 @@ describe("agent-api", () => {
     const testIdentity2 = createIdentity(testPrincipal1);
 
     const mockAgent = mock<HttpAgent>();
-    let agent1: Agent;
-    let agent2: Agent;
 
     beforeEach(async () => {
       utilsCreateAgentSpy.mockResolvedValue(mockAgent);
-
-      expect(utilsCreateAgentSpy).not.toBeCalled();
-
-      agent1 = await createAgent(testIdentity1);
-      agent2 = await createAgent(testIdentity2);
-
-      expect(utilsCreateAgentSpy).toBeCalledTimes(1);
     });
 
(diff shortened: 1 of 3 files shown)


frontend/src/tests/lib/api/agent.api.spec.ts
```diff
@@ -42,21 +55,388 @@ describe("agent-api", () => {
...
+    const mockAgent = mock<HttpAgent>();
+    let agent1: Agent;
+    let agent2: Agent;
+
+    beforeEach(async () => {
+      utilsCreateAgentSpy.mockResolvedValue(mockAgent);
+
+      expect(utilsCreateAgentSpy).not.toBeCalled();
+
+      agent1 = await createAgent(testIdentity1);
```
lmuntaner, line 335: Personally, I find more readable to create these in each test case.

But not a strong opinion.

Otherwise, I find myself wondering, what is the variable and where does it come from?

lmuntaner (approved):
This looks pretty good to me!

I wouldn't wait for me testing with different entities in the code.

At the moment, everything works. We will just improve by using different identites.

frontend/src/tests/lib/api/agent.api.spec.ts
dskloetd: Done

PR #3173, opened 2023-08-22
Fix ordered dependence in html.utils.spec.ts

dskloetd:
# Motivation

Test `"should call markedjs/marked with custom renderers"` literally depended on the test before it calling `markdownToHTML()`.

# Changes

1. Change `beforeAll` to `beforeEach` to give each test a clean slate.
2. Reset `renderer` in `beforeEach`.
3. Call `markdownToHTML()` in both tests rather than one test relying on the other.

# Tests

Passed with 5 different randomize seeds.

# Todos

Already covered


lmuntaner (approved):
Thanks!

PR #3174, opened 2023-08-22
Fix order dependence of layout.spec.ts

dskloetd:
# Motivation

The test expecting a call to `toastsStore.reset` didn't wait long enough for the call to have happened but it passed because the call also happened in an earlier test.

# Changes

1. Use `beforeEach` instead of `beforeAll` and `afterAll`.
2. Remove `jest.clearAllMock` because `jest.restoreAllMocks` does the same and more.
3. Expect calls not to have happened yet at the start of the test.
4. Use `runResolvedPromises` to make sure all calls that will happen, have happened.

# Tests

Each test passes individually.

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3177, opened 2023-08-23
Don't use `get_mut()` for accounts.

bitdivine:
# Motivation
When we change the accounts data storage, accounts will no longer be mutable memory on the heap.  They will have to be pulled out of storage, parsed , edited and then written back.  That means that we will no longer be able to use `get_mut()` to edit accounts in-place.

The preferred method for making changes in the accounts db API is called `db_try_with_account()` however basic getters and setters are available to both the old HashMap and to the new database, so when we transition, using only getters and setters lets us translate the code literally, so keeps the transition simple.  Once the new API is in use, the getters and setters can be replaced.

# Changes
* Replace every `accounts.get_mut` with `accounts.get()` followed by an insert if the data is changed.

# Tests
* There is no new or altered functionality so existing tests _should_ suffice.

# Todos


dskloetd (approved):
This PR is fine but I have some questions for the future.

Is the account definitely changed in all the cases where we insert it back into `accounts`? Or are there cases where it's only potentially changed? When writing becomes more expensive, does it make sense to make sure we only do it if the account has definitely changed?

And if using getters and setters works and is simpler, why would we ever want to use `db_try_with_account()` instead?

PR #3178, opened 2023-08-23
Fix order dependence in vote-registration.services.spec.ts

dskloetd:
# Motivation

* `"should update store with a new vote registration"` requires `voteRegistrationStore` to be reset but depending on the order in which tests were run, it wasn't guaranteed to be reset.
* `"should show error.register_vote_unknown on not nns-js-based error"` required a mock implementation for `neuronsServices.listNeurons` but depending on the order, it wasn't guaranteed to have a mock implementation.
* `"should call the api to register multiple votes"` expected 3 calls to `governanceApi.registerVote` but didn't clear the mock before the test so it could see calls from other tests.

# Changes

1. Put all cleanup in a top-level `beforeEach`.
2. Put setup that isn't specific to any test in the top-level `beforeEach` as well.
3. Put cleanup before setup to make sure none of the setup is removed by cleanup.
4. Do `jest.clearAllMocks` instead of clearing individual mocks to make sure all mocks are cleared before each test.
5. When expecting calls on mocks, expect the calls not to have happened yet earlier in the test.
6. Remove all `beforeAll`, `afterAll`, and `afterEach` in favor of `beforeEach`.

# Tests

Ran the test with 20 different randomized orderings.

# Todos

covered by existing entry

### DIFF
--- a/frontend/src/tests/lib/services/vote-registration.services.spec.ts
+++ b/frontend/src/tests/lib/services/vote-registration.services.spec.ts
@@ -3,7 +3,6 @@
  */
 
 import * as governanceApi from "$lib/api/governance.api";
-import * as proposalsApi from "$lib/api/proposals.api";
 import { OWN_CANISTER_ID } from "$lib/constants/canister-ids.constants";
 import * as authServices from "$lib/services/auth.services";
 import * as neuronsServices from "$lib/services/neurons.services";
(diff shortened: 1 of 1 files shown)


frontend/src/tests/lib/services/vote-registration.services.spec.ts
```diff
@@ -126,32 +109,6 @@ describe("vote-registration-services", () => {
...
-        jest
-          .spyOn(neuronsServices, "listNeurons")
-          .mockImplementation(() => Promise.resolve());
-        jest
-          .spyOn(proposalsApi, "queryProposal")
-          .mockImplementation(() => Promise.resolve(proposal));
-
-        jest
-          .spyOn(governanceApi, "registerVote")
-          .mockImplementation(() => waitForMilliseconds(10));
```
dskloetd, line 139: Was there a specific reason to wait 10 ms?

mstrasinskis (approved):
Thanks!

PR #3180, opened 2023-08-23
Fix orderer dependency in ckbtc.utils.spec.ts

dskloetd:
# Motivation

One of the tests sets `mockedConstants.FORCE_CALL_STRATEGY = "query"` but doesn't reset it.

# Changes

Call `resetMockedConstants()` in `beforeEach`.

# Tests

passed

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3184, opened 2023-08-23
Fix: Render accoungs landing page in login

lmuntaner:
# Motivation

Render the accounts landing page in `/`.

# Changes

* Render the same components as in `/accounts` when user is not logged in in `(login)/+page.svelte`.

# Tests

* Test that it renders the proper UI.

# Todos

### DIFF
--- a/frontend/src/lib/pages/SignInAccounts.svelte
+++ b/frontend/src/lib/pages/SignInAccounts.svelte
@@ -11,7 +11,7 @@
   // TODO(GIX-1071): this static pre-rendering component should be improved with some more information and shiny design
 </script>
 
-<main class="sign-in">
+<main class="sign-in" data-tid="accounts-landing-page">
   <h1>{$i18n.auth_accounts.title}</h1>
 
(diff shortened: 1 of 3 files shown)


lmuntaner:

peterpeterparker (changes requested):
Just blocking the PR for making it really obvious.

I don't think this PR would work.

- There are few components that would be missing in the layout when the accounts would be loaded from route (`Busy` and `Toasts`)
- Root will always load Nns regardless if root is access with a Sns or ckBTC query param

Instead of this PR I propose following PR, WDYT @lmuntaner ?

👉 https://github.com/dfinity/nns-dapp/pull/3187

Btw. I also noticed a UI glitch following the introduction of the new landing pages, so I also propose a PR to improve that:

👉 https://github.com/dfinity/nns-dapp/pull/3186

lmuntaner:
I agree, closing this one in favor of https://github.com/dfinity/nns-dapp/pull/3187

PR #3187, opened 2023-08-23
Render accounts on root "/" page

peterpeterparker:
# Motivation

We noticed that using a `goto(accounts, replace)` on testnet leads to slowing down the boot time of the dapp.
While we can display a spinner while loading, this would still feel like "being slow".

Therefore and instead, we are proposing to render the accounts at the root of the dapp `/` as on the route `/accounts`.

# Changes

Routing and layout:

- create a new subgroup `(app)/(u)/(accounts)`
- move `(app)/(u)/accounts/+layout` to `(app)/(u)/(accounts)/+layout` that way both root and /accounts share the custom layout and still inherit all other layout
- move and copy `(app)/(u)/accounts/+page` to `(app)/(u)/(accounts)/+page` and `(app)/(u)/(accounts)/accounts/+page`

Changes:

- remove enum `AppPath.Authentication`
- use `AppPath.Accounts` as root


lmuntaner (approved):
LGTM!

PR #3188, opened 2023-08-23
Accounts database trait

bitdivine:
# Motivation
We would like to migrate account storage.  The migration plan calls for a consistent way of accessing accounts, regardless of storage.

# Changes
- Define an account database interface as a Rust trait.
- Define a mock implementation of the trait.
- Define generic test functions that can be run against any implementation of the account storage trait.

# Tests
- I have added Rust tests of the mock implementation.

# Todos


rs/backend/src/accounts_store/schema.rs
dskloetd, line 1: Do you expect to have multiple types of storage for a period of time during migrations?
Will accounts have to be written to both types of storage during this time?
Will that be handled inside one such a DB implementation or will the caller have to manage writing to multiple DB implementations?

dskloetd (approved):

rs/backend/src/accounts_store/schema.rs
bitdivine: The caller has just one db to work with.  That DB is a virtual DB, it delegates to the actual storage mechanism.  During a transition, that DB has two DBs and handles reads and writes to them.

bitdivine: Closing so that I can merge but happy to discuss further.

dskloetd: Thanks

PR #3189, opened 2023-08-23
Fix order dependency in BitcoinEstimatedFee.spec.ts

dskloetd:
# Motivation

One test was expecting a specific number of calls to a mock but the mock was not being cleared between tests.

# Changes

1. Call `jest.clearAllMocks()` in `beforeEach`.
2. Expect no calls at the start of the test.

# Tests

pass

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3190, opened 2023-08-23
Static loading spinner

peterpeterparker:
# Motivation

We have no control on the way the chunks are loaded. As a result, it's possible that it waits until most chunks are loaded before displaying the hints - the spinner - that the dapp is booting which is leading to an unfriendly UX as a blank screen might be displaying for a while.

By rendering a spinner within the static html page and removing it once most of the JS files are loaded, we can overcome this "race condition".

# Changes

- add (copy) spinner in `app.html`
- when the auth is ready, which indicates most JS resources are loaded, remove the spinner

# Tests

https://kgqse-raaaa-aaaaa-aactq-cai.nnsdapp.dfinity.network/
https://kgqse-raaaa-aaaaa-aactq-cai.nnsdapp.dfinity.network/accounts

https://xnjld-hqaaa-aaaal-qb56q-cai.icp0.io/

# Screenshots

Mainnet - current main - Fast 3g network => spinner appears for a milli second when all js is loaded

mainnet-no-spinner

Mainnet - this PR - Fast 3g network

mainnet-spinner

Mainnet this PR - full speed network

full-speed

### DIFF
--- a/frontend/src/routes/(app)/+layout.svelte
+++ b/frontend/src/routes/(app)/+layout.svelte
@@ -15,6 +15,8 @@
   import { referrerPathStore } from "$lib/stores/routes.store";
   import { referrerPathForNav } from "$lib/utils/page.utils";
   import { authStore } from "$lib/stores/auth.store";
+  import { isNullish } from "@dfinity/utils";
+  import {browser} from "$app/environment";
 
   onMount(async () => await Promise.all([initAppAuth(), initAppPublicData()]));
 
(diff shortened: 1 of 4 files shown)


frontend/src/routes/(app)/+layout.svelte
```diff
@@ -29,12 +30,25 @@
...
   );
+
+  // To improve the UX while the app is loading on mainnet we display a spinner which is attached statically in the index.html files.
+  // Once the authentication has been initialized we know most JavaScript resources has been loaded and therefore we can hide the spinner, the loading information.
+  $: (() => {
+    if (!browser) {
+      return;
+    }
+
+    if ($authStore === undefined) {
```
lmuntaner, line 41: Do we need to check for this? Or can we remove it when this component loads anyway?

lmuntaner (approved):
Already approving though.

frontend/src/routes/(app)/+layout.svelte
peterpeterparker: Good idea, let me check

peterpeterparker: I added a comment. Actually we need this check because we want to preserve the current feature of displaying a spinner until the authentication is loaded.

Spinner -> Either landing page or accounts

If we remove it then it could potentially be blank for few milli seconds

Spinner -> briefly blank -> Either landing page or accounts

PR #3191, opened 2023-08-24
Fix: Safari page banner

lmuntaner:
# Motivation

PageBanner in Safari doesn't render properly.

In this PR, it fixes the overflow.

In another PR: upgrade gix-components and svg should be visible.

# Changes

* Wrap the PageBanner within a div. Having a grid inside a flexbox doesn't work well for Safari.

# Tests

Only UI changes.

# Todos

Already covered by the changelog entry of the new landing pages. They have not yet been released.

### DIFF
--- a/frontend/src/lib/pages/SignInAccounts.svelte
+++ b/frontend/src/lib/pages/SignInAccounts.svelte
@@ -9,27 +9,14 @@
     title: $i18n.wallet.title,
     header: "",
   });
-
-  let innerWidth = 0;
-
-  let size = 86;
-
-  const onWindowSizeChange = (innerWidth: number) => {
-    // Change size of the icon
-    size = innerWidth > 768 ? 144 : 86;
-  };
-
-  $: onWindowSizeChange(innerWidth);
 </script>
-
-<svelte:window bind:innerWidth />
 
 <main class="sign-in" data-tid="accounts-landing-page">
   <!-- Safari doesn't handle well grid inside flexbox -->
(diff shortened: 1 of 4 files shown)


lmuntaner:

frontend/src/lib/pages/SignInAccounts.svelte
```diff
@@ -12,12 +12,17 @@
...
 <main class="sign-in" data-tid="accounts-landing-page">
-  <PageBanner>
-    <IconAccountsPage slot="image" />
-    <svelte:fragment slot="title">{$i18n.auth_accounts.title}</svelte:fragment>
-    <p class="description" slot="description">{$i18n.auth_accounts.text}</p>
-    <SignIn slot="actions" />
-  </PageBanner>
+  <!-- Safari doesn't handle well grid inside flexbox -->
+  <!-- https://stackoverflow.com/questions/62075401/safari-grid-in-flexbox-produces-height-overflow -->
+  <div>
```
peterpeterparker, line 17: Shouldn't we add that `div` within the `PageBanner` in gix-cmp to avoid duplicating code?

If yes, we can for sure do that in other PRs.

peterpeterparker (approved):
LGTM, just one question

frontend/src/lib/pages/SignInAccounts.svelte
lmuntaner: Mhm, it's kind of ugly, but this is also ugly. Sure, it make it makes sense. I'll do it, but still merge this PR to not block the RC more.

PR #3192, opened 2023-08-24
Fix order dependence in HardwareWalletName.spec.ts

dskloetd:
# Motivation

One of the tests updates global state in `addAccountStoreMock` and a test (that normally runs before it) expects the global state to be empty.
The global state was only set in `beforeAll`.

# Changes

2. Remove `afterAll`.

# Tests

pass

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3194, opened 2023-08-24
Fix order dependence in NnsSelectAccount.spec.ts

dskloetd:
# Motivation

Some tests mocked `icpAccountsStore` while other tests tried to set the real `icpAccountsStore`.
Setting the real `icpAccountsStore` had no effect for tests that tried to do that and those test only passed by coincidence of what the last test that mocked `icpAccountsStore` happened to mock it with.

# Changes

Stop mocking `icpAccountsStore` in `frontend/src/tests/lib/components/accounts/NnsSelectAccount.spec.ts` and instead set the real store in each of the tests.

# Tests

Pass with 20 different random seeds.

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3196, opened 2023-08-24
Fix order dependence in NnsProposal.spec.ts

dskloetd:
# Motivation

One test sets `referrerPathStore` but it's never reset, which can affect other tests depending on the order.

# Changes

Reset `referrerPathStore` in `beforeEach`.

# Tests

Pass with 20 different random seeds.

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3198, opened 2023-08-24
Fix order dependence in Summary.spec.ts

dskloetd:
# Motivation

Some tests used `page.mock` and some tests relied on the initial value of the `page` store without resetting it.
In fact, there was no way provided to reset it.

# Changes

1. Add `reset` method to mock `page` store.
2. In `frontend/src/tests/lib/components/summary/Summary.spec.ts`, call `page.reset()` in `beforeEach`.
3. In individual tests, change `beforeAll` to `beforeEach` because the top-level `beforeEach` is called after the sub-level `beforeAll`.
4. Remove `afterAll`s and instead fo `jest.clearAllMocks` in top-level `beforeEach`. (drive-by)

# Tests

Pass with random seed 1-20.

# Todos

covered?


lmuntaner (approved):
Thanks!

PR #3199, opened 2023-08-24
Fix order denpendence in SelectUniverseDropdown.spec.ts

dskloetd:
# Motivation

One test relies on a clean `snsAccountsStore` but another test sets data on `snsAccountsStore` and it wasn't reset between tests.

# Changes

1. Reset `snsAccountsStore` in top-level `beforeEach`.
2. Change sub-level `beforeAll` to `beforeEach` to make sure it's run after the top-level `beforeEach`.
3. Move other cleanup from `beforeAll` and `afterAll` to `beforeEach`. (drive-by)

# Tests

Pass with random seeds 1-20.

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3200, opened 2023-08-24
Fix: Ordering test HardwareWalletNeuronAddHotkeyModal

lmuntaner:
# Motivation

Fix test HardwareWalletNeuronAddHotkeyModal.spec that fails when randomized.

# Changes

* Resetting all mocks in a beforeEach fixed the problem and cleaned up the code.

# Tests

Only test changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/tests/lib/modals/accounts/HardwareWalletNeuronAddHotkeyModal.spec.ts
+++ b/frontend/src/tests/lib/modals/accounts/HardwareWalletNeuronAddHotkeyModal.spec.ts
@@ -24,26 +24,26 @@ jest.mock("$lib/proxy/icp-ledger.services.proxy");
 
 describe("HardwareWalletNeuronAddHotkeyModal", () => {
   const props = { testComponent: HardwareWalletNeuronAddHotkeyModal };
+  let spyAddHotkey;
+  let spyGetNeuron;
 
-  const spyAddHotkey = jest
-    .spyOn(api, "addHotkey")
-    .mockImplementation(() => Promise.resolve());
-
-  const spyGetNeuron = jest
-    .spyOn(api, "queryNeuron")
-    .mockImplementation(() => Promise.resolve(mockNeuron));
-
-  jest.spyOn(authStore, "subscribe").mockImplementation(mockAuthStoreSubscribe);
-
-  beforeAll(() => {
+  beforeEach(() => {
     (getLedgerIdentityProxy as jest.Mock).mockImplementation(() =>
       Promise.resolve(mockIdentity)
     );
-  });
-
-  afterAll(() => {
     jest.clearAllMocks();
     jest.restoreAllMocks();
+    spyAddHotkey = jest
+      .spyOn(api, "addHotkey")
+      .mockImplementation(() => Promise.resolve());
+
+    spyGetNeuron = jest
+      .spyOn(api, "queryNeuron")
+      .mockImplementation(() => Promise.resolve(mockNeuron));
+
+    jest
+      .spyOn(authStore, "subscribe")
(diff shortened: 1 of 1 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/modals/accounts/HardwareWalletNeuronAddHotkeyModal.spec.ts
```diff
@@ -24,26 +24,26 @@ jest.mock("$lib/proxy/icp-ledger.services.proxy");
...
-  beforeAll(() => {
+  beforeEach(() => {
     (getLedgerIdentityProxy as jest.Mock).mockImplementation(() =>
       Promise.resolve(mockIdentity)
     );
-  });
-
-  afterAll(() => {
     jest.clearAllMocks();
     jest.restoreAllMocks();
```
dskloetd, line 35: Can you put these 2 lines at the top so the cleanup comes before the setup?

Currently the `restoreAllMocks` undoes the `getLedgerIdentityProxy` setup on lines 31-33 so either that's not necessary or it actually broke the tests and they no longer test what they should test.

Also `restoreAllMocks` is a stronger version of `clearAllMocks` so you don't need both.

frontend/src/tests/lib/modals/accounts/HardwareWalletNeuronAddHotkeyModal.spec.ts
```diff
@@ -24,26 +24,26 @@ jest.mock("$lib/proxy/icp-ledger.services.proxy");
 
 describe("HardwareWalletNeuronAddHotkeyModal", () => {
   const props = { testComponent: HardwareWalletNeuronAddHotkeyModal };
+  let spyAddHotkey;
+  let spyGetNeuron;
```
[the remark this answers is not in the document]
lmuntaner: Ah, we always did it the other way. I don't have a strong opinion.

frontend/src/tests/lib/modals/accounts/HardwareWalletNeuronAddHotkeyModal.spec.ts
lmuntaner: True, weird that it works then... I'l check it out.

lmuntaner: This mock was not necessary because it was not being tested.

Yet, I added a check and changed to mock to return another identity. This way the test makes sure the hotkey is added properly.

I also added a call to `resetIdentity` to make it explicit that the mockIdentity is the logged in user.

lmuntaner:

frontend/src/tests/lib/modals/accounts/HardwareWalletNeuronAddHotkeyModal.spec.ts
```diff
@@ -103,13 +107,30 @@ describe("HardwareWalletNeuronAddHotkeyModal", () => {
...
+        neuronId: mockNeuronStake.neuronId,
+        principal: mockIdentity.getPrincipal(),
+        // It uses the ledger identity to add the hotkey
+        identity: mockIdentity2,
+      })
+    );
+    await waitFor(() =>
+      expect(spyGetNeuron).toBeCalledWith({
+        neuronId: mockNeuronStake.neuronId,
+        // Onece the hotkey is added, it uses the current identity to query the neuron
```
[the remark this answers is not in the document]
lmuntaner: fixed, thanks!

PR #3201, opened 2023-08-24
Fix order dependence in Warnings.spec.ts

dskloetd:
# Motivation

`metricsCallback` is initialized while rendering the component.
One of the tests explicitly waited for this but the others didn't and may have reused the value from the other test.

# Changes

1. Reset `metricsCallback` before each test.
2. Wait for `metricsCallback` to be initialized in each test.

# Tests

Pass with random seed 1 - 20.

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3202, opened 2023-08-24
Fix order dependence in selectable-universes.derived.spec.ts

dskloetd:
# Motivation

One test mocked `snsProjectsCommittedStore` but other tests expected to be empty.

# Changes

1. Restore mocks in top-level `beforeEach`.
2. Change `beforeAll` to `beforeEach` so it runs after the top-level `beforeEach`.
3. Remove `clearAllMocks` which became redundant with `restoreAllMocks`.

# Tests

Passes with random seed 1 - 20.

# Todos

covered

### DIFF
--- a/frontend/src/tests/lib/derived/selectable-universes.derived.spec.ts
+++ b/frontend/src/tests/lib/derived/selectable-universes.derived.spec.ts
@@ -12,13 +12,13 @@ import { get } from "svelte/store";
 
 describe("selectable universes derived stores", () => {
   beforeEach(() => {
+    jest.restoreAllMocks();
+
     page.mock({
       routeId: AppPath.Accounts,
       data: { universe: OWN_CANISTER_ID.toText() },
     });
   });
-
-  afterAll(() => jest.clearAllMocks());
 
   it("should return Nns, ckBTC and ckTESTBTC (flag for test is true) per default", () => {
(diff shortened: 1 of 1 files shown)


frontend/src/tests/lib/derived/selectable-universes.derived.spec.ts
```diff
@@ -44,13 +44,11 @@ describe("selectable universes derived stores", () => {
   });
 
   describe("with projects", () => {
-    beforeAll(() =>
+    beforeEach(() => {
       jest
```
lmuntaner, line 48: How does this fix the issue?

Shouldn't the mock be used only where is needed and if the other required it to be empty, we should explicitly reset that store?

Sorry, but I don't understand the explanation with the changes.

dskloetd: The store is not filled, but mocked.
So the way to reset it is with `restoreAllMocks` (line 15).

But if that was the only thing I changed, the `restoreAllMocks()` would be called after the store is mocked here, because `beforeEach` is run after `beforeAll`. So I needed to also change `beforeAll` to `beforeEach`, which is better anyway.

Does it make sense now?

lmuntaner (approved):

PR #3203, opened 2023-08-24
Fix order dependence in intersection.directives.spec.ts

dskloetd:
# Motivation

There are 2 tests: one tests with `IsIntersecting = true` and the other with `IsIntersecting = false`.
But `IsIntersecting = true` is assumed implicitly as default value.

# Changes

Set `IsIntersecting` explicitly in both tests.

# Tests

Tests now pass in both orders.

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3206, opened 2023-08-24
Bump ic-js and agent-js

peterpeterparker:
# Motivation

See PR https://github.com/dfinity/ic-js/pull/397 for the reason behind bumping agent-js and de facto ic-js.


mstrasinskis (approved):
Thanks!

PR #3207, opened 2023-08-24
Prevent deep linking ckTESTBTC with only ckBTC enabled

dskloetd:
# Motivation

I was looking at the test because its success depended on the order in which tests were run.
But then I realized the test could actually fail while it shouldn't because `ENABLE_CKTESTBTC` is `true`.
Then I realized that the code that checks if the ckBTC or ckTESTBTC universe should be enabled doesn't distinguish between the two feature flags.

# Changes

1.  Instead of allowing both universes if either flag is enabled, only allow each universe if its flag is enabled.
2. In `beforeEach` reset `overrideFeatureFlagsStore` to make sure each test has a clean slate.
3. Test that ckBTC is not available with `ENABLE_CKTESTBTC` enabled.
4. Corresponding tests for ckTESTBTC.

# Tests

Manually tested that the ckBTC universe can still be selected when the feature flag is enabled.

# Todos

not worth it


frontend/src/lib/derived/selected-universe.derived.ts
```diff
@@ -48,11 +52,23 @@ export const selectedUniverseIdStore: Readable<Principal> = derived<
...
+      return OWN_CANISTER_ID;
+    }
+    if (
+      canisterId.toText() === CKBTC_UNIVERSE_CANISTER_ID.toText() &&
+      !$ENABLE_CKBTC
+    ) {
+      return OWN_CANISTER_ID;
+    }
+    if (
+      canisterId.toText() === CKTESTBTC_UNIVERSE_CANISTER_ID.toText() &&
```
peterpeterparker, line 65: You can use `isUniverseCkTESTBTC()`

frontend/src/lib/derived/selected-universe.derived.ts
```diff
@@ -48,11 +52,23 @@ export const selectedUniverseIdStore: Readable<Principal> = derived<
   [pageUniverseIdStore, pageStore, ENABLE_CKBTC, ENABLE_CKTESTBTC],
   ([canisterId, page, $ENABLE_CKBTC, $ENABLE_CKTESTBTC]) => {
     // ckBTC is only available on Accounts therefore we fallback to Nns if selected and user switch to another view
-    if (($ENABLE_CKBTC || $ENABLE_CKTESTBTC) && pathSupportsCkBTC(page)) {
-      return canisterId;
+    if (isUniverseCkBTC(canisterId) && !pathSupportsCkBTC(page)) {
+      return OWN_CANISTER_ID;
+    }
+    if (
+      canisterId.toText() === CKBTC_UNIVERSE_CANISTER_ID.toText() &&
```
peterpeterparker, line 59: You can use `isUniverseCkBTC() && !isUniverseCkTESTBTC()`

frontend/src/lib/derived/selected-universe.derived.ts
dskloetd: Done

frontend/src/lib/derived/selected-universe.derived.ts
dskloetd: I find it confusing that `isUniverseCkBTC()` also returns `true` for `CKTESTBTC_UNIVERSE_CANISTER_ID` and I think it's clearer to compare to a single ID that combine these 2 checks, one of them negative.

peterpeterparker (approved):
LGTM

PR #3214, opened 2023-08-25
Fix order dependence in Settings.spec.ts

dskloetd:
# Motivation

`"should render a skeleton while loading expired session time"` expects `authRemainingTimeStore` to start out empty but it's never reset when other tests fill it.

# Changes

Reset `authRemainingTimeStore` in top-level `beforeEach`.

# Tests

Pass with random seeds 1 - 20.

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3215, opened 2023-08-25
Fix order dependence in ProposalDetail.spec.ts

dskloetd:
# Motivation

When tests run in a different order, both `beforeAll`s are run before either test and then both tests are run.

# Changes

2. Remove `afterEach`.

# Tests

The two tests pass in both orders.

# Todos

covered


lmuntaner (approved):
Thanks!

PR #3216, opened 2023-08-25
 Fix order dependence in NnsProposals.spec.ts

dskloetd:
# Motivation

`"should render proposals"` expects `neuronsStore` to be empty.
So if `"should hide proposal card if already voted"`, which mocks `neuronStore`, is run before it, `"should render proposals"` fails.

`neuronStore` is already reset in `beforeEach` but resetting the real store doesn't help once the store is mocked.
So we either need to restore the mock, or use the real store instead of mocking it.
Why not both?

Also `"should render proposals"` expects `proposalsFiltersStore` to be empty so it needs to be reset after other tests fill it.

Lastly, the tests in `"log in and out"` expect `authStore` and `GovernanceCanister` to be already mocked, so those setups need to go to the top-level as well.

# Changes

1. Use real `neuronStore` instead of mocking it.
2. Restore all mocks in top-level `beforeEach` to give each test a clean slate.
3. Reset `proposalsFiltersStore` in top-level `beforeEach`.
4. Move mocking of `authStore` and `GovernanceCanister` to top-level `beforeEach`.
5. Create `proposalsFiltersStore.reload` spy inside `beforeEach` so it doesn't get restored before it's used.
6. Remove redundant `jest.clearAllMocks`.

# Tests

Passes with random seed 1 - 20

# Todos

covered

### DIFF
--- a/frontend/src/tests/lib/pages/NnsProposals.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsProposals.spec.ts
@@ -21,10 +21,7 @@ import {
 } from "$tests/mocks/auth.store.mock";
 import { MockGovernanceCanister } from "$tests/mocks/governance.canister.mock";
 import en from "$tests/mocks/i18n.mock";
-import {
-  buildMockNeuronsStoreSubscribe,
-  mockNeuron,
-} from "$tests/mocks/neurons.mock";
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!!

PR #3217, opened 2023-08-25
Fix order dependence in NnsWallet.spec.ts

dskloetd:
# Motivation

`"should display receive modal information"` expects `icpAccountsStore` to be filled with certified data to avoid calling `queryAccount` via `pollAccounts`.
The certified test data was put in `icpAccountsStore` in `beforeAll` rather than `beforeEach`.
If `"should reload account after finish receiving tokens"` was run first, it would end up with non-certified data in `icpAccountsStore` and so `"should display receive modal information"` would fail when doing `queryAccount` which wasn't mocked.

The solution is to put the certified data in `icpAccountsStore` in `beforeEach` so that each test gets a fresh store.

# Changes

2. Move all cleanup to top-level `beforeEach` (drive-by)

# Tests

Passes with random seeds 1 - 20

# Todos

covered

### DIFF
--- a/frontend/src/tests/lib/pages/NnsWallet.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsWallet.spec.ts
@@ -55,6 +55,10 @@ describe("NnsWallet", () => {
 
   beforeEach(() => {
     jest.clearAllMocks();
+    jest.clearAllTimers();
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!

PR #3219, opened 2023-08-28
Rename MockAccountsDb to AccountsDbAsMap

bitdivine:
# Motivation
The first production implementation of the accounts DB trait will be the status quo: Accounts are stored in a hash map.

We actually already have an implementation of the accounts db trait as a map in the mock, so we can just rename it and move it out of the tests file.

# Changes
* Rename `MockAccountsDb` to `AccountsDbAsMap`.

# Tests
* The existing tests have been updated.

# Todos

  - Not needed, as this is minor and doesn't make any user-facing changes.

### DIFF
--- a/rs/backend/src/accounts_store/schema/tests.rs
+++ b/rs/backend/src/accounts_store/schema/tests.rs
@@ -2,34 +2,10 @@
 
 use super::super::{AccountIdentifier, CanisterId, NamedCanister, PrincipalId};
 use super::*;
-use std::collections::{BTreeMap, HashMap};
-
-/// Toy accounts database for testing.
-#[derive(Default)]
-pub struct MockAccountsDb {
-    accounts: BTreeMap<Vec<u8>, Account>,
-}
-
-impl AccountsDbTrait for MockAccountsDb {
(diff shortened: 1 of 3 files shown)


rs/backend/src/accounts_store/schema/tests.rs
```diff
@@ -2,34 +2,10 @@
...
-        self.accounts.get(account_key).cloned()
-    }
-    fn db_remove_account(&mut self, account_key: &[u8]) {
-        self.accounts.remove(account_key);
-    }
-    fn db_accounts_len(&self) -> u64 {
-        self.accounts.len() as u64
-    }
-}
+use std::collections::HashMap;
```
bitdivine, line 6: Note: These functions, below, are to be reused on ever implementation of the accounts db trait, so are kept here.  But as we see they need to be marked `pub`.

mstrasinskis (approved):
Rubber stamp, thanks!

frederikrothenberger (approved):
Not very familiar yet with the NNS dapp code base, but this seems like a sensible refactoring. 👍

PR #3222, opened 2023-08-28
GIX-1841: queryFinalizationStatus api function

lmuntaner:
# Motivation

We want to add a new status in project detail: "Finalizing".

In this PR, add the api function to get the value that will be used to check whether an SNS project is "finalizing".

# Changes

* New error util: `isMethodNotSupportedError`.
* New sns sale api: `queryFinalizationStatus`.

# Tests

* Test the new api function `queryFinalizationStatus`.
* Test new error util `isMethodNotSupportedError`.

# Todos

Not yet necessary. No new functionality has been added.


lmuntaner:

frontend/src/lib/api/sns-sale.api.ts
```diff
@@ -114,3 +116,37 @@ export const notifyParticipation = async ({
...
+  });
+
+  try {
+    const response = await getFinalizationStatus({});
+
+    return response;
+  } catch (err) {
+    // If the method is not available, return undefined
+    if (
+      isMethodNotSupportedError({ err, method: "get_auto_finalization_status" })
```
dskloetd, line 144: Where does this magic string `"get_auto_finalization_status"` come from?
I'm guessing somewhere in `ic-js` we use this to make the call and it needs to be the same?
Can we organize it so we don't need to have this magic string?
Is it important that the string is there? If we get an error that says that some method is not supported, can't we assume it's the one we just called?
Or can the error maybe be detected in `ic-js` and then it throws a properly typed error?

frontend/src/lib/utils/error.utils.ts
```diff
@@ -164,3 +164,28 @@ export const isPayloadSizeError = (err: unknown): boolean => {
...
+export const isMethodNotSupportedError = ({
+  err,
+  method,
+}: {
+  err: unknown;
+  method: string;
+}): boolean => {
+  if (typeof err === "object" && nonNullish(err) && "message" in err) {
+    const message = err.message as string;
+    return message.includes("has no update method") && message.includes(method);
```
dskloetd, line 188: How about this?
```
return message.includes(`has no update method ${method}`);
```

frontend/src/tests/lib/api/sns-sale.api.spec.ts
```diff
@@ -130,4 +136,64 @@ describe("sns-sale.api", () => {
       confirmation_text: [confirmationText],
     });
   });
+
+  describe("queryFinalizationStatus", () => {
+    beforeEach(() => {
+      expect(finalizationStatusSpy).toBeCalledTimes(0);
```
dskloetd, line 142: I think it makes more sense to have this in every test.
The point is that you can look at the test and trust that it does what you think it does.
If you accidentally have `beforeAll` here instead of `beforeEach`, this won't be executed before each test.
But also, I didn't actually see this while reading the tests so I was already asking to add it in the test and only then I noticed this.

frontend/src/lib/api/sns-sale.api.ts
lmuntaner: Mhm, it's an interesting idea to add this at the ic-js level. I think I also like it better.

Let me do it there. That's also where the information about the "magic" string is. So it makes sense to have it there.

frontend/src/tests/lib/api/sns-sale.api.spec.ts
lmuntaner: ok

lmuntaner:

dskloetd (approved):

PR #3224, opened 2023-08-28
test: e2e Filter proposals by Votable only

mstrasinskis:
# Motivation

E2e test to check "Show only proposals you can still vote for" feature.
image

# Changes

- missing test id
- extend POs
- `Filter proposals by Votable only` test

# Tests

Manually by running e2e test.

# Todos


frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -32,6 +40,64 @@ test("Test neuron voting", async ({ page, context }) => {
...
+    ).map((card) => card.getProposalId())
+  );
+
+  expect(createdProposalIds).toContain(proposalIdForVoting);
+
+  // Hide proposals that are not votable
+  await appPo
+    .getProposalsPo()
+    .getNnsProposalFiltersPo()
+    .setVotableProposalsOnlyValue(true);
```
dskloetd, line 83: Minor suggestion:
When you expect the voted proposal to still be visible on line 77, you are implicitly assuming that the checkbox is not yet checked. So you could also just click it instead of explicitly setting it to true. Or if you'd rather not make the implicit assumption, you could either assert the value or explicitly set it to false (and wait for content loaded) before line 68.

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -32,6 +40,64 @@ test("Test neuron voting", async ({ page, context }) => {
...
+    .getProposalDetailPo()
+    .getNnsProposalPo()
+    .getVotingCardPo()
+    .voteYes();
+
+  // Back to proposals list
+  await appPo.goBack();
+  await appPo.getProposalsPo().getNnsProposalListPo().waitForContentLoaded();
+
+  // Get all proposals that are currently votable
```
dskloetd, line 67: You haven't checked the box yet so this also includes proposals that are not votable, doesn't it?

mstrasinskis: I expect it to be unchecked because it is `false` by default (https://github.com/dfinity/nns-dapp/blob/70fce4797f276f7aa5487a4d8c9faba07b5310f5/frontend/src/lib/constants/proposals.constants.ts#L22)

dskloetd: So the comment is wrong, no?

mstrasinskis: Yep, the flow was updated and the comment removed. Thanks!

frontend/src/tests/e2e/proposals.spec.ts
mstrasinskis: The flow was updated a bit to not rely on the default filter values:
1. set "votable only" to be true
2. vote for proposal X
3. check that X is hidden
4. set "votable only" to be false
5. check that X is visible again

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -5,6 +5,22 @@ import { signInWithNewUser, step } from "$tests/utils/e2e.test-utils";
 import { ProposalStatus, Topic } from "@dfinity/nns";
 import { expect, test } from "@playwright/test";
 
+const visibleProposalIds = async ({
```
dskloetd, line 8: I think it's good for functions to have a verb in the name to distinguish with variables.
`visibleProposalIds` sounds like a variable holding an array.
How about `getVisibleProposalIds`?

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -5,6 +5,22 @@ import { signInWithNewUser, step } from "$tests/utils/e2e.test-utils";
...
+  appPo: AppPo;
+  proposerNeuronId: string;
+}): Promise<string[]> =>
+  await Promise.all(
+    (
+      await appPo
+        .getProposalsPo()
+        .getNnsProposalListPo()
+        .getProposalCardPosForProposer(proposerNeuronId)
+    ).map((card) => card.getProposalId())
```
dskloetd, line 21: Actually, how about adding this function to `NnsProposalListPo` instead of here?

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -40,65 +56,67 @@ test("Test neuron voting", async ({ page, context }) => {
...
 
-  step(`Filter proposals by Votable only`);
-
+  step("Filter proposals by Votable only");
+  // Hide proposals that are not votable
+  await appPo
+    .getProposalsPo()
+    .getNnsProposalFiltersPo()
+    .setVotableProposalsOnlyValue(true);
+  await appPo.getProposalsPo().getNnsProposalListPo().waitForContentLoaded();
```
dskloetd, line 65: You could store `appPo.getProposalsPo().getNnsProposalListPo()` in a variable so you don't need to get it every time.

frontend/src/tests/e2e/proposals.spec.ts
mstrasinskis: Agree 👍

frontend/src/tests/e2e/proposals.spec.ts
mstrasinskis: Good point, done!

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -47,26 +31,27 @@ test("Test neuron voting", async ({ page, context }) => {
...
     .setVotableProposalsOnlyValue(true);
-  await appPo.getProposalsPo().getNnsProposalListPo().waitForContentLoaded();
-  const proposalIdsBeforeVoting = await visibleProposalIds({
-    appPo,
-    proposerNeuronId,
-  });
+  await nnsProposalListPo.waitForContentLoaded();
+  const proposalIdsBeforeVoting = await appPo
+    .getProposalsPo()
+    .getNnsProposalListPo()
```
dskloetd, line 53: Replace
```
appPo
    .getProposalsPo()
    .getNnsProposalListPo()
```
with `nnsProposalListPo`?

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -85,11 +70,11 @@ test("Test neuron voting", async ({ page, context }) => {
...
   await appPo.goBack();
-  await appPo.getProposalsPo().getNnsProposalListPo().waitForContentLoaded();
-  const proposalIdsAfterVoting = await visibleProposalIds({
-    appPo,
-    proposerNeuronId,
-  });
+  await nnsProposalListPo.waitForContentLoaded();
+  const proposalIdsAfterVoting = await appPo
+    .getProposalsPo()
+    .getNnsProposalListPo()
```
dskloetd, line 76: same?

frontend/src/tests/e2e/proposals.spec.ts
```diff
@@ -103,26 +88,25 @@ test("Test neuron voting", async ({ page, context }) => {
...
 
   step("Voted proposal should be visible again");
   expect(
-    await visibleProposalIds({
-      appPo,
-      proposerNeuronId,
-    })
+    await appPo
+      .getProposalsPo()
+      .getNnsProposalListPo()
```
dskloetd, line 98: same

frontend/src/tests/e2e/proposals.spec.ts
mstrasinskis: Thanks,done.

dskloetd (approved):

PR #3226, opened 2023-08-28
Remove unused filter from NnsDestinationAddress

dskloetd:
# Motivation

`NnsDestinationAddress` has a prop `filterIdentifier` which is not passed in from anywhere so it's always `undefined`.

# Changes

1. Remove `filterIdentifier` from `NnsDestinationAddress`.
2. Remove the test that tests the filter.

# Tests

Remaining tests still pass.

# Todos

not worth it


peterpeterparker (approved):
Thx

PR #3227, opened 2023-08-28
Improve NnsDestinationAddress.spec.ts

dskloetd:
# Motivation

I want to bring `NnsDestinationAddress`, which is only used in `DisburseNnsNeuronModal`, more in line with the account picker in the transaction modal.
But before I do, I want to make sure there is proper test coverage so I don't break anything.

# Changes

1. Use page objects in `frontend/src/tests/lib/components/accounts/NnsDestinationAddress.spec.ts`.
2. Create spies in `beforeEach` instead of directly in the `describe` block.
3. Add a test where it selects an account from the list.
4. Add a test where it enters an identifier in the input field.
5. Add required functionality to page objects.

# Tests

Only tests

# Todos

not worth it

### DIFF
--- a/frontend/src/tests/lib/components/accounts/NnsDestinationAddress.spec.ts
+++ b/frontend/src/tests/lib/components/accounts/NnsDestinationAddress.spec.ts
@@ -8,6 +8,8 @@ import {
   mockAccountsStoreSubscribe,
   mockSubAccount,
 } from "$tests/mocks/icp-accounts.store.mock";
+import { NnsDestinationAddressPo } from "$tests/page-objects/NnsDestinationAddress.page-object";
+import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
 import { render } from "@testing-library/svelte";
 
 describe("NnsDestinationAddress", () => {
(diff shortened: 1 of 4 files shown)


frontend/src/tests/lib/components/accounts/NnsDestinationAddress.spec.ts
```diff
@@ -16,28 +18,72 @@ describe("NnsDestinationAddress", () => {
...
-    .spyOn(icpAccountsStore, "subscribe")
-    .mockImplementation(
-      mockAccountsStoreSubscribe([mockSubAccount, mockSubAccount2])
-    );
+  let onAccountSelectedSpy: jest.Mock;
 
-  it("should render an input to enter an address", () => {
-    const { container } = render(NnsDestinationAddress);
+  beforeEach(() => {
+    jest
```
lmuntaner, line 24: Why do you rely on mocking the store? Didn't we say that it was better to fill the store?

frontend/src/tests/lib/components/accounts/NnsDestinationAddress.spec.ts
```diff
@@ -16,28 +18,72 @@ describe("NnsDestinationAddress", () => {
...
-  jest
-    .spyOn(icpAccountsStore, "subscribe")
-    .mockImplementation(
-      mockAccountsStoreSubscribe([mockSubAccount, mockSubAccount2])
-    );
+  let onAccountSelectedSpy: jest.Mock;
 
-  it("should render an input to enter an address", () => {
-    const { container } = render(NnsDestinationAddress);
+  beforeEach(() => {
```
lmuntaner, line 23: Don't you need to clearMocks?

lmuntaner (approved):
A few comments.

I thought that you wanted to remove this component and use instead the `SelectDestinationAddress`.

Approving because this adds more testing, but I'm a little bit confused with where this is going.

frontend/src/tests/lib/components/accounts/NnsDestinationAddress.spec.ts
dskloetd: I don't *need* to, because all the mocks are fully specified in the `beforeEach`. But it doesn't hurt either.
Done.

frontend/src/tests/lib/components/accounts/NnsDestinationAddress.spec.ts
dskloetd: I just didn't touch this part of the test. It was already mocked and it didn't affect the test so I didn't decide to change it this time.
But I've changed it now.

dskloetd:
> I thought that you wanted to remove this component and use instead the SelectDestinationAddress.

I think `SelectDestinationAddress` will be used inside this component. Because using `SelectDestinationAddress` requires a lot of overhead and I didn't want to put all of that directly in the DisburseNeuronModal. Maybe it ends up not making sense.

frontend/src/tests/lib/components/accounts/NnsDestinationAddress.spec.ts
lmuntaner: Thanks!

PR #3228, opened 2023-08-29
Scipt to increment the patch version

bitdivine:
# Motivation
The nns dapp release process includes incrementing the patch version of the frontend and cargo versions.  Let's automate that.

# Changes
- Add a script that increments the patch versions.

# Tests
- The effect of running this command can be seen here: https://github.com/dfinity/nns-dapp/compare/bump-patch...test-patch
- A minimal test is included and exercised in CI.

# Todos

- [ ] Create full tests.  I don't want to spend the time now.
- [ ] Use this in the release procedure.  https://www.notion.so/dfinityorg/Checklist-YYYY-MM-DD-0e1db3aadfc140ab9bd9c2f3957ba25d

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -237,6 +237,10 @@ jobs:
         run: sudo apt-get update -yy && sudo apt-get install -yy moreutils && command -v sponge
       - name: Bumping the package versions works
         run: ./scripts/nns-dapp/bump-patch
+      - name: Set git user
+        run: |
+                  git config --global user.email "gix-bot@dfinity.org"
+                  git config --global user.name "GIX bot"
       - name: Command creates commits
         run: |
           original_commit="$(git rev-parse HEAD)"
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -227,6 +227,25 @@ jobs:
       - uses: actions/checkout@v3
       - name: docker-build prints help
         run: ./scripts/docker-build --help | grep -i usage
+  minor-version-bump-works:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install yq
+        run: sudo snap install yq
+      - name: Bumping the package versions works
+        run: ./scripts/nns-dapp/bump-patch
+      - name: Command creates commits
+        run: |
+          original_commit="$(git rev-parse HEAD)"
+          ./scripts/nns-dapp/bump-patch --commit
+          num_commits="$(git log --oneline "$original_commit..HEAD" | wc -l)"
+          (( num_commits == 2 )) || {
+                  echo "Expected two commits.  Got:"
+                  git log --oneline "$original_commit..HEAD"
+                  exit 1
+          } >&2
+          # TODO: Verify that the commits contain the expected changes.
   download-nns-dapp-ci-wasm:
     runs-on: ubuntu-20.04
     steps:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -239,8 +239,8 @@ jobs:
         run: ./scripts/nns-dapp/bump-patch
       - name: Set git user
         run: |
-                  git config --global user.email "gix-bot@dfinity.org"
-                  git config --global user.name "GIX bot"
+          git config --global user.email "gix-bot@dfinity.org"
+          git config --global user.name "GIX bot"
       - name: Command creates commits
         run: |
           original_commit="$(git rev-parse HEAD)"
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -233,6 +233,8 @@ jobs:
       - uses: actions/checkout@v3
       - name: Install yq
         run: sudo snap install yq
+      - name: Install sponge
+        run: sudo apt-get update -yy && sudo apt-get install -yy moreutils && command -v sponge
       - name: Bumping the package versions works
         run: ./scripts/nns-dapp/bump-patch
       - name: Command creates commits
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -236,7 +236,7 @@ jobs:
       - name: Install sponge
         run: sudo apt-get update -yy && sudo apt-get install -yy moreutils && command -v sponge
       - name: Bumping the package versions works
-        run: ./scripts/nns-dapp/bump-patch
+        run: ./scripts/nns-dapp/bump-patch.test
       - name: Set git user
         run: |
           git config --global user.email "gix-bot@dfinity.org"
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -240,6 +240,7 @@ jobs:
       - name: Command creates commits
         run: |
           original_commit="$(git rev-parse HEAD)"
+          git checkout .
           ./scripts/nns-dapp/bump-patch --commit
           num_commits="$(git log --oneline "$original_commit..HEAD" | wc -l)"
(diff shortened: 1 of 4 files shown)


scripts/nns-dapp/bump-patch
```diff
@@ -0,0 +1,41 @@
...
+  echo "Changes:"
+  git status --untracked-files=no --porcelain | head
+  exit 1
+} >&2
+pushd frontend
+npm version patch
+npm install
+test -z "${COMMIT_CHANGES:-}" || git commit -a -m "Increment nns-dapp frontend patch version"
+popd
+awk -F'"' '/^version *= *"[0-9.]+"/{split($2,v,".");v[3]++;printf "version = \"%s.%s.%s\"\n", v[1], v[2], v[3];next}{print}' Cargo.toml | sponge Cargo.toml
```
dskloetd, line 39: Maybe add a comment explaining in words what this tries to do in case it ever needs to be modified?

scripts/nns-dapp/bump-patch.test
dskloetd, line 1: Did you also want to invoke this from the workflow?

scripts/nns-dapp/bump-patch
```diff
@@ -0,0 +1,41 @@
...
+  echo "       Please do so before re-running this command."
+  echo
+  echo "Changes:"
+  git status --untracked-files=no --porcelain | head
+  exit 1
+} >&2
+pushd frontend
+npm version patch
+npm install
+test -z "${COMMIT_CHANGES:-}" || git commit -a -m "Increment nns-dapp frontend patch version"
```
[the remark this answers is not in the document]
bitdivine: > But in the workflow I see it being executed without arguments and it expects commits.

Where do you see this?  I see it only in the step "Command creates commits" and there the `--commit` flag is passed.

.github/workflows/checks.yml
```diff
@@ -227,6 +227,32 @@ jobs:
       - uses: actions/checkout@v3
       - name: docker-build prints help
         run: ./scripts/docker-build --help | grep -i usage
+  minor-version-bump-works:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install yq
+        run: sudo snap install yq
+      - name: Install sponge
```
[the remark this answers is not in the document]
bitdivine: Sponge is in the standard set of tools for the nns-dapp  repo, installed by `./scripts/setup`, so I would stick with the default.  It's definitely worth being able to script that in situations where dependencies are painful, though.

scripts/nns-dapp/bump-patch.test
```diff
@@ -0,0 +1,49 @@
...
+	1	1	Cargo.toml
+	2	2	frontend/package-lock.json
+	1	1	frontend/package.json
+	EOF
+}
+expected_file_changes="$(expected_file_changes)"
+
+(
+  printf "########\n## %s\n" "Verifies that the expected files are changed"
+  git stash # Make sure that git is clean.
```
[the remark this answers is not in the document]
bitdivine: Git stash saves any changes, so it would be hard or impossible for it to be destructive in itself.  But it could confuse the user.  We could require git to be clean before running the test, or notify the user that changes have been stashed.  I'd be fine with either.

scripts/nns-dapp/bump-patch
```diff
@@ -0,0 +1,41 @@
...
+popd
+awk -F'"' '/^version *= *"[0-9.]+"/{split($2,v,".");v[3]++;printf "version = \"%s.%s.%s\"\n", v[1], v[2], v[3];next}{print}' Cargo.toml | sponge Cargo.toml
+cargo check # Updates cargo lock but saves time by not creating binaries.
```
bitdivine, line 40: Actually, cargo fetch is a lot faster still and works as well for our purposes.

scripts/nns-dapp/bump-patch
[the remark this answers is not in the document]
dskloetd: I guess I must have been looking at line 239 in the step `Bumping the package versions works`.
But I see now it is called again on line 248 with `--commit`.

Anyway, I would still appreciate `if then` (throughout) so I don't have to make an extra step in my mind to verify I'm not crazy.

scripts/nns-dapp/bump-patch.test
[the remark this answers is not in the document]
dskloetd: I'm fine with requiring git to be clean but would not be comfortable with stash (even when notified) unless it's guarded by an explicit flag.

scripts/nns-dapp/bump-patch.test
dskloetd: ping

I'm really confused why you implemented a separate test in the workflow instead of invoking this test.

bitdivine: This test _is_ invoked from the workflow.  I  was not comfortable having the test make commits in a local repo.

Ah - in github the first test is missing the `.test` suffix.

scripts/nns-dapp/bump-patch
bitdivine: Already done.  See above.

bitdivine: Added a step-by-step explanation of what the awk does.

scripts/nns-dapp/bump-patch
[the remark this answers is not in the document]
bitdivine: Otherwise one might ban e.g. `forEach` in javascript because C programmers don't  have `forEach`.  Yes, every `forEach` _can_ be written as a for loop but it is not an improvement.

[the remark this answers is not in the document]
dskloetd: I think it's worth considering that most people on the team have little bash experience.
When anyone else responds to my bash PRs I usually get something like "rubber stamping because I don't understand bash".

scripts/nns-dapp/bump-patch
```diff
@@ -0,0 +1,59 @@
...
+#
+# Note: The code uses awk to edit the Cargo.toml.  Alternatives considered:
+#  - I looked for a cargo extension similar to `npm version`.  I found this:
+#      https://github.com/filipstefansson/cargo-semver
+#    It works, but has not released version 1.0.0 yet, is not used much, the install instructions
+#    on crates.io are out of date and the author has not responded to recent issues.  If that
+#    changes, we could use `cargo semver bump patch`.  This also updates the Cargo.lock.
+#  - I checked `cargo metadata` but that does not seem to have an option for editing the metadata.
+#  - yq currently parses TOML but does not emit TOML, so we cannot currently use that to edit Cargo.toml.
+#    Again, if that changes we can use that.
```
dskloetd, line 50: These 9 lines are not useful here. Just put them in the PR description if you really want to.
Did you think I asked for this?
Are you making fun of my review comments?

scripts/nns-dapp/bump-patch
dskloetd: You added it after I commented for the second time. Why did you say it's already done?

scripts/nns-dapp/bump-patch.test
dskloetd: > I was not comfortable having the test make commits in a local repo.

I don't understand. Isn't it making commits in the local repo on line 248?
```
./scripts/nns-dapp/bump-patch --commit
```

dskloetd (approved):

scripts/nns-dapp/bump-patch
bitdivine: On the contrary.  If a user wants to edit the code below, they might want to know that there are potentially better solutions out there.

PR #3231, opened 2023-08-29
Don't emit values from webworker timer if is has been stopped

peterpeterparker:
# Motivation

Do not emit values from the web worker timer if the job has been stopped.

Resolves https://github.com/dfinity/nns-dapp/pull/3230 for any worker that uses the timer utils.

# Changes

- extract the `postMessage` function to the timer utils
- within that function, add a checks to verify that timer is not idle and if so, do not call postMessage
- move and write new test with vitest
- copy and adapt related utils (`timers.test-utils.ts` and `utils.test-utils.ts`) for vitest


frontend/src/vitests/lib/worker-utils/timer.worker-utils.spec.ts
```diff
@@ -239,5 +240,74 @@ describe("timer.worker-utils", () => {
...
+
+      const job = async () => worker.postMsg(testMsg);
+
+      await worker.start({
+        job,
+        data: {},
+        interval: 5000,
+      });
+
+      expect(spyPostMessage).toHaveBeenCalledTimes(3);
```
dskloetd, line 260: Why do we expect 3 calls?
Can you add 3 `toHaveBeenCalledWith` calls instead of just 1, similar to the test below?

frontend/src/vitests/utils/timers.test-utils.ts
dskloetd, line 1: Can you add a comment mentioning that this is a vitest specific fork of `frontend/src/tests/utils/timers.test-utils.ts` and also add a comment to `frontend/src/tests/utils/timers.test-utils.ts` pointing to here?

frontend/src/vitests/utils/utils.test-utils.ts
dskloetd, line 1: Can you link this one to the original as well?

frontend/src/lib/worker-utils/timer.worker-utils.ts
```diff
@@ -76,6 +77,15 @@ export class TimerWorkerUtils {
     this.setStatus("idle");
   }
 
+  postMsg<T>(data: { msg: PostMessageResponse; data: T }) {
+    if (this.isIdle()) {
+      // The worker was stopped between the start of the execution and the actual completion of the job it runs.
```
dskloetd, line 82: Is this in any way related to the vitests migration or did you include that just because you don't want to write new jest tests?

frontend/src/vitests/lib/worker-utils/timer.worker-utils.spec.ts
peterpeterparker: Sure

frontend/src/vitests/utils/timers.test-utils.ts
peterpeterparker: I've added a comment and todo in original file.

frontend/src/vitests/utils/utils.test-utils.ts
peterpeterparker: I've added a comment and todo in original file.

frontend/src/lib/worker-utils/timer.worker-utils.ts
peterpeterparker: It's a comment which is suppose to explain why we return without further notice. It's covered by test `"should not allow call to postMessage when timer is already idle"`. If comment is not clear I can remove it if you rather like?

peterpeterparker:
FYI @dskloetd while implementing your feedback I noticed Webstorm IDE did actually not resolved the pseudo imports `$vitests` so I set the config for those in https://github.com/dfinity/nns-dapp/pull/3231/commits/da54f187b9b7937e5f59c33d224a8c4abcca7276

frontend/src/lib/worker-utils/timer.worker-utils.ts
dskloetd: Sorry, I was unclear.
I meant to ask if not emitting values on idle workers is related to the vitest migration.
My question wasn't about this specific comment but about the thing this comment is also about.

peterpeterparker: No it is not related to vitest.

It is what we discussed when you solved https://github.com/dfinity/nns-dapp/pull/3230 by setting one particular callback to undefined. When a timer is stopped - idle - no postMessage is throw so no callback on destroyed component is per extension called.

dskloetd: OK, for next time it would be nice (raise fewer questions) if unrelated changes are in different PRs.

dskloetd (approved):

PR #3233, opened 2023-08-29
Use the upstream slack action directly

bitdivine:
# Motivation
We have a copy of the II slack action in this repo.  We can actually refer to the upstream code directly.

# Changes
- Delete the local copy of the II slack action
- Call the upstream action instead

# Tests
- Tested for one action in the snsdemo repo.  It created  a notification.

# Todos


nmattia (approved):
LGTM!

PR #3234, opened 2023-08-29
Remove i18n key disburse_maturity

peterpeterparker:
# Motivation

I don't know if correct but locally, I always get this difference when building main. So I am guessing  a PR removed the key from `en.json` and we do not need `disburse_maturity: string` anymore in i18n.


dskloetd:
> So I am guessing a PR removed the key from en.json and we do not need disburse_maturity: string anymore in i18n.

Isn't it here? https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/i18n/en.json#L595

dskloetd:
Actually, we have 2! Also this one: https://github.com/dfinity/nns-dapp/blob/main/frontend/src/lib/i18n/en.json#L635

`frontend/src/lib/types/i18n.d.ts` also has 2 and 1 is being removed.
I think we should remove one from `en.json` as well but which one?

mstrasinskis (approved):
Thank you!

mstrasinskis:
A redundant label was removed and manually tested. Everything looks as expected.

PR #3236, opened 2023-08-30
Format dropdown options without whitespace

dskloetd:
# Motivation

In `frontend/src/lib/components/accounts/SelectAccountDropdown.svelte` dropdown options were formatted such that extra whitespace appears in the option text. This makes it harder to find the right option in tests and is simply unnecessary.

# Changes

1. Format the options such that no additional whitespace appears in the options.
2. Remove a `trim()` which is now unnecessary.

# Tests

Still pass.

# Todos

not necessary


mstrasinskis (approved):
LGTM, thanks!

PR #3237, opened 2023-08-30
Use page objects in DisburseNnsNeuronModal.spec.ts

dskloetd:
# Motivation

I plan to change `DisburseNnsNeuronModal` to use `SelectDestinationAddress` to bring it more in line with `TransactionModal`.
The test will be easier to adapt when it uses page objects already.

# Changes

1. Use page objects in `DisburseNnsNeuronModal.spec.ts`.
2. When the test expects a promise to have not yet resolved (when loading accounts) use an explicitly not-yet-resolved promise.

# Tests

only tests

# Todos

not necessary


mstrasinskis (approved):
Thank you!

PR #3239, opened 2023-08-30
E2E: Only toggle the address select if necessary

dskloetd:
# Motivation

There is a TODO in `frontend/src/tests/page-objects/SelectDestinationAddress.page-object.ts` to only toggle the input method if necessary.
I'm working on a PR to reuse the `SelectDestinationAddress` component in the `DisburseNnsNeuronModal` component and for that's it's useful to finally do this TODO.

# Changes

1. Fix the type of `isChecked`. There is no reason for it to return `null`.
2. Add functionality to `frontend/src/tests/page-objects/Toggle.page-object.ts` to check the value and set a specific value.
3. Add functions to `frontend/src/tests/page-objects/SelectDestinationAddress.page-object.ts` to specifically enable select input or text input.
4. Specifically choose select input before selecting an account.

# Tests

test changes only

# Todos

not necessary

### DIFF
--- a/frontend/src/tests/page-objects/SelectDestinationAddress.page-object.ts
+++ b/frontend/src/tests/page-objects/SelectDestinationAddress.page-object.ts
@@ -24,9 +24,16 @@ export class SelectDestinationAddressPo extends BasePageObject {
     return this.getTogglePo().toggle();
   }
 
+  async enableSelect(): Promise<void> {
+    return this.getTogglePo().setDisabled();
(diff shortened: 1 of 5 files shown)


lmuntaner (approved):
Thanks! This is still used only for e2e tests, right?

dskloetd:
> Thanks! This is still used only for e2e tests, right?

Yes, but that will soon change.

PR #3240, opened 2023-08-30
Accounts db

bitdivine:
# Motivation
We would like to move account storage to stable memory.  The plan involves creating a virtual database that proxies calls to one or more underlying database implementations.

# Changes
- Place a proxy layer between database calls and the current storage.
  Note: Currently accounts are stored on the heap, in a BTreeMap.  This is unchanged.  However access to the map has to go through the proxy now.

# Tests
* I think that existing tests should suffice.

# Todos

  - Covered by the existing comment: Add a database abstraction layer, preparing for migration.


rs/backend/src/accounts_store.rs
```diff
@@ -298,7 +299,7 @@ pub enum AddPendingTransactionResponse {
 impl AccountsStore {
     pub fn get_account(&self, caller: PrincipalId) -> Option<AccountDetails> {
         let account_identifier = AccountIdentifier::from(caller);
-        if let Some(account) = self.accounts.get(&account_identifier.to_vec()) {
+        if let Some(account) = self.accounts_db.db_get_account(&account_identifier.to_vec()) {
```
bitdivine, line 302: All the remaining changes replace direct map access for database API calls.

rs/backend/src/accounts_store.rs
```diff
@@ -738,25 +748,26 @@ impl AccountsStore {
...
                 if account_identifier == request.account_identifier {
-                    &account.default_account_transactions
+                    account.default_account_transactions
                 } else if let Some(hardware_wallet_account) = account
                     .hardware_wallet_accounts
                     .iter()
                     .find(|a| request.account_identifier == AccountIdentifier::from(a.principal))
                 {
-                    &hardware_wallet_account.transactions
+                    hardware_wallet_account.transactions.clone()
```
frederikrothenberger, line 764: Do you need to introduce the `clone` here, if you match on `Some(ref account)`?

frederikrothenberger (approved):
LGTM, just one small comment!

rs/backend/src/accounts_store.rs
bitdivine: I actually wanted to do that at first but backed out of it.  That said, the code has had some changes since, so let me check whether the reason is still valid.

bitdivine: Whatever the original reason, the change works now.  Good catch!  Thanks.

PR #3241, opened 2023-08-30
GIX-1823: Use ENABLE_SNS_AGGREGATOR_STORE in sns.services

lmuntaner:
# Motivation

Final motivation: set ENABLE_SNS_AGGREGATOR_STORE to `true`.

Do not set the snsQueryStore in the services unless `ENABLE_SNS_AGGREGATOR_STORE` is false.

Otherwise, the tests need to also fill the `snsQueryStore` and we want to remove it.

# Changes

* Use ENABLE_SNS_AGGREGATOR_STORE in `loadSnsDerivedState` and `loadSnsLifecycle`.

# Tests

# Todos

### DIFF
--- a/frontend/src/tests/lib/services/sns.services.spec.ts
+++ b/frontend/src/tests/lib/services/sns.services.spec.ts
@@ -8,7 +8,7 @@ import { WATCH_SALE_STATE_EVERY_MILLISECONDS } from "$lib/constants/sns.constant
 import * as services from "$lib/services/sns.services";
 import { snsDerivedStateStore } from "$lib/stores/sns-derived-state.store";
 import { snsLifecycleStore } from "$lib/stores/sns-lifecycle.store";
-import { snsQueryStore, snsSwapCommitmentsStore } from "$lib/stores/sns.store";
+import { snsSwapCommitmentsStore } from "$lib/stores/sns.store";
 import {
   mockIdentity,
   mockPrincipal,
@@ -18,7 +18,7 @@ import {
   mockSnsSwapCommitment,
   principal,
 } from "$tests/mocks/sns-projects.mock";
-import { setSnsProjects } from "$tests/utils/sns.test-utils";
+import { resetSnsProjects, setSnsProjects } from "$tests/utils/sns.test-utils";
 import {
   advanceTime,
   runResolvedPromises,
@@ -29,7 +29,6 @@ import type {
   SnsGetLifecycleResponse,
 } from "@dfinity/sns";
 import { SnsSwapLifecycle } from "@dfinity/sns";
-import { fromNullable } from "@dfinity/utils";
 import { waitFor } from "@testing-library/svelte";
 import { get } from "svelte/store";
 
@@ -51,7 +50,7 @@ describe("sns-services", () => {
     jest.clearAllTimers();
     jest.clearAllMocks();
     snsSwapCommitmentsStore.reset();
-    snsQueryStore.reset();
+    resetSnsProjects();
     snsDerivedStateStore.reset();
   });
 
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/services/sns.services.spec.ts
```diff
@@ -109,84 +111,128 @@ describe("sns-services", () => {
   });
 
   describe("loadSnsDerivedState", () => {
```
dskloetd, line 113: The diff below here is really hard to read.
Can you describe the changes below?
Is it simply duplicating all the tests with and without the flag?
And with some expectations removed?

> The difference is only that when it's false, we don't expect the snsQueryStore to be filled.

Does that mean only expectations are removed or do we actively expect `snsQueryStore` not to be filled when it shouldn't have been?

lmuntaner: I know. Let me add some comments then.

I removed the expectations, but I can also check that the snsQueryStore is not filled.

frontend/src/tests/lib/services/sns.services.spec.ts
```diff
@@ -109,84 +111,128 @@ describe("sns-services", () => {
...
-    it("should call api to get total commitments and load them in stores", async () => {
```
lmuntaner, line 112: This withuot touching has been moved under `with ENABLE_SNS_AGGREGATOR_STORE false`.

The one in `with ENABLE_SNS_AGGREGATOR_STORE true` is the same but without the expectations in snsQueryStore.

frontend/src/tests/lib/services/sns.services.spec.ts
```diff
@@ -109,84 +111,128 @@ describe("sns-services", () => {
...
+
+        expect(
+          get(snsDerivedStateStore)[rootCanisterId1.toText()]?.derivedState
+        ).toEqual(derivedState);
+      });
 
-      expect(
-        get(snsDerivedStateStore)[rootCanisterId1.toText()]?.derivedState
-      ).toEqual(derivedState);
+      it("should call api with the strategy passed", async () => {
```
lmuntaner, line 152: This was already existing before. Nothing has changed.

frontend/src/tests/lib/services/sns.services.spec.ts
```diff
@@ -109,84 +111,128 @@ describe("sns-services", () => {
...
+    describe("with ENABLE_SNS_AGGREGATOR_STORE false", () => {
+      beforeEach(() => {
+        overrideFeatureFlagsStore.setFlag("ENABLE_SNS_AGGREGATOR_STORE", false);
       });
-
-      expect(spy).toBeCalledWith({
-        rootCanisterId: mockPrincipal.toText(),
-        identity: mockIdentity,
-        certified: true,
+      it("should call api to get total commitments and load them in stores", async () => {
```
lmuntaner, line 182: This is exactly the same that was before.

frontend/src/tests/lib/services/sns.services.spec.ts
```diff
@@ -330,48 +380,92 @@ describe("sns-services", () => {
   });
 
   describe("loadSnsLifecycle", () => {
-    it("should call api to get lifecycle and load them in store", async () => {
```
lmuntaner, line 333: This exact test has been moved inside `with ENABLE_SNS_AGGREGATOR_STORE false`

frontend/src/tests/lib/services/sns.services.spec.ts
```diff
@@ -330,48 +380,92 @@ describe("sns-services", () => {
...
-      )?.swap[0].lifecycle;
-      expect(initLifecycle).toEqual(dataLifecycle);
+    describe("with ENABLE_SNS_AGGREGATOR_STORE true", () => {
+      beforeEach(() => {
+        overrideFeatureFlagsStore.setFlag("ENABLE_SNS_AGGREGATOR_STORE", true);
+      });
 
-      await services.loadSnsLifecycle({
-        rootCanisterId: rootCanisterId1.toText(),
+      it("should call api to get lifecycle and load them in store", async () => {
```
lmuntaner, line 388: This is the same test as before, but without the snsQueryStore expectations.

frontend/src/tests/lib/services/sns.services.spec.ts
```diff
@@ -330,48 +380,92 @@ describe("sns-services", () => {
...
-      expect(updatedLifecycle).toEqual(newLifeCycle);
+    describe("with ENABLE_SNS_AGGREGATOR_STORE false", () => {
+      beforeEach(() => {
+        overrideFeatureFlagsStore.setFlag("ENABLE_SNS_AGGREGATOR_STORE", false);
+      });
 
-      expect(get(snsLifecycleStore)[rootCanisterId1.toText()].data).toEqual(
-        lifeCycleResponse
-      );
+      it("should call api to get lifecycle and load them in store", async () => {
```
lmuntaner, line 426: This is exactly the same file as before

dskloetd (approved):

frontend/src/tests/lib/services/sns.services.spec.ts
lmuntaner: I added it.

PR #3243, opened 2023-08-30
GIX-1823: Refactor snsQueryStoreIsLoading

lmuntaner:
# Motivation

Stop using snsQueryStore. `snsQueryStoreIsLoading` has snsQueryStore as its only dependency.

In this PR, use also the snsAggregatorStore inside `snsQueryStoreIsLoading` to return whether data is loading or not.

# Changes

* Add dependency of snsAggregator store in snsQueryStoreIsLoading.

# Tests

* Add a describe for only `snsQueryStoreIsLoading` with two test cases. Instead of having one test case for it.

# Todos

Not necessary

### DIFF
--- a/frontend/src/lib/stores/sns.store.ts
+++ b/frontend/src/lib/stores/sns.store.ts
@@ -282,9 +282,13 @@ const initSnsQueryStore = (): SnsQueryStore => {
  */
 export const snsQueryStore = initSnsQueryStore();
 
-export const snsQueryStoreIsLoading = derived<SnsQueryStore, boolean>(
-  snsQueryStore,
-  (data: SnsQueryStoreData) => isNullish(data)
+export const snsQueryStoreIsLoading = derived<
+  [SnsQueryStore, SnsAggregatorStore],
+  boolean
+>(
+  [snsQueryStore, snsAggregatorStore],
+  ([snsQueryStoreData, aggregatorData]) =>
+    isNullish(snsQueryStoreData) && isNullish(aggregatorData.data)
 );
 
 /**
--- a/frontend/src/tests/lib/stores/sns.store.spec.ts
+++ b/frontend/src/tests/lib/stores/sns.store.spec.ts
@@ -151,19 +151,6 @@ describe("sns.store", () => {
       expect(store).toBeUndefined();
     });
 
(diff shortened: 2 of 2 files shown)


frontend/src/lib/stores/sns.store.ts
```diff
@@ -282,9 +282,13 @@ const initSnsQueryStore = (): SnsQueryStore => {
...
-export const snsQueryStoreIsLoading = derived<SnsQueryStore, boolean>(
-  snsQueryStore,
-  (data: SnsQueryStoreData) => isNullish(data)
+export const snsQueryStoreIsLoading = derived<
+  [SnsQueryStore, SnsAggregatorStore],
+  boolean
+>(
+  [snsQueryStore, snsAggregatorStore],
+  ([snsQueryStoreData, aggregatorData]) =>
+    isNullish(snsQueryStoreData) && isNullish(aggregatorData.data)
```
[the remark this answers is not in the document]
lmuntaner: Ah, ok, good point. I can check depending on the flag.

frontend/src/lib/stores/sns.store.ts
```diff
@@ -282,9 +282,15 @@ const initSnsQueryStore = (): SnsQueryStore => {
  */
 export const snsQueryStore = initSnsQueryStore();
 
-export const snsQueryStoreIsLoading = derived<SnsQueryStore, boolean>(
-  snsQueryStore,
-  (data: SnsQueryStoreData) => isNullish(data)
+export const snsQueryStoreIsLoading = derived<
```
dskloetd, line 285: Sorry, I'm just having another look and the name of this thing `snsQueryStoreIsLoading` is explicitly about the `snsQueryStore`. Does it really make sense to add additional logic here?
Shouldn't clients just stop using `snsQueryStoreIsLoading` as they stop using `snsQueryStore` as well?

lmuntaner: It's used only in one place. I'll rename it, I agree it's not accurate.

I'm also afraid of doing too much in one PR because then we lose sight of the change.

lmuntaner:

dskloetd (approved):

PR #3249, opened 2023-08-30
Change account storage from HashMap to BTreeMap

bitdivine:
# Motivation
We wish to be able to iterate over accounts in a consistent way so that we can migrate accounts from one schema to another in chunks.  `HashMap` does not guarantee a stable order for iterating whereas BTree does.  Let us convert, but be cautious of any change in cycle consumption.

Notes:

- The Candid serialization of `HashMap` and `BTreeMap` is the same: As a vector of `(key, value)` tuples.
- The pre-upgrade cycle count was last measured as 5% of the limit, so it is almost certainly not a concern.  It would be nice to have this available as a statistic in PRs though.
- `BTreeMap` does support [`Cursor`](https://doc.rust-lang.org/std/collections/btree_map/struct.Cursor.html) but unfortunately all the useful cursor API methods are still labelled experimental.  Why is this relevant?  Suppose that accounts up to key `x` have been migrated.  One could then look up the [entry](https://doc.rust-lang.org/std/collections/btree_map/enum.Entry.html) for `x` and use the cursor to get the next 100 accounts for migration.  This would be a lot more efficient than iterating over all keys until `x` is found.  Going forward, it would also be useful to have cursors on `StableBTreeMap`s.

# Changes
- Change the account storage type from `HashMap` to `BTreeMap`.

# Tests
Here are the cycle counts from two CI runs, one with BTreeMap and one with HashMap.

### HashMap
Screenshot from 2023-08-31 02-15-11
[The CI run.](https://github.com/dfinity/nns-dapp/actions/runs/6025533342)

### BTreeMap
Screenshot from 2023-08-31 02-14-53
[The CI run](https://github.com/dfinity/nns-dapp/actions/runs/6026859114)

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -17,6 +17,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Changed
 
 * Bigger icon and description first on Sns project page.
+* Change accounts storage heap structure from `HashMap` to `BTreeMap`.
 
 #### Deprecated
 #### Removed
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -17,7 +17,7 @@ use itertools::Itertools;
 use on_wire::{FromWire, IntoWire};
 use serde::Deserialize;
 use std::cmp::{min, Ordering};
-use std::collections::{HashMap, HashSet, VecDeque};
+use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 use std::ops::RangeTo;
 use std::time::{Duration, SystemTime};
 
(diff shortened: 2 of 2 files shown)


frederikrothenberger:
> Going forward, it would also be useful to have cursors on StableBTreeMaps.

The `range` function of the stable BTreeMap is not experimental and works great (we use it in II): https://github.com/dfinity/stable-structures/blob/e6dcf1b54698ba8484d57f0743df1ba8ee0114b2/src/btreemap.rs#L844

frederikrothenberger (approved):
LGTM thanks!

dskloetd (approved):

bitdivine:
@frederikrothenberger Re range: Cool, thanks for the tip!  `std::collections::BTreeMap` has `range()` as well, so that is super useful.

PR #3251, opened 2023-08-31
Update example

bitdivine:
# Motivation
The rustdoc for the AccountsDb trait needed to be updated after 1bb96c0b288f98981d7bb1093044e54f8daf8cc7.

Note: This was not caught in CI because the schema module is still behind test configuration.  It probably make sense to remove the test flag now, as the code will be used soon and doing so catches errors like this.

# Changes
* Add the schema module to the production codebase.  It is not used yet so the optimizer will probably remove it.  This is fine.
* Update the rustdoc.

# Tests
- See CI.

# Todos

  I think this is minor enough not to need a changelog entry.

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -21,7 +21,6 @@ use std::collections::{HashMap, HashSet, VecDeque};
 use std::ops::RangeTo;
 use std::time::{Duration, SystemTime};
 
-#[cfg(test)] // TODO: Make available in prod when ready.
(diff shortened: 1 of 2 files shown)


frederikrothenberger (approved):
Ah, I missed the outdated example on the previous reviews...
Thanks for fixing it 👍

bitdivine:
Thank you for the review, @frederikrothenberger .  The example wasn't tested yet otherwise we would have caught it then.

PR #3252, opened 2023-08-31
GIX-1841: Add sns finalization status stores

lmuntaner:
# Motivation

We want to add a new sns project status: Finalizing.

In this PR, I introduce the store where the data from the endpoint will be saved and a derived store with the logic to return whether the project is finalizing or not.

# Changes

* New dynamic store creator: `getOrCreateSnsFinalizationStatusStore`.
* New dynamic derived store: `createIsSnsFinalizingStore`.

# Tests

* Test new stores.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/lib/utils/sns.utils.ts
+++ b/frontend/src/lib/utils/sns.utils.ts
@@ -17,6 +17,7 @@ import { mapOptionalToken } from "$lib/utils/icrc-tokens.utils";
 import { AccountIdentifier, SubAccount } from "@dfinity/nns";
 import { Principal } from "@dfinity/principal";
 import type {
+  SnsGetAutoFinalizationStatusResponse,
   SnsGetMetadataResponse,
   SnsParams,
   SnsSwap,
@@ -293,4 +294,22 @@ export const parseSnsSwapSaleBuyerCount = (
       ?.split(/\s/)?.[1]
   );
   return isNaN(value) ? undefined : value;
+};
+
+/**
+ * An SNS is in finalization state if:
+ *
+ * 1. `has_auto_finalize_been_attempted` is true
+ * 2. `auto_finalize_swap_response` is empty
+ */
+export const isSnsFinalizing = (
+  finalizationStatus: SnsGetAutoFinalizationStatusResponse
+): boolean => {
+  const { has_auto_finalize_been_attempted, auto_finalize_swap_response } =
+    finalizationStatus;
+
+  return (
+    Boolean(fromNullable(has_auto_finalize_been_attempted)) &&
+    isNullish(fromNullable(auto_finalize_swap_response))
+  );
 };
--- a/frontend/src/tests/lib/stores/sns-finalization-status.store.spec.ts
+++ b/frontend/src/tests/lib/stores/sns-finalization-status.store.spec.ts
@@ -1,0 +1,142 @@
+import {
+  createIsSnsFinalizingStore,
+  getOrCreateSnsFinalizationStatusStore,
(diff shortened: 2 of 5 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/stores/sns-finalization-status.store.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+      store2.setData({
+        data: anotherResponse,
+        certified: true,
+      });
+
+      expect(get(store2).data).toEqual(anotherResponse);
+    });
+  });
+
+  describe("createIsSnsFinalizingStore", () => {
```
dskloetd, line 81: Maybe I'm missing it but I don't see a test where it's false because the method is not supported.

frontend/src/tests/mocks/sns-finalization-status.mock.ts
```diff
@@ -0,0 +1,28 @@
...
+          set_dapp_controllers_call_result: [],
+          settle_community_fund_participation_result: [],
+          error_message: [],
+          set_mode_call_result: [],
+          sweep_icp_result: [],
+          claim_neuron_result: [],
+          sweep_sns_result: [],
+        },
+      ],
+  has_auto_finalize_been_attempted: isFinalizing ? [true] : [false],
```
dskloetd, line 27: If `isFinalizing` is `false`, you have `has_auto_finalize_been_attempted` but also `auto_finalize_swap_response` has a value. Isn't that an unrealistic state to have? How can you have a response without even trying?

frontend/src/tests/lib/stores/sns-finalization-status.store.spec.ts
lmuntaner: Good point, I'll add that test.

If it's not supported, it will be `undefined`, and this will return `false`.

frontend/src/tests/mocks/sns-finalization-status.mock.ts
lmuntaner: Yes, good point. That would be a weird state. I'll change it.

lmuntaner:

I moved the logic of "finalizing" to a util to have it more isolated.

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -294,3 +295,21 @@ export const parseSnsSwapSaleBuyerCount = (
   );
   return isNaN(value) ? undefined : value;
 };
+
+/**
+ * An SNS is in finalization state if:
+ *
+ * 1. `has_auto_finalize_been_attempted` is true
+ * 2. `auto_finalize_swap_response` is empty
```
dskloetd, line 303: By the way, if finalization is attempted but then fails and is then attempted again, is the first response removed?
Or is it possible that a second finalization is being attempted but we wouldn't recognize it because of the first response?

dskloetd (approved):

frontend/src/lib/utils/sns.utils.ts
lmuntaner: That's a good question.

Check the slack thread for more details: https://dfinity.slack.com/archives/C0175CHJ0P6/p1692951861506129

Some quotes from it.

> this api was designed with the assumption that it won’t fail

> to my knowledge, the main reason it might fail in an “unexpected” way is if it hits the instruction limit somehow or we run out of cycles. IIRC this happened once (and the bug that caused it was fixed). Another could be a bug that causes a panic. In that case you won’t be able to tell what happened with this API

PR #3255, opened 2023-08-31
Support wallet.~ domains

dskloetd:
# Motivation

We want to make NNS Dapp accessible on `wallet.ic0.app` and `wallet.internetcomputer.org`.

# Changes

1. Add both domains to `frontend/static/.well-known/ic-domains`. This allows the canister to be accessible on those domain.
2. Add both origins to `frontend/static/.well-known/ii-alternative-origins`. This allows people to get the same principal from Internet Identity on those domains as they get on `nns.ic0.app`.
3. Make `isNnsAlternativeOrigin` return `true` for the new origins. This causes the app to pass a `derivationOrigin` of `"https://nns.ic0.app"` to Internet Identity on line 90 of `frontend/src/lib/stores/auth.store.ts`.
4. Use `identity.ic0.app` (instead of `identity.internetcomputer.org`) for any domain ending in `.ic0.app`, not just `nns.ic0.app`.
5. Add missing test that we use `identity.internetcomputer.org` by default.
6. Add tests for the other changes.

# Tests

Unit tests pass.
Not sure how to test if it actually works, short of releasing.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -13,6 +13,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Added
 
 * Disburse maturity of sns neurons.
+* Make NNS Dapp accessible via wallet.ic0.app and wallet.internetcomputer.org.
 
 #### Changed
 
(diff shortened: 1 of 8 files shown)


frederikrothenberger (approved):
Changes to the `.well-known` files look good. 👍

lmuntaner (approved):
Thanks, LGTM!

PR #3256, opened 2023-08-31
GIX-1759: Remove unnecessary variant prop

lmuntaner:
# Motivation

A "variant" prop was introduced to reused the same buttons when implementing a new UI for neuron details.

The old UI has already been removed.

In this PR, I remove the "variant" prop which is not needed anymore and set the buttons to "secondary".

# Changes

* Remove "variant" prop in neuron related buttons.

# Tests

* Remove tests regarding the variant prop.

# Todos

Not necessary. Already covered by the changelog entry specifying cleaning components.

### DIFF
--- a/frontend/src/tests/lib/components/neuron-detail/actions/IncreaseStakeButton.spec.ts
+++ b/frontend/src/tests/lib/components/neuron-detail/actions/IncreaseStakeButton.spec.ts
@@ -32,16 +32,4 @@ describe("IncreaseStakeButton", () => {
 
     expect(increaseStake).toBeCalledTimes(1);
   });
-
-  it("renders variant primary by default", async () => {
-    const po = await renderComponent({});
-    expect(await po.getVariant()).toBe("primary");
-  });
-
-  it("renders variant secondary if set", async () => {
-    const po = await renderComponent({
-      variant: "secondary",
-    });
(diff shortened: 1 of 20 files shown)


frontend/src/tests/lib/components/neuron-detail/actions/IncreaseStakeButton.spec.ts
```diff
@@ -32,16 +32,4 @@ describe("IncreaseStakeButton", () => {
...
-  it("renders variant primary by default", async () => {
-    const po = await renderComponent({});
-    expect(await po.getVariant()).toBe("primary");
-  });
-
-  it("renders variant secondary if set", async () => {
-    const po = await renderComponent({
-      variant: "secondary",
-    });
-    expect(await po.getVariant()).toBe("secondary");
```
dskloetd, line 45: Can you also remove the `getVariant()` method from the PO?

dskloetd (approved):

frontend/src/tests/lib/components/neuron-detail/actions/IncreaseStakeButton.spec.ts
lmuntaner: Yes, and a related test also.

PR #3258, opened 2023-08-31
GIX-1823: Set sns aggregator store flag true for tests

lmuntaner:
# Motivation

Set ENABLE_SNS_AGGREGATOR_STORE `true` for tests.

# Changes

* Set ENABLE_SNS_AGGREGATOR_STORE to `true` in jest.setup.ts
* Use snsQueryStore in sns services spec file. This should have never been setSnsProjects.
* Add more parameters to `aggregatorSnsMockWith`.
* Use aggregatorStore in setSnsProjects test util.

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/jest-setup.ts
+++ b/frontend/jest-setup.ts
@@ -36,7 +36,7 @@ jest.mock("./src/lib/utils/env-vars.utils.ts", () => ({
       ENABLE_CKTESTBTC: true,
       ENABLE_ICP_ICRC: false,
       ENABLE_INSTANT_UNLOCK: true,
-      ENABLE_SNS_AGGREGATOR_STORE: false,
+      ENABLE_SNS_AGGREGATOR_STORE: true,
       ENABLE_DISBURSE_MATURITY: true,
       TEST_FLAG_EDITABLE: true,
       TEST_FLAG_NOT_EDITABLE: true,
--- a/frontend/src/tests/lib/pages/SnsProposals.spec.ts
+++ b/frontend/src/tests/lib/pages/SnsProposals.spec.ts
@@ -37,8 +37,6 @@ describe("SnsProposals", () => {
     )[0];
 
   const rootCanisterId = mockPrincipal;
-  const functionName = "test_function";
-  const functionId = BigInt(3);
 
   beforeEach(() => {
     jest.clearAllMocks();
@@ -50,7 +48,14 @@ describe("SnsProposals", () => {
   });
 
   describe("logged in user", () => {
+    const functionName = "test_function";
+    const functionId = BigInt(3);
     beforeEach(() => {
+      fakeSnsGovernanceApi.addNervousSystemFunctionWith({
+        rootCanisterId,
+        name: functionName,
+        id: functionId,
+      });
       jest
         .spyOn(authStore, "subscribe")
         .mockImplementation(mockAuthStoreSubscribe);
@@ -138,6 +143,9 @@ describe("SnsProposals", () => {
         identity: new AnonymousIdentity(),
         rootCanisterId,
       });
+      fakeSnsGovernanceApi.addNervousSystemFunctionWith({
+        rootCanisterId,
+      });
     });
 
     describe("Matching results", () => {
@@ -183,6 +191,10 @@ describe("SnsProposals", () => {
         rootCanisterId,
         ...proposals[1],
         action: functionId,
+      });
+      fakeSnsGovernanceApi.addNervousSystemFunctionWith({
+        rootCanisterId,
+        id: functionId,
       });
     });
 
--- a/frontend/src/tests/lib/pages/SnsProposals.spec.ts
+++ b/frontend/src/tests/lib/pages/SnsProposals.spec.ts
@@ -37,6 +37,8 @@ describe("SnsProposals", () => {
     )[0];
 
   const rootCanisterId = mockPrincipal;
+  const functionName = "test_function";
+  const functionId = BigInt(3);
 
   beforeEach(() => {
     jest.clearAllMocks();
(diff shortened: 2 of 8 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/routes/NeuronDetail.spec.ts
```diff
@@ -123,10 +123,8 @@ describe("NeuronDetail", () => {
       // Load SNS projects after rendering to make sure we don't load
       // NnsNeuronDetail instead, which was a bug we had.
       await loadSnsProjects();
-      expect(await po.isContentLoaded()).toBe(false);
-      await waitFor(async () => {
-        expect(await po.isContentLoaded()).toBe(true);
-      });
+      expect(await po.isContentLoaded()).toBe(true);
```
lmuntaner, line 126: Loading the project was faster and the first expect was already failing.

There is a previous check that content is not loaded. Therefore, it's covered.

frontend/jest-setup.ts
```diff
@@ -36,7 +36,7 @@ jest.mock("./src/lib/utils/env-vars.utils.ts", () => ({
       ENABLE_CKTESTBTC: true,
       ENABLE_ICP_ICRC: false,
       ENABLE_INSTANT_UNLOCK: true,
-      ENABLE_SNS_AGGREGATOR_STORE: false,
+      ENABLE_SNS_AGGREGATOR_STORE: true,
```
dskloetd, line 39: I'm not really comfortable with this.
I'd rather keep the test coverage of the disabled flag until the flag is actually enabled.
But any tests that have different behavior with or without the flag should ideally have a twin where the flag is enabled explicitly.

frontend/src/tests/lib/pages/SnsProposals.spec.ts
```diff
@@ -49,21 +52,21 @@ describe("SnsProposals", () => {
...
     ]);
     // Reset to default value
     page.mock({ data: { universe: rootCanisterId.toText() } });
   });
 
   describe("logged in user", () => {
-    const functionName = "test_function";
-    const functionId = BigInt(3);
     beforeEach(() => {
-      fakeSnsGovernanceApi.addNervousSystemFunctionWith({
```
dskloetd, line 62: Was this change necessary?
Don't we still have a separate store that is loaded this way?

frontend/src/tests/lib/services/sns.services.spec.ts
```diff
@@ -124,16 +126,17 @@ describe("sns-services", () => {
           direct_participant_count: [],
           cf_neuron_count: [],
         };
-        setSnsProjects([
```
dskloetd, line 127: How about making the behavior of `setSnsProjects` depend on the flag?

Otherwise at least keep the existing function with a different name. Reverting back to `snsQueryStore.setData(responses);` now seems silly.

frontend/jest-setup.ts
lmuntaner: I can also change it in dfx.json except for mainnet. That was my next plan.

frontend/src/tests/lib/pages/SnsProposals.spec.ts
lmuntaner: The functions are not loaded if they already exist in the snsAggregatorStore.

Therefore, the governance api is not called because we set the projects before.

frontend/src/tests/lib/services/sns.services.spec.ts
lmuntaner: This is not reverting back, this should not have been setSnsProjects. This tests literally deal with snsQueryStore. Not with SNSes or snsSummaries.

lmuntaner: I can change this in another PR if you prefer.

frontend/jest-setup.ts
lmuntaner: I think we should change the test before changing mainnet. I'm more comfortable with that.

frontend/src/tests/lib/pages/SnsProposals.spec.ts
dskloetd: Can you keep the old tests and disable the feature flag in it to make sure it continues to work until we have actually enabled the feature flag in mainnet?

frontend/src/tests/lib/services/sns.services.spec.ts
dskloetd: It *is* reverting back. You're saying that reverting back is intentional. But the other function was more readable.
And the whole point of that function was that you don't have to change individual calls to it.

frontend/jest-setup.ts
dskloetd: When do you plan to set the flag on mainnet?
At least make sure that every test that had to change has a twin where the flag is still disabled.

lmuntaner: Only the sns.services needed it and they are added here.

The other functionality uses snsSummariesStore, which has also tests with the feature flag enabled and disabled.

lmuntaner: I want to try the next release again, and I believe that next week we could turn it on.

frontend/src/tests/lib/pages/SnsProposals.spec.ts
lmuntaner: Yes, sounds good.

frontend/src/tests/lib/services/sns.services.spec.ts
lmuntaner: This tests deal direclty with snsQueryStore, not with setting SNSes. Therefore, it makes sense that here it specifically sets snsQueryStore.

I won't have to change individual calls to it after changing the flag, because this doesn't deal with the snsAggregatorStore. The only changes regardless of the feature flag are about setting or not setting the snsQueryStore (directly). But the snsAggregatorStore doesn't play any role in this tests.

lmuntaner:

dskloetd (approved):

PR #3259, opened 2023-08-31
Proxy db

bitdivine:
# Motivation
The plan for migrating data involves having one or more account databases and a virtual database that sits on top of these other databases and deals with the migration and database calls during the migration.

# Changes
- Define a proxy DB.

# Tests
- The standard db tests are applied to the proxy DB.

# Todos


rs/backend/src/accounts_store/schema/map.rs
```diff
@@ -30,6 +31,26 @@ impl AccountsDbTrait for AccountsDbAsMap {
     }
 }
 
+impl AccountsDbAsMap {
+    /// Creates a db from a map of accounts.
```
bitdivine, line 35: In the `post_upgrade` hook, a large blob of candid is parsed and the result includes a hashmap of all accounts.  `from_map()` lets us create a database from that data.

rs/backend/src/accounts_store/schema/map.rs
```diff
@@ -30,6 +31,26 @@ impl AccountsDbTrait for AccountsDbAsMap {
...
+    #[allow(dead_code)] // TODO: Remove allow when this is used in production.
+    pub fn from_map(map: BTreeMap<Vec<u8>, Account>) -> Self {
+        Self { accounts: map }
+    }
+    /// Provides the DB contents as a map.
```
bitdivine, line 40: This is used in the pre-upgrade hook when serializing data to candid.

rs/backend/src/accounts_store/schema/proxy.rs
```diff
@@ -0,0 +1,76 @@
...
+#[derive(Default, Debug)]
+pub struct AccountsDbAsProxy {
+    map: AccountsDbAsMap,
+}
+
+impl AccountsDbAsProxy {
+    /// Creates a db from a hash map of accounts.
+    #[allow(dead_code)] // TODO: Remove allow when this is used in production.
+    pub fn from_map(map: BTreeMap<Vec<u8>, Account>) -> Self {
+        Self {
```
bitdivine, line 28: Note: Currently we have only one underlying implementation, so we can proxy calls straight through to the underlying `AccountsDbAsMap`.

rs/backend/src/accounts_store/schema/map.rs
frederikrothenberger: Shouldn't this be used somewhere as part of a test allowing you to replace `#[allow(dead_code)]` with `#[cfg(test)]`?

rs/backend/src/accounts_store/schema/proxy.rs
```diff
@@ -0,0 +1,76 @@
...
+///
+/// Notes:
+/// - The proxy manages migrations from one implementation to another, if applicable.
+/// - The proxy code itself will specify which databases are currently in
+///   use and how to migrate from one database to another.
+/// - It is the responsibility of the post-install hook to look at any
+///   version information and set up the db accordingly.
+///
+/// # Current data storage
+/// - Accounts are
```
frederikrothenberger, line 18: There seems to be something missing here?

frederikrothenberger (approved):
Not a huge fan of the `#[allow(dead_code)]` attributes. I think it would be better to use these methods in tests too.

bitdivine:
@frederikrothenberger Thank you for the review.  I put the `from_map()` and `as_map()` methods in a trait, which makes the complaints about dead code go away.  I also tested that trait on both the proxy  and map databases.  You have approved, but in view of the changes you may want to have another look.

frederikrothenberger (approved):
Thanks, LGTM!

PR #3260, opened 2023-08-31
Add QrWizardModal component

dskloetd:
# Motivation

Using the QR code scanner requires adding an extra step to a `WizardModal` and switching back and forth between them.
This is currently done in `TransactionModal`.
I want to reuse the QR code scanner in the `DisburseNnsNeuronModal` so I want to streamline its usage.

# Changes

1. Add `QrWizardModal` which manages going to and from the QR Code step.
2. Use `QrWizardModal` in `TransactionModal`.
3. Tests for `QrWizardModal`.

# Tests

Tested manually on local that scanning a QR code when making a transaction still works.

# Todos

not necessary for a small refactoring

### DIFF
--- a/frontend/src/lib/modals/transaction/QrWizardModal.svelte
+++ b/frontend/src/lib/modals/transaction/QrWizardModal.svelte
@@ -19,15 +19,7 @@
     },
   ];
 
-  let modal: WizardModal;
-
-  export const next = () => {
-    modal.next();
-  };
-
-  export const back = () => {
-    modal.back();
-  };
+  export let modal: WizardModal;
 
   export const set = (step: number) => {
     modal.set(step);
--- a/frontend/src/lib/modals/transaction/QrWizardModal.svelte
+++ b/frontend/src/lib/modals/transaction/QrWizardModal.svelte
@@ -25,7 +25,7 @@
     modal.set(step);
   };
 
-  const setStepName = (stepName: string) => {
+  const goStep = (stepName: string) => {
     const stepNumber = stepsPlusQr.findIndex(({ name }) => name === stepName);
     set(stepNumber);
   };
@@ -36,14 +36,14 @@
   export const scanQrCode = async () => {
     const prevStep = currentStep;
 
-    setStepName(STEP_QRCODE);
+    goStep(STEP_QRCODE);
 
     return new Promise<string | undefined>((resolve) => {
       resolveQrCodePromise = resolve;
     }).finally(() => {
       // TypeScript can't know that currentStep is defined before scanQrCode
       // will be called.
-      setStepName(prevStep?.name || steps[0].name);
+      goStep(prevStep?.name || steps[0].name);
     });
   };
 
--- a/frontend/src/lib/modals/transaction/QrWizardModal.svelte
+++ b/frontend/src/lib/modals/transaction/QrWizardModal.svelte
@@ -49,10 +49,12 @@
 
   const onQRCode = ({ detail: value }: CustomEvent<string>) => {
     resolveQrCodePromise?.(value);
+    resolveQrCodePromise = undefined;
   };
 
   const onCancel = () => {
     resolveQrCodePromise?.(undefined);
+    resolveQrCodePromise = undefined;
   };
 </script>
 
--- a/frontend/src/lib/modals/transaction/QrWizardModal.svelte
+++ b/frontend/src/lib/modals/transaction/QrWizardModal.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import type { QrResponse, QrResult } from "$lib/types/qr-wizard-modal";
   import TransactionQRCode from "$lib/components/transaction/TransactionQRCode.svelte";
   import { toastsError } from "$lib/stores/toasts.store";
   import {
@@ -37,15 +38,6 @@
     set(stepNumber);
   };
 
-  type QrResult = "success" | "canceled" | "token_incompatible";
-
-  type QrResponse = {
-    result: QrResult;
-    identifier?: string;
-    token?: string;
-    amount?: number;
-  };
-
   let resolveQrCodePromise:
     | (({ result, value }: { result: QrResult; value?: string }) => void)
     | undefined = undefined;
@@ -62,6 +54,8 @@
     if (result !== "success") {
       return { result };
     }
+    // Because `result === "success"`, value is defined.
+    // This is just here to convince TypeScript.
     assertNonNullish(value);
 
     const payment = decodePayment(value);
--- a/frontend/src/lib/modals/transaction/QrWizardModal.svelte
+++ b/frontend/src/lib/modals/transaction/QrWizardModal.svelte
@@ -65,19 +65,17 @@
   };
 </script>
 
-{#if nonNullish(stepsPlusQr)}
-  <WizardModal
-    {testId}
-    steps={stepsPlusQr}
-    bind:currentStep
-    bind:this={modal}
-    on:nnsClose
-    {disablePointerEvents}
-  >
-    <slot name="title" slot="title" />
-    <slot />
-    {#if currentStep?.name === STEP_QRCODE}
-      <TransactionQRCode on:nnsCancel={onCancel} on:nnsQRCode={onQRCode} />
-    {/if}
-  </WizardModal>
-{/if}
+<WizardModal
+  {testId}
+  steps={stepsPlusQr}
+  bind:currentStep
+  bind:this={modal}
+  on:nnsClose
+  {disablePointerEvents}
+>
+  <slot name="title" slot="title" />
+  <slot />
+  {#if currentStep?.name === STEP_QRCODE}
+    <TransactionQRCode on:nnsCancel={onCancel} on:nnsQRCode={onQRCode} />
+  {/if}
+</WizardModal>
--- a/frontend/src/lib/modals/transaction/QrWizardModal.svelte
+++ b/frontend/src/lib/modals/transaction/QrWizardModal.svelte
@@ -2,7 +2,6 @@
   import { WizardModal } from "@dfinity/gix-components";
   import type { WizardStep, WizardSteps } from "@dfinity/gix-components";
   import TransactionQRCode from "$lib/components/transaction/TransactionQRCode.svelte";
-  import { nonNullish } from "@dfinity/utils";
 
   export let testId: string | undefined = undefined;
   export let steps: WizardSteps;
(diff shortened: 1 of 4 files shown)


frontend/src/lib/modals/transaction/QrWizardModal.svelte
```diff
@@ -0,0 +1,83 @@
...
+
+  export const back = () => {
+    modal.back();
+  };
+
+  export const set = (step: number) => {
+    modal.set(step);
+  };
+
+  const setStepName = (stepName: string) => {
```
peterpeterparker, line 37: in `TransactionModal`  we use the wording `goStep`. Does it make sense to change it?

frontend/src/lib/modals/transaction/QrWizardModal.svelte
```diff
@@ -0,0 +1,83 @@
...
+
+  let resolveQrCodePromise: ((value: string | undefined) => void) | undefined =
+    undefined;
+
+  export const scanQrCode = async () => {
+    const prevStep = currentStep;
+
+    setStepName(STEP_QRCODE);
+
+    return new Promise<string | undefined>((resolve) => {
```
peterpeterparker, line 50: Not sure I understand this complexity. Why not just using the events `nnsCancel` and `nnsQRCode` of the `TransactionQRCode` and forward the events with a dispatcher after the previous step has been set?

Pseudo code, we have that somewhere in the code base:

```
const onCancel = ($event) => {
   setStepName(prevStep?.name || steps[0].name);
   dispatch('nnsCancel', $event);
}
```

frontend/src/lib/modals/transaction/QrWizardModal.svelte
```diff
@@ -0,0 +1,83 @@
...
+  const onQRCode = ({ detail: value }: CustomEvent<string>) => {
+    resolveQrCodePromise?.(value);
+  };
+
+  const onCancel = () => {
+    resolveQrCodePromise?.(undefined);
+  };
+</script>
+
+{#if nonNullish(stepsPlusQr)}
```
peterpeterparker, line 68: We never use accross the code base the `WizardModal` within a `if`. Is there a good reason to do so here?

peterpeterparker (commented):
In addition to my questions and observations, should this component be part of the gix-cmp?

frontend/src/lib/modals/transaction/QrWizardModal.svelte
```diff
@@ -0,0 +1,83 @@
...
+    ...steps,
+    {
+      name: STEP_QRCODE,
+      title: "",
+    },
+  ];
+
+  let modal: WizardModal;
+
+  export const next = () => {
```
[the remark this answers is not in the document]
dskloetd: I'm not sure I understand.
If the user of `QrWizardModal` binds `bind:this={modal}`, then what they get is the `QrWizardModal` instance, not the `WizardModal` instance, right?
So in order to still be able to do `modal.next()` etc., `QrWizardModal` has to forward those calls, right?

I'd rather not have them, so if I'm missing something, I'd like to know.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
dskloetd: Oops, I tried this when something wasn't working and forgot to revert it later.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
dskloetd: I think it's nice for the client to invoke the QR code scanner, and get the result, in the same call, rather than having to separate that into seemingly unrelated mechanisms. Does that make sense?

frontend/src/lib/modals/transaction/QrWizardModal.svelte
dskloetd: I got confused about when a number is passed, when a step is passed and when a string is passed. I thought this made it clearer.

I'll change it to `goStep` if you want, or I can change `goStep` to `setStepName` or both to something else that you think is clearer.

dskloetd:
I can add it to `gix-component` but let's finish the conversation here first.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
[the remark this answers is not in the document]
peterpeterparker: I meant:

QrWizardModal:

```
<script lang="ts">
   export let modal
</script>

<WizardModal bind:this={modal}/>
```

Consumers:

```
<script lang="ts">
  let modal;
</script>

<QrWizardModal bind:modal />
```

Would that work?

frontend/src/lib/modals/transaction/QrWizardModal.svelte
peterpeterparker: To be honest, I am asking because I did not understood the purpose of `setStepName` at first. I thought it was meant to change the name of a step. So `goStep` because existing sounds more clear to me but, I'm also open to another name. Just one that express what it really does, going to a particular step in the modal by using a step name.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
peterpeterparker: Ah I understand. Have you check if this is compatible with the `TransactionModal` - i.e. is this pattern easy to integrate with the existing or will it requires many changes?

If easy to integrate, then here are two suggestions:

1. instead of returning `string | undefined` would it be interesting to returns also the reason with the promise? e.g. `Promise<{result: "code" | "cancelled"; value?: string}>`. The code of the consumer will be a bit more clear when processing the answer and also it can be handy if in the future we extend the results with for example an error.

2. Should the `resolveQrCodePromise` be resetted to `null` once the transition to previous step has been called?

peterpeterparker:
Actually I mislead it. We cannot move this component to gix-cmp as it wraps `TransactionQRCode` which is specific to NNS-dapp. 

Given this wizard will be integrated in `DisburseNnsNeuronModal`, we might want to rename `TransactionQRCode` afterwards to something more generic I guess.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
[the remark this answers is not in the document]
dskloetd: Yes, it works.
Although now I can no longer do `modal.scanQrCode` so I had to add a separate binding for `scanQrCode`.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
dskloetd: OK, I've changed it to `goStep`.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
dskloetd: > Have you check if this is compatible with the `TransactionModal` - i.e. is this pattern easy to integrate with the existing or will it requires many changes?

Have you seen that `TransactionModal.svelte` has been changed in this PR?

1. Yes, I can do that. Or alternatively I could bind the scanned address to the component and do all the token/amount/error handing in `QrWizardModal` as well. What do you think.
2. I don't think it's strictly necessary but it might help avoid confusion. Done.

dskloetd:
> we might want to rename TransactionQRCode afterwards to something more generic I guess.

In my mind disbursing a neuron is also a transaction (from the neuron account to your own account). Is that not how we think of it for naming?

frontend/src/lib/modals/transaction/QrWizardModal.svelte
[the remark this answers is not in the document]
peterpeterparker: Sounds good, thanks.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
peterpeterparker: > es, I can do that. Or alternatively I could bind the scanned address to the component and do all the token/amount/error handing in QrWizardModal as well. What do you think.

That's a good idea too. Which one you rather like? I'll follow your call on this.

dskloetd: I have moved the decoding and error handling to the `QrWizardModal` but I kept it as an async function.
I thought the binding only made sense if the `QrWizardModal` could also handle the `nnsOpenQRCodeReader` event but I don't think it can listen for that on its slot.

dskloetd (commented):
Please have another look.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
```diff
@@ -0,0 +1,143 @@
...
+    requiredToken,
+  }: {
+    result: QrResult;
+    value?: string | undefined;
+    requiredToken?: Token | undefined;
+  }): QrResponse => {
+    if (result !== "success") {
+      return { result };
+    }
+    assertNonNullish(value);
```
peterpeterparker, line 65: Maybe we should extend `QrResult` with a specific type of error for this particular `assertNonNullish` as well, that way the consumer does not have to wrap the decoding within a `try catch`?

frontend/src/lib/modals/transaction/TransactionModal.svelte
```diff
@@ -75,13 +75,15 @@
...
     },
-    {
-      name: STEP_QRCODE,
-      title: "",
-    },
   ];
 
   let modal: WizardModal;
+  let scanQrCode: ({ requiredToken }: { requiredToken: Token }) => Promise<{
+    result: "success" | "canceled" | "token_incompatible";
```
peterpeterparker, line 82: maybe extract the types to global or to `src/types/qr-wizard-modal.ts` that way we don't have to redeclare it?

frontend/src/lib/modals/transaction/TransactionModal.svelte
```diff
@@ -94,49 +96,34 @@
...
-      }
-
-      goForm();
+    if (result !== "success") {
       return;
     }
 
-    selectedDestinationAddress = value;
-    goForm();
+    if (nonNullish(paymentAmount)) {
```
peterpeterparker, line 113: maybe preserve the same order or the assigment?

```
selectedDestinationAddress = identifier;

if (nonNullish(paymentAmount)) {
   amount = paymentAmount;
}
```

not sure it is that important

peterpeterparker (approved):
LGTM, thanks a lot!

Just added few new minor comments if you can have a look before merge.

frontend/src/lib/modals/transaction/QrWizardModal.svelte
peterpeterparker: Code looks good, thanks!

PR #3264, opened 2023-09-01
GIX-1760: Fix spacings nns neuron detail

lmuntaner:
# Motivation

The different sections of NnsNeuronDetails page had the vertical spacing set in different places. This made it easy to have a different spacing between them.

In this PR, all the spacing is set in the parent and the children have no vertical spacing.

# Changes

* Add `noMargin` prop to CardInfo.
* Add `noMargin` prop to SkeletonCard.
* All sections and separators in NnsNeuronDetails are siblings in the html. Just like it looks like in the UI.
* Use the `noMargin` prop and remove the separator in NeuronFollowingCard, NeuronProposalsCard, NeuronVotingHistoryCard and NnsNeuronHotkeysCard.
* Moved the testnet only MakeProposalsCard above the NeuronVotingHistoryCard to have the beginning like the mainnet. I didn't move it below NeuronVotingHistoryCard because this one has an infinite scroll.

# Tests

Only CSS changes.

# Todos

- [ ] Add entry to changelog (if necessary).
Not necessary. No UI nor UX change.

PR #3267, opened 2023-09-01
GIX-1761: Align spacings sns neuron detail

lmuntaner:
# Motivation

The different sections of SnsNeuronDetails page had the vertical spacing set in different places. This made it easy to have a different spacing between them.

In this PR, all the spacing is set in the parent and the children have no vertical spacing.

# Changes

* All sections and separators in SnsNeuronDetails are siblings in the html. Just like it looks like in the UI.
* Use the noMargin prop and remove the separator in SnsNeuronPermissionsCard, SnsNeuronFollowingCard, SnsNeuronHotkeysCard.

# Tests

Only CSS changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/pages/SnsNeuronDetail.svelte
+++ b/frontend/src/lib/pages/SnsNeuronDetail.svelte
@@ -28,7 +28,7 @@
   import { snsSelectedTransactionFeeStore } from "$lib/derived/sns/sns-selected-transaction-fee.store";
   import type { Token } from "@dfinity/utils";
   import { snsTokenSymbolSelectedStore } from "$lib/derived/sns/sns-token-symbol-selected.store";
-  import { nonNullish } from "@dfinity/utils";
+  import { nonNullish, isNullish } from "@dfinity/utils";
   import { IS_TESTNET } from "$lib/constants/environment.constants";
   import SnsNeuronProposalsCard from "$lib/components/neuron-detail/SnsNeuronProposalsCard.svelte";
   import SnsPermissionsCard from "$lib/components/neuron-detail/SnsPermissionsCard.svelte";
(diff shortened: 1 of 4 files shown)


lmuntaner:
UI:

sns-neuron-detail

lmuntaner:
@dskloetd please review

frontend/src/lib/pages/SnsNeuronDetail.svelte
```diff
@@ -150,55 +150,51 @@
...
+          <Separator spacing="none" />
+          <SnsNeuronAdvancedSection
+            neuron={$selectedSnsNeuronStore.neuron}
+            {governanceCanisterId}
+            {parameters}
+            {token}
+            {transactionFee}
+          />
+          <Separator spacing="none" />
           <SnsNeuronFollowingCard />
```
dskloetd, line 191: Previously the `<SnsNeuronFollowingCard />` would still be rendered if `parameters`, `token`, `$selectedSnsNeuronStore.neuron` or `transactionFee` is nullish.
But now no longer.
1. Is this intended?
2. Does it have anything to do with the purpose of this PR?

dskloetd (commented):
I'm regretting I didn't get to merge the NNS and SNS versions of these component yet.

frontend/src/lib/pages/SnsNeuronDetail.svelte
lmuntaner: No, it wouldn't be shown because it always had the `{#if loading}` at the top that rendered the skeletons.

This PR doesn't change this functionality. When `loading` is false, the if statements are `true`.

lmuntaner:
> I'm regretting I didn't get to merge the NNS and SNS versions of these component yet.

Which versions? I don't understand it sorry.

dskloetd:
> Which versions? I don't understand it sorry.

I was referring to the fact that you need 2 PRs (this one and https://github.com/dfinity/nns-dapp/pull/3264) because there are separate components for NNS and SNS while I think they should be the same UI component.

dskloetd (approved):

PR #3270, opened 2023-09-04
Use page objects in SelectUniverseCard.spec.ts

dskloetd:
# Motivation

It's easier to modify the test when it uses page objects.
I want to modify the test in another PR in order to show the universe selector, but without balances, even when the user is not logged in.

# Changes

1. Add required functionality to page objects.
2. Add required `data-tid` to `frontend/src/lib/components/universe/UniverseAccountsBalance.svelte`.
3. Use page objects in `frontend/src/tests/lib/components/universe/SelectUniverseCard.spec.ts`
4. Use real `icpAccountsStore` instead of mocking it.
5. Set and expect explicit values for the balances.

# Tests

yes

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/universe/UniverseAccountsBalance.svelte
+++ b/frontend/src/lib/components/universe/UniverseAccountsBalance.svelte
@@ -22,7 +22,7 @@
       : undefined;
 </script>
 
-<div class="amount">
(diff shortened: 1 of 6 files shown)


mstrasinskis (approved):
Thanks!

PR #3272, opened 2023-09-04
Rename test-utils

peterpeterparker:
# Motivation

Pattern to exclude test utils from prod build is `**/*test-utils.ts`.

# Changes

- rename `transaction-modal.test.utils.ts` to `transaction-modal.test-utils.ts` to match pattern


frontend/src/tests/lib/modals/accounts/CkBTCTransactionModal.spec.ts
```diff
@@ -32,7 +32,7 @@ import {
   testTransferFormTokens,
   testTransferReviewTokens,
   testTransferTokens,
-} from "$tests/utils/transaction-modal.test.utils";
+} from "$tests/utils/transaction-modal.test-utils";
```
dskloetd, line 35: This is of course fine, but shouldn't everything under `frontend/src/tests` already be excluded from the prod build?

dskloetd (approved):

frontend/src/tests/lib/modals/accounts/CkBTCTransactionModal.spec.ts
peterpeterparker: I just adapted it for the current `tsconfig.json`:

```
"exclude": [
    "**/*.spec.ts",
    "jest.setup.ts",
    "**/*.mock.ts",
    "**/*.fake.ts",
    "**/*.page-object.ts",
    "**/*Test.svelte",
    "**/*.test-utils.ts"
  ]
```

Indeed maybe we can try to to update the config.

PR #3273, opened 2023-09-04
More stats

bitdivine:
# Motivation

# Changes

# Tests

# Todos


rs/backend/src/accounts_store/histogram.rs
```diff
@@ -0,0 +1,106 @@
...
+            *self.sub_account_transactions(sub_account.transactions.len()) += 1;
+        });
+        *self.hardware_wallet_accounts(rhs.hardware_wallet_accounts.len()) += 1;
+        *self.canisters(rhs.canisters.len()) += 1;
+        self
+    }
+}
+
+/// Determines which log base 2 bucket a count falls into.
+fn log2_bucket(mut count: usize) -> u32 {
```
frederikrothenberger, line 99: Is there a reason to not use the [built-in function](https://doc.rust-lang.org/std/primitive.usize.html#method.ilog2)?

rs/backend/src/accounts_store/histogram.rs
```diff
@@ -0,0 +1,106 @@
...
+    ///
+    /// Note: The buckets are logarithmic, as with `default_account_transactions`.
+    hardware_wallet_accounts: BTreeMap<u32, u64>,
+    /// A histogram of the number of canisters per account.
+    ///
+    /// Note: The buckets are logarithmic, as with `default_account_transactions`.
+    canisters: BTreeMap<u32, u64>,
+}
+
+// Getter sand setters for the histogram fields that ensure that data is placed in the right columns.
```
mstrasinskis, line 44:
```suggestion
// Getters and setters for the histogram fields that ensure that data is placed in the right columns.
```

rs/backend/src/accounts_store.rs
```diff
@@ -1071,6 +1073,14 @@ impl AccountsStore {
         stats.transactions_to_process_queue_length = self.multi_part_transactions_processor.get_queue_length();
     }
 
+    pub fn get_histogram(&self) -> AccountsStoreHistogram {
```
frederikrothenberger, line 1076: So iterating over everything whenever you want to know the stats is fast enough? Also in production?

frederikrothenberger (approved):
LGTM, only minor comments.

rs/backend/src/accounts_store/histogram.rs
bitdivine: I didn't  know about it.  Thanks for the pointer.  I would use it with some changes, e.g. the builtin panics at zero, and I chose the histogram keys to be `2**n-1` rather than the exponent as this seemed more natural, so we need to round up rather than down.  But this should give the same results: `(1 << usize::ilog2(count*2+1))-1`.  Let me plug that in...

bitdivine: .. and added a test just to be sure my math isn't wrong, although I guess the bucket test would catch an error.

PR #3274, opened 2023-09-04
Use page objects in SelectUniverseDropdown.spec.ts

dskloetd:
# Motivation

It's easier to modify the test when it uses page objects.
I want to modify the test in another PR in order to show the universe selector, but without balances, even when the user is not logged in.

# Changes

1. Add test ID to `frontend/src/lib/components/universe/SelectUniverseDropdown.svelte`
2. Add `SelectUniverseDropdownPo`.
3. Add `isButton` to `CardPo`.
4. Allow passing a test ID to `.isPresent()` to check presence of a child with a specific ID.
5. Change `SelectUniverseDropdown.spec.ts` to use page object and specify and expect explicit values.

# Tests

Manually checked that the dropdown still looks the same.

# Todos

not necessary


mstrasinskis (approved):
LGTM

PR #3276, opened 2023-09-04
Spellcheck yaml files

bitdivine:
# Motivation
The YAML files have various spelling mistakes and formatting errors.

# Changes
- Fix spelling.
- Tidy and commit the command I used to find the errors.

# Tests
None.  The spell checker should not be called in CI; it isn't ready for that.

# Todos

  - Not significant enough for a changelog entry, in my opinion.


.github/actions/start_dfx_snapshot/action.yaml
```diff
@@ -6,7 +6,7 @@ inputs:
   snsdemo_ref:
     description: "The commit at which to use the snsdemo scripts"
     required: false
-    # Version from 2023-06-06 with dfx v 0.14.1
+    # Version from 2023-06-06 with `dfx v 0.14.1`
```
dskloetd, line 9: `dfx v 0.14.1` is not a command so I'm not sure why should have this formatting.
Did `spellcheck` do this or am I right in assuming that the backticks in this PR were all added by hand?

In `.github/actions/build_nns_dapp/action.yaml` there is no space after the `v` and here there is.

scripts/spellcheck-yaml
```diff
@@ -0,0 +1,43 @@
...
+  grep -E '[.]ya?ml$'
+}
+
+check_file() {
+  local filename="$1"
+  local comments_file="$filename.spellcheck.md"
+  # Put the comments into another file, preserving line numbers:
+  sed -n 's/^ *#//g;ta;s/.*//g;:a;p' "$filename" >"$comments_file"
+  # Check spelling, but convert the filename in any problem report back to the original file.
+  # Hunspell does not provide the character offset in each line so we do not need to handle that.
```
dskloetd, line 27: What is "Hunspell"? I see no other mention of it in the PR.

bitdivine: Hunspell is a spell checker. It is used, moderately widely, but the _format_ of the hunspell dictionaries is used in many more programs.  Nuspell spings to mind but more relevant here are `cargo spellcheck` which uses hunspell and likewise [the MS code spell checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker#:~:text=Open%20up%20VS%20Code%20and,and%20reload%20window%20to%20enable.) - I just picked the first popular one with reasonable ratings.  I am sure there are other options if you want them.  Chances are that if you pick up a random spell-checker, it will support custom dictionaries in the hunspell format.

.github/actions/start_dfx_snapshot/action.yaml
bitdivine: All of these changes are made by hand.  I tend to put anything that isn't English into backticks, not just commands but also code snippets and machine output. So I would write that the th eoutput of `dfx --version` is `dfx 0.14.3`, for example, when I pay attention anyway! :-)

Regarding the `v` above, I didn't change the spacing but the line of text has been removed in another PR.  The ref is now passed in as a parameter, so this is moot.

scripts/spellcheck-yaml
bitdivine: I checked - the MS code spell cehcker seems to use the hunspell format but not hunspell itself, if I see corerctly.  In the simplest case, which applies here, the hunspell format just puts a number on the first line and words on the remaining lines, but more complicated patters are supported.  We just don't need any, yet.  We might want a pattern for e.g. v$semantic_version, for example.

.github/actions/start_dfx_snapshot/action.yaml
dskloetd: Can you update the branch so I can see how it looks now?

scripts/spellcheck-yaml
dskloetd: But what is relevant here is that `cargo spellcheck` does not provide character offsets, right?
What Hunspell does or whether `cargo spellcheck` uses Hunspell isn't relevant, if I understand correctly.
So mentioning Hunspell, rather than `cargo spellcheck` here is a bit confusing.

bitdivine: I see.  Yes, sure.  I'll clarify the text.

.github/actions/build_nns_dapp/action.yaml
```diff
@@ -52,7 +52,7 @@ runs:
       with:
         repository: 'dfinity/snsdemo'
         path: 'snsdemo'
-        # Version from 2023-06-06 with dfx v0.14.1
+        # Version from 2023-06-06 with `dfx v0.14.1`
         ref: '0e61f618018db099a01b1686535fff8eff980d1c'
```
bitdivine, line 56: Interesting.  This `snsdemo` version should be pointing at `dfx.json`.

.github/actions/start_dfx_snapshot/action.yaml
bitdivine: Done.

dskloetd (approved):

PR #3268, opened 2023-09-01
Make DisburseNnsNeuronModal more consistent

dskloetd:
# Motivation

When making a normal transaction, there is a dropdown account picker.
When disbursing a neuron, there is a completely different UI for picking an account.
We want to make it more consistent by reusing `SelectDestinationAddress`.

`NnsDestinationAddress` is a modal step used in `DisburseNnsNeuronModal` to select the account.

# Changes

1. In `NnsDestinationAddress` use `SelectDestinationAddress` and a submit button instead of `NnsAddress` and `NnsSelectAccount`.
2. In `DisburseNeuronModal` use `QrWizardModal` to handle QR code scanning which is supported by `SelectDestinationAddress`.
3. Update tests and page objects to work with the new UI.
4. Implement `isChecked` and `selectOption` on `JestPageObjectElement` (which were previously only implemented for Playwright).
5. Install `@testing-library/user-event` to support using the `<select>` element in tests.

# Tests

Tested manually with dropdown, text input and QR code with and without token specified.

https://github.com/dfinity/nns-dapp/assets/122978264/8767d9d5-8f66-4728-9c4a-fa0471c1c489

image

https://github.com/dfinity/nns-dapp/assets/122978264/a73a67c4-b9a7-4409-8c75-765f2032ba12

image

# Todos

- [x] Add entry to changelog (if necessary).

PR #3277, opened 2023-09-04
Removed unused prop hash from AccountCard

dskloetd:
# Motivation

Since https://github.com/dfinity/nns-dapp/pull/3268, every component that uses `AccountCard`, sets the `hash` prop.
So there is no reason to have it.

# Changes

Remove the `hash` prop from `AccountCard` and assume it's always true.

# Tests

pass

# Todos

not necessary


mstrasinskis (approved):
Thanks!

PR #3278, opened 2023-09-04
Update snsdemo

bitdivine:
# Motivation
We would like to keep the `nns-dapp` test environment, provided by `snsdemo`, up to date.

# Changes
- Add a workflow to update the `snsdemo` commit.  Updating the `snsdemo` also requires potentially updating other values such as `IC_COMMIT` and `dfx` version.
  - Note: Updates _can_ and frequently do cause some compilation issues that need top be fixed manually, so the PRs generated by this bot will still require manual intervention.  As long as the bot updates the versions in `dfx.json`, that is the most important step.  The more we can automate the work needed after that, the better, but it is not currently possible to automate everything.

# Tests

- This workflow generated this PR: https://github.com/dfinity/nns-dapp/pull/3289
  Note: The PR is made against the branch from which the workflow was run, so the cron job will create the PR against main, or when testing the PR will be created against this development branch.  This way the PR contains just the changes created by updating the commit and not any changes in the workflow development branch.

# Todos


dskloetd:
I'm concerned that this disguises PRs that require careful manual changes and careful review as automated.

bitdivine:
@dskloetd  How does it disguise?  The PRs generated by this bot explicitly state in the PR body manual work is needed.  The way I see it is this:  IC updating is fait accompli - it updates in the wild - we cannot pretend that the NNS canister APIs are other than they are and expect other canisters to work according to our expectation.  What we must do is to face the update and make any necessary changes.  So the bot updates the API, any adaption we still have to do ourselves, as the proposal body states explicitly.

CHANGELOG-Nns-Dapp.md
```diff
@@ -33,6 +33,12 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 ### Operations
 
 #### Added
+* Add "Filter proposals by Votable only" e2e test.
+
+* `--import-from-index-html` flag on `scripts/canister_ids` to get canister IDs from an existing (testnet) release.
+* A dictionary for spell-checking.
+* New "finalizing" status in SNS project detail page.
+* Cron job to update `snsdemo` weekly.
```
dskloetd, line 41: Looks like something went wrong. I see 5 new entries here.

dskloetd:
I think I just don't really get how this will work.
A bot creates a PR but the PR is incomplete so requires extra work. Who will do the extra work?
And why does it have to happen when the bot decides?
Does it perhaps make more sense to have a script that prepares the changes that whoever is going to do the extra work can call at their convenience, rather than from a workflow on a schedule?

Does [the example PR](https://github.com/dfinity/nns-dapp/pull/3278) also still need more changes? Could you add those changes with an additional commit so I can see the difference between what the bot/script does vs. the extra changes required?

dskloetd:
After a long offline discussion my opinion is now that the snsdemo release and the IC commit should be updated in separate PRs.

bitdivine:
I would rather have the .did files match the test environment.  I don't see that there is significant benefit in turning an snsdemo update into a fuzz test.  If one wanted to test the hypothesis that small differences in canister version break the nns-dapp, one could test that separately and consistently.

dskloetd:
Separate changes should ideally be separate PRs.
The other points are just extra benefit.
Yes, we can do better things to get those benefits but that's never an argument against doing separate things in separate PRs.
The point is just that there is no reason to want them to always be in sync at the PR level because they aren't in mainnet either.

dskloetd:
FYI: There is also an IC commit in Cargo.toml which defines the types we use to parse proposals in the nns dapp canister:
https://github.com/dfinity/nns-dapp/blob/main/rs/backend/Cargo.toml

and we also import did files into ic-js, which is how the nns-dapp frontend communicates with other canisters:
https://github.com/dfinity/ic-js/blob/main/scripts/import-candid

PR #3279, opened 2023-09-05
Bump ic-js next

peterpeterparker:
# Motivation

We released a version of ic-js so this PR integrates the next version (numbers) of ic-js.


mstrasinskis (approved):
Rubber stamp. Thanks!

PR #3281, opened 2023-09-06
Get the snsdemo version from dfx.json

bitdivine:
# Motivation
Most tool versions are specified in `dfx.json`.  The `snsdemo` commit has hitherto not been standardized.  The standardization is now forced on us as otherwise the bot that updates the `snsdemo` cannot update the commit.  The bot does not have permission to update the workflows but can update versions in `dfx.json`.

# Changes
- Specify the `snsdemo` commit in `dfx.json`.
- Use the commit in `dfx.json`.

# Tests
- See CI

# Todos

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -638,6 +638,7 @@
         "IDL2JSON_VERSION": "0.8.8",
         "OPTIMIZER_VERSION": "0.3.6",
         "DIDC_VERSION": "2023-07-25",
+        "SNSDEMO_REF": "0e61f618018db099a01b1686535fff8eff980d1c",
         "IC_COMMIT": "06f339b83ce37e3fc9571e1b4251fbcf5c1a8239"
       },
       "packtool": ""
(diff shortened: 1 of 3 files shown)


bitdivine:
Note: Discussed options with @dskloetd , including asking for more power for the gix bot so that it can update workflows.  This seems the preferable option.

dfx.json
```diff
@@ -638,6 +638,7 @@
         "IDL2JSON_VERSION": "0.8.8",
         "OPTIMIZER_VERSION": "0.3.6",
         "DIDC_VERSION": "2023-07-25",
+        "SNSDEMO_REF": "0e61f618018db099a01b1686535fff8eff980d1c",
```
dskloetd, line 641: It was called `ref` because that's what the parameter to `actions/checkout@v3` is called. But maybe in this context `_COMMIT` is clearer?

dskloetd (approved):

dfx.json
bitdivine: I don't mind.  Technically entering say a tag works fine here, as well, which is why I used the more general term.  I'll change it.  We can change it again if we ever want to use a different sort of ref.

PR #3282, opened 2023-09-06
Spellcheck - another small step

bitdivine:
# Motivation
Another small step towards having spell checking in the nns-dapp repo.

# Changes
* Add scripts to run spell checks.
* Fix a few of the spelling mistakes.

# Tests
None.

# Todos


scripts/spellcheck-changelog
```diff
@@ -0,0 +1,27 @@
...
+}
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
+for file in CHANGELOG-*; do
+  unreleased="$file.unreleased.md" # File with unreleased changelog entries.  Must end in `.md`.
```
nmattia, line 23: suggestion: use `mktemp`?

scripts/spellcheck-md
```diff
@@ -0,0 +1,42 @@
...
+
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
+# shellcheck disable=SC2120 # Filters will be passed to match the fmt and lint commands.
+list_files() {
+  git ls-files "${@}" | filter
```
nmattia, line 28: how are the arguments supposed to be user (`${@}`)? If unused, then you could do this instead:
```
git ls-files --- '*.md'
```

scripts/spellcheck-md
```diff
@@ -0,0 +1,42 @@
...
+# Selects eligible files; filenames are read from stdin, one line per filename.
+filter() {
+  grep -E '[.]md$' | grep -v CHANGELOG
+}
+
+EXIT_CODE=0
+readarray -t files < <(list_files | filter)
+for filename in "${files[@]}"; do
+  test -s "$filename" || # Skip empty files
+    cargo spellcheck check "$filename" || EXIT_CODE=1
```
nmattia, line 40: nit: `if` would be clearer here

scripts/spellcheck-md
```diff
@@ -0,0 +1,42 @@
...
+  git ls-files "${@}" | filter
+}
+
+# Selects eligible files; filenames are read from stdin, one line per filename.
+filter() {
+  grep -E '[.]md$' | grep -v CHANGELOG
+}
+
+EXIT_CODE=0
+readarray -t files < <(list_files | filter)
```
nmattia, line 37: suggestion: you could also pipe everything directly into `while read ...` (since you don't seem to be using `files` elsewhere):

```bash
    while read -r filename
    do
      ...
    done < <(list_files | filter)
```

nmattia (approved):
LGTM, couple of suggestions

scripts/spellcheck-md
bitdivine: Cool, I did not know that.  In similar commands, but not yet implemented here, we allow the user to pass in filters such as `-m` to check just the modified files, or specific files or directories.  We will still need the `${@}` then.  We do also still need the filter  to exclude the changelogs, however we could use the `-- '*.md'` so that `filter` doesn't  have to apply that filter.

Testing:
* I can confirm that `git ls-files -m -- '*.md'` works; if I change `README.md` it returns just that.
* If I filter by paths with e.g. `git ls-files frontend/ README.md -- '*.md'` it unfortunately lists non-markdown files under frontend as well.  It looks as if it is just another pattern to match against, it doesn't filter.  If we wanted to use git wildcards we would have to convert that into something like `git ls-files frontend/**/*.md README.md`

Let me keep this wildcarding in mind, read the docs and think through the possibilities, repercussions and how that would fit into these spelling and linting commands. It's definitely nice to know but I'm not sure how well it fits into this use case.

scripts/spellcheck-md
bitdivine: Unfortulately not with bash as `EXIT_CODE` needs to be updated.  In bash, in a pipeline, the first part of the pipeline runs in the current shell, others run in subshells that can use but not change values in the parent shell.  In ksh it is the other way around: The last part of a pipeline runs in the current shell, which is much more useful!

Bash:
```
$ echo 1 | read -r foo ; echo $foo

$
```
Ksh:
```
$ echo 1 | read -r foo ; echo $foo
1
$ 
```

If writing a script that needs to self-detect which shell it is running, while being super subtle about it, this is one difference you can use to split up the possibilities.

scripts/spellcheck-md
bitdivine: Actually, on re-reading the original is wrong anyway, so the pattern of a sequence of `||` until we hit a truthy return value doesn't apply.  Fixed.

scripts/spellcheck-changelog
bitdivine: Good idea.  Changing the filename back becomes slightly more complicated but it is more robust with mktemp.

scripts/spellcheck-md
bitdivine: I propose to merge this and follow up with the flags that are planed to be supported.  Then we can then look at the options more closely.

PR #3283, opened 2023-09-06
Apply pre-migration limits

bitdivine:
# Motivation
We are about to migrate the account storage.  The migration is affected by the number of accounts so we want to make sure we don't have too many.

# Changes
- Prevent a large number of accounts from being created pre-migration.

# Tests
None.

# Todos

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -27,6 +27,10 @@ use schema::{proxy::AccountsDbAsProxy, AccountsDbBTreeMapTrait, AccountsDbTrait}
 
 type TransactionIndex = u64;
 
+/// The data migration is more complicated if there are too many accounts.  With below this many
+/// accounts we avoid some complications.
+const PRE_MIGRATION_LIMIT: u64 = 220_000;
+
 /// Accounts, transactions and related data.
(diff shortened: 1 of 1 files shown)


rs/backend/src/accounts_store.rs
```diff
@@ -1513,6 +1519,14 @@ impl AccountsStore {
             _ => {}
         };
     }
+    fn assert_pre_migration_limit(&self) {
+        let db_accounts_len = self.accounts_db.db_accounts_len();
+        assert!(
+            db_accounts_len < PRE_MIGRATION_LIMIT,
+            "Pre migration account limit exceeded {}",
```
peterpeterparker, line 1526: Shouldn't we return a distinctive error message that way the frontend can interpret it easily and display gracefully an explanation of the error to the users?

peterpeterparker: Side note: I'm guessing we `trap` because we do not want to change the interface and return a `Result`, correct?

bitdivine: Potentially, yes, but we have time to implement that.

peterpeterparker (approved):
LGTM, thx

PR #3287, opened 2023-09-07
Add tests for SetDissolveDelay component

dskloetd:
# Motivation

Currently we don't allow setting a dissolve delay on a neuron that would not result in voting power because the delay is too short.
We want to start allowing this.

But currently `SetDissolveDelay` is very complicated and also has several bugs:

1. Error messages are only updated on text input and not on slider input.
2. When the current dissolve delay is not a whole number of days, increasing it by less than a day to a whole number of days is allowed. But we still show an error message that it's not allowed even though the button is enabled.
3. The slider has second granularity but we only display days so we might display X days but that will not actually be the dissolve delay you get.

I want to simplify the code and fix the bugs but I want to make sure I don't break the component.

But currently the component has very limited test coverage.

So this PR increases the test coverage and adds comments about tests that had to be changed because the bugs need to be fixed first.

# Changes

1. Add and use page object for `InputRange`.
2. Fix a bug in `frontend/src/tests/page-objects/InputWithError.page-object.ts` where it treated a `Promise<boolean>` as a `boolean` which was always truthy, even if the promise would resolve to `false`.
3. Add more functionality to `SetDissolveDelayPo` page object.
4. Add tests to `frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts`

# Tests

test only

# Todos

not necessary

### DIFF
--- a/frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
+++ b/frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
@@ -3,40 +3,276 @@
  */
 import SetDissolveDelay from "$lib/components/neurons/SetDissolveDelay.svelte";
 import {
+  SECONDS_IN_DAY,
   SECONDS_IN_EIGHT_YEARS,
   SECONDS_IN_HALF_YEAR,
 } from "$lib/constants/constants";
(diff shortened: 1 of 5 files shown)


mstrasinskis (approved):
LGTM, thanks!

PR #3290, opened 2023-09-07
E2E: Wait for proposal card instead of skeleton card

dskloetd:
# Motivation

Once again we had an [e2e flake](https://github.com/dfinity/nns-dapp/actions/runs/6110153913/job/16582777946?pr=3280) because no proposal cards were present even though the skeleton cards had disappeared.
So I want to try a different approach for `waitForContentLoaded`.

# Changes

1. Instead of waiting for the skeleton card to go away, wait for either at least one proposal card or the "no proposals" message to be present.
2. Support passing a `tid` parameter to `waitFor` to specify a specific child element, same as with `isPresent`, `click`, etc.

# Tests

test only

# Todos

not necessary


mstrasinskis (approved):
Nice,thank you!

PR #3291, opened 2023-09-07
Snsdemo release

bitdivine:
# Motivation
The snsdemo version is currently specified as a commit in `dfx.json`.  Unfortunately this version is not applied to the snsdemo snapshot.

# Changes
- Use a release tag to specify which version the nns-dapp should use, instead of a commit.
- Use the release tag to get the sns snapshot.

# Tests
See CI.

# Todos

  I think this is too minor a tweak to warrant a chnagelog entry.  No frontend or canister changes and only a minor operations tweak.

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -638,7 +638,7 @@
         "IDL2JSON_VERSION": "0.8.8",
         "OPTIMIZER_VERSION": "0.3.6",
         "DIDC_VERSION": "2023-07-25",
-        "SNSDEMO_COMMIT": "0e61f618018db099a01b1686535fff8eff980d1c",
+        "SNSDEMO_RELEASE": "release-2023-07-07",
         "IC_COMMIT": "06f339b83ce37e3fc9571e1b4251fbcf5c1a8239"
       },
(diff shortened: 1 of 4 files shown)


dfx.json
```diff
@@ -638,7 +638,7 @@
         "IDL2JSON_VERSION": "0.8.8",
         "OPTIMIZER_VERSION": "0.3.6",
         "DIDC_VERSION": "2023-07-25",
-        "SNSDEMO_COMMIT": "0e61f618018db099a01b1686535fff8eff980d1c",
+        "SNSDEMO_RELEASE": "release-2023-07-07",
```
bitdivine, line 641: Note: The commit on the left is a bit older than the snapshot:
```
Tue Jun 6 08:41:19 2023 +0200
```

bitdivine:
I think the most efficient way of merging this to main is to cherry-pick the change across to main.  This way the changes are identical to those in this PR and unaffected by any tweaks to the automated updates.  If we have another automated update in say a week, any changes to the patch file are likely to be small.

PR #3292, opened 2023-09-08
Fix and simplify SetDissolveDelay.svelte

dskloetd:
# Motivation

`SetDissolveDelay.svelte` has several bugs:
1. Error messages are only updated on text input and not on slider input.
2. When the current dissolve delay is not a whole number of days, increasing it by less than a day to a whole number of days is allowed. But we still show an error message that it's not allowed even though the button is enabled.
3. The slider has second granularity but we only display days so we might display X days but that will not actually be the dissolve delay you get.

It is also overly complicated because:
1. The slider has granularity in seconds while the text input has granularity in days and both numbers need to be kept in sync.
2. Disabling the button has logic based on seconds while the error message has separate logic based on days.
3. There is a `minDelayInSeconds` prop which is not actually the minimum allowed value but rather all allowed values are strictly greater than this value.

# Changes

1. Use days for all logic. And only convert back to seconds (as a whole multiple of the number of seconds in a day) to export the output prop.
2. Determine whether the button is disable based on the error message rather than separate logic.
3. Use the `neuronDissolveDelaySeconds` for the lower bound and remove the redundant `minDelayInSeconds` prop.
4. Update the error message also on slider input.
5. Update the tests to use days for the range input instead of seconds.
6. Fix the tests that were forced to have wrong expectations because of the existing bugs.
7. Add missing test coverage for min/max buttons and prop syncing.

# Tests

Unit tests added and updated.
Also tested manually.

# Todos

### DIFF
--- a/frontend/src/lib/components/neurons/SetDissolveDelay.svelte
+++ b/frontend/src/lib/components/neurons/SetDissolveDelay.svelte
@@ -1,24 +1,26 @@
 <script lang="ts">
   import { createEventDispatcher } from "svelte";
   import { i18n } from "$lib/stores/i18n";
-  import { daysToSeconds, secondsToDays } from "$lib/utils/date.utils";
+  import {
+    daysToSeconds,
+    secondsToDays,
+    secondsToDaysRoundedDown,
+  } from "$lib/utils/date.utils";
   import { formatToken } from "$lib/utils/token.utils";
   import { formatVotingPower } from "$lib/utils/neuron.utils";
   import { replacePlaceholders } from "$lib/utils/i18n.utils";
   import { InputRange, Html } from "@dfinity/gix-components";
-  import { isDefined, valueSpan } from "$lib/utils/utils";
+  import { valueSpan } from "$lib/utils/utils";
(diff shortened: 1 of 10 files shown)


frontend/src/lib/components/neurons/SetDissolveDelay.svelte
```diff
@@ -27,66 +29,71 @@
...
   let minDelayInDays = 0;
-  $: minDelayInDays = secondsToDays(minDelayInSeconds);
+  $: minDelayInDays = secondsToDays(Number(neuronDissolveDelaySeconds) + 1);
 
   let minProjectDelayInDays = 0;
   $: minProjectDelayInDays = secondsToDays(minProjectDelayInSeconds);
 
   let maxDelayInDays = 0;
-  $: maxDelayInDays = secondsToDays(maxDelayInSeconds);
+  $: maxDelayInDays = secondsToDaysRoundedDown(maxDelayInSeconds);
```
mstrasinskis, line 45: Nice 👍

mstrasinskis (approved):
LGTM, thanks!

PR #3298, opened 2023-09-11
GIX-1862 Allow setting shorter dissolve delays

dskloetd:
# Motivation

Even though a neuron requires a certain dissolve delay to have voting power, it's allowed to set shorter dissolve delays as well.
But currently this is not possible in the NNS dapp UI. We want to make it possible.

# Changes

1. Change the error message when the dissolve delay is shorter than what's required for voting power, to make it clear that the neuron won't have voting power.
2. If the selected dissolve delay is shorter than the current dissolve delay AND shorter than what's required for voting power, show the error about the current dissolve delay because it's now stricter as it actually disables setting the dissolve delay.
3. Don't disable the update button if the dissolve delays is not enough to have voting power.
4. Update the tests.

# Tests

Unit tests have been updated.
Manually tested at: https://fswcu-uqaaa-aaaaa-aadwq-cai.nnsdapp.dfinity.network/

# Todos


lmuntaner (approved):
LGTM! Thanks

PR #3301, opened 2023-09-12
Feat: Add metadata logo path to SNS aggregator data

lmuntaner:
# Motivation

The logo to the SNS Metadata needs to be hardcoded by the frontend following a specific pattern.

In this PR, I add the relative path of the logo so that the logic of where the logo lives lies only in the SNS aggregator canister.

Another reason is because the logo is optional. Therefore, this way it's easier to know whether the logo is present or not. Otherwise, we would need to try to fetch the asset and check it.

# Changes

* New field `logo` in `SlowMetadata`.
* Change `SlowMetadata::from` to use `UpstreamData` to access the root canister id.

# Tests

Tests are skipped.

# Todos

### DIFF
--- a/rs/sns_aggregator/src/types/slow.rs
+++ b/rs/sns_aggregator/src/types/slow.rs
@@ -44,13 +44,15 @@ pub struct SlowSnsData {
     pub lifecycle: Option<GetLifecycleResponse>,
 }
 
+/// Transforms the canister response GetMetadataResponse into the Aggregator data SlowMetadata.
+/// It substitutes the logo in base 64 by the path to the logo asset if present.
 pub fn from_get_metadata_response(upstream: &GetMetadataResponse, root_canister_id: Option<Principal>) -> SlowMetadata {
     SlowMetadata {
         url: upstream.url.clone(),
         name: upstream.name.clone(),
         description: upstream.description.clone(),
         logo: match (upstream.logo.clone(), root_canister_id) {
-            (Some(_), Some(canister_id)) => Some(format!("/sns/root/{}/logo.png", canister_id.to_string())),
+            (Some(_), Some(canister_id)) => Some(format!("/sns/root/{}/logo.png", canister_id.to_text())),
             _ => None,
         },
     }
--- a/rs/sns_aggregator/src/types/slow.rs
+++ b/rs/sns_aggregator/src/types/slow.rs
@@ -1,5 +1,5 @@
 //! Slowly changing information about an SNS
-use crate::types::ic_sns_governance::{GetMetadataResponse, ListNervousSystemFunctionsResponse};
+use crate::types::ic_sns_governance::ListNervousSystemFunctionsResponse;
 use crate::types::ic_sns_root::ListSnsCanistersResponse;
 use crate::types::ic_sns_swap::{
     DerivedState, GetDerivedStateResponse, GetInitResponse, GetLifecycleResponse, GetSaleParametersResponse,
@@ -8,7 +8,7 @@ use crate::types::ic_sns_swap::{
 use crate::types::ic_sns_wasm::DeployedSns;
 use crate::types::upstream::UpstreamData;
 use crate::Icrc1Value;
-use candid::{CandidType, Nat, Principal};
+use candid::{CandidType, Nat};
 use num_traits::ToPrimitive;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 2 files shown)


rs/sns_aggregator/src/types/slow.rs
```diff
@@ -77,16 +89,8 @@ pub struct SlowMetadata {
     /// Same as upstream.
     /// Note: The description can also be quite long and isn't needed for the initial view of all the SNSs
     pub description: Option<String>,
-}
-
-impl From<&GetMetadataResponse> for SlowMetadata {
```
peterpeterparker, line 82: Instead of a new function that takes two parameters why don't you transform this `impl` into a `impl From<&UpstreamData> for SlowMetadata`?

peterpeterparker: Curiosity question :)

lmuntaner: I couldn't pass the extra parameter `rootCanisterId`. I also preferred that, but that didn't like it.

I wanted to ask @bitdivine about it. That's also why this is in Draft yet.

peterpeterparker: > but that didn't like it.

Ah it did not build, weird. Thanks for the feedback. Again was strictly a curiosity question.

rs/sns_aggregator/src/types/slow.rs
```diff
@@ -44,13 +44,27 @@ pub struct SlowSnsData {
...
+}
+
 impl From<&UpstreamData> for SlowSnsData {
     fn from(upstream: &UpstreamData) -> Self {
         SlowSnsData {
             index: upstream.index,
             canister_ids: upstream.canister_ids.clone(),
             list_sns_canisters: upstream.list_sns_canisters.clone(),
-            meta: SlowMetadata::from(&upstream.meta),
+            meta: from_get_metadata_response(&upstream.meta, upstream.list_sns_canisters.root),
```
lmuntaner, line 67: @bitdivine the following was not possible, it didn't compile:

```
SlowMetadata::from(&upstream.meta, upstream.list_sns_canisters.root),
```

It didn't like to have a second parameter. Was I doing something wrong?

rs/sns_aggregator/src/types/slow.rs
```diff
@@ -44,13 +44,27 @@ pub struct SlowSnsData {
...
 
+/// Transforms the canister response GetMetadataResponse into the Aggregator data SlowMetadata.
+/// It substitutes the logo in base 64 by the path to the logo asset if present.
+pub fn from_get_metadata_response(upstream: &GetMetadataResponse, root_canister_id: Option<Principal>) -> SlowMetadata {
+    SlowMetadata {
+        url: upstream.url.clone(),
+        name: upstream.name.clone(),
+        description: upstream.description.clone(),
+        logo: match (upstream.logo.clone(), root_canister_id) {
+            (Some(_), Some(canister_id)) => Some(format!("/sns/root/{}/logo.png", canister_id.to_text())),
```
lmuntaner, line 55: @bitdivine should we add the version here?

I there a way to ensure that this matches where the asset is stored?

lmuntaner:
@bitdivine please review

I added some questions / comments in the code changes.

rs/sns_aggregator/src/types/slow.rs
peterpeterparker: It is indeed not possible to had a `::from` with two parameters but, I think you can have such an implementation:

```
impl From<&UpstreamData> for SlowMetadata {
   fn from(upstream: &UpstreamData) -> Self {
       SlowMetadata {
          url: upstream.meta.url.clone(),
          name: upstream.meta.name.clone(),
          description: upstream.meta.description.clone(),
          logo: match (upstream.meta.logo.clone(), upstream.root_canister_id) {
            (Some(_), Some(canister_id)) => Some(format!("/sns/root/{}/logo.png", canister_id.to_text())),
            _ => None,
        },
```

lmuntaner: oh, good point!

rs/sns_aggregator/src/types/slow.rs
lmuntaner: @peterpeterparker what do you think about this? Should we add the version in this path as well? Or is that not part of it do you think?

peterpeterparker: I don't have opinion. I just randomly spotted your PR and was only curious about why you implemented a new function instead of modifying the impl stricly curiosity about the code because I'm a Rust noob.

lmuntaner: But as a developer, would you expect the relative path to include the version?

peterpeterparker: I would maybe expect the absolute path (`https://....raw.icp0.io/sns/......jpg`). I'm out of the loop so do not mind me, really was just curiosity about the code 😃.

rs/sns_aggregator/src/types/slow.rs
```diff
@@ -77,14 +77,20 @@ pub struct SlowMetadata {
...
+    fn from(upstream: &UpstreamData) -> Self {
         SlowMetadata {
-            url: upstream.url.clone(),
-            name: upstream.name.clone(),
-            description: upstream.description.clone(),
+            url: upstream.meta.url.clone(),
+            name: upstream.meta.name.clone(),
+            description: upstream.meta.description.clone(),
+            logo: match (upstream.meta.logo.clone(), upstream.list_sns_canisters.root) {
+                (Some(_), Some(canister_id)) => Some(format!("/sns/root/{}/logo.png", canister_id.to_text())),
```
bitdivine, line 91: Please note that the full URL path has a version number, such as `v1`, at the beginning:
```
https://3r4gx-wqaaa-aaaaq-aaaia-cai.ic0.app/v1/sns/root/u67kc-jyaaa-aaaaq-aabpq-cai/logo.png
```
Maybe this could be added as a comment?

bitdivine (approved):
LGTM, thanks!

rs/sns_aggregator/src/types/slow.rs
lmuntaner: Yes, but for that we would also need to know the domain from within the sns aggregator. That's a bigger task. Maybe we can expand it.

rs/sns_aggregator/src/types/slow.rs
lmuntaner: I'll add the v1 in the relative path and add this comment, thanks!

PR #3306, opened 2023-09-12
GIX-1878: Fix TVL text color

lmuntaner:
# Motivation

TVL title text is not clearly visible in light theme. See screenshot of "Before".

The reason is that `--text2-color` was not present in the light theme.

The solution is to use `--description-color`. `--text2-color` will be removed because it's not used anywhere else.

# Changes

* Change the color of TotalValueLocked to `--description-color`.

# Tests

* Only CSS changes

# Todos


lmuntaner:
Before:

Screenshot 2023-09-12 at 13 43 46

Now:

Screenshot 2023-09-12 at 14 18 54
Screenshot 2023-09-12 at 14 19 01

peterpeterparker (approved):
Thx

PR #3307, opened 2023-09-12
GIX-1872 More and readable tests for secondsToDuration and daysToDuration

dskloetd:
# Motivation

When you set a dissolve delay, the way the duration is rendered between the different steps is often off by 1 day.
This is because we use `daysToDuration` and `secondsToDuration` which have different inconsistent implementations.
Especially `secondsToDuration` seem wrong because of how it deals with leap days.
This was not caught by tests because the tests are very lenient and only expect the right words (not numbers) to be present in the rendered strings.
Before fixing this I want to make sure there are good tests in place so it will be clear how the behavior changes when the code is fixed.

This PR adds tests (often with wrong expectations) which show what is currently implemented.

# Changes

Change the tests to expect that actual output strings and add a lot of tests close to important boundaries to clearly demonstrate where the code is broken and to catch future bugs after the code is fixed.

# Tests

yes

# Todos

not necessary


lmuntaner (approved):
Thanks!

Could you shortly write the expected somewhere that we can discuss it?

dskloetd:
Shall we discuss on the ticket? https://dfinity.atlassian.net/browse/GIX-1872
The issue I'm trying to fix is described there.

The only thing I really expect is to render the same duration in the same way in different places in the UI.
Everything else I would try to do as it seems it was intended but I don't think it matters that much.

lmuntaner:
Sounds good, I started watching the issue.

PR #3310, opened 2023-09-12
GIX-1872 Make secondsToDuration consistent with daysToDuration

dskloetd:
# Motivation

We use both `secondsToDuration` and `daysToDuration` to render the same time period so they should be consistent but currently they are not.

This PR fixes https://dfinity.atlassian.net/browse/GIX-1872.
It does not aim to fix https://dfinity.atlassian.net/browse/GIX-1770 which is still being discussed.
I think it's worthwhile to fix GIX-1872 and put a test in place that the rendering is consistent before we have a decision on GIX-1770.

# Changes

1. Reimplement `secondsToDuration`.
2. Update tests for `secondsToDuration` for the new implementation.
3. Test that both functions return the same string for periods that are a whole number of days between 1 and 3000.

# Tests

Extensive unit test are updated.

# Todos

### DIFF
--- a/frontend/src/lib/utils/date.utils.ts
+++ b/frontend/src/lib/utils/date.utils.ts
@@ -23,6 +23,7 @@ const createLabel = (labelKey: LabelKey, amount: bigint): LabelInfo => ({
 // Returns how many days there are in the given number of years, adding a leap
 // day for every 4 years.
 const daysInYears = (years: bigint): bigint => {
+  // Use integer division.
   const leapDays = years / BigInt(4);
   return years * BigInt(DAYS_IN_NON_LEAP_YEAR) + leapDays;
 };
@@ -30,6 +31,7 @@ const daysInYears = (years: bigint): bigint => {
 // Returns how many full years, requiring a leap day for every 4 full years,
 // there are in the given number of days.
 const fullYearsInDays = (days: bigint): bigint => {
+  // Use integer division.
   let years = days / BigInt(DAYS_IN_NON_LEAP_YEAR);
   while (daysInYears(years) > days) {
     years--;
--- a/frontend/src/lib/utils/date.utils.ts
+++ b/frontend/src/lib/utils/date.utils.ts
@@ -1,6 +1,8 @@
 import {
+  DAYS_IN_NON_LEAP_YEAR,
+  HOURS_IN_DAY,
+  MINUTES_IN_HOUR,
   SECONDS_IN_DAY,
-  SECONDS_IN_HOUR,
   SECONDS_IN_MINUTE,
   SECONDS_IN_MONTH,
   SECONDS_IN_YEAR,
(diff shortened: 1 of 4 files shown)


frontend/src/lib/utils/date.utils.ts
```diff
@@ -18,13 +20,37 @@ const createLabel = (labelKey: LabelKey, amount: bigint): LabelInfo => ({
   amount: Number(amount),
 });
 
+// Returns how many days there are in the given number of years, adding a leap
+// day for every 4 years.
+const daysInYears = (years: bigint): bigint => {
+  const leapDays = years / BigInt(4);
```
lmuntaner, line 26: Are you relying on bigint division returning always an integer?

Maybe add a comment? It might seem a mistake otherwise.

lmuntaner (commented):
Why didn't you use the `daysToDuration` within `secondsToDuration`?

Don't we want that?

frontend/src/lib/utils/date.utils.ts
dskloetd: Ehm, to me this is normal integer division. I'm always confused when JavaScript doesn't do this by default. :)
You are only allowed to do arithmetic with `bigint`s if both sides are `bigint` and the result will also be `bigint`.
The previous code relies heavily on this as well, but if you think it might be confusing, I'll add a comment.

frontend/src/lib/utils/date.utils.ts
```diff
@@ -18,13 +20,37 @@ const createLabel = (labelKey: LabelKey, amount: bigint): LabelInfo => ({
...
+// there are in the given number of days.
+const fullYearsInDays = (days: bigint): bigint => {
+  let years = days / BigInt(DAYS_IN_NON_LEAP_YEAR);
+  while (daysInYears(years) > days) {
+    years--;
+  }
+  return years;
+};
+
 export const secondsToDuration = (seconds: bigint): string => {
```
dskloetd, line 40: > Why didn't you use the daysToDuration within secondsToDuration?

Do you mean the opposite? Use `secondsToDuration` within `daysToDuration`?
I think that makes sense later but here I wanted to make it clear that `secondsToDuration` is consistent with `daysToDuration` without changing `daysToDuration`. So that it's completely clear why the behavior is like this.
Happy to make `daysToDuration` call `secondsToDuration` in a follow-up PR.

If you really did mean "use `daysToDuration` within `secondsToDuration`", I'm sure how you imagine that exactly.

lmuntaner: Oops, yes, I meant "use secondsToDuration within daysToDuration".

Ok in another PR. I just assumed that's how you were going to make them work the same way.

Specially, if we fix a bug in one, it's fixed in both.

lmuntaner (approved):

PR #3312, opened 2023-09-13
Fix: Partial different derived state in snsQueryStore

lmuntaner:
# Motivation

Part of the response of `get_state` is the derived state, which we can also get from `get_derived_state`. Yet, the endpoint returns different fields and different precision for one of the fields (`sns_tokens_per_icp`).

When we were populating the snsQueryStore in the service `loadSnsProjects`, we were using first the data from the `get_state`, and right after we were updating it with the data from `get_derived_state`.

This caused that the sns summary from snsQueryStore and from snsAggregatorStore didn't match for some time.

# Changes

* Use the data from `get_derived_state` in the first place instead of updating it afterward.

# Tests

* The tests didn't cover the moment when the stores were different because it happens in the middle of the function. Yet, when I removed the call to `updateDerivedState`, the test failed. Therefore, it was checking that both derived states from snsAggregator and snsQueryStore where the same. Then, I fixed the test by changing the value in the field `derived` for the snsQueryStore. 

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/services/$public/sns.services.ts
+++ b/frontend/src/lib/services/$public/sns.services.ts
@@ -22,7 +22,7 @@ import { convertDtoData } from "$lib/utils/sns-aggregator-converters.utils";
 import { ProposalStatus, Topic, type ProposalInfo } from "@dfinity/nns";
 import { Principal } from "@dfinity/principal";
 import type { SnsNervousSystemFunction } from "@dfinity/sns";
-import { nonNullish, toNullable } from "@dfinity/utils";
+import { fromNullable, nonNullish, toNullable } from "@dfinity/utils";
 import { get } from "svelte/store";
 import { getCurrentIdentity } from "../auth.services";
 
@@ -77,7 +77,18 @@ export const loadSnsProjects = async (): Promise<void> => {
         ),
         indexCanisterId: Principal.fromText(sns.canister_ids.index_canister_id),
         swap: toNullable(sns.swap_state.swap),
-        derived: toNullable(sns.swap_state.derived),
+        // The endpoint `get_state` and `get_derived_state` return different fields and decimal precision.
+        // We want to use the info in `derived_state` immediately
+        // instead of changing it afterwards to avoid having a partial different data in the snsQueryStore.
+        derived: toNullable({
+          sns_tokens_per_icp:
+            fromNullable(sns.derived_state.sns_tokens_per_icp) ?? 0,
+          buyer_total_icp_e8s:
+            fromNullable(sns.derived_state.buyer_total_icp_e8s) ?? 0n,
+          cf_participant_count: sns.derived_state.cf_participant_count,
+          direct_participant_count: sns.derived_state.direct_participant_count,
+          cf_neuron_count: sns.derived_state.cf_neuron_count,
+        }),
       })),
     ];
     snsQueryStore.setData(snsQueryStoreData);
@@ -123,14 +134,6 @@ export const loadSnsProjects = async (): Promise<void> => {
           }),
           {} as TokensStoreData
         )
-    );
-    cachedSnses.forEach(
-      ({ canister_ids: { root_canister_id }, derived_state }) => {
-        snsQueryStore.updateDerivedState({
-          rootCanisterId: root_canister_id,
-          derivedState: derived_state,
-        });
-      }
     );
     // TODO: PENDING to be implemented, load SNS parameters.
   } catch (err) {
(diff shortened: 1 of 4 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/services/$public/sns.services.ts
```diff
@@ -77,7 +77,18 @@ export const loadSnsProjects = async (): Promise<void> => {
         ),
         indexCanisterId: Principal.fromText(sns.canister_ids.index_canister_id),
         swap: toNullable(sns.swap_state.swap),
-        derived: toNullable(sns.swap_state.derived),
+        // The endpoint `get_state` and `get_derived_state` return different fields and decimal precision.
+        // We want to use the info in `derived_state` immediately
+        // instead of changing it afterwards to avoid having a partial different data in the snsQueryStore.
+        derived: toNullable({
+          sns_tokens_per_icp:
+            fromNullable(sns.derived_state.sns_tokens_per_icp) ?? 0,
```
dskloetd, line 85: Why do we treat `null` as `0`? Don't they mean different things?

frontend/src/lib/services/$public/sns.services.ts
```diff
@@ -77,7 +77,18 @@ export const loadSnsProjects = async (): Promise<void> => {
         ),
         indexCanisterId: Principal.fromText(sns.canister_ids.index_canister_id),
         swap: toNullable(sns.swap_state.swap),
-        derived: toNullable(sns.swap_state.derived),
+        // The endpoint `get_state` and `get_derived_state` return different fields and decimal precision.
```
dskloetd, line 80: The comment speaks of `get_state` and `get_derived_state` but in the code I see `sns.swap_state.derived` (old) and `sns.derived_state` (new) but no mention of `get_state` and `get_derived_state` so it's not really clear to me how the comment relates to the code below or what the code is trying to clarify.

frontend/src/lib/services/$public/sns.services.ts
```diff
@@ -77,7 +77,18 @@ export const loadSnsProjects = async (): Promise<void> => {
...
+        // instead of changing it afterwards to avoid having a partial different data in the snsQueryStore.
+        derived: toNullable({
+          sns_tokens_per_icp:
+            fromNullable(sns.derived_state.sns_tokens_per_icp) ?? 0,
+          buyer_total_icp_e8s:
+            fromNullable(sns.derived_state.buyer_total_icp_e8s) ?? 0n,
+          cf_participant_count: sns.derived_state.cf_participant_count,
+          direct_participant_count: sns.derived_state.direct_participant_count,
+          cf_neuron_count: sns.derived_state.cf_neuron_count,
+        }),
```
dskloetd, line 91: Do/should we have a converter function that converts `sns.derived_state` to this object?

frontend/src/lib/services/$public/sns.services.ts
```diff
@@ -124,14 +135,6 @@ export const loadSnsProjects = async (): Promise<void> => {
...
         )
     );
-    cachedSnses.forEach(
-      ({ canister_ids: { root_canister_id }, derived_state }) => {
-        snsQueryStore.updateDerivedState({
-          rootCanisterId: root_canister_id,
-          derivedState: derived_state,
-        });
-      }
-    );
```
dskloetd, line 134: Did this old code here do the same things as the new code above? If so, why doesn't it do the same field conversion? Is that performed somewhere else? Can that be shared?

frontend/src/lib/services/$public/sns.services.ts
dskloetd, line 1: The PR description mentions a test being fixed but I don't see any test changes in the PR?

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: I'll expand the comment.

frontend/src/lib/services/$public/sns.services.ts
lmuntaner: Done

lmuntaner:

frontend/src/lib/stores/sns.store.ts
```diff
@@ -213,22 +216,13 @@ const initSnsQueryStore = (): SnsQueryStore => {
...
-        derivedState.buyer_total_icp_e8s
-      );
-      // We don't update the store if any of the derived state mandatory fields is undefined.
-      if (
-        sns_tokens_per_icp === undefined ||
-        buyer_total_icp_e8s === undefined
-      ) {
+      const newDerivedState =
+        convertDerivedStateResponseToDerivedState(derivedState);
+      // Ignore updating ths store if the mandatory fields are not present.
```
dskloetd, line 221: typo: ths

dskloetd (approved):

frontend/src/lib/stores/sns.store.ts
lmuntaner: Thanks!

PR #3313, opened 2023-09-13
Chore: Add warn log if summaries are different

lmuntaner:
# Motivation

We want to stop using snsQueryStore and use snsAggregatorStore instead.

To be sure this is safe, we will log a warning if they create different summaries.

PENDING: Merge #3312 

# Changes

* Compare summaries from snsQueryStore and snsAggregatorStore in snsSummariesStore.
* New project util `differentSummaries`.

# Tests

* Test new project util `differentSummaries`.
* Test that `snsSummariesStore` logs warning if summaires don't match.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/stores/sns.store.ts
+++ b/frontend/src/lib/stores/sns.store.ts
@@ -1,5 +1,6 @@
 import type { SnsSummary, SnsSwapCommitment } from "$lib/types/sns";
 import type { QuerySnsMetadata, QuerySnsSwapState } from "$lib/types/sns.query";
+import { differentSummaries } from "$lib/utils/projects.utils";
 import { convertDtoToSnsSummary } from "$lib/utils/sns-aggregator-converters.utils";
 import {
   convertDerivedStateResponseToDerivedState,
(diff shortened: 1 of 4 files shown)


lmuntaner:
@dskloetd please review

third and final PR to add a log when summaries don't match.

This is ready for review but tests fail because some add a warning because summaries don't match due to what's fixed in #3312

frontend/src/lib/stores/sns.store.ts
```diff
@@ -393,24 +394,50 @@ export const snsSummariesStore = derived<
     lifecycles,
     enableSnsAggregatorStore,
   ]) => {
-    if (!enableSnsAggregatorStore) {
```
dskloetd, line 396: As discussed: Keep this check up here and only do the comparison when the feature flag is enabled.

frontend/src/lib/utils/projects.utils.ts
```diff
@@ -373,3 +374,21 @@ export const participateButtonStatus = ({
...
+  summaries1: SnsSummary[],
+  summaries2: SnsSummary[]
+): SnsSummary[] =>
+  summaries1.filter((summary1) => {
+    const summary2 = summaries2.find(
+      ({ rootCanisterId }) =>
+        rootCanisterId.toText() === summary1.rootCanisterId.toText()
+    );
+    // We compare the inner fields because the order when stringifying it might be different
+    return (
```
dskloetd, line 388: `SnsSummary` has 9 fields. Why not compare the canisterIds as well just to be sure?

lmuntaner: done

lmuntaner:

PR #3315, opened 2023-09-13
Update downgrade-upgrade test

bitdivine:
# Motivation
The downgrade-upgrade test needs to be able to verify that accounts are preserved across schema updates.

# Changes
- Make it possible to specify a wasm to downgrade to.  This way the test can be run with a wasm that implements the future schema `S0.early` and another, `undefined.late` that is not yet in production.
- Get account data; this was already present but commented out as prod did not have account data in a stable, usable form.

# Tests

# Todos

### DIFF
--- a/.config/spellcheck.dic
+++ b/.config/spellcheck.dic
@@ -5,6 +5,7 @@ ICP
 schemas
 WASM
 Wasm
+Wasms
 OWASP
 <
 >
--- a/scripts/nns-dapp/downgrade-upgrade-test
+++ b/scripts/nns-dapp/downgrade-upgrade-test
@@ -3,8 +3,6 @@ set -euo pipefail
 SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 PATH="$SOURCE_DIR:$PATH"
 
-PROD_WASM="nns-dapp-prod.wasm.gz" # TODO: Switch to a test build, when available for the prod tag.
-
 help_text() {
   cat <<-EOF
 	Tests:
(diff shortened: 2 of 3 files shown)


.config/spellcheck.dic
```diff
@@ -5,6 +5,7 @@ ICP
 schemas
 WASM
 Wasm
+Wasms
```
bitdivine, line 8: Plural of Wasm.  By the way, technically, [WASM is incorrect according to the Wasm org](https://webassembly.org/) but we use it widely and that is not the worst issue we have with text.

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -32,9 +31,16 @@ get_current_wasm() {
 }
 
 get_prod_wasm() {
-  # TODO: Use the test build once available for the prod release.
   # Note: Using the test build allows us to get arbitrary test accounts from the canister.
-  curl -sL https://github.com/dfinity/nns-dapp/releases/download/prod/nns-dapp.wasm.gz >"$PROD_WASM"
+  if test -n "${PROD_WASM:-}"; then
+    test -f "${PROD_WASM}" || {
```
lmuntaner, line 36: Sorry my ignorance, but it's hard to find on the internet with this name.

What does the "test" command do? Check for presence?

lmuntaner (approved):
LGTM, thanks!

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: `test` is a bash builtin for evaluating [an expression](https://www.gnu.org/software/bash/manual/bash.html#Bash-Conditional-Expressions).  E.g. `-n` checks whether the string is non-empty and `-f` checks whether the path exists and is  a file.

PR #3317, opened 2023-09-14
GIX-1889: Fix blank modal disburse maturity

lmuntaner:
# Motivation

There was a bug when NeuronSelectPercentage was used for maturity above 1000.

That's because now we started calculating the maturity of the percentage. Yet, the maturity was passed as a string and then converted to a number. If the number was larger than 1000, then the string couldn't be converted successfully to a number.

The solution was to expect a number instead of a string.

# Changes

* Change `formttedMaturity` in NeuronSelectPercentage for `availableMaturityE8s`. Change accordingly.
* Change where the NeuronSelectPercentage was used to pass the maturity as e8s instead of as a string.
* Remove unnecessary util `maturityPercentageToE8s`. Now it's a simple multiplication and division. No need to use an util for that. There is no logic to reuse.

# Tests

* I added a test that fails in `main` but succeeds in this PR.

# Todos

### DIFF
--- a/frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
+++ b/frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
@@ -38,6 +38,7 @@ describe("SnsDisburseMaturityModal", () => {
 
   beforeEach(() => {
     authStore.setForTesting(mockIdentity);
+    jest.clearAllMocks();
   });
 
   it("should display total maturity", async () => {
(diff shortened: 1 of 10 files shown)


lmuntaner:
@dskloetd please review

My plan is to add tests for NeuronSelectPercentage in another PR.

frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
```diff
@@ -65,24 +66,32 @@ describe("SnsDisburseMaturityModal", () => {
...
+    id: [1],
+    maturity: 110_000_000_000n,
+  });
+  const neurons = [mockSnsNeuron, neuron1100maturity];
 
-    // precondition
-    expect(disburseMaturity).not.toBeCalled();
-    expect(reloadNeuron).not.toBeCalled();
+  neurons.forEach((neuron) => {
+    it("should call disburse maturity api and reloadNeuron", async () => {
```
dskloetd, line 76: Can you add 2 separate tests? One of them should make it clear in the description that it's about maturity being > 1000.
You can call a helper function from both tests to still not have code duplication without `forEach`.

lmuntaner: done

lmuntaner:

dskloetd (approved):

PR #3319, opened 2023-09-15
Add feature flag ENABLE_STAKE_NEURON_ICRC1

dskloetd:
# Motivation

We want to start using ICRC-1 for the ledger transfers on staking a neuron.
We want to test it before making it default.

# Changes

Add feature flag ENABLE_STAKE_NEURON_ICRC1. In a later PR this will be used to enable using ICRC-1 ledger transfers on staking a neuron.

# Tests

no

# Todos

I will add an entry when I actually use the flag.


dskloetd:
I haven't decided yet which other flows will use feature flags. Stake neuron is more complicated because the change is in another repo, but it probably makes sense to have a feature flag for all of them. But feature flags are cheap and if one flow works and others don't we should already release what works. It also might take longer to finish all the flows and I wouldn't want to block enabling some flows because others aren't finished yet. So I see no reason not to have multiple feature flags.

lmuntaner:
> I haven't decided yet which other flows will use feature flags. Stake neuron is more complicated because the change is in another repo, but it probably makes sense to have a feature flag for all of them. But feature flags are cheap and if one flow works and others don't we should already release what works. It also might take longer to finish all the flows and I wouldn't want to block enabling some flows because others aren't finished yet. So I see no reason not to have multiple feature flags.

Well, adding feature flags is adding complexity and the more feature flags the easier it is that we misplace one and release something we might not want. That's just a risk though. I expect our pipeline, tests and processes to check that.

PR #3321, opened 2023-09-15
Add sendIcpIcrc1 to ledger API

dskloetd:
# Motivation

We want to stop generating account identifiers in the client and use ICRC-1 instead.
We have a generic `icrcTransfer` API method but for cases where we want to send ICP it's convenient to have a method similar to `sendICP` that does an ICRC-1 transfer without having to specify the canister ID and the fee.

This will be used in future PRs for swap participation and canister top-ups.

# Changes

1. Add `sendIcpIcrc1` to `frontend/src/lib/api/icp-ledger.api.ts`.
2. Call `sendIcpIcrc1` from `frontend/src/lib/services/icp-accounts.services.ts` instead of calling `icrcTransfer` directly.
3. Use `DEFAULT_TRANSACTION_FEE_E8S` directly in the API, instead of using `mainTransactionFeeE8sStore` in the service. `mainTransactionFeeE8sStore` is anyway just populated with `DEFAULT_TRANSACTION_FEE_E8S`, [here](https://github.com/dfinity/nns-dapp/blob/fe5a7e8f67e780cfa696f658fe81f2e4293b0b84/frontend/src/lib/stores/transaction-fees.store.ts#L33).

Drive by: Fix confusing `logWithTimestamp` message.

# Tests

1. Update `frontend/src/tests/lib/services/icp-accounts.services.spec.ts` to expect a call to `sendIcpIcrc1` instead of `icrcTransfer`.
2. Add tests for `sendIcpIcrc1` to `frontend/src/tests/lib/api/icp-ledger.api.spec.ts`.
3. Tested manually on local with ICRC-1 addresses and legacy account identifiers.

# Todos

not necessary yet, still just refactoring

### DIFF
--- a/frontend/src/lib/api/icp-ledger.api.ts
+++ b/frontend/src/lib/api/icp-ledger.api.ts
@@ -1,15 +1,18 @@
 import { createAgent } from "$lib/api/agent.api";
+import { icrcTransfer } from "$lib/api/icrc-ledger.api";
 import type { SubAccountArray } from "$lib/canisters/nns-dapp/nns-dapp.types";
 import { LEDGER_CANISTER_ID } from "$lib/constants/canister-ids.constants";
 import { HOST } from "$lib/constants/environment.constants";
+import { DEFAULT_TRANSACTION_FEE_E8S } from "$lib/constants/icp.constants";
 import { isLedgerIdentityProxy } from "$lib/proxy/icp-ledger.services.proxy";
 import type { IcpAccountIdentifierText } from "$lib/types/account";
 import { nowInBigIntNanoSeconds } from "$lib/utils/date.utils";
 import { logWithTimestamp } from "$lib/utils/dev.utils";
 import type { Agent, Identity } from "@dfinity/agent";
+import type { IcrcAccount, IcrcBlockIndex } from "@dfinity/ledger";
 import type { BlockHeight } from "@dfinity/nns";
 import { AccountIdentifier, LedgerCanister } from "@dfinity/nns";
-import type { TokenAmount } from "@dfinity/utils";
+import { ICPToken, type TokenAmount } from "@dfinity/utils";
 
 /**
  * Transfer ICP between accounts.
@@ -39,6 +42,10 @@ export const sendICP = async ({
   logWithTimestamp(`Sending icp call...`);
   const { canister } = await ledgerCanister({ identity });
 
+  if (amount.token !== ICPToken) {
+    throw new Error("Token should be ICP");
+  }
+
   const response = await canister.transfer({
     to: AccountIdentifier.fromHex(to),
     amount: amount.toE8s(),
@@ -48,6 +55,47 @@ export const sendICP = async ({
   });
   logWithTimestamp(`Sending icp complete.`);
   return response;
+};
+
+/**
+ * Transfer ICP between accounts.
+ *
+ * @param {Object} params
+ * @param {Identity} params.identity user identity
+ * @param {string} params.to send ICP to destination address - an account identifier
+ * @param {ICP} params.amount the amount to be transferred in ICP
+ * @param {number[] | undefined} params.fromSubAccount the optional subaccount that would be the source of the transaction
+ * @param {bigint | undefined} params.createdAt the optional timestamp of the transaction. Used to avoid deduplication.
+ */
+export const sendIcpIcrc1 = async ({
+  identity,
+  to,
+  amount,
+  memo,
+  fromSubAccount,
+  createdAt,
+}: {
+  identity: Identity;
+  to: IcrcAccount;
+  amount: TokenAmount;
+  memo?: Uint8Array;
+  fromSubAccount?: SubAccountArray;
+  createdAt?: bigint;
+}): Promise<IcrcBlockIndex> => {
+  if (amount.token !== ICPToken) {
+    throw new Error("Token should be ICP");
+  }
+
+  return icrcTransfer({
+    identity,
+    to,
+    amount: amount.toE8s(),
+    memo,
+    fromSubAccount,
+    createdAt: createdAt ?? nowInBigIntNanoSeconds(),
+    fee: BigInt(DEFAULT_TRANSACTION_FEE_E8S),
+    canisterId: LEDGER_CANISTER_ID,
+  });
 };
 
 /**
--- a/frontend/src/lib/api/icp-ledger.api.ts
+++ b/frontend/src/lib/api/icp-ledger.api.ts
@@ -62,8 +62,8 @@ export const sendICP = async ({
  *
  * @param {Object} params
  * @param {Identity} params.identity user identity
- * @param {string} params.to send ICP to destination address - an account identifier
- * @param {ICP} params.amount the amount to be transferred in ICP
+ * @param {IcrcAccount} params.to destination account
+ * @param {TokenAmount} params.amount the amount to be transferred in ICP
  * @param {number[] | undefined} params.fromSubAccount the optional subaccount that would be the source of the transaction
  * @param {bigint | undefined} params.createdAt the optional timestamp of the transaction. Used to avoid deduplication.
  */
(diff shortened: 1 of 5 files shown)


frontend/src/lib/api/icp-ledger.api.ts
```diff
@@ -50,6 +57,47 @@ export const sendICP = async ({
...
+  }
+
+  return icrcTransfer({
+    identity,
+    to,
+    amount: amount.toE8s(),
+    memo,
+    fromSubAccount,
+    createdAt: createdAt ?? nowInBigIntNanoSeconds(),
+    fee: BigInt(DEFAULT_TRANSACTION_FEE_E8S),
```
lmuntaner, line 96: Why would we prefer to do this instead of expecting it as a parameter?

This is exposed from the ledger canister api. Just because we decided to not fetch it everytime to skip a call, that doesn't mean the code shouldn't be ready for it.

Maybe now that we use only query calls we can introduce the call to get the fee properly.

This is not only a refactor change, but a functionality change.

frontend/src/lib/api/icp-ledger.api.ts
```diff
@@ -1,15 +1,18 @@
 import { createAgent } from "$lib/api/agent.api";
+import { icrcTransfer } from "$lib/api/icrc-ledger.api";
```
peterpeterparker, line 2: this introduce a dependency between two apis - `icp-ledger.api` and `icrc-ledger.api` - which I think we should rather avoid.

This new function is probably best suited in a `service`.

frontend/src/lib/api/icp-ledger.api.ts
```diff
@@ -50,6 +57,47 @@ export const sendICP = async ({
...
+ */
+export const sendIcpIcrc1 = async ({
+  identity,
+  to,
+  amount,
+  memo,
+  fromSubAccount,
+  createdAt,
+}: {
+  identity: Identity;
```
peterpeterparker, line 78: this should probably extends (partially at least) the declaration of `IcrcTransferParams` given the fact that `icrcTransfer` is used within the method

frontend/src/lib/api/icp-ledger.api.ts
lmuntaner: That's an interesting point, yes. I like the idea of adding it in another layer, not the API layer.

frontend/src/lib/api/icp-ledger.api.ts
dskloetd: It's already a constant in our code. I'm just using a constant in a less convoluted way. Changing the code to fetch the constant from the canister instead of using a constant is a much bigger change outside the scope of this project.

frontend/src/lib/api/icp-ledger.api.ts
dskloetd: Isn't the purpose of `icp-ledger.api.ts` to provide an API to the ICP ledger canister? 
The ICP ledger canister does offer an [ICRC-1 interface](https://github.com/dfinity/ic/blob/def91661e9d7eb18303b850f88a903112b1c6b6a/rs/rosetta-api/icp_ledger/ledger.did#L440) so I do believe that this functionality does belong in this API.

If we can agree on that, we can separately discuss how much we want to duplicate code or reuse code between the ICP ledger API and the general ICRC-1 ledger API and how to do that.

peterpeterparker: I thought `service` was best suited but, not strong opinion on the level.

On the contrary, I'm afraid that adding this dependency will ultimately lead to a circular dependency. That's why generally we avoid referencing `.api` within each others.

Of course if we can setup a test that checks for circular dependency, then my concern is solved as well.

dskloetd: It seems unlikely that the ICRC-1 ledger API would depend on the ICP ledger API but I guess it's always possible that some things get unnecessarily tangled up. What aspect of circular dependencies are you specifically afraid of? Is it something we can deal with if it happens or can it cause problems silently?

peterpeterparker: - silent warning at build time (`npm run build`)
- reproducibility issue

frontend/src/lib/api/icp-ledger.api.ts
dskloetd: `IcrcTransferParams` has
```
  fee: bigint;
  transfer: (params: TransferParams) => Promise<IcrcBlockIndex>;
```
which I don't have here. And I'm not a fan of using `Omit` as to me it indicates that the type we really need just doesn't exist.
But I admit I also don't really understand the purpose of `IcrcTransferParams`, especially the `transfer` function.

peterpeterparker: The all ICRC ledger implementation follow this approach. For consistency reason I would appreciate if we continue to use the same approach.

peterpeterparker: That said if you rather not, no strong opinion neither and feel free to close this comment.

frontend/src/lib/api/icp-ledger.api.ts
dskloetd: > * silent warning at build time (`npm run build`)

I just tried adding a deliberate circular dependency (between `frontend/src/lib/utils/date.utils.ts` and `frontend/src/lib/services/sns-sale.services.ts`) but I don't see any warning. Do you know what it looks like or do I need to enable it with a flag. If there is indeed a warning, we can probably also add a test for it.

> * reproducibility issue

I thought about that, but in my experience the build gets more reproducible if we have more dependencies. Hence [this line](https://github.com/dfinity/nns-dapp/blob/be7f26ce62d668ad359d097a186c5276366ae5df/frontend/src/routes/%2Blayout.ts#L12).

peterpeterparker: > Do you know what it looks like 

With rollup it's a yellow warning printed in the console. Don't exactly remember the phrasing.

Anyway, as you wish. Feel free to close this comments.

frontend/src/lib/api/icp-ledger.api.ts
lmuntaner: The transaction fee will be fetched and added in the store. Yet, that store uses the default value as initial value.

By using the store, we ensure that everything will work once we enable again getting the transaction fee from the canister.

frontend/src/lib/api/icp-ledger.api.ts
lmuntaner: @peterpeterparker you want to add a new library `icp-ledger`, right? That would expose the api to make ICP transfers with ICRC.

Maybe in that case, better use a service so that then the service can use the new library.

frontend/src/lib/api/icp-ledger.api.ts
dskloetd: I'm trying to replace `sendICP`, which does not have a transaction fee parameter.
In fact, it does not even pass a transaction fee to `canister.transfer()`.
Because the transaction fee is hard coded in `ic-js`:
https://github.com/dfinity/ic-js/blob/90e8d0110529ffeba3341f8fba9fbb5059ffea94/packages/nns/src/canisters/ledger/ledger.request.converts.ts#L39

Maybe, instead of using the IcrcLedger API here, I should add another method on the ICP LedgerCanister in ic-js and use the hardcoded fee from there the same as the existing `transfer` method does?

frontend/src/lib/api/icp-ledger.api.ts
dskloetd: There already is a service (actually, there are several). But the service calls the API. So the question is which API should it call. Currently it calls `sendICP`, which does not use ICRC-1. So doesn't it make sense to have a parallel which does use ICRC-1 (or eventually replace the existing `sendICP`)?

dskloetd:
This is replaced by https://github.com/dfinity/nns-dapp/pull/3360

PR #3322, opened 2023-09-15
Make daysToDuration call secondsToDuration

dskloetd:
# Motivation

We want `daysToDuration` and `secondsToDuration` to be consistent when the number of seconds is a whole number of times the number of seconds in a day. So there is no need to have 2 completely separate implementations.

# Changes

Make `daysToDuration` call `secondsToDuration`.

# Tests

Existing test coverage is good and passes.

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #3323, opened 2023-09-15
Include the schema in the stats

bitdivine:
# Motivation
When changing the schema, a migration completes at some time after the deployment of a new canister.  It is necessary to have visibility of when that migration completes.

The most important information is what the canonical storage schema is.  Some sort of progress-meter would be nice but not essential and depends rather on the details of a specific migration.  A countdown to zero is reasonably general and the meaning of the count can be specified in a different way for every migration.

# Changes
* Add the current schema to the metrics.

# Tests
The metric only becomes interesting when we have migrations.  We can however verify that artificial changes in the schema are passed through to the metrics and API:

Artificial changes:
```
 pub enum SchemaLabel {
-    Map = 0,
+    Map = 1,
 }
 
     pub fn migration_countdown(&self) -> u32 {
-        0
+        1234
     }
```
This yields this metrics:
```
# HELP nns_dapp_schema The nns-dapp schema version
# TYPE nns_dapp_schema gauge
nns_dapp_schema 1 1694800609059
# HELP nns_dapp_migration_countdown When non-zero, a migration is in progress.
# TYPE nns_dapp_migration_countdown gauge
nns_dapp_migration_countdown 1234 1694800609059
```
and stats:
```
max@sinkpad:~/dfn/nns-dapp/branches/migration-status (1:46:10)$ dfx canister call nns-dapp get_stats
(
  record {
    migration_countdown = opt (1_234 : nat32);
   [SNIP]
    schema = opt (1 : nat32);
```

# Todos


rs/backend/src/accounts_store/schema/proxy.rs
```diff
@@ -51,6 +58,9 @@ impl AccountsDbTrait for AccountsDbAsProxy {
     fn values(&self) -> Box<dyn Iterator<Item = Account> + '_> {
         self.map.values()
     }
+    fn schema_label(&self) -> SchemaLabel {
+        self.map.schema_label()
```
bitdivine, line 62: We have no migrations yet, which makes this simple in this case.

rs/backend/src/stats.rs
```diff
@@ -48,6 +48,8 @@ pub struct Stats {
     //       production, CI and snsdemo populate these fields.
     pub stable_memory_size_bytes: Option<u64>,
     pub wasm_memory_size_bytes: Option<u64>,
+    pub schema: Option<u32>,              // The numeric form of a SchemaLabel.
+    pub migration_countdown: Option<u32>, // When non-zero, a migration is in progress.
```
frederikrothenberger, line 52: I think this metric is a bit confusing... So is it supposed to count down from some value that indicate how many things are yet to be migrated, right?
What about: 
1) only exposing the metric if a migration is in progress?
2) make it a progress percentage? This way you can write generic alerts for the generic metric rather than having a value of 37 mean entirely different things depending on the specific migration.

scripts/nns-dapp/downgrade-upgrade-test
```diff
@@ -104,6 +104,26 @@ check_num_transactions() {
   fi
 }
 
+# Wait for the migration countdown to complete.
+wait_for_migration() {
```
frederikrothenberger, line 108: This test runs against the local replica, right?
Just because as soon as multiple replicas are involved, this test / check could become flaky: If you hit a slightly behind replica, you might still observe `0` from before the upgrade.

But might not be worth the effort to do anything about it here. One solution would be to also expose the last upgrade timestamp, query the metrics before the upgrade and the wait with this check until that timestamp has changed.

rs/backend/src/accounts_store/schema.rs
```diff
@@ -87,6 +87,17 @@ pub trait AccountsDbTrait {
...
+    /// Gets the label of the storage schema.
+    fn schema_label(&self) -> SchemaLabel;
+}
+
+/// A label to identify the schema.
+///
+/// Note: The numeric representations of these labels are guaranteed to be stable.
+#[repr(u32)]
+pub enum SchemaLabel {
+    Map = 0,
```
frederikrothenberger, line 100: Maybe adding a comment here, what is meant by the "Map" schema would be nice (i.e. does not use stable memory, etc.).

frederikrothenberger (approved):
LGTM, left only minor comments.

rs/backend/src/stats.rs
bitdivine: I thought about the percentage but wasn't very comfortable with a rather arbitrary mapping of work done to percentage.  I reckoned that if we decided that we really want a percentage, we can always start the countdown at 100 so this code does not exclude percentages.

Having the metric present only when a migration is in progress is an interesting idea.  Essentially, as soon as the countdown reached 0, or whatever the final value is, it would disappear.  I would want to check how Prometheus would handle a metric that is present only sometimes.

scripts/nns-dapp/downgrade-upgrade-test
bitdivine: Correct, the test runs only on a one-replica subnet.

~A 0 from before the upgrade is distinguishable from a 0 after upgrade by checking the schema.~  Actually, no, most downgrade/upgrade tests don't change the schema so have no migration.

Given that we are currently testing on one replica only I will leave this but we may need to return to this if we change the testing environment.

rs/backend/src/stats.rs
bitdivine: This will need more research.  Thank you for the ideas.  I will merge but if the tests all work out well then maybe we can try dropping the countdown when no migration is in progress.  The data type is already opt, so we just need to adapt the prometheus code to hide the metric when the countdown is None and set None when not migrating.  The data types don'tr need to change.

PR #3324, opened 2023-09-18
Remove transfer from IcrcTransferParams

dskloetd:
# Motivation

The type `IcrcTransferParams` is defined as
```
export interface IcrcTransferParams {
  to: IcrcAccount;
  amount: bigint;
  memo?: Uint8Array;
  fromSubAccount?: SubAccountArray;
  createdAt?: bigint;
  fee: bigint;
  transfer: (params: TransferParams) => Promise<IcrcBlockIndex>;
}
```

The field `transfer` is only used by `executeIcrcTransfer` and every other place where `IcrcTransferParams` is used, it's used as `Omit<IcrcTransferParams, "transfer">`.

# Changes

Remove `transfer` from `IcrcTransferParams` and at it back only to the params of `executeIcrcTransfer`.

# Tests

`npm run build`
`npm run check`
`npm run test`

All still pass.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/api/icrc-ledger.api.ts
+++ b/frontend/src/lib/api/icrc-ledger.api.ts
@@ -81,7 +81,6 @@ export interface IcrcTransferParams {
   fromSubAccount?: SubAccountArray;
   createdAt?: bigint;
   fee: bigint;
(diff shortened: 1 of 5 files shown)


peterpeterparker (approved):
Nice, thanks!

PR #3332, opened 2023-09-18
Bump bump ic-js

peterpeterparker:
# Motivation

Bump ic-js which was updated with last did files of the CMC, ledger and index canisters.

# Changes

- bump lib
- update anonymize function to leverage new `approve` type.

# Notes

No particular test added for `approve` and `spender` fields because those are related to ICRC-2 which NNS dapp does not implement yet in any features.

### DIFF
--- a/frontend/src/lib/utils/anonymize.utils.ts
+++ b/frontend/src/lib/utils/anonymize.utils.ts
@@ -311,6 +311,32 @@ const anonymizeTransfer = async (
     amount: await anonymizeAmount(data.amount),
   };
 };
+type Approve = Transfer & {
+  expected_allowance: [] | [bigint];
+  expires_at: [] | [bigint];
+  spender: SnsAccountRaw;
+};
+type ApproveOpt = [] | [Approve];
+const anonymizeApprove = async (
+  transfer: ApproveOpt
+): Promise<{ [key in keyof Approve]: unknown } | undefined> => {
+  const data = fromNullable(transfer);
+  if (isNullish(data)) {
+    return undefined;
+  }
+
+  return {
+    to: await anonymizeSnsAccount(data.to),
+    from: await anonymizeSnsAccount(data.from),
+    spender: await anonymizeSnsAccount(data.spender),
+    fee: data.fee,
+    memo: data.memo,
+    created_at_time: data.created_at_time,
+    amount: await anonymizeAmount(data.amount),
+    expected_allowance: data.expected_allowance,
+    expires_at: data.expires_at,
+  };
+};
 
 const anonymizeSnsTransaction = async (
   tx: IcrcTransaction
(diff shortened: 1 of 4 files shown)


frontend/src/lib/utils/anonymize.utils.ts
```diff
@@ -321,6 +347,7 @@ const anonymizeSnsTransaction = async (
     burn: await anonymizeTransfer(tx.burn as TransferOpt),
     mint: await anonymizeTransfer(tx.mint as TransferOpt),
     transfer: await anonymizeTransfer(tx.transfer as TransferOpt),
+    approve: await anonymizeApprove(tx.approve as ApproveOpt),
```
dskloetd, line 350: Does `approve` need its own anonymizing function?
Burn, mint and transfer all go through the same `anonymizeTransfer`.

frontend/src/lib/utils/anonymize.utils.ts
```diff
@@ -311,6 +311,32 @@ const anonymizeTransfer = async (
...
+
+  return {
+    to: await anonymizeSnsAccount(data.to),
+    from: await anonymizeSnsAccount(data.from),
+    spender: await anonymizeSnsAccount(data.spender),
+    fee: data.fee,
+    memo: data.memo,
+    created_at_time: data.created_at_time,
+    amount: await anonymizeAmount(data.amount),
+    expected_allowance: data.expected_allowance,
```
dskloetd, line 336: Isn't this like an amount that should be anonymized like an amount?

dskloetd (approved):

frontend/src/lib/utils/anonymize.utils.ts
peterpeterparker: The `approve` field have few more types, three to be precise.

frontend/src/lib/utils/anonymize.utils.ts
peterpeterparker: Sure good idea.

frontend/src/lib/utils/anonymize.utils.ts
dskloetd: `Burn` also has a `spender` field, for example, but apparently we don't care:
https://github.com/dfinity/ic-js/blob/4c8dd0b22416f86f75b5a957fa6a210b44fd5156/packages/ledger/candid/icrc1_index.d.ts#L23

Fine with me either way, but I was confused by it when I was trying to fix it myself.

peterpeterparker: I don't mind neither so I removed the new code and used `anonymizeTransfer`.

frontend/src/lib/utils/anonymize.utils.ts
peterpeterparker: Code has been removed so...

PR #3334, opened 2023-09-19
Feat: Minor change disbursing maturity UI

lmuntaner:
# Motivation

Final UI improvements for Disbursing maturity flow.

# Changes

* Rename buttons to "Stake" and "Disburse".
* Change font-size of description in confirmation screen.

# Tests

Only UI changes.

# Todos

- [ ] Add entry to changelog (if necessary).
Covered by disbursing maturity entry.

PR #3335, opened 2023-09-19
Fix: SnsStakeMaturityButton spec file

lmuntaner:
# Motivation

One test for SnsStakeMaturityButton was not correct. The test passed because the expected text (title of the modal) was the same text as the button.

In this PR, I added a testId to the modal and checked the presence of the modal with test id. The test failed.

The test also failed because it required a token in the store to be present. That is not necessary for the modal, but it was inside an `if` checking the presence of the token.

Therefore, in this PR, I also refactored the SnsNeuronModals to check for presence next to the type of the modal. This way it's easier to understand what is strictly necessary in each modal.

# Changes

* Add testId to StakeMaturityModal.
* Move modals and conditions in SnsNeuronModals.

# Tests

* Use `SnsNeuronContextTest` in `SnsStakeMaturityButton.spec` to check that modal is opened.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/tests/lib/components/sns-neuron-detail/actions/SnsStakeMaturityButton.spec.ts
+++ b/frontend/src/tests/lib/components/sns-neuron-detail/actions/SnsStakeMaturityButton.spec.ts
@@ -3,9 +3,10 @@
  */
 
 import SnsStakeMaturityButton from "$lib/components/sns-neuron-detail/actions/SnsStakeMaturityButton.svelte";
-import en from "$tests/mocks/i18n.mock";
+import { mockPrincipal } from "$tests/mocks/auth.store.mock";
 import { mockSnsNeuron } from "$tests/mocks/sns-neurons.mock";
 import { fireEvent, render, waitFor } from "@testing-library/svelte";
+import SnsNeuronContextTest from "../SnsNeuronContextTest.svelte";
 
 describe("SnsStakeMaturityButton", () => {
   afterEach(() => {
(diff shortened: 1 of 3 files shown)


lmuntaner:
@dskloetd please review

This fixes the test in #3334

frontend/src/tests/lib/components/sns-neuron-detail/actions/SnsStakeMaturityButton.spec.ts
```diff
@@ -3,30 +3,33 @@
...
         },
+        testComponent: SnsStakeMaturityButton,
+        passPropNeuron: true,
+        rootCanisterId: mockPrincipal,
       },
     });
 
     fireEvent.click(getByTestId("stake-maturity-button") as HTMLButtonElement);
 
     await waitFor(() =>
```
dskloetd, line 31: Do you need `waitFor`?

The pattern I'd like to see is:
```
expect modal is not present
fire event
expect modal is present
```
Because then you know if the expectation succeeds for a different reason.

But that doesn't work with `waitFor`.
Would be great if it just works without `waitFor`, but otherwise I expect it probably works with `await runResolvedPromises()`. In that case we should do that both before expecting the modal not to be there and before expecting the modal to be there.

lmuntaner: Done! Thanks for the suggestion.

dskloetd (approved):

PR #3337, opened 2023-09-20
Chore: Enable ENABLE_DISBURSE_MATURITY

lmuntaner:
# Motivation

Enable disburse maturity for mainnet.

# Changes

* Set feature flag `ENABLE_DISBURSE_MATURITY` to `true` everywhere.

# Tests

* Adapt feature flag tests.

# Todos


lmuntaner:
@dskloetd please review

Deployed here: https://gzaux-siaaa-aaaaa-aad4a-cai.nnsdapp.dfinity.network/

dskloetd:
I played a bit with it and I noticed that the estimated disbursed tokens range is rounded to 2 decimals. And the rounding is to nearest neighbor (not always up or always down).
This means that we might say that we estimate newly minted ICP between 2.07-2.28 and then in some cases you actually get 2.066 ICP. Since we're already giving a range, shouldn't it include all possible values?
If we don't want to display full decimals, shall we at least round the minimum down and the maximum up so we definitely include all possible values?

lmuntaner:
> I played a bit with it and I noticed that the estimated disbursed tokens range is rounded to 2 decimals. And the rounding is to nearest neighbor (not always up or always down). This means that we might say that we estimate newly minted ICP between 2.07-2.28 and then in some cases you actually get 2.066 ICP. Since we're already giving a range, shouldn't it include all possible values? If we don't want to display full decimals, shall we at least round the minimum down and the maximum up so we definitely include all possible values?

Thanks, good point. Yes, that makes sense. Let me prepare a PR. Good catch!

lmuntaner:
@dskloetd can we merge this after #3340 ?

dskloetd (approved):

PR #3338, opened 2023-09-20
Feat: Improve message disabled disburse maturity button

lmuntaner:
# Motivation

The message of the disabled Disburse button when the neuron had no maturity could be confusing.

In this PR, we introduce two messages to disable the button. When there is no maturity, or not enough to disburse.

# Changes

* Two i18n keys instead of one.
* Check the maturity in SnsDisburseMaturityButton to select the disabled text.

# Tests

* Add a new test case to check the tooltip text when no maturity i SnsDisburseMaturityButton.

# Todos

Not necessary, covered by the disburse maturity entry.


lmuntaner:
Old UI (and still used when not maturity is lower than needed but more than 0)

Screenshot 2023-09-20 at 10 41 21

New message when no maturity

Screenshot 2023-09-20 at 10 41 11

lmuntaner:
@dskloetd please review

There is an issue with the tooltip not being fully visible. I'll prepare another PR for it.

dskloetd (approved):

PR #3339, opened 2023-09-20
Fix: Tooltip is hidden if container has overflow hidden

lmuntaner:
# Motivation

Part of the tooltip of the disabled "Disburse" button is hidden behind the island.

That's because the tooltip uses the window to calculate whether there is enough space to the left or right. The problem happens when the container where the tooltip lives has `overflow: hidden`.

The tooltip already received a containerSelector. Therefore, it can use the container to calculate the space to the left and right.

# Changes

* Use the container element instead of the width to calculate the space in the left or right of the tooltip.

# Tests

* Purely CSS changes.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -30,6 +30,7 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 * Fix sent transaction icon background color dark theme.
 * Improve text color of total value locked's label.
 * Make duration rendering consistent.
+* Improve Tooltip location when container hides part of it.
 
 #### Security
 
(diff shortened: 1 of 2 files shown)


lmuntaner:
Before:

Screenshot 2023-09-20 at 10 41 21

Now:

Screenshot 2023-09-20 at 11 14 08

lmuntaner:

lmuntaner:
I found one that still doesn't look good on mobile: Merging neurons.

Screenshot 2023-09-20 at 11 20 52

I'll take a look in another PR.

peterpeterparker (approved):
LGTM, thx

dskloetd:
FYI, the proper way to deal with this is to place the tooltip at the root of the DOM where it's not inside any container with `overflow: hidden`.

PR #3340, opened 2023-09-20
Fix: Better min and max range in disburse maturity

lmuntaner:
# Motivation

The range of the predicted tokens in the disbursing flow is an approximation of two decimals. By default, the decimals were rounded the standard way. That meant that the min range could be rounded up and give a wrong min, or the max range rounded down and give a wrong max.

In this PR, we add the `roundingMode` parameter to `formatToken` and use it for the range.

# Changes

* Add `roundingMode` param to `formatToken`.
* Use param `roundingMode` for the range in DisburseMaturityModal confirmation screen.

# Tests

* Add skipped tests because the new parameter is only supported in node 19 and we are in node 18.

# Todos

Not necessary. Covered by an entry.

### DIFF
--- a/frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
+++ b/frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
@@ -99,12 +99,19 @@ describe("SnsDisburseMaturityModal", () => {
   });
 
   it("should display summary information in the last step", async () => {
-    const po = await renderSnsDisburseMaturityModal();
+    const neuron = createMockSnsNeuron({
+      id: [1],
+      maturity: 1_233_123_112n,
+    });
+    const po = await renderSnsDisburseMaturityModal(neuron);
     await po.setPercentage(50);
     await po.clickNextButton();
 
     expect(await po.getConfirmPercentage()).toBe("50%");
-    expect(await po.getConfirmTokens()).toBe("0.48-0.53 TST");
+    expect(await po.getConfirmTokens()).toBe("5.86-6.47 TST");
+    expect(await po.getConfirmTokensDetailed()).toBe(
+      "5.85733478-6.47389634 TST"
+    );
     expect(await po.getConfirmDestination()).toBe("Main");
   });
 
--- a/frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
+++ b/frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
@@ -101,14 +101,14 @@ describe("SnsDisburseMaturityModal", () => {
   it("should display summary information in the last step", async () => {
     const neuron = createMockSnsNeuron({
       id: [1],
-      maturity: 1_233_123_112n,
+      maturity: 1_000_000_000n,
     });
     const po = await renderSnsDisburseMaturityModal(neuron);
     await po.setPercentage(50);
     await po.clickNextButton();
 
     expect(await po.getConfirmPercentage()).toBe("50%");
-    expect(await po.getConfirmTokens()).toBe("5.86-6.47 TST");
+    expect(await po.getConfirmTokens()).toBe("4.75-5.25 TST");
     expect(await po.getConfirmDestination()).toBe("Main");
   });
 
@@ -123,6 +123,9 @@ describe("SnsDisburseMaturityModal", () => {
     await po.setPercentage(100);
     await po.clickNextButton();
 
+    // with 123123213n maturity
+    // -5% is 1,169670524, which should show as 1.16 with the rounding mode "floor"
+    // +5% is 1,292793737, which should show as 1.30 with the rounding mode "ceil"
     expect(await po.getConfirmTokens()).toBe("1.16-1.30 TST");
   });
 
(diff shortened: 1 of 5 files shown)


lmuntaner:
New UI:

The actual numbers are:

* min: 1,169670524
* max: 1,292793737

Screenshot 2023-09-20 at 13 13 18

lmuntaner:
@dskloetd please review

frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
```diff
@@ -99,15 +99,33 @@ describe("SnsDisburseMaturityModal", () => {
...
+      id: [1],
+      maturity: 1_233_123_112n,
+    });
+    const po = await renderSnsDisburseMaturityModal(neuron);
     await po.setPercentage(50);
     await po.clickNextButton();
 
     expect(await po.getConfirmPercentage()).toBe("50%");
-    expect(await po.getConfirmTokens()).toBe("0.48-0.53 TST");
+    expect(await po.getConfirmTokens()).toBe("5.86-6.47 TST");
```
dskloetd, line 111: If I calculate I get
```
1_233_123_112 * .50 * .95 / 100000000
> 5.857334782
```
So why isn't it rounded down to 5.85?

frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
```diff
@@ -99,15 +99,33 @@ describe("SnsDisburseMaturityModal", () => {
...
     expect(await po.getConfirmDestination()).toBe("Main");
   });
 
+  // NodeJS supports roundingMode since v19
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#browser_compatibility
+  it.skip("should display range with floor and ceil rounding", async () => {
```
dskloetd, line 117: skip?

dskloetd, line 117: It's not really clear from the test why this tests floor and ceil rounding.

frontend/src/tests/lib/utils/token.utils.spec.ts
```diff
@@ -102,6 +102,14 @@ describe("token-utils", () => {
     ).toEqual(`2'000'000.00000000`);
   });
 
+  // NodeJS supports roundingMode since v19
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#browser_compatibility
+  it.skip("should use roundingMode", () => {
```
dskloetd, line 107: skip?

frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
lmuntaner: Node doesn't support the roundingMode, Let me change the amount so that it's not affected by it.

frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
lmuntaner: I added a comment on top to explain it.

lmuntaner: Let me add a comment.

lmuntaner:

frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
dskloetd: It's easy to forget to enable a skipped test.
How about instead commenting out the expectation below and adding another expectation with the actual values and a comment explaining why they are wrong?
Then when Node is upgraded, the test will fail and we'll automatically be reminded to fix it to be what we want it to be.
```
    // with 123123213n maturity
    // -5% is 1,169670524, which should show as 1.16 with the rounding mode "floor"
    // +5% is 1,292793737, which should show as 1.30 with the rounding mode "ceil"
    // roundingMode is only supported by NodeJS from version 19 so we need to expect incorrect values here until we upgrade
    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#browser_compatibility
    // expect(await po.getConfirmTokens()).toBe("1.16-1.30 TST");
    expect(await po.getConfirmTokens()).toBe("1.17-1.29 TST");
```

lmuntaner: Sounds good

lmuntaner: done

lmuntaner:

Couldn't you have approved after the last comment?

frontend/src/tests/lib/modals/sns/SnsDisburseMaturityModal.spec.ts
dskloetd: The test is still skipped as far as I can see.

lmuntaner: I just pushed. I thought I already pushed it when I added the comment, my bad.

PR #3341, opened 2023-09-20
Paginated account store

bitdivine:
# Motivation
We have a few huge accounts and many tiny accounts.  The current fixed size stable memory structures are ill suited to this.  Work _is_ progressing on unbounded stable memory data structures, however for the time being we cannot assume that they will be stable and well tested by the time we migrate.  Until then, we have to split our data into pages and store them in a "classic" stable memory map.

It is not impossible that we will be able to switch to the unbounded maps.  It is too early to tell.

The plan is to proceed with our own pagination and then, when everything else is ready, potentially switch to the new stable structures.

# Changes
* Define how accounts will be serialized and stored in pages.
* Create a mock data store with this pagination for unit tests.

# Tests
* There are tests that save and restore data from the mock account store.

# Todos


rs/backend/src/accounts_store/schema/s1.rs
```diff
@@ -0,0 +1,292 @@
+//! Data storage schema `S1`: Accounts data is stored in a `StableBTreeMap`,
```
bitdivine, line 1: Note: This PR covers only the pagination, which is tested with a normal BTreeMap.

rs/backend/src/accounts_store/schema/s1.rs
```diff
@@ -0,0 +1,306 @@
...
+    // TODO: Add a small cache to limit the cost of getting and parsing data to once per API call.
+
+    /// Every account is serialized and stored in between 1 and 2**16 pages.
+    const MAX_PAGES_PER_ACCOUNT: usize = (u16::MAX as usize) + 1;
+
+    // Low level methods to get and set pages.
+    /// Gets a page of memory.
+    fn s1_get_account_page(&self, account_storage_key: &AccountStorageKey) -> Option<AccountStoragePage>;
+
+    /// Inserts a page of memory.
```
frederikrothenberger, line 46: Please expand the comment to explain the return value.

rs/backend/src/accounts_store/schema/s1.rs
```diff
@@ -0,0 +1,306 @@
...
+    fn s1_insert_account_page(
+        &mut self,
+        account_storage_key: AccountStorageKey,
+        account: AccountStoragePage,
+    ) -> Option<AccountStoragePage>;
+
+    /// Checks whether a page of memory exists.
+    fn s1_contains_account_page(&self, account_storage_key: &AccountStorageKey) -> bool;
+
+    /// Removes a page of memory.
```
frederikrothenberger, line 56: Dito.

rs/backend/src/accounts_store/schema/s1.rs
```diff
@@ -0,0 +1,306 @@
...
+    fn s1_remove_account_page(&mut self, account_storage_key: &AccountStorageKey) -> Option<AccountStoragePage>;
+
+    /// Checks whether to get the next page.
+    /// - If the very last page is full, getting the next page will return None. That is expected.
+    fn s1_is_last_page(last_page_maybe: &Option<AccountStoragePage>) -> bool {
```
frederikrothenberger, line 61: This is an unexpected implementation for me:
* I would not have expected it to take an option, because `None` is _not_ the last page.
* It gives wrong results for the last page of accounts that align exactly with the pagination.

I would either rename the function to better convey what it does, or handle the implementation in a different way. I.e. you could also do a lookup on the next index and decide based on its existence.

frederikrothenberger: I see you use this to decide when you exit a loop that iterates over potentially every page. You could do this with an range scan. In that case, you would not need this function at all.

frederikrothenberger (commented):
I think, the implementation of pagination might benefit from using [range](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) scans, which is available in both the normal BTreeMap and the StableBTreeMap. See comments.

rs/backend/src/accounts_store/schema/s1.rs
```diff
@@ -0,0 +1,306 @@
...
+    const MAX_SIZE: u32 = Self::SIZE as u32;
+    const IS_FIXED_SIZE: bool = true;
+}
+impl AccountStorageKey {
+    /// Location of the page number in the key bytes.
+    ///
+    /// Note: When an account is serialized, it is split into pages and the pages are stored in a
+    /// BTreeMap.  The first page for that account has page number 0, the second 1 and so on.  The
+    /// `PAGE_NUM` field referred to here is where the page number appears in the lookup key.
+    const PAGE_NUM_OFFSET: usize = 0;
```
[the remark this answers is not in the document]
bitdivine: I did have similar thoughts and on balance I put the index first.  The deciding factor was that the account identifier has variable length and it may make sense to switch the key field to variable length.  There are several excellent arguments the other way.  I was not persuaded that having the leaves close together had any performance impact or structural advantage in this case.  Once we have `Unbounded` in stable structures the key becomes just the account identifier, so we are really only looking at very immediate advantages and I could not see any the other way around.  I could be wrong.

image

rs/backend/src/accounts_store/schema/s1/mock.rs
```diff
@@ -0,0 +1,85 @@
...
+mod test_db {
+    //! Tests for the [`MockS1DataStorage`] implementation of the [`AccountsDbTrait`].
+    use super::MockS1DataStorage;
+    use crate::accounts_store::schema::tests::test_accounts_db;
+
+    test_accounts_db!(MockS1DataStorage::default());
+}
+
+#[derive(Default)]
+struct MockS1DataStorage {
```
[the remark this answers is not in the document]
bitdivine: True, that would be even closer to prod.  Do you mind if I make that switch in a separate PR?

rs/backend/src/accounts_store/schema/s1/mock/test_s1.rs
```diff
@@ -0,0 +1,91 @@
...
+
+/// Creates a tiny account with minimal activity.
+fn tiny_account(account_index: u64) -> Account {
+    let size = ToyAccountSize::default();
+    toy_account(account_index, size)
+}
+
+#[test]
+fn large_account_uses_several_pages() {
+    let yo = large_account(1);
```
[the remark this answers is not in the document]
bitdivine: I don't know.  I guess a name is just a name.  Ask your mother if you want to be sure. 😉

Someone once published an NPM crate called `yomoma` that was something like of 2GB of garbage, just to mess around with npm devs who were making ever larger packages.  It was taken down but it was fun while it lasted.  Terribly improper but who cares?

rs/backend/src/accounts_store/schema/s1/mock/test_s1.rs
```diff
@@ -0,0 +1,91 @@
...
+    );
+}
+
+#[test]
+fn tiny_account_uses_one_page() {
+    let yo = tiny_account(1);
+    let num_pages = AccountStoragePage::pages_from_account(&yo).len();
+    assert_eq!(
+        1, num_pages,
+        "A large test account should use several pages of memory but has only: {}",
```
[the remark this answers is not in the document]
bitdivine: Good catch.

rs/backend/src/accounts_store/schema/s1/mock/test_s1.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#[test]
+fn test_account_storage() {
+    let mut storage = MockS1DataStorage {
+        accounts_storage: BTreeMap::new(),
+    };
+    let account_key = vec![1, 2, 3];
+    let account = tiny_account(1);
+    // TODO: Check that this spans several pages.
```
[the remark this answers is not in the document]
bitdivine: Implemented by the large account check above, so the TODO can be removed.

rs/backend/src/accounts_store/schema/s1/mock/test_s1.rs
```diff
@@ -0,0 +1,91 @@
...
+    storage.s1_remove_account(&account_key);
+    assert!(!storage.s1_contains_account(&account_key));
+    assert_eq!(storage.s1_get_account(&account_key), None);
+}
+
+/// We assume that new account identifiers are 32 bytes long.
+///
+/// Note: Old-style account identifiers were 28 bytes long.
+#[test]
+fn account_identifier_has_32_bytes() {
```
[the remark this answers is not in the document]
bitdivine: Yes, if we get shiny new account IDs that consume more bytes, I want to know about it before anything breaks.

rs/backend/src/accounts_store/schema/s1.rs
```diff
@@ -0,0 +1,306 @@
...
+use dfn_candid::Candid;
+use ic_stable_structures::storable::{BoundedStorable, Storable};
+use on_wire::{FromWire, IntoWire};
+use std::borrow::Cow;
+use std::convert::TryInto;
+
+#[cfg(test)]
+mod mock;
+
+pub trait AccountsDbS1Trait {
```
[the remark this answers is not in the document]
frederikrothenberger: Yeah, I agree with numbers being fragile. Also I like the expressive name of `AccountsInStableMemory`. So I would go with that variant. 👍

rs/backend/src/accounts_store/schema/s1.rs
bitdivine: I dropped this and used a range scan instead, as discussed.

rs/backend/src/accounts_store/schema/s1.rs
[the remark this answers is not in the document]
bitdivine: Discussed.  @frederikrothenberger changed my mind!  I have changed the order of the fields to put the account identifier at the beginning and made the page index big endian.

rs/backend/src/accounts_store/schema/accounts_in_stable_memory.rs
```diff
@@ -33,76 +33,66 @@ use std::convert::TryInto;
 #[cfg(test)]
 mod mock;
 
-pub trait AccountsDbS1Trait {
+pub trait AccountsInStableMemoryTrait {
     // TODO: Add a small cache to limit the cost of getting and parsing data to once per API call.
 
-    /// Every account is serialized and stored in between 1 and 2**16 pages.
-    const MAX_PAGES_PER_ACCOUNT: usize = (u16::MAX as usize) + 1;
+    /// Every account is serialized and stored in between `1` and `2**16-1` pages.
```
frederikrothenberger, line 39: Where does the `-1` come from?

rs/backend/src/accounts_store/schema/accounts_in_stable_memory/mock/test_accounts_in_stable_memory.rs
```diff
@@ -63,19 +63,57 @@ fn test_account_storage() {
...
+    );
+    assert_eq!(
+        storage.aism_pages_len(),
+        1,
+        "Expected to be using one page after inserting a tiny account."
+    );
+    assert!(storage.aism_contains_account(&account_key));
+    assert_eq!(storage.aism_get_account(&account_key), Some(account.clone()));
+
+    // Updates the account to a large one that uses lots of pages
```
frederikrothenberger, line 85: It would be nice if this block had an assert to check that the large account did indeed take up more pages than just 1.

frederikrothenberger (approved):
LGTM, just two nits! 👍

rs/backend/src/accounts_store/schema/accounts_in_stable_memory.rs
bitdivine: Good catch.  I considered for a while reserving the last page but then discarded the idea.

rs/backend/src/accounts_store/schema/accounts_in_stable_memory/mock/test_accounts_in_stable_memory.rs
bitdivine: Done.

PR #3344, opened 2023-09-21
E2E: Work around images not loading

dskloetd:
# Motivation

We have an e2e test that tests that images load.
Sometimes images don't load and the test times out:
https://github.com/dfinity/nns-dapp/actions/runs/6245207559?pr=3337

I'm not sure why they don't always load on CI but let's retry the test when they don't.

# Changes

1. Set a timeout on waiting for images to be loaded.
2. Retry the test up to 2 times when it fails.

# Tests

Still pass, hopefully less flaky.

# Todos

not necessary


lmuntaner (approved):
LGTM!

PR #3347, opened 2023-09-21
Add stakeNeuronIcrc1 to governance API service

dskloetd:
# Motivation

We want to stake neurons using an ICRC-1 transfer when possible.
We access the governance API through the governance API service, which caches certified responses.

# Changes

Add stakeNeuronIcrc1 to governance API service, which calls the governance API.

# Tests

Unit tests added

# Todos

Will add when it's all hooked up


lmuntaner (approved):
Thanks!

PR #3351, opened 2023-09-21
Add ENABLE_SWAP_ICRC1 feature flag

dskloetd:
# Motivation

We want to use ICRC-1 transfer when possible, including when participating in an SNS swap.

# Changes

Add feature flag `ENABLE_SWAP_ICRC1` which will be used to guard using ICRC-1 transfers when participating in an SNS swap.

# Tests

no

# Todos

not necessary


lmuntaner (approved):
LGTM!

PR #3353, opened 2023-09-22
GIX-1911: Split participation in progress bar

lmuntaner:
# Motivation

We want to differentiate between direct and Neurons' Fund participation.

In this PR, the differentiation is done only in the progress bar.

This is not yet supported because it's missing the field in the API response. Therefore, I used a getter that returns undefined.

I don't think there is any need for a feature flag. As soon as the field is available, we should support it.

# Changes

* Add a new sns summary getter `getNeuronsFundParticipation` that returns `undefined`.
* Change prop `value` in CommitmentProgressBar to two props `directParticipation` and `nfParticipation`.
* Use the new `segments` prop of the ProgressBar instead of the `value`.
* Use the new getter `getNeuronsFundParticipation` in ProjectCommitment to use new props in CommitmentProgressBar.

# Tests

* Adapt tests of CommitmentProgressBar to new props and add a test case.
* Add a test case in ProjectCommitment.

# Todos

Not necessary yet because there is no change for the user until the field is supported by the API.

### DIFF
--- a/frontend/src/lib/components/project-detail/CommitmentProgressBar.svelte
+++ b/frontend/src/lib/components/project-detail/CommitmentProgressBar.svelte
@@ -2,10 +2,14 @@
   import { TokenAmount, ICPToken } from "@dfinity/utils";
   import { i18n } from "$lib/stores/i18n";
   import AmountDisplay from "../ic/AmountDisplay.svelte";
-  import { ProgressBar } from "@dfinity/gix-components";
+  import {
+    ProgressBar,
+    type ProgressBarSegment,
+  } from "@dfinity/gix-components";
 
   export let max: bigint;
-  export let value: bigint;
+  export let directParticipation: bigint;
+  export let nfParticipation: bigint;
   export let minimumIndicator: bigint | undefined = undefined;
 
   let width: number | undefined;
@@ -14,9 +18,19 @@
     minimumIndicator !== undefined && width !== undefined
       ? (Number(minimumIndicator) / Number(max)) * width
       : undefined;
+
+  let segments: ProgressBarSegment[] = [];
+  $: segments = [
+    { value: Number(nfParticipation), color: "var(--positive-emphasis)" },
+    {
+      value: Number(directParticipation),
+      color: "var(--warning-emphasis)",
+    },
+  ];
 </script>
 
-<ProgressBar max={Number(max)} value={Number(value)} color="warning">
+<!-- TODO: Remove value={0} after https://github.com/dfinity/gix-components/pull/298 is merged -->
+<ProgressBar max={Number(max)} value={0} {segments}>
   <div class="info" slot="top">
     <p class="right">
       <span>
--- a/frontend/src/tests/lib/components/project-detail/CommitmentProgressBar.spec.ts
+++ b/frontend/src/tests/lib/components/project-detail/CommitmentProgressBar.spec.ts
@@ -7,7 +7,7 @@ import { render } from "@testing-library/svelte";
 
 describe("CommitmentProgressBar", () => {
   const props = {
-    directParticipation: 1327n,
+    directParticipation: 1500n,
     nfParticipation: 0n,
     max: BigInt(3000),
     minimumIndicator: BigInt(1500),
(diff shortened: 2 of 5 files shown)


lmuntaner:
New UI (by mocking a nfParticipation in the getter):

Screenshot 2023-09-22 at 09 10 29

In the end, we also want to show the two values above (the one in the right):

image (1)

The split in the values will come in another PR.

lmuntaner:
@dskloetd please review

frontend/src/lib/components/project-detail/CommitmentProgressBar.svelte
```diff
@@ -14,9 +18,19 @@
     minimumIndicator !== undefined && width !== undefined
       ? (Number(minimumIndicator) / Number(max)) * width
       : undefined;
+
+  let segments: ProgressBarSegment[] = [];
+  $: segments = [
+    { value: Number(nfParticipation), color: "var(--positive-emphasis)" },
```
dskloetd, line 24: I'm surprised that we're putting the NF participation first and in green.
Isn't the "main" participation the direct participation and the NF participation more like a secondary/derived participation?

frontend/src/tests/lib/components/project-detail/CommitmentProgressBar.spec.ts
```diff
@@ -41,4 +43,17 @@ describe("CommitmentProgressBar", () => {
...
+  it("should display the value adding NF and direct commitments", async () => {
+    const nfParticipation = 1000n;
+    const { container } = render(CommitmentProgressBar, {
+      props: {
+        ...props,
+        nfParticipation,
+      },
+    });
+    expect(container.querySelector("progress").value).toBe(
+      Number(props.directParticipation + nfParticipation)
```
dskloetd, line 56: Where do you test the individual segments are correct?

frontend/src/lib/components/project-detail/CommitmentProgressBar.svelte
lmuntaner: I also raised the colors question with Artem. It might change before we go live. Also the color of the triangle that sets the minimum is the same color as direct participation. We might want to change that also.

I'll do those changes in a separate PR.

frontend/src/tests/lib/components/project-detail/CommitmentProgressBar.spec.ts
lmuntaner: I do not. That's show in the CSS. But maybe I can look at the inline-style. That has the information of the gradient.

It's an implementation detail.

Or should we add a test attribute in the ProgressBar with the segments information?

dskloetd: > It's an implementation detail.

Huh, it's the actual thing that matters, right? If it wouldn't show up as different colored segments, there would be no reason to set separate values.
If it's expressed via an inline style, I guess that's what we should have an expectation on.
Ideally that would be encapsulated in a page object. But that can be a separate PR.

lmuntaner: I'm waiting for the testId to be available to set it as a PO. I'll do it after #3354

lmuntaner: New test case added.

frontend/src/tests/lib/components/project-detail/ProjectCommitment.spec.ts
```diff
@@ -109,4 +117,29 @@ describe("ProjectCommitment", () => {
...
+          buyer_total_icp_e8s: directCommitment + nfCommitment,
+        },
+      },
+      swapCommitment: mockSnsFullProject.swapCommitment as SnsSwapCommitment,
+      Component: ProjectCommitment,
+    });
+
+    expect(container.querySelector("progress").value).toBe(
+      Number(directCommitment + nfCommitment)
+    );
```
dskloetd, line 143: Once you have the PO, please test for the individual segments here as well.

dskloetd (approved):

frontend/src/tests/lib/components/project-detail/ProjectCommitment.spec.ts
lmuntaner: Sure

PR #3355, opened 2023-09-22
GIX-1831 Fix CanisterCard CSS

dskloetd:
# Motivation

1. The tooltip on the canister ID hash is cut off and only a tiny sliver of it is visible.
2. On a very long canister name, it doesn't show ellipses.

# Changes

1. Remove 
`@include card.title` from the start slot in the `CanisterCard` because it comes with `overflow: hidden` which cuts off the tooltip.
2. Add `width: 100%` to the start slot, to propagate the available width for `overflow: hidden` in a descendant. Otherwise the `display:block` child in the `display:inline-flex` parent will take up space according to its content, rather than its parent width.
3. Add `@include card.title` to the title-block in `CanisterCardTitle` to get `overflow: hidden` and `text-overflow: ellipsis`.
4. Remove `display: inline-flex` from `h4` because otherwise it will be sized according to its content instead of its parent and you don't get to see the ellipsis.

# Tests

Try it out here: https://fswcu-uqaaa-aaaaa-aadwq-cai.nnsdapp.dfinity.network/canisters/?u=fswcu-uqaaa-aaaaa-aadwq-cai

Before:
image

After:
image

# Todos


peterpeterparker:
Is the card overflow issue on mobile devices introduced by this PR or it's an existing issue?

I do not use the canister's name function therefore cannot compare with production.

Capture d’écran 2023-09-26 à 10 05 10

dskloetd:
Good catch, but...
It appears that in narrow window mode, the card never gets smaller than to fit its content.
The same happens with a subaccount card with name "MMMMMMMMMMMMMMMMMMMMMMMM" although account names have a somewhat smaller font than canister names so it's a bit less notieable.
Anyway, this can be reproduced on the release candidate and on prod with subaccount names to it's unrelated to my change.

peterpeterparker (approved):
LGTM, thanks for the PR and for double checking.

Can you take care of fixing the display issue we discosvered in another PR?

dskloetd:
> Can you take care of fixing the display issue we discosvered in another PR?

I don't know honestly. It took quite a while before I got to fixing the issue from this PR.

PR #3356, opened 2023-09-22
Update rust

bitdivine:
# Motivation
The rust updater tries to update to the wrong version.  It gets the latest release from GitHub, however that does not correspond to the latest stable version.  In fact, it currently matches an older release that presumably had a patch release.

# Changes
- Determine the latest stable version directly from the channel toml.
  - Note: This is the authoritative file used by [`rustup`](https://rustup.rs/) (see [the code](https://github.com/rust-lang/rustup/blob/d4c684485cb5c817d00aa6c6c0aee44da80450c7/src/test/mock/clitools.rs#L989)), [rust-forge](https://forge.rust-lang.org/) (see [the code](https://github.com/rust-lang/rust-forge/blob/533d65ea37246388d914c1fcff2fb79e2556d23f/blacksmith/src/lib.rs#L17)).

# Tests

# Todos

### DIFF
--- a/.github/workflows/update-rust.yml
+++ b/.github/workflows/update-rust.yml
@@ -6,6 +6,10 @@ on:
     # check for new rust versions weekly
     - cron: '30 3 * * FRI'
   workflow_dispatch:
+  push:
+    branches:
+      # The development branch for this workflow:
+      - "update-rust"
 jobs:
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!

PR #3359, opened 2023-09-22
GIX-1911: Split total commitment

lmuntaner:
# Motivation

We want to add the deails of how much the NF is participating in a swap.

# Changes

* New project util `getProjectCommitmentSplit` that returns a variant with the commitments split or only the total commitment.
* Add the details of NF and direct commitments below the current overall commitment in ProjectCommitment if NF commitment is present.

# Tests

* Test new project util.
* Add test cases for new functionality.

# Todos

Not yet worth an entry because the functionality is not available.

### DIFF
--- a/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
@@ -14,10 +14,7 @@
   import { snsSwapMetricsStore } from "$lib/stores/sns-swap-metrics.store";
   import { nonNullish } from "@dfinity/utils";
   import { swapSaleBuyerCount } from "$lib/utils/sns-swap.utils";
-  import {
-    getNeuronsFundParticipation,
-    isNeuronsFundParticipationPresent,
-  } from "$lib/getters/sns-summary";
+  import { getNeuronsFundParticipation } from "$lib/getters/sns-summary";
 
   const { store: projectDetailStore } = getContext<ProjectDetailContext>(
     PROJECT_DETAIL_CONTEXT_KEY
--- a/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
@@ -14,7 +14,10 @@
   import { snsSwapMetricsStore } from "$lib/stores/sns-swap-metrics.store";
   import { nonNullish } from "@dfinity/utils";
   import { swapSaleBuyerCount } from "$lib/utils/sns-swap.utils";
-  import { getNeuronsFundParticipation } from "$lib/getters/sns-summary";
+  import {
+    getNeuronsFundParticipation,
+    isNeuronsFundParticipationPresent,
+  } from "$lib/getters/sns-summary";
 
   const { store: projectDetailStore } = getContext<ProjectDetailContext>(
     PROJECT_DETAIL_CONTEXT_KEY
@@ -40,6 +43,15 @@
     token: ICPToken,
   });
 
+  let directCommitmentE8s: bigint;
+  $: directCommitmentE8s = buyersTotalCommitment - neuronsFundCommitmentE8s;
+
+  let directCommitmentIcp: TokenAmount;
+  $: directCommitmentIcp = TokenAmount.fromE8s({
+    amount: directCommitmentE8s,
+    token: ICPToken,
+  });
+
   let saleBuyerCount: number | undefined;
   $: saleBuyerCount = swapSaleBuyerCount({
     rootCanisterId: $projectDetailStore?.summary?.rootCanisterId,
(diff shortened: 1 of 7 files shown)


lmuntaner:
New UI:

Screenshot 2023-09-22 at 14 07 58

lmuntaner:
@dskloetd please review

frontend/src/lib/components/project-detail/ProjectCommitment.svelte
```diff
@@ -66,11 +88,35 @@
...
+
+    <AmountDisplay slot="value" amount={directCommitmentIcp} singleLine />
+  </KeyValuePair>
+{/if}
 <div data-tid="sns-project-commitment-progress">
   <CommitmentProgressBar
-    directParticipation={buyersTotalCommitment - neuronsFundParticipation}
-    nfParticipation={neuronsFundParticipation}
+    directParticipation={directCommitmentE8s}
+    nfParticipation={neuronsFundCommitmentE8s}
```
dskloetd: Won't this look misleading if `isNeuronsFundCommitmentAvailable` is `false`?

frontend/src/lib/getters/sns-summary.ts
```diff
@@ -15,3 +15,8 @@ export const getConditionsToAccept = (
 export const getNeuronsFundParticipation = (
   _summary: SnsSummary
 ): bigint | undefined => undefined;
+
+// TODO: https://dfinity.atlassian.net/browse/GIX-1909 check if nf participation field is present
+export const isNeuronsFundParticipationPresent = (
+  _summary: SnsSummary
+): boolean => false;
```
dskloetd: Suggestion: If this calls `getNeuronsFundParticipation`, you only need to mock `getNeuronsFundParticipation` and the answers are always consistent.
Now you have to be careful not to mock `getNeuronsFundParticipation` with a number and `isNeuronsFundParticipationPresent` with false or the other way around.

frontend/src/lib/components/project-detail/ProjectCommitment.svelte
```diff
@@ -47,8 +47,27 @@
...
+    amount: neuronsFundCommitmentE8s ?? 0n,
+    token: ICPToken,
+  });
+
+  let isNeuronsFundCommitmentAvailable: boolean;
+  $: isNeuronsFundCommitmentAvailable = nonNullish(neuronsFundCommitmentE8s);
+
+  let directCommitmentE8s: bigint;
+  $: directCommitmentE8s =
+    buyersTotalCommitment - (neuronsFundCommitmentE8s ?? 0n);
```
dskloetd, line 64: I don't think it's accurate to claim to know the `directCommitmentE8s` if you don't know the `neuronsFundCommitmentE8s`. Shouldn't this also be `undefined` if `neuronsFundCommitmentE8s` is `undefined`?

frontend/src/lib/components/project-detail/ProjectCommitment.svelte
lmuntaner: Yes, agree. That's why I changed to add here the `?? on` instead.

frontend/src/lib/getters/sns-summary.ts
lmuntaner: Precisely the improvement I thought when taking out the trash as well.

lmuntaner:

I added an util to manage the logic of whether there is split or not `getProjectCommitmentSplit`.

frontend/src/lib/utils/projects.utils.ts
```diff
@@ -399,3 +402,35 @@ export const differentSummaries = (
...
+  nfCommitmentE8s: bigint;
+};
+export type ProjectCommitmentSplit =
+  | { totalCommitmentE8s: bigint }
+  | FullProjectCommitmentSplit;
+
+export const isFullProjectCommitmentSplit = (
+  commitment: ProjectCommitmentSplit
+): commitment is FullProjectCommitmentSplit =>
+  nonNullish((commitment as FullProjectCommitmentSplit).directCommitmentE8s) &&
```
dskloetd, line 418: Isn't the proper way to test this without casting to do `"directCommitmentE8s" in commitment`?

dskloetd (approved):

frontend/src/lib/utils/projects.utils.ts
lmuntaner: I looked at this docs https://www.typescriptlang.org/docs/handbook/advanced-types.html

But I like better your change. Done.

frontend/src/lib/components/project-detail/ProjectCommitment.svelte
lmuntaner: If we decide to change the color of the direct vs when it's total commitment, in that case yes, I'll add another prop.

But for simplicity, I considered the full commitment to be direct commitment to keep the current UI working without changing.

But yes, maybe the color for direct should be different. The colors are still being reviewed by Artem.

PR #3360, opened 2023-09-22
Use icrc1Transfer on ICP ledger canister

dskloetd:
# Motivation

When you use an ICRC-1 address, we make an ICRC-1 transfer but currently we use the generic ICRC-1 ledger interface for that, specifying the ICP ledger canister ID.
But the ICP ledger interface has its own ICRC-1 transfer method.

# Changes

Use the direct `icrc1Transfer` method on the ICP ledger API instead of using the generic ICRC-1 ledger API.

# Tests

Unit test updated.
Tested manually and noticed that for HW transactions, an empty memo is rendered as 422710814078008.
This will be fixed with https://github.com/dfinity/ic-js/pull/423

# Todos


lmuntaner (approved):
Thanks!

PR #3364, opened 2023-09-25
GIX-1881: toAccount disburse maturity service & api

lmuntaner:
# Motivation

Add `toAccount` to disburse maturity service and api.

# Changes

* Add parameter `toAccount` to disbrse maturity sns governance api.
* Add `toAccountAddress` to disburseMaturity sns service. Convert string to ICRC account to send it to api.

# Tests

* Add tests for disburseMaturity sns governance api (it was missing).
* Add test case for disburseMaturity sns service.
* I moved the `jest.clearMocks` to a main `beforeEach` in both files.

# Todos

Not yet necessary. I'll add an entry when the new parameter is used from the UI.

### DIFF
--- a/frontend/src/lib/services/sns-neurons.services.ts
+++ b/frontend/src/lib/services/sns-neurons.services.ts
@@ -61,6 +61,7 @@ import {
   fromDefinedNullable,
   fromNullable,
   isNullish,
+  nonNullish,
 } from "@dfinity/utils";
 import { get } from "svelte/store";
 import { getAuthenticatedIdentity } from "./auth.services";
@@ -776,19 +777,26 @@ export const disburseMaturity = async ({
   neuronId,
   rootCanisterId,
   percentageToDisburse,
+  toAccountAddress,
 }: {
   neuronId: SnsNeuronId;
   rootCanisterId: Principal;
   percentageToDisburse: number;
+  toAccountAddress?: string;
 }): Promise<{ success: boolean }> => {
   try {
     const identity = await getSnsNeuronIdentity();
+
+    const toAccount = nonNullish(toAccountAddress)
+      ? decodeIcrcAccount(toAccountAddress)
+      : undefined;
 
     await disburseMaturityApi({
       neuronId,
       rootCanisterId,
       percentageToDisburse,
       identity,
+      toAccount,
     });
 
     return { success: true };
--- a/frontend/src/tests/lib/api/sns-governance.api.spec.ts
+++ b/frontend/src/tests/lib/api/sns-governance.api.spec.ts
@@ -7,6 +7,7 @@ import {
   autoStakeMaturity,
   claimNeuron,
   disburse,
+  disburseMaturity,
   getNervousSystemFunctions,
   getNeuronBalance,
   getSnsNeuron,
@@ -89,6 +90,7 @@ describe("sns-api", () => {
   const autoStakeMaturitySpy = jest.fn().mockResolvedValue(undefined);
   const listProposalsSpy = jest.fn().mockResolvedValue(proposals);
   const getProposalSpy = jest.fn().mockResolvedValue(mockSnsProposal);
+  const disburseMaturitySpy = jest.fn().mockResolvedValue(undefined);
   const nervousSystemFunctionsMock: SnsListNervousSystemFunctionsResponse = {
     reserved_ids: new BigUint64Array(),
     functions: [nervousSystemFunctionMock],
@@ -99,6 +101,10 @@ describe("sns-api", () => {
   const nervousSystemParametersSpy = jest
     .fn()
     .mockResolvedValue(snsNervousSystemParametersMock);
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
 
   beforeAll(() => {
     jest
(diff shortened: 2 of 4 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/services/sns-neurons.services.ts
```diff
@@ -776,19 +777,26 @@ export const disburseMaturity = async ({
   neuronId,
   rootCanisterId,
   percentageToDisburse,
+  toAccountAddress,
 }: {
   neuronId: SnsNeuronId;
   rootCanisterId: Principal;
   percentageToDisburse: number;
+  toAccountAddress?: string;
```
dskloetd, line 785: What will the UI be like for this?
Will there be only a dropdown with your own accounts or will there be a text input where you can enter any address?

If it's text input, will it only allow ICRC-1 address, even though we don't even display ICRC-1 address in the UI? If it will allow non-ICRC-1 addresses, how can those be decoded to an `IcrcAccount`?

If it's not text input but only a dropdown, can we use something more type safe than `string`, here?

frontend/src/tests/lib/api/sns-governance.api.spec.ts
```diff
@@ -400,4 +402,46 @@ describe("sns-api", () => {
...
+    it("should disburse maturity", async () => {
+      const res = await disburseMaturity({
+        identity: mockIdentity,
+        rootCanisterId: rootCanisterIdMock,
+        neuronId: mockSnsNeuron.id[0],
+        percentageToDisburse: 33,
+      });
+
+      expect(res).toBeUndefined();
+      expect(disburseMaturitySpy).toBeCalledTimes(1);
```
dskloetd, line 416: Can you expect it not to be called at the beginning of the test?

frontend/src/tests/lib/api/sns-governance.api.spec.ts
```diff
@@ -400,4 +402,46 @@ describe("sns-api", () => {
...
+      };
+      await disburseMaturity({
+        identity: mockIdentity,
+        rootCanisterId: rootCanisterIdMock,
+        neuronId: mockSnsNeuron.id[0],
+        percentageToDisburse: 33,
+        toAccount,
+      });
+
+      expect(disburseMaturitySpy).toBeCalledTimes(1);
```
dskloetd, line 439: Can you expect it not to be called at the beginning of the test?

frontend/src/tests/lib/services/sns-neurons.services.spec.ts
```diff
@@ -790,6 +789,35 @@ describe("sns-neurons-services", () => {
         identity,
       });
     });
+
+    it("should decode the destination address to an ICRC account", async () => {
+      const ownerText =
+        "k2t6j-2nvnp-4zjm3-25dtz-6xhaa-c7boj-5gayf-oj3xs-i43lp-teztq-6ae";
+      const owner = Principal.fromText(ownerText);
+      const subaccount = Uint8Array.from([...Array(32)].map((_, i) => i + 1));
```
dskloetd, line 797: Why not use a more realistic subaccount like 1 or 2?

dskloetd: To add to this: a subaccount with leading 0 bytes is actually slightly more complicated because the leading zeros have to be removed in the text representation, so I think even though it looks simpler it provides slightly more test coverage.
But also in general I like to use realistic data in tests because it helps me conceptualize what's going on.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: The UI will use `SelectDestinationAddress`. Like you added it to some disburse flow.

The input will only allow ICRC-1. We display the ICRC-1 account in the SNS tokens.

This is an SNS service, it expects only ICRC-1 accounts. But good point, let me add a try catch around the decode function and show a toast error if it can't be decoded.

frontend/src/tests/lib/services/sns-neurons.services.spec.ts
lmuntaner: That's a good point. I used the same example as in the decodeAccount.

I wanted to check that it was decoding properly.

But this test is not about decoding all edge cases. Just that it decodes. I'll add a simpler and more realistic account address.

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: I added a test case for the service. There was already a try/catch

frontend/src/tests/lib/api/sns-governance.api.spec.ts
lmuntaner: done

frontend/src/tests/lib/api/sns-governance.api.spec.ts
lmuntaner: done

frontend/src/tests/lib/services/sns-neurons.services.spec.ts
lmuntaner: Done

frontend/src/lib/services/sns-neurons.services.ts
dskloetd: Of course, thanks for clarifying. Will there be similar functionality for NNS?

dskloetd (approved):

frontend/src/lib/services/sns-neurons.services.ts
lmuntaner: Yes, but at the moment NNS doesn't support disburse maturity. It uses Spawn Neuron instead. Who know when we'll implement it for NNS.

PR #3365, opened 2023-09-25
Chore: CommitmentProgressPo

lmuntaner:
# Motivation

Introduce a page object to improve testing: CommitmentProgressPo

# Changes

* New PO CommitmentProgressPo.
* Change the CommitmentProgress.spec.ts to use the new PO.

# Tests

Only test changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/components/project-detail/CommitmentProgressBar.svelte
+++ b/frontend/src/lib/components/project-detail/CommitmentProgressBar.svelte
@@ -29,8 +29,11 @@
   ];
 </script>
 
-<!-- TODO: Remove value={0} after https://github.com/dfinity/gix-components/pull/298 is merged -->
-<ProgressBar max={Number(max)} value={0} {segments}>
+<ProgressBar
+  max={Number(max)}
+  {segments}
+  testId="commitment-progress-bar-component"
+>
   <div class="info" slot="top">
     <p class="right">
       <span>
--- a/frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
+++ b/frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
@@ -1,0 +1,56 @@
+import type { PageObjectElement } from "$tests/types/page-object.types";
+import { BasePageObject } from "./base.page-object";
+
+export class CommitmentProgressBarPo extends BasePageObject {
+  private static TID = "commitment-progress-bar-component";
+
+  static under(element: PageObjectElement): CommitmentProgressBarPo {
+    return new CommitmentProgressBarPo(
+      element.byTestId(CommitmentProgressBarPo.TID)
+    );
+  }
+
+  hasMinCommitmentIndicator(): Promise<boolean> {
+    return this.root.byTestId("commitment-min-indicator").isPresent();
+  }
+
+  hasMaxCommitmentIndicator(): Promise<boolean> {
+    return this.root.byTestId("commitment-max-indicator").isPresent();
+  }
+
+  async getMinCommitment(): Promise<string> {
+    return (await this.getText("commitment-min-indicator-value")).trim();
+  }
+
+  async getMaxCommitment(): Promise<string> {
+    return (await this.getText("commitment-max-indicator-value")).trim();
+  }
+
+  getProgressBarValue(): Promise<string> {
+    // Getting the `value` directly isn't working, we get the max value instead.
+    return this.root.querySelector("progress").getAttribute("value");
+  }
+
+  /**
+   * Uses the slices of the linear gradient to determine the segments of the progress bar.
+   *
+   * Example of inline-style: "--progress-bar-background: linear-gradient(to right, var(--positive-emphasis) 0% 25%, var(--warning-emphasis) 25% 100%);"
+   */
+  async getProgressBarSegments(): Promise<string[][]> {
+    const inlineStyle = await this.root
+      .querySelector("progress")
+      .getAttribute("style");
+    const percentageRegex = /\d+(\.\d+)?%/g;
+    return (
+      inlineStyle
+        // We assume that there is only a single css variable `linear-gradient` in the style.
+        .split(",")
+        // Remove everything before the `to right` part.
+        .slice(1)
+        // Match and extract the percentages of each segment.
+        .map<string[]>((segment) =>
+          Array.from(segment.matchAll(percentageRegex)).map(([match]) => match)
+        )
+    );
+  }
+}
--- a/frontend/src/tests/lib/components/project-detail/CommitmentProgressBar.spec.ts
+++ b/frontend/src/tests/lib/components/project-detail/CommitmentProgressBar.spec.ts
@@ -3,6 +3,8 @@
  */
 
 import CommitmentProgressBar from "$lib/components/project-detail/CommitmentProgressBar.svelte";
+import { CommitmentProgressBarPo } from "$tests/page-objects/CommitmentProgressBarPo.page-object";
+import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
 import { render } from "@testing-library/svelte";
 
 describe("CommitmentProgressBar", () => {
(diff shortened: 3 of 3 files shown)


frontend/src/lib/components/project-detail/CommitmentProgressBar.svelte
```diff
@@ -29,8 +29,11 @@
   ];
 </script>
 
-<!-- TODO: Remove value={0} after https://github.com/dfinity/gix-components/pull/298 is merged -->
-<ProgressBar max={Number(max)} value={0} {segments}>
```
lmuntaner, line 33: Removing the `value` prop is not strictly part of this PR. But I thought it was ok to remove it here. I can revert.

lmuntaner:
@dskloetd please review

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
```diff
@@ -0,0 +1,56 @@
...
+   * Example of inline-style: "--progress-bar-background: linear-gradient(to right, var(--positive-emphasis) 0% 25%, var(--warning-emphasis) 25% 100%);"
+   */
+  async getProgressBarSegments(): Promise<string[][]> {
+    const inlineStyle = await this.root
+      .querySelector("progress")
+      .getAttribute("style");
+    const percentageRegex = /\d+(\.\d+)?%/g;
+    return (
+      inlineStyle
+        // We assume that there is only a single css variable `linear-gradient` in the style.
```
dskloetd, line 46: Can we just use a regex that matches this exactly?
```
linear-gradient(to right, var(--positive-emphasis) 0% X%, var(--warning-emphasis) Y% 100%);"
```
And then we just assert `X == Y` and return that value?

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
```diff
@@ -0,0 +1,56 @@
...
+        .split(",")
+        // Remove everything before the `to right` part.
+        .slice(1)
+        // Match and extract the percentages of each segment.
+        .map<string[]>((segment) =>
+          Array.from(segment.matchAll(percentageRegex)).map(([match]) => match)
+        )
+    );
+  }
```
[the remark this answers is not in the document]
lmuntaner: Is this another method, or do you mean instead of `getProgressBarSegments`?

Then we could have `getDirectCommitmentPercentage` and `getNFCommitmentPercentage`.

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
dskloetd: I want the interface of the PO to just give percentages. One function per percentage. So `getNeuronFundCommitmentPercentage`, `getDirectParticipationPercentage`.

And I want the implementation to actually verify that the style attribute is what the implementation assumes the style attribute to be.

lmuntaner: I changed the methods and implementation. Let me know if that's what you had in mind.

lmuntaner:

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
```diff
@@ -0,0 +1,67 @@
...
+      /--progress-bar-background: linear-gradient\(to right, var\(.*\) 0% (\d+(?:\.\d+)?)%, var\(.*\) (\d+(?:\.\d+)?)% 100%\);/g;
+
+    const inlineStyle = await this.root
+      .querySelector("progress")
+      .getAttribute("style");
+
+    const matches = inlineStyleRegex.exec(inlineStyle);
+
+    if (matches === null) {
+      throw new Error("Invalid inline style");
```
dskloetd, line 50: How about including the actual style in the error for easier debugging?

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  async getMinCommitment(): Promise<string> {
+    return (await this.getText("commitment-min-indicator-value")).trim();
+  }
+
+  async getMaxCommitment(): Promise<string> {
+    return (await this.getText("commitment-max-indicator-value")).trim();
+  }
+
+  getProgressBarValue(): Promise<string> {
```
dskloetd, line 29: Would this be equal to the total (direct + NF) commitment? In ICP or in E8s?

Shall we use a method name that expresses that rather than what we see in the DOM?

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const [_inlineStyle, firstPercentage, secondPercentage] = matches;
+
+    if (firstPercentage !== secondPercentage) {
+      throw new Error(
+        `The first and second percentage should be the same: ${firstPercentage} !== ${secondPercentage}`
+      );
+    }
+
+    return Number(firstPercentage);
```
dskloetd, line 61: Thanks! This is exactly what I meant.

But now that I see it, I'm wondering if we could make it even more convenient by multiplying it by `getProgressBarValue()` to get the actual amount. Is that correct?

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
```diff
@@ -0,0 +1,56 @@
...
+  async getMinCommitment(): Promise<string> {
+    return (await this.getText("commitment-min-indicator-value")).trim();
+  }
+
+  async getMaxCommitment(): Promise<string> {
+    return (await this.getText("commitment-max-indicator-value")).trim();
+  }
+
+  getProgressBarValue(): Promise<string> {
+    // Getting the `value` directly isn't working, we get the max value instead.
```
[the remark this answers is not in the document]
dskloetd: I don't know why you're seeing what you're seeing. My explanation was for the sake of terminology.
If I open NNS dapp and manually go into the DOM and set either the `value` attribute or the `value` property on the `progress` element, they seem to be synced between each other.
My best guess is that either, the testing library doesn't mimic the real browser perfectly, or something else was going on when you changed this that seemed to fix it but it was perhaps unrelated.

If they're always in sync, you might ask why I'm saying they're different things.
For example if you set the `foo` attribute instead of the `value` attribute, and then you check the `foo` property, you'll see those are not in sync. So `value` is in sync because it has special logic. Not because attributes and properties always reflect each other.

dskloetd (approved):

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
lmuntaner: done

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
lmuntaner: done

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
lmuntaner: done

frontend/src/tests/page-objects/CommitmentProgressBarPo.page-object.ts
```diff
@@ -26,17 +26,20 @@ export class CommitmentProgressBarPo extends BasePageObject {
     return (await this.getText("commitment-max-indicator-value")).trim();
   }
 
-  getProgressBarValue(): Promise<string> {
+  async getProgressBarTotalCommitmentE8s(): Promise<bigint> {
```
dskloetd, line 29: Given that the class already has "ProgressBar" in the name, having it in the method name is well may be unnecessary.

dskloetd (approved):
Thanks!

PR #3367, opened 2023-09-25
GIX-1881: Add destination disburse maturity

lmuntaner:
# Motivation

Users can select a destination when disbursing maturity.

# Changes

* Use `QrWizardModal` and `SelectDestinationAddress` in DisburseMaturityModal.
* Add props `rootCanisterId`, `token` and `selectedNetwork` to DisburrseMaturityModal.
* SnsDisburseMaturityModal passes new props to DisburseMaturityModal. `selectedNetwork` is left `undefined` on purpose because it means using ICRC accounts.
* Add `destinationAddress` to the `nnsDisburseMaturity` event caught by SnsDisburseMaturityModal.
* Use the new parameter `toAccount` of the disburse maturity service.

# Tests

* Add a test case that user can disburse maturity to another account and that it gets an error if a non-ICRC account is used.
* Add `blur` method to jest-page-object.
* Add a few methods to the related POs.

# Todos

### DIFF
--- a/frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
+++ b/frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
@@ -14,7 +14,7 @@
   } from "@dfinity/gix-components";
   import { formatToken } from "$lib/utils/token.utils";
   import { formatMaturity } from "$lib/utils/neuron.utils";
-  import QrWizardModal from "../transaction/QrWizardModal.svelte";
+  import QrWizardModal from "$lib/modals/transaction/QrWizardModal.svelte";
   import SelectDestinationAddress from "$lib/components/accounts/SelectDestinationAddress.svelte";
   import type { Principal } from "@dfinity/principal";
   import { assertNonNullish, type Token } from "@dfinity/utils";
@@ -28,7 +28,7 @@
   export let rootCanisterId: Principal;
   export let token: Token;
   export let minimumAmountE8s: bigint;
-  // Using `undefined` allows only ICRC accounts.
+  // Using `undefined` allows only ICRC
   export let selectedNetwork: TransactionNetwork | undefined = undefined;
 
   const steps: WizardSteps = [
--- a/frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
+++ b/frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
@@ -19,7 +19,7 @@
   import type { Principal } from "@dfinity/principal";
   import { assertNonNullish, type Token } from "@dfinity/utils";
   import type { QrResponse } from "$lib/types/qr-wizard-modal";
-  import { TransactionNetwork } from "$lib/types/transaction";
+  import type { TransactionNetwork } from "$lib/types/transaction";
   import { getAccountByRootCanister } from "$lib/utils/accounts.utils";
   import { universesAccountsStore } from "$lib/derived/universes-accounts.derived";
 
@@ -28,6 +28,8 @@
   export let rootCanisterId: Principal;
   export let token: Token;
   export let minimumAmountE8s: bigint;
+  // Using `undefined` allows only ICRC accounts.
+  export let selectedNetwork: TransactionNetwork | undefined = undefined;
 
   const steps: WizardSteps = [
     {
@@ -167,7 +169,7 @@
           {rootCanisterId}
           bind:selectedDestinationAddress
           bind:showManualAddress
-          selectedNetwork={TransactionNetwork.ICP}
+          {selectedNetwork}
           on:nnsOpenQRCodeReader={goQRCode}
         />
       </div>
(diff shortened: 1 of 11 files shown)


lmuntaner:

mstrasinskis (approved):
LGTM, just a minor comment.

frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
```diff
@@ -14,11 +14,22 @@
   } from "@dfinity/gix-components";
   import { formatToken } from "$lib/utils/token.utils";
   import { formatMaturity } from "$lib/utils/neuron.utils";
+  import QrWizardModal from "../transaction/QrWizardModal.svelte";
```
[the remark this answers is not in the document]
lmuntaner: Done! Thanks!

frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
```diff
@@ -14,11 +14,22 @@
...
+  import { universesAccountsStore } from "$lib/derived/universes-accounts.derived";
 
   export let availableMaturityE8s: bigint;
   export let tokenSymbol: string;
-
+  export let rootCanisterId: Principal;
+  export let token: Token;
   export let minimumAmountE8s: bigint;
+  // Using `undefined` allows only ICRC
+  export let selectedNetwork: TransactionNetwork | undefined = undefined;
```
dskloetd, line 32: Does this need to be exported? Does it even need to be a variable rather than just passing `selectedNetwork={undefined}`.

(And should we maybe have a less confusing interface to tell `SelectDestinationAddress` that we want to use ICRC-1?)

frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
```diff
@@ -53,8 +64,12 @@
       : undefined;
 
   const dispatcher = createEventDispatcher();
-  const disburseNeuronMaturity = () =>
-    dispatcher("nnsDisburseMaturity", { percentageToDisburse });
+  const disburseNeuronMaturity = () => {
+    dispatcher("nnsDisburseMaturity", {
```
dskloetd, line 68: Why is the event called `nnsDisburseMaturity` while this is specifically for SNS and not NNS?

frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
```diff
@@ -14,11 +14,22 @@
...
+  import SelectDestinationAddress from "$lib/components/accounts/SelectDestinationAddress.svelte";
+  import type { Principal } from "@dfinity/principal";
+  import { assertNonNullish, type Token } from "@dfinity/utils";
+  import type { QrResponse } from "$lib/types/qr-wizard-modal";
+  import type { TransactionNetwork } from "$lib/types/transaction";
+  import { getAccountByRootCanister } from "$lib/utils/accounts.utils";
+  import { universesAccountsStore } from "$lib/derived/universes-accounts.derived";
 
   export let availableMaturityE8s: bigint;
   export let tokenSymbol: string;
```
dskloetd, line 27: Now that `token` is passed in, should we derive `tokenSymbol` from `token`?
I assume we don't want it to be possible for them to not match?

frontend/src/tests/page-objects/simple-base.page-object.ts
```diff
@@ -35,4 +35,8 @@ export class SimpleBasePageObject {
   getText(tid: string | undefined = undefined): Promise<string> {
     return this.getElement(tid).getText();
   }
+
+  async blur(): Promise<void> {
```
dskloetd, line 39: I'm wondering if this is overkill. For example `typeText` is also not here.

frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
lmuntaner: Because now we only use it for SNS, we could have used `selectedNetwork={undefined}`. But precisely because I found it confusing I wanted to add it. So that when we do NnsDisburseMaturityModal we remember about it.

Yes, I believe a better name or another prop would make it clearer. But this is the prop used in `SelectDestinationAddress` and I preferred to keep consistency.

frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
lmuntaner: We always add the suffix `nns`. Now it's confusing, but when we started we didn't think there would be SNSes.

It means more "event of nns dapp".

frontend/src/lib/modals/neurons/DisburseMaturityModal.svelte
lmuntaner: Good point! Done: https://github.com/dfinity/nns-dapp/pull/3374

frontend/src/tests/page-objects/simple-base.page-object.ts
lmuntaner: Do you mean to add it in the `TextInput` directly? Sounds good!

Done: https://github.com/dfinity/nns-dapp/pull/3375

PR #3370, opened 2023-09-25
Create release branch without upstream

dskloetd:
# Motivation

When you do `git checkout -b Release_XYZ origin/main` it not only creates a branch starting from the current state of `origin/main` but it also set "upstream" to `origin/main`, which means that if you do `git push` without additional arguments, it tries to push to `main` on GitHub.

# Changes

Instead of creating the release branch with `git checkout -b "$BRANCH_NAME" origin/main`, create it with `checkout -b "$BRANCH_NAME" "$RELEASE_COMMIT"` where `RELEASE_COMMIT=$(git rev-parse origin/main)`.

This way the release branch won't have any upstream and you'll have to set it manually before you can `git push` without additional arguments.

# Tests

Test have been updated.

# Todos


scripts/nns-dapp/release-sop
```diff
@@ -342,13 +342,14 @@ echo Checklist file: "$CHECKLIST_FILE"
 record "Start time" date +"%Y-%m-%d %H:%M:%S"
 confirm_cmd git fetch --tags --force
 record "Release SOP script is newest version" is_release_sop_script_newest_version
-confirm_cmd git checkout -b "$BRANCH_NAME" origin/main
+record "Release commit" git rev-parse origin/main
+RELEASE_COMMIT="$(checklist_must_get "Release commit")"
+confirm_cmd git checkout -b "$BRANCH_NAME" "$RELEASE_COMMIT"
```
bitdivine, line 347: This is not correct.  There is already a PR with a correct solution [here](https://github.com/dfinity/nns-dapp/pull/3368).

bitdivine (changes requested):

bitdivine:
>  you'll have to set it manually before you can git push without additional arguments.

This should be automated, not done manually.

dskloetd:
> This should be automated, not done manually.

Yes, the entire second half of the release process should still be automated. That's not what this is about.

PR #3376, opened 2023-09-26
Fix changelog

bitdivine:
# Motivation
The command that splits the changelog after a release added both proposals at that commit, not just the `nns-dapp` proposal.

# Changes
- Delete the aggregator proposal number from the `nns-dapp` changelog.

# Tests
Note: There is a CI check that ensures that the changelog is not modified.  That will have to be overridden.

# Todos

  - Not applicable

### DIFF
--- a/CHANGELOG-Nns-Dapp.md
+++ b/CHANGELOG-Nns-Dapp.md
@@ -42,7 +42,6 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Security
 
 ## Proposal 124787
-124788
 
 ### Application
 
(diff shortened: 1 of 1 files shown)


CHANGELOG-Nns-Dapp.md
```diff
@@ -42,7 +42,6 @@ The NNS Dapp is released through proposals in the Network Nervous System. Theref
 #### Security
 
 ## Proposal 124787
```
lmuntaner, line 44: Can you move the entry "* Select destination when disbursing maturity." above?

It's below the Proposal 124787 but it should be in Unreleased

bitdivine: Done

lmuntaner (approved):
Thanks!

PR #3378, opened 2023-09-26
Check Rust spelling in CI

bitdivine:
# Motivation
Tooling for checking for spelling mistakes has been in the `nns-dapp` repository for some weeks now.  The checks for spelling mistakes in `rustdoc` have been consistently near zero and the tooling seems stable, so it is ripe for use in CI.

# Changes
- Add `cargo spellcheck` to CI.

# Tests
- See CI

# Todos


rs/backend/src/accounts_store/schema.rs
```diff
@@ -97,7 +97,7 @@ pub trait AccountsDbTrait {
 /// Note: The numeric representations of these labels are guaranteed to be stable.
 #[repr(u32)]
 pub enum SchemaLabel {
-    /// Data is stored on the heap in a BTreeMap and serialized to stable memory on upgrade.
+    /// Data is stored on the heap in a `BTreeMap` and serialized to stable memory on upgrade.
```
bitdivine, line 100: The spellchecker complains when code isn't  quoted.

.github/workflows/checks.yml
```diff
@@ -33,6 +33,19 @@ jobs:
...
                   exit 1
           }
+  spelling:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install spellcheck
+        run: |
+          cd /
```
nmattia, line 43: nit: why move to the root?

.github/workflows/checks.yml
```diff
@@ -33,6 +33,19 @@ jobs:
...
+          sudo apt-get update
+          sudo apt-get install -yy hunspell hunspell-en-gb hunspell-en-us clang libclang-dev
+          cargo install --target x86_64-unknown-linux-gnu cargo-spellcheck
```
nmattia, line 46: nit: isn't the target the default? this seems to make the code less portable for no apparent reason

nmattia (approved):
LGTM, thanks!

.github/workflows/checks.yml
bitdivine: It defaults to `musl` which is normally a fine choice but in this particular case it fails to compile.

bitdivine:
Could you re-approve, please, @nmattia ?  I had to merge in main and fix more spelling and under the merge rules we are trialling, that requires a fresh review.

nmattia (approved):

PR #3379, opened 2023-09-26
Chore: ProjectCommitmentPO

lmuntaner:
# Motivation

Improve testing with a new PO for ProjectCommitment component.

# Changes

* Wrap ProjectCommitment with TestIdWrapper.
* New PO: ProjectCommitmentPo.
* Trim the value in getValueText of KeyValuePairPo and change the return value. It was wrong always saying `null`. I don't know why TS didn't catch it.
* Add new params to `createSummary` helper.
* Use new PO and `createSummary` helper in ProjectCommitment.spec file.

# Tests

Only test changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
+++ b/frontend/src/lib/components/project-detail/ProjectCommitment.svelte
@@ -19,6 +19,7 @@
     isFullProjectCommitmentSplit,
     type ProjectCommitmentSplit,
   } from "$lib/utils/projects.utils";
+  import TestIdWrapper from "../common/TestIdWrapper.svelte";
 
   const { store: projectDetailStore } = getContext<ProjectDetailContext>(
     PROJECT_DETAIL_CONTEXT_KEY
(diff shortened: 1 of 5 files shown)


frontend/src/lib/components/project-detail/ProjectCommitment.svelte
```diff
@@ -55,69 +56,73 @@
   });
 </script>
 
-{#if nonNullish(saleBuyerCount)}
-  <KeyValuePair testId="sns-project-current-sale-buyer-count">
+<TestIdWrapper testId="project-commitment-component">
```
lmuntaner, line 59: The only change in this file is wrapping it with TestIdWrapper.

GH is not very good at explaining these kind of changes.

frontend/src/tests/lib/components/project-detail/ProjectCommitment.spec.ts
```diff
@@ -81,140 +78,94 @@ describe("ProjectCommitment", () => {
...
-      Number(directCommitment + nfCommitment)
+    const po = renderComponent(summary);
+    const progressBarPo = po.getCommitmentProgressBarPo();
+    expect(await progressBarPo.getTotalCommitmentE8s()).toBe(
+      directCommitment + nfCommitment
     );
   });
 
-  it("should not render detailed participation if neurons fund participation is not available", () => {
+  it("should render a progress bar with different participations", async () => {
```
lmuntaner, line 112: This test is new as agreed in another PR.

frontend/src/tests/page-objects/KeyValuePair.page-object.ts
```diff
@@ -16,7 +17,8 @@ export class KeyValuePairPo extends BasePageObject {
     return this.root.querySelector("dt").getText();
   }
 
-  async getValueText(): Promise<string> {
-    return this.root.querySelector("dd").getText();
+  async getValueText(): Promise<string | null> {
```
lmuntaner, line 20: This wasn't caught by TS but it because a problem when I tried to trim the value.

lmuntaner:
@dskloetd please review

frontend/src/tests/page-objects/KeyValuePair.page-object.ts
```diff
@@ -16,7 +17,8 @@ export class KeyValuePairPo extends BasePageObject {
...
+    const text = await this.root.querySelector("dd").getText();
+    return nonNullish(text) ? text.trim() : null;
```
dskloetd, line 22: Trimming the value is necessary because the `KeyValuePair` template is wrong.
It has
```
  <dd>
    <slot name="value" />
  </dd>
```
but it should be something like
```
  <dd
    ><slot name="value"/></dd
  >
```
Not for this PR but would be good to fix some time and then remove this trim().
It's very similar to https://github.com/dfinity/nns-dapp/pull/3236/files

dskloetd (approved):
Thanks!

frontend/src/tests/page-objects/KeyValuePair.page-object.ts
lmuntaner: Thanks! I'll take a look in another PR then.

lmuntaner: I tested it in gix-components, and the problem doesn't happen there. See screenshot below.

Screenshot 2023-09-26 at 13 20 16

It happens when the value is split into two elements. For example, when the value is from the AmountDisplay component, it uses one span for the value, one for the token symbol.

See the screenshot below that when we get the content of the whole, it comes with the extra space. If we get the value of each span individually, there is no such space. See screenshot below.

Screenshot 2023-09-26 at 13 25 54

I'm not sure where this comes from. Any ideas?

dskloetd: Yes, then it comes from whatever is in the slot, so the `AmountDisplay`.
Here's the fix: https://github.com/dfinity/nns-dapp/pull/3391

PR #3381, opened 2023-09-26
Get binstall version from dfx.json

bitdivine:
# Motivation
There have been several breakages in the last few hours caused by the latest `binstall` being unavailable.

# Changes
* Specify the version of `binstall` in `dfx.json`, with the other tool versions.
* Create an action to install `binstall` with the specified version and use it, rather than other methods, wherever `binstall` is installed in GitHub workflows.
* Elsewhere, use the version from `dfx.json` directly.

# Tests
- See CI

# Todos


lmuntaner (approved):
Thanks!

PR #3385, opened 2023-09-26
More stats

bitdivine:
# Motivation
We would like to model accounts more accurately, for the purpose of sizing the account store optimally.  Our current stats are useful but deficient in some areas:

We have a distribution of transactions per subaccount, but don't have a count of subaccount transactions per main account.  For hardware wallets we have even less data.

# Changes
* Add stats for:
  * `total_sub_account_transactions` per main account.
  * `hardware_wallet_transactions` per per hardware wallet.
  * `hardware_wallet_transactions` per main acccount.

# Tests
The histogram tests have been extended to account for these additional statistics.

# Todos

- [x] Update the `.did` file.


lmuntaner (approved):
LGTM!

PR #3386, opened 2023-09-27
GIX-1859: New components for the new Canister Details page header

lmuntaner:
# Motivation

Change the canister details header to be more consistent with the new header in neuron details.

In this PR, I introduce, but not use the components needed to render it.

In another PR, I'll use them in the CanisterDetails and I'll change the tests in there as well.

# Changes

* New component CanisterHeadingTitle.
* New component CanisterPageHeader.
* New component CanisterPageHeading that uses new component CanisterHeadingTitle.
* Check whether the "subtitle" slot is present in the PageHeading. Otherwise, there is an empty space because it uses flex and gap with an empty element.

# Tests

* Test files and POs for CanisterHeadingTitle, CanisterPageHeader, CanisterPageHeading

# Todos

Not yet. I'll add it with the PR that uses the components.

### DIFF
--- a/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
+++ b/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
@@ -4,31 +4,33 @@
   import type { CanisterDetails } from "$lib/canisters/ic-management/ic-management.canister.types";
   import { i18n } from "$lib/stores/i18n";
   import { SkeletonText } from "@dfinity/gix-components";
+  import TestIdWrapper from "../common/TestIdWrapper.svelte";
 
   export let canisterDetails: CanisterDetails | undefined;
-  export let controller: boolean | undefined;
+  export let isController: boolean | undefined;
 </script>
 
-{#if controller === undefined}
-  <div class="skeleton">
-    <SkeletonText tagName="h1" />
-  </div>
-{:else if !controller}
-  <h1>Balance unavailable</h1>
-{:else if nonNullish(canisterDetails)}
-  <AmountDisplay
-    amount={TokenAmount.fromE8s({
-      amount: canisterDetails.cycles,
-      token: { name: "cycles", symbol: $i18n.canisters.t_cycles },
-    })}
-    size="huge"
-    singleLine
-  />
-{:else}
-  <div class="skeleton">
-    <SkeletonText tagName="h1" />
-  </div>
-{/if}
+<TestIdWrapper testId="canister-heading-title-component">
+  {#if nonNullish(canisterDetails)}
+    <AmountDisplay
+      amount={TokenAmount.fromE8s({
+        amount: canisterDetails.cycles,
+        token: { name: "cycles", symbol: $i18n.canisters.t_cycles },
+      })}
+      size="huge"
+      singleLine
+    />
+    <!-- Only when we have loaded the data and we know whether the user is the controller -->
+  {:else if isController === false}
+    <h1 data-tid="caniter-title-balance-unavailable">
+      {$i18n.canister_detail.balance_unavailable}
+    </h1>
+  {:else}
+    <div data-tid="skeleton" class="skeleton">
+      <SkeletonText tagName="h1" />
+    </div>
+  {/if}
+</TestIdWrapper>
 
 <style lang="scss">
   @use "@dfinity/gix-components/dist/styles/mixins/media";
--- a/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
+++ b/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
@@ -1,20 +1,20 @@
 <script lang="ts">
   import { TokenAmount, nonNullish } from "@dfinity/utils";
-  import AmountDisplay from "../ic/AmountDisplay.svelte";
+  import AmountDisplay from "$lib/components/ic/AmountDisplay.svelte";
   import type { CanisterDetails } from "$lib/canisters/ic-management/ic-management.canister.types";
   import { i18n } from "$lib/stores/i18n";
   import { SkeletonText } from "@dfinity/gix-components";
   import TestIdWrapper from "../common/TestIdWrapper.svelte";
 
-  export let canisterDetails: CanisterDetails | undefined;
+  export let details: CanisterDetails | undefined;
   export let isController: boolean | undefined;
 </script>
 
 <TestIdWrapper testId="canister-heading-title-component">
-  {#if nonNullish(canisterDetails)}
+  {#if nonNullish(details)}
     <AmountDisplay
       amount={TokenAmount.fromE8s({
-        amount: canisterDetails.cycles,
+        amount: details.cycles,
         token: { name: "cycles", symbol: $i18n.canisters.t_cycles },
       })}
       size="huge"
@@ -40,6 +40,7 @@
   }
 
   .skeleton {
+    // This is a width for the skeleton that looks good on desktop and mobile.
     width: 300px;
     max-width: 90%;
   }
--- a/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
+++ b/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
@@ -1,0 +1,43 @@
+<script lang="ts">
+  import { TokenAmount, nonNullish } from "@dfinity/utils";
+  import AmountDisplay from "../ic/AmountDisplay.svelte";
+  import type { CanisterDetails } from "$lib/canisters/ic-management/ic-management.canister.types";
+  import { i18n } from "$lib/stores/i18n";
+  import { SkeletonText } from "@dfinity/gix-components";
+
+  export let canisterDetails: CanisterDetails | undefined;
+  export let controller: boolean | undefined;
+</script>
+
+{#if controller === undefined}
+  <div class="skeleton">
+    <SkeletonText tagName="h1" />
+  </div>
+{:else if !controller}
+  <h1>Balance unavailable</h1>
+{:else if nonNullish(canisterDetails)}
+  <AmountDisplay
+    amount={TokenAmount.fromE8s({
+      amount: canisterDetails.cycles,
+      token: { name: "cycles", symbol: $i18n.canisters.t_cycles },
+    })}
+    size="huge"
+    singleLine
+  />
+{:else}
+  <div class="skeleton">
+    <SkeletonText tagName="h1" />
+  </div>
+{/if}
+
+<style lang="scss">
+  @use "@dfinity/gix-components/dist/styles/mixins/media";
+
+  .skeleton {
+    width: 90%;
+
+    @include media.min-width(small) {
+      width: 40%;
+    }
+  }
+</style>
--- a/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
+++ b/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
@@ -35,11 +35,12 @@
 <style lang="scss">
   @use "@dfinity/gix-components/dist/styles/mixins/media";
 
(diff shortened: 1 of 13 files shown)


lmuntaner:
New UI:

Explanation of the components:
canister-header-components

Example when the user is controller and the canister has a name:
controller-name

Example when the user is the controller but the canister has no name:
controller-not-name

Example when the user is not the controller and the canister has a name:
not-controller-name

Example when the user is not the controller and the canister has no name:
not-controller-not-name

lmuntaner:
@dskloetd please review

peterpeterparker:
UI/UX feedback if I may:

- In my opinion, the text "Balance unavailable" can lead the user to ask him/her self if there is an issue with the IC or the canister. So I would rephrase with a message that does not imply issue or add an explanation "Balance unavailable. Only controllers can fetch the status of the canister"

- Have you try to display the canister name closer to the balance, like a subtitle, and having more space underneath? With the current spacing it feels like the name is related to the action buttons.

lmuntaner:
> UI/UX feedback if I may:
> 
> * In my opinion, the text "Balance unavailable" can lead the user to ask him/her self if there is an issue with the IC or the canister. So I would rephrase with a message that does not imply issue or add an explanation "Balance unavailable. Only controllers can fetch the status of the canister"

Good point. I will change the text below (not part of the component) to explain it better.

> * Have you try to display the canister name closer to the balance, like a subtitle, and having more space underneath? With the current spacing it feels like the name is related to the action buttons.

The spacings are the same, the problem comes with how text has extra white space because of line-height and because the name used in the screenshot is all lowercase. Leaving more space above than below.

I'll bring it up with @artkorotkikh-dfinity and maybe we can put the title and subtitle of the page heading one spacing and another one for the action / tags.

artkorotkikh-dfinity:
Thanks @lmuntaner! 

Visually, the only thing that grabbed my attention - there is less space between buttons and "subtitle" than between "title" and "subtitle". This kind of breaks the grouping of visual elements. 

Is it possible to add more space? Attaching example, also feel free to double check Figma values (I know they might be a bit different from what we have in the code, but they should give a good picture for spacings) 
 
image

lmuntaner:
@peterpeterparker ready for another review!

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
```diff
@@ -0,0 +1,46 @@
+<script lang="ts">
+  import { TokenAmount, nonNullish } from "@dfinity/utils";
+  import AmountDisplay from "../ic/AmountDisplay.svelte";
+  import type { CanisterDetails } from "$lib/canisters/ic-management/ic-management.canister.types";
+  import { i18n } from "$lib/stores/i18n";
+  import { SkeletonText } from "@dfinity/gix-components";
+  import TestIdWrapper from "../common/TestIdWrapper.svelte";
+
+  export let canisterDetails: CanisterDetails | undefined;
```
peterpeterparker, line 9: idea: prop name can be shortened to `details`

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
```diff
@@ -0,0 +1,46 @@
+<script lang="ts">
+  import { TokenAmount, nonNullish } from "@dfinity/utils";
+  import AmountDisplay from "../ic/AmountDisplay.svelte";
```
peterpeterparker, line 3: use `$lib` instead of '../'

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
```diff
@@ -0,0 +1,46 @@
...
+
+<style lang="scss">
+  @use "@dfinity/gix-components/dist/styles/mixins/media";
+
+  h1 {
+    margin: 0;
+  }
+
+  .skeleton {
+    width: 300px;
```
peterpeterparker, line 43: where does these values come from? why 300px and why 90%?

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
```diff
@@ -0,0 +1,46 @@
...
+        amount: canisterDetails.cycles,
+        token: { name: "cycles", symbol: $i18n.canisters.t_cycles },
+      })}
+      size="huge"
+      singleLine
+    />
+    <!-- Only when we have loaded the data and we know whether the user is the controller -->
+  {:else if isController === false}
+    <h1 data-tid="caniter-title-balance-unavailable">
+      {$i18n.canister_detail.balance_unavailable}
```
peterpeterparker, line 26: should we also rename the i18n key as we reworked its text?

frontend/src/lib/components/canister-detail/CanisterPageHeader.svelte
```diff
@@ -0,0 +1,28 @@
...
+  import PageHeader from "../common/PageHeader.svelte";
+  import IdentifierHash from "../ui/IdentifierHash.svelte";
+  import UniversePageSummary from "../universe/UniversePageSummary.svelte";
+
+  export let canister: CanisterInfo;
+</script>
+
+<PageHeader testId="canister-page-header-component">
+  <UniversePageSummary slot="start" universe={NNS_UNIVERSE} />
+  <span
```
peterpeterparker, line 13: semantically speaking, is it allowed to have a title - `IdentifierHash` === `h3` per default - within a `span`?

frontend/src/lib/components/canister-detail/CanisterPageHeading.svelte
```diff
@@ -0,0 +1,34 @@
...
+  import UnlinkCanisterButton from "./UnlinkCanisterButton.svelte";
+  import RenameCanisterButton from "./RenameCanisterButton.svelte";
+  import CanisterHeadingTitle from "./CanisterHeadingTitle.svelte";
+
+  export let canisterDetails: CanisterDetails | undefined;
+  export let canister: CanisterInfo;
+  export let isController: boolean | undefined;
+</script>
+
+<!-- We can't set conditional slots. -->
```
peterpeterparker, line 14: why not using a`svelte:fragment`?

frontend/src/lib/components/common/PageHeading.svelte
```diff
@@ -3,14 +3,19 @@
 
   let hasTags: boolean;
   $: hasTags = $$slots.tags !== undefined;
+
+  let hasSubtitle: boolean;
+  $: hasSubtitle = $$slots.subtitle !== undefined;
```
peterpeterparker, line 8: `nonNullish`?

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
lmuntaner: We didn't rework this text. But the one below. That will happen in another PR.

This is a title, explaining that only the controller can see the cycles will be part of the text below.

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
lmuntaner: Nowhere, just a width good enough for the skeleton to look good.

Because it was a skeleton and I wanted a fixed width I considered it was ok.

I'll add a comment. But I'm also happy to change and do it somehow else if you have any suggestion.

frontend/src/lib/components/canister-detail/CanisterPageHeading.svelte
lmuntaner: Because the fragment is still true with the slot. And I want to avoid the slot being true to avoid an extra spacing.

I believe the problem is forcing the PageHeading subtitle slot inside an `h4` but then expecting a slot instead of a text.

Maybe we can discuss this in another PR.

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
peterpeterparker: Thanks for the explanation. Maybe use a `calc(100% - var(--padding-2x))` or something for the `max-width` which I would personally read as though to match a particular size related to the grid design.

I'm fine with 300, can maybe be replaced by using `$breakpoint-xsmall` but all good.

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
peterpeterparker: Gotcha

frontend/src/lib/components/canister-detail/CanisterPageHeading.svelte
peterpeterparker: That's the current code

```
{#if canister.name.length === 0}
  <PageHeading testId="canister-page-heading-component">
    <CanisterHeadingTitle
      slot="title"
      details={canisterDetails}
      {isController}
    />
    <svelte:fragment slot="tags">
      <UnlinkCanisterButton canisterId={canister.canister_id} />
      <RenameCanisterButton />
    </svelte:fragment>
  </PageHeading>
{:else}
  <PageHeading testId="canister-page-heading-component">
    <CanisterHeadingTitle
      slot="title"
      details={canisterDetails}
      {isController}
    />
    <span slot="subtitle" data-tid="subtitle">
      {canister.name}
    </span>
    <svelte:fragment slot="tags">
      <UnlinkCanisterButton canisterId={canister.canister_id} />
      <RenameCanisterButton />
    </svelte:fragment>
  </PageHeading>
{/if}
```

That's how it can be simplified I think:

```
<PageHeading testId="canister-page-heading-component">
    <CanisterHeadingTitle
            slot="title"
            details={canisterDetails}
            {isController}
    />
    <svelte:fragment>
        {#if canister.name.length === 0}
            <span slot="subtitle" data-tid="subtitle">
          {canister.name}
        {/if}
    </span>
    </svelte:fragment>
    <svelte:fragment slot="tags">
        <UnlinkCanisterButton canisterId={canister.canister_id} />
        <RenameCanisterButton />
    </svelte:fragment>
</PageHeading>
```

Second avoid lots of duplicatation and feel more readable to me.

peterpeterparker: Or I am missing something?

peterpeterparker: > Because the fragment is still true with the slot. And I want to avoid the slot being true to avoid an extra spacing.

Ah gotcha now

peterpeterparker: > Maybe we can discuss this in another PR.

Yeah would be great to improve this in another PR.

peterpeterparker (approved):
LGTM

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
lmuntaner: done

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
lmuntaner: done

frontend/src/lib/components/canister-detail/CanisterPageHeader.svelte
lmuntaner: good point, I used a `div`

frontend/src/lib/components/common/PageHeading.svelte
lmuntaner: done

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
lmuntaner: done!

PR #3388, opened 2023-09-27
Fix past-changelog-test.test on Mac and test more scripts on macos on CI

dskloetd:
# Motivation

Somehow our scripts keep changing in ways that make them break on Mac.
`past-changelog-test.test` was broken by https://github.com/dfinity/nns-dapp/pull/3176

# Changes

1. Revert the change to `past-changelog-test.test` from https://github.com/dfinity/nns-dapp/pull/3176. I've tried using double quotes with escaping, which worked on my laptop but didn't work on CI. The [documentation](https://www.shellcheck.net/wiki/SC1004) of the warning says that it was retired so it seems safe to ignore.
2. Run the test on macos on CI.
3. Run some other tests also on macos on CI.

It would be good to run even more of these tests on macos on CI but others require more changes so they'll have to wait for another time.

# Tests

CI

# Todos


.github/workflows/checks.yml
```diff
@@ -180,9 +180,18 @@ jobs:
         run: scripts/mk_nns_patch.test
   config-check:
     name: Config is as expected
-    runs-on: ubuntu-20.04
+    strategy:
```
lmuntaner, line 183: Can you change the title of the PR?

Or do a PR only with the fix and adding that test to the CI. This PR adds many other tests to the CI.

lmuntaner (approved):
LGTM, adding a comment.

.github/workflows/checks.yml
dskloetd: I've updated the title of the PR.

PR #3391, opened 2023-09-27
Remove extra space from AmountDisplay

dskloetd:
# Motivation

https://github.com/dfinity/nns-dapp/pull/3379 had to add a `trim()` which shouldn't be necessary.
It was necessary because `AmountDisplay` renders an unnecessary space.
`AmountDisplay` has an optional copy icon. If the copy icon is rendered, we want there to be a space between the amount and the copy icon. But if there is no copy icon, there also shouldn't be a space.
I think this might be a bug in Svelte template parsing but we can work around it.

# Changes

1. Render the space separating the amount and the copy icon inside the `{#if copy}` block.
2. Remove the `trim()` from `frontend/src/tests/page-objects/KeyValuePair.page-object.ts`

# Tests

Pass
Also manually checked that it renders correctly with copy icon (in the wallet summary) and without (in the transaction history).
Try it here: https://fswcu-uqaaa-aaaaa-aadwq-cai.nnsdapp.dfinity.network/

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/ic/AmountDisplay.svelte
+++ b/frontend/src/lib/components/ic/AmountDisplay.svelte
@@ -33,9 +33,9 @@
     class:tabular-num={detailed === "height_decimals"}
     >{`${sign}${formatToken({ value: amount.toE8s(), detailed })}`}</span
   >
-  <span class="label">{label !== undefined ? label : amount.token.symbol}</span>
-
-  {#if copy}
+  <span class="label">{label !== undefined ? label : amount.token.symbol}</span
+  >{#if copy}
+    {" "}
     <Copy value={formatToken({ value: amount.toE8s(), detailed: true })} />
   {/if}
 </div>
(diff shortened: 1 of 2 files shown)


lmuntaner (commented):
Wouldn't it be safer to wrap the `Copy` with a `<span>` and add a `margin-left` to it?

dskloetd:
> Wouldn't it be safer to wrap the Copy with a <span> and add a margin-left to it?

I don't know. That's a different kind of change. I'm just trying to avoid a space that shouldn't be there in the first place.

dskloetd:
Also note that it should be aligned with the address below it. So if you want to change it, it will cascade to other components.
image

dskloetd:
I don't know. It's hard to guarantee without having the same DOM structure. Maybe on one browser a space is 3 pixels and on another browser it's 4 pixels? Maybe it depends on which fonts are available? Maybe we decide to change the font one day?

But it's orthogonal to this change. We already have a space in the DOM there. I'm not changing that. I'm just removing it in the case where we don't have the copy icon.

lmuntaner (approved):
Fair point, this PR just removes the need to trim.

PR #3395, opened 2023-09-27
Feat: Job to ONLY update IC commit

lmuntaner:
# Motivation

We'd like a CI job to update the types of the SNS Aggregator.

# Changes

* Rename script update_ic_commit to set-ic-commit-update-agg-types
* New script update-ic-commit that only changes the SNS Aggregator types and sets the IC_COMMIT in dfx.json.
* New Github workflow that uses `update-ic-commit`.

# Tests

Not specifically for these scripts.

# Todos


scripts/set-ic-commit-update-agg-types
dskloetd, line 1: Not necessarily for this PR but:
I had a look at the file to understand what it does to see if the new name made sense and I noticed that the help text says
```
	Gets stable data from an sns_aggregator.
```
I'm not sure what that means, but given that there are 2 other files with the same help text, I'm guessing it wasn't supposed to say that:
```
$ git grep "Gets stable data from an sns_aggregator."
scripts/sns/aggregator/get_log: Gets stable data from an sns_aggregator.
scripts/sns/aggregator/get_stable_data: Gets stable data from an sns_aggregator.
scripts/update_ic_commit:       Gets stable data from an sns_aggregator.
```

scripts/update-ic-commit
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+set -euo pipefail
+SOURCE_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
+PATH="$SOURCE_DIR:$PATH"
+
```
dskloetd, line 5: Can you define a help text that explains what the script does?

scripts/update-ic-commit
```diff
@@ -0,0 +1,39 @@
...
+# Source the clap.bash file ---------------------------------------------------
+source "$SOURCE_DIR/clap.bash"
+# Define options
+clap.define short=d long=dir desc="The path to an snsdemo repo checked out at the desired release" variable=IC_REPO
```
dskloetd, line 9: Do you mean ic repo instead of snsdemo repo?

scripts/update-ic-commit
```diff
@@ -0,0 +1,39 @@
...
+# Source the output file ----------------------------------------------------------
+source "$(clap.build)"
+
+# Check that the snsdemo repository directory exists
```
dskloetd, line 13: Do you mean "ic repository" instead of "snsdemo repository"?

scripts/update-ic-commit
```diff
@@ -0,0 +1,39 @@
...
+IC_REPO="${IC_REPO:-${1:-}}"
+test -e "${IC_REPO:-}" || {
+  echo "ERROR: Please supply the path to the snsdemo repository with e.g. --dir ../some/path/" >&2
```
dskloetd, line 16: same?

scripts/update-ic-commit
```diff
@@ -0,0 +1,39 @@
...
+  exit 1
+} >&2
+
+# Check that the SNSDEMO repo is checked out at the correct release
```
dskloetd, line 20: same?

scripts/update-ic-commit
```diff
@@ -0,0 +1,39 @@
...
+pushd "$IC_REPO"
+IC_COMMIT="$(git rev-parse HEAD)"
+
+# Check that all values were found.
+for var in IC_COMMIT; do
```
dskloetd, line 25: I'm not sure what the intention if of this "loop" but `$var` will only have have 1 value and that's the string `"IC_COMMIT"`.
Then
```
  test -n "${!var}" || {
    echo "ERROR: Could not find $var" >&2
    exit 1
  } >&2
```
is just a more arcane way of writing
```
  if [ "${!var}" = "" ]; then
    echo "ERROR: Could not find $var" >&2
    exit 1
  fi
```
And because `$var` will only have the value `"IC_COMMIT"`, this whole loop can just be replaced with:
```
if [ "$IC_COMMIT" = "" ]; then
  echo "ERROR: Could not find IC_COMMIT" >&2
  exit 1
fi
```

scripts/update-ic-commit
```diff
@@ -0,0 +1,39 @@
...
+  test -n "${!var}" || {
+    echo "ERROR: Could not find $var" >&2
+    exit 1
+  } >&2
```
dskloetd, line 29: The only echo in this block is already done to stdout (`&2`), so this `>&2` here is redundant.

scripts/update-ic-commit
```diff
@@ -0,0 +1,39 @@
...
+done
+
+# Move to the root of the nns-dapp repository
+cd "$SOURCE_DIR/.."
+
+# Update the IC commit to match snsdemo
+scripts/set-ic-commit-update-agg-types --ic_commit "$IC_COMMIT"
```
dskloetd, line 36: Huh, I thought this script was made separate from `set-ic-commit-update-agg-types`, because that script does 2 things and this script should only do 1 thing (hence the shorter name). But now I see that this script calls `set-ic-commit-update-agg-types` so now I'm confused about the goal of this script.

.github/workflows/update-ic-commit-sns-aggregator.yml
```diff
@@ -0,0 +1,62 @@
...
+          body: |
+            # Motivation
+            We would like to update the types in the SNS Aggregator.
+
+            # Changes
+
+            * Candid files in the `declarations` directory.
+            * Update types and patches in `rs/sns_aggregator/src/types`.
+            * TODO: The code will probably NOT compile if there have been significant changes to the NNS-dapp APIs.
+              Please check the Rust code and make minimal changes to make the code compile.
```
dskloetd, line 59: If manual changes need to be made, the PR shouldn't be presented as made by a bot.
Let's provide instructions for what to do (what script to call, etc.) if a PR with manual changes needs to be made and instruct the use make their own PR and close this one.

Or maybe we can automatically test that the PR will require manual changes (because the code doesn't compile or something?) and then skip creating the PR (which will be closed anyway) and send a message with instructions directly to Slack?

.github/workflows/update-ic-commit-sns-aggregator.yml
```diff
@@ -0,0 +1,62 @@
...
+          git config --global user.name "GIX bot"
+          git commit -a -m "Update IC Commit"
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GIX_BOT_PAT }}
+          commit-message: Update IC Commit
+          committer: GitHub <noreply@github.com>
+          author: gix-bot <gix-bot@users.noreply.github.com>
+          branch: bot-ic-update
```
dskloetd, line 46: Using a fixed branch will be problematic if the workflow runs again while the previous branch/PR is still open.
Let's put a date/timestamp in the branch name.

.github/workflows/update-ic-commit-sns-aggregator.yml
```diff
@@ -0,0 +1,62 @@
...
+
+            # Changes
+
+            * Candid files in the `declarations` directory.
+            * Update types and patches in `rs/sns_aggregator/src/types`.
+            * TODO: The code will probably NOT compile if there have been significant changes to the NNS-dapp APIs.
+              Please check the Rust code and make minimal changes to make the code compile.
+
+            # Tests
+              - [ ] Please check the API updates for any breaking changes that affect our code.
```
dskloetd, line 62: What does this mean?

.github/workflows/update-ic-commit-sns-aggregator.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Get IC repo
+        uses: actions/checkout@v3
+        with:
+          repository: 'dfinity/ic'
+          path: 'ic'
+      - name: Update IC commit and SNS Aggregator types
+        run: |
+          set -x
+          # Install sponge
+          sudo apt-get update -yy && sudo apt-get install -yy moreutils && command -v sponge
```
dskloetd, line 31: Optional:
FYI The functionality of sponge can be provided by a 4 line function.
https://github.com/dfinity/nns-dapp/blob/a0f4c9b1632bd027cc6320388783c428852f639f/scripts/nns-dapp/split-changelog#L41
So installing sponge can be easily avoided if we want.

.github/workflows/update-ic-commit-sns-aggregator.yml
dskloetd, line 59: Let's include the exact command to compile the code as a reminder and for easy copy/pasting.

.github/workflows/update-ic-commit-sns-aggregator.yml
```diff
@@ -0,0 +1,62 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install didc
+        run: |
+          USER_BIN="$HOME/.local/bin"
+          mkdir -p "$USER_BIN"
+          echo "$USER_BIN" >> $GITHUB_PATH
+          version="$(jq -r .defaults.build.config.DIDC_VERSION dfx.json)"
+          # TODO: Make `didc` support `binstall`, then use `binstall` here.
+          curl -Lf "https://github.com/dfinity/candid/releases/download/${version}/didc-linux64" | install -m 755 /dev/stdin "$USER_BIN/didc"
```
dskloetd, line 21: Not for this PR but:
I think we have quite a few workflows that install didc.
It could be nice to make 1 github action to standardize it.
That will also make it easy to switch over all of them once binstall is supported.

bitdivine:
Note that sometimes the snsdemo and IC commit need to be updated together, in which case these separate actions would have to be run simultaneously, which is practically impossible.  A better option might be to provide options to the existing update function, so that the caller can choose to update one, the other or both.  Also, that would be a PR of about 5 lines.

.github/workflows/update-ic-commit-sns-aggregator.yml
bitdivine: Note that that polyfill leaves behind the tempfile if the pipe fails.  Doesn't matter in most circumstances but I would be wary of casually using it everywhere.

lmuntaner:
I'm closing this PR.

After discussing with @dskloetd and @bitdivine, the first step is to list all the jobs we want to automate. Then, which automation level we want in each, and then which priority we give to each of them.

PR #3398, opened 2023-09-27
Update SNS aggregator changelog

bitdivine:
# Motivation
A release was made of the SNS aggregator; the unreleased changelog entries have now been released.

I note that the earlier changelog entries are somewhat broken; a proposal is listed twice.  However I propose to make the release cut and then fix the earlier proposal entry in a separate PR.

# Changes
- Cut the changelog for the last SNS aggregator release.

# Tests
The proposal in the section title is correct and the link is the right one.

# Todos

  - Not applicable


lmuntaner (approved):
Thanks!

PR #3399, opened 2023-09-27
Escape raw html tag in markdown by default

mstrasinskis:
# Motivation

Escape raw html tags in markdown content to increase the security.

# Changes

- add optional `escapeRawHtmlTags` to `markdownToHTML` function

# Tests

TBD

# Todos

# Screenshots

There is a raw html in the [provided sample](https://dashboard.internetcomputer.org/proposal/104084):

## before
Screenshot 2023-09-27 at 15 20 21

## after

| NNS-DAPP | Dashboard |
|--------|--------|
| image | image |

### DIFF
--- a/frontend/src/lib/utils/html.utils.ts
+++ b/frontend/src/lib/utils/html.utils.ts
@@ -51,15 +51,28 @@ export const renderer = (marked: Marked): Renderer => {
   return renderer;
 };
 
+export const escapeHtml = (html: string): string =>
+  html.replace(/</g, "&lt;").replace(/>/g, "&gt;");
+
 /**
- * Uses markedjs
+ * Uses markedjs.
+ * Escape raw HTML tags by default (<script> -> &lt;script&gt;)
  * @see {@link https://github.com/markedjs/marked}
  */
-export const markdownToHTML = async (text: string): Promise<string> => {
+export const markdownToHTML = async ({
+  text,
+  escapeRawHtmlTags = true,
+}: {
+  text: string;
+  escapeRawHtmlTags?: boolean;
+}): Promise<string> => {
+  if (escapeRawHtmlTags)
+    text = text.replaceAll("<", "&lt;").replaceAll(">", "&gt;");
+
   const url = "/assets/libs/marked.min.js";
   // The dynamic import cannot be analyzed by Vite. As it is intended, we use the /* @vite-ignore */ comment inside the import() call to suppress this warning.
   const { marked }: { marked: Marked } = await import(/* @vite-ignore */ url);
-  return marked(text, {
+  return marked(escapeRawHtmlTags ? escapeHtml(text) : text, {
     renderer: renderer(marked),
   });
 };
--- a/frontend/src/lib/components/ui/Markdown.svelte
+++ b/frontend/src/lib/components/ui/Markdown.svelte
@@ -8,7 +8,7 @@
   let error = false;
   const transform = async (text: string) => {
     try {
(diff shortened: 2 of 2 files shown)


frontend/src/lib/utils/html.utils.ts
```diff
@@ -42,24 +43,57 @@ export const imageToLinkRenderer = (
...
+ *
+ * @param marked
+ */
+const proposalSummaryRenderer = (marked: Marked): Renderer => {
   const renderer = new marked.Renderer();
-  // custom link renderer
+
   renderer.link = targetBlankLinkRenderer;
   renderer.image = imageToLinkRenderer;
+  renderer.html = htmlRenderer;
```
mstrasinskis, line 82: @peterpeterparker, the logic was updated. Now only `` tags are processed to harmonize approach with markdown images.

frontend/src/lib/utils/html.utils.ts
```diff
@@ -42,24 +43,57 @@ export const imageToLinkRenderer = (
...
 
-export const renderer = (marked: Marked): Renderer => {
+export const escapeHtml = (html: string): string =>
+  html.replace(/</g, "&lt;").replace(/>/g, "&gt;");
+
+/**
+ * Escape <img> tags or convert them to links
+ */
+const transformImg = (img: string): string => {
+  const src = img.match(/src="([^"]+)"/)?.[1];
```
peterpeterparker, line 53: What happens if other attributes - e.g. `loading="lazy"` or `class` - are provided?

frontend/src/lib/utils/html.utils.ts
```diff
@@ -42,24 +43,57 @@ export const imageToLinkRenderer = (
...
+  const imageHtml = shouldEscape
+    ? escapeHtml(img)
+    : imageToLinkRenderer(src, title, alt);
+
+  return imageHtml;
+};
+
+/** Avoid <img> tags; instead, apply the same logic as for markdown images by either escaping them or converting them to links. */
+export const htmlRenderer = (html: string): string =>
+  /<img\s+[^>]*>/gi.test(html) ? transformImg(html) : html;
```
peterpeterparker, line 66: I suck at regex: it works also if the `` has a closing `/` not only ``?

frontend/src/lib/utils/html.utils.ts
peterpeterparker: Good idea

peterpeterparker: That said, it works until the day someone add an image tag within a piece of code I guess.

frontend/src/lib/utils/html.utils.ts
mstrasinskis: It depends on the src value. If it's `data:image` the whole image tag will be escaped and rendered as text. In case of a url in src, only the link will be shown (line #57)

frontend/src/lib/utils/html.utils.ts
mstrasinskis: Yes, will work for both.

mstrasinskis:
@peterpeterparker, re-review please.

frontend/src/lib/utils/html.utils.ts
peterpeterparker: Coolio

frontend/src/lib/utils/html.utils.ts
peterpeterparker: Oki doki

PR #3402, opened 2023-09-28
IC release 2023 06 07 23 01

bitdivine:
# Motivation
The IC commit in the nns-dapp backend is from April; it needs to be updated.  Ditto with the aggregator.

Updating to a recent commit requires a very large number of changes.  More than can be reasoned about well.  I therefore propose to update gradually.  The IC repo releases are convenient stepping stones.

# Changes
* Update the nns-dapp  backend IC repo dependencies to IC repo `release-2023-06-07_23-01`.

# Consequences and fixes
### `RefreshBuyerTokensRequest` has a new field: `confirmation_text`.

Addressed by:
```
RefreshBuyerTokensRequest {
        confirmation_text: None,
...
```

## Resolver
```
0.271 warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`                                                                                                                                                                                                                                                                                                                                                                                                                                          
0.271 note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
0.271 note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest 
...
66.97 error[E0432]: unresolved import `crate::sys::IoSourceState`
```
Addresed by:
```
 [workspace]
+resolver = "2"
```
This alone did not solve the issue for `release-2023-05-18_23-01` so `workspace.package.resolver` is also set explicitly to 2.  My understanding is that this will be needed only during the transition of cargo packages.

## `SignedTokens` -> `Tokens`?
`SignedTokens` were removed in [this commit](https://github.com/dfinity/ic/commit/3466378e9dbc62a66e5fc433757735bac3703bcd):
```
commit 3466378e9dbc62a66e5fc433757735bac3703bcd
Author: Maciej Modelski <maciej.modelski@dfinity.org>
Date:   Thu Jun 1 09:02:38 2023 +0000

    feat(ledger_core): update approvals to icrc2 standard, remove SignedTokens
```
We use `SignedTokens` in stored transactions, so if we switch we need to ensure that upgrades to this new type will not break.

Update: ICRC2 `Tokens` are NOT compatible with the earlier `SignedTokens`.

### What is the difference in the types?
[`SignedTokens`](https://github.com/dfinity/ic/commit/3466378e9dbc62a66e5fc433757735bac3703bcd#diff-0e8883ac86598c0f2bc7f5a4b9f3eecbd36b5f0c58eb45557b072f83d17580b7L263) have both positive an negative `e8s` whereas [`Tokens`](https://docs.rs/ic-ledger-types/0.1.0/src/ic_ledger_types/lib.rs.html#35-37) have only positive `e8s`.

As such, `Tokens` can be cast to `SignedTokens` without loss, but not the other way around.

#### Proposal
Keep storing `SignedTokens`, as this is compatible with the data we have.  When required, cast `Tokens` to `SignedTokens`.

# Tests
See CI

# Todos

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -769,17 +769,17 @@ dependencies = [
  "dfn_core",
  "dfn_http_metrics",
  "dfn_protobuf",
- "ic-base-types",
+ "ic-base-types 0.8.0 (git+https://github.com/dfinity/ic?rev=release-2023-06-07_23-01)",
  "ic-certified-map 0.3.4",
  "ic-crypto-getrandom-for-wasm",
  "ic-crypto-tree-hash",
- "ic-ic00-types",
- "ic-ledger-core",
+ "ic-ic00-types 0.8.0 (git+https://github.com/dfinity/ic?rev=release-2023-06-07_23-01)",
+ "ic-ledger-core 0.8.0 (git+https://github.com/dfinity/ic?rev=release-2023-06-07_23-01)",
  "ic-metrics-encoder",
  "ic-nervous-system-common-build-metadata",
  "ic-nns-common",
  "ic-nns-constants",
- "ic-protobuf",
+ "ic-protobuf 0.8.0 (git+https://github.com/dfinity/ic?rev=release-2023-06-07_23-01)",
  "ic-types",
  "ic-xrc-types",
  "icp-ledger",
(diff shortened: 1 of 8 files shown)


lmuntaner (commented):
Changes LGTM so far.

bitdivine:
Outdated; the code has been refactored to make this easier.

PR #3411, opened 2023-09-29
GIX-1930: Use new fields from sns aggregator

lmuntaner:
# Motivation

We want to expose the new fields related to NF  and already in the SNS Aggregator response to the nns-dapp components.

In this PR, the converters of the SNS aggregator data use the fields in the response to populate the converted data, instead of returning empty arrays.

# Changes

* Add the new fields to the "CachedXxx" types.
* Create new CachedNeuronsFundParticipationConstraints and CachedLinearScalingCoefficient types.
* New converted `convertNeuronsFundParticipationConstraints`.
* Use the new fields in the relevant converters.

# Tests

* Test that an aggregator response with the new fields, includes them in the SnsSummary.

# Todos

Not necessary yet.

### DIFF
--- a/frontend/src/lib/types/sns-aggregator.ts
+++ b/frontend/src/lib/types/sns-aggregator.ts
@@ -95,11 +95,11 @@ export type CachedSwapParamsDto = {
   sale_delay_seconds?: number;
 };
 
-export interface CachedLinearScalingCoefficient {
+interface CachedLinearScalingCoefficient {
   slope_numerator?: number | null;
+  slope_denominator?: number | null;
   intercept_icp_e8s?: number | null;
   from_direct_participation_icp_e8s?: number | null;
-  slope_denominator?: number | null;
   to_direct_participation_icp_e8s?: number | null;
 }
 
--- a/frontend/src/lib/types/sns-aggregator.ts
+++ b/frontend/src/lib/types/sns-aggregator.ts
@@ -95,6 +95,20 @@ export type CachedSwapParamsDto = {
   sale_delay_seconds?: number;
 };
 
+export interface CachedLinearScalingCoefficient {
+  slope_numerator?: number | null;
+  intercept_icp_e8s?: number | null;
+  from_direct_participation_icp_e8s?: number | null;
+  slope_denominator?: number | null;
+  to_direct_participation_icp_e8s?: number | null;
+}
+
+export type CachedNeuronsFundParticipationConstraints = {
+  coefficient_intervals: Array<CachedLinearScalingCoefficient>;
+  max_neurons_fund_participation_icp_e8s?: number | null;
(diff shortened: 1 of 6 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/types/sns-aggregator.ts
```diff
@@ -95,6 +95,20 @@ export type CachedSwapParamsDto = {
   sale_delay_seconds?: number;
 };
 
+export interface CachedLinearScalingCoefficient {
+  slope_numerator?: number | null;
+  intercept_icp_e8s?: number | null;
+  from_direct_participation_icp_e8s?: number | null;
+  slope_denominator?: number | null;
```
dskloetd, line 102: What determines the order of these fields?
It seems odd not to thave `slope_numerator` and `slope_denominator` next to each other.

frontend/src/lib/types/sns-aggregator.ts
```diff
@@ -95,6 +95,20 @@ export type CachedSwapParamsDto = {
   sale_delay_seconds?: number;
 };
 
+export interface CachedLinearScalingCoefficient {
```
dskloetd, line 98: Do we need these in the frontend?
We aren't going to calculate the dynamic NF participation in the frontend, right?
I hope that's just provided by the backend canister, regardless of how it's calculated.

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
```diff
@@ -399,6 +399,81 @@ describe("sns aggregator converters utils", () => {
         convertDtoToSnsSummary(aggregatorMissingSwapParams)
       ).toBeUndefined();
     });
+
+    it("converts fields related to NF participation", () => {
+      const aggregatorNewNeuronsFundFields: CachedSnsDto = {
```
dskloetd, line 404: These fields will not always be "new", right? Do you plan to rename this later?

frontend/src/lib/types/sns-aggregator.ts
lmuntaner: I copied it from candid file. I moved them together.

frontend/src/lib/types/sns-aggregator.ts
lmuntaner: No, we're not using it. I removed the `export`, thanks!

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
lmuntaner: Good point, I renamed it.

lmuntaner:
@dskloetd please review

dskloetd (approved):

PR #3412, opened 2023-09-29
Proposals lib

bitdivine:
# Motivation
If we split the nns-dapp backend code into several crates it should be more manageable.  Potentially we could split out `accounts`, `neurons`, `proposals`, `canisters` and keep just the HTTP serving code in the main crate.  However here we have just an experiment with the proposals, as the proposals rendering needs to be updated to handle the latest and greatest proposal types.

# Changes
* Move the `proposals` module into a separate crate.
  * Remove unused dependencies from the `nns-dapp` `backend` `Cargo.toml`.
  * Move just the `get_proposal_info` method from `canisters::governance` into the proposals crate.
  * Update `Dockerfile` to match.

# Tests
See CI

# Todos

- [x] Check that any `chrono` lib imported by the `proposals` crate is a safe version; In the `nns-dapp` crate it is forced to a safe version.  (although it still has an audit warning)
- [x] Compare with [a build on main](https://github.com/dfinity/nns-dapp/actions/runs/6348394781):
  - [x] Wasm size - almost indistinguishable
  - [x] Cycle count


lmuntaner (commented):
I like the idea!!

To then move to its own canister would be easy!

mstrasinskis (approved):
Rubber stamp! Great idea!

mstrasinskis (approved):
Re-approve!

PR #3413, opened 2023-09-29
Fix: SkeletonHeading width

lmuntaner:
# Motivation

The width of the SkeletonHeading was almost zero.

Problem: the width was set as %. But the parent's width is determined by the content.

# Changes

* Set a width in pixels for title text in `SkeletonHeading`.

# Tests

Not necessary.

# Todos

- [ ] Add entry to changelog (if necessary).
Not necessary. New skeletons were recently introduced and this fix is covered by the entry of those.

PR #3414, opened 2023-09-29
GIX-1859: Implement new header Canister detail page

lmuntaner:
# Motivation

Change the header of the canister detail page to be more consistent with the new neuron details page header.

In this PR, we use the components introduced in #3386 in the CanisterDetails.svelte

# Changes

* Use CanisterPageHeader and CanisterPageHeading in CanisterDetails.svelte
* Manage the spacing between components in the `section` in CanisterDetails instead of each child setting the spacings.
* Set Unlink button to secondary. Buttons in the header are secondary.
* Change the text when the canister is not the controller to better explain the "Balance unavailable" title.
* Remove CyclesCard.svelte

# Tests

* Adapt tests in CanisterDetails page to new UI and logic. I didn't move ALL tests to use POs. Only the ones related to the changes.
* Add some methods to related POs.

# Todos


lmuntaner:
New UI:

Controller and name:
Screenshot 2023-09-29 at 09 25 03

Controller and not name:
Screenshot 2023-09-29 at 09 26 19

Not controller and name:
Screenshot 2023-09-29 at 09 25 13

Not controller and not name:
Screenshot 2023-09-29 at 09 25 44

Loading state:
Screenshot 2023-09-29 at 10 09 09

The issue with the title is fixed in #3413

lmuntaner:

dskloetd:
In https://github.com/dfinity/nns-dapp/pull/3386 @peterpeterparker has some concerns about the "Balance unavailable" text.
But I see it again in the screenshots here.
What was the decision regarding that?

dskloetd:
I also see a very narrow header skeleton in the screenshot here.
Will that also be fixed be https://github.com/dfinity/nns-dapp/pull/3413 ?

lmuntaner:
> In #3386 @peterpeterparker has some concerns about the "Balance unavailable" text. But I see it again in the screenshots here. What was the decision regarding that?

To improve the text below. As done here: https://github.com/dfinity/nns-dapp/pull/3414/files#diff-cab8170fd7b4abad14b90b471915ffa043fe4f41180aa147f22423a43a66f976

"You are not the controller of this canister. Only controllers have access to its cycles and controllers."

lmuntaner:
> I also see a very narrow header skeleton in the screenshot here.
> Will that also be fixed be https://github.com/dfinity/nns-dapp/pull/3413

Yes. There is a comment right below the screenshot precisely saying that: "The issue with the title is fixed in #3413". Maybe it was not clear enough, sorry.

frontend/src/tests/lib/pages/CanisterDetail.spec.ts
```diff
@@ -50,14 +63,9 @@ describe("CanisterDetail", () => {
...
-      const { queryByTestId } = render(CanisterDetail, props);
-
-      await runResolvedPromises();
-
-      expect(
-        queryByTestId("canister-card-title-compoment").textContent.trim()
-      ).toEqual(mockCanister.canister_id.toText());
+    it("should not render subitle", async () => {
+      const po = await renderComponent();
+      expect(await po.hasSubtitle()).toBe(false);
```
dskloetd, line 68: Can you also expect it to be true in another test?
Or make the PO such that it's guaranteed that `hasSubtitle()` and `getSubtitle()` always refer to the same thing?
Right now, if `hasSubtitle()` is changed to always return `false`, or if there is a typo in the TID, or if the TID changes and we update `getSubtitle` but forget to update `hasSubtitle`, all tests will continue to pass.

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
```diff
@@ -26,9 +27,19 @@ export class CanisterDetailPo extends BasePageObject {
...
-      await this.root.byTestId("canister-card-title-compoment").getText()
-    ).trim();
+    return this.getCanisterPageHeading().getTitle();
+  }
+
+  async getCanisterSubtitle(): Promise<string> {
+    return this.getCanisterPageHeading().getSubtitle();
+  }
+
+  async hasSubtitle(): Promise<boolean> {
```
dskloetd, line 42: Why does `getCanisterSubtitle` have "canister" in the name, but `hasSubtitle` doesn't?
Given that `CanisterDetailPo` is already about the canister, it's probably not necessary, but mostly I think it should be consistent.

frontend/src/tests/lib/pages/CanisterDetail.spec.ts
```diff
@@ -115,6 +118,26 @@ describe("CanisterDetail", () => {
...
+          name: "canister name",
+        },
+      ]);
+    });
+
+    it("shuold render cycles balance as title", async () => {
+      jest.spyOn(canisterApi, "queryCanisterDetails").mockResolvedValue({
+        ...mockCanisterDetails,
+        cycles: 100_000_000n,
+      });
```
dskloetd, line 135: Why is the `queryCanisters` setup in `beforeEach` but `queryCanisterDetails` in the test?
I'm not saying it shouldn't. Maybe one is more generic and the other is more specific to the test.
But it's a bit hard to tell with just 1 test in this block.

frontend/src/lib/pages/CanisterDetail.svelte
dskloetd, line 1: It still seems odd that "Balance unavailable" is the most prominent text on the page.
It only makes sense if you know that it would show the balance there in other cases.
But when you are looking at a single page by itself, surely some information that's not available can't be the most important information?
Even just saying "Canister" seems more useful somehow.

frontend/src/tests/lib/pages/CanisterDetail.spec.ts
lmuntaner: I wanted the part that was related to the test to be defined inside the test for visibility.

The cycles is set here, and the expect is related directly to the cycles.

frontend/src/lib/pages/CanisterDetail.svelte
lmuntaner: We didn't find a better text.

https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=283-1652&mode=design&t=BsB98WdJtVL1gHTk-0

We agreed to it with @artkorotkikh-dfinity 

I'm open to other ideas.

frontend/src/tests/lib/pages/CanisterDetail.spec.ts
lmuntaner: I renamed the method. I also refactored the getSubtitle and hasSubtitle of the `CanisterPageHeadingPo` to refer to each other. I'm not sure what else you mean.

There are already test that check `expect(await po.getSubtitle()).toBe(canisterName);`. Isn't that enough?

frontend/src/tests/page-objects/CanisterDetail.page-object.ts
lmuntaner: Renamed.

lmuntaner:
@dskloetd please review

frontend/src/tests/lib/pages/CanisterDetail.spec.ts
dskloetd: Makes sense. Thanks.

frontend/src/lib/pages/CanisterDetail.svelte
dskloetd: As I said, I think "Canister" is better there than "Balance unavailable".
If you see on the mocks in sequence, "Balance unavailable" makes sense because you see the balance in the other mocks.
But in real life you always see just 1 UI.
If the canister has a name, then it could be the name.

frontend/src/tests/lib/pages/CanisterDetail.spec.ts
dskloetd: It wasn't before (because it mentioned the TID separately), but now with the refactoring it is. Thanks.

dskloetd (approved):

frontend/src/lib/pages/CanisterDetail.svelte
lmuntaner: But the name is in the subtitle. It would be inconsistent to put it in one place in one case and in another place in other cases.

Let me bring discuss with @artkorotkikh-dfinity

dskloetd: Yes, but since you only see 1 UI, I think UX is more important than consistency.
Thanks for bringing it up. This is not blocking of course.

lmuntaner: let's discuss and solve in another PR. I agree.

PR #3416, opened 2023-09-29
Fix release-sop test

dskloetd:
# Motivation

There is a `release-sop` tests what happens when HTML is found.
But the mocking behavior for the test didn't cause that step to fail with non-zero exit code.

# Changes

1. Fix the behavior of `no_html_in_templates` to return 1 instead of 0.
2. Update the expected output with `scripts/nns-dapp/release-sop.test --update`

# Tests

Change is test only.

# Todos

not necessary


mstrasinskis (approved):
Rubber stamp. Thanks!

PR #3417, opened 2023-09-29
Chore: Update Launchpad Prod Test JSON files

lmuntaner:
# Motivation

We want to make sure that the NNS Dapp works with the data in the SNS Aggregator from mainnet.

# Changes

* Two new JSON files inside the `test/workflows/Launchpad` that are the same response from the SNS Aggregator today at 15.45.
* Use the two new JSON files as the mock data for the `ProdLaunchpad.spec.ts` test.

# Tests

Only test changes.

# Todos


lmuntaner:
@dskloetd please review

frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
```diff
@@ -27,11 +28,11 @@ describe("Launchpad", () => {
     mockFetch
       .mockResolvedValueOnce({
         ok: true,
-        json: () => Promise.resolve(tenAggregatedSnses),
+        json: () => Promise.resolve(snsPage0),
```
dskloetd, line 31: Would it be difficult to return one or the other depending on the arguments passed to fetch?

frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
```diff
@@ -27,11 +28,11 @@ describe("Launchpad", () => {
...
       })
-      .mockReturnValueOnce({
+      .mockResolvedValueOnce({
         ok: true,
-        json: () => Promise.resolve([]),
+        json: () => Promise.resolve(snsPage1),
```
dskloetd, line 35: Is there still a test that explicitly tests that it works when the last page is full and the next page is empty?

dskloetd (approved):
Sorry, I thought this was a generic mock but I see now it's just a mock for 1 specific test.

frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
lmuntaner: Not specifically.

We have a test that checks whether it still works if the second page fails. Which was the bug we had.

We don't have a test with an empty array. I can add it in another PR.

frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
lmuntaner: done

lmuntaner:
@dskloetd can you approve again?

dskloetd (approved):

PR #3422, opened 2023-10-02
Don't trim newlines on mock outputs for release-sop

dskloetd:
# Motivation

When `release-sop` is tested, we provide mock behavior for the commands it executes.
Newlines are removed from the output provided in the mock behavior, using `| tr -d '\n'`.
But this is not necessary because we also pass `-n` to `echo` when outputting the output.
And it turns out that `| tr -d '\n'` actually removes inner newlines as well which is interfering with other improvements I want to make to the script.

# Changes

Remove `| tr -d '\n'` applied to the mock output.

# Tests

Still pass

# Todos

not necessary

### DIFF
--- a/scripts/nns-dapp/release-sop
+++ b/scripts/nns-dapp/release-sop
@@ -56,8 +56,8 @@ execute() {
     echo "No mock behavior for: $cmd" >&2
     exit 1
   fi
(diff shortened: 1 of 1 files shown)


mstrasinskis (approved):
Rubber stamp. Thanks!

PR #3424, opened 2023-10-02
Feat: Change canister detail page title

lmuntaner:
# Motivation

The title "Balance unavailable" was not very user friendly when the user was not the controller.

Instead, the title will be now the canister name or the id when the user is not the controller. In that case, there is no need for the subtitle.

# Changes

* Change the title in CanisterHeadingTitle when the user is not the controller. To access the name, I added the prop `canister`.
* Pass the `canister` prop to CanisterHeadingTitle in CanisterPageHeading.
* Add the logic of not the controller to render the PageHeading without title in CanisterPageHeading.
* Remove the `balance_unavailable` i18n key.

# Tests

* Change and add test cases for new functionality in CanisterHeadingTItle.spec
* Change and add test cases for new functionality in CanisterPageHeading.spec
* Change and add test cases for new functionality in CanisterDetail.spec

# Todos

Not necessary. Covered by the entry that there is a new canister header.

### DIFF
--- a/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
+++ b/frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
@@ -2,11 +2,14 @@
   import { TokenAmount, nonNullish } from "@dfinity/utils";
   import AmountDisplay from "$lib/components/ic/AmountDisplay.svelte";
   import type { CanisterDetails } from "$lib/canisters/ic-management/ic-management.canister.types";
+  import type { CanisterDetails as CanisterInfo } from "$lib/canisters/nns-dapp/nns-dapp.types";
   import { i18n } from "$lib/stores/i18n";
   import { SkeletonText } from "@dfinity/gix-components";
   import TestIdWrapper from "../common/TestIdWrapper.svelte";
+  import { shortenWithMiddleEllipsis } from "$lib/utils/format.utils";
 
   export let details: CanisterDetails | undefined;
+  export let canister: CanisterInfo;
   export let isController: boolean | undefined;
 </script>
 
@@ -23,7 +26,11 @@
     <!-- Only when we have loaded the data and we know whether the user is the controller -->
   {:else if isController === false}
     <h1 data-tid="caniter-title-balance-unavailable">
-      {$i18n.canister_detail.balance_unavailable}
+      {#if canister.name.length === 0}
+        {shortenWithMiddleEllipsis(canister.canister_id.toText())}
+      {:else}
+        {canister.name}
+      {/if}
     </h1>
   {:else}
     <div data-tid="skeleton" class="skeleton">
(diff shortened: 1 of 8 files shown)


lmuntaner:
New UI:

Not controller, nor name:
Screenshot 2023-10-02 at 11 26 29

Not controller with name:
Screenshot 2023-10-02 at 11 26 37

lmuntaner:

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
```diff
@@ -23,7 +26,11 @@
     <!-- Only when we have loaded the data and we know whether the user is the controller -->
   {:else if isController === false}
     <h1 data-tid="caniter-title-balance-unavailable">
-      {$i18n.canister_detail.balance_unavailable}
+      {#if canister.name.length === 0}
```
peterpeterparker, line 29: Not for this PR, just an idea: `.name.length === 0` finds place multiple time in the code base, maybe worth a utils that either checks if canister has a name or provide name or id

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
```diff
@@ -23,7 +26,11 @@
...
+        {shortenWithMiddleEllipsis(canister.canister_id.toText())}
```
peterpeterparker, line 30: there isn't enough space to print out the all canister id?

peterpeterparker (approved):
LGTM, thanks

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
lmuntaner: Yeah, good point. It's almost a business logic, so it makes sense to have it in a separate function. I'll make myself a note. Thanks!

frontend/src/lib/components/canister-detail/CanisterHeadingTitle.svelte
lmuntaner: There is, but depends on the canister id it would be very big. I wasn't sure about this either to be honest.

Let's show all of it, and if people complain we'll make it smaller.

peterpeterparker: Principal can be up to 32 bytes but, generally canister ids are `aaaaa-bbbbb-ccccc-ddddd-cai`. So if there is enough space, I would display the canister id without the utils but, use CSS to add elipsis in case of overlfow.

Just my two cents though.

lmuntaner: I believe two lines is better than ellipsis actually.

peterpeterparker (approved):
> We changed a setting that requires new approvals after new commits.

Why? Sounds a bit ultra parioniac and I've got other things to do that accepting a PR twenty times.

That said, looks good to me. Thanks 👍

lmuntaner:
> > We changed a setting that requires new approvals after new commits.
> 
> Why? Sounds a bit ultra parioniac and I've got other things to do that accepting a PR twenty times.
> 
> That said, looks good to me. Thanks 👍

Oh yes, that was not out intention.

We just want that the last pusher is not the approver. But Github doesn't explain these settings very well.

Maximilian is sick today and he's the only one with access to those settings.

PR #3425, opened 2023-10-02
Chore: Disable NNS proposal args

lmuntaner:
# Motivation

`scripts/dfx-nns-proposal-args.test` is failing.

Disable until we fix it.

# Changes

* Comment the lines in the workflow that run the script.

# Tests

Tests should be now green.

# Todos

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -220,9 +220,9 @@ jobs:
         run: sudo apt-get update -yy && sudo apt-get install -yy moreutils && command -v sponge
       - name: Install dfx
         run: DFX_VERSION="$(jq -r .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
-      # TODO: Fix so that it doesn't rely on mainnet?
-      # - name: Test getting proposal args
-      #   run: scripts/dfx-nns-proposal-args.test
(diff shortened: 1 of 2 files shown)


dskloetd (approved):
Thanks!

.github/workflows/checks.yml
```diff
@@ -220,8 +220,9 @@ jobs:
         run: sudo apt-get update -yy && sudo apt-get install -yy moreutils && command -v sponge
       - name: Install dfx
         run: DFX_VERSION="$(jq -r .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
-      - name: Test getting proposal args
-        run: scripts/dfx-nns-proposal-args.test
+        # TODO: Fix so that it doesn't rely on mainnet?
```
dskloetd, line 223: Actually, it's probably better to just add
```
if: false
```
on line 203.
The additional indentation here make it look like this step is part of the previous step which might be confusing when reenabling again.

Also there's no point to do the whole setup for a test we don't do.

dskloetd (approved):

.github/workflows/checks.yml
lmuntaner: done, thanks!

dskloetd (approved):

dskloetd (approved):

PR #3431, opened 2023-10-03
Use clap in did2rs

bitdivine:
# Motivation
`did2rs` currently hand-parses command line arguments.  With `clap` this parsing can be standardized.

# Changes
- Add the `clap` CLI parser to `did2rs`
- Add a flag to specify the canister name, but support the former positional argument as well.
- Add a flag to override the default location of the `.did` file.

# Tests
See CI - usage should work exactly as before.

# Todos


peterpeterparker (approved):
LGTM

PR #3436, opened 2023-10-03
Specialize did2rs for the aggregator

bitdivine:
# Motivation
We wish to use `did2rs` to derive rust code for use in proposals rendering.  Currently `did2rs` has parameters hard-wired for use in the SNS aggregator.

# Changes
- Replace hard-wired constants in `did2rs` with command line arguments.
- Put the constants applicable to the SNS aggregator in `scripts/sns/aggregator/did2rs`; this can be used as a drop-in replacement for the former `did2rs.sh`.
- Update other commands to use the aggregator-specific `did2rs`.
  - Note: Some of these other files should also be moved into `scripts/sns/aggregator/` but that is not done in this PR in order to keep the PR size small.

# Tests
Existing functionality is unchanged, so existing tests should suffice.

# Todos


scripts/sns/aggregator/did2rs
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: The constants specifc to the SNS aggregator have been moved to here, from the main `did2rs`.

lmuntaner (approved):
LGTM!

PR #3442, opened 2023-10-03
Robust traits

bitdivine:
# Motivation
We will need to import different sets of traits to different files.  To do that, we need to parameterize the set of traits.  Additionally, the current trait insertion is fragile in that changes to didc can break it easily; that needs to become robust.

# Changes
- Move the inserted traits to the beginning of each `#[derive(...` line, rather than depending on didc inserting a specific trait.
- Extract the inserted traits as a variable.

# Tests
- Existing tests should suffice, as the rust code is identical apart from the ordering of the traits.

# Todos

  - Covered by existing changelog entries.


rs/sns_aggregator/src/types/ic_sns_governance.patch
```diff
@@ -1,22 +1,22 @@
 diff --git b/rs/sns_aggregator/src/types/ic_sns_governance.rs a/rs/sns_aggregator/src/types/ic_sns_governance.rs
-index 484f5f709..ee951741f 100644
+index 7edf9991d..d779ceeaa 100644
 --- b/rs/sns_aggregator/src/types/ic_sns_governance.rs
 +++ a/rs/sns_aggregator/src/types/ic_sns_governance.rs
 @@ -517,7 +517,7 @@ pub struct GetMaturityModulationResponse {
- #[derive(CandidType, Deserialize, Serialize, Clone, Debug)]
+ #[derive(Serialize, Clone, Debug, CandidType, Deserialize)]
```
bitdivine, line 6: The order of the traits has changed, with `Serialize, Clone, Debug` now at the beginning rather than after `Deserialize`.  This is repeated many times.  The change that causes this is at the very end of this PR.

rs/sns_aggregator/src/types/ic_sns_ledger.rs
```diff
@@ -9,10 +9,10 @@ use ic_cdk::api::call::CallResult;
 // This is an experimental feature to generate Rust binding from Candid.
 // You may want to manually adjust some of the types.
 // #![allow(dead_code, unused_imports)]
-// use candid::{self, CandidType, Decode, Deserialize, Serialize, Clone, Debug, Encode, candid::Principal};
```
bitdivine, line 12: Detail: The old code also injected traits in this commented-out import line.  Not that it mattered, but the new trait is more accurate.

scripts/did2rs.sh
```diff
@@ -113,7 +116,7 @@ cd "$GIT_ROOT"
             s/^    [a-z].*:/    pub&/;s/^( *pub ) *pub /\1/;
 
 	    # Add traits
-            s/([{( ]Deserialize)([,})])/\1, Serialize, Clone, Debug\2/;
+            s/#\[derive\(/&'"${TRAITS:-}${TRAITS:+, }"'/;
```
bitdivine, line 119: Here we insert the traits at the beginning of `#[derive(...`, as that is more robust than adding them after `Deserialize` that might not even be a trait annotation.

scripts/did2rs.sh
```diff
@@ -48,6 +48,9 @@ RUST_PATH="${GIT_ROOT}/rs/sns_aggregator/src/types/ic_${CANISTER_NAME}.rs"
 PATCH_PATH="${GIT_ROOT}/rs/sns_aggregator/src/types/ic_${CANISTER_NAME}.patch"
 DID_PATH="${DID_PATH:-${GIT_ROOT}/declarations/${CANISTER_NAME}/${CANISTER_NAME}.did}"
 
+# TODO: Provide this as an argument.
+TRAITS="Serialize, Clone, Debug"
```
bitdivine, line 52: Here the list of traits has been extracted as a variable.  We can now set that as a command line variable.

lmuntaner (approved):
Thanks!

PR #3443, opened 2023-10-03
Fix aggregator canister dependency list

bitdivine:
# Motivation
The command that creates sns aggregator patch files currently produces patch files even for canisters that are not used by the SNS aggregator.

# Changes
- Create patch files only for canisters used by the SNS aggregator.
- Add to the help text.

# Tests
Previously, after running `mk_nns_patch`, there would be untracked files:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	rs/sns_aggregator/src/types/ic_nns_governance.rs
```
No more.

# Todos

  - Too minor to need an entry, I think.


lmuntaner (approved):
LGTM!

PR #3444, opened 2023-10-04
Init vitest

peterpeterparker:
# Motivation

We have to migrate from jest to vitest; otherwise, we are stuck with Svelte v3 forever. I did not find any ways to set up jest for ESM in our complex dependencies context. ESM which is required by Svelte v4 and the testing library.

As it is complicated to migrate everything in a single go (see PR #2475), this PR instantiates and configures [vitest](https://vitest.dev/). That way, we can develop new tests with it, and we can iteratively migrate to vitest. Either by migrating particular test when we develop new features in existing functions or by providing PRs with batchs of tests migrated.

# Changes

- configure vitest
- setup vitest and jest to be run on `npm run test`
- migrate 2-3 tests to vitest to validate the configuration


dskloetd:
> That way, we can develop new tests with it, and we can

Did you mean to write more there in the PR description?

TESTING.md
```diff
@@ -43,6 +54,23 @@ All the mocks used in different tests are in `frontend/src/tests/mocks`.
 
 Found in `frontend/src/tests/utils`.
 
+### Migration to vitest
+
+To migrate a test from jest to vitest, aside from moving it between folders, the following changes can be made:
+
+- Replace the prefix `jest.` with `vi.`
+- Remove the `jsdom` annotation, as all vitest runs in a simulated browser environment
```
dskloetd, line 62: Doesn't this make it much slower for unit tests that don't require DOM?
Will we only migrate tests that don't depend on Svelte or will we migrate all tests?

peterpeterparker:
> Did you mean to write more there in the PR description?

Thanks, yes somehow my last sentence was trimmed.

frontend/src/vitests/lib/api/accounts.api.spec.ts
dskloetd, line 1: Can we migrate to Svelte 5 after only migrating the component tests and not the API tests?
Should we focus on first migrating the component tests so we can migrate to Svelte 5 and then decide if we should migrate the rest as well?

dskloetd (approved):

frontend/src/vitests/lib/api/accounts.api.spec.ts
peterpeterparker: Spontaneously I would say it does not gonna work because I would bet some components uses API and therefore it will fails loading does because of the ESM requirements. Does not mean that's a good approach and we can try that way.

peterpeterparker: That said, new tests regardless of the topic should not be written with jest, that would mean double work.

TESTING.md
peterpeterparker: > Doesn't this make it much slower for unit tests that don't require DOM?

A bit I would guess but as far as I tested, I had to set it that way. If you want to have a look, not against.

> Will we only migrate tests that don't depend on Svelte or will we migrate all tests?

All tests because we need ESM.

PR #3450, opened 2023-10-04
Fix: Do not hard-wire traits

bitdivine:
# Motivation
Traits are now provided to `did2rs` using a command line flag.  Unfortunately the traits are still hard wired.

Annoying, as I removed the hard-wired code but probably in the wrong branch or failed to commit the change. :-(

# Changes
No longer hard-wire the traits.

# Tests
The traits passed in by `./scripts/sns/aggregator/did2rs` are the same as the formerly hard-wired ones.  This should mean that there are no changes to the aggregator source code and patch files.  Furthermore consistency of this is checked in CI.

Locally, I am able to generate Rust code with different traits.

# Todos

  - Not needed, I think.


lmuntaner (approved):
Thanks!

PR #3451, opened 2023-10-04
style: add full width layout behind the flag

mstrasinskis:
# Motivation

Full-width proposal details layout (behind the flag). The first step for the page redesign.

# Changes

- add alternative `SplitBlock` based view.

# Tests

- Current tests should work.

# Todos

Will be updated after more changes,

# Screenshots

image

image

image


frontend/src/lib/components/proposal-detail/ProposalVotingSection.svelte
```diff
@@ -21,10 +21,8 @@
   $: total = yes + no;
 </script>
 
-<div class="content-cell-island">
```
mstrasinskis, line 24: Unwrap from `content-cell-island`

lmuntaner (approved):
Thanks for going with small PRs!

Do you have a plan or can you explain me in a call where you're going and the plan?

PR #3457, opened 2023-10-05
Migrate non features components related tests to vitest

peterpeterparker:
# Changes

- migrate non features components related tests to vitest
- document `done()` deprecation in TESTING.md
- remove `.only` in `it.only("should render a total balance for ckBTC",` because CI / vitest throwed an error asking if it was expected that only one test was active


TESTING.md
```diff
@@ -136,6 +136,21 @@ vi.mock("$lib/workers/transactions.worker?worker", () => ({
 
 - The pattern `await waitFor(expect().toBeNull)` seems to require an explicit arrow function call for vitest, like this: `await waitFor(() => expect().toBeNull())`.
 
+- [done() callback](https://vitest.dev/guide/migration.html#done-callback) is deprecated and should be replaced by promise. is deprecated and should be replaced with a promise. It's worth noting that this isn't an issue per sé, but it does trigger a console.log, which is not allowed by our test suite.
```
dskloetd, line 139: "is deprecated and should be replaced with a promise." appears twice.

frontend/src/vitests/lib/components/header/AccountSyncIndicator.spec.ts
```diff
@@ -189,13 +185,17 @@ describe("AccountSyncIndicator", () => {
         state: "idle",
       });
 
-      await waitFor(() => expect(queryByRole("menu")).toBeNull());
+      rerender({});
```
dskloetd, line 188: Can you point me to the documentation of rerender?
I searched on Google but I couldn't find it.
I'd like to understand what it does and why it's necessary.
And I was thinking it might be async, in which case we should put `await` in front of it here.

peterpeterparker: I had a second thought about it while coming to the office and using rerender is actually a bad idea here.
So I just reverted to original code and the test should fail in the CI. Locally running those is flaky.
Do you have any idea what can be the root cause of the issue?

e.g.

```
await waitFor(() => expect(queryByRole("menu")).toBeNull());
```

This fails, claiming that menu is actually not null. If locally for example I add a timeout, it solve problem. I remove timeout, test still works. Really weird.

TESTING.md
peterpeterparker: thx

frontend/src/vitests/lib/components/header/AccountSyncIndicator.spec.ts
dskloetd: Can you try if it helps to do `await runResolvedPromises();` instead of `rerender({});`?

Like this:
https://github.com/dfinity/nns-dapp/blob/71b4fb86613b6683bb6635629e736b8986d55a07/frontend/src/vitests/routes/layout.spec.ts#L49

peterpeterparker: Did not work neither unfortunately. Something with the binding to the gix components and hiding the popover that makes the test flaky, not sure why.

Instead I modified the two tests as following for now - i.e. testing that the property in NNS dapp to hide the popovers becomes false.

```
// TODO: flaky test
// expect(queryByRole("menu")).toBeNull()
expect(component.$$.ctx[component.$$.props["visible"]]).toBeFalsy();
```

Do you have a better idea?

dskloetd: I think it's important to really understand what's going on before moving on.
I don't really know the component so I don't have specific ideas but personally I would add a lot of logging in the test and in the component to really understand where it is getting stuck.

peterpeterparker: Give that's two tests, should we remove those of the PR so that we can move on with the PR and we have a look separatly or should we merge it with the TODO?

peterpeterparker: I'm a bit out of ideas regarding those, can you have a look and help about this?

dskloetd: Let's revert those 2 tests and merge the PR. I'll try to have a look.

peterpeterparker: Done. Thanks a lot.

dskloetd (approved):

PR #3465, opened 2023-10-05
Chore: Remove update:next script

lmuntaner:
# Motivation

`npm run update:next` doesn't install the next dependencies for ledger-icp and ledger-icrc.

Instead, we can upgrade using `npm run upgrade:next`. There is no reason to have two scripts that do the same thing anyway, so I'm removing one.

# Changes

* Remove update:next from packgage.json scripts

# Tests

Not necessary.

# Todos


lmuntaner:

peterpeterparker (approved):
Thx

PR #3466, opened 2023-10-05
Proposals didc governance gen

bitdivine:
# Motivation
We would like to free ourselves of the dependency tree of other canisters, such as the `nns_governance` canister.

# Changes
- Add code to derive Rust types exposed by the governance canister.

# Tests
- None yet.  This code does not even include the generated code as a module.

# Todos

  - I suggest adding a changelog entry when we switch to using this generated code.


rs/proposals/src/canisters/nns_governance/api.rs
bitdivine, line 1: This is the main file of this PR - the generated Rust code.

scripts/proposals/did2rs
bitdivine, line 1: This is the command that generates the Rust code, and that would be used to regenerate the Rust code if it is updated.

scripts/proposals/did2rs-list
bitdivine, line 1: Currently we generate only `nns_governance` but more canisters are coming.  This gets a list of the canisters we can generate.  I optimistically assume that we will import everything we can.  If that turns out to be over-optimistic, we can print a hard-wired list of canisters to generate.

mstrasinskis (approved):
Rubber stamp.

PR #3470, opened 2023-10-05
Increase local Playwright timeout

dskloetd:
# Motivation

Some tests have become quite long.
We also have a faster machine on CI then we used to have in the beginning.
So having a different timeout on CI vs. locally isn't really warranted anymore.

# Changes

Set the Playwright timeout to 5 minutes regardless of whether the test runs locally or on CI.

# Tests

CI passes

# Todos

not necessary


lmuntaner (approved):
LGTM

PR #3471, opened 2023-10-05
feat (proposal rendering): Get sns_wasm types from .did file

bitdivine:
# Motivation
We would like to free the nns-dapp from the code and libraries pulled in by the nns-sns-wasm canister.

# Changes
* Remove the dependency on the `sns_wasm` canister from the proposals rendering crate.
* Derive the `nns-sns-wasm` canister types from `.did`.

Note: The `.didc` is from 2023-09-27 whereas the removed library is from `2023-08-01`.

# Tests
## Does proposal rendering still work?
There are existing Rust tests for proposal rendering; these should suffice.
## Is this actually reducing dependencies?
We can get a rough measure of the number of dependencies in the proposal rendering crate with:
```
cargo tree -p proposals | wc -l
```
For the current release candidate, i.e. a recent commit on main, this yields 1555 entries.  In this branch this has dropped to 1172 and, critically, no longer contains the timer crate.

# Todos

### DIFF
--- a/rs/proposals/src/canisters/sns_wasm/api.rs
+++ b/rs/proposals/src/canisters/sns_wasm/api.rs
@@ -1,0 +1,408 @@
+#![allow(clippy::all)]
+#![allow(clippy::missing_docs_in_private_items)]
+#![allow(non_camel_case_types)]
+#![allow(dead_code, unused_imports)]
+use candid::{self, CandidType, Decode, Deserialize, Encode, Principal};
+use ic_cdk::api::call::CallResult;
+use serde::Serialize;
+
+#[derive(Serialize, CandidType, Deserialize)]
+pub struct EmptyRecord {}
+// This is an experimental feature to generate Rust binding from Candid.
+// You may want to manually adjust some of the types.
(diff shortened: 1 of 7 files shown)


rs/proposals/src/canisters/sns_wasm/api.rs
```diff
@@ -0,0 +1,408 @@
+#![allow(clippy::all)]
```
bitdivine, line 1: This file is generated with `did2rs`

rs/proposals/src/lib.rs
```diff
@@ -215,13 +215,13 @@ fn debug<T: Debug>(value: T) -> String {
 
 mod def {
     use crate::canister_arg_types;
+    use crate::canisters::sns_wasm::api::{SnsUpgrade, SnsVersion};
     use crate::{decode_arg, Json};
-    use candid::CandidType;
+    use candid::{CandidType, Principal};
```
bitdivine, line 220: The generated code uses `Principal` rather than `PrincipalId`.  The two represent the same type, though, and `Principal` has the advantage of not depending on the `ic` repo.

PR #3474, opened 2023-10-06
Migrate canisters and launchpad components to vitest

peterpeterparker:
# Changes

- migrate to vitests
- in `CanisterCardCycles.spec.ts`, `await tick` replaced with `waitFor` to wait for the initialization to be ready

# Notes

`RemoveCanisterControllerButton.spec.ts` not moved yet as it appears to be flaky. Got the feeling it's the same reason as what we discussed in #3457 - i.e. it happens in two tests were we tests if the Gix-cmp modal is closed.


frontend/src/vitests/lib/components/canister-detail/RenameCanisterButton.spec.ts
dskloetd, line 1: Any idea why GitHub shows this as a new file instead of a moved file?

frontend/src/vitests/lib/components/canisters/CanisterCardCycles.spec.ts
```diff
@@ -48,7 +45,7 @@ describe("CanisterCardCycles", () => {
   it("should render canister cycles information", async () => {
     const { getByTestId } = render(CanisterCardCycles, props);
 
-    await tick();
+    await waitFor(() => expect(cyclesCallback).not.toBeUndefined());
```
dskloetd, line 48: Use `.toBeDefined()` instead of `.not.toBeUndefined()`?

below as well

dskloetd (approved):

frontend/src/vitests/lib/components/canisters/CanisterCardCycles.spec.ts
peterpeterparker: Sure

frontend/src/vitests/lib/components/canister-detail/RenameCanisterButton.spec.ts
peterpeterparker: No idea, it's weird because I definitely moved it 🤷‍♂️

PR #3476, opened 2023-10-06
GIX-1860: Refactor PageHeading subtitle slot

lmuntaner:
# Motivation

Avoid the problem with the PageHeading spacing when the subtitle was not defined.

At the same time, we learned that the PageHeading might get two subtitles; therefore, a slot without being wrapped makes it more flexible.

# Changes

* Remove the `h4` wrapping the "subtitle" slot in PageHeading.
* New component "HeadingSubtitle" which has the style of the `h4` inside the PageHeading.
* Use the new component as subtitle slot for PageHeading in NnsNeuronPageHeading, SnsNeuronPageHeading and CanisterPageHeading.

# Tests

Only refactor of purely UI components. Functionality didn't change which is confirmed by tests passing without extra changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/components/common/HeadingSubtitle.svelte
+++ b/frontend/src/lib/components/common/HeadingSubtitle.svelte
@@ -1,0 +1,12 @@
+<script lang="ts">
+  export let testId: string | undefined = undefined;
+</script>
+
+<h4 data-tid={testId} class="description"><slot /></h4>
+
+<style lang="scss">
+  h4 {
+    margin: 0;
+    font-weight: normal;
+  }
+</style>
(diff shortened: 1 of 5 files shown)


frontend/src/lib/components/common/HeadingSubtitle.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+  export let testId: string | undefined = undefined;
+</script>
+
+<h4 data-tid={testId} class="description"><slot /></h4>
+
+<style lang="scss">
+  h4 {
+    margin: 0;
+    font-weight: normal;
```
peterpeterparker, line 10: not related to the pr: worth checking if this can actually be a global thing? maybe all >= h4 should be normal?

peterpeterparker (approved):
Thanks!

frontend/src/lib/components/common/HeadingSubtitle.svelte
lmuntaner: Good point. I'm not sure at the moment. I hope we'll look into global variables together with David Aerne and Artem. We can consider that then.

PR #3477, opened 2023-10-06
Proposals didc registry

bitdivine:
# Motivation
We need to parse recent registry canister proposal types.  In order to do that, we need to update the registry types.

# Changes
- Derive the registry canister types from the registry canister .did file.
- Remove the direct import of the registry canister.

## Alternatives considered
- Given that the timer conflicts are now gone in main, I tried updating the canister commit.  This passes cargo check with considerable effort but then fails the Wasm build with different conflicts, involving tokio which is imported by the registry.  Deriving types from the .did files again avoids these problems by shedding dependencies.

# Tests
There are existing proposal test vectors.  Presumably they cover all proposal types so should suffice.

Note: With this PR, the dependency count given by:
```
cargo tree -p proposals | wc -l
```
falls further from the 1555 in the last release candidate to 789.

# Todos


scripts/update_ic_commit
```diff
@@ -42,6 +42,7 @@ source "$(clap.build)"
     curl -sfL --retry 5 "$IC_URL$upstream_path" >"${local_path}"
   }
   did_curl nns_governance /rs/nns/governance/canister/governance.did
```
lmuntaner, line 44: Question: Didn't we say that we would try to have a script to update the IC dependencies separately?

This also updates the SNS Aggregator did files.

Shouldn't we have a script for the proposals and one for SNS aggregator? Are you working on it?

bitdivine: At the moment, this is the one place that updates the did files in `/dependencies`.  But, yes, this needs to be split per our agreement.  Splitting is not the focus of this PR.  I'm hoping to get new proposal types rendered before I go on holiday, so I'm limiting myself to just what is needed for that. :-D

mstrasinskis (approved):
Rubber stamp. Thank you!

scripts/update_ic_commit
lmuntaner: Ok, understood, thanks

PR #3480, opened 2023-10-06
test: migrate sns and proposals components to vitest

peterpeterparker:
# Changes

- move sns and proposals components to vitest
- make `mockAuthStoreSubscribe` and `mockTokenStore` accept optional runner
- move `destroyed` check in tooltip at the top of the debounced function to avoid "document is not defined" error in test


lmuntaner (approved):
Thanks!

PR #3482, opened 2023-10-06
CLI tool for proposal rendering

bitdivine:
# Motivation
The proposals crate has quite a few test vectors, however I still wanted some confidence that real proposals would render correctly.

Hence a command line tool, that allows running the proposals renderer against any proposals payload that has been downloaded from mainnet.

I downloaded hundreds of proposals and rendered them with the current code and with older versions of the proposals renderer, to check for differences.

Adding hundreds of proposal payloads to the repository does not make sense, as many are large.  It is quite a useful tool none the less.  Perhaps we can devise a test suite that uses it.

# Changes
- Add a `main.rs` to the proposals crate, to create a proposals rendering command line tool.
- Add a script that downloads a proposal and renders the payload.

# Tests
None yet.  Maybe this could form the basis of a test suite in future, though.

# Todos


scripts/proposals/mk-test-vector
```diff
@@ -0,0 +1,104 @@
...
+print_help() {
+  cat <<-"EOF"
+
+	Creates a test vector for proposal argument rendering:
+
+	- Gets a proposal from mainnet or another network.
+	- Renders the proposal to JSON.
+	- Saves both in a directory of test vectors.
+
+	Note: The rendered JSON MUST be reviewed before being merged into main.
```
bitdivine, line 15: Note: I wrote that before I realized how large the data would be.  Maybe let's fine somewhere else to store it.

rs/proposals/src/main.rs
lmuntaner, line 1: This file requires that we download a proposal with `mk-test-vector` first, right?

scripts/proposals/mk-test-vector
```diff
@@ -0,0 +1,107 @@
...
+prettyprint_proposal() {
+  (
+    cd "$SOURCE_DIR/.."
+    EXE="target/debug/proposals"
+    test -x "$EXE" || {
+      echo "ERROR: Please build the proposals CLI."
+      echo "       cargo build -p proposals"
+      exit 1
+    } >&2
+    "$EXE" <"$(proposal_path).raw" >"$(proposal_path).payload.json"
```
lmuntaner, line 101: I'm sorry for the stupid question, but how do we know this is a "valid" JSON?

bitdivine: Not a silly question.  The `.payload.json` is what the proposals library returns, unmodified in any way.  Given that the proposals API is _meant_ to return JSON it would be bad if it returned something else!  Let's check that it does.

The API that is called is here, which has return type Json: https://github.com/dfinity/nns-dapp/blob/main/rs/proposals/src/lib.rs#L134

However given that [Json is just an alias for String](https://github.com/dfinity/nns-dapp/blob/main/rs/proposals/src/lib.rs#L18) that means nothing.  At least I think we should call it JsonString.

The actual encoding is here: https://github.com/dfinity/nns-dapp/blob/main/rs/proposals/src/lib.rs#L139

So we either have the actual payload as JSON or the JSON string `"Proposal has no payload"`.

rs/proposals/src/main.rs
bitdivine: You can get the proposal in many ways. The script downloads the proposal and then runs this.  A minimal alternative would be:
```
dfx canister call nns-governance --network ic get_proposal_info "($PROPOSAL : nat64)" --query --output raw
```
You could pipe that straight into the `proposals` executable, as mentioned in the rustdoc.  By the way, I wish I had named the proposals crate `proposal-summary` or some such.  Maybe we can rename it another time.

I'll add the full command line to the rustdoc.

lmuntaner: proposal-payload better?

scripts/proposals/mk-test-vector
lmuntaner: `mk-test-vector` uses `main.rs`? where?

bitdivine: On line 95:
```
EXE="target/debug/proposals"
```
and then EXE is checked and executed.

lmuntaner (approved):

scripts/proposals/mk-test-vector
bitdivine: Sure thing.  Thank you for the review.

PR #3487, opened 2023-10-06
Migrate remaining components to vitests

peterpeterparker:
# Motivation

Still the modals, routes etc. todo but those are the last `lib/components` tests.

# Changes

- migrate remaining components to vitest
- comment `SnsVotingCard.reloadProposal` which ends up in timeout. Will ask @mstrasinskis to have a look afterwards to not block this PR.


lmuntaner (approved):
Thanks!

PR #3493, opened 2023-10-09
GetTokens: Don't check SNS balance for ckBTC

dskloetd:
# Motivation

We have a Get Tokens button which gives either ICP or SNS tokens.
It only gives SNS tokens if an SNS universe is selected and the faucet address has positive balance for that SNS.
Currently we also try to fetch the SNS faucet balance if the universe is not an SNS universe, for example with ckBTC.
This results in an error on the console.

# Changes

Only try to fetch SNS faucet balance if the selected universe is an SNS.

# Tests

Manually tried getting ICP while ckBTC or ICP or SNS without faucet balance universe is select.
And getting SNS tokens while an SNS with balance is selected.

# Todos

not necessary


lmuntaner (approved):
Thanks!

Sorry I didn't solve it. I got carried away by other things, I didn't know you were expecting it.

PR #3494, opened 2023-10-09
Fix: Separators for nothing in ProjectDetail

lmuntaner:
# Motivation

Separators were displayed even when there was nothing inside them.

# Changes

* Rename ProjecTimeline to ProjectTimelineUserCommitment.
* Move the rendering of the user commitment to ProjectTimeline.
* Add data as props for ProjectTimelineUserCommitment instea of relying in the context.
* Render Separators if needed in ProjectTimelineUserCommitment.

# Tests

* Add tests regarding the user commitment in ProjectTimelineUserCommitment.

# Todos


lmuntaner:

lmuntaner:
New UI:

No data rendered:
Screenshot 2023-10-09 at 14 41 40

With deadline:
Screenshot 2023-10-09 at 14 41 45

With commitment:
Screenshot 2023-10-09 at 14 42 49

With commitment and deadline
Screenshot 2023-10-09 at 14 42 57

frontend/src/lib/components/project-detail/ProjectStatusSection.svelte
```diff
@@ -58,28 +54,18 @@
 
 <!-- Because information might not be displayed once loaded - according the state - we do no display a spinner or skeleton -->
 
-{#if displayStatusSection}
+{#if displayStatusSection && nonNullish($projectDetailStore.summary)}
```
peterpeterparker, line 57: why not extending the subscribe `$: displayStatusSection = ...` with  `nonNullish($projectDetailStore.summary)}`?

frontend/src/lib/components/project-detail/ProjectTimelineUserCommitment.svelte
```diff
@@ -0,0 +1,83 @@
...
+  let hasParticipated: boolean;
+  $: hasParticipated =
+    nonNullish(myCommitment) && myCommitment.toE8s() > BigInt(0);
+
+  let dataIsRendered: boolean;
+  $: dataIsRendered = isOpen || isAdopted || hasParticipated;
+</script>
+
+<!-- The data is separated by Separators. But we don't want them if nothing is rendered. -->
+{#if dataIsRendered}
```
peterpeterparker, line 50: Ummm feel like it's a display thing that should be solve with CSS, like a `::before` and `::after` rather than using JavaScript to display a border bottom you know what I mean? Would that be possible here?

frontend/src/lib/components/project-detail/ProjectStatusSection.svelte
lmuntaner: Because it doesn't understand that `$projectDetailStore.summary` is set anyway...

frontend/src/lib/components/project-detail/ProjectTimelineUserCommitment.svelte
lmuntaner: Yes, but I find this easier to read.

When there was the bug, I could easily find the problem and solve it.

I prefer that it something shows in the website, it also shows in the html.

peterpeterparker: Mmmh not a fan of it I have to say, using js for stricly ui display purpose is generally slow but, shall be it

PR #3495, opened 2023-10-09
Get BTC button

dskloetd:
# Motivation

We want to be able to test ckBTC locally and on CI.
We can get ckBTC by requesting a BTC deposit address from the minter and sending BTC to it.
Using the mock bitcoin canister we can fake send BTC to the deposit address.

# Changes

1. When the ckBTC universe is selected show a "Get BTC" button instead of the "Get ICP" button.
2. When the button is clicked, we request a BTC address and tell the mock bitcoin canister to fake a transation to that address.

After this the use can click "Refresh Balance" in the ckBTC wallet to mint ckBTC into their account.

It's not much code so I didn't make it a dynamic import, but let me know if I should make it a dynamic import instead.

# Tests

Tested manually with a new snsdemo snapshot.

# Todos


lmuntaner (approved):
Thanks!

PR #3497, opened 2023-10-09
fix flaky "should trigger register-vote and call reloadProposal" test

mstrasinskis:
# Motivation

Fix and enable broken test (more info in comments)

# Tests

Works

# Todos

Not necessary


peterpeterparker (approved):
Thanks!!

PR #3499, opened 2023-10-09
test: revert last commit

peterpeterparker:
# Motivation

Tests are stuck so even if CI was green, try to run tests without last commit of main.

# Changes

# Tests

# Todos


peterpeterparker:
There probably was a GitHub incident on going that's why we had issue earlier trying to run the tests against PRs.

PR #3502, opened 2023-10-10
Revert jest for test with open handles

peterpeterparker:
# Motivation

Revert test which has open handles according:

- https://github.com/dfinity/nns-dapp/actions/runs/6461596460/job/17541597083
- https://github.com/dfinity/nns-dapp/actions/runs/6465195750/job/17550971767?pr=3496


lmuntaner (approved):
Thanks

PR #3504, opened 2023-10-10
Use child_process strategy to spread tests

peterpeterparker:
# Motivation

Locally I can reproduce the issue with open handles running `ProjectCommitment.spec.ts`.

When I switch the threading strategy to a file system based with `child_process` the issue is solved. Therefore this PR provide such configuration.

# Documentation

I don't know if vitest documentation is up-to-date regarding pool and threads. Either it is newer or older than the vitest version we are using, which is the last one. child_process is defined there https://vitest.dev/config/#forks


peterpeterparker:
Not a good idea. It works when a single test is executed but, not when all test are because it does not reset the mocks between those.

PR #3505, opened 2023-10-10
Chore: Use new Tag instead of nnsdapp HeadingTag

lmuntaner:
# Motivation

Be consistent with tags.

In this PR, remove the HeadingTag and instead use the Tag with size "large"

# Changes

* In NnsNeuroPageHeading and SnsNeuronPageHeading use a Tag with size "large" instead of HeadingTag.

# Tests

Tests still passing. Test ids were preserved.

# Todos

Not necessary.


lmuntaner:
UI has changed slightly, but not that it matters, we want the new tags.

Screenshot 2023-10-10 at 08 14 06
Screenshot 2023-10-10 at 08 14 12

lmuntaner:

frontend/src/lib/components/neuron-detail/NnsNeuronPageHeading.svelte
dskloetd, line 1: What is the purpose of `HeadigTag` if not to be used in a heading?

lmuntaner: I'm removing the `HeadingTag`, I'm not sure I understand the question.

dskloetd (approved):

PR #3513, opened 2023-10-10
Chore: Migrate ProjectDetail

lmuntaner:
# Motivation

Migreate tests to vitest.

In this PR, migrate ProjectDetail page to vitest.

# Changes

* Common changes to migrate the test.
* `blockedApiPaths` only for the module used in a fake. I don't understand why it was working before. It seems wrong.
* Close the ParticipateSwapModal immediately if test environment. This was the problem for all the participation tests. The modal wasn't closing, no matter how much time it was advanced. Like this, the modal closes and the tests pass.

# Tests

Only test changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -19,7 +19,7 @@
     "test": "npm run vitest && npm run deprecated-jest-test",
     "test-e2e": "tsc --noEmit -p src/tests/e2e/tsconfig.json && playwright test",
     "deprecated-jest-test": "TZ=UTC jest",
-    "vitest": "TZ=UTC vitest --config ./vitest.config.ts --sequence.hooks='list'",
+    "vitest": "TZ=UTC vitest --config ./vitest.config.ts",
     "update:agent": "npm rm @dfinity/agent @dfinity/auth-client @dfinity/candid @dfinity/identity @dfinity/principal @dfinity/nns @dfinity/nns-proto @dfinity/cmc @dfinity/sns @dfinity/utils @dfinity/ledger-icrc @dfinity/ledger-icp @dfinity/ckbtc @dfinity/ic-management && npm i @dfinity/agent @dfinity/auth-client @dfinity/candid @dfinity/identity @dfinity/principal @dfinity/utils@next @dfinity/ledger-icrc@next @dfinity/ledger-icp@next @dfinity/nns@next @dfinity/nns-proto@next @dfinity/cmc@next @dfinity/sns@next @dfinity/ckbtc@next @dfinity/ic-management@next && ../scripts/revert-next.sh",
     "update:gix": "npm update @dfinity/gix-components",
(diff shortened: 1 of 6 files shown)


dskloetd:
Let me debug a bit before you continue with this.
`blockedApiPaths` was not wrong and I'd like to keep it if possible.

lmuntaner:
> Let me debug a bit before you continue with this. `blockedApiPaths` was not wrong and I'd like to keep it if possible.

`blockedApiPaths` is necessary only if we use a `fake`, isn't it?

dskloetd:
`blockedApiPaths` predates the fakes :)
It's useful any time you mock every function that's called. It makes sure no other functions are being called. Without it, those functions would silently return `undefined` and if that doesn't cause the test to fail, you wouldn't know.

If found the issue.
With Jest, the `beforeEach` from `blockAllCallsTo` is called before the `beforeEach` from the test, which makes sense because its also set up first. But with vitest it's not run in the same order.
I'm not sure yet what to do about it. I'm still looking.

dskloetd:
I also found where to put the config:
https://github.com/dfinity/nns-dapp/commit/ab47d0c6f90ff14882dbfca85233277008c18544

dskloetd:
The issue with the modal timeout, I fixed here: https://github.com/dfinity/nns-dapp/commit/48dd0dde8500e1464b79f4722db67df349c9a4b8

peterpeterparker:
@dskloetd coolio. Are you going to provide two PRs for those or you are still debugging?

dskloetd:
I'm done debugging. I was expecting Llorenc to update this PR :)

lmuntaner:
Done!

I couldn't add the setting in the config file, I had to add it in the package.json.

The type `UserConfig` that we use doesn't have the `sequence` field. We can fix that in another PR.

lmuntaner:
@dskloetd @peterpeterparker ready for another review

frontend/package.json
```diff
@@ -19,7 +19,7 @@
     "test": "npm run vitest && npm run deprecated-jest-test",
     "test-e2e": "tsc --noEmit -p src/tests/e2e/tsconfig.json && playwright test",
     "deprecated-jest-test": "TZ=UTC jest",
-    "vitest": "TZ=UTC vitest --config ./vitest.config.ts",
+    "vitest": "TZ=UTC vitest --config ./vitest.config.ts --sequence.hooks='list'",
```
peterpeterparker, line 22: > I couldn't add the setting in the config file, I had to add it in the package.json.

Why?

Did you try the wrong config file? I cann add it to `vitest.config`

Capture d’écran 2023-10-10 à 16 45 10

peterpeterparker:
@dskloetd in the commit you linked https://github.com/dfinity/nns-dapp/commit/48dd0dde8500e1464b79f4722db67df349c9a4b8 the timer utils is also modified, should we do that as well?

```
// Currently
if (nonNullish(millis)) {
    vi.advanceTimersByTime(millis);
  } else {
    vi.runOnlyPendingTimers();
  }

// In your commit
if (nonNullish(millis)) {
    await vi.advanceTimersByTimeAsync(millis);
  } else {
    await vi.runOnlyPendingTimersAsync();
  }
```

lmuntaner:
> @dskloetd in the commit you linked [48dd0dd](https://github.com/dfinity/nns-dapp/commit/48dd0dde8500e1464b79f4722db67df349c9a4b8) the timer utils is also modified, should we do that as well?
> 

Yes, I did that.

frontend/package.json
lmuntaner: done!

peterpeterparker (approved):
LGTM, thanks a tone to both of you!

PR #3515, opened 2023-10-10
GIX-1936: Add "init" to SnsSummary

lmuntaner:
# Motivation

Support the NF enhancements. The new parameters will be added to the `get_init` endpoint.

In this PR, I introduce a field `init` in the SnsSummary with the data from `get_init`.

The aggregator already supports it, but the NNS Dapp was not using it until now.

Even the oldest SNS supports the endpoint: https://3r4gx-wqaaa-aaaaq-aaaia-cai.icp0.io/v1/sns/root/zxeu2-7aaaa-aaaaq-aaafa-cai/slow.json

# Changes

* Refactor `convertSwapInitParams` to return the type and convert to nullable if needed.
* Add field in SnsSummary `init`.
* Read the related field from the aggregator data, convert it to sns-js type and add it when creating SnsSummary.

# Tests

* Test that the SnsSummary contains the new field from `convertDtoToSnsSummary`.

# Todos

Not yet necessary.


frontend/src/lib/types/sns.ts
```diff
@@ -74,6 +74,11 @@ export interface SnsSummary {
    * Derived information about the sale such as the current total of ICP all buyers have invested so far
    */
   derived: SnsSwapDerivedState;
+
+  /**
+   * Data from `get_init` call.
+   */
+  init?: SnsSwapInit;
```
dskloetd, line 81: How does it relate to the field on line 40?
Is the comment on line 39 still accurate?

lmuntaner: Yes, it's still accurate. That init we don't use.

The fields we used were not in `init`, they were in the sale params. But with the 1-proposal, they moved everything to `init`. And copied them in the swap params for compatibility.

dskloetd (approved):

PR #3516, opened 2023-10-11
Fix hanging process after test

peterpeterparker:
# Motivation

`ProjectCommitment.spec.ts`  leads to an hanging process when tested with vitest:

> # FILEHANDLE
> node:internal/async_hooks:202
> 
> # FILEHANDLE
> node:internal/async_hooks:202
> close timed out after 30000ms
> Failed to terminate worker while running /Users/daviddalbusco/projects/dfinity/nns-dapp/frontend/src/vitests/lib/components/project-detail/ProjectCommitment.spec.ts.
> Tests closed successfully but something prevents Vite server from exiting
> You can try to identify the cause by enabling "hanging-process" reporter. See https://vitest.dev/config/#reporters

If found a related issue in vitests with many comments: https://github.com/vitest-dev/vitest/issues/2008
One of [those](https://github.com/vitest-dev/vitest/issues/2008#issuecomment-1434789112) mention that the issue was related to a debounce. So I thought we might face similar situation and advanced the time on after each which resolved the issue.

### DIFF
--- a/frontend/src/vitests/lib/components/project-detail/ProjectCommitment.spec.ts
+++ b/frontend/src/vitests/lib/components/project-detail/ProjectCommitment.spec.ts
@@ -9,6 +9,7 @@ import {
 import { renderContextCmp } from "$tests/mocks/sns.mock";
 import { ProjectCommitmentPo } from "$tests/page-objects/ProjectCommitment.page-object";
 import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
+import { advanceTime } from "$vitests/utils/timers.test-utils";
 import { SnsSwapLifecycle } from "@dfinity/sns";
 
 describe("ProjectCommitment", () => {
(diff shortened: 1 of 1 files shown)


frontend/src/vitests/lib/components/project-detail/ProjectCommitment.spec.ts
```diff
@@ -29,8 +30,15 @@ describe("ProjectCommitment", () => {
 
   beforeEach(() => {
     vi.clearAllMocks();
+
+    vi.clearAllTimers();
```
lmuntaner, line 34: But this component doesn't use any timeout, that doesn't make sense...

lmuntaner (commented):
I fear that the problem must be in the setup or something. This is a very simple component with no timers, nor date, not even mocks.

peterpeterparker:
We've got a cleaner and more global solution with #3518.

PR #3517, opened 2023-10-11
T Cycles display on canister page

peterpeterparker:
# Motivation

Cycles are not displayed as T Cycles on canister detail page - i.e. value is 10000 too high.

# Changes

- use cycles utils to format value

# Screenshots

Is:

Capture d’écran 2023-10-11 à 07 43 55

Fix:

Capture d’écran 2023-10-11 à 07 44 07


lmuntaner (approved):
Thanks!

My changes were the same.

I think you also need to change a test in CanisterDetail.spec.ts and CanisterPageHeadging.spec.ts

PR #3518, opened 2023-10-11
build: use atomics to synchronize vitest threads

peterpeterparker:
# Motivation

Another possible solution to try to solve #3516.

Source: https://github.com/vitest-dev/vitest/issues/2008#issuecomment-1436415644

### DIFF
--- a/frontend/vitest.config.ts
+++ b/frontend/vitest.config.ts
@@ -71,6 +71,19 @@ export default defineConfig(
       sequence: {
         hooks: "list",
       },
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Let's try it!

Can you ad a comment above `useAtomcs` explaining why and that is tinypool's problem?

peterpeterparker:
> Can you ad a comment above useAtomcs explaining why and that is tinypool's problem?

Good idea

PR #3524, opened 2023-10-11
Marked moved to an explicit dependency

peterpeterparker:
# Motivation

Our bundler and chunking strategy has evolved, therefore we can integrate `marked` as a dependency as any other dependencies.

In addition, this allow us to test the library with vitest without mock.

# Changes

- remove marked assed
- add dependency marked
- update lazy import (replace url with lib)
- add marked to chunking lazy strategy
- move and adapt test for vitest
- adapt test that uses marked per extension as library is not mocked anymore

# Todos


mstrasinskis (approved):
Nice! Thank you!

PR #3528, opened 2023-10-11
Migrate stores, keys, identities and derived to vitest

peterpeterparker:
# Motivation

Migrate stores, keys, identities and derived to vitest.

# Changes

- migrate test
- set `browser` mock environment variable globally to `true` as we are testing in browser mode
- remove local storage mock from vitest configuration because it is not required
- update `featureFlags` in vitest setup to reflect flags that were modified few days ago in jest setup


peterpeterparker:
This PR tests will probably pass when we have migrates the services. So let's wait for #3534

PR #3532, opened 2023-10-11
Migrate canisters test package to vitest

peterpeterparker:
# Motivation

Migrate canisters test package to vitest.


lmuntaner (approved):
Thanks!

PR #3540, opened 2023-10-12
Migrate remaining route test to vitest

peterpeterparker:
# Motivation

Migrate `Neurons.spec` route test to vitest.

# Notes

It does not seem to need particular changes and our improvement of the configuration seems to handle hanging process, so let's see in the CI.


lmuntaner (approved):
Great!

PR #3542, opened 2023-10-12
style: apply alignIconRight to proposal key-value blocks

mstrasinskis:
# Motivation

Redesign info icon alignment on proposal page ~~(behind the flag)~~.
- [x] [Related `gix-component` pr](https://github.com/dfinity/gix-components/pull/309)

# Changes

Add `alignIconRight` prop to KeyValueInfo

# Tests

Works

# Todos

# Screenshots

## Mobile Sns
image

# Desktop Nns
image


frontend/src/lib/components/proposal-detail/ProposalSystemInfoEntry.svelte
```diff
@@ -2,14 +2,15 @@
...
+  import { ENABLE_FULL_WIDTH_PROPOSAL } from "$lib/stores/feature-flags.store";
 
   export let labelKey: string;
   export let testId: string;
   export let value: string;
   export let description: string | undefined;
 </script>
 
-<KeyValuePairInfo {testId}>
+<KeyValuePairInfo {testId} alignIconRight={$ENABLE_FULL_WIDTH_PROPOSAL}>
```
lmuntaner, line 13: Why do we want this behind the flag? Isn't it already an improvement that works on its own?

mstrasinskis: On the proposal page the descriptions describe the values, not keys. In other places it could be different. I would prefer to start with the flag and drop it if needed later.

mstrasinskis: The task for this was created: https://dfinity.atlassian.net/browse/GIX-1975

lmuntaner: But this component is only for proposals `ProposalSystemInfoEntry`. Therefore, the description is about the value and the align to the right is correct.

The redesign doesn't change that, does it?

mstrasinskis: Yep, the new design has the icons on right. I don't have a strong opinion, so the restyling is not behind the flag anymore.

lmuntaner (approved):
Thanks!

PR #3543, opened 2023-10-12
Chore: Migrate sns-sale services

lmuntaner:
# Motivation

Migrate sns-sale services to vitest.

# Changes

* Normal migration changes. Plus an extra `spyOnToastsShow.mockReturnValue(Symbol("toast-id"));` on beforeEach because the services rely on the return value of `toastsShow`. And with vi it was `undefined`.

# Tests

Only test changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/services/sns-sale.services.ts
+++ b/frontend/src/lib/services/sns-sale.services.ts
@@ -14,7 +14,12 @@ import {
 import { getConditionsToAccept } from "$lib/getters/sns-summary";
 import { getCurrentIdentity } from "$lib/services/auth.services";
 import { loadBalance } from "$lib/services/icp-accounts.services";
-import { toastsError, toastsHide, toastsShow } from "$lib/stores/toasts.store";
+import {
+  toastsError,
+  toastsHide,
+  toastsShow,
+  toastsSuccess,
+} from "$lib/stores/toasts.store";
 import { transactionsFeesStore } from "$lib/stores/transaction-fees.store";
 import type { Account } from "$lib/types/account";
 import { ApiErrorKey } from "$lib/types/api.errors";
@@ -68,7 +73,6 @@ import { get } from "svelte/store";
 import { DEFAULT_TOAST_DURATION_MILLIS } from "../constants/constants";
 import { SALE_PARTICIPATION_RETRY_SECONDS } from "../constants/sns.constants";
 import { snsTicketsStore } from "../stores/sns-tickets.store";
-import { toastsSuccess } from "../stores/toasts.store";
 import { nanoSecondsToDateTime } from "../utils/date.utils";
 import { logWithTimestamp } from "../utils/dev.utils";
 import { formatToken } from "../utils/token.utils";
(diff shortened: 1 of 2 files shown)


frontend/src/lib/services/sns-sale.services.ts
```diff
@@ -68,7 +73,6 @@ import { get } from "svelte/store";
 import { DEFAULT_TOAST_DURATION_MILLIS } from "../constants/constants";
 import { SALE_PARTICIPATION_RETRY_SECONDS } from "../constants/sns.constants";
 import { snsTicketsStore } from "../stores/sns-tickets.store";
-import { toastsSuccess } from "../stores/toasts.store";
```
lmuntaner, line 71: This was not necessary, but it made sense to put them together.

frontend/src/vitests/lib/services/sns-sale.services.spec.ts
```diff
@@ -155,21 +152,22 @@ describe("sns-api", () => {
     spyOnSendICP.mockReset();
     spyOnNotifyParticipation.mockReset();
     spyOnToastsShow.mockReset();
+    spyOnToastsShow.mockReturnValue(Symbol("toast-id"));
```
lmuntaner, line 155: This was the change needed apart from normal migration changes.

frontend/src/vitests/lib/services/sns-sale.services.spec.ts
```diff
@@ -183,7 +181,7 @@ describe("sns-api", () => {
       certified: true,
     });
 
-    (importSnsWasmCanister as jest.Mock).mockResolvedValue({
+    (importSnsWasmCanister as vi.Mock).mockResolvedValue({
```
peterpeterparker, line 184: here you probably gonna have to transform `vi.Mock` into `Mock` and import the type otherwise `npm run lint` will complain and Playwright won't build.

peterpeterparker (approved):
LGTM, thanks!

frontend/src/vitests/lib/services/sns-sale.services.spec.ts
```diff
@@ -155,21 +152,25 @@ describe("sns-api", () => {
     spyOnSendICP.mockReset();
     spyOnNotifyParticipation.mockReset();
     spyOnToastsShow.mockReset();
+    // `mockReset` seems to add a mock not calling the actual function and not only reset the spy.
```
dskloetd, line 155: According to the [documentation](https://vitest.dev/api/mock.html#mockreset), this is what mockReset does:
> makes inner implementation an empty function (returning undefined, when invoked).

Jest has the same [documentation](https://jestjs.io/docs/mock-function-api#mockfnmockreset) but it doesn't seem to be what Jest actually does.

`mockRestore` is the function that restores the original behavior but it seems that in vitest it also stops the spy from spying.

lmuntaner: Thanks for checking it!

PR #3545, opened 2023-10-12
Don't define mocks directly in describe blocks.

dskloetd:
# Motivation

Code in `describe` blocks is called before any of the tests are executed.
For example the `jest.spyOn(` at the top of the `"removeFollowee "` block is called after the `jest.spyOn(` at the top of the `"addFollowee "` but before any of the tests in the `"addFollowee "` blocks.

This can cause confusing behavior if the tests depend on code from its own `describe` block.

Additionally, when we migrate to vitest, defining mock behavior on the same function again, prevents an earlier spy from spying. This means the following fails in vitest while the equivalent does not fail in Jest:
```
      const obj = { func: function() { } };
      const spy1 = vi.spyOn(obj, 'func');
      const spy2 = vi.spyOn(obj, 'func').mockImplementation(() => { });
      obj.func();
      expect(spy2).toBeCalled();
      expect(spy1).toBeCalled();
```

# Changes

In `frontend/src/tests/lib/services/sns-neurons.services.spec.ts` move all `spyOn`s that are directly in a `describe` block to a `beforeEach` block isntead.

After this change, migrating to vitest should be straightforward.

# Tests

pass

# Todos

not necessary

### DIFF
--- a/frontend/src/tests/lib/services/sns-neurons.services.spec.ts
+++ b/frontend/src/tests/lib/services/sns-neurons.services.spec.ts
@@ -143,12 +143,12 @@ describe("sns-neurons-services", () => {
     });
 
     describe("when sns parameters are loaded", () => {
-      const spyOnNervousSystemParameters = jest
-        .spyOn(governanceApi, "nervousSystemParameters")
-        .mockRejectedValue("should not be called");
+      let spyOnNervousSystemParameters;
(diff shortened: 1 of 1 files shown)


peterpeterparker (approved):
LGTM, thanks!!!!

PR #3546, opened 2023-10-12
Fix: Finalizing status in ProjectCard

lmuntaner:
# Motivation

Issue reported by a user: https://forum.dfinity.org/t/bug-projectpage-sns-launchpad-direct-url/23029/7?u=lmuntaner

Add the "Finalizing" status in the Launchpad, when we show all the SNSes.

Yet, to do that, we need to fetch the finalizing status of all the projects.

To not make unnecessary calls, I added a check that the project has finished within the last week. This way we don't make calls to for open or old projects.

# Changes

* Changes in `loadSnsFinalizationStatus` to not call the API unless it's possible that the project is finalizing. Also change the argument to accept another parameter `forceFetch`.
* Call `loadSnsFinalizationStatus` on mounting the ProjectDetailCard.
* Use `createIsSnsFinalizingStore` on ProjectCardSwap to render "Finalizing" instead of "Completed" when needed.
* Force fetching autofinalization after a participation. At that moment the project might not be loaded as Committed locally yet, but it might be Committed already.
* Fix the `SnsFinalizationStatusStore` that might be `undefined`.
* New sns util `swapEndedMoreThanOneWeekAgo` for convenience.

# Tests

* Add `swapDueTimestampSeconds` parameter when creating an SNS.
* Add `getStatus` to ProjectCardPo.
* New ProjectCardSwapInfoPo.
* Add a test case in ProjectCard that shows the Finalizing status. Also adapt it so that it has SNS setup. Until now it was not needed because it was using the props only.
* Add a test case in ProjectCardSwapInfo for the Finalizing status.
* Fix the swap ending time in ProjectDetail so that it triggers the call to get finalizing data.
* Edit tests in sns-finalization.services.spec after the changes in `loadSnsFinalizationStatus`.
* Test new sns util `swapEndedMoreThanOneWeekAgo`.

# Todos

### DIFF
--- a/frontend/src/lib/components/launchpad/ProjectCard.svelte
+++ b/frontend/src/lib/components/launchpad/ProjectCard.svelte
@@ -12,6 +12,9 @@
   import { nonNullish } from "@dfinity/utils";
   import { onMount } from "svelte";
   import { loadSnsFinalizationStatus } from "$lib/services/sns-finalization.services";
+  import type { Readable } from "svelte/store";
+  import { createIsSnsFinalizingStore } from "$lib/stores/sns-finalization-status.store";
+  import type { Principal } from "@dfinity/principal";
 
   export let project: SnsFullProject;
 
@@ -21,7 +24,8 @@
 
   let summary: SnsSummary;
   let swapCommitment: SnsSwapCommitment | undefined;
-  $: ({ summary, swapCommitment } = project);
+  let rootCanisterId: Principal;
+  $: ({ summary, swapCommitment, rootCanisterId } = project);
 
   let logo: string;
   let name: string;
@@ -39,6 +43,9 @@
 
   let href: string;
   $: href = `${AppPath.Project}/?project=${project.rootCanisterId.toText()}`;
+
+  let isFinalizingStore: Readable<boolean>;
+  $: isFinalizingStore = createIsSnsFinalizingStore(rootCanisterId);
 </script>
 
 <Card
@@ -53,7 +60,7 @@
 
   <p class="value description">{description}</p>
 
-  <ProjectCardSwapInfo {project} />
+  <ProjectCardSwapInfo isFinalizing={$isFinalizingStore} {project} />
 
   <SignedInOnly>
     <!-- TODO L2-751: handle fetching errors -->
--- a/frontend/src/lib/components/launchpad/ProjectCard.svelte
+++ b/frontend/src/lib/components/launchpad/ProjectCard.svelte
@@ -10,8 +10,14 @@
   import { getCommitmentE8s } from "$lib/utils/sns.utils";
   import SignedInOnly from "$lib/components/common/SignedInOnly.svelte";
   import { nonNullish } from "@dfinity/utils";
+  import { onMount } from "svelte";
+  import { loadSnsFinalizationStatus } from "$lib/services/sns-finalization.services";
 
   export let project: SnsFullProject;
+
+  onMount(() => {
+    loadSnsFinalizationStatus(project.rootCanisterId);
+  });
 
   let summary: SnsSummary;
   let swapCommitment: SnsSwapCommitment | undefined;
--- a/frontend/src/lib/components/launchpad/ProjectCard.svelte
+++ b/frontend/src/lib/components/launchpad/ProjectCard.svelte
@@ -16,7 +16,7 @@
   export let project: SnsFullProject;
 
   onMount(() => {
-    loadSnsFinalizationStatus(project.rootCanisterId);
+    loadSnsFinalizationStatus({ rootCanisterId: project.rootCanisterId });
   });
 
   let summary: SnsSummary;
--- a/frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
+++ b/frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
@@ -17,17 +17,15 @@
   import ProjectUserCommitmentLabel from "$lib/components/project-detail/ProjectUserCommitmentLabel.svelte";
   import { getCommitmentE8s } from "$lib/utils/sns.utils";
   import { nonNullish } from "@dfinity/utils";
-  import type { Readable } from "svelte/store";
-  import { createIsSnsFinalizingStore } from "$lib/stores/sns-finalization-status.store";
-  import type { Principal } from "@dfinity/principal";
   import TestIdWrapper from "../common/TestIdWrapper.svelte";
 
   export let project: SnsFullProject;
+  // The data to know whether it's finalizing or not is not in the SnsFullProject.
+  export let isFinalizing: boolean;
 
   let summary: SnsSummary;
   let swapCommitment: SnsSwapCommitment | undefined;
-  let rootCanisterId: Principal;
-  $: ({ summary, swapCommitment, rootCanisterId } = project);
+  $: ({ summary, swapCommitment } = project);
 
   let swap: SnsSummarySwap;
   $: ({ swap } = summary);
@@ -42,9 +40,6 @@
 
   let durationTillStart: bigint | undefined;
   $: durationTillStart = durationTillSwapStart(swap);
-
-  let isFinalizingStore: Readable<boolean>;
-  $: isFinalizingStore = createIsSnsFinalizingStore(rootCanisterId);
 
   let myCommitment: TokenAmount | undefined = undefined;
   $: {
@@ -64,7 +59,7 @@
     {#if lifecycle === SnsSwapLifecycle.Committed}
       <dt class="label">{$i18n.sns_project_detail.status_completed}</dt>
       <!-- is finalizing is not a lifecycle, it can be Committed and Finalizing or Not Finalizing -->
-      {#if $isFinalizingStore}
+      {#if isFinalizing}
         <dd class="value">{$i18n.sns_project_detail.status_finalizing}</dd>
       {:else}
         <dd class="value">{$i18n.sns_project_detail.completed}</dd>
--- a/frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
+++ b/frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
@@ -20,6 +20,7 @@
   import type { Readable } from "svelte/store";
   import { createIsSnsFinalizingStore } from "$lib/stores/sns-finalization-status.store";
   import type { Principal } from "@dfinity/principal";
+  import TestIdWrapper from "../common/TestIdWrapper.svelte";
 
   export let project: SnsFullProject;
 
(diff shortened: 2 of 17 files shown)


frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte
```diff
@@ -51,23 +59,30 @@
...
+        <dd class="value">{$i18n.sns_project_detail.completed}</dd>
+      {/if}
+    {/if}
 
-  <!-- Sale is adopted -->
-  {#if lifecycle === SnsSwapLifecycle.Adopted && durationTillStart !== undefined}
-    <dt class="label">{$i18n.sns_project_detail.starts}</dt>
-    <dd class="value">{secondsToDuration(durationTillStart)}</dd>
-  {/if}
+    <!-- Sale is adopted -->
```
lmuntaner, line 74: These lines are just indented, no other changes here.

lmuntaner:
@dskloetd please review

It seems unbelievable how much had to be done to change a word.

frontend/src/lib/components/launchpad/ProjectCard.svelte
```diff
@@ -10,9 +10,15 @@
...
   import SignedInOnly from "$lib/components/common/SignedInOnly.svelte";
   import { nonNullish } from "@dfinity/utils";
+  import { onMount } from "svelte";
+  import { loadSnsFinalizationStatus } from "$lib/services/sns-finalization.services";
 
   export let project: SnsFullProject;
 
+  onMount(() => {
+    loadSnsFinalizationStatus({ rootCanisterId: project.rootCanisterId });
+  });
```
dskloetd, line 20: Would it make sense to put this in `frontend/src/lib/components/launchpad/ProjectCardSwapInfo.svelte` since that's where it's used?

frontend/src/lib/services/sns-finalization.services.ts
```diff
@@ -1,12 +1,41 @@
...
+  );
+  // We don't want to do calls that we know are unnecessary.
+  // The project starts finalizing right after the sale ends.
+  // It might take a few hours.
+  // Waiting one week ensures that the project is not finalizing anymore.
+  if (
+    !forceFetch &&
+    (isNullish(summary) ||
+      summary.swap.lifecycle !== SnsSwapLifecycle.Committed ||
+      swapEndedMoreThanOneWeekAgo({ summary, nowInSeconds: nowInSeconds() }))
```
dskloetd, line 35: Is there a benefit to passing in `nowInSeconds` rather than `swapEndedMoreThanOneWeekAgo` calling `nowInSeconds()`?

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
```diff
@@ -132,6 +150,38 @@ describe("ProjectCard", () => {
...
+
+    it("should render finalizing status if swap is finalizing", async () => {
+      vi.spyOn(saleApi, "queryFinalizationStatus").mockResolvedValue(
+        createFinalizationStatusMock(true)
+      );
+
+      const { container } = render(ProjectCard, {
+        props: {
+          project: {
+            ...mockSnsFullProject,
```
dskloetd, line 162: How about making a `createMockSnsFullProject` helper function for convenience?

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
```diff
@@ -132,6 +150,38 @@ describe("ProjectCard", () => {
...
+              },
+            },
+          },
+        },
+      });
+
+      const po = ProjectCardPo.under(new JestPageObjectElement(container));
+      await runResolvedPromises();
+
+      await waitFor(async () =>
```
dskloetd, line 181: Do you need `waitFor`?

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
```diff
@@ -132,6 +150,38 @@ describe("ProjectCard", () => {
...
+        expect(await po.getStatus()).toBe("Status Finalizing")
```
dskloetd, line 182: Is there a corresponding test where the status is not finalizing?

frontend/src/vitests/lib/components/launchpad/ProjectCardSwapInfo.spec.ts
```diff
@@ -120,4 +128,26 @@ describe("ProjectCardSwapInfo", () => {
...
+          summary: summaryForLifecycle(SnsSwapLifecycle.Committed),
+        },
+      },
+    });
+
+    const po = ProjectCardSwapInfoPo.under(
+      new JestPageObjectElement(container)
+    );
+
+    expect(await po.getStatus()).toBe("Status Finalizing");
```
dskloetd, line 151: Is there a corresponding test with a different status?

frontend/src/vitests/lib/services/sns-finalization.services.spec.ts
```diff
@@ -16,36 +19,110 @@ describe("sns-finalization-services", () => {
...
+            rootCanisterId,
+            lifecycle: SnsSwapLifecycle.Open,
+            swapDueTimestampSeconds: nowInSeconds - SECONDS_IN_MONTH,
+          },
+        ]);
+      });
 
-      await loadSnsFinalizationStatus(rootCanisterId);
+      it("should call not load finalization status in store nor call api", async () => {
+        await loadSnsFinalizationStatus({ rootCanisterId });
```
dskloetd, line 120: With forceFetch it would still load even if it was more than a week ago, right?
Is there a test for that?

frontend/src/lib/components/launchpad/ProjectCard.svelte
lmuntaner: I see `ProjectCardSwapInfo` more as a helper component that we might easily change and refactor. It doesn't mean anything in UI.

That's why I don't want to add any extra logic there.

It makes sense to me that the ProjectCard loads the data needed inside it. Maybe I should pass instead the status as a prop to `ProjectCardSwapInfo`. That would make more sense even. But maybe in another PR.

frontend/src/lib/services/sns-finalization.services.ts
lmuntaner: Just to make the `util` a pure function.

I can revert though. I was thinking of having utils as pure functions.

But maybe we want a time sensitive utils. We used that a lot.

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: I can add that in another PR. Although ideally I would like to remove this idea of SnsFullProject.

It's one more type we need to maintain.

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: Yes, because initially it's Completed, until the request of the finalizing status finishes.

I guess I could wait for the store to be filled also.

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: No. I can add it in another PR, I didn't want to add it for this PR. But I can also do it if needed. This PR is already big enough, that's why I didn't do it.

frontend/src/vitests/lib/components/launchpad/ProjectCardSwapInfo.spec.ts
lmuntaner: Yes, for all other statuses. But they don't use the page object.

frontend/src/vitests/lib/services/sns-finalization.services.spec.ts
lmuntaner: Yes, it will. Let me add the test.

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
dskloetd: Having a corresponding test makes both tests much easier to review because you can focus on the differences. "If this one thing in the setup changes, this one expectation changes."
So I think it makes sense to have both in the same PR.

frontend/src/vitests/lib/components/launchpad/ProjectCardSwapInfo.spec.ts
dskloetd: In the future, try to use a consistent pattern for tests. (Ideally moving a whole file to POs first.)
As mentioned above similar tests make each other easier to understand because you can focus on their differences.

frontend/src/lib/services/sns-finalization.services.ts
lmuntaner: Yes, I'm also not super happy about it. I think we should differentiate better what's a pure function util and what not.

frontend/src/vitests/lib/components/launchpad/ProjectCardSwapInfo.spec.ts
lmuntaner: In general, I agree. But it's quite some work to convert some tests to use PO and the best practices of mocking only the API for example.

I'm not so sure I can commit to it.

frontend/src/lib/components/launchpad/ProjectCard.svelte
lmuntaner: true, done!

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: Done. Maybe it's too early for x-mas tress...

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: Maybe Svelte hasn't changed the component yet. The api might finish, but it still needs to update the store, and that will trigger a rerender.

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: done

lmuntaner:

frontend/src/vitests/lib/components/launchpad/ProjectCardSwapInfo.spec.ts
dskloetd: At least the tests that are similar/corresponding to the one you're adding, should be easy to convert/add, right? That would still help provide context for the new test (in the future).

frontend/src/vitests/lib/components/launchpad/ProjectCard.spec.ts
dskloetd: The purpose of `await runResolvedPromises()` is that all of that happens and we don't need `waitFor`.
Unless there is an explicit delay (?), but in that case we should be using fake timers so the test doesn't actually wait.

We should really try to avoid `waitFor` in tests because the same test could pass with both
```
      await waitFor(async () =>
        expect(await po.getStatus()).toBe("Status Finalizing")
      );
 ```
 and
 ```
      await waitFor(async () =>
        expect(await po.getStatus()).toBe("Status Completed")
      );
 ```
if the status changes during the `waitFor`.

dskloetd: I checked out your branch, removed the `waitFor`, and ran the test. And it passed.

lmuntaner: Weird, it does. Good. Maybe with the change of the store being moved up?

frontend/src/vitests/lib/components/launchpad/ProjectCardSwapInfo.spec.ts
lmuntaner: For here yes, I can do that in another PR. Or do you mean I should do it in this PR?

dskloetd (approved):

PR #3551, opened 2023-10-13
GIX-1937: Use min max direct participation fields from aggregator

lmuntaner:
# Motivation

Add functionality after Neurons' Fund enhancements.

In this PR, we use the new fields from the SNS Aggregator when we convert the data to SnsSummary.

# Changes

* Populate new fields `min_direct_participation_icp_e8s` and `max_direct_participation_icp_e8s` from the different places into the SnsSummary. Those fields are in the `Init` from `get_init`, but they are also in the `init` inside the `swap` and also in the `params` of the `swap`.

# Tests

* Add the fields in the test case for the NF's enhancements fields. The other tests have those fields not populated and are converted to `[]` as it was hardcoded before.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
+++ b/frontend/src/lib/utils/sns-aggregator-converters.utils.ts
@@ -210,8 +210,12 @@ const convertSwapInitParams = (
               )
             )
           : [],
-        min_direct_participation_icp_e8s: [],
-        max_direct_participation_icp_e8s: [],
+        min_direct_participation_icp_e8s: toNullable(
+          convertOptionalNumToBigInt(init.min_direct_participation_icp_e8s)
+        ),
+        max_direct_participation_icp_e8s: toNullable(
+          convertOptionalNumToBigInt(init.max_direct_participation_icp_e8s)
+        ),
       }
     : undefined;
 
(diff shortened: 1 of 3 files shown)


frontend/src/lib/utils/sns-aggregator-converters.utils.ts
```diff
@@ -233,8 +237,12 @@ const convertSwapParams = (params: CachedSwapParamsDto): SnsParams => ({
...
     convertOptionalNumToBigInt(params.sale_delay_seconds)
   ),
-  min_direct_participation_icp_e8s: [],
-  max_direct_participation_icp_e8s: [],
+  min_direct_participation_icp_e8s: toNullable(
+    convertOptionalNumToBigInt(params.min_direct_participation_icp_e8s)
+  ),
+  max_direct_participation_icp_e8s: toNullable(
+    convertOptionalNumToBigInt(params.max_direct_participation_icp_e8s)
+  ),
```
dskloetd, line 245: Isn't SnsParams deprecated? Do you really need these fields in both places?

lmuntaner: I don't "need" them, but because I convert those types to the expected sns-js type, I would need to explicitly leave it as `[]`. Instead, it's more consistent to populate them.

It's not so much extra work and this way there are no weird rules to remember.

dskloetd: Thanks for clarifying. The whole SwapParams type will be deleted soon-ish, right?

dskloetd (approved):

frontend/src/lib/utils/sns-aggregator-converters.utils.ts
lmuntaner: As soon as I have time to move the access to data to the other fields all the `swap` field should disappear.

PR #3552, opened 2023-10-13
GIX-1962: Proof Of Concept for My Tokens page

lmuntaner:
# Motivation

This is a proof of concept for a new idea on how to split the page with functionality and UI components.

Main idea: Be able to render the whole page for designers to work only with CSS.

The route component will load all the stores with initial data with the related services.

All subsequent data and actions is dispatched from components as custom events, then handled from the parent.

# Changes

* Route component that renders a Page component.
* Page component and children do not interact directly with the API. (IMPORTANT)
* Page component and children do not read from the store or the context.

# Tests

* We could easily do screenshot testing of the static pages.

# Todos

### DIFF
--- a/frontend/src/routes/(app)/(nns)/my-tokens/+page.svelte
+++ b/frontend/src/routes/(app)/(nns)/my-tokens/+page.svelte
@@ -1,0 +1,32 @@
+<script lang="ts">
+  import MyTokens from "$lib/pages/MyTokens.svelte";
+  import { ActionTypes, type Action } from "$lib/types/actions";
+
+  // Call services to load necessary data. Maybe one service to rule them all?
+
+  // This would be a derived store.
+  const data = [
+    { name: "ICP", balance: "3.14", id: "1" },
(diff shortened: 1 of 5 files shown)


frontend/src/routes/(app)/(nns)/my-tokens/+page.svelte
```diff
@@ -0,0 +1,32 @@
...
+        return;
+      }
+      default: {
+        console.log("Unknown action");
+      }
+    }
+  };
+</script>
+
+<MyTokens {data} on:nnsAction={manageActions} />
```
mstrasinskis, line 32: Does this mean that all the container-like components should bubble up this event?

lmuntaner: Yes, all actions that interact with the API or load more data should trigger an event.

Kind of bringing the Redux like functionality.

It doesn't need to be an event. It needs to be something that we can easily "disable" when we render it in "static" mode.

lmuntaner:
Closing, PoC done

PR #3553, opened 2023-10-13
GIX-1937: Use fields in min/max participation getters

lmuntaner:
# Motivation

Add support for the new Neurons' Fund enhancements.

In this PR, the getters `getMinDirectParticipation` and `getMaxDirectParticipation` return the actual value from the SNS project.

# Changes

* Return the values in the `init` field of the SnsSummary.

# Tests

* Add two fields to createSummary helper.
* Remove the mocks of the getters and instead create a summary with the desired values.

# Todos


lmuntaner:
@dskloetd please review

The last of the PRs to use the new min and max fields. We can't yet test them because the sns cli doesn't support the new fields.

dskloetd (approved):

PR #3562, opened 2023-10-17
GIX-1986: Make SnsTransactionModal page agnostic

lmuntaner:
# Motivation

Implement the MyTokens page.

In this PR, the SnsTransactionModal is decoupled from the page it is rendered. Instead, it receives the data needed as props.

# Changes

* Add props `rootCanisterId`, `token` and `transactionFee` to SnsTransactionModal.
* Pass the new props where the modal was used: SnsWallet and SnsAccountsFooter.

# Tests

* Pass props in the SnsTransactionModal and remove related mocks.
* Add sns setup in routes/Accounts.spec

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/pages/SnsWallet.svelte
+++ b/frontend/src/lib/pages/SnsWallet.svelte
@@ -33,6 +33,7 @@
   import SnsBalancesObserver from "$lib/components/accounts/SnsBalancesObserver.svelte";
   import WalletPageHeader from "$lib/components/accounts/WalletPageHeader.svelte";
   import WalletPageHeading from "$lib/components/accounts/WalletPageHeading.svelte";
+  import { snsSelectedTransactionFeeStore } from "$lib/derived/sns/sns-selected-transaction-fee.store";
 
   let showModal: "send" | undefined = undefined;
 
@@ -179,11 +180,14 @@
   </Footer>
 </Island>
 
-{#if showModal}
+{#if showModal && nonNullish($snsOnlyProjectStore) && nonNullish(token)}
   <SnsTransactionModal
     on:nnsClose={() => (showModal = undefined)}
     selectedAccount={$selectedAccountStore.account}
+    rootCanisterId={$snsOnlyProjectStore}
     loadTransactions
+    {token}
+    transactionFee={$snsSelectedTransactionFeeStore}
   />
 {/if}
 
--- a/frontend/src/lib/pages/SnsWallet.svelte
+++ b/frontend/src/lib/pages/SnsWallet.svelte
@@ -180,7 +180,7 @@
   </Footer>
 </Island>
 
-{#if showModal && nonNullish($snsOnlyProjectStore) && nonNullish(token)}
+{#if showModal && nonNullish($snsOnlyProjectStore)}
   <SnsTransactionModal
     on:nnsClose={() => (showModal = undefined)}
     selectedAccount={$selectedAccountStore.account}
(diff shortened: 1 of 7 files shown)


lmuntaner:

frontend/src/lib/pages/SnsWallet.svelte
```diff
@@ -179,11 +180,14 @@
   </Footer>
 </Island>
 
-{#if showModal}
+{#if showModal && nonNullish($snsOnlyProjectStore) && nonNullish(token)}
```
mstrasinskis, line 183: Probably `&& nonNullish(token)` is redundant since the Modal already handles `token: undefined`.

mstrasinskis (approved):
LGTM, thanks.

frontend/src/tests/lib/modals/accounts/SnsTransactionModal.spec.ts
```diff
@@ -28,33 +19,28 @@ vi.mock("$lib/services/sns-accounts.services", () => {
...
       },
     });
 
-  beforeAll(() =>
-    vi.spyOn(authStore, "subscribe").mockImplementation(mockAuthStoreSubscribe)
-  );
-
   beforeEach(() => {
+    resetIdentity();
     vi.spyOn(snsAccountsStore, "subscribe").mockImplementation(
```
dskloetd, line 41: How does the modal depend on the `snsAccountsStore`? It doesn't seem to import it.

dskloetd (commented):
Would it be possible to test that `SnsWallet` and `SnsAccountsFooter` pass the correct props to `SnsTransactionModal`?

frontend/src/tests/lib/modals/accounts/SnsTransactionModal.spec.ts
lmuntaner: Not always, I moved it from beforeEach to one of the tests to make it clear. And I filled the store instead of mocking it.

lmuntaner:
@dskloetd is it ok for you if I merge?

dskloetd:
I would appreciate an answer to my earlier question:

> Would it be possible to test that `SnsWallet` and `SnsAccountsFooter` pass the correct props to `SnsTransactionModal`?

lmuntaner:
> I would appreciate an answer to my earlier question:
> 
> > Would it be possible to test that `SnsWallet` and `SnsAccountsFooter` pass the correct props to `SnsTransactionModal`?

Why would I do that? I can add a test for a transaction, but testing that props are passed is testing implementation details.

dskloetd:
I don't mean that you should literally test the props being passed. I mean that there should be a test that would fail if you pass the wrong props. Those props serve a purpose so they should be correct so some test should fail if they are not correct.

lmuntaner:
> I don't mean that you should literally test the props being passed. I mean that there should be a test that would fail if you pass the wrong props. Those props serve a purpose so they should be correct so some test should fail if they are not correct.

Pending on #3567

lmuntaner:
> I don't mean that you should literally test the props being passed. I mean that there should be a test that would fail if you pass the wrong props. Those props serve a purpose so they should be correct so some test should fail if they are not correct.

Done, I added a test on SnsWallet that makes a transaction and checked that the api was called with the proper parameters. Some of them are the ones passed as props.

lmuntaner:

dskloetd (approved):
Thanks! I think SnsAccountsFooter should have a similar test because it separately passes props to SnsTransactionModal.

lmuntaner:
> Thanks! I think SnsAccountsFooter should have a similar test because it separately passes props to SnsTransactionModal.

Let me add it in another PR, I'll change the test to mock only the API as well.

PR #3564, opened 2023-10-17
Bump ic-js

dskloetd:
# Motivation

We want to be able to use `retrieveBTCWithApproval` on the ckBTC minter canister, which was added to ic-js in https://github.com/dfinity/ic-js/pull/454.

# Changes

1. Updated `frontend/package-lock.json` by running `npm run upgrade:next`.
2. In `frontend/src/lib/utils/anonymize.utils.ts` added `isGenesis` field to return value of `anonymizeNeuronInfo` and `anonymizeFullNeuron` to match the new types for `Neuron` and `NeuronInfo`, changed in https://github.com/dfinity/ic-js/pull/453.

# Tests

Manually created an anonymized log by clicking 6 times and typing `c`.

# Todos

- [ ] Add entry to changelog (if necessary).
not necessary

PR #3566, opened 2023-10-17
Feat: Upgrade next dependencies

lmuntaner:
# Motivation

Upgrade ic-js depenencies.

# Changes

## Automatic changes

* Changes in package.lock.json. Using npm script `npm run upgrade:next`.

## Manual changes

* Add `isGenesis` to anonymize.
* Return `undefined` from anonymize helpers if `neuron` is undefined or null.

# Tests

Still passing

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/utils/anonymize.utils.ts
+++ b/frontend/src/lib/utils/anonymize.utils.ts
@@ -147,6 +147,7 @@ export const anonymizeNeuronInfo = async (
     votingPower: await anonymizeAmount(votingPower),
     ageSeconds,
     fullNeuron: await anonymizeFullNeuron(fullNeuron),
+    isGenesis: neuron.isGenesis,
   };
 };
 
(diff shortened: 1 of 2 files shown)


frontend/src/lib/utils/anonymize.utils.ts
```diff
@@ -147,6 +147,7 @@ export const anonymizeNeuronInfo = async (
     votingPower: await anonymizeAmount(votingPower),
     ageSeconds,
     fullNeuron: await anonymizeFullNeuron(fullNeuron),
+    isGenesis: neuron.isGenesis,
```
dskloetd: The other fields from `neuron` are assigned on lines 126-137.
Let's be consistent with that, like here: https://github.com/dfinity/nns-dapp/pull/3564/files

frontend/src/lib/utils/anonymize.utils.ts
```diff
@@ -214,6 +215,7 @@ export const anonymizeFullNeuron = async (
     spawnAtTimesSeconds,
     dissolveState,
     followees: await mapPromises(followees, anonymizeFollowees),
+    isGenesis: neuron.isGenesis,
```
dskloetd: same

dskloetd (approved):
Thanks!

PR #3570, opened 2023-10-17
GIX-1994: Preserve provenance

bitdivine:
# Motivation
We would like to break updates into smaller chunks.  This means that .did files and derived Rust may not all be from the same IC commit, so we need to track, for each file, where it came from.

# Changes
* Introduce provenance headers that declare where each .did or derived rust file came from.

# Tests
* CI tests that generating the current .did and derived rust files yields the same result as the committed code.

# Todos


declarations/nns_governance/nns_governance.did
```diff
@@ -1,3 +1,4 @@
+//! nns_governance .did file obtained from: <https://raw.githubusercontent.com/dfinity/ic/2d57e93dabc5f13258d0dee1ffb2363ddce7fe62/rs/nns/governance/canister/governance.did>
```
bitdivine, line 1: Note: The <> around the link corresponds to [the Markdown autolink convention](https://sgmljs.net/docs/markdown-autolink-examples.html).  When the .did file is converted to Rust, the comment is included in the rustdoc and there the autolink behaves correctly.

scripts/did2rs.sh
```diff
@@ -24,6 +24,9 @@ print_help() {
 	EOF
 }
 
+# This command, with paths stripped down to the basename.
+STRIPPED_COMMAND="scripts/did2rs.sh $(printf " %q" scripts/did2rs.sh "${@##*/}")"
```
bitdivine, line 28: Note: Stripping the paths makes the output reproducible as the paths can contain e.g. the absolute path of the repo, which differs between execution environments.  The shorter output I also find more readable.

I did consider, as an alternative, making all the paths relative, however backed out of that as there are many corner cases that would, I think, be much more confusing than the basename.

lmuntaner: Isn't the script duplicated?

```
scripts/did2rs.sh scripts/did2rs.sh --canister sns_governance --out ic_sns_governance.rs --header did2rs.header --traits Serialize\,\ Clone\,\ Debug
```

`scripts/did2rs.sh` appears twice at the beginning.

scripts/update_ic_commit
```diff
@@ -39,7 +39,10 @@ source "$(clap.build)"
     local upstream_path="$2"
     local local_path="declarations/${canister}/${canister}.did"
     mkdir -p "$(dirname "$local_path")"
-    curl -sfL --retry 5 "$IC_URL$upstream_path" >"${local_path}"
+    {
+      echo "//! $canister .did file obtained from: <$IC_URL$upstream_path>"
```
lmuntaner, line 43: Can you add here also, "Created by scripts/update_ic_commit"?

lmuntaner (commented):
A couple of comments, thanks! That looks much nicer!

scripts/update_ic_commit
bitdivine: Sure thing.  It might make sense to put come context in as otherwise we have two "Created" lines.  Better might be:
```
//! Rust code created from candid by:  `scripts/did2rs.sh --canister sns_swap --out ...
//! Candid for canister `foo` obtained by `scripts/update_ic_commit` from: <http://...>
```
What do you think?

lmuntaner (approved):
Thanks!

PR #3571, opened 2023-10-17
List playwright results

dskloetd:
# Motivation

I've noticed many times that the playwright-failure-results artifact contains only an index.html file and nothing else, which makes it useless for debugging.
When deliberately making an e2e fail, I was not able to reproduce this.
I'm not sure if Playwright fails to generate the right files or if GitHub fails to upload them.

# Changes

List the files generated by Playwright, so that we compare the logs with the artifact to see if the problem is with Playwright or with Github.

# Tests

I made a test fail deliberately and checked the GitHub Actions log.

# Todos

not necessary


bitdivine (approved):
LGTM.  Thanks!

PR #3579, opened 2023-10-18
Chore: Add sns transaction test in SnsAccountsFooter

lmuntaner:
# Motivation

Add a missing test when making the SnsTransactionModal agnostic of the page.

In this PR, add a test that makes an SNS transaction in SnsAccountsFooter.

# Changes

All changes in SnsAccountsFooter.spec
* Remove mock of `syncSnsAccounts` and instead mock sns ledger api.
* Expect calls to api instead to `syncSnsAccounts`.
* Add test case for SNS transaction

# Tests

Only test changes.

# Todos

Not necessary.


frontend/src/tests/lib/components/accounts/SnsAccountsFooter.spec.ts
```diff
@@ -74,10 +85,48 @@ describe("SnsAccountsFooter", () => {
...
+    };
+    await testTransferTokens({
+      result,
+      amount: "2",
+      destinationAddress: encodeIcrcAccount(destinationAccount),
+    });
+
+    await runResolvedPromises();
+
+    expect(snsLedgerApi.snsTransfer).toHaveBeenCalledTimes(1);
```
dskloetd, line 122: The previous test uses `toBeCalledTimes` but here you use `toHaveBeenCalledTimes`.
Personally I prefer the shorter version but let's at least be consistent within a file.

dskloetd (approved):
Thanks!

frontend/src/tests/lib/components/accounts/SnsAccountsFooter.spec.ts
lmuntaner: done

PR #3583, opened 2023-10-19
Compile before running test

peterpeterparker:
# Motivation

Detect compilation errors before running test. That way we do not rely on playwright job to detect those as for example in this [job](https://github.com/dfinity/nns-dapp/actions/runs/6561947531/job/17823049086?pr=3580)

# Changes

- `tsc` before running the test suite
- create a seprate `vitest` script that way we can run a test locally without building everything


frontend/package.json
```diff
@@ -16,7 +16,8 @@
     "lint": "prettier --check './**/*.{ts,js,mjs,json,scss,svelte,html}' && eslint --max-warnings 0 .",
     "format": "prettier --write './**/*.{ts,js,mjs,json,scss,svelte,html}'",
     "format:modified": "prettier --write $(git diff --diff-filter=ACM --name-only | egrep \"\\.(ts|js|mjs|json|scss|svelte|html)$\" | cut -d'/' -f2-)",
-    "test": "TZ=UTC vitest --config ./vitest.config.ts",
+    "test": "tsc --noEmit -p src/tests/e2e/tsconfig.json && npm run vitest",
```
dskloetd, line 19: Does it make sense to point specifically to the `e2e` config here? It seems confusing.
Do other configs provide less coverage?

peterpeterparker: No you are right, it should point to to the spec file directly. Updated.

dskloetd (approved):
Thanks!

PR #3588, opened 2023-10-19
Use props instead of Context CkBTCWallet

dskloetd:
# Motivation

Using a context makes it difficult to see what is used where.
Some of these props might not be necessary and when we stop using them it should be clear that we can remove them completely.

# Changes

1. Remove `CkBTCWalletContext`.
2. Pass `store, reloadAccount, reloadAccountFromStore` directly as props from `CkBTCWallet` to `CkBTCWalletFooter`

# Tests

`CkBTCWalletContextTest.svelte` has been updated to make the existing test pass again.
Manual testing seems to work.
You can try it here: https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #3591, opened 2023-10-20
Add trait

bitdivine:
# Motivation
Patch files are a general solution for editing autogenerated rust code.  They are, however, fragile, so we wish to move towards more specific, targeted, robust changes.

Most of the changes made by patch files now add traits to types.  We can make a command that does precisely that and delete most of the patch lines.

# Changes
- Add a command that adds traits to Rust types.
- Use the above command instead of patch files to edit traits.

# Tests
###  Expectation: The Rust files should not change.
See this PR.  There are no changes to Rust code.

### Expectation: Updates should be more robust
This update succeeds now, where previously it failed:
```
scripts/update_ic_commit --ic_commit dd51544944987556c978e774aa7a1992e5c11542
cargo check
```

# Todos

- Replace the rust-editing script with a proper parser.


lmuntaner (approved):
I like the idea!

Can we do this for all the needed changes and ditch the patch files?

PR #3594, opened 2023-10-23
Fix: Allow robots to crawl

lmuntaner:
# Motivation

We want search engines to crawl the NNS Dapp.

There was a bug with the logic of the robots confused. We only want them to crawl mainnet.

# Changes

* Change the `config.sh` to set as ROBOTS `""` on mainnet.

# Tests

* Change the config test expected value for mainnet.

# Todos

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -95,7 +95,7 @@ local_deployment_data="$(
   export TVL_CANISTER_ID
 
   : "Define the robots text, if any"
-  if [[ "$DFX_NETWORK" == "local" ]] || [[ "$DFX_NETWORK" == "testnet" ]]; then
+  if [[ "$DFX_NETWORK" == "mainnet" ]]; then
     ROBOTS=''
   else
     # shellcheck disable=SC2089 # yes, we really want the backslash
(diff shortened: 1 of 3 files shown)


lmuntaner:
@peterpeterparker @bitdivine please review

config.sh
```diff
@@ -95,7 +95,7 @@ local_deployment_data="$(
   export TVL_CANISTER_ID
 
   : "Define the robots text, if any"
-  if [[ "$DFX_NETWORK" == "local" ]] || [[ "$DFX_NETWORK" == "testnet" ]]; then
```
bitdivine, line 98: Actually, there is already an if/or clause [here](https://github.com/dfinity/nns-dapp/blob/main/config.sh#L97).  I recommend updating that instead.  It looks as if it was my bug - when I moved the logic across I inverted the NOT. :-(

lmuntaner: Aren't both the same line?

bitdivine: Ah, yes they are.

bitdivine (approved):

PR #3595, opened 2023-10-23
GIX-1982: Tokens page layout

lmuntaner:
# Motivation

Implement the new Tokens page.

In this PR, setup the layout as expected in Figma.

This is the first time for this layout, but it seems to follow the rules of the current detail pages. Therefore, I applied the same rules and created a ticket to confirm them with Artem once he's back.

# Changes

* New component `MainWrapper.svelte`. For now used only in the Tokens page and therefore only in `components/tokens`. It might be moved in the future.
* New i18n key for the Tokens page title.
* New `MainWrapper` in the `Tokens` page.
* Use the new i18n key for LayoutList in tokens layout.

# Tests

* Add test for tokens layout.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/components/tokens/MainWrapper.svelte
+++ b/frontend/src/lib/components/tokens/MainWrapper.svelte
@@ -1,0 +1,41 @@
+<script lang="ts">
+  export let testId: string;
+</script>
+
+<main data-tid={testId}>
+  <div class="content">
+    <slot />
+  </div>
+</main>
+
+<style lang="scss">
+  @use "@dfinity/gix-components/dist/styles/mixins/media";
+
+  // TODO: Confirm spacings https://dfinity.atlassian.net/browse/GIX-2005
+  main {
+    display: flex;
+    align-items: stretch;
+    justify-content: center;
+    height: 100%;
+    width: 100%;
+
+    padding: var(--padding-2x) var(--padding-2x) var(--padding)
+      var(--padding-2x);
+
+    @include media.min-width(large) {
+      padding: var(--padding-4x) 0 0 0;
+    }
+
+    & .content {
+      width: 100%;
+      height: 100%;
+
+      overflow-x: hidden;
+      overflow-y: auto;
+
+      @include media.min-width(large) {
+        width: var(--island-width);
+      }
+    }
+  }
+</style>
--- a/frontend/src/lib/pages/Tokens.svelte
+++ b/frontend/src/lib/pages/Tokens.svelte
@@ -1,2 +1,8 @@
+<script>
+  import MainWrapper from "$lib/components/tokens/MainWrapper.svelte";
+</script>
+
 <!-- TODO: GIX-1976 Render Tokens table with data -->
-<div data-tid="tokens-page-component">My Tokens</div>
+<MainWrapper testId="tokens-page-component">
+  <div>My Tokens</div>
+</MainWrapper>
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -1008,5 +1008,8 @@
     "status_idle_detailed": "Synchronization is currently idle.",
     "status_error_detailed": "An error occurred while synchronizing data. Please try reloading the page. If the problem persists, please contact support for further assistance.",
     "status_in_progress_detailed": "Checking for new transactions."
+  },
+  "tokens": {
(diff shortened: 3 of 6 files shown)


lmuntaner:
Layout:

Screenshot 2023-10-23 at 09 51 04
Screenshot 2023-10-23 at 09 51 10

lmuntaner:

frontend/src/lib/components/tokens/MainWrapper.svelte
```diff
@@ -0,0 +1,41 @@
...
+  export let testId: string;
+</script>
+
+<main data-tid={testId}>
+  <div class="content">
+    <slot />
+  </div>
+</main>
+
+<style lang="scss">
```
dskloetd, line 11: This is a lot of styling for 2 words.
Do you have another branch with more content where this looks good?
Or did this come directly from Figma?

frontend/src/lib/components/tokens/MainWrapper.svelte
```diff
@@ -0,0 +1,41 @@
...
+  @use "@dfinity/gix-components/dist/styles/mixins/media";
+
+  // TODO: Confirm spacings https://dfinity.atlassian.net/browse/GIX-2005
+  main {
+    display: flex;
+    align-items: stretch;
+    justify-content: center;
```
dskloetd, line 18: The screenshot you shared doesn't look centered. Does this do anything?

frontend/src/lib/pages/Tokens.svelte
```diff
@@ -1,2 +1,8 @@
+<script>
+  import MainWrapper from "$lib/components/tokens/MainWrapper.svelte";
+</script>
+
 <!-- TODO: GIX-1976 Render Tokens table with data -->
-<div data-tid="tokens-page-component">My Tokens</div>
+<MainWrapper testId="tokens-page-component">
+  <div>My Tokens</div>
```
dskloetd, line 7: Did you mean to use the i18n field or is this a placeholder?

dskloetd (approved):

frontend/src/lib/components/tokens/MainWrapper.svelte
lmuntaner: What do you mean a lot of styling for 2 words?

frontend/src/lib/components/tokens/MainWrapper.svelte
lmuntaner: The div with class content is centered within `main`.

I added on more screenshot in the PR to make it clear. Let me know if that's ok.

frontend/src/lib/pages/Tokens.svelte
lmuntaner: This is a placeholder.

frontend/src/lib/components/tokens/MainWrapper.svelte
dskloetd: I mean I only see "My Tokens" in the screenshot. I wouldn't expect 30 lines of styling until more content is added. Am I missing something?

frontend/src/lib/components/tokens/MainWrapper.svelte
dskloetd: Thanks. Makes sense.

dskloetd (approved):

frontend/src/lib/components/tokens/MainWrapper.svelte
lmuntaner: Haha, ok, I understand.

This is to place a div that we want to stretch until the bottom in the middle of the page area.

I agree it seems like a lot of lines, but some are to manage the responsiveness. Because for mobile, the two words are not placed the same way.

Some other rules might be set early (height, overflow), but we will need then when there are more tokens that the height of the screen. Not really tested in this PR, but I already know I'll need them.

This might become a shared component like `Island`, but I can't use Island in this case.

dskloetd: > but I already know I'll need them.

Thanks for clarifying.
That's what I had in mind with "Do you have another branch with more content where this looks good?"

lmuntaner: If I combine this and #3601 it will look as expected in Figma.

PR #3598, opened 2023-10-23
E2E: Wait for content loaded before getting account names

dskloetd:
# Motivation

We noticed some flakiness in the account e2e test when checking the account names after reloading the page.

# Changes

1. Await the reload and wait for the window load event.
2. Wait for the component content to be loaded before getting the account names.

# Tests

1. I applied these changes to branch `build/svelte-v4`
2. I added a loop around around lines 32-33 in `accounts.spec.ts`.
3. Without the changes, the loop would fail after a couple of times.
4. With the changes the loop would pass 20 iterations without failing.

# Todos

not necessary


peterpeterparker (approved):
Cool, thanks

PR #3608, opened 2023-10-23
fix: Join URLs with spaces

bitdivine:
# Motivation
NPM audit provides issues as links.  Joining links with commas makes them into a single, long URL to a non-existent location.

# Changes
- Join npm audit URLs with spaces

# Tests
None: It doesn't appear to be possible to run the npm audit step without running everything before it in the release command.

# Todos


dskloetd (approved):
thanks

PR #3609, opened 2023-10-24
Fix newlines

bitdivine:
# Motivation
This, taken from the last release, is meant to have line breaks:
```
Update the 'Playground' bookmark on the #eng-nns-dapp Slack channel to point to the new release candidate.\n\nSlack channel: https://dfinity.slack.com/channels/eng-nns-dapp\nNew Playground bookmark: https://eojib-pqaaa-aaaaa-aaepa-cai.nnsdapp.dfinity.network/
```

# Changes
- Add printf to the not-mocked commands; printf is better suited to multi-line printouts than echo.
- Adapt the relevant step to use printf rather than echo.
- Update the test.

# Tests
See the newlines in the change-detector test. 

# Todos

  - Too minor


dskloetd (approved):
thanks!

PR #3610, opened 2023-10-24
Delete the nightly II updates

bitdivine:
# Motivation
Currently the nightly job attempts to update II.  This is at odds with the way other updates are run, and doesn't allow the update job to be run on its own.

It is not clear to me that we need II in the NNS-dapp at all.  Do we need more than snsdemo provides?  Maybe just corresponding .did files?  If we can just delete, that would be great.  If not, let's move.

# Changes
- Delete the II update

# Tests
- No tests for no code.

# Todos


CHANGELOG-Nns-Dapp-unreleased.md
```diff
@@ -20,14 +20,14 @@ proposal is successful, the changes it released will be moved from this file to
 
 #### Changed
```
bitdivine, line 21: `spellcheck-changelog` failed, even though it appears to be in CI so should be clean on main.

lmuntaner (approved):

PR #3612, opened 2023-10-24
Chore: Refactor ProjectCard and ProjectCardSwapInfo tests

lmuntaner:
# Motivation

Improve testing.

In this PR, migrate the test files for ProjectCard and ProjectCardSwapInfo to use page objects.

# Changes

* Add a couple of data-tid to elements in ProjectCard and ProjectCardSwapInfo.
* Add methods to ProjectCardPo and ProjectCardSwapInfoPo.
* Add new fields to test helper `createMockSnsFullProject`.
* Rewrite tests that weren't using PO and the helper.

# Tests

Only test changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
+++ b/frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
@@ -18,11 +18,12 @@ import { render } from "@testing-library/svelte";
 
 describe("ProjectCard", () => {
   const rootCanisterId = rootCanisterIdMock;
-  const now = 1698139468065;
+  const now = 1698139468000;
   const nowInSeconds = Math.round(now / 1000);
   const yesterdayInSeconds = nowInSeconds - SECONDS_IN_DAY;
   beforeEach(() => {
     vi.clearAllMocks();
+    vi.useFakeTimers().setSystemTime(now);
     setSnsProjects([
       {
         rootCanisterId: mockSnsFullProject.rootCanisterId,
--- a/frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
+++ b/frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
@@ -18,9 +18,8 @@ import { render } from "@testing-library/svelte";
 
 describe("ProjectCard", () => {
   const rootCanisterId = rootCanisterIdMock;
-  const now = Date.now();
-  // Round up to the next second to avoid flaky tests
-  const nowInSeconds = Math.ceil(now / 1000);
+  const now = 1698139468065;
+  const nowInSeconds = Math.round(now / 1000);
   const yesterdayInSeconds = nowInSeconds - SECONDS_IN_DAY;
   beforeEach(() => {
     vi.clearAllMocks();
--- a/frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
+++ b/frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
@@ -1,13 +1,8 @@
 import * as saleApi from "$lib/api/sns-sale.api";
 import ProjectCard from "$lib/components/launchpad/ProjectCard.svelte";
 import { SECONDS_IN_DAY, SECONDS_IN_MONTH } from "$lib/constants/constants";
-import { authStore } from "$lib/stores/auth.store";
-import {
-  authStoreMock,
-  mockIdentity,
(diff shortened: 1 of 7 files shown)


lmuntaner:
@dskloetd please review

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
```diff
@@ -22,12 +17,10 @@ import { SnsSwapLifecycle } from "@dfinity/sns";
...
 
 describe("ProjectCard", () => {
-  vitest
-    .spyOn(authStore, "subscribe")
-    .mockImplementation(mutableMockAuthStoreSubscribe);
-
+  const rootCanisterId = rootCanisterIdMock;
   const now = Date.now();
-  const nowInSeconds = Math.floor(now / 1000);
+  // Round up to the next second to avoid flaky tests
```
dskloetd, line 22: How does a fractional second cause flaky tests?

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
```diff
@@ -22,12 +17,10 @@ import { SnsSwapLifecycle } from "@dfinity/sns";
 import { render } from "@testing-library/svelte";
 
 describe("ProjectCard", () => {
-  vitest
-    .spyOn(authStore, "subscribe")
-    .mockImplementation(mutableMockAuthStoreSubscribe);
-
+  const rootCanisterId = rootCanisterIdMock;
   const now = Date.now();
```
dskloetd, line 21: Is there any reason you need to use the actual current time?
Why not just use a hardcoded time?

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
```diff
@@ -166,13 +136,8 @@ describe("ProjectCard", () => {
...
 
-      const { container } = render(ProjectCard, {
-        props: {
-          project: snsFulProject,
-        },
-      });
+      const po = await renderCard(snsFulProject);
 
-      const po = ProjectCardPo.under(new JestPageObjectElement(container));
       await runResolvedPromises();
```
dskloetd, line 141: `renderCard` calls `runResolvedPromises();` so this is redundant now.

dskloetd (commented):
Thanks!

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: This was easier. Is there a reason to use a hardcoded time?

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: It was using `floor`, which means that the number of seconds could be rounded down. And when 1 day was added, it was missing that second and showing `32h and 59 minutes` instead of `1 day` for durations.

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: done, thanks!

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
dskloetd: Yes, unit tests should be deterministic.

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
dskloetd: So it was failing 99.9% of the time, and only passing when now() returned a whole number of seconds? Did you actually observe flakiness (i.e. sometimes passing and sometimes failing)?

The comment as is, doesn't really tell me the problem, so I would either just remove it, or really explain what it's about.

lmuntaner: I'll set a hardcoded now, it's a good reason to do so.

frontend/src/tests/lib/components/launchpad/ProjectCard.spec.ts
lmuntaner: and the comment shows it, good point.

lmuntaner:

dskloetd (approved):

PR #3613, opened 2023-10-24
Bump stable structures

bitdivine:
# Motivation
We would like to use the variable memory sizes introduced in `ic-stable-structures 0.6.0`

# Changes
- Update `stable-structures` to version `0.6.0`
- Update the code to match:
  - Instead of having a separate `BoundedStorable` trait, `stable-structures` now has a `BOUND` in the `Storable` trait.

# Tests
Existing tests should suffice.

# Todos

### DIFF
--- a/rs/backend/src/accounts_store/schema/accounts_in_stable_memory.rs
+++ b/rs/backend/src/accounts_store/schema/accounts_in_stable_memory.rs
@@ -25,7 +25,7 @@
 
 use crate::accounts_store::Account;
 use dfn_candid::Candid;
-use ic_stable_structures::storable::{BoundedStorable, Storable};
+use ic_stable_structures::storable::{Bound, Storable};
 use on_wire::{FromWire, IntoWire};
 use std::borrow::Cow;
 use std::convert::TryInto;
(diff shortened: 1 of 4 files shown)


lmuntaner (approved):
Rubber stamp! Thanks!

rs/backend/src/accounts_store/schema/accounts_in_stable_memory.rs
```diff
@@ -157,6 +157,10 @@ pub struct AccountStorageKey {
     bytes: [u8; AccountStorageKey::SIZE],
 }
 impl Storable for AccountStorageKey {
+    const BOUND: Bound = Bound::Bounded {
```
peterpeterparker, line 160: Out of curiosity, why not using `Unbounded`?

bitdivine: Unbounded is really new.  Planning to get all the other code ready, buying time to test unbounded and let other projects test it as well, then we can decide at the last moment which we trust more.

peterpeterparker: Isn't Unbounded already used on mainnet by some ckBTC canister?

PR #3614, opened 2023-10-24
Fix: Dissolve Delay Max

lmuntaner:
# Motivation

There is a bug raised in the [forum](https://forum.dfinity.org/t/changing-future-max-lock-up-for-sns-1-dkp/23576/34).

This screen has already been fixed with other bugs. The source of the bugs is the duality of seconds and days and synchronizing the slider with the input.

In this PR, we remove the slider and show a read-only progress bar instead.

We use days to be user friendly, but the code uses the values in seconds.

# Changes

* New component RangeDissolveDelay instead of the InputRange.
* Days are not managed in SetDissolveDelay, instead in the child components DayInput and RangeDissolveDelay.
* New props and functionality in DayInput, like managing Min and Max buttons.
* Change delayInSeconds of ConfirmDissolveDelay to `bigint` to avoid issues with decimals.
* Remove rounding in `secondsToDays`.

# Tests

* SetDissolveDelay.spec. Remove tests regarding slider and check progress bar instead.
* DayInput.spec. Add new test cases for new functionality.
* IncreaseDissolveDelayModal.spec. Use the input instead of range to set delay.
* NnsStakeNeuronModal.spec. Use the input instead of range to set delay.
* IncreaseSnsDissolveDelayModal.spec. Use the input instead of range to set delay. Remove test for slider.
* New PO RangeDissolveDelayPo
* Edit SetDissolveDelayPo accordingly

# Todos

### DIFF
--- a/frontend/src/lib/components/neurons/SetDissolveDelay.svelte
+++ b/frontend/src/lib/components/neurons/SetDissolveDelay.svelte
@@ -1,21 +1,15 @@
 <script lang="ts">
   import { createEventDispatcher } from "svelte";
   import { i18n } from "$lib/stores/i18n";
-  import {
-    daysToSeconds,
-    secondsToDays,
-    secondsToDaysRoundedDown,
-  } from "$lib/utils/date.utils";
   import { formatToken } from "$lib/utils/token.utils";
-  import { formatVotingPower } from "$lib/utils/neuron.utils";
   import { replacePlaceholders } from "$lib/utils/i18n.utils";
-  import { InputRange, Html } from "@dfinity/gix-components";
+  import { Html } from "@dfinity/gix-components";
   import { valueSpan } from "$lib/utils/utils";
   import NeuronStateRemainingTime from "$lib/components/neurons/NeuronStateRemainingTime.svelte";
   import DayInput from "$lib/components/ui/DayInput.svelte";
-  import { daysToDuration } from "$lib/utils/date.utils";
   import type { NeuronState } from "@dfinity/nns";
   import { type TokenAmount, nonNullish } from "@dfinity/utils";
+  import InputRangeDissolveDelay from "./InputRangeDissolveDelay.svelte";
 
   export let neuronState: NeuronState;
   export let neuronDissolveDelaySeconds: bigint;
@@ -29,68 +23,48 @@
 
   const dispatch = createEventDispatcher();
 
-  let delayInDays = secondsToDays(delayInSeconds);
-
-  $: delayInSeconds = daysToSeconds(delayInDays);
-
-  let neuronDelayInDays = secondsToDays(Number(neuronDissolveDelaySeconds));
-
-  let minDelayInDays = 0;
-  $: minDelayInDays = secondsToDays(Number(neuronDissolveDelaySeconds) + 1);
-
-  let minProjectDelayInDays = 0;
-  $: minProjectDelayInDays = secondsToDays(minProjectDelayInSeconds);
-
-  let maxDelayInDays = 0;
-  $: maxDelayInDays = secondsToDaysRoundedDown(maxDelayInSeconds);
-
   let votingPower: number;
   $: votingPower = calculateVotingPower(delayInSeconds);
 
-  let inputError: string | undefined;
-
   let disableUpdate: boolean;
-  $: disableUpdate = shouldUpdateBeDisabled(delayInDays);
+  $: disableUpdate = shouldUpdateBeDisabled(delayInSeconds);
 
   const keepDelaysInBounds = () => {
-    if (delayInDays < neuronDelayInDays) {
-      delayInDays = neuronDelayInDays;
+    if (delayInSeconds < Number(neuronDissolveDelaySeconds)) {
+      delayInSeconds = Number(neuronDissolveDelaySeconds);
     }
 
-    if (delayInDays > maxDelayInDays) {
-      delayInDays = maxDelayInDays;
+    if (delayInSeconds > maxDelayInSeconds) {
+      delayInSeconds = maxDelayInSeconds;
     }
   };
 
   const setMin = () => {
-    delayInDays = Math.max(minDelayInDays, minProjectDelayInDays);
-    updateInputError();
+    delayInSeconds = Math.max(
+      Number(neuronDissolveDelaySeconds),
+      minProjectDelayInSeconds
+    );
   };
 
   const setMax = () => {
-    delayInDays = maxDelayInDays;
-    updateInputError();
+    delayInSeconds = maxDelayInSeconds;
   };
 
-  const getInputError = (delayInDays: number) => {
-    if (delayInDays > maxDelayInDays) {
+  const getInputError = (delayInSeconds: number) => {
+    if (delayInSeconds > maxDelayInSeconds) {
       return $i18n.neurons.dissolve_delay_above_maximum;
     }
-    if (delayInDays < minDelayInDays) {
+    if (delayInSeconds < neuronDissolveDelaySeconds) {
       return $i18n.neurons.dissolve_delay_below_current;
     }
-    if (delayInDays < minProjectDelayInDays) {
+    if (delayInSeconds < minProjectDelayInSeconds) {
       return $i18n.neurons.dissolve_delay_below_minimum;
     }
     return undefined;
   };
 
-  const updateInputError = () => {
-    inputError = getInputError(delayInDays);
-  };
-
-  const shouldUpdateBeDisabled = (delayInDays: number): boolean => {
-    const error = getInputError(delayInDays);
+  const shouldUpdateBeDisabled = (delayInSeconds: number): boolean => {
+    const error = getInputError(delayInSeconds);
     // It's allowed to set the dissolve delay below the project minimum but we
     // still show a warning message to the user.
     return (
@@ -100,7 +74,6 @@
 
   const onRangeInput = () => {
     keepDelaysInBounds();
-    updateInputError();
   };
 
   const cancel = () => dispatch("nnsCancel");
@@ -146,40 +119,22 @@
     <p class="subtitle">{$i18n.neurons.dissolve_delay_label}</p>
     <div>
       <DayInput
-        bind:days={delayInDays}
+        bind:seconds={delayInSeconds}
+        maxInSeconds={maxDelayInSeconds}
         on:nnsMin={setMin}
         on:nnsMax={setMax}
-        on:nnsInput={updateInputError}
-        on:blur={updateInputError}
-        errorMessage={inputError}
         placeholderLabelKey="neurons.dissolve_delay_placeholder"
         name="dissolve_delay"
+        {getInputError}
       />
     </div>
     <div class="range">
-      <InputRange
-        ariaLabel={$i18n.neuron_detail.dissolve_delay_range}
-        min={0}
-        max={maxDelayInDays}
-        bind:value={delayInDays}
-        handleInput={onRangeInput}
+      <InputRangeDissolveDelay
+        {maxDelayInSeconds}
+        bind:delayInSeconds
+        {onRangeInput}
+        {votingPower}
       />
-      <div class="details">
-        <div>
-          <p class="label">
-            {formatVotingPower(votingPower)}
-          </p>
-          <p>{$i18n.neurons.voting_power}</p>
-        </div>
-        <div>
-          {#if delayInSeconds > 0}
-            <p class="label">{daysToDuration(delayInDays)}</p>
-          {:else}
-            <p class="label">{$i18n.neurons.no_delay}</p>
-          {/if}
-          <p>{$i18n.neurons.dissolve_delay_title}</p>
-        </div>
-      </div>
     </div>
   </div>
 
@@ -209,11 +164,5 @@
 
   .select-delay-container {
     width: 100%;
-
-    .details {
-      margin-top: var(--padding);
-      display: flex;
-      justify-content: space-around;
-    }
   }
 </style>
--- a/frontend/src/lib/components/neurons/SetDissolveDelay.svelte
+++ b/frontend/src/lib/components/neurons/SetDissolveDelay.svelte
@@ -29,32 +29,11 @@
   let disableUpdate: boolean;
   $: disableUpdate = shouldUpdateBeDisabled(delayInSeconds);
 
-  const keepDelaysInBounds = () => {
-    if (delayInSeconds < Number(neuronDissolveDelaySeconds)) {
-      delayInSeconds = Number(neuronDissolveDelaySeconds);
-    }
-
-    if (delayInSeconds > maxDelayInSeconds) {
-      delayInSeconds = maxDelayInSeconds;
-    }
-  };
-
-  const setMin = () => {
-    delayInSeconds = Math.max(
-      Number(neuronDissolveDelaySeconds),
-      minProjectDelayInSeconds
-    );
-  };
-
-  const setMax = () => {
-    delayInSeconds = maxDelayInSeconds;
-  };
-
   const getInputError = (delayInSeconds: number) => {
     if (delayInSeconds > maxDelayInSeconds) {
       return $i18n.neurons.dissolve_delay_above_maximum;
     }
-    if (delayInSeconds < neuronDissolveDelaySeconds) {
+    if (delayInSeconds <= neuronDissolveDelaySeconds) {
       return $i18n.neurons.dissolve_delay_below_current;
     }
     if (delayInSeconds < minProjectDelayInSeconds) {
@@ -70,10 +49,6 @@
     return (
       nonNullish(error) && error !== $i18n.neurons.dissolve_delay_below_minimum
     );
-  };
-
-  const onRangeInput = () => {
-    keepDelaysInBounds();
   };
 
   const cancel = () => dispatch("nnsCancel");
@@ -121,8 +96,10 @@
       <DayInput
         bind:seconds={delayInSeconds}
         maxInSeconds={maxDelayInSeconds}
-        on:nnsMin={setMin}
-        on:nnsMax={setMax}
+        minInSeconds={Math.max(
+          Number(neuronDissolveDelaySeconds),
+          minProjectDelayInSeconds
+        )}
         placeholderLabelKey="neurons.dissolve_delay_placeholder"
         name="dissolve_delay"
         {getInputError}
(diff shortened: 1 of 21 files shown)


lmuntaner:
New UI:

Playing with NNS dissolve delay
https://github.com/dfinity/nns-dapp/assets/4550653/b297640a-3cdf-4e4d-8df4-e59b111572ed

Showing the rounding up
https://github.com/dfinity/nns-dapp/assets/4550653/949c587c-e928-4cf3-a6c9-1c22d57bb6ff

Creating the state of the [issue from the forum](https://forum.dfinity.org/t/changing-future-max-lock-up-for-sns-1-dkp/23576/34) and showing how the issue is not there anymore.
https://github.com/dfinity/nns-dapp/assets/4550653/fdfe0c46-c463-4b13-8890-adee998970e7

lmuntaner:

dskloetd:
Can I try it somewhere?

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
```diff
@@ -146,40 +94,19 @@
...
-        errorMessage={inputError}
+        bind:seconds={delayInSeconds}
+        maxInSeconds={maxDelayInSeconds}
+        minInSeconds={Math.max(
+          Number(neuronDissolveDelaySeconds),
+          minProjectDelayInSeconds
+        )}
         placeholderLabelKey="neurons.dissolve_delay_placeholder"
         name="dissolve_delay"
+        {getInputError}
```
dskloetd, line 105: Why do we pass the function instead of the message?

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
```diff
@@ -25,7 +25,7 @@
     Number(
       neuronVotingPower({
         neuron,
-        newDissolveDelayInSeconds: BigInt(delayInSeconds),
+        newDissolveDelayInSeconds: BigInt(Math.round(delayInSeconds)),
```
dskloetd, line 28: Why does it need to be rounded now but not before?

frontend/src/lib/components/ui/DayInput.svelte
```diff
@@ -1,29 +1,46 @@
...
-  const dispatch = createEventDispatcher();
-  const setMin = () => dispatch("nnsMin");
-  const setMax = () => dispatch("nnsMax");
+  // Round up the first time to not show a lot of decimal places.
+  let days: number = Math.min(Math.ceil(secondsToDays(seconds)), maxInSeconds);
+  $: seconds = daysToSeconds(days);
+
+  let errorMessage: string | undefined;
+  const showError = () => {
+    // This is called with before `$: seconds = daysToSeconds(days);`
```
dskloetd, line 20: I don't understand this sentence.

frontend/src/lib/components/ui/DayInput.svelte
```diff
@@ -1,29 +1,46 @@
...
+  export let minInSeconds: number;
   export let placeholderLabelKey = "core.amount";
   export let name = "amount";
+  export let getInputError: (value: number) => string | undefined;
 
-  const dispatch = createEventDispatcher();
-  const setMin = () => dispatch("nnsMin");
-  const setMax = () => dispatch("nnsMax");
+  // Round up the first time to not show a lot of decimal places.
+  let days: number = Math.min(Math.ceil(secondsToDays(seconds)), maxInSeconds);
```
dskloetd, line 15: The `Math.min` is a good catch! Is there a test for it?

frontend/src/lib/components/ui/DayInput.svelte
```diff
@@ -1,29 +1,46 @@
...
   {placeholderLabelKey}
   {name}
   bind:value={days}
-  {max}
+  max={secondsToDays(maxInSeconds)}
   inputType="number"
   {errorMessage}
-  on:nnsInput
-  on:blur
+  on:blur={showError}
```
dskloetd, line 42: Do we need to update the error on blur if we already update it on input?

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
```diff
@@ -57,15 +57,14 @@ describe("SetDissolveDelay", () => {
...
-  it("should initialize text and slider correctly", async () => {
+  it("should initialize text", async () => {
     const neuronDissolveDelaySeconds = 90 * SECONDS_IN_DAY;
     const po = renderComponent({
       neuronDissolveDelaySeconds: BigInt(neuronDissolveDelaySeconds),
       delayInSeconds: neuronDissolveDelaySeconds,
     });
 
     expect(await po.getDays()).toBe(90);
-    expect(await po.getSliderDays()).toBe(90);
```
dskloetd, line 68: Why not still test the value of the progress bar?

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
```diff
@@ -77,31 +76,18 @@ describe("SetDissolveDelay", () => {
     });
 
     expect(await po.getDays()).toBe(91);
-    expect(await po.getSliderDays()).toBe(91);
```
dskloetd, line 80: Can we keep an equivalent expectation?

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
```diff
@@ -77,31 +76,18 @@ describe("SetDissolveDelay", () => {
...
   });
 
   it("should update slider on text input", async () => {
     const po = renderComponent();
     await po.enterDays(1);
-    expect(await po.getSliderDays()).toBe(1);
+    expect(await po.getProgressBarSeconds()).toBe(1 * SECONDS_IN_DAY);
```
dskloetd, line 84: For convenience we can continue to have a method that returns days, no?

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
dskloetd, line 1: Can you add some tests with fractional day input?

frontend/src/tests/lib/components/ui/DayInput.spec.ts
```diff
@@ -1,35 +1,131 @@
...
+      target: { value: days },
+    });
+
+    await tick();
+    expect(queryByTestId("seconds")?.textContent).toBe(
+      String(SECONDS_IN_DAY * days)
+    );
+  });
+
+  it("shuold update seconds and input on Min/Max click", async () => {
```
dskloetd, line 93: typo: shuold

frontend/src/tests/lib/utils/date.utils.spec.ts
```diff
@@ -190,6 +190,11 @@ describe("daysToDuration", () => {
       });
     }
   });
+
+  it("should return fractions of day", () => {
+    expect(daysToDuration(1.5)).toBe("");
```
dskloetd, line 195: Why is this `""`?

frontend/src/tests/lib/utils/date.utils.spec.ts
```diff
@@ -190,6 +190,11 @@ describe("daysToDuration", () => {
...
+    expect(daysToDuration(365.125)).toBe("1 year");
```
dskloetd, line 196: Shouldn't this be "1 year, 3 hours" or something?

frontend/src/tests/page-objects/RangeDissolveDelay.page-object.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { PageObjectElement } from "$tests/types/page-object.types";
+
+export class RangeDissolveDelayPo extends BasePageObject {
+  private static readonly TID = "range-dissolve-delay-component";
+
+  static under(element: PageObjectElement): RangeDissolveDelayPo {
+    return new RangeDissolveDelayPo(element.byTestId(RangeDissolveDelayPo.TID));
+  }
+
+  async getProgressBarValue(): Promise<number> {
```
dskloetd, line 11: `RangeDissolveDelay` is already specifically about seconds, right? Shall we reflect that in the name here as well?

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
lmuntaner: I use the `on:blur` and `on:nnsInput` to show the error.

Those are called before the seconds are updated, and `DayInput` doesn't expose the value of the input `days`.

```
  const showError = () => {
    // This is called with before `$: seconds = daysToSeconds(days);`
    errorMessage = getInputError(daysToSeconds(days));
  };
```

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
lmuntaner: Before it couldn't be a decimal because from integer days to seconds, it was always an integer.

But now that we allow fractions of days, this might be also a fraction.

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
lmuntaner: We have a few test with fractional:

* `"fractional days get rounded up"`
* `"can set same number of days when current number of days is fractional"`

I added on expect in `"when text input below or equal to current delay"`

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
lmuntaner: Yes, but I find that convenience in expense of understanding what it is.

Apart from convenience, is there a reason to use that? Then we might have to deal again with decimals...

frontend/src/tests/page-objects/RangeDissolveDelay.page-object.ts
lmuntaner: No, only the progress bar is. Below the progress bar the voting power and the delay in a human friendly format are shown.

frontend/src/lib/components/ui/DayInput.svelte
lmuntaner: Added!

frontend/src/lib/components/ui/DayInput.svelte
lmuntaner: It doesn't look like it. I left it because it was there before. Removed.

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
lmuntaner: done

frontend/src/tests/lib/components/ui/DayInput.spec.ts
lmuntaner: done

frontend/src/tests/lib/utils/date.utils.spec.ts
lmuntaner: fixed

frontend/src/tests/lib/utils/date.utils.spec.ts
lmuntaner: fixed

frontend/src/lib/components/ui/DayInput.svelte
dskloetd: Oh, I didn't realize it was there before because it was in another file.
In that case, feel free to keep it.

frontend/src/lib/components/ui/DayInput.svelte
```diff
@@ -12,12 +12,16 @@
...
+  let days: number = Math.min(
+    Math.ceil(secondsToDays(seconds)),
+    secondsToDays(maxInSeconds)
+  );
   $: seconds = daysToSeconds(days);
 
   let errorMessage: string | undefined;
   const showError = () => {
-    // This is called with before `$: seconds = daysToSeconds(days);`
+    // This is called with before we update the `seconds` variable
```
dskloetd, line 23: I don't understand what this is saying.

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
lmuntaner: I would prefer DayInput to not expose two units of the same value. That can lead to errors. It's already hard to keep in sync two values of different units within one component. Exposing them it will increase the possibility of errors.

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
lmuntaner: Mhm, good idea, maybe it can expose `bigint` instead of `number`. But if it has to expose `number`, I won't be able to type safe that it's always a whole number. Let me try.

lmuntaner: Ok, I realized I would have to change a lot more things to try that. `SetDissolveDelay` uses `number` as props. Therefore, I would have to change also the usages of `SetDissolveDelay`.

Are you ok with adding much more changes here, or shall I try that in another PR?

frontend/src/tests/page-objects/RangeDissolveDelay.page-object.ts
lmuntaner: Done

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
lmuntaner: Yes, that's true.

Added one to check that the user can enter fractional values, and added a loop of tests for pairs of min/max values.

frontend/src/lib/components/ui/DayInput.svelte
lmuntaner: reverted

lmuntaner:

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
dskloetd: Let's keep it `number`, but still make it an integer (non-fraction) number.
(You can make it a `bigint` later if you want, or not.)

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
dskloetd: DayInput doesn't expose days and I'm not asking for it to expose days.
I don't see what this has to do with passing the error message vs. passing the error function.

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
```diff
@@ -212,45 +227,53 @@ describe("SetDissolveDelay", () => {
...
-    });
-    const po = SetDissolveDelayPo.under(new JestPageObjectElement(container));
+  const minMaxDaysPairs = [
+    { min: 50, max: 500 },
+    { min: 50.5, max: 500.5 },
+    { min: 50, max: 500.2 },
+    { min: 50.1, max: 500 },
+  ];
+  for (const { min, max } of minMaxDaysPairs) {
+    it("should update prop, text, progress bar on Min/Max", async () => {
```
dskloetd, line 237: Can you put the parameters in the test name?

frontend/src/lib/components/neurons/SetNnsDissolveDelay.svelte
lmuntaner: Done. I added a test also for daysToSeconds to make sure it returns integers.

But I'd rather not remove this code here. There is no type assurance and when converting to bigint, it makes sense to round always.

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
lmuntaner: done

lmuntaner:
@dskloetd please review

frontend/src/tests/lib/utils/date.utils.spec.ts
```diff
@@ -316,3 +317,16 @@ describe("secondsToTime", () => {
...
+
+describe("daysToSeconds", () => {
+  it("returns the days in seconds", () => {
+    expect(daysToSeconds(1)).toBe(SECONDS_IN_DAY);
+    expect(daysToSeconds(2)).toBe(SECONDS_IN_DAY * 2);
+    expect(daysToSeconds(3)).toBe(SECONDS_IN_DAY * 3);
+  });
+
+  it("returns integers only", () => {
+    expect(daysToSeconds(1.11)).not.toBe(SECONDS_IN_DAY * 1.11);
```
dskloetd, line 329: Does this work? `SECONDS_IN_DAY * 1.11` is actually an integer because `SECONDS_IN_DAY` is a multiple of 100.
I think you need some more decimals for it to work, or at least you need more decimals for the test not to be very confusing.

lmuntaner: done!

frontend/src/lib/components/neurons/SetDissolveDelay.svelte
lmuntaner: I tried, the issue was the first error message.

We don't want to trigger an error message until the input changes. For that, we use `on:nnsInput` instead of the reactive `$: errorMessage = ...`.

I could expose `on:nnsInput` on `DayInput`, but that is triggered no days changing, not on seconds. I would need to convert from days to seconds in SetDissolveDelay. Which is what I want to avoid.

I can't use `on:nnsInput` with the exposed `seconds`, because `on:nnsInput` is triggered when days change, but before seconds change.

lmuntaner: Does that help understand the issue?

For example, if we were ok showing the error on the first show, then we could use `$: errorMessage = getError(...)` on SetDissolveDelay and pass the error message to DayInput.

But that would trigger an error message when the current dissolve delay is a whole amount of days.

Like a test well caught when I tried to move `errorMessage` to SetDissolveDelay. I had forgotten exactly why I did it like that.

I will add a comment.

dskloetd: Thanks!

dskloetd (approved):

frontend/src/lib/components/ui/DayInput.svelte
lmuntaner: I added a comment below to better explain it. Is still not clear?

dskloetd: I can't even parse this as a sentence. I stumble over "with before".

lmuntaner: oh, that's a typo, sorry

PR #3618, opened 2023-10-25
Remove debounce on deriving network from BTC address

dskloetd:
# Motivation

When sending ckBTC you can enter either a ckBTC address or a BTC address.
We derive the "network" (Internet Computer/Bitcoin) from the address.
This is debounced so there is a small delay.
The derivation does not require any network requests, and once the network is derived we don't change it (if you later change to an address to a different network, we consider it invalid unless you change the network manually), so this delay does not prevent any jarring experience.
So I see no reason to have the debounce there.
Additionally it makes unit testing slightly harder so there is a benefit in removing the debounce.

# Changes

Remove the debounce on deriving the network from the address.

# Tests

Removing it didn't cause any tests to fail so it wasn't tested before.
Manual testing looks good and can be tried here: 

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp-unreleased.md
+++ b/CHANGELOG-Nns-Dapp-unreleased.md
@@ -27,6 +27,8 @@ proposal is successful, the changes it released will be moved from this file to
 #### Deprecated
 #### Removed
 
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Sounds good to me

PR #3620, opened 2023-10-25
Use page objects in CkBTCWallet.spec.ts

dskloetd:
# Motivation

Making the test more readable.

# Changes

1. Use page objects in `CkBTCWallet.spec.ts`.
2. Add test IDs where necessary.
3. Add necessary methods to exiting page objects.
4. Add new page objects.
5. In `frontend/src/tests/page-objects/SelectDestinationAddress.page-object.ts` check whether the dropdown input is present before trying to check the state of the checkbox. In some contexts there is no checkbox and if the input is already the type of input we need (dropdown vs. text), this is fine.
6. Add missing types to `frontend/src/tests/page-objects/WalletPageHeader.page-object.ts`
7. Add `innerHtmlForDebugging` to PageObjectElement to help probe the state of an element when debugging.

# Tests

yes

# Todos

not necessary

### DIFF
--- a/frontend/src/tests/lib/pages/CkBTCWallet.spec.ts
+++ b/frontend/src/tests/lib/pages/CkBTCWallet.spec.ts
@@ -11,20 +11,20 @@ import * as transactionsServices from "$lib/services/ckbtc-transactions.services
 import { authStore } from "$lib/stores/auth.store";
 import { icrcAccountsStore } from "$lib/stores/icrc-accounts.store";
 import { tokensStore } from "$lib/stores/tokens.store";
-import { TransactionNetwork } from "$lib/types/transaction";
-import { formatToken } from "$lib/utils/token.utils";
 import { page } from "$mocks/$app/stores";
 import CkBTCAccountsTest from "$tests/lib/components/accounts/CkBTCAccountsTest.svelte";
 import { mockAuthStoreSubscribe } from "$tests/mocks/auth.store.mock";
-import {
-  mockCkBTCMainAccount,
-  mockCkBTCToken,
-} from "$tests/mocks/ckbtc-accounts.mock";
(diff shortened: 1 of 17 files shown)


frontend/src/tests/lib/pages/CkBTCWallet.spec.ts
```diff
@@ -171,135 +199,79 @@ describe("CkBTCWallet", () => {
...
-        expect(
-          getByTestId("universe-page-summary-component").textContent.trim()
-        ).toBe("ckTESTBTC")
-      );
+      expect(await po.getWalletPageHeaderPo().getUniverse()).toBe("ckTESTBTC");
     });
 
     it("should hide spinner when selected account is loaded", async () => {
-      const { queryByTestId } = render(CkBTCWallet, props);
+      const po = await renderWallet();
```
lmuntaner, line 208: How do I know that the selected account is loaded?

Does it make sense to merge this and `"should render a spinner while loading"` to show spinner before `runResolvedPromises` and no afterwards?

lmuntaner (approved):
LGTM!

Just a comment, but already approving.

frontend/src/tests/lib/pages/CkBTCWallet.spec.ts
dskloetd: It's a good question.
I just changed the tests to use page objects without paying much attention to what they do.
But the test with spinner is actually in a different group, without accounts.
In this group, the accounts are already loaded.
Both tests implicitly do `runResolvedPromises` as part of `renderWallet`.

I do want to look at also mocking the API layer instead of the services. Maybe that would make it easier to do what you suggest. But in this PR I'd like to stick to just using page objects.

Also you should never explicitly expect a different state before `runResolvedPromises` than after because whatever happens during `runResolvedPromises` can easily also happen earlier already so such a test would be very fragile.

PR #3625, opened 2023-10-26
Keep a list of exceptional transactions

bitdivine:
# Motivation
Some special, zero-value, transaction types currently fail to parse.  We would like visibility of which transactions fail to parse.

# Changes
* Keep a bounded vector of exceptional transaction IDs.

# Tests
There is a unit test that verifies that the list does not exceed the limit and that it retains the most recent transactions.

# Todos

### DIFF
--- a/rs/backend/src/ledger_sync.rs
+++ b/rs/backend/src/ledger_sync.rs
@@ -104,6 +104,11 @@ async fn get_blocks(from: BlockIndex, tip_of_chain: BlockIndex) -> Result<Vec<(B
                             &dummy,
                             err
                         );
+                        STATE.with(|s| {
+                            s.performance
+                                .borrow_mut()
+                                .record_exceptional_transaction_id(range.start() + (index as u64))
+                        });
                         dummy
                     }
                 },
--- a/rs/backend/src/perf.rs
+++ b/rs/backend/src/perf.rs
@@ -34,6 +34,7 @@ impl PerformanceCount {
 #[derive(CandidType, Deserialize, Clone, Debug, Default, Eq, PartialEq)]
 pub struct PerformanceCounts {
     pub instruction_counts: VecDeque<PerformanceCount>,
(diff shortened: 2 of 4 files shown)


rs/backend/src/perf.rs
```diff
@@ -49,6 +50,21 @@ impl PerformanceCounts {
...
+    /// Saves an exceptional transaction ID
+    pub fn record_exceptional_transaction_id(&mut self, transaction_id: u64) {
+        if self.exceptional_transactions.is_none() {
+            self.exceptional_transactions = Some(VecDeque::new());
+        }
+        if let Some(exceptional_transactions) = &mut self.exceptional_transactions {
+            while exceptional_transactions.len() >= Self::MAX_EXCEPTIONAL_TRANSACTIONS {
+                exceptional_transactions.pop_front();
+            }
+            exceptional_transactions.push_back(transaction_id);
```
bitdivine, line 64: TODO: Are transaction IDs parsed only once?  Do we need logic to ensure that there are no dupes?  Can dupes include entries in the middle of the vector?

rs/backend/src/ledger_sync.rs
```diff
@@ -104,6 +104,11 @@ async fn get_blocks(from: BlockIndex, tip_of_chain: BlockIndex) -> Result<Vec<(B
                             &dummy,
                             err
                         );
+                        STATE.with(|s| {
```
bitdivine, line 107: TODO: Is STATE used by the caller?  If there is a risk of this, we may need to store the list separately.

bitdivine: No, STATE is not used by the caller so this should be fine but it would still be safer to pass in STATE explicitly rather than acquiring the lock in this code.

MarioDfinity (approved):

dskloetd (approved):

PR #3629, opened 2023-10-26
Add NNS ICP ledger candid to declarations directory

bitdivine:
# Motivation
We would like to make calls to the NNS ICP ledger for development and testing purposes.

# Changes
* Add the NNS Ledger candid file to the `declarations` directory.
* Update related code.

# Tests
- CI checks that the candid files imported from the IC repo match the commit in `dfx.json`.

# Todos

- [x] Add entry to changelog (if necessary).

PR #3630, opened 2023-10-26
Include exceptional transactions in stats

bitdivine:
# Motivation
We would like to have visibility of the number of exceptional transactions and potentially alerts for their number.

# Changes
Provide a metric with a count of the number of stored exceptional transactions.

Note: If the exceptional transactions log ever becomes full, this metric will plateau.  If that happens before we switch to the index canister we may need revise this statistic, e.g. with a dedicated counter.

# Tests
Awaiting #3629 to write an e2e test.

# Todos

  - Covered by an existing changelog entry


lmuntaner (approved):
Thanks!

PR #3634, opened 2023-10-27
Use transaction type Transfer for Send/Receive derived from TransferFrom

dskloetd:
# Motivation

[Here](https://github.com/dfinity/nns-dapp/blob/5642420e1cadc0fe4c80c11ed2be9d34293c8630/rs/backend/src/accounts_store.rs#L794) we treat TransferFrom the same as Transfer and turn into a `Send` or a `Receive`. But [here](https://github.com/dfinity/nns-dapp/blob/5642420e1cadc0fe4c80c11ed2be9d34293c8630/rs/backend/src/accounts_store.rs#L821) we pass the transaction type which might be TransferFrom . But [in the frontend](https://github.com/dfinity/nns-dapp/blob/5642420e1cadc0fe4c80c11ed2be9d34293c8630/frontend/src/lib/canisters/nns-dapp/nns-dapp.idl.js#L87), we don't know TransactionType TransferFrom and thus it fails to decode the transaction.

# Changes

Use transaction type `Transfer` on the `Send` or `Receive` created from a `TransferFrom`.

# Tests

Added a test with a `TransferFrom` and check that it get transaction type `Transfer`.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp-unreleased.md
+++ b/CHANGELOG-Nns-Dapp-unreleased.md
@@ -34,6 +34,7 @@ proposal is successful, the changes it released will be moved from this file to
 
 * Remove robots meta tag to allow search engines to crawl NNS Dapp.
 * Fix i18n key in merge neurons summary screen.
+* Display `TransferFrom` as a normal receive instead of failing to load transactions.
 
 #### Security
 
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
Thanks!

PR #3637, opened 2023-10-30
Update E2E tests to new version of Internet Identity

dskloetd:
# Motivation

The current used version of Internet Identity is from April. It has a bug where it tries to connect to mainnet when it doesn't recognize the FE URL. This prevents us from using it on a DevEnv.

# Changes

1. Use the latest snsdemo snapshot which uses version 2023-10-27 of Internet Identity.
2. Update `signInWithNewUser` to the new ii UI.

# Tests

Pass on CI

# Todos


mstrasinskis (approved):
Thanks!

PR #3638, opened 2023-10-30
Mock API layer in CkBTCWallet.spec.ts

dskloetd:
# Motivation

It's easier to refactor code if the test mocks lower layers because we don't care which service calls the API.

# Changes

1. Stop mocking several services in `CkBTCWallet.spec.ts`
2. Instead mock the necessary APIs and block non-mocked calls to those APIs
3. In `beforeEach`, reset stores that are filled by the service layer
4. Use `resetIdentity` instead of `mockAuthStoreSubscribe`
5. Merge the tests that the spinner is visible/not visible into one test where the spinner starts visible and then goes away
6. Fix an error message about using `jest.mock` to say `vi.mock` instead

# Tests

Pass

# Todos

not necessary


mstrasinskis (approved):
Nice, thanks!

PR #3639, opened 2023-10-31
Payload redesign

mstrasinskis:
# Motivation

Redesign of proposal payload and action block.

# Changes

- toggle to switch between tree and raw json view
- json as tree preview
- raw json preview

# Tests
Current tests are not broken. Tests for the new components moved to [a separate task](https://dfinity.atlassian.net/browse/GIX-2030).

# Todos

# Screenshots

| Tree | Raw |
|--------|--------|
| image | image |


frontend/src/lib/components/common/JsonPreview.svelte
```diff
@@ -0,0 +1,79 @@
...
+  import TreeJson from "$lib/components/common/TreeJson.svelte";
+  import RawJson from "$lib/components/common/RawJson.svelte";
+  import { fade } from "svelte/transition";
+
+  export let json: unknown | undefined = undefined;
+
+  let expandedData: unknown;
+  $: expandedData = isNullish(json)
+    ? json
+    : (expandObject(json as Record<string, unknown>) as object);
```
peterpeterparker, line 16: lint is not complaining about the use of `object`?

frontend/src/lib/components/common/JsonPreview.svelte
```diff
@@ -0,0 +1,79 @@
...
+
+  let expandAll: boolean = false;
```
peterpeterparker, line 18: lint will probably complains about `let expandAll: boolean = false;` to have it as `let expandAll = false;`

frontend/src/lib/components/common/JsonPreview.svelte
```diff
@@ -0,0 +1,79 @@
...
+  $: expandAll = getObjMaxDepth(expandedData) < 2 ? true : expandAll;
```
peterpeterparker, line 19: worth to add a comment about `2` or extract a local const?

frontend/src/lib/components/common/TreeJson.svelte
```diff
@@ -0,0 +1,162 @@
+<script lang="ts">
+  import { i18n } from "$lib/stores/i18n";
+  import { IconExpandMore } from "@dfinity/gix-components";
+  import TreeJsonValue from "$lib/components/common/TreeJsonValue.svelte";
+  import { getTreeJsonValueType } from "$lib/utils/json.utils";
+  import { fade } from "svelte/transition";
+
+  export let json: unknown | undefined = undefined;
+  export let defaultExpandedLevel = Infinity;
+  export let _key: string | undefined = undefined;
```
peterpeterparker, line 10: out of curiosity, why are those props prefixed with `_`?

often prefixing with `_` means "this is not used"

frontend/src/lib/components/common/TreeJsonValue.svelte
```diff
@@ -0,0 +1,77 @@
...
+      : stringifyJson(data);
+
+  let title: string | undefined;
+  $: title = valueType === "hash" ? (data as number[]).join() : undefined;
+</script>
+
+{#if valueType === "base64Encoding"}
+  <!-- base64 encoded image (use <Html> to sanitize the content from XSS) -->
+  <Html
+    text={`<img class="value ${valueType}" alt="${key}" src="${value}" />`}
```
peterpeterparker, line 26: `loading="lazy"`?

frontend/src/lib/components/proposal-detail/ProposalProposerActionsEntry.svelte
```diff
@@ -1,33 +1,66 @@
...
 >
-  <h2
-    class="content-cell-title"
-    data-tid="proposal-proposer-actions-entry-title"
-  >
-    {actionKey ?? ""}
-  </h2>
+  {#if $ENABLE_FULL_WIDTH_PROPOSAL}
+    <div class="header">
+      <h2
```
peterpeterparker, line 22: semantically, shouldn't it be a div with a h2 and copy inside - i.e. copy should not be part of the h2?

frontend/src/lib/stores/json-representation-mode.store.ts
```diff
@@ -0,0 +1,13 @@
+import { StoreLocalStorageKey } from "$lib/constants/stores.constants";
+import { writableStored } from "./writable-stored";
+
+type JsonRepresentationMode = "tree" | "raw";
```
peterpeterparker, line 4: should we use an object that contains the mode rather than a string?

thinking about future improvements and migrations, in case we add more options that should be persisted in the local storage.

frontend/src/lib/utils/json.utils.ts
```diff
@@ -0,0 +1,28 @@
+import { isHash, isPrincipal } from "$lib/utils/utils";
+
+export type TreeJsonValueType =
+  | "bigint"
+  | "boolean"
+  | "function"
+  | "null"
```
peterpeterparker, line 7: should we used a keyword lile "nullish" instead of "null" to avoid unexpected issue with a string that would be really null but not "null" at runtime?

frontend/src/lib/utils/json.utils.ts
```diff
@@ -0,0 +1,28 @@
...
+  | "number"
+  | "object"
+  | "principal"
+  | "hash"
+  | "string"
+  | "symbol"
+  | "base64Encoding"
+  | "undefined";
```
peterpeterparker, line 15: same as for my above comment about null

frontend/src/lib/utils/json.utils.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const getTreeJsonValueType = (value: unknown): TreeJsonValueType => {
```
peterpeterparker, line 17: took me a bit to understand what's the goal of this function, now it's pretty clear but, I did not get it instantly. So not sure if there would be a better function's name that would make more obvious what it does

frontend/src/lib/components/common/TreeJson.svelte
```diff
@@ -0,0 +1,162 @@
...
+      </button>
+      {keyLabel}
+    </div>
+  {/if}
+  {#if !collapsed}
+    <!-- children of expandable-key -->
+    <ul class:root class:is-array={isArray} in:fade>
+      {#each children as [key, value]}
+        <li class:root>
+          <svelte:self
```
peterpeterparker, line 70: nice!!!

frontend/src/lib/components/proposal-detail/JsonRepresentationModeToggle.svelte
```diff
@@ -0,0 +1,61 @@
+<script lang="ts">
+  import { Toggle } from "@dfinity/gix-components";
+  import { i18n } from "$lib/stores/i18n";
+  import { jsonRepresentationModeStore } from "$lib/stores/json-representation-mode.store";
+
+  let checked: boolean;
+  $: checked = $jsonRepresentationModeStore === "raw";
+
+  const setTree = () => ($jsonRepresentationModeStore = "tree");
+  const setRaw = () => ($jsonRepresentationModeStore = "raw");
```
peterpeterparker, line 10: should we write `jsonRepresentationModeStore.set("raw")` for consistency reason? if I am not wrong we mostly use `set`

frontend/src/lib/components/proposal-detail/JsonRepresentationModeToggle.svelte
```diff
@@ -0,0 +1,61 @@
...
+  const toggle = () =>
+    $jsonRepresentationModeStore === "tree" ? setRaw() : setTree();
+</script>
+
+<div class="data-representation-toggle" data-tid="data-representation-toggle">
+  <div class="toggle">
```
peterpeterparker, line 16: you need that div?

peterpeterparker (approved):
LGTM, minor comments

frontend/src/lib/components/common/JsonPreview.svelte
mstrasinskis: No complains, but `as object` is definitely redundant here.

frontend/src/lib/components/common/JsonPreview.svelte
mstrasinskis: Same here. No complains but agree and updated 👍

frontend/src/lib/components/common/JsonPreview.svelte
mstrasinskis: Good point, done.

frontend/src/lib/components/common/TreeJson.svelte
mstrasinskis: Because of recursive usage the `private` proposals are prefixed with `_`.

frontend/src/lib/components/common/TreeJsonValue.svelte
mstrasinskis: Currently only `base64` images are supported here, but let's add it for the future proof.

frontend/src/lib/components/proposal-detail/ProposalProposerActionsEntry.svelte
mstrasinskis: It's done for tags economy, but not a strong opinion. Let's have a wrapper here.

frontend/src/lib/stores/json-representation-mode.store.ts
mstrasinskis: Good point. Updating.

frontend/src/lib/utils/json.utils.ts
mstrasinskis: @peterpeterparker, here "null" means only null `if (value === null) return "null";` Or you do you mean something else?

peterpeterparker: @mstrasinskis I mean that I am afraid to use a string `"null"` because I don't trust bundler and javascript enough so was just suggesting to use another value, like `"nullyolo"` or `"null_"` or `"n_U-l$l"` or anything that is just not only "null"

peterpeterparker: Probably more an irrational  based guts feeling suggestion than something that make technically sense

frontend/src/lib/utils/json.utils.ts
mstrasinskis: Renamed + comments

frontend/src/lib/components/proposal-detail/JsonRepresentationModeToggle.svelte
mstrasinskis: @peterpeterparker was updated, could you please re-review store changes?

peterpeterparker: Looks cool 👍

frontend/src/lib/components/proposal-detail/JsonRepresentationModeToggle.svelte
mstrasinskis: removed 👍

frontend/src/lib/utils/json.utils.ts
mstrasinskis: Let's keep the current version for now, since this is a copy/paste from the sibling component that will be deprecated soon but I'll keep it in mind, thanks 👍
image

PR #3641, opened 2023-10-31
update: Use checksummed schema

bitdivine:
# Motivation
Context: According to the schema migration plan, the schema should be recorded in the first bytes of stable memory.  The schema version is checksummed, so if there is no valid checksum we assume that we have stable memory according to the old format that is `<len:u32>||<data serialized as candid>`.

Currently this is not implemented:  The placeholder function that gets the schema currently always returns None without looking at the stable memory.

# Changes
- Read the schema version from stable memory.

# Tests
- The downgrade upgrade test will verify that the `None` schema is still recognized correctly.
- Other serializations are not defined yet.

# Todos


rs/backend/src/state/tests.rs
```diff
@@ -25,3 +28,20 @@ fn state_can_be_restored_from_stable_memory() {
     assert_eq!(toy_state.assets, parsed.assets, "Assets have changed");
     assert_eq!(toy_state.asset_hashes, parsed.asset_hashes, "Asset hashes have changed");
 }
+
+#[test]
+fn schema_can_be_read_from_memory() {
+    let raw_memory = vec![];
+    let memory: VectorMemory = Rc::new(RefCell::new(raw_memory));
```
frederikrothenberger, line 35: You could you `VectorMemory::default()` here.

rs/backend/src/state.rs
```diff
@@ -63,9 +65,21 @@ impl StableState for State {
...
     /// The schema version, determined by the last version that was saved to stable memory.
-    fn schema_version_from_stable_memory() -> Option<u32> {
-        None // The schema is currently unversioned.
+    fn schema_version_from_stable_memory() -> Option<SchemaLabel> {
+        let memory = DefaultMemoryImpl::default();
+        Self::schema_version_from_memory(memory)
     }
+
+    /// The schema version, as stored in an arbitrary memory.
+    fn schema_version_from_memory<M>(memory: M) -> Option<SchemaLabel>
```
frederikrothenberger, line 74: Reading the schema from memory should probably not consume the memory?
It would also save you a clone in the test if you only took a reference here.

frederikrothenberger (approved):
LGTM!

rs/backend/src/state.rs
bitdivine: It clones just the reference though, not the whole memory.

bitdivine: It seems a bit strange to pass a reference of a reference.  Are there any advantages?  I guess that with Rc there is a potential for deadlocks, so if code has once acquired the lock it's worth keeping it.  Not really relevant in this case but it wouldn't hurt either.  Ok, fair enough.

rs/backend/src/state/tests.rs
bitdivine: True.  When testing upgrades it is useful to have the raw memory and then write according to partition schema A and then read according to partitioning scheme B but here that is not needed.

rs/backend/src/state.rs
bitdivine: Done

rs/backend/src/state/tests.rs
bitdivine: Done

PR #3642, opened 2023-10-31
Add tests for withdrawing BTC

dskloetd:
# Motivation

When withdrawinig BTC, the transactions can reloaded twice. Once while the modal is open and showing the 4 steps (reloading is the last ste) and once after the modal is closed.
The second reload is unnecessary.
Before I fix this issue, I want to add a test that demonstrates it so it's clear that the issue is fixed when it is.

# Changes

Add tests that check that the balance and transactions are reloaded after making a BTC withdrawal. These test mirror the 2 tests before it which test the same for making ckBTC transactions.

The final expectation shows the additional unnecessary reload.

# Tests

Only tests

# Todos

not necessary


mstrasinskis (approved):
LGTM, thanks!

PR #3643, opened 2023-10-31
feat: Add schema to init args

bitdivine:
# Motivation
Testing schema migrations is a lot easier if the code for both schemas is included in the code and the switch is controlled by argument.

# Changes
- Include the preferred schema as an argument.
  Note:  This argument is as yet unused.

# Tests
- The downgrade-upgrade tests ensure that the argument does not interfere with current deployments.
- The argument currently does nothing, so there is no further functionality to check.

# Todos


rs/backend/src/arguments.rs
```diff
@@ -12,11 +12,18 @@ use regex::{Captures, Regex};
...
 use std::collections::HashMap;
 
+use crate::accounts_store::schema::SchemaLabel;
+
 /// `init` and `post_upgrade` arguments
 #[derive(Debug, Default, Eq, PartialEq, CandidType, Serialize, Deserialize)]
 pub struct CanisterArguments {
     /// Values that are to be set in the web front end, by injecting them into JavaScript.
     pub args: Vec<(String, String)>,
+    /// The preferred schema.  If there is existing data in another schema, it will be converted to this schema.
```
frederikrothenberger, line 22: Is this actually true yet? If I understand correctly, the `AccountsDbAsProxy` should in the future do the migration, but does not do it yet.

Maybe in that case it would be more accurate to leave a note about planned future behaviour?

frederikrothenberger (approved):
LGTM, however I don't know the infrastructure enough to tell whether you updated all relevant places where the argument is used.

rs/backend/src/arguments.rs
bitdivine: Correct, this is the variable that _will_ be used but isn't used yet.  I can add a note to that effect.  I was worried about forgetting the note. :-D

bitdivine: Done.

PR #3644, opened 2023-10-31
Check sns_aggregator schema

bitdivine:
# Motivation
The `sns_aggregator` code and `.did` file should be in sync.  The current preferred approach is to decide on an API, express it in a .did file and then verify that the code matches the .did "spec".

# Changes
- Adda  cargo unit test 

# Tests

# Todos

### DIFF
--- a/rs/sns_aggregator/Cargo.toml
+++ b/rs/sns_aggregator/Cargo.toml
@@ -29,5 +29,8 @@ serde_derive = "1.0.126"
 serde_json = "1.0.91"
 sha2 = "0.10.6"
 
+[dev-dependencies]
+pretty_assertions = "1.4.0"
+
 [features]
 reconfigurable = []
--- a/rs/sns_aggregator/src/lib.rs
+++ b/rs/sns_aggregator/src/lib.rs
@@ -91,8 +91,7 @@ fn tail_log(limit: Option<u16>) -> String {
 /// Web server
 #[candid_method(query)]
 #[export_name = "canister_query http_request"]
-fn http_request(request: HttpRequest) -> HttpResponse
-{
+fn http_request(request: HttpRequest) -> HttpResponse {
     ic_cdk::setup();
     match request.url.as_ref() {
         "/__candid" => HttpResponse::from(__export_service()),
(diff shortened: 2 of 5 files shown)


rs/sns_aggregator/src/lib.rs
```diff
@@ -240,3 +238,50 @@ export_service!();
 fn interface() -> String {
     __export_service()
 }
+
```
bitdivine, line 241: Note: The code block below is from [the ic repo](https://github.com/dfinity/ic/blob/master/rs/nns/handlers/root/impl/canister/canister.rs#L237), with minor adjutments.

rs/sns_aggregator/src/lib.rs
```diff
@@ -165,7 +163,7 @@ fn post_upgrade(config: Option<Config>) {
 ///
 /// Note: This _could_ be exposed in production if limited to the controllers
 ///  - Controllers can be obtained by the async call: `agent.read_state_canister_info(canister_id, "controllers")`
-#[cfg(feature = "reconfigurable")]
+#[cfg(any(test, feature = "reconfigurable"))]
```
bitdivine, line 166: The `.did` file declares the "reconfigurable" interface.  We test that the declared interface matches the spec.  We probably need a different .did file for every set of features and tests to match.  To avoid making this PR too large, I propose enabling the method in test builds (not in prod!) then making a .did file per feature and adjusting the tests.  

An alternative would be to remove the method declaration from the .did file, leave this as it is, and include the method in specialized .did files.  However given that e2e tests use `reconfigure()` they are likely to break without a declaration.

rs/sns_aggregator/Cargo.toml
```diff
@@ -29,5 +29,8 @@ serde_derive = "1.0.126"
 serde_json = "1.0.91"
 sha2 = "0.10.6"
 
+[dev-dependencies]
+pretty_assertions = "1.4.0"
```
bitdivine, line 33: `assert_eq!` produces unreadable output when long strings are compared.  `pretty_assertions::assert_eq!` prints a nice diff.

PR #3646, opened 2023-11-01
Bump gix-cmp to v3.1.0

peterpeterparker:
# Motivation

Bump gix-cmp to v3.1.0@next because we just released v3.0.0.


mstrasinskis (approved):
Thanks

PR #3647, opened 2023-11-01
fix: expandObject to respect arrays

mstrasinskis:
# Motivation

Add array support to `expandObject` util.

# Changes

- Returns array as array, not an object.

# Tests

Extended.

# Todos

Not necessary

# Screenshots

## Before
image

## After
image

### DIFF
--- a/frontend/src/lib/utils/utils.ts
+++ b/frontend/src/lib/utils/utils.ts
@@ -378,7 +378,7 @@ export const expandObject = (value: unknown): unknown => {
   if (typeof value === "string") {
     try {
       return JSON.parse(value);
-    } catch (e) {
+    } catch (e: unknown) {
       return value;
     }
   }
--- a/frontend/src/lib/utils/utils.ts
+++ b/frontend/src/lib/utils/utils.ts
@@ -386,12 +386,15 @@ export const expandObject = (value: unknown): unknown => {
     return value.map(expandObject);
   }
   if (typeof value === "object") {
-    Object.keys(value).forEach(
+    // to avoid mutating original object
+    const result = { ...value };
+    Object.keys(result).forEach(
       (key) =>
-        ((value as Record<string, unknown>)[key] = expandObject(
-          (value as Record<string, unknown>)[key]
+        ((result as Record<string, unknown>)[key] = expandObject(
+          (result as Record<string, unknown>)[key]
         ))
     );
+    return result;
   }
   return value;
 };
--- a/frontend/src/lib/utils/utils.ts
+++ b/frontend/src/lib/utils/utils.ts
@@ -371,30 +371,30 @@ export const isPngAsset = (
  * @param obj
  * @returns parsed object
  */
-export const expandObject = (
-  obj: Record<string, unknown>
-): Record<string, unknown> =>
-  Object.keys(obj).reduce(
-    (acc, key) => {
-      const value = obj[key];
-      if (typeof value === "string") {
-        try {
-          acc[key] = JSON.parse(value);
-        } catch (e) {
-          acc[key] = value;
-        }
-      } else if (typeof value === "object") {
-        acc[key] =
-          value !== null
-            ? expandObject(value as Record<string, unknown>)
-            : value;
-      } else {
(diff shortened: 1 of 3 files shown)


frontend/src/lib/utils/utils.ts
```diff
@@ -371,30 +371,30 @@ export const isPngAsset = (
...
-            : value;
-      } else {
-        acc[key] = value;
-      }
-      return acc;
-    },
-    {} as Record<string, unknown>
-  );
+export const expandObject = (value: unknown): unknown => {
+  if (value === null || value === undefined) {
```
peterpeterparker, line 375: `isNullish`?

frontend/src/lib/utils/utils.ts
```diff
@@ -371,30 +371,30 @@ export const isPngAsset = (
...
+    return value;
+  }
+  if (typeof value === "string") {
+    try {
+      return JSON.parse(value);
+    } catch (e) {
```
peterpeterparker, line 381: `e: unknown`?

frontend/src/lib/utils/utils.ts
```diff
@@ -371,30 +371,30 @@ export const isPngAsset = (
...
-    },
-    {} as Record<string, unknown>
-  );
+export const expandObject = (value: unknown): unknown => {
+  if (value === null || value === undefined) {
+    return value;
+  }
+  if (typeof value === "string") {
+    try {
+      return JSON.parse(value);
```
peterpeterparker, line 380: should we use the reviver here as well?

```
JSON.parse(value, jsonReviver)
```

frontend/src/lib/utils/utils.ts
```diff
@@ -371,30 +371,30 @@ export const isPngAsset = (
...
+    } catch (e) {
+      return value;
+    }
+  }
+  if (Array.isArray(value)) {
+    return value.map(expandObject);
+  }
+  if (typeof value === "object") {
+    Object.keys(value).forEach(
```
peterpeterparker, line 389: isn't a `return` missing here?

frontend/src/lib/utils/utils.ts
mstrasinskis: The main goal of this function is to parse leaf nodes. It worked like this before, so I don't think it worth to extend it.

frontend/src/lib/utils/utils.ts
mstrasinskis: Because of the ts bug with `isNullish` there will be an error on the line 389 (`Object.keys(value).forEach(`):
` Argument of type 'object | null' is not assignable to parameter of type '{}'.`

frontend/src/lib/utils/utils.ts
mstrasinskis: Updated.

frontend/src/lib/utils/utils.ts
mstrasinskis: It updated keys of the `value` object and return is on the line 396.

frontend/src/lib/utils/utils.ts
peterpeterparker: Ah yeah, gotcha

frontend/src/lib/utils/utils.ts
peterpeterparker: Works for me

frontend/src/lib/utils/utils.ts
peterpeterparker: I see, then, from a functional perspective and given the fact that value is the parameter of the function, it would maybe be cleaner to spread/clone the value before modifying?

```
if (typeof value === "object") {
   let result = {...value};
   Object.keys(result).forEach(
      etc.
  return result;
}
```

peterpeterparker (approved):
LGTM, thanks

frontend/src/lib/utils/utils.ts
mstrasinskis: Thanks, good point, updated!

PR #3651, opened 2023-11-02
Add convertCkBTCToBtcIcrc2 to ckbtc-convert.services

dskloetd:
# Motivation

We want to use ICRC-2 approve transfer for BTC withdrawal.
This PR adds a service function `convertCkBTCToBtcIcrc2` to be used in `CkBTCTransactionModal.svelte` in a later PR.

# Changes

1. Add `convertCkBTCToBtcIcrc2` in `frontend/src/lib/services/ckbtc-convert.services.ts` which approves the transfer by the minter canister and than asks the minter to make the withdrawal.
2. Add a new `APPROVE_TRANSFER` step in `ConvertBtcStep`. This will be used in a later PR to render the ICRC-2 steps correctly if the feature flag is enabled.
3. Separate a `reload` function from `retrieveBtcAndReload` to reuse in `convertCkBTCToBtcIcrc2`.

# Tests

Added unit tests.
Manually tested in a separate branch which has the service function hooked up already.

# Todos

Will add when hooked up.


frontend/src/lib/constants/constants.ts
```diff
@@ -26,3 +26,4 @@ export const SECONDS_IN_EIGHT_YEARS = SECONDS_IN_YEAR * 8;
 export const DAYS_IN_NON_LEAP_YEAR = 365;
 
 export const NANO_SECONDS_IN_MILLISECOND = 1_000_000;
+export const NANO_SECONDS_IN_MINUTE = NANO_SECONDS_IN_MILLISECOND * 1000 * 60;
```
peterpeterparker, line 29: `1_000` since we use `1_000_000` above?

peterpeterparker (approved):
LGTM, thx

frontend/src/lib/constants/constants.ts
dskloetd: Thanks. done.

PR #3653, opened 2023-11-02
Don't call global.fetch from unit tests.

dskloetd:
# Motivation

This [test failure](https://github.com/dfinity/nns-dapp/actions/runs/6730359901/job/18292918007) seems to be the result of a unit test making a real request:
```
Server returned an error:
   ❯ src/tests/lib/components/launchpad/ProjectCard.spec.ts > ProjectCard > not signed in > should not display a spinner when the swapCommitment is not loaded
  Code: 400 (Bad Request)
  Body: Specified ingress_expiry not within expected range: Minimum allowed expiry: 2023-11-02 09:03:58.649355448 UTC, Maximum allowed expiry: 2023-11-02 09:09:28.649355448 UTC, Provided expiry:        2023-10-24 09:28:28 UTC
  Retrying request.
  ```
Unit test should never (try to) make actual network requests.
This PR tries to catch it when a unit test tries to call `global.fetch`. Strangely there seems to be some interference between tests that makes it so that the check doesn't work when all tests are run together. Even though individual test files should be isolated from each other.
Still it seems better to have the check than not to have it.

# Changes

1. Add code in `vitest.setup.ts` to disallow calling `global.fetch`.
2. Mock `sns-sale.api` in `ProjectCard.spec.ts` so it doesn't try to make actual requests.

# Tests

pass

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp-unreleased.md
+++ b/CHANGELOG-Nns-Dapp-unreleased.md
@@ -56,6 +56,7 @@ proposal is successful, the changes it released will be moved from this file to
 * Formatting for `Cargo.toml` files.
 * Add test to check that the nns-dapp cargo and npm versions match.
 * Script to deploy nns-dapp on `DevEnv`.
+* Try to prevent calls to global.fetch in unit tests.
 
 #### Changed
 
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
Thanks!

How come it didn't fail earlier?

dskloetd:
> How come it didn't fail earlier?

It's all a bit mysterious. One possibility is that usually the test passes before the failing fetch manages to cause problems.

PR #3658, opened 2023-11-02
refactor: Action to clone snsdemo

bitdivine:
# Motivation
Checking out the `snsdemo` repo requires several steps.  These steps are repeated across the code.  I am about to add another workflow that needs these steps.  Rather than copy and paste a block, why not factor it out?

# Changes
* Define a GitHub action that ensures that the snsdemo repo is checked out.
* Update two existing code blocks that do the same.  There are more but they are a bit less obvious so I would rather deal with the complicated cases in a separate PR.

# Tests
- See CI

# Todos


CHANGELOG-Nns-Dapp-unreleased.md
```diff
@@ -22,6 +22,7 @@ proposal is successful, the changes it released will be moved from this file to
 
 #### Changed
 
+* Factor out the `snsdemo` installation.
```
dskloetd, line 25: This doesn't affect the wasm we release, right?
So it should go under `### Operations` below.

bitdivine: Good catch!  Done.

dskloetd (approved):

PR #3665, opened 2023-11-03
Fix back navigation from sns creation proposal

mstrasinskis:
# Motivation

Fix back button.

# Changes

- adds new store to preserve query params (to jump back to the same sns project)
- update back button logic
- fix tiny style issue (screenshot)

# Tests

works

# Todos

# Screenshots

Screen Recording 2023-11-03 at 11 31 11

### DIFF
--- a/frontend/src/lib/stores/routes.store.ts
+++ b/frontend/src/lib/stores/routes.store.ts
@@ -2,3 +2,4 @@ import type { AppPath } from "$lib/constants/routes.constants";
 import { writable } from "svelte/store";
 
 export const referrerPathStore = writable<AppPath | undefined>(undefined);
+export const referrerQueryParamsStore = writable<string | undefined>(undefined);
--- a/frontend/src/lib/stores/routes.store.ts
+++ b/frontend/src/lib/stores/routes.store.ts
@@ -2,4 +2,3 @@ import type { AppPath } from "$lib/constants/routes.constants";
 import { writable } from "svelte/store";
 
 export const referrerPathStore = writable<AppPath | undefined>(undefined);
-export const referrerQueryParamsStore = writable<string | undefined>(undefined);
--- a/frontend/src/routes/(app)/(u)/(detail)/proposal/+layout.svelte
+++ b/frontend/src/routes/(app)/(u)/(detail)/proposal/+layout.svelte
@@ -14,7 +14,7 @@
       return Promise.resolve();
     }
 
-    goto(
+    return goto(
       $referrerPathStore === AppPath.Launchpad
         ? $referrerPathStore
         : $proposalsPathStore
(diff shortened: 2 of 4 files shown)


frontend/src/routes/(app)/(u)/(detail)/proposal/+layout.svelte
```diff
@@ -5,12 +5,16 @@
...
+    referrerQueryParamsStore,
+  } from "$lib/stores/routes.store";
 
   const back = (): Promise<void> =>
     goto(
-      $referrerPathStore === AppPath.Launchpad
-        ? $referrerPathStore
+      $referrerPathStore === AppPath.Launchpad ||
+        $referrerPathStore === AppPath.Project
+        ? `${$referrerPathStore}${$referrerQueryParamsStore ?? ""}`
```
peterpeterparker, line 17: `history.back()` would not work?

peterpeterparker: Ah yeah I see, we always do `goto` for back, maybe something we can improve in another PR.

mstrasinskis: I just have a feeling that some svelte state could be lost, but for simplicity let's use your suggestion. Thanks!

frontend/src/lib/stores/routes.store.ts
```diff
@@ -2,3 +2,4 @@ import type { AppPath } from "$lib/constants/routes.constants";
 import { writable } from "svelte/store";
 
 export const referrerPathStore = writable<AppPath | undefined>(undefined);
+export const referrerQueryParamsStore = writable<string | undefined>(undefined);
```
peterpeterparker, line 5: we should maybe not introduce a new store but rather enhance `referrerPathStore` this because `referrerQueryParamsStore` never gets resetted so we might end up have an unexpected side effect where "back" is performance on an unrelated query param value

peterpeterparker: WDYT?

frontend/src/routes/(app)/(u)/(detail)/proposal/+layout.svelte
peterpeterparker: @mstrasinskis maybe not in this PR because we should also handle the case where the user lands on the page from an external place

In Juno I do

```
export const back = async () => {
	const stack: number | undefined = history.state?.['sveltekit:index'];

	if (!stack || stack === 0) {
		await goBack();
		return;
	}

	history.back();
};
```

But it's meh. So let's try in another PR I would say

frontend/src/lib/stores/routes.store.ts
peterpeterparker: Solve with history.back

PR #3668, opened 2023-11-03
Create dependabot.yml

bitdivine:
# Motivation
We would like to keep the Rust dependencies up to date.  Some are simple to manage, third party packages with semantic versions and these can be taken care of by dependabot.  Others, in particular updates of crates in the IC repo, require a tailored approach.  This PR addresses simple third party packages.

Note: Dependabot opens PRs when updates are available.

### References
- https://dfinity.atlassian.net/browse/GIX-1964
- https://dfinity.atlassian.net/browse/GIX-1967

# Changes
- Enable dependabot for rust crate updates.

# Tests

# Todos

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
+        # These are peer deps of Cargo and should not be automatically bumped
+        - dependency-name: "semver"
+        - dependency-name: "crates-io"
+    rebase-strategy: "disabled"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
     schedule:
       interval: "weekly"
     ignore:
-        # These are peer deps of Cargo and should not be automatically bumped
-        - dependency-name: "semver"
-        - dependency-name: "crates-io"
+      # These are peer deps of Cargo and should not be automatically bumped
+      - dependency-name: "semver"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
+      # These are peer deps of Cargo and should not be automatically bumped
+      - dependency-name: "semver"
+      - dependency-name: "crates-io"
```
lmuntaner, line 13: Don't you have to tell it to ignore the IC dependencies?

lmuntaner (commented):
Does this update SNS aggregator and NNS Dapp dependencies?

.github/dependabot.yml
bitdivine: I believe that the answer is "ic dependencies will be ignored because they specify a version as a git commit".  I am not super-familiar with dependabot so I cannot answer for sure, so I planned to run it with the default configuration and make changes as necessary.  Via the PRs we do have human eyes on the proposed changes.

What I have read:
- The documentation by GitHub for that configuration file: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
- [The dependabot cargo driver (skim read)](https://github.com/dependabot/dependabot-core/blob/94848f7abaaef1d137f2cb539e478bca75db73b0/cargo/lib/dependabot/cargo/file_parser.rb#L45).
- I looked at the usage of dependabot in dfinity, for example, [in the motoko repository](https://github.com/dfinity/motoko/blob/7fbe0135b18de3e04dc65164754762028c323400/.github/dependabot.yml#L4).

lmuntaner: Ok, makes sense.

lmuntaner (approved):

PR #3670, opened 2023-11-03
GIX-2022: Update proposals

bitdivine:
# Motivation
We wish to keep the proposal payload rendering code up to date.

# Changes
- Run the code that derives the proposal rust types from candid regularly.

# Tests

# Todos


.github/workflows/update-proposals.yml
```diff
@@ -0,0 +1,68 @@
+# A GitHub Actions workflow that regularly updates the proposals rendering code
+# and creates a PR for any changes.
+name: Update proposals
+on:
+  schedule:
+    - cron: '30 3 * * *'
```
lmuntaner, line 6: Why so often?

bitdivine: So that the code is updated reasonably quickly after a did change.  Most of the time there are no .did changes and so this does nothing.  The maximum frequency we will get, if upstream makes a lot of changes, is one PR per week, unless we do runs manually.

lmuntaner: That makes sense

lmuntaner (approved):

PR #3680, opened 2023-11-06
Use ICRC-2 in BTC withdrawal

dskloetd:
# Motivation

By using ICRC-2 for BTC withdrawal we can't transfer to the wrong account because it's the minter itself that makes the transfer.

# Changes

When `ENABLE_CKBTC_ICRC2` is enabled use `convertCkBTCToBtcIcrc2` instead of `convertCkBTCToBtc` and tell the `ConvertBtcInProgress` component to use the ICRC-2 steps.

# Tests

1. In `frontend/src/tests/lib/modals/accounts/CkBTCTransactionModal.spec.ts` duplicate the tests that expect a call to `convertCkBTCToBtc` and expect a call to `convertCkBTCToBtcIcrc2` instead if `ENABLE_CKBTC_ICRC2` is enabled.
2. In `frontend/src/tests/lib/pages/CkBTCWallet.spec.ts` duplicate the tests that expect a call to `icrcLedgerApi.icrcTransfer` and expect calls to `icrcLedgerApi.approveTransfer` and `ckbtcMinterApi.retrieveBtcWithApproval` instead if `ENABLE_CKBTC_ICRC2` is enabled.

Also tested manually [here](https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/) by sending some ckBTC to `1ASLxsAMbbt4gcrNc6v6qDBW4JkeWAtTeh`.

# Todos

### DIFF
--- a/frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
+++ b/frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
@@ -3,6 +3,7 @@
   import { startBusy, stopBusy } from "$lib/stores/busy.store";
   import { i18n } from "$lib/stores/i18n";
   import { toastsSuccess } from "$lib/stores/toasts.store";
+  import { ENABLE_CKBTC_ICRC2 } from "$lib/stores/feature-flags.store";
   import type { NewTransaction, TransactionInit } from "$lib/types/transaction";
   import { TransactionNetwork } from "$lib/types/transaction";
   import type { ValidateAmountFn } from "$lib/types/transaction";
@@ -18,6 +19,7 @@
   import type { CkBTCAdditionalCanisters } from "$lib/types/ckbtc-canisters";
   import {
     convertCkBTCToBtc,
+    convertCkBTCToBtcIcrc2,
     type ConvertCkBTCToBtcParams,
     retrieveBtc,
   } from "$lib/services/ckbtc-convert.services";
@@ -109,6 +111,8 @@
     }
   };
 
+  let useIcrc2ForProgress = true;
+
   const convert = async ({
     detail: { sourceAccount, amount, destinationAddress },
   }: CustomEvent<NewTransaction>) => {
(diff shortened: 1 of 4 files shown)


frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
```diff
@@ -109,6 +111,8 @@
     }
   };
 
+  let useIcrc2ForProgress = true;
```
lmuntaner, line 114: Why not setting it to `$ENABLE_CKBTC_ICRC2`?

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
```diff
@@ -124,12 +128,24 @@
...
 
-    const { success } = withdrawalAccount
-      ? await retrieveBtc(params)
-      : await convertCkBTCToBtc({
-          source: sourceAccount,
-          ...params,
-        });
+    let success = false;
+
+    if (withdrawalAccount) {
```
lmuntaner, line 133: Is there a reason to check this instead of first `$ENABLE_CKBTC_ICRC2`?

I would find it easier to understand if we first check for the feature flag. I expect a feature flag to split to different code paths as soon as possible.

Not a strong opinion, just trying to understand whether I was missing something.

lmuntaner (approved):
Just a couple of comments, but already approving.

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
dskloetd: If you use the old flow but reload the page before the withdrawal is finished, you have ckBTC sitting in the minter's withdrawal account.
It looks like this:
image

If you click on that card, it opens the `CkBTCTransactionModal` such that `withdrawalAccount` is true. If you don't click on that card, your ckBTC will just set there in the withdrawal account until you do.

If we then enable the feature flag, you still have tokens in the withdrawal account that can only be withdrawn through the old flow. So we have to check that situation regardless of the feature flag.

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
dskloetd: If you withdrawal from a minter withdrawal account, we still have to use the old flow, even if the feature flag is enabled.

frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
lmuntaner: Ahh, ok, thanks for the explanation!

PR #3681, opened 2023-11-06
Remove unused parameter loadTransactions

dskloetd:
# Motivation

Parameter `loadTransactions` of `ckBTCTransferTokens` only ever gets the value `false` so it's less confusing to just remove the parameter.

# Changes

Remove the `loadTransactions` parameter from `ckBTCTransferTokens` and all the places where it is passed from.

# Tests

Removed a test where `true` is passed and updated other tests to no longer pass the value.

# Todos

not necessary


dskloetd:
There is more that can be removed so marking as draft.

dskloetd:
I think it's ready now.

peterpeterparker (approved):
LGTM, thx

PR #3689, opened 2023-11-07
fix: Make the prod releases public

bitdivine:
# Motivation
The `prod` releases should be public.  In particular, CI expects the `aggregator-prod` to be public.

# Changes
- Add the two `prod` releases to the set  of releases that should be public.

# Tests
None

# Todos


lmuntaner (approved):
Thanks!

PR #3690, opened 2023-11-07
Add balance not available TokensTable

lmuntaner:
# Motivation

When the user is not logged in, we want to tease them with the same Tokens view. But instead of balances, we want to show a potential balance if they login.

# Changes

* In UserTokenData: add field `token` and type `"not-available"` for `balance`.
* Manage when balance is `"not-available"` in DesktopTokensTableRow.
* Add a `"not-available"` in the mock data.

# Tests

* Test the balance value if `"not-available"` in DesktopTokensTable.spec

# Todos

Not necessary yet.


lmuntaner:
* main
  * **PR #3690** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3690" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3690?utm_source=stack-comment).

lmuntaner:
New UI:

lmuntaner:

frontend/src/lib/components/tokens/DesktopTokensTable/DesktopTokensTableRow.svelte
```diff
@@ -47,7 +47,13 @@
   </td>
   <td>
     <div class="universe-balance">
-      <AmountDisplay singleLine amount={userTokenData.balance} />
+      {#if userTokenData.balance === "not-available"}
+        <span data-tid="token-value-label"
+          >{`--/-- ${userTokenData.token}`}</span
```
dskloetd, line 52: What does `--/--` mean? Why those specific symbols?

frontend/src/lib/types/tokens-page.ts
```diff
@@ -10,7 +10,8 @@ export enum UserTokenAction {
 export type UserTokenData = {
   universeId: Principal;
   title: string;
-  balance: TokenAmount;
+  balance: TokenAmount | "not-available";
   logo: string;
+  token: string;
```
dskloetd, line 15: Is this only needed when there is no token balance because otherwise we get the token from `TokenAmount`?
What if these don't match?
Can we do something else?
Maybe
```
balance: TokenAmount | UnavailableToken
```
Where `UnavailableToken` has a token field?

frontend/src/lib/components/tokens/DesktopTokensTable/DesktopTokensTableRow.svelte
lmuntaner: Sorry, it's `-/-`.

And it's from Figma. That's how they want to present when a balance is not available because the user is not logged in.

https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?node-id=408%3A2789&mode=dev

frontend/src/lib/types/tokens-page.ts
lmuntaner: Good idea, done

lmuntaner:

dskloetd (approved):

PR #3693, opened 2023-11-08
Fix table background

lmuntaner:
# Motivation

Use the themed background variable for the row background in the tokens table.

# Changes

* Use the commented background color because now the variable is already available in gix-components.

# Tests

No tests for this change.

# Todos

Not necessary

### DIFF
--- a/frontend/src/lib/components/tokens/DesktopTokensTable/DesktopTokensTableRow.svelte
+++ b/frontend/src/lib/components/tokens/DesktopTokensTable/DesktopTokensTableRow.svelte
@@ -69,8 +69,7 @@
   tr {
     @include interaction.tappable;
 
-    // background-color: var(--input-focus-background);
-    background-color: var(--purple-75);
+    background-color: var(--input-focus-background);
 
     &:hover {
       background-color: var(--input-background);


lmuntaner:
* main
  * **PR #3693** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3693" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3693?utm_source=stack-comment).

lmuntaner:
New UI (dark)

lmuntaner:

frontend/src/lib/components/tokens/DesktopTokensTable/DesktopTokensTableRow.svelte
```diff
@@ -69,8 +69,7 @@
   tr {
     @include interaction.tappable;
 
-    // background-color: var(--input-focus-background);
-    background-color: var(--purple-75);
+    background-color: var(--input-focus-background);
```
dskloetd, line 72: Why "input focus"? What does that refer to?

dskloetd: OK, but what does a table row have to to with input or focus?
If this is the new and improved process, we better understand what we're doing and why.

Is it possible that the designer first decided on the color they wanted and then just picked some arbitrary variable with that color without taking into consideration the meaning of the variable?
If I look in [this file](https://github.com/dfinity/gix-components/blob/36eeb97f80fc5eb8b2787325ea3dfac8e93a054d/src/lib/styles/themes/light.scss#L12), there is for example also `--card-background` which I think is already closer to the meaning of this element. But maybe we should define a table background color explicitly and use that.
Also, looking at the design, one of the table rows is "focused" and that one has color "Input/Background" while the others have color "Input/Focus Background" so the meaning of "focus" seems to be inverted?
And then the Dragginz row has background color "Popup/Background tint", for which I can't find a corresponding variable, but surely we don't want to use a different color variable (with the same color) for the Dragginz row than for the other rows.

So I think we need to be more deliberate here and ask the same from the designers if we want the code to reflect the design like this.

lmuntaner: Those are good questions.

The `--card-background` will be deprecated. The idea is to use the colors in the local variables [here](https://www.figma.com/file/P4JbeFs6oGt2WmKeB4PHFc/Updated-NNS-Styles-%26-Components?type=design&node-id=514-220550&mode=design&t=I8mRm6HgwcW02Ef7-0).

When we move to design tokens, which is why David Aerne is doing, we will then transform those variables into more meaningful variables. We're not trying to do it here yet, just to start using the new variables also used in Figma.

The point with Dragginz is interesting, I didn't see that. I looked into the variables and "Popup/Background tint" uses the same colors than "Input/Focus Background".

I like beter the variable "Popup/Background tint". Therefore, I'll expose that variable in gix-components and use it here. I'll close this PR instead.

dskloetd: I must be doing something wrong because I don't see those variable anywhere in Figma. If I follow your link, I just see things like "Purple/250".

This table is not a popup so using a variable with "popup" in the name isn't really better than using a variable with "input" in the name.

I think there needs to be a more generic background variable and then a more specific table background variable which is defined to be equal to the more general background variable. And then the other (popup/input) background variable should also be declared to be equal to one of the generic background variables.

PR #3701, opened 2023-11-09
mapTransaction prop on IcrcTransactionsList

dskloetd:
# Motivation

For ckBTC withdrawal transactions we want to render the BTC address as the "To:" address.
This address is encoded in the memo of burn transactions and we want to place it in the `to:` of the transaction rendered in the `TransactionCard`.
This can be done in `mapIcrcTransaction` but the logic is specific to ckBTC so shouldn't be applied to all tokens.
My solution is to be able to pass a different `mapTransaction` function per token.

This PR only makes it possible to pass a `mapTransaction` to `IcrcTransactionsList` but does not use it for any custom mapping for ckBTC.

# Changes

1. Add `mapTransaction` prop to `IcrcTransactionsList` and pass it on to `IcrcTransactionCard` which will call the passed function instead of `mapIcrcTransaction`.
2. Pass `mapIcrcTransaction` from `CkBTCTransactionsList` and `SnsTransactionsList`. This is for now. In a future PR we'll pass a custom function from `CkBTCTransactionsList`.

# Tests

Add tests for `IcrcTransactionsList` and `IcrcTransactionCard` that the passed in mapping function is used.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
@@ -12,6 +12,7 @@
   import {
     getSortedTransactionsFromStore,
     isIcrcTransactionsCompleted,
+    mapIcrcTransaction,
   } from "$lib/utils/icrc-transactions.utils";
   import IcrcTransactionsList from "$lib/components/accounts/IcrcTransactionsList.svelte";
   import type { UniverseCanisterId } from "$lib/types/universe";
@@ -106,5 +107,6 @@
     {completed}
     {descriptions}
     {token}
+    mapTransaction={mapIcrcTransaction}
   />
 </CkBTCWalletTransactionsObserver>
--- a/frontend/src/lib/components/accounts/IcrcTransactionCard.svelte
+++ b/frontend/src/lib/components/accounts/IcrcTransactionCard.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import type { Account } from "$lib/types/account";
-  import { mapIcrcTransaction } from "$lib/utils/icrc-transactions.utils";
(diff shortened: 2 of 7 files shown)


lmuntaner (commented):
Have you thought instead of using a different component with that logic and add a slot to IcrcTransactionList to render the list of transactions?

```
<TransactionList ...>
  {#each transactions as transaction}
    <CkBTCTransactionCard ... />
  {/each>
</>
```

And for others:
```
<TransactionList ...>
  {#each transactions as transaction}
    <IcrcTransactionCard ... />
  {/each>
</>
```

What do you think?

dskloetd:
Right now the loop is in `IcrcTransactionsList`. If I understand correctly you are proposing to duplicate that logic. But that doesn't seem necessary. If I want to use a custom card component, I would consider passing the component type to use as a prop to `IcrcTransactionsList` so it could render a custom component without duplicating the code around it.
I might still do that but right now the code path that I want to customize is in `mapIcrcTransaction` so duplicating the card component seems premature. What do you think? @lmuntaner

lmuntaner:
> Right now the loop is in `IcrcTransactionsList`. If I understand correctly you are proposing to duplicate that logic. But that doesn't seem necessary. If I want to use a custom card component, I would consider passing the component type to use as a prop to `IcrcTransactionsList` so it could render a custom component without duplicating the code around it. I might still do that but right now the code path that I want to customize is in `mapIcrcTransaction` so duplicating the card component seems premature. What do you think? @lmuntaner

My feeling is that `TransactionList` is doing too much. I know that there is a loop duplication. But I believe that this duplication comes with more better maintainability and flexibility of the component.

I'm afraid of adding props to current components, that's how we end up with god components. I prefer to split them.

But we can do it later.

lmuntaner (approved):

PR #3702, opened 2023-11-09
Update snsdemo to release-2023-11-08

sa-github-api:
# Motivation
We would like to keep the testing environment, provided by snsdemo, up to date.

# Changes
* Updated `snsdemo` version in `dfx.json`.
* Ensured that the `dfx` version and IC commit in `dfx.json` match `snsdemo`.

# Tests
  - [ ] Please check the API updates for any breaking changes that affect our code.
  - [ ] Please check for new proposal types.

PR #3703, opened 2023-11-09
Use a unique branch name for snsdemo updates

dskloetd:
# Motivation

If the update workflow runs again while we are still considering a previous update, we don't want the previous update to be overwritten.
It might have additional manual changes or a custom PR description that we don't want to lose.
By using a fixed branch name, we might lose these changes, so we want to use a unique branch name.
The GitHub action to create a PR already supports this out of the box:
https://github.com/peter-evans/create-pull-request#alternative-strategy---always-create-a-new-pull-request-branch
Note that the documentation does mention the follow:
> This strategy is not recommended because if not used carefully it could result in multiple pull requests being created unnecessarily.

Since this workflow runs only once a week or when triggered manually, I don't think creating multiple pull requests is a concern.

Here is an example PR created this way: https://github.com/dfinity/nns-dapp/pull/3702

This PR is just for the snsdemo update flow but when it is working as expected we should update the other update flows as well.

# Changes

Add `branch-suffix: timestamp` to the inputs of the `create-pull-request` action.

# Tests

Ran manually and it created this PR: https://github.com/dfinity/nns-dapp/pull/3702

# Todos

### DIFF
--- a/.github/workflows/update-snsdemo.yml
+++ b/.github/workflows/update-snsdemo.yml
@@ -63,6 +63,7 @@ jobs:
           committer: GitHub <noreply@github.com>
           author: gix-bot <gix-bot@users.noreply.github.com>
           branch: bot-snsdemo-update
+          branch-suffix: timestamp
           add-paths: .github/*/*ml
           delete-branch: true
           title: 'Update snsdemo to ${{ steps.update.outputs.release }}'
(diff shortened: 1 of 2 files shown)


bitdivine:
Note:

> By using a fixed branch name, we might lose these changes, so we want to use a unique branch name.

There are other, in my opinion better, ways of achieving the same end.  If you want to discuss, let's have a call.

lmuntaner (approved):
Thanks!

We can always come back to unique branch name if we find problems with this approach.

PR #3705, opened 2023-11-10
Checkout at the specified location

bitdivine:
# Motivation
The snsdemo needs to be checked out outside the nns-dapp repository.

# Changes
- In the snsdemo checkout action, accept a `path` variable, similar to the vanilla checkout action.
- Improve logging in the snsdemo checkout action.
- Fix the error handling and improve logging in the error cases.

# Tests
- See CI for current use
- For checking the case where snsdemo is already checked out at the requested commit, see: https://github.com/dfinity/nns-dapp/actions/runs/6821083632/job/18550979415
- For the case where snsdemo is already checked out but at the wrong commit: https://github.com/dfinity/nns-dapp/actions/runs/6829080386/job/18574535473

# Todos

### DIFF
--- a/.github/actions/checkout_snsdemo/action.yaml
+++ b/.github/actions/checkout_snsdemo/action.yaml
@@ -37,11 +37,16 @@ runs:
       if: ${{ steps.have_snsdemo.outputs.have_snsdemo }} == 'true'
       shell: bash
       run: |
-        expected_commit="$(cd "${{ inputs.path }}" && git rev-parse "${{ steps.snsdemo_ref.outputs.ref }}" || true)"
-        actual_commit="$(cd "${{ inputs.path }}" && git rev-parse HEAD)"
+        set -euxo pipefail
+        cd "${{ inputs.path }}"
+        git fetch --tags || true
+        git tag
+        expected_commit="$(git rev-parse "${{ steps.snsdemo_ref.outputs.ref }}" || true)"
+        actual_commit="$(git rev-parse HEAD)"
         [[ "$expected_commit" == "$actual_commit" ]] || {
           echo "ERROR: snsdemo is already checked out but not at the requested commit."
           echo "Snsdemo path:     ${{ inputs.path }}"
+          echo "              ==  $PWD"
           echo "Requested ref:    ${{ steps.snsdemo_ref.outputs.ref }}"
           echo "Requested commit: $expected_commit  (May be missing in the case of a partial clone)"
           echo "Actual commit:    $actual_commit"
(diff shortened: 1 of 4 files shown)


.github/actions/checkout_snsdemo/action.yaml
```diff
@@ -1,14 +1,17 @@
...
-    out at the expected commit.  If not, this fails with an error merssage.
+    out at the expected commit.  If not, this fails with an error message.
   * Otherwise: Clones the snsdemo repository at the commit specified in `dfx.json`
 inputs:
   token:
     description: "Github access token used to clone"
     required: true
+  path:
+    description: "Where to check out the repository"
+    default: "snsdemo"
```
bitdivine, line 14: Alternative considered: Set the default to `../snsdemo`.  Decided against it as it has surprise factor for anyone used to the standard checkout action.  Better make the path explicit where the action is called, I think.

.github/actions/checkout_snsdemo/action.yaml
```diff
@@ -35,9 +38,10 @@ runs:
       shell: bash
       run: |
         expected_commit="$(git rev-parse "${{ steps.snsdemo_ref.outputs.ref }}" || true)"
-        actual_commit="$(git rev-parse HEAD)"
+        actual_commit="$(cd "${{ inputs.path }}" && git rev-parse HEAD)"
```
bitdivine, line 41: The code on the left actually omits the cd which is a bug.

lmuntaner (commented):
How are you going to use this? What's the default commit? I can't find it...

bitdivine:
> How are you going to use this?

The action is used in `.github/actions/build_nns_dapp/action.yaml` and `.github/workflows/deploy-to-app.yaml`.  The places where the action is used is not expected to change.

The bulk of this PR is actually bug fixes and they stand on their own even if nothing else is changed.  At root, the bugs were caused by not using the same path everywhere, so making it a variable showed me the bugs.  The original motivation was however to move the snsdemo to somewhere outside the nns-dapp repo and I do still intend to do this, to address some of the many warnings we are getting,  but in a separate PR.

> What's the default commit?

The default  commit is unchanged.  It's described in the action description at the top of the file:
```
Otherwise: Clones the snsdemo repository at the commit specified in `dfx.json`
```

lmuntaner (approved):
Thanks for the explanations.

Is this related to the comment you made that the snsdemo is used for two different things: utilities and the snapshot?

bitdivine:
> Is this related to the comment you made that the snsdemo is used for two different things: utilities and the snapshot?

The checkouts are used to get the utilities.  This PR changes nothing in that regard.

PR #3708, opened 2023-11-10
Fix: Aggregator with devenv

lmuntaner:
# Motivation

When connecting to `devenv`, we can't use the `raw` in the url.

# Changes

* Add an extra check in the environments constants to build the aggregator url.

# Tests

Only dev environment, this is not tested.

# Todos

### DIFF
--- a/frontend/src/lib/constants/environment.constants.ts
+++ b/frontend/src/lib/constants/environment.constants.ts
@@ -12,7 +12,7 @@ const snsAggregatorUrlEnv = envVars.snsAggregatorUrl ?? "";
 const snsAggregatorUrl = (url: string) => {
   try {
     const { hostname } = new URL(url);
-    if (isLocalhost(hostname)) {
+    if (isLocalhost(hostname) || hostname.includes("devenv")) {
       return url;
     }
 
--- a/frontend/src/lib/constants/environment.constants.ts
+++ b/frontend/src/lib/constants/environment.constants.ts
@@ -12,7 +12,12 @@ const snsAggregatorUrlEnv = envVars.snsAggregatorUrl ?? "";
 const snsAggregatorUrl = (url: string) => {
   try {
     const { hostname } = new URL(url);
-    if (isLocalhost(hostname) || hostname.includes("devenv")) {
+    if (isLocalhost(hostname)) {
+      return url;
+    }
+
+    // We don't want to add `raw` to the URL if we are running against a devenv environment.
+    if (url.includes("devenv")) {
       return url;
     }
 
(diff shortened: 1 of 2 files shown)


lmuntaner:
* main
  * **PR #3708** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3708?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3708?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

frontend/src/lib/constants/environment.constants.ts
```diff
@@ -16,6 +16,11 @@ const snsAggregatorUrl = (url: string) => {
       return url;
     }
 
+    // We don't want to add `raw` to the URL if we are running against a devenv environment.
+    if (url.includes("devenv")) {
```
dskloetd, line 20: Is this for the case when you are connecting from `localhost:5173` or from `https://qsgjb-riaaa-aaaaa-aaaga-cai.llorenc-ingress.devenv.dfinity.network/` or both?

Referencing `devenv` in production code feels too specific so maybe we can take a step back and look at when exactly we need `.raw.` to come up with a more generic rule.

Can you remind me why we need `.raw.` in the first place? Does the canister add certain headers when you don't use `.raw.` which are not added when you use `.raw.`? Or how exactly does `.raw.` avoid the CORS issues?

dskloetd: `.raw.` is now working on my DevEnv and should work on yours soon.

lmuntaner: If we think that static testnets are not coming back, we can remove `addRawToUrl`.

If `raw` will work, then that's great.

As you say, I would think again whether testnets are coming back or not. I guess we can do quite some clean up if not....

lmuntaner:
Closing this issue. This is fixed upstream

frontend/src/lib/constants/environment.constants.ts
dskloetd: But what's the difference between static testnets and devenv that raw is required on testnet but not on devenv?
Maybe it's just not required at all anymore and we just didn't notice it?
Or am I missing something?

lmuntaner: I also don't know. It was weird in the first place that `raw` was needed for the SNS aggregator in the first place...

PR #3709, opened 2023-11-10
Update sent/received icons

dskloetd:
# Motivation

We have new icons for sent/received transactions in the [ckBTC UI update design](https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=475-5782&mode=design&t=DZsyUIvKZoKJmlPj-4).
I asked Artem if these icons are specific to ckBTC or if we should use them everywhere and he said we should use them everywhere.

# Changes

1. Replace the diagonal arrows for sent/received transaction with up/down arrows.
2. The diagonal arrows were the same icons but differently rotated so I also removed the rotation.

# Tests

Old:
image

New:
image

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp-unreleased.md
+++ b/CHANGELOG-Nns-Dapp-unreleased.md
@@ -25,6 +25,7 @@ proposal is successful, the changes it released will be moved from this file to
 #### Changed
 
 * Read schema version from stable memory.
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Thanks!

PR #3710, opened 2023-11-10
GIX-1979: Load ckBTC tokens

lmuntaner:
# Motivation

We want to show the tokens table also for non-logged in users.

Therefore, we need to fetch the token information about CkBTC.

In this PR, I introduce a service `loadCkBTCTokens` that will be used for the tokens route to load the token data.

# Changes

* New ckBTC token service `loadCkBTCTokens`.

# Tests

* Test new service

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/services/ckbtc-tokens.services.ts
+++ b/frontend/src/lib/services/ckbtc-tokens.services.ts
@@ -62,13 +62,16 @@ export const loadCkBTCToken = async ({
       if (!certified && notForceCallStrategy()) {
         return;
       }
+
       // Explicitly handle only UPDATE errors
       toastsError({
         labelKey: "error.token_not_found",
         err,
       });
+
       // Hide unproven data
       tokensStore.resetUniverse(universeId);
+
       handleError?.();
     },
   });
--- a/frontend/src/lib/services/ckbtc-tokens.services.ts
+++ b/frontend/src/lib/services/ckbtc-tokens.services.ts
@@ -45,7 +45,7 @@ export const loadCkBTCToken = async ({
 
   return queryAndUpdate<IcrcTokenMetadata, unknown>({
     strategy: FORCE_CALL_STRATEGY,
-    identityType: "anonymous",
+    identityType: "current",
     request: ({ certified, identity }) =>
       getCkBTCToken({
         identity,
--- a/frontend/src/lib/services/ckbtc-tokens.services.ts
+++ b/frontend/src/lib/services/ckbtc-tokens.services.ts
@@ -1,12 +1,33 @@
 import { getCkBTCToken } from "$lib/api/ckbtc-ledger.api";
+import {
+  CKBTC_UNIVERSE_CANISTER_ID,
+  CKTESTBTC_UNIVERSE_CANISTER_ID,
+} from "$lib/constants/ckbtc-canister-ids.constants";
 import { FORCE_CALL_STRATEGY } from "$lib/constants/mockable.constants";
-import { queryAndUpdate } from "$lib/services/utils.services";
+import {
+  ENABLE_CKBTC,
+  ENABLE_CKTESTBTC,
+} from "$lib/stores/feature-flags.store";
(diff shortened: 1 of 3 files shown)


lmuntaner:
* main
  * **PR #3710** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3710?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3710?utm_source=stack-comment).

frontend/src/lib/services/ckbtc-tokens.services.ts
```diff
@@ -24,6 +45,7 @@ export const loadCkBTCToken = async ({
 
   return queryAndUpdate<IcrcTokenMetadata, unknown>({
     strategy: FORCE_CALL_STRATEGY,
+    identityType: "current",
```
lmuntaner, line 48: The default is `authorized` and we also want to call this when users are not logged in.

The endpoint doesn't care about the caller anyway.

lmuntaner:
@dskloetd please review

frontend/src/tests/lib/services/ckbtc-tokens.services.spec.ts
```diff
@@ -73,4 +72,95 @@ describe("ckbtc-tokens-services", () => {
...
+      );
+    });
+
+    describe("no ckBTC enabled", () => {
+      beforeEach(() => {
+        overrideFeatureFlagsStore.setFlag("ENABLE_CKBTC", false);
+        overrideFeatureFlagsStore.setFlag("ENABLE_CKTESTBTC", false);
+      });
+
+      it("should load the ckTESTBTC token if enabled", async () => {
```
dskloetd, line 100: What does "if enabled" mean here in the `describe` group with both disabled?

Similar for the 3 tests below.

frontend/src/lib/services/ckbtc-tokens.services.ts
```diff
@@ -1,12 +1,33 @@
...
+} from "$lib/stores/feature-flags.store";
 import { toastsError } from "$lib/stores/toasts.store";
 import { tokensStore } from "$lib/stores/tokens.store";
 import type { IcrcTokenMetadata } from "$lib/types/icrc";
 import type { UniverseCanisterId } from "$lib/types/universe";
 import { notForceCallStrategy } from "$lib/utils/env.utils";
 import { get } from "svelte/store";
+import { queryAndUpdate } from "./utils.services";
+
+export const loadCkBTCTokens = async () => {
```
dskloetd, line 19: This is fine to have but I can't help but wonder if the table should have a list of all tokens it wants to show and load all of them, regardless of whether they are ckbtc or other kinds of tokens?

Is there anything ckBTC specific about the `loadCkBTCToken` function or would it work with any universeId that's also a ICRC-1 ledger canister ID? Probably easier said than done but worth keeping in mind if you see an opportunity.

I think the SNS wrappers might be holding us back and identifying the SNS ledgers by their ledger canister ID instead of root canister ID might allow for a lot of streamlining.

lmuntaner: That's a good point.

I think that the most maintainable way is to decouple loading the data from rendering the data.

This way, when we want to add tokens manually, we could add a service to load them, and then a derived store would read from where they are stored and then convert them to the data needed by the component.

Another idea was to use the callback when stores are initialized to load data. But I believe that was discarded and the loading services was the preferred pattern.

frontend/src/tests/lib/services/ckbtc-tokens.services.spec.ts
lmuntaner: Oops, that was before I had the current structure of `describe`.

I changed it.

dskloetd (approved):

PR #3713, opened 2023-11-10
Global network

bitdivine:
# Motivation

# Changes

# Tests

# Todos

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -4,6 +4,8 @@ set -euo pipefail
 export PATH="$PWD/scripts:$PATH"
 
 cd "$(dirname "$(realpath "$0")")" || exit
+# Get commands for network configuration.
+. scripts/network-config
 
 help_text() {
   cat <<-"EOF"
@@ -56,10 +58,7 @@ while (($# > 0)); do
   *)
     DFX_NETWORK="$env"
     # Check that the network is valid.
-    DFX_NETWORK="$env" jq -e '.networks[env.DFX_NETWORK]' dfx.json || {
-      echo "ERROR: Network '$env' is not listed in dfx.json"
-      exit 1
-    } >&2
+    assert_dfx_network_var_is_configured || exit 1
     ;;
   esac
 done
@@ -82,10 +81,6 @@ first_not_null() {
     fi
   done
   echo "null"
-}
-
-static_host() {
-  jq -re '.networks[env.DFX_NETWORK].config | .STATIC_HOST // .HOST' dfx.json
 }
 
 canister_static_url_from_id() {
--- a/scripts/propose
+++ b/scripts/propose
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 cd "$(dirname "$(realpath "$0")")/.."
+. scripts/network-config
 PROPOSALS_DIR="scripts/propose-to"
 PATH="$PWD/target/ic:$PATH"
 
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -87,6 +87,7 @@ RUN cargo binstall --no-confirm "ic-wasm@$(cat config/ic_wasm_version)" && comma
 FROM builder AS configurator
 SHELL ["bash", "-c"]
 COPY dfx.json config.sh canister_ids.jso[n] /build/
+COPY scripts/network-config /build/scripts/network-config
```
lmuntaner, line 90: Why is this needed in the docker build? Because of the `config.sh`?

lmuntaner: Do you want to set the global networks in the docker build also?

deploy.sh
```diff
@@ -84,10 +83,6 @@ first_not_null() {
   echo "null"
 }
 
-static_host() {
```
lmuntaner, line 87: Is this part of this PR? Or just that you realized it wasn't used and removed it?

scripts/propose
```diff
@@ -98,16 +99,7 @@ while (($# > 0)); do
...
-      cat <<-EOF
-
-	ERROR: No such network in dfx.json.
-	       Available networks are:
-	       $(jq -er '.networks | keys | join(" ")' dfx.json)
-
-	EOF
-      exit 1
-    }
+    DFX_NETWORK="$DFX_NETWORK" assert_dfx_network_var_is_configured || exit 1
```
lmuntaner, line 102: Ditto that this comes from `network-config`

scripts/sns/swap/start_swap
```diff
@@ -95,7 +81,7 @@ while (($# > 0)); do
 done
 
 dfx_network_check_specified
-dfx_network_check_valid
+DFX_NETWORK="$DFX_NETWORK" assert_dfx_network_var_is_configured || exit 1
```
lmuntaner, line 84: Ditto that this comes from `network-config`.

Dockerfile
bitdivine: > Why is this needed in the docker build? Because of the config.sh?

Yes, exactly.

Global networks are another kettle of fish as the global config has to be available for reproducible builds by people out in the wild.  We can't expect people to set a global config so we need to use defaults and local data.  The local data doesn't have to be in `dfx.json` in the current location.  However moving things around is outside the scope of this PR.  This PR allows us to use other networks but doesn't go as far as deleting local networks.

deploy.sh
```diff
@@ -56,10 +58,7 @@ while (($# > 0)); do
   *)
     DFX_NETWORK="$env"
     # Check that the network is valid.
-    DFX_NETWORK="$env" jq -e '.networks[env.DFX_NETWORK]' dfx.json || {
-      echo "ERROR: Network '$env' is not listed in dfx.json"
-      exit 1
-    } >&2
+    assert_dfx_network_var_is_configured || exit 1
```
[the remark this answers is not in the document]
bitdivine: Done.

deploy.sh
bitdivine: This has been moved to `network_config`.  I actually expected it to be needed repeatedly but it is used just once so it can go back, if you want, in its updated form.

scripts/propose
bitdivine: Done

scripts/sns/swap/start_swap
bitdivine: Done

Dockerfile
bitdivine: Let me look into this a bit more.

bitdivine: If I recall correctly docker cannot access files outside of the local directory, so e.g. `COPY ~/.config/dfx/networks` would not work but we could read the global config into a variable or local file and then, inside docker, write the global config in the configuration stage.  Let me look into that a bit more.

bitdivine: Testing an implementation of "copy  gloabl config into dockerfile".  I'll add it to this config when I'm done.

lmuntaner: Thanks!

deploy.sh
lmuntaner: No, it's ok. I wanted to understand the change. Thanks!

lmuntaner (approved):
LGTM! Thanks

Dockerfile
bitdivine: Done.  Tested locally.  I will add some tests, but maybe in another PR now as by now this is already quite big.

PR #3731, opened 2023-11-13
GIX-1979: TokensListBase derived store

lmuntaner:
# Motivation

# Changes

# Tests

# Todos


lmuntaner:
* main
  * **PR #3731** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3731?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3731?utm_source=stack-comment).

lmuntaner:

frontend/src/lib/derived/tokens-list-base.derived.ts
```diff
@@ -0,0 +1,63 @@
...
+  isUniverseCkTESTBTC,
+  isUniverseNns,
+  isUniverseRealCkBTC,
+} from "$lib/utils/universe.utils";
+import { Principal } from "@dfinity/principal";
+import { isNullish, nonNullish } from "@dfinity/utils";
+import { derived, get, type Readable } from "svelte/store";
+import { universesStore } from "./universes.derived";
+
+const convertUniverseToVisitorTokenData =
```
dskloetd, line 23: Does "visitor" mean not signed-in? Is this existing terminology or did you invent it here for the first time?

frontend/src/lib/derived/tokens-list-base.derived.ts
```diff
@@ -0,0 +1,63 @@
...
+      ? i18nKeys.ckbtc.title
+      : universe.summary?.metadata.name;
+    const token = tokensData[universe.canisterId]?.token;
+    const logo = isUniverseNns(universeId)
+      ? IC_LOGO_ROUNDED
+      : isUniverseCkTESTBTC(universeId)
+      ? CKTESTBTC_LOGO
+      : isUniverseRealCkBTC(universeId)
+      ? CKBTC_LOGO
+      : universe.summary?.metadata.logo;
```
dskloetd, line 42: Hard coding the title and logo here doesn't feel right. Could we give NNS and ckBTC a summary as well? Or at least put these `?:` chains in separate utility functions?

frontend/src/lib/derived/tokens-list-base.derived.ts
```diff
@@ -0,0 +1,63 @@
...
+    return {
+      universeId,
+      title,
+      balance: new UnavailableTokenAmount(token),
+      logo,
+      actions: [],
+    };
+  };
+
+export const tokensListBaseStore = derived<
```
dskloetd, line 56: What is "base" indicating here? Will there also be a `tokensListStore`?

frontend/src/lib/derived/tokens-list-base.derived.ts
lmuntaner: I wanted to call it like this initially, that's why the converter is called like this. But then I changed to call it "Base".

I'll rename the helper.

Thanks for asking!

frontend/src/lib/derived/tokens-list-base.derived.ts
lmuntaner: We would need a common "summary" type or even the universe not having any summary at all, but all the needed fields instead.

I can work on it in another PR. For now, I'll move it to a helper.

frontend/src/lib/derived/tokens-list-base.derived.ts
lmuntaner: Yes, correct. The derived store that I mention in the description:

> The derived store when the user is logged in.

dskloetd: What does "the derived store" mean? Which derived store?

lmuntaner: A Svelte store created with `derived`. A derived store doesn't hold data, instead, it uses other stores to combine or transform data.

lmuntaner: This store will be used for visitors, because it has only the universes data, no user data (balance).

Another task will be to create the derived store with the user data. This other derived store will combine `tokensListBaseStore` and the different account stores to return the tokens with the balance.

And it will also add the user actions.

lmuntaner:

dskloetd (approved):

PR #3732, opened 2023-11-13
feat: Separate branches everywhere

bitdivine:
# Motivation
Keep bots consistent.  In #3703, one bot was changed to create a separate branch and PR every time the bot was run.

Note: I am concerned that this will make the bots spammier, which in turn will want developers to run the bots less frequently, which in turn will increase the time between e.g. a new proposal type appearing in the nns and being supported in the nns-dapp.  But we can see and revert if necessary.

# Changes
- Convert all the nns-dapp bots to branch-per-run.

# Tests
None

# Todos


lmuntaner (approved):
Thanks!

PR #3735, opened 2023-11-13
Add mapCkbtcTransaction to decode cbor memo

dskloetd:
# Motivation

In the ckBTC wallets, we want to render burn transactions as withdrawal transactions.
To accomodate this, https://github.com/dfinity/nns-dapp/pull/3701 adds a `mapTransaction` parameter to `IcrcTransactionsList` such that we can use a different transaction mapping function per token.
This PR introduces the alternative mapping function for ckBTC.
But it is not yet used. It will be used in a follow-up PR.

# Changes

1. Install `cbor-web` to be able to decode cbor encoded memos on ckBTC burn transactions.
2. Add `mapCkbtcTransaction` which behaves like `mapIcrcTransaction` for most transactions but decodes the withdrawal address from the memo for burn transactions.

# Tests

1. Add `createBurnTransaction` to conveniently create burn transactions with specific properties.
2. Create a function `testMapTransactionCommon` which adds all the tests that are identical between `mapIcrcTransaction` and `mapCkbtcTransaction` to a `describe` block.
3. Add specific additional tests for burn transactions.

# Todos

Will add when it's used.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -28,7 +28,6 @@
         "@ledgerhq/hw-transport-webhid": "^6.27.8",
         "@zondax/ledger-icp": "^0.6.1",
         "buffer": "^6.0.3",
-        "cbor-web": "^9.0.1",
         "marked": "^9.1.0"
       },
       "devDependencies": {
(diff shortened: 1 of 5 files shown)


dskloetd:
I'm looking at using a different cbor library. You can wait with reviewing until after that or already review. I expect almost all the code to remain unchanged.

lmuntaner (approved):
LGTM!

Did you check for vulnerabilities in the new npm package?

dskloetd:
> Did you check for vulnerabilities in the new npm package?

In the end I was able to use `Cbor` exported by `agent-js` after all. So no new packages needed!

PR #3742, opened 2023-11-14
Fix add-paths lines

bitdivine:
# Motivation
The snsdemo repo was added accidentally.  (Not the whole repo but as a link.)

This looks as if it was caused by an incorrect pattern.

# Changes
- Remove the spurious directory.
- Fix incorrect patterns.

# Tests
We can trigger workflows but given that the workflows do nothing most of the time, the only real test will be to wait for changes that cause non-trivial changes and check whether the bot creates correct outputs.

# Todos

### DIFF
--- a/.github/workflows/update-ic.yml
+++ b/.github/workflows/update-ic.yml
@@ -78,6 +78,7 @@ jobs:
         with:
           token: ${{ secrets.GIX_BOT_PAT }}
           base: main
+          add-paths: dfx.json
           commit-message: Update ic candid files
           committer: GitHub <noreply@github.com>
           author: gix-bot <gix-bot@users.noreply.github.com>
--- a/.github/workflows/update-ic.yml
+++ b/.github/workflows/update-ic.yml
@@ -78,6 +78,7 @@ jobs:
         with:
           token: ${{ secrets.GIX_BOT_PAT }}
           base: main
+          # Note: Please be careful when updating the add-paths field.  We have had the snsdemo committed by accident, with a pattern that matches nothing seemingly committing everything.
           add-paths: |
             dfx.json
             declarations/*/*.did
--- a/.github/workflows/update-ic.yml
+++ b/.github/workflows/update-ic.yml
@@ -78,7 +78,7 @@ jobs:
         with:
           token: ${{ secrets.GIX_BOT_PAT }}
           base: main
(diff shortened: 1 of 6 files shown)


.github/workflows/update-proposals.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           author: gix-bot <gix-bot@users.noreply.github.com>
           branch: bot-proposals-update
           branch-suffix: timestamp
-          add-paths: .github/*/*ml
+          add-paths: rs/proposals/src/canisters/*/api.rs
```
bitdivine, line 39: This bot should only be generating rust API code for proposal rendering.

Example: None, as this bot hasn't made any PRs yet.

.github/workflows/update-snsdemo.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           author: gix-bot <gix-bot@users.noreply.github.com>
           branch: bot-snsdemo-update
           branch-suffix: timestamp
-          add-paths: .github/*/*ml
+          add-paths: dfx.json
```
bitdivine, line 67: The bot should only be updating the snsdemo version in dfx.json 

Example: https://github.com/dfinity/nns-dapp/pull/3717/files

.github/workflows/update-ic.yml
```diff
@@ -78,6 +78,9 @@ jobs:
         with:
           token: ${{ secrets.GIX_BOT_PAT }}
           base: main
+          add-paths: |
+            dfx.json
+            declarations/*/*.did
           commit-message: Update ic candid files
```
bitdivine, line 84: This should be updating only the IC version in dfx.json and the .did files associated with that commit.

Example: https://github.com/dfinity/nns-dapp/pull/3719/files

nmattia (approved):
LGTM!

You could nonetheless check out the snsdemo repo in a different direction to be safe and avoid anything being overwritten.

Thought: the comments you've added here on GitHub could be valuable in the code as well

bitdivine:
@nmattia I've started a PR to move the code.  It looks as if to checkout side by side, I also have to change the way the main repo is checked out.  https://github.com/actions/checkout#checkout-multiple-repos-side-by-side so that will be a slightly bigger PR.

PR #3749, opened 2023-11-14
Better logging for update actions

bitdivine:
# Motivation
Update actions have insufficient logging for rapid debugging.

# Changes
- Log the git status, so that the log can be compared with the the PR generated by the action.
- Include the version being updated to in the PR title, where applicable, to make it easier to find.

# Tests

# Todos

### DIFF
--- a/.github/workflows/update-ic.yml
+++ b/.github/workflows/update-ic.yml
@@ -42,7 +42,7 @@ jobs:
         run: |
           if test -d snsdemo
           then echo have_snsdemo=true >> "$GITHUB_OUTPUT"
-          else have_snsdemo=false >> "$GITHUB_OUTPUT"
+          else echo have_snsdemo=false >> "$GITHUB_OUTPUT"
           fi
       - name: Determine snsdemo ref
         if: ${{ steps.have_snsdemo.outputs.have_snsdemo }} == 'false'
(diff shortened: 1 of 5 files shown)


.github/workflows/update-ic.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         run: |
           if test -d snsdemo
           then echo have_snsdemo=true >> "$GITHUB_OUTPUT"
-          else have_snsdemo=false >> "$GITHUB_OUTPUT"
+          else echo have_snsdemo=false >> "$GITHUB_OUTPUT"
```
bitdivine, line 45: Bug: Missing echo

.github/workflows/update-snsdemo.yml
```diff
@@ -49,17 +49,17 @@ jobs:
...
           # Update
           ./scripts/update-snsdemo --dir ./snsdemo --release "${{ steps.update.outputs.release }}" --verbose
+          # Show changes
+          echo "Git status:"
+          git status
           echo "Changes:"
           git diff
-          git config --global user.email "gix-bot@users.noreply.github.com"
-          git config --global user.name "GIX bot"
-          git commit -a -m "Update snsdemo to ${{ steps.update.outputs.release }}"
```
bitdivine, line 56: Not needed - the action below makes a commit.  Also a deviation from the standard used in all these actions is going to lead to surprises, so should be avoided.

lmuntaner (approved):
LGTM! Thanks

PR #3750, opened 2023-11-14
GIX-1979: Login action visitors

lmuntaner:
# Motivation

Tokens table is also interactive for non logged in users.

The requirement was to send them to login.

# Changes

* New derived store for the tokens table: `tokensListVisitorsStore`.
* Listen for the event `nnsAction` in the Tokens route and SignInTokens.
* Trigger a login in any `nnsAction` if the user is not logged in.

# Tests

* Add a test in the tokens route that the user is sent to login when clicking in a row.
* Test new store `tokensListVisitorsStore`.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/routes/(app)/(nns)/tokens/+page.svelte
+++ b/frontend/src/routes/(app)/(nns)/tokens/+page.svelte
@@ -15,8 +15,9 @@
   import { UserTokenAction, type UserTokenData } from "$lib/types/tokens-page";
   import type { Action } from "$lib/types/actions";
   import { UnavailableTokenAmount } from "$lib/utils/token.utils";
-  import { tokensListBaseStore } from "$lib/derived/tokens-list-base.derived";
   import { loadCkBTCTokens } from "$lib/services/ckbtc-tokens.services";
+  import { tokensListVisitorsStore } from "$lib/derived/tokens-list-visitors.derived";
+  import { login } from "$lib/services/auth.services";
 
   onMount(() => {
     if (!$ENABLE_MY_TOKENS) {
@@ -42,9 +43,11 @@
     },
   ];
 
-  const handleAction = ({ detail }: { detail: Action }) => {
-    console.log("action", detail.type);
-    console.log(detail.data);
+  const handleAction = ({ detail: _ }: { detail: Action }) => {
+    // Any action from non-signed in user should be trigger a login
+    if (!$authSignedInStore) {
+      login();
+    }
   };
 </script>
 
@@ -52,6 +55,9 @@
   {#if $authSignedInStore}
     <Tokens userTokensData={data} on:nnsAction={handleAction} />
   {:else}
-    <SignInTokens userTokensData={$tokensListBaseStore} />
+    <SignInTokens
+      on:nnsAction={handleAction}
+      userTokensData={$tokensListVisitorsStore}
+    />
   {/if}
 </TestIdWrapper>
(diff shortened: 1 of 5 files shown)


lmuntaner:
* main
  * **PR #3750** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3750?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3750?utm_source=stack-comment).

lmuntaner:
New UX:

[uploading Screen Recording 2023-11-14 at 15.53.54.mov...]

lmuntaner:
New UX:

            <a href="https://app.graphite.dev/media/video/W99H6e2WWjPgtes6VOJI/4276508e-e17c-4a93-b61b-591e40c96c24.mov">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/W99H6e2WWjPgtes6VOJI/4276508e-e17c-4a93-b61b-591e40c96c24.mov">Screen Recording 2023-11-14 at 15.53.54.mov</video>

lmuntaner:

dskloetd:
Do we really want this? This honestly looks pretty irritating to me. There is a big sign in button at the top of the page, which I can press if I want to sign in. If the rows are also clickable, I'm going to assume they'll give me more information about the token or something. What's the value in having the entire table have the same functionality as the sign-in button which is already there?

lmuntaner:
> Do we really want this? This honestly looks pretty irritating to me. There is a big sign in button at the top of the page, which I can press if I want to sign in. If the rows are also clickable, I'm going to assume they'll give me more information about the token or something. What's the value in having the entire table have the same functionality as the sign-in button which is already there?

Irritating? Interesting. I wouldn't have thought that way.

There are plenty of places that show user functionality just to redirect to the login page.

We can always change if users complain, but ideally we would have done some user testing instead of the team deciding on this functionality.

@artkorotkikh-dfinity what do you think?

dskloetd:
> There are plenty of places that show user functionality just to redirect to the login page.

The only one that comes to mind is swap participation but there the button clearly says "Sign in to participate".
Going through the UI, I don't see any others.
Can you give some examples where it looks functional to begin with but then redirects to a sign-in instead?

artkorotkikh-dfinity:
@lmuntaner, @dskloetd my original idea was to make the 'token details' page accessible to non-signed-in users as well, and to include a login button there for better continuity. However, I then realized that prompting them to sign in might be a viable short-term solution without significantly increasing the scope.

I discussed this with Mary, and she was okay with the suggestion. Can we try this and see if we receive similar feedback about it being irritating?

dskloetd (approved):

frontend/src/routes/(app)/(nns)/tokens/+page.svelte
```diff
@@ -40,16 +41,21 @@
     },
   ];
 
-  const handleAction = ({ detail }: { detail: Action }) => {
-    console.log("action", detail.type);
-    console.log(detail.data);
+  const handleAction = ({ detail: _ }: { detail: Action }) => {
+    // Any action from non-signed in user should be trigger a login
+    if (!$authSignedInStore) {
+      login();
```
dskloetd, line 47: How do we actually open the requested token details once the login is complete?

PR #3757, opened 2023-11-15
Change Source: to From:

dskloetd:
# Motivation

The opposite of "To" is "From".
The opposite of "Source" is "Destination".
Currently we write "To: address" for Sent transaction but "Source: address" to Received transactions.
But we also write "From: BTC Network" rather than "Source: BTC Network" for ckBTC mint transactions.
And new UI mocks (here) use "From:" everywhere.

# Changes

Change "Source:" to "From:" for consistency.

# Tests

Existing tests are updated.

# Todos

### DIFF
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -484,7 +484,7 @@
     "title": "Account",
     "address": "Address",
     "principal": "Principal",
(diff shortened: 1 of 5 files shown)


lmuntaner (approved):
I agree, thanks!

PR #3758, opened 2023-11-15
GIX-1987: Load Tokens User Data

lmuntaner:
# Motivation

Render the tokens of the logged in user.

In this PR, we load the balances of SNS and ckBTC projects in the tokens route.

I used the same way that they are loaded in accounts. It has been working well.

# Changes

* Load SNS and ckBTC balances in tokens route.

# Tests

* Add a test to Tokens route that checks balances of the user when logged in.
* Add various methods to tokens table related page objects.

# Todos

Not yet necessary.


lmuntaner:
* main
  * **PR #3758** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3758?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3758?utm_source=stack-comment).

lmuntaner:
Same UI, but loading data:

lmuntaner:

dskloetd:
> I used the same way that they are loaded in accounts. It has been working well.

Would it make sense to share the code instead of duplicating it?

frontend/src/tests/page-objects/TokensTableRow.page-object.ts
```diff
@@ -20,6 +20,12 @@ export class TokensTableRowPo extends BasePageObject {
     return this.getText("token-value-label");
   }
 
+  async getData(): Promise<Record<string, string>> {
```
dskloetd, line 23: Why `Record<string, string>` and not `{ projectName: string; balance: string }`?

frontend/src/tests/routes/app/tokens/page.spec.ts
```diff
@@ -34,6 +42,33 @@ describe("Tokens route", () => {
...
+      );
+      vi.spyOn(snsLedgerApi, "getSnsAccounts").mockImplementation(
+        async ({ rootCanisterId }) => {
+          if (rootCanisterId.toText() === rootCanisterId1.toText()) {
+            return [mockSnsMainAccount];
+          }
+          return [
+            {
+              ...mockSnsMainAccount,
+              balanceE8s: 314000000n,
```
dskloetd, line 56: I'm missing something. Can you explain what's going on here?
Why do we set the balance except for a specific canisterId?
Do we want to set a different balance for `rootCanisterId1`?
Why should the balance (but nothing else) depend on the canisterId?

frontend/src/tests/page-objects/TokensTableRow.page-object.ts
lmuntaner: I wanted it to be generic...

But it makes sense to be specific actually.

frontend/src/tests/routes/app/tokens/page.spec.ts
lmuntaner: Yes, I want to set a different balance based on rootCanisterId.

the account is:
```
  identifier: encodeIcrcAccount({
    owner: mockPrincipal,
  }),
  balanceE8s: 890156712340000n,
  principal: mockPrincipal,
  type: "main",
```

Nothing is specific to the SNS. The difference is more where it's located, than in the data itself.

I want a difference balance, just to test that it works with multiple SNSes.

lmuntaner:

lmuntaner:
> Would it make sense to share the code instead of duplicating it?

I will be removing the code there as part of the clean up once the feature flag is enabled in prod.

frontend/src/tests/routes/app/tokens/page.spec.ts
lmuntaner: Yes, good point. I did the same for ckbtc and icp also.

frontend/src/tests/routes/app/tokens/page.spec.ts
```diff
@@ -43,6 +51,33 @@ describe("Tokens route", () => {
...
+            },
+          ];
+        }
+      );
+      vi.spyOn(snsLedgerApi, "getSnsToken").mockImplementation(
+        async ({ rootCanisterId }) => {
+          if (rootCanisterId.toText() === rootCanisterId1.toText()) {
+            return mockSnsToken;
```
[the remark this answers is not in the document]
lmuntaner: done

lmuntaner:

dskloetd (approved):

PR #3760, opened 2023-11-15
fix: countdown label visibility

mstrasinskis:
# Motivation

Remove the label when there is no active countdown.

# Changes

- expose countdown visible state
- limit label visibility

# Tests

Tested manually
| Before | After |
|--------|--------|
| image | image |

# Todos


frontend/src/lib/components/proposal-detail/VotesResults.svelte
```diff
@@ -36,11 +38,16 @@
...
+        {#if isCountdownVisible}
+          <span class="caption description"
+            >{$i18n.proposal_detail__vote.expiration}</span
+          >
+        {/if}
         <div class="caption value">
-          <Countdown {deadlineTimestampSeconds} />
+          <Countdown
+            {deadlineTimestampSeconds}
+            bind:visible={isCountdownVisible}
```
lmuntaner, line 49: Wouldn't it be better to not rely on the value inside here, and set it at this level instead?

```
$: isCountdownVIsible = deadlineTimestampSeconds > nowInSeconds()
```

I'm guessing that's the logic...

frontend/src/lib/components/proposal-detail/VotesResults.svelte
lmuntaner, line 1: Can you add a test for this also?

frontend/src/lib/components/proposal-detail/VotesResults.svelte
mstrasinskis: Good point! Updated.

frontend/src/lib/components/proposal-detail/VotesResults.svelte
mstrasinskis: Sure ✔️

lmuntaner (approved):
Thanks!

PR #3763, opened 2023-11-15
chore: Bump account threshold limit

bitdivine:
# Motivation
A limit is set on the maximum number of accounts that may be created.  We are approaching that limit.  The limit is very conservative and safe to raise.

# Changes
- Increase the account limit

# Tests
- The tests have been updated to simulate future behaviour near the limit, but not actually on the limit as that can cause spurious errors where test account creation fails.

# Todos


lmuntaner (approved):
Thanks!

PR #3764, opened 2023-11-15
fix: stuck loading of the sns proposal

mstrasinskis:
# Motivation

Fix sns proposal initial update. For undiscovered reason data changes in `SnsProposalDetail` don't trigger auto-subsribers (maybe the lazy loading of the component on slow internet provokes this behaviour)

# Changes

- wait for component finished current updates before applying new data changes.

# Tests

Manually tested with network throttling.

# Todos


peterpeterparker (approved):
LGTM, thx

We debugged this together and I also concluded that the issue can either be solved with a `tick` or by imperatively updating all the states within a callback because an auto-subscriber was not triggered.

PR #3768, opened 2023-11-15
Fix deploy to app

bitdivine:
# Motivation
Test deployments to the app subnet currently fail.

# Changes
- Add logging
- Change provider to match what dfx 15 expects.  See [this thread on the forum](https://forum.dfinity.org/t/replica-error-not-allowed-to-call-ic00-method-provisional-create-canister-with-cycles/23709/6) for details.

# Tests
- See this deployment: https://github.com/dfinity/nns-dapp/actions/runs/6879490121

# Todos

### DIFF
--- a/dfx.json
+++ b/dfx.json
@@ -213,7 +213,7 @@
         }
       },
       "providers": [
-        "https://icp-api.io/"
+        "https://icp0.io"
       ],
       "type": "persistent"
     },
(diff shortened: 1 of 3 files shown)


dfx.json
```diff
@@ -213,7 +213,7 @@
         }
       },
       "providers": [
-        "https://icp-api.io/"
+        "https://icp0.io"
```
bitdivine, line 216: dfx now treats this URL as mainnet.  The app subnet is also affected.

lmuntaner (approved):
Thanks!

PR #3769, opened 2023-11-15
feat: do not hide voting buttons after voting

mstrasinskis:
# Motivation

Keep adapt/reject buttons visible but disabled after voting. Part of voting section redesign.

# Changes

- update visibility logic for nns and sns proposals

# Tests

Works

# Todos

# Screenshots

image


lmuntaner (approved):
I guess that the disabling of the buttons is already somewhere else, right?

mstrasinskis:
> I guess that the disabling of the buttons is already somewhere else, right?
Correct. It is inside `VotingConfirmationToolbar` and it's logic already covers removed from visibility cases.

PR #3771, opened 2023-11-16
Error if network is defined but not exported

bitdivine:
# Motivation
`DFX_NETWORK` is typically provided to dfx build commands as an exported environment variable.  When checking that `DFX_NETWORK` is provided, we need to check specifically that it is exported as otherwise it is possible to get very confusing behaviour in which the network is defined in one part of the deployment but not another.

# Changes
- Check that `DFX_NETWORK` is exported.

# Tests
- Tests have been added that the error message specifically mentions `DFX_NETWORK` not being exported.

# Todos


lmuntaner (approved):
Thanks!

PR #3772, opened 2023-11-16
Fix local deployments

bitdivine:
# Motivation
With dfx 0.15.1 this fails with a confusing error message:
```
./deploy.sh --nns-dapp local
```
This is due to the network being defined but not exported.

# Changes
- Export `DFX_NETWORK`.

# Tests
* Tested locally.  CI tests are underway but I am unlikely to get them finished this morning so this is to unblock developers.
* Fixes and tests for appropriate error messages are in #3771 

# Todos


lmuntaner (approved):
Thanks!

PR #3773, opened 2023-11-16
Replace secondsToDuration with utils move to ic-js

peterpeterparker:
# Motivation

`secondsToDuration` has been moved to `@dfinity/utils` therefore this PR replaces the usage of the function by the one in the library. It also remove the duplicate implementation within NNS dapp which becomes unused.

# PRs

- [x] Function moved in https://github.com/dfinity/ic-js/pull/467


mstrasinskis (approved):
Thank you!

PR #3775, opened 2023-11-16
GIX-1980: Setup icp accounts with tokens table

lmuntaner:
# Motivation

When we migrate to the tokens page, we also want a similar view for the ICP accounts, subaccounts and hardware wallet accounts.

In this PR, I introduce the TokensTable for NNS Accounts.

# Changes

* Render `TokensTable` in `NnsAccounts` depending on feature flag.
* Expect an optional list of `UserTokenData` in `NnsAccounts`.
* Hard code a list of `UserTokenData` in the Accounts route and pass it to `NnsAccounts`.

# Tests

* Add test in NnsAccounts.spec that the tokens table is there if feature flag is enabled. I added myself some TODOs comments to remember improving the test before adding more features here.
* Add methods to NnsAccountsPo.
* Add test in Accounts route to check that tokens table is rendered if flag is enabled.

# Todos

### DIFF
--- a/frontend/src/tests/lib/pages/NnsAccounts.spec.ts
+++ b/frontend/src/tests/lib/pages/NnsAccounts.spec.ts
@@ -3,6 +3,7 @@ import * as nnsDappApi from "$lib/api/nns-dapp.api";
 import { SYNC_ACCOUNTS_RETRY_SECONDS } from "$lib/constants/accounts.constants";
 import NnsAccounts from "$lib/pages/NnsAccounts.svelte";
 import { cancelPollAccounts } from "$lib/services/icp-accounts.services";
+import { overrideFeatureFlagsStore } from "$lib/stores/feature-flags.store";
 import { icpAccountsStore } from "$lib/stores/icp-accounts.store";
 import { formatToken } from "$lib/utils/token.utils";
 import { resetIdentity } from "$tests/mocks/auth.store.mock";
@@ -12,6 +13,8 @@ import {
   mockMainAccount,
   mockSubAccount,
 } from "$tests/mocks/icp-accounts.store.mock";
+import { NnsAccountsPo } from "$tests/page-objects/NnsAccounts.page-object";
+import { JestPageObjectElement } from "$tests/page-objects/jest.page-object";
 import {
   advanceTime,
   runResolvedPromises,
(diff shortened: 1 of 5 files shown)


lmuntaner:
* main
  * **PR #3775** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3775?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3775?utm_source=stack-comment).

lmuntaner:
New UI:

lmuntaner:

dskloetd:
I thought the point of the tokens table was to have all accounts in a single table. Are we actually going to have a table per token? I'm confused.

lmuntaner:
Not per token, only for ICP.

This is how we will support subaccounts and hardware wallet (for now).

Hardware wallet will move to the main view in another iteration of the feature.

Subaccounts might move to the main view also if we add filtering to the main view. But filtering and sorting have their own problems.

Therefore, as an MVP, we agreed with product and design to have the subaccounts and HW in a separate table.

dskloetd:
Wait, do we never want to support subaccounts and hardware wallets for other tokens? I thought we just didn't have time for that yet. But now we are closing the door on having subaccounts and hardware wallets by making this design decision?

lmuntaner:
The opposite.

By doing this decision we can easily support subaccouts on other tokens.

`/accounts?u=...` will have this table view with accounts of only one token.

What I said is that it's not clear whether we want subaccounts here or all in a single view in the long term.

dskloetd:
> `/accounts?u=...` will have this table view with accounts of only one token.

So a table per token then, as I said in my first comment?

frontend/src/tests/lib/pages/NnsAccounts.spec.ts
```diff
@@ -23,14 +26,38 @@ vi.mock("$lib/api/nns-dapp.api");
...
+    overrideFeatureFlagsStore.setFlag("ENABLE_MY_TOKENS", false);
   });
 
-  afterEach(() => {
-    vi.clearAllMocks();
+  describe("when tokens flag is enabled", () => {
+    beforeEach(() => {
+      overrideFeatureFlagsStore.setFlag("ENABLE_MY_TOKENS", true);
+      // TODO: Return from mocked api layer instead of setting store directly
+      icpAccountsStore.setForTesting({
```
dskloetd, line 45: As far as I can see, the `NnsAccounts` component doesn't use the `icpAccountsStore` when the feature flag is enabled. Am I missing something or why is it necessary to fill the store here?

lmuntaner:
> So a table per token then, as I said in my first comment?

Not now because we won't support subaccounts. But in the future, yes. We could do that.

frontend/src/tests/lib/pages/NnsAccounts.spec.ts
lmuntaner: Because the component triggers `pollAccounts`, and I don't want that depending on the feature flag.

That's why I added the comment above. I will refactor this test file to mock the api layer instead of loading the store.

dskloetd: Thanks for the clarification. I didn't make the link from "mocking the API" to "poll accounts will not make requests if we populate the store in advance".

But now that you mention it, why not make `pollAccount` depend on the feature flag? With the feature flag the data will come from the input prop, rather than from the store, right?

lmuntaner: Yes good point. And it's the rout that should `pollAccounts` then. I'll do that.

lmuntaner: done

lmuntaner:

dskloetd (approved):

PR #3780, opened 2023-11-16
Pass a single prop to TransactionCard

dskloetd:
# Motivation

We currently pass `mapTransaction` from `CkbtcTransactionsList` to `IcrcTransactionsList` to `IcrcTransactionCard` and then we apply the mapping and pass a lot of props to `TransactionCard`.
My goal is to simplify this and give the higher level component more direct control over how the transactions are rendered.

In this PR I introduce the `UiTransaction` type which contains all information necessary to render the `TransactionCard` component. This allows us to pass a single prop to `TransactionCard`.
My goal is to eventually return `UiTransaction` from `mapTransaction` and possibly retire the intermediate `Transaction` type.
Also, I want to map all the transactions in the higher level component to avoid sending down all these different props.

This is the first step.

# Changes

1. Introduce a new type `UiTransaction` which has the information needed to render a transaction in the transaction list.
2. Change `TransactionCard` to accept 1 prop of type `UiTransaction` and remove the logic that is made obsolete by that.
3. Add `toUiTransaction` which converts `Transaction`, plus some extra params, to a `UiTransaction`.
4. Change `IcrcTransactionCard` and `NnsTransactionCard` to pass a `UiTransaction` to `TransactionCard` instead of a collection of props.

# Tests

Unit tests updated and pass.
Manually at: https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/types/transaction.ts
+++ b/frontend/src/lib/types/transaction.ts
@@ -84,7 +84,7 @@ export interface UiTransaction {
   isIncoming: boolean;
   headline: string;
   // Where the amount is going to or coming from.
-  otherParty: string;
+  otherParty?: string;
   // TODO: Remove fallbackDescription and always use otherParty.
   fallbackDescription?: string;
   // Always positive.
--- a/frontend/src/lib/types/transaction.ts
+++ b/frontend/src/lib/types/transaction.ts
@@ -83,11 +83,11 @@ export interface Transaction {
 export interface UiTransaction {
   isIncoming: boolean;
   headline: string;
-  // Where to amount is going to or coming from.
+  // Where the amount is going to or coming from.
   otherParty: string;
   // TODO: Remove fallbackDescription and always use otherParty.
   fallbackDescription?: string;
-  // Positing for incoming, negative for outgoing.
+  // Always positive.
   amount: bigint;
   token: Token;
   timestamp: Date;
--- a/frontend/src/lib/types/transaction.ts
+++ b/frontend/src/lib/types/transaction.ts
@@ -1,4 +1,5 @@
 import type { IcrcTransactionWithId } from "@dfinity/ledger-icrc";
+import type { Token } from "@dfinity/utils";
 import type { Account } from "./account";
 
 export type NewTransaction = {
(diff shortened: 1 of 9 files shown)


frontend/src/lib/types/transaction.ts
```diff
@@ -79,6 +80,19 @@ export interface Transaction {
...
 
+export interface UiTransaction {
+  isIncoming: boolean;
+  headline: string;
+  // Where the amount is going to or coming from.
+  otherParty?: string;
+  // TODO: Remove fallbackDescription and always use otherParty.
+  fallbackDescription?: string;
+  // Always positive.
+  amount: bigint;
```
lmuntaner, line 91: How about having the amount already in `TokenAmount`?

frontend/src/lib/types/transaction.ts
```diff
@@ -79,6 +80,19 @@ export interface Transaction {
   date: Date;
 }
 
+export interface UiTransaction {
+  isIncoming: boolean;
+  headline: string;
+  // Where the amount is going to or coming from.
+  otherParty?: string;
```
lmuntaner, line 87: Will this always be an identifier or address? Or do we leave it as `string` to have flexibility?

frontend/src/lib/utils/transactions.utils.ts
```diff
@@ -166,6 +167,41 @@ export const mapNnsTransaction = ({
   };
 };
 
+export const toUiTransaction = ({
```
lmuntaner, line 170: There are no tests for this new util.

frontend/src/lib/types/transaction.ts
dskloetd: I'm hoping to put "BTC Network" here instead of having to put "From: <span>BTC Network</span>" in the `fallbackDescription`. And then get rid of `fallbackDescription`.

frontend/src/lib/types/transaction.ts
dskloetd: done

frontend/src/lib/utils/transactions.utils.ts
dskloetd: oops, done

frontend/src/lib/types/transaction.ts
lmuntaner: Sounds good

lmuntaner (approved):
Thanks!

PR #3787, opened 2023-11-17
Update rust workflow text

bitdivine:
# Motivation
The title and body of PRs created by the Rust update action are bad.  [Example](https://github.com/dfinity/nns-dapp/pull/3781)

- Title tries to declare which version is updated to but the version number is missing
- The PR has no meaningful body

# Changes
- Fix the title (by making the version number available unconditionally)
- Add a body

# Tests
This PR was generated by this action: #3788

# Todos

### DIFF
--- a/.github/workflows/update-rust.yml
+++ b/.github/workflows/update-rust.yml
@@ -1,6 +1,6 @@
 # A GitHub Actions workflow that regularly checks for new Rust toolchain release
 # and creates a PR on new versions.
-name: Rust Update
+name: Update rust
 on:
   schedule:
     # check for new rust versions weekly
@@ -36,6 +36,7 @@ jobs:
           fi
 
           echo "latest rust version '$latest_rust_version'"
+          echo "version=$latest_rust_version" >> "$GITHUB_OUTPUT"
 
           if [ "$current_rust_version" != "$latest_rust_version" ]
           then
@@ -44,7 +45,6 @@ jobs:
             echo "updated=1" >> "$GITHUB_OUTPUT"
           else
             echo "updated=0" >> "$GITHUB_OUTPUT"
-            echo "version=$latest_rust_version" >> "$GITHUB_OUTPUT"
           fi
 
           cat ./rust-toolchain.toml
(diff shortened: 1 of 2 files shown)


.github/workflows/update-rust.yml
```diff
@@ -1,6 +1,6 @@
 # A GitHub Actions workflow that regularly checks for new Rust toolchain release
 # and creates a PR on new versions.
-name: Rust Update
+name: Update rust
```
bitdivine, line 3: Note: This matches all the other `Update X` actions, so makes it easier to find.

.github/workflows/update-rust.yml
```diff
@@ -36,6 +36,7 @@ jobs:
           fi
 
           echo "latest rust version '$latest_rust_version'"
+          echo "version=$latest_rust_version" >> "$GITHUB_OUTPUT"
```
bitdivine, line 39: Bug: This variable was exposed only when there was _no_ update, so wasn't available for the PR text.

.github/workflows/update-rust.yml
```diff
@@ -69,9 +69,19 @@ jobs:
...
+
+            # Automated changes
+            * Update Rust version.
+
+            # Manual changes
+            - [ ] Fix clippy lints (if necessary)
+            - [ ] Write a changelog entry.
       - name: Notify Slack on failure
+        # Since the this is a scheduled job, a failure won't be shown on any
+        # PR status. To notify the team, we send a message to our Slack channel on failure.
```
bitdivine, line 84: Comment moved down as it relates to this step, not the one above.

.github/workflows/update-rust.yml
```diff
@@ -69,9 +69,19 @@ jobs:
...
+          body: |
+            # Motivation
+            A new verion of Rust is available.
+
+            # Automated changes
+            * Update Rust version.
+
+            # Manual changes
+            - [ ] Fix clippy lints (if necessary)
+            - [ ] Write a changelog entry.
```
bitdivine, line 81: It would be nice if all the bots created their own changelog entries.  But that is outside the scope of this PR.

lmuntaner (approved):
Thanks!

PR #3799, opened 2023-11-20
Create UiTransactions in IcrcTransactionsList

dskloetd:
# Motivation

To simplify transactions list rendering and give the higher level components more control over it, I'm moving the conversion to `UiTransaction` up the component chain.
The ultimate goal is to create the list of `UiTransaction`s directly from the list of `TransactionWithId`s without intermediate `IcrcTransactionData` or `Transaction`. Then it will also be easier to render multiple transactions as 1 (to hide the approval step) or 1 as multiple (such as with a to-self transaction).

In this PR we move creation of `UiTransactions` from the IcrcTransactionCard` to the `IcrcTransactionsList`.

# Changes

1. Add a field `domKey` to `UiTransaction` to keep consistent list rendering.
2. Pass transaction ID to `toUiTransaction` to use for the `domKey`.
3. Remove all props from `IcrcTransactionCard` and keep only 1 prop: `transaction: UiTransaction`.
4. In `IcrcTransactionsList` convert the entire list of transactions to a list of `UiTransaction`s before rendering.

# Tests

1. Remove most tests from `IcrcTransactionCard.spec.ts` because the component became a very simple pass-through components.
2. Add a test that was missing that no transactions are rendered when `token` is not available yet.
3. Add `domKey` to unit tests.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
@@ -2,12 +2,18 @@
   import type { Account } from "$lib/types/account";
   import type { Principal } from "@dfinity/principal";
   import { InfiniteScroll, Spinner } from "@dfinity/gix-components";
+  import type { IcrcTransactionWithId } from "@dfinity/ledger-icrc";
+  import { isNullish } from "@dfinity/utils";
   import { i18n } from "$lib/stores/i18n";
   import IcrcTransactionCard from "./IcrcTransactionCard.svelte";
   import SkeletonCard from "../ui/SkeletonCard.svelte";
-  import type { IcrcTransactionData } from "$lib/types/transaction";
+  import type {
+    IcrcTransactionData,
+    UiTransaction,
+  } from "$lib/types/transaction";
   import type { IcrcTokenMetadata } from "$lib/types/icrc";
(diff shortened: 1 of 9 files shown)


frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
```diff
@@ -18,6 +24,40 @@
   export let descriptions: Record<string, string> | undefined = undefined;
   export let token: IcrcTokenMetadata | undefined;
   export let mapTransaction: mapIcrcTransactionType;
+
+  let uiTransactions: UiTransaction[] = [];
+  $: uiTransactions = transactions.flatMap(
```
lmuntaner, line 29: How about moving this to an `util` so that we can add specific tests for this?

lmuntaner (commented):
One comment, looks good, thanks!

frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
dskloetd: I'll keep moving and changing this up the stack. I'm not sure it's worth adding and removing tests in each step.

lmuntaner (approved):
Thanks for the explanation!

PR #3802, opened 2023-11-20
Chore: Upgrade next dependencies

lmuntaner:
# Motivation

Upgrade NNS Dapp to new ic-js dependencies.

The new version include breaking changes. That's why this PR comes right after the ic-js branch with the breaking changes was merged.

# Changes

## Automatic changes

* package-lock.json

## Manual changes

subaccount and memo can now also be `number[]`

* Add the extra type in some helpers.
* Add the extra type in `SnsSummarySwap`.
* Convert to `Uint8Array` if necessary for some other utils.

# Tests

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/api/sns-governance.api.ts
+++ b/frontend/src/lib/api/sns-governance.api.ts
@@ -342,7 +342,7 @@ export const claimNeuron = async ({
   identity: Identity;
   memo: bigint;
   controller: Principal;
-  subaccount: Uint8Array;
+  subaccount: Uint8Array | number[];
 }): Promise<SnsNeuronId> => {
   logWithTimestamp(`Claiming neuron call...`);
 
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -249,9 +249,9 @@
       }
     },
     "node_modules/@dfinity/ckbtc": {
-      "version": "1.1.0-next-2023-11-15",
-      "resolved": "https://registry.npmjs.org/@dfinity/ckbtc/-/ckbtc-1.1.0-next-2023-11-15.tgz",
-      "integrity": "sha512-nrRjXdNw1cu7fDcfeSbD7ftHJ0AMNXwhzRilc0RV4qo2nCvjd9VAw1wJzhaxzdSguAP5Iftr0t6ZRRRvDfO6Dw==",
+      "version": "1.1.0-next-2023-11-20.1",
+      "resolved": "https://registry.npmjs.org/@dfinity/ckbtc/-/ckbtc-1.1.0-next-2023-11-20.1.tgz",
+      "integrity": "sha512-8Q9VwC8E9en+1lafJlocUiU3A1QibESWLk8Dmn58BQvE8/gUoozSQpMF8mhdHuAcZW/upeWuSC36a67VTbPQUQ==",
       "dependencies": {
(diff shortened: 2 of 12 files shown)


lmuntaner:
* main
  * **PR #3802** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3802?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3802?utm_source=stack-comment).

lmuntaner:
@dskloetd @peterpeterparker pleaes review

frontend/src/lib/api/sns-governance.api.ts
```diff
@@ -342,7 +342,7 @@ export const claimNeuron = async ({
   identity: Identity;
   memo: bigint;
   controller: Principal;
-  subaccount: Uint8Array;
+  subaccount: Uint8Array | number[];
```
dskloetd, line 345: Do we need to accept `number[]` here or can we convert on the calling side?
Or do we actually want to accept more?
I'm on the fence but I lean to being consistent in what we use rather than allowing both.

If we don't make a decision now, I'm afraid the burden of having to support both will keep spreading even if in practice we only use `Uint8Array` everywhere.

Same question applies in other places.

bitdivine:
First PR I have seen using graphite.  I would be interested to hear what your experiences are, after you have tried it for a while.

frontend/src/lib/api/sns-governance.api.ts
lmuntaner: My decision was to match with the entity.

If it's a subaccount, allow both, because that's how the type in ic-js is. Like in this case.

In helpers like `uint8ArrayToHexString`, I converted to `Uint8Array` instead.

I think that it makes sense to keep the type from the source instead of changing it.

I agree it's a burden, but maybe we need to ask the candid team to make up their minds. Or provide a flag to be consistent. Or they could provide a type for the union.

lmuntaner:
> First PR I have seen using graphite. I would be interested to hear what your experiences are, after you have tried it for a while.

So far, I'm happy. Although I still feel I work with my old flow of separating PRs in parallel instead of focusing on one task and splitting it into stacked PRs.

frontend/src/lib/api/sns-governance.api.ts
dskloetd: > I agree it's a burden, but maybe we need to ask the candid team to make up their minds.

The candid team has decided to take on the burden of supporting both. That's nice of them because they have many users and they don't know who prefers what.
That does not mean that we have to do the same in `ic-js` but we can if we want to be nice to other users of `ic-js`.
But if we don't need it in `nns-dapp`, there is absolutely no reason to impose this burden on ourselves.

lmuntaner: We would need to convert the inputs as well as the outputs. The inputs might change as well.

Setting one type here doesn't mean it will always be like that unless we also convert it in to source. Which means we would need to have a duplicated type in NNS Dapp with only one of the two types.

I think that the burden of going out of sync with ic-js is bigger than the burden or supporting a union type.

dskloetd (approved):

PR #3803, opened 2023-11-20
refactor: Mv aggregator updaters

bitdivine:
# Motivation
Previously only the aggregator had Rust code derived automatically from candid.  Now that the proposal payload rendering code does so as well, it is confusing having the aggregator derivation commands at the top level of the scripts directory.

# Changes
- Move the commands for deriving aggregator code into `scripts/sns/aggregator` with the other aggregator-specific commands.

# Tests
See CI

# Todos


lmuntaner (approved):
Thanks!

PR #3808, opened 2023-11-20
Remove unnecessary showTransactionFee

dskloetd:
# Motivation

The logic in `showTransactionFee` is unnecessary.
Burn and Mint transactions don't have a fee so even if we tried to add it, there's nothing to add.
If `toSelfTransaction` is true, then `isReceive` is also true, so there's no need to check `toSelfTransaction` separately.

# Changes

1. Remove `showTransactionFee` and just use `!isReceive` instead.
2. Remove test for `showTransactionFee`.

# Tests

Existing unit tests pass.
Transaction list, including Burn, Mint, Approve and TransferFrom, looks correct.
I also changed `transactionDisplayAmount` to add a fee when it shouldn't or not add a fee when it should and in both cases there was a failing test so that confirms there is test coverage.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/utils/icrc-transactions.utils.ts
+++ b/frontend/src/lib/utils/icrc-transactions.utils.ts
@@ -22,7 +22,6 @@ import {
   nonNullish,
   uint8ArrayToHexString,
 } from "@dfinity/utils";
-import { showTransactionFee } from "./transactions.utils";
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
That's great, thanks!

PR #3809, opened 2023-11-20
Fix: global config fallback

bitdivine:
# Motivation
There is a typo in the global config fallback, causing docker builds to fail on machines that have never set a global config.

Note: Local configs have been deprecated since February 2022, so most active developers will have a global config, but not all.

Note: In CI we call docker directly so that we can cache builds use the buildx action.  Maybe we could have tests that are run only when the `release-candidate` tag is pushed.  That way we could test the docker build script pre-release without slowing down every PR.

# Changes
- Fix the typo - with a var so that the same filename is used on both branches.

# Tests
- Built locally with and without a global config.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp-unreleased.md
+++ b/CHANGELOG-Nns-Dapp-unreleased.md
@@ -28,6 +28,7 @@ proposal is successful, the changes it released will be moved from this file to
 
 #### Fixed
 
+* Fix docker builds when there is no global config.
 * Add missing "Rename" button in the subaccount page.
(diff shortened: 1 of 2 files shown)


lmuntaner (approved):
Oh, that's subtle!

I'm in favor of having some tests run only for specific tags.

bitdivine:
[Ticket made for the release-candidate workflow](https://dfinity.atlassian.net/browse/GIX-2101).  I thought there was a way of setting a priority but I don't see one.

PR #3813, opened 2023-11-21
Map directly to UiTransaction

dskloetd:
# Motivation

In `IcrcTransactionsList` we map `IcrcTransactionData`s to `Transaction`s and then to `UiTransaction`s.
This is not necessary and by doing it in 1 step we can give the custom `mapTransaction` function more control over the rendering of the transactions.
This will make it easier to implement ckBTC UI improvements.

# Changes

1. Change `mapIcrcTransaction` and `mapCkbtcTransaction` to return `UiTransaction` instead of `Transaction`.
2. In `IcrcTransactionsList` map transactions with the new single mapping function. Then filter out `undefined` transactions.
3. 

# Tests

1. Accept optional `timestamp` and `memo` params to `createIcrcTransactionWithId`.
2. Change unit tests for `mapTransaction` to expect the full result instead of just a few properties. Define default values so we still only have to specify the properties relevant to the test.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
@@ -3,7 +3,7 @@
   import type { Principal } from "@dfinity/principal";
   import { InfiniteScroll, Spinner } from "@dfinity/gix-components";
   import type { IcrcTransactionWithId } from "@dfinity/ledger-icrc";
-  import { isNullish } from "@dfinity/utils";
+  import { nonNullish } from "@dfinity/utils";
   import { i18n } from "$lib/stores/i18n";
   import IcrcTransactionCard from "./IcrcTransactionCard.svelte";
   import SkeletonCard from "../ui/SkeletonCard.svelte";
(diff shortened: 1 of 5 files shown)


frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
```diff
@@ -13,7 +13,6 @@
   } from "$lib/types/transaction";
   import type { IcrcTokenMetadata } from "$lib/types/icrc";
   import type { mapIcrcTransactionType } from "$lib/utils/icrc-transactions.utils";
-  import { toUiTransaction } from "$lib/utils//transactions.utils";
```
lmuntaner, line 16: I guess you can't remove this one yet, right?

lmuntaner (approved):
LGTM! Thanks!

frontend/src/lib/components/accounts/IcrcTransactionsList.svelte
dskloetd: It's still used for NNS transactions. Hopefully I get around to doing the same refactorings for NNS transactions but it's less of a priority right now.

PR #3821, opened 2023-11-22
Remove the fallbackDescription slot from TransactionCard

dskloetd:
# Motivation

We use the `fallbackDescription` on `UiTransaction` only to render "From: BTC Network" on ckBTC mint transactions.
We can accomplish the same thing by setting `otherParty` to "BTC Network".

# Changes

1. Remove the `fallbackDescription` field from `UiTransaction`.
2. Remove the description slot from `TransactionCard`.
3. Stop passing `$i18n.ckbtc_transaction_names` from `CkBTCTransactionList` to `IcrcTransactionsList` and to `mapTransaction`.
4. Pass the full `i18n` object to `mapTransaction` instead of just `transaction_names`. This gives it additional access to `ckbtc.btc_network` to use as `otherParty`.
5. In `ColumnRow` use `justify-content: space-between` instead of `@include display.space-between` on larger screens.  [@include display.space-between](https://github.com/dfinity/gix-components/blob/ff7f9beb27b15f91fdee2bbe640d6c7dba28d1e2/src/lib/styles/mixins/_display.scss#L1) comes additionally with `display: inline-flex` and `width: 100%`. We don't need these extra styles and the `display: inline-flex` has the strange effect that the transaction row containing the To/From and date gets taller when the To/From slot isn't there. You can see it in the video in [this comment](https://github.com/dfinity/nns-dapp/pull/3645#discussion_r1378616357). I spent several hours debugging this and I still don't understand why this happens. But removing the include seems cleaner than rendering a zero-width space when there is no To/From. `ColumnRow` is only use in `TransactionCard` and in `CardInfo`. `CardInfo` is used in many places but as far as I can tell, the `ColumnRow` is only relevant to it when it has an `end` slot, which it only has in `NeuronCardContainer` and in `SkeletonCard`. But `NeuronCardContainer` itself isn't used anywhere with an `end` slot. I've looked at `SkeletonCard` and it still looks fine to me.
6. Remove now unused `"ckbtc_transaction_names"` from `i18n`.
7. Set `uiTransaction.otherParty` to "BTC Network" in `mapCkbtcTransaction` for Mint and Burn transactions.
8. Remove `getDescription` from the page object and only check `getIdentifier` in tests.

# Tests

Updated and passing.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/ui/ColumnRow.svelte
+++ b/frontend/src/lib/components/ui/ColumnRow.svelte
@@ -21,6 +21,7 @@
     @include media.min-width(small) {
       @include display.space-between;
       flex-direction: row;
+      align-items: normal;
 
       margin: 0 0 var(--padding);
     }
--- a/frontend/src/lib/components/ui/ColumnRow.svelte
+++ b/frontend/src/lib/components/ui/ColumnRow.svelte
@@ -9,7 +9,6 @@
 
 <style lang="scss">
   @use "@dfinity/gix-components/dist/styles/mixins/media";
-  @use "@dfinity/gix-components/dist/styles/mixins/display";
 
   div {
     display: flex;
@@ -19,9 +18,8 @@
     column-gap: var(--padding-2x);
 
     @include media.min-width(small) {
-      @include display.space-between;
+      justify-content: space-between;
       flex-direction: row;
-      align-items: normal;
 
       margin: 0 0 var(--padding);
     }
--- a/frontend/src/lib/utils/icrc-transactions.utils.ts
+++ b/frontend/src/lib/utils/icrc-transactions.utils.ts
@@ -167,16 +167,14 @@ export const mapIcrcTransaction = ({
   toSelfTransaction,
   governanceCanisterId,
   token,
-  transactionNames,
-  fallbackDescriptions,
+  i18n,
 }: {
   transaction: IcrcTransactionWithId;
   account: Account;
   toSelfTransaction: boolean;
   governanceCanisterId?: Principal;
   token: Token | undefined;
-  transactionNames: I18nTransaction_names;
-  fallbackDescriptions?: Record<string, string>;
+  i18n: I18n;
 }): UiTransaction | undefined => {
   try {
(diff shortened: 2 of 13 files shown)


frontend/src/lib/components/ui/ColumnRow.svelte
```diff
@@ -19,7 +18,7 @@
     column-gap: var(--padding-2x);
 
     @include media.min-width(small) {
-      @include display.space-between;
+      justify-content: space-between;
```
lmuntaner, line 21: This is not strictly related to this PR, right?

It's ok to have it in this PR, it was just for me to understand it. I read the comment, so I'm assuming it's not related.

frontend/src/lib/utils/icrc-transactions.utils.ts
```diff
@@ -244,24 +238,27 @@ export const mapCkbtcTransaction = (params: {
...
   if (isNullish(mappedTransaction)) {
     return mappedTransaction;
   }
   const {
+    i18n,
     transaction: { transaction },
   } = params;
-  if (transaction.burn.length === 1) {
+  if (transaction.mint.length === 1) {
+    mappedTransaction.otherParty = i18n.ckbtc.btc_network;
```
lmuntaner, line 252: Did you add a test for this transaction? I can't see it.

lmuntaner (approved):
Thanks!

Just one comment about a (maybe) missing test.

Already approving though.

frontend/src/lib/utils/icrc-transactions.utils.ts
dskloetd: Good point. There was a failing test without this, so it was covered, but it should have its own dedicated test. Added.

frontend/src/lib/components/ui/ColumnRow.svelte
lmuntaner: Ah, I thought that the problem was already present now, not due to removing the description.

Thanks for the explanation.

PR #3822, opened 2023-11-22
GIX-2102: Add column props to TokensTable

lmuntaner:
# Motivation

# Changes

# Tests

# Todos


lmuntaner:
* main
  * **PR #3822** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3822?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3822?utm_source=stack-comment).

frontend/src/lib/i18n/en.json
```diff
@@ -1040,6 +1040,9 @@
     "status_in_progress_detailed": "Checking for new transactions."
   },
   "tokens": {
-    "title": "Tokens"
+    "title": "Tokens",
+    "projects_header": "Projects",
+    "balance_header": "Balance",
+    "accounts_header": "Accounts"
```
lmuntaner, line 1046: Not used yet, but will be used in following PRs

lmuntaner:
Not in this PR, but this is where the changes apply:

lmuntaner:

frontend/src/lib/components/tokens/TokensTable/TokensTable.svelte
```diff
@@ -1,15 +1,24 @@
 <script lang="ts">
+  import { i18n } from "$lib/stores/i18n";
   import type { UserTokenData } from "$lib/types/tokens-page";
   import Row from "./TokensTableRow.svelte";
 
   export let userTokensData: UserTokenData[];
+  export let columnHeaders: [string, string] = [
```
dskloetd, line 7: Should it really be an array or a prop per specific column?
The data is in a specific format so we can't really choose to render something else in a specific column, right?

And maybe we don't need to be able to control the "Balance" header since it can't render anything else?

frontend/src/lib/components/tokens/TokensTable/TokensTable.svelte
```diff
@@ -1,15 +1,24 @@
...
+    $i18n.tokens.projects_header,
+    $i18n.tokens.balance_header,
+  ];
```
dskloetd, line 10: The plan was to have the table be fully controlled by its props, right?
Doesn't that mean that it shouldn't depend on any stores?
Is i18n an exception?
How does that interact with the idea of the store book?

lmuntaner: The store book? What is that?

I wanted to add some default props, why shouldn't they depend on i18n?

frontend/src/lib/components/tokens/TokensTable/TokensTable.svelte
lmuntaner: I wanted to give more flexibility. But it's true that the second colum will always be a balance.

lmuntaner:

frontend/src/lib/components/tokens/TokensTable/TokensTable.svelte
lmuntaner: Ahh, good point.

I'll remove the default prop in an upcoming PR.

Keeping it like this make it backwards compatible and nicer to review.

dskloetd (approved):

PR #3825, opened 2023-11-22
Refactor: Create SnsSummaryWrapper

lmuntaner:
# Motivation

Stop using the information within `swap` of `SnsSummary` because it comes from the deprecated endpoint `get_state`.

In this PR, I introduce a class SnsSummaryWrapper that implements the same interface as `SnsSummary`.

# Changes

* New directory `wrappers` where I put the first one: `SnsSummaryWrapper`.
* `convertDtoToSnsSummary` returns an instance of `SnsSummaryWrapper`.
* `snsSummariesStore` returns instances of `SnsSummaryWrapper`.

# Tests

* Test that `convertDtoToSnsSummary` creates instances of `SnsSummaryWrapper`.
* Test that `snsSummariesStore` returns instances of `SnsSummaryWrapper`.

# Todos

Not necessary.


lmuntaner:
* main
  * **PR #3825** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3825?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3825?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

The plan is:
* Change all the places where `SnsSummary` type is used for the class type.
* Rename class to `SnsSummary` and remove the type.
* Move the getters to the `SnsSummary`.
* Start using getters for the information within `swap` field.

frontend/src/lib/wrappers/sns-summary.wrappers.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.governanceCanisterId = governanceCanisterId;
+    this.ledgerCanisterId = ledgerCanisterId;
+    this.indexCanisterId = indexCanisterId;
+    this.metadata = metadata;
+    this.token = token;
+    this.swap = swap;
+    this.derived = derived;
+    this.init = init;
+    this.swapParams = swapParams;
+    this.lifecycle = lifecycle;
```
dskloetd, line 68: To me this does not look like a wrapper. This is a copy.
A wrapper would be
```
export class SnsSummaryWrapper implements SnsSummary {
  private readonly summary: SnsSumamry;

  constructor(summary: SnsSummary) {
    this.summary = summary;
  }

  get rootCanisterId() {
    return this.summary.rootCanisterId();
  }

  // etc...
}
```

And that's also what I had in mind.
If a copy is what you want, I'm also not sure what's the point in using a class. Then you can just do
```
  snsSummaryCopy = {
    ...snsSummary,
    // plus whatever changes and overrides you want
  }
```

If I'm not mistaken, the static method below would work just as well with the `SnsSummary` type instead of the Wrapper class, right?

frontend/src/tests/lib/stores/sns.store.spec.ts
```diff
@@ -136,6 +137,12 @@ describe("sns.store", () => {
       expect(get(snsSummariesStore)).toHaveLength(1);
     });
 
+    it("returns instances of SnsSummaryWrapper", () => {
+      snsAggregatorStore.setData([aggregatorSnsMockDto]);
+
+      expect(get(snsSummariesStore)[0]).toBeInstanceOf(SnsSummaryWrapper);
```
dskloetd, line 143: I'm not sure about the value of checking the type. What matter is that it does what it should do.

frontend/src/lib/wrappers/sns-summary.wrappers.ts
lmuntaner: Done!

I had to do a few more changes like this, but it still works.

dskloetd: I see only 1 commit in the PR.

frontend/src/lib/types/sns-summary-wrapper.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  static overrideDerivedState(
+    currentSummary: SnsSummaryWrapper,
+    newDerivedState: SnsSwapDerivedState
+  ): SnsSummaryWrapper {
+    return new SnsSummaryWrapper({
+      ...currentSummary.summary,
+      derived: newDerivedState,
+    });
+  }
```
dskloetd, line 67: What is the reason for having this static method instead of just doing
```
   new SnsSummaryWrapper({
      ...currentSummary.summary,
      derived: newDerivedState,
    });
```
directly?

frontend/src/lib/types/sns-summary-wrapper.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  static overrideLifecycle(
+    currentSummary: SnsSummaryWrapper,
```
dskloetd, line 70: It usually doesn't make sense to have a static method which takes the same class as parameter.
You'd make it a non-static method instead with 1 fewer parameter.

That would work here as well, right?

frontend/src/lib/wrappers/sns-summary.wrappers.ts
lmuntaner: I see 4 now, but maybe something got messed up?

frontend/src/tests/lib/stores/sns.store.spec.ts
lmuntaner: removed, indeed. The type has changed, no need for this.

lmuntaner:

frontend/src/lib/stores/sns.store.ts
```diff
@@ -126,37 +127,23 @@ const overrideDerivedState =
...
   (lifecycleStore: SnsLifecycleData) =>
-  (summary: SnsSummary | undefined): SnsSummary | undefined => {
+  (summary: SnsSummaryWrapper | undefined): SnsSummaryWrapper | undefined => {
     if (isNullish(summary)) {
       return undefined;
     }
     const projectData = lifecycleStore[summary.rootCanisterId.toText()];
-    const lifecycle = fromNullable(projectData?.data.lifecycle ?? []);
-    if (isNullish(lifecycle)) {
+    if (isNullish(projectData?.data)) {
```
dskloetd, line 143: There is a very similar nullish check inside `SnsSummaryWrapper.overrideLifecycle`.
Does it make sense to move this check there as well?

frontend/src/tests/lib/stores/sns.store.spec.ts
dskloetd: Can we just remove this test or is there something that you are worried about that we can maybe test in a better way?

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
```diff
@@ -286,158 +287,160 @@ describe("sns aggregator converters utils", () => {
...
-              min_participant_icp_e8s: [],
-              min_participants: [],
-              neuron_basket_construction_parameters: [],
-              neuron_minimum_stake_e8s: [400000000n],
-              neurons_fund_participants: [],
-              nns_governance_canister_id: "rrkah-fqaaa-aaaaa-aaaaq-cai",
-              nns_proposal_id: [],
-              restricted_countries: [
+      expect(convertDtoToSnsSummary(mockData)).toEqual(
+        new SnsSummaryWrapper({
```
dskloetd, line 291: Is the difference below only in indentation?

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
```diff
@@ -286,158 +287,160 @@ describe("sns aggregator converters utils", () => {
...
-              min_icp_e8s: [],
-              min_participant_icp_e8s: [],
-              min_participants: [],
-              neuron_basket_construction_parameters: [],
-              neuron_minimum_stake_e8s: [400000000n],
-              neurons_fund_participants: [],
-              nns_governance_canister_id: "rrkah-fqaaa-aaaaa-aaaaq-cai",
-              nns_proposal_id: [],
-              restricted_countries: [
+      expect(convertDtoToSnsSummary(mockData)).toEqual(
```
dskloetd, line 290: Does this check also pass if we don't change it?
Do we need to care that it's an instance of SnsSummaryWrapper or is it enough that the getters return the same values?

frontend/src/lib/types/sns-summary-wrapper.ts
lmuntaner: Yes, maybe. It would be like a `copy()` method actually.

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
lmuntaner: Yes, correct.

frontend/src/tests/lib/utils/sns-aggregator-converters.utils.spec.ts
lmuntaner: Yes, it matters. It doesn't pass otherwise.

frontend/src/tests/lib/stores/sns.store.spec.ts
lmuntaner: removed!

lmuntaner:

frontend/src/lib/stores/sns.store.ts
lmuntaner: That's not the same.

`isNullish(projectData?.data)` is to check that there is a response from the endpoint.

The one below is checking a field within the response called `lifecycle`.

```
if (isNullish(lifecycle)) {
    return this;
}
```

When I said: "It feels weird to call overrideLifecycle if I don't have a lifecycle." I meant without the LifecycleResponse.

Maybe I can rename the method to make it clearer.

dskloetd: I understand it's not exactly the same. But in both cases there is no life cycle data to override with and you end up returning the original data instead of overriding something. So I still don't understand why it's different in a way that matters.

I'm guessing I'm missing something. Would it be more complicated than moving these 3 lines? Or would it be inconvenient or confusing or something?

It's not super important but if I'm missing something, I'd like to understand.

lmuntaner: It's not about the technical changes, but conceptual.

In one case I'm missing the whole object, therefore, why call the method.

The second case is, I have the whole object, but because one part is optional there, but I don't want it optional in another place, I can't override if it's empty.

PR #3830, opened 2023-11-23
GIX-2062: Add loading balance

lmuntaner:
# Motivation

Give a better user experience in the tokens table pages.

In this PR, I introduce two new types `UserTokenBase` and `UserTokenLoading`. Both are accepted in the props of TokenTable. But only the initial one `UserTokenData` can be used to dispatch actions.

I guess the loading could also dispatch events, but we should use another event name, not `nnsAction`. This way, we know for certain that when `nnsAction` is triggered, it has the token and fee available. This is important to open modals at the moment, but could be important for other actions in the future.

# Changes

* Add two new types `UserTokenBase` and `UserTokenLoading`.
* Render a Spinner if UserTokenLoading in TokensTableRow.
* Support a union of `UserTokenLoading` and `UserTokenData` in TokensTable and TokensTableRow

# Tests

* Add a test in TokensTable.spec for the new loading value

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
+++ b/frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
@@ -1,5 +1,9 @@
 <script lang="ts">
-  import { UserTokenAction, type UserTokenData } from "$lib/types/tokens-page";
+  import {
+    UserTokenAction,
+    type UserTokenData,
+    type UserTokenLoading,
+  } from "$lib/types/tokens-page";
   import {
     SvelteComponent,
     createEventDispatcher,
@@ -15,8 +19,9 @@
   import { nonNullish } from "@dfinity/utils";
   import { i18n } from "$lib/stores/i18n";
   import { Spinner } from "@dfinity/gix-components";
+  import { isUserTokenData } from "$lib/utils/user-token.utils";
 
-  export let userTokenData: UserTokenData;
+  export let userTokenData: UserTokenData | UserTokenLoading;
   export let index: number;
 
   const dispatcher = createEventDispatcher();
@@ -30,11 +35,18 @@
     [UserTokenAction.Send]: SendButton,
   };
 
-  const handleClick = () =>
-    dispatcher("nnsAction", {
-      type: ActionType.GoToTokenDetail,
-      data: userTokenData,
-    });
+  let userToken: UserTokenData | undefined;
+  $: userToken = isUserTokenData(userTokenData) ? userTokenData : undefined;
+
+  const handleClick = () => {
+    if (nonNullish(userToken)) {
+      // Actions can only be dispatched with `UserTokenData`
+      dispatcher("nnsAction", {
+        type: ActionType.GoToTokenDetail,
+        data: userToken,
+      });
+    }
+  };
 </script>
 
 <div
@@ -62,13 +74,15 @@
       </div>
     </div>
     <div class="title-actions actions mobile-only">
-      {#each userTokenData.actions as action}
-        <svelte:component
-          this={actionMapper[action]}
-          userToken={userTokenData}
-          on:nnsAction
-        />
-      {/each}
+      {#if nonNullish(userToken)}
+        {#each userToken.actions as action}
+          <svelte:component
+            this={actionMapper[action]}
+            {userToken}
+            on:nnsAction
+          />
+        {/each}
+      {/if}
     </div>
   </div>
   <div role="cell" class="mobile-row-cell left-cell">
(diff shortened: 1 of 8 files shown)


lmuntaner:
* main
  * **PR #3830** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3830?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3830?utm_source=stack-comment).

lmuntaner:

lmuntaner:
New UI: (not used in this PR though)

There is misalignment of both icons. This is due to the button with the icon not centering vertically the icon. I will fix that in a PR in gix-components.

lmuntaner:
@dskloetd now it's ready for review

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -75,18 +90,24 @@
       <span data-tid="token-value-label"
         >{`-/- ${userTokenData.balance.token.symbol}`}</span
       >
+    {:else if userTokenData.balance === "loading"}
```
dskloetd, line 93: How about defining
```
$: isLoading = isNullish(userToken)
```
above?

frontend/src/lib/types/tokens-page.ts
```diff
@@ -8,14 +8,29 @@ export enum UserTokenAction {
...
+ * but they are not present while loading because the data is not yet there.
+ *
+ */
+export type UserTokenLoading = UserTokenBase & {
+  balance: "loading";
+  actions: [];
+};
+
+export type UserTokenData = UserTokenBase & {
+  balance: TokenAmount | UnavailableTokenAmount | "loading";
```
dskloetd, line 32: Why can `balance` be `"loading"` here? Wouldn't we use the `UserTokenLoading` type for that case?

frontend/src/tests/page-objects/TokensTableRow.page-object.ts
```diff
@@ -25,6 +25,11 @@ export class TokensTableRowPo extends BasePageObject {
     return this.getText("token-value-label");
   }
 
+  hasBalanceSpinner(): Promise<boolean> {
+    const balanceElement = this.root.byTestId("token-value-label");
+    return balanceElement.byTestId("spinner").isPresent();
```
dskloetd, line 30: This can be
```
return balanceElement.isPresent("spinner");
```

lmuntaner: Nops...

```
export interface PageObjectElement {
  //...
  isPresent(): Promise<boolean>;
  //...
}
```

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: But then I need an `else if userTokenData.balance instanceof TokenAmount` below because TS doesn't know that's the only missing type.

I'd rather keep it like this, because if we add a new type to the union in the future, I'd like TS to fail.

lmuntaner:
New UI:

lmuntaner:
New UI: (no action icon now) no alignment to take care of.

io

frontend/src/lib/components/tokens/TokensTable/actions/GoToDetailButton.svelte
```diff
@@ -10,7 +10,7 @@
 </script>
 
 <button
-  class="icon"
+  class="icon-only"
```
lmuntaner, line 13: This was a mistake. The class icon does not exist on button, it's `icon-only`. That's why I had the misalignment in the first place.

frontend/src/lib/types/tokens-page.ts
lmuntaner: done!

lmuntaner:

dskloetd (approved):

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -75,18 +90,24 @@
...
+      <span data-tid="token-value-label" class="balance-spinner"
+        ><Spinner inline size="tiny" /></span
+      >
     {:else}
       <AmountDisplay singleLine amount={userTokenData.balance} />
     {/if}
```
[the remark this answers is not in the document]
dskloetd: Doesn't have to be this PR.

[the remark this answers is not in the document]
lmuntaner: Yes, maybe. This component is already getting hard to read.

PR #3831, opened 2023-11-23
Draft Migration test

bitdivine:
# Motivation
Testing migration is very similar to the downgrade-upgrade test.  The downgrade-upgrade test can almost certainly be expanded to make it appropriate for this use case.  However the schema change tests are likely to need a lot of modification over the next few weeks, so I would prefer to have a separate test for schema changes that can "move fast and experiment" and then migrate useful code back into the original, eventually deleting the experimental code.

Why not keep the test in a branch?  Most development is on Rust code.  Importing tests into every rust development branch has been a real pain.

# Changes
* Make a copy of the downgrade upgrade test dedicated to schema changes:
  * Copy the test script
  * Mark the new test as unstable.
  * Split the test into small chunks, as the file is huge!
  * Add argument linters & backtraces to help with debugging.
* Adapt the command line arguments to make them suitable for schema migration tests:
  * Previously the downgrade-upgrade test was always between the default current build and prod.  Added flags to allow the downgrade-upgrade test to work against any pair of wasms and argfiles.  This way, we can e.g. upgrade to self, just changing the schema in the schema.
  * Add flags to make it easy to change the schema without writing a new argfile.  We will need this a lot! :-)

# Tests
* Run locally.  Marked the most urgent missing unit tests with TODOs in the code.

# Todos

### DIFF
--- a/scripts/nns-dapp/migration-test
+++ b/scripts/nns-dapp/migration-test
@@ -19,8 +19,8 @@ help_text() {
 # Source the clap.bash file ---------------------------------------------------
 source "$SOURCE_DIR/../clap.bash"
 # Define options
-clap.define short=w long=wasm1 desc="A test build of the first wasm.  May be a .wasm.gz file or a named release." variable=FIRST_WASM default="out/nns-dapp_test.wasm.gz"
-clap.define short=W long=wasm2 desc="A test build of the second wasm.  May be a .wasm.gz file or a named release." variable=SECOND_WASM default="prod"
+clap.define short=w long=wasm1 desc="A test build of the first wasm.  May be a _test.wasm.gz file or a named release." variable=FIRST_WASM default="out/nns-dapp_test.wasm.gz"
+clap.define short=W long=wasm2 desc="A test build of the second wasm.  May be a _test.wasm.gz file or a named release." variable=SECOND_WASM default="prod"
 clap.define short=a long=args1 desc="A file containing the first init arguments." variable=FIRST_ARGFILE default="out/nns-dapp-arg-local.did"
 clap.define short=A long=args2 desc="A file containing the second init arguments." variable=SECOND_ARGFILE default="out/nns-dapp-arg-local.did"
 clap.define short=s long=schema1 desc="The schema for the first deployment.  Overrides the value in the argument file." variable=FIRST_SCHEMA default=""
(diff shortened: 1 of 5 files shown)


scripts/nns-dapp/migration-test.canister
```diff
@@ -0,0 +1,161 @@
...
+      printf "  Expected at least: %10d\n" "$expected_min_transactions"
+      printf "  Actual:            %10d\n" "$transactions_count"
+      printf "  Note: Transactions MAY be pruned but if a lot are being pruned, there is insufficient memory."
+      exit 1
+    } >&2
+  fi
+}
+
+# Wait for the migration countdown to complete.
+wait_for_migration() {
```
bitdivine, line 76: TODO: This waits for the migration "progress meter" but wqe also need to check that when the progress meter has finished, the schema is as expected.

scripts/nns-dapp/migration-test.canister
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: Note:  All the code in this file is directly as in the downgrade-upgrade test, but with relevant functionality grouped and moved into a small file.

scripts/nns-dapp/migration-test
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: This file is the downgrade-upgrade test with:
- Code moved out into libraries
- Flags added to support arbitrary pairs of wasm, arguments and schema.
- Debug stack traces.

scripts/nns-dapp/migration-test.linters
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bash
```
bitdivine, line 1: All the code in this file is new.  It has argument linters so that errors can be caught early, rather than waiting for something to break downstream and then having to figure out what caused the breakage.

lmuntaner (approved):
This is not yet used in CI, is that right?

Thanks!

bitdivine:
@lmuntaner  Correct, this is exercised only locally.

I have to update the memory manager (move it to offset 0) and a couple of other changes like that before I can make  meaningful CI test.

Thank you for the review.

PR #3838, opened 2023-11-24
Check schemas in the migration test

bitdivine:
# Motivation
We wish to be sure that when a migration has taken place, the data is stored according to the correct schema.

Note: The `get_stats` API provides the current schema.

# Changes
- Define bash functions to convert the numeric schema index provided in `get_stats()` into a name and back.
- If `--schema` has been declared explicitly, check that the `get_stats()` reports the correct schema after migration.
- Update docs.

# Tests
- Unit tests are included.
- System tests probably don't make sense yet in CI but I have run the overall test locally.

# Todos

  - Covered by an existing changelog entry


scripts/nns-dapp/migration-test
```diff
@@ -24,25 +24,25 @@ help_text() {
 	EOF
 }
```
bitdivine, line 26: Moved up so that the list of schemas is available to the help text.

.github/workflows/checks.yml
```diff
@@ -364,7 +373,27 @@ jobs:
       - name: Getting network config works
         run: scripts/network-config.test
   checks-pass:
```
bitdivine, line 375: The list got too long to read in this format.

lmuntaner (approved):
LGTM, rubber stamp

PR #3839, opened 2023-11-24
Support rendering pending transactions

dskloetd:
# Motivation

When converting BTC to and from ckBTC, it takes a while before the transaction on one network results in a transaction on the other network.
We want to render this as a "pending" transaction, which should look like this:

image

# Changes

1. Add a field `isPending` to the type `UiTransaction`.
2. Make the field `timestamp` optional in `UiTransaction`. Since pending transaction haven't happened yet, they don't have a timestamp.
3. In `TransactionCard` make the icon orange, by adding class `pending`, when the transaction is pending.
4. In `TransactionCard` don't render a timestamp if it's absent, and render "Pending..." if it's pending.
5. Add `isPending: false` to existing instances.
 
# Tests

1. Add a unit test for rendering a pending transaction.
2. Test that a non-pending transaction does not have a pending icon.
3. Add `hasPendingIcon` to page object.

# Todos

will add when it's used

### DIFF
--- a/frontend/src/lib/components/accounts/TransactionCard.svelte
+++ b/frontend/src/lib/components/accounts/TransactionCard.svelte
@@ -14,9 +14,10 @@
   let headline: string;
   let tokenAmount: TokenAmount;
   let isIncoming: boolean;
+  let isPending: boolean;
   let otherParty: string | undefined;
-  let timestamp: Date;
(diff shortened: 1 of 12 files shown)


mstrasinskis (approved):
LGTM, thanks!

PR #3853, opened 2023-11-27
Map PendingUtxo to UiTransaction

dskloetd:
# Motivation

When converting BTC to ckBTC, the minter only credits the ckBTC account when the incoming BTC transaction has 12 confirmations. When a UTXO has fewer than 12 confirmations, we consider it "pending" and we want to render it as a pending transaction.

In this PR I add a function to convert `PendingUtxo` to `UiTransaction` so that we can use `TransactionCard` to render it (already implemented in https://github.com/dfinity/nns-dapp/pull/3839).
The mapping function will be used in a future PR.

# Changes

1. Add `mapCkbtcPendingUtxo` which converts a `PendingUtxo` to a `UiTransaction`.
2. Add a unit test. The conversion is so straightforward that a single test seemed sufficient, but let me know if you agree.
3. Drive-by: Use a ckBTC token instead of `ICPToken` in another ckBTC related unit test. This doesn't matter to the test but it might be confusing to use ICP instead of ckBTC.

# Tests

Unit tests pass.
Tested manually in another branch which has the working code to rendering pending UTXOs.

# Todos

will add when it's used

### DIFF
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -999,7 +999,8 @@
     "checking_incomplete_btc_transfers": "Checking for incomplete BTC transfers",
     "all_btc_transfers_complete": "All transfers are complete",
     "click_to_complete_btc_transfers": "Click here to complete BTC transfer for $amount ckBTC",
-    "btc_network": "BTC Network"
+    "btc_network": "BTC Network",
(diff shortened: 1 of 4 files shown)


lmuntaner (approved):
LGTM, test seems sufficient. Thanks!

PR #3841, opened 2023-11-24
Json preview tests

mstrasinskis:
# Motivation

Update new proposal UI related tests before removing the `ENABLE_FULL_WIDTH_PROPOSAL` flag support.

# Changes

- remove unused test-ids from `TreeJson.svelte`
- add missing testIds
- `JsonPreview.spec`
- `JsonRepresentationModeToggle.spec`
- create related POs
- remove `ENABLE_FULL_WIDTH_PROPOSAL` from tests.
- fix usage of not expanded data in raw view

# Tests

Yes

# Todos

- [ ] Add entry to changelog (if necessary).
Will be updated when the flag is removed.

PR #3855, opened 2023-11-27
test: JsonPreview + po

mstrasinskis:
# Motivation

Unit tests for `JsonPreview` component

# Changes

- test ids
- POs
- spec file
- ~~fix usage of not expanded data in [json raw view](https://github.com/dfinity/nns-dapp/pull/3855/files#diff-d3221440ffd52395ad348f65c868bc110230266fb5e4db91a12eab755fe5d5b1R66)
  Some fields can be delivered in json-stringified format. There is a util function `expandObject` to parse all object fields if needed. For the raw-view there was rendered an object w/o applied `expandObject`, which could leave to not perfect data formatting.~~

# Tests

Yes

# Todos

Not necessary.


frontend/src/lib/components/common/JsonPreview.svelte
```diff
@@ -49,13 +56,14 @@
...
           <TreeJson
+            testId="tree-json"
             json={expandedData}
             defaultExpandedLevel={isAllExpanded ? Number.MAX_SAFE_INTEGER : 1}
           />
         </div>
       {:else}
         <div in:fade>
-          <RawJson {json} />
+          <RawJson testId="raw-json" json={expandedData} />
```
dskloetd, line 66: Did you mean to change the prop from `json` to `expandedData`?
This is a non-test change, right?

frontend/src/lib/components/common/TreeJson.svelte
```diff
@@ -37,7 +38,7 @@
     root = _level === 0;
     // ignore 0 level wrapper
     keyRoot = _level <= 1;
-    testId = root ? "json" : undefined;
+    testIdValue = root ? testId : undefined;
```
dskloetd, line 41: Can this be removed? Now that we pass the `testId` explicitly, we just don't pass one when it's not the root, right?

frontend/src/lib/components/common/TreeJson.svelte
```diff
@@ -73,7 +74,7 @@
   {/if}
   {#if !collapsed}
     <!-- children of expandable-key -->
-    <ul class:root class:is-array={isArray} in:fade>
+    <ul class:root class:is-array={isArray} in:fade data-tid={testIdValue}>
```
dskloetd, line 77: Is it not possible for the `data-tid` on line 61 and the one here to both be in the DOM?
How about collapsing lines 74-75 to
```
{:else if !collapsed}
```
to make it obvious?

frontend/src/tests/lib/components/common/JsonPreview.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    const po = renderComponent({ test: "hello world" });
+    expect(await po.getTreeText()).toBe('test "hello world"');
+  });
+
+  it("should render data in raw view", async () => {
+    jsonRepresentationStore.setMode("raw");
+    const po = renderComponent({ test: "hello world" });
+    expect(await po.getRawObject()).toEqual({
+      test: "hello world",
+    });
```
dskloetd, line 40: Can we just compare the text instead of parsing the JSON?

frontend/src/lib/components/common/JsonPreview.svelte
mstrasinskis: Correct, this is a minor bug fix that was catch by a new test from the old [pr](https://github.com/dfinity/nns-dapp/pull/3841). Just to keep less PRs the fix is there. I'll update the description.

frontend/src/lib/components/common/TreeJson.svelte
mstrasinskis: This is the [recursive component](https://github.com/dfinity/nns-dapp/pull/3855/files#diff-add258a073f68bf0fcb3160532a621f11a2fd2b8708481341873c68ce13d646cL79), so this is done only to not have the same code snippet in every place the `testId` is used in the template.

frontend/src/lib/components/common/TreeJson.svelte
mstrasinskis: Not really, both blocks can be rendered for the level 1+

frontend/src/lib/components/common/JsonPreview.svelte
dskloetd: Can you explain a bit more (ideally in the PR description) why we do this kind of expansion? What was the behavior before and why was this a problem? And what is the new behavior?

frontend/src/lib/components/common/TreeJson.svelte
dskloetd: I understand that's the reason why this was here before.
But now `testId` is a prop, and we don't pass `testId` on line 80.
So I still think this is not needed anymore.

mstrasinskis: Then this part `root ? testId : undefined` should be inside the html template as `data-tid={root ? testId : undefined}` instead of `data-tid={testIdValue}`. Or I got it wrong?

mstrasinskis: Yes, you're right! I was focused on the wrong part. Thanks, will be updated.

frontend/src/lib/components/common/JsonPreview.svelte
mstrasinskis: Yes.

frontend/src/lib/components/common/TreeJson.svelte
```diff
@@ -38,7 +37,6 @@
     root = _level === 0;
```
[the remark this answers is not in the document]
mstrasinskis: It is used also here `<li class:root class:key-is-index={keyIsIndex}>`.

PR #3856, opened 2023-11-27
test: for JsonRepresentationModeToggle

mstrasinskis:
# Motivation

Test for `JsonRepresentationModeToggle` component.

# Changes

- po
- test

# Tests

Yes

# Todos

Not necessary

### DIFF
--- a/frontend/src/lib/components/proposal-detail/JsonRepresentationModeToggle.svelte
+++ b/frontend/src/lib/components/proposal-detail/JsonRepresentationModeToggle.svelte
@@ -15,7 +15,7 @@
 
 <div
   class="data-representation-toggle"
-  data-tid="data-representation-mode-toggle-component"
+  data-tid="json-representation-mode-toggle-component"
 >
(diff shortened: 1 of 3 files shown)


frontend/src/lib/components/proposal-detail/JsonRepresentationModeToggle.svelte
```diff
@@ -13,7 +13,10 @@
     $jsonRepresentationModeStore === "tree" ? setRaw() : setTree();
 </script>
 
-<div class="data-representation-toggle" data-tid="data-representation-toggle">
+<div
+  class="data-representation-toggle"
+  data-tid="data-representation-mode-toggle-component"
```
dskloetd, line 18: If there is an existing `data-tid`, I usually just use that without changing it.
But if you want to change it to the pattern that I use when there is not yet a `data-tid`, the name should match the component. Specifically, it should start with `json` rather than `data`.

mstrasinskis: Good point, thanks!

PR #3862, opened 2023-11-27
fix: preserve undefined in stringifyJson

mstrasinskis:
# Motivation

Preserve undefined in `stringifyJson` util. Since the function is used in `RawJson` component it's important to preserve all value, even empty.

# Changes

- preserve undefined

# Tests

Extended

# Todos


lmuntaner (approved):
Thanks!

I thought this was already the case...

mstrasinskis:
> I thought this was already the case...

@lmuntaner, before there was only one mode, now it works as expected in `tree view` but was wrong in `raw view`, since `JSON.stringify` removes all `undefine` fields.

PR #3868, opened 2023-11-27
GIX-2117: Load ckETH on Accounts page

lmuntaner:
# Motivation

Load ckETH and ckETHSepolia (if needed) in Accounts page.

# Changes

* Use services `loadCkETHCanisters` and `loadIcrcAccounts` in Accounts page.

# Tests

* Test that the store is filled with proper data in Accounts route.

I added this test because the Accounts page doesn't render the token selector sidebar. Therefore, I can't test that the token is present in the UI from that component.

This will change with the new Tokens page. Therefore, there is no need to improve it for now.

# Todos

Not necessary

### DIFF
--- a/frontend/src/lib/routes/Accounts.svelte
+++ b/frontend/src/lib/routes/Accounts.svelte
@@ -26,6 +26,13 @@
   import AccountsModals from "$lib/modals/accounts/AccountsModals.svelte";
   import CkBTCAccountsModals from "$lib/modals/accounts/CkBTCAccountsModals.svelte";
   import { icpTokensListUser } from "$lib/derived/icp-tokens-list-user.derived";
+  import {
+    icrcCanistersStore,
+    type IcrcCanistersStoreData,
+  } from "$lib/stores/icrc-canisters.store";
+  import { loadIcrcAccounts } from "$lib/services/icrc-accounts.services";
+  import { onMount } from "svelte";
+  import { loadCkETHCanisters } from "$lib/services/cketh-canisters.services";
 
   // TODO: This component is mounted twice. Understand why and fix it.
 
@@ -34,6 +41,10 @@
 
   let loadSnsAccountsBalancesRequested = false;
   let loadCkBTCAccountsBalancesRequested = false;
+
+  onMount(() => {
+    loadCkETHCanisters();
+  });
 
   const loadSnsAccountsBalances = async (projects: SnsFullProject[]) => {
     // We start when the projects are fetched
@@ -73,10 +84,19 @@
     });
   };
 
+  const loadIcrcs = (icrcCanisters: IcrcCanistersStoreData) => {
+    const ledgerCanisterIds = Object.values(icrcCanisters).map(
+      ({ ledgerCanisterId }) => ledgerCanisterId
+    );
+
+    loadIcrcAccounts({ ledgerCanisterIds, certified: false });
+  };
+
   $: (async () =>
     await Promise.allSettled([
       loadSnsAccountsBalances($snsProjectsCommittedStore),
       loadCkBTCAccountsBalances($ckBTCUniversesStore),
+      loadIcrcs($icrcCanistersStore),
     ]))();
 </script>
 
--- a/frontend/src/lib/routes/Accounts.svelte
+++ b/frontend/src/lib/routes/Accounts.svelte
@@ -84,7 +84,7 @@
     });
   };
 
-  const loadIcrcs = (icrcCanisters: IcrcCanistersStoreData) => {
+  const loadIcrcTokenAccounts = (icrcCanisters: IcrcCanistersStoreData) => {
     const ledgerCanisterIds = Object.values(icrcCanisters).map(
       ({ ledgerCanisterId }) => ledgerCanisterId
     );
@@ -96,7 +96,7 @@
     await Promise.allSettled([
       loadSnsAccountsBalances($snsProjectsCommittedStore),
       loadCkBTCAccountsBalances($ckBTCUniversesStore),
-      loadIcrcs($icrcCanistersStore),
+      loadIcrcTokenAccounts($icrcCanistersStore),
     ]))();
 </script>
 
--- a/frontend/src/lib/services/cketh-canisters.services.ts
+++ b/frontend/src/lib/services/cketh-canisters.services.ts
@@ -9,16 +9,26 @@ import {
   ENABLE_CKTESTBTC,
 } from "$lib/stores/feature-flags.store";
 import { icrcCanistersStore } from "$lib/stores/icrc-canisters.store";
+import { isNullish } from "@dfinity/utils";
 import { get } from "svelte/store";
 
 export const loadCkETHCanisters = async () => {
-  if (get(ENABLE_CKETH)) {
+  const storeData = get(icrcCanistersStore);
+  // To avoid rerendering the UI and possibly triggering new requests
+  // We don't change the store if it's already filled.
+  if (
+    get(ENABLE_CKETH) &&
+    isNullish(storeData[CKETH_LEDGER_CANISTER_ID.toText()])
+  ) {
     icrcCanistersStore.setCanisters({
       ledgerCanisterId: CKETH_LEDGER_CANISTER_ID,
       indexCanisterId: CKETH_INDEX_CANISTER_ID,
     });
   }
-  if (get(ENABLE_CKTESTBTC)) {
+  if (
+    get(ENABLE_CKTESTBTC) &&
+    isNullish(storeData[CKETHSEPOLIA_LEDGER_CANISTER_ID.toText()])
+  ) {
     icrcCanistersStore.setCanisters({
       ledgerCanisterId: CKETHSEPOLIA_LEDGER_CANISTER_ID,
       indexCanisterId: CKETHSEPOLIA_INDEX_CANISTER_ID,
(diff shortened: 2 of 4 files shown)


lmuntaner:
* main
  * **PR #3868** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3868?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3868?utm_source=stack-comment).

lmuntaner:

frontend/src/lib/routes/Accounts.svelte
```diff
@@ -35,6 +42,10 @@
   let loadSnsAccountsBalancesRequested = false;
   let loadCkBTCAccountsBalancesRequested = false;
 
+  onMount(() => {
+    loadCkETHCanisters();
```
dskloetd, line 46: Don't you want to guard this with the feature flag?

frontend/src/lib/routes/Accounts.svelte
```diff
@@ -73,10 +84,19 @@
     });
   };
 
+  const loadIcrcs = (icrcCanisters: IcrcCanistersStoreData) => {
```
dskloetd, line 87: > `loadIcrcs`

ICRC-21 and ICRC-25 are also ICRCs but have nothing to do with tokens.
I find it confusing enough that we don't use the number for ICRC tokens, but let's at least use some noun to indicate what this is about.

dskloetd (approved):

frontend/src/lib/routes/Accounts.svelte
lmuntaner: No, feature flags are inside `loadCkETHCanisters`.

frontend/src/lib/routes/Accounts.svelte
lmuntaner: Good point

frontend/src/lib/routes/Accounts.svelte
peterpeterparker, line 46: don't think it's correct to locate thise here, because we also gonna need those on the wallet page for example.

frontend/src/lib/routes/Accounts.svelte
```diff
@@ -73,10 +84,19 @@
     });
   };
 
+  const loadIcrcTokenAccounts = (icrcCanisters: IcrcCanistersStoreData) => {
+    const ledgerCanisterIds = Object.values(icrcCanisters).map(
```
peterpeterparker, line 88: there isn't a race condition here? like the canister store not being ready while this is executed?

lmuntaner: It will rerun when the canisters store is filled. But you raised a good point that we don't want to fill the canisters store multiple times and trigger multiple times this if not necessary.

frontend/src/lib/routes/Accounts.svelte
lmuntaner: Do you mean checking the feature flag, or the call to `loadCkETHCanisters`?

lmuntaner:
@dskloetd I added some code after your review.

The one in frontend/src/lib/services/cketh-canisters.services.ts and frontend/src/tests/lib/services/cketh-canisters.services.spec.ts

frontend/src/lib/routes/Accounts.svelte
peterpeterparker: the call to `loadCkETHCanisters`

I a user open `https://nns.ic0..../wallet/?id=....` directly for example

lmuntaner: I expect this call to be also in the Wallet page.

frontend/src/lib/routes/Accounts.svelte
peterpeterparker: You gonna fix it in this PR or another ?

frontend/src/lib/routes/Accounts.svelte
lmuntaner: We do the same for the SNS accounts. We load them in the specific wallet or accounts. Basically on demand. Why do you see this different?

frontend/src/lib/routes/Accounts.svelte
lmuntaner: I added some changes in the last comment.

The ones in frontend/src/lib/services/cketh-canisters.services.ts and frontend/src/tests/lib/services/cketh-canisters.services.spec.ts

frontend/src/lib/routes/Accounts.svelte
peterpeterparker: To me loading those canister ids is closer to loading the sns projects than loading a particular feature such as the accounts.

lmuntaner: I had it in the init of the app and you said that you didn't like it there. Did you change your mind then?

I can probably change it in another PR. It's no biggie.

peterpeterparker: > I had it in the init of the app and you said that you didn't like it there. 

Don't remember having seen that PR today for the canister ids store.

Anyway like I said, go on, won't block the PR.

frontend/src/lib/services/cketh-canisters.services.ts
```diff
@@ -9,19 +9,25 @@ import {
...
 import { get } from "svelte/store";
 
 export const loadCkETHCanisters = async () => {
+  const storeData = get(icrcCanistersStore);
   if (get(ENABLE_CKETH)) {
-    icrcCanistersStore.setCanisters({
-      ledgerCanisterId: CKETH_LEDGER_CANISTER_ID,
-      indexCanisterId: CKETH_INDEX_CANISTER_ID,
-    });
+    if (isNullish(storeData[CKETH_LEDGER_CANISTER_ID.toText()])) {
```
peterpeterparker, line 18: if {
   if {
   
you can inline the if

maybe we can also have a derived store

peterpeterparker (approved):
LGTM, I guess we can iterate from here anyway if needed. Thanks.

frontend/src/lib/routes/Accounts.svelte
peterpeterparker: Ah I remember, it's the fact that it was loaded publicaly that I disliked because I assumed it is not something we need yet publicaly.

lmuntaner: I think we need it, ckBTC appears in the token selector when the user is not logged in for example.

frontend/src/lib/services/cketh-canisters.services.ts
lmuntaner: I know... but it's temporary because it relates to feature flags. In the end, there will be only one `if` to check whether the store is loaded.

But maybe a derived store makes sense. It would make it more readable.

dskloetd, line 18: Why does it matter if we set the store again?

lmuntaner: It will trigger a rerender of the components and it will trigger a fetch of the balances and token again.

I never know what's our strategy regarding avoiding fetching again or not. When it was with Update calls, the theory was to avoid extra update calls at all costs. But these are query calls...

dskloetd: Then it would be good if the test tested that we didn't do an additional fetch (because that's what we care about) rather than spying on the store set.
If that's not easy, can you add a comment in the test explaining that's why we don't want to additional store set?

dskloetd (approved):

frontend/src/lib/services/cketh-canisters.services.ts
dskloetd: Make it clear in the comment what exactly we try to avoid so someone later can easily check if it's still the case.

lmuntaner: Done, I added both. A comment and a test in Account page: `"should not refetch ckETH accounts if ckETH canisters are already loaded"`

lmuntaner:

frontend/src/tests/lib/routes/Accounts.spec.ts
```diff
@@ -340,6 +346,30 @@ describe("Accounts", () => {
...
+
+    render(Accounts);
+
+    await runResolvedPromises();
+
+    // It's called once when the component is mounted
+    expect(icrcLedgerApi.queryIcrcToken).toHaveBeenCalledTimes(1);
+    expect(icrcLedgerApi.queryIcrcBalance).toHaveBeenCalledTimes(1);
+
+    await runResolvedPromises();
```
dskloetd, line 365: You already ran all resolved promises on line 359 and then didn't change anything about the component or the store. There can't be any new resolved promises by the time you reach this line so I don't think anything in the test from this line onwards can be relevant.

frontend/src/tests/lib/routes/Accounts.spec.ts
```diff
@@ -340,6 +346,30 @@ describe("Accounts", () => {
     });
   });
 
+  it("should not refetch ckETH accounts if ckETH canisters are already loaded", async () => {
```
dskloetd, line 349: Does this test fail if you remove the `isNullish(storeData[CKETH_LEDGER_CANISTER_ID.toText()])` check above?

lmuntaner: Yes, it fails.

frontend/src/tests/lib/routes/Accounts.spec.ts
dskloetd: Maybe the name is bad but I made `runResolvedPromises` because I was tired of having to do multiple steps. The point is that after a single `runResolvedPromises` everything that could have happened has happened.

PR #3876, opened 2023-11-28
Include ckETH canisters in env config

dskloetd:
# Motivation

The NNS dapp needs to be aware of the ckETH ledger and index canister IDs.

# Changes

1. Include entries for `cketh_ledger` and `cketh_index` in `dfx.json`. The don't have mainnet IDs yet.
2. Update `./config.sh` to include entries for ckETH similar as for ckBTC but without minter.

# Tests

Ran `scripts/nns-dapp/test-config --update` to update golden files.

# Todos


lmuntaner (approved):
Thanks!!!

After this that's it, right?

dskloetd:
> After this that's it, right?

Still need the TypeScript code to include the env vars.

PR #3885, opened 2023-11-28
fix: warning "Void function return value is used"

peterpeterparker:
# Motivation

Editor throw a warning "Void function return value is used " because the function is not a promise.


lmuntaner (approved):
Thanks

PR #3889, opened 2023-11-28
Include cketh canisters when importing IDs from URL

dskloetd:
# Motivation

When staging a release, one of the steps is importing canister IDs from a URL so they can be used to deploy the release candidate.

# Changes

1. Include ckETH canister IDs when importing from URL in `scripts/canister_ids`.
2. Output empty string instead of failing when a field is not present in the `index.html`.

# Tests

Test was updated.

# Todos

### DIFF
--- a/scripts/canister_ids
+++ b/scripts/canister_ids
@@ -117,7 +117,7 @@ get_value_from_html() {
   pattern="$key=\"([^\"]+)\""
   # E.g. find a line like `data-own-canister-id="lf43c-fyaaa-aaaaa-aacva-cai"`
   # and outputs the value between quotes.
-  grep -oE "$pattern" "$file" | sed -E "s/$pattern/\\1/" | head -1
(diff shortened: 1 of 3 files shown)


scripts/canister_ids
```diff
@@ -117,7 +117,7 @@ get_value_from_html() {
   pattern="$key=\"([^\"]+)\""
   # E.g. find a line like `data-own-canister-id="lf43c-fyaaa-aaaaa-aacva-cai"`
   # and outputs the value between quotes.
-  grep -oE "$pattern" "$file" | sed -E "s/$pattern/\\1/" | head -1
+  grep -oE "$pattern" "$file" | sed -E "s/$pattern/\\1/" | head -1 || true
```
lmuntaner, line 120: why this change?

lmuntaner (approved):
Thanks!

Just a question but already approving

PR #3890, opened 2023-11-28
GIX-2121: Fix snsOnlyProjectStore

lmuntaner:
# Motivation

snsOnlyProjectStore should not return universes of ICRC Tokens. It's used in specifically SNS only places.

# Changes

* Fix `snsOnlyProjectStore` to check whether the current universe is an ICRC Token universe.

I know that SNS universes are actually ICRC Tokens. But in our current code, that is not the case, and until we refactor to use the same, we consider them differently.

# Tests

* Add a test case for `snsOnlyProjectStore`.

Test was failing without the fix.

# Todos

Not necessary yet.

### DIFF
--- a/frontend/src/lib/derived/sns/sns-selected-project.derived.ts
+++ b/frontend/src/lib/derived/sns/sns-selected-project.derived.ts
@@ -1,13 +1,14 @@
-import {
-  isIcrcTokenUniverseStore,
-  selectedUniverseIdStore,
-} from "$lib/derived/selected-universe.derived";
+import { selectedUniverseIdStore } from "$lib/derived/selected-universe.derived";
 import {
   snsProjectsCommittedStore,
   snsProjectsStore,
   type SnsFullProject,
 } from "$lib/derived/sns/sns-projects.derived";
-import { isUniverseCkBTC, isUniverseNns } from "$lib/utils/universe.utils";
+import {
+  snsAggregatorStore,
+  type SnsAggregatorData,
+  type SnsAggregatorStore,
+} from "$lib/stores/sns-aggregator.store";
 import type { Principal } from "@dfinity/principal";
 import { derived, type Readable } from "svelte/store";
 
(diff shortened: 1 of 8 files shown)


lmuntaner:
* main
  * **PR #3890** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3890?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3890?utm_source=stack-comment).

lmuntaner:

frontend/src/lib/derived/sns/sns-selected-project.derived.ts
```diff
@@ -12,13 +15,16 @@ import { derived, type Readable } from "svelte/store";
...
-  isUniverseNns($selectedUniverseIdStore) ||
-  isUniverseCkBTC($selectedUniverseIdStore)
-    ? undefined
-    : $selectedUniverseIdStore
+>(
+  [selectedUniverseIdStore, isIcrcTokenUniverseStore],
+  ([$selectedUniverseIdStore, isIcrcTokenUniverse]: [Principal, boolean]) =>
+    isUniverseNns($selectedUniverseIdStore) ||
+    isUniverseCkBTC($selectedUniverseIdStore) ||
+    isIcrcTokenUniverse
```
dskloetd, line 25: If I'm on a ckETH page and then I reload the page, my page universe will still be the same, but until the token universes are loaded, this logic will not know that it's not an SNS universe, right?

Can we use the aggregator data instead to know if a universe is an SNS universe?

lmuntaner: Yes, that makes more sense actually.

lmuntaner: DOne

lmuntaner:

PR #3893, opened 2023-11-28
Display pending BTC deposits above other ckBTC transactions

dskloetd:
# Motivation

We want to display incoming BTC deposits as pending transactions so the user doesn't have to stress for the full 12 confirmations whether their money is gone or not.

Note that we don't yet populate the `ckbtcPendingUtxosStore` so there is no user visible change yet.

# Changes

1. Convert `PendingUtxo`s from `ckbtcPendingUtxosStore` to `UiTransaction`s.
2. Prepend those transaction to the transactions from the ckBTC ledger and pass the combined list to `IcrcTransactionsList`.

# Tests

Unit test is added.
Also tested manually in another branch which populates the store.

# Todos

will add when the store is populated


lmuntaner (approved):
LGTM!

PR #3894, opened 2023-11-28
Extract content from ckBTCWallet & IcrcWallet === ckETH Wallet page

peterpeterparker:
# Motivation

Extract a wallet page from CkBTCWallet that can be use to create the Icrc wallet page.

# Notes

No function changes in term of loading or interacting with backend.

With this PR we've got an Icrc Wallet page that display header with balance, list of transactions and interact with the web workers that periodically sync the balance and transaction. This PR also handles loading the parameters of the url, redirect is unknow. Everything with existing code.

i.e. this is the wallet page for ckETH.

# Changes

- extract a new cmp `IcrcWalletPage` from `CkBTCWallet`
- use new cmp in `CkBTCWallet`
- add `IcrcWallet` that uses the new component
- display `IcrcWallet` on wallet if Icrc.

# Demo

demo


lmuntaner (approved):
It looks great!! Thanks!

PR #3895, opened 2023-11-28
Refactor ckBTC sync/load tokens and accounts to wallet

peterpeterparker:
# Motivation

This PR allow us in #3894 to extract a common WalletPageContent that can be used by Icrc and ckBTC wallet pages without the need of duplicating any loading logic.

# Changes

- extract sync/load functions from `ckbtc-accounts.services` -> `wallet-accounts.services`
- extract load function from `ckbtc-token.services` -> `wallet-token.services` (I'm not sure here we do not have a duplicate but moving it was a must to be able to move above functions. we can clean up in another PR if duplicate)
- rename current `ckbtc-accounts.services` to `ckbtc-uncertified-accounts.services` to resolve name clashes
- remove ckBTC in function names - e.g. `syncCkBTCAccounts` -> `syncAccounts` (one of the function that will be use in the generic Wallet Page)

No functional changes.


lmuntaner (approved):
Thanks!

PR #3897, opened 2023-11-29
GIX-2122: Send CkETH Tokens

lmuntaner:
# Motivation

User can send ICRC Tokens

It's a fairly big PR, but the functionality was better tested once all was in.

# Changes

* Add new event name `nnsIcrcTokenModal` to HTMLAttributes.
* Implement the button to open the Send modal.
* New component IcrcTokenAccountsModals to render the Icrc Token modals. For now only `IcrcTokenTransactionModal`. The type of the data needed differes from ckBTC and SNS; therefore, I couldn't easily reuse any. I bet we can further improve this, but in another PR.
* New IcrcTokenTransactionModal that renders a TransactionModal and makes the ICRC Token transfer.
* Render IcrcTokenAccountsModals in Accounts
* New icrc accounts service `icrcTransferTokens`. There is some duplication with ckBTC. We can improve this further.
* New types for icrc-accounts modals.
* New modal util `openIcrcTokenModal`.

# Tests

* Test new IcrcTokenTransactionModal
* Test that users can perform transactions from Accounts.
* Test new ICRC service `icrcTransferTokens`.
* Add new PO IcrcTokenTransactionModalPo
* Add necessary methods to POs

# Todos

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcTokenAccountsFooter.svelte
+++ b/frontend/src/lib/components/accounts/IcrcTokenAccountsFooter.svelte
@@ -4,28 +4,28 @@
   import { isNullish, nonNullish } from "@dfinity/utils";
   import { selectedIcrcTokenUniverseIdStore } from "$lib/derived/selected-universe.derived";
   import { openIcrcTokenModal } from "$lib/utils/modals.utils";
-  import type { Principal } from "@dfinity/principal";
   import { tokensStore } from "$lib/stores/tokens.store";
   import type { IcrcTokenMetadata } from "$lib/types/icrc";
   import { toastsError } from "$lib/stores/toasts.store";
+  import type { UniverseCanisterId } from "$lib/types/universe";
 
-  let ledgerCanisterId: Principal | undefined;
-  $: ledgerCanisterId = $selectedIcrcTokenUniverseIdStore;
+  let universeId: UniverseCanisterId | undefined;
+  $: universeId = $selectedIcrcTokenUniverseIdStore;
 
   let token: IcrcTokenMetadata | undefined;
-  $: token = nonNullish(ledgerCanisterId)
-    ? $tokensStore[ledgerCanisterId.toText()]?.token
+  $: token = nonNullish(universeId)
+    ? $tokensStore[universeId.toText()]?.token
     : undefined;
 
   const openSendModal = () => {
-    if (isNullish(ledgerCanisterId) || isNullish(token)) {
+    if (isNullish(universeId) || isNullish(token)) {
       toastsError({ labelKey: "error.icrc_token_load" });
       return;
     }
     openIcrcTokenModal({
       type: "icrc-send",
       data: {
-        universeId: ledgerCanisterId,
+        universeId,
         token,
         loadTransactions: false,
       },
--- a/frontend/src/lib/components/accounts/IcrcTokenAccountsFooter.svelte
+++ b/frontend/src/lib/components/accounts/IcrcTokenAccountsFooter.svelte
@@ -1,15 +1,43 @@
 <script lang="ts">
   import { i18n } from "$lib/stores/i18n";
   import Footer from "$lib/components/layout/Footer.svelte";
-  import { nonNullish } from "@dfinity/utils";
+  import { isNullish, nonNullish } from "@dfinity/utils";
   import { selectedIcrcTokenUniverseIdStore } from "$lib/derived/selected-universe.derived";
+  import { openIcrcTokenModal } from "$lib/utils/modals.utils";
+  import type { Principal } from "@dfinity/principal";
+  import { tokensStore } from "$lib/stores/tokens.store";
+  import type { IcrcTokenMetadata } from "$lib/types/icrc";
+  import { toastsError } from "$lib/stores/toasts.store";
+
+  let ledgerCanisterId: Principal | undefined;
+  $: ledgerCanisterId = $selectedIcrcTokenUniverseIdStore;
+
+  let token: IcrcTokenMetadata | undefined;
+  $: token = nonNullish(ledgerCanisterId)
+    ? $tokensStore[ledgerCanisterId.toText()]?.token
+    : undefined;
+
+  const openSendModal = () => {
+    if (isNullish(ledgerCanisterId) || isNullish(token)) {
+      toastsError({ labelKey: "error.icrc_token_load" });
+      return;
+    }
+    openIcrcTokenModal({
+      type: "icrc-send",
+      data: {
+        universeId: ledgerCanisterId,
+        token,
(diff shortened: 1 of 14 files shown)


lmuntaner:
* main
  * **PR #3897** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3897?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3897?utm_source=stack-comment).

lmuntaner:
UX:

            <a href="https://app.graphite.dev/media/video/W99H6e2WWjPgtes6VOJI/f2f6bb16-9dea-4306-90cf-bbb80f09c590.mov">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/W99H6e2WWjPgtes6VOJI/f2f6bb16-9dea-4306-90cf-bbb80f09c590.mov">Screen Recording 2023-11-29 at 09.20.14.mov</video>

lmuntaner:

frontend/src/lib/components/accounts/IcrcTokenAccountsFooter.svelte
```diff
@@ -1,15 +1,43 @@
...
+  $: ledgerCanisterId = $selectedIcrcTokenUniverseIdStore;
+
+  let token: IcrcTokenMetadata | undefined;
+  $: token = nonNullish(ledgerCanisterId)
+    ? $tokensStore[ledgerCanisterId.toText()]?.token
+    : undefined;
+
+  const openSendModal = () => {
+    if (isNullish(ledgerCanisterId) || isNullish(token)) {
+      toastsError({ labelKey: "error.icrc_token_load" });
```
peterpeterparker, line 22: does the error msg match the undefined ledger canister id?

frontend/src/lib/components/accounts/IcrcTokenAccountsFooter.svelte
```diff
@@ -1,15 +1,43 @@
...
-  import { nonNullish } from "@dfinity/utils";
+  import { isNullish, nonNullish } from "@dfinity/utils";
   import { selectedIcrcTokenUniverseIdStore } from "$lib/derived/selected-universe.derived";
+  import { openIcrcTokenModal } from "$lib/utils/modals.utils";
+  import type { Principal } from "@dfinity/principal";
+  import { tokensStore } from "$lib/stores/tokens.store";
+  import type { IcrcTokenMetadata } from "$lib/types/icrc";
+  import { toastsError } from "$lib/stores/toasts.store";
+
+  let ledgerCanisterId: Principal | undefined;
```
peterpeterparker, line 12: I would either do `let universeId: UniverseCanisterId | undefined` given that `universeId` is the parameter of the modal or rename the paramter of the modal to `ledgerCanisterId`. WDYT?

frontend/src/lib/services/icrc-accounts.services.ts
```diff
@@ -212,3 +213,34 @@ export const transferTokens = async ({
     return { blockIndex: undefined };
   }
 };
+
+export const icrcTransferTokens = async ({
```
peterpeterparker, line 217: To follow what I've done in the other PRs to avoid duplicate code, instead of this function, I suggest that we move `ckBTCTransferTokens` from `ckbtc-accounts.services` to `wallet-accounts.services` and rename it fo `transferTokens`. 

That way, we avoid code duplication and also do no inject new functions in the code base.

I can take care of this in another PR. Can you add a TODO?

frontend/src/lib/services/icrc-accounts.services.ts
```diff
@@ -212,3 +213,34 @@ export const transferTokens = async ({
...
+        identity: Identity;
+      } & IcrcTransferParams
+    ) =>
+      await icrcTransfer({
+        ...params,
+        canisterId: ledgerCanisterId,
+      }),
+    reloadAccounts: async () =>
+      await loadIcrcAccount({ ledgerCanisterId, certified: true }),
+    // TODO: Reload transactions after transfer in Wallet page
```
peterpeterparker, line 243: You can remove this TODO, the web workers take care of refreshing the transactions.

peterpeterparker (approved):
LGTM, few suggestions.

frontend/src/lib/components/accounts/IcrcTokenAccountsFooter.svelte
lmuntaner: More or less:

```
An error occurred while loading the token balance or metadata.
```

I think it's generic enough for such an edge case. It's more a TS issue. The button is not available unless that's present, but TS is not smarth enough.

frontend/src/lib/components/accounts/IcrcTokenAccountsFooter.svelte
lmuntaner: Yes, it makes sense in this case. Thanks!

frontend/src/lib/services/icrc-accounts.services.ts
lmuntaner: Done! Thanks!

frontend/src/lib/services/icrc-accounts.services.ts
lmuntaner: Oh, great. I added a comment precisely saying that.

That means we can remove them from `transferTokens` as well then, right?

dskloetd: The web worker can take 10 or 20 seconds, no?
Having the transaction appear after some time is not the same as having the transaction appear as you make it.

peterpeterparker: An Icrc Index Canister takes some times too. The ICP ledger has fast refresh rate but per default an Icrc Index canister has a refresh rate of 1min.

peterpeterparker: We can keep to the TODO, I don't mind, I would just suggest to not blindly implement a refresh transactions feature but double check that it makes sense first.

dskloetd: I've seen them appear much faster. Is that only in the test environment?
Also, the reload function has a [deliberate 4 second delay](https://github.com/dfinity/nns-dapp/blob/8d8e985a6486f799552f64bd7a348afc41a60b03/frontend/src/lib/components/accounts/IcrcWalletTransactionsList.svelte#L66) which supposedly is to give the index canister time to catch up.
Can the delay be removed if it serves no purpose? The delay is even there on initial load when no transaction was just made.

PR #3898, opened 2023-11-29
GIX-2144: Load ckETH canisters on init public

lmuntaner:
# Motivation

CkETH will also be visible for non-logged in users.

In this PR, load the ckETH canisters on public init along the SNS projects.

This is not enough. We also need to load the tokens. That will come in another PR.

# Changes

* Move `loadCkETHCanisters` to app.services and remove it from Accounts.svelte.

# Tests

* Add a test util `setCkETHCanisters`.
* The Accounts tests now need to ckETH set because it's not loading them.
* Add tests for `initAppPublicData` to check that loads the ckETH canisters accordingly.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/services/$public/app.services.ts
+++ b/frontend/src/lib/services/$public/app.services.ts
@@ -11,15 +11,12 @@ import { loadCkETHCanisters } from "../cketh-canisters.services";
  * These data can be read by any users without being signed-in.
  */
 export const initAppPublicData = (): Promise<
-  [PromiseSettledResult<void[]>, PromiseSettledResult<void[]>]
+  [PromiseSettledResult<void>, PromiseSettledResult<void>]
 > => {
-  const initNns: Promise<void>[] = [loadCkETHCanisters()];
-  const initSns: Promise<void>[] = [loadSnsProjects()];
-
   /**
-   * If Nns load but Sns load fails it is "fine" to go on because Nns are core features.
+   * If one of the promises fails, we don't want to block the app.
    */
-  return Promise.allSettled([Promise.all(initNns), Promise.all(initSns)]);
+  return Promise.allSettled([loadCkETHCanisters(), loadSnsProjects()]);
 };
 
 const syncAuthStore = async () => {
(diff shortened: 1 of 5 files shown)


lmuntaner:
* main
  * **PR #3898** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3898?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3898?utm_source=stack-comment).

lmuntaner:

frontend/src/lib/services/$public/app.services.ts
```diff
@@ -12,7 +13,7 @@ import { toastsError } from "$lib/stores/toasts.store";
 export const initAppPublicData = (): Promise<
   [PromiseSettledResult<void[]>, PromiseSettledResult<void[]>]
 > => {
-  const initNns: Promise<void>[] = [];
+  const initNns: Promise<void>[] = [loadCkETHCanisters()];
```
dskloetd, line 16: Is ckETH really "Nns". Why are initNns and initSns separate?
And why do we have these lists of lists of 1 item each?
Can we just have
```
return Promise.allSettled([loadCkETHCanisters(), loadSnsProjects()]);
```
Not important if I'm missing something or it's not simple but it seems like it should just work and simplify things.

frontend/src/tests/lib/services/_public/app.services.spec.ts
```diff
@@ -8,13 +19,48 @@ vi.mock("$lib/services/$public/sns.services", () => {
...
 
   it("should init Sns", async () => {
     await initAppPublicData();
 
     await expect(loadSnsProjects).toHaveBeenCalledTimes(1);
   });
+
+  describe("when ENABLE_CKETH is enabled", () => {
+    beforeEach(() => {
+      overrideFeatureFlagsStore.setFlag("ENABLE_CKETH", true);
```
dskloetd, line 37: How about a test where both features are disabled?

dskloetd (approved):

frontend/src/lib/services/$public/app.services.ts
lmuntaner: Done, it makes sense.

frontend/src/tests/lib/services/_public/app.services.spec.ts
lmuntaner: Done, good point!

PR #3828, opened 2023-11-23
Pass UiTransaction[] to IcrcTransactionList

dskloetd:
# Motivation

In CkBTCTransactions list I want to render multiple transactions (such as Approve + transfer) as a single transaction.
This is easier if the conversion from `IcrcTransactionData` to `UiTransaction` happens closer to the CkBTCTransactionsList.

In this PR I move the conversion from `IcrcTransactionData[]` to `UiTransaction[]` from `IcrcTransactionsList` to `CkBTCTransactionsList` and `SnsTransactionsList`.

# Changes

1. In `CkBTCTransactionsList` convert `IcrcTransactionData[]` to `UiTransaction[]` and pass that to `IcrcTransactionsList`. No longer pass `account` and `token` which were only passed to do the conversion.
2. In `SnsTransactionsList` convert `IcrcTransactionData[]` to `UiTransaction[]` and pass that to `IcrcTransactionsList`. No longer pass `governanceCanisterId`, `account` and `token` which were only passed to do the conversion.
3. Remove the conversion step an unused props from `IcrcTransactionsList`.
4. Drive-by: Remove `fallbackDescription` from `UiTransaction`, which I thought I did in a previous PR.

# Tests

1. Removed irrelevant tests from `frontend/src/tests/lib/components/accounts/IcrcTransactionsList.spec.ts`.
2. Added `createMockUiTransaction` for convenience.

# Todos

- [ ] Add entry to changelog (if necessary).
not necessary

PR #3899, opened 2023-11-29
Map the list of transactions instead of individual transactions

dskloetd:
# Motivation

We want the ckBTC wallet component to have full control over the transactions list in order to:
1. Include pending transactions mapped from `PendingUtxo`s
2. Merge some transactions into a single transaction, such as Approve+Transfer or Failed+Reimbursed.

This control was accomplished with https://github.com/dfinity/nns-dapp/pull/3828 by passing a `UITransaction[]` from `CkBTCTransactionsList` to `IcrcTransactionsList`.

Unfortunately, before we could make use of this, the effort was reversed in https://github.com/dfinity/nns-dapp/pull/3884 by passing only `mapCkbtcTransaction` (which maps individual transactions) from `CkBTCTransactionsList` to `IcrcWalletTransactionsList`.

In this PR we, once again, give `CkBTCTransactionsList` control over the full list by passing a function that maps a list of transactions instead of passing a function that maps individual transactions.

This is a pure refactoring. The PR does not yet do anything other than mapping each individual transaction.

# Changes

1. In `IcrcWalletTransactionsList` remove the `mapTransaction` (singular) prop and add the `mapTransactions` (plural) prop. It gets a default value which does the mapping that was already the default in that component.
2. From `CkBTCTransactionsList`, pass a function which maps all transactions through `mapCkbtcTransaction`.

# Tests

Existing tests pass.
Added a unit test for `IcrcWalletTransactionsList` that tests that the custom mapping function is used. It previously did not have such a test for the singular mapping function.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
@@ -2,6 +2,7 @@
 <svelte:options accessors />
 
 <script lang="ts">
+  import { i18n } from "$lib/stores/i18n";
   import type { Account } from "$lib/types/account";
   import { mapCkbtcTransaction } from "$lib/utils/icrc-transactions.utils";
   import type { UniverseCanisterId } from "$lib/types/universe";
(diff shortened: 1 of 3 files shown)


peterpeterparker (approved):
Thx

PR #3902, opened 2023-11-29
 Show skeletons while loading if mapTransactions return non-empty

dskloetd:
# Motivation

In a future PR `CkBTCTransactionsList` will add pending UTXO transactions at the top of the list, using `mapTransactions` to add additional transactions not mapped from any actual ICRC ledger transactions.
We don't want these transactions to appear before the "real" transaction list finished loading.
So while no "real" transactions have been loaded yet, we show skeletons regardless of what `mapTransactions` returns.

`IcrcTransactionsList` shows skeletons while `loading` is `true` and there are 0 transactions. So while `loading` is `true` and there are 0 transaction before mapping, we make sure to pass 0 UI transaction after mapping.

# Changes

If `transactions.length === 0 && loading`, pass an empty list to `IcrcTransactionsList`.

# Tests

Unit test is added which simulates the described scenario.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcWalletTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/IcrcWalletTransactionsList.svelte
@@ -3,19 +3,20 @@
 
 <script lang="ts">
   import type { Account } from "$lib/types/account";
-  import type { UiTransaction } from "$lib/types/transaction";
+  import type {
+    UiTransaction,
+    IcrcTransactionData,
+  } from "$lib/types/transaction";
   import {
     loadWalletNextTransactions,
     loadWalletTransactions,
   } from "$lib/services/wallet-transactions.services";
-  import type { IcrcTransactionData } from "$lib/types/transaction";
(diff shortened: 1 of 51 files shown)


frontend/src/lib/components/accounts/IcrcWalletTransactionsList.svelte
```diff
@@ -103,8 +103,14 @@
...
   });
 
+  let mappedTransactions: UiTransaction[];
+  $: mappedTransactions = mapTransactions(transactions);
+
+  // If transactions haven't loaded yet, we want to display the skeletons,
+  // even if mapTransactions returns a non-empty list.
   let uiTransactions: UiTransaction[];
-  $: uiTransactions = mapTransactions(transactions);
+  $: uiTransactions =
```
peterpeterparker, line 112: > In a future PR

How distant is that PR? before if I get it right, this PR has for effect to display the skeleton also when the user has no transaction right?

peterpeterparker (approved):
Thx

PR #3905, opened 2023-11-29
Set ckETH canister IDs in dfx.json

dskloetd:
# Motivation

Include ckETH canister IDs in mainnet module args.

# Changes

1. Set mainnet canister IDs for ckETH canisters in `dfx.json`.
2. Run `scripts/nns-dapp/test-config --update`.

# Tests

Configs were updated by `scripts/nns-dapp/test-config --update` so the canister IDs were picked up.

# Todos

not necessary


lmuntaner (approved):
Thanks

PR #3908, opened 2023-11-29
GIX-2144: Watch new IcrcToken canisters load token data

lmuntaner:
# Motivation

Non-logged in users should also have available the ICRC Tokens.

What was missing was loading the token metadata in the store.

To accomplish that, I added a listener on icrcCanistersStore that loads the tokensStore for each ledgerCanisterId added there.

# Changes

* New icrc token service `watchIcrcTokensLoadTokenData`.
* Set the watcher in the service `initAppPublicData`.
* Allow querying the token from `loadIcrcToken` for non logged in users.

# Tests

* Test that tokens are loaded in `initAppPublicData`.
* Test new service `watchIcrcTokensLoadTokenData`.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
+++ b/frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
@@ -2,7 +2,6 @@ import * as icrcLedgerApi from "$lib/api/icrc-ledger.api";
 import { watchIcrcTokensLoadTokenData } from "$lib/services/icrc-tokens.services";
 import { icrcCanistersStore } from "$lib/stores/icrc-canisters.store";
 import { tokensStore } from "$lib/stores/tokens.store";
-import { resetIdentity } from "$tests/mocks/auth.store.mock";
 import { mockToken, principal } from "$tests/mocks/sns-projects.mock";
 import { runResolvedPromises } from "$tests/utils/timers.test-utils";
 import { get } from "svelte/store";
@@ -32,7 +31,7 @@ describe("icrc-tokens.services", () => {
     });
 
     it("should load tokens when icrcCanistersStore is updated", async () => {
-      watchIcrcTokensLoadTokenData({ certified: false });
+      watchIcrcTokensLoadTokenData();
 
       icrcCanistersStore.setCanisters({
         ledgerCanisterId: ledgerCanisterId1,
@@ -48,7 +47,7 @@ describe("icrc-tokens.services", () => {
     });
 
     it("should load multiple tokens when icrcCanistersStore is updated multiple times", async () => {
-      watchIcrcTokensLoadTokenData({ certified: false });
+      watchIcrcTokensLoadTokenData();
 
       icrcCanistersStore.setCanisters({
         ledgerCanisterId: ledgerCanisterId1,
@@ -70,23 +69,6 @@ describe("icrc-tokens.services", () => {
       expect(get(tokensStore)[ledgerCanisterId2.toText()]).toEqual({
         certified: false,
         token: token2,
-      });
-    });
-
-    it("should load tokens with certified data when logged in", async () => {
-      resetIdentity();
-      watchIcrcTokensLoadTokenData({ certified: true });
-
-      icrcCanistersStore.setCanisters({
-        ledgerCanisterId: ledgerCanisterId1,
-        indexCanisterId,
-      });
-
-      await runResolvedPromises();
-
-      expect(get(tokensStore)[ledgerCanisterId1.toText()]).toEqual({
-        certified: true,
-        token: token1,
       });
     });
   });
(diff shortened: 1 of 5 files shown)


lmuntaner:
* main
  * **PR #3908** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3908?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3908?utm_source=stack-comment).

lmuntaner:

lmuntaner:
UI for non logged-in users:

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+        ledgerCanisterId: ledgerCanisterId1,
+        indexCanisterId,
+      });
+
+      icrcCanistersStore.setCanisters({
+        ledgerCanisterId: ledgerCanisterId2,
+        indexCanisterId,
+      });
+
+      await waitFor(() =>
```
dskloetd, line 63: This seems fragile. If the loading is async, is it guaranteed that if token1 is loaded, token2 is also loaded?

And is perhaps a single call to `runResolvedPromises()` enough for both to be loaded with the need for `waitFor`?

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+        })
+      );
+
+      expect(get(tokensStore)[ledgerCanisterId2.toText()]).toEqual({
+        certified: false,
+        token: token2,
+      });
+    });
+
+    it.only("should load tokens with certified data when logged in", async () => {
```
dskloetd, line 76: only?

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      resetIdentity();
```
dskloetd, line 77: Does this make a difference? I thought this is the default.
The reason you get `certified` in the store is because you passed `certified` to `watchIcrcTokensLoadTokenData`, isn't it?
I don't see how it's related to being logged in.

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      watchIcrcTokensLoadTokenData({ certified: true });
+
+      icrcCanistersStore.setCanisters({
+        ledgerCanisterId: ledgerCanisterId1,
+        indexCanisterId,
+      });
+
+      await waitFor(async () =>
```
dskloetd, line 85: Is `waitFor` needed or do `runResolvedPromises` work?

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
lmuntaner: I tried `runResolvedPromises` as well as `tick`, but non did the trick.

No, there is no guarantee that both will be loaded. I will add a wait fore below as well

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
lmuntaner: We use `queryAndUpdate` and it doesn't allow to make update calls unless the user is logged in.

It seems very arbitrary, but that's a rule we added when we wanted to easily reduce the number of update calls.

Now pretty useful and confusing...

lmuntaner:

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
lmuntaner: oops, removed

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
dskloetd: I checked out your branch and tried it for myself and I was able to replace all instances of `waitFor` with `runResolvedPromises` and all passed.

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
dskloetd: Thanks for clarifying.
The test description still seems wrong though.
The call is not certified because the user is logged in, is it?
We just make sure the user is logged in before we manually make a certified call.

By the way, do you ever plan to use this with `certified: true`? Or should we rather remove the `certified` parameter from the `watchIcrcTokensLoadTokenData` function?

dskloetd (approved):

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
lmuntaner: I can remove it actually, I guess it makes no sense for now.

frontend/src/tests/lib/services/icrc-tokens.services.spec.ts
lmuntaner: done

PR #3909, opened 2023-11-29
Add pending UTXOs to store after updateBalance

dskloetd:
# Motivation

In order to get a BTC deposit credited as ckBTC, we need to call `update_balace` on the minter.
If (and only if!) there are no new deposits with enough confirmations, the minter returns (as an error) any deposits that do not yet have enough confirmations.
This means that if there are new completed deposits, we need to call `update_balace` again to know if there are also any pending deposits.

# Changes

1. In `frontend/src/lib/services/ckbtc-minter.services.ts` call `update_balance` until we get pending transactions (if there are any) and then put them in the store.
2. Pass `universeId` to `updateBalance` and to components that call `updateBalance` because it's used as a key in the store.

# Tests

Unit tests are added.
Test manually at https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/accounts/?u=oq4ec-3maaa-aaaaa-qabeq-cai

# Todos

### DIFF
--- a/frontend/src/lib/services/ckbtc-minter.services.ts
+++ b/frontend/src/lib/services/ckbtc-minter.services.ts
@@ -8,11 +8,13 @@ import { getAuthenticatedIdentity } from "$lib/services/auth.services";
 import { queryAndUpdate } from "$lib/services/utils.services";
 import { bitcoinAddressStore } from "$lib/stores/bitcoin.store";
 import { startBusy, stopBusy } from "$lib/stores/busy.store";
+import { ckbtcPendingUtxosStore } from "$lib/stores/ckbtc-pending-utxos.store";
 import { i18n } from "$lib/stores/i18n";
 import { toastsError, toastsSuccess } from "$lib/stores/toasts.store";
 import type { CanisterId } from "$lib/types/canister";
 import { CkBTCErrorKey, CkBTCSuccessKey } from "$lib/types/ckbtc.errors";
 import type { IcrcAccountIdentifierText } from "$lib/types/icrc";
+import type { UniverseCanisterId } from "$lib/types/universe";
 import { toToastError } from "$lib/utils/error.utils";
 import { waitForMilliseconds } from "$lib/utils/utils";
 import {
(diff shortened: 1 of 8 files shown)


frontend/src/lib/services/ckbtc-minter.services.ts
```diff
@@ -99,12 +103,51 @@ export const estimateFee = async ({
...
+}): Promise<{ completed: UpdateBalanceOk; pending: PendingUtxo[] }> => {
+  const identity = await getAuthenticatedIdentity();
+  const completedUtxos: UpdateBalanceOk = [];
+
+  try {
+    // The minter only returns pending UTXOs (as an error) if there are no
+    // completed UTXOs. So we need to keep calling until it throws an error.
+    // To avoid an infinite loop we stop after 3 attempts.
+    for (let i = 0; i < 3; i++) {
+      const response = await updateBalanceAPI({
```
lmuntaner, line 123: If this fails, we go to the catch, right?

frontend/src/tests/lib/services/ckbtc-minter.services.spec.ts
```diff
@@ -202,8 +205,82 @@ describe("ckbtc-minter-services", () => {
...
+
+    it("should stop after 3 times success", async () => {
+      const updateBalanceSpy = vi
+        .spyOn(minterApi, "updateBalance")
+        .mockResolvedValue(mockUpdateBalanceOk);
+
+      const result = await services.updateBalance(params);
+
+      expect(result).toEqual({ success: true });
+      expect(updateBalanceSpy).toBeCalledTimes(3);
```
lmuntaner, line 283: Why not checking the toast success here?

lmuntaner (approved):
LGTM!! Just a couple of questions

frontend/src/lib/services/ckbtc-minter.services.ts
dskloetd: Yes, and if the failure is not `MinterNoNewUtxosError`, it's rethrown on line 131 and then caught on 186 to produce a toast.

PR #3910, opened 2023-11-29
Smoothness universes list appearance

peterpeterparker:
# Motivation

I find the appearance of the universes in the list a bit rough on the edges given that now not only Snses are displayed with a delay but also ckETH. Therefore I suggest to blur the appearance of the card.

# Changes

- `in:blur` in universe card

# Screenshots

Without blur

without

With blur

with

### DIFF
--- a/frontend/src/lib/components/universe/SelectUniverseList.svelte
+++ b/frontend/src/lib/components/universe/SelectUniverseList.svelte
@@ -4,6 +4,7 @@
   import SelectUniverseCard from "$lib/components/universe/SelectUniverseCard.svelte";
   import { createEventDispatcher } from "svelte";
   import { selectedUniverseIdStore } from "$lib/derived/selected-universe.derived";
+  import { blur } from "svelte/transition";
 
   export let role: "link" | "button" = "link";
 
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Cool!

peterpeterparker:
@dskloetd I did not refresh the screenshot to make this PR pass because, I'm not sure if that would make the CI becomes inconsistent. This PR introduces a blur effect, should we add a delay in the test that way it compares things one the visual effect is over?

dskloetd:
Is it possible to disable animations in the screenshot test?

Another question: In my experience the SNSes appear before ckETH, this makes the SNSes jump down to insert ckETH. Is this also animated now?

dskloetd:
> That's a good idea. There is a flag I can check in the code to disable the animation?

I'm not sure. I don't have experience with animations in the nns-dapp. If there isn't, you can add a feature flag for it?

I think SNSes are loaded based on aggregator data which is delivered statically, while ckETH needs to load some token metadata from a canister.

peterpeterparker:
> I think SNSes are loaded based on aggregator data which is delivered statically

Right.

I'll think about which flag we can use. Even if this universe selector as currently display vanish, I can think that we will slowly add more animation to make the dapps more smooth.

dskloetd:
Alternatively, if there is some class that is added or removed when the animation ends, we can tell Playwright to wait for that before taking the screenshot.

If we go with a feature flag, it could be ENABLE_ANIMATION (default true) and we could gate more and more animations by that flag as required.

lmuntaner:
> Alternatively, if there is some class that is added or removed when the animation ends, we can tell Playwright to wait for that before taking the screenshot.
> 
> If we go with a feature flag, it could be ENABLE_ANIMATION (default true) and we could gate more and more animations by that flag as required.

That sounds like a good idea. Or use the environment? We have an `IS_TESTNET` but we probably want to test them before releasing. So not the best idea.

peterpeterparker:
Won't be proceeeded further.

PR #3911, opened 2023-11-29
feat: Struct for partitions

bitdivine:
# Motivation
We wish to use a memory manager.  For that we need to store virtual memory IDs and address some requirements not covered by the memory manager.

# Changes
* Define a structure holding a memory manager and virtual memory IDs.
* Check the magic bytes for a memory manager, so that we don't erase candid data directly in raw memory when upgrading or downgrading.
  * Note: The `MemoryManager` does have internal code for checking whether memory has the right magic bytes and uses it to decide whether to load an existing memory manager or whether to write  anew memory manager to stable memory.  It does not support the sue case of: "If it's not a memory manager, preserve the memory".
* Support returning the raw memory, discarding the managing struct.
  *  Note: MemoryManager takes ownership of memory but does not have a method for returning the memory.  Now, it is possible in the case of canister memory to simply create another `DefaultMemoryImpl` and it will point at the same stable memory, although one might well have qualms about having multiple references to the same data.  In the case of vector memory used in tests, however, creating a new DefaultMemoryImpl allocates a new data structure, it doesn't point to the same memory.

# Tests
* Unit tests are included.

# Todos

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -276,9 +276,9 @@
       }
     },
     "node_modules/@dfinity/gix-components": {
-      "version": "3.1.0-next-2023-11-28",
-      "resolved": "https://registry.npmjs.org/@dfinity/gix-components/-/gix-components-3.1.0-next-2023-11-28.tgz",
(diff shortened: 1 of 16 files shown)


bitdivine:
Absolutely.  Thanks for the reviews.

PR #3912, opened 2023-11-29
Fix non governance navigation with additional flag

peterpeterparker:
# Motivation

ckETH is displayed on governance pages because we know if a path support governance or not but, we don't if a universe is actually meant for governance or not. It works with ckBTC because we have some additional checks but, it is not generic.

```
// We know we are on Neurons page
// We know it is not ckBTC
// We don't know (in the future) if the Icrc universe is an Icrc token related to a Sns project or just token
isNonGovernanceTokenPath(page) || !isUniverseCkBTC(canisterId)
```

Moreover we face another problem when user select ckETH and switch to proposal. If we do nothing, the selected universe is not reseted and nothing is shown on proposal. For ckBTC we can detect this and fallback to NNS but, in case of generic Icrc token or Sns, we cannot know neither.

```
// Current code:
//
// ckBTC is only available on Accounts therefore we fallback to Nns if selected and user switch to another view
    if (isUniverseCkBTC(canisterId) && !isNonGovernanceTokenPath(page)) {
      return OWN_CANISTER_ID;
}
```

Therefore, I suggest to add a flag `governance` in the universe. Open to better idea!

# Changes

# Screenshot

Issue 1 - ckETH on neuron

Capture d’écran 2023-11-29 à 16 35 28

Issue 2 - empty proposal, no fallback

Capture d’écran 2023-11-29 à 16 51 17

### DIFF
--- a/frontend/src/lib/derived/ckbtc-universe.derived.ts
+++ b/frontend/src/lib/derived/ckbtc-universe.derived.ts
@@ -10,5 +10,6 @@ export const ckBTCUniverseStore: Readable<Universe> = derived(
     canisterId: CKBTC_UNIVERSE_CANISTER_ID.toText(),
     title: $i18n.ckbtc.title,
     logo: CKBTC_LOGO,
+    governance: false,
   })
 );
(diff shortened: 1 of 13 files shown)


peterpeterparker:
Won't be proceeeded further.

PR #3916, opened 2023-11-30
Icrc receive modal on wallet page

peterpeterparker:
# Motivation

Add receive modal on Wallet page.

# Notes

I did not manage to set up the test `should reload on close receive modal` correctly and ultimately thought we should not block moving forward because of it.
@lmuntaner can you have a look at the test after merge? or before merge?

# Changes

- refactor `reloadAccount` from `CkBTCWallet` to `IcrcWallet`
- integrate `ReceiveButton` in icrc wallet footer
- connect existing reload account action with new integrated button

No function change in the reload process.

# Screenshots

tmp

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcWalletPage.svelte
+++ b/frontend/src/lib/components/accounts/IcrcWalletPage.svelte
@@ -7,10 +7,7 @@
   import { findAccount, hasAccounts } from "$lib/utils/accounts.utils";
   import { icrcAccountsStore } from "$lib/stores/icrc-accounts.store";
   import { TokenAmount, isNullish, nonNullish } from "@dfinity/utils";
-  import {
-    loadAccounts,
-    syncAccounts as syncWalletAccounts,
-  } from "$lib/services/wallet-accounts.services";
+  import { syncAccounts as syncWalletAccounts } from "$lib/services/wallet-accounts.services";
   import { toastsError } from "$lib/stores/toasts.store";
   import { replacePlaceholders } from "$lib/utils/i18n.utils";
   import { i18n } from "$lib/stores/i18n";
@@ -42,7 +39,6 @@
       return;
     }
 
-    await loadAccounts({ universeId: selectedUniverseId });
     await loadAccount(selectedUniverseId);
 
     // transactions?.reloadTransactions?.() returns a promise.
--- a/frontend/src/lib/components/accounts/IcrcWalletPage.svelte
+++ b/frontend/src/lib/components/accounts/IcrcWalletPage.svelte
@@ -7,30 +7,53 @@
   import { findAccount, hasAccounts } from "$lib/utils/accounts.utils";
   import { icrcAccountsStore } from "$lib/stores/icrc-accounts.store";
   import { TokenAmount, isNullish, nonNullish } from "@dfinity/utils";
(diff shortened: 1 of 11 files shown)


frontend/src/lib/components/accounts/IcrcWalletPage.svelte
```diff
@@ -25,13 +28,29 @@
...
 
   const goBack = (): Promise<void> => goto(AppPath.Accounts);
 
+  // e.g. is called from "Receive" modal after user click "Done"
+  export const reloadAccount = async () => {
+    if (isNullish(selectedUniverseId)) {
+      return;
+    }
+
+    await loadAccounts({ universeId: selectedUniverseId });
```
lmuntaner, line 45: Why both `load`?

frontend/src/lib/routes/Wallet.svelte
```diff
@@ -39,6 +39,7 @@
     <CkBTCAccountsModals />
   {:else if $isIcrcTokenUniverseStore}
     <IcrcTokenAccountsModals />
+    <AccountsModals />
```
lmuntaner, line 42: Can you add a TODO with GIX-2150 to remove the `IcrcTokenAccountsModals` and use only `AccountsModals`?

lmuntaner (approved):
LGTM!

Just a few minor comments.

I'll take a look after merging and finishing the feature. I agree it's not a blocker and thanks for referencing the issue GIX-2150.

frontend/src/lib/components/accounts/IcrcWalletPage.svelte
peterpeterparker: Good point. I reverted the code so that loadAccounts stays in ckBTCWallet.

frontend/src/lib/components/accounts/IcrcWalletPage.svelte
```diff
@@ -25,13 +28,29 @@
...
+    await loadAccount(selectedUniverseId);
+
+    // transactions?.reloadTransactions?.() returns a promise.
+    // However, the UI displays skeletons while loading and the user can proceed with other operations during this time.
+    // That is why we do not need to wait for the promise to resolve here.
+    reloadTransactions();
+  };
+
   export const setSelectedAccount = () => {
```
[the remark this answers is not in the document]
peterpeterparker: We should probably review when and not when to reload. I does not arm but, we can probably optimize.

The idea I have with this IcrcWallet is to refactor the CkBTCWallet that way we do not change a single thing in term of logic and functionnality. Kind of just refactoring. Afterwards we can improve.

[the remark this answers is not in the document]
peterpeterparker: Therefore suggesting to do the same in the other PR. But for sure feel free to also review the feature, not against at all.

frontend/src/lib/routes/Wallet.svelte
peterpeterparker: There is one that summarize that in `icrc-modal`

```
// TODO: https://dfinity.atlassian.net/browse/GIX-2150 to be refactored:
// 1. Integrate those types in accounts.modal.ts
// 2. Remove and integrate IcrcTokenAccountsModal into AccountsModal
// 3. Remove usage in Accounts.svelte and Wallet
export type IcrcTokenModalType = "icrc-send";
```

PR #3936, opened 2023-12-01
Pass ulps to transferTokens

dskloetd:
# Motivation

`transferTokens` currently takes a `amount: number` and optional `token` to convert the number to `e8s`.
When we use `TokenAmountV2`, the `token` can't be optional because if we need to use a fallback we can't be certain that we use the token when we have to.

So instead pass `ulps` to `transferToken` and convert on the calling side. The calling side knows better whether it's safe to use `numberToE8s` or whether numbsToUlps` with `token` should be used.

This PR also includes most of the changes from https://github.com/dfinity/nns-dapp/pull/3931

# Changes

1. Change the parameters of `transferToken` to accept `amountUlps: bigint` and don't use `numberToE8s`.
2. Change functions that used to share the params type with `transferTokens` but which now accept `amount: number` instead of `amountUlps: bigint` to explicitly state their param types and convert amount, either with `numberToE8s` for SNS or ckBTC, or with `numberToUlps` for IcrcToken.
3. Add `numberToUlps`, which requires 8 decimal places until we use `TokenAmountV2`.

# Tests

1. Copied the unit test from https://github.com/dfinity/nns-dapp/pull/3935 but adjust to start failing when amountToUlps start working instead of skipping that test.
2. Other unit tests are adjust to work with the new param types.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/services/ckbtc-convert.services.ts
+++ b/frontend/src/lib/services/ckbtc-convert.services.ts
@@ -42,10 +42,9 @@ import {
 import type { IcrcTransferTokensUserParams } from "./icrc-accounts.services";
 import { loadWalletTransactions } from "./wallet-transactions.services";
 
-export type ConvertCkBTCToBtcParams = Omit<
-  IcrcTransferTokensUserParams,
-  "source"
-> & {
+export type ConvertCkBTCToBtcParams = {
+  destinationAddress: string;
+  amount: number;
   universeId: UniverseCanisterId;
   canisters: CkBTCAdditionalCanisters;
   updateProgress: (step: ConvertBtcStep) => void;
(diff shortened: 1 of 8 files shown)


frontend/src/lib/services/ckbtc-convert.services.ts
```diff
@@ -236,13 +235,15 @@ const retrieveBtcAndReload = async ({
...
   updateProgress,
   blockIndex,
-}: Omit<IcrcTransferTokensUserParams, "source"> &
-  Partial<Pick<IcrcTransferTokensUserParams, "source">> & {
-    universeId: UniverseCanisterId;
-    canisters: CkBTCAdditionalCanisters;
-    updateProgress: (step: ConvertBtcStep) => void;
-    blockIndex?: bigint;
-  }): Promise<{
+}: {
```
lmuntaner, line 238: Thanks!

lmuntaner (approved):
Thanks!

PR #3939, opened 2023-12-03
Fix errors with ckETH decimals

dskloetd:
# Motivation

# Changes

# Tests

# Todos

### DIFF
--- a/frontend/src/lib/utils/icrc-transactions.utils.ts
+++ b/frontend/src/lib/utils/icrc-transactions.utils.ts
@@ -20,6 +20,7 @@ import { encodeIcrcAccount } from "@dfinity/ledger-icrc";
 import type { Principal } from "@dfinity/principal";
 import {
   TokenAmount,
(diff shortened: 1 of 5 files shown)


frontend/src/lib/utils/icrc-transactions.utils.ts
```diff
@@ -286,7 +285,7 @@ export const mapCkbtcPendingUtxo = ({
     isPending: true,
     headline: i18n.ckbtc.receiving_btc,
     otherParty: i18n.ckbtc.btc_network,
-    tokenAmount: TokenAmount.fromE8s({
+    tokenAmount: TokenAmountV2.fromUlps({
```
dskloetd, line 288: ckBTC uses e8s so this one isn't necessary.
If it doesn't cause extra contagion, it's good to have, but that's the reason I didn't have it in my commit.

lmuntaner (approved):

PR #3942, opened 2023-12-04
Chore: Use decimals to get test tokens

lmuntaner:
# Motivation

# Changes

# Tests

# Todos

### DIFF
--- a/frontend/src/lib/services/dev.services.ts
+++ b/frontend/src/lib/services/dev.services.ts
@@ -18,7 +18,7 @@ import {
 } from "$lib/stores/sns-accounts.store";
 import type { IcpAccount } from "$lib/types/account";
 import type { Principal } from "@dfinity/principal";
-import { nonNullish } from "@dfinity/utils";
+import { nonNullish, type Token } from "@dfinity/utils";
 import { get } from "svelte/store";
 import { syncAccounts } from "./icp-accounts.services";
 import { loadIcrcAccount } from "./icrc-accounts.services";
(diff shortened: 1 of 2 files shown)


lmuntaner:
@dskloetd please review

frontend/src/lib/services/dev.services.ts
```diff
@@ -115,7 +117,7 @@ export const getIcrcTokens = async ({
   }
 
   await acquireIcrcTokens({
-    e8s: BigInt(tokens * E8S_PER_ICP),
+    e8s: BigInt(tokens * 10 ** token.decimals),
```
dskloetd, line 120: Why not use `numberToUlps`?

By the way, when you multiply large numbers, you should make them `bigint`s first, otherwise you store an intermediate result in a type that can't hold that result in full precision and relying on the rounding not to mess things up.
```
BigInt(tokens) * 10n ** BigInt(token.decimals)
```

dskloetd, line 120: Can we rename the parameter of `acquireIcrcTokens` to `ulps`? Or is that too contagious to include in this PR?

lmuntaner: I forgot about the helper, added it here.

lmuntaner: yes, good point, done

frontend/src/lib/services/dev.services.ts
```diff
@@ -115,7 +118,7 @@ export const getIcrcTokens = async ({
   }
 
   await acquireIcrcTokens({
-    e8s: BigInt(tokens * E8S_PER_ICP),
+    ulps: numberToUlps({ amount: tokens, token }),
```
dskloetd, line 121: FYI, this violates your proposed rule not to use plain numbers in services but only in components.
Up to you if you already want to use that rule.

lmuntaner: Not necessary yet.

PR #3946, opened 2023-12-04
Remove ICRC-1 neuron staking logic

dskloetd:
# Motivation

ICRC-1 transfers use a different memo.
We use the memo when staking an NNS neuron to make sure we can complete the transaction.
So currently, using ICRC-1 for neuron staking wouldn't be safe.
We should move directly to ICRC-2.

# Changes

Remove the logic for using ICRC-1 when staking a neuron.

The feature flag itself will be clean up, together with some other unused feature flags, in a separate PR.

# Tests

Corresponding tests have been removed.

# Todos


lmuntaner (approved):
Thanks!

PR #3950, opened 2023-12-04
Remove .gitattributes

dskloetd:
# Motivation

We have `/CHANGELOG* merge=union` in `.gitattributes`.
This means we have fewer merge conflicts because git will just happily add everything together in `CHANGELOG*`.
But it has caused change log entries that were already remove to come back many times.
Removing this might make merging PRs harder, but I think it's worth a try, given how many problems it has already caused.
We can always revert the change.

# Changes

Remove `.gitattributes`

# Tests

no

# Todos


lmuntaner (approved):
I agree. Thanks!

PR #3951, opened 2023-12-04
Use page objects in BitcoinAddress.spec.ts

dskloetd:
# Motivation

Make the test more maintainable.

# Changes

1. Add a test ID to `BitcoinAddress`
2. Add `BitcoinAddressPo`.
3. Use `BitcoinAddressPo` in `BitcoinAddress.spec.ts`.
4. Remove `waitFor` from `BitcoinAddress.spec.ts`.

# Tests

only

# Todos

- [ ] Add entry to changelog (if necessary).
not necessary

PR #3953, opened 2023-12-05
Add new info card to ckBTC wallet

dskloetd:
# Motivation

To implement the new design with ckBTC UI improvements from [Figma](https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=475-5799&mode=design&t=NWId7h3xZKkR43qb-4).

# Changes

1. Add a new "info card" slot to the `IcrcWalletPage` component.
2. Move the `BitcoinAddress` component from the `"page-content"` slot to the new `"info-card"` slot.
3. Update the `BitcoinAddress` component with additional text, a QR code and the BTC address.
4. Use the refresh balance component as a button instead of inline.
5. Use CSS grid to place the QR code on the side or inline depending on the viewport.
6. Remove "Refresh Balance" button from CkBTCWallet heading, since we have the same button in the info card.
7. Add additional getters to page object.
8. Test that QR code, skeleton and address are there when they should be and not when they shouldn't.
9. Test that the component is used in the `CkBTCWallet` component.

# Tests

Unit tests updated.
Tested manually at: https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/wallet/?u=oq4ec-3maaa-aaaaa-qabeq-cai&account=nrqjo-jlkev-i2w42-r4ek4-jz6wm-tpflh-oyr5y-wql2a-knuwm-hr24a-gqe

Before:
image

After:
image

# Todos


dskloetd:
Artem asked for some changes, so the screenshot doesn't look exactly how it will be anymore.
He'll also give slightly different text, which I will still update and then I will update the DevEnv canister.

frontend/src/lib/components/accounts/BitcoinAddress.svelte
```diff
@@ -100,7 +147,65 @@
...
+      align-items: center;
+      gap: var(--padding);
+      margin: 0;
+    }
+  }
+
+  .qr-code {
+    grid-area: qr;
+    background: white;
+    width: calc(16 * var(--padding));
```
lmuntaner, line 179: Why 16?

frontend/src/lib/components/accounts/BitcoinAddress.svelte
```diff
@@ -100,7 +147,65 @@
...
+    height: calc(16 * var(--padding));
+    border: var(--padding) solid white;
+    border-radius: var(--border-radius);
+
+    .logo {
+      width: calc(5 * var(--padding));
```
lmuntaner, line 185: why 5?

frontend/src/lib/i18n/en.json
```diff
@@ -990,10 +990,12 @@
     "about_thirty_minutes": "Typically 30-60min",
     "transaction_success_about_thirty_minutes": "Transaction completed. Estimated time ≈30 minutes.",
     "loading_address": "Loading the address",
-    "incoming_bitcoin_network_part_1": "Incoming Bitcoin network transactions require $min confirmations. Then click",
-    "incoming_bitcoin_network_part_2": "to update your ckBTC balance. Check status on a",
+    "receive_btc_title": "Receive from Bitcoin Network (BTC)",
+    "ckbtc_buzz_words": "ckBTC — a multi-chain bitcoin twin, trustlessly created by chain-key cryptography and Internet Computer smart contracts that directly hold raw bitcoin.",
```
lmuntaner, line 994: "ckbtc_buzz_words" 😂

frontend/src/tests/lib/components/accounts/BitcoinAddress.spec.ts
```diff
@@ -181,7 +195,7 @@ describe("BitcoinAddress", () => {
       const po = await renderComponent();
 
       expect(await po.getText()).toContain(
-        "Incoming Bitcoin network transactions require  confirmations. Then click"
+        "Incoming Bitcoin network transactions require  confirmations."
```
lmuntaner, line 198: Is it correct like this? There are two spaces between "require  confirmations"

frontend/src/lib/components/accounts/BitcoinAddress.svelte
```diff
@@ -51,34 +53,79 @@
...
+          <Copy value={btcAddress} />
+        </div>
+      {:else}
+        <div class="skeleton">
+          <SkeletonText />
+        </div>
+      {/if}
+    </div>
+  </div>
+  <div class="content-cell-details">
```
lmuntaner, line 119: This is inside the grid, but I don't see it being placed anywhere. Should we make it explicit instead of relying on how the browser will implement it?

I don't know whether there are differences among browsers, but it feels safer to also place it within the grid.

lmuntaner (approved):
Just a few comments, but looks good to me. Thanks!

frontend/src/lib/components/accounts/BitcoinAddress.svelte
dskloetd: I'm not really sure how to answer this other than it's the integer multiple of 8 closest to what we have in Figma.
Should I just put a number instead of a multiple of `var(--padding)`?
Should I make a constant for it here in the file and then immediately use that constant?
I'm trying to do what I see in other places, for example here: https://github.com/dfinity/nns-dapp/blob/4619857a316be5b47a950c6b8461ffcaa1a3dde3/frontend/src/lib/components/accounts/ReceiveAddressQRCode.svelte#L105
But doing CSS always feels more like hacking than engineering to me, so if you have an opinion on what to do instead, please share.

frontend/src/tests/lib/components/accounts/BitcoinAddress.spec.ts
dskloetd: That's what `innerText` gives me. It's because it renders `require ${min_confirmations} confirmations` where `${min_confirmations}` is empty in this test (before it's loaded).
I guess the `exact: false` that was there before https://github.com/dfinity/nns-dapp/pull/3951 took care of that but now we use page objects to get the actual text.
I'm not sure why we even want to render it like this but it was like that already so I kept it.

frontend/src/lib/components/accounts/BitcoinAddress.svelte
dskloetd: Good point. It was in the `address-section` grid area before but I moved it out to get better spacing.
I moved it back in and now instead made 2 separate `<div class="content-cell-details">s inside the grid area.

dskloetd:
@peterpeterparker 

CSS is challenging. I'm trying to use the `content-cell-island` from Gix components to avoid putting too much custom styling. If I'm using it wrong, or I should be using something else, can you comment in the code?

I've already put extra space below the paragraph before the QR code. About the rest Artem said it could be better but we should prefer using standard "components" and not change those standard component until we do a whole UI overhaul.

Regarding the colors, please have a look at the [Figma design](https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=708-7440&mode=design&t=NWId7h3xZKkR43qb-4), and let me know if you think something is different from that.

frontend/src/tests/lib/components/accounts/BitcoinAddress.spec.ts
lmuntaner: Ok, understood.

frontend/src/lib/components/accounts/BitcoinAddress.svelte
lmuntaner: Ok, that's good as well.

PR #3954, opened 2023-12-05
Create a dummy global config

bitdivine:
# Motivation
The docker build breaks when there is no global config available.  None is available for the cache-filling workflow.

A simple, reliable solution is to create the desired file.  A nicer solution is to make Docker accept that the file may be absent.  The latter is typically done with a hack - docker requires arguments to be present unless the path is a wildcard or charset, so it may be done, cautiously, with something like this:

```
COPY global-config.jso[n] /root/.config/dfx/networks.json
```

However these always need to be tested carefully in all the places where the dockerfile is run.  I don't  have the time for that now.  I can test it slowly over the next few days and _if_ it all works out seamlessly, submit a PR with that in a few days.

# Changes
- Create the requested file in the cache-filling workflow.

# Tests
- See CI

# Todos


lmuntaner (approved):
This solution looks good enough for now.

I wouldn't spend any more time on this, there are other priorities.

PR #3958, opened 2023-12-05
Require an iterator over account db entries

bitdivine:
# Motivation
We need to be able to iterate over all entries in an accounts database.

# Changes
- Add `iter()` to the `AccountsDb` trait.

# Tests
None - I think this is pretty trivial!

# Todos

### DIFF
--- a/rs/backend/src/accounts_store/schema/proxy.rs
+++ b/rs/backend/src/accounts_store/schema/proxy.rs
@@ -40,24 +40,35 @@ impl AccountsDbBTreeMapTrait for AccountsDbAsProxy {
 }
 
 impl AccountsDbTrait for AccountsDbAsProxy {
+    // Inserts into all the underlying databases.
     fn db_insert_account(&mut self, account_key: &[u8], account: Account) {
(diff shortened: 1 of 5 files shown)


rs/backend/src/accounts_store/schema/proxy.rs
```diff
@@ -40,24 +40,35 @@ impl AccountsDbBTreeMapTrait for AccountsDbAsProxy {
 }
 
 impl AccountsDbTrait for AccountsDbAsProxy {
+    // Inserts into all the underlying databases.
```
frederikrothenberger, line 43: If you use `///` it becomes a [rustdoc](https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html) comment.

frederikrothenberger (approved):
LGTM!
On small comment.

rs/backend/src/accounts_store/schema/proxy.rs
bitdivine: Thanks!  These were intended to be `///`.

PR #3960, opened 2023-12-05
GIX-2150: Remove E8S_PER_ICP part 2

lmuntaner:
# Motivation

# Changes

# Tests

# Todos


lmuntaner:
* main
  * **PR #3960** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3960?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3960?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

frontend/src/lib/services/dev.services.ts
```diff
@@ -71,13 +74,14 @@ export const getTokens = async ({
   const store: SnsAccountsStoreData = get(snsAccountsStore);
   const { accounts } = store[rootCanisterId.toText()];
   const main = accounts.find((account) => account.type === "main");
+  const token = get(tokensStore)[rootCanisterId.toText()]?.token;
 
-  if (!main) {
-    throw new Error("No account found to send tokens");
+  if (!main || !token) {
+    throw new Error("No account or token found to send tokens");
```
dskloetd, line 80: If would make it 2 if blocks with 2 errors, so the error is more meaningful when it happens.

frontend/src/lib/services/dev.services.ts
```diff
@@ -88,13 +92,14 @@ export const getTokens = async ({
...
   const identity = await getAuthenticatedIdentity();
+  const token = get(tokensStore)[CKBTC_UNIVERSE_CANISTER_ID.toText()]?.token;
   const btcAddress = await getBTCAddress({
     canisterId: CKBTC_MINTER_CANISTER_ID,
     identity,
   });
   await receiveMockBtc({
     btcAddress,
-    amountE8s: BigInt(amount * E8S_PER_ICP),
+    amountE8s: numberToUlps({ amount, token }),
```
dskloetd, line 102: This actually generates a BTC UTXO.
Should we make the implicit assumption that BTC has the same number of decimals as ckBTC?
While it's the case, it's not strictly necessary so I'm a bit conflicted here about using decimals from ckBTC vs. hardcoding 8 for BTC.

frontend/src/lib/utils/neuron.utils.ts
```diff
@@ -858,13 +863,15 @@ export const neuronVoting = ({
 export const validTopUpAmount = ({
   amount,
   neuron,
+  token,
 }: {
   neuron: NeuronInfo;
   amount: number;
+  token: Token;
```
dskloetd, line 870: Isn't it strange to pass a token while this can only be used for ICP?
Maybe just pass ICPToken to`numberToUlps` from here?
Or even have specific function for ICP-only context?

frontend/src/lib/services/dev.services.ts
lmuntaner: Mhm, that's a good point. I assumed it would be the same number of decimals because it uses the ledger, but it's true that it's not necessarily the case.

I will add a constant with a "fake" btc token in the same file.

frontend/src/lib/utils/neuron.utils.ts
lmuntaner: Yes, I also thought about it.

It's ICP-only because of the `NeuronInfo` type already.

I'll pass IcpToken from here.

frontend/src/tests/lib/services/dev.services.spec.ts
```diff
@@ -10,7 +10,7 @@ describe("dev-services", () => {
   });
 
   it("should throw an error if the environment is not testnet", async () => {
-    await expect(getICPs(0.00000002)).rejects.toThrowError(
+    await expect(getICPs(2)).rejects.toThrowError(
```
lmuntaner, line 13: This discovered an error upstream in TokenAmountV2

I added a test upstream in ic-js that we should fix.

https://github.com/dfinity/ic-js/pull/496

This test, has nothing to do with fixing that; therefore, I don't want it to block merging this PR.

lmuntaner (commented):

frontend/src/lib/services/dev.services.ts
```diff
@@ -52,7 +52,7 @@ export const getICPs = async (icps: number) => {
   }
 
   await acquireICPTs({
-    e8s: BigInt(icps * E8S_PER_ICP),
+    e8s: numberToUlps({ amount: icps, token: ICPToken }),
```
dskloetd, line 55: I feel a bit conflicted about this.
This looks dangerous because you're passing ulps to something that only accepts e8s.
On the other hand, we know that ICPToken has decimals = 8.
But if we're relying on that, should we really use a generic ulps function, rather than hard coding `numberToE8s`?
What do you think?

frontend/src/lib/services/dev.services.ts
```diff
@@ -71,13 +71,17 @@ export const getTokens = async ({
...
   if (!main) {
     throw new Error("No account found to send tokens");
   }
+  if (!token) {
+    throw new Error("No token found to send tokens");
+  }
 
   await acquireSnsTokens({
-    e8s: BigInt(tokens * E8S_PER_ICP),
+    e8s: numberToUlps({ amount: tokens, token }),
```
dskloetd, line 84: This looks dangerous because you're passing ulps to something that only accepts e8s.
Can you add a TODO on the parameter name in the function to change it to `ulps`?

frontend/src/lib/services/dev.services.ts
```diff
@@ -86,6 +90,15 @@ export const getTokens = async ({
   await loadSnsAccounts({ rootCanisterId });
 };
 
+// Not clear whether BTC should use the same token as ckBTC, that's why we have a separate token here.
+// Not exported to not be used outside the dev services.
+const BTC_TOKEN = {
+  decimals: 8,
+  id: "btc",
```
dskloetd, line 97: What is this for?
The `Token` type doesn't have an `id` field.
https://github.com/dfinity/ic-js/blob/ac1bbf01f6f53d441474d02784c0aa6acf4c74f4/packages/utils/src/parser/token.ts#L106

frontend/src/tests/lib/utils/token.utils.spec.ts
```diff
@@ -499,6 +499,8 @@ describe("token-utils", () => {
       expect(numberToUlps({ amount: 1, token })).toBe(100_000_000n);
       expect(numberToUlps({ amount: 3.14, token })).toBe(314_000_000n);
       expect(numberToUlps({ amount: 0.14, token })).toBe(14_000_000n);
+      // TODO: enable after fixing it upstream in ic-js
```
dskloetd, line 502: Can you enable it now but expect it to throw, so it's impossible to forget?

frontend/src/lib/services/dev.services.ts
lmuntaner: That it uses only e8s is a matter of the decimals, which is information in the token.

I think we should remove all reference to e8s even though it might use only e8s.

Using ICPToken in the wrong place seems less likely than using an e8s helper.

frontend/src/lib/services/dev.services.ts
lmuntaner: Oops, that was Copilot filling up the data.

frontend/src/tests/lib/utils/token.utils.spec.ts
lmuntaner: Good idea!

frontend/src/lib/services/dev.services.ts
lmuntaner: No need for a todo, I can change the name of the field.

lmuntaner:

dskloetd (approved):

PR #3963, opened 2023-12-05
Fix function name

bitdivine:
# Motivation
The name of the help text function is incorrect in the migration test, so the help does not include the intended text.

# Changes
- Rename the help text function to the correct value.

# Tests
- Run locally, `$ scripts/nns-dapp/migration-test --help` now shows the desired text.

# Todos

  - Too minor to warrant a changelog entry


lmuntaner (approved):
👍

PR #3964, opened 2023-12-06
Update aggregator

sa-github-api:
# Motivation
We would like to parse all the latest SNS data.

# Changes
* Updated the Rust code derived from `.did` files in the aggregator.

# Tests


bitdivine (approved):
No significant change but that is useful information too, I guess!

PR #3966, opened 2023-12-06
GIX-2086: Change UserTokenBase type

lmuntaner:
# Motivation

We need to support a new type of Row in the TokensTable.

It's an action that covers the whole row and has a label. The design is [here](https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=408-2886&mode=design&t=wRdtQmzMVgBEbjOF-0).

The idea is to use a UserTokenBase with actions.

Yet, just the action is not enough because we'll need also a label for the action.

In this PR, I change the type of the field `actions` in `UserTokenBase` to `UserTokenActionData[]` with a `type` and a `label`.

# Changes

* Convenient new type UserTokenTableData for the list of data for TokensTable.
* Change type `actions` in `UserTokenBase` to `UserTokenActionData`.
* Change access to the action type in TokensTableRow.

# Tests

* Adapt tests and mocks to new type.
* No new functionality, no new tests.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/types/tokens-page.ts
+++ b/frontend/src/lib/types/tokens-page.ts
@@ -8,12 +8,17 @@ export enum UserTokenAction {
   Receive = "receive",
 }
 
+export type UserTokenActionData = {
+  type: UserTokenAction;
+  label?: string;
+};
+
 export type UserTokenBase = {
   universeId: Principal;
   title: string;
   subtitle?: string;
   logo: string;
-  actions: UserTokenAction[];
+  actions: UserTokenActionData[];
 };
 
 /**
@@ -34,3 +39,7 @@ export type UserTokenData = UserTokenBase & {
   // Fees are included in the metadata of ICRC tokens, but this is not a list of only ICRC tokens
   fee: TokenAmount;
 };
+
+export type UserTokenTableData = Array<
+  UserTokenData | UserTokenLoading | UserTokenBase
+>;
--- a/frontend/src/lib/types/tokens-page.ts
+++ b/frontend/src/lib/types/tokens-page.ts
@@ -40,6 +40,4 @@ export type UserTokenData = UserTokenBase & {
   fee: TokenAmount;
 };
 
-export type UserTokenTableData = Array<
-  UserTokenData | UserTokenLoading | UserTokenBase
->;
+export type UserTokenTableData = Array<UserTokenData | UserTokenLoading>;
--- a/frontend/src/lib/utils/user-token.utils.ts
+++ b/frontend/src/lib/utils/user-token.utils.ts
@@ -1,11 +1,7 @@
-import type {
-  UserTokenBase,
-  UserTokenData,
-  UserTokenLoading,
-} from "$lib/types/tokens-page";
+import type { UserTokenData, UserTokenLoading } from "$lib/types/tokens-page";
 
 export const isUserTokenData = (
-  userToken: UserTokenData | UserTokenLoading | UserTokenBase
+  userToken: UserTokenData | UserTokenLoading
 ): userToken is UserTokenData => {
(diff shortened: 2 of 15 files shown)


lmuntaner:
* main
  * **PR #3966** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3966?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3966?utm_source=stack-comment).

dskloetd:
Does it really need to be a row in the table? Wouldn't it make more sense to just render it as a component below the table (with 0 margin and no rounded corners between).

frontend/src/lib/types/tokens-page.ts
```diff
@@ -8,12 +8,17 @@ export enum UserTokenAction {
...
+  label?: string;
+};
+
 export type UserTokenBase = {
   universeId: Principal;
   title: string;
   subtitle?: string;
   logo: string;
-  actions: UserTokenAction[];
+  actions: UserTokenActionData[];
```
dskloetd, line 21: Here we would never use the label, right? So why not just keep using `UserTokenAction` instead of `UserTokenActionData`?

frontend/src/lib/utils/user-token.utils.ts
```diff
@@ -1,7 +1,11 @@
-import type { UserTokenData, UserTokenLoading } from "$lib/types/tokens-page";
+import type {
+  UserTokenBase,
+  UserTokenData,
+  UserTokenLoading,
+} from "$lib/types/tokens-page";
 
 export const isUserTokenData = (
-  userToken: UserTokenData | UserTokenLoading
+  userToken: UserTokenData | UserTokenLoading | UserTokenBase
```
dskloetd, line 8: `UserTokenData` and `UserTokenLoading` both extend `UserTokenBase` so
```
userToken: UserTokenData | UserTokenLoading | UserTokenBase
```
is the same as just
```
userToken: UserTokenBase
```

dskloetd:
I don't think removing the rounded corners to allow it to fit next to another component, is a hack.
But introducing a whole new type of row, just to render a button at the bottom, I do think is a hack.
But having a separate prop to rendering something at the bottom of the table is fine with me.

frontend/src/lib/types/tokens-page.ts
lmuntaner: No, I think that this could very well happen in the future if instead of an icon, we want a button with text for an action.

If I use the prop to render something at the bottom, I would use the type `UserTokenBase`. It might not be shown in the UI, but the data will be useful to open the modal of any action.

We could even consider the possibility of having multiple actions at the bottom of a row.

lmuntaner:

lmuntaner:
> I don't think removing the rounded corners to allow it to fit next to another component, is a hack. But introducing a whole new type of row, just to render a button at the bottom, I do think is a hack. But having a separate prop to rendering something at the bottom of the table is fine with me.

I meant removing corners with a prop.

But anyway, when I said it I liked it as well. A prop for bottom row makes sense to me.

frontend/src/lib/types/tokens-page.ts
dskloetd: Why would you use `UserTokenBase` to render 1 button?
You would only need `UserTokenActionData`, no? Why have 4 extra fields?

lmuntaner (commented):
We are using a slot for the last row action. Therefore, this is not needed and a premature optimization, and the root of all evil.

PR #3967, opened 2023-12-06
Rename CkBTCWalletActions to CkBTCUpdateBalanceButton

dskloetd:
# Motivation

`CkBTCWalletActions` only contains a single button and is only used inside the ckBTC info card and not in a generic wallet location.
So the name doesn't make much sense anymore.

# Changes

1. Rename CkBTCWalletActions to CkBTCUpdateBalanceButton
2. Fix `CkBTCWalletFooter.spec.ts` which has a function named `renderWalletActions`, which renders the footer.

# Tests

updated

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #3968, opened 2023-12-06
Rename BitcoinAddress to CkBTCInfoCard

dskloetd:
# Motivation

The component defined in `BitcoinAddress.svelte` now renders the contents of the ckBTC info card added in https://github.com/dfinity/nns-dapp/pull/3953
It already was more than just a Bitcoin address before that PR but now even more so.

# Changes

Rename `BitcoinAddress` to `CkBTCInfoCard`.

# Tests

updated

# Todos

not necessary


lmuntaner (approved):
Thanks!

I was actually going to ask for a rename in the PR with the new UI.

PR #3971, opened 2023-12-06
GIX-2150: Remove E8S_PER_ICP part 3

lmuntaner:
# Motivation

Remove `E8S_PER_ICP` from NNS Dapp.

In this PR, remove usage in SNS neurons and use it internally in token.utils

# Changes

* Use `ulpsToNumber` instead of dividing by `E8S_PER_ICP` in two components.
* Local variable `E8S_PER_ICP` in token.utils instead of importing the constant.

# Tests

No new functionality, tests still pass.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/lib/modals/sns/neurons/SnsStakeNeuronModal.svelte
+++ b/frontend/src/lib/modals/sns/neurons/SnsStakeNeuronModal.svelte
@@ -16,8 +16,8 @@
   import { snsParametersStore } from "$lib/stores/sns-parameters.store";
   import { mapNervousSystemParameters } from "$lib/utils/sns-parameters.utils";
   import type { SnsNervousSystemParameters } from "@dfinity/sns";
-  import { E8S_PER_ICP } from "$lib/constants/icp.constants";
   import { nonNullish } from "@dfinity/utils";
+  import { ulpsToNumber } from "$lib/utils/token.utils";
 
   export let token: Token;
   export let rootCanisterId: Principal;
@@ -44,9 +44,10 @@
   let minimumStake: number | undefined;
   $: minimumStake =
     parameters !== undefined
-      ? Number(
-          mapNervousSystemParameters(parameters).neuron_minimum_stake_e8s
-        ) / E8S_PER_ICP
+      ? ulpsToNumber({
+          ulps: mapNervousSystemParameters(parameters).neuron_minimum_stake_e8s,
+          token,
+        })
       : undefined;
   let checkMinimumStake: ValidateAmountFn;
   $: checkMinimumStake = ({ amount }) => {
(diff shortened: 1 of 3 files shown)


lmuntaner:
* main
  * **PR #3971** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3971?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3971?utm_source=stack-comment).

frontend/src/lib/modals/sns/neurons/SnsStakeNeuronModal.svelte
```diff
@@ -44,9 +44,10 @@
   let minimumStake: number | undefined;
   $: minimumStake =
     parameters !== undefined
-      ? Number(
-          mapNervousSystemParameters(parameters).neuron_minimum_stake_e8s
-        ) / E8S_PER_ICP
+      ? ulpsToNumber({
+          ulps: mapNervousSystemParameters(parameters).neuron_minimum_stake_e8s,
```
dskloetd, line 48: Given that the name `neuron_minimum_stake_e8s` comes from the candid interface, we should assume this to be in e8s, regardless of the number of decimals, in the token, wouldn't you aggree?

lmuntaner: As discussed, I believe that a single rule "Do not use E8S_PER_ICP" is easier to follow than the rule, use E8S_PER_ICP only if the value is and will always be in e8s.

dskloetd: I'm fine with not using E8S_PER_ICP. Because this is an SNS component and has nothing to do with ICP.

But this parameter is part of a public API and the parameter literally has e8s in the name. If I was in the NNS team trying to support more decimals for SNS tokens, I would also argue that this parameter must remain in e8s because that's what the name says. So it's absolutely wrong to assume that this parameter matches the decimals of the SNS token and at the same time assume that that might not be 8.
So if you want to make the code here generic, it should correctly convert from e8s to ulps with the given number of decimals.

Now this is probably overkill because there are no plans to support SNS with decimals !== 8. This is why I think it might be better to consistently stick to e8s for SNS for now than to have some strange complicated mix.

lmuntaner: I'm not so attached to this and I'd rather spend time on My Tokens page actually. I want to finish it before christmas holidays.

I'll close this PR, it's not a biggie and we'll talk about the best approach without an open PR to rush us into decisions.

PR #3972, opened 2023-12-06
GIX-2150: Use TokenAmountV2 part 1

lmuntaner:
# Motivation

# Changes

# Tests

# Todos

### DIFF
--- a/frontend/src/lib/pages/SnsWallet.svelte
+++ b/frontend/src/lib/pages/SnsWallet.svelte
@@ -21,7 +21,12 @@
     snsOnlyProjectStore,
     snsProjectSelectedStore,
   } from "$lib/derived/sns/sns-selected-project.derived";
-  import { TokenAmount, isNullish, nonNullish } from "@dfinity/utils";
+  import {
+    TokenAmount,
+    TokenAmountV2,
+    isNullish,
+    nonNullish,
+  } from "@dfinity/utils";
   import { selectedUniverseStore } from "$lib/derived/selected-universe.derived";
   import { loadSnsAccountTransactions } from "$lib/services/sns-transactions.services";
   import { replacePlaceholders } from "$lib/utils/i18n.utils";
@@ -34,6 +39,7 @@
   import WalletPageHeading from "$lib/components/accounts/WalletPageHeading.svelte";
   import { snsSelectedTransactionFeeStore } from "$lib/derived/sns/sns-selected-transaction-fee.store";
   import IC_LOGO from "$lib/assets/icp.svg";
+  import { toTokenAmountV2 } from "$lib/utils/token.utils";
 
   let showModal: "send" | undefined = undefined;
 
@@ -119,6 +125,11 @@
   $: token = nonNullish($snsOnlyProjectStore)
     ? $tokensStore[$snsOnlyProjectStore.toText()]?.token
     : undefined;
+
+  let transactionFee: TokenAmountV2 | undefined = undefined;
+  $: transactionFee = nonNullish($snsSelectedTransactionFeeStore)
+    ? toTokenAmountV2($snsSelectedTransactionFeeStore)
+    : undefined;
 </script>
 
 <Island>
@@ -184,7 +195,7 @@
     rootCanisterId={$snsOnlyProjectStore}
     loadTransactions
     {token}
-    transactionFee={$snsSelectedTransactionFeeStore}
+    {transactionFee}
   />
 {/if}
 
--- a/frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
+++ b/frontend/src/lib/components/accounts/SnsAccountsFooter.svelte
@@ -5,7 +5,7 @@
   import SnsTransactionModal from "$lib/modals/accounts/SnsTransactionModal.svelte";
   import ReceiveButton from "$lib/components/accounts/ReceiveButton.svelte";
   import { syncSnsAccounts } from "$lib/services/sns-accounts.services";
-  import { isNullish, nonNullish } from "@dfinity/utils";
+  import { TokenAmountV2, isNullish, nonNullish } from "@dfinity/utils";
   import { snsOnlyProjectStore } from "$lib/derived/sns/sns-selected-project.derived";
   import { toastsError } from "$lib/stores/toasts.store";
   import {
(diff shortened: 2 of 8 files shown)


lmuntaner:
* main
  * **PR #3972** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3972?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3972?utm_source=stack-comment).

frontend/src/lib/pages/SnsWallet.svelte
```diff
@@ -119,6 +125,11 @@
   $: token = nonNullish($snsOnlyProjectStore)
     ? $tokensStore[$snsOnlyProjectStore.toText()]?.token
     : undefined;
+
+  let transactionFee: TokenAmountV2 | undefined = undefined;
+  $: transactionFee = nonNullish($snsSelectedTransactionFeeStore)
+    ? toTokenAmountV2($snsSelectedTransactionFeeStore)
+    : undefined;
```
dskloetd, line 132: This seems tedious. It would be nicer if you could just pass `undefined` and it would return `undefined`.
The problem is then you can't use it anymore in places where you need the value not to be `undefined`.
So I asked GPT-4 about this:
````
In TypeScript I have 2 classes: TokenAmount and TokenAmountV2.
I also have a function toTokenAmountV2, which takes a parameter of type `TokenAmount | TokenAmountV2` and return `TokenAmountV2`.
But now I want it to also be possible to pass in undefined and then it should return undefined.
But I want the type system to know that if I pass in a value that's not undefined, then the return value is also not undefined.
Is this possible in TypeScript and if so how?

Answer:

Yes, it's possible using function overloads in TypeScript. Here's how you can define the `toTokenAmountV2` function with the desired behavior:

```typescript
class TokenAmount {}
class TokenAmountV2 {}

// Function overloads
function toTokenAmountV2(amount: TokenAmount): TokenAmountV2;
function toTokenAmountV2(amount: TokenAmountV2): TokenAmountV2;
function toTokenAmountV2(amount: undefined): undefined;

// Function implementation
function toTokenAmountV2(amount: TokenAmount | TokenAmountV2 | undefined): TokenAmountV2 | undefined {
  if (amount === undefined) {
    return undefined;
  }
  // Your conversion logic here
  return new TokenAmountV2(); // Assuming you have a way to convert TokenAmount to TokenAmountV2
}

// Usage
const result1 = toTokenAmountV2(new TokenAmount()); // result1 is of type TokenAmountV2
const result2 = toTokenAmountV2(new TokenAmountV2()); // result2 is of type TokenAmountV2
const result3 = toTokenAmountV2(undefined); // result3 is of type undefined
```

With these overloads, TypeScript knows that if you pass in a `TokenAmount` or `TokenAmountV2`, you'll get a `TokenAmountV2` back, and if you pass in `undefined`, you'll get `undefined` back.
````

And these function overloads seem to be a real thing:
https://www.tutorialsteacher.com/typescript/function-overloading

What do you think? Shall we try this?

**Comment by lmuntaner** [comment]

File: frontend/src/lib/utils/token.utils.ts [hunk 2]
```diff
@@ -251,15 +251,29 @@ export const ulpsToNumber = ({
   return Number(e8s) / 100_000_000;
 };
 
-export const toTokenAmountV2 = (tokenAmount: TokenAmount | TokenAmountV2) => {
+export function toTokenAmountV2(amount: TokenAmount): TokenAmountV2;
+export function toTokenAmountV2(amount: TokenAmountV2): TokenAmountV2;
+export function toTokenAmountV2(amount: undefined): undefined;
+export function toTokenAmountV2(
+  amount: TokenAmount | TokenAmountV2
+): TokenAmountV2;
```
dskloetd, line 259: Isn't this one covered by the ones on lines 254 and 255?

frontend/src/lib/utils/token.utils.ts
```diff
@@ -251,15 +251,29 @@ export const ulpsToNumber = ({
...
+  }
   if (tokenAmount instanceof TokenAmountV2) {
     return tokenAmount;
   }
   return TokenAmountV2.fromUlps({
     amount: tokenAmount.toE8s(),
     token: tokenAmount.token,
   });
-};
+}
```
dskloetd, line 276: If you prefer to stick with arrow functions, it might also be possible (I don't really care, though it would be more consistent).
https://stackoverflow.com/questions/39187614/typescript-overload-arrow-functions
```
type ToTokenAmountV2 = {
    (amount: TokenAmount): TokenAmountV2;
    (amount: TokenAmountV2): TokenAmountV2;
    (amount: undefined): undefined;
}

const toTokenAmountV2: ToTokenAmountV2 = (param: tokenAmount: TokenAmount | TokenAmountV2 | undefined): TokenAmountV2 | undefined => {
  ...
}
```

dskloetd (approved):

frontend/src/lib/utils/token.utils.ts
```diff
@@ -251,15 +251,29 @@ export const ulpsToNumber = ({
...
-export const toTokenAmountV2 = (tokenAmount: TokenAmount | TokenAmountV2) => {
+export function toTokenAmountV2(amount: TokenAmount): TokenAmountV2;
+export function toTokenAmountV2(amount: TokenAmountV2): TokenAmountV2;
+export function toTokenAmountV2(amount: undefined): undefined;
+export function toTokenAmountV2(
+  amount: TokenAmount | TokenAmountV2
+): TokenAmountV2;
+export function toTokenAmountV2(
+  amount: TokenAmount | TokenAmountV2 | undefined
+): TokenAmountV2 | undefined;
```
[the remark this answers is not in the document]
lmuntaner: It's either this or `amount: TokenAmount | undefined` and `amount: TokenAmountV2 | undefined`.

Some cases we pass type `TokenAmount | undefined`, such as when using `snsSelectedTransactionFeeStore`. Those need to be specifically typed in the overloads.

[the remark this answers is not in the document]
dskloetd: I was hoping that it would know to merge the output types when we merge the input types. But if it doesn't, this is fine.

frontend/src/lib/utils/token.utils.ts
lmuntaner: No, it's not apparently.

For example in TransactionForm we do:

```
toTokenAmountV2(transactionFee).toUlps()
```

And transactionFee is of type `TokenAmount | TokenAmountV2`.

frontend/src/lib/utils/token.utils.ts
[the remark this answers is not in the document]
lmuntaner: No, they don't merge the types, you have to specify all the combinations you want to handle.

frontend/src/lib/utils/token.utils.ts
lmuntaner: I couldn't make it work with `type`.

Maybe worth a try in another example. I like consistency, but overload functions is more important.

PR #3973, opened 2023-12-06
Rm paginated store

bitdivine:
# Motivation
The unbounded stable structures are working well enough that I think we can discard the earlier paginated implementation, rather than keeping them both open as options.  If we really want to resurrect the paginated store, we can look in git history.

# Changes
- Delete the paginated account store.

# Tests
See CI.

Note: The store was behind a test flag, so this should not affect anything in production.

# Todos

  - Too minor.  Let's announce when the feature is ready!


frederikrothenberger (approved):
LGTM, nice line count 😄

PR #3974, opened 2023-12-06
ckBTC: Merge Approve and Burn transactions into one

dskloetd:
# Motivation

When withdrawing ckBTC to BTC with ICRC-2 we first make an Approve transaction and then the minter makes a Burn transaction. Since this is part of a single process, we want to render it as a single transaction in the wallet.

# Changes

1. Pass both the previous `IcrcTransactionWithId` and previously returns `UiTransaction` to `mapCkbtcTransaction`. Note that transactions are mapped in reverse chronological order, so the previous transaction happened later in time.
2. In `mapCkbtcTransaction`, if the current transaction is an Approve transaction, check if it was followed by a Burn transaction, and if so, add the fee to the amount burned and remove the Approve transaction from the list.

# Tests

1. Added a test for `mapCkbtcTransaction` with a transaction that should be merged and with one that shouldn't be merged.
2. Added a test for `CkBTCTransactionsList`.
3. Added a convenience method to create Approve transactions.

Before:
image

After:
image

# Todos

### DIFF
--- a/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
@@ -10,7 +10,7 @@
   import { ckbtcPendingUtxosStore } from "$lib/stores/ckbtc-pending-utxos.store";
   import type { Account } from "$lib/types/account";
   import {
-    mapCkbtcTransaction,
+    mapCkbtcTransactions,
     mapCkbtcPendingUtxo,
   } from "$lib/utils/icrc-transactions.utils";
   import type {
@@ -22,8 +22,7 @@
   import type { IcrcTokenMetadata } from "$lib/types/icrc";
   import IcrcWalletTransactionsList from "$lib/components/accounts/IcrcWalletTransactionsList.svelte";
   import type { PendingUtxo } from "@dfinity/ckbtc";
-  import type { IcrcTransactionWithId } from "@dfinity/ledger-icrc";
-  import { nonNullish, isNullish, TokenAmountV2 } from "@dfinity/utils";
+  import { isNullish } from "@dfinity/utils";
 
   export let indexCanisterId: CanisterId;
   export let universeId: UniverseCanisterId;
@@ -76,41 +75,12 @@
   const mapTransactions = (
     transactionData: IcrcTransactionData[]
   ): UiTransaction[] => {
-    let prevTransaction: IcrcTransactionWithId | undefined = undefined;
-    let prevUiTransaction: UiTransaction | undefined = undefined;
-    // Note: Transaction are expected to be mapped in reverse chronological
-    // order. Some transactions might merge themselves into previous
-    // transactions, because they are conceptually part of the same event, so
-    // it's important that they appear in the expected order.
-    const completedTransactions = transactionData
-      .map(({ transaction, toSelfTransaction }: IcrcTransactionData) => {
-        if (
-          transaction.transaction.approve.length === 1 &&
-          transaction.transaction.approve[0].fee.length === 1 &&
-          prevTransaction?.transaction.burn.length === 1 &&
-          prevUiTransaction?.tokenAmount instanceof TokenAmountV2
-        ) {
-          prevUiTransaction.tokenAmount = TokenAmountV2.fromUlps({
-            amount:
-              prevUiTransaction.tokenAmount.toUlps() +
-              transaction.transaction.approve[0].fee[0],
-            token: prevUiTransaction.tokenAmount.token,
-          });
-          prevTransaction = transaction;
-          return undefined;
-        }
-        const uiTransaction = mapCkbtcTransaction({
-          transaction,
-          toSelfTransaction,
-          account,
-          token,
-          i18n: $i18n,
-        });
-        prevTransaction = transaction;
-        prevUiTransaction = uiTransaction;
-        return uiTransaction;
-      })
-      .filter(nonNullish);
+    const completedTransactions = mapCkbtcTransactions({
+      transactionData,
+      account,
+      token,
+      i18n: $i18n,
+    });
     return [...pendingTransactions, ...completedTransactions];
   };
 </script>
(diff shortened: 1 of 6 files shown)


frontend/src/lib/utils/icrc-transactions.utils.ts
```diff
@@ -263,6 +269,21 @@ export const mapCkbtcTransaction = (params: {
       console.error("Failed to decode ckBTC burn memo", memo, err);
       mappedTransaction.otherParty = i18n.ckbtc.btc_network;
     }
+  } else if (transaction.approve.length === 1) {
+    const { prevTransaction, prevUiTransaction } = params;
+    if (
+      transaction.approve[0].fee.length === 1 &&
+      prevTransaction?.transaction.burn.length === 1 &&
+      prevUiTransaction?.tokenAmount instanceof TokenAmountV2
```
lmuntaner, line 277: Can it happen that it's something different?

frontend/src/lib/utils/icrc-transactions.utils.ts
```diff
@@ -263,6 +269,21 @@ export const mapCkbtcTransaction = (params: {
...
+    ) {
+      prevUiTransaction.tokenAmount = TokenAmountV2.fromUlps({
```
lmuntaner, line 279: So, `mapCkbtcTransaction` mutates a parameter and returns nothing.

Let's maybe rename the function?

frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
```diff
@@ -75,15 +76,23 @@
   const mapTransactions = (
     transactionData: IcrcTransactionData[]
   ): UiTransaction[] => {
+    let prevTransaction: IcrcTransactionWithId | undefined = undefined;
+    let prevUiTransaction: UiTransaction | undefined = undefined;
     const completedTransactions = transactionData
-      .map((transaction: IcrcTransactionData) =>
-        mapCkbtcTransaction({
-          ...transaction,
+      .map(({ transaction, toSelfTransaction }: IcrcTransactionData) => {
```
lmuntaner, line 82: What if we passed the nextTransaction instead?

```
.map(({ ... }), index, elements) => {
  const next = elements[index + 1];
}
```

lmuntaner (commented):
A comment about a different approach maybe.

frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
dskloetd: We pass both the previous `IcrcTransactionWithId` and the corresponding previous `UiTransaction`.
We can pass the next `IcrcTransactionWithId` but the next `UiTransaction` doesn't exist yet.

What would be the benefit of passing the next transaction instead of the previous transaction?

frontend/src/lib/utils/icrc-transactions.utils.ts
dskloetd: If there is no `prevUiTransaction` (because this is the first transaction, or the previous transaction was excluded), then `prevUiTransaction?.tokenAmount` will be `undefined`.
Also, the type of `UiTransaction` allows for `prevUiTransaction?.tokenAmount` to be `TokenAmount` (v1) but in the context of ckBTC it will never be `TokenAmount` (v1), but TypeScript doesn't know that. But we need TS to know it to allow `prevUiTransaction.tokenAmount.toUlps()` below.

frontend/src/lib/utils/icrc-transactions.utils.ts
dskloetd: It almost always returns a `UiTransaction`. But even if it returns `undefined`, this is explicitly to signal that this transaction does not result in a corresponding transaction in the resulting `UiTransaction` array. So I wouldn't call that "nothing".

It's already possible for `mapCkbtcTransaction` to return `undefined` on line 254 when `mapIcrcTransaction` returns `undefined`, which happens when `token` is `undefined` or when the mapping throws an error.

For this reason we already filter out nullish transactions from the resulting array, here: https://github.com/dfinity/nns-dapp/blob/fca69c8daa55891c5c8aeba715130788c7072090/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte#L87

There are many possible ways to go about merging some transactions. But I haven't found one that isn't at least somewhat tricky. Other options I've considered are:
1. First map all the transactions 1:1 and then do a separate pass to merge some transaction. The problem is that then you only have `UiTransaction`s and you need to recognize burns transaction by the different headline (or whatever) instead of the type of the transaction.
2. Make `mapCkbtcTransaction` potentially map multiple transactions in a single invocation. This would require returning a (potentially empty) array and passing in the full array of transactions, and index of which transaction to process next, so that the function can choose how many of them to consume. It would also need to return how many it consumed so the outer mapping function knows how much to advance the index. I actually had the beginning of an implementation of this but it seems too complicated to understand.
3. Inline `mapCkbtcTransaction` into `CkBTCTransactionsList` to just have one big mapping loop which has full access to everything. This is a reasonable option which I'm happy to try if you want.

Or do you just want me to rename the function to `mapOrMergeCkbtcTransaction`?

frontend/src/lib/utils/icrc-transactions.utils.ts
lmuntaner: Thanks for the explanation!

frontend/src/lib/utils/icrc-transactions.utils.ts
lmuntaner: Replied in the other comment.

I like option 3, but what I like the least is mutating a parameter.

dskloetd: I'll try option 3 in a separate PR so you can compare.

dskloetd: Actually, I made a small change in this PR moving just the part that mutates the previous transaction out of `mapCkbtcTransaction` and into `CkBTCTransactionsList`.
If you like it like this, I propose moving the entire loop out of the component and back into the util as `mapCkbtcTransactions` (plural) so that it can be tested separately.
Let me know what you think.

lmuntaner: I like it better, thanks!

I think it's a good idea to move the entire loop out of the component for better testing.

lmuntaner (approved):

PR #3981, opened 2023-12-07
GIX-2110: Use AnchorTag for row if needed

lmuntaner:
# Motivation

The whole row on the Tokens page is really a link to another page.

In this PR, introduce a new optional field in UserTokenData: `"rowHref"`. Optional, because in the icp tokens page the rows are not clickable.

# Changes

* Add optional field `rowHref` in UserTokenData type.
* TokensTableRow uses the new field to render a link with href or not.

# Tests

* Test that Tokens table renders row with href when present.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
+++ b/frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
@@ -49,7 +49,9 @@
   };
 </script>
 
-<div
+<svelte:element
+  this={nonNullish(userTokenData.rowHref) ? "a" : "div"}
+  href={userTokenData.rowHref}
   role="row"
   tabindex={index + 1}
   on:keypress={handleClick}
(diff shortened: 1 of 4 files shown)


dskloetd:
Why are the rows not clickable in the ICP tokens page?

lmuntaner:
> Why are the rows not clickable in the ICP tokens page?

Ah, no, they are actually. They will redirect to the wallet page. I can do it mandatory after populating the field, this way the PRs can be smaller though.

dskloetd:
I don't understand. In the PR description it says:

> Optional, because in the icp tokens page the rows are not clickable.

Do you mean that it's only optional temporarily to split the PR?
After another PR it will no longer be optional?

dskloetd:
Maybe there is a typo in this sentence because I'm not sure how to parse this:

>  I can do it mandatory after populating the field

lmuntaner:
> Do you mean that it's only optional temporarily to split the PR?
> After another PR it will no longer be optional?

Initially I thought I would leave it as optional. But it's true that the rows in the ICP tables also redirect somewhere. So, yes, at this moment, this is temporarily optional.

dskloetd (approved):

PR #3985, opened 2023-12-07
GIX-2171: Use Tokens in "/"

lmuntaner:
# Motivation

When Tokens is enabled, the home page should be the tokens page.

The current home page is the one in `routes/(app)/(u)/(accounts)/+page.svelte`.

In this PR, I change the logic of this component to render the Tokens route instead of the current Accounts route.

In another PR I'll move the layout and this home page to `routes/(app)/(home)`.

# Changes

* Change logic in `(app)/(u)/(accounts)/+page.svelte` to render either the Accounts route or the Tokens route.

# Tests

* New test for this page to check which component is rendered depending on the flag.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/routes/(app)/(u)/(accounts)/+page.svelte
+++ b/frontend/src/routes/(app)/(u)/(accounts)/+page.svelte
@@ -1,15 +1,12 @@
 <script lang="ts">
-  import SignInAccounts from "$lib/pages/SignInAccounts.svelte";
-  import Accounts from "$lib/routes/Accounts.svelte";
-  import { authSignedInStore } from "$lib/derived/auth.derived";
-  import TestIdWrapper from "$lib/components/common/TestIdWrapper.svelte";
-  import { icpTokensListVisitors } from "$lib/derived/icp-tokens-list-visitors.derived";
+  import { ENABLE_MY_TOKENS } from "$lib/stores/feature-flags.store";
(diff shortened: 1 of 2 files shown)


frontend/src/routes/(app)/(u)/(accounts)/+page.svelte
```diff
@@ -1,15 +1,12 @@
...
-  import Accounts from "$lib/routes/Accounts.svelte";
-  import { authSignedInStore } from "$lib/derived/auth.derived";
-  import TestIdWrapper from "$lib/components/common/TestIdWrapper.svelte";
-  import { icpTokensListVisitors } from "$lib/derived/icp-tokens-list-visitors.derived";
+  import { ENABLE_MY_TOKENS } from "$lib/stores/feature-flags.store";
+  import TokensRoute from "../../(nns)/tokens/+page.svelte";
+  import AccountsRoute from "./accounts/+page.svelte";
 </script>
 
-<TestIdWrapper testId="accounts-plus-page-component">
```
lmuntaner, line 9: This was the same that was in `./accounts/+page.svelte`. Which I render as one component now to not duplicate logic.

I didn't know you could import a `+page.svelte`, but apparently you can.

lmuntaner:
UX:

In "/":



In "/accounts":

lmuntaner:
@dskloetd please review

PR #3988, opened 2023-12-08
fix: CI

bitdivine:
# Motivation
CI is broken

## Version update test failing
The node version is not specified in one of the tests, so fails now that GitHub is running a node version that uses an NPM version that is out the supported range:

```
npm ERR! notsup Required: {"npm":">=9.0.0 <10.0.0","node":">=18.0.0 <19.0.0"}
npm ERR! notsup Actual:   {"npm":"10.2.3","node":"v18.19.0"}
```

## Spellcheck build failing
It looks as if the spellcheck build should work.  Perhaps some dependencies are being updated though as `--locked` is not passed to `cargo install`.

# Changes
- Install the specified node version for the version update test.
- Lock the spellcheck version.
  - Note: There have been no new versions recently but better safe than sorry.
 - Require dependencies to be used as specified in the spellcheck Cargo.lock file.

# Tests
- See CI

# Todos

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -42,14 +42,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Install cargo binstall
-        uses: ./.github/actions/install_binstall
       - name: Install spellcheck
         run: |
           cd /
           sudo apt-get update
           sudo apt-get install -yy hunspell hunspell-en-gb hunspell-en-us clang libclang-dev
-          cargo binstall --pkg-url="{ repo }/releases/download/v{ version }/{ name }-v{ version }-{ target }" --strategies crate-meta-data --version 0.13.1 cargo-spellcheck -y --force --pkg-fmt bin
+          cargo install --locked --target x86_64-unknown-linux-gnu cargo-spellcheck@0.13.1
       - name: Spellcheck Rust
         run: cargo spellcheck
       - name: Spellcheck changelog


.github/workflows/checks.yml
```diff
@@ -47,7 +47,7 @@ jobs:
           cd /
           sudo apt-get update
           sudo apt-get install -yy hunspell hunspell-en-gb hunspell-en-us clang libclang-dev
-          cargo install --target x86_64-unknown-linux-gnu cargo-spellcheck
+          cargo install --locked --target x86_64-unknown-linux-gnu cargo-spellcheck@0.13.1
```
bitdivine, line 50: Note: The build is actually quite slow in CI.  Locally it is super fast.  It might make sense to get a binary installation working.  Can get binstall configured upstream or install manually with:
```
curl -sSLf https://github.com/drahnr/cargo-spellcheck/releases/download/v0.13.1/cargo-spellcheck-v0.13.1-x86_64-unknown-linux-gnu | install -m 755 /dev/stdin ~/.cargo/bin/cargo-spellcheck
```
(Not tested)

bitdivine: Tested locally:
```
cargo binstall --pkg-url="{ repo }/releases/download/v{ version }/{ name }-v{ version }-{ target }" --strategies crate-meta-data --version 0.13.1 cargo-spellcheck -y --force --pkg-fmt bin
```

bitdivine: Note: Bumping the Ubuntu version to 22.04 would fix the glibc issue.  But now is probably not the time...

mstrasinskis (approved):
Thank you!

PR #3989, opened 2023-12-08
fix: Cargo install locked

bitdivine:
# Motivation
The spellcheck build is broken.

# Changes
- Lock the spellcheck version.
  - Note: There have been no new versions recently but better safe than sorry.
 - Require dependencies to be used as specified in the spellcheck Cargo.lock file.

# Tests
See CI

# Todos


bitdivine:
In another PR

PR #3995, opened 2023-12-11
Renamed Account.balanceE8s to balanceUlps

dskloetd:
# Motivation

The nns-dapp `Account` type has a field called `balanceE8s` but this type is also used for ckETH which has 18 decimals.
So the name `balanceE8s` is misleading in that case.

# Changes

1. Renamed `balanceE8s` to `balanceUlps`.
2. Added a few TODOs of related things that should be renamed.

# Tests

pass

# Todos

not necessary


lmuntaner (commented):
Looking good. I'll wait for final review

lmuntaner (approved):
Thanks!

PR #3999, opened 2023-12-11
GIX-2110: Populate rowHref Tokens page

lmuntaner:
# Motivation

Users can navigate to Wallet or ICP Accounts page from the Tokens page.

# Changes

* Populate the rowHref field in `tokensListUserStore`.

# Tests

* Add new field in `tokensListUserStore` test files. Rename some local variables to `xxxNoBalance` to better represent what they are.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/lib/derived/tokens-list-user.derived.ts
+++ b/frontend/src/lib/derived/tokens-list-user.derived.ts
@@ -1,11 +1,19 @@
 import { OWN_CANISTER_ID_TEXT } from "$lib/constants/canister-ids.constants";
 import {
+  authStore,
+  type AuthStore,
+  type AuthStoreData,
+} from "$lib/stores/auth.store";
+import {
   tokensStore,
   type TokensStore,
   type TokensStoreData,
 } from "$lib/stores/tokens.store";
 import { UserTokenAction, type UserTokenData } from "$lib/types/tokens-page";
+import { buildAccountsUrl, buildWalletUrl } from "$lib/utils/navigation.utils";
 import { UnavailableTokenAmount } from "$lib/utils/token.utils";
+import { isUniverseNns } from "$lib/utils/universe.utils";
+import { encodeIcrcAccount } from "@dfinity/ledger-icrc";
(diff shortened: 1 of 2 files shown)


lmuntaner:
* main
  * **PR #3999** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/3999?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/3999?utm_source=stack-comment).

frontend/src/lib/derived/tokens-list-user.derived.ts
```diff
@@ -47,15 +55,39 @@ const addActions = (userTokenData: UserTokenData): UserTokenData => ({
...
 
+const addHref = ({
+  userTokenData,
+  authData,
+}: {
+  userTokenData: UserTokenData;
+  authData: AuthStoreData;
+}): UserTokenData => ({
+  ...userTokenData,
+  rowHref: isNullish(authData.identity)
```
lmuntaner, line 66: This derived store is not called when there is no identity, but still, we need to check this.

lmuntaner:
@dskloetd please review

frontend/src/lib/derived/tokens-list-user.derived.ts
dskloetd: Since the plan is for `rowHref` to be optional only temporarily, how do you plan to satisfy the type system here when it becomes mandatory?
Maybe it makes sense to do that already?

lmuntaner: As I mentioned [here](https://github.com/dfinity/nns-dapp/pull/4000#issuecomment-1849623996), I'm not sure yet it should be mandatory.

Here I would throw an error or force a logout probably. It's an edge case that shouldn't happen.

PR #4000, opened 2023-12-11
GIX-2110: Populate rowHref icp accounts

lmuntaner:
# Motivation

Users can navigate to the wallet page from the ICP accounts page with the tokens table.

# Changes

* Populate `rowHref` in `icpTokensListUser` derived store.

# Tests

* Add new field to the `icpTokensListUser` test data.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/lib/derived/icp-tokens-list-user.derived.ts
+++ b/frontend/src/lib/derived/icp-tokens-list-user.derived.ts
@@ -7,6 +7,7 @@ import {
 import type { Account, AccountType } from "$lib/types/account";
 import { UserTokenAction, type UserTokenData } from "$lib/types/tokens-page";
 import type { Universe } from "$lib/types/universe";
+import { buildWalletUrl } from "$lib/utils/navigation.utils";
 import { UnavailableTokenAmount } from "$lib/utils/token.utils";
 import { Principal } from "@dfinity/principal";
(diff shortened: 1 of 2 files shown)


lmuntaner:
* main
  * **PR #4000** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4000?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4000?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

I realized that I prefer to keep the `rowHref` as optional because I don't want it populated while the balance is loading, no?

dskloetd:
> I realized that I prefer to keep the `rowHref` as optional because I don't want it populated while the balance is loading, no?

Why not? Does the link depend on the balance?

lmuntaner:
> > I realized that I prefer to keep the `rowHref` as optional because I don't want it populated while the balance is loading, no?
> 
> Why not? Does the link depend on the balance?

No, it doesn't. But I prefer the UX to be fully loaded before we offer the functionality. We do the same to send and receive. Although there makes more sense to wait for the balance.

I was just thinking out loud, though. It doesn't affect this PR.

dskloetd (approved):

PR #4003, opened 2023-12-11
chore: Keep dfx server logs

bitdivine:
# Motivation
When debugging failures in CI, it is useful to be able to see the logs of the running dfx server.

# Changes
- Save the `dfx start` logs to a file and upload them, so that they can be examined.
- Where a workflow runs dfx start several times, use different names for different runs.
  - Note: It _may_ make sense to make the default name include the job name, however in practice I found some issues with this approach.  I think setting a name manually is fine for now.  For a cleverer default, I would rather wait until there is a solution that really works well.

# Tests
- See CI for sample outputs.
- Usability:  I used this while developing the stable structures migration.

# Todos


lmuntaner (approved):
LGTM, thanks!

PR #4005, opened 2023-12-11
Placeholder for migration state

bitdivine:
# Motivation
When migrating from one database to another, we need to store the migration state.

# Changes
- Replace the secondary db with a container that holds both the second db and the migration state.

# Tests
None - there are no functional changes.

# Todos


frederikrothenberger (approved):
LGTM

PR #4010, opened 2023-12-12
Fix: TokensTable actions triggering link

lmuntaner:
# Motivation

There was a navigation and full refresh when the user clicked directly in one of the action buttons.

We need to prevent the default behavior of the link as well as stop propagation.

# Changes

* Add `preventDefault` directive to ReceiveButton and SendButton.

The GoToDetailButton will be refactoring into a single span with no clicking functionality because we should rely on the rowHref instead.

# Tests

* Test that when clicking Send or Receive, the default is prevented.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/components/tokens/TokensTable/actions/ReceiveButton.svelte
+++ b/frontend/src/lib/components/tokens/TokensTable/actions/ReceiveButton.svelte
@@ -12,7 +12,7 @@
 <button
   class="icon"
   data-tid="receive-button-component"
-  on:click|stopPropagation={() => {
+  on:click|preventDefault|stopPropagation={() => {
     dispatcher("nnsAction", { type: ActionType.Receive, data: userToken });
   }}
 >
--- a/frontend/src/lib/components/tokens/TokensTable/actions/SendButton.svelte
+++ b/frontend/src/lib/components/tokens/TokensTable/actions/SendButton.svelte
@@ -12,8 +12,7 @@
 <button
   class="icon"
   data-tid="send-button-component"
-  on:click|preventDefault|stopPropagation={() => {
-    console.log("userToken", userToken);
+  on:click|stopPropagation|preventDefault={() => {
     dispatcher("nnsAction", { type: ActionType.Send, data: userToken });
   }}
 >
--- a/frontend/src/lib/components/tokens/TokensTable/actions/SendButton.svelte
+++ b/frontend/src/lib/components/tokens/TokensTable/actions/SendButton.svelte
@@ -12,7 +12,8 @@
 <button
   class="icon"
   data-tid="send-button-component"
-  on:click|stopPropagation={() => {
+  on:click|preventDefault|stopPropagation={() => {
+    console.log("userToken", userToken);
     dispatcher("nnsAction", { type: ActionType.Send, data: userToken });
   }}
 >
--- a/frontend/src/tests/lib/components/tokens/TokensTable.spec.ts
+++ b/frontend/src/tests/lib/components/tokens/TokensTable.spec.ts
@@ -1,5 +1,6 @@
 import TokensTable from "$lib/components/tokens/TokensTable/TokensTable.svelte";
 import { OWN_CANISTER_ID } from "$lib/constants/canister-ids.constants";
+import { AppPath } from "$lib/constants/routes.constants";
 import { ActionType } from "$lib/types/actions";
 import {
   UserTokenAction,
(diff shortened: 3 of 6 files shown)


lmuntaner:
* main
  * **PR #4010** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4010?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4010?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

frontend/src/lib/components/tokens/TokensTable/actions/ReceiveButton.svelte
dskloetd, line 1: Can you add a test?

lmuntaner: I want to, but I need to research quite a bit how to trigger that kind of click.

I didn't find anything so easily, let me spend some more time.

Otherwise, I'll create a ticket to add a test and research that further.

This prevents us from using the functionality and I want it to be testeable after the next release behind a feature flag. So I'd like this to be merged before tomorrow.

frontend/src/lib/components/tokens/TokensTable/actions/SendButton.svelte
```diff
@@ -12,7 +12,8 @@
 <button
   class="icon"
   data-tid="send-button-component"
-  on:click|stopPropagation={() => {
+  on:click|preventDefault|stopPropagation={() => {
+    console.log("userToken", userToken);
```
dskloetd, line 16: Should this be here?

frontend/src/lib/components/tokens/TokensTable/actions/ReceiveButton.svelte
dskloetd: What do you mean by "that kind of click"? What is different here than any other test?

lmuntaner: The click event with `fireEvent` doesn't trigger the parent anchor tag. Otherwise, we would have caught this in our current tests.

I had to do a workaround and test explicitly that `preventDefault` was called instead with the `event.defaultPrevented`

frontend/src/lib/components/tokens/TokensTable/actions/SendButton.svelte
lmuntaner: Nops, removed, thanks.

lmuntaner:

dskloetd (approved):

frontend/src/lib/components/tokens/TokensTable/actions/ReceiveButton.svelte
lmuntaner: I agree, it's quite unfortunate. That's the only way I was able to find.

I wonder whether e2e click is similar to firing an event of the click actually triggers also the parent click.

We do have some e2e test to send, so when I migrate them I will check without the preventDefault.

frontend/src/tests/lib/components/tokens/TokensTable.spec.ts
```diff
@@ -272,6 +273,60 @@ describe("TokensTable", () => {
...
+      onAction: vi.fn(),
+    });
+
+    const rows = await po.getRows();
+    const rowPo = rows[0];
+
+    const sendButton = rowPo.getSendButton();
+
+    let isClicked = false;
+    sendButton.addEventListener("click", (event) => {
```
dskloetd, line 293: I don't recall seeing all these changes.
Was this added after I approved?
I don't like adding `addEventListener` on the page object element.
Why can't you just test for the consequences of clicking the button?

dskloetd: Looks like I approved after you added the test.
So apologies for implying anything.
I must have not looked closely.
But my opinion and question still stand.

lmuntaner: Because what we are testing here is that a click on the button doesn't trigger the click on the parent anchor link.

Triggering a click event in the button doesn't trigger the parent click, nor can I check that the default behaviour is prevented.

We discussed that the ideal test is an e2e test.

PR #4017, opened 2023-12-13
feat: Add range to AccountsDb trait

bitdivine:
# Motivation
To migrate data from one database to another, we wish to use ranges to get the next N accounts.

# Changes
- Add `range()` to the `AccountsDb` trait.

# Tests
- None.  These are straight proxies to the `BTreeMap` so tests probably aren't warranted.

# Todos


frederikrothenberger (approved):
LGTM!

PR #4019, opened 2023-12-13
Render ckBTC reimbursement transactions

dskloetd:
# Motivation

When withdrawaing ckBTC to BTC fails, the minter will mint the previously burned amount (minus fees) back into the user's account.
We want to render these transactions as Reimbursement transactions.

# Changes

1. Add `isReimbursement` field to `UiTransaction`.
2. Decode the memo on mint transactions. If we can decode them as failed KYT, mark the transaction as a reimbursed transaction.
3. Render `isReimbursement` transactions with `IconReimbursed` and alert colors.

# Tests

# Todos


mstrasinskis (approved):
LGTM, thank you.

PR #4020, opened 2023-12-13
fix: Truncate error

bitdivine:
# Motivation
We would like to make proposal payload rendering more robust.

- If schema types do not match, fall back to idl2json.
- If there is still an error, it should not be so large that it fails to return.

This PR addresses the latter.  As an example we have:

```
./scripts/proposals/mk-test-vector --network mainnet --proposal 126392
```
which yields a huge 3Mb error message:
```
$ wc -c /home/max/dfn/nns-dapp/rs/proposals/src/tests/proposals/mainnet/proposals/126392.payload.json
3279306
```

# Changes
Limit the size of the error message to 400 chars.

# Tests
Running the above comamnd on the same proposal now yields a short error:
```
$ cat /home/max/dfn/nns-dapp/rs/proposals/src/tests/proposals/mainnet/proposals/126392.payload.json
"Unable to deserialize payload: Custom(Fail to decode argument 0 from table0 to record {\n  arg : vec nat8;\n  wasm_module : vec nat8;\n  stop_before_installing : bool;\n  mode : variant { reinstall; upgrade; install };\n  canister_id : " <-- Truncated
```

# Todos


lmuntaner (approved):
Thanks! This is still good to have.

PR #4021, opened 2023-12-13
Fix: Proposal payload rendering

lmuntaner:
# Motivation

A proposal failed to render the payload: https://nns.ic0.app/proposal/?u=qoctq-giaaa-aaaaa-aaaea-cai&proposal=126392

The problem was a [breaking change in the IC for the payload types](https://github.com/dfinity/ic/blob/master/rs/nervous_system/root/src/change_canister.rs#L137). One field was removed from two types and also renamed.

Yet, updating the IC depedency is becoming a huge pain. Therefore, to solve the issue in the short term, I hardcoded the types in the PR instead of importing them.

# Changes

* Do not import AddNnsCanisterProposal but hardcode it.
* Do not import ChangeNnsCanisterProposal but hardcode it.
* Remove the field `authz_changes` from both `XxxTrimmed` type and transform functions.

# Tests

* I downloaded the proposal. It was failing withuot this changes, but with this changes it was working.

I'm trying to make such a proposal in a local environment and test it there.

# Todos

- [ ] Add entry to changelog (if necessary).

PR #4022, opened 2023-12-13
Proposal rendering fallback

bitdivine:
# Motivation
Proposal schemas change, and not always in a forwards and backwards compatible fashion.  If that happens, the proposal rendering needs to not break.

# Changes
- If proposals fail to parse strictly with the expected types, use `idl2json` for a best-effort conversion.

# Tests
This proposal previously failed to parse:
```
$ ./scripts/proposals/mk-test-vector --network mainnet --proposal 126392
```
It now parses as:
```
[{"arg":"4449444c00017f","canister_id":"rdmx6-jaaaa-aaaaa-aaadq-cai","compute_allocation":[],"memory_allocation":[],"mode":{"upgrade":null},"query_allocation":[],"stop_before_installing":true,"wasm_module":"Bytes with sha256: 1950912f23acea10a63ea1c0a70ac316f2cea6055acde3044e24cb79682c6d40"}]
```

# Todos

### DIFF
--- a/rs/proposals/src/lib.rs
+++ b/rs/proposals/src/lib.rs
@@ -142,7 +142,7 @@ const IDL2JSON_OPTIONS: Idl2JsonOptions = Idl2JsonOptions {
     prog: Vec::new(), // These are the type definitions used in proposal payloads.  If we have them, it would be nice to use them.  Do we?
 };
 
-/// Rust types can include things such as functions.  IDL types, sent over a wire, cannot.  Given that we want an IDLType from a more general type we need toconvert th egeneral type to the more specialized type.
+/// Rust types can include things such as functions.  IDL types, sent over a wire, cannot.  Given that we want an IDLType from a more general type we need toconvert the general type to the more specialized type.
 fn type_2_idltype(ty: Type) -> Result<IDLType, String> {
     match ty {
         Type::Null => Ok(IDLType::PrimT(parser_types::PrimType::Null)),
--- a/rs/proposals/src/lib.rs
+++ b/rs/proposals/src/lib.rs
@@ -134,7 +134,7 @@ fn decode_arg(arg: &[u8], arg_types: IDLTypes) -> String {
 pub fn process_proposal_payload(proposal_info: ProposalInfo) -> Json {
     if let Some(Action::ExecuteNnsFunction(f)) = proposal_info.proposal.as_ref().and_then(|p| p.action.as_ref()) {
         transform_payload_to_json(f.nns_function, &f.payload)
-            .unwrap_or_else(|e| serde_json::to_string(&format!("Unable to deserialize payload: {e}")).unwrap())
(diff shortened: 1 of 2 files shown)


rs/proposals/src/lib.rs
```diff
@@ -140,8 +136,78 @@ pub fn process_proposal_payload(proposal_info: ProposalInfo) -> Json {
     }
 }
 
+const IDL2JSON_OPTIONS: Idl2JsonOptions = Idl2JsonOptions {
+    bytes_as: Some(BytesFormat::Hex),
+    long_bytes_as: Some((256, BytesFormat::Sha256)),
+    prog: Vec::new(), // These are the type definitions used in proposal payloads.  If we have them, it would be nice to use them.  Do we?
```
lmuntaner, line 142: Do you mean Rust types?

lmuntaner (commented):
This change is also worth it. But it we do it like this, we might also lose the rendering of the arguments within the payload, right?

The one that happens here in [canister_arg_types](https://github.com/dfinity/nns-dapp/blob/d60a84c3f07bd8e90fac59cd329c88e7c6635272/rs/proposals/src/lib.rs#L103)

Let's get both PRs in, also #4021

rs/proposals/src/lib.rs
bitdivine: IDL types.  This isn't all that interesting here.  A .did file may mention types that are not defined in the .did file, in which case other .did files with the missing definitions can be loaded and provided in that vector.  The comment is old - I haven't changed it - but would be happy to delete it.  It would become interesting again if and when we get proposal types dynamically rather than having them copied in.

bitdivine:
> But it we do it like this, we might also lose the rendering of the arguments within the payload, right?

Correct.  Good point.  Let me think about that.

rs/proposals/src/lib.rs
lmuntaner: No, it's ok, leave it. Maybe add "These are the IDL types"

lmuntaner:
> > But it we do it like this, we might also lose the rendering of the arguments within the payload, right?
> 
> Correct. Good point. Let me think about that.

thanks!

rs/proposals/src/lib.rs
bitdivine: Done.

bitdivine:
Regarding install arguments, I think that there are nice solutions in idl2json itself.  We could say there, for example: "If you encounter bytes, check if they are candid and, if so, render them, if not, render the bytes as usual."  Or some other more refined ways of dealing with serializations inside serializations.  However that will not get finished today.  I feel as if this PR improves the case where there is a type mismatch reasonably well though, so I would be inclined to merge it.  What do you think?

lmuntaner (approved):
Thanks!

PR #4027, opened 2023-12-13
fix: local OWN_CANISTER_ID

bitdivine:
# Motivation
The nns-dapp configuration is wrong for local deployments:  The local nns-dapp canister ID is overridden.

# Changes
- Use the term `OWN_CANISTER_ID` consistently to make this easier to debug.  Previously sometimes `CANISTER_ID` was used and that gives many false hits when searching.
- Make the canister IDs provided by dfx override those in configuration files.

# Tests
- I added a test for teh output of config.sh for local, with two different deployments of the nns-dapp.
  - Note: There is an existing test that covers mainnet and the app subnet, but these both have fixed canister IDs.  The test needed to be adapted a little to cope with a variable canister ID.
- I read through all uses of these variables I could find and am reasonably confident on that basis that these changes are correct, not just that the tests pass.  Additional vigilance will also be needed in the next release to make sure that the arguments are correct there.

# Todos


config.sh
```diff
@@ -160,11 +152,11 @@ local_deployment_data="$(
 : "- construct ledger and governance canister URLs"
 # TODO: I believe that the following can be discarded now.
 json=$(HOST=$(dfx-canister-url --network "$DFX_NETWORK" --type api) jq -s --sort-keys '
-  (.[0].defaults.network.config // {}) * .[1] * (.[2].config // {}) |
+  (.[0].defaults.network.config // {}) * (.[1].config // {}) * .[2] |
   .DFX_NETWORK = env.DFX_NETWORK |
   . as $config |
   .HOST=env.HOST
-    ' dfx.json <(echo "$local_deployment_data") <(network_config))
+    ' dfx.json <(network_config) <(echo "$local_deployment_data"))
```
bitdivine, line 159: This is the actual fix.

config.sh
```diff
@@ -48,9 +48,9 @@ local_deployment_data="$(
   : "- may be set by dfx as an env var"
   : "- may be deployed locally"
   LOCALLY_DEPLOYED_NNS_CANISTER_ID="$(dfx canister --network "$DFX_NETWORK" id nns-dapp 2>/dev/null || true)"
-  test -n "${CANISTER_ID:-}" || CANISTER_ID="$LOCALLY_DEPLOYED_NNS_CANISTER_ID"
-  export CANISTER_ID
-  test -n "${CANISTER_ID:-}" || unset CANISTER_ID
+  test -n "${OWN_CANISTER_ID:-}" || OWN_CANISTER_ID="$LOCALLY_DEPLOYED_NNS_CANISTER_ID"
+  export OWN_CANISTER_ID
+  test -n "${OWN_CANISTER_ID:-}" || unset OWN_CANISTER_ID
```
bitdivine, line 53: This `CANISTER_ID` is hard to search for in logs as it appears in other places and as a substring of other variables.  `OWN_CANISTER_ID` is unique.  We might even switch to `CANISTER_ID_NNS_DAPP` as used by `dfx`.  https://internetcomputer.org/docs/current/references/cli-reference/dfx-envars  But that would touch more files and so belongs in another PR.

dfx.json
```diff
@@ -299,7 +299,6 @@
     },
     "local": {
       "config": {
-        "FETCH_ROOT_KEY": true,
         "HOST": "http://localhost:8080",
```
bitdivine, line 302: Not needed as it matches the default and increases noise.

config.test
```diff
@@ -16,7 +16,17 @@ clean_config() {
...
+  echo "Compare generated config with local deployments"
+  # Note: Unlike on mainnet, canister IDs can change.  The test will adapt the expected values with the actual canister ID.
+  # We will deploy twice to make sure that the local canister ID changes.
+  dfx start --clean --background
+  for ((i = 0; i < 2; i++)); do
+    dfx canister create nns-dapp
+    scripts/nns-dapp/test-config --network local
+    dfx canister delete nns-dapp --yes
+  done
+)
```
bitdivine, line 29: We test local deployments twice to make sure that we test different canister IDs.

config.sh
lmuntaner: I also like the idea of `CANISTER_ID_NNS_DAPP`. It's clearer.

config.sh
```diff
@@ -120,16 +120,9 @@ local_deployment_data="$(
   STATIC_HOST="$(dfx-canister-url --network "$DFX_NETWORK" --type static)"
   export API_HOST STATIC_HOST
 
-  : "Determine whether we need to fetch the root key"
```
lmuntaner, line 123: Why did you remove this?

config.sh
lmuntaner: Thanks for pointing it out!

lmuntaner (approved):
LGTM! thanks!

config.sh
bitdivine: It gives the wrong answer for e.g. the app subnet "network".  `scripts/network-config` gives the right answer.

bitdivine: Resolving to merge but happy to discuss further if wanted.

PR #4028, opened 2023-12-14
Bump the account limit

bitdivine:
# Motivation
We are getting more users.

# Changes
* Increase the account limit.

# Tests
- Existing tests should suffice.

# Todos

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -32,7 +32,7 @@ type TransactionIndex = u64;
 
 /// The data migration is more complicated if there are too many accounts.  With below this many
 /// accounts we avoid some complications.
-const PRE_MIGRATION_LIMIT: u64 = 250_000;
+const PRE_MIGRATION_LIMIT: u64 = 300_000;
 
 /// Accounts, transactions and related data.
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Maybe bring it up to 300k?

It's holiday season and who knows, maybe there is a bigger spike and we will be a few weeks without a new release.

The account creation duplicated from 125 to 250 in one day. If we reach 1000k per day, we might need to make another release otherwise.

rs/backend/src/accounts_store.rs
```diff
@@ -32,7 +32,7 @@ type TransactionIndex = u64;
 
 /// The data migration is more complicated if there are too many accounts.  With below this many
 /// accounts we avoid some complications.
-const PRE_MIGRATION_LIMIT: u64 = 230_000;
+const PRE_MIGRATION_LIMIT: u64 = 250_000;
```
lmuntaner, line 35: Increase to 300_000? See the previous comment

lmuntaner (approved):

rs/backend/src/accounts_store.rs
bitdivine: Done.

PR #4032, opened 2023-12-14
Fix: Wallet back button tokens enabled

lmuntaner:
# Motivation

There was a bug that the user was redirected to the Tokens page from an ICP wallet.

Instead, the user should go to the ICP Accounts.

# Changes

* Change implementation of `buildAccountsUrl` to build an accounts page for NNS only.

# Tests

* Add tests for `buildAccountsUrl`.
* Add tests for `accountsPathStore` which uses `buildAccountsUrl`.
* Add a test file for the wallet layout component which is were the bug was.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/lib/utils/navigation.utils.ts
+++ b/frontend/src/lib/utils/navigation.utils.ts
@@ -1,3 +1,4 @@
+import { OWN_CANISTER_ID_TEXT } from "$lib/constants/canister-ids.constants";
 import {
   ACCOUNT_PARAM,
   AppPath,
@@ -7,6 +8,8 @@ import {
   UNIVERSE_PARAM,
 } from "$lib/constants/routes.constants";
 import type { NeuronId, ProposalId } from "@dfinity/nns";
+import { Principal } from "@dfinity/principal";
+import { isUniverseNns } from "./universe.utils";
 import { isArrayEmpty } from "./utils";
 
 // If the previous page is a particular detail page and if we have data in store, we don't reset and query the data in store after the route is mounted.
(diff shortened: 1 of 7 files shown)


lmuntaner:
* `main`
  * **PR #4032** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4032?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4032?utm_source=stack-comment).

lmuntaner:
UX:

            <a href="https://app.graphite.dev/media/video/W99H6e2WWjPgtes6VOJI/b2b6add9-da99-4ead-8e6e-e86a7d3d968e.mov">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/W99H6e2WWjPgtes6VOJI/b2b6add9-da99-4ead-8e6e-e86a7d3d968e.mov">Screen Recording 2023-12-14 at 11.17.19.mov</video>

lmuntaner:
@dskloetd please review

frontend/src/lib/utils/navigation.utils.ts
```diff
@@ -53,10 +56,19 @@ export const buildAccountsUrl = ({
 }: {
   universe: string;
   tokensEnabled?: boolean;
-}) =>
-  tokensEnabled
-    ? AppPath.Tokens
-    : buildUrl({ path: AppPath.Accounts, universe });
+}) => {
+  try {
+    // `Principal.fromText` throws if the text is not a valid principal
```
dskloetd, line 61: isUniverseNns just compares the string to `OWN_CANISTER_ID.toText()`.
You don't need to parse the string to do that.

lmuntaner: I have been thinking that I want to change the approach.

`accountsPathStore` and `buildAccountsPath` should be about Accounts.

Instead of changing the loginc inside there, I shuold change the usage and use the Tokens path when applicable.

What do you think?

For example, in the layout:

```
const back = (): Promise<void> => goto($accountsPathStore);
```

It should be instead:

```
const back = (): Promise<void> => {
  goto($ENABLE_MY_TOKENS && !$isNnsUniverseStore ? AppPath.Tokens : $accountsPathStore);
}
```

What do you think?

dskloetd: I've read this 4 times now, but I don't really understand. What problem are you trying to solve with this proposal?

lmuntaner: The PR is about a bug.

The approach to solve the bug in this PR is to change the implementation of `buildAccountsPath`.

My proposal now is that this approach is misleading. `buildAccountsPath` should be about Accounts. It should not be used for the Tokens page. Changing the implementation to return the Tokens page was a mistake.

The proposal is to revert this mistake and instead, move the logic of the back button up.

Is that clearer now?

dskloetd: Ah, yes, it makes sense that a function called `buildAccountsUrl` only builds URLs for Accounts.

lmuntaner: done

lmuntaner:

dskloetd (approved):

PR #4038, opened 2023-12-14
Add retrieveBtcStatusV2ByAccount to API

dskloetd:
# Motivation

We want to fetch the status of BTC retrievals from the ckBTC minter canister so that we can display which transactions are pending and which have failed.

# Changes

Add `retrieveBtcStatusV2ByAccount` to ckbtc-minter api to call `retrieveBtcStatusV2ByAccount` from ic-js.

# Tests

Unit test was added.

# Todos

not yet


lmuntaner (approved):
LGTM

PR #4039, opened 2023-12-14
chore: Show which invalid label

bitdivine:
# Motivation
Labels are serialized as `u32`.  On parsing, a u32 may not correspond to a valid schema.  There is an error type for tyhis, but it does not indicate what the invalid u32 is, which can be unhelpful.

# Changes
- Include the u32 in the error.

# Tests
- The corresponding test has been updated.

# Todos

  - Too minor


frederikrothenberger (approved):
LGTM!

PR #4040, opened 2023-12-15
clean: Implement Eq for tests only

bitdivine:
# Motivation
Unit tests require `Eq` to verify that state, accounts etc have the expected content.  However the implementations can be prohibitively inefficient with large data so we would prefer to have them present only in test code.

# Changes
- Configure `Eq` and `PartialEq` to be created only in test code.
- For the ProxyDb, the implementation of PartialEq previously depended on converting an arbitrary database format into a `BTreeMap`.  This is unnecessarily expensive as we can iterate of the elements of the two databases regardless of the underlying implementation.

# Tests
- Existing tests should suffice.

# Todos

  Too minor.


frederikrothenberger (approved):
LGTM!

PR #4041, opened 2023-12-15
Tests for migration

bitdivine:
# Motivation

# Changes

# Tests

# Todos

### DIFF
--- a/rs/backend/src/accounts_store/schema/proxy.rs
+++ b/rs/backend/src/accounts_store/schema/proxy.rs
@@ -1,11 +1,15 @@
 //! Accounts DB that delegates API calls to underlying implementations.
 //!
 //! The proxy manages migrations from one implementation to another.
-use std::collections::BTreeMap;
-mod enum_boilerplate;
 use super::{map::AccountsDbAsMap, Account, AccountsDbBTreeMapTrait, AccountsDbTrait, SchemaLabel};
 use core::fmt;
 use core::ops::RangeBounds;
+use std::collections::BTreeMap;
+
+mod enum_boilerplate;
+mod migration;
+#[cfg(test)]
+mod tests;
 
 /// An accounts database delegates API calls to underlying implementations.
 ///
@@ -33,9 +37,9 @@ impl Default for AccountsDbAsProxy {
 struct Migration {
     /// The database being migrated to
     db: AccountsDb,
-    // This is used in the next PR that implements the migration:
-    // /// The next account to migrate.
-    // next_to_migrate: Option<Vec<u8>>,
+    /// The next account to migrate.
+    #[cfg(test)]
+    next_to_migrate: Option<Vec<u8>>,
 }
 
 impl fmt::Debug for Migration {
@@ -51,6 +55,7 @@ pub enum AccountsDb {
     Map(AccountsDbAsMap),
 }
 
+// Constructors
 impl AccountsDbAsProxy {
     /// Creates a new proxy usinga map as the underlying storage.
     pub fn new_with_map() -> Self {
@@ -58,11 +63,6 @@ impl AccountsDbAsProxy {
             authoritative_db: AccountsDb::Map(AccountsDbAsMap::default()),
             migration: None,
         }
-    }
-
-    /// Migration countdown; when it reaches zero, the migration is complete.
-    pub fn migration_countdown(&self) -> u32 {
-        0
     }
 }
 
@@ -138,16 +138,3 @@ impl PartialEq for AccountsDbAsProxy {
     }
 }
 impl Eq for AccountsDbAsProxy {}
-
-#[cfg(test)]
-mod tests {
-    use super::super::tests::{assert_map_conversions_work, test_accounts_db};
-    use super::AccountsDbAsProxy;
-
-    test_accounts_db!(AccountsDbAsProxy::default());
-
-    #[test]
-    fn map_conversions_should_work() {
-        assert_map_conversions_work::<AccountsDbAsProxy>();
-    }
-}
--- a/rs/backend/src/accounts_store/schema/proxy/migration.rs
+++ b/rs/backend/src/accounts_store/schema/proxy/migration.rs
@@ -1,0 +1,78 @@
+//! Code for migration from the authoritative database to a new database.
+use super::*;
+
+impl AccountsDbAsProxy {
+    /// The number of accounts to move per heartbeat.
+    pub const MIGRATION_STEP_SIZE: u32 = 10;
+    /// The progress meter count reserved for finalizing a migration.
+    /// Note: This must be positive and should correspond to a reasonable estimate of the number of blocks needed to complete the migration.
+    pub const MIGRATION_FINALIZATION_BLOCKS: u32 = 1;
+
+    /// Migration countdown; when it reaches zero, the migration is complete.
+    ///
+    /// Note: This is a rough estimate of the number of blocks needed to complete the migration.
+    pub fn migration_countdown(&self) -> u32 {
+        self.migration.as_ref().map_or(0, |migration| {
+            Self::MIGRATION_FINALIZATION_BLOCKS
+                + u32::try_from(
+                    self.authoritative_db
+                        .db_accounts_len()
+                        .saturating_sub(migration.db.db_accounts_len()),
+                )
+                .expect("Huge difference in accounts count")
+                    / Self::MIGRATION_STEP_SIZE
+        })
+    }
+    /// Starts a migration, if needed.
+    ///
+    /// # Panics
+    /// - If the new database is not empty.
+    #[cfg(test)]
+    pub fn start_migrating_accounts_to(&mut self, accounts_db: AccountsDb) {
+        assert!(
+            accounts_db.db_accounts_len() == 0,
+            "The DB to migrate to should be empty"
+        );
+        let migration = Migration {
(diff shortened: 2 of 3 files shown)


rs/backend/src/accounts_store/schema/proxy.rs
```diff
@@ -59,11 +64,6 @@ impl AccountsDbAsProxy {
             migration: None,
         }
     }
```
bitdivine, line 66: Moved and expanded.  Please see the new implementation in migration.rs

rs/backend/src/accounts_store/schema/proxy.rs
```diff
@@ -138,16 +138,3 @@ impl PartialEq for AccountsDbAsProxy {
...
 }
 impl Eq for AccountsDbAsProxy {}
-
```
bitdivine, line 141: Moved.  Please see tests.rs.

rs/backend/src/accounts_store/schema/proxy/tests.rs
```diff
@@ -0,0 +1,77 @@
...
+test_accounts_db!(AccountsDbAsProxy::default());
+
+#[test]
+fn map_conversions_should_work() {
+    assert_map_conversions_work::<AccountsDbAsProxy>();
+}
+
+fn migration_steps_should_work(accounts_db: &mut AccountsDbAsProxy, new_accounts_db: AccountsDb) {
+    // During the migration, the accounts db should behave as if no migration were in progress,
+    // regardless of what CRUD operations are performed.  We check this by running another database
```
frederikrothenberger, line 16: Shouldn't you actually do some CRUD operation to check that the migration is indeed correct, even if you mess with the DB contents while it is in progress?

rs/backend/src/accounts_store/schema/proxy/migration.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn step_migration(&mut self) {
+        if let Some(migration) = &mut self.migration {
+            if let Some(next_to_migrate) = &migration.next_to_migrate {
+                dfn_core::api::print(format!(
+                    "Stepping migration: {:?} -> {:?}",
+                    self.authoritative_db, migration.db
+                ));
+                let mut range = self.authoritative_db.range(next_to_migrate.clone()..);
+                for (key, account) in (&mut range).take(Self::MIGRATION_STEP_SIZE as usize) {
+                    migration.db.db_insert_account(&key, account);
```
frederikrothenberger, line 60: Doesn't this logic create a mismatch between the reported steps necessary and the actual steps taken?
The scenario is:
1. Start migration
2. Insert new accounts with account id > `next_to_migrate`
  * the DB proxy will already insert it into the new DB
  * `migration_countdown` will count that account as already migrated
  * `step_migration` will treat that account as non-migrated and insert it again into the new db

As far as I can tell, it will only throw off the reporting but not the actual contents of the db (as re-inserting the same account should not be harmful). However, you want to give this some thought as well (or document your findings, if you have considered this)...

rs/proposals/src/lib.rs
```diff
@@ -142,7 +142,7 @@ const IDL2JSON_OPTIONS: Idl2JsonOptions = Idl2JsonOptions {
     prog: Vec::new(), // These are the type definitions used in proposal payloads.  If we have them, it would be nice to use them.  Do we?
 };
 
-/// Rust types can include things such as functions.  IDL types, sent over a wire, cannot.  Given that we want an IDLType from a more general type we need toconvert the general type to the more specialized type.
+/// Rust types can include things such as functions.  `IDL` types, sent over a wire, cannot.  Given that we want an `IDLType` from a more general type we need to convert the general type to the more specialized type.
```
frederikrothenberger, line 145: This seems to be a drive-by improvement?
So unrelated question: is the comment actually correct? `Func` exists. So isn't this rather a case of "`IDL` types, sent over a wire, containing Func are not supported."?

rs/backend/src/accounts_store/schema/proxy/migration.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// - If the migration countdown is too large to fit in a `u32`.
+    pub fn migration_countdown(&self) -> u32 {
+        self.migration.as_ref().map_or(0, |migration| {
+            let accounts_to_migrate = self
+                .authoritative_db
+                .db_accounts_len()
+                .saturating_sub(migration.db.db_accounts_len());
+            let blocks_for_migration = accounts_to_migrate.div_ceil(u64::from(Self::MIGRATION_STEP_SIZE));
+            let blocks_for_migration = u32::try_from(blocks_for_migration).expect("Migration will take forever");
+            Self::MIGRATION_FINALIZATION_BLOCKS + blocks_for_migration
```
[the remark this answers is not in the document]
bitdivine: You are right.  I can do a saturating add... and then put my feet up for a long time :-D

rs/backend/src/accounts_store/schema/proxy/migration.rs
bitdivine: The intended logic is:
1. Start migration. (next_to_migrate is set to the first account, if any)
2. Insert accounts with id >= `next_to_migrate`, so if IDs were sequential natural numbers this would migrate 0..9 inclusive.  At the end `next_to_migrate` is set to the next index which woudl be 10 in this example. 
3. Repeat, this time migrating 10..19 inclusive.  And so on.

Looking through the code I believe that it works as expected.  I could update the test to use big endian natural numbers as the keys for the toy data and then it would be reasonably easy to confirm - or discover the bug! :-D

rs/backend/src/accounts_store/schema/proxy/tests.rs
bitdivine: Yes, I'll add a test, but keep this as the vanilla version.

rs/proposals/src/lib.rs
bitdivine: Yes, it's a drive by. There is a spell checker that reports typos and unquoted code.  I didn't read deeper than that.

But interesting question.  Can there be a Func in a canister call argument?  There aren't in any of the types this is used on but could there be one, in future?  I guess that one could argue:  If the type supports it, it can happen!  So yes, I can add support for functions.

rs/backend/src/accounts_store/schema/proxy/migration.rs
```diff
@@ -0,0 +1,80 @@
...
+        ));
+        self.migration = Some(migration);
+    }
+
+    /// Advances the migration by one step.
+    #[cfg(test)]
+    pub fn step_migration(&mut self) {
+        if let Some(migration) = &mut self.migration {
+            if let Some(next_to_migrate) = &migration.next_to_migrate {
+                dfn_core::api::print(format!(
```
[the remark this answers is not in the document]
bitdivine: How do you make sure that `std::println` doesn't get used by accident, e.g. by forgetting to import `ic_cdk::println` ?  `std::println` will compile but print nothing in prod.  Do you  say run a modern alternative to `wasm-nm` to list which function calls are made and fail the wasm build if there are calls to std::println?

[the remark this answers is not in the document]
bitdivine: Note: This is how I have implemented this policy change: https://github.com/dfinity/nns-dapp/pull/4046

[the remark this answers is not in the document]
frederikrothenberger: Disallowing `std::println` via clippy seems like a good solution to me. 👍

rs/proposals/src/lib.rs
frederikrothenberger: > Can there be a Func in a canister call argument?

Yes, arguments can have function pointers in them. Which is actually pretty cool.
One application of that is in the [HTTP gateway protocol](https://internetcomputer.org/docs/current/references/http-gateway-protocol-spec#canister-http-interface), where the `StreamingStrategy` gives you a function pointer to call for the next chunk.

rs/backend/src/accounts_store/schema/proxy/migration.rs
```diff
@@ -0,0 +1,80 @@
...
+                    "Stepping migration: {:?} -> {:?}",
```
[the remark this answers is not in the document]
bitdivine: I haven't found printing the range to be useful and it adds noise.

rs/backend/src/accounts_store/schema/proxy/migration.rs
bitdivine: The vanilla test actually already verifies that 10 accounts are migrated at every step, so we know that we aren't double-moving any accounts.

rs/backend/src/accounts_store/schema/proxy/tests.rs
bitdivine: Done.  Please see the proptest below, starting with `Operation` on line 85.  Yay, I haven't written any proptests since writing the crypto core!

rs/proposals/src/lib.rs
bitdivine: Covered in another PR so closing this.

frederikrothenberger (approved):
LGTM, thanks. As discussed `migration_countdown` might underestimate the amount of work that needs to be done, but this should not harm correctness of the implementation.

PR #4043, opened 2023-12-15
GIX-2181: Work around frozen sns canister

dskloetd:
# Motivation

One SNS's root canister was frozen and it broke all SNSes in nns-dapp.

# Changes

Don't throw an error when one of the SNSes is broken but just filter it out of the wrapper.

# Tests

Tested manually by calling `dfx stop` on one of the root canisters.
Without the fix:
https://drive.google.com/file/d/1xF5zRp62HMpQTmlIWTcuWyl9r5JU0ZT7/view

With the fix:
https://drive.google.com/file/d/19VpUbd27HImQDAe0ywrB78tpFc8zQq9D/view

# Todos


mstrasinskis (approved):
LGTM, thank you!

PR #4046, opened 2023-12-16
Use ic-cdk println

bitdivine:
# Motivation
Move more code from `dfn_core` to `ic_cdk`.

Note: Apparently `ic_cdk::println` defaults to printing to stdout in e.g. unit tets.  Could be useful.

# Changes
- Change calls to `dfn_core..print` to `ic_cdk::println!`.
- Given that `std::println!` does nothing in production, prevent accidental usage of `std::println` in wasms by:
  - Prohibiting `std::println!` and similar macros in `clippy.toml`
  - Running clippy only against `wasm32-unknown-unknown` as `ic_cdk::println` is `std::println` when run natively, so triggers the lint.
    - Note: It would be nicer if the prohibition in `clippy.toml` could be applied only when the target arch is `wasm32`.  I do not see a way of doing that.  It may perhaps be possible in other ways.

# Tests
- See CI

# Todos

- [x] What happens if `std::println` is used by accident?  Does it
  - Fail to build for wasm32? - Nope, it builds.
  - Build but fail at runtime? - Nope
  - Build and simply do nothing at runtime? <-- Yes
  - Something else?
- [x] Make some kind of static test that prevents `std::println` from being used instead of `ic_cdk::println`.

### DIFF
--- a/CHANGELOG-Nns-Dapp-unreleased.md
+++ b/CHANGELOG-Nns-Dapp-unreleased.md
@@ -30,6 +30,7 @@ proposal is successful, the changes it released will be moved from this file to
 * Update Rust version: `1.74.0` -> `1.74.1`
 * Provide space for migration state in the `ProxyDb`.
 * Rename "Launch Pad" to "Launchpad".
+* Use `ic_cdk::println` instead of the `dfn_core` equivalent.
 
 #### Deprecated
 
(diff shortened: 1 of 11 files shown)


frederikrothenberger (approved):
LGTM, thanks!

PR #4048, opened 2023-12-17
Update proposal rendering

sa-github-api:
# Motivation

# Changes

# Tests


lmuntaner:
There is a more recent PR #4290

PR #4057, opened 2023-12-18
Bump SvelteKit v2

peterpeterparker:
# Changes

- bump SvelteKit related dependencies
- adapt `useAtomics` flag in vitest config
- adapt test pattern for new file ashes
- vite build option [assetsInlineLimit](https://vitejs.dev/config/build-options.html#build-assetsinlinelimit) set to 0 for backwards compatibility but, it's maybe something we want to revisit someday


lmuntaner (approved):
LGTM!

PR #4059, opened 2023-12-18
fix: error visibility after clicking on min/max

mstrasinskis:
# Motivation

Solves the remaining error message after clicking on min/max button.
Steps:
- Add a too big number for the dissolve delay.
- Click on the Min/Max
- Input updated to a valid value, but the error is still displayed.

# Changes

- update error visibility after min/max click

# Tests

- "should update error message after Min/Max click"

# Todos

### DIFF
--- a/frontend/src/tests/lib/components/ui/DayInput.spec.ts
+++ b/frontend/src/tests/lib/components/ui/DayInput.spec.ts
@@ -143,7 +143,36 @@ describe("DayInput", () => {
     expect(inputElement.value).toBe("30");
   });
 
-  it("should take into accout the max when rounding up", async () => {
+  it("should update error message after Min/Max click", async () => {
+    const initialSeconds = SECONDS_IN_DAY;
+    const minInSeconds = SECONDS_IN_DAY * 30;
+    const maxInSeconds = SECONDS_IN_YEAR;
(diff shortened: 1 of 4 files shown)


frontend/src/tests/lib/components/ui/DayInput.spec.ts
```diff
@@ -143,7 +143,36 @@ describe("DayInput", () => {
     expect(inputElement.value).toBe("30");
   });
 
-  it("should take into accout the max when rounding up", async () => {
+  it("should update error message after Min/Max click", async () => {
```
lmuntaner, line 146: Could you add a test also in SetDissolveDelay. That would be more accurate to how the user saw the error.

mstrasinskis: Thanks, updated.

lmuntaner (commented):
Let me know when it's ready.

frontend/src/tests/lib/components/ui/DayInput.spec.ts
mstrasinskis: Could you please take a look once again?

lmuntaner (approved):
LGTM! Thanks!

PR #4060, opened 2023-12-18
GIX-2180: Add accountIdentifier UserTokenData

lmuntaner:
# Motivation

To open the TransactionModal or ReceiveModal in the ICP tokens table, I need to know which account to use based on the row clicked.

I could use the title, but that is fragile because the title might change and not be the name of the account.

Instead, I added an optional field to UserTokenData: accounIdentifier.

# Changes

* New field `accountIdentifier` in `UserTokenData`.
* Populate the new field in `icpTokensListUser`.
* Populate the new field in `tokensListUserStore`. I changed the implementation to read all the universes accounts instead of the balance. I needed it to get the account identifier.
* Use function overloading in helper `sumAccounts`. Convenient to make sure the return is a `bigint` if there are accounts. I also changed the functionality slightly to return 0 when the list is empty. That made more sense to me.

# Tests

* Adapt the tests expectation for `icpTokensListUser`.
* Adapt the tests expectation for `tokensListUserStore`.
* Add tests `sumAccounts` edge cases.

# Todos

Not yet necessary.

### DIFF
--- a/frontend/src/lib/derived/tokens-list-user.derived.ts
+++ b/frontend/src/lib/derived/tokens-list-user.derived.ts
@@ -14,29 +14,27 @@ import {
   type UserToken,
   type UserTokenBase,
 } from "$lib/types/tokens-page";
+import { sumAccounts } from "$lib/utils/accounts.utils";
 import { buildAccountsUrl, buildWalletUrl } from "$lib/utils/navigation.utils";
 import { isUniverseNns } from "$lib/utils/universe.utils";
 import { encodeIcrcAccount } from "@dfinity/ledger-icrc";
 import { isNullish, TokenAmountV2 } from "@dfinity/utils";
 import { derived, type Readable } from "svelte/store";
+import type { UniversesAccounts } from "./accounts-list.derived";
 import { tokensListBaseStore } from "./tokens-list-base.derived";
-import {
-  universesAccountsBalance,
-  type UniversesAccountsBalanceReadableStore,
-} from "./universes-accounts-balance.derived";
+import { universesAccountsStore } from "./universes-accounts.derived";
 
 const convertToUserTokenData = ({
-  balances,
+  accounts,
   tokens,
   baseTokenData,
   authData,
 }: {
-  balances: UniversesAccountsBalanceReadableStore;
+  accounts: UniversesAccounts;
   tokens: TokensStoreData;
   baseTokenData: UserTokenBase;
   authData: AuthStoreData;
 }): UserToken => {
-  const balanceUlps = balances[baseTokenData.universeId.toText()]?.balanceUlps;
   const token = tokens[baseTokenData.universeId.toText()]?.token;
   const rowHref = isNullish(authData.identity)
     ? undefined
@@ -48,7 +46,9 @@ const convertToUserTokenData = ({
           owner: authData.identity.getPrincipal(),
         }),
       });
-  if (isNullish(token) || isNullish(balanceUlps)) {
+  const accountsList = accounts[baseTokenData.universeId.toText()];
+  const mainAccount = accountsList?.find(({ type }) => type === "main");
+  if (isNullish(token) || isNullish(accountsList) || isNullish(mainAccount)) {
     return {
       ...baseTokenData,
       balance: "loading",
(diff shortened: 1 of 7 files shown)


lmuntaner:
* `main`
  * **PR #4060** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4060?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4060?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

frontend/src/lib/derived/tokens-list-user.derived.ts
```diff
@@ -48,15 +46,28 @@ const convertToUserTokenData = ({
...
     return {
       ...baseTokenData,
       balance: "loading",
       actions: [],
     };
   }
   const fee = TokenAmountV2.fromUlps({ amount: token.fee, token });
-  const balance = TokenAmountV2.fromUlps({ amount: balanceUlps, token });
+  // For ICP we show the sum of all ICP accounts.
+  // For other tokens we show the main account balance because subaccounts are not yet supported.
```
dskloetd, line 60: I would expect `sumAccounts` to work either way because for other tokens `accountsList` is just a list with a single element. Is that not the case?

lmuntaner: Yes, that is the case, but I wanted to be explicit about what's the balance in the other rows.

At the moment, both would calculate the same balance, but on the other tokens, we don't support subaccounts and this lines represent the Main account, not the balance of all the accounts.

frontend/src/lib/derived/tokens-list-user.derived.ts
```diff
@@ -48,15 +46,28 @@ const convertToUserTokenData = ({
...
+  const balance = TokenAmountV2.fromUlps({
+    amount: isUniverseNns(baseTokenData.universeId)
+      ? sumAccounts(accountsList)
+      : mainAccount.balanceUlps,
+    token,
+  });
+  // For ICP, the row represents all the ICP accounts. Therefore, we don't want to set on accountIdentifier.
```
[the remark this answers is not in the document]
lmuntaner: yes, thanks. I changed it to "the", sounds even better

frontend/src/lib/derived/tokens-list-user.derived.ts
```diff
@@ -67,22 +78,28 @@ const convertToUserTokenData = ({
...
   UserToken[]
 >(
-  [tokensListBaseStore, universesAccountsBalance, tokensStore, authStore],
+  [tokensListBaseStore, universesAccountsStore, tokensStore, authStore],
   ([tokensList, balances, tokens, authData]) =>
     tokensList.map((baseTokenData) =>
-      convertToUserTokenData({ baseTokenData, balances, tokens, authData })
+      convertToUserTokenData({
+        baseTokenData,
+        accounts: balances,
```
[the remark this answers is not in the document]
lmuntaner: oops, I forgot to rename. Done

lmuntaner:

frontend/src/lib/derived/tokens-list-user.derived.ts
dskloetd: But if we ever do support subaccounts for other tokens, then in all likelihood we want want their rows to show the balances of all their accounts as well, right?

For me, adding this unnecessary conditional logic just adds confusion.

dskloetd (approved):

frontend/src/lib/derived/tokens-list-user.derived.ts
lmuntaner: We will need to change the actions if we support subaccounts (depending on how we support them). The actions are for one wallet. If we support subaccounts, the actions will then be like ICP, a "GoToDetail" icon.

Or maybe we'll add all main and subaccounts in this same table. At the moment, the meaning is Main account, that's how I understand it and that's what I want the code to do.

dskloetd: But that is also what the code will do if you don't add all the extra logic.

lmuntaner: Yes, but when I read it it won't be clear what the intention of the code was and whether it was a bug as a feature or on purpose.

PR #4061, opened 2023-12-18
Fix: GoToDetail icon instead of button

lmuntaner:
# Motivation

Clicking on the GoToDetail has the same behavior as clicking on the whole row.

Therefore, I refactored the button to be only an icon with no specific behavior.

# Changes

* Rename GoToDetailButton to GoToDetailIcon and render a `span` instead of a `button`.
* Rename the imported components in TokensTableRow

# Tests

* Rename method `hasGoToDetailButton` and `getGoToDetailButton`.
* Remove test of clicking in the clickGoToDetail directly.

# Todos

Not yet necessary.


lmuntaner:
* `main`
  * **PR #4061** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4061?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4061?utm_source=stack-comment).

frontend/src/lib/components/tokens/TokensTable/actions/GoToDetailIcon.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+  import type { UserTokenData } from "$lib/types/tokens-page";
+  import { IconRight } from "@dfinity/gix-components";
+
+  // svelte-ignore unused-export-let
```
lmuntaner, line 5: We need to same interface in all the action components to have them typed in the mapper:

```
  const actionMapper: Record<
    UserTokenAction,
    ComponentType<SvelteComponent<{ userToken: UserTokenData }>>
  > = {
    [UserTokenAction.GoToDetail]: GoToDetailIcon,
    [UserTokenAction.Receive]: ReceiveButton,
    [UserTokenAction.Send]: SendButton,
  };
```

lmuntaner:

dskloetd (approved):

dskloetd:
I aslo added a comment but it seems to have disappeared.
I was thinking we should have a test that clicking on the icon actually works, so results in going to the details.
So testing that the bug is actually fixed.
It might have to be an e2e test.

PR #4064, opened 2023-12-18
Fix min dissolve delay

mstrasinskis:
# Motivation

The Min button set available minimum value.

# Changes

- min button logic (+1)
- disable editing when maximum was set

# Tests

- SetDissolveDelay should disable input when maximum is already selected
- DayInput supports disabled prop

# Todos


frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
```diff
@@ -227,6 +227,19 @@ describe("SetDissolveDelay", () => {
     expect(getDelayInSeconds(component)).toBe(200 * SECONDS_IN_DAY);
   });
 
+  it("should disable input when maximum is already selected", async () => {
```
lmuntaner, line 230: How about a test for clicking in the minimum twice?

lmuntaner (approved):
Looks good, just a comment but already approving.

frontend/src/tests/lib/components/neurons/SetDissolveDelay.spec.ts
mstrasinskis: Good idea, the test was added.

PR #4067, opened 2023-12-19
build: bump agent-js v0.20.2

krpeacock:
# Motivation

Bump agent-js v0.20.2 without breaking changes - i.e. by disabling per default `verifyQuerySignatures`

# PRs

- [ ] https://github.com/dfinity/ic-js/pull/499

# Changes

# Tests

# Todos


krpeacock:
As it turns out, two changes were required:
* changing `private` `rawKey` and `derKey` into public
* updating the signature of a Princpal's JSON encoding (per https://github.com/dfinity/agent-js/pull/770)

I also have included the following optional suggestions:
* make use of `bufEquals` and `concat` utils from agent-js
* add a `from` constructor on `Secp256k1PublicKey`
* Use the `Secp256k1PublicKey.from` constructor instead of casting using `as`
* Inject the `__derEncodedPublicKey__` value for `DerEncodedPublicKey`s instead of relying on just the typing

peterpeterparker:
Due to the timezone difference and the fact that this PR originates from a fork, I took the liberty to open a new PR to complete the implementation based on your code. I copied your key implementation without making any further changes. I hope this makes sense.

This continues there 👉 #4068

PR #4002, opened 2023-12-11
build: bump agent-js v0.20.2

peterpeterparker:
# Motivation

Bump agent-js v0.20.2 without breaking changes - i.e. by disabling per default `verifyQuerySignatures`

# PRs

- [ ] https://github.com/dfinity/ic-js/pull/499

# Changes

# Tests

# Todos

- [ ] Add entry to changelog (if necessary).

PR #4068, opened 2023-12-19
build: bump agent-js v0.20.2

peterpeterparker:
# Motivation

Bump agent-js v0.20.2 without breaking changes - i.e. by disabling per default `verifyQuerySignatures`

# PR

ic-js:

- [x] https://github.com/dfinity/ic-js/pull/499

This PR follow-up:

- [x] #4002
- [x] #4067

# Changes

- bump ic-js
- replace few `new HttpAgent` with `createAgent` utils and set the remaining contructor to not use `verifyQuerySignatures`
- adapt `secp256k1.ts` as provided by @krpeacock in #4067 (see [description](https://github.com/dfinity/nns-dapp/pull/4067#issuecomment-1861919968)) (except principal JSON)
- adapt `utils.stringifyJson` to parse principals in a backwards compatible manner given that we display those in proposal rendering
- mock agent-js in two tests that were trying to use global.fetch

# Tests

# Todos


frontend/src/lib/api/dev.api.ts
```diff
@@ -359,6 +363,7 @@ export const receiveMockBtc = async ({
 }) => {
   const agent = new HttpAgent({
```
lmuntaner, line 364: Why not using `createAgentUtils` here?

frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
```diff
@@ -21,6 +24,8 @@ describe("Launchpad", () => {
       Promise.resolve([])
     );
 
+    vi.spyOn(agent, "createAgent").mockResolvedValue(mock<HttpAgent>());
```
lmuntaner, line 27: Can you add a todo? It would be good to find out which api function is trying to be called here and mock that one specifically.

That you had to mock this means that there might be something going on that we don't know in this test.

lmuntaner (approved):
LGTM!

When do we set `verifyQuerySignatures` to true and stop relying on update calls?

frontend/src/lib/api/dev.api.ts
peterpeterparker: I would had had to modify the code by passing an identity as parameter which by reading the code, isn't clear if it's a signed in or anonymous identity. Long story short, felt like a change not strictly related to this PR.

frontend/src/tests/workflows/Launchpad/ProdLaunchpad.spec.ts
peterpeterparker: Makes sense. I added the TODOs

frontend/src/lib/api/dev.api.ts
lmuntaner: Makes sense

PR #4069, opened 2023-12-19
A store for BTC retrieval statuses

dskloetd:
# Motivation

We should store them somewhere after fetching them so we can later use them for rendering.

# Changes

Add a store to put the `RetrieveBtcStatusV2WithId` objects that will be fetched by the minter in a separate PR.

# Tests

Added unit tests for the store.
Manually tested in another branch.

# Todos

not necessary


lmuntaner (approved):
LGTM

PR #4070, opened 2023-12-19
Remove past-changelog-test

dskloetd:
# Motivation

Until recently we had unreleased changes and released changes in the same changelog file.
This often resulted in incorrectly resolved merge conflict resulting in new entries ending up as part of old releases.
A test was added to prevent this.
But now that we have unreleased change and released changes in different file, this is unlikely to happen again.
We still have some merge issues with the changelog but they all happen in the unreleased file and not in the released file.
This test now only fails as a false positive when you start a PR before a new release and try to merge it after the new release.

# Changes

1. Remove `scripts/past-changelog-test`.
2. Remove `scripts/past-changelog-test.test`. 
3. Remove them from CI.

# Tests

removed

# Todos


lmuntaner (approved):
Good point, I didn't notice it!

PR #4071, opened 2023-12-19
GIX-2180: Move IcpTransactionModal

lmuntaner:
# Motivation

Users can open the Send ICP modal from the tokens table in the ICP tokens page.

In this PR, I move the IcpTransactionModal from the NnsAccountsFooter to the AccountsModals.

This will make it easy to then open it when clicking in the icon of the tokens table.

# Changes

* Add `"nns-send"` to AccountsModalsType.
* Remove IcpTransactionModal from NnsAccountsFooter and dispatch an event instead.
* Add IcpTransactionModal to AccountsModals.

# Tests

* Added a test in the Accounts route that the user can send ICP by clicking in the button of the footer.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/tests/lib/routes/Accounts.spec.ts
+++ b/frontend/src/tests/lib/routes/Accounts.spec.ts
@@ -27,7 +27,10 @@ import { snsAccountsStore } from "$lib/stores/sns-accounts.store";
 import { tokensStore } from "$lib/stores/tokens.store";
 import { transactionsFeesStore } from "$lib/stores/transaction-fees.store";
 import { page } from "$mocks/$app/stores";
-import { mockAuthStoreSubscribe } from "$tests/mocks/auth.store.mock";
+import {
+  mockAuthStoreSubscribe,
+  mockIdentity,
+} from "$tests/mocks/auth.store.mock";
 import en from "$tests/mocks/i18n.mock";
 import {
   mockAccountDetails,
@@ -51,6 +54,7 @@ import { setCkETHCanisters } from "$tests/utils/cketh.test-utils";
 import { setSnsProjects } from "$tests/utils/sns.test-utils";
 import { runResolvedPromises } from "$tests/utils/timers.test-utils";
 import { SnsSwapLifecycle } from "@dfinity/sns";
+import { ICPToken, TokenAmount } from "@dfinity/utils";
 import { fireEvent, waitFor } from "@testing-library/dom";
 import { render } from "@testing-library/svelte";
 import WalletTest from "../pages/AccountsTest.svelte";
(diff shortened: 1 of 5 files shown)


lmuntaner:
* `main`
  * **PR #4071** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4071?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4071?utm_source=stack-comment).

lmuntaner:

frontend/src/tests/lib/routes/Accounts.spec.ts
```diff
@@ -720,6 +722,46 @@ describe("Accounts", () => {
...
+          destinationAddress: mockSubAccount.identifier,
+          amount: 2,
+        });
+
+        await runResolvedPromises();
+
+        expect(await tablePo.getRowData("Main")).toEqual({
+          balance: "1.14 ICP",
+          projectName: "Main",
+        });
```
dskloetd, line 762: What is the purpose of changing the balance and then checking that the balance has changed?
The changing balance is not a result of the transaction so this doesn't test that the transaction happened and seems entirely orthogonal?

It could make sense if there was a fake that updated the balance based on the transaction, but like this the 2 things are completely separate and you'd have to check a spy to know if the transaction happened.

Although in this test it seems enough to check that the modal opened and doing the transaction may be overkill?

lmuntaner: I can also add expect to the transfer calls.

What is and what is not an overkill is hard to tell.

In this case, this test was missing at this level. There is logic that is passed to the modal, for example, no account is passed here and the main is used as default. Yet, in another one, when clicking in the Send of a row, I want the modal to have the account of the row where it was clicked. Those tests are not anywhere else. We could check that the modal is open with that specific account as well. But here there is also the logic of updating the balance after a transacion. Which is not necessarily tested in the modal in isolation.

I like this level of tests to be honest. Specially because they are easy to write thanks to your page objects.

I will add expect on the transfer calls.

lmuntaner:

frontend/src/tests/lib/routes/Accounts.spec.ts
dskloetd: Thanks. If the test was missing then it's not overkill. I meant that from the perspective of just the changes made in this PR.

You mentioned once that you don't really see the value of the fake APIs.
I think this is a good example.
If we had a fake ledger API, you could actually just check that the balance had changed (without changing it explicitly in the test) and then you knew that the transaction had happened.

lmuntaner: I know, I was actually thinking about it while writing the test.

We would need to centralize all the fakes and keep state. I wonder also the complexity it would bring. But yes, it would make more sense in this scenario.

dskloetd: > We would need to centralize all the fakes

I'm not sure what you mean by that. Do you mean that different APIs (like index and ledger) need to be backed by the same state so they need a single fake together?

> and keep state.

The fakes already keep state. That's kind of the whole point.

lmuntaner: Yes, but the state might need to be centralized. Some flows touch different APIs that share a state. That probably happens now because our API layer is not exactly the canister one. We would need to change that first. Then it would be easier to add the fakes per canister.

PR #4074, opened 2023-12-19
refactor: WASM_PAGE_BYTES is provided by the cdk

bitdivine:
# Motivation
We define `WASM_PAGE_SIZE` but it is already provided by the CDK, so we can deduplicate.

# Changes
- Use the ic_cdk definition of Wasm page size.

# Tests
- I have verified that the two constants have the same value.

# Todos

  - Too minor


frederikrothenberger (approved):
LGTM

PR #4075, opened 2023-12-19
feat: Implement Storable for Account

bitdivine:
# Motivation
We wish to store accounts in stable memory.

# Changes
- Implement the `Storable` trait for `Account`s

# Tests
- A unit test is included.

# Todos

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -13,11 +13,13 @@ use ic_ledger_core::timestamp::TimeStamp;
 use ic_ledger_core::tokens::SignedTokens;
 use ic_nns_common::types::NeuronId;
 use ic_nns_constants::{CYCLES_MINTING_CANISTER_ID, GOVERNANCE_CANISTER_ID};
+use ic_stable_structures::{storable::Bound, Storable};
 use icp_ledger::Operation::{self, Approve, Burn, Mint, Transfer, TransferFrom};
 use icp_ledger::{AccountIdentifier, BlockIndex, Memo, Subaccount, Tokens};
 use itertools::Itertools;
 use on_wire::{FromWire, IntoWire};
 use serde::Deserialize;
+use std::borrow::Cow;
 use std::cmp::{min, Ordering};
 use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
 use std::fmt;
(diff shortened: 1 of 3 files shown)


rs/backend/src/accounts_store.rs
```diff
@@ -96,6 +98,20 @@ pub struct Account {
     canisters: Vec<NamedCanister>,
 }
 
+impl Storable for Account {
+    const BOUND: Bound = Bound::Unbounded;
+    fn to_bytes(&self) -> Cow<'_, [u8]> {
+        let account_serialized = Candid((self,)).into_bytes().expect("Failed to serialize account");
```
peterpeterparker, line 104: I remember that I've heard once that using Candid for this particular operation is less approriate than other serializer/deserializer. That's notably why I use ciborium (`use ciborium::{from_reader, into_writer};`) as displayed in the stable-structure repo.

Should we do so here as well?

peterpeterparker: Feel free to ignore my comment, just randomly spoted the PR so it's more a "out of curiosity" comment than call to action.

peterpeterparker: It's more like a vague souvenir but, maybe it has to do with the size being more predictable.

Anyway just noticed that in Oisy we use candid (https://github.com/dfinity/oisy-wallet/blob/main/src/backend/src/lib.rs) so never mind my comment. I should have ckeched more examples.

bitdivine: CBOR includes field names in the encoding, so here I would expect the CBOR encoding to be less efficient than Candid.  If you have loosely typed data, sure, CBOR is probably a better choice but here we have the exact same type 220,000 times.

I added the required traits and checked the above (toy) example:
```
candid len: 712 cbor len: 904
```
so CBOR took 27% more space.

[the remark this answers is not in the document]
bitdivine: Done.

rs/backend/src/accounts_store.rs
```diff
@@ -96,6 +98,20 @@ pub struct Account {
...
+        account_serialized.to_owned().into()
+    }
+    fn from_bytes(bytes: Cow<'_, [u8]>) -> Self {
+        let (account,) = Candid::from_bytes(bytes.as_ref().to_owned())
```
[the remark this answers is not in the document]
bitdivine: Done

PR #4076, opened 2023-12-19
Service method to load BTC retrieval statuses

dskloetd:
# Motivation

# Changes

Add a service method `loadRetrieveBtcStatuses` which loads BTC retrieval statuses and puts them in the store.

# Tests

Unit tests were added.

# Todos

will add when it's used


frontend/src/lib/services/ckbtc-minter.services.ts
```diff
@@ -251,3 +255,22 @@ export const getWithdrawalAccount = async ({
     return undefined;
   }
 };
+
+export const loadRetrieveBtcStatuses = async ({
```
lmuntaner, line 259: Do we want to call this once per user session or can the user trigger it and we want to still hit the endpoint?

Just wondering whether you want to avoid calling the endpoint if the store is filled.

lmuntaner (approved):
LGTM, just a question

frontend/src/lib/services/ckbtc-minter.services.ts
dskloetd: I want to call it when opening the wallet and when reloading transactions.
The status can change even if the transactions don't change so I don't want to skip it.

PR #4079, opened 2023-12-19
Rename formatToken to formatTokenE8s

dskloetd:
# Motivation

Now that we have ckETH which has 18 decimals, it's important to be clear when a function only supports 8 decimals.
`formatToken` only supports 8 decimals. For a different number of decimals, we should use `formatTokenV2` and `TokenAmountV2`.

# Changes

Rename `formatToken` to `formatTokenE8s`.

# Tests

existing tests pass

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #4082, opened 2023-12-19
Check icons in TransactionCard test

dskloetd:
# Motivation

`TransactionCard` renders both an icon image and class.
So far we only tested the class.

# Changes

Also test that we have the right icon image in `TransactionCard.page-object.ts`.

# Tests

only tests

# Todos

not necessary

### DIFF
--- a/frontend/src/tests/page-objects/TransactionCard.page-object.ts
+++ b/frontend/src/tests/page-objects/TransactionCard.page-object.ts
@@ -37,13 +37,32 @@ export class TransactionCardPo extends BasePageObject {
     return this.getAmountDisplayPo().getAmount();
   }
 
-  async hasPendingIcon(): Promise<boolean> {
+  async hasIconClass(className: string): Promise<boolean> {
     const classNames = await this.root.byTestId("icon").getClasses();
-    return classNames.includes("pending");
+    return classNames.includes(className);
+  }
+
+  async hasSentIcon(): Promise<boolean> {
+    const hasIcon = await this.isPresent("icon-up");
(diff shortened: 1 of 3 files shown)


frontend/src/tests/page-objects/TransactionCard.page-object.ts
```diff
@@ -37,13 +37,32 @@ export class TransactionCardPo extends BasePageObject {
...
-  async hasPendingIcon(): Promise<boolean> {
+  async hasIconClass(className: string): Promise<boolean> {
     const classNames = await this.root.byTestId("icon").getClasses();
-    return classNames.includes("pending");
+    return classNames.includes(className);
+  }
+
+  async hasSentIcon(): Promise<boolean> {
+    const hasIcon = await this.isPresent("icon-up");
+    const hasClass = await this.hasIconClass("send");
```
lmuntaner, line 47: Why do we want to check the class? Do we test UI in unit tests? Would it make more sense a screenshot test?

lmuntaner (approved):
LGTM, just one question.

frontend/src/tests/page-objects/TransactionCard.page-object.ts
dskloetd: The class is added based on [logic, here](https://github.com/dfinity/nns-dapp/blob/6a241399f2560ffbd664a1e4984ae4308de1c1f0/frontend/src/lib/components/accounts/TransactionCard.svelte#L53). So we're testing that logic.

PR #4083, opened 2023-12-19
fix: Spellcheck

bitdivine:
# Motivation
The spellcheck in CI does not prevent typos from being merged into main.

On investigation, `cargo spellcheck` and `cargo spellcheck check` both return 0 even if errors are found.  The manual shows that `--code 1` or similar needs to be added to return non-zero if errors are found.

# Changes
- Add `--code 1` to calls to `cargo spellcheck`.
- Fix spelling and typographic issues.

# Tests
See CI.

# Todos


rs/proposals/src/canisters/nns_governance/api.rs
```diff
@@ -1,4 +1,4 @@
-//! Rust code created from candid by: scripts/did2rs.sh --canister nns_governance --out api.rs --header did2rs.header --traits Serialize
+//! Rust code created from candid by: `scripts/did2rs.sh --canister nns_governance --out api.rs --header did2rs.header --traits Serialize`
```
bitdivine, line 1: Note: This is generated code.  I updated the code that generates this and re-ran the generation command.

lmuntaner (approved):
Thanks!

PR #4084, opened 2023-12-19
Take retrieval status into account in mapCkbtcTransaction

dskloetd:
# Motivation

These statuses are not part of the transactions themselves but come from the ckBTC minter and then put in a store.

In this PR we change the mapping function to take the status into account when creation the `UiTransaction`.

# Changes

1. Add a new `isFailed` field to `UiTransaction` to indicate a failed transaction.
2. Add an optional parameter to `mapCkbtcTransaction` to pass the retrieval status if there is one.
3. If there is a status, output a pending or failed transaction if appropriate.
4. Add i18n strings for the transaction headlines.

# Tests

Unit tests added.
Tested manually in another branch with more changes.

# Todos

not yet

### DIFF
--- a/frontend/src/lib/i18n/en.json
+++ b/frontend/src/lib/i18n/en.json
@@ -1025,7 +1025,9 @@
     "btc_sent": "BTC Sent",
     "btc_network": "BTC Network",
     "receiving_btc": "Receiving BTC",
-    "reimbursement": "Reimbursement"
+    "reimbursement": "Reimbursement",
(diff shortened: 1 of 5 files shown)


lmuntaner (approved):
LGTM!

I guess there still needs the logic of mapping the status to a specific transaction, right?

dskloetd:
> I guess there still needs the logic of mapping the status to a specific transaction, right?

That's correct!

PR #4085, opened 2023-12-20
Update SNS aggregator url in documention

peterpeterparker:
# Motivation

For consistency, let's always use the extension `icp0.io` which is the one we also used in the documentation and in the code.


mstrasinskis (approved):
Rubber stamp, thank you!

PR #4089, opened 2023-12-20
GIX-2184: Fix padding icp tokens page

lmuntaner:
# Motivation

There was a visual bug in the ICP Tokens page because the outer paddings weren't the same as in the Tokens page. This bug was visible in mobile devices.

The problem came from an extra padding added by a `main`'s default padding inside the Accounts route. This `main` will be removed when the flag is removed, but for now, I fixed it by passing a class and setting the padding to 0 when the flag is enabled.

# Changes

* Add a new class `noPadding` to the `main` in Accounts route and set it when ENABLE_MY_TOKENS is enabled.

# Tests

Only UI changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/routes/Accounts.svelte
+++ b/frontend/src/lib/routes/Accounts.svelte
@@ -34,6 +34,7 @@
   import { icrcCanistersStore } from "$lib/stores/icrc-canisters.store";
   import IcrcTokenAccountsFooter from "$lib/components/accounts/IcrcTokenAccountsFooter.svelte";
   import IcrcTokenAccountsModals from "$lib/modals/accounts/IcrcTokenAccountsModals.svelte";
+  import { ENABLE_MY_TOKENS } from "$lib/stores/feature-flags.store";
 
   // TODO: This component is mounted twice. Understand why and fix it.
 
@@ -117,7 +118,8 @@
 </script>
 
 <TestIdWrapper testId="accounts-component">
-  <main>
+  <!-- TODO: Remove the `main` element and the rest of unused elements when removing flag ENABLE_MY_TOKENS -->
+  <main class:noPadding={$ENABLE_MY_TOKENS}>
     <SummaryUniverse />
 
     {#if $isNnsUniverseStore}
@@ -154,5 +156,9 @@
 <style lang="scss">
   main {
     padding-bottom: var(--footer-height);
+
+    &.noPadding {
+      padding: 0;
+    }
   }
 </style>


lmuntaner:
* `main`
  * **PR #4089** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4089?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4089?utm_source=stack-comment).

lmuntaner:
Before:



Now:



The page still has the padding if the feature is not enabled:

lmuntaner:

frontend/src/lib/routes/Accounts.svelte
```diff
@@ -117,7 +118,8 @@
 </script>
 
 <TestIdWrapper testId="accounts-component">
-  <main>
+  <!-- TODO: Remove the `main` element and the rest of unused elements when removing flag ENABLE_MY_TOKENS -->
+  <main class:noPadding={$ENABLE_MY_TOKENS}>
```
dskloetd, line 122: Why does it only happen when `ENABLE_MY_TOKENS`?
It makes me wonder if there is a more proper place to fix it where it doesn't depend on a feature flag.

lmuntaner: Yes, I could change the layout.svelte of this page to accomodate for this extra padding. But that layout is shared with the Tokens page; therefore, I would need a special prop there.

I see a change here more adequate because when we remove the flag, I will also remove thie `main` from here and nothing else should change. If I move the logic to the layout I won't be able to remove this `main` without touching the layout.

In conclusion, by making the change in this file, I ensure that only this files gets changed when we remove the flag and the rest will stay the same. It will give a cleaner diff when we remove the flag.

dskloetd: Since I still wasn't sure about the answer to my question (why it only happens with `ENABLE_MY_TOKENS`), I did some digging myself.
It looks like with `ENABLE_MY_TOKENS`, you add an extra `main` [here](https://github.com/dfinity/nns-dapp/blob/b9b7d54938ed6a171128276718a4b2b7603beec8/frontend/src/routes/(app)/(u)/(accounts)/%2Blayout.svelte#L19), which causes extra padding compared to when `ENABLE_MY_TOKENS` is disabled (and also results in nested `main` tags which seems strange).

Why do we need that `main`? Could we solve this issue by removing that `main` instead of having conditional logic here?

Also, just for my understanding, is it true that with `ENABLE_MY_TOKENS`, this page will only be for the NNS universe and not any of the others listed here?

lmuntaner: I expect `MainWrapper` to be part of all the pages where we have a tokens table. That's why I put it in the Layout.

What we also have here is weird because the `main` in this case is in a third level.

`src/routes/(app)/(u)/(accounts)/accounts/+page.svelte` -> `src/lib/routes/Accounts.svelte` -> `src/lib/pages/NnsAccounts.svelte`.

I find this nesting quite confusing. For example, the `main` is missing in the other components, it's only in `NnsAccounts`. I'm thinking on the way to have it cleaner after the refactor.

After the refactor I see `src/lib/pages/NnsAccounts.svelte` disappearing and instead have the following in `src/lib/routes/Accounts.svelte`:

```
<TokensTable ... />
<Footer />
<Modals />
```

Another option is to repeat the `MainWrapper` where it's used instead of the layout:

```
<MainWrapper />
  // With the logic of signin or not here
  <TokensTable />
</MainWrapper>
```

We don't have a written rule yet.

dskloetd: OK, thanks for explaining.
So I think another option now would be to have this, correct?
```
<TestIdWrapper testId="accounts-component">
  <!-- TODO: Remove the `main` element and the rest of unused elements when removing flag ENABLE_MY_TOKENS -->
  {#if $ENABLE_MY_TOKENS}
    <SummaryUniverse />
    {#if $isNnsUniverseStore}
      <NnsAccounts userTokensData={$icpTokensListUser} />
    {/if}
  {/else}
    <main>
      <SummaryUniverse />
      {#if $isNnsUniverseStore}
        <NnsAccounts userTokensData={$icpTokensListUser} />
      {:else if $isCkBTCUniverseStore}
        <CkBTCAccounts />
      {:else if $isIcrcTokenUniverseStore}
        <IcrcTokenAccounts />
      {:else if nonNullish($snsProjectSelectedStore)}
        <SnsAccounts />
      {/if}
    </main>
  {/if}
  ...
```
Although in that case I guess you could still deep link to accounts of other tokens and they would just be empty?
Anyway, up to you.

dskloetd (approved):

frontend/src/lib/routes/Accounts.svelte
lmuntaner: Yes, that's correct.

I have an issue to redirect to `/tokens` if Accounts is rendered with anything but NNS universe.

I'll merge this and we can improve once we remove the flag.

PR #4096, opened 2023-12-20
GIX-2079: Take retrieval status into account in mapCkbtcTransactions

dskloetd:
# Motivation

These statuses are not part of the transactions themselves but come from the ckBTC minter and then put in a store.

In this PR we pass the status from `mapCkbtcTransactions` (plural) to `mapCkbtcTransaction` singular.
We don't yet pass the statuses to `mapCkbtcTransactions` (plural).

# Changes

1. Add a `retrieveBtcStatuses` parameter to `mapCkbtcTransactions`.
2. Convert the array to a map and use the map to pass the correct status to each invocation of `mapCkbtcTransaction`.

# Tests

1. Extract some test values from test and put them in a larger scope so they can be used.
2. Add a unit test with multiple statuses for different transactions and test that they are applied correctly.

# Todos

not yet

### DIFF
--- a/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
@@ -80,6 +80,8 @@
       account,
       token,
       i18n: $i18n,
+      // TODO GIX-2079: Pass actual statuses.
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
LGTM!

PR #4098, opened 2023-12-20
feat: Don't save a map on the heap except when it makes sense

bitdivine:
# Motivation
When accounts are stored in stable memory, it no longer makes sense to save the accounts with all the other data on the heap.

# Changes
- Serialize accounts in the `pre_upgrade` hook only when accounts are stored as a `BTreeMap` on the heap.

# Tests
- Existing tests for `Map` to `Map` upgrades still suffice.

# Todos


frederikrothenberger (approved):
LGTM 👍

PR #4099, opened 2023-12-21
Spellcheck

bitdivine:
# Motivation
The spell checker has a lot of false hits in markdown files, which makes it hard to see genuine errors.  The false hits are primarily from unformatted code.

# Changes
- Format code as code, so that spellcheck ignores it.
- Add legitimate well known words to the dictionary.

# Tests
I have NOT added the markdown spellchecker to CI as it chokes on the fancy HTML in `README.html`.  Probably want to address that first.  Just having less noise is a huge improvement.

# Todos

  - Too minor


lmuntaner (approved):
Thanks!

PR #4100, opened 2023-12-21
Save stats

bitdivine:
# Motivation
Currently, in the post_upgrade hook stats are computed over all accounts.  This will become too expensive once accounts are stored in stable memory.

Instead we would like to preserve these stats on upgrade.

# Changes
- Save the `accounts_db` stats in the `pre_upgrade()` hook.
- In the `post_upgrade()` hook, compute stats only if they are not already available.

# Tests
- The downgrade-upgrade test verifies that this is forwards and backwards compatible.

# Todos

### DIFF
--- a/rs/backend/src/accounts_store.rs
+++ b/rs/backend/src/accounts_store.rs
@@ -50,9 +50,7 @@ pub struct AccountsStore {
     neuron_accounts: HashMap<AccountIdentifier, NeuronDetails>,
     block_height_synced_up_to: Option<BlockIndex>,
     multi_part_transactions_processor: MultiPartTransactionsProcessor,
-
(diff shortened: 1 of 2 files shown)


rs/backend/src/accounts_store.rs
```diff
@@ -1607,6 +1611,7 @@ impl StableState for AccountsStore {
             MultiPartTransactionsProcessor,
             u64,
             u64,
+            Option<AccountsDbStats>,
```
bitdivine, line 1614: Note: The field is optional so that upgrades from previous versions, without this field, succeed.

frederikrothenberger (approved):
LGTM 👍

PR #4101, opened 2023-12-21
Fix SnsWallet test logo expectation

dskloetd:
# Motivation

I was trying to clean up `SnsWallet.spec.ts` by using page objects but I found out that the test is broken. I'm fixing the test in a separate PR because I didn't want it to be hidden in the cleanup PR. So this PR is only the fix and not the cleanup.

`SnsWallet.spec.ts` has a test for the alt text on the logo on the receive QR code.
But this test does not do what it seems to do.
Because there is a logo already present in the wallet component and when the test does `getByTestId("logo")` it actually gets that logo and not the logo from the receive QR code.
The logo from the receive QR code has a different alt text which is just the token symbol.

Additionally the receive QR code is not even present in the DOM at the time of the expectation because the `ReceiveModal` component waits for the `introend` event, [here](https://github.com/dfinity/nns-dapp/blob/bf19f8108b2bc13e35f10be3aec75f8beeb29f1d/frontend/src/lib/modals/accounts/ReceiveModal.svelte#L40), before rendering the QR code.

Lastly, the token that is rendered on the QR code comes from SNS project store (normally populated by the sns aggregator) rather than from the response of `getSnsToken`. And in the test these differed because `getSnsToken` was mocked to return `mockSnsToken`, while `setSnsProjects` uses a different default token from [aggregatorSnsMockDto](https://github.com/dfinity/nns-dapp/blob/bf19f8108b2bc13e35f10be3aec75f8beeb29f1d/frontend/src/tests/mocks/sns-aggregator.mock.ts#L23).

# Changes

1. Specify an explicit token to use in the test rather than relying on mock values defined in other files.
2. Use the explicit token value in both the mock `getSnsToken` and `setSnsProjects`.
3. In `"should open receive modal with sns logo"` wait for the QR code to be rendered.
4. Then query for the actual logo in the receive modal rather than just any logo.
5. And expect just the token symbol as alt text because that's the actual alt text we render in the QR code.

# Tests

This is a test-only PR.

# Todos

not necessary

### DIFF
--- a/frontend/src/tests/lib/pages/SnsWallet.spec.ts
+++ b/frontend/src/tests/lib/pages/SnsWallet.spec.ts
@@ -57,6 +57,15 @@ vi.mock("$lib/services/worker-balances.services", () => ({
 }));
 
 describe("SnsWallet", () => {
+  const testTokenSymbol = "OOO";
+  const testTokenName = "Out of office";
+
+  const testToken = {
+    ...mockSnsToken,
+    name: testTokenName,
+    symbol: testTokenSymbol,
+  };
+
   const props = {
     accountIdentifier: mockSnsMainAccount.identifier,
   };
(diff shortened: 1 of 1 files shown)


mstrasinskis (approved):
Nice, thank you!

PR #4102, opened 2023-12-21
Render failed BTC withdrawal transactions

dskloetd:
# Motivation

We want to render failed "Sending BTC" transactions as such.
This PR has just the UI changes to render failed transaction.
Detecting failed transactions is done in other PRs.

# Changes

1. If `UiTransaction.isFailed` is true render a different icon.
2. Use a CSS class that renders the icon in a red color. This is the same color that reimbursed transactions also use.
3. Add a method to the page object to check for the failed icon and class.
4. Test that `isFailed` results in using the failed icon.

# Tests

Unit test is included.

# Todos

not yet.


mstrasinskis (approved):
Thanks!

PR #4103, opened 2023-12-21
Use page objects in SnsWallet.spec.ts

dskloetd:
# Motivation

Make the test easier to maintain and less fragile.

# Changes

1. Add test IDs to existing components.
2. Add necessary page object functionality.
3. Use page objects in `frontend/src/tests/lib/pages/SnsWallet.spec.ts`.

# Tests

only

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/pages/SnsWallet.svelte
+++ b/frontend/src/lib/pages/SnsWallet.svelte
@@ -11,6 +11,7 @@
     type WalletContext,
     type WalletStore,
   } from "$lib/types/wallet.context";
+  import TestIdWrapper from "$lib/components/common/TestIdWrapper.svelte";
   import Footer from "$lib/components/layout/Footer.svelte";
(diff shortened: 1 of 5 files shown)


frontend/src/lib/pages/SnsWallet.svelte
```diff
@@ -122,72 +123,74 @@
...
-            })}
-            accountName={$selectedAccountStore.account.name ??
-              $i18n.accounts.main}
-          />
-
-          <Separator spacing="none" />
-
-          <SnsTransactionsList
+<TestIdWrapper testId="sns-wallet-component">
+  <Island>
```
dskloetd, line 127: This is just indented. There are no other changes than adding the `TestIdWrapper`.

mstrasinskis (approved):
Thank you!

PR #4108, opened 2023-12-22
Pass BTC retrieval statuses when mapping ckBTC transactions

dskloetd:
# Motivation

These statuses are not part of the transactions themselves but come from the ckBTC minter and will be put in a store (in a future PR).

In this PR we take the statuses from the store and pass them to `mapCkbtcTransactions` so they can be used to map the transactions.

# Changes

In `CkBTCTransactionsList.svelte`, pass statuses from `ckbtcRetrieveBtcStatusesStore` to `mapCkbtcTransactions`.

# Tests

Added a unit test where we put a status in the store and verify that it renders the corresponding transaction according to the status.

# Todos

will add when we fill the store

### DIFF
--- a/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCTransactionsList.svelte
@@ -2,6 +2,7 @@
 <svelte:options accessors />
 
 <script lang="ts">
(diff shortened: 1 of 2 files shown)


mstrasinskis (approved):
LGTM, thanks!

PR #3801, opened 2023-11-20
Sns type filtering

mstrasinskis:
# Motivation

Update `snsFilteredProposalsStore` logic to make sns  proposal `types` filter work.

# Changes

- update `isSns(Native|Generic)NervousSystemFunction` to rely on `MIN_VALID_SNS_GENERIC_NERVOUS_SYSTEM_FUNCTION_ID`. To have the same same logic as `isSnsGenericNervousSystemTypeProposal`.
- `isSnsGenericNervousSystemTypeProposal` util
- update `snsFilteredProposalsStore` logic to filter by types.

# Tests

- add `action` param to `createSnsProposal` mock
- `isGenericNervousSystemTypeProposal`
- `loadSnsProposals` provides `excludeType` parameter
- `SnsProposals` component has working types filtering
- `snsFilteredProposalsStore` should filter by types

# Todos

- [ ] Add entry to changelog (if necessary).
Not yet

PR #4109, opened 2023-12-22
Ns function helpers

mstrasinskis:
# Motivation

Helpers to get the nsFunction type.
Part of https://github.com/dfinity/nns-dapp/pull/3801

# Changes

- extend nsFunction mock
- `isNativeNervousSystemFunction`
- `isGenericNervousSystemFunction`

# Tests

- tests for new utils

# Todos

Not necessary

### DIFF
--- a/frontend/src/lib/utils/sns.utils.ts
+++ b/frontend/src/lib/utils/sns.utils.ts
@@ -178,12 +178,9 @@ export const swapEndedMoreThanOneWeekAgo = ({
  */
 export const isNativeNervousSystemFunction = (
   nsFunction: SnsNervousSystemFunction
-): boolean => {
-  return (
-    Object.keys(fromNullable(nsFunction.function_type) ?? {})?.[0] ===
-    "NativeNervousSystemFunction"
-  );
-};
+): boolean =>
+  "NativeNervousSystemFunction" in
+  (fromNullable(nsFunction.function_type) ?? {});
 
 /**
  * Returns true if the FunctionType is GenericNervousSystemFunction (custom per sns).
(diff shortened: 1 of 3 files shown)


frontend/src/lib/utils/sns.utils.ts
```diff
@@ -171,3 +172,24 @@ export const swapEndedMoreThanOneWeekAgo = ({
...
+  );
+};
+
+/**
+ * Returns true if the FunctionType is GenericNervousSystemFunction (custom per sns).
+ */
+export const isGenericNervousSystemFunction = (
+  nsFunction: SnsNervousSystemFunction
+): boolean =>
+  Object.keys(fromNullable(nsFunction.function_type) ?? {})?.[0] ===
```
dskloetd, line 194: same as above

frontend/src/lib/utils/sns.utils.ts
```diff
@@ -171,3 +172,24 @@ export const swapEndedMoreThanOneWeekAgo = ({
...
 };
+
+/**
+ * Returns true if the FunctionType is NativeNervousSystemFunction (same for all same-version snses).
+ */
+export const isNativeNervousSystemFunction = (
+  nsFunction: SnsNervousSystemFunction
+): boolean => {
+  return (
+    Object.keys(fromNullable(nsFunction.function_type) ?? {})?.[0] ===
```
dskloetd, line 183: You use `{}` as fallback if `fromNullable(nsFunction.function_type)` is `null`.
But `{}` has no keys so then `Object.keys(fromNullable(nsFunction.function_type) ?? {})` would be the empty list and you can't get `[0]` from the empty list.

Do we need to use `Objecy.keys` or could we do this instead?
```
  return "NativeNervousSystemFunction" in (fromNullable(nsFunction.function_type) ?? {});
```

It seems like that would be simpler as well.

mstrasinskis: Good point, using `in` is much cleaner, thanks.

dskloetd (approved):

PR #4110, opened 2023-12-22
feat: sns types filter button

mstrasinskis:
# Motivation

Prepare data needed for the `types` filtering and display the `Types` filter button. 

In order to filter by SNS proposal types, the following components are still missing:

1. update `snsFiltersStore` with types and add button to open the filter modal (this pr)
2. filter modal
3. update sns proposal service to utilise the types filter data

Since there are no entities such as 'topics' (a topic being a group of types) and there are no plans to present them, the 'topics' filter type was renamed to 'types'.

Part of: https://github.com/dfinity/nns-dapp/pull/3801

# Changes

- rename not used "topics" to "types"
- util `generateSnsProposalTypeFilterData`
  - ignores `ALL_SNS_PROPOSAL_TYPES_NS_FUNCTION_ID`
  - replace all generic ns functions with a single filter item
- Update types filter store in SnsProposalsFilters.
- Add `FiltersButton`component to `SnsProposalsFilters` (temporary behind `IS_TEST_ENV`)

# Tests

- `generateSnsProposalTypeFilterData`
- SnsProposalsFilters should update the store and display a button

# Todos

Not yet

# Screenshot

When enabled
image


frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
```diff
@@ -28,6 +46,17 @@
 </script>
 
 <FiltersWrapper>
+  {#if IS_TEST_ENV}
```
dskloetd, line 49: If `IS_TEST_ENV`? Is this a test feature like "Make dummy proposals"?
Or should it have a feature flag?

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
```diff
@@ -10,15 +10,33 @@
...
   import SnsFilterRewardsModal from "$lib/modals/sns/proposals/SnsFilterRewardsModal.svelte";
+  import type { SnsNervousSystemFunction } from "@dfinity/sns";
+  import { nonNullish } from "@dfinity/utils";
+  import { generateSnsProposalTypesFilterData } from "$lib/utils/sns-proposals.utils";
+  import { IS_TEST_ENV } from "$lib/constants/mockable.constants";
 
-  let modal: "topics" | "rewards" | "status" | undefined = undefined;
+  export let nsFunctions: SnsNervousSystemFunction[] | undefined;
+
+  let modal: "types" | "rewards" | "status" | undefined = undefined;
```
dskloetd, line 20: I see `"topics"` was an option before but not anymore?
Is the feature ready enough not to have `"topics"` as an option anymore?
I don't see anything about "topics" in the PR description.
In fact, the word "proposal" doesn't even appear in the "Motivation" section.
I think I need some more context and the PR description would be the ideal place to provide it.

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
mstrasinskis: A feature flag could be a cleaner solution, but since we don't plan to test this feature on the mainnet, creating one would be overkill. 'IS_TEST_ENV' was added temporarily. It will remain until the other parts of the story are merged, ensuring a deploy-ready state of the main.

dskloetd: Adding a feature flag is cheap and easy. I don't see in what way it would be overkill.
It would also make it easier to clean up later because you can just search for the feature flag instead of having to consider whether something should stay in the test environment.
Ensuring the deploy ready state of main while developing a feature is exactly what feature flags are for.

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
mstrasinskis: Good point, the description was updated. Please let me know if you need more context.

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
mstrasinskis: Replaced with `ENABLE_SNS_TYPES_FILTER` feature flag.

mstrasinskis:
Obsoleted.

PR #4111, opened 2023-12-22
Load BTC retrieval statuses in ckBTC wallet

dskloetd:
# Motivation

In this PR we call the service method that loads the statuses and puts them in the store.
This is the final PR that ties it all together.

# Changes

1. Call `loadRetrieveBtcStatuses` when opening the `CkBTCWallet` component.
2. Also call `loadRetrieveBtcStatuses` after making a transaction.

# Tests

2 unit tests added.

# Todos

### DIFF
--- a/frontend/src/lib/pages/CkBTCWallet.svelte
+++ b/frontend/src/lib/pages/CkBTCWallet.svelte
@@ -13,6 +13,7 @@
   import { CKBTC_ADDITIONAL_CANISTERS } from "$lib/constants/ckbtc-additional-canister-ids.constants";
   import CkBTCInfoCard from "$lib/components/accounts/CkBTCInfoCard.svelte";
   import type { TokensStoreUniverseData } from "$lib/stores/tokens.store";
+  import { loadRetrieveBtcStatuses } from "$lib/services/ckbtc-minter.services";
   import { loadCkBTCInfo } from "$lib/services/ckbtc-info.services";
   import IcrcWalletPage from "$lib/components/accounts/IcrcWalletPage.svelte";
   import { writable } from "svelte/store";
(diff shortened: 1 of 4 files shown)


frontend/src/lib/pages/CkBTCWallet.svelte
```diff
@@ -47,7 +48,15 @@
   // transactions?.reloadTransactions?.() returns a promise.
   // However, the UI displays skeletons while loading and the user can proceed with other operations during this time.
   // That is why we do not need to wait for the promise to resolve here.
-  const reloadTransactions = () => transactions?.reloadTransactions?.();
+  const reloadTransactions = () => {
+    $selectedCkBTCUniverseIdStore &&
+      canisters &&
```
mstrasinskis, line 53: Nitpick: check for `minterCanisterId` presence before usage
```
      canisters?.minterCanisterId &&
```

frontend/src/lib/pages/CkBTCWallet.svelte
```diff
@@ -73,6 +82,13 @@
       universeId: $selectedCkBTCUniverseIdStore,
       minterCanisterId: canisters?.minterCanisterId,
     }))();
+
+  $: $selectedCkBTCUniverseIdStore &&
+    canisters &&
```
mstrasinskis, line 87: Same here.

mstrasinskis (approved):
Thank you! Just a single comment.

frontend/src/lib/pages/CkBTCWallet.svelte
dskloetd: It's actually not optional.
https://github.com/dfinity/nns-dapp/blob/24859b2029d8fb5e1fc36a276227b43effdad6c7/frontend/src/lib/types/ckbtc-canisters.ts#L5

I've removed the question mark on line 56 to avoid confusion.

frontend/src/lib/pages/CkBTCWallet.svelte
dskloetd: same

PR #4112, opened 2023-12-22
GIX-2187: Test incorrect account identifier on ICRC wallet page

dskloetd:
# Motivation

When you go to a wallet page without an account identifier in the URL, this is invalid and you are redirected to the accounts page.
We want to instead show the main account for that universe.
Currently there are no tests covering this behavior at all.
So before changing the behavior I'm adding missing test coverage.

This PR adds the missing test coverage for ICRC wallet pages (ckBTC and ckETH).

# Changes

1. Add tests for navigation on missing, invalid and correct account identifiers.

# Tests

only tests

# Todos

not necessary


mstrasinskis (approved):
LGTM, thanks!

PR #4114, opened 2023-12-22
Use concise bigint notation

dskloetd:
# Motivation

`bigint`s can be written as `123n` but in many places we write `BigInt(123)`.
The former is more concise and also more accurate if the number actually requires a `bigint` for full accuracy.

# Changes

1. Replace `BigInt(???)` with `???n`.
2. Add underscore separators for numbers with more than 3 digits if they weren't there already.

This PR only does non-test files. There are a lot more cases in tests and it would make this PR too big.

# Tests

Still pass

# Todos

not necessary


mstrasinskis (approved):
Nice, thank you!

PR #4127, opened 2024-01-04
GIX-2187: Test incorrect account identifier for NNS wallet

dskloetd:
# Motivation

When you go to a wallet page without an account identifier in the URL, this is invalid and you are redirected to the accounts page.
We want to instead show the main account for that universe.
Currently there are no tests covering this behavior for the NNS wallet.
So before changing the behavior in all wallets I'm adding missing test coverage.

This PR adds the missing test coverage for NNS wallet.

# Changes

1. Add tests for navigation on missing, invalid and correct account identifiers.
2. Fix a typo in a related test description that has been copied around.

# Tests

pass

# Todos

not necessary


mstrasinskis (approved):
LGTM, thanks!

PR #4128, opened 2024-01-04
GIX-2187: Default to main account when accountIdentifier is missing

dskloetd:
# Motivation

People want to be able to link from external websites to the ckBTC wallet page, which has information about converting BTC to ckBTC.
This should work whether the user is signed-in or not.
When linking from an external website, the account identifier of the user is not known so it can't be included in the URL.
To make this work we want the wallet deep-link URL work even when know account identifier is specified.
In that case we default to the main account for the given wallet.
For consistency we do the same SNS and NSS wallets.

# Changes

1. Default to the main account when no account identifier is specified in the `IcrcWallet` component.
2. Default to the main account when no account identifier is specified in the `SnsWallet` component. 
3. Default to the main account when no account identifier is specified in the `NnsWallet` component. 

# Tests

Adjust unit tests for each wallet where no account identifier is specified to default to the main account instead of navigating to the accounts page.

Fix `"should render a spinner while loading"` in `NnsWallet.spec.ts`. It previously faked the loading state by not specifying an account identifier. Now we put it in loading state by pausing the API.

Remove `"new transaction action should be disabled while loading"` in `NnsWallet.spec.ts`. It did not simulate the loading state correctly and the test below it covers the same functionality.

# Todos

### DIFF
--- a/frontend/src/lib/pages/NnsWallet.svelte
+++ b/frontend/src/lib/pages/NnsWallet.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import type { Account } from "$lib/types/account";
   import { onDestroy, onMount, setContext } from "svelte";
   import { i18n } from "$lib/stores/i18n";
   import TestIdWrapper from "$lib/components/common/TestIdWrapper.svelte";
@@ -23,6 +24,7 @@
   import {
     accountName,
     findAccount,
+    mainAccount,
     isAccountHardwareWallet,
   } from "$lib/utils/accounts.utils";
   import {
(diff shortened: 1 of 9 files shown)


frontend/src/lib/pages/NnsWallet.svelte
```diff
@@ -112,15 +114,32 @@
...
 
+  const setSelectedAccount = ({
+    identifier,
+    accounts,
+  }: {
+    identifier: string | undefined | null;
+    accounts: Account[];
+  }) => {
+    // If there is no accountIdentifier specified, default to the main account.
+    const account = isNullish(identifier)
```
mstrasinskis, line 125: Just thinking out loud, is it possible to have this logic in an util function to avoid repetitive code.

mstrasinskis (approved):
Thank you!

frontend/src/lib/pages/NnsWallet.svelte
dskloetd: Done. Although the gains or a bit minor.

PR #4132, opened 2024-01-05
GIX-2187: Fix wallet title when balance is undefined

dskloetd:
# Motivation

The `WalletPageHeading` component has logic to show the account name and balance in the layout title when they scroll out of the viewport (i.e. no longer "intersect" with the viewport). The logic is such that if there is no balance, it always puts the balance in the layout title, even if it's not out of the viewport.
This seems backwards since then it renders as "0 undefined".
So I'm guessing it was intended the other way around, where if there is no balance, it is never (rather than always) rendered in the title.

Note that currently we never pass an `undefined` balance to the `WalletPageHeading` component, so I'm not even sure why this logic exists, but we plan to make the wallet page accessible when not signed in so then we do need `WalletPageHeading` to work with undefined balance.

# Changes

1. Invert the logic for showing the balance in the title to how it was (probably) intended.
2. Add a test.
3. Remove `waitFor` from the test. It's usually bad to use `waitFor` but in this case it can't accomplish anything because the value of `title` is determined before the call to `waitFor` and so can never change while we wait.

# Tests

Unit test added.

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/accounts/WalletPageHeading.svelte
+++ b/frontend/src/lib/components/accounts/WalletPageHeading.svelte
@@ -1,5 +1,10 @@
 <script lang="ts">
-  import { nonNullish, type TokenAmount, TokenAmountV2 } from "@dfinity/utils";
+  import {
+    isNullish,
+    nonNullish,
+    type TokenAmount,
+    TokenAmountV2,
+  } from "@dfinity/utils";
   import PageHeading from "../common/PageHeading.svelte";
   import { SkeletonText } from "@dfinity/gix-components";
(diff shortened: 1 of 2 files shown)


frontend/src/lib/components/accounts/WalletPageHeading.svelte
```diff
@@ -34,7 +39,7 @@
     layoutTitleStore.set({
       title: $i18n.wallet.title,
       header:
-        intersecting && nonNullish(balance)
+        intersecting || isNullish(balance)
```
mstrasinskis, line 42: Make sense 👍

mstrasinskis (approved):
Thanks!

PR #4133, opened 2024-01-05
GIX-2187: Allow CkBTCInfoCard without account

dskloetd:
# Motivation

We want to allow deep linking into the ckBTC wallet page so people can learn about BTC -> ckBTC conversion.
This should work even when users are not signed in.
The components on the wallet page currently assume that the user is always signed in or the wallet page wouldn't be shown. So this needs to change.

In this PR we support not being signed in for the `CkBTCInfoCard` component by allowing to not pass the `account` prop.

When signed in:

image

When not signed in:

image

A placeholder for the address QR code will be added later.

This change is not yet user visible because it's still not possible to see the wallet page without being signed in.

# Changes

1. Change the types of `account` and `identifier` to include `undefined.
2. Load the BTC address whenever the identifier is defined, instead of only on mount. This is for the case where the user signs in while on the wallet page.
3. Exclude the sentence with the block explorer link if there is no account.
4. Exclude the QR code of the BTC address if there is no account.
5. Instead of showing the BTC address, show a message to tell the user to sign in.
6. Exclude the "Check for incoming BTC" button if there is no account.

# Tests

1. Unit tests that various elements are not rendered.
2. Unit test that the sign in message is rendered.
3. Additional unit test to make sure that setting the account after rendering works, for the case where the user signs in on the wallet page.

# Todos

not yet

### DIFF
--- a/frontend/src/lib/components/accounts/CkBTCInfoCard.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCInfoCard.svelte
@@ -20,23 +20,22 @@
     BITCOIN_BLOCK_EXPLORER_MAINNET_URL,
     BITCOIN_BLOCK_EXPLORER_TESTNET_URL,
   } from "$lib/constants/bitcoin.constants";
-  import { onMount } from "svelte";
   import { loadBtcAddress } from "$lib/services/ckbtc-minter.services";
   import CkBTCUpdateBalanceButton from "$lib/components/accounts/CkBTCUpdateBalanceButton.svelte";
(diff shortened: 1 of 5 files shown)


mstrasinskis (approved):
LGTM, thanks!

mstrasinskis:
Is there a plan to add a Sign In button after this message?
image

dskloetd:
There will be a sign-in button earlier on the page. Like you can see here:
https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/wallet/?u=oq4ec-3maaa-aaaaa-qabeq-cai

PR #4141, opened 2024-01-08
Allow absent wallet address in WalletPageHeader

dskloetd:
# Motivation

We want to make deep links to wallet pages, render a logged-out version of the wallet page.
When the user is not logged in, there is no wallet address known.
So we need to be able to render the `WalletPageHeader` component without wallet address.

This PR does not change anything user visible because it's not yet possible to see the wallet page while not logged in.

# Changes

Allow not passing a wallet address to the `WalletPageHeader` component. Then don't render the `IdentifierHash` component.

# Tests

1. Merge the existing tests for the wallet address and universe name.
2. Add an extra test where the wallet address is absent.

# Todos

not yet


mstrasinskis (approved):
Thanks!

PR #4142, opened 2024-01-08
Sns type filter store init

mstrasinskis:
# Motivation

In order to filter SNS proposals by types, we prepare the data needed for the `types` filtering in the `SnsProposalsFilter` component.

Part of: https://github.com/dfinity/nns-dapp/pull/3801

# Changes

- util `generateSnsProposalTypeFilterData`. 
  - ignore `ALL_SNS_PROPOSAL_TYPES_NS_FUNCTION_ID`
  - replace all generic ns functions with a single filter item (`native` - same for all snses; `generic` - unique per sns). We combined all generic filters into a single filter entry. 
- Update types filter store in `loadSnsFilters`.

# Tests

- `generateSnsProposalTypeFilterData` util
- `loadSnsFilters` should update the store

# Todos

Not yet


frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
```diff
@@ -10,15 +10,33 @@
...
+
+  let modal: "types" | "rewards" | "status" | undefined = undefined;
 
   let rootCanisterId: Principal;
   $: rootCanisterId = $selectedUniverseIdStore;
   let filtersStore: ProjectFiltersStoreData | undefined;
   $: filtersStore = $snsFiltersStore[rootCanisterId.toText()];
 
-  const openFilters = (filtersModal: "topics" | "rewards" | "status") => {
+  $: if (nonNullish(nsFunctions)) {
```
dskloetd, line 27: Why is it this component's job to put the `nsFunctions` in a global store?
Shouldn't that be the job of the service that loads the data (or where do these `nsFunctions` come from)?
It just seems strange to do this every time the component is mounted, or not at all if the component is not mounted, while `snsFiltersStore` is a global store.

frontend/src/lib/i18n/en.json
```diff
@@ -949,6 +949,9 @@
     "select": "Select",
     "universe_logo": "$universe logo"
   },
+  "sns_types": {
+    "sns_specific": "All SNS-specific proposals"
```
dskloetd, line 953: I had to think for a moment what this meant.
First I thought it meant proposals specific to SNSes rather than NNS.
But I guess it means proposal types specific to one particular SNS.
Like proposals that only OpenChat or only Kinic has, right?
Maybe we should put the name of the SNS as a placeholder, so it would say
"All Open Chat specific proposals" for example.
What do you think?

frontend/src/lib/utils/sns-proposals.utils.ts
```diff
@@ -472,6 +479,49 @@ export const getUniversalProposalStatus = (
...
+// Generate new "types" filter data, but preserve the checked state of the current filter state
+// `nsFunctions` can be changed on the backend, and to display recently created proposal types, new entries should be preselected.
+export const generateSnsProposalTypesFilterData = ({
+  nsFunctions,
+  typesFilterState,
+}: {
+  nsFunctions: SnsNervousSystemFunction[];
+  typesFilterState: Filter<SnsProposalTypeFilterId>[];
+}): Filter<SnsProposalTypeFilterId>[] => {
+  // transfer only unchecked entries to preselect new items that are not in the current filter state
```
dskloetd, line 491: I understood this comment only after reading the test.
How about
```
// New proposal types are checked by default so only keep unchecked those types that were already unchecked.
```

frontend/src/lib/utils/sns-proposals.utils.ts
```diff
@@ -472,6 +479,49 @@ export const getUniversalProposalStatus = (
...
+  const getCheckedState = (id: string) =>
+    typesFilterState.find(({ id: stateId }) => id === stateId)?.checked !==
+    false;
+  const nativeNsFunctionEntries: Filter<SnsProposalTypeFilterId>[] = nsFunctions
+    .filter(isNativeNervousSystemFunction)
+    // ignore { 0n: "All Topics"}
```
dskloetd, line 497: What does it mean to ignore this entry? Is this user not allow to filter based on this entry?
And why do we want to ignore it?

frontend/src/lib/utils/sns-proposals.utils.ts
```diff
@@ -472,6 +479,49 @@ export const getUniversalProposalStatus = (
...
+    }))
+    .map(({ id, name }) => ({
+      id,
+      value: id,
+      name: name,
+      // transfer only unchecked entries to preselect new items that are not in the current filter state
+      checked: getCheckedState(id),
+    }));
+  const genericNsFunctionEntries: Filter<SnsProposalTypeFilterId>[] =
+    nsFunctions.filter(isGenericNervousSystemFunction).length > 0
```
dskloetd, line 511: If you use `some` instead of `filter`, it will stop as soon as it finds one and you don't need to check the `length`.
```
nsFunctions.some(isGenericNervousSystemFunction)
```

frontend/src/lib/utils/sns-proposals.utils.ts
```diff
@@ -472,6 +479,49 @@ export const getUniversalProposalStatus = (
...
+    .map((entry) => ({
+      ...entry,
+      id: `${entry.id}`,
+    }))
+    .map(({ id, name }) => ({
+      id,
+      value: id,
+      name: name,
+      // transfer only unchecked entries to preselect new items that are not in the current filter state
+      checked: getCheckedState(id),
```
dskloetd, line 508: I'm wondering if the checked state should really be in the same store.
I would have expected a separate store with the checked types and then a derived store that mixes them.
Then when you update the store of existing types, you don't need to mess what is selected and the derived store will just take care of it.

frontend/src/tests/lib/components/sns-proposals/SnsProposalsFilters.spec.ts
```diff
@@ -29,4 +46,24 @@ describe("SnsProposalsFilters", () => {
...
   });
+
+  it("should update snsFiltersStore with types", async () => {
+    const setTypesSpy = vi.spyOn(snsFiltersStore, "setTypes");
+    expect(setTypesSpy).toBeCalledTimes(0);
+
+    renderComponent();
+
+    expect(setTypesSpy).toBeCalledTimes(1);
+    expect(setTypesSpy).toBeCalledWith({
```
dskloetd, line 57: Why not just check the contents of the store, rather than spying on the setter?

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
```diff
@@ -638,15 +647,172 @@ describe("sns-proposals utils", () => {
...
+      checked: true,
+    };
 
-    it("should not break when no percentage provided", () => {
-      expect(fromPercentageBasisPoints(undefined)).toBe(undefined);
+    describe("generateSnsProposalTypeFilterData", () => {
+      it("should use nsFunctions to create filter entries", () => {
+        const nsFunctions: SnsNervousSystemFunction[] = [
+          nativeNervousSystemFunctionMock,
+          nativeNervousSystemFunctionMock,
```
dskloetd, line 668: Normally we wouldn't have the same entry twice in this array, right?
How about having 2 different instances with different IDs or something?

frontend/src/lib/utils/sns-proposals.utils.ts
```diff
@@ -472,6 +479,49 @@ export const getUniversalProposalStatus = (
...
+    .filter(isNativeNervousSystemFunction)
+    // ignore { 0n: "All Topics"}
+    .filter(({ id }) => id !== ALL_SNS_PROPOSAL_TYPES_NS_FUNCTION_ID)
+    .map((entry) => ({
+      ...entry,
+      id: `${entry.id}`,
+    }))
+    .map(({ id, name }) => ({
+      id,
+      value: id,
```
dskloetd, line 505: Why have 2 fields ('id' and `value`) with the same value?

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
```diff
@@ -28,6 +46,9 @@
 </script>
 
 <FiltersWrapper>
+  {#if $ENABLE_SNS_TYPES_FILTER}
+    <!-- TBD: Type filter button -->
+  {/if}
```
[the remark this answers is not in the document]
mstrasinskis: Removed.

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
```diff
@@ -10,15 +10,33 @@
...
+  $: if (nonNullish(nsFunctions)) {
+    // Always update type filters in case of backend changes
+    snsFiltersStore.setTypes({
+      rootCanisterId,
+      types: generateSnsProposalTypesFilterData({
+        nsFunctions,
+        typesFilterState: filtersStore?.types ?? [],
+      }),
+    });
+    filtersStore = $snsFiltersStore[rootCanisterId.toText()];
```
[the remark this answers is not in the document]
mstrasinskis: Unfortunately w/o this line the store won't be updated.

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
```diff
@@ -10,15 +10,33 @@
...
 
   let rootCanisterId: Principal;
   $: rootCanisterId = $selectedUniverseIdStore;
   let filtersStore: ProjectFiltersStoreData | undefined;
   $: filtersStore = $snsFiltersStore[rootCanisterId.toText()];
 
-  const openFilters = (filtersModal: "topics" | "rewards" | "status") => {
+  $: if (nonNullish(nsFunctions)) {
+    // Always update type filters in case of backend changes
+    snsFiltersStore.setTypes({
```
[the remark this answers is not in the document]
mstrasinskis: Initially it's populated in `loadSnsFilters()`. This function provides `rewardStatus` and `decisionStatus` entries (enum based). To generate `types` some `nsFunctions` needs to be loaded, that's why it's not possible to initiate `types` in `loadSnsFilters` function.

frontend/src/lib/utils/sns-proposals.utils.ts
mstrasinskis: Good point, updated 👍

frontend/src/lib/i18n/en.json
mstrasinskis: Thanks, I like the idea.

frontend/src/lib/utils/sns-proposals.utils.ts
mstrasinskis: Thanks, updated.

frontend/src/lib/utils/sns-proposals.utils.ts
mstrasinskis: Because having "All Topics" in the filter list will break the current selection logic.

frontend/src/tests/lib/components/sns-proposals/SnsProposalsFilters.spec.ts
mstrasinskis: Good point, the test was moved to `SnsProposals`.

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
mstrasinskis: Moved to `SnsProposals`, next to `loadSnsFilters`.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
mstrasinskis: Thank you for pointing that out; I'll keep it in mind for the future.

frontend/src/lib/utils/sns-proposals.utils.ts
mstrasinskis: By the definition of `Filter` generic https://github.com/dfinity/nns-dapp/blob/f1be6c051282bb4d5ae90babbf05ada54587ac8b/frontend/src/lib/types/filters.ts#L9, `id` is a `string`. I would avoid using the raw 'nsFunction' ID as a 'bigint' value to prevent extra serialization on every LocalStorage I/O, as the entire structure is stored there.

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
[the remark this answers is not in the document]
dskloetd: Why not? Isn't that what the `$:` syntax is for?

frontend/src/lib/utils/sns-proposals.utils.ts
dskloetd: OK, not for this PR, but could we just remove either `id` or `value` and just always use one of those fields? Or is it conceivable that they would have a different value in some circumstances?

frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -31,6 +37,15 @@
         loadSnsNervousSystemFunctions(selectedProjectCanisterId),
         loadSnsFilters(selectedProjectCanisterId),
       ]);
+
+      if (isNullish($nsFunctionsStore)) {
+        throw new Error("no nsFunctions");
```
dskloetd, line 42: Is it OK to throw here? Do we have reason to assume this can't happen?

frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -31,6 +37,15 @@
...
+      }
+      updateSnsTypeFilter({
+        rootCanisterId: selectedProjectCanisterId,
+        nsFunctions: $nsFunctionsStore,
+        snsName: $snsProjectSelectedStore?.summary?.metadata?.name ?? "",
```
dskloetd, line 47: We really don't want this to be empty, right?
Can we also throw if there is no name?
Or wait with updating until there is a name?

frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -31,6 +37,15 @@
         loadSnsNervousSystemFunctions(selectedProjectCanisterId),
         loadSnsFilters(selectedProjectCanisterId),
       ]);
+
+      if (isNullish($nsFunctionsStore)) {
+        throw new Error("no nsFunctions");
+      }
+      updateSnsTypeFilter({
```
lmuntaner, line 44: How about using the `loadSnsFilters` to do so?

`onSnsProjectChanged` and `loadSnsFilters` could take the `nsFunctionsStore` as parameter to make sure it's called every time the store changes.

Then, inside `loadSnsFilters` we have the logic of merging the filters and setting the parts of the store according to `nsFunctionsStore`.

```
loadSnsFilters({ nsFunctions, rootCanisterId: selectedProjectCanisterId });
```

Then inside the `loadSnsFilters` we can use `generateSnsProposalTypesFilterData`.

frontend/src/lib/services/sns-filters.services.ts
```diff
@@ -78,3 +80,25 @@ export const loadSnsFilters = async (rootCanisterId: Principal) => {
...
+  nsFunctions,
+  snsName,
+}: {
+  rootCanisterId: Principal;
+  nsFunctions: SnsNervousSystemFunction[];
+  snsName: string;
+}) => {
+  const typesFilterState =
+    get(snsFiltersStore)?.[rootCanisterId.toText()]?.types ?? [];
+  const types = generateSnsProposalTypesFilterData({
```
dskloetd, line 95: `typesFilterState` and `types` are basically the same thing.
Just `typesFilterState` is the old state and `types` is the new/merged state, right?
Maybe you can reflect that in the names?

frontend/src/tests/lib/pages/SnsProposals.spec.ts
```diff
@@ -103,6 +104,25 @@ describe("SnsProposals", () => {
...
       });
 
+      it("should init types filter", async () => {
+        const getFiltersStoreData = () =>
+          get(snsFiltersStore)[rootCanisterId.toText()];
+
+        expect(getFiltersStoreData()?.types).toEqual(undefined);
+        render(SnsProposals);
+
+        await waitFor(() =>
```
dskloetd, line 114: `waitFor` is brittle. Try not to use it.
Can you do without?
If not, can you use `runResolvedPromises` instead?

frontend/src/tests/lib/pages/SnsProposals.spec.ts
```diff
@@ -103,6 +104,25 @@ describe("SnsProposals", () => {
...
+          expect(getFiltersStoreData()?.types).toEqual([
+            {
+              checked: true,
+              id: "3",
+              name: "test_function",
```
dskloetd, line 119: Use the `functionName` constant defined above?

frontend/src/tests/lib/services/sns-filters.services.spec.ts
```diff
@@ -52,8 +58,34 @@ describe("sns-filters services", () => {
...
+        rootCanisterId: mockPrincipal,
+        nsFunctions: [nativeNervousSystemFunctionMock],
+        snsName: "sns-name",
+      });
+
+      expect(getFiltersStoreData().types).toEqual([
+        {
+          checked: true,
+          id: "1",
+          name: "Motion",
```
dskloetd, line 85: Where does `"Motion"` come from? I don't see it anywhere else in this file.

frontend/src/lib/pages/SnsProposals.svelte
mstrasinskis: It's only about the place of the logic and this wouldn't make the code more readable. There is a `if` the `loadSnsFilters`
```
  const filtersProjectStoreData = get(snsFiltersStore)[rootCanisterId.toText()];
  if (nonNullish(filtersProjectStoreData)) {
    return;
  }
```
which blocks store changes if there is already some data, otherwise it will be reset. Perhaps I misunderstood something.

frontend/src/lib/components/sns-proposals/SnsProposalsFilters.svelte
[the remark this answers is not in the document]
mstrasinskis: Yes it is, but after running logging it, I just saw that it wasn't updated. Anyway this is already an outdated part and don't exist anymore.

frontend/src/lib/pages/SnsProposals.svelte
mstrasinskis: It is here just to provide more info which `loadSnsNervousSystemFunctions` fail. I don't have a strong opinion here. What would you suggest?

frontend/src/lib/pages/SnsProposals.svelte
lmuntaner: This block should be removed, yes indeed.

I have the feeling that we are leaking implementation details into how filters work and are set setting it like this.

The component should care only about setting the filters, not on how the filters are set.

That's why I believe it's nicer to have the logic in `loadSnsFilters`.

Wdyt @dskloetd ?

dskloetd: I think what @lmuntaner says makes sense. And having this logic here in the component doesn't feel right.

mstrasinskis: @lmuntaner, do you mean something like this: https://github.com/dfinity/nns-dapp/pull/4142/files#diff-0bf688db9e3080ae8ff14ae04ebbcfd28f34be71e6fdac4efd17d4890a2d2f6eR92

frontend/src/lib/services/sns-filters.services.ts
mstrasinskis: Thanks, renamed.

frontend/src/lib/pages/SnsProposals.svelte
lmuntaner: Yes!

mstrasinskis: Thank you, the tests have been updated accordingly.

frontend/src/lib/utils/sns-proposals.utils.ts
mstrasinskis: I guess, since the `value` is generic `id` is needed for entries comparison - https://github.com/dfinity/nns-dapp/blob/38dd1b83a0b4980c4074d8a583df5346bf7b022f/frontend/src/lib/modals/common/FilterModal.svelte#L52.

frontend/src/tests/lib/services/sns-filters.services.spec.ts
mstrasinskis: From the `nativeNervousSystemFunctionMock` https://github.com/dfinity/nns-dapp/blob/38dd1b83a0b4980c4074d8a583df5346bf7b022f/frontend/src/tests/mocks/sns-functions.mock.ts#L21

frontend/src/tests/lib/pages/SnsProposals.spec.ts
mstrasinskis: Thanks, `runResolvedPromises` is enough.

frontend/src/lib/utils/sns-proposals.utils.ts
dskloetd: Ahhhh! I didn't realize `value` is generic. Now it makes sense. Thanks.

frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -10,31 +10,51 @@
...
+      await loadSnsFilters({
+        rootCanisterId,
+        nsFunctions: $nsFunctionsStore,
+        snsName,
+      });
     }
   };
 
-  $: onSnsProjectChanged($snsOnlyProjectStore);
+  $: if (nonNullish($snsProjectSelectedStore?.summary?.metadata?.name)) {
```
dskloetd, line 52: This `if` can also go inside `onSnsProjectChanged`, right?
(As long as you pass the name as a parameter so the function will be called again if it changes.)

frontend/src/lib/pages/SnsProposals.svelte
```diff
@@ -10,31 +10,51 @@
...
-      ]);
+  const onSnsProjectChanged = async ({
+    rootCanisterId,
+    snsName,
+  }: {
+    rootCanisterId: Principal | undefined;
+    snsName: string;
+  }) => {
+    currentProjectCanisterId = rootCanisterId;
+    if (nonNullish(rootCanisterId)) {
```
dskloetd, line 36: If you do
```
if (isNullish(rootCanisterId)) {
  return;
}
```
then the rest of the function can be less indented.

frontend/src/lib/services/sns-filters.services.ts
```diff
@@ -2,79 +2,103 @@ import { i18n } from "$lib/stores/i18n";
...
-    types: [],
-    rewardStatus: [],
-    decisionStatus: [],
-  };
 
-  snsFiltersStore.setTypes({
-    rootCanisterId,
-    types: defaultFiltersProjectData.types,
-  });
+  // Set initial filters state
```
dskloetd, line 27: This function is already quite long. How about making the initial state block a separate function?

frontend/src/tests/lib/services/sns-filters.services.spec.ts
dskloetd: I see you made it explicit now. Thanks!

frontend/src/lib/pages/SnsProposals.svelte
mstrasinskis: Good point, updated.

frontend/src/lib/services/sns-filters.services.ts
mstrasinskis: Agree, has been splitted into 3.

frontend/src/lib/pages/SnsProposals.svelte
mstrasinskis: Done.

dskloetd (approved):
Thanks!

frontend/src/lib/pages/SnsProposals.svelte
mstrasinskis: Outdated.

PR #4149, opened 2024-01-09
Revert "Bump SvelteKit v2 (#4057)"

dskloetd:
This reverts commit a1598bc03bcebdf8460fcda3ccb67d98a14fd6c9.

# Motivation

It breaks NNS dapp on iOS 15.X.

# Changes

1. Ran `git revert a1598bc03bcebdf8460fcda3ccb67d98a14fd6c9`.
2. Added an entry to the changelog.

# Tests

Installed on DevEnv and verified with an iPad running iOS 15.8 that it fixes the issue.

# Todos

- [x] Add entry to changelog (if necessary).

PR #4150, opened 2024-01-09
Fix import lint

dskloetd:
# Motivation

Somehow this is only caught after reverting SvelteKit. https://github.com/dfinity/nns-dapp/pull/4149

# Changes

Fix import lint.

# Tests

no

# Todos

no


mstrasinskis (approved):
Thanks!

PR #4151, opened 2024-01-09
Fix balances observers

dskloetd:
# Motivation

We want to be able to render wallet pages to signed-out users.
When the user is not signed in, there is no balance to observer so the observer should be separated from the content that can be displayed to signed-out users. So the content should not be passed as a slot of the observer, but should be a sibling of the observer.
After I made this change, I tested if the observer (still) worked, and I realized that the observer was actually already broken before.
It was broken by https://github.com/dfinity/nns-dapp/pull/3995 which changed `Account.balanceE8s` to `Account.balanceUlps` while the observers kept setting the `balanceE8s` field. This was not caught by the test because the test also checked that the `balanceE8s` field was set.
So in addition to fixing the usage of the field, I wanted to add a test that would have caught the issue.
And while adding the test I ran into what seems like a bug in Svelte where it would not update values render inside a slot.
Since my plan was to move the content outside the slot of the observer, I decided not to investigate further and just do the fix and the move in the same PR.

# Changes

1. In `IcrcBalancesObserver.svelte` and `SnsBalancesObserver.svelte` set `balanceUlps` instead of `balanceE8s`.
2. In `IcrcWalletPage.svelte` and `SnsWallet.svelte` move the content outside the respective Observer slot.

# Tests

1. Update the Observer tests to check the `balanceUlps` field instead of the `balanceE8s` field.
2. Add wallet tests to check that the balances passed from the observer callback are actually rendered.

# Todos

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcWalletPage.svelte
+++ b/frontend/src/lib/components/accounts/IcrcWalletPage.svelte
@@ -136,35 +136,34 @@
           universeId={selectedUniverseId}
           accounts={[$selectedAccountStore.account]}
           reload={reloadOnlyAccountFromStore}
+        />
+        <WalletPageHeader
+          universe={$selectedUniverseStore}
+          walletAddress={$selectedAccountStore.account.identifier}
+        />
+        <WalletPageHeading
+          accountName={$selectedAccountStore.account.name ??
+            $i18n.accounts.main}
+          balance={TokenAmountV2.fromUlps({
+            amount: $selectedAccountStore.account.balanceUlps,
+            token,
+          })}
         >
-          <WalletPageHeader
-            universe={$selectedUniverseStore}
-            walletAddress={$selectedAccountStore.account.identifier}
-          />
-          <WalletPageHeading
-            accountName={$selectedAccountStore.account.name ??
-              $i18n.accounts.main}
-            balance={TokenAmountV2.fromUlps({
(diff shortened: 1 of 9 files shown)


frontend/src/lib/components/accounts/IcrcWalletPage.svelte
```diff
@@ -136,35 +136,34 @@
           universeId={selectedUniverseId}
           accounts={[$selectedAccountStore.account]}
           reload={reloadOnlyAccountFromStore}
+        />
```
dskloetd, line 139: This looks like a large diff but it's only moving what was in the slot of `IcrcBalancesObserver` to below the `IcrcBalancesObserver`.

frontend/src/lib/pages/SnsWallet.svelte
```diff
@@ -155,28 +155,27 @@
...
-
-            <Separator spacing="none" />
-
-            <SnsTransactionsList
-              rootCanisterId={$snsOnlyProjectStore}
-              account={$selectedAccountStore.account}
-              {token}
-            />
-          </SnsBalancesObserver>
+          />
```
dskloetd, line 158: This looks like a large diff but it's only moving what was in the slot of `SnsBalancesObserver ` to below the `SnsBalancesObserver `.

peterpeterparker (approved):
LGTM, thx

PR #4157, opened 2024-01-10
test: Use the new migration test in CI

bitdivine:
# Motivation
The migration to stable memory needs a more tests than our downgrade-upgrade tests provide.  In previous PR,  a copy of the downgrade-upgrade test was made and refactored to allow it to test any migration rather than just to prod and back.  This test was used locally during stable memory development but not (yet) exercised in CI.  Let's test the simplest case first - just upgrade to self - to gain confidence that the test works as expected.

Note: This does NOT yet replace the downgrade-upgrade test.  Given  that there have been many changes to the test it would be nice to know that it works reliably in the simplest setting possible before using it for more complicated cases.

# Changes
- Exercise the new migration test, just for a simple upgrade-to-self with classic stable-memory-on-heap storage.

# Tests
- See CI

# Todos

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -118,6 +118,39 @@ jobs:
         run: ./scripts/nns-dapp/downgrade-upgrade-test -w nns-dapp_test.wasm.gz
       - name: Count upgrade cycles
         run: scripts/nns-dapp/estimate-upgrade-cycles | tee -a $GITHUB_STEP_SUMMARY
+  test-upgrade-map:
+    needs: build
+    runs-on: ubuntu-20.04
+    timeout-minutes: 40
+    steps:
+      - name: Checkout nns-dapp
+        uses: actions/checkout@v3
+      - name: Get nns-dapp_test
+        uses: actions/download-artifact@v3
+        with:
+          name: out
+          path: out
+      - name: Install ic-wasm
+        uses: ./.github/actions/install_ic_wasm
+      - name: Install tools
+        run: |
+          sudo apt-get update -yy && sudo apt-get install -yy moreutils && command -v sponge
+          DFX_VERSION="$(jq -r .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
+          cargo binstall --no-confirm "idl2json_cli@$(jq -r .defaults.build.config.IDL2JSON_VERSION dfx.json)" && idl2json --version
+      - name: Start dfx
+        run: dfx start --clean --background &>test-upgrade-map-dfx.log
+      - name: Put assets in the default location
+        run: |
+          ls
+          ls out
+      - name: Upgrade to self, keeping the storage schema as map
+        run: ./scripts/nns-dapp/migration-test --schema1 Map --schema2 Map --accounts 10000
+      - name: Upload dfx logs
+        if: failure()
+        uses: actions/upload-artifact@v3
+        with:
+          name: test-upgrade-map-dfx.log
+          path: test-upgrade-map-dfx.log
   test-test-account-api:
     needs: build
     runs-on: ubuntu-20.04
--- a/scripts/nns-dapp/migration-test.canister
+++ b/scripts/nns-dapp/migration-test.canister
@@ -16,7 +16,7 @@ migration_test_canisters_help() {
 
 	A list of functions:
 	EOF
-  declare -F | awk '{print "        " $3}' | grep -vw "$FUNCNAME"
+  declare -F | awk '{print "        " $3}' | grep -vw "${FUNCNAME[0]}"
 }
 
 # Verifies that the nns-dapp canister is healthy
--- a/scripts/nns-dapp/migration-test.canister
+++ b/scripts/nns-dapp/migration-test.canister
@@ -75,7 +75,7 @@ check_num_transactions() {
 # Wait for the migration countdown to complete.
 wait_for_migration() {
   local migration_countdown
-  for ((i = 0; i < 120; i++)); do
+  for ((i = 0; i < 1200; i++)); do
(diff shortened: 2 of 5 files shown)


scripts/nns-dapp/migration-test.canister
```diff
@@ -75,7 +75,7 @@ check_num_transactions() {
 # Wait for the migration countdown to complete.
 wait_for_migration() {
   local migration_countdown
-  for ((i = 0; i < 120; i++)); do
+  for ((i = 0; i < 1200; i++)); do
```
bitdivine, line 78: Testing locally, for larger migrations I needed more time.

scripts/nns-dapp/migration-test.canister
```diff
@@ -96,14 +96,15 @@ wait_for_migration() {
 # - Creates the canister if it doesn't exist
 # - Installs the canister even if the wasm is the same
```
bitdivine, line 97: Minor cleanup and use --upgrade-unchanged, needed for upgrade to self when all that has changed is the arguments.  The arguments include the desired schema so this is an important case!

scripts/nns-dapp/migration-test.canister
lmuntaner: Will you revert this change then?

.github/workflows/build.yml
```diff
@@ -118,6 +118,39 @@ jobs:
...
+          DFX_VERSION="$(jq -r .dfx dfx.json)" sh -ci "$(curl -fsSL https://sdk.dfinity.org/install.sh)"
+          cargo binstall --no-confirm "idl2json_cli@$(jq -r .defaults.build.config.IDL2JSON_VERSION dfx.json)" && idl2json --version
+      - name: Start dfx
+        run: dfx start --clean --background &>test-upgrade-map-dfx.log
+      - name: Put assets in the default location
+        run: |
+          ls
+          ls out
+      - name: Upgrade to self, keeping the storage schema as map
+        run: ./scripts/nns-dapp/migration-test --schema1 Map --schema2 Map --accounts 10000
```
lmuntaner, line 147: Could you add a comment in the migration-test.canister explaining which arguments it supports?

lmuntaner (approved):
LGTM, just a couple of comments, but already approving.

scripts/nns-dapp/migration-test.canister
bitdivine: I don't plan to.  Rather I plan to add stable memory migration tests to CI.  Then CI will need the extra time.  Stable memory is slow!

.github/workflows/build.yml
bitdivine: Migration-test.canister isn't meant to be called directly.  `migration-test` supports `--help` which includes a description and arguments:
```
$ ./scripts/nns-dapp/migration-test --help
Tests:
  - Populating a given schema with large scale data
  - Migrating from one Wasm+schema to another Wasm+schema and back.

Note:
  - This test is primarily experimental at this time.
  - This test is based on the 'downgrade-upgrade' test.
  - Some or all of these changes may be migrated into the downgrade-upgrade test.

Needs:
  - dfx
  - ic-wasm
  - idl2json
  - wasm files for the two deployments
  - candid argument files
  - a running replica

usage: migration-test [OPTIONS]

OPTIONS:
        
	-w --wasm1:                  A test build of the first wasm. May be a _test.wasm.gz file or a named release. [default:out/nns-dapp_test.wasm.gz]
	-W --wasm2:                  A test build of the second wasm. May be a _test.wasm.gz file or a named release. [default:out/nns-dapp_test.wasm.gz]
	-a --args1:                  A file containing the first init arguments. [default:out/nns-dapp-arg-local.did]
	-A --args2:                  A file containing the second init arguments. [default:out/nns-dapp-arg-local.did]
	-s --schema1:                The schema for the first deployment. Overrides the value in the argument file. Options: Map AccountsInStableMemory
	-S --schema2:                The schema for the second deployment. Overrides the value in the argument file. Options: Map AccountsInStableMemory
	-a --accounts:               Accounts will be created until there are at least this many. [default:220000]
	-c --chunk:                  The accounts are created in chunks of this size. [default:10000]

        -? --help  :  usage

        --verbose  :  show debug info
```

bitdivine: There is also support for `--help` in `migration-test.canister` but it says, pretty much: "source this":
```
$ ./scripts/nns-dapp/migration-test.canister --help
A library of methods for getting and manipulating the nns-canister state.

Usage:
- Source this file from your code.
- You can also call this file directly.  This may be useful for
  ad-hoc runs but is not intended for production use.

      Usage: scripts/nns-dapp/migration-test.canister <function> [arguments..]
      e.g.:  scripts/nns-dapp/migration-test.canister verify_healthy

      A list of functions:
        assert_schema_is
        check_num_transactions
        check_state_size
        get_schema_name
        get_transactions_count
        get_upgrade_invariant_stats
        get_upgrade_invariants
        install_nnsdapp
        sample_toy_accounts
        schema_index
        schema_name
        upgrade_nnsdapp
        verify_healthy
        wait_for_migration
```

bitdivine: Does this provide the sort of information that you were looking for?

bitdivine: Maybe it would help if there were a comment at the top of these files saying: "For usage information about this script, please run `--help`"?  That would steer readers towards the information they are looking for.

lmuntaner: That is enough, thanks for the info!

bitdivine: I moved the help text for migration-test.* into functions and put the functions at the top of the file.  I think that covers both someone reading the code and command line --help. I will move ahead with merging but am happy to come back to this if needed.

PR #4159, opened 2024-01-10
GIX-1984: Redirect from /accounts to /tokens accordingly

lmuntaner:
# Motivation

When the new tokens page is enabled, the `/accounts` page is only accessible for NNS universe.

# Changes

* Redirect to `/tokens` from `/accounts` if the universe is not NNS.

# Tests

* Add tests for the redirection logic.

# Todos

Not yet necesary.

### DIFF
--- a/frontend/src/lib/routes/Accounts.svelte
+++ b/frontend/src/lib/routes/Accounts.svelte
@@ -35,8 +35,17 @@
   import IcrcTokenAccountsFooter from "$lib/components/accounts/IcrcTokenAccountsFooter.svelte";
   import IcrcTokenAccountsModals from "$lib/modals/accounts/IcrcTokenAccountsModals.svelte";
   import { ENABLE_MY_TOKENS } from "$lib/stores/feature-flags.store";
+  import { goto } from "$app/navigation";
+  import { AppPath } from "$lib/constants/routes.constants";
 
   // TODO: This component is mounted twice. Understand why and fix it.
+
+  $: {
+    // When using the new Tokens table, the Accounts page is enabled only for NNS.
+    if ($ENABLE_MY_TOKENS && !$isNnsUniverseStore) {
+      goto(AppPath.Tokens, { replaceState: true });
+    }
+  }
 
   // Selected project ID on mount is excluded from load accounts balances. See documentation.
   let selectedUniverseId = $selectedUniverseIdStore;
--- a/frontend/src/tests/lib/routes/Accounts.spec.ts
+++ b/frontend/src/tests/lib/routes/Accounts.spec.ts
@@ -868,9 +868,11 @@ describe("Accounts", () => {
       },
     });
 
+    expect(get(pageStore)?.path).toEqual(AppPath.Accounts);
+
     render(Accounts);
 
-    await waitFor(() => expect(get(pageStore)?.path).toEqual(AppPath.Tokens));
+    expect(get(pageStore)?.path).toEqual(AppPath.Tokens);
   });
 
   it("should not redirect to Tokens page when tokens page is not enabled and universe is not NNS", async () => {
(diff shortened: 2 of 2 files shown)


lmuntaner:
* `main`
  * **PR #4159** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4159?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4159?utm_source=stack-comment).

frontend/src/lib/routes/Accounts.svelte
```diff
@@ -35,9 +35,18 @@
...
   import { ENABLE_MY_TOKENS } from "$lib/stores/feature-flags.store";
+  import { goto } from "$app/navigation";
+  import { AppPath } from "$lib/constants/routes.constants";
 
   // TODO: This component is mounted twice. Understand why and fix it.
 
+  $: {
+    // When using the new Tokens table, the Accounts page is enabled only for NNS.
+    if ($ENABLE_MY_TOKENS && !$isNnsUniverseStore) {
+      goto(AppPath.Tokens, { replaceState: true });
```
lmuntaner, line 46: I replace the state so that the back button doesn't start a loop of redirections.

lmuntaner:
The only way to test this is by changing the URL.

Video of the UX.

            <a href="https://app.graphite.dev/media/video/W99H6e2WWjPgtes6VOJI/c55dcc13-15fa-4d01-8384-fd0107609a57.mov">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/W99H6e2WWjPgtes6VOJI/c55dcc13-15fa-4d01-8384-fd0107609a57.mov">Screen Recording 2024-01-10 at 11.11.43.mov</video>

lmuntaner:
@dskloetd please review

frontend/src/lib/routes/Accounts.svelte
dskloetd: Should we actually do that on all redirects?

frontend/src/tests/lib/routes/Accounts.spec.ts
```diff
@@ -855,4 +857,67 @@ describe("Accounts", () => {
...
+    page.mock({
+      data: {
+        universe: CKBTC_UNIVERSE_CANISTER_ID.toText(),
+        routeId: AppPath.Accounts,
+      },
+    });
+
+    render(Accounts);
+
+    await waitFor(() => expect(get(pageStore)?.path).toEqual(AppPath.Tokens));
```
dskloetd, line 873: You don't need `waitFor`. I checked out your branch and this just works:
```
    expect(get(pageStore)?.path).toEqual(AppPath.Accounts);

    render(Accounts);

    expect(get(pageStore)?.path).toEqual(AppPath.Tokens);
```

`waitFor` is really brittle. Please always try without `waitFor`, using `runResolvedPromises()` if you have to.

dskloetd, line 873: Can you check that the value of the `pageStore` before rendering so that it's clear that it changed?

frontend/src/lib/routes/Accounts.svelte
lmuntaner: I'm not sure on all, but in the error ones is probably a good idea, yes.

frontend/src/tests/lib/routes/Accounts.spec.ts
lmuntaner: I removed it from the others tests as well. I checked that the expect of the redirection fails when expecting `AppPath.Accounts`. Therefore, it's safe to expect it when there is no redirection.

lmuntaner: Done

lmuntaner:

frontend/src/tests/lib/routes/Accounts.spec.ts
dskloetd: Thanks! Yes, the other tests were my main concern. I found the double `waitFor`s unreadable. But if the `waitFor` would have been needed here, the other tests can't do without. So I just tested this one.

dskloetd (approved):

PR #4161, opened 2024-01-10
Update nns-dapp to dfx 0.15.3

dskloetd:
# Motivation

dfx 0.14.4 is quite old but we had been stuck because newer versions did work remotely from DevEnv.
dfx 0.15.3 finally does support connecting remotely so it's time to upgrade.

After upgrading to dfx 0.15.3, the installation of canisters started timing out when attempted immediately after starting the snapshot. I started a thread with the SDK team to find out why but in the meantime I added retrying logic to work around this.

# Changes

1. Change the dfx version in `dfx.json`.
2. Change the `SNSDEMO` release to one created with dfx 0.15.3.
3. Add a loop+sleep to work around the canister installation timeout.

# Tests

CI passes.

# Todos


.github/actions/start_dfx_snapshot/action.yaml
```diff
@@ -82,9 +82,23 @@ runs:
...
+          sleep 5
+        done
+        echo "Couldn't install nns-dapp canister"
+        exit 1
     - name: Install sns_aggregator
       if: ${{ inputs.sns_aggregator_wasm }}
       shell: bash
       run: |
-        dfx canister install sns_aggregator --wasm ${{ inputs.sns_aggregator_wasm }} --upgrade-unchanged --mode ${{ inputs.sns_aggregator_install_mode }} --yes
+        for ((try = 0; try < 5; try++)); do
```
lmuntaner, line 97: Can you add a TODO to remember to remove it?

.github/actions/start_dfx_snapshot/action.yaml
```diff
@@ -82,9 +82,23 @@ runs:
         export DFX_NETWORK=local
         ./config.sh
         echo "Install:"
-        dfx canister install nns-dapp --wasm ${{ inputs.nns_dapp_wasm }} --upgrade-unchanged --mode ${{ inputs.nns_dapp_install_mode }} --yes --argument "$(cat nns-dapp-arg-${DFX_NETWORK}.did)"
+        for ((try = 0; try < 5; try++)); do
```
lmuntaner, line 85: Can you add a TODO to remember to remove it?

lmuntaner (approved):
Thanks!

.github/actions/start_dfx_snapshot/action.yaml
dskloetd: Done

.github/actions/start_dfx_snapshot/action.yaml
dskloetd: Done

PR #4163, opened 2024-01-10
Balance placeholder instead of skeleton in wallet page

dskloetd:
# Motivation

We want to be able to render a wallet page (with placeholders) to a user who's not signed in.
Currently if you don't pass a balance to `WalletPageHeading` it renders a skeleton.
But the code currently never renders a `WalletPageHeading` when no balance is available.
When we show a wallet page while the user is not signed in, we want to render `-/-` as a wallet place holder.

# Changes

Remove the code to render a skeleton balance and replace it with rendering `-/-` when no balance is passed to `WalletPageHeading`.

This is not yet user visible because it's not yet possible to visit a wallet page without being signed in.

# Tests

Unit test is updated.

# Todos

not yet


lmuntaner (approved):
LGTM

PR #4167, opened 2024-01-11
Delete .bak file after splitting changelog

dskloetd:
# Motivation

When replacing an existing file, `sed` leaves behind the original with a specified extension.
We should clean it up because we don't need it.

# Changes

Delete `CHANGELOG-Nns-Dapp.md.bak` after `sed` creates it.

# Tests

Unit test updated to check that the file is deleted.

# Todos

not necessary


mstrasinskis (approved):
Thank you!

PR #4176, opened 2024-01-11
Use QR code placeholder icon on ckBTC wallet page

dskloetd:
# Motivation

We want to be able to render the wallet pages when the user is not signed in.
On the ckBTC wallet page we show a QR code with the BTC deposit address.
When the user is not signed in, there is no deposit address so we want to show a placeholder icon instead.

Screenshot 2024-01-11 at 14 57 09

image

# Changes

1. Remove the white background for the QR code if the user is not signed in.
2. Show the QR code icon instead.

# Tests

Existing tests updated.

# Todos

not user visible yet

### DIFF
--- a/frontend/src/lib/components/accounts/CkBTCInfoCard.svelte
+++ b/frontend/src/lib/components/accounts/CkBTCInfoCard.svelte
@@ -15,6 +15,7 @@
     QRCode,
     Copy,
     Html,
+    IconQRCode,
   } from "@dfinity/gix-components";
   import {
     BITCOIN_BLOCK_EXPLORER_MAINNET_URL,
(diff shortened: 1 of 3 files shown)


frontend/src/lib/components/accounts/CkBTCInfoCard.svelte
```diff
@@ -216,6 +221,13 @@
     }
   }
 
+  .qr-code-placeholder {
+    --qr-code-size: calc(18 * var(--padding));
```
mstrasinskis, line 225: Could you please also add a mobile screenshot?

mstrasinskis (approved):
LGTM, thank you!

frontend/src/lib/components/accounts/CkBTCInfoCard.svelte
dskloetd: I've added it to the PR description.

PR #4182, opened 2024-01-12
Fix: UI Improvements tokens table pages

lmuntaner:
# Motivation

# Changes

# Tests

# Todos


lmuntaner:
* `main`
  * **PR #4182** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4182?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4182?utm_source=stack-comment).

lmuntaner:
Issue 1:



Fixed 1:



Issue 2:



Fixed 2:



Issue 3:



Fixed 3:

lmuntaner:
@dskloetd please review

frontend/src/lib/components/tokens/MainWrapper.svelte
```diff
@@ -18,8 +18,7 @@
     height: 100%;
     width: 100%;
 
-    padding: var(--padding-2x) var(--padding-2x) var(--padding)
-      var(--padding-2x);
+    padding: var(--padding-2x) var(--padding-2x) 0 var(--padding-2x);
```
dskloetd, line 21: I think we might be scrolling the wrong container here.
If you compare the location of the scrollbar here:
image

to the scrollbar on the wallet page:
image

On the tokens page there is too much padding to the right of the scrollbar.
And if you look in the DOM, the MainWrapper itself is inside another div which has `class="scrollable-content"` so I'm guess it shouldn't be the `main` element doing the scrolling.

frontend/src/lib/pages/Tokens.svelte
```diff
@@ -1,16 +1,22 @@
...
   <TokensTable
     {userTokensData}
     on:nnsAction
     firstColumnHeader={$i18n.tokens.projects_header}
   />
-</TestIdWrapper>
+</div>
+
+<style lang="scss">
+  div {
```
dskloetd, line 18: Please use a class or it will get confusing if another `div` is added for whatever reason.

frontend/src/lib/components/layout/Footer.svelte
```diff
@@ -30,6 +30,9 @@
     background: var(--footer-background);
     --button-secondary-background: var(--focus-background);
 
+    // The footer should be on top of the content.
```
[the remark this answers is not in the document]
lmuntaner: Good catch. That's the style of the button.

I also wonder why the `z-index` was added there.

https://github.com/dfinity/gix-components/pull/352

frontend/src/lib/components/tokens/MainWrapper.svelte
lmuntaner: The scrollbar of the tokens page is on the table's container, not on the `main` container.

I believe that's another improvement, whether it should be the `main` or the table's container scrolling.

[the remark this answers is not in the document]
lmuntaner: It's ok to have padding at the top. What i didn't like was the padding below because it's not clear that there is more table.

frontend/src/lib/components/layout/Footer.svelte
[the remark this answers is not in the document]
lmuntaner: I removed issue 3 from this PR. That will be fixed by upgrading gix-components.

frontend/src/lib/pages/Tokens.svelte
lmuntaner: done

frontend/src/lib/components/tokens/MainWrapper.svelte
dskloetd: I don't think it's another improvement. I think both issues stem from using the wrong container for scrolling. I think it should be consistent with what other views do and I think doing it right will both make it consistent and solve the issues you're seeing.

dskloetd: Can you try if this solves the issues you're trying to solve with this PR?
https://github.com/dfinity/nns-dapp/compare/main...kloet/tokens-scroll

You can try it here:
https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/

lmuntaner: Yes, that looks good! Make a PR and I'll take a look at the diff and we can merge yours.

PR #4183, opened 2024-01-12
GIX-2080: Migrate e2e accounts.spec to tokens page

lmuntaner:
# Motivation

Enable my tokens flag for mainnet.

In this PR, set the flag to true for the accounts.spec e2e test and change the expects and interactions to adapt to the new UI.

# Changes

* Set flag `ENABLE_MY_TOKENS` to `true` in accounts.spec and adapt it to the new UI.
* Add convenient methods to different POs to adapt the e2e test to the new UI.

# Tests

Only test changes.

# Todos

### DIFF
--- a/frontend/src/tests/e2e/accounts.spec.ts
+++ b/frontend/src/tests/e2e/accounts.spec.ts
@@ -23,7 +23,7 @@ test("Test accounts requirements", async ({ page, context }) => {
   const accountsPo = appPo.getAccountsPo();
   const nnsAccountsPo = accountsPo.getNnsAccountsPo();
   const tokensTablePo = nnsAccountsPo.getTokensTablePo();
-  const mainAccountRow = await tokensTablePo.findRowByName(mainAccountName);
+  const mainAccountRow = tokensTablePo.findRowByName(mainAccountName);
   expect(await mainAccountRow.isPresent()).toBe(true);
 
   step("AU002: The user MUST be able to create an additional account");
@@ -47,7 +47,7 @@ test("Test accounts requirements", async ({ page, context }) => {
   // We wait until the table is loaded.
   await tokensTablePo.waitFor();
   // We wait until the subaccount row is loaded.
-  const subaccountRow = await tokensTablePo.findRowByName(subAccountName);
+  const subaccountRow = tokensTablePo.findRowByName(subAccountName);
   await subaccountRow.waitFor();
 
   expect(await accountNames()).toEqual([mainAccountName, subAccountName]);
@@ -57,7 +57,7 @@ test("Test accounts requirements", async ({ page, context }) => {
   // Get some ICP to be able to transfer
   await appPo.getIcpTokens(20);
   // Go back to /accounts page
-  const icRow = await appPo
+  const icRow = appPo
     .getTokensPo()
     .getTokensPagePo()
     .getTokensTable()
--- a/frontend/src/tests/e2e/accounts.spec.ts
+++ b/frontend/src/tests/e2e/accounts.spec.ts
@@ -23,8 +23,8 @@ test("Test accounts requirements", async ({ page, context }) => {
   const accountsPo = appPo.getAccountsPo();
   const nnsAccountsPo = accountsPo.getNnsAccountsPo();
   const tokensTablePo = nnsAccountsPo.getTokensTablePo();
-  const mainAccountRow = tokensTablePo.findRowByName(mainAccountName);
-  expect(await mainAccountRow.isPresent()).toBe(true);
+  const mainAccountRow = tokensTablePo.getRowByName(mainAccountName);
+  await mainAccountRow.waitFor();
 
   step("AU002: The user MUST be able to create an additional account");
   const subAccountName = "My second account";
@@ -47,7 +47,7 @@ test("Test accounts requirements", async ({ page, context }) => {
   // We wait until the table is loaded.
   await tokensTablePo.waitFor();
   // We wait until the subaccount row is loaded.
-  const subaccountRow = tokensTablePo.findRowByName(subAccountName);
+  const subaccountRow = tokensTablePo.getRowByName(subAccountName);
   await subaccountRow.waitFor();
 
   expect(await accountNames()).toEqual([mainAccountName, subAccountName]);
@@ -61,7 +61,7 @@ test("Test accounts requirements", async ({ page, context }) => {
     .getTokensPo()
     .getTokensPagePo()
     .getTokensTable()
-    .findRowByName("Internet Computer");
+    .getRowByName("Internet Computer");
   await icRow.click();
 
   await tokensTablePo.waitFor();
@@ -70,9 +70,12 @@ test("Test accounts requirements", async ({ page, context }) => {
   expect(await mainAccountRow.getBalance()).toEqual("20.00 ICP");
   expect(await subaccountRow.getBalance()).toEqual("0 ICP");
 
+  const subAccountAddress = await accountsPo.getAccountAddress(subAccountName);
+
   await mainAccountRow.click();
   await appPo.getWalletPo().getNnsWalletPo().transferToAccount({
     accountName: subAccountName,
+    expectedAccountAddress: subAccountAddress,
     amount: 5,
   });
   await appPo.goBack({ waitAbsent: false });
@@ -83,6 +86,8 @@ test("Test accounts requirements", async ({ page, context }) => {
   step(
     "AU005: The user MUST be able to see the transactions of a specific account"
   );
+  const mainAccountAddress =
+    await accountsPo.getAccountAddress(mainAccountName);
   await subaccountRow.click();
   const transactionList = appPo
     .getWalletPo()
@@ -93,5 +98,6 @@ test("Test accounts requirements", async ({ page, context }) => {
   expect(transactions).toHaveLength(1);
   const transaction = transactions[0];
 
+  expect(await transaction.getIdentifier()).toBe(`From: ${mainAccountAddress}`);
   expect(await transaction.getAmount()).toBe("+5.00");
 });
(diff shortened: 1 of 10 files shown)


lmuntaner:
* `main`
  * **PR #4183** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4183?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4183?utm_source=stack-comment).

frontend/src/tests/page-objects/TokensTable.page-object.ts
```diff
@@ -44,6 +45,21 @@ export class TokensTablePo extends BasePageObject {
     return undefined;
   }
 
+  async waitForRowByName(projectName: string): Promise<void> {
```
lmuntaner, line 48: This was needed because the table takes some time to load the data and I needed a way to way for the presence of a specific row.

frontend/src/tests/utils/utils.test-utils.ts
```diff
@@ -44,3 +44,8 @@ export const isNotVisible = (element: Element): boolean => {
     return nonNullish(heightString) && Number(heightString) === 0;
   }
 };
+
+export const waitForMilliseconds = (milliseconds: number): Promise<void> =>
```
lmuntaner, line 48: We have this util for production, but it was not usable within the e2e tests. Therefore, I copied within `tests` directory.

lmuntaner:
@dskloetd please review

frontend/src/tests/page-objects/TokensTable.page-object.ts
```diff
@@ -44,6 +45,21 @@ export class TokensTablePo extends BasePageObject {
...
+    let row = await this.findRowByName(projectName);
+    let counter = 1;
+    const MAX_TRIES = 10;
+    const WAIT_MILLISECONDS = 1_000;
+    while (isNullish(row) && counter < MAX_TRIES) {
+      await waitForMilliseconds(WAIT_MILLISECONDS * counter);
```
dskloetd, line 54: I think it would be cleaner to use the existing `querySelectorCount` or `countByTestId` and wait for rows one by one until you have one that matches.

lmuntaner: Do you mean instead of `findRowByName`?

How would the count help me find the specific row?

Do you mean to create a `waitForRowsCount(amount)`?

Another idea I had was to use the projectName in the testId of the row. In that case I could select it directly and use `waitFor`.

dskloetd: I mean to avoid explicit `waitForMilliseconds`.
Something like
```
for (let i = 1;; i++) {
  const row = TokensTableRowPo.countUnder({ element: this.root, count: i}).at(-1);
  if (await row.getProjectName() === projectName) {
    return row;
  }
}
```
Implementation of `TokensTableRowPo.countUnder` would be similar to `SelectUniverseCardPo.countUnder`.

Putting the token symbol in the test ID is also fine with me but it might interfere with an existing test ID.

lmuntaner: I think it's nicer to use another attribute for the project name: `data-title`.

I implemented it and it looks cleaner IMO.

Let me know what you think.

frontend/src/tests/page-objects/TokensTableRow.page-object.ts
```diff
@@ -17,6 +17,18 @@ export class TokensTableRowPo extends BasePageObject {
     );
   }
 
+  static async byTitle({
```
lmuntaner, line 20: @dskloetd I added this static method to the row.

frontend/src/tests/page-objects/TokensTable.page-object.ts
```diff
@@ -34,14 +34,7 @@ export class TokensTablePo extends BasePageObject {
...
   ): Promise<TokensTableRowPo | undefined> {
-    const rows = await this.getRows();
-    for (const row of rows) {
-      const name = await row.getProjectName();
-      if (name === projectName) {
-        return row;
-      }
-    }
-    return undefined;
+    return TokensTableRowPo.byTitle({ element: this.root, title: projectName });
```
lmuntaner, line 37: Just select directly instead of looping.

lmuntaner:

frontend/src/tests/page-objects/TokensTableRow.page-object.ts
```diff
@@ -17,6 +17,18 @@ export class TokensTableRowPo extends BasePageObject {
...
+    element,
+    title,
+  }: {
+    element: PageObjectElement;
+    title: string;
+  }): Promise<TokensTableRowPo> {
```
dskloetd, line 26: This doesn't need to be async, does it?
Same for the caller and the caller of the caller.

frontend/src/tests/e2e/accounts.spec.ts
```diff
@@ -14,55 +20,70 @@ test("Test accounts requirements", async ({ page, context }) => {
...
+    (await tokensTablePo.getRowsData()).map(({ projectName }) => projectName);
   expect(await accountNames()).toEqual([mainAccountName, subAccountName]);
 
   // The linked account should still be present after refresh
   await page.reload({ waitUntil: "load" });
+
+  // We wait until the table is loaded.
+  await tokensTablePo.waitFor();
+  // We wait until the subaccount row is loaded.
+  const subaccountRow = await tokensTablePo.findRowByName(subAccountName);
```
dskloetd, line 50: I think this doesn't need `await` anymore.

frontend/src/tests/page-objects/TokensTable.page-object.ts
```diff
@@ -34,14 +34,7 @@ export class TokensTablePo extends BasePageObject {
   async findRowByName(
```
dskloetd, line 34: Does `find` still make sense with the current implementation or should it be `get`?
This will return something regardless of whether row is there yet so "find" seems inaccurate.

frontend/src/tests/e2e/accounts.spec.ts
```diff
@@ -72,6 +93,5 @@ test("Test accounts requirements", async ({ page, context }) => {
   expect(transactions).toHaveLength(1);
   const transaction = transactions[0];
 
-  expect(await transaction.getIdentifier()).toBe(`From: ${mainAccountAddress}`);
```
dskloetd, line 75: Why is this removed?

frontend/src/tests/e2e/accounts.spec.ts
```diff
@@ -14,55 +20,70 @@ test("Test accounts requirements", async ({ page, context }) => {
...
+  expect(await mainAccountRow.getBalance()).toEqual("20.00 ICP");
+  expect(await subaccountRow.getBalance()).toEqual("0 ICP");
 
-  const subAccountAddress =
-    await nnsAccountsPo.getAccountAddress(subAccountName);
-  await nnsAccountsPo.getMainAccountCardPo().click();
+  await mainAccountRow.click();
   await appPo.getWalletPo().getNnsWalletPo().transferToAccount({
     accountName: subAccountName,
-    expectedAccountAddress: subAccountAddress,
```
dskloetd, line 49: Is there no way to see the account address once we launch the my tokens table? Isn't it visible when you click on the receive icon?

lmuntaner: Yes, it's there.

But is it necessary for this test?

Maybe we can add another e2e test, no? Or is there a reason to have such a test here?

frontend/src/tests/e2e/accounts.spec.ts
lmuntaner: Because we don't have the address in the Tokens table.

frontend/src/tests/page-objects/TokensTableRow.page-object.ts
lmuntaner: True, removed.

frontend/src/tests/page-objects/TokensTable.page-object.ts
lmuntaner: done

frontend/src/tests/e2e/accounts.spec.ts
lmuntaner: reverted

frontend/src/tests/e2e/accounts.spec.ts
lmuntaner: Done

lmuntaner:

frontend/src/tests/page-objects/Accounts.page-object.ts
```diff
@@ -74,4 +74,16 @@ export class AccountsPo extends BasePageObject {
...
+
+  async getAccountAddress(accountName: string): Promise<string> {
+    const row = this.getNnsAccountsPo()
+      .getTokensTablePo()
+      .getRowByName(accountName);
+    await row.clickReceive();
+    const receiveModalPo = this.getReceiveModalPo();
+    await receiveModalPo.waitFor();
+    const address = await receiveModalPo.getAddress();
+    await receiveModalPo.clickFinish();
```
dskloetd, line 86: "Finish" does an extra unnecessary reload of the transactions, right?
We could use "close" instead to avoid that.

dskloetd (approved):

frontend/src/tests/page-objects/Accounts.page-object.ts
lmuntaner: done

PR #4184, opened 2024-01-12
Use page object in TransactionList.spec.ts

dskloetd:
# Motivation

Make the test more maintainable and easier to read.

# Changes

Use a page object in `TransactionList.spec.ts`.

# Tests

only

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #4185, opened 2024-01-12
Transaction list placeholder

dskloetd:
# Motivation

[The design](https://www.figma.com/file/x8dk5axQUf8Bi43uUsqCHc/NNS-Production-%26-Hand-over?type=design&node-id=1319-1085&mode=design&t=BiIOfY7BQjiYMj8P-4) for the logged-out wallet page has a new placeholder for when there are no transactions.
We should use this also when the user is logged in but there are no transactions.

This PR introduces a simple component to use as the placeholder when there are no transactions and uses it just for the logged-in case.

# Changes

1. Add `NoTransactions` component.
2. Use it instead of just text when there are no transactions.

# Tests

Unit tests and page objects were updated.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp-unreleased.md
+++ b/CHANGELOG-Nns-Dapp-unreleased.md
@@ -17,6 +17,7 @@ proposal is successful, the changes it released will be moved from this file to
 #### Changed
 
 * New light and dark theme colors.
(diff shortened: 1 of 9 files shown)


lmuntaner (approved):
LGTM! Thanks!

PR #4194, opened 2024-01-15
Remove double scroll container from My Tokens table

dskloetd:
# Motivation

The `main` element of the My Tokens table is a scrolling container, but it exists in another scroll container.
This results in a strange cut-off at the top and bottom and strange placement of the scroll bar.
See https://github.com/dfinity/nns-dapp/pull/4182 for details.

# Changes

1. Remove `height` and `overflow-y` from the `main` element.
2. Remove custom padding that was necessary because of the double scroll container.

# Tests

Tested manually at https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/

# Todos

not necessary

### DIFF
--- a/frontend/src/lib/components/tokens/MainWrapper.svelte
+++ b/frontend/src/lib/components/tokens/MainWrapper.svelte
@@ -17,13 +17,6 @@
     justify-content: center;
     width: 100%;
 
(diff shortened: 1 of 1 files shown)


lmuntaner (approved):
Thanks!

PR #4196, opened 2024-01-15
GIX-2187: Make ICRC wallet renderable when not signed in.

dskloetd:
# Motivation

This PR makes it possible to render the `IcrcWallet.svelte` page when the user is not logged in.

# Changes

1. Don't load data until the user is logged in.
2. Don't render `IcrcBalancesObserver` if the user is not logged in.
3. Treat `selectedAccountStore.account` as nullable because we want to render most of the page when there is no account.
4. Use `SignInGuard` to render a sign in button if the user is not logged in.
5. Render the no-transactions placeholder instead of the `IcrcWalletTransactionsList` component if there is no account.

# Tests

Unit tests are added.

# Todos

not yet

### DIFF
--- a/frontend/src/lib/components/accounts/IcrcWalletPage.svelte
+++ b/frontend/src/lib/components/accounts/IcrcWalletPage.svelte
@@ -1,5 +1,7 @@
 <script lang="ts">
+  import { authSignedInStore } from "$lib/derived/auth.derived";
   import { Island, Spinner } from "@dfinity/gix-components";
+  import SignInGuard from "$lib/components/common/SignInGuard.svelte";
   import Separator from "$lib/components/ui/Separator.svelte";
   import type { Writable } from "svelte/store";
   import type { WalletStore } from "$lib/types/wallet.context";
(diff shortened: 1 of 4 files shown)


frontend/src/lib/components/accounts/IcrcWalletPage.svelte
```diff
@@ -129,31 +143,42 @@
...
+          balance={nonNullish($selectedAccountStore.account) &&
+          nonNullish(token)
+            ? TokenAmountV2.fromUlps({
+                amount: $selectedAccountStore.account.balanceUlps,
+                token,
+              })
+            : undefined}
         >
           <slot name="header-actions" />
+          <SignInGuard />
```
lmuntaner, line 181: Why not use directly the `SignIn` button?

The `SignInGuard` renders the children when the user is logged in but here you want only the button, right?

lmuntaner (approved):
LGTM! Just a minor comment.

frontend/src/lib/components/accounts/IcrcWalletPage.svelte
dskloetd: It want the `SignIn` button but only `{#if $authSignedInStore}` which is what `SignInGuard` provides.
Otherwise I need an extra 2 lines for the `if` here.

lmuntaner: Ok, thanks for the explanation!

PR #4197, opened 2024-01-15
Fix BTC address QR code spinner in dark theme

dskloetd:
# Motivation

When using the `Spinner` component, it renders in the current color.
By default, this is dark in the light theme and light in the dark theme.
But the QR code for the BTC deposit address is always rendered on a white background, so it should always have a dark spinner.

# Changes

Set `color` to `black` in the `qr-code` element.

# Tests

Tested manually:

image

# Todos


lmuntaner (approved):
LGTM.

PR #4198, opened 2024-01-15
GIX-2187: Make ckBTC wallet renderable when not signed in.

dskloetd:
# Motivation

This PR makes it possible to render the `CkBTCWallet.svelte` page when the user is not logged in.

# Changes

1. Hide Send/Receive buttons.
2. Don't load data.
3. Don't require `nonNullish($selectedAccountStore.account)` to show the info card.
4. Show transaction list placeholder.

The rest of the functionality comes from reusing `IcrcWalletPage.svelte`.

# Tests

Unit tests added.
Also added missing tests for Send/Receive button for `IcrcWalletPage` that should have been in the [previous PR](https://github.com/dfinity/nns-dapp/pull/4196).
Test manually in a separate branch.

# Todos

not yet


lmuntaner (approved):
LGTM

PR #4204, opened 2024-01-16
GIX-2187: Make NNS wallet renderable when not signed in.

dskloetd:
# Motivation

This PR makes it possible to render the `NnsWallet.svelte` page when the user is not logged in.

# Changes

1. Don't `pollAccounts` until the user is logged in.
2. Treat `selectedAccountStore.account` as nullable because we want to render most of the page when there is no account.
3. Use SignInGuard to render a sign in button if the user is not logged in.
4. Render the no-transactions placeholder instead of the SnsWalletTransactionsList component if there is no account.
5. Hide Send/Receive buttons if there is no account.

# Tests

Unit tests are added.

Drive-by: Use the real auth store instead of `mockAuthStoreSubscribe`.

# Todos

not yet

### DIFF
--- a/frontend/src/lib/pages/NnsWallet.svelte
+++ b/frontend/src/lib/pages/NnsWallet.svelte
@@ -1,7 +1,9 @@
 <script lang="ts">
+  import { authSignedInStore } from "$lib/derived/auth.derived";
   import type { Account } from "$lib/types/account";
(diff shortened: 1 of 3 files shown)


lmuntaner (approved):
LGTM

PR #4208, opened 2024-01-16
GIX-2187: Render wallet page when not signed in.

dskloetd:
# Motivation

People should be able to see the wallet page while not being signed in.
In particular, this will allow us to deep link to the ckBTC wallet so that people can read info about ckBTC and then convert after signing in.

# Changes

In `frontend/src/routes/(app)/(u)/(detail)/wallet/+page.svelte` always render `Wallet` and never `SignInAccounts`.

# Tests

The existing unit test only checked that a sign-in button was rendered. It still passed but because the wallet also renders a sign-in page.
The current component doesn't have any logic anymore, except for passing on the account identifier.
So I just added a unit test to check that the account identifier makes it into the wallet component.

Also tested manually [here](https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/).
You can navigate to a wallet and then log out, or use the deep links for [ICP](https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/wallet/?u=qsgjb-riaaa-aaaaa-aaaga-cai), [SNS](https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/wallet/?u=be2us-64aaa-aaaaa-qaabq-cai), [ckETH](https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/wallet/?u=omy6t-mmaaa-aaaaa-qabgq-cai) and [ckBTC](https://qsgjb-riaaa-aaaaa-aaaga-cai.dskloet-ingress.devenv.dfinity.network/wallet/?u=oz7p6-neaaa-aaaaa-qabfa-cai) (all different implementations).

# Todos


lmuntaner (commented):
Should we add also a test that it renders correctly without an account identifier?

dskloetd:
> Should we add also a test that it renders correctly without an account identifier?

Those tests exist for the individual wallets.
Do you think this component needs the same tests?
It's the individual wallets that have logic depending on whether an account identifier is passed or not.
This component doesn't have any such logic so I'm not sure it makes sense to have such a test for this component.
I wasn't even sure whether to add a test at all since there is no logic in this component anymore.
Am I missing a scenario that could break if we don't have such a test here as well?

dskloetd:
> Before there was the test of logged in and not logged in. In the new scenario, the distinction is done by passing an account identifier or not.

No, passing an account identifier is orthogonal to whether the user is logged in. This is just the account identifier that comes from the URL and is passed on to the component. There is no check whether the user is loggin in to decide whether to pass on the account identifier. It is simply always passed on directly from the URL without any logic.

All 4 combinations are possible:
1. not signed-in, no account identifier
2. not signed-in, yes account identifier
3. yes signed-in, no account identifier
4. yes signed-in, yes account identifier

All cases are handled by the wallet components.
When no account identifier is passed, we default to the default Main account.

The only thing that the page could do wrong, in my view, is not passing on the account identifier from the URL to the prop. As long as it's passed on to the component, all logic account on the account identifier happens in the child components.

> The logic is to pass an account or not.

What do you mean by "or not"? There is no logic in the `+page.svelte`. The account identifier is passed exactly as received without any logic.

lmuntaner (approved):
Thanks for the explanation!

PR #4209, opened 2024-01-16
Feat: Remove duplicated actions cell

lmuntaner:
# Motivation

There was an issue with the e2e tests clicking the Receive button in the tokens table. The reason was that the selector was getting the mobile button is hidden in desktop.

In this PR, I remove the duplicated button while maintaining the flexibility of adding new columns.

# Changes

* Set the grid style using inline style in the TokensTableRow.svelte. Because inline style is more than a media query, I needed to use `!important`.

# Tests

No new tests, this is a refactor of the same current functionality.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
+++ b/frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
@@ -46,6 +46,15 @@
       });
     }
   };
+
+  // Should be the same as the area in the classes `rows-count-X`.
+  const cellAreaName = (index: number) => `cell-${index}`;
+  // This will allow us to have different number of rows depending on the number of columns.
+  // It's not really necessary for the TokensTable becuase we know we want only 1 row.
+  // But this should be moved when we make the generic table.
+  const mobileTemplateClass = (rowsCount: number) => {
+    return `rows-count-${rowsCount}`;
+  };
 </script>
 
 <svelte:element
@@ -56,6 +65,7 @@
   on:keypress={handleClick}
   on:click={handleClick}
   data-tid="tokens-table-row-component"
+  class={mobileTemplateClass(2)}
 >
   <div role="cell" class="title-cell">
     <div class="title-logo-wrapper">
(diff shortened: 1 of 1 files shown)


lmuntaner:
* `main`
  * **PR #4209** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4209?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4209?utm_source=stack-comment).

lmuntaner:
@dskloetd it adds some complexity, but at least there is no html duplication now.

Moreover, I think we'll need the inline style anyway for the generic table (when we create it).

lmuntaner:
@dskloetd ready for review. No `!important` used.

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -128,15 +127,30 @@
...
 
     text-decoration: none;
 
+    &.rows-count-1 {
+      grid-template-areas:
+        "first-cell last-cell"
+        "cell-0 cell-0";
+    }
+
+    &.rows-count-2 {
```
dskloetd, line 143: I'm a bit confused by the template names. I a template with 2 rows and one with 3 rows. Shouldn't they have `2` and `3` in the name, rather than `1` and `2`?

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -128,15 +127,30 @@
     // If we use grid-template-areas, we need to specify all the areas.
     // That makes it hard to have dynamic columns.
     // Instead, we duplicate the actions. Once as the last cell, another within the title cell.
-    display: flex;
+    display: grid;
+    // `grid-template-areas` set with inline style
```
dskloetd, line 131: where is the inline style?

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
dskloetd, line 1: The change looks OK to me but I can't render it in my head.
Have you tried testing it with multiple rows/columns?
If not, maybe we shouldn't make it generic until we can actually test it?
Or comment out the extra bits and just leave as a suggestion while mentioning that it's not tested?

lmuntaner: Yes, I tried adding another column:

I needed to add the column as well in the TokensTable `grid-template-column` which will be another PR to make it dynamic of number of columns.

Screenshot 2024-01-17 at 08 25 29
Screenshot 2024-01-17 at 08 25 35

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: I see the confusion. I counted after the first row. I changed it.

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: Removed. This was from an old version.

lmuntaner:

dskloetd (approved):

PR #4210, opened 2024-01-16
fix: use all neurons in voting card

mstrasinskis:
# Motivation

Fix missing NF-neurons in the SNS proposal details voting block.

# Changes

- switch to use all user neurons `snsSortedNeuronStore`.

# Tests

- `SnsVotingCard` should display NF neurons as well

# Todos


lmuntaner (approved):
Thanks!

PR #4214, opened 2024-01-17
Keep local replica accessible locally

dskloetd:
# Motivation

`dfx` has a flag `--domain` which you need to specify to make the replica accessible remotely at that domain.
But when you do, this overrides the default value, which is `localhost` so then the replica is no longer accessible from the machine itself at `localhost`.
To make the replica accessible on both domains, we need to pass the `--domain` flag twice.

# Changes

Pass `--domain localhost` to replica when specifying an additional domain to `dfx-snapshot-start`.

# Tests

Tested manually on my DevEnv.

# Todos


lmuntaner (approved):
Thanks

PR #4219, opened 2024-01-17
GIX-2172: Remove old BTC withdrawal flow

dskloetd:
# Motivation

There are currently 3 BTC withdrawal flows:
1. Transfer ckBTC to minter withdrawal account, then ask minter to convert (only with ENABLED_CKBTC_ICRC2 disabled)
2. Have existing ckBTC in withdrawal account and ask the minter to convert (regardless of ENABLED_CKBTC_ICRC2)
3. Approve a transfer, then ask the minter to convert directly from your account (only with ENABLED_CKBTC_ICRC2 enabled)

This PR removes the first and keeps the other 2.
The second flow will be remove soon in another PR.

# Changes

1. Remove the possibility of rendering the "transfer to withdrawal account" step from `ConvertBtcInProgress.svelte
`.
2. Remove the check on `ENABLE_CKBTC_ICRC2` in `CkBTCTransactionModal.svelte` and determine the flow based only on whether we are trying to use an already funded withdrawal address or not.
3. Remove the service method for transferring to a withdrawal account and then withdrawing.
4. Remove tests for the removed flow.

# Tests

Unit tests removed.
Tested manually that ICRC2 flow still works.

# Todos

### DIFF
--- a/frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
+++ b/frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
@@ -3,7 +3,6 @@
   import { startBusy, stopBusy } from "$lib/stores/busy.store";
   import { i18n } from "$lib/stores/i18n";
   import { toastsSuccess } from "$lib/stores/toasts.store";
-  import { ENABLE_CKBTC_ICRC2 } from "$lib/stores/feature-flags.store";
   import type { NewTransaction, TransactionInit } from "$lib/types/transaction";
   import { TransactionNetwork } from "$lib/types/transaction";
   import type { ValidateAmountFn } from "$lib/types/transaction";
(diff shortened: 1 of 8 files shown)


frontend/src/lib/modals/accounts/CkBTCTransactionModal.svelte
```diff
@@ -132,18 +130,12 @@
     if (withdrawalAccount) {
       useIcrc2ForProgress = false;
       ({ success } = await retrieveBtc(params));
-    } else if ($ENABLE_CKBTC_ICRC2) {
```
lmuntaner, line 135: Can we also remove `ENABLE_CKBTC_ICRC2`?

lmuntaner (approved):
Thanks!

PR #4222, opened 2024-01-17
Console feature to temporarily add ICRC-1 token

dskloetd:
# Motivation

Sometimes people accidentally send tokens to their NNS dapp address while those tokens are not supported by NNS dapp.
It would be nice if there was a way to recover those tokens.

The experience could definitely be improved but I don't want to spend more time on this now, so please review based on whether this is an improvement or not.

# Changes

1. Provide a hard coded logo for unknown tokens.
2. Add a method to the `window` object to add a ledger and index canister to the store.

# Tests

1. Adjusted an existing unit test for an unknown token.
2. Tested manually adding an unknown ledger.

Currently it happens to work with and without index canister but there are no guarantees:
```
__experimentalAddIcrc1Token('m7h5t-euaaa-aaaaa-qabia-cai');
__experimentalAddIcrc1Token('m7h5t-euaaa-aaaaa-qabia-cai', 'myg3h-jmaaa-aaaaa-qabiq-cai');
```
image

# Todos

### DIFF
--- a/frontend/src/lib/stores/icrc-canisters.store.ts
+++ b/frontend/src/lib/stores/icrc-canisters.store.ts
@@ -1,5 +1,7 @@
+import { browser } from "$app/environment";
 import type { UniverseCanisterIdText } from "$lib/types/universe";
-import type { Principal } from "@dfinity/principal";
+import { Principal } from "@dfinity/principal";
+import { nonNullish } from "@dfinity/utils";
 import type { Readable } from "svelte/store";
 import { writable } from "svelte/store";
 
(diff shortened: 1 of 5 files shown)


frontend/src/lib/stores/icrc-canisters.store.ts
```diff
@@ -54,3 +56,30 @@ const initIcrcCanistersStore = (): IcrcCanistersStore => {
 };
 
 export const icrcCanistersStore = initIcrcCanistersStore();
+
+// Useful to help people record coins they accidentally sent to NNS dapp, but
+// which are not officially supported by the NNS dapp.
+//
+// In the devtools console, run:
+// __experimentalAddIcrc1Token(ledgerCanisterId, indexCanisterId)
+if (browser) {
```
mstrasinskis, line 65: Will using the `triggerDebugReport` improve the ux a bit?

dskloetd: Also, we should really just implement this properly for everybody.
This is just to be able to support for lost coins. I don't care about the UX tbh.

mstrasinskis: No objections :)

mstrasinskis (approved):
The code looks good to me. Thanks!

lmuntaner (approved):
LGTM

PR #4228, opened 2024-01-18
GIX-2138: Remove ckBTC withdrawal account flow

dskloetd:
# Motivation

There are 3 BTC withdrawal flows:
1. Transfer ckBTC to minter withdrawal account, then ask minter to convert (only with ENABLED_CKBTC_ICRC2 disabled)
2. Have existing ckBTC in withdrawal account and ask the minter to convert (regardless of ENABLED_CKBTC_ICRC2)
3. Approve a transfer, then ask the minter to convert directly from your account (only with ENABLED_CKBTC_ICRC2 enabled)

Number 1 was removed in https://github.com/dfinity/nns-dapp/pull/4219.
This PR removes number 2.

# Changes

# Tests

# Todos


frontend/src/lib/pages/CkBTCAccounts.svelte
```diff
@@ -30,6 +29,4 @@
   testId="ckbtc-accounts-body"
   selectedUniverseId={$selectedCkBTCUniverseIdStore}
   token={token?.token}
->
-  <CkBTCWithdrawalAccount slot="additional-accounts" />
-</IcrcAccountsPage>
+></IcrcAccountsPage>
```
lmuntaner, line 32: This means we could use IcrcAccounts if we set up the ckbtc like any other ICRC account, right?

lmuntaner (approved):
Love it!

frontend/src/lib/pages/CkBTCAccounts.svelte
dskloetd: Do you mean if we move `loadCkBTCInfo` somewhere else?
I could look into it but won't this be obsolete anyway when the My Tokens table launches?

PR #4233, opened 2024-01-18
Add exclude type param

mstrasinskis:
# Motivation

Add `excludeType` parameter support. It will be used to filter sns proposals by types.

# Changes

- pass down `nsFunctions` to the `loadSnsProposals` service (needs to calculate `excludeType` param).
- added `toExcludeTypeParameter` function.
- calculate `excludeType` and add it to the request

# Tests

- `toExcludeTypeParameter`
- add nsFunctions argument to sns proposal service specs

# Todos

Not yet

### DIFF
--- a/frontend/src/lib/utils/sns-proposals.utils.ts
+++ b/frontend/src/lib/utils/sns-proposals.utils.ts
@@ -14,8 +14,8 @@ import type {
 import { replacePlaceholders } from "$lib/utils/i18n.utils";
 import { getSnsNeuronIdAsHexString } from "$lib/utils/sns-neuron.utils";
 import {
-  isGenericNervousSystemFunction,
-  isNativeNervousSystemFunction,
+  isSnsGenericNervousSystemFunction,
+  isSnsesNativeNervousSystemFunction,
 } from "$lib/utils/sns.utils";
 import { basisPointsToPercent } from "$lib/utils/utils";
 import { Vote } from "@dfinity/nns";
@@ -496,7 +496,7 @@ export const toExcludeTypeParameter = ({
     return [];
   }
   const nativeNsFunctionIds = snsFunctions
-    .filter(isNativeNervousSystemFunction)
+    .filter(isSnsesNativeNervousSystemFunction)
     .map(({ id }) => id);
   const isNativeNsFunctionSelected = (nsFunctionId: bigint) =>
     filter.some(({ id, checked }) => id === nsFunctionId.toString() && checked);
@@ -504,12 +504,12 @@ export const toExcludeTypeParameter = ({
     (id) => !isNativeNsFunctionSelected(id)
   );
   const genericNsFunctionIds = snsFunctions
-    .filter(isGenericNervousSystemFunction)
+    .filter(isSnsGenericNervousSystemFunction)
     .map(({ id }) => id);
-  const isGenericNsFunctionChecked = filter.some(
+  const isSnsGenericNsFunctionChecked = filter.some(
     ({ id, checked }) => id === ALL_SNS_GENERIC_PROPOSAL_TYPES_ID && checked
   );
-  const excludedGenericNsFunctionIds = isGenericNsFunctionChecked
+  const excludedGenericNsFunctionIds = isSnsGenericNsFunctionChecked
     ? []
     : genericNsFunctionIds;
 
@@ -536,7 +536,7 @@ export const generateSnsProposalTypesFilterData = ({
     typesFilterState.find(({ id: stateId }) => id === stateId)?.checked !==
     false;
   const nativeNsFunctionEntries: Filter<SnsProposalTypeFilterId>[] = nsFunctions
-    .filter(isNativeNervousSystemFunction)
+    .filter(isSnsesNativeNervousSystemFunction)
     // ignore { 0n: "All Topics"}
     .filter(({ id }) => id !== ALL_SNS_PROPOSAL_TYPES_NS_FUNCTION_ID)
     .map((entry) => ({
@@ -557,7 +557,7 @@ export const generateSnsProposalTypesFilterData = ({
     }
   );
   const genericNsFunctionEntries: Filter<SnsProposalTypeFilterId>[] =
-    nsFunctions.some(isGenericNervousSystemFunction)
+    nsFunctions.some(isSnsGenericNervousSystemFunction)
       ? // Replace all generic entries w/ a single "All Generic"
         [
           {
--- a/frontend/src/lib/utils/sns-proposals.utils.ts
+++ b/frontend/src/lib/utils/sns-proposals.utils.ts
@@ -480,6 +480,46 @@ export const getUniversalProposalStatus = (
   return statusType;
 };
 
+/**
+ * Returns the proposal type ids that should be excluded from the response.
+ * snsFunctions are needed to process "All Generic" entry (when not selected, include all generic ids).
+ */
+export const toExcludeTypeParameter = ({
+  filter,
+  snsFunctions,
+}: {
+  filter: Filter<SnsProposalTypeFilterId>[];
+  snsFunctions: SnsNervousSystemFunction[];
+}): bigint[] => {
+  // If no filter is selected, return all functions
+  if (filter.length === 0) {
+    return [];
+  }
+  const nativeNsFunctionIds = snsFunctions
+    .filter(isNativeNervousSystemFunction)
+    .map(({ id }) => id);
+  const isNativeNsFunctionSelected = (nsFunctionId: bigint) =>
+    filter.some(({ id, checked }) => id === nsFunctionId.toString() && checked);
+  const excludedNativeNsFunctionIds = nativeNsFunctionIds.filter(
+    (id) => !isNativeNsFunctionSelected(id)
+  );
+  const genericNsFunctionIds = snsFunctions
+    .filter(isGenericNervousSystemFunction)
+    .map(({ id }) => id);
+  const isGenericNsFunctionChecked = filter.some(
+    ({ id, checked }) => id === ALL_SNS_GENERIC_PROPOSAL_TYPES_ID && checked
+  );
+  const excludedGenericNsFunctionIds = isGenericNsFunctionChecked
+    ? []
+    : genericNsFunctionIds;
+
+  return (
+    [...excludedNativeNsFunctionIds, ...excludedGenericNsFunctionIds]
+      // never exclude { 0n: "All Topics"}
+      .filter((id) => id !== ALL_SNS_PROPOSAL_TYPES_NS_FUNCTION_ID)
+  );
+};
+
 // Generate new "types" filter data, but preserve the checked state of the current filter state
 // `nsFunctions` can be changed on the backend, and to display recently created proposal types, new entries should be preselected.
 export const generateSnsProposalTypesFilterData = ({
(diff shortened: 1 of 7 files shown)


frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
```diff
@@ -830,6 +831,91 @@ describe("sns-proposals utils", () => {
...
+          filter: [snsSpecificFilterEntry],
+          snsFunctions,
+        }).find((id) => id === 0n)
+      ).toBe(undefined);
+    });
+
+    it("should exclude all except selected native ns function", () => {
+      expect(
+        toExcludeTypeParameter({
+          filter: [nativeFilterEntry],
```
lmuntaner, line 903: Could you add also non selected to check that they not excluded? and also the filter for SNS_SPECIFIC

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
```diff
@@ -830,6 +831,91 @@ describe("sns-proposals utils", () => {
...
+      ).toStrictEqual([1001n]);
+    });
+
+    it('should exclude generic when "SNS_SPECIFIC" is selected', () => {
+      expect(
+        toExcludeTypeParameter({
+          filter: [snsSpecificFilterEntry],
+          snsFunctions,
+        })
+      ).toStrictEqual([1n]);
```
lmuntaner, line 915: What is `1n`? Is not clear to me what should be the exclude type when "SNS_SPECIFIC" is selected.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
```diff
@@ -830,6 +831,91 @@ describe("sns-proposals utils", () => {
...
+      ).toBe(undefined);
+    });
+
+    it("should exclude all except selected native ns function", () => {
+      expect(
+        toExcludeTypeParameter({
+          filter: [nativeFilterEntry],
+          snsFunctions,
+        })
+      ).toStrictEqual([1001n]);
```
lmuntaner, line 906: Ditto, can you reference maybe the nervous function that will be excluded?

frontend/src/tests/lib/services/_public/sns-proposals.services.spec.ts
lmuntaner, line 1: Will you add a test with some exclude functionality?

frontend/src/lib/utils/sns-proposals.utils.ts
```diff
@@ -480,6 +480,46 @@ export const getUniversalProposalStatus = (
...
+    .map(({ id }) => id);
+  const isNativeNsFunctionSelected = (nsFunctionId: bigint) =>
+    filter.some(({ id, checked }) => id === nsFunctionId.toString() && checked);
+  const excludedNativeNsFunctionIds = nativeNsFunctionIds.filter(
+    (id) => !isNativeNsFunctionSelected(id)
+  );
+  const genericNsFunctionIds = snsFunctions
+    .filter(isGenericNervousSystemFunction)
+    .map(({ id }) => id);
+  const isGenericNsFunctionChecked = filter.some(
```
lmuntaner, line 509: I find confusing the `Generic` term here without SNS in front. Do you mean the SNS Specific functions?

frontend/src/lib/utils/sns-proposals.utils.ts
```diff
@@ -480,6 +480,46 @@ export const getUniversalProposalStatus = (
...
+  }
+  const nativeNsFunctionIds = snsFunctions
+    .filter(isNativeNervousSystemFunction)
+    .map(({ id }) => id);
+  const isNativeNsFunctionSelected = (nsFunctionId: bigint) =>
+    filter.some(({ id, checked }) => id === nsFunctionId.toString() && checked);
+  const excludedNativeNsFunctionIds = nativeNsFunctionIds.filter(
+    (id) => !isNativeNsFunctionSelected(id)
+  );
+  const genericNsFunctionIds = snsFunctions
```
lmuntaner, line 506: Ditto with the name, these are SNS specific nervous functions, right?

lmuntaner (commented):
Thanks! That the API expectes the "excluded_type" makes everything harder to read...

I added some comments. Mainly for readibility. I believe that adding more tests could also make understanding the function better.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
mstrasinskis: - for "non selected" there is "should return empty list if nothing checked".
- for selection of "SNS_SPECIFIC" - "should exclude generic when "SNS_SPECIFIC" is selected"
Probably after I rename the IDs it will be more clear where is what.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
mstrasinskis: Expected values were replaced with the const. There are two types of nsFunctions - `generic` and `native`
- `native` nsFunctions are same between SNSes;
- each sns has it's own(or none) `generic` nsFunctions.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
mstrasinskis: Probably after I rename the IDs it will be more clear where is what.

frontend/src/lib/utils/sns-proposals.utils.ts
mstrasinskis: How about `isSnsGenericNervousSystemFunction` and `isSnsesNativeNervousSystemFunction`?

frontend/src/lib/utils/sns-proposals.utils.ts
mstrasinskis: Yes, @lmuntaner. I find `native` and `generic` a bit confusing to understand, but not sure about replacing `generic` with `specific` everywhere in the FE. Then should we also replace `native` w/ ... `common`?

frontend/src/tests/lib/services/_public/sns-proposals.services.spec.ts
mstrasinskis: Good point, added https://github.com/dfinity/nns-dapp/pull/4233/files#diff-ec2dbb4dfc0a6b866ff4caef8c6c28d8d3d8058f1bb864e073f4d442f26c9d5bR206

frontend/src/lib/utils/sns-proposals.utils.ts
lmuntaner: I believe singuler makes more sense to me. But isSnsesNativeNervousSystemFunction is also ok.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
lmuntaner: Yes, but I'd like to test also how non-selected filters affect or not the outcome.

And maybe rename `nativeFilterEntry` to `nativeFilterEntryChecked`?

frontend/src/lib/utils/sns-proposals.utils.ts
lmuntaner: Or do we need to remember that "Generic" means specific to the SNS and "Native" means common to all SNSes. Is that the meaning?

Maybe we can ask the NNS team the reason for these names. It might help us remember it.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
mstrasinskis: Good point, https://github.com/dfinity/nns-dapp/pull/4233/files#diff-a7f8c28adec3867db2ed1ce1c0952391a2ff7e7e11bb0652a8ea0d69cd5d5433R927

frontend/src/lib/utils/sns-proposals.utils.ts
mstrasinskis: Ok, after reading the comments in `ic` repo it starts to make more sense. It's all about the target canisters:
- `native` are predefined functions that can be executed only on it's SNS canisters
- `generic` can execute a method outside the SNS canisters (any canister)

mstrasinskis: I would suggest to keep native/generic in the code and use "specific" in the labels only (copied from the dashboard). The comments were updated next to `isSnsNativeNervousSystemFunction` and `isSnsGenericNervousSystemFunction`.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
```diff
@@ -830,6 +832,127 @@ describe("sns-proposals utils", () => {
...
+      ).toBe(undefined);
+    });
+
+    it("should exclude all except selected native ns function", () => {
+      expect(
+        toExcludeTypeParameter({
+          filter: [nativeFilterEntry1Checked, nativeFilterEntry2Checked],
+          snsFunctions,
+        })
+      ).toStrictEqual([genericNsFunctionId]);
```
lmuntaner, line 924: Shouldn't the `genericNsFunction` be in the filter array?

Or change the description of the test to say that functions not present in the filter are considered as unchecked and therefore excluded.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
```diff
@@ -830,6 +832,127 @@ describe("sns-proposals utils", () => {
...
+              checked: false,
+            },
+            {
+              ...allGenericProposalTypesFilterEntryChecked,
+              checked: false,
+            },
+          ],
+          snsFunctions,
+        })
+      ).toStrictEqual([nativeNsFunctionId2, genericNsFunctionId]);
```
lmuntaner, line 943: Could you use `nativeFilterEntry2Checked.id` and `genericNsFunctionId.id`?

It would make it clearer in my opinion.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
```diff
@@ -830,6 +832,127 @@ describe("sns-proposals utils", () => {
...
+    });
+
+    it('should exclude generic when "All $snsName specific proposals" is selected', () => {
+      expect(
+        toExcludeTypeParameter({
+          filter: [allGenericProposalTypesFilterEntryChecked],
+          snsFunctions,
+        })
+      ).toStrictEqual([nativeNsFunctionId1, nativeNsFunctionId2]);
```
lmuntaner, line 952: Ditto, shouldn't this be in the filter as not checked?

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
mstrasinskis: That is a good point. Even though missing elements are treated as unselected, it is better to include them in the list to more accurately reflect the actual state of the data.

frontend/src/tests/lib/utils/sns-proposals.utils.spec.ts
mstrasinskis: Thanks, switched to functions.

lmuntaner (approved):
Thanks!!! Looks much better and even understandable despite the convoluted logic needed.

PR #4241, opened 2024-01-19
Use page objects in NnsStakeNeuronModal.spec.ts

dskloetd:
# Motivation

Make the tests easier to read and maintain.

# Changes

1. Move some test setup to the top-level `beforeEach` because other tests also depend on it and would fail when run in isolation.
2. Use page objects for the tests.
3. Add some additional expectations to make sure we are testing what we think we're testing.

# Tests

only

# Todos

not necessary


lmuntaner (approved):
That's great, thanks!

PR #4243, opened 2024-01-19
Clean up SnsNeuronsFooter.spec.ts

dskloetd:
# Motivation

# Changes

1. Remove existing setup from `beforeEach` because it wasn't used.
2. Replace setup inside the test by instead of mocking the store, filling the real stores, and put this setup in `beforeEach`.
3. Use page object.

# Tests

only tests

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #4246, opened 2024-01-19
Use page objects in SnsNeurons.spec.ts

dskloetd:
# Motivation

# Changes

1. Use page objects
2. Remove some redundant tests that check for static content that doesn't depend on logic.

# Tests

yes

# Todos

not necessary


lmuntaner (approved):
LGTM!

PR #4248, opened 2024-01-19
GIX-2080: Enable tokens flag for sns-governance e2e test

lmuntaner:
# Motivation

Enable tokens feature flat for mainnet.

In this PR, set flag to true for sns-governance test.

# Changes

* Set tokens flag to true in sns-governance.spec and change the flow to adapt to the tokens page.
* Add choosing the dropdown option in `getSnsTokens`. This worked before because it was in the same universe as the SNS and was initially selected.
* Add helper method `clickOnSnsUniverse` to SelectUniverseList.
* Add method to get SNS rows in TokensTable PO and helper constructor in TokensTableRow PO.

# Tests

Only test changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/tests/e2e/sns-governance.spec.ts
+++ b/frontend/src/tests/e2e/sns-governance.spec.ts
@@ -31,9 +31,10 @@ test("Test SNS governance", async ({ page, context }) => {
   expect(snsProjectName).toMatch(/[A-Z]{5}/);
 
   step("Acquire tokens");
-  await appPo.getSnsTokens({ amount: 20, name: snsProjectName });
+  const askedAmount = 20;
+  await appPo.getSnsTokens({ amount: askedAmount, name: snsProjectName });
 
-  expect(await snsUniverseRow.getBalanceNumber()).toEqual("20.00");
+  expect(await snsUniverseRow.getBalanceNumber()).toEqual(askedAmount);
 
   step("Stake a neuron");
   await appPo.goToNeurons();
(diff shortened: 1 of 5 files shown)


lmuntaner:
* `main`
  * **PR #4248** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4248?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4248?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

frontend/src/tests/e2e/sns-governance.spec.ts
```diff
@@ -1,45 +1,47 @@
...
-      .getMainAccountCardPo()
-      .getBalance()
-  ).toEqual("20.00");
+  expect(await snsUniverseRow.getBalanceNumber()).toEqual(askedAmount);
 
   step("Stake a neuron");
   await appPo.goToNeurons();
+
+  await appPo.openUniverses();
+  await appPo.getSelectUniverseListPo().clickOnSnsUniverse(snsProjectName);
```
dskloetd, line 43: Why does clicking on the universe card as before no longer work?

lmuntaner: Before it was getting the name from the selector by getting the first card, storing the name and clicking on it.

But now I get the name from the first SNS in the tokens table.

I don't want to rely that the first universe will be the same.

I could have also removed the check of the SNS in the accounts page and move to the neurons earlier. But that kind of changed the coveraga and I wanted to keep it as similar as possible.

dskloetd: It wasn't really in the accounts page but rather in the side bar, which used to be the same in the accounts page and the neurons page. So to me this feels less similar. But it's fine with me.

dskloetd (approved):

frontend/src/tests/e2e/sns-governance.spec.ts
lmuntaner: There is no sidebar in the tokens page though.

PR #4264, opened 2024-01-22
GIX-2080: Enable tokens page in multi-tab e2e test

lmuntaner:
# Motivation

Enable tokens page in mainnet.

In this PR, enable the flag for the multi-tag e2e test.

# Changes

* Enable tokens page for multi-tab e2e and fix new expectations.

# Tests

Only test changes.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/tests/e2e/multi-tab-auth.spec.ts
+++ b/frontend/src/tests/e2e/multi-tab-auth.spec.ts
@@ -19,6 +19,13 @@ const expectSignedInAccountsPage = async (appPo: AppPo) => {
   expect(await appPo.getSignInPo().isPresent()).toBe(false);
 };
 
+const expectSignedInTokensPage = async (appPo: AppPo) => {
+  await appPo.getTokensPo().getTokensPagePo().waitFor();
+  expect(await appPo.getTokensPo().getSignInTokensPagePo().isPresent()).toBe(
+    false
+  );
+};
+
 test("Test multi-tab auth", async ({ page: page1, context }) => {
   await page1.goto("/accounts");
   await expect(page1).toHaveTitle("My ICP Tokens / NNS Dapp");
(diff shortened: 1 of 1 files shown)


lmuntaner:
* `main`
  * **PR #4264** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4264?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4264?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

frontend/src/tests/e2e/multi-tab-auth.spec.ts
```diff
@@ -40,9 +50,9 @@ test("Test multi-tab auth", async ({ page: page1, context }) => {
   await page2.reload();
   await expectSignedInAccountsPage(appPo2);
 
-  // When signed in, the landing page redirects to the accounts page.
+  // When signed in, the landing page shows the tokens page.
   await page1.goto("/");
-  await expectSignedInAccountsPage(appPo1);
+  await appPo1.getTokensPo().waitFor();
```
dskloetd, line 55: Why did you change this line but not the previous line with `expectSignedInAccountsPage` or rather the body of `expectSignedInAccountsPage` itself?

Oh, I see, this is a different path.
Can you change the path instead?
I think it's more important that we use a consistent way to check the the user is still signed in, than to use a different path.
(And wouldn't `getTokensPo()` be present when not signed in as well?)

lmuntaner: Yes, good point. I added a helper like the others and added the expect there.

lmuntaner:

frontend/src/tests/e2e/multi-tab-auth.spec.ts
dskloetd: I don't see the point in using different URLs but I do think there is value in using the same helper in both cases, so I'd prefer just changing the URL instead of having an extra helper.

dskloetd (approved):

frontend/src/tests/e2e/multi-tab-auth.spec.ts
lmuntaner: Discussed offline.

PR #4269, opened 2024-01-22
GIX-2060: Remove deprecated transactionsFeeStore

lmuntaner:
# Motivation

Remove deprecated `transactionsFeesStore`.

In this PR, remove the usage in the derived stores that returns the fee of the ICP.

# Changes

* `mainTransactionFeeStore` uses `tokensStore`.
* `mainTransactionFeeE8sStore` uses `tokensStore`.
* `mainTransactionFeeStoreAsToken` uses `tokensStore`.

# Tests

* Adapt `mainTransactionFeeStoreAsToken` spec file to use tokensStore.
* Adapt `mainTransactionFeeStore` test case to use tokensStore.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/derived/main-transaction-fee.derived.ts
+++ b/frontend/src/lib/derived/main-transaction-fee.derived.ts
@@ -1,8 +1,16 @@
-import { transactionsFeesStore } from "$lib/stores/transaction-fees.store";
-import { ICPToken, TokenAmount } from "@dfinity/utils";
+import { OWN_CANISTER_ID_TEXT } from "$lib/constants/canister-ids.constants";
+import { NNS_TOKEN_DATA } from "$lib/constants/tokens.constants";
+import { tokensStore } from "$lib/stores/tokens.store";
+import { TokenAmount, isNullish } from "@dfinity/utils";
 import { derived } from "svelte/store";
 
-export const mainTransactionFeeStoreAsToken = derived(
-  transactionsFeesStore,
-  ($store) => TokenAmount.fromE8s({ amount: $store.main, token: ICPToken })
-);
+export const mainTransactionFeeStoreAsToken = derived(tokensStore, ($store) => {
+  const icpToken = $store[OWN_CANISTER_ID_TEXT].token;
+  if (isNullish(icpToken)) {
+    return TokenAmount.fromE8s({
+      amount: NNS_TOKEN_DATA.fee,
+      token: NNS_TOKEN_DATA,
+    });
+  }
+  return TokenAmount.fromE8s({ amount: icpToken.fee, token: icpToken });
+});
--- a/frontend/src/lib/derived/main-transaction-fee.derived.ts
+++ b/frontend/src/lib/derived/main-transaction-fee.derived.ts
@@ -1,5 +1,4 @@
 import { OWN_CANISTER_ID_TEXT } from "$lib/constants/canister-ids.constants";
-import { NNS_TOKEN_DATA } from "$lib/constants/tokens.constants";
 import { tokensStore } from "$lib/stores/tokens.store";
 import { TokenAmount, isNullish } from "@dfinity/utils";
 import { derived } from "svelte/store";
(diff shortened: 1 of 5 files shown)


lmuntaner:
* `main`
  * **PR #4269** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4269?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4269?utm_source=stack-comment).

frontend/src/lib/derived/main-transaction-fee.derived.ts
```diff
@@ -1,8 +1,16 @@
...
+import { TokenAmount, isNullish } from "@dfinity/utils";
 import { derived } from "svelte/store";
 
-export const mainTransactionFeeStoreAsToken = derived(
-  transactionsFeesStore,
-  ($store) => TokenAmount.fromE8s({ amount: $store.main, token: ICPToken })
-);
+export const mainTransactionFeeStoreAsToken = derived(tokensStore, ($store) => {
+  const icpToken = $store[OWN_CANISTER_ID_TEXT].token;
+  if (isNullish(icpToken)) {
```
dskloetd, line 9: Is this necessary?

frontend/src/lib/stores/transaction-fees.store.ts
```diff
@@ -101,12 +104,14 @@ export const transactionsFeesStore = initTransactionFeesStore();
...
+    Number($store[OWN_CANISTER_ID_TEXT]?.token.fee) ??
+    Number(NNS_TOKEN_DATA.fee)
 );
 
-export const mainTransactionFeeE8sStore = derived(
-  transactionsFeesStore,
-  ($store) => $store.main
+export const mainTransactionFeeE8sStore: Readable<bigint> = derived(
+  tokensStore,
+  ($store) => $store[OWN_CANISTER_ID_TEXT]?.token.fee ?? NNS_TOKEN_DATA.fee
```
dskloetd, line 116: Is the fallback necessary?

frontend/src/lib/stores/transaction-fees.store.ts
```diff
@@ -101,12 +104,14 @@ export const transactionsFeesStore = initTransactionFeesStore();
...
  * @deprecated prefer mainTransactionFeeE8sStore to use e8s for amount of tokens instead of Number.
  */
-export const mainTransactionFeeStore = derived(
-  transactionsFeesStore,
-  ($store) => Number($store.main)
+export const mainTransactionFeeStore: Readable<number> = derived(
+  tokensStore,
+  ($store) =>
+    Number($store[OWN_CANISTER_ID_TEXT]?.token.fee) ??
+    Number(NNS_TOKEN_DATA.fee)
```
dskloetd, line 111: * `Number(null) == 0`
* `0 ?? 5 == 0`

Did you mean `Number($store[OWN_CANISTER_ID_TEXT]?.token.fee ?? NNS_TOKEN_DATA.fee))` instead?

But if we have the data hardcoded anyway, why do we need the store?

frontend/src/lib/derived/main-transaction-fee.derived.ts
lmuntaner: For TS reasons yes. But not really because the `$store` will have the hardcoded value.

I thought it was better to return the hard-coded value than raise an error here though.

frontend/src/lib/stores/transaction-fees.store.ts
lmuntaner: Ditto, `$store[OWN_CANISTER_ID_TEXT]` is not defined according to TS.

lmuntaner:

frontend/src/lib/stores/transaction-fees.store.ts
dskloetd: > Ideally we would be fetching the ICP fee from the canister and storing it.

Why? If we trust the hard coded value to be correct, why even add the complication of a store?
If we don't trust it to be correct, we should not fall back on it.

dskloetd: I think hard coding is fine.
I think putting it in a store is fine.
But this mix here doesn't really make sense to me.

lmuntaner: Throwing makes more sense to you then?

dskloetd: If we believe it not being present is a bug, yes.

frontend/src/lib/derived/main-transaction-fee.derived.ts
lmuntaner: Error raised.

dskloetd (approved):

PR #4271, opened 2024-01-22
GIX-2080: Fix design.spec e2e test

lmuntaner:
# Motivation

Enable tokens page in mainnet.

In this PR, fix the design e2e test. The screenshot was taken from `/accounts` instead of the expected `/tokens`.

I have also changed it to use `/` instead of `/tokens`.

# Changes

* Go again to `/` after setting the flag.
* Replace content with XXX before the screenshot.

# Tests

Only test changes.

# Todos

Not necessary.


lmuntaner:
* `main`
  * **PR #4271** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4271?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4271?utm_source=stack-comment).

lmuntaner:
@dskloetd please review

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -89,9 +89,12 @@
   <div role="cell" class={`mobile-row-cell left-cell ${cellAreaName(0)}`}>
     <span class="mobile-only">{$i18n.tokens.balance_header}</span>
     {#if userTokenData.balance instanceof UnavailableTokenAmount}
-      <span data-tid="token-value-label"
-        >{`-/- ${userTokenData.balance.token.symbol}`}</span
-      >
+      <!-- The label is within its own element because it will be replaced in screenshot testing. -->
+      <div data-tid="token-value-label" class="balance-unavailable">
+        <!-- The two spans are needed so that when we get the text from "token-value-label" there is a space between amount and token.  -->
+        <span>{`-/-`}</span>
```
dskloetd, line 95: You can use normal double quotes now that there are no variables in this string.

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -89,9 +89,12 @@
   <div role="cell" class={`mobile-row-cell left-cell ${cellAreaName(0)}`}>
     <span class="mobile-only">{$i18n.tokens.balance_header}</span>
     {#if userTokenData.balance instanceof UnavailableTokenAmount}
-      <span data-tid="token-value-label"
-        >{`-/- ${userTokenData.balance.token.symbol}`}</span
-      >
+      <!-- The label is within its own element because it will be replaced in screenshot testing. -->
+      <div data-tid="token-value-label" class="balance-unavailable">
+        <!-- The two spans are needed so that when we get the text from "token-value-label" there is a space between amount and token.  -->
```
dskloetd, line 94: Can you just use an explicit space similar to line 39 of AmountDisplay? It can probably even be inside `{"-/- "}`, right?

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -89,9 +89,12 @@
...
+        <span>{`-/-`}</span>
+        <span data-tid="label">{userTokenData.balance.token.symbol}</span>
```
dskloetd, line 96: Did you mean `symbol` instead of `label`?

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -235,6 +238,15 @@
     align-items: center;
   }
 
+  .balance-unavailable {
+    // display: inline-flex;
+    // gap: var(--padding-0_5x);
```
dskloetd, line 243: Why is this here?

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -235,6 +238,15 @@
...
+    display: inline-grid;
```
dskloetd, line 244: Do we really need a grid for something so simple?

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -235,6 +238,15 @@
...
+    grid-template-columns: repeat(2, auto);
+    grid-gap: var(--padding-0_5x);
```
dskloetd, line 246: Doesn't having the space already give us the distance we need? I feel like I'm missing something.

frontend/src/tests/e2e/screenshots/design.spec.ts-Design-Login-1-Google-Chrome-linux.png
dskloetd, line 1: Is it really worth going through all this effort to add those few rows, which we then partly block out?
It doesn't look like the extra rows add any extra styling that we potentially might not catch breakages in without them.

dskloetd, line 1: I see changes in multiple tests. How come only 1 screenshot needed to be updated?

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: Done

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: DOne

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: It was from a test that I did and forgot to remove. The space does the trick, Therefore, no need for any specific style.

Removed.

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: Removed.

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: Yes. The space was not in the string itself and the blank space between spans didn't show a space in the html. Only when getting the text from it.

frontend/src/tests/e2e/screenshots/design.spec.ts-Design-Login-1-Google-Chrome-linux.png
lmuntaner: That also requires extra effort, no? We would need to remove the rows below it.

Whatever we do, we need to manipulate the DOM.

Ideally we would set the same SNS names.

dskloetd: I'm must be seriously missing something.
In the diff of the screenshot, I see on the left a screenshot without all the extra rows.
That's the screenshot how it is right now, isn't it?
Why is it like that?
Why can't we keep that?

lmuntaner: Ahh, I see. I mentioned in the PR description:

> In this PR, fix the design e2e test. The screenshot was taken from /accounts instead of the expected /tokens.

It was taking a screenshot of `/accounts`, which is only ICP.

dskloetd: Is it important that the screenshot is taken from `/tokens`? Why don't we continue to take it from `/accounts`, given that the only difference doesn't really add important styling that we need to test and is causing a hassle?

lmuntaner: Done.

lmuntaner:
@dskloetd ready for another review.

The screenshot didn't change because it's the same as in `/accounts`.

The changes are to make it more stable by waiting for the proper things and not relying on the redirection.

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
```diff
@@ -89,9 +89,13 @@
   <div role="cell" class={`mobile-row-cell left-cell ${cellAreaName(0)}`}>
     <span class="mobile-only">{$i18n.tokens.balance_header}</span>
     {#if userTokenData.balance instanceof UnavailableTokenAmount}
-      <span data-tid="token-value-label"
-        >{`-/- ${userTokenData.balance.token.symbol}`}</span
-      >
+      <!-- The label is within its own element because it will be replaced in screenshot testing. -->
```
dskloetd, line 92: Is this still true?

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: reverted

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: reverted

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: reverted

frontend/src/lib/components/tokens/TokensTable/TokensTableRow.svelte
lmuntaner: No, and I can keep it as it was before. I reverted.

lmuntaner:

dskloetd (approved):
Thanks!

PR #4215, opened 2024-01-17
Refactor: Migrate to TokenAmountV2

lmuntaner:
# Motivation

Migrate to TokenAmountV2 which supports decimals different than 8.

In this PR, I change `snsSelectedTransactionFeeStore` to return a `TokenAmountV2`.

# Changes

* Set `snsSelectedTransactionFeeStore` to return `TokenAmountV2`.
* Change dependencies accordingly.
* Remove unnecessary `toTokenAmountV2`.

# Tests

* Change mock accordingly.
* Fix test data.

# Todos

- [ ] Add entry to changelog (if necessary).

Not necessary.

PR #4272, opened 2024-01-22
GIX-2150: Use TokenAmount for fee in SnsNeuronDetail

lmuntaner:
# Motivation

Use TokenAmountV2 instead of TokenAmount.

In this PR, change the type of the fee in SnsNeuronDetail to TokenAmount from bigint. This will allow us to then migrate to TokenAmountV2.

The following changes were triggered in [this comment]([url](https://github.com/dfinity/nns-dapp/pull/4215#discussion_r1455261670)).

# Changes

* Set local variable `transactionFee` type to `TokenAmount` instead of `E8s`.
* Fix the necessary props in the way.

# Tests

* Fix the tests with the new prop type and name.

# Todos

Not necessary.

### DIFF
--- a/frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
@@ -9,10 +9,10 @@
   import { replacePlaceholders } from "$lib/utils/i18n.utils";
   import { formatTokenE8s } from "$lib/utils/token.utils";
   import { i18n } from "$lib/stores/i18n";
-  import type { TokenAmount } from "@dfinity/utils";
+  import type { TokenAmountV2 } from "@dfinity/utils";
 
   export let neuron: SnsNeuron;
-  export let fee: TokenAmount;
+  export let fee: TokenAmountV2;
 
   let enoughMaturity: boolean;
   $: enoughMaturity = hasEnoughMaturityToDisburse({ neuron, fee: fee.toE8s() });
@@ -25,7 +25,7 @@
           $i18n.neuron_detail.disburse_maturity_disabled_tooltip_non_zero,
           {
             $amount: formatTokenE8s({
-              value: minimumAmountToDisburseMaturity(fee.toE8s()),
+              value: minimumAmountToDisburseMaturity(fee.toUlps()),
             }),
           }
         )
--- a/frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
@@ -13,12 +13,11 @@
 
   export let neuron: SnsNeuron;
   export let fee: TokenAmountV2;
-  feeE8s;
 
   let enoughMaturity: boolean;
   $: enoughMaturity = hasEnoughMaturityToDisburse({
     neuron,
-    fee: fee.toE8s(),
+    feeE8s: fee.toE8s(),
   });
 
   let disabledText: string | undefined = undefined;
--- a/frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
+++ b/frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
@@ -15,7 +15,10 @@
   export let fee: TokenAmountV2;
 
   let enoughMaturity: boolean;
-  $: enoughMaturity = hasEnoughMaturityToDisburse({ neuron, fee: fee.toE8s() });
+  $: enoughMaturity = hasEnoughMaturityToDisburse({
+    neuron,
+    fee: fee.toUlps(),
+  });
 
   let disabledText: string | undefined = undefined;
   $: disabledText = !enoughMaturity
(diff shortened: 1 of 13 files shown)


lmuntaner:
* `main`
  * **PR #4272** <a href="https://app.graphite.dev/github/pr/dfinity/nns-dapp/4272?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/dfinity/nns-dapp/4272?utm_source=stack-comment).

dskloetd:
What is the benefit of going to `TokenAmount` first instead of directly to `TokenAmountV2`?

lmuntaner:
> What is the benefit of going to `TokenAmount` first instead of directly to `TokenAmountV2`?

I didn't want to change `snsSelectedTransactionFeeStore` yet. But I forgot that I can use `toTokenAmountV2`. Therefore, all components use `TokenAmountV2` now.

Thanks!

frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
```diff
@@ -9,12 +9,16 @@
...
   import { i18n } from "$lib/stores/i18n";
+  import type { TokenAmountV2 } from "@dfinity/utils";
 
   export let neuron: SnsNeuron;
-  export let feeE8s: bigint;
+  export let fee: TokenAmountV2;
 
   let enoughMaturity: boolean;
-  $: enoughMaturity = hasEnoughMaturityToDisburse({ neuron, feeE8s });
+  $: enoughMaturity = hasEnoughMaturityToDisburse({
```
dskloetd, line 18: `hasEnoughMaturityToDisburse` only supports e8s. If we pass in generic ulps we'd have to convert it to e8s or be sure that the ulps are e8s.

I've been thinking it might make sense to add `toE8s` to `TokenAmountV2` (which converts to actual e8s regardless of the number of decimals) for cases like this. What do you think?

frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
```diff
@@ -24,7 +28,7 @@
           $i18n.neuron_detail.disburse_maturity_disabled_tooltip_non_zero,
           {
             $amount: formatTokenE8s({
-              value: minimumAmountToDisburseMaturity(feeE8s),
+              value: minimumAmountToDisburseMaturity(fee.toUlps()),
```
dskloetd, line 31: This is passed to `formatTokenE8s` so it's not safe to use generic ulps.
How about changing `minimumAmountToDisburseMaturity` to take and return `TokenAmountV2`?

frontend/src/lib/components/sns-neuron-detail/actions/SplitSnsNeuronButton.svelte
```diff
@@ -29,7 +28,7 @@
   let enoughStakeToSplit: boolean;
   $: enoughStakeToSplit = hasEnoughStakeToSplit({
```
dskloetd, line 29: `hasEnoughStakeToSplit` requires e8s.

frontend/src/lib/components/sns-neuron-detail/actions/SplitSnsNeuronButton.svelte
```diff
@@ -51,7 +50,7 @@
       {
         $amount: formatTokenE8s({
```
dskloetd, line 51: `formatTokenE8s` requires e8s.

frontend/src/lib/utils/sns-neuron.utils.ts
```diff
@@ -513,12 +513,11 @@ export const hasEnoughMaturityToStake = (
  */
 export const hasEnoughMaturityToDisburse = ({
   neuron: { maturity_e8s_equivalent },
-  feeE8s,
+  fee,
```
dskloetd, line 516: Please keep this as `feeE8s` as long as it's compared to something that's also e8s.

frontend/src/lib/components/sns-neuron-detail/actions/SnsDisburseMaturityButton.svelte
lmuntaner: I updated the PR to use only toE8s because all the places here are either maturity which is always e8s or formatted as e8s.

lmuntaner:
@dskloetd could you take another look? Thanks!

frontend/src/lib/components/sns-neuron-detail/actions/SplitSnsNeuronButton.svelte
```diff
@@ -10,15 +10,14 @@
...
+  import type { Token, TokenAmountV2 } from "@dfinity/utils";
   import VestingTooltipWrapper from "../VestingTooltipWrapper.svelte";
 
   export let neuron: SnsNeuron;
   export let parameters: SnsNervousSystemParameters;
-  export let transactionFee: E8s;
+  export let transactionFee: TokenAmountV2;
   export let token: Token;
 
   let neuronMinimumStake: bigint;
```
dskloetd, line 23: How about renaming this to put `E8s` in the name to make it more obvious?

dskloetd (approved):

frontend/src/lib/components/sns-neuron-detail/actions/SplitSnsNeuronButton.svelte
lmuntaner: Good point, done

frontend/src/lib/utils/sns-neuron.utils.ts
lmuntaner: done

PR #4273, opened 2024-01-22
feat: Partition schema getter and setter

bitdivine:
# Motivation
The memory partitioning scheme needs labels to identify the scheme with which data has been persisted.

# Changes
- Add partition scheme label getter, setter, and a constructor that initializes partitions with a given schema.

# Tests
Unit tests are included.

# Todos


rs/backend/src/state/partitions.rs
```diff
@@ -83,6 +85,34 @@ impl Partitions {
...
+        let current_pages = memory.size();
+        if current_pages < min_pages {
+            memory.grow(min_pages - current_pages);
+        }
+        memory.write(offset, bytes)
+    }
+
+    /// Reads a buffer, if possible.
+    #[cfg(test)]
+    pub fn try_read(&self, memory_id: MemoryId, offset: u64, buffer: &mut [u8]) -> Result<(), &'static str> {
```
bitdivine, line 106: `read_exact` is actually more idiomatic.  It matches similar functionality in the Read trait.  I'll rename.

bitdivine: Done

rs/backend/src/state/partitions.rs
```diff
@@ -83,6 +85,34 @@ impl Partitions {
     pub fn into_memory(self) -> DefaultMemoryImpl {
         self.memory
     }
+
+    /// Writes, growing the memory if necessary.
+    #[cfg(test)]
+    pub fn growing_write(&self, memory_id: MemoryId, offset: u64, bytes: &[u8]) {
+        let memory = self.get(memory_id);
+        let min_pages = u64::try_from(
+            (usize::try_from(offset).unwrap() + bytes.len() + WASM_PAGE_SIZE_IN_BYTES - 1) / WASM_PAGE_SIZE_IN_BYTES,
```
bitdivine, line 94: I think there is div_ceil.  I'll use that instead.

rs/backend/src/state/partitions.rs
```diff
@@ -83,6 +85,34 @@ impl Partitions {
...
+    }
+
+    /// Reads the exact number of bytes needed to fill `buffer`.
+    #[cfg(test)]
+    pub fn read_exact(&self, memory_id: MemoryId, offset: u64, buffer: &mut [u8]) -> Result<(), &'static str> {
+        let memory = self.get(memory_id);
+        let bytes_in_memory =
+            memory.size() * u64::try_from(WASM_PAGE_SIZE_IN_BYTES).expect("Wasm page size is too large");
+        if offset + u64::try_from(buffer.len()).unwrap() >= bytes_in_memory {
+            return Err("Insufficient memory to read");
```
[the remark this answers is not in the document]
bitdivine: Good idea.  Added more detail, including offset and buffer len.  Added a unit tests as well, as there weren't any for these functions; they were exercised in tests only indirectly.

rs/backend/src/state/partitions.rs
```diff
@@ -83,6 +85,29 @@ impl Partitions {
...
+        }
+        memory.write(offset, bytes)
+    }
+
+    /// Reads the exact number of bytes needed to fill `buffer`.
+    pub fn read_exact(&self, memory_id: MemoryId, offset: u64, buffer: &mut [u8]) -> Result<(), &'static str> {
+        let memory = self.get(memory_id);
+        let bytes_in_memory =
+            memory.size() * u64::try_from(WASM_PAGE_SIZE_IN_BYTES).expect("Wasm page size is too large");
+        if offset + u64::try_from(buffer.len()).unwrap() >= bytes_in_memory {
```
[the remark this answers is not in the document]
bitdivine: Done.  When someone passes in a buffer > 2**64 long I will buy you a beer. :smile:  It could happen if memory has holes in it or is defined by a generator, so the memory doesn't have to be allocated, so I am not entirely safe from having to pour a glass of Schwarzer Kristall for that dark day. :wink:

PR #4276, opened 2024-01-22
Stake neuron to vote

mstrasinskis:
# Motivation

Update voting block to provide more information for the new users how to start with voting.

# Changes

- StakeNeuronToVote component.
- Adopt/Reject buttons are visible also when user doesn't has a neuron to vote.

# Tests

- StakeNeuronToVote displays info and navigation works (nns/sns)
- Adopt/Reject buttons are visible

# Todos

# Screenshots

image

image

image


frontend/src/lib/components/proposal-detail/VotingCard/StakeNeuronToVote.svelte
```diff
@@ -0,0 +1,98 @@
...
+      $token: token,
+    });
+
+  const gotoNeurons = () => goto($neuronsPathStore);
+</script>
+
+<TestIdWrapper testId="stake-neuron-to-vote-component">
+  {#if nonNullish(label)}
+    <div class="container" in:fade>
+      <Collapsible
```
lmuntaner, line 35: The UI is different than the `Collapsible` then? Can't we have the same one?

frontend/src/tests/lib/components/proposal-detail/VotingCard/StakeNeuronToVote.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+import { runResolvedPromises } from "$tests/utils/timers.test-utils";
+import { SnsSwapLifecycle } from "@dfinity/sns";
+import { render } from "@testing-library/svelte";
+import { get } from "svelte/store";
+
+describe("StakeNeuronToVote", () => {
+  const renderComponent = async () => {
+    const { container } = render(StakeNeuronToVote);
+
+    await runResolvedPromises();
```
lmuntaner, line 19: Is this needed? The component doesn't make any api calls nor rely on anything async, right?

frontend/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
```diff
@@ -146,6 +163,22 @@ describe("VotingCard", () => {
       expect(await po.getIneligibleNeurons().isPresent()).toBe(false);
     });
 
+    it('should not display "stake a neuron" when user has neurons', async () => {
+      const po = await renderComponent();
+
+      expect(await po.getStakeNeuronToVotePo().isPresent()).toBe(false);
```
lmuntaner, line 169: Do we want a test where this is present as well?

lmuntaner (commented):
Looking good! I added a few comments.

frontend/src/lib/components/proposal-detail/VotingCard/StakeNeuronToVote.svelte
mstrasinskis: It's the same UI that is used in the voting block. We can't use the default `Collapsibles` because normally it has an info on the right side:
image

frontend/src/lib/components/proposal-detail/VotingCard/StakeNeuronToVote.svelte
```diff
@@ -0,0 +1,98 @@
...
+            on:click|stopPropagation={toggleContent}
+            data-tid="expand-icon"
+          >
+            <IconExpandCircleDown />
+          </button>
+        </div>
+        <p class="description" data-tid="stake-neuron-description">
+          {$i18n.proposal_detail__vote.no_neurons_description}
+        </p>
+        <button
```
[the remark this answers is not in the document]
mstrasinskis: I thought initially about using an anchor, but in other places in the `dapp` for this kind of navigation the button was used and there is no like-a-button styling for the `a` tag available. Not sure it will bring any value to make it openable in a new tab. Please let me know if you have another opinion.

frontend/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
mstrasinskis: It's done here: https://github.com/dfinity/nns-dapp/pull/4276/files#diff-36701b8cfc7831da244fde8c8f869ba5812f05df269691c58fafb6a8aaf307efR179

frontend/src/lib/components/proposal-detail/VotingCard/StakeNeuronToVote.svelte
lmuntaner: Yes, but then should create a component to be shared in the voting pages?

frontend/src/lib/components/proposal-detail/VotingCard/StakeNeuronToVote.svelte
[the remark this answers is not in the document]
lmuntaner: Definitely not to open in a new tab. But why wouldn't a link be used for paths within the same application?

I have precisely a PR to add button-like anchor tags: https://github.com/dfinity/gix-components/pull/362

I used it in the Buy ICP modal. That opened a new tab, so it made even more sense I'd say.

Where else do we use it like this? The MenuItems for example are `<a>`.

frontend/src/tests/lib/components/proposal-detail/VotingCard/VotingCard.spec.ts
lmuntaner: oops, I missed it sorry!

lmuntaner (approved):
Approving already. I don't have a strong opinion about anchor or button, it was more a consideration to be consistent in the application.

frontend/src/lib/components/proposal-detail/VotingCard/StakeNeuronToVote.svelte
mstrasinskis: The task was created.

frontend/src/lib/components/proposal-detail/VotingCard/StakeNeuronToVote.svelte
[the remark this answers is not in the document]
mstrasinskis: I noticed it mostly in the back buttons, which utilize the same navigation logic. Since you've already prepared the pr for the `a` tag, let's use it! Thank you!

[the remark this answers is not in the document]
lmuntaner: Sounds good

PR #4278, opened 2024-01-23
preparation: Implement state traits manually

bitdivine:
# Motivation
We would like to restrict changes to stable memory, so that such changes are made only as part of updates to the state structure.  We will do this by adding a reference to memory in state and making changes only via that pointer.   However we will need to implement traits such as Debug and Eq manually.

# Changes
- Implement Debug manually.
  - Alternative considered: Implement the Debug trait for stable memory.  Rejected as the Debug printout is already much too verbose.  It seems better to limit the debug printout to what is actually useful.
- Implement `Default` manually, so that the default can later include the stable memory.
- Implement `Eq` manually, as I would like to ignore changes to the stable memory pointer when comparing state e.g. pre and post upgrade.  I would like to restrict access to memory but carry on not caring how the state is represented, as long as the meaningful content of the state is the same.

# Tests
- Tests are unchanged.

# Todos

### DIFF
--- a/rs/backend/src/state.rs
+++ b/rs/backend/src/state.rs
@@ -13,8 +13,6 @@ pub mod partitions;
 #[cfg(test)]
 pub mod tests;
 
-#[derive(Default, Debug)]
-#[cfg_attr(test, derive(Eq, PartialEq))]
 pub struct State {
     // NOTE: When adding new persistent fields here, ensure that these fields
     // are being persisted in the `replace` method below.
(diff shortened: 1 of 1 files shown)


rs/backend/src/state.rs
```diff
@@ -13,15 +13,49 @@ pub mod partitions;
...
     pub performance: RefCell<PerformanceCounts>,
+    // TODO: Take ownership of stable memory.
+}
+
+#[cfg(test)]
+impl PartialEq for State {
+    /// Compares essential content of two states for equality.
+    ///
+    /// Excluded from the comparison:
+    /// - Raw stable memory owned by the state.
```
przydatek, line 31: nit: maybe remove this comment as the state doesn't have raw stable memory yet

przydatek (approved):
LGTM, modulo two suggestions.

rs/backend/src/state.rs
bitdivine: Done.

rs/backend/src/state.rs
```diff
@@ -13,15 +13,49 @@ pub mod partitions;
...
+            performance: RefCell::new(PerformanceCounts::default()),
+        }
+    }
+}
+
+impl core::fmt::Debug for State {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "State {{")?;
+        writeln!(f, "  accounts: {:?}", &self.accounts_store.borrow())?;
+        writeln!(f, "}}")
```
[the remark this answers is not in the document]
bitdivine: Done.  I destructured self so that it is hard to accidentally forget a field and summarized the omitted fields.

PR #4285, opened 2024-01-23
Chore: Add reviewers to update PRs by CI

lmuntaner:
# Motivation

We have some nice workflows that create PRs to update dependencies. They are accumulating because there are no reviewers.

In this PR, add reviewers for the udpate workflows.

Example PR: https://github.com/dfinity/nns-dapp/pull/4284

# Changes

* Add `reviewers: mstrasinskis, dskloetd, lmuntaner` to the `update-*.yml` CI workflows.

# Tests

Not necessary.

# Todos

### DIFF
--- a/.github/workflows/update-aggregator.yml
+++ b/.github/workflows/update-aggregator.yml
@@ -35,6 +35,7 @@ jobs:
           committer: GitHub <noreply@github.com>
           author: gix-bot <gix-bot@users.noreply.github.com>
           base: main
+          reviewers: mstrasinskis, dskloetd, lmuntaner
           # Note: Please be careful when updating the add-paths field.  We have had the snsdemo committed by accident, with a pattern that matches nothing seemingly committing everything.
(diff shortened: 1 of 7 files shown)


lmuntaner:

.github/workflows/update-aggregator.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           committer: GitHub <noreply@github.com>
           author: gix-bot <gix-bot@users.noreply.github.com>
           base: main
+          reviewers: mstrasinskis, dskloetd, lmuntaner
```
dskloetd, line 38: I wonder if it makes sense to put this list in 1 place and reference it from there.
But approving already because it's not urgent.

dskloetd (approved):

.github/workflows/update-aggregator.yml
lmuntaner: I added the three of us for now, but ideally we should split the reviews.

PR #4289, opened 2024-01-23
Remove input error outline for dissolve delay below project minimum

dskloetd:
# Motivation

Requested in point 14 of https://www.notion.so/dfinityorg/Wording-Changes-4411020138ce4d09b6605ad1fdbb9555

When the dissolve delay is not enough to have voting power, we should an error message but we don't disable the "Set Delay" button.
Additionally to not disabling the button we should also not put a red outline on the input field.

# Changes

1. Add an extra prop `warningMessage` on `InputWithError`. Setting the warning message show the message similar to an error message but it does not cause the red outline.
2. Use the `warningMessage` prop for the warning about not having voting power.

# Tests

1. Use page objects for existing tests in `InputWithError.spec.ts`.
2. Add unit tests for `warningMessage`.
3. Add expects for the error outline in `SetDissolveDelay.spec.ts`.

# Todos

### DIFF
--- a/CHANGELOG-Nns-Dapp-unreleased.md
+++ b/CHANGELOG-Nns-Dapp-unreleased.md
@@ -27,6 +27,7 @@ proposal is successful, the changes it released will be moved from this file to
 #### Removed
 
 * ckBTC withdrawal account.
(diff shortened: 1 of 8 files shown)


lmuntaner (approved):
LGTM!

PR #4290, opened 2024-01-24
Update proposal rendering

sa-github-api:
# Motivation

# Changes

# Tests


dskloetd:
> Please check the API updates for any breaking changes that affect our code.
> Please check for new proposal types and add tests for them.

Instructions are not specific enough to be actionable.

lmuntaner:
> Please check the API updates for any breaking changes that affect our code.
> Please check for new proposal types and add tests for them.

I believe the new proposal types can be added in a separate PR. We can merge this.

And there are no breaking changes in the API we use.

@bitdivine can you confirm?

bitdivine:
For the backend, this looks safe to me.  I see no mention of the deleted `SeedAccount(s)` or `update_api_boundary_node_domain` in the frontend code either.  So I agree, this looks fine to me.

bitdivine (approved):

PR #4293, opened 2024-01-24
Do not hide vote in progress proposals

mstrasinskis:
# Motivation

Because "Show only proposals you can still vote for" can be unchecked or voting actions may take a few seconds, it is expected that the proposal will be accessible through navigation until it is complete. 

1. Start from nns proposal list page
2. Uncheck “`Show only proposals you can still vote for`" if checked
3. Start voting process on proposal with id N
4. Quickly click “Older” to navigate to N-1
5. Right after navigation click “Newer”

Expected navigation to proposal N, not N+1.

# Changes

- ignore the temporal voting status (from `vote registrations store`) in `is proposal hidden` logic. The proposal entry remains in the `voteRegistrationStore` only till the voting is done.
- drop `VoteRegistrationStatus.complete` since it's not used anymore (It's a relic that we still rely on it).

# Tests

- update relevant tests

# Todos

# Screenshots

Issue w/ selected checkbox
https://github.com/dfinity/nns-dapp/assets/98811342/8fc2bf51-3b71-4380-83b0-aa466743a7ee

Issue w/o selected checkbox
https://github.com/dfinity/nns-dapp/assets/98811342/f6b166a3-c6af-4d5c-9a27-55c12629cbc2

After the fix
https://github.com/dfinity/nns-dapp/assets/98811342/3c5a1183-6964-4882-875c-1524c5abf41f

### DIFF
--- a/frontend/src/lib/stores/vote-registration.store.ts
+++ b/frontend/src/lib/stores/vote-registration.store.ts
@@ -8,10 +8,7 @@ import type { Vote } from "@dfinity/nns";
 import type { SnsVote } from "@dfinity/sns";
 import { writable, type Readable } from "svelte/store";
 
-export type VoteRegistrationStatus =
-  | "vote-registration"
-  | "post-update"
-  | "complete";
+export type VoteRegistrationStatus = "vote-registration" | "post-update";
 
 export interface VoteRegistrationStoreEntry {
   status: VoteRegistrationStatus;
--- a/frontend/src/tests/lib/stores/vote-registration.store.spec.ts
+++ b/frontend/src/tests/lib/stores/vote-registration.store.spec.ts
@@ -44,25 +44,19 @@ describe("voting-store", () => {
 
       voteRegistrationStore.updateStatus({
         proposalIdString: voteA.proposalIdString,
-        status: "complete",
+        status: "post-update",
         canisterId: OWN_CANISTER_ID,
       });
 
       expect(
         get(voteRegistrationStore).registrations[OWN_CANISTER_ID.toText()][0]
           .status
-      ).toEqual("complete");
+      ).toEqual("post-update");
     });
 
     it("should remove completed voting items", () => {
       voteRegistrationStore.add({ ...voteA, canisterId: OWN_CANISTER_ID });
       voteRegistrationStore.add({ ...voteB, canisterId: OWN_CANISTER_ID });
-
-      voteRegistrationStore.updateStatus({
-        status: "complete",
-        proposalIdString: voteA.proposalIdString,
-        canisterId: OWN_CANISTER_ID,
-      });
 
       voteRegistrationStore.remove({
         proposalIdString: voteA.proposalIdString,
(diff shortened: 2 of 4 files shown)


lmuntaner:
Do you mean "Check" in?

> Uncheck “Show only proposals you can still vote for" if checked

After the vote is finished, it doesn't go to that one, right?

frontend/src/tests/lib/stores/vote-registration.store.spec.ts
```diff
@@ -44,26 +44,20 @@ describe("voting-store", () => {
...
       });
 
       expect(
         get(voteRegistrationStore).registrations[OWN_CANISTER_ID.toText()][0]
           .status
-      ).toEqual("complete");
+      ).toEqual("post-update");
     });
 
     it("should remove completed voting items", () => {
```
lmuntaner, line 57: Why isn't the status anymore to remove them?

frontend/src/lib/stores/vote-registration.store.ts
```diff
@@ -8,10 +8,7 @@ import type { Vote } from "@dfinity/nns";
 import type { SnsVote } from "@dfinity/sns";
 import { writable, type Readable } from "svelte/store";
 
-export type VoteRegistrationStatus =
-  | "vote-registration"
-  | "post-update"
-  | "complete";
+export type VoteRegistrationStatus = "vote-registration" | "post-update";
```
lmuntaner, line 11: Is this related to the PR?

This is not used anywhere else, right?

lmuntaner (commented):
A couple of comments but for my understanding.

frontend/src/tests/lib/stores/vote-registration.store.spec.ts
mstrasinskis: It is likely that this was an error that occurred during copy-pasting, as I did not find any need for a status within the "remove" implementation.

frontend/src/lib/stores/vote-registration.store.ts
mstrasinskis: It's not directly related to the fix. It was added only for the purpose of code-reviews economy. Should the pr be split?

mstrasinskis:
> Do you mean "Check" in?
> 
> > Uncheck “Show only proposals you can still vote for" if checked
> 
> After the vote is finished, it doesn't go to that one, right?

True that in the screenshots section there is a sample with selected checkbox “Show only proposals you can still vote for". But in case of unchecked button it was a bit worse because during voting the proposal is not reachable (via newer/older), but after the voting is complete the proposal is reachable again (added a video).

frontend/src/lib/stores/vote-registration.store.ts
lmuntaner: No need. It was for my understanding.

frontend/src/tests/lib/stores/vote-registration.store.spec.ts
lmuntaner: Ok, thanks for the explanation.

lmuntaner (approved):
Thanks for the explanations.

PR #4294, opened 2024-01-24
Add ModalPo base page object

dskloetd:
# Motivation

To avoid duplicate implementations of modal functionality in page objects and provide convenience methods to all modal page objects.

# Changes

1. Add a base class for modal page objects.
2. Extend `ModalPo` in modal page objects.
3. Remove duplicate modal functionality from modal page object subclasses.
4. Update tests that need to use a new common method name.

# Tests

only tests

# Todos

not necessary


lmuntaner (approved):
Thanks!

PR #4296, opened 2024-01-24
Use page objects in FollowSnsNeuronsModal.spec.ts

dskloetd:
# Motivation

# Changes

1. Add `FollowSnsNeuronsModalPo`.
2. Accept `number` instead of `Topic` so it can also  be used for NS function IDs.
3. Use page object in `FollowSnsNeuronsModal.spec.ts`.

# Tests

only tests

# Todos

not necessary


lmuntaner (approved):
LGTM!

PR #4300, opened 2024-01-25
Update proposal rendering

sa-github-api:
# Motivation

# Changes

# Tests


lmuntaner:
No changes. Closing.

@bitdivine should we move this to a weekly basis or confirm that there are changes before making the PR?

dskloetd:
There's more than a week between 2024-01-09 and 2024-01-18.
And there is some (if little) benefit in knowing that things didn't break (because they didn't change) with the newest release.

lmuntaner:
> There's more than a week between 2024-01-09 and 2024-01-18. And there is some (if little) benefit in knowing that things didn't break (because they didn't change) with the newest release.

Should we still merge this then?

dskloetd:
It's not important but it wouldn't hurt either.

dskloetd:
Probably easiest not to worry about it and just do it.

lmuntaner (approved):

dskloetd:
Although for all those update jobs that might require manual work, I agree we don't want a bot making daily new PRs while we figure out the manual work to do. Weekly seem more than good enough.

PR #4303, opened 2024-01-25
Neuron tag tests

mstrasinskis:
# Motivation

Add tests for `NeuronTag` tooltip.

# Changes

- add `NeuronTagPo` to return separately tag text and tooltip text
- `getNeuronTags` to return `NeuronTagPos` instead of raw text
- update current tests to use  `NeuronTagPo`
- check for correct tooltip text
- additionally added "check for no tags" and "check for multiple tags" rendering in `NnsNeuronCard` 

# Tests

Yes

# Todos

Not necessary

### DIFF
--- a/frontend/src/tests/lib/components/neurons/NnsNeuronCard.spec.ts
+++ b/frontend/src/tests/lib/components/neurons/NnsNeuronCard.spec.ts
@@ -157,7 +157,11 @@ describe("NnsNeuronCard", () => {
 
     const po = NnsNeuronCardPo.under(new JestPageObjectElement(container));
 
-    expect(await po.getNeuronTags()).toEqual(["Hardware Wallet Controlled"]);
+    expect(await po.getNeuronTags()).toHaveLength(1);
+
+    const tag = (await po.getNeuronTags())[0];
+    expect(await tag.getTagText()).toBe("Hardware Wallet Controlled");
+    expect(await tag.getTooltip().isPresent()).toBe(false);
   });
 
   it("renders neuron type tag when not default", async () => {
@@ -180,7 +184,12 @@ describe("NnsNeuronCard", () => {
 
     const po = NnsNeuronCardPo.under(new JestPageObjectElement(container));
 
-    expect(await po.getNeuronTags()).toEqual(["Seed Seed Neuron"]);
+    expect(await po.getNeuronTags()).toHaveLength(1);
+
+    const tag = (await po.getNeuronTags())[0];
+    expect(await tag.getTagText()).toBe("Seed");
+    expect(await tag.getTooltip().isPresent()).toBe(true);
+    expect(await tag.getTooltipText()).toBe("Seed Neuron");
   });
 
   it("renders proper text when status is LOCKED", async () => {
--- a/frontend/src/tests/lib/components/neurons/NnsNeuronCard.spec.ts
+++ b/frontend/src/tests/lib/components/neurons/NnsNeuronCard.spec.ts
@@ -136,6 +136,24 @@ describe("NnsNeuronCard", () => {
     expect(getByText(en.neurons.hotkey_control)).toBeInTheDocument();
   });
 
+  it("renders no tags by default", async () => {
+    icpAccountsStore.setForTesting({
+      main: mockMainAccount,
+      subAccounts: [],
+    });
+    const { container } = render(NnsNeuronCard, {
+      props: {
+        neuron: {
+          ...mockNeuron,
+        },
+      },
+    });
+
+    const po = NnsNeuronCardPo.under(new JestPageObjectElement(container));
+
+    expect(await po.getNeuronTags()).toHaveLength(0);
+  });
+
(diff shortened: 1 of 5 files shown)


frontend/src/tests/page-objects/NnsNeuronCardTitle.page-object.ts
```diff
@@ -12,8 +13,7 @@ export class NnsNeuronCardTitlePo extends BasePageObject {
     return this.root.querySelector("[data-tid=neuron-id]").getText();
   }
 
-  async getNeuronTags(): Promise<string[]> {
-    const elements = await this.root.allByTestId("neuron-tag");
-    return Promise.all(elements.map((tag) => tag.getText()));
+  getNeuronTags(): Promise<NeuronTagPo[]> {
```
dskloetd, line 16: For consistency, this should be named `getNeuronTagPos`.

frontend/src/tests/page-objects/NnsNeuronCard.page-object.ts
```diff
@@ -46,7 +47,7 @@ export class NnsNeuronCardPo extends BasePageObject {
     return Number(await this.getText("token-value"));
   }
 
-  getNeuronTags(): Promise<string[]> {
+  getNeuronTags(): Promise<NeuronTagPo[]> {
```
dskloetd, line 50: For consistency, this should be named `getNeuronTagPos`.

frontend/src/tests/page-objects/NeuronTag.page-object.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  static async allUnder(element: PageObjectElement): Promise<NeuronTagPo[]> {
+    return Promise.all(
+      Array.from(await element.allByTestId(NeuronTagPo.TID)).map(async (el) => {
+        return new NeuronTagPo(el);
+      })
+    );
+  }
+
+  getTag(): TagPo {
```
dskloetd, line 17: This should be called `getTagPo`.

frontend/src/tests/page-objects/NeuronTag.page-object.ts
```diff
@@ -0,0 +1,32 @@
...
+    return TagPo.under(this.root);
+  }
+
+  getTagText(): Promise<string> {
+    return this.getTag().getText();
+  }
+
+  getTooltip(): TooltipPo {
```
dskloetd, line 25: This should be called `getTooltipPo`.

frontend/src/tests/page-objects/NeuronTag.page-object.ts
```diff
@@ -0,0 +1,32 @@
...
+    return TooltipPo.under(this.root);
+  }
+
+  getTooltipText(): Promise<string> {
+    return TooltipPo.under(this.root).getText();
```
dskloetd, line 30: reuse `this.getTooltipPo()`

frontend/src/tests/lib/components/neurons/NnsNeuronCard.spec.ts
```diff
@@ -180,7 +205,45 @@ describe("NnsNeuronCard", () => {
...
 
-    expect(await po.getNeuronTags()).toEqual(["Seed Seed Neuron"]);
+    expect(await po.getNeuronTags()).toHaveLength(1);
+
+    expect(await (await po.getNeuronTags())[0].getTagText()).toBe("Seed");
+    expect(await (await po.getNeuronTags())[0].getTooltip().isPresent()).toBe(
+      true
+    );
+    expect(await (await po.getNeuronTags())[0].getTooltipText()).toBe(
+      "Seed Neuron"
```
dskloetd, line 215: This makes me wonder why we need a separate tooltip instead of just `Seed Neuron` being the tag text. It would still be much shorter than `Hardware Wallet Controller`.

frontend/src/tests/page-objects/NnsNeuronCard.page-object.ts
dskloetd, line 50: Can we keep the existing `getNeuronTags` along side `getNeuronTagPos` for convenience?

frontend/src/tests/lib/components/neurons/NnsNeuronCard.spec.ts
mstrasinskis: Let me discuss this w/ designer.

frontend/src/tests/page-objects/NnsNeuronCard.page-object.ts
mstrasinskis: Good point.`getNeuronTags` was returned. If we remove tooltips, `getNeuronTagPos` can also be removed.

frontend/src/tests/page-objects/NnsNeuronCardTitle.page-object.ts
mstrasinskis: Good point, thanks.

mstrasinskis:
The change is no longer necessary because the tooltips have been removed from the neuron tag.

PR #4306, opened 2024-01-26
Update aggregator

sa-github-api:
# Motivation
We would like to parse all the latest SNS data.

# Changes
* Updated the Rust code derived from `.did` files in the aggregator.

# Tests

### DIFF
--- a/rs/sns_aggregator/src/types/ic_sns_ledger.rs.orig
+++ b/rs/sns_aggregator/src/types/ic_sns_ledger.rs.orig
@@ -120,15 +120,17 @@ pub struct BlockRange {
     pub blocks: Vec<Block>,
 }
 
-pub type QueryBlockArchiveFn = candid::Func;
-#[derive(CandidType, Deserialize)]
+candid::define_function!(pub QueryBlockArchiveFn : (GetBlocksArgs) -> (
+    BlockRange,
+  ) query);
+#[derive(Serialize, Clone, Debug, CandidType, Deserialize)]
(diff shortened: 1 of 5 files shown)


rs/sns_aggregator/src/types/ic_sns_ledger.rs.orig
lmuntaner, line 1: @bitdivine what's this file doing here?

Is this correct?

dskloetd: Looks like something git left behind. Almost certainly not correct.

bitdivine: Agreed.  Question is why was it left behind?

bitdivine: I checked.  Context first:
- patch creates .orig files if there is a problem applying a patch.
- The patch for the ledger  was applied successfully but at an offset of 7 lines.  Presumably that was enough for patch to keep a backup just in case that was wrong.

I  would recommend deleting this file from this PR and merging.  If we decide that we never want these `.orig` files we can add `*.orig` to `.gitignore`.  What are your thoughts?

lmuntaner: I agree. If they are created by the script and used temporarily let's ignore them.

lmuntaner:
Closing because there is a more recent PR #4317

PR #4311, opened 2024-01-26
fix: participation button width inside the tooltip on mobile

mstrasinskis:
# Motivation

The tooltip container resets the `toolbar` stretching style. 

# Changes

- To avoid code repetition, the disabled buttons have been merged into a single block (the only difference was the tooltip text).
- Added `full-width-mobile` and applied also inside the tooltip.
- Removed redundant `participate` class.

# Tests

Not necessary.

# Todos

| Before | After |
|--------|--------|
| image | image |


mstrasinskis:
Replaced w/ the mixin.

PR #4324, opened 2024-01-29
Add TooltipIcon component

dskloetd:
# Motivation

We have one place where we have an info icon with a tooltip and we want to add more.
For convenience and consistency I'm factoring out the existing one into a separate component.
I'm also giving it the standard icon color and padding.
This makes the existing usage consistent with other info icons.

# Changes

1. Create `TooltipIcon` component which has an `IconInfo` with a `Tooltip`.
2. Give it a left margin of `--padding` so the parent component doesn't have to style the gap.
3. Give it a color and make it center aligned.
4. Use the `TooltipIcon` in `SnsNeuronHotkeysCard.svelte` to replace the current ad-hoc implementation.

# Tests

Unit tests added.

The affected icon is the one next to "Hotkeys".
"Following" is included for comparison.

Before:
image

After:
image

# Todos

### DIFF
--- a/frontend/src/lib/components/ui/TooltipIcon.svelte
+++ b/frontend/src/lib/components/ui/TooltipIcon.svelte
@@ -16,6 +16,5 @@
     display: inline-flex;
     align-items: center;
     color: var(--tertiary);
-    margin-left: var(--padding);
   }
 </style>
--- a/frontend/src/lib/components/ui/TooltipIcon.svelte
+++ b/frontend/src/lib/components/ui/TooltipIcon.svelte
@@ -1,0 +1,21 @@
+<script lang="ts">
+  import { IconInfo, Tooltip } from "@dfinity/gix-components";
+
+  export let tooltipId: string;
+  export let text: string;
+</script>
+
+<div class="wrapper" data-tid="tooltip-icon-component">
+  <Tooltip id={tooltipId} {text}>
+    <IconInfo />
+  </Tooltip>
+</div>
+
+<style lang="scss">
+  .wrapper {
+    display: inline-flex;
+    align-items: center;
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/ui/TooltipIcon.svelte
```diff
@@ -0,0 +1,21 @@
...
+    <IconInfo />
+  </Tooltip>
+</div>
+
+<style lang="scss">
+  .wrapper {
+    display: inline-flex;
+    align-items: center;
+    color: var(--tertiary);
+    margin-left: var(--padding);
```
lmuntaner, line 19: This means that the TooltipInco has to go on the right side.

I think that it makes more sense for the parent to manage this space. It might mean more code, but the component is more independent of the context.

dskloetd: Done.
Although the "on the right side" problem can also easily be fixed by having a margin on both sides.
And it's not just more code without, it's also more likely to be inconsistent, as demonstrated by how it was before this PR.

lmuntaner: Inconsistent with what? The other place where it's used?

That might mean we need a shared component between those use cases, no?

lmuntaner (approved):
Thanks!

frontend/src/lib/components/ui/TooltipIcon.svelte
dskloetd: The shared use case of "having something with a tooltip icon next to it"?

lmuntaner: Yes, maybe using a slot to position it though.

PR #4326, opened 2024-01-29
Add tooltip icon to dissolve delay item action

dskloetd:
# Motivation

Requested in point 22 of https://www.notion.so/dfinityorg/Wording-Changes-4411020138ce4d09b6605ad1fdbb9555

# Changes

1. Add fields to `CommonItemAction` to add an optional tooltip next to the title.
2. Use the fields in `NnsNeuronDissolveDelayItemAction.svelte` and `SnsNeuronDissolveDelayItemAction.svelte` to explain the dissolve delay.
3. Pass the token to `SnsNeuronDissolveDelayItemAction` to use in the tooltip message.

# Tests

Added a unit test that the tooltip text includes the token symbol.

image

# Todos

Added an entry which will also cover the next PR.

### DIFF
--- a/frontend/src/lib/components/ui/CommonItemAction.svelte
+++ b/frontend/src/lib/components/ui/CommonItemAction.svelte
@@ -1,7 +1,11 @@
 <script lang="ts">
+  import TooltipIcon from "./TooltipIcon.svelte";
   import { ItemAction } from "@dfinity/gix-components";
+  import { nonNullish } from "@dfinity/utils";
 
   export let testId: string;
+  export let tooltipText: string | undefined = undefined;
+  export let tooltipId: string | undefined = undefined;
 </script>
 
 <ItemAction {testId}>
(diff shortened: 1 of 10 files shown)


frontend/src/lib/components/ui/CommonItemAction.svelte
```diff
@@ -1,15 +1,24 @@
...
 
 <ItemAction {testId}>
   <div class="icon" slot="icon">
     <slot name="icon" />
   </div>
   <div class="content">
-    <h4 data-tid="staked-maturity"><slot name="title" /></h4>
+    <h4>
+      <slot name="title" />
+      {#if nonNullish(tooltipText) && nonNullish(tooltipId)}
```
lmuntaner, line 18: Do we need a test for this?

Until now this component had not logic, but now it does. I wonder whether it's needed or the test in the usage is enough.

lmuntaner (approved):
Just a question, already approving.

frontend/src/lib/components/ui/CommonItemAction.svelte
dskloetd: I wasn't sure since the CommonItemAction isn't really used by itself and adding a test component and page object felt like overkill.
But having a test is better than not having a test so I added one which uses the PO of `SnsNeuronDissolveDelayItemAction` with a comment explaining why.

PR #4331, opened 2024-01-29
Fix dissolve delay item action tooltip

dskloetd:
# Motivation

In https://github.com/dfinity/nns-dapp/pull/4326, I accidentally mixed up points 22 and 23 of https://www.notion.so/dfinityorg/Wording-Changes-4411020138ce4d09b6605ad1fdbb9555 and put the tooltip text for the neuron state on the item action for the dissolve delay.
In this PR we put the correct tooltip in the dissolve delay item action.
The intended behavior also requires additional logic because the intended tooltip text includes the dissolve delay duration and the tooltip should be hidden if the neuron is unlocked.

# Changes

1. Include the duration in the dissolve delay item action tooltip.
2. Hide the tooltip if there is no dissolve delay.

# Tests

Unit tests added.

image

# Todos

covered

### DIFF
--- a/frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
+++ b/frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
@@ -136,4 +136,28 @@ describe("NnsNeuronDissolveDelayItemAction", () => {
 
     expect(await po.hasIncreaseDissolveDelayButton()).toBe(true);
   });
+
+  it("should not render a tooltip without dissolve delay", async () => {
+    const neuron: NeuronInfo = {
+      ...controlledNeuron,
+      state: NeuronState.Dissolved,
+      dissolveDelaySeconds: 0n,
+    };
+    const po = renderComponent(neuron);
+
+    expect(await po.getTooltipIconPo().isPresent()).toBe(false);
+  });
+
+  it("should render the dissolve duration in the tooltip text", async () => {
+    const neuron: NeuronInfo = {
+      ...controlledNeuron,
+      state: NeuronState.Locked,
+      dissolveDelaySeconds: BigInt(SECONDS_IN_YEAR * 2),
+    };
+    const po = renderComponent(neuron);
+
+    expect(await po.getTooltipIconPo().getText()).toContain(
+      "available in 2 years, 12 hours"
+    );
+  });
 });
--- a/frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
+++ b/frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
@@ -148,7 +148,7 @@ describe("NnsNeuronDissolveDelayItemAction", () => {
     expect(await po.getTooltipIconPo().isPresent()).toBe(false);
   });
 
-  it("should render the dissolve duration in the tooltip text", async () => {
+  it("should render tooltip when locked", async () => {
     const neuron: NeuronInfo = {
       ...controlledNeuron,
       state: NeuronState.Locked,
@@ -156,12 +156,12 @@ describe("NnsNeuronDissolveDelayItemAction", () => {
     };
     const po = renderComponent(neuron);
 
-    expect(await po.getTooltipIconPo().getText()).toContain(
-      "available in 2 years, 12 hours"
+    expect(await po.getTooltipIconPo().getText()).toBe(
+      "Dissolve delay is the minimum amount of time you have to wait for the neuron to unlock, and ICP to be available again. If your neuron is dissolving, your ICP will be available in 2 years, 12 hours."
     );
   });
 
-  it("should render the dissolve duration in the tooltip text when dissolving", async () => {
+  it("should render tooltip when dissolving", async () => {
     const neuron: NeuronInfo = {
       ...controlledNeuron,
       state: NeuronState.Dissolving,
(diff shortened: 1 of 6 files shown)


frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
```diff
@@ -136,4 +136,28 @@ describe("NnsNeuronDissolveDelayItemAction", () => {
...
+      ...controlledNeuron,
+      state: NeuronState.Dissolved,
+      dissolveDelaySeconds: 0n,
+    };
+    const po = renderComponent(neuron);
+
+    expect(await po.getTooltipIconPo().isPresent()).toBe(false);
+  });
+
+  it("should render the dissolve duration in the tooltip text", async () => {
```
lmuntaner, line 151: If it's dissolving there is also a tooltip, is that correct?

Do we want to add a test for that case?

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronDissolveDelayItemAction.spec.ts
```diff
@@ -148,14 +148,41 @@ describe("SnsNeuronDissolveDelayItemAction", () => {
...
+  it("should render the dissolve duration in the tooltip text", async () => {
+    const neuron = createMockSnsNeuron({
+      id: [1],
+      state: NeuronState.Locked,
+      permissions: [noDissolvePermissions],
+      dissolveDelaySeconds: BigInt(SECONDS_IN_FOUR_YEARS),
+    });
+    const po = renderComponent(neuron);
+
+    expect(await po.getTooltipIconPo().getText()).toContain(
```
lmuntaner, line 184: Why didn't you add this expect in the previous one?

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronDissolveDelayItemAction.spec.ts
```diff
@@ -148,14 +148,41 @@ describe("SnsNeuronDissolveDelayItemAction", () => {
...
+      state: NeuronState.Locked,
+      permissions: [noDissolvePermissions],
+      dissolveDelaySeconds: BigInt(SECONDS_IN_FOUR_YEARS),
     });
     const po = renderComponent(neuron);
 
     expect(await po.getTooltipIconPo().getText()).toContain(token.symbol);
   });
+
+  it("should render the dissolve duration in the tooltip text", async () => {
```
lmuntaner, line 175: Ditto, do we need one with the Dissolving state?

lmuntaner (approved):
Looks good, a few comments on the tests but already approving.

frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
dskloetd: The neuron state is irrelevant. There is a tooltip if the dissolve delay is  greater than zero.
But, sure I can add another test in case the logic ever changes.

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronDissolveDelayItemAction.spec.ts
dskloetd: Which previous one? The one called `"should render the token symbol in the tooltip text"`?

frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
lmuntaner: So there is no tooltip when `whenDissolvedTimestampSeconds` is populated. Only when it's locked and `dissolveDelaySeconds` is present within `dissolve_state`?

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronDissolveDelayItemAction.spec.ts
lmuntaner: Yes, correct. They both test the tooltip with the same setup but different parts. Would it make sense to put the expects together?

frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
dskloetd: I'm not sure. There is a tooltip if `remainingTimeSeconds > 0n`.
In turn
```
remainingTimeSeconds =
    dissolvingTime ?? spawningTime ?? neuron.dissolveDelaySeconds;
```
and those values are in turn determined by helper functions in other files.
Does this sound reasonable or do I need to dig down into the logic of those functions?

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronDissolveDelayItemAction.spec.ts
dskloetd: Done

frontend/src/tests/lib/components/sns-neuron-detail/SnsNeuronDissolveDelayItemAction.spec.ts
dskloetd: OK, I've merged the tests.

dskloetd (commented):
Please take another look

frontend/src/tests/lib/components/neuron-detail/NnsNeuronDissolveDelayItemAction.spec.ts
lmuntaner: That sounds reasonable, but then better add a test that with dissolving state and `whenDissolvedTimestampSeconds` there is also the expected text.

The neuron state is not irrelevant because the neuron state is calculated based on the `dissolve_state`.

dskloetd: Discussed on Slack. This is covered by the added tests.

lmuntaner (approved):
Thanks!

